PR #67, opened 2020-04-03
RUMM-346 Unify dependency manager tests

ncreated:
### What and why?

📦 This PR unifies Carthage, Cocoa Pods and SPM test projects.

### How?

Now, all projects behave the same:
* `cd <test-project> && make` sources project to use current git branch;
* `bitrise.yml` runs unit tests and UI tests for each project and exports artifacts to Bitrise;
* `testCallingLogicThatLoadsSDK()` unit test calls basic SDK logic;
* `testDisplayingUI()` UI test launches the app and asserts that "Testing ..." label is visible;
* dedicated Build Phase runs linter;
* deployment target is set to `iOS 13.0`;
* Code Signing is configured for app target, unit test target and UI test target.

**App Target**
Every app target is just "Single View" app with default code + invocation of `Datadog` SDK.

**UI Tests**
Every test project has this basic UI and asserts the label is visible:
Screenshot 2020-04-03 at 17 07 08

**Unit Tests**
Unit tests just invoke `vc.viewDidLoad()`, which guarantees that SDK is linked, loaded and run with no runtime error.

**Code Signing**
3 `.xcconfigs` are shared between all projects:

Screenshot 2020-04-03 at 17 12 17

where `datadog.xcodeproj` is git-ignored as it might contain our secrets.

### Review checklist


buranmert:
ℹ️ let's merge https://github.com/DataDog/dd-sdk-ios/pull/68 before this one

dependency-manager-tests/carthage/.gitignore
```diff
@@ -1,4 +1,6 @@
 .DS_Store
+/.build
+xcuserdata/
 /Carthage
 
 # Ignore `Cartfile.resolved` and `Cartfile` as they will be re-created for every test run.
```
buranmert, line 6: IMO one single `.gitignore` in the parent directory (`/dep-manager-tests`) could simpler and easier to maintain

dependency-manager-tests/cocoapods/CPProject/ViewController.swift
```diff
@@ -7,7 +7,9 @@
 import UIKit
 import Datadog
 
-class ViewController: UIViewController {
+internal class ViewController: UIViewController {
+    private var logger: Logger! // swiftlint:disable:this implicitly_unwrapped_optional
```
buranmert, line 11: why do we need an instance prop?

dependency-manager-tests/spm/SPMProject/AppDelegate.swift
```diff
@@ -5,74 +5,19 @@
 */
 
 import UIKit
-import CoreData
```
buranmert, line 8: i can imagine removing this line made you happy 😄

dependency-manager-tests/xcconfigs/unit-tests-target.xcconfig
```diff
@@ -0,0 +1,2 @@
+#include "app-target.xcconfig"
+PROVISIONING_PROFILE_SPECIFIER[sdk=iphoneos*]=// overwrite to empty, so Xcode picks "Automatic"
```
buranmert, line 2: what does `overwrite to empty` mean? 
• should we leave it empty? or
• Xcode picks `automatic` if we leave it empty so that we should provide a value here?

dependency-manager-tests/spm/SPMProject/AppDelegate.swift
ncreated: Totally 😄. I also noticed that the "+" button in navigation bar of the skeleton app was not working - it was crashing the app. Turns out that creating "basic Core Data" example requires some custom logic to be implemented.

Anyway, I changed all examples to "Single View App".

dependency-manager-tests/xcconfigs/unit-tests-target.xcconfig
ncreated: Here we need to overwrite it with empty string, so it doesn't use the setting from `app-target.xcconfig`. Then Xcode uses it's default value (auto). I tried overwriting it with `"auto"`, `"Auto"` and `"Automatic"`, but none of them worked.

Alternatively, we could split it into two xcconfigs:
```
// app-target.xcconfig
DEVELOPMENT_TEAM[sdk=iphoneos*]=xxx
CODE_SIGN_IDENTITY[sdk=iphoneos*]=yyy
PROVISIONING_PROFILE_SPECIFIER[sdk=iphoneos*]=zzz
```
```
// unit-tests-target.xcconfig
DEVELOPMENT_TEAM[sdk=iphoneos*]=xxx
CODE_SIGN_IDENTITY[sdk=iphoneos*]=yyy
```
But this would requires more hassle and two separate `datadog-app-target.xcconfig` and `datadog-unit-tests-target.xcconfig`. All will duplicate parts of the setup. Current approach optimizes number of duplication.

dependency-manager-tests/cocoapods/CPProject/ViewController.swift
ncreated: No reason 🤷‍♂, this will work anyway even if `Logger` is allocated only for the current stack.

dependency-manager-tests/carthage/.gitignore
ncreated: Hmm, maybe let's wait for RUMM-334 where we want to merge `dep-manager-tests` with `instrumented-tests` and we'll see if single `.gitignore` will be practical? IMO it might not, as for Integration and Benchmark we might want to ignore different parts of the setup - at least, each folder has different `.gitignore` now. WDYT?

buranmert: I prefer less `.gitignore` files since they are hidden and not discoverable so i prefer _discover-once-understand-all_

we can have project specific entries as well:
```
...
/carthage/.build
/cocoapods/Pods
...
```

dependency-manager-tests/xcconfigs/unit-tests-target.xcconfig
buranmert: okay i see 👍 
but i think the current comment message is not clear enough.

_this line overrides `app-target.xcconfig` and sets prov profile to `Auto` in its target_
what do you think?

ncreated: Will do 👍

dependency-manager-tests/carthage/.gitignore
ncreated: KK, will do 👍

dependency-manager-tests/xcconfigs/unit-tests-target.xcconfig
ncreated: Done ✅

dependency-manager-tests/carthage/.gitignore
ncreated: Done ✅. I also updated SPM test to use `project.pbxproj.src` and gitignored `SPMProject/project.pbxproj`. This way git status is empty after running the test. Now it's also consistent with `Podfile.src` and `Cartfile.src`.

PR #68, opened 2020-04-03
RUMM-346 DatadogPrivate module is hidden from public

buranmert:
### What and why?

DDPrivate module is created to use ObjC from Swift

### How?

It is not intended for public use
We used private.modulemap to hide it from public

### Review checklist


buranmert:
Xcode doesn't autocomplete for me but yes, `_Datadog_Private` seems to be still public somehow.

I'm going ahead as this is good enough for us at this stage ✅

PR #72, opened 2020-04-08
RUMM-311 Remove flakiness in unit tests

ncreated:
### What and why?

🚚 This PR reduces flakiness in unit tests on CI.

BEFORE: flakiness **~10%** (measured on 30 unit test runs).
AFTER: flakiness **~0.6% or even less*** (measured on 150 unit test runs).

_* we can expect even less than 0.6%, as I doubled the timeout responsible for this single failure in 150 test runs probe without later retrying the probe._ 

### How?

There were two main sources of flakiness:
* 1/ missing synchronization code for asynchronous work completion;
* 2/ too optimistic timeouts and too short intervals for `Thread.sleep()`.

Both were resulting in random failures on CI, which performs tests significantly slower than our laptops (which is why almost no flakiness was observed locally).

**1/ Missing synchronization code.**

In few places I added explicit synchronization with appropriate note in the comment, i.e.:
```swift
// DataUploadWorkerTests.swift

// ...
uploaderQueue.sync {} // wait until last "process upload" operation completes (to make sure "delete file" was requested)
fileReadWriteQueue.sync {} // wait until last scheduled "delete file" operation completed
// ...
```

Majority of synchronization issues was coming from our custom `URLSessionMock`, `URLSessionDataTaskMock` and `URLSessionDataTaskReferences` as we were trying to re-do internal synchronization done in `URLSession` in our own mocks layer.

To mitigate this, I replaced custom mocks with in-build `URLProtocol` subclass that intercepts network traffic. This behaviour is abstracted by `ServerMock` component:
```swift
func testXYZ() {
   let server = ServerMock(delivery: .success(response: .mockResponseWith(statusCode: 200)))

   // The `server.urlSession` is now intercepted by `ServerMock`...
   let httpClient = HTTPClient(session: server.urlSession)

   // ...and recorded requests can be retrieved by a blocking method call:
   let recordedRequests = server.waitAndReturnRequests(count: 3, timeout: timeout)
   // ✨ At this point (after calling `server.wait*()`) all pending completion closures 
   // are deallocated, so no asynchronous callback is leaked outside the unit test scope.
}
```

**2/ Too optimistic timeouts.**
CI workers perform significantly slower than our laptops, so I adjusted some arbitrary timeouts:
```swift
// Set the timeout to 40 times more than expected.
// In `RUMM-311` we observed 0.66% of flakiness for 150 test runs on CI with arbitrary value of `20`.
let timeout = DatadogInstanceMock.dataUploadInterval * Double(numberOfLogsSent) * 40
```

## Further note on flakiness

If we still observe annoying flakiness, we might consider setting [should_retry_test_on_fail: "yes"](https://github.com/bitrise-steplib/steps-xcode-test/blob/2f36e5fc43f2d1f1133e69c8fa095a3c7dfaa544/step.yml#L259) for `xcode-test` step in `bitrise.yml`. This will re-do the step one more time (no more) if it failed.

To make sure our tests code is thread-safe, I also enabled Thread Sanitizer for unit tests.

### Review checklist


Tests/DatadogTests/Datadog/Mocks/ServerMock.swift
```diff
@@ -0,0 +1,167 @@
...
+        precondition(ServerMock.activeInstance != nil, "Request was stopped while no `ServerMock` instance is active.")
+    }
+}
+
+/// All unit tests use this shared `URLSession`.
+private let sharedURLSession: URLSession = {
+    let configuration = URLSessionConfiguration.ephemeral
+    configuration.protocolClasses = [ServerMockProtocol.self]
+    return URLSession(configuration: configuration)
+}()
```
buranmert, line 65: very nice 👍 

maybe we could rename it to smth like `sharedTestsURLSession` (or anything with `test` in it) to help with its discoverability while writing new test cases?

ncreated: I can rename it for clarity in this file 👍. Although this won't impact the discoverability, as this is defined `private`:
```swift
private let sharedURLSession: URLSession = ...
```
and this instance is later exposed internally by `ServerMock`:
```swift
class ServerMock {
   let urlSession: URLSession = sharedURLSession
}
```

buranmert: sorry i missed that it was `private`, yes you are right 👍

Issue #71, opened 2020-04-08
Carthage Compilation Error: 'module.modulemap' should be inside the 'include' directory


joeydong:
**Describe what happened**

Hi, I'm trying to install the latest release dd-sdk-ios 1.0.1 using Carthage 0.34.0 and Xcode 11.3.1 but this dependency fails to compile with this error:

```
$ /usr/bin/xcrun xcodebuild -workspace /Carthage/Checkouts/dd-sdk-ios/instrumented-tests/Integration/Integration.xcworkspace CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES -list
Build settings from command line:
    CARTHAGE = YES
    CODE_SIGN_IDENTITY =
    CODE_SIGNING_REQUIRED = NO

Resolve Package Graph

Resolved source packages:
  dd-sdk-ios: /Carthage/Checkouts/dd-sdk-ios
  HTTPServerMock: /Carthage/Checkouts/dd-sdk-ios/instrumented-tests/http-server-mock

xcodebuild: error: Could not resolve package dependencies:
  package has unsupported layout; modulemap '/Carthage/Checkouts/dd-sdk-ios/Datadog/DatadogPrivate/module.modulemap' should be inside the 'include' directory
```

I got this xcrun command from the failed Carthage output:

```
A shell task (/usr/bin/xcrun xcodebuild -workspace /Carthage/Checkouts/dd-sdk-ios/instrumented-tests/Integration/Integration.xcworkspace CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES -list) failed with exit code 74:
xcodebuild: error: Could not resolve package dependencies:
  package has unsupported layout; modulemap '/Carthage/Checkouts/dd-sdk-ios/Datadog/DatadogPrivate/module.modulemap' should be inside the 'include' directory
```

Any ideas?

**Steps to reproduce the issue:**

Looks like I can reproduce this in Xcode 11.3.1 by:

1. Clone this repo
1. Open `/dd-sdk-ios/examples/iOS-app-example-spm/iOS-app-example-spm.xcodeproj`
1. Cmd + B

<img width="1895" alt="Screen Shot 2020-04-07 at 5 31 51 PM" src="https://user-images.githubusercontent.com/434294/78731859-c6737f00-78f5-11ea-92c5-5e8c7074f511.png">

[...]

PR #73, opened 2020-04-08
RUMM-369 Fix Carthage for Xcode 11.3

ncreated:
### What and why?

🐞 Fixing #71, where it was reported that `carthage update` doesn't work for Xcode 11.3. 

### How?

`Xcode 11.3` expects `*.modulemap` to be put inside `include/` directory for `DatadogPrivate`.
This is different in `Xcode 11.4` where `*.modulemap` can be put anywhere.

In result `carthage update` was failing with:
```
/usr/bin/xcrun xcodebuild -workspace /Carthage/Checkouts/dd-sdk-ios/instrumented-tests/Integration/Integration.xcworkspace CODE_SIGNING_REQUIRED=NO 

...

xcodebuild: error: Could not resolve package dependencies:
  package has unsupported layout; modulemap '/Carthage/Checkouts/dd-sdk-ios/Datadog/DatadogPrivate/module.modulemap' should be inside the 'include' directory
```

💡 Why `carthage update` builds `Integration.xcworkspace`? Well, Carthage scans recursively its `Checkouts/` directory and builds every `.xcodeproj` and `.xcworkspace` it finds and [there is no option to make it ignore particular schemes](https://github.com/Carthage/Carthage/pull/1990).

🎁 I also updated `spm-example` project to use `1.0.2` pre-release tag, so the same problem doesn't happen there when building `Datadog` dependency.

🧪 I validated all dependency managers by running unit + UI tests:
* `Xcode 11.3.1 (11C505)` - OK for Carthage, Cocoapods and SPM both on 📱 and Simulator;
* `Xcode 11.4 (11E146)` - OK for Carthage, Cocoapods and SPM both on 📱 and Simulator.

💡 This PR also sets the `exactVersion` of `Datadog` SPM package used by `spm-example` project to a tag on this branch. This is to make this PR correct for Xcode 11.3. We're OK with merging this to `master` as `spm-example` will be re-done soon.

### Review checklist


ncreated:
- [ ] I forgot to bump the `sdkVersion`, doing it right now.

PR #76, opened 2020-04-10
RUMM-296 Implement "Set a `Span` tag" part of OT interface

ncreated:
### What and why?

📦 This PR allows adding tags to a `Span`, so _"Set a `Span` tag"_ part of [OT spec](https://github.com/opentracing/specification/blob/master/specification.md#set-a-span-tag).

### How?

According to OT spec, tags can be associated with `Span` in two ways:
* when starting the span with `Tracer`:
```swift
let span = tracer.startSpan(
   operationName: "operation", 
   tags: ["tag1": "value1"]
)
```
* when span already exists:
```swift
span.setTag(key: "key1", value: "value1")
```

💡 In this PR I don't implement any thread-safety for `span.tags` dictionary, as this will come later in `RUMM-340`. The reason for this is that I'm not sure about the impact of span serialization on the threading model for `DDTracer` and `DDSpan`. A lot of things might change when implementing span writer, so instead of doing thread-safety now, I just leave `// TODO: RUMM-340` comments in appropriate places.

### Review checklist

### DIFF
--- a/Datadog/Datadog.xcodeproj/project.pbxproj
+++ b/Datadog/Datadog.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		6109A21B244087A600BFA8EE /* Casting.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6109A21A244087A600BFA8EE /* Casting.swift */; };
 		61133B8C242393DE00786299 /* Datadog.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 61133B82242393DE00786299 /* Datadog.framework */; };
 		61133B93242393DE00786299 /* Datadog.h in Headers */ = {isa = PBXBuildFile; fileRef = 61133B85242393DE00786299 /* Datadog.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		61133BCA2423979B00786299 /* EncodableValue.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61133BA02423979B00786299 /* EncodableValue.swift */; };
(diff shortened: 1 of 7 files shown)


Datadog/Datadog.xcodeproj/project.pbxproj
```diff
@@ -265,17 +260,6 @@
...
-		61C3637F2436164B00C4D4E6 /* ObjcExceptionHandlerTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ObjcExceptionHandlerTests.swift; sourceTree = "<group>"; };
-		61C3638224361BE200C4D4E6 /* DatadogPrivateMocks.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = DatadogPrivateMocks.swift; sourceTree = "<group>"; };
-		61C3638424361E9200C4D4E6 /* Globals.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = Globals.swift; sourceTree = "<group>"; };
-		61C3646F243B5C8300C4D4E6 /* ServerMock.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ServerMock.swift; sourceTree = "<group>"; };
-		9E085879242519FF001A3583 /* NetworkPathMonitor.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = NetworkPathMonitor.swift; sourceTree = "<group>"; };
-		9E36D92124373EA700BFBDB7 /* SwiftExtensionsTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SwiftExtensionsTests.swift; sourceTree = "<group>"; };
```
ncreated, line 278: @buranmert I don't understand well those removals done by Xcode, but they are added automatically when I add new files to the project. I think it's the echo of rebase I made with `tracing` over `master`. Let's revisit this after we update repository layout. Maybe we'll just re-add all the tracing files again and this will be gone.

buranmert: i noticed them [here](https://github.com/DataDog/dd-sdk-ios/pull/74/files#diff-58b21cfb4754fd977f9037494c84df07) as well but i just went by since i assumed that was harmless.

as we discussed today, let's merge `tracing` first and then we can move and re-add these files to the project and it should be alright ✅

Sources/Datadog/Traces/DDSpan.swift
```diff
@@ -9,14 +9,16 @@ import Foundation
 
 internal class DDSpan: Span {
     private(set) var operationName: String
-    /// The `Tracer` which created this span.
-    internal let issuingTracer: DDTracer
+    private(set) var tags: [String: Codable]
```
buranmert, line 12: why is `tags` a `Dictionary` instead of an `Array`?
(i guess this is part of OT spec and i assume this is fine. i'm just wondering)

Sources/Datadog/Traces/DDSpan.swift
```diff
@@ -9,14 +9,16 @@ import Foundation
...
     internal let startTime: Date
     internal var isFinished: Bool = false
+    /// The `Tracer` which created this span.
+    private let issuingTracer: DDTracer
 
-    init(tracer: DDTracer, operationName: String, parentSpanContext: DDSpanContext?, startTime: Date) {
+    init(tracer: DDTracer, operationName: String, parentSpanContext: DDSpanContext?, tags: [String: Codable], startTime: Date) {
         self.issuingTracer = tracer
```
buranmert, line 19: if `Tracer` must be created by another `Tracer`, shouldn't its API be similar to the following?
```swift
private init(...) { ... } // no public/internal constructor
public func createTrace(...) -> Trace { ... } //public instance method to create a new Trace
```
the design above seems to enforce the correctness of `issuingTracer`

Sources/Datadog/Traces/DDSpan.swift
ncreated: [Open Tracing defines](https://github.com/opentracing/specification/blob/master/specification.md#set-a-span-tag) tags as key-value pairs:
> Required parameters
> * The tag key, which must be a string
> * The tag value, which must be either a string, a boolean value, or a numeric type

and because of this dictionary is more efficient than array. To me those are rather "attributes" than "tags", but OT standard needs to be language/technology agnostic, that's why "tags" is more general lingo I thing.

And yes, `[String: Codable]` is [enforced by `opentracing-swift`](https://github.com/opentracing/opentracing-swift/blob/e2b0551de2fd9a5cd9a3d8dff5390440f724283c/Sources/OpenTracing/Span.swift#L21).

Sources/Datadog/Traces/DDSpan.swift
ncreated: I'm not sure if I understand, but the `DDSpan` cannot be created by the user. In fact, `DDSpan` is an `internal` type and users will only see `OpenTracing.Span`.

It is the `DDTracer` that creates `OpenTracing.Span` with `tracer.startSpan(...) -> OpenTracing.Span`.

Does it bring any answer 🤔?

buranmert: the problem with my proposal would be the following
```swift
// Span
fileprivate init(tracer, params...) { ... }
// Tracer extension
public issueSpan(params...) { ... }
```
we need to duplicate `params`. i saw this case as `Builder` pattern but it is not quite the same.

in short: let's go with the current code 👍💯

Issue #52, opened 2020-03-20
Send logs immediately in app extension?


lmramirez:
Hi, first off thanks to you all for creating this sdk. While it's not in our production app yet, it's working great for us in testing.

One thing I'm trying to figure out: We have a broadcast extension which has a short lifecycle. Is there a way to configure the datadog sdk to send logs immediately? Our extension can be killed at any time and I want to make sure all of our logs make it to datadog. I was poking through the code but it wasn't obvious if this was possible.

Thanks!

ncreated:
Hello @lmramirez 👋 thanks for trying out the SDK 🙂. As for now, there's no public API to configure the data upload frequency. We rather avoid exposing it so that users do not end up with a configuration that is ineffective or kills the app performance. Instead, with "immediate write" and "deferred upload" strategy implemented by the SDK, apps should not care about enforcing immediate uploads. We do our best to make sure the logs are delivered in a performant way.

The upload frequency is dynamic - it might vary from `1s` to `20s` depending on the logging pressure. Logs are written to the file system and they should be uploaded in the next session if app was terminated or suspended.

I think that your use case of short-lived app extension falls somehow beyond our upload heuristic. Could you give us more insight into this extension lifecycle?
* How long does it run in average?
* How often do your users come back to the extension?

We are going investigate the use case of app extensions for the SDK, so I put the enhancement label on this issue. Stay tuned we'll try to address it.

lmramirez:
@ncreated thank you for the detailed response.

The app allows users to make screen recordings of their iOS devices, so the time the extension runs ranges from about 15 seconds to maybe an hour in some cases. 
Some users come back frequently and make daily recordings and others do not return to the extension for weeks. Users come back to our host application more often than they come back to our extension, since the host application providers other functionality as well.

I haven't dug too deep into the datadog implementation,  but is there a way to configure the location logs are written to on disk? It would be nice to have the option to configure this to point at a location in our App Group directory, so that it could be read and uploaded by the host application maybe?

Again, thanks to you and the datadog team for working on this! We have some hard to debug issues in our app extension and this sdk is already helping us a lot in testing.

ncreated:
Thanks for details 👌. So as I understand you manage to upload some logs from the extension, but some number of short & unfrequent sessions is lost. One thing to comment here is that some percent of logs will be lost even when we apply the fix - due to our internal constraints we don't process logs that are significantly old. But even with this, the enhancement should give you a viable benefit.

[...]

PR #79, opened 2020-04-15
RUMM-319 Improve logging for app extensions

ncreated:
### What and why?

🚚 This PR addresses the enhancement discussed in #52 - sending logs immediately. 

The origin of this enhancement idea was the use case of running the SDK in short-lived app extension. Default timing configuration for persistence and upload was not effective when running in such environment so the SDK did not always have enough time to upload logs to Datadog during app extension lifetime.

However, instead of implementing "send logs immediately" feature we decided to **adjust timing configuration if SDK is running in app extension**.

### How?

An `Environment` abstraction is introduced:
```swift
/// An environment running the SDK.
internal enum Environment: Equatable {
    /// iOS app
    case app
    /// iOS app extension
    case appExtension
}
```

The `environment` is determined automatically during the `Datadog.initialize()` call and provides custom timing configuration for each case. The `.app` environment defaults to previous configuration and `.appExtension` brings distinct values.

The `.appExtension` environment:
* Issues the initial upload right away: `0.5s` after the SDK is initialized. This enables even very-short lived extensions to upload logs collected but not-yet-uploaded in the previous session.
* Defaults to `5s` for max upload delay. This means, every check will be performed in less than `5s` (vs `20s` in `.app` environment).
* Delay decrease factor is more aggressive: `0.5` (vs `0.9` in `.app`). This significantly reduces the time for next upload if the extension performs intensive logging.

### A note on integration tests

💡 This PR adds only unit tests and necessary changes to benchmarks. It doesn't ship the integration test for app extension. Doing this will require changing the nature of our integration tests - from being a unit test target to UI Test target which launches the app and app extension. Not done in this PR, as I don't want to conflict with other open PRs.

### Review checklist


buranmert:
The concerns of `Envrionment` and `Strategy` types seem to be kind of mixed up to me.
I'd like to propose another way for this change:
```swift
// no logic inside, simply an indicator of which environment we are in
enum Environment {
  case app
  case extension
}

struct DefaultLogsUploadStrategy: LogsUploadStrategy {
  let internal = default_interval // normal/default interval, eg: for app
  ...
}

struct IntensiveLogsUploadStrategy: LogsUploadStrategy {
  let internal = very_short_interval // a very short interval, eg: for app ext
  ...
}
// this could be LogsUploadStrategyBuilder or just an internal function instead of an extension
extension LogsUploadStrategy {
  static func build(for environment: Envrionment) -> LogsUploadStrategy {
    if app { return DefaultLogsUploadStrategy() }
    else { return IntensiveLogsUploadStrategy() }
  }
}
```

IMO that keeps concerns separate and strategies extensible (ie: currently, what happens if we'd like intensive loggin in apps?)

what do you think?

ncreated:
Closed in favour of https://github.com/DataDog/dd-sdk-ios/pull/84

PR #82, opened 2020-04-20
RUMM-392 SDK is set dynamic in SPM

buranmert:
### What and why?

[Swift Package Manager encourages us to leave library type for itself to decide either `dynamic` or `static`](https://developer.apple.com/documentation/swift_packages/product/2878196-library)
Yet, it doesn't change to use `dynamic` when needed.

### Example case

We have 3 targets:
1. `App`: links `Datadog`, `Framework1` and `Framework2`
2. `Framework1`: links `Datadog`
3. `Framework2`: links `Datadog`

When we `build App`, Xcode emits the following **error**:

> Swift package product 'Datadog' is linked as a static library by 'Framework1' and 'Framework2'. This will result in duplication of library code

### Solution

Either
1. we can change our SDK's type in `Package.swift` to `dynamic` or
2. our users can use `DISABLE_DIAMOND_PROBLEM_DIAGNOSTIC` build setting to silence the error

### Important note

Xcode builds **App Extensions** and alikes _without this problem_

That makes the example case above an edge case IMO

### Conclusion

Given that the error occurs only in an edge case, giving up on static linking may not be the best solution. 
I suggest **not fixing the issue** and
1. waiting for Xcode/SwiftPackageManager to fix the issue on their end
2. [providing information about possible workarounds to our users](https://forums.swift.org/t/adding-a-package-to-two-targets-in-one-projects-results-in-an-error/35007/2) (eg: in `README.md` or `TROUBLESHOOT.md`)

### Review checklist


dependency-manager-tests/spm/project.pbxproj.src
```diff
@@ -1,652 +0,0 @@
-// !$*UTF8*$!
```
ncreated, line 1: 👌 Just double checking - this file is not needed as we use `SPMProject.xcodeproj.src`, true?

buranmert: yes, correct 👍 

```
// Makefile
@rm -rf SPMProject.xcodeproj
@cp -r SPMProject.xcodeproj.src SPMProject.xcodeproj
@sed "s|REMOTE_GIT_BRANCH|${GIT_BRANCH}|g" SPMProject.xcodeproj.src/project.pbxproj > SPMProject.xcodeproj/project.pbxproj
```

PR #86, opened 2020-04-21
RUMM-281 Development flag fixed

buranmert:
### What and why?

`DD_SDK_DEVELOPMENT` flag gives SDK developers useful insights while debugging, but it wasn't working since `carthage` support was introduced

### How?

This PR adds `DD_SDK_DEVELOPMENT`
• as a `Swift active compilation flag`
• to `Datadog` target
• in `DEBUG` build configuration

That makes it enabled for those who build it from source (eg: SDK devs) while disabled for those who use any package manager (eg: app devs)

### Review checklist


buranmert:
@ncreated It doesn't bloat CI logs ✅

PR #88, opened 2020-04-22
RUMM-301 Shopist project created

buranmert:
### What and why? 🛒

We needed an example app to showcase the features and usage of SDK

#### Scope
This is only the groundwork of Shopist app
Other screens are yet to come 🔜 

#### SDK development
This also comes handy for SDK development: 
1. now we can open `Shopist.xcodeproj` and 
2. debug in an actual app target

#### Legacy examples
We had a few example projects in the repo
We removed them as they are not needed anymore ♻️ 

### Review checklist


Shopist/Makefile
```diff
@@ -0,0 +1,15 @@
+all: secret
+
+LOCAL_XCCONFIG="./Shopist/shopist-secrets.local.xcconfig" 
+ifeq ($(secret),)
+	secret="my-secret-token"
+endif
+
+secret:
```
ncreated, line 8: - [ ] Also, we need to remove this from the root `Makefile`:
```
examples:
		@echo "⚙️  Generating 'examples/examples-secret.xcconfig' file..."
		./tools/config/generate-examples-config-template.sh
		@echo "OK 👌"
```
Which also means those two can be removed:
- [ ] `tools/config/examples-secret-template.xcconfig`
- [ ] `generate-examples-config-template.sh`

bitrise.yml
```diff
@@ -12,12 +12,9 @@ workflows:
     - _make_dependencies
     - run_linter
     - run_unit_tests
+    - run_ui_tests
     - run_integration_tests
     - check_dependency_managers
-    # I disable it for now, as this would require supporting local pod installation
-    # of `DatadogSDK.podspec` in CP example project. This effort will be lost as in `RUMM-334`
-    # we're going to re-do sample apps layout and not use local pod anymore.
-    # - check_example_projects
```
ncreated, line 20: We can also remove the `check_example_projects` workflow definition now 🚀

PR #89, opened 2020-04-23
RUMM-298 Write spans to file

ncreated:
### What and why?

🚚 This PR enables writing OT spans to files. Later, those files will be uploaded to Datadog backend 🚀.

### How?

The Tracing architecture for spans serialization follows the one already done for Logging.

`Log` serialization:
* `LogFileOutput` bundles `LogBuilder` and `FileWriter`;
* `LogBuilder` creates the `Log: Encodable` object;
* `FileWriter` writes the `Log` to underlying file.

`Span` serialization:
* `SpanFileOutput` bundles `SpanBuilder` and `FileWriter`;
* `SpanBuilder` creates the `Span: Encodable` object;
* `FileWriter` writes the `Span` to underlying file.

Both `Log` and `Span` implement custom serialization in `LogEncoder` and `SpanEncoder`. Serialized span looks like this:

```json
{
  "trace_id": "1D",
  "span_id": "1",
  "parent_id": "13E",
  "name": "operation",
  "duration": 500000000,
  "type": "custom",
  "service": "abc",
  "error": 0,
  "start": 1576404000000000000,
  "resource": "operation",
  "meta.network.client.sim_carrier.name": "abc",
  "meta.network.client.available_interfaces": ["wifi"],
  "meta.network.client.sim_carrier.iso_country": "abc",
  "meta.network.client.is_expensive": true,
  "meta.network.client.sim_carrier.technology": "LTE",
  "meta.network.client.reachability": "maybe",
  "meta.network.client.supports_ipv4": true,
  "meta.network.client.supports_ipv6": true,
  "meta.network.client.sim_carrier.allows_voip": false,
  "meta.network.client.is_constrained": true,
  "meta._dd.source": "mobile",
  "meta.application.version": "1.0.0",
  "meta.tracer.version": "1.1.0"
}
```

**Notes on testing**: 
* To make assertions on serialized `Data`, we used to use `LogMatcher`. Now, generic logic was extracted to base `JSONDataMatcher` and reused among `LogMatcher` and `SpanMatcher`.
* Some important tests on `Span` JSON format will be added after upload part is finished. This is to balance number of unit tests - instead of testing full format in writting components, we can test it comprehensively with the upload feature (exactly as it is done for logs, where JSON format is tested in `LoggerTests` - with assertions on data delivered to mock server).

**Notes on implementation**: 
* This implementation doesn't yet pass the `DDSpan` to `SpanFileOutput` when `span.finish()` is called - this requires parts of the upload stack to be present, so will be done together with the upload feature.
* Because both Logging and Tracing write/upload stacks are similar, some code can be reused (like `getApplicationVersion()` in builders or `UserInfo` encoding in encoders). Appropriate `TODO` comments were left to do this after we have more comprehensive unit tests for Tracing feature.
* Some assumptions about the JSON format were taken from [SpanSerializer.kt](https://github.com/DataDog/dd-sdk-android/blob/master/dd-sdk-android/src/main/kotlin/com/datadog/android/tracing/internal/domain/SpanSerializer.kt), [DDTracer.java](https://github.com/DataDog/dd-sdk-android/blob/master/dd-sdk-android/src/main/java/datadog/opentracing/DDTracer.java) and other components from [`dd-sdk-android`](https://github.com/DataDog/dd-sdk-android).

### Review checklist

### DIFF
--- a/Sources/Datadog/Tracing/Span/SpanEncoder.swift
+++ b/Sources/Datadog/Tracing/Span/SpanEncoder.swift
@@ -1,0 +1,150 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+/// `Encodable` representation of span.
+internal struct Span: Encodable {
+    let traceID: TracingUUID
+    let spanID: TracingUUID
+    let parentID: TracingUUID?
+    let operationName: String
+    let serviceName: String
+    let resource: String
+    let startTime: Date
+    let duration: TimeInterval
+    let isError: Bool
+
+    // MARK: - Meta
+
+    let tracerVersion: String
+    let applicationVersion: String
+    let networkConnectionInfo: NetworkConnectionInfo?
+    let mobileCarrierInfo: CarrierInfo?
+    let userInfo: UserInfo
+
+    func encode(to encoder: Encoder) throws {
+        try SpanEncoder().encode(self, to: encoder)
+    }
+}
+
+/// Encodes `Span` to given encoder.
+internal struct SpanEncoder {
+    /// Coding keys for permanent `Log` attributes.
+    enum StaticCodingKeys: String, CodingKey {
+        case traceID = "trace_id"
+        case spanID = "span_id"
+        case parentID = "parent_id"
+        case operationName = "name"
+        case serviceName = "service"
+        case resource
+        case type
+        case startTime = "start"
+        case duration
+        case isError = "error"
+
+        // MARK: - Metrics
+
+        case isRootSpan = "metrics._top_level"
+
+        // MARK: - Meta
+
+        case source = "meta._dd.source"
+
+        // MARK: - Application info
+
+        case applicationVersion = "meta.application.version"
+
+        // MARK: - Tracer info
+
+        case tracerVersion = "meta.tracer.version"
+
+        // MARK: - User info
+
+        case userId = "meta.usr.id"
+        case userName = "meta.usr.name"
+        case userEmail = "meta.usr.email"
+
+        // MARK: - Network connection info
+
+        case networkReachability = "meta.network.client.reachability"
+        case networkAvailableInterfaces = "meta.network.client.available_interfaces"
+        case networkConnectionSupportsIPv4 = "meta.network.client.supports_ipv4"
+        case networkConnectionSupportsIPv6 = "meta.network.client.supports_ipv6"
+        case networkConnectionIsExpensive = "meta.network.client.is_expensive"
+        case networkConnectionIsConstrained = "meta.network.client.is_constrained"
+
+        // MARK: - Mobile carrier info
+
+        case mobileNetworkCarrierName = "meta.network.client.sim_carrier.name"
+        case mobileNetworkCarrierISOCountryCode = "meta.network.client.sim_carrier.iso_country"
+        case mobileNetworkCarrierRadioTechnology = "meta.network.client.sim_carrier.technology"
+        case mobileNetworkCarrierAllowsVoIP = "meta.network.client.sim_carrier.allows_voip"
+    }
+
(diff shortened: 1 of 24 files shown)


ncreated:
> I am a bit confused of the relationship between `OpenTracing.Span`, `DDSpan` and `Span`
I was wondering if we could achieve this PR without `Span`?

@buranmert The OT interface [requires](https://github.com/opentracing/specification/blob/master/specification.md#span) mutability of `OpenTracing.Span`, i.e.: 
* `func setOperationName(_ operationName:)`)
* `func setTag(key:, value:)`)
* `func finish(at time: Date)`

Hence, `DDSpan` is a **mutable** class conforming to `OpenTracing.Span`.

Our internal `struct Span: Encodable` is an **immutable** and serializable representation of a finished `DDSpan`. It gets created when `finish(at:)` is called on `DDSpan` and passed further to the SDK serialization stack. This means: all further mutations of `DDSpan` are ignored by design ✨.

We can think of `DDSpan` being a mutable object owned by the user (through OT interface), whereas `Span` is an internal representation of a **finished** `DDSpan`. This makes the identity difference between those two types, so although they have similar name, they represent distinct beings.

Also, the serializable `Span` contains more informations retrieved from different providers running inside SDK (network info, user info, carrier info, ...), so its interface is way different than `DDSpan`.

> As DDSpan is internal and exposed as OpenTracing.Span to public, could we add extra properties from Span as optionals?

This will make the serialization part way more difficult as we will be **serializing a mutable data** Going this direction... we will probably end up with creating an immutable snaphost of `DDSpan`, so ... `Span` 🙂.

Also, having no `SpanBuilder` will mean that `DDSpan` may become a blob type responsible for:
* responding to user calls;
* retrieving SDK-internal metadata from providers;
* doing serialization.

Now `SpanBuilder`, `SpanEncoder` and `DDSpan` abstract single and distinct responsibilities which can be unit tested.

Last, but not least, this is exactly the same principle as we use for the Logging part, where:
* `message: String` and `attributes: [AttributeKey: AttributeValue]` are the raw input received from the user through `Logger` interface (i.e. `logger.info(message:attributes:)`). This input is later turned into `Log: Encodable` inside `LogBuilder`.
* `DDSpan` is the raw input received from the user. This input is later turned into `Span: Encodable` inside `SpanBuilder`.

So we have symmetric architecture, same performance and similar lingo for `Logging` and `Tracing`. Also, the same will scale up to `RUM` product, where `RUMMonitor` will be producing events as user input, and they can be later armed with provider's metadata inside `*Builder` and passed to `*Encoder`.

Sources/Datadog/Tracing/Span/SpanEncoder.swift
```diff
@@ -0,0 +1,150 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+/// `Encodable` representation of span.
+internal struct Span: Encodable {
```
buranmert, line 10: can we rename it to something like `PersistedSpan`, `EncodableSpan` or `FinishedSpan`?

ncreated:
> except we do message -> log in Logging ✅
> while span -> span here ❓

The "span → span" is only true if it comes to naming, `DDSpan` vs `Span`. Those types represent different things:

> We can think of DDSpan being a mutable object owned by the user (through OT interface), whereas Span is an internal representation of a finished DDSpan. This makes the identity difference between those two types, so although they have similar name, they represent distinct beings.

Sources/Datadog/Tracing/Span/SpanEncoder.swift
ncreated: Hmm, I don't like this idea from the architectural glance:

```swift
Log
LogBuilder
LogEncoder
LogSanitizer
```

```swift
PersistedSpan // 😬
SpanBuilder
SpanEncoder
SpanSanitizer  // eventually if we need to sanitize tags
```

WDYT?

buranmert: We asked other mobile devs for opinion and the verdict is `Span`
resolved ✅

PR #91, opened 2020-04-29
RUMM-300 Auto-instrumented URLSession added

buranmert:
### What and why?

SDK needs to provide an easy way to
1. intercept and modify `URLRequest`s in order to inject `tracing` headers
2. observe `URLSessionTask`s to **start** and **end** tracing `span`s

### How?

Given that `URLSession` **cannot be** `subclass`ed as its constructor methods are not inherited, we considered 3 solutions:

1. ❌ **URLProtocol:** allows us to be notified as a new request is being made and modify the request, however this doesn't meet our needs
2. ❓ **Method swizzling:** all the other similar libraries go this way, still researching this solution
3. ✅ **The unnamed way:** this PR implements _the unnamed way_

### Usage

Our users will continue to use `URLSession` in the same way

```swift
let session = URLSession.tracedSession(configuration: config)
let task = session.dataTask(request: request) { data, res, err in ... }
task.resume()
```

### Under the hood

```
caller --- dataTask(url: url) >>> HookedSession
                                       |
                        is the method implemented in HookedSession?
                            / yes                      no \
      execute own implementation          forward method call to self.session
```

1. Although `tracedSession()` returns `URLSession`, the actual returned object is an `NSObject` subclass named `HookedSession`
2. `HookedSession` keeps a real `URLSession` instance as **private property**.
3. `HookedSession` implements methods such as `dataTask(with:completion:)`, `uploadTask(...)`, etc. yet it doesn't implement any other `URLSession` methods.
4. It forwards unimplemented method calls to its `session` private property.

That way, it functions as a `URLSession` transparently while adding an extra layer between calling site and `URLSession`.
In this layer, we can 
1. intercept requests ✅ 
2.observe tasks ✅ 

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/HookedURLSession.swift
+++ b/Sources/Datadog/Core/HookedURLSession.swift
@@ -1,0 +1,280 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+public extension URLSession {
+    internal typealias RequestInterceptor = HookedSession.RequestInterceptor
+    internal typealias TaskObserver = HookedSession.TaskObserver
+
+    static func tracedSession(
+        configuration: URLSessionConfiguration = .default,
+        delegate: URLSessionDelegate? = nil,
+        delegateQueue: OperationQueue? = nil
+    ) -> URLSession {
+        let session = URLSession(
+            configuration: configuration,
+            delegate: delegate,
+            delegateQueue: delegateQueue
+        )
+
+        let requestInterceptor: RequestInterceptor = {
+            // TODO: RUMM-300 Inject tracing headers
+            return $0
+        }
+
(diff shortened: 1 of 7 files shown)


Sources/Datadog/Core/HookedURLSession.swift
```diff
@@ -0,0 +1,280 @@
...
+            previousState = observed.state
+
+            switch observed.state {
+            case .canceling, .completed:
+                self.observations[observed.taskIdentifier] = nil
+            default:
+                break
+            }
+        }
+        observations[task.taskIdentifier] = observation
```
ncreated, line 122: Continuing my thought on threading vs completion callbacks, this for instance would require additional synchronisation of `self.observations` as it might be accessed by different threads. Because `TaskObserver` uses `URLSessionTask`, we'd have to pass the `task` to another thread, possibly changing its lifespan. Some client's app might rely on `task` deallocation callback and here we'd be changing this behaviour 💥. This is too invasive IMO, and again reminds me problems we struggled with before #72.

buranmert:
This code is re-written entirely, PR description is completely out-dated.
I'm closing it, I will open a new PR

PR #92, opened 2020-04-29
RUMM-299 Upload spans to server

ncreated:
### What and why?

📦 This PR adds trace upload, so we can visualize spans on Datadog → APM → Traces:

Screenshot 2020-04-29 at 12 42 37 2

### How?

As `Span` serialization was already done (#89), here we just configure the upload stack for tracing.

Technically, it was possible to reuse the `DataUploadWorker` and share the same principles between both features by making the upload stack more feature-agnostic.
```
| TracingFeature:
| → Storage: FileWriter, FileReader
| → Upload: DataUploadWorker

| LoggingFeature:
| → Storage: FileWriter, FileReader
| → Upload: DataUploadWorker
```

This includes:
* Introduction of `TracesEndpoint` (`LogsEndpoint` counterpart).
* Refactoring of `HTTPHeaders` as tracing feature uses distinct headers.
* Decoupling of `DataFormat` from `FileWriter` and `FileReader` as tracing http payload is way different than the one used for logging.

#### HTTP payload: Logging  vs Tracing
Both features use different way of sending data through HTTP.
* Sending 3 `Logs` in single HTTP message:
```
Content-Type: "application/json"

[{ /* log1 json */ },{ /* log2 json */ },{ /* log3 json */ }]
```
* Sending 3 `Spans` in single HTTP message:
```
Content-Type: "text/plain;charset=UTF-8"

{"spans": [{/* span 1 json */}], "env": "staging"}
{"spans": [{/* span 2 json */}], "env": "staging"}
{"spans": [{/* span 3 json */}], "env": "staging"}
```

This required changes to both how we write and read data from files. This is why `DataFormat` was introduced and two separate instances are used:
```swift
// LoggingFeature.swift
let dataFormat = DataFormat(prefix: "[", suffix: "]", separator: ",")

// TracingFeature.swift
let dataFormat = DataFormat(prefix: "", suffix: "", separator: "\n")
```
This new `DataFormat` becomes now a dependency for `FileWriter` and `FileReader`. Logging feature uses `,` separator when writing serialized logs and `[` prefix + `]` suffix when reading batches. Tracing feature uses `\n` and empty prefix + empty suffix.

💡 A good question is **why we send only one span json per row in tracing HTTP message?** The backend contract requires that all `Spans` within a row must share the same `trace_id`. This will require batching and segregation of `Spans` before they are persisted, which is against the principle of our mobile SDK: _write data as fast as possible so we don't lost it_ (due to crash or process being suspended).

#### Additional things done in this PR

I had to update the way our `SpanMatcher` works in unit tests:
```swift
// before
matcher.assertStartTime(equals: ...)
matcher.assertDuration(equals: ...)

// after
XCTAssertEqual(try matcher.startTime(), ...)
XCTAssertEqual(try matcher.duration(), ...)
```
This was necessary to test parent-child relationship between `Spans`, e.x.:
```swift
XCTAssertGreaterThan(try childMatcher.startTime(), try parentMatcher.startTime())
XCTAssertEqual(try childMatcher.parentSpanID(), try parentMatcher.spanID())
```
As a bonus, it removed `XCTest` dependency from `SpanMatcher` and simplified its code. I think similar refactoring can be done to `LogMatcher` in the future.

The traces upload contract requires that all `span.meta.*` JSON values are `String` and all `span.metrics.*` values are numeric. Instead of adding unit tests for this, I added set of utility protocols in `SpanMatcher` to constraint it clearly.

#### Things coming in next PR(s)

To keep this PR relatively small (still, it's not 😬😞) I detached following to next PR(s):
- [ ] Adding integration tests for tracing with `HTTPMockServer`.
- [x] Adding debug screen to send traces from `Shopist` project.
- [x] Updating `PerformancePreset` so it's not coupled only with logging performance.

### Review checklist


Sources/Datadog/Tracing/TracingFeature.swift
```diff
@@ -14,13 +19,156 @@ internal final class TracingFeature {
...
+                readConditions: ReadableFileConditions(performance: performance),
+                dateProvider: dateProvider
+            )
+
+            // NOTE: any change to tracing data format requires updating the tracing directory url to be unique
+            let dataFormat = DataFormat(
+                prefix: "",
+                suffix: "",
+                separator: "\n"
+            )
```
buranmert, line 61: maybe we can pass `DataFormat` to `obtainTracingFeatureDirectory` function in order not to rely on a in-code comment only
```swift
func obtainTracingFeatureDirectory(for dataFormat: DataFormat) -> Directory {
  let hash = dataFormat.deterministicHash
  let directory = {a path with hash value}
  return directory
}
```
`Hashable` doesn't work for this case as it randomizes seeds at launch.
i'm leaving this idea here only, we may improve in the future

Sources/Datadog/Tracing/TracingFeature.swift
```diff
@@ -14,13 +19,156 @@ internal final class TracingFeature {
...
+            self.uploader = DataUploadWorker(
+                queue: uploadQueue,
+                fileReader: storage.reader,
+                dataUploader: dataUploader,
+                uploadConditions: uploadConditions,
+                delay: DataUploadDelay(performance: performance),
+                featureName: "tracing"
+            )
+        }
+    }
```
buranmert, line 124: could `Upload` class be turned into an `DataUploadWorker` extension?
does it do something extra than `DataUploadWorker`?

Sources/Datadog/Core/Persistence/FileReader.swift
```diff
@@ -43,7 +42,7 @@ internal final class FileReader {
         if let file = orchestrator.getReadableFile(excludingFilesNamed: Set(filesRead.map { $0.name })) {
             do {
                 let fileData = try file.read()
-                let batchData = openingBracketData + fileData + closingBracketData
+                let batchData = dataFormat.prefixData + fileData + dataFormat.suffixData
```
buranmert, line 45: how about turning `DataFormat` into `DataFormatter` and doing `let batchData = formatter.finalize(fileData)` instead of making `FileReader` know the implementation details?

i'm not requesting this change 
because i see `try write(dataFormat.separatorData)` below may not be that easy 🤔

Sources/Datadog/Tracing/TracingFeature.swift
ncreated: Yes, using hash function seems interesting, but it's not much practical when it comes to data cleanup when format changes (when we change it, the SDK should do `previousFormatDirectories.forEach { directory in directory.delete() }`.

I was rather thinking `let version: Int` on the `DataFormat` itself. This would be more correct, as it also enables migration to any past format in the next version (`A` → `B` → `A`).

Anyway, I decided to keep it simple, as we don't yet need any migration nor format change. Also, I'm not sure how the upcoming RUM format will look like, so IMO simple is better for now.

Sources/Datadog/Tracing/TracingFeature.swift
ncreated: The `Upload` struct segregates dependencies on features, configures `DataUploadWorker` with feature-related HTTP headers, creates data upload conditions based on device kind and names the feature:
```swift
if let mobileDevice = appContext.mobileDevice { // mobile device
    httpHeaders = HTTPHeaders(
        headers: [
            .contentTypeHeader(contentType: .textPlainUTF8),
            .userAgentHeader(for: mobileDevice, appName: appContext.executableName, appVersion: appContext.bundleVersion)
        ]
    )
    uploadConditions = DataUploadConditions(
        batteryStatus: BatteryStatusProvider(mobileDevice: mobileDevice),
        networkConnectionInfo: networkConnectionInfoProvider
    )
} else { // other device (i.e. iOS Simulator)
    httpHeaders = HTTPHeaders(
        headers: [
            .contentTypeHeader(contentType: .textPlainUTF8)
            // UA http header will default to the one produced by the OS
        ]
    )
    uploadConditions = DataUploadConditions(
        batteryStatus: nil, // uploads do not depend on battery status
        networkConnectionInfo: networkConnectionInfoProvider
    )
}

self.uploader = DataUploadWorker(
    // ...
    dataUploader: dataUploader,
    uploadConditions: uploadConditions,
    // ...
    featureName: "tracing"
)
```

I don't see a good reason to make `DataUploadWorker` to depend on more things than it should and perform this configuration inside it. Also, this will just move the code around and make mocking more difficult - now we just mock all the dependencies for the feature and it results with consistent `Upload` + `Storage` stacks, but if we make the feature depend on `DataUploadWorker` directly, we'd have to consider corner cases like "mocking Tracing feature with Logging upload stack".

Sources/Datadog/Core/Persistence/FileReader.swift
ncreated: Honestly, I find:
```swift
let batchData = dataFormat.prefixData + fileData + dataFormat.suffixData
```
more readable and straightforward than
```swift
let batchData = formatter.finalize(fileData)
```
no?

buranmert: it was more about taking the responsibility of formatting data out of `FileWriter` and giving it to `DataFormatter`

IMO either way is fine 👍

PR #93, opened 2020-04-30
RUMM-299 Update `PerformancePreset` to not reference to logging feature

ncreated:
### What and why?

🚚 This PR makes the `PerformancePreset` feature-agnostic.

### How?

The main thing is renaming of properties and comments in `PerformancePreset`.

I also did the refactoring which we discussed some time ago:
```diff
- struct WritableFileConditions
- struct ReadableFileConditions
+ protocol StoragePerformancePreset
+ protocol UploadPerformancePreset
```

Now, `PerformancePreset's` interface is segregated using two protocols:
```swift
protocol StoragePerformancePreset {
   var maxFileSize: UInt64 { get }
   // ...
}

protocol UploadPerformancePreset {
   var initialUploadDelay: TimeInterval { get }
   // ...
}

struct PerformancePreset: StoragePerformancePreset, UploadPerformancePreset {
   let maxFileSize: UInt64
   // ...

   let initialUploadDelay: TimeInterval
   // ...
}
```

### Review checklist


Sources/Datadog/Core/Upload/DataUploadDelay.swift
```diff
@@ -15,12 +15,12 @@ internal struct DataUploadDelay {
...
-        self.minDelay = performance.minLogsUploadDelay
-        self.maxDelay = performance.maxLogsUploadDelay
-        self.decreaseFactor = performance.logsUploadDelayDecreaseFactor
-        self.delay = performance.initialLogsUploadDelay
+    init(performance: UploadPerformancePreset) {
+        self.defaultDelay = performance.defaultUploadDelay
+        self.minDelay = performance.minUploadDelay
+        self.maxDelay = performance.maxUploadDelay
+        self.decreaseFactor = performance.uploadDelayDecreaseFactor
+        self.delay = performance.initialUploadDelay
```
buranmert, line 23: it seems like `DataUploadDelay` is made redundant as a duplicate of `UploadPerformancePreset`
they have the exact same properties with almost the same names.

could we remove it?

ncreated: `DataUploadDelay` is more than just a container for constants:
```swift
/// Mutable interval used for periodic data uploads.
internal struct DataUploadDelay {
   // ...

   mutating func nextUploadDelay() -> TimeInterval
   mutating func decrease()
   mutating func increaseOnce()
}
```

It calculates the next upload delay which changes within a ranges described by `UploadPerformancePreset`. It's a separate responsibility and I wouldn't mix it with `UploadPerformancePreset`.

PR #96, opened 2020-05-04
RUMM-437 Encoding doesn't ignore miliseconds anymore

### LINKED ISSUE #95: No sub-second precision in logs
**Describe what happened**
image
Thanks for developing an iOS first party SDK!
We've been successful integrating it and it has been super helpful.
Unfortunately I recently discovered that sub-second precision was missing from the log explorer, and that logs within a same second appear to be randomly ordered.

**Steps to reproduce the issue:**
Log multiple events within a second.

**Expected behaviour:**
Subsecond precision is reported to the log explorer.

**Actual behaviour:**
Subsecond precision is truncated, and it appears that the log sequence within a given second is random 🙀

**Additional context**
 - OS version and device model: Confirmed on `XS @ iOS 13.4.1`
 - Datadog SDK version: `1.0.2`
 - We are piping our logs to `Datadog` using `SwiftyBeaver` with an additional log destination. The sequence of logs is preserved in our other `SwiftyBeaver` destinations: console output, OSLog...

Some code in sequence:
image

You can see that the two logs in the `14:31:51` second are not in the right sequence.

buranmert:
Closes #95 

### What and why?

SDK collects `Log`s and writes them onto disk, then uploads them as raw data
It relies on `JSONEncoder` with `dateEncodingStrategy = .iso8601`
However, it turns out `IS08601` doesn't enforce miliseconds and `JSONEncoder` omits them

```swift
let date = {2020-05-04T18:19:57.109Z}
jsonEncoder.dateEncodingStrategy = .iso8601
jsonEncoder.encode(date) // becomes {2020-05-04T18:19:57Z}
```

The result is very well described in the original issue #95

### How?

There are 2 main solutions:

#### `ISO8601DateFormatter` and `dateEncodingStrategy = .custom`

🍎 Apple recommends using `ISO8601DateFormatter` when dealing with `ISO8601` format.
❌ However, that enforces us to use `dateEncodingStrategy = .custom`
Because the logical strategy option, `.formatted(DateFormatter)`, doesn't accept it as it's not a subclass of `DateFormatter`

#### Customized `DateFormatter` and `dateEncodingStrategy = .formatted`

Although this is not the official way, this method is used in many places/projects.

Basically, we describe the `ISO8601` standard in our own `DateFormatter` instance
Then we pass the instance to `dateEncodingStrategy = .formatted(dateFormatter)`
✅ Now we have miliseconds in the encoded data

## UPDATE:

I turned to the first option: `ISO8601DateFormatter` and `dateEncodingStrategy = .custom`

### Questions?

How can we test **encoding** without making types `Decodable`, since they don't need to be decoded in the main target?

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/Persistence/FileWriter.swift
+++ b/Sources/Datadog/Core/Persistence/FileWriter.swift
@@ -19,11 +19,7 @@ internal final class FileWriter {
     init(orchestrator: FilesOrchestrator, queue: DispatchQueue) {
         self.orchestrator = orchestrator
         self.queue = queue
-        self.jsonEncoder = JSONEncoder()
-        jsonEncoder.dateEncodingStrategy = .iso8601
-        if #available(iOS 13.0, OSX 10.15, *) {
-            jsonEncoder.outputFormatting = [.withoutEscapingSlashes]
-        }
+        self.jsonEncoder = JSONEncoder.default()
     }
 
     // MARK: - Writing data
--- a/Datadog/Datadog.xcodeproj/project.pbxproj
+++ b/Datadog/Datadog.xcodeproj/project.pbxproj
@@ -97,6 +97,9 @@
 		9E08587A242519FF001A3583 /* NetworkPathMonitor.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9E085879242519FF001A3583 /* NetworkPathMonitor.swift */; };
 		9E2FB2722447660E001C9B7B /* Datadog.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 61133B82242393DE00786299 /* Datadog.framework */; };
 		9E36D92224373EA700BFBDB7 /* SwiftExtensionsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9E36D92124373EA700BFBDB7 /* SwiftExtensionsTests.swift */; };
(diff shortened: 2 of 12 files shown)


Sources/Datadog/Core/Persistence/FileWriter.swift
```diff
@@ -20,7 +20,19 @@ internal final class FileWriter {
...
+            // TODO: RUMM-437 Find a way to test encoded values
+            // Taken from https://useyourloaf.com/blog/swift-codable-with-custom-dates/
+            let formatter = DateFormatter()
+            formatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSS"
+            formatter.calendar = Calendar(identifier: .iso8601)
+            formatter.timeZone = TimeZone(secondsFromGMT: 0)
+            formatter.locale = Locale(identifier: "en_US_POSIX")
+            return formatter
+        }()
+        jsonEncoder.dateEncodingStrategy = .formatted(iso8601Formatter)
```
ncreated, line 34: Same should be done in `LogConsoleOutput` to also include factorial seconds in logs printed to the console if users configure the JSON format. I think we could decouple the `iso8601Formatter` to a global constant.

buranmert: done ✅

buranmert:
@ncreated I was thinking of a more complete encoding testing strategy (not only `Date`s) but I couldn't come up with any idea
I guess the test case that you suggested should be fine 👌

PR #99, opened 2020-05-06
RUMM-300 URLSession auto-instrumentation

buranmert:
### What and why?

We'd like to provide our users a convenient way to trace their network operations.
The usage that we aim for is the following:

```swift
let mySession: URLSession = URLSession.shared
Datadog.trace(mySession)
```

### How?

Subclassing `URLSession` is not an option as its `init` methods are not inherited.
However, there are still a few different ways to achieve that:

1. ❌ [**`forwardingTargetForSelector` hack**](https://github.com/DataDog/dd-sdk-ios/pull/91)
  * we considered this highly dangerous since it relies on `unsafeBitCast(:)`
2. ❌ **Method swizzling**
  * this makes changes globally, which we'd like to avoid
3. ✅ **isa swizzling**
  * this PR implements this approach

#### How it works?

In a nutshell, **isa swizzling** is creating a brand new class at runtime and setting it as the actual class of the given object.

Here is how `Swizzler.isaSwizzle(:)` method works step by step

1. It supports only 2 types: `URLSession` and `URLSessionTask`
2. It constructs the new dynamic class name and checks if it already exits
  * if exists, returns it
  * otherwise, proceeds to the next step
3. It allocates the new class **whose superclass is the actual class of the given object**
4. **It adds methods from the appropriate template class to the new class**
  * _Template class of `URLSession`: `HookedURLSession`_
  * _Template class of `URLSessionTask`: `HookedURLSessionTask`_
4. It registers the new class, from now on the runtime is ready to use this class
5. Calls `object_setClass(object, dynamicClass)`

```swift
session -> class = _URLLocalSession
_URLLocalSession -> superclass = URLSession

Swizzler.isaSwizzle(session)

session -> class = dynamicClass
dynamicClass -> superclass = _URLLocalSession
```

**Note that template classes (`HookedURLSession/Task`) are never used!**
They only function as a container for methods. `Swizzler` copies those `Method`s and add them to dynamically created classes.

### Review checklist

- [ ] Find a convenient way to inject code into hooked methods


buranmert:
New unexpected problem: `objc_msgSendSuper` crashes when there is a block among parameters

Sources/Datadog/Core/SessionTracer.swift
```diff
@@ -0,0 +1,54 @@
...
+        let taskObserver: TaskObserver = { task in
+            let observation: NSKeyValueObservation
+            var oldState: URLSessionTask.State? = nil
+            // URLSessionTask.state does NOT support NSKeyValueObservedChange
+            observation = task.observe(\.state) { observedTask, _ in
+                let newState = observedTask.state
+                switch (oldState, newState) {
+                case (.some(.running), .running): // was running, is still running
+                    break
+                case (_, .running): // TODO: RUMM-300 Start span
```
ncreated, line 29: Among [all the possible `task` states](https://developer.apple.com/documentation/foundation/urlsessiontask/state#topics): `running`, `suspended`, `canceling`, `completed`, why `(_, .running)` for determining the "initial" one to start the span? Shouldn't this be `(.suspended, .running)` or `(nil, .running)`?

Sources/Datadog/Core/SessionTracer.swift
```diff
@@ -0,0 +1,54 @@
...
+            // TODO: RUMM-300 Set tracer HTTP header values
+            return originalRequest
+        }
+        let taskObserver: TaskObserver = { task in
+            let observation: NSKeyValueObservation
+            var oldState: URLSessionTask.State? = nil
+            // URLSessionTask.state does NOT support NSKeyValueObservedChange
+            observation = task.observe(\.state) { observedTask, _ in
+                let newState = observedTask.state
+                switch (oldState, newState) {
```
ncreated, line 26: Let's decouple this transition logic to separate function, so we can unit test all possible transitions.

Tests/DatadogTests/Datadog/Core/SessionTracerTests.swift
```diff
@@ -0,0 +1,106 @@
...
+import Datadog
+import _Datadog_Private
+
+private extension String {
+    init(fromUnsafePtr unsafePtr: UnsafePointer<Int8>) {
+        self.init(format: "%s", unsafePtr)
+    }
+}
+
+class SessionTracerTests: XCTestCase {
```
ncreated, line 17: I think we rather have `Swizzler` tests in there as all of them assert things with reflection. We also need the instrumentation unit tests which use `DDTracer()` and `SpanOutputMock()` for recording and asserting spans:
```swift
func testGivenInstrumentedURLSession_whenResponseIsReceived_itSendsSpan() {
   // test 200 OK response
   // test error response
}
```

Sources/_Datadog_Private/Swizzler.m
```diff
@@ -0,0 +1,212 @@
...
+    NSURLRequest *modifiedRequest = [self injected_interceptRequest:request];
+    /*
+     IMPORTANT NOTE: "super" keyword is resolved at compile-time,
+     so [super dataTaskWithRequest:modifiedRequest] would resolve to NSURLSession.dataTaskWithRequest:
+     as NSURLSession is the superclass of this class at compile-time.
+     However, superclass of this method may be different at runtime.
+     Eg: [NSURLSession shared] is _NSURLLocalSession, which is different than NSURLSession.
+     */
+    NSURLSessionDataTask *dataTask = [self super_dataTaskWithRequest:modifiedRequest];
+    [self injected_observeTask:dataTask];
```
ncreated, line 59: `DDSpan` will be created in `injected_observeTask`, but the headers must be modified in `injected_interceptRequest` which is called before. I think this will not let us pass the `trace_id`, no?

Sources/Datadog/Core/SessionTracer.swift
```diff
@@ -0,0 +1,54 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+import _Datadog_Private
+
+// TODO: RUMM-300 This should probably be a part of DDTracer
```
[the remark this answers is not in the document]
buranmert: no, i was thinking of using `Tracer` as a namespace

for example,
```swift
let session = makeASession()
DDTracer.trace(session) // trace(:) is a static function
```

it could be under `Datadog` namespace as well.
it just shouldn't be a free function for the sake of discoverability

Sources/Datadog/Core/SessionTracer.swift
buranmert: i will probably remove this block before merging this PR as it cannot do anything now without `tracing` branch 🧹 

you are right, i think especially `stop` event is problematic here: what if we `downloadTask.suspend()`?
but i just left this naive way of catching start/stop events to showcase how it will be used in the future

Sources/_Datadog_Private/Swizzler.m
ncreated: I was thinking about this for a while. If it's not possible to follow tracer's public API order:
```swift
// on request started
let span = tracer.startSpan()

// on request mutation
let traceHeaders = DDHTTPHeadersWriter()
Global.sharedTracer.inject(spanContext: span.context, writer: traceHeaders)
request.allHTTPHeaderFields += traceHeaders.tracePropagationHTTPHeaders

// on request completed
span.finish()
```

we might add some internal APIs to follow the `URLSession` order:
```swift
// on request mutation
let spanContext = DDSpanContext(...)
let traceHeaders = DDHTTPHeadersWriter()
Global.sharedTracer.inject(spanContext: spanContext, writer: traceHeaders)
request.allHTTPHeaderFields += traceHeaders.tracePropagationHTTPHeaders

// on request started
let span = tracer.startSpan(with: spanContext) // internal tracer API

// on request completed
span.finish()
```

However, if possible, that would be nice to have only one tracer API to use 🙂.

Sources/Datadog/Core/SessionTracer.swift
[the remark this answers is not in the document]
ncreated: I see 👍. I think `Datadog.` is a better choice because of what I discussed [here](https://github.com/DataDog/dd-sdk-ios/pull/99#discussion_r423833722).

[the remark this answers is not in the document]
buranmert: I'd say let's see how to plug tracing and/or RUM in another PR and keep the scope of this PR only to general isa swizzling, what do you think?

Sources/Datadog/Core/SessionTracer.swift
buranmert: 🎗️ I left a `todo` for now, we can implement it when we plug `tracer`

Sources/Datadog/Core/SessionTracer.swift
buranmert: 🎗️ Again, I left a `todo` for now, we can implement it when we plug `tracer`

Sources/Datadog/Core/SessionTracer.swift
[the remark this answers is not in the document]
buranmert: I remove this todo as we will revisit this class with the introduction of `tracer`

Sources/Datadog/Core/SessionTracer.swift
```diff
@@ -0,0 +1,54 @@
...
+public extension Datadog {
+    internal typealias RequestInterceptor = (URLRequest) -> URLRequest
+    internal typealias TaskObserver = (URLSessionTask) -> Void
+
+    static func trace(_ session: URLSession) throws {
```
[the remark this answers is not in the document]
buranmert: As far as I understand, `tracing` and `rum` will require different `requestInterceptor` and/or `taskObserver` blocks.

We can simply pass appropriate interceptor and/or observers to `Swizzler`

```swift
extension Datadog {
  static func intercept(_ session: URLSession, config: Trace/Monitor) {
    if (config == Trace) {
      let interceptor = TracingInterceptor()
    else if (config == Monitor) {
      let interceptor = MonitorInterceptor()
    }
    Swizzler.swizzle(session, requestInterceptor: interceptor, ...)
  }
}
```
We can implement public API as we prefer: either `Datadog.trace()`, `Datadog.intercept()` or any other way. Those calls will only change the configuration of swizzling

Sources/_Datadog_Private/Swizzler.m
buranmert: This is a major issue that we need to discuss, I create another ticket for this

Tests/DatadogTests/Datadog/Core/SessionTracerTests.swift
buranmert: Moved them to `SwizzlerTests` ✅

buranmert:
I suggest keeping swizzling and plugging tracer into `URLSession` as 2 different tasks, therefore I leave discussions related to tracing/RUM unresolved here

I'm closing this PR in favor of #108 , which is rebased onto `tracing` branch

PR #100, opened 2020-05-06
RUMM-332 Add public initializer for `DDTracer`

ncreated:
### What and why?

📦 This PR adds the `public` way of initializing the `DDTracer`

### How?

Following API was introduced:
```swift
let tracer = DDTracer.initialize(
   configuration: DDTracer.Configuration(
      serviceName: "app-tracer"
   )
)
```

As default `serviceName` will soon (`RUMM-409`) come from `Datadog.Configuration`, the defaults resolution was decoupled to a separate `internal` type:
```swift
ResolvedConfiguration(tracerConfiguration: Configuration)
```
After default service name is provided by `Datadog.Configuration`, it will become:
```swift
ResolvedConfiguration(
    tracerConfiguration: Configuration, 
    defaultConfiguration: datadog.instance.configuration
)
```

### Review checklist

### DIFF
--- a/Sources/Datadog/DDTracerConfiguration.swift
+++ b/Sources/Datadog/DDTracerConfiguration.swift
@@ -1,0 +1,32 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+extension DDTracer {
+    /// Datadog Tracer configuration.
+    public struct Configuration {
+        /// The service name that will appear in traces (if not provided or `nil`, the SDK default `serviceName` will be used).
+        public var serviceName: String?
+
+        /// Initializes the Datadog Tracer configuration.
+        /// - Parameter serviceName: the service name that will appear in traces (if not provided or `nil`, the SDK default `serviceName` will be used).
+        public init(serviceName: String? = nil) {
+            self.serviceName = serviceName
+        }
+    }
+
+    /// Datadog Tracer configuration resolved with defaults coming from the other source.
+    internal struct ResolvedConfiguration {
+        internal let serviceName: String
+
+        /// TODO: RUMM-409 Defaults should be resolved using `Datadog.instance`. I don't do it now, because the SDK-wide `serviceName`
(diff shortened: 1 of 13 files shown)


buranmert:
As far as I see, we will have configuration with multiple layers:
1. Tracer config
2. Datadog singleton config
3. Default values

The following should work as we discussed in Slack:
```swift
struct DatadogConfig {
  var serviceName: String = { getDefaultServiceName() }
}
struct TracerConfig {
  var serviceName: String?
}
class Tracer {
  init(config: TracerConfig) {
    let serviceName = config.serviceName ?? DatadogConfig.serviceName
  }
}
```

In order to test the config resolution, we can resolve `TracerConfig` somewhere else, eg:
```swift
extension TracerConfig {
  func resolved(with datadogConfig: DatadogConfig = DatadogConfig.shared) -> TracerConfig {
     return mergedConfigWithDatadogConfig()
  }
}
```
then we can test this method

Sources/Datadog/DDTracerConfiguration.swift
```diff
@@ -0,0 +1,32 @@
...
+    /// Datadog Tracer configuration resolved with defaults coming from the other source.
+    internal struct ResolvedConfiguration {
+        internal let serviceName: String
+
+        /// TODO: RUMM-409 Defaults should be resolved using `Datadog.instance`. I don't do it now, because the SDK-wide `serviceName`
+        ///  must be first done on `master` branch, then merged to `tracing` and update here.
+        init(tracerConfiguration: Configuration) {
+            self.serviceName = tracerConfiguration.serviceName ?? "ios"
+        }
+    }
```
buranmert, line 31: couldn't we use `DDTracer.Configuration` for this? why do we need another type?
(i forgot if we'd already discussed about that)

ncreated: The `DDTracer.Configuration` contains optional `serviceName: String?`. Here we resolve it to non-optional `serviceName: String` to cut-off optionality consideration in other SDK components.

However, this is likely to change after #102 (PR to `master`), so let's assume the `ResolvedConfiguration` will be removed in favour of `Datadog.ValidConfiguration` from #102. I think that when #102 is approved and merged, I'll create a separate, `master` → `tracing` PR to highlight this change and review it separately, as it will be non-trivial merge.

PR #101, opened 2020-05-07
RUMM-403 Add tags to span

ncreated:
### What and why?

🚚 This PR adds support to `Span` tags. Tags can be specified when starting the span:
```swift
let span = tracer.startSpan(..., tags: ["foo": "bar"], ...)
```
or anytime on the `Span` object:
```swift
span.setTag(key: "number", value: 10)
```

Tag consist of a `String` key and value which can be any `Codable` type - `String`, `Int`, `URL`, custom `Codable` structure etc.

### How?

APM API enforces that all tags are encoded as `span.meta.*` JSON attributes **with `String` values**. This means that whatever `Codable` we get from the user, we must translate it into `String`.

This `Codable` → `String` translation happens in `JSONStringEncodableValue`. It turns `Codable` values into their JSON string representation. Given this:
```swift
span.setTag(key: "string", value: "abc")
span.setTag(key: "number", value: 1)
span.setTag(key: "boolean", value: true)
span.setTag(key: "person", value: Person(name: "foo", surname: "bar"))
```
are values are sent as `String`. This is how they are rendered at [app.datadoghq.com](https://app.datadoghq.com):

Screenshot 2020-05-07 at 20 35 20

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/Utils/EncodableValue.swift
+++ b/Sources/Datadog/Core/Utils/EncodableValue.swift
@@ -18,3 +18,35 @@ internal struct EncodableValue: Encodable {
         try value.encode(to: encoder)
     }
 }
+
+/// Value type converting any `Encodable` to its lossless JSON string representation.
+///
+/// For example:
+/// * it encodes `"abc"` string as `"abc"` JSON string value
+/// * it encodes `1` integer as `"1"` JSON string value
+/// * it encodes `true` boolean as `"true"` JSON string value
+/// * it encodes `Person(name: "foo")` encodable struct as `"{\"name\": \"foo\"}"` JSON string value
+///
+/// This encoding doesn't happen instantly. Instead, it is deferred to the actual `encoder.encode(jsonStringEncodableValue)` call.
+internal struct JSONStringEncodableValue: Encodable {
+    /// Encoder used to encode `encodable` as JSON String value.
+    /// It is invoked lazily at `encoder.encode(jsonStringEncodableValue)` so its encoding errors can be propagated in master-type encoding.
+    private let jsonEncoder: JSONEncoder
+    private let encodable: EncodableValue
+
+    init(_ value: Encodable, encodedUsing jsonEncoder: JSONEncoder) {
+        self.jsonEncoder = jsonEncoder
+        self.encodable = EncodableValue(value)
+    }
+
+    func encode(to encoder: Encoder) throws {
+        if let stringValue = encodable.value as? String {
+            try stringValue.encode(to: encoder)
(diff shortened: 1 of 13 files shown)


Sources/Datadog/Tracing/Span/SpanEncoder.swift
```diff
@@ -99,6 +102,15 @@ internal struct SpanEncoder {
         case mobileNetworkCarrierAllowsVoIP = "meta.network.client.sim_carrier.allows_voip"
     }
 
+    /// Coding keys for dynamic `Span` attributes specified by user.
+    private struct DynamicCodingKey: CodingKey {
+        var stringValue: String
+        var intValue: Int?
+        init?(stringValue: String) { self.stringValue = stringValue }
+        init?(intValue: Int) { return nil }
```
buranmert, line 110: why do we need this line?

Sources/Datadog/Core/Utils/EncodableValue.swift
```diff
@@ -15,6 +15,56 @@ internal struct EncodableValue: Encodable {
...
+            try stringValue.encode(to: encoder)
+        } else if let urlValue = encodable.value as? URL {
+            // Switch to encode `url.absoluteString` directly - see the comment in `EncodableValue`
+            try urlValue.absoluteString.encode(to: encoder)
+        } else {
+            let jsonData = try jsonEncoder.encode(encodable)
+            if let stringValue = String(data: jsonData, encoding: .utf8) {
+                try stringValue.encode(to: encoder)
+            }
+        }
```
buranmert, line 68: i'm not sure but we might remove `JSONStringEncodableValue` type by using `lazy` collections somehow
the current version is 👍 as well

Sources/Datadog/Tracing/Span/SpanEncoder.swift
ncreated: This is required to conform to [`CodingKey` protocol](https://developer.apple.com/documentation/swift/codingkey). It's taken from `LogEncoder`, where I didn't found different way around 😞.

Sources/Datadog/Core/Utils/EncodableValue.swift
ncreated: I've been playing with this for quite a time and this `JSONStringEncodableValue` was the best I came up with. 

This part is already well unit-tested (including special `URL` handling, as I found out by default it uses objc key-values to encode `{"relative": absoluteString}` instead of `"absoluteString"`) so we can refactor it once visiting next time if we have a good reason 👌 .

PR #102, opened 2020-05-11
RUMM-409 Unify `service`, `env` and `version` across all SDKs

ncreated:
### What and why?

📦 This PR unifies 3 common attributes across all Datadog SDKs. Similar changes were already applied to Android SDK ([#248](https://github.com/DataDog/dd-sdk-android/pull/248), [#241](https://github.com/DataDog/dd-sdk-android/pull/241), [#240](https://github.com/DataDog/dd-sdk-android/pull/240)).

#### 1/ Environment:
This attribute is specified by the user when initializing the SDK and cannot be overwritten any later:
```swift
let confBuilder = Datadog.Configuration.builderUsing(clientToken:environment:)
```
It is sent to Datadog as default tag in log JSON, i.e.:
```js
{
   /* ... */
   ddtags: ["env:production"]
}
```
💡 NOTE: this changes the public API by introducing mandatory `environment:` argument.

#### 2/ Service Name:
The default value of this attribute can be optionally specified by the user when initializing the SDK:
```swift
let confBuilder = Datadog.Configuration.builderUsing(clientToken:environment:)
confBuilder.set(serviceName: "my-default-service-name")
```
If not given, it defaults to the app bundle identifier (i.e. `com.datadoghq.app`).

The service name can be overwritten per `Logger` instance using `Logger.Builder`:
```swift
let logger = Logger.builder
   .set(serviceName: "custom-service-name")
   .build()
```
It is sent do Datadog as `.service` attribute in log JSON, i.e.:
```js
{
   /* ... */
   service: "my-default-service-name"
}
```

💡 NOTE: this changes the `service` default value, which previously was evaluated to `"ios"`.

#### 3/ Application version:
This attribute is automatically inferred from application `Bundle`. It is sent to Datadog as as `.version` attribute in log JSON, i.e.:
```js
{
   /* ... */
   version: "3.1.4"
}
```
💡 NOTE: this changes the coding key for application version (before, it was `application.version`).

### How?

All three attributes are resolved using both arguments from the public `Datadog.initialize()` API:
```swift
// Datadog.swift

public static func initialize(appContext: AppContext, configuration: Configuration) {
   // ...
}
```

Their resolution can lead to many corner cases, i.e.:
* `configuration.environment` contains invalid characters (e.x. `"*(#@"`).
* when `configuration.serviceName` is not given, it should be inferred from `appContext.bundleIdentifier`, which can also be `nil`.
* `appContext.bundleVertsion` can be `nil` so fallback must be provided.

To not leak this corner cases to other SDK components, I created a type-safe representation of valid SDK configuration:
```swift
/// Valid SDK configuration, passed to the features.
///
/// It takes two types received from the user: `Datadog.Configuration` and `AppContext` and blends them together
/// with resolving defaults and ensuring the configuration consistency.
internal struct ValidConfiguration {
        internal let applicationName: String
        internal let applicationVersion: String
        internal let applicationBundleIdentifier: String
        internal let serviceName: String
        internal let environment: String

        internal let logsUploadURLWithClientToken: URL

        init(configuration: Datadog.Configuration, appContext: AppContext) throws {
            // defaults resolution, consistency and input validation
        }
}
```

The `ValidConfiguration` is now the only dependency passed down to features, which cuts off many edge cases consideration and centralizes unit testing to `ValidConfiguration`.

### Other things done in this PR

It was possible to make `LogBuilder` independent of the public `AppContext`. Now it consumes only plain, non-optional values:
```swift
internal struct LogBuilder {
    let applicationVersion: String
    let environment: String
    let serviceName: String
    let loggerName: String

    // ...
}
```
This turned out to be highly effective, as the `LoggerBuilderTests` do not require full-feature mocking and assertions can be made on `LogOutput` and `LogBuilder`.

Newly introduced `LoggerBuilderTests` unrevealed one bug that we had:
```diff
// Logger
public func sendNetworkInfo(_ enabled: Bool) -> Builder {
-    sendNetworkInfo = true
+    sendNetworkInfo = enabled
    return self
}
```
It's also fixed now 👍.

### Review checklist

- [ ] Update documentation in `docs/` folder, get it reviewed


Sources/Datadog/DatadogConfiguration.swift
```diff
@@ -31,44 +31,126 @@ extension Datadog {
...
+            environment: String,
+            logsUploadURLWithClientToken: URL
+        ) {
+            self.applicationName = applicationName
+            self.applicationVersion = applicationVersion
+            self.applicationBundleIdentifier = applicationBundleIdentifier
+            self.serviceName = serviceName
+            self.environment = environment
+            self.logsUploadURLWithClientToken = logsUploadURLWithClientToken
+        }
```
buranmert, line 132: if we did
```swift
extension ValidConfiguration {
  init(configuration: Datadog.Configuration, appContext: AppContext) { ... }
}
```
we wouldn't need to write this `init(applicationName: ...) { ... }` as it would be generated by the compiler

Sources/Datadog/DatadogConfiguration.swift
```diff
@@ -31,44 +31,126 @@ extension Datadog {
...
+    }
+}
+
+private func ifValid(environment: String) throws -> String {
+    let regex = #"^[a-zA-Z0-9_]+$"#
+    guard environment.range(of: regex, options: .regularExpression, range: nil, locale: nil) != nil else {
+        throw ProgrammerError(description: "`environment` contains illegal characters (only alphanumerics and `_` are allowed)")
+    }
```
buranmert, line 140: `if environment.range(of: ...) == nil { throw Error(...) }` could be simpler/easier to read

Sources/Datadog/DatadogConfiguration.swift
```diff
@@ -31,44 +31,126 @@ extension Datadog {
...
+    return environment
+}
+
+private func ifValid(endpointURL: String, clientToken: String) throws -> URL {
+    guard !endpointURL.isEmpty, let endpointURL = URL(string: endpointURL) else {
+        throw ProgrammerError(description: "`endpointURL` cannot be empty.")
+    }
```
buranmert, line 147: `let endpointURL = URL(string: endpointURL)`
could we use a different name when we change the type?

---

> "`endpointURL` cannot be empty."

`endpointURL` function parameter could be non-nil yet invalid so that `ProgrammerError` with this message would be thrown although `endpointURL` was not empty

ncreated: 👌 Changed to:
```swift
guard let endpointURL = URL(string: endpointURLString) else { ... }
```
which covers 🎸 both things.

Sources/Datadog/DatadogConfiguration.swift
ncreated: Yay, double negation 👿. Done ✅.

Sources/Datadog/DatadogConfiguration.swift
ncreated: Done ✅. Although I don't like using extensions this way - it forces to put some part of the `struct` somewhere else in the file, and doesn't play well for readability in some cases (i.e. nested structs). Here it was relatively fine, as `ValidConfiguration` is the last type in this class extension.

buranmert: Definitely 💯 but this use case is measurably (metric: amount of less code) good without any downside, as long as we keep extension right below struct declaration

IMO this is one of the few (maybe the only one) good use cases of extensions

PR #103, opened 2020-05-12
RUMM-340 Add thread safety to Tracing API

ncreated:
### What and why?

📮 This PR adds additional thread-safety to `public` APIs exposed by Tracing feature.

### How?

I added unit test to randomly and call various public APIs from concurrent threads. Later, fixed all the crashes and Thread Sanitizer issues by using synchronization queue inside `DDTracer`.

This is a counterpart to thread safety done in logging feature, where the `Logger` owns a `DispatchQueue` used to synchronize its public interface.

### Review checklist

### DIFF
--- a/Sources/Datadog/DDTracer.swift
+++ b/Sources/Datadog/DDTracer.swift
@@ -54,7 +54,10 @@ public class DDTracer: Tracer {
 
     internal init(spanOutput: SpanOutput) {
         self.spanOutput = spanOutput
-        self.queue = DispatchQueue(label: "com.datadoghq.tracer", qos: .userInteractive)
+        self.queue = DispatchQueue(
+            label: "com.datadoghq.tracer",
+            target: .global(qos: .userInteractive)
+        )
     }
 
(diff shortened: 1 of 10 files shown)


Sources/Datadog/DDTracer.swift
```diff
@@ -52,6 +54,7 @@ public class DDTracer: Tracer {
 
     internal init(spanOutput: SpanOutput) {
         self.spanOutput = spanOutput
+        self.queue = DispatchQueue(label: "com.datadoghq.tracer", qos: .userInteractive)
```
buranmert, line 57: this way of creating queues is discouraged in [Apple docs](https://developer.apple.com/documentation/dispatch/dispatchqueue)
could we set the target to a global dispatch queue as suggested?

ncreated: 👍 switched to:
```swift
self.queue = DispatchQueue(
    label: ...,
    target: .global(qos: .userInteractive)
)
```
both for `DDTracer` and `Logger`.

PR #105, opened 2020-05-12
RUMM-409 Update public docs with the new, `environment` setting

ncreated:
### What and why?

📦 As we added new property to the public API, this PR updates the online docs accordingly.

I also went through the other content in there and:
* made attributes naming correct;
* unified some method names to industry-way of documenting Swift code.

### Review checklist

### DIFF
--- a/docs/log_collection.md
+++ b/docs/log_collection.md
@@ -53,7 +53,7 @@ github "DataDog/dd-sdk-ios"
 Datadog.initialize(
     appContext: .init(),
     configuration: Datadog.Configuration
-        .builderUsing(clientToken: config.clientToken)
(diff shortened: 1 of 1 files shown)


buranmert:
Can you confirm that also requires an update to https://app.datadoghq.com/logs/onboarding/client ? 
If yes, I can do it

ncreated:
> Can you confirm that also requires an update to https://app.datadoghq.com/logs/onboarding/client ?
> If yes, I can do it

@buranmert good catch 👌. Yes we need this update, indeed. We'd need to synchronize the PR merges. Could you please update it and link the PR here?

PR #108, opened 2020-05-13
RUMM-300 Swizzler is introduced

buranmert:
### What and why?

We'd like to provide our users a convenient way to trace their network operations.
The usage that we aim for is the following:

```swift
let mySession: URLSession = URLSession.shared
Datadog.trace(mySession)
```

#### `TemplateURLSession`

In order to have as much type safety as we could, we introduced a real `NSURLSession` subclass called `TemplateURLSession`

It is hoped to help understanding the runtime behavior of swizzled objects

#### `Swizzler`

`Swizzler.swizzle:` method takes an object, it creates a new class dynamically whose `superclass` is the class of the object.
Then it decorates the dynamic class with method implementations from `TemplateURLSession` class.

In the end, what we have is a dynamic class who uses `TemplateURLSession` methods and whose superclass is the original class of object

_**Before:**_ object -> `_NSURLLocalSession` -> `NSURLSession`
**_After_** object -> `__Datadog_NSURLLocalSession` -> `_NSURLLocalSession`  -> `NSURLSession`

#### `SwizzlerError`

Various errors that may occur during swizzling

### Next step

In the next PR, we will plug `Tracer` into swizzled `session` objects 🔜 

### Review checklist

#### KVO on `URLSessionTaskState`

`Swizzler` does NOT rely on KVO.
However, we are planning to use it to start and stop spans.
Until a few years ago, it was being reported that KVO on `URLSessionTaskState` was dangerous.

For example, AFNetworking switched to method swizzling on `URLSessionTask`: https://github.com/AFNetworking/AFNetworking/issues/1477
However, they had some other issues with method swizzling as well although all seems to have been solved by now ✅ 

**_UPDATE:_** KVO in Swift is officially dangerous, we should go with swizzling `URLSessionTaskState`
https://github.com/apple/swift/pull/20103
❓ ~Shall we at least dogfood KVO on `URLSessionTask.State`?~
➕ ~KVO gives us every state changes with very few lines of code~
➕ ~KVO in Swift has a different API, _which seems like more than a thin wrapper_~
➕ ~Even the most recent related issue is from 2017, the issue might have been fixed in KVO by Apple~

#### Inject request interceptor and task observer via blocks

As we are facing a couple of random crashes, we consider that passing blocks and relying on them retaining their enclosing variables might be an issue.
If this gets proven to be the issue, then we will re-visit `Swizzler` and injecting method implementations into it via blocks

I don't think we should change `Swizzler` right now but I'm leaving this note here in case it may be needed in the future


Sources/_Datadog_Private/include/Swizzler.h
```diff
@@ -0,0 +1,30 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface Swizzler : NSObject
+
++ (BOOL)    swizzle:(NSURLSession *)session
+ requestInterceptor:(NSURLRequest *(^)(NSURLRequest *))requestInterceptor
+       taskObserver:(void(^)(NSURLSessionTask *))taskObserver
+enforceDynamicClassCreation:(BOOL)enforceCreateNewClass
+              error:(NSError **)error;
```
ncreated, line 19: This violates the open-closed principle and won't scale for other features (like RUM auto-instrumentation for `UIViewController`), as swizzling new method requires:
* ~100 lines of very complex code to be added to `Swizzler`;
* even more code for tests;
* testing each swizzling individually.

Current implementation of `Swizzler` is tightly coupled with `URLSession`, but what we need is a domain-agnostic swizzling.


#### Domain-agnostic swizzling

What we need is to:
* _inherit part of a "source object" interface with methods coming from a "template object"_.

A specialisation of this requirement would be what we need toauto-instrumentation the `URLSession`:
* _inherit part of the `URLSession` with methods coming from the `TemplateURLSession`_.

 This sort of API would be properly opened for extensions and closed for modification:
```objc
/// Swizzles `sourceObject` to call corresponding method in `templateObject`
///  before executing the base method in `sourceObject`.
+(BOOL)  inherit:(NSObject *)sourceObject 
byTemplateObject:(NSObject *)templateObject
           error:(NSError **)error;
```

And the `URLSession` use case will be reduced to this:
```swift
let session = URLSession.shared
let sessionTemplate = URLSessionTemplate(requestInterceptor: RequestInterceptor())

try Swizzler.inherit(session, byTemplateObject: sessionTemplate)
```

With this approach, the unit tests will be reduced to:
```swift
class SourceObject: NSObject {
   var someMethodWasCalled = false

   func someMethod() { someMethodWasCalled = true }
}

class TemplateObject: NSObject {
   var someMethodWasCalled = false

   func someMethod() { someMethodWasCalled = true }
}

func testWhenCallingMethodOnASourceObject_theCorrespondingMethodIsCalledOnTemplateObejct() {
   // Given
   let sourceObject = SourceObject()
   let templateObject = TemplateObject()

   try Swizzler.inherit(sourceObject, byTemplateObject: templateObject)

   // When
   sourceObject.method1()

   // Then
   XCTAssertTrue(sourceObject.someMethodWasCalled)
   XCTAssertTrue(templateObject.someMethodWasCalled)
}
```
---
Future use cases in RUM will leverage the code reuse:
```swift
try Swizzler.inherit(viewController, byTemplateObject: viewControllerTemplate)
try Swizzler.inherit(navigationController, byTemplateObject: navigationController)
try Swizzler.inherit(button, byTemplateObject: uiControlTemplate)
```

Sources/Datadog/Core/SessionTracer.swift
```diff
@@ -0,0 +1,46 @@
...
+                .OBJC_ASSOCIATION_RETAIN_NONATOMIC
+            )
+             */
+        }
+
+        try Swizzler.swizzle(
+            session,
+            requestInterceptor: requestInterceptor,
+            taskObserver: taskObserver,
+            enforceDynamicClassCreation: false
```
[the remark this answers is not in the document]
buranmert: 1. If you swizzle `let obj1: Foo`, its dynamic class will always be `__Datadog_Foo`
2. if you swizzle `let obj2: Foo`, `Swizzler` will lookup `__Datadog_Foo` instead of creating a new class

This is not a problem **as long as you use the same `requestInterceptor`/`taskObserver`s**

`enforceDynamicClassCreation` is an attempt to make this case more explicit: if it is `true`, fetching previously created dynamic classes will throw

Another solution could be creating a new class for each swizzled object, but that wouldn't be practical for our actual use case here.

Sources/_Datadog_Private/include/Swizzler.h
buranmert: ### `NSURLSession` Coupling
`Swizzler.h` might be coupled to `URLSession` but `Swizzler.m` is not really. The only part which is `URLSession`-specific is where we add `injected_` methods (ie: `requestInterceptor/taskObserver`), the rest doesn't have `URLSession` references.

`Swizzler.h` has `URLSession` references for the clarity; it would still work if we renamed it to `swizzle:(id)object...`

### Testing

Since we don't have type safety, we need to test every swizzled class/method one by one; there is no shortcut.
If the `Swizzler` could take any kind of object we would need to test every kind of objects; as we only need to swizzle `URLSession` it only takes `session` and we only test `session`s

As a side-note: `TemplateURLSession` class is written only for providing some sort of type safety at compile time, we could write `Swizzler` without it either but then we would need to rely on free functions with probably less type safety.

### Reusability

It's hard to foresee the other use cases for swizzling. `Swizzler` does `isa swizzling`, we will probably not need it for `UIViewController` and go with `method swizzling` as we don't change the execution path unlike in `URLSession`

#### If we swizzle methods of `UIViewController`

We will probably add a new method for method swizzling to `Swizzler` but won't share much between method and isa swizzling.

#### If we `isa` swizzle `UIViewController`

We will need to modify `injected_` methods (as we won't intercept requests in `ViewController`) and keep using the rest of `Swizzler` code

Either way we will need to test every swizzled method, subclass, etc.
Having a generic swizzler won't change that unfortunately.

## Wrap up

At first, injecting blocks into a dynamic class to customize behavior sounded like a good idea to me but now I find it confusing. We can implement them in `TemplateClass` at compile-time and that would be less confusing.
When I do that, `Swizzler` will be working as you described above (since that was the only difference)

However, I prefer keeping `TemplateClass` as a class constant, making it a variable would increase complexity by a lot.
Maybe we can subclass `Swizzler` and override `TemplateClass` such as

```objc
class URLSessionSwizzler: Swizzler {
  override class var templateClass: AnyClass { return URLSession.self }
}
class FooSwizzler: Swizzler {
  override class var templateClass: AnyClass { return Foo.self }
}
```

I'm making the change now 🛠

Sources/Datadog/Core/SessionTracer.swift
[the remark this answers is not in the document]
buranmert: The comment below also addresses the issue here 🙌 
Without `requestInterceptor/taskObserver`, we don't need `enforceDynamicClassCreation` neither 🎉

Sources/_Datadog_Private/include/Swizzler.h
ncreated: > Swizzler.h might be coupled to URLSession but Swizzler.m is not really. The only part which is URLSession-specific is where we add injected_ methods (ie: requestInterceptor/taskObserver), the rest doesn't have URLSession references.

> Swizzler.h has URLSession references for the clarity; it would still work if we renamed it to swizzle:(id)object...

ncreated: > Swizzler.h might be coupled to URLSession but Swizzler.m is not really. The only part which is URLSession-specific is where we add injected_ methods (ie: requestInterceptor/taskObserver), the rest doesn't have URLSession references.

> Swizzler.h has URLSession references for the clarity; it would still work if we renamed it to swizzle:(id)object...

So it sounds like decoupling `injected_` methods from `Swizzler.m` and moving `URLSession` references away from `Swizzler.h` will make it generic ✨. I think the solution with subclassing the `Swizzler` may make this thing right 👌.

> Since we don't have type safety, we need to test every swizzled class/method one by one; there is no shortcut.
If the Swizzler could take any kind of object we would need to test every kind of objects; as we only need to swizzle URLSession it only takes session and we only test sessions

This is totally fine 👌. Let's just create easy solution for testing swizzled things.

The first step IMO is to separate "`Swizzler` tests", from "`URLSession` swizzling tests". Now everything is mixed together in `SwizzlerTests`. Once we have separate test suites, we may look into it and see if some generic helpers can be introduced to make next swizzlings in our codebase trivial. How about 3 levels of tests:
* 1/ `SwizzlerTests` - **tests swizzling** of basic `NSObject`;
* 2/ `URLSessionSwizzlingTests` - **tests the type safety** of swizzled methods - here generic helpers can be a real booster;
* 3/ `URLSessionInterceptorTests` - **tests the logic of intercepting** `URLSession` calls and adding trace information.

> Either way we will need to test every swizzled method, subclass, etc.
Having a generic swizzler won't change that unfortunately.

With the above testing model, new use cases will require only 2nd layer for type safety and 3rd layer for logic tests.

> **If we swizzle methods of UIViewController**
> We will probably add a new method for method swizzling to Swizzler but won't share much between method and isa swizzling.

You're right 👍. Method swizzling is a separate thing that may be added to `Swizzler`, or we may want to separate `MethodSwizzler` from `ISASwilzzler`.

> **If we isa swizzle UIViewController**
> We will need to modify injected_ methods (as we won't intercept requests in ViewController) and keep using the rest of Swizzler code

Yes, so the `Swizzler` is not open for extensions and not closed for modifications. That's why let's generalize it 🚀.

> At first, injecting blocks into a dynamic class to customize behavior sounded like a good idea to me but now I find it confusing. We can implement them in TemplateClass at compile-time and that would be less confusing.
When I do that, Swizzler will be working as you described above (since that was the only difference)

IMO it's better to inject an object (`RequestInterceptor` or sth), instead of separate closures for different things. Single object will be much easier to work with in tests, and way easier to add thread safety if necessary.

> However, I prefer keeping TemplateClass as a class constant, making it a variable would increase complexity by a lot.

Yes, totally, I really like the template class 🙌 . Just keep in mind that we'd probably need more template classes. Even in current use case, I think we need `URLSessionTemplate` and `URLSessionTaskTemplate`, so this way we can reuse `Swizzler` to swizzle `task.resume()`. Although, I'm not yet sure how the completion may be captured on `task`.

buranmert:
Current implementation has 2 main parts:
1. `Swizzler: NSObject`
2. `URLSessionSwizzler: NSObject`

#### `Swizzler` (ObjC)

A proxy/helper class for ObjectiveC runtime and swizzling.
Any swizzler should use its methods instead of `import ObjectiveC.runtime`
 
#### SwizzlerError (ObjC)

`Swizzler` methods usually throws. This file is the collection of those errors.

#### URLSessionSwizzler (Swift)

isa swizzles URLSession instances with methods from `Swizzler`
It takes typed implementations for interceptor and observer as parameters, that's how we can inject `DDTracer` into swizzled class/method implementations.

#### TemplateURLSession (ObjC)

`URLSessionSwizzler` is tightly-coupled to it. It has to be written in ObjC, otherwise Swift doesn't use `message forwarding` and we can't change implementations of methods properly.

---

Previously, unit test cases were covering `URLSessionSwizzler`
Now
- [ ] add thread safety to `Swizzler` (or `URLSessionSwizzler`)
- [ ] add unit tests for `Swizzler` methods

Tests/DatadogTests/DatadogPrivate/URLSessionSwizzlerTests.swift
```diff
@@ -0,0 +1,410 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import XCTest
+import _Datadog_Private
+@testable import Datadog
+
+class URLSessionSwizzlerTests: XCTestCase {
```
ncreated, line 11: Some of the tests in this suite should be a `Swizzler` tests, and here we should be only testing the interesting aspects of `URLSession` swizzling.

What IMO should be moved away:
* `testThirdPartySwizzler()`
* `testOriginalClassGetsSwizzledBy3rdParty()`
* `testSwizzlingTheSwizzled()`
* `testSwizzleMultipleTimes()`

The first two do test if our swizzling plays well with any other swizzling done on the object. The latter couple checks if an object can be swizzled multiple times.

Testing those on a concrete implementations of swizzling (`URLSessionSwizzler`) would introduce too many unit tests for the same logic (I mean, we'd be repeating the same in `UIViewControllerSwizzler`, `UIControlSwizzler` etc.).

Tests/DatadogTests/DatadogPrivate/URLSessionSwizzlerTests.swift
```diff
@@ -0,0 +1,410 @@
...
+            disposeDynamicClass: false
+        )
+        try! URLSessionSwizzler.unswizzle(
+            session2,
+            disposeDynamicClass: true
+        )
+        tearDownTracedSession(session3)
+    }
+
+//    func testConcurrentSwizzling() {
```
ncreated, line 227: Let's uncomment this? Ideally, the thread safety shouldn't be considered on `URLSessionSwizzler`, but on `Swizzler`, so we do not have to repeat the thread-safety logic in all swizzlers. Is it doable?

Tests/DatadogTests/DatadogPrivate/URLSessionSwizzlerTests.swift
```diff
@@ -0,0 +1,410 @@
...
+    // MARK: - URLSessionDataTask methods
+    func testDataTaskWithURL() {
+        let url = URL(string: "http://foo.bar")!
+        let task = setupTracedSession().dataTask(with: url)
+
+        XCTAssertEqual(task.originalRequest?.allHTTPHeaderFields, injectedHeaders)
+
+        tearDownTracedSession()
+    }
+    func testDataTaskWithURLRequest() {
```
ncreated, line 259: Missing newline.

Tests/DatadogTests/DatadogPrivate/URLSessionSwizzlerTests.swift
```diff
@@ -0,0 +1,410 @@
...
+//                swizzledSessions[iteration].1.fulfill()
+//            } catch {
+//                XCTAssertNil(error)
+//            }
+//        }
+//        waitForExpectations(timeout: 0.5, handler: nil)
+//    }
+
+    // MARK: - URLSessionDataTask methods
+    func testDataTaskWithURL() {
```
ncreated, line 251: Missing newline.

Tests/DatadogTests/DatadogPrivate/URLSessionSwizzlerTests.swift
```diff
@@ -0,0 +1,410 @@
...
+    func testDataTaskWithURLRequest() {
+        let url = URL(string: "http://foo.bar")!
+        let request = URLRequest(url: url)
+        let task = setupTracedSession().dataTask(with: request)
+
+        XCTAssertEqual(task.originalRequest?.allHTTPHeaderFields, injectedHeaders)
+
+        tearDownTracedSession()
+    }
+    func testDataTaskWithURLCompletionHandler() {
```
ncreated, line 268: Missing newline.

Tests/DatadogTests/DatadogPrivate/URLSessionSwizzlerTests.swift
```diff
@@ -0,0 +1,410 @@
...
+        let url = URL(string: "http://foo.bar")!
+        let task = setupTracedSession().dataTask(with: url) { _,_,_ in }
+
+        XCTAssertEqual(task.originalRequest?.allHTTPHeaderFields, injectedHeaders)
+
+        tearDownTracedSession()
+    }
+    func testDataTaskWithURLRequestCompletionHandler() {
```
ncreated, line 276: Missing newline.

Tests/DatadogTests/DatadogPrivate/URLSessionSwizzlerTests.swift
```diff
@@ -0,0 +1,410 @@
...
+
+        XCTAssertEqual(dataTaskURL.originalRequest?.allHTTPHeaderFields, injectedHeaders)
+        XCTAssertEqual(dataTaskRequest.originalRequest?.allHTTPHeaderFields, injectedHeaders)
+        XCTAssertEqual(dataTaskURLCompletion.originalRequest?.allHTTPHeaderFields, injectedHeaders)
+        XCTAssertEqual(dataTaskRequestCompletion.originalRequest?.allHTTPHeaderFields, injectedHeaders)
+
+        tearDownTracedSession()
+    }
+
+    func testSwizzledBackgroundSession() {
```
ncreated, line 304: Why do we need to test `.background` configuration for `URLSession` separately? It's still the same `URLSession` object, using the same API, so this test covers `URLSession` logic, not our swizzling.

Sources/_Datadog_Private/Swizzler.m
```diff
@@ -0,0 +1,149 @@
...
+
++ (nullable Class)dynamicClassWith:(NSString *)dynamicClassPrefix
+                        superclass:(Class)superklass
+                         configure:(BOOL(^)(Class newClass))configure {
+    // Get dynamic class if already registered at runtime
+    NSString *superclassName = NSStringFromClass(superklass);
+    const char *dynamicClassName = [self string:superclassName
+                                     withPrefix:dynamicClassPrefix].UTF8String;
+    Class dynamicClass = objc_lookUpClass(dynamicClassName);
+    if (dynamicClass == nil) {
```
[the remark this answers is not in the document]
buranmert: I left a comment above about that:

> Currently URLSessionSwizzler creates only 1 dynamic class at runtime. This was a design decision, avoiding class creation should be faster. We'll see if that goes well with plugging Tracer into dynamic URLSession class; if needed we will create new class per instance.

Firebase uses `UUID()` for every class creation, I was thinking of adding a counter to `URLSessionSwizzler` instead. 
`Swizzler.m` is just a toolkit, it shouldn't be responsible for randomizing class names.

But for now, it seems like every swizzled `URLSession` can share the same behavior and we may not need to create a new class each time. We can change that if needed.

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
```diff
@@ -0,0 +1,81 @@
...
+        target: DispatchQueue.global(qos: .userInteractive)
+    )
+    private static let templateClass: AnyClass = TemplateURLSession.self
+    private static let dynamicClassPrefix: String = "__Datadog"
+
+    static func swizzle(
+        _ session: URLSession,
+        requestInterceptor: @escaping RequestInterceptorIMP,
+        taskObserver: @escaping TaskObserverIMP,
+        enforceNewClassConfiguration: Bool = false
```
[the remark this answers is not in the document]
buranmert: As explained above, we can remove `enforceNewClassConfiguration` parameter if we need different behavior per instance

Tests/DatadogTests/DatadogPrivate/SwizzlerTests.swift
```diff
@@ -0,0 +1,160 @@
...
+        } catch {
+            XCTAssertNotNil(error)
+        }
+    }
+
+    func testUnswizzleSwizzledObject() {
+        let obj = NSObject()
+        try! Swizzler.swizzle(obj, with: DummySubclass.self)
+        do {
+            try Swizzler.unswizzle(obj, ifPrefixed: nil, andDisposeDynamicClass: false)
```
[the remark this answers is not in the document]
buranmert: yes, good idea 👍

Tests/DatadogTests/DatadogPrivate/URLSessionSwizzlerTests.swift
buranmert: First 2 cases test if swizzled `URLSession` forwards `super` calls properly; technically that tests if `super_xxx` methods in `TemplateURLSession` are implemented properly.
`Swizzler` cannot solve this problem, therefore this cannot be its responsibility.

Last 2 cases again should not be `Swizzler`s responsibility.

As far as I see, we have 2 main misunderstandings:

1. `Swizzler` is not a base class nor an abstract class; it's more of a wrapper over `ObjectiveC.runtime`, set of helper/utility methods for swizzling. Having a base class would require us to streamline every kind of swizzling in the same way whereas we may need different kinds of swizzling behavior for different domains, we don't know yet.
2. We need to repeat almost all the test cases for every `XXXSwizzler` implementation. There is no way to know if swizzling is well-implemented without unit tests as we don't have compile-time safety at all (`TemplateClass` approach is _just an attempt_ to improve readability and compile-time safety)

Tests/DatadogTests/DatadogPrivate/URLSessionSwizzlerTests.swift
buranmert: I had 2 issues when I tried to add thread-safety to `Swizzler` instead of `URLSessionSwizzler`

First problem was, `dispatch_a/sync` and `(NSError **)error` don't go well together because of ARC/memory management. Compiler complains about it, although there is a certain syntax to make it work it didn't feel safe to me

But tha main issue was that dispatching one block and getting things done in one go vs. dispatching multiple blocks
As `Swizzler` is a set of methods and `XXXSwizzler`s are usually one or two methods, I prefer syncing the whole operation in one go at `XXXSwizzler` level.
It came simpler, safer and faster. Downside is having to write one more test case for each `XXXSwizzler`, but IMO this is fine

Tests/DatadogTests/DatadogPrivate/URLSessionSwizzlerTests.swift
buranmert: This test case was just a nice-to-have IMO, `.background` sessions don't support completion handlers and they raise an exception. I wanted to see if our swizzled methods work fine in that case as well.

(I wondered about that because I think I saw a few questions/threads somewhere regarding swizzling background session weren't working fine or something like that)

Sources/_Datadog_Private/TemplateURLSession.m
```diff
@@ -0,0 +1,93 @@
...
+// Ivars change the memory layout of the class and needs to be added very carefully unless if it MUST be added
+@implementation TemplateURLSession
+
+- (NSURLRequest *)injected_interceptRequest:(NSURLRequest *)request {
+    NSAssert(false, @"This method should not be run at runtime!");
+    return request;
+}
+- (void)injected_observeTask:(NSURLSessionTask *)task {
+    NSAssert(false, @"This method should not be run at runtime!");
+}
```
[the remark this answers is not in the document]
buranmert: That can be a good idea 👍 
I didn't think about it as `ivar`s is usually a no-go with dynamic classes but associated object could work, let me check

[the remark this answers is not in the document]
ncreated: We can revisit this when we'll be injecting tracer into this swizzling. But my worry is that without a separate object, there won't be a good place to install a thread safety queue required by `requestInterceptor/taskObserver` to operate on the `Span` 😞.

But in general, this level of complexity for adding new swizzling is definitely too much. Hard to imagine the amount of code we'd have to produce for swizzling various components for RUM feature 😥.

Do you see any way we could make it more straightforward?

Sources/_Datadog_Private/Swizzler.m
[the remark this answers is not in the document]
ncreated: I see. Although I wouldn't optimize it for performance, as normally there is from 1 to few `URLSession` instances per app, so the impact would be negligible even if we create different dynamic overload for each.

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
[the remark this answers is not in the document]
ncreated: I'd vote to remove it and provide sustainable default behaviour: creating new instance for each swizzling, as optimizing this is IMO premature.

Having [flag argument](https://martinfowler.com/bliki/FlagArgument.html), brings much more trouble, especially on the testing part, where we have two separate logical branches starting from the top-most abstraction.

Tests/DatadogTests/DatadogPrivate/URLSessionSwizzlerTests.swift
ncreated: Ah, I see 👌👍.

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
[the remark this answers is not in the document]
buranmert: I would probably remove it in the next PR but yes let's do it now 👍

buranmert:
I'm closing this PR in favor of https://github.com/DataDog/dd-sdk-ios/pull/113

`isa` swizzling is considered too complex although it is safe and it works.
We still can go back if we encounter an unexpected problem with method swizzling.

Issue #106, opened 2020-05-13
Crash in FileWriter.init


gransell:
**Describe what happened**
The SDK is crashing on initialisation in FileWriter.init

**Steps to reproduce the issue:**
We have not been able to reproduce this in a dev environment but we are getting reports from our production app. It's currently only happening to a limited amount of users but for them it looks like it crashes on each launch.

Here is the code for initialising:
```swift
Datadog.initialize(appContext: .init(),
    configuration: Datadog.Configuration
        .builderUsing(clientToken: <client-token>)
        .build())
```

And here is the top of the call stack:
```
Crashed: com.apple.main-thread
SIGABRT ABORT 0x000000010491a418

Crashed: com.apple.main-thread
0  ???                            0x10491a418 (Missing)
1  ???                            0x104919a58 (Missing)
2  ???                            0x104919a88 (Missing)
3  ???                            0x1048dc8f8 (Missing)
4  ???                            0x1048dca14 (Missing)
5  libdyld.dylib                  0x1c0b7c848 dyld_stub_binder + 60
6  Datadog                        0x106d286b0 specialized FileWriter.init(orchestrator:queue:maxWriteSize:) + 460 (<compiler-generated>:460)
7  Datadog                        0x106d4b1f8 specialized static LogsPersistenceStrategy.defalut(using:) + 1000 (FileWriter.swift:1000)
8  Datadog                        0x106d31a98 specialized Datadog.__allocating_init(appContext:logsUploadURLProvider:dateProvider:userInfoProvider:networkConnectionInfoProvider:carrierInfoProvider:) + 64 (<compiler-generated>:64)
9  Datadog                        0x106d31fcc specialized static Datadog.initializeOrThrow(appContext:configuration:) + 716 (<compiler-generated>:716)
10 Datadog                        0x106d31478 static Datadog.initialize(appContext:configuration:) + 32 (<compiler-generated>:32)
11 Qapital                        0x1044e4c04 DataDogDestination.init(runtime:dataModel:) + 4379347972 (<compiler-generated>:4379347972)
12 Qapital                        0x1044d6310 specialized static LogController.setupDataDogDestination(runtime:dataModel:) + 4379288336 (DataDogDestination.swift:4379288336)

```
**Expected behaviour:**

**Actual behaviour:**

**Additional context**
 - OS version and device model
   Does not seem to be limited to a specific device model or OS version.
 - Datadog SDK version
   1.1.0
[...]

Issue #70, opened 2020-04-07
`NetworkConnectionInfoProvider` EXC_BAD_ACCESS crash on data upload


00FA9A:
**Describe what happened**
```
Crashed: com.datadoghq.ios-sdk-logs-upload
EXC_BAD_ACCESS KERN_INVALID_ADDRESS 0x0000000000000020
```

We are seeing two separate but likely related crashes. Limited stack traces included below.
```
1
Datadog
$s7Datadog29NetworkConnectionInfoProviderC7currentAA0bcD0Vvg + 152
2
Datadog
$s7Datadog20DataUploadConditionsV010canPerformC0SbyF + 180
3
Datadog
$s7Datadog16DataUploadWorkerC012scheduleNextC004_2E3G27D1790FB5C692514696637CC9BCBLL5afterySd_tFyycfU_ + 600
```

and
```
2
Datadog
$s7Network13NWPathMonitorC7DatadogE15currentPathInfoAD09NWCurrentfG0VyF + 132
3
Datadog
$s7Datadog29NetworkConnectionInfoProviderC7currentAA0bcD0Vvg + 152
4
Datadog
$s7Datadog20DataUploadConditionsV010canPerformC0SbyF + 180
5
Datadog
$s7Datadog16DataUploadWorkerC012scheduleNextC004_2E3G27D1790FB5C692514696637CC9BCBLL5afterySd_tFyycfU_ + 600
```

Will try to provide more detailed stack traces if available.

**Steps to reproduce the issue:**
Unknown

**Expected behaviour:**
No crash

**Actual behaviour:**
Crash; numbers till limited, but not negligible 

**Additional context**
 - OS version and device model: across various models os versions
 - Datadog SDK version: latest beta
 - an explanation of what might cause the bug and/or how it can be fixed: unknown, but about 40-45% of crashes are in background

ncreated:
Hello @00FA9A , thanks for all the details 👍. 

> Datadog SDK version: latest beta

Do you have any numbers from any prior version? In [`1.0.0-beta3`](https://github.com/DataDog/dd-sdk-ios/releases/tag/1.0.0-beta3) we modified `NetworkConnectionInfoProvider` which appears in the stack trace, this would be very handy to know if this is a regression 🙏.

Second, does it happen on `iOS 11` or it's `iOS 12+`? We have different implementation of network info provider for both versions.

00FA9A:
[...]

PR #109, opened 2020-05-15
RUMM-448 Improve `NWPathMonitor` thread safety

ncreated:
### What and why?

🔬 This PR attempts to fix issues with `NWPathMonitor` that were reported by several people. Unfortunately, I wasn't able to reproduce any of reports, so all the changes are based on my research, investigation and intuition.

#### GH Issue #110 

Provided crash logs clearly says we have a problem with some parts of `NWPath` being destroyed while we're accessing it inside `NWPathMonitor.current`:
```
0  libswiftCore.dylib             0x195690314 _swift_release_dealloc + 16
1  libswiftNetwork.dylib          0x1bd5d4d88 destroy for NWPath + 44
2  libswiftNetwork.dylib          0x1bd5d4d88 destroy for NWPath + 44
3  Datadog                        0x1052e79fc NWPathMonitor.current.getter + 22 (NetworkPathMonitor.swift:22)
4  Datadog                        0x1052c2234 DataUploadConditions.canPerformUpload() + 48 (NetworkConnectionInfoProvider.swift:48)
```
Even when we capture the value of `struct NWPath` on local property, this code seems to not be thread safe:
```swift
let info: NWPath = currentPath
_ = info.supportsIPv4
_ = info.supportsIPv6
```
My suspicion is that the `NWPath` internals are based on mutable state, thus it crashes when the `NWPath` is updated by the OS.

#### Issue reported by @jegnux:

`swift_unknownObjectRelease` crash during background → foreground transition (and probably network connection status change - not sure):

image

#### GH Issue #70:
This is clearly the same problem:
```
EXC_BAD_ACCESS KERN_INVALID_ADDRESS 0x0000000000000020

Datadog
$s7Datadog29NetworkConnectionInfoProviderC7currentAA0bcD0Vvg + 152
```
```
EXC_BAD_ACCESS KERN_INVALID_ADDRESS 0x0000000000000020

Datadog
$s7Network13NWPathMonitorC7DatadogE15currentPathInfoAD09NWCurrentfG0VyF + 132
```

#### Other issues

There is a small chance that the `NWPathMonitor` might be the root cause of #106. In this case, we don't have a symbolicated crashlog, so we can't really tell if [`dyld_stub_binder`](https://github.com/opensource-apple/dyld/blob/master/src/dyld_stub_binder.s) attempts to evaluate a symbol coming from `Darwin` → `Network` image, but this is IMO possible, as `NWPathMonitor's` `start()` is invoked at the SDK initialization time.

#### Debugging

To reproduce the issue, I was trying different combinations of following:
* Using `DispatchQueue.concurrentPerform(iterations: 10_000)` to stress test `nwPathMonitor.currentPath` access from multiple threads.
* Transitioning the app between background and foreground while running tests on a real device.
* Switching between wifi, cellular and no Internet connection.

After many runs, I still didn't catch the crash.

#### The fix

First, I got rid of `fetch {}` and `cancel {}` closures, which were the only things retaining `NWPathMonitor` instance. Instead, the monitor is now retained on the property explicitly. Even if it doesn't help, it simplifies the code and should result with cleaner / different stack trace if the issue is back.

Second, looking at [the same problem solved in](https://github.com/rwbutler/Connectivity/issues/22#issuecomment-517235384) [`rwbutler/Connectivity`](https://github.com/rwbutler/Connectivity) project, I decided to update our way of handling the `NWPathMonitor`. This monitor provides to ways of accessing the `NWPath`:
* pulling: `monitor.currentPath`;
* pushing: `monitor.pathUpdateHandler = { path in _ }`.

The pulling model is very convenient for other parts of the SDK, but its documentation doesn't give any guidelines on how to deal with a thread safety (same, when we look at [this implementation](https://github.com/apple/swift/blob/bbe2f6e735e98c7b1d21aa91b804705e750a532f/stdlib/public/Darwin/Network/NWPath.swift#L237), we can only guess what happens internally). 

The pushing model is more explicit - it promises to deliver updates on the `queue` passed to `monitor.start(queue:)`. To leverage both, the thread safety of the pushing model, and keep the convenience of the pulling model, I created a wrapper which exposes `.current` value while using `.pathUpdateHandler = { _ in }` internally.

Because [`rwbutler/Connectivity`](https://github.com/rwbutler/Connectivity) successfully uses `monitor.pathUpdateHandler = { _ in }` to read network info, I hope this will ultimately solve also our problem.

In addition to this PR, I opened a thread on Apple Developer Forum, to hopefully get some more insights on `NWPath` thread safety:

https://forums.developer.apple.com/message/420389#420389

### Review checklist

- [x] I point this PR to `tracing` branch, as we have better setup for running unit tests on device. When approved, I will rebase it onto `master`. ⚠️


ncreated:
In addition to this PR, I opened a thread on Apple Developer Forum, to hopefully get some more insights on `NWPath` thread safety:

https://forums.developer.apple.com/message/420389#420389

Sources/Datadog/Core/System/NetworkPathMonitor.swift
```diff
@@ -7,25 +7,34 @@
...
+        // NOTE: RUMM-448
+        // We experienced a crash on `info.supportsIPv6` during the app transition from background to foreground.
+        //
+        // By looking at the `NWPath` implementation in `Darwin/Network`:
+        // https://github.com/apple/swift/blob/5c4bbf3de96b733154229a5cfdf16096bbdfdb51/stdlib/public/Darwin/Network/NWPath.swift#L174-L198
+        // we suspect that `nw_path_*` functions might behave unreliable for some `nw_path_get_status()` values.
+        //
+        // This is why we resolve `currentPath` details only on the status values that we were able to
+        // stress test with heavy thread concurrency on real device (`.staisfied` & `.unsatisfied`).
+        let isCurrentPathReliable = info.status != .requiresConnection
```
buranmert, line 23: I couldn't understand how `isCurrentPathReliable` variable may help us

Btw, source code of `NWPathMonitor` is very useful for understanding, thanks for finding it 🙏 

Source code shows that once we call `start` with a `queue`, `NWPathMonitor` sets its `currentPath` occassionally within `updateHandler` in `queue` while we  read `currentPath` from other queues.

What we need to do is 
1. the queue is serial (which is already the case ✅ )
2. ensuring that we always read `currentPath` from the same queue with writer queue ❌ 

```swift
class NetworkMonitor {
  private(set) var current: NetworkConnectionInfo?

  init(monitor: NWPathMonitor) {
    ...
    provider.pathUpdateHandler = { [weak self] path in
      // don't read currentPath outside of pathUpdateHandler
      self.current = path.current
    }
    ...
  }
```

Sources/Datadog/Core/System/NetworkConnectionInfoProvider.swift
```diff
@@ -40,12 +40,16 @@ internal protocol NetworkConnectionInfoProviderType {
...
+    private let queue = DispatchQueue(
+        label: "com.datadoghq.network-connection-info",
+        target: .global(qos: .utility)
+    )
+    /// The `NetworkConnectionInfoProviderType` depending on the OS version.
+    private let wrappedProvider: NetworkConnectionInfoProviderType
 
     var current: NetworkConnectionInfo {
-        return fetchBlock()
+        queue.sync { wrappedProvider.current }
```
buranmert, line 52: I just saw that after writing the first comment :)

Yes, this should ensure thread safety 👍 

But if we use `pathUpdateHandler` to read `currentPath`, we can 
1. remove `queue` property from this class
2. reduce amount of `queue.sync {...}` calls
3. make thread safety the responsibility of `NWPathMonitor`; regardless of the queue we provide, we will be reading from the write queue

ncreated:
@buranmert Yes, #110 brought some new light to the problem. We should avoid `NWPath` processing by our own, and consider this to be not thread safe:
```
let info: NWPath = currentPath
_ = info.supportsIPv4
_ = info.supportsIPv6
```
 I updated the PR and description accordingly.

Sources/Datadog/Core/System/NetworkConnectionInfoProvider.swift
```diff
@@ -35,54 +35,152 @@ internal struct NetworkConnectionInfo {
...
+            reachability: NetworkConnectionInfo.Reachability(from: flags),
+            availableInterfaces: Array(fromReachabilityFlags: flags),
+            supportsIPv4: nil,
+            supportsIPv6: nil,
+            isExpensive: nil,
+            isConstrained: nil
         )
-        self.queue = queue
-        self.fetchBlock = {
-            return provider.current
```
buranmert, line 69: As far as I understand, the problem is that `NWPathMonitor` writes into `currentPath` in its `queue` and we read it from other queues which leads to the crash because of disintegrated `currentPath` object.

If that's correct, adding `queue.sync { return provider.current }` to this line should solve it. 

At first I thought using `pathHandler` could be a good idea too but now it seems like it only complicates the problem, it gives a mutable `unsafeCurrentNetworkConnectionInfo` that we don't seem to need

In fact, maybe we don't even need to capture reader `queue` separately, we can do
```swift
self.fetchBlock = { [capturedMonitor = provider] in
  capturedMonitor.queue.sync { return capturedMonitor.currentPath }
}
```
What do you think?

ncreated: > If that's correct, adding `queue.sync { return provider.current }` to this line should solve it.

Synchronizing the `provider.current` on the same queue as we pass to `NWPathMonitor` is exactly what I did in the first iteration of this PR ([first 3 commits](https://github.com/DataDog/dd-sdk-ios/pull/109/files/87ee84bf2441ed3313b5d1796aedcb83af090a6a)). But later, I thought that this is somehow depending on the implementation detail of the `NWPathMonitor`. Yes, the queue used for `pathHandler` seems to be the same which writes the `.currentPath` in [this part of `NWPathMonitor` implementation](https://github.com/apple/swift/blob/bbe2f6e735e98c7b1d21aa91b804705e750a532f/stdlib/public/Darwin/Network/NWPath.swift#L284-L289
):

```swift
nw_path_monitor_set_update_handler(self.nw) { (newPath) in
   self.currentPath = NWPath(newPath)
   if let handler = self._pathUpdateHandler {
      handler(self.currentPath)
   }
}
```

but what I see from the public API docs is:
```swift
/// Start the path monitor and set a queue on which path updates
/// will be delivered.
public func start(queue: DispatchQueue)

/// Get queue used for delivering the pathUpdateHandler block.
public var queue: DispatchQueue?
```
vs:
```swift
/// Access the current network path tracked by the monitor
public var currentPath = NWPath(nil)
```

Which means:
* clear threading information for `pathUpdateHandler` in public API doc;

vs:

* an implicit promise on `.currentPath` threading model based on the current implementation.

buranmert: 1. `NWPathMonitor.currentPath` is a readonly property (it's a `var` but `NWPath` doesn't have `public init`)
2. As it sets `currentPath` internally, it should've made sure its getter/setter were using the same thread
3. But it doesn't do so and we need to deal with it

To my understanding, there is clearly a problem with `NWPathMonitor` implementation. It wasn't even needed to state thread safety of `currentPath` explicitly, it was supposed to be by default.

Anyway, that `pathUpdateHandler` work does add a bit of complexity but it's still at an acceptable level ✅ 
Let's keep it that way and see how it goes 🤞

PR #113, opened 2020-05-20
RUMM-300 Method swizzling

buranmert:
### What and why?

We needed to change the behavior of `URLSession` instances so that our users will be able to trace network calls automatically.

We considered 3 different solutions, this is the 3rd one.

#### `URLProtocol`

A custom `URLProtocol` allows us to detect new network requests, change them and detect when the operation is finished.

However, `URLProtocol` isn't designed to be an observer; it needs to do the actual work by starting the operation. If our custom protocol claims responsibility for a given request, we can't use default protocols (such as `HTTPURLProtocol`) anymore.

#### `NSProxy` approach

https://github.com/DataDog/dd-sdk-ios/pull/91

```swift
let tracedSession = URLSession.traced(session)
```

`URLSession.traced(:)` takes an actual `URLSession` instance, wraps it into a proxy type (eg: `DDURLSessionProxy`) and `unsafeBitcast` it to `URLSession`.
Returned value is `DDURLSessionProxy` at runtime although it is `URLSession` at compile-time.

1. This approach requires `unsafeBitcast`
2. Adds many extra redirections to ObjC message forwarding
3. Doesn't work with `URLSession.shared`

#### `isa` swizzling

https://github.com/DataDog/dd-sdk-ios/pull/108

This solutions is safe and it works. However, it is considered too complex to read/understand/maintain

#### Method swizzling ✅ 

This PR implements `MethodSwizzler`
This is thought to be the simplest working solution to our problem.

⚠️ ⚠️ ⚠️ It still requires `unsafeBitcast` in order to call original method implementation in the swizzled implementation, this is based on an assumption how Swift makes method calls and that might be subject to change. We should keep an eye on it

### Review checklist

### DIFF
--- a/Sources/Datadog/Tracing/Utils/MethodSwizzler.swift
+++ b/Sources/Datadog/Tracing/Utils/MethodSwizzler.swift
@@ -1,0 +1,89 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+internal enum SwizzlingError: LocalizedError {
+    case unknown
+    case classIsNotNSObjectSubclass(className: String)
+    case methodNotFound(selector: String, className: String)
+    case methodWasNotSwizzled(selector: String, className: String)
+}
+
+internal class MethodSwizzler {
+    static let shared = MethodSwizzler()
+
+    private let queue = DispatchQueue(label: "com.datadoghq.methodSwizzlerQueue", target: DispatchQueue.global(qos: .userInteractive))
+    private var implementationCache = [String: IMP]()
+
+    func currentImplementation<TypedIMP>(of selector: Selector, in klass: AnyClass) throws -> TypedIMP {
+        try queue.sync {
+            let foundMethod = try method(with: selector, in: klass)
+            return unsafeBitCast(method_getImplementation(foundMethod), to: TypedIMP.self)
+        }
+    }
+
+    func swizzle(selector: Selector, in klass: AnyClass, with implementation: IMP) throws {
+        try queue.sync {
+            if !isKindOfNSObject(klass) {
+                throw SwizzlingError.classIsNotNSObjectSubclass(className: NSStringFromClass(klass))
+            }
+            let foundMethod = try method(with: selector, in: klass)
+            let cacheKey = methodIdentifier(foundMethod, in: klass)
+            if implementationCache[cacheKey] == nil {
+                implementationCache[cacheKey] = method_getImplementation(foundMethod)
+            }
+            method_setImplementation(foundMethod, implementation)
+        }
+    }
+
+    func unswizzle(selector: Selector, in klass: AnyClass) throws {
+        try queue.sync {
+            let foundMethod = try method(with: selector, in: klass)
+            let cacheKey = methodIdentifier(foundMethod, in: klass)
+            guard let originalImp = implementationCache[cacheKey] else {
+                throw SwizzlingError.methodWasNotSwizzled(selector: NSStringFromSelector(selector), className: NSStringFromClass(klass))
+            }
+            method_setImplementation(foundMethod, originalImp)
+            implementationCache[cacheKey] = nil
+        }
+    }
+
+    private func method(with selector: Selector, in klass: AnyClass) throws -> Method {
+        var methodsCount: UInt32 = 0
+        guard let methods: UnsafeMutablePointer<Method> = class_copyMethodList(klass, withUnsafeMutablePointer(to: &methodsCount) { $0 }) else {
+            throw SwizzlingError.methodNotFound(selector: NSStringFromSelector(selector), className: NSStringFromClass(klass))
+        }
+        defer {
+            free(methods)
+        }
+        for index in 0..<Int(methodsCount) {
+            let method = methods.advanced(by: index).pointee
+            if method_getName(method) == selector {
+                return method
+            }
+        }
+        throw SwizzlingError.methodNotFound(selector: NSStringFromSelector(selector), className: NSStringFromClass(klass))
+    }
+
+    private func methodIdentifier(_ method: Method, in klass: AnyClass) -> String {
+        let klassName = NSStringFromClass(klass)
+        let methodName = NSStringFromSelector(method_getName(method))
+        return "\(klassName)|||\(methodName)"
+    }
+
+    private func isKindOfNSObject(_ klass: AnyClass) -> Bool {
+        let NSObjectClassName: String = NSStringFromClass(NSObject.self)
+        var head: AnyClass? = klass
+        while let headClass = head {
+            if NSStringFromClass(headClass) == NSObjectClassName {
+                return true
+            }
+            head = class_getSuperclass(head)
+        }
+        return false
+    }
+}
--- a/Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
+++ b/Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
@@ -1,0 +1,104 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+// TODO: RUMM-300 plug DDTracer into swizzled methods
+
+internal class URLSessionSwizzler {
+    static let subjectClass = URLSession.self
+
+    private typealias CompletionHandler = (Data?, URLResponse?, Error?) -> Void
+
+    static let sel_DataTaskWithURL = #selector(URLSession.dataTask(with:) as (URLSession) -> (URL) -> URLSessionDataTask)
+    static let sel_DataTaskWithRequest = #selector(URLSession.dataTask(with:) as (URLSession) -> (URLRequest) -> URLSessionDataTask)
+    static let sel_DataTaskWithURLCompletion = #selector(URLSession.dataTask(with:completionHandler:) as (URLSession) -> (URL, @escaping CompletionHandler) -> URLSessionDataTask)
+    static let sel_DataTaskWithRequestCompletion = #selector(URLSession.dataTask(with:completionHandler:) as (URLSession) -> (URLRequest, @escaping CompletionHandler) -> URLSessionDataTask)
+
+    private static let swizzler = MethodSwizzler.shared
+
+    private typealias DataTaskWithURL_C = @convention(c) (AnyObject, Selector, URL) -> URLSessionDataTask
+    private typealias DataTaskWithURL_Block = @convention(block) (AnyObject, URL) -> URLSessionDataTask
+    private typealias DataTaskWithRequest_C = @convention(c) (AnyObject, Selector, URLRequest) -> URLSessionDataTask
+    private typealias DataTaskWithRequest_Block = @convention(block) (AnyObject, URLRequest) -> URLSessionDataTask
+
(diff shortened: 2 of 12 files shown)


Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
```diff
@@ -0,0 +1,104 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+// TODO: RUMM-300 plug DDTracer into swizzled methods
+
+internal class URLSessionSwizzler {
```
ncreated, line 11: I really like the way `URLSessionSwizzler` implementation looks like 👌🏅! It's easy to read and clearly shows how to add new swizzlings.

However, this kind of interface:
```swift
static func swizzle() throws
static func unswizzle() throws
static func swizzleDataTaskWithURL() throws
static func swizzleDataTaskWithRequest() throws
static func swizzleDataTaskWithURLCompletionHandler() throws
static func swizzleDataTaskWithRequestCompletionHandler() throws
```
will force us to break the single responsibility of `URLSessionSwizzler` and bloat it with our custom logic - today for `URLRequest` tracing, but later for RUM product features (resources tracing). It will become impossible to test different pieces of logics individually as the `URLSessionSwizzlerTests` will be touching different products (mocking hell 🔥 ).

The problem with testing will only reveal a wider, architectural issue: as adding new features won't be based on composition, it will require touching the existing stuff, thus may break them i.e. the tracing auto-instrumentation may be broken when adding resource tracing for RUM.

Last, but not least - this is the first `*Swizzler` that we add. For other, most likely, we'll follow the same patterns, and the composition problem will be escalated. Single `UIViewControllerSwizzler` will become giant because of implementing all the RUM auto-instrumentation.

Let's make the interface like this:
```swift
class URLSessionSwizzler {
    static func interceptDataTaskWithURL(_ callback: (...) -> Void) throws
    static func interceptDataTaskWithRequest(_ callback: (...) -> Void) throws
    static func interceptDataTaskWithURLCompletionHandler(_ callback: (...) -> Void) throws
    static func interceptDataTaskWithRequestCompletionHandler(_ callback: (...) -> Void) throws
}
```
so it can be composed:
```swift
class URLSessionTracingInterceptor {
    static func initialize() -> URLSessionTracingInterceptor {
        let interceptor = URLSessionTracingInterceptor()
        URLSessionSwizzler.interceptDataTaskWithRequestCompletionHandler(interceptor.handleURLSessionCalll)
        // ...
        return interceptor
    }

    func handleURLSessionCalll(task: URLSessionTask, completion: (Data?, URLResponse?, Error?)) {}
    // ...
}
```
This way:
* we can unit test Swizzling separately in `URLSessionSwizzlerTests`;
* we can unit test tracing instrumentation separately in `URLSessionTracingInterceptorTests`;
* the `initialize()` method can be tested in as a loosely coupled integration test.

WDYT @buranmert ?

Sources/Datadog/Tracing/Utils/MethodSwizzler.swift
```diff
@@ -0,0 +1,77 @@
...
+            let foundMethod = try method(with: selector, in: klass)
+            let cacheKey = methodIdentifier(foundMethod, in: klass)
+            if implementationCache[cacheKey] == nil {
+                implementationCache[cacheKey] = method_getImplementation(foundMethod)
+            }
+            method_setImplementation(foundMethod, implementation)
+        }
+    }
+
+    func unswizzle(selector: Selector, in klass: AnyClass) throws {
```
ncreated, line 43: We don't need the unswizzling in production code. Let's make it an `internal extension` in `DatadogTests` target. We'll move it to `Datadog` once needed

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    static func swizzleDataTaskWithRequestCompletionHandler() throws {
+        typealias TypedIMP = DataTaskWithRequestCompletion_C
+        typealias TypedBlockIMP = DataTaskWithRequestCompletion_Block
+
+        let sel = sel_DataTaskWithRequestCompletion
+        let typedOriginalImp: TypedIMP = try swizzler.currentImplementation(of: sel, in: subjectClass)
+
+        let newImpBlock: TypedBlockIMP = { [impSelector = sel] impSelf, impURLRequest, impCompletion -> URLSessionDataTask in
```
ncreated, line 98: Don't we need also `URLSessionDataTaskSwizzler` which will allow us intercepting `resume()` and `cancel()` calls?

Tests/DatadogTests/Datadog/Tracing/Utils/MethodSwizzlerTests.swift
```diff
@@ -0,0 +1,163 @@
...
+        let newImp = imp_implementationWithBlock(newBlockImp)
+
+        do {
+            try swizzler.swizzle(selector: sel, in: klass, with: newImp)
+        } catch SwizzlingError.methodNotFound(let unfoundSelector, let searchedClassName) {
+            XCTAssertEqual(unfoundSelector, NSStringFromSelector(sel))
+            XCTAssertEqual(searchedClassName, NSStringFromClass(klass))
+        } catch {
+            XCTAssertNil(error)
+        }
```
ncreated, line 117: I don't understand this assertions 🤔. There should be only one valid control flow for asserting, but now it looks like this test is valid in two scenarios. What are we testing in this test and why we don't assert on `swizzledReturnValue`? Maybe we can also rename this test to `given/when/then` so we increases its readability.

If the goal is to test the error, we should use `XCTAssertThrowsError`:
```swift
XCTAssertThrowsError(try swizzler.swizzle(selector: sel, in: klass, with: newImp)) { error in
    guard case SwizzlingError.methodNotFound(let unfoundSelector, let searchedClassName) = error else {
        XCTFail("Expected `SwizzlingError.methodNotFound` but got \(error)")
        return
    }
    XCTAssertEqual(unfoundSelector, NSStringFromSelector(sel))
    XCTAssertEqual(searchedClassName, NSStringFromClass(klass))
}
```

Also, if the `swizzledReturnValue` is not important for this test, we should use `.mockAny()` to communicate that this can be any string:
```swift
let newBlockImp: TypedBlockIMP = { impSelf -> String in
    return .mockAny()
}
```

Tests/DatadogTests/Datadog/Tracing/Utils/MethodSwizzlerTests.swift
```diff
@@ -0,0 +1,163 @@
...
+        let newImp = imp_implementationWithBlock(newBlockImp)
+
+        do {
+            try swizzler.swizzle(selector: sel, in: klass, with: newImp)
+        } catch SwizzlingError.classIsNotNSObjectSubclass(let className) {
+            XCTAssertEqual(className, NSStringFromClass(klass))
+        } catch {
+            XCTAssertNil(error)
+        }
```
ncreated, line 138: Same thing as for `testSwizzleSuperclassMethod()` - let's increase this test's readability.

tools/lint/sources.swiftlint.yml
```diff
@@ -30,7 +30,7 @@ whitelist_rules: # we enable lint rules explicitly - only the ones listed below
   - force_try
   - force_unwrapping
   - function_default_parameter_at_end
-  - identifier_name
+  # - identifier_name
```
ncreated, line 33: If we need to bypass some linter rules, no reason to disable it globally - Swiftlint has [Disable Rules in Code](https://github.com/realm/SwiftLint#disable-rules-in-code) feature in-build. Let's do either:
```swift
// swiftlint:disable identifier_name
let a: String = ...
// swiftlint:enable identifier_name
```
or:
```swift
let a: String = ... // swiftlint:disable:this identifier_name
```

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
buranmert: This is out of scope of this PR, there is another ticket for this.
This PR basically implements the swizzler which doesn't blow up.

We need to talk about how `Tracer` and `RUM` will be used.
I was thinking of passing `SharedTracer` and `SharedRUM` to `swizzle` methods, this will be easier to read than passing blocks.
Passing blocks was necessary in isa swizzling (because of ObjC<->Swift) but now it's not needed.

Sources/Datadog/Tracing/Utils/MethodSwizzler.swift
buranmert: then we would need to expose `implementationCache` which is worse.
this keeps everything in one place

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
buranmert: yes, if we could KVO it would be trivial but it seems like KVO in Swift is broken due to threading issues.

i'm still trying to figure out how we can do that as `cancel/suspend` don't always get called as far as i see

tools/lint/sources.swiftlint.yml
buranmert: this is a very good point: when i search for `swiftlint:disable:this identifier_name` in the project, there are too many of them. if we need to disable this more than once, let's disable it for good.?
and `identifier_name` doesn't make the code safer unlike `forced_unwrapping` (which also has quite a few exceptions in code)

buranmert:
@ncreated New Relic problem is/was not matching method name with selector, this happens when you change selectors (for example, exchanging implementations)
we keep the selectors, so it should be fine.
_note: testing it in Swift is tricky as there is no `_cmd` argument anymore_

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
ncreated: > This is out of scope of this PR, there is another ticket for this.

OK, we can refactor this in separate PR, as I don't see clearly the idea for using `Tracer` for interception.

---

> I was thinking of passing SharedTracer and SharedRUM to swizzle methods, this will be easier to read than passing blocks.

If we use features directly in the swizzled methods, the `*Swizzler` won't follow the SRP, which will result with the problems I explained here:

> (This) will force us to break the single responsibility of URLSessionSwizzler and bloat it with our custom logic - today for URLRequest tracing, but later for RUM product features (resources tracing). It will become impossible to test different pieces of logics individually as the URLSessionSwizzlerTests will be touching different products (mocking hell 🔥 ).

> The problem with testing will only reveal a wider, architectural issue: as adding new features won't be based on composition, it will require touching the existing stuff, thus may break them i.e. the tracing auto-instrumentation may be broken when adding resource tracing for RUM.

Sources/Datadog/Tracing/Utils/MethodSwizzler.swift
ncreated: OK I see, the `implementationCache` is necessary for unswizzling which we need in tests. Although I don't like it - it runs the code which we need only for tests, I don't see any good way to abstract it just for tests.

tools/lint/sources.swiftlint.yml
ncreated: OK, let's disable this 👌. We keep an eye on code verbosity in code reviews, so I'm OK to remove this rule. But please also update the code and delete all `// swiftlint:disable:this identifier_name` along with deleting (not disabling by `#` comment) the `- identifier_name` rule.

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
buranmert: ✅ Injecting blocks is an option
ℹ️ i'm resolving this thread as we will have this discussion in the next PR

PR #115, opened 2020-05-25
RUMM-442 Tracing dogfooding

ncreated:
### What and why?

📦 This PR brings necessary changes to start dogfooding Tracing product.

### How?

First, it updates the `sdkVersion` so we can differentiate data in Dashboard.

Second, it adds missing public initializer to `DDHTTPHeadersWriter`.

### Review checklist

### DIFF
--- a/Sources/Datadog/Tracing/Propagation/HTTPHeadersWriter.swift
+++ b/Sources/Datadog/Tracing/Propagation/HTTPHeadersWriter.swift
@@ -14,6 +14,8 @@ public class DDHTTPHeadersWriter: HTTPHeadersWriter {
         // https://github.com/DataDog/dd-trace-java/blob/4ba0ca0f9da748d4018310d026b1a72b607947f1/dd-trace-ot/src/main/java/datadog/opentracing/propagation/DatadogHttpCodec.java#L23
     }
 
+    public init() {}
+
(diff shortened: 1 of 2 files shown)


Sources/Datadog/Tracing/Propagation/HTTPHeadersWriter.swift
```diff
@@ -14,6 +14,8 @@ public class DDHTTPHeadersWriter: HTTPHeadersWriter {
         // https://github.com/DataDog/dd-trace-java/blob/4ba0ca0f9da748d4018310d026b1a72b607947f1/dd-trace-ot/src/main/java/datadog/opentracing/propagation/DatadogHttpCodec.java#L23
     }
 
+    public init() {}
```
buranmert, line 17: question regarding `OpenTracing` API: the user always needs to `init` a non-OT type (`let writer = DDHTTPHeadersWriter()` in this case), we cannot make this usage only-OT, correct?

ncreated: Yes, it's user who decides on which `OT.Writer` to use, as [there are multiple ones](https://github.com/opentracing/opentracing-swift/blob/e2b0551de2fd9a5cd9a3d8dff5390440f724283c/Sources/OpenTracing/Format.swift#L13-L34). As Datadog we only support `HTTPHeaders` 👍.

PR #48, opened 2020-03-17
RUMM-290 Make `Logger` tags and attributes thread safe

ncreated:
### What and why?

📦 This fixes #47 

`Logger` tags and attributes were not thread safe, so when mutated parallelly by different threads, the SDK was resulting with a crash on `loggerTags` and `loggerAttributes` access.

### How?

I added basic thread safety using logger-specific `DispatchQueue` with `.sync {}` (writes) and `.async {}` (reads) synchronization.

Following this great [post from Matt Gallagher](http://www.cocoawithlove.com/blog/2016/06/02/threads-and-mutexes.html) this might be not the most performant way of synchronization, but I checked benchmark results and they are still acceptable:

Screenshot 2020-03-17 at 18 48 48

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference

PR #116, opened 2020-05-26
RUMM-452 URLSessionSwizzler implemented

buranmert:
### What and why?

We need to inject hooks into `URLSession` methods so that we can inject `Tracer` methods into them. This PR implements `URLSessionSwizzler` which injects `RequestInterceptor` into `dataTask` creation methods

### How?

Tracing auto-instrumentation has 3 main components:
1. `MethodSwizzler`: generic base class for swizzling methods
2. `URLSessionSwizzler`: swizzles specific `URLSession` methods
3. `URLSessionHooks`: blocks to be injected into `URLSession` methods via `URLSessionSwizzler`

This PR is focused on 2nd component: `URLSessionSwizzler`
This component takes a `RequestInterceptor` block and injects it into 
1. `dataTaskWithURL:completion:`
2. `dataTaskWithRequest:completion` methods

• `RequestInterceptor` takes 1 parameter: `originalRequest`
• It modifies the original request and returns a modified request, if interception is needed.
• Otherwise returns nil and `dataTask` is created as if never swizzled.
• `RequestInterceptor` also returns a block which is executed in `resume` and `completion` of created `task` object: that way we will be able to `start/finish` tracer spans

We don't swizzle methods without completion handler because
1. we would need to implement more advanced/more intrusive techniques, such as isa-swizzling `URLSession.delegate` in order to detect `task.resume()` and `completion` events
2. we assumed that methods with completion blocks are the most common ones

Given that we will need a very similar solution for **monitoring RUM** events, 
1. either, we will need to implement a new `RequestInterceptor` or 
2. we will need to pass an array of `interceptors` to `URLSessionSwizzler`

### Review checklist


Sources/Datadog/DDTracer.swift
```diff
@@ -103,6 +103,17 @@ public class DDTracer: Tracer {
 
     // MARK: - Internal
 
+    internal func createSpanContext() throws -> DDSpanContext {
```
ncreated, line 106: Let's use this method to create the context in `startSpanOrThrow()`, so there's only one control flow for the `DDSpanContext` creation. Otherwise, we'll have to test it separately.

Also, it's better for this method to accept `tracing: TracingFeature` as an argument.
* this will centralize the error throwing (less testing - now we'd have to test as well if this method really throws);
* this will clearly mean that the `URLSession` auto-instrumentation depends on the `TracingFeature.instance`.

Sources/Datadog/Tracing/Utils/URLRequestInterceptor.swift
```diff
@@ -0,0 +1,84 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+import OpenTracing
+
+// TODO: RUMM-452 URLSessionHooks has incomplete implementation
+
+internal enum URLSessionHooks {
```
ncreated, line 12: Why not making this object-oriented instead of a set of static functions (aka structured programming)?

To me, this could be simply:
```swift
internal class URLSessionTracingAutoInstrumentation {
   private let tracingFeature: TracingFeature
   private let domainWhitelist: Set<URL>

   init(domainWhitelist: Set<URL>) throws {
      guard let tracing = TracingFeature.instance else {
         throw ProgrammerError(...) // centralized error throwing API misuse
      }
      self.tracingFeature = tracing
      // swizzle `URLSession`
   }

   deinit {
      // unswizzle `URLSession`
   }

   // ... object-oriented methods, not static functions
}
```
This will nicely connect with the public API by using the same error-logging convention as we do for other things:
```swift
public class Datadog {
    // ...

   static func enableURLSessionInstrumentation(domainWhitelist whitelist: Set<URL>) {
      do {
         TracingFeature.instance.urlSessionInstrumentation = try URLSessionTracingAutoInstrumentation(whitelist)
      } catch {
         consolePrint("\(error)")
      }
   }
}
```

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
```diff
@@ -6,102 +6,146 @@
...
+                let modifiedCompletion: CompletionHandler = { origData, origResponse, origError in
+                    impCompletion(origData, origResponse, origError)
+                    if let completedTask = blockTask {
+                        taskObserver(completedTask, .completed)
+                    }
+                }
+
+                let task = typedOriginalImp(impSelf, sel, interceptionResult.request, modifiedCompletion)
+                blockTask = task
+                taskObserver(task, .created(spanContext: interceptionResult.spanContext))
```
ncreated, line 108: `URLSessionSwizzler` shouldn't know about `spanContext`. This leaks the abstraction.

Sources/Datadog/Tracing/Utils/URLRequestInterceptor.swift
buranmert: I've never heard of the term structured programming but the aim here is/was having immutable stateless static blocks.
currently, the only captured reference is `Global.sharedTracer`, which is good enough yet not perfect.

if we need `TracingFeature`, yes we will probably need to change this design to a singleton because there can be only one instrumentation instance only as we swizzle methods globally.
but if possible, I'd prefer keeping static blocks rather than singletons.

Note on `deinit { unswizzle() }`, we can never call unswizzle because this will unswizzle 3rd party libraries as well

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
buranmert: yes you are right, that part already took me a while before raising this PR

here is the problem:
1. we need `spanContext` in `task.resume()`
2. therefore, we need to carry `spanContext` from task creation step, after `let task = dataTaskWith...`
3. we create `spanContext` at request creation, in `intercept(originalRequest)` call

in short, we need to carry `spanContext` between 3 non-contagious calling sites.
spilling `DDSpanContext` was the simplest working solution i came up with.
i'll be happy to change if you see a better solution

ncreated: How about replacing closures with `URLSessionDataTaskInterceptor` object, which is attached to `URLSessionTask` through associated-object? Like this:
```swift
class URLSessionDataTaskInterceptor {
    private let configuration: TracedDomains
    private var spanContext: SpanContext?
    private var span: SpanContext?

    /// Called right after the task is created.
    func configure(_ request: inout URLRequest) {
        guard configuration.shouldRequestBeIntercepted(request) else {
            return
        }

        let spanContext = tracer.createSpanContext()
        let headersWriter = DDHTTPHeadersWriter()
        headersWriter.inject(spanContext: spanContext)

        for (headerField, value) in headersWritter.tracePropagationHTTPHeaders {
            request.addValue(value, forHTTPHeaderField: headerField)
        }

        self.spanContext = spanContext
    }

    /// Called when task is resumed (so it starts)
    func notifyTaskResumed() {
        if let spanContext = self.spanContext {
            span = tracer.startSpan(with: spanContext!)
        }
    }

    /// Called when task completes
    func notifyTaskCompleted(data: Data?, response: URLResponse?, error: Error?) {
        if let span = self.span {
            span.setTag("status.code": response.statusCode)
            span.finish()
        }
    }
}
```

This can be easily plugged into existing Swizzler's method with even simplifying the implementation and signature (I did not it 🤚):
```swift
internal class URLSessionSwizzler {
    static func interceptDataTaskWithRequestCompletionHandler(
        using interceptor: URLSessionDataTaskInterceptor
    ) {
        typealias TypedIMP = TypedIMPDataTaskWithURLRequestCompletion
        typealias TypedBlockIMP = TypedBlockDataTaskWithRequestCompletion
        let sel = SelectorDataTaskWithRequestCompletion

        let typedOriginalImp: TypedIMP = try swizzler.currentImplementation(of: sel, in: subjectClass)
        let newImpBlock: TypedBlockIMP = { impSelf, impURLRequest, impCompletion -> URLSessionDataTask in
            var mutableRequest = impURLRequest
            interceptor.configure(&mutableRequest)

            let modifiedCompletion: CompletionHandler = { origData, origResponse, origError in
                impCompletion(origData, origResponse, origError)
                interceptor.notifyTaskCompleted(data: origData, response: origResponse, error: origError)
            }

            let task = typedOriginalImp(impSelf, sel, mutableRequest, modifiedCompletion)

            // associate Interceptor with task, so `interceptor.resume()` can be called appropriately:
            task.interceptor = interceptor

            return task
        }

        let newImp: IMP = imp_implementationWithBlock(newImpBlock)
        try swizzler.swizzle(selector: sel, in: subjectClass, with: newImp)
    }
}
```

This way we can simplify the number of closures and we have a good piece of logic: `URLSessionDataTaskInterceptor` to unit test the span instrumentation logic (without even mocking the `URLSession`).

This also scales up for adding more tracing features (like setting error on `Span` object). For RUM product, we could change `URLSessionDataTaskInterceptor` to protocol and provide two separate implementations (one for tracing, one for RUM), so this is composable.

buranmert: but we never need to mock `URLSession`, do we? 🤔 

there are a few problems with that approach

1. why would `URLSessionTask` need an associated object with `tracedDomains`?
2. we would need to create a new `Interceptor` object per `task`; if we'd like to use a singleton `interceptor` then we would need to implement a `request->task` map or something like that
3. if we group those 3 blocks in the same type, that won't win us much and we would need to mock that type for different scenarios (eg: never intercept, intercept and fail at modifying request, intercept and succeed, resume task with X event, etc.)

if i understand well, you want to proxy `DDSpanContext` with a container type so that `URLSessionSwizzler` won't have any reference to `tracing`, but the code above doesn't address that problem unfortunately

currently, we don't actually need further abstraction. i'm thinking of revisiting this problem when we start implementing swizzling for RUM, when we may need more abstraction for more general use cases.

ncreated: > 1. why would URLSessionTask need an associated object with tracedDomains?

That's why I suggested `TracedDomains` instead of `Set<URL>` so we can pass the `tracedDomains` as a reference type. This will add only the size of the reference pointer to each task - negligible memory impact.

> 2. we would need to create a new Interceptor object per task;

Why this could be a problem? With this approach, the task observation state (`spanContext` + `span`) is centralized within an object, instead of being distributed across closure callbacks. Callbacks force us to leak the abstraction like in the discussed line. The `URLSessionDataTaskInterceptor ` is one way to fix it.

> if we'd like to use a singleton interceptor then we would need to implement a request->task map or something like that

I don't see any reason for `request` → `task` mapping. `Task` becomes the owner of the interceptor (through associated object). No other component component needs to communicate with `URLSessionDataTaskInterceptor`. And even if some does in the future, we can inject a delegate through interceptor `init()`.

> if we group those 3 blocks in the same type, that won't win us much and we would need to mock that type for different scenarios (eg: never intercept, intercept and fail at modifying request, intercept and succeed, resume task with X event, etc.)

Do you mean testing? We can unit test the `URLSessionDataTaskInterceptor` like this:
```swift
func testGivenValidURL_itAddsTracingHeadersToTheRequest() {
   // given
   let url = URL(...)
   let urlRequest: URLRequest = URLRequest(url: url)
   let interceptor = URLSessionDataTaskInterceptor(domains: TracedDomains([url.domain]))

   // when
   interceptor.configure(&urlRequest)

   // then
   XCTAssertNotNil(urlRequest.allHeaderFields["x-trace-id-header"])
   XCTAssertNotNil(urlRequest.allHeaderFields["x-span-id-header"])
}
```
Other scenarios seem easy as well. For the "whole thing" testing, we can leverage integration tests, so no need to even swizzle `URLSession` for unit tests suite (and this is what I meant, when saying "mock `URLSession`", sorry 🙌).

> if i understand well, you want to proxy `DDSpanContext` with a container type so that `URLSessionSwizzler` won't have any reference to tracing, but the code above doesn't address that problem unfortunately

Why doesn't? This gets rid of `spanContext` in `URLSessionSwizzler` at all. And scales up to injecting other types of interceptors if we need some for RUM product. When going with closure-based approach, we'd have to mix `tracingSpanContext` with `rumSpanContext` inside `URLSessionSwizzler` 🙅‍♂️ (both optionals, as features might be disabled / enabled).

> currently, we don't actually need further abstraction. i'm thinking of revisiting this problem when we start implementing swizzling for RUM, when we may need more abstraction for more general use cases.

Disagree. I don't see why we may not want to do scalable piece of architecture now and avoid re-thinking the `URLSession` interception concept when RUM product arrives. I wouldn't highlight this, if we don't see an obvious clash between Tracing and RUM with the current approach. And this clash starts in the discussed line:
```swift
taskObserver(task, .created(spanContext: interceptionResult.spanContext))
```
When adding resources interception for RUM, this will become:
```swift
taskObserver(task, .created(
   tracingSpanContext: tracingInterceptionResult.spanContext,
   rumSpanContext: rumInterceptionResult.spanContext
))
```
Which means we need to re-do `URLSwizzling`. That in turn will be very sensitive and dangerous, as tracing auto-instrumentation will be released and rolled out to production by many apps. Instead, we should leverage this and only once go through the `URLSwizzling` roll-out process.

buranmert:
```swift
func swizzle(using interceptor: Interceptor) {
  let newImplementation = {
    interceptor.doSomething()
  }
}
```
if we do that, every call to the swizzled method will use the same `interceptor` object
if we associate this `interceptor` to a `task`, X number of `task`s will share the same `interceptor` object.

we can do this instead:
```swift
func swizzle(using interceptorProvider: (() -> Interceptor)) {
  let newImplementation = {
    let newInterceptor = interceptorProvider()
    newInterceptor.doSomething()
  }
}
```
(or passing `Interceptor.shared` in your proposal and implement a map)
or to keep it simple, we can do this:
```swift
internal struct Payload { // a proxy, container type
  var span: Span?
  var spanContext: SpanContext?
}
...
enum TaskObservationEvent: Equatable {
  case created(payload: Payload)
}
```

or we still can add `RUM` without any change by passing an array `RequestInterceptor`s to swizzled method:
```swift
enum TaskObservationEvent: Equatable {
  case tracingContextAdded(context: DDSpanContext)
  case rumContextAdded(context: RUMContext)
}
...
func swizzle(with interceptors: [RequestInterceptor]) {
  let newImp = {
    let interceptions = interceptors.reduce(originalRequest) { // pass request through interceptor chain }
    let newRequest = interceptions.last.request
    let createdContexts = interceptions.map { $0.context }
    createdContexts.forEach { taskObserver(.contextCreated($0)) }
    ...
  }
}
```

_correction:_ what i mean by revisiting this with RUM is that now we don't have the actual implementation of RUM, we can only educated-guess how it will look, what use-cases we will have. if we make a bad guess now, that can probably lead to a bad abstraction

ncreated: > if we do that, every call to the swizzled method will use the same interceptor object

You're right, I missed this part 👌, but what you propose with:
```swift
func swizzle(using interceptorProvider: (() -> Interceptor))
```
is the easy and correct way to fix it.

> (or passing Interceptor.shared in your proposal and implement a map)

I'd avoid having a map. First, the map would have to be thread safe, so we'd need a `DispatchQueue` to synchronize it, which will open the door for threading issues. Second,  IMO the "one task - one interceptor" abstraction provides much easier mental model to work with.

> or to keep it simple, we can do this: (`struct Payload`)

This is IMO not easier 🤔. This problem requires a state: `spanContext` + `span`. With closures + `Payload` approach, we distribute this state across callbacks, which will be harder to manage.

> what i mean by revisiting this with RUM is that now we don't have the actual implementation of RUM, we can only educated-guess how it will look, what use-cases we will have. if we make a bad guess now, that can probably lead to a bad abstraction

That's why I don't want to go deeply into RUM implementation details 🙌, definitely not 🙅. But what we know even today is that RUM will require separate request interceptor. And what we can quickly grasp from the Android SDK is a visible difference between [`RUMRequestInterceptor`](https://github.com/DataDog/dd-sdk-android/blob/master/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/net/RumRequestInterceptor.kt) and [`TracingRequestInterceptor`](https://github.com/DataDog/dd-sdk-android/blob/master/dd-sdk-android/src/main/kotlin/com/datadog/android/tracing/internal/net/TracingRequestInterceptor.kt). That's why IMO injecting an object handling the interception logic is a way safer strategy than basing the interception on set of tracing-coupled callbacks. Object can be subclassed / abstracted / even duplicated, but callbacks can be only refactored.

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
```diff
@@ -6,102 +6,167 @@
...
+            let newImpBlock: TypedBlockIMP = { impSelf, impURL, impCompletion -> URLSessionDataTask in
+                if let interceptionResult = interceptor(impSelf.urlRequest(with: impURL)) {
+                    weak var blockTask: URLSessionDataTask? = nil
+                    let modifiedCompletion: CompletionHandler = { origData, origResponse, origError in
+                        impCompletion(origData, origResponse, origError)
+                        blockTask?.payload?(.completed)
+                    }
+                    let task = typedRedirectedImp(impSelf, sel, interceptionResult.request, modifiedCompletion)
+                    // TODO: RUMM-452 report error?
+                    try? Resume.swizzle(in: task)
```
ncreated, line 68: I think the error handling will be easier to reason about if we centralize the swizzling code in a single place. Now, the swizzling is done in a deep-search way:
```swift
swizzleURLSession()
   → DataTask_URL_Completion.swizzle()
         → Resume.swizzleIfNotSwizzled()
   → DataTask_Request_Completion.swizzle()
         → Resume.swizzleIfNotSwizzled()
```

If we cut off the nesting and do it in a flat way:
```swift
swizzleURLSession()
   → let swizzleOnce: (URLSessionTask) -> Void = { task in URLSessionTaskSwizzler(task).swizzle() }
   → DataTask_URL_Completion.swizzle(: ..., followWith: { task in swizzleOnce(task) })
   → DataTask_Request_Completion.swizzle(: ..., followWith: { task in swizzleOnce(task) })
```
this will be IMO much easier to follow. Other benefits:
* the `swizzleIfNotSwizzled()` method will be no longer needed (we can easily make `swizzleOnce` closure being called only once in the local scope);
* the code will be more decoupled, into: `URLSessionSwizzler` + `URLSessionTaskSwizzler`;
* `URLSessionTask` swizzling and its errors can be tested separately.

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
```diff
@@ -6,102 +6,167 @@
...
+}
+
+internal func swizzleURLSession(interceptor: @escaping RequestInterceptor) throws {
+    do {
+        try URLSessionSwizzler.DataTask_URL_Completion.swizzle(interceptor: interceptor)
+        try URLSessionSwizzler.DataTask_Request_Completion.swizzle(interceptor: interceptor)
+    } catch {
+        try URLSessionSwizzler.DataTask_URL_Completion.unswizzle()
+        try URLSessionSwizzler.DataTask_Request_Completion.unswizzle()
+        throw error
```
ncreated, line 27: How will we handle the `unswizzle()` errors? It seems that we cannot guarantee a non-intrusive behaviour for the customer app with the current implementation.

Tests/DatadogTests/Datadog/Tracing/Utils/URLSessionSwizzlerTests.swift
```diff
@@ -7,107 +7,242 @@
...
+
+        let task = session.dataTask(with: URLRequest.mockAny()) { _, _, _ in }
+
+        XCTAssertEqual(task.originalRequest?.url, modifiedURL)
+    }
+}
+
+// MARK: - 3rd party swizzlers
+
+class ThirdPartySwizzlingTests: XCTestCase {
```
ncreated, line 161: If we need to test 3rd party swizzling against a specific swizzler implementation: `URLSessionSwizzler`, does it mean we'll need it for other swizzlerts as well?
* `UIViewControllerSwizzler` + `ThirdPartyUIViewControllerSwizzlingTests`,
* `UIViewNavigationControllerSwizzler` + `ThirdPartyUIViewNavigationControllerSwizzlerTests`,

Sources/Datadog/Tracing/Utils/MethodSwizzler.swift
```diff
@@ -6,56 +6,118 @@
...
+
+    private func sync<T>(block: () throws -> T) throws -> T {
+        objc_sync_enter(self)
+        defer { objc_sync_exit(self) }
+        return try block()
+    }
+
+    private func set(newIMP: IMP, for found: FoundMethod) {
+        let cacheKey = methodIdentifier(for: found)
+        if implementationCache[cacheKey] == nil {
```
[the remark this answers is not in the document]
buranmert: `implementationCache` is private, therefore it doesn't need to be explicitly thread-safe as long as every non-private method is wrapped with `sync`

`set(newIMP:)` is also private and never should be called outside of `sync { ... }`

Sources/Datadog/Tracing/Utils/MethodSwizzler.swift
```diff
@@ -86,4 +160,18 @@ internal class MethodSwizzler {
         }
         return false
     }
+
+    // MARK: - DEBUG only
+
+    // TODO: RUMM-452 only for unit tests?
+    #if DEBUG
+    func unswizzleALL() {
```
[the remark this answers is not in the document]
buranmert: exposing `implementationCache` will also expose private `FoundMethod` type, which is not needed.

this method should not have any long-term effect as it is a one-off method (doesn't impose a pattern)
it shouldn't make a huge difference in code coverage neither, but if this is a major concern we can just remove this `#if DEBUG #endif`?

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
```diff
@@ -6,102 +6,167 @@
...
-        static let DataTaskWithRequest = #selector(URLSession.dataTask(with:) as (URLSession) -> (URLRequest) -> URLSessionDataTask)
-        static let DataTaskWithURLCompletion = #selector(URLSession.dataTask(with:completionHandler:) as (URLSession) -> (URL, @escaping CompletionHandler) -> URLSessionDataTask)
-        static let DataTaskWithRequestCompletion = #selector(URLSession.dataTask(with:completionHandler:) as (URLSession) -> (URLRequest, @escaping CompletionHandler) -> URLSessionDataTask)
+// hook into URLSessionTask creation
+internal typealias RequestInterceptor = (URLRequest) -> InterceptionResult?
+
+// modifying URLRequest for task creation, observer block to run at resume/completion
+internal typealias InterceptionResult = (request: URLRequest, taskPayload: TaskObserver)
+internal typealias TaskObserver = (TaskObservationEvent) -> Void
+internal enum TaskObservationEvent: Int, Equatable {
```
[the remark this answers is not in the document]
buranmert: this PR shouldn't have contained this change, we can decide on that in the next PR where we implement a `taskObserver`

i'm removing this change ✅

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
```diff
@@ -6,102 +6,167 @@
...
-    enum Selectors {
-        static let DataTaskWithURL = #selector(URLSession.dataTask(with:) as (URLSession) -> (URL) -> URLSessionDataTask)
-        static let DataTaskWithRequest = #selector(URLSession.dataTask(with:) as (URLSession) -> (URLRequest) -> URLSessionDataTask)
-        static let DataTaskWithURLCompletion = #selector(URLSession.dataTask(with:completionHandler:) as (URLSession) -> (URL, @escaping CompletionHandler) -> URLSessionDataTask)
-        static let DataTaskWithRequestCompletion = #selector(URLSession.dataTask(with:completionHandler:) as (URLSession) -> (URLRequest, @escaping CompletionHandler) -> URLSessionDataTask)
+// hook into URLSessionTask creation
+internal typealias RequestInterceptor = (URLRequest) -> InterceptionResult?
+
+// modifying URLRequest for task creation, observer block to run at resume/completion
+internal typealias InterceptionResult = (request: URLRequest, taskPayload: TaskObserver)
```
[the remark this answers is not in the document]
buranmert: right 👍

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
buranmert: i considered that but it didn't turn out to be as good as i imagine

1. currently, the new implementation of `dataTask...` method is in the same place as a whole. we don't need to look at different places to understand what it does. IMO this is easier to read.
2. swizzled` resume` implementation relies on `task.payload` associated object, payload is set by `dataTask...` implementation; so they always have this implicit coupling

i think i didn't understand how this proposal would help with decoupling 🤔 can you explain a bit more please?

i would avoid `swizzleOnce` because there is no public API guarantee that the `dataTask` type will always be the same: `URLSession` has different configs and i suppose different configs may produce different data task types somehow

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
buranmert: you are right 👍 

it should be
```swift
catch {
  try? URLSessionSwizzler.DataTask_URL_Completion.unswizzle()
  try? URLSessionSwizzler.DataTask_Request_Completion.unswizzle()
  throw error
}
```
as we attempt to revert, but if that fails too we can't do anything further and we should give the user what went wrong in the first place

Tests/DatadogTests/Datadog/Tracing/Utils/URLSessionSwizzlerTests.swift
buranmert: correct, 3rd party swizzling tests make sure that new implementation doesn't swallow old implementation of method. wherever we provide a new implementation for a method, ideally we should test it with a 3rd party swizzler

but when needed, we can make 3rd party swizzler more generic
currently it's more of a simplicity-first class

Sources/Datadog/Tracing/Utils/MethodSwizzler.swift
```diff
@@ -6,56 +6,118 @@
...
-        try queue.sync {
-            if !isKindOfNSObject(klass) {
-                throw SwizzlingError.classIsNotNSObjectSubclass(className: NSStringFromClass(klass))
-            }
-            let foundMethod = try method(with: selector, in: klass)
-            let cacheKey = methodIdentifier(foundMethod, in: klass)
-            if implementationCache[cacheKey] == nil {
-                implementationCache[cacheKey] = method_getImplementation(foundMethod)
+        try sync {
+            assert(isKindOfNSObject(klass), "\(klass) is not NSObject subclass: swizzling may not work!")
```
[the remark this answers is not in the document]
buranmert: yes, good idea 👍 
actually, we shouldn't even check that. if someone comes up with an idea to swizzle non-NSObject class one day, we shouldn't be preventing them from doing that.

i'm removing this check ✅

Sources/Datadog/Tracing/Utils/MethodSwizzler.swift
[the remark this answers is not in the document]
ncreated: My point is that even the private code should be fault tolerant. I don't see any thread safety tests in `MethodSwizzlerTests`. That would be well enough if the thread safety is enforced by the implementation (it is not).

Ideally, we should have both - solid implementation and thread safety tests - missing them was the root cause of the prod issue fixed in #48 .

Sources/Datadog/Tracing/Utils/MethodSwizzler.swift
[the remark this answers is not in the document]
ncreated: The `implementationCache` is `[String: IMP]()`, so it doesn't depend on `FoundMethod`, neither `classSelectorPair(from:)` does.

> it is a one-off method (doesn't impose a pattern)

I prefer to be ultimately consequent for such things. If we allow for this once, why not another time in another place?

Sources/Datadog/Tracing/Utils/MethodSwizzler.swift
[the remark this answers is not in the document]
buranmert: #48 is obviously a different case, a public property wasn't thread safe.
there is no way `implementationCache` can be accessed outside of `sync {...}` block in the current implementation. 

do you mean if one day someone changes the implementation and accesses it outside of `sync`?

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
ncreated: But `try?` will leave the customer app in the inconsistent state as it only ignores the failure ⚠️.

The problem is that we only know if the swizzling error happens while we're swizzling, which already modifies the things:
```swift
try URLSessionSwizzler.DataTask_URL_Completion.swizzle(interceptor: interceptor)
```

A typical approach to this is to separate configuration from actual logic:
```swift
// instead of:
do {
   try doTheThing()
} catch {
   // 🤷‍♂️ can't undo, the thing is PARTIALLY done
}

// make it:
do {
   let preparation = try prepareTheThing()
   doTheThing(preparation)
} catch {
   // 👍 tell the user we can't do the thing
   // ⭐️ nothing to undo as the thing was not done
}
```

Sources/Datadog/Tracing/Utils/MethodSwizzler.swift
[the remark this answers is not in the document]
buranmert: sorry, my bad: yes, `classSelectorPair(from:)` returns another private tuple type, not `FoundMethod`

what i mean by one-off is this method is not part a pattern, we won't need to implement a similar method if we scale this design up to other things.
IMO we should be pragmatic, if `#if DEBUG` meets our needs nicely then we should use it. fortunately, that doesn't happen often.

but to resolve this discussion, i'm removing `#if DEBUG` ✅

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
ncreated: > i think i didn't understand how this proposal would help with decoupling 🤔 can you explain a bit more please?

I was thinking that if both:
* `session.dataTaskWithURLCompletion()`
* and `session.dataTaskWithURLRequestCompletion()`

can result with either "swizzling OK" or "swizzling FAIL" and the `task.resume()` swizzling can result with "OK" or "FAIL" this builds up 8 control flows to test. By separating task swizzling from session swizzling, there would be only 2 scenarios in the prior and 4 in the latter. But turns out that my assumption is not true, as "session FAIL, resume OK" is not a possible control flow.

Let's keep it the way it is 👌.

> i would (NOT) avoid swizzleOnce because there is no public API guarantee that the dataTask type will always be the same: URLSession has different configs and i suppose different configs may produce different data task types somehow

I see, sounds fair enough 👍.

Tests/DatadogTests/Datadog/Tracing/Utils/URLSessionSwizzlerTests.swift
ncreated: 👍 ok, let's keep it concrete type now and we'll try to make it generic in future use cases.

Sources/Datadog/Tracing/Utils/MethodSwizzler.swift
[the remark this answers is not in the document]
ncreated: > do you mean if one day someone changes the implementation and accesses it outside of sync?

Yup, I mean potential regression, which could be avoided.

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
buranmert: looks like a good idea 👍 
i made the change but that led to a non-trivial refactoring, you may wanna review this PR once again

`MethodSwizzler` was throwing only in the case of "method not found".
I changed its API in a way that methods take `FoundMethod`, that way methods become non-throwing.
Now we first find the method, only if method is found then we go on with swizzling; so no need to undo swizzling in case of error anymore ✅

Sources/Datadog/Tracing/Utils/MethodSwizzler.swift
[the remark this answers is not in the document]
buranmert: i don't think that is necessary but given it'll be harmless, i'll do that ✅

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
```diff
@@ -6,102 +6,195 @@
...
+            methodToSwizzle: method_dataTask_request
+        )
+        swizzler.set(newIMP: newImp_request, for: method_dataTask_request)
 
-    static func swizzleDataTaskWithRequest() throws {
-        typealias TypedIMP = TypedIMPs.DataTaskWithRequest
-        typealias TypedBlockIMP = TypedBlocks.DataTaskWithRequest
-        let sel = Selectors.DataTaskWithRequest
+        hasSwizzledBefore = true
+    }
```
ncreated, line 55: I found this `IMP`-dance very fragile and error-prone. In fact, I tried to mess it up randomly, and it compiled, so even if we enforce the type-safety for `IMP` blocks, we don't leverage it:

Screenshot 2020-06-03 at 13 09 02

I gave it a shot and tried if an object-oriented approach can help. Here's what I came up with:

```swift
static func swizzleOnce(using interceptor: @escaping RequestInterceptor) throws {
    if hasSwizzledBefore {
        return
    }
    defer { hasSwizzledBefore = true }

    guard let dataTaskWithURL = DataTaskWithURL(), let dataTaskWithURLRequest = DataTaskWithURLRequest() else {
        throw InternalError(description: "URLSession methods could not be found, thus not swizzled")
    }

    dataTaskWithURL.intercept(using: interceptor, byRedirectingTo: dataTaskWithURLRequest)
    dataTaskWithURLRequest.intercept(using: interceptor)
}
```

Modeling methods as objects is natural when solvign this problem in OOP. It's easy to follow and we keep a full control over what is passed where. It is impossible to use it wrong. See the code [here](https://github.com/DataDog/dd-sdk-ios/blob/ncreated/RUMM-452-propose-swizzling-refactoring/Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift).

Sources/Datadog/Tracing/Utils/MethodSwizzler.swift
```diff
@@ -6,56 +6,126 @@
...
 import Foundation
 
-internal enum SwizzlingError: LocalizedError {
-    case unknown
-    case classIsNotNSObjectSubclass(className: String)
-    case methodNotFound(selector: String, className: String)
-    case methodWasNotSwizzled(selector: String, className: String)
-}
-
 internal class MethodSwizzler {
```
ncreated, line 9: Looking at the `MethodSwizzler` interface, we can see that it's suffering from the lack of OOP:
```swift
class MethodSwizzler {
   func findMethodRecursively(with selector: Selector, in klass: AnyClass) -> FoundMethod?
   func currentImplementation<TypedIMP>(of found: FoundMethod) -> TypedIMP
   func originalImplementation<TypedIMP>(of found: FoundMethod) -> TypedIMP
   func swizzleIfNonSwizzled(foundMethod: FoundMethod, with implementation: @autoclosure () -> IMP) -> Bool
   func set(newIMP: IMP, for found: FoundMethod)
}
```
The `FoundMethod` appears in **all** internal method signatures. Using OOP, we can basically turn the `FoundMethod` into an object and simplify this code a lot. I gave it a try on [this branch](https://github.com/DataDog/dd-sdk-ios/blob/ncreated/RUMM-452-propose-swizzling-refactoring/Sources/Datadog/Tracing/Utils/MethodSwizzler.swift), resulting with half the code removed, no global `implementationCache` to maintain and no threading consideration.

The trick is to turn all above methods into instance methods on `FoundMethod` (I renamed it to `MethodSwizzler` as IMO it sounds more correct for the problem we solve):
```swift
class MethodSwizzler {
    /// The `Method` controlled by this swizzler.
    let method: Method

    static func `for`(selector: Selector, inClass class: AnyClass) -> MethodSwizzler?

    func originalImplementation<TypedIMP>() -> TypedIMP
    func currentImplementation<TypedIMP>() -> TypedIMP
    func swizzle(to newIMP: IMP)
    func swizzleIfNotSwizzled(to newIMP: IMP)
}
```

---

I think this also solves one problem that we will fall into in a future: If we need to swizzle the same method (e.g. `URLSession.dataTaskWithURLRequest`) in two places, e.g. in `TracingInterceptor` and for `RUMInterceptor`, the `implementationCache` hashing function will be colliding, because the `className & methodName` hash will be the same for swizzlings done in `TracingInterceptor` and `RUMInterceptor`.

My proposed way solves this, as the swizzling is identified by `MethodSwizzler` instance.

---

Last, but not least, because we will be swizzling from multiple places in the app (from different features), IMO the most easy and performant way of ensuring thread safety would be to share the same `swizzlingQueue` between all swizzlers (including `URLSessionSwizzler`). This way,
each of them can perform its logic sequentially on the same thread. This should be way more performant than wrapping each `swizzle()` in a critical section (1 critical section per Swizzler vs many critical sections per Swizzler). That's why I removed the `sync {}` method.

WDYT @buranmert ?

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
buranmert: correct 👍 those were leaked during the last refactor yesterday, this looks better; making the change ✅

Sources/Datadog/Tracing/Utils/MethodSwizzler.swift
buranmert: #### `FoundMethod`

you can think of this type as an `IMP` with its `class`, as Cocoa doesn't provide such a type we introduced our own. i don't see any problem with using it in `internal` interface.? 🤔 

#### `impCache` and thread safety

this is incorrect.
you call that OOP but this type modifies the runtime under the hood. the runtime is a shared resource, we always should consider access from multiple threads.
so, we cannot bypass threading issues with the proposed approach.
and we lose the track of swizzlings we made at runtime without `impCache`, that would lead to swizzle the same class (eg: `__NSURLSessionDataTask`) over and over again ❌ 

the problem with the proposed approach is proxying a singleton (ie: runtime) with a non-singleton (ie: `MethodSwizzler`) ❌ 
from my own experience, that usually ends up bad.

#### `FoundMethod` collision

this is incorrect. `impCache` keeps original implementations of methods. if we swizzle the same method in the same class, that doesn't change its value in `impCache` and we still can get its original imp from the cache.
in short, we can swizzle the same methods multiple times ✅ 
plus, we can attempt to undo it if anything ever goes wrong during swizzling 🦺 

**_not in the scope of this PR_** but i think we should minimize the number of swizzling we make. if we know, at runtime `Tracing` and `RUM` both require swizzling IMO the best option would be 
1. swizzling once with multiple `requestInterceptors` or 
2. a special `Tracing&RUM` interceptor

#### Queue vs lock

i first considered a serial private queue as [`Firebase` did](https://github.com/firebase/firebase-ios-sdk/blob/master/GoogleUtilities/MethodSwizzler/GULSwizzler.m#L33)

but that didn't really make sense to me: 
we can do swizzling in a private queue, but that doesn't mean the swizzled method will be called in that private queue. what problem does a private queue solve in this case?
that's why i used `objc_sync_enter/exit` only to restrict access from multiple threads

i'm not sure if there is huge perf diff between those 2, i should benchmark them 🛠️

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
buranmert: `dataTaskWithURL/Request` swizzlers are %~100 self-contained, their types or properties cannot be mixed anymore ✅ 

also i made another change that simplifies the code:
`dataTaskWithURL` does redirection with normal method call now
```swift
let task = impSelf.dataTask(with: interceptionResult.modifiedRequest, completionHandler: modifiedCompletion)
```

as stated in commit message,
🙌 Pros: we don't swallow further swizzling on `dataTaskWithRequest`
😞 Cons: we will execute requestInterceptor twice for the same URL
wdyt?

Sources/Datadog/Tracing/Utils/MethodSwizzler.swift
ncreated: > you call that OOP but this type modifies the runtime under the hood.

The principle of OOP is to use abstraction to hide the implementation details. In case of abstracting the swizzling logic, runtime modification is an implementation detail.

The [InterposeKit](https://github.com/steipete/InterposeKit) is a good example of abstracting the Swizzling by objects:
```swift
let swizzling = try Interpose(URLSession.self) {
    try $0.hook(#selector(URLSession.dataTaskWithURL) { 
        /* provide swizzled implementation */
    }
}

swizzling.revert() // unswizzle
```
I makes the swizzling conceptually easy to digest: _"this object manages one method swizzling"_. Similarly to what was proposed, the underlying swizzlings are modeled as objects: [`Task`](https://github.com/steipete/InterposeKit/blob/f8daba1194b3118b819261374e318c8a958975fa/Sources/InterposeKit/InterposeKit.swift#L95-L106), which validates it as a working solution to the problem.

But I don't insist on going into `InterposeKit's` design (I don't even don't know if it supports what we need). I only say that encapsulation with objects is more elegant and reusable, in contrast to direct manipulation on `IMPs`:
```swift
private static func newIMP(using interceptor: @escaping RequestInterceptor) throws -> IMP {
let newImp_URL = DataTask_URL_Completion.newIMP(
    using: interceptor,
    methodToSwizzle: method_dataTask_URL,
    methodToRedirect: method_dataTask_request
)
swizzler.set(newIMP: newImp_URL, for: method_dataTask_URL)

let newImp_request = DataTask_Request_Completion.newIMP(
    using: interceptor,
    methodToSwizzle: method_dataTask_request
)
swizzler.set(newIMP: newImp_request, for: method_dataTask_request)
```

But the way it was updated in recent commit is good enough for readability :+1:. Sadly, we won't be able to reuse it for RUM.

---

> `impCache` keeps original implementations of methods. if we swizzle the same method in the same class, that doesn't change its value in impCache and we still can get its original imp from the cache.

OK, so this is the part that I didn't get, but this is an implementation detail in any abstraction we take.

> if we know, at runtime Tracing and RUM both require swizzling IMO the best option would be
> 1. swizzling once with multiple requestInterceptors or
> 2. a special Tracing&RUM interceptor

The first option sounds fine. The second one will mix up the responsibility of shared interceptor, making it creak when testing (meaning a bad abstraction).

> but that didn't really make sense to me:
> we can do swizzling in a private queue, but that doesn't mean the swizzled method will be called in that private queue. what problem does a private queue solve in this case?
that's why i used objc_sync_enter/exit only to restrict access from multiple threads

Sure, we don't want to change the execution thread of swizzled methods.

The problem is not only **how to synchronize**, but more importantly, **where to synchronize**.

#### Where to synchronize?

There are two choices:
* in `URLSessionSwizzler`,
* in `MethodSwizzler` - the current solution.

If we synchronize in `URLSessionSwizzler`, the critical section will be executed `N * number of tasks` times and won't grow when the number of `swizzler` access grows:
```
URLSessionSwizzler.installSwizzling() { // <- one critical section access
    swizzleDataTaskWithURLRequest()
    swizzleDataTaskWithURL()
}

// later, each time when a `URLSessionTask` is created:

swizzleDataTaskResume() // <- one critical section access
```

If we synchronize in `MethodSwizzler`, it will be `3 * N * number of tasks` times and will grow by `N` with every new `swizzler` access we need:
```swift
URLSessionSwizzler.installSwizzling() {
    swizzleDataTaskWithURLRequest() {
        // ...
        swizzler.currentImplementation() // <- 1st critical section access
        // ...
        swizzler.set(newIMP: newImp, for: method) // <- 2nd critical section access
        // ...
    }
    swizzleDataTaskWithURL() {
        // ...
        swizzler.currentImplementation() // <- 3rd critical section access
        // ...
        swizzler.set(newIMP: newImp, for: method) // <- 4th critical section access
        // ...
    }
}

// later, each time when a `URLSessionTask` is created:

swizzleDataTaskResume() {
    // ...
    swizzler.findMethodRecursively() // <- 1st critical section access
    // ...
    swizzler.swizzleIfNonSwizzled // <- 2nd critical section access
    // ...
    swizzler.currentImplementation(of: foundMethod) // <- 3rd critical section access
    // ...
}
```

Although I skipped it, the impact of initial swizzlings is not negligible, as we may have many of them done in the RUM product.

#### How to synchronize?

There are two choices:
* sync,
* async.

If we do `sync`, we will block the application thread (almost always the main thread) for the time of critical section access. Given that we currently do 3 accesses per task, this might result in a visible FPS drop.

If we do `async`, we won't be dropping any application frames, as swizzling will be non-blocking. The only problem with `async` which I see is the `task.resume()` which might be called before we swizzle. That can be easily overcome by checking the `task.status` before swizzling.

#### Picking the synchronization tool

Only given "where" and "how" we can discuss the implementation of critical section. 

Current approach with `objc_sync_enter(self)` is **blocking**, so the main thread will be impacted. Yes, it is more performant than `DispatchQueue` according to [some benchmarks](https://www.cocoawithlove.com/blog/2016/06/02/threads-and-mutexes.html), but it doesn't give us the off-main-thread execution.

That's why IMO using a `DispatchQueue` is more relevant to our problem: conceptually easier, quite performant and makes async execution possible.

buranmert: in fact, we can reduce the number of critical region access with block-based approach 👍 
```swift
func swizzle<TypedIMP>(method: FoundMethod, impProvider: (currentImp: TypedIMP) -> IMP) {
  sync {
    ...
  }
}
```
that way `MethodSwizzler` provides `currentImplementation` within `impProvider` block
i couldn't do that before because we needed to get `currentImp` separately in order to do redirection in `dataTaskWithURL`
i'll make that change now ✅ 

Firebase is swizzling in its own private queue in `sync`, even they don't seem to cause perf issues (benchmarking should help anyway though)
`async` cannot be an option, the app needs to wait while the runtime is being changed 🤷‍♂️ 
`objc_sync_enter/exit` still looks like the best option to me ✅ 

If we model a single swizzling as an object:
1. what does it give us?
2. this is incorrect abstraction
let's say we made `swizzle1` and then `swizzle2`; if we now `swizzle1.revert` that will also revert `swizzle2`

p.s: this thread is too long to follow now, can we start a new one if we still have questions?

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
ncreated: > we will execute requestInterceptor twice for the same URL

Sounds like this will create two spans for single request, which is not acceptable. Even if not, "this block is called twice" is definitely not what we want.

Sources/Datadog/Tracing/Utils/MethodSwizzler.swift
ncreated: > If we model a single swizzling as an object:
> 1. what does it give us?

First and crucial, it allows us to get rid of threading synchronization for `IMP` storage at all (!), because the `originalIMP` and `currentIMP` will be bound to the `MethodSwizzler` object, instead of being stored in a **global, shared** `implementationCache: [FoundMethod: IMP]` which requires synchronization (as "global + shared resources" usually do).

Second, it is much cleaner, as it enables the _"Is this method already swizzled?"_ reasoning on the right level of abstraction: inside the `URLSessionSwizzler`. This vs _"I will swizzle this method, but maybe it won't require swizzling and something internally will handle it"_.

Third, this is more performant, not only because it doesn't force the thread synchronization for **all swizzlings**, but also because it gives us the freedom to selectively decide which swizzlings we want to synchronize. Considering two swizzlers:
* `UIViewControllerSwizzler`
* `URLSessionSwizzler`

**Why would we need to synchronize the `UIViewController.viewWillLoad` swizzlings?** We can ask the users to call `Datadog.autoinstrumentViewControllers()` early on the main thread and this will be benefitial for both sides: zero thread-switch penalty in the running app and much easier threading model in the SDK code (well: single threading model).

**Why do we need to synchronize `dataTaskWithURL` and `dataTaskWithURLRequest` swizzlings in `URLSessionSwizzler`?** Again, asking the user to call `Datadog.autoinstrumentURLSession()` early on the main thread is a win-win. Things are different if it comes to `task.resume()` swizzling, as that one is applied later in the running app, so that one requires synchronization (and even this can be solved - what I describe later).

So, using object-oriented approach:
* is not forcing any thread-switch for **every swizzling** (no FPS drop),
* allows for selectively synchronizing **some swizzlings** if necessary (and so far we only have one).

---

Also, here's my proposal to do `task.resume()` swizzling on the main thread. Using the object-oriented approach, it might look this:
```swift
class URLSessionSwizzler {
    /// Object applying `URLSession.dataTaskWithURL()` swizzling.
    private let dataTaskWithURLSwizzling: DataTaskWithURLSwizzling
    
    /// Object applying `URLSession.dataTaskWithURLRequest()` swizzling.
    private let dataTaskWithRequestSwizzling: DataTaskWithURLRequestSwizzling

    /// Object applying `URLSessionTask.resume()` swizzling.
    private let dataTaskResumeSwizzling: DataTaskResumeSwizzling

    init() {
        do {
            // check if we can swizzle:
            self.dataTaskWithURLSwizzling = try DataTaskWithURLSwizzling()
            self.dataTaskWithRequestSwizzling = try DataTaskWithURLRequestSwizzling()
            self.dataTaskResumeSwizzling = try DataTaskResumeSwizzling()

            // swizzle:
            self.dataTaskWithURLSwizzling.apply()
            self.dataTaskWithRequestSwizzling.apply()
            self.dataTaskResumeSwizzling.apply()
        } catch {
            print("Sorry, we cannot swizzle `URLSession` because: \(error)")
        }
    }

    /// Applies the `URLSession.dataTaskWithURL()` swizzling.
    private struct DataTaskWithURLSwizzling {
        private let swizzling: MethodSwizzling

        init() throws {
            try MethodSwizzling(
                class: URLSession.self,
                selector: #selector(URLSession.dataTask(with:completionHandler:) as ...
            )
        }

        func apply() {
            swizzling.addImplementation { ... in
                // ...
            }
        }
    }

    /// Applies the `URLSession.dataTaskWithURLRequest()` swizzling.
    private struct DataTaskWithURLRequestSwizzling {
        private let swizzling: MethodSwizzling

        init() throws {
            try MethodSwizzling(
                class: URLSession.self,
                selector: #selector(URLSession.dataTask(with:completionHandler:) as ...
            )
        }

        func apply() {
            swizzling.addImplementation { ... in
                // ...
            }
        }
    }

    /// Applies the `URLSessionTask.resume()` swizzling.
    private struct DataTaskResumeSwizzling {
        private let swizzling: MethodSwizzling

        init() throws {
            let dummyURLSession = URLSession()
            let task = dummyURLSession.dataTaskWithRequests(...)

            self.swizzling = try MethodSwizzling(
                class: object_getClass(task),
                selector: #selector(URLSessionTask.resume as ...
            )

            task.cancel()
        }

        func apply() {
            swizzling.addImplementation { ... in
                // ...
            }
        }
    }
}
```
```swift
class UIViewControllerSwizzling {
    /// Object applying `UIViewController.viewWillAppear()` swizzling.
    private let viewWillAppearSwizzling: ViewWillAppearSwizzling
    
    /// Object applying `UIViewController.viewDidDisappear()` swizzling.
    private let viewDidDisappearSwizzling: ViewDidDisappearSwizzling

    init() {
        // ...
    }

    /// Applies the `UIViewController.viewWillAppear()` swizzling.
    private struct ViewWillAppearSwizzling {
        // ...
    }    

    /// Applies the `UIViewController.viewDidDisappear()` swizzling.
    private struct ViewDidDisappearSwizzling {
        // ...
    }    
}
```

> 2. this is incorrect abstraction let's say we made swizzle1 and then swizzle2; if we now swizzle1.revert that will also revert swizzle2

First, we don't need to unswizzle our swizzlings in production code. And punishing the customer apps for unswizzling need in unit tests is not a way to go.

Second, all this can be done in object-oriented way. I don't see why `MethodSwizzler` object couldn't track all swizzlings that were done to a given selector? The same way as we track them in `implementationCache: [FoundMethod: IMP]`. The difference is crucial: it will track it locally and only when necessary, instead of using global, shared and synchronized cache. This abstraction is not making it impossible in any way. Again, [`InterposeKit`](https://github.com/steipete/InterposeKit) uses something similar, with success.

---

Wrapping up:
* object-oriented abstraction lets us pick the execution thread per swizzling,
* global-cache abstraction forces the same execution thread for all swizzlings.

And less key factor: the OO abstraction I proposed is half the number of code less than we need for global-cache.

Sources/Datadog/Tracing/Utils/MethodSwizzler.swift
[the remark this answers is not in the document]
buranmert: update: we don't wrap it as it's inaccessible from outside

Sources/Datadog/Tracing/Utils/MethodSwizzler.swift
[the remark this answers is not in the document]
buranmert: update: this check is removed as it was not needed

Sources/Datadog/Tracing/Utils/MethodSwizzler.swift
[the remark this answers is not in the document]
buranmert: update: we leave this method with `unsafe_` prefix

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
buranmert: i'm a bit confused, your quote my comment with adding `NOT` in the most important part :) but i'm assuming we are clear on this
otherwise, feel free to unresolve 👍

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
buranmert: update: method discovery (note: the only failure case is unfound method) and swizzling are separate now ✅

Tests/DatadogTests/Datadog/Tracing/Utils/URLSessionSwizzlerTests.swift
buranmert: i'm assuming this discussion is resolved by now ✅

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
buranmert: update: this discussion was resolved in Slack. we never call `interceptor` twice ✅

Sources/Datadog/Tracing/Utils/MethodSwizzler.swift
buranmert: summary: this discussion went quite long because we were not talking about the same thing.

we have 2 main use cases: 
1. static: swizzling `URLSession` only once
2. dynamic: swizzling every `URLSessionTask` created (if needed)
(i made up the names "static" and "dynamic", i hope they make sense)

this PR covers both of them, static and dynamic. 
@ncreated argues that dynamic swizzling can be covered per-case basis as a special case.

i assume this is resolved by now

Tests/DatadogTests/Datadog/Tracing/Utils/URLSessionSwizzlerTests.swift
```diff
@@ -7,107 +7,233 @@
...
+    }
+}
 
-        let url = URL(string: "http://foo.bar")!
-        let request = URLRequest(url: url)
-        let task = session.dataTask(with: request)
+// MARK: - 3rd party swizzlers
 
-        XCTAssertEqual(task.originalRequest?.url, url)
+class ThirdPartySwizzlingTests: XCTestCase {
```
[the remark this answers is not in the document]
buranmert: after discussing in Slack, we decided that 3rd party swizzler tests are not needed
resolving the discussion ✅

Tests/DatadogTests/Datadog/Tracing/Utils/MethodSwizzlerTests.swift
```diff
@@ -13,146 +13,160 @@ private class EmptySubclass: BaseClass { }
...
-    func swizzleMethodToSwizzle() {
-        typealias TypedIMP = @convention(c) (AnyObject, Selector) -> String
-        typealias TypedBlockIMP = @convention(block) (AnyObject) -> String
+        // before
+        XCTAssertNotEqual(obj.perform(selToSwizzle)?.takeUnretainedValue() as? String, String.mockAny())
+        // swizzle
+        let foundMethod = try swizzler.findMethod(with: selToSwizzle, in: BaseClass.self)
+        swizzler.swizzle(foundMethod, impSignature: IMP.self) { _ in newIMPReturnString }
+        // after
+        XCTAssertEqual(obj.perform(selToSwizzle)?.takeUnretainedValue() as? String, String.mockAny())
```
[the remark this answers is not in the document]
buranmert: return value changed to `originalImp + String.mockAny()`, that should prove that the original imp is being invoked correctly ✅

Tests/DatadogTests/Datadog/Tracing/Utils/MethodSwizzlerTests.swift
```diff
@@ -13,146 +13,160 @@ private class EmptySubclass: BaseClass { }
...
-        }
-        let newImp = imp_implementationWithBlock(newBlockImp)
+        let afterOrigIMP: IMP = swizzler.originalImplementation(of: foundMethod)
 
         let obj = BaseClass()
-
-        XCTAssertNoThrow(try swizzler.swizzle(selector: sel, in: BaseClass.self, with: newImp))
-        XCTAssertEqual(obj.perform(sel)?.takeUnretainedValue() as? String, String.mockAny())
+        XCTAssertEqual(obj.perform(selToSwizzle)?.takeUnretainedValue() as? String, secondSwizzlingReturnValue)
+        XCTAssertEqual(beforeOrigIMP, afterOrigIMP)
```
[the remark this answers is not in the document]
buranmert: this test case changed to test the immutability of `originalImp` ✅

Sources/Datadog/Tracing/Utils/URLSessionSwizzler.swift
```diff
@@ -6,102 +6,157 @@
...
+        }
+    }
+}
 
-        let typedOriginalImp: TypedIMP = try swizzler.currentImplementation(of: sel, in: subjectClass)
+private extension URLSession {
+    /// This method is used in swizzled dataTaskWithURL implementation
+    /// We create URLRequest from URL and pass it to interceptor
+    func urlRequest(with url: URL) -> URLRequest {
+        return URLRequest(url: url, cachePolicy: configuration.requestCachePolicy, timeoutInterval: configuration.timeoutIntervalForRequest)
```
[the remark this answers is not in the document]
ncreated: What about starting with the safest possible solution:
* 1/ `session.dataTask(with: request, completion: {})` → send span + propagate trace
* 2/ `session.dataTask(with: url, completion: {})` → send span

With extension to:
* 3/ `session.dataTask(with: request)` → redirect to 1, so: send span + propagate trace
* 4/ `session.dataTask(with: url)` → redirect to 2, so: send span

This is the most safe & valuable implementation I can come up with. WDYT @buranmert ?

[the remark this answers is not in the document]
buranmert: methods without completion handler (3&4) use `session.delegate` for events, changing their execution path will break their behavior, so i'd say let's leave them out of this.

what would be confusing, distributed tracing for requests but not for urls.
some SDKs provide `DistributedTracing` as a configurable feature.

if we add `DistributedTracing` to `TracingConfiguration`, 
then we can provide the same behavior (span + trace) for both methods 
with an emergency brake ("if it breaks anything, disable `DistributedTracing`")
wdyt @ncreated ?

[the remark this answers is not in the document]
ncreated: > methods without completion handler (3&4) use session.delegate for events, changing their execution path will break their behavior, so i'd say let's leave them out of this.

Ah I see, totally right 👌.

> if we add DistributedTracing to TracingConfiguration,
then we can provide the same behavior (span + trace) for both methods
with an emergency brake ("if it breaks anything, disable DistributedTracing")
wdyt @ncreated ?

As concluded on Slack with product - let's start simple with
- `session.dataTask(with: request, completion: {})` → send span + propagate trace
- `session.dataTask(with: url, completion: {})` → send span

and we'll consider the propagation for plain `URLs` later, if requested. This requires a round of simplification in `URLSessionSwizzler` and its tests 🚀.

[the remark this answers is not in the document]
buranmert: 1. `dataTaskWithURL` doesn't redirect to `dataTaskWithRequest` anymore
2. `RequestInterceptor` changed to `(URL?) -> (TaskObserver, HTTPHeaders)?`
    * so that we still have only one way to intercept requests

pushed ✅

Sources/Datadog/Core/AutoInstrumentation/URLSessionSwizzler.swift
```diff
@@ -0,0 +1,184 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+/// Block type to hook into URLSession methods:
+/// dataTaskWithURL:completion: and dataTaskWithRequest:completion:
+/// Takes original URLRequest and returns modified URLRequest with TaskObserver
```
ncreated, line 11: This comment is now invalid.

Sources/Datadog/Core/AutoInstrumentation/URLSessionSwizzler.swift
```diff
@@ -0,0 +1,184 @@
...
+                onlyIfNonSwizzled: true
+            )
+        }
+    }
+}
+
+private extension URLRequest {
+    func merging(_ httpHeaders: [String: String]) -> URLRequest {
+        let mergedHeaders: [String: String]
+        if let currentHeaders = allHTTPHeaderFields {
```
ncreated, line 160: We should rather use the precise [`addValue(_:forHTTPHeaderField:)`](https://developer.apple.com/documentation/foundation/urlrequest/2011522-addvalue) from `URLRequest`. It does [internal synchronization](https://github.com/apple/swift-corelibs-foundation/blob/1c1e52abee4efc2e24def6ed117751909bbb32cb/Sources/FoundationNetworking/URLRequest.swift#L165-L173) of header fields dictionary, so it's thread safe. I think this is to not conflict with internal processing of request by `URLSession`.

PR #120, opened 2020-06-01
RUMM-401 Add error information to span

ncreated:
### What and why?

🚚 With this PR, it is possible to add error information to the `Span`, which leads to dedicated appearance on [app.datadoghq.com](https://app.datadoghq.com).

Open Tracing defines two distinct groups of attributes regarding error:
* [`error: Bool` span tag](https://github.com/opentracing/specification/blob/master/semantic_conventions.md#span-tags-table),
* [`error.kind`, `error.object`, `event`, `message` and `stack` log fields](https://github.com/opentracing/specification/blob/master/semantic_conventions.md#log-fields-table)

and lets tracer authors decide on how to handle them in a given language.

### How?

No new API is introduced. Instead, the implementation handles necessary OT tags and log fields.

#### Setting `error` tag
```swift
span.setTag("error": true)
```
This will result with only error icon and label being added to the `Span`. No details are given:

Screenshot 2020-06-01 at 16 53 14

#### Sending `event: "error"` log
```swift
span.log(
    fields: [
        "event": "error",
        "error.kind": "Simulated error",
        "message": "Describe what happened",
        "stack": "Foo.swift:42",
    ]
)
```
This will not only send the `error` log using `Logging` feature (and link it to the `Trace`), but will also leverage custom rendering in our platform, so the error information is nicely digested:

Screenshot 2020-06-01 at 17 06 43

---

Supporting both required a new piece in the `Tracing` feature's architecture. Now, `SpanBuilder` uses `SpanTagsReducer` to reduce `span.tags` and `span.logAttributes` received from the user. If interesting fields are found, custom values are extracted by the chain of reducers.

This is a simplified version of the [`AbstractDecorator` concept](https://github.com/DataDog/dd-sdk-android/blob/master/dd-sdk-android/src/main/java/datadog/opentracing/decorators/DDDecoratorsFactory.java) from the Android SDK. The `SpanTagsReducer` will be used to process other attributes which are given through span tags, but require encoding other than `span.meta.*` (i.e. `span.resource`).

### Review checklist


Sources/Datadog/Tracing/Span/SpanTagsReducer.swift
```diff
@@ -0,0 +1,69 @@
...
+            }
+        }
+    }
+
+    private static func extractErrorFromTags(_ reducer: inout SpanTagsReducer) {
+        if reducer.spanTags[OpenTracingTagKeys.error] as? Bool == true {
+            reducer.extractedIsError = true
+        }
+    }
+}
```
buranmert, line 69: this file is quite hard to understand IMO: 
• where do we mutate? 
• why do we reduce the reducer `reducers.forEach { reduce in reduce(&reducer) }`?
• we pass `spanTags` to reducer and then we use `reducer.spanTags`, it's not that clear what's changed?
• is the order of `reducers` important?

if i understood well, we take `logFields` and construct a special set of `spanTags` and we extract various info from `spanTags`
pseudo-code:
```swift
let logFields = someLogFields()
let spanTags = someSpanTags()

let datadogEncodedSpanTags = spanTags.merge(with: logsFields.datadogSpanTags)
let hasError = datadogEncodedSpanTags.hasError
let resourceName = datadogEncodedSpanTags.extractResourceName() // mutating func
```

above, `datadogSpanTags`, `hasError` and `extractResourceName()` can be implemented in extensions for `spanTag` and `logFields` types (different kinds of `Dict`s)

what do you think?

ncreated: Yes, this piece of logic is tricky. Generally speaking, this problem can be simplified to a **pure function**:
```swift
func computeStateBasedOn(dictionaryA: [String: Codable], dictionaryB: [String: Codable]) -> State
```
We take the `logFields` and `spanTags` (dictionaries `A` and `B`)  and compute a `State` which describes following:
* `spanTags` (may be different than the input `spanTags`),
* `isError` flag,
* `resourceName` value,
* other values in the future (likely: HTTP status codes).

And what we do now, conforms to a pure function signature:
```swift
let state = SpanTagsReducer.reduce(spanTags: ddspan.tags, logFields: ddspan.logFields)
_ = state.spanTags
_ = state.extractedIsError
_ = state.extractedResourceName
```

We can't merge the dictionaries initially, as proposed:
```swift
let datadogEncodedSpanTags = spanTags.merge(with: logsFields.datadogSpanTags)
```
because it is important from which dictionary, the entry comes from (1) and both dictionaries may contain the same keys (2). In both cases, we will lost the information.

Also, we can't provide mutating getters, as proposed:
```swift
let resourceName = datadogEncodedSpanTags.extractResourceName() // mutating func
```
because state computation changes the input dictionaries, so this will give a different results depending on the order of method calling:
```swift
var mergedDictionaries = spanTags.merge(with: logsFields.datadogSpanTags)

// this:
let datadogEncodedSpanTags = mergedDictionaries
let hasError = mergedDictionaries.hasError

// will not mean the same as this:
let hasError = mergedDictionaries.hasError
let datadogEncodedSpanTags = mergedDictionaries
```

And lastly, IMO the `.isError` extension to a `[String: Codable]` type is a very bad practise - it will add the `.isError` interface on all other dictionaries of this kind (even if it's declared privately in the same file, it's kind of abuse IMO).

buranmert: the usage of this type is 👍 
however, its implementation seems to be over-abstracted for now
we can start simple, without `reducers` array and `inout` parameters
then we can progress when needed

ncreated: Fixed in https://github.com/DataDog/dd-sdk-ios/pull/121/commits/b49d1b1b98ea135bf26ec4e9b6ae447dfe3b4a02, #121 to avoid rebasing 🙌.

PR #122, opened 2020-06-02
RUMM-341 Fix linter not failing on missing license header

ncreated:
### What and why?

⚙️ The `run_linter` CI job was not failing when license header was not found.

### How?

The `exit 1` was not properly evaluated in `check-license.sh` script. I made the bash script more straightforward by removing the `while` loop nested with pipe operator `|`.

### Review checklist

~~- [ ] Feature or bugfix MUST have appropriate tests (unit, integration)~~

### DIFF
--- a/tools/license/check-license.sh
+++ b/tools/license/check-license.sh
@@ -8,7 +8,7 @@ IFS=$'\n'
 
 function files {
 	find -E . \
-		-iregex '.*\.(swift|h|m)$' \
+		-iregex '.*\.(swift|h|m|py)$' \
 		-type f \( ! -name "Package.swift" \) \
 		-not -path "*/.build/*" \
 		-not -path "*Pods*" \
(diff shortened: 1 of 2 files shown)


tools/license/check-license.sh
```diff
@@ -4,21 +4,30 @@ if [ ! -f "Package.swift" ]; then
...
+		-not -path "*Carthage/Build/*" \
+		-not -path "*Carthage/Checkouts/*"
 }
 
-files | while IFS= read -r -d '' file; do
-  if ! grep -q "Apache License Version 2.0" "$file"
-  then
-  	echo "No license in $file"
-  	exit 1
-  fi
```
buranmert, line 21: `then` was never executed somehow because of `while` above?

ncreated: It was executed, but the `exit 1` was not exiting the entire process. I'm not the bash expert, but I think the `exit 1` was only exiting the `while's` scope as there's a pipe `|` operator used, which seems to create subprocess 🤷. Anyway, now it works and we can get feedback on multiple broken files, not only the first one.

PR #133, opened 2020-06-12
RUMM-506 Prevent malforming payload due to I/O exception 

ncreated:
### What and why?

🐛  Some payloads send from SDK were malformed.

The reason was not performing writes atomically:
```swift
// pseudo-code

let separatorData = ",".utf8Data
try writer.write(separatorData)
try writer.write(logData)
```
In rare cases, when the `I/O` exception was thrown for the second write, it was leaving the file's content malformed.

### How?

The fix is to perform atomic writes:
```swift
// pseudo-code

let logWithSeparatorData = ",".utf8Data + logData
try writer.write(logWithSeparatorData)
```

The red unit test for payload malformation was added, then it was made green with a fix.

### Later improvements

The `File` interface is not resistant for this sort of bugs, as it doesn't perform transactional write even if it looks like:
```swift
// FileTests.swift

try file.append { write in
    try write(Data([0x42, 0x42, 0x42, 0x42, 0x42])) // 5 bytes
    try write(Data([0x41, 0x41, 0x41, 0x41, 0x41])) // 5 bytes
}
```

As this requires additional refactoring, to not impact the hotfix scope, I will open a separate PR changing the `WritableFile` interface from:
```swift
func append(transaction: ((Data) throws -> Void) throws -> Void) throws
```
to simply:
```swift
func append(data Data) throws
```

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/Persistence/FileWriter.swift
+++ b/Sources/Datadog/Core/Persistence/FileWriter.swift
@@ -42,9 +42,9 @@ internal final class FileWriter {
                     try write(data)
                 }
             } else {
+                let atomicData = commaSeparatorData + data
                 try file.append { write in
-                    try write(commaSeparatorData)
-                    try write(data)
+                    try write(atomicData)
                 }
             }
         } catch {
--- a/DatadogSDK.podspec
+++ b/DatadogSDK.podspec
@@ -1,7 +1,7 @@
 Pod::Spec.new do |s|
   s.name         = "DatadogSDK"
   s.module_name  = "Datadog"
-  s.version      = "1.2.1"
+  s.version      = "1.2.2"
   s.summary      = "Official Datadog Swift SDK for iOS."
   
   s.homepage     = "https://www.datadoghq.com"
--- a/DatadogSDKObjc.podspec
+++ b/DatadogSDKObjc.podspec
@@ -1,7 +1,7 @@
 Pod::Spec.new do |s|
   s.name         = "DatadogSDKObjc"
   s.module_name  = "DatadogObjc"
-  s.version      = "1.2.1"
+  s.version      = "1.2.2"
   s.summary      = "Official Datadog Objective-C SDK for iOS."
   
   s.homepage     = "https://www.datadoghq.com"
--- a/Sources/Datadog/Datadog.swift
+++ b/Sources/Datadog/Datadog.swift
@@ -8,7 +8,7 @@ import Foundation
 
 /// SDK version associated with logs.
 /// Should be synced with SDK releases.
-internal let sdkVersion = "1.2.1"
+internal let sdkVersion = "1.2.2"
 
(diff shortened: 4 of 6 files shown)


Sources/Datadog/Core/Persistence/FileWriter.swift
```diff
@@ -42,9 +42,9 @@ internal final class FileWriter {
                     try write(data)
                 }
             } else {
+                let atomicData = commaSeparatorData + data
                 try file.append { write in
-                    try write(commaSeparatorData)
-                    try write(data)
+                    try write(atomicData)
                 }
```
buranmert, line 48: we can also reduce `try write()` calls
```swift
let rawData = try jsonEncoder.encode(value)
let data: Data = try file.size() == 0 ? (rawData) : (commaSeparator + data)
try file.append { write in try write(data) }
```
wdyt?

ncreated: Good idea 👍. However, to keep the hotfix scope as small as possible, let's do it in a separate PR to `master` (within the "Later improvements" I list in the PR description).

buranmert:
@ncreated i just noticed that in some cases the log has `\u0000` instead of extra `,`.
i guess that's because the second `try write(data)` throws and leaves `\u0000` there, can this be true? 
if so, how can we make sure that we don't leave corrupt chars in the file?

buranmert:
note: we still can't figure out why `\u0000` appears, we couldn't reproduce it and it seems very rare (possibly a few in a million logs)

we will ship this hotfix version and keep monitoring the issue

ncreated:
I added the test to prevent also malformations in the middle of the file: https://github.com/DataDog/dd-sdk-ios/pull/133/commits/792c74cd886372c3d002305a2eb970c218bbac34 This is to additionally check if the `0x00` byte is not inserted.

buranmert:
i suspect that when `FileHandler.write(data)` raises a certain kind of exception, it somehow changes `offsetInFile` without adding non-null bytes.

can we add something like this?
```swift
func append(transaction: ((Data) throws -> Void) throws -> Void) throws {
  let fileHandle = try FileHandle(forWritingTo: url)
  let initialOffset = fileHandle.offsetInFile
  defer {
    let finalOffset = fileHandle.offsetInFile
    if errorThrown && (initialOffset != finalOffset) {
      userLogger.error("Houston, we have a problem...")
    }
    fileHandle.closeFile()
  }
  ...
}       
```

ncreated:
> can we add something like this?

@buranmert even if it works, this does neither fix the problem nor empower us with reporting. I think it's better to iterate on this issue and first see if we'll be receiving malformed logs after applying this hotfix. WDYT?

buranmert:
yes, fixing null char issue doesn't seem trivial; let's go with this hotfix and keep monitoring 👍

PR #134, opened 2020-06-12
RUMM-468 Add public api to enable / disable features

ncreated:
### What and why?

📦 This PR adds two public methods to toggle logging and tracing features:
```swift
/// Enables or disables the logging feature.
///
/// This option is meant to opt-out from using Datadog Logging entirely, no matter of your environment or build configuration. If you need to
/// disable logging only for certain scenarios (e.g. in `DEBUG` build configuration), use `sendLogsToDatadog(false)` available
/// on `Logger.Builder`.
///
/// If `enableLogging(false)` is set, the SDK won't instantiate underlying resources required for
/// running the logging feature. This will give you additional performance optimization if you only use tracing, but not logging.
///
/// **NOTE**: If you use logging for tracing (`span.log(fields:)`) keep the logging feature enabled. Otherwise the logs
/// you send for `span` objects won't be delivered to Datadog.
///
/// - Parameter enabled: `true` by default
public func enableLogging(_ enabled: Bool) -> Builder

/// Enables or disables the tracing feature.
///
/// This option is meant to opt-out from using Datadog Tracing entirely, no matter of your environment or build configuration. If you need to
/// disable tracing only for certain scenarios (e.g. in `DEBUG` build configuration), do not set `OpenTracing.Global.sharedTracer` to `DDTracer`,
/// and your app will be using the no-op tracer instance provided by `OpenTracing`.
///
/// If `enableTracing(false)` is set, the SDK won't instantiate underlying resources required for
/// running the tracing feature. This will give you additional performance optimization if you only use logging, but not tracing.
///
/// - Parameter enabled: `true` by default
public func enableTracing(_ enabled: Bool) -> Builder
```

### How?

Internally, the feature status is determined by its `.instance` optional, so:
```swift
let loggingEnabled = LoggingFeature.instance != nil
let tracingEnabled = TracingFeature.instance != nil
```
It's simple, and powerful, as we expect non-optional feature instances in appropriate places, making the error handling and consistency consideration super simple, e.g. this reads as `TracingFeature` depends optionally on `LoggingFeature`:
```swift
internal final class TracingFeature {
    let loggingFeatureAdapter: LoggingForTracingAdapter?
}
```
but this:
```swift
internal struct LoggingForTracingAdapter {
    private let loggingFeature: LoggingFeature
}
```
means that `LoggingForTracingAdapter` requires the logging feature to be enabled.

All possible API misuses were covered with user-facing error messages:
* if `Logger` is instantiated with logging feature disabled;
* if `DDTracer` is instantiated with tracing feature disabled;
* if `span.log(fields:)` is used with logging feature disabled.

Also, because playing with global components (`.instance`, `consolePrint`, `userLogger`, ...) might be error prone in unit tests, I enabled "Randomize execution order" for tests target. This way, if we introduce flakiness due to global state inconsistency, we will be able to spot it early and fix.

### Review checklist

### DIFF
--- a/Sources/Datadog/Datadog.swift
+++ b/Sources/Datadog/Datadog.swift
@@ -110,44 +110,51 @@ public class Datadog {
         let httpClient = HTTPClient()
         let mobileDevice = MobileDevice.current
 
-        let logging = LoggingFeature(
-            directory: try obtainLoggingFeatureDirectory(),
-            configuration: validConfiguration,
-            performance: performance,
-            mobileDevice: mobileDevice,
-            httpClient: httpClient,
-            logsUploadURLProvider: UploadURLProvider(
-                urlWithClientToken: validConfiguration.logsUploadURLWithClientToken,
-                queryItemProviders: [
-                    .ddsource(),
-                    .batchTime(using: dateProvider)
-                ]
-            ),
-            dateProvider: dateProvider,
-            userInfoProvider: userInfoProvider,
-            networkConnectionInfoProvider: networkConnectionInfoProvider,
-            carrierInfoProvider: carrierInfoProvider
-        )
+        var logging: LoggingFeature?
+        var tracing: TracingFeature?
 
-        let tracing = TracingFeature(
-            directory: try obtainTracingFeatureDirectory(),
-            configuration: validConfiguration,
-            performance: performance,
-            loggingFeatureAdapter: LoggingForTracingAdapter(loggingFeature: logging),
-            mobileDevice: mobileDevice,
-            httpClient: httpClient,
-            tracesUploadURLProvider: UploadURLProvider(
-                urlWithClientToken: validConfiguration.tracesUploadURLWithClientToken,
-                queryItemProviders: [
-                    .batchTime(using: dateProvider)
-                ]
-            ),
-            dateProvider: dateProvider,
-            tracingUUIDGenerator: DefaultTracingUUIDGenerator(),
-            userInfoProvider: userInfoProvider,
-            networkConnectionInfoProvider: networkConnectionInfoProvider,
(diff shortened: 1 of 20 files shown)


Sources/Datadog/Datadog.swift
```diff
@@ -92,49 +92,69 @@ public class Datadog {
...
+                carrierInfoProvider: carrierInfoProvider
+            )
+        }
+
+        if configuration.tracingEnabled {
+            tracing = TracingFeature(
+                directory: try obtainTracingFeatureDirectory(),
+                configuration: validConfiguration,
+                performance: performance,
+                loggingFeatureAdapter: logging.flatMap { LoggingForTracingAdapter(loggingFeature: $0) },
```
buranmert, line 142: do we need `flatMap` here?

ncreated: Yup, the `logging` is an optional `LoggingFeature?`, so it's either `flatMap {}` or we'd have to wrap the entire 18-lines `TracingFeature()` setup within a `if let`, which is IMO less readable.

buranmert: if you are happy with `flatMap` we can keep it, otherwise we can do this instead:
```swift
var adapter: Adapter? = nil
if let someLogging = logging {
  adapter = Adapter(someLogging)
}
...
loggingFeatureAdapter: adapter
...
```

ncreated: IMO it's fine with, we don't get `flatMap` wrong:
> Evaluates the given closure when this `Optional` instance is not `nil`, passing the unwrapped value as a parameter.

https://developer.apple.com/documentation/swift/optional/1540500-flatmap

PR #135, opened 2020-06-15
RUMM-506 Update `File` interface to promote atomic writes

ncreated:
### What and why?

⚙️ The cause of `1.2.2` hotfix was a non-atomic write performed on the `File` object. It was fixed in #133. This PR brings additional improvement to the `File` interface to make it cleaner.

### How?

The `File` interface was simplified:
```diff
-    func append(transaction: ((Data) throws -> Void) throws -> Void) throws {
+    func append(data: Data) throws {
```
so it doesn't erroneously mention the "transaction" word.

### Review checklist


ncreated:
> those FileHandle methods (read/write) are deprecated in iOS 13
and new methods are throwing 🙌 we can start migrating to new methods*, wdyt?

Good catch 👌. Definitely, we should use them. And Tracing beta is a good moment to roll-it-out safely.

I updated the PR to use throwing interface of `FileHandler`. I had to make the `FileWriterTests` not depend on the `objcExceptionHandler` anymore. To hide `objcExceptionHandler` detail, I forge the I/O exceptions by randomly making the file read-only on secondary thread.

The `testWhenIOExceptionsHappenRandomly_theFileIsNeverMalformed()` passes fine with the right coverage on `iOS 12.4` and `13.5`.

Sources/Datadog/Core/Persistence/Files/File.swift
```diff
@@ -43,23 +43,27 @@ internal struct File: WritableFile, ReadableFile {
...
-                fileHandle.write(data)
+                fileHandle.seekToEndOfFile()
             }
-        }
 
-        try transaction { chunkOfData in
-            try appendData(chunkOfData)
+            try objcExceptionHandler.rethrowToSwift {
+                fileHandle.write(data)
+            }
```
buranmert, line 66: we can put `fileHandle.seekToEndOfFile()` and `write(data)` into the same `rethrowToSwift` block, wdyt?

Sources/Datadog/Core/Persistence/Files/File.swift
```diff
@@ -43,23 +43,27 @@ internal struct File: WritableFile, ReadableFile {
...
         let fileHandle = try FileHandle(forWritingTo: url)
-        defer { fileHandle.closeFile() }
 
-        try objcExceptionHandler.rethrowToSwift {
-            fileHandle.seekToEndOfFile()
-        }
+        if #available(iOS 13.0, *) {
+            try fileHandle.seekToEnd()
+            try fileHandle.write(contentsOf: data)
+            try fileHandle.close()
```
buranmert, line 52: i guess we still need to `defer { try file.close() }`, no?

ncreated: Hmm, inside `defer {}` we can't `try`. The trade-off will be `defer { try? file.close }`, but I think it's still better so will do it 👌.

Sources/Datadog/Core/Persistence/Files/File.swift
ncreated: 👍

PR #137, opened 2020-06-16
RUMM-500 Unify feature groups in `xcodeproj`

ncreated:
### What and why?

📦 This PR updates files and folders layout in `Datadog.xcodeproj`. After we added second feature, it's easy to spot similarities between Logging and Tracing and move some pieces of code across files and folders, so it is makes single convention when adding next features.

### How?

Few things are done:
* `Logs` is renamed to `Logging`, so now we name folders by feature: `Logging`, `Tracing`, ...
* various mocks shredded between `*Mocks` files are now clearly grouped by feature or system framework for easy discovery:
Screenshot 2020-06-16 at 09 58 27

* swizzling test timeouts are updated, as I noticed several timeout failures on my local machine (and we had some yesterday on CI),
* linter was fixed as it was giving false-positives in `URLSessionSwizzler` - [`swiftlint` bug](https://github.com/realm/SwiftLint/issues/3112).

### Review checklist


Sources/Datadog/Core/AutoInstrumentation/URLSessionSwizzler.swift
```diff
@@ -34,8 +34,9 @@ internal class URLSessionSwizzler {
     }
 
     static var hasSwizzledBefore = false
+
     @discardableResult
-    func swizzleOnce(using interceptor: @escaping RequestInterceptor) -> Bool {
+    func swizzleOnce(using interceptor: @escaping RequestInterceptor) -> Bool { // swiftlint:disable:this attributes
```
buranmert, line 39: this line doesn't give me warning, which version of swiftlint are you using?

ncreated: Yes, it's weird - CI is also good, but if I run the CI locally it fails.
```
❯ swiftlint version
0.38.0
```

Screenshot 2020-06-16 at 09 40 48

buranmert: looks like you need to update your swiftlint, mine is `0.39.1`

ncreated: 👍 works in `0.39.1`. Removed that change.

PR #139, opened 2020-06-16
RUMM-489 TracingRequestInterceptor added

buranmert:
### What and why?

We need to trace every network request automatically as part of auto-instrumentation feature.
[First](https://github.com/DataDog/dd-sdk-ios/pull/113) and [second](https://github.com/DataDog/dd-sdk-ios/pull/116) parts are already implemented.
This is the last part of tracing auto-instrumentation.

### How?

With previous PRs we implemented a way to inject our custom code into `URLSession` methods.
Now we inject the code to:
1. create&inject tracing headers to network requests
2. create span context at network task creation `dataTaskWithURL/Request`
3. create span at starting network operation `task.resume`

### Review checklist

- [x] Feature or bugfix MUST have appropriate unit tests
- [x] Feature or bugfix MUST have appropriate integration tests

### DIFF
--- a/Datadog/Example/AppConfig.swift
+++ b/Datadog/Example/AppConfig.swift
@@ -12,6 +12,8 @@ protocol AppConfig {
     var serviceName: String { get }
     /// SDK configuration
     var datadogConfiguration: Datadog.Configuration { get }
+    /// Endpoint for arbitrary network requests
+    var sourceEndpoint: URL { get }
 }
 
 struct ExampleAppConfig: AppConfig {
@@ -19,6 +21,7 @@ struct ExampleAppConfig: AppConfig {
     let serviceName = "ios-sdk-example-app"
     /// Configuration for uploading logs to Datadog servers
     let datadogConfiguration: Datadog.Configuration
+    let sourceEndpoint = URL(string: "https://app.datadoghq.com/")!
 
     init() {
         guard let clientToken = Bundle.main.infoDictionary?["DatadogClientToken"] as? String, !clientToken.isEmpty else {
@@ -32,6 +35,7 @@ struct ExampleAppConfig: AppConfig {
 
         self.datadogConfiguration = Datadog.Configuration
             .builderUsing(clientToken: clientToken, environment: "tests")
+            .setTracedHosts([sourceEndpoint])
             .build()
     }
 }
@@ -41,12 +45,18 @@ struct UITestAppConfig: AppConfig {
     let serviceName = "ui-tests-service-name"
     /// Configuration for uploading logs to mock servers
     let datadogConfiguration: Datadog.Configuration
+    let sourceEndpoint: URL
 
-    init(mockLogsEndpoint: String, mockTracesEndpoint: String) {
+    init() {
+        let mockLogsEndpoint = ProcessInfo.processInfo.environment["DD_MOCK_LOGS_ENDPOINT_URL"]!
+        let mockTracesEndpoint = ProcessInfo.processInfo.environment["DD_MOCK_TRACES_ENDPOINT_URL"]!
+        let sourceEndpointString = ProcessInfo.processInfo.environment["DD_MOCK_SOURCE_ENDPOINT_URL"]!
+        sourceEndpoint = URL(string: sourceEndpointString)!
         self.datadogConfiguration = Datadog.Configuration
             .builderUsing(clientToken: "ui-tests-client-token", environment: "integration")
             .set(logsEndpoint: .custom(url: mockLogsEndpoint))
             .set(tracesEndpoint: .custom(url: mockTracesEndpoint))
+            .setTracedHosts([sourceEndpoint])
             .build()
     }
 }
@@ -54,10 +64,7 @@ struct UITestAppConfig: AppConfig {
 /// Returns different `AppConfig` when running in UI Tests or launching directly.
 func currentAppConfig() -> AppConfig {
     if ProcessInfo.processInfo.arguments.contains("IS_RUNNING_UI_TESTS") {
-        return UITestAppConfig(
-            mockLogsEndpoint: ProcessInfo.processInfo.environment["DD_MOCK_LOGS_ENDPOINT_URL"]!,
-            mockTracesEndpoint: ProcessInfo.processInfo.environment["DD_MOCK_TRACES_ENDPOINT_URL"]!
-        )
+        return UITestAppConfig()
     } else {
         return ExampleAppConfig()
     }
--- a/Datadog/Example/AppConfig.swift
+++ b/Datadog/Example/AppConfig.swift
@@ -12,8 +12,9 @@ protocol AppConfig {
     var serviceName: String { get }
     /// SDK configuration
     var datadogConfiguration: Datadog.Configuration { get }
-    /// Endpoint for arbitrary network requests
-    var sourceEndpoint: URL { get }
+    /// Endpoints for arbitrary network requests
+    var arbitraryNetworkURL: URL { get }
+    var arbitraryNetworkRequest: URLRequest { get }
 }
 
 struct ExampleAppConfig: AppConfig {
@@ -21,7 +22,14 @@ struct ExampleAppConfig: AppConfig {
     let serviceName = "ios-sdk-example-app"
     /// Configuration for uploading logs to Datadog servers
     let datadogConfiguration: Datadog.Configuration
-    let sourceEndpoint = URL(string: "https://app.datadoghq.com/")!
+
+    let arbitraryNetworkURL = URL(string: "https://status.datadoghq.com")!
+    let arbitraryNetworkRequest: URLRequest = {
+        var request = URLRequest(url: URL(string: "https://status.datadoghq.com/bad/path")!)
+        request.httpMethod = "POST"
+        request.addValue("dataTaskWithRequest", forHTTPHeaderField: "creation-method")
+        return request
+    }()
 
     init() {
         guard let clientToken = Bundle.main.infoDictionary?["DatadogClientToken"] as? String, !clientToken.isEmpty else {
@@ -35,7 +43,7 @@ struct ExampleAppConfig: AppConfig {
 
         self.datadogConfiguration = Datadog.Configuration
             .builderUsing(clientToken: clientToken, environment: "tests")
-            .setTracedHosts([sourceEndpoint])
+            .set(tracedHosts: [arbitraryNetworkURL.host!, "foo.bar"])
             .build()
     }
 }
@@ -45,19 +53,29 @@ struct UITestAppConfig: AppConfig {
     let serviceName = "ui-tests-service-name"
     /// Configuration for uploading logs to mock servers
     let datadogConfiguration: Datadog.Configuration
-    let sourceEndpoint: URL
+    let arbitraryNetworkURL: URL
+    let arbitraryNetworkRequest: URLRequest
 
     init() {
         let mockLogsEndpoint = ProcessInfo.processInfo.environment["DD_MOCK_LOGS_ENDPOINT_URL"]!
         let mockTracesEndpoint = ProcessInfo.processInfo.environment["DD_MOCK_TRACES_ENDPOINT_URL"]!
-        let sourceEndpointString = ProcessInfo.processInfo.environment["DD_MOCK_SOURCE_ENDPOINT_URL"]!
-        sourceEndpoint = URL(string: sourceEndpointString)!
+        let sourceEndpoint = ProcessInfo.processInfo.environment["DD_MOCK_SOURCE_ENDPOINT_URL"]!
+        let tracedhost = URL(string: sourceEndpoint)!.host!
         self.datadogConfiguration = Datadog.Configuration
             .builderUsing(clientToken: "ui-tests-client-token", environment: "integration")
             .set(logsEndpoint: .custom(url: mockLogsEndpoint))
             .set(tracesEndpoint: .custom(url: mockTracesEndpoint))
-            .setTracedHosts([sourceEndpoint])
+            .set(tracedHosts: [tracedhost, "foo.bar"])
             .build()
+
+        let url = URL(string: sourceEndpoint)!
+        self.arbitraryNetworkURL = URL(string: url.deletingLastPathComponent().absoluteString + "inspect")!
+        self.arbitraryNetworkRequest = {
+            var request = URLRequest(url: url)
+            request.httpMethod = "POST"
+            request.addValue("dataTaskWithRequest", forHTTPHeaderField: "creation-method")
+            return request
+        }()
     }
 }
 
--- a/Datadog/Example/IntegrationTestFixtures/SendTracesFixtureViewController.swift
+++ b/Datadog/Example/IntegrationTestFixtures/SendTracesFixtureViewController.swift
@@ -57,15 +57,17 @@ internal class SendTracesFixtureViewController: UIViewController {
 
         viewAppearingSpan.finish()
 
-        // Step #2: Auto Instrumentation
-        // Send two requests which will be automatically traced as tracing auto-instrumentation is enabled
-        let tracedHost = currentAppConfig().sourceEndpoint.absoluteString
-        let url = URL(string: "\(tracedHost)/dataTaskWithURL")!
-        var request = URLRequest(url: URL(string: "\(tracedHost)/dataTaskWithRequest")!)
-        request.httpMethod = "POST"
+        // Send requests which will be automatically traced as tracing auto-instrumentation is enabled
+        let url = currentAppConfig().arbitraryNetworkURL
+        let request = currentAppConfig().arbitraryNetworkRequest
+        let dnsErrorURL = URL(string: "https://foo.bar")!
+        // Step #2: Auto-instrumentated request with URL to succeed
         URLSession.shared.dataTask(with: url) { _, _, _ in
-            // Step #3
-            URLSession.shared.dataTask(with: request) { _, _, _ in }.resume()
+            // Step #3: Auto-instrumentated request with Request to fail
+            URLSession.shared.dataTask(with: request) { _, _, _ in
+                // Step #4: Auto-instrumentated request to return NSError
+                URLSession.shared.dataTask(with: dnsErrorURL) { _, _, _ in }.resume()
+            }.resume()
         }.resume()
     }
 
--- a/Datadog/Example/IntegrationTestFixtures/SendTracesFixtureViewController.swift
+++ b/Datadog/Example/IntegrationTestFixtures/SendTracesFixtureViewController.swift
@@ -29,6 +29,7 @@ internal class SendTracesFixtureViewController: UIViewController {
         dataDownloadingSpan.setTag(key: "data.url", value: URL(string: "https://example.com/image.png")!)
         dataDownloadingSpan.setTag(key: DDTags.resource, value: "GET /image.png")
 
+        // Step #1: Manual tracing with complex hierarchy
         downloadSomeData { [weak self] data in
             // Simulate logging download progress
             dataDownloadingSpan.log(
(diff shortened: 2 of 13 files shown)


Sources/Datadog/Tracing/AutoInstrumentation/TracingURLSessionHooks.swift
```diff
@@ -0,0 +1,75 @@
...
+                return
+            }
+            previousEvent = observedEvent
+
+            switch observedEvent {
+            case .starting:
+                startedSpan = tracer.startSpan(
+                    spanContext: spanContext,
+                    operationName: operationName
+                )
```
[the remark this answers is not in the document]
buranmert: 👍 
let's merge https://github.com/DataDog/opentracing-swift/pull/1/files first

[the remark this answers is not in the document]
ncreated: It's merged 👌. Still we need to merge https://github.com/DataDog/dd-sdk-ios/pull/141 to use standard tags in SDK.

buranmert:
As it turns out to be a bit tricky, I'll expose this in public API in the next PR
Let's keep this one only for `TracingRequestInterceptor` implementation

Sources/Datadog/Tracing/TracingAutoInstrumentation.swift
```diff
@@ -0,0 +1,103 @@
...
+        guard let someHosts = tracedHosts else {
+            return nil
+        }
+        do {
+            self.tracedHosts = someHosts
+            self.swizzler = try URLSessionSwizzler()
+        } catch {
+            userLogger.warn("🔥 Network requests won't be traced automatically for \(String(describing: tracedHosts)): \(error)")
+            developerLogger?.warn("🔥 Network requests won't be traced automatically for \(String(describing: tracedHosts)): \(error)")
+            return nil
```
ncreated, line 25: It's missing tests. By convention, we test all user facing features, including printed error messages if possible. See example in:

https://github.com/DataDog/dd-sdk-ios/blob/c93f9687a7e18199d91f6f38e1a9944f1c3ed738/Tests/DatadogTests/Datadog/TracerTests.swift#L571-L588

Sources/Datadog/Tracing/TracingAutoInstrumentation.swift
```diff
@@ -0,0 +1,103 @@
...
+            case .completed(let response, let error):
+                guard let completedSpan = startedSpan else {
+                    break
+                }
+                if let statusCode = (response as? HTTPURLResponse)?.statusCode {
+                    completedSpan.setTag(key: OTTags.httpStatusCode, value: statusCode)
+                }
+                if let someError = error {
+                    completedSpan.logError(someError)
+                }
```
ncreated, line 89: This is a critical part for trace rendering on https://app.datadoghq.com, but it's missing tests.

Sources/Datadog/Tracing/DDSpan.swift
```diff
@@ -121,3 +121,23 @@ internal class DDSpan: OTSpan {
...
+            OTLogFields.stack: String(describing: error)
+        ]
+        if let nsError = (error as NSError?) {
+            let errorKind = "\(nsError.domain) - \(nsError.code)"
+            logFields[OTLogFields.errorKind] = errorKind
+            let errorMessage = nsError.localizedDescription
+            logFields[OTLogFields.message] = errorMessage
+        }
+        log(fields: logFields, timestamp: timestamp)
+    }
```
ncreated, line 142: Marking `span` error this way will only work if Logging feature is enabled. If it's disabled, tracing auto-instrumentation will result with no error reporting.

Instead of depending on Logging For Tracing integration, we can set appropriate tags directly on `span`. This is exactly how LFT works:

https://github.com/DataDog/dd-sdk-ios/blob/c93f9687a7e18199d91f6f38e1a9944f1c3ed738/Sources/Datadog/Tracing/Span/SpanTagsReducer.swift#L54-L56

It's less code, and works independently of Logging feature.

Sources/Datadog/Tracing/TracingAutoInstrumentation.swift
```diff
@@ -0,0 +1,103 @@
...
+        let observer: TaskObserver = { observedEvent in
+            switch observedEvent {
+            case .starting(let request):
+                if let ongoingSpan = startedSpan {
+                    userLogger.warn("\(String(describing: request)) is starting a new trace but it's already started a trace before: \(ongoingSpan)")
+                    developerLogger?.warn("\(String(describing: request)) is starting a new trace but it's already started a trace before: \(ongoingSpan)")
+                }
+                let span = tracer.startSpan(
+                    spanContext: spanContext,
+                    operationName: operationName
```
ncreated, line 73: To unify it with the way network `spans` [are named in Android SDK](https://github.com/DataDog/dd-sdk-android/blob/d512922887ac0d214bfc0b5808ac05c226f90235/dd-sdk-android/src/main/kotlin/com/datadog/android/tracing/internal/net/TracingRequestInterceptor.kt#L57), the `operationName` should be constant: `"urlsession.request"`.

Also, we should set the `resourceName` to `request.url` as [we do in Android SDK](https://github.com/DataDog/dd-sdk-android/blob/d512922887ac0d214bfc0b5808ac05c226f90235/dd-sdk-android/src/main/kotlin/com/datadog/android/tracing/internal/net/TracingRequestInterceptor.kt#L63).

Tests/DatadogIntegrationTests/TracingIntegrationTests.swift
```diff
@@ -69,11 +93,22 @@ class TracingIntegrationTests: IntegrationTests {
...
         XCTAssertEqual(try spanMatchers[2].resource(), try spanMatchers[2].operationName())
+        XCTAssertEqual(try autoTracedWithURL.resource(), try autoTracedWithURL.operationName())
+        XCTAssertEqual(try autoTracedWithRequest.resource(), try autoTracedWithRequest.operationName())
+
+        // assert baggage item:
+        XCTAssertEqual(try spanMatchers[0].meta.custom(keyPath: "meta.class"), "SendTracesFixtureViewController")
+        XCTAssertEqual(try spanMatchers[1].meta.custom(keyPath: "meta.class"), "SendTracesFixtureViewController")
+        XCTAssertEqual(try spanMatchers[2].meta.custom(keyPath: "meta.class"), "SendTracesFixtureViewController")
+        XCTAssertNil(try? autoTracedWithURL.meta.custom(keyPath: "meta.class"))
+        XCTAssertNil(try? autoTracedWithRequest.meta.custom(keyPath: "meta.class"))
```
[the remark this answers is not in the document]
buranmert: yep, i was running with the convention
removed ✅

Tests/DatadogIntegrationTests/UITestsHelpers.swift
```diff
@@ -14,12 +14,15 @@ class ExampleApplication: XCUIApplication {
     }
 
     /// Launches the app by mocking feature endpoints separately.
-    func launchWith(mockLogsEndpointURL: URL, mockTracesEndpointURL: URL) {
+    func launchWith(mockLogsEndpointURL: URL, mockTracesEndpointURL: URL, mockSourceEndpointURL: URL? = nil) {
```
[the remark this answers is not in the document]
buranmert: we use `nil` in `LogsFixtures`

Sources/Datadog/Tracing/TracingAutoInstrumentation.swift
buranmert: good catch
adding 👍

Sources/Datadog/DatadogConfiguration.swift
```diff
@@ -127,6 +129,17 @@ extension Datadog {
                 return self
             }
 
+            /// Sets the hosts to be automatically traced.
+            ///
+            /// Example, if tracedHosts is ["https://foo.bar"], then every network request such as "https://foo.bar/some/path" will be automatically traced.
+            /// If nil, automatic tracing is disabled.
+            ///
+            /// - Parameter tracedHosts: `nil` by default
+            public func setTracedHosts(_ tracedHosts: Set<URL>?) -> Builder {
```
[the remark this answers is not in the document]
buranmert: i also just realized there was a case where we may swizzle without intercepting (if `tracedHosts.isEmpty` is `true`)
we should make it non-optional 👍 

IMO `set(tracedHosts:spanDecorator:)` is following the other examples in `Builder` interface 👌 
`automaticallyTrace` would be an exception and also it's a verb whereas it just changes the config really
so overall `set(tracedHosts:)` feels slightly better to me, wdyt?

(again IMO we should get rid of `Builder` and its tests when we can 🙏 😄 )

Sources/Datadog/Tracing/DDSpan.swift
buranmert: so we should prefer `DatadogTagKeys` over `OTLogFields`, is this correct?

Sources/Datadog/Tracing/TracingAutoInstrumentation.swift
buranmert:
```swift
let span = tracer.startSpan(
  spanContext: spanContext,
  operationName: "urlsession.request"
)
span.setTag(key: DDTags.resource, value: url)
```
✅

Tests/DatadogIntegrationTests/UITestsHelpers.swift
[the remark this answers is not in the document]
buranmert: i can add `mockSourceEndpointURL: mockServerURL` to `launchWith(mockServerURL:)` method to make it non-nil but that may lead to arbitrary requests made to `Logs/Traces` intake
i prefer keeping it `nil` and explicit, wdyt?

Sources/Datadog/Tracing/DDSpan.swift
buranmert: summary of a quick Zoom: yes, we prefer `DDTags` in internal code over `OTLogFields` ✅

Sources/Datadog/Tracing/TracingAutoInstrumentation.swift
ncreated: Ah I see 👌. For critical & complex things we may use component injection through protocol and provide throwing mock implementation in tests, but I think this part is not that complex, so it's clear and safe from the code. Let's categorise it as "not easy possible" and keep no coverage for it 👍.

Sources/Datadog/Tracing/TracingAutoInstrumentation.swift
```diff
@@ -0,0 +1,113 @@
...
+                startedSpan = span
+            case .completed(let response, let error):
+                guard let completedSpan = startedSpan else {
+                    break
+                }
+                if let statusCode = (response as? HTTPURLResponse)?.statusCode {
+                    completedSpan.setTag(key: OTTags.httpStatusCode, value: statusCode)
+                }
+                if let someError = error {
+                    completedSpan.handleError(someError)
```
ncreated, line 87: When testing by running the `Example` app on Simulator and browsing seeded traces on app.datadoghq.com I spotted another misalignment with Android SDK. Because the app tries to download data from `https://app.datadoghq.com//dataTaskWithURL`, it results with `HTTP 404` error. Following the [Android SDK implementation](https://github.com/DataDog/dd-sdk-android/blob/master/dd-sdk-android/src/main/kotlin/com/datadog/android/tracing/internal/net/TracingRequestInterceptor.kt#L78-L93), this should have a separate handling based on the HTTP status codes:
```kotlin
span.setTag(Tags.HTTP_STATUS.key, statusCode)
if (statusCode in 400..499) {
   (span as? MutableSpan)?.isError = true
}
if (statusCode == 404) {
   // limit spam on 404 resources
   (span as? MutableSpan)?.resourceName = "404"
}
```
Such implementation will:
* mark the span with error label in Traces Explorer,
* group all `404s` under single `"404"` RESOURCE facet for easy discovery by users.

Datadog/Example/IntegrationTestFixtures/SendTracesFixtureViewController.swift
```diff
@@ -55,6 +56,17 @@ internal class SendTracesFixtureViewController: UIViewController {
         super.viewDidAppear(animated)
 
         viewAppearingSpan.finish()
+
+        // Step #2: Auto Instrumentation
+        // Send two requests which will be automatically traced as tracing auto-instrumentation is enabled
+        let tracedHost = currentAppConfig().sourceEndpoint.absoluteString
+        let url = URL(string: "\(tracedHost)/dataTaskWithURL")!
+        var request = URLRequest(url: URL(string: "\(tracedHost)/dataTaskWithRequest")!)
```
ncreated, line 64: This is OK for running the Example app in `UITests`. But when launching it on Simulator or device to seed data to Traces Explorer, this results with hitting the `404` resource: `https://app.datadoghq.com//dataTaskWithURL`. This misaligns the behaviour between tests vs regular launch:
* Tests: `200` OK for `GET`,  `200` OK for `POST`,
* Regular Launch: `404` error for `GET`, `404` error for `POST`.

Because of this, we don't seed enough trace data to Traces Explorer for lookup and reference, as it's done for `"view appearing"` trace. The `"view appearing"` trace contains 3 spans, from which 2 are OK and 1 is error.  **By "not enough trace data" I mean: we don't even know how both, the "ok" and "error" spans send from tracing interceptor will look in Traces Explorer** and this is IMO critical to implement.

I think we can easily fix it by differentiating the `GET` and `POST` urls in the `Example` app:
* for tests, pass two distinct Python server `.recordingURL` (by obtaining two server sessions),
* for regular launch, just configure two URLs in `ExampleAppConfig`.

As this will still produce two `200s` in tests (vs ` GET → 200` and `POST → 4xx` in regular launch), we can further improve it (in separate JIRA), so that Python server can return `4xx` for predefined POST request.

WDYT @buranmert ?

Datadog/Example/AppConfig.swift
```diff
@@ -12,13 +12,16 @@ protocol AppConfig {
...
+    /// Endpoint for arbitrary network requests
+    var sourceEndpoint: URL { get }
 }
 
 struct ExampleAppConfig: AppConfig {
     /// Service name used for logs and traces.
     let serviceName = "ios-sdk-example-app"
     /// Configuration for uploading logs to Datadog servers
     let datadogConfiguration: Datadog.Configuration
+    let sourceEndpoint = URL(string: "https://app.datadoghq.com/")!
```
ncreated, line 24: As our backend monitoring intensively watches the DD entry point: https://app.datadoghq.com/, maybe we can use more "technical" URL for this? How about https://status.datadoghq.com?

Sources/Datadog/Tracing/TracingAutoInstrumentation.swift
buranmert: 💡 we could have tested this by removing `dataTaskWithURL/Request` methods from `URLSession` but `ObjCRuntime` doesn't provide `removeMethod`

Sources/Datadog/Tracing/TracingAutoInstrumentation.swift
buranmert: copied Android's imp ✅

Datadog/Example/IntegrationTestFixtures/SendTracesFixtureViewController.swift
buranmert: summary: basically that reads we should send one OK and one ERROR span in `ExampleApp`; duplicate this in tests when mock server can return failure as well.

we can make 2 requests: `www.good-url.com` and `www.good-url.com/bad-path`, that result 1 OK and 1 ERROR in App; test always has 2 OKs for now.

Sources/Datadog/DatadogConfiguration.swift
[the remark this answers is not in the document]
ncreated: > IMO `set(tracedHosts:spanDecorator:)` is following the other examples in Builder interface 👌

Sounds fine, let's change to `set(tracedHosts:)` 👍

> (again IMO we should get rid of Builder and its tests when we can 🙏 😄 )

Yay, one day for `2.0.0` 👌  - now it's too intrusive breaking change.

Datadog/Example/IntegrationTestFixtures/SendTracesFixtureViewController.swift
buranmert: as `MockServer` doesn't accept `GET` requests, i went with a sub-optimal solution/workaround

`AppConfig` has `arbitraryNetworkURL/Request`, `ExampleAppConfig` and `UITestsConfig` provide different values

Datadog/Example/AppConfig.swift
buranmert: set to Datadog logo image URL ✅

Sources/Datadog/DatadogConfiguration.swift
```diff
@@ -127,6 +129,17 @@ extension Datadog {
                 return self
             }
 
+            /// Sets the hosts to be automatically traced.
+            ///
+            /// Example, if tracedHosts is ["https://foo.bar"], then every network request such as "https://foo.bar/some/path" will be automatically traced.
+            /// If nil, automatic tracing is disabled.
+            ///
+            /// - Parameter tracedHosts: `nil` by default
```
ncreated, line 137: This public API comment needs a fix - the `https://foo.bar` is not a valid input. Let's use a realistic example, like `example.com` + `api.example.com` instead of `foo.bar`.

It also doesn't mention `URLSession`, nor distributed tracing and doesn't explain what "automatically traced" means. IMO the [comment in Android SDK](https://github.com/DataDog/dd-sdk-android/blob/34c6d35ccf4c1f88fb64df47861ffbed5dc06ca8/dd-sdk-android/src/main/kotlin/com/datadog/android/tracing/TracingInterceptor.kt#L16-L39) is a good reference, let's elevate it to that level (?) 📈.

buranmert: IMO `tracedHost: example.com` including `api.example.com` is already very confusing in iOS as those URLs have different `url.host`s. we can follow Android and not mention this behavior in comments, wdyt?

```swift
Sets the hosts to be automatically traced.

Every request made to a traced host will create its Span with related information; such as url, method, status code, error (if any).

Example, if tracedHosts is ["https://example.com"], then every network request such as "https://example.com/some/path" will be automatically traced and generate a span.
If tracedHosts is empty, automatic tracing is disabled.

IMPORTANT: Non-empty tracedHosts modifies implementation of URLSession methods, in case your app relies on URLSession internals please refer to URLSessionSwizzler.swift file
```
i'm not so sure of mentioning distributed tracing as we support it only in some cases. does that sound good?

ncreated: > IMO tracedHost: example.com including api.example.com is already very confusing in iOS as those URLs have different url.hosts. we can follow Android and not mention this behavior in comments, wdyt?

We want the same behaviour as Android SDK, which is:
```python
if tracedHosts == ["example.com"], then:

https://api.example.com/v1/oauth     - traced
https://example.com/api/v1/oauth     - traced
http://www.example.com/api/v1/oauth  - traced
https://foo.com/                     - not traced


if tracedHosts == ["api.example.com"], then:

https://api.example.com/v1/oauth     - traced
https://example.com/api/v1/oauth     - not traced
http://www.example.com/api/v1/oauth  - not traced
https://foo.com/                     - not traced
```

If we expect the user to provide schema (`https`/`http`), then this concept gets broken or becomes non-obvious, no?

> i'm not so sure of mentioning distributed tracing as we support it only in some cases. does that sound good?

IMO we should mention it, as it's a powerful integration. If it doesn't work for everyone, we can say _"If your backend is instrumented with Datadog, ..."_.

buranmert: oh yes, i'm removing `https` from comments 👍 

i mean it doesn't work for `dataTaskWithURL` in iOS 13, shall we mention this feature anyway?

Sources/Datadog/Tracing/TracingAutoInstrumentation.swift
```diff
@@ -0,0 +1,129 @@
...
+        }
+        do {
+            /// pattern = "^(.*\\.)*tracedHost1|^(.*\\.)*tracedHost2|..."
+            let pattern = tracedHosts
+                .map {
+                    let escaped = NSRegularExpression.escapedPattern(for: $0)
+                    return "^(.*\\.)*\(escaped)$"
+                }
+                .joined(separator: "|")
+            let regex = try NSRegularExpression(pattern: pattern, options: [.caseInsensitive, .anchorsMatchLines])
```
ncreated, line 27: Suggestion: the `TracingAutoInstrumentation` becomes more and more complex. If we want to reduce the complexity, hosts processing could be decoupled to a basic type:
```swift
let allowedHosts = AllowedHosts(hosts: ["v2.api.com"])

internal struct AllowedHosts {
    private let regex: String

    init(hosts: [String]) {}

    func matches(url: URL) -> Bool {}
}
```
This way, the `AllowedHosts` can be tested as separate unit. Currently all host-matching tests depend on the complex condition in `TracingRequestInterceptor`:
```swift
guard let tracer = Global.sharedTracer as? DDTracer,
      let someURL = urlRequest.url,
      tracedHostsRegex.allows(someURL),
      HTTPHeadersWriter.canInject(to: urlRequest) else {
         return nil
}
```
and may give false-positives due to wrong configuration.

buranmert: could be a good idea, but i don't think we need it today

PR #140, opened 2020-06-16
RUMM-333 Set boundaries for trace and span ID generation

ncreated:
### What and why?

📦 This PR adds boundaries to the generated `trace_id` and `span_id` so it stays compatible with Datadog infrastructure.

### How?

Following the [dd-trace-java/DDId.java](https://github.com/DataDog/dd-trace-java/blob/master/dd-trace-api/src/main/java/datadog/trace/api/DDId.java) implementation, the boundaries are:
* the `id` should be higher or equal `1`,
* the `id` should be less or equal `2 ^ 63 - 1`.

### Review checklist

### DIFF
--- a/Sources/Datadog/Tracing/UUIDs/TracingUUIDGenerator.swift
+++ b/Sources/Datadog/Tracing/UUIDs/TracingUUIDGenerator.swift
@@ -11,9 +11,25 @@ internal protocol TracingUUIDGenerator {
 }
 
 internal struct DefaultTracingUUIDGenerator: TracingUUIDGenerator {
+    /// Lower boundary of `TracingUUID`.
+    /// `0` is reserved for historical reason: 0 == "unset", ref: dd-trace-java:DDId.java.
+    internal static let min: UInt64 = 1
+    /// Upper boundary of `TracingUUID`.
+    /// It equals to `2 ^ 63 - 1` because some tracers can't handle the `2 ^ 64 -1` range, ref: dd-trace-java:DDId.java.
+    internal static let max = UInt64.max >> 1
+
+    internal let min: UInt64
+    internal let max: UInt64
+
+    init(
+        lowerBoundary: UInt64 = DefaultTracingUUIDGenerator.min,
+        upperBoundary: UInt64 = DefaultTracingUUIDGenerator.max
+    ) {
+        self.min = lowerBoundary
+        self.max = upperBoundary
+    }
+
     func generateUnique() -> TracingUUID {
-       // TODO: RUMM-333 Add boundaries to trace & span ID generation, keeping in mind that `0` is reserved (ref: DDTracer.java#L600)
-       // NOTE: RUMM-340 Consider thread safety if the generation will depend on local state
-       return TracingUUID(rawValue: .random(in: 1...UInt64.max))
+        return TracingUUID(rawValue: .random(in: min...max))
    }
 }
--- a/Sources/Datadog/Tracing/UUIDs/TracingUUIDGenerator.swift
+++ b/Sources/Datadog/Tracing/UUIDs/TracingUUIDGenerator.swift
@@ -11,25 +11,18 @@ internal protocol TracingUUIDGenerator {
 }
 
 internal struct DefaultTracingUUIDGenerator: TracingUUIDGenerator {
-    /// Lower boundary of `TracingUUID`.
-    /// `0` is reserved for historical reason: 0 == "unset", ref: dd-trace-java:DDId.java.
-    internal static let min: UInt64 = 1
-    /// Upper boundary of `TracingUUID`.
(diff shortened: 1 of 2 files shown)


Sources/Datadog/Tracing/UUIDs/TracingUUIDGenerator.swift
```diff
@@ -11,9 +11,25 @@ internal protocol TracingUUIDGenerator {
...
+    internal let min: UInt64
+    internal let max: UInt64
+
+    init(
+        lowerBoundary: UInt64 = DefaultTracingUUIDGenerator.min,
+        upperBoundary: UInt64 = DefaultTracingUUIDGenerator.max
+    ) {
+        self.min = lowerBoundary
+        self.max = upperBoundary
+    }
```
buranmert, line 30: we can use 1 `ClosedRange<UInt64>` instead of 2 `UInt64` instances
that way we don't need to worry about `lowerBoundary > upperBoundary` case as it's handled by `ClosedRange`

```swift
internal static let defaultRange = 1...(UInt64.max << 1) // some comments

internal let range: ClosedRange<UInt64>
init(range: ClosedRange<UInt64> = Self.defaultRange) { self.range = range }
```

Sources/Datadog/Tracing/UUIDs/TracingUUIDGenerator.swift
```diff
@@ -11,9 +11,25 @@ internal protocol TracingUUIDGenerator {
...
+
     func generateUnique() -> TracingUUID {
-       // TODO: RUMM-333 Add boundaries to trace & span ID generation, keeping in mind that `0` is reserved (ref: DDTracer.java#L600)
-       // NOTE: RUMM-340 Consider thread safety if the generation will depend on local state
-       return TracingUUID(rawValue: .random(in: 1...UInt64.max))
+        return TracingUUID(rawValue: .random(in: min...max))
```
buranmert, line 33: why don't we keep an internal counter and generate `UUID`s with it?
```swift
private var lastUUID: UInt64 = 0
func generateUnique() -> TracingUUID {
  let newUUID = lastUUID + 1
  defer { lastUUID = newUUID }
  return TracingUUID(newUUID)
}
```
this would give us real uniqueness instead of randomness, better testability (as deterministic) and also cheaper in terms of performance

is randomness part of OpenTracing?

Sources/Datadog/Tracing/UUIDs/TracingUUIDGenerator.swift
ncreated: Good idea 🏅, done ✅.

Sources/Datadog/Tracing/UUIDs/TracingUUIDGenerator.swift
ncreated: This will not work in distributed tracing 🤔. Given two apps launched about the same time, the trace IDs they generate will be the same (!), so the trace collector on backend won't be able to match the distributed span IDs with the right trace ID as there will be multiple traces with the same ID.

The idea behind distributed tracing is to uniformly generate unique IDs. Hence the reasonably high range: `1` - `2 ^63 -1` to avoid collisions.

buranmert: hmm i see, then we can do that:
```swift
private let lastUUID: UInt64 = .random(in: range)
func generateUnique() -> TracingUUID {
  defer { lastUUID += 1 }
  return TracingUUID(lastUUID)
}
```
basically, randomizing the counter at init and then increment one by one
that way we won't need to run this method 1000 times in tests
wdyt?

ncreated: > ```swift
>  private let lastUUID: UInt64 = .random(in: range)
>  func generateUnique() -> TracingUUID {
>     defer { lastUUID += 1 }
>     return TracingUUID(lastUUID)
> }
> ```

This won't lead to [uniform distribution](https://en.wikipedia.org/wiki/Uniform_distribution_(continuous)). Instead, there will be groups of UUIDs with higher density, which may overlap considering that we may have million of tracers, each generating hundreds of `trace_id`. The risk of `trace_id` collision is mitigated by the uniform randomization.

> basically, randomizing the counter at init and then increment one by one
that way we won't need to run this method 1000 times in tests

Why is it simpler or better for us? This will not only break the distributed tracing concept, but will also incorporate the need of `lastUUID` synchronization between threads (performance impact).

buranmert: it could be simpler/better because we could test uniqueness and wouldn't have to run 1000 times.

> The risk of trace_id collision is mitigated by the uniform randomization
i see 👍

ncreated: We should really test the distribution of IDs (e.g. if there's statistically same amount of IDs generated below and above `UInt64.max / 2`). But we trust `UInt64.random` to be uniformed.

Issue #142, opened 2020-06-16
Cannot access Logger class with Datadog.Logger


amyslies:
With our own Logger class in Logger.swift, we can't access Datadog's Logger class with Datadog.Logger.

Already tried building a logger by doing the following:
```
import Datadog
import Datadog.Logger
import Datadog.Datadog
...

Datadog.Logger.builder... // still doesn't work

```

ncreated:
Hello @amyslies 👋. This seems to be duplicate of #127. Until we fix it, could you please try the workaround and create a separate `.swift` file that only imports `Datadog` and defines `typealias DatadogLogger = Logger`?

```swift
// DatadogLoggerAlias.swift

import class Datadog.Logger

typealias DatadogLogger = Logger
```

Then you should be good to go with `DatadogLogger.builder` instead of `Datadog.Logger.builder`.

Let me know if it helps 🙂.

ncreated:
`DDLogger` typealias was added in [`1.3.0`](https://github.com/DataDog/dd-sdk-ios/releases/tag/1.3.0).

Issue #127, opened 2020-06-05
The name of the target shouldn't be the same as the name of a main class


krzysztofzablocki:
`Logger` is such a common name that almost every project I've worked with had a class with that name.

Now because both the SDK target and the class it bundles are called `Datadog` you can't get rid of `ambiguous for type lookup in this context` error to reference the Datadog Logger type since `Datadog.Logger` won't work as it will think it's looking for a subtype of `Datadog` class and not module.

This should be corrected, either by some renaming or scope changes.

PS. The workaround is to create a file that only `imports Datadog` module and create a typealias there `typealias DatadogLogger = Logger` but given how generic `Logger` name is I really think this should be corrected in the SDK itself.

ncreated:
Hey @krzysztofzablocki 👋. Thanks for reaching out. The point on the `Logger` name popularity and conflicts is totally correct.

Renaming sounds hard as it's not backward compatible, same for changing the scope. However, I did a quick POC and adding an extra `typealias` should work:
```swift
public typealias DatadogLogger = Logger
```
We will do additional tests and make it work in future releases.

> PS. The workaround is to create a file that only imports Datadog module and create a typealias there typealias DatadogLogger = Logger but given how generic Logger name is I really think this should be corrected in the SDK itself.

Thanks for sharing 💪. Small improvements which can be made to this separate-file approach is importing the `Datadog.Logger` symbol exclusively: `import class Datadog.Logger`.

ncreated:
Available in [`1.3.0`](https://github.com/DataDog/dd-sdk-ios/releases/tag/1.3.0).

PR #143, opened 2020-06-17
RUMM-501 Add `DDLogger` typealias to avoid compiler ambiguity

ncreated:
### What and why?

🚚 This PR addresses #127 and #142 where compiler ambiguity is reported in projects defining their own `Logger` class.

### How?

A public `DDLogger` typealias is exposed and can be used in projects which define their own `Logger` class.

### Review checklist

### DIFF
--- a/Sources/Datadog/Logger.swift
+++ b/Sources/Datadog/Logger.swift
@@ -73,6 +73,21 @@ public typealias AttributeKey = String
 ///
 public typealias AttributeValue = Encodable
 
+/// Because `Logger` is a common name widely used across different projects, the `Datadog.Logger` may conflict when
+/// using `Logger.builder`. In such case, following `DDLogger` typealias can be used to avoid compiler ambiguity.
+///
+/// Usage:
+///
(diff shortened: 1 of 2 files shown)


ncreated:
> if i had 2 Logger types in my project, this PR would solve my Datadog.Logger problem but i would still have TheOther.Logger colliding with Datadog.Logger, is this correct?

Yes, if any other module exposes `Logger`, this will be ambiguous:
```swift
import Datadog
import OtherModule

let logger: Logger
```
This is also ambiguous:
```swift
import Datadog

let logger: Datadog.Logger
```
as we define `public class Datadog` in SDK.

Hence, `DDLogger` is made to address all issues of this kind.

buranmert:
so we need to rename `Logger` to `DDLogger`, not only `typealias`ing it as we still make `TheOther.Logger` ambigious
plus, we have `DDTracer`; `DDLogger` would be consistent 👌 

let's add this to backlog for next major version, wdyt?

PR #146, opened 2020-06-18
RUMM-492 Encode log levels as lowercase strings

ncreated:
### What and why?

📦 With this PR and its counterpart DataDog/dd-sdk-android#288, we're changing log levels send from mobile SDKs to be lowercased strings.

### How?

Internal change to `LogEncoder` and appropriate tests.

### Review checklist

### DIFF
--- a/Sources/Datadog/Logging/Log/LogEncoder.swift
+++ b/Sources/Datadog/Logging/Log/LogEncoder.swift
@@ -9,12 +9,12 @@ import Foundation
 /// `Encodable` representation of log. It gets sanitized before encoding.
 internal struct Log: Encodable {
     enum Status: String, Encodable {
-        case debug = "debug"
-        case info = "info"
-        case notice = "notice"
-        case warn = "warn"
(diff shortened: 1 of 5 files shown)


Sources/Datadog/Logging/Log/LogEncoder.swift
```diff
@@ -9,12 +9,12 @@ import Foundation
...
-        case notice = "NOTICE"
-        case warn = "WARN"
-        case error = "ERROR"
-        case critical = "CRITICAL"
+        case debug = "debug"
+        case info = "info"
+        case notice = "notice"
+        case warn = "warn"
+        case error = "error"
+        case critical = "critical"
```
buranmert, line 17: we can remove `"..."` as they match their case value, that way we have one less opportunity to make a mistake
```swift
case debug
case info
...
```

ncreated: Done 👍

PR #147, opened 2020-06-18
RUMM-486 Fix span encoding error on iOS versions prior to 13.0

ncreated:
### What and why?

📦 Prior to `iOS13.0` the SDK was unable to send traces, as `Span` serialization was always failing. This PR fixes the behaviour on `iOS11` and `iOS12`.

### How?

As discussed in https://bugs.swift.org/browse/SR-6163, the `JSONEncoder` doesn't support primitive types encoding. This was fixed in `iOS13`.

We need primitive types encoding to make following part of Open Tracing interface compatible with Intake requirement: _"all tags must be send as `String` values"_.
```swift
func setTag(key: String, value: Codable)
```
As we can't fix `JSONEncoder`, a workaround was added - the tag `value` is encoded as array and two surrounding bytes are removed from encoded `Data`. This is the fastest and most performant workaround that I found.

### Other things done in this PR

This PR is part of `RUMM-486`, so I made all production-code unit tests green on `iOS11.1`, `iOS12.4` and `iOS13.5`. I hope this will be enough to automate tests execution on all supported platforms (coming in next PR).

I've also discovered that `URLSessionSwizzlingTests` are failing prior to `iOS13.0`. As this is uncompleted and non-public feature, I just disabled this tests temporary, so we can deliver working beta faster. This should be fixed in "tracing autoinstrumentation" scope.

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/Utils/EncodableValue.swift
+++ b/Sources/Datadog/Core/Utils/EncodableValue.swift
@@ -61,7 +61,33 @@ internal struct JSONStringEncodableValue: Encodable {
             // Switch to encode `url.absoluteString` directly - see the comment in `EncodableValue`
             try urlValue.absoluteString.encode(to: encoder)
         } else {
-            let jsonData = try jsonEncoder.encode(encodable)
+            let jsonData: Data
+
+            if #available(iOS 13.0, *) {
+                jsonData = try jsonEncoder.encode(encodable)
+            } else {
+                // Prior to `iOS13.0` the `JSONEncoder` is unable to encode primitive values - it expects them to be
+                // wrapped inside top-level JSON object (array or dictionary). Reference: https://bugs.swift.org/browse/SR-6163
+                //
+                // As a workaround, we serialize the `encodable` as a JSON array and then remove `[` and `]` bytes from serialized data.
+                let temporaryJsonArrayData = try jsonEncoder.encode([encodable])
+
+                let subdataStartIndex = temporaryJsonArrayData.startIndex.advanced(by: 1)
+                let subdataEndIndex = temporaryJsonArrayData.endIndex.advanced(by: -1)
+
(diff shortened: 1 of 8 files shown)


Sources/Datadog/Core/Utils/EncodableValue.swift
```diff
@@ -61,7 +61,33 @@ internal struct JSONStringEncodableValue: Encodable {
...
+                jsonData = try jsonEncoder.encode(encodable)
+            } else {
+                // Prior to `iOS13.0` the `JSONEncoder` is unable to encode primitive values - it expects them to be
+                // wrapped inside top-level JSON object (array or dictionary). Reference: https://bugs.swift.org/browse/SR-6163
+                //
+                // As a workaround, we serialize the `encodable` as a JSON array and then remove `[` and `]` bytes from serialized data.
+                let temporaryJsonArrayData = try jsonEncoder.encode([encodable])
+
+                let subdataStartIndex = temporaryJsonArrayData.startIndex.advanced(by: 1)
+                let subdataEndIndex = temporaryJsonArrayData.endIndex.advanced(by: -1)
```
buranmert, line 76: how does this work? 
i assume `advanced(by: 1)` means advancing 1 byte, how do we know "[/]" is always 1 byte? 
it must be `utf8` encoding for them to be 1 byte, where does `utf8` encoding come from?

buranmert: is this part of JSON spec?

ncreated: Yes, `UTF-8` is part of JSON standard ([RFC-8259](https://tools.ietf.org/html/rfc8259)). Whenever we speak about JSON, we talk about `UTF-8`.

PR #151, opened 2020-06-19
RUMM-500 Fix pod installation snippet

ncreated:
### What and why?

📦 Until our contribution in  https://github.com/opentracing/opentracing-swift/pull/11 is considered OR we go with another `.podspec` for our [OT fork](https://github.com/DataDog/opentracing-swift), the classic `pod DatadogSDK` installation will not work. We need a workaround.

### How?

The workaround is to reference `OpenTracingSwift` and `DatadogSDK` using `git` + `tag` syntax in `Podfile`.

@DataDog/documentation - it's purely technical update, nothing to review 🙂.

### Review checklist


ncreated:
> this is hard to manage for us and the users
let's have a beta branch so that they don't need to update their Podfile every now and then, wdyt?

@buranmert separate branch will be even harder for us (rebasing / merging), and is not guaranteed to be stable in our release model. Logging feature required 3-betas iteration, not a big deal. And I hope we'll manage the dependency thing until `-beta2`, then we're ultimately good 👍.

PR #157, opened 2020-06-25
RUMM-339 Expose tracing-related configuration for Objective-C

ncreated:
### What and why?

🚚 This PR exposes new tracing configuration APIs to Objective-C.

### How?

Two APIs were exposed using our way of Objective-C → Swift bridging:
```swift
.enableTracing(_ enabled:)
.set(tracesEndpoint:)
```

### Review checklist

### DIFF
--- a/Tests/DatadogTests/Datadog/Core/AutoInstrumentation/URLSessionSwizzlerTests.swift
+++ b/Tests/DatadogTests/Datadog/Core/AutoInstrumentation/URLSessionSwizzlerTests.swift
@@ -106,7 +106,7 @@ class URLSessionSwizzlerTests: XCTestCase {
                 mock2.interceptionExpectation,
                 mock3.interceptionExpectation
             ],
-            timeout: 1
+            timeout: 2
         )
 
         task.resume()
@@ -120,7 +120,7 @@ class URLSessionSwizzlerTests: XCTestCase {
         ]
         wait(
             for: resumeExpectations,
-            timeout: 1,
+            timeout: 2,
             enforceOrder: true
         )
     }
--- a/Tests/DatadogTests/Datadog/Core/AutoInstrumentation/URLSessionSwizzlerTests.swift
+++ b/Tests/DatadogTests/Datadog/Core/AutoInstrumentation/URLSessionSwizzlerTests.swift
@@ -106,7 +106,7 @@ class URLSessionSwizzlerTests: XCTestCase {
                 mock2.interceptionExpectation,
                 mock3.interceptionExpectation
             ],
-            timeout: 2
+            timeout: 1
         )
 
         task.resume()
@@ -120,7 +120,7 @@ class URLSessionSwizzlerTests: XCTestCase {
         ]
         wait(
             for: resumeExpectations,
-            timeout: 2,
+            timeout: 1,
             enforceOrder: true
         )
     }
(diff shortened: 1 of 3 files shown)


Tests/DatadogTests/Datadog/Core/AutoInstrumentation/URLSessionSwizzlerTests.swift
```diff
@@ -120,7 +120,7 @@ class URLSessionSwizzlerTests: XCTestCase {
         ]
         wait(
             for: resumeExpectations,
-            timeout: 1,
+            timeout: 2,
```
buranmert, line 123: do they fail in your local as well? if so, we may have a problem 🤔

ncreated: Yes 😞, recently I got at least 2 failures on this locally.

buranmert: okay, i can look into it 👍 
did they fail both of them at the same time? were you running all the tests or some classes/cases only?

let's keep `timeout: 1` in this PR, okay?

ncreated: It failed when running all tests. I can bring it back to `1`, but that will increase the flakiness on CI, no? We have a separate JIRA to look into flakiness, so we may tackle it in that scope. Or you think it's not a flakiness?

buranmert: IMO we can keep it `1` here 👌 
we don't make more than a simple dns query for "https://foo.bar" in test cases and as that fails, we don't make any real HTTPS connection to any server. `timeout: 1` should be more than enough for those tasks, especially in your local.

my educated-guess is unswizzling goes wrong in some other test cases and that affects those cases as well.
in short, i don't think this is flakiness. 
i'll try to fail while running them individually, if they don't then it means this is an unswizzling problem in another place.

do you remember which expectations failed? `interceptionExpectation` or `observationExpectation`?

ncreated: I just realised that `URLSessionSwizzlerTests` do real network requests 😬. That would be flaky anyway (i.e. if no network connection is available). We should be using [`ServerMock`](https://github.com/DataDog/dd-sdk-ios/blob/master/Tests/DatadogTests/Datadog/Mocks/ServerMock.swift) which mocks the session's `URLProtocol`, as we do for other unit tests.

Anyway, I set the `timeout` back to `1` ✅ and this should be changed in separate PR (to depend on a mock, instead of a real network).

> do you remember which expectations failed? interceptionExpectation or observationExpectation?

Unfortunately, nope 😞.

buranmert: yes, the idea was simulating an almost-real-world case with a real URL (whose DNS query fails)
but apparently, that _simple DNS query_ wasn't that simple 😅 
i could reproduce timeouts (heavy load ~5000 iteration & `timeout: 0.5`) even while running those cases individual 👌 

i'm raising a PR 🏗️  let's continue there 🔜

PR #158, opened 2020-06-26
RUMM-478 Skip sanitization of log attributes created by Tracing

ncreated:
### What and why?

📦 The SDK [performs sanitization](https://github.com/DataDog/dd-sdk-ios/blob/master/Sources/Datadog/Logging/Log/LogSanitizer.swift) for `Log` attributes created by the user. This includes list of reserved attributes, that get removed when given by the user. This PR adds `dd.trace_id` and `dd.span_id` to that list.

### How?

The reason why it was not done in #117 is that this includes additional refactoring. It is the subject of this PR.

The `LogOutput` was not differentiating if the attribute comes from the user, or is set internally by the `LoggingForTracingAdapter`:
```swift
internal protocol LogOutput {
    func writeLogWith(level: LogLevel, message: String, date: Date, attributes: [String: Encodable], tags: Set<String>)
}
```

This is now changed by introducing `LogAttributes` type:
```swift
internal struct LogAttributes {
    /// Log attributes received from the user. They are subject for sanitization.
    let userAttributes: [String: Encodable]
    /// Log attributes added internally by the SDK. They are not a subject for sanitization.
    let internalAttributes: [String: Encodable]?
}

internal protocol LogOutput {
    func writeLogWith(level: LogLevel, message: String, date: Date, attributes: LogAttributes, tags: Set<String>)
}
```

The `.userAttributes` are considered and `.internalAttributes` are skipped for sanitization.

### Benchmarks

I tested it with benchmarks - no impact, but in theory it should be even more performant, as more logic is now moved to `LogEncoder`, which runs on the background thread.

I also include benchmark baselines for my iPhone X deleted in #63, as a reference point for future changes.

### Review checklist

### DIFF
--- a/Datadog/Datadog.xcodeproj/xcshareddata/xcbaselines/61441C6724619FE4003D8BB8.xcbaseline/7C373BCE-2B91-4706-AD99-F9FD342891D3.plist
+++ b/Datadog/Datadog.xcodeproj/xcshareddata/xcbaselines/61441C6724619FE4003D8BB8.xcbaseline/7C373BCE-2B91-4706-AD99-F9FD342891D3.plist
@@ -1,0 +1,75 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>classNames</key>
+	<dict>
+		<key>LoggingBenchmarkTests</key>
+		<dict>
+			<key>testCreatingOneLog()</key>
+			<dict>
+				<key>com.apple.XCTPerformanceMetric_WallClockTime</key>
+				<dict>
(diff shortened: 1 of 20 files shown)


ncreated:
> as an alternative idea: we could give Keys more context
> that way we can handle more special cases with more enum cases instead of more LogAttributes properties

Maybe, if we have way more `Log` attribute kinds than "user defined" and "internal", but I don't see such use case right now nor in the near future. As `.kind` check will require additional processing and more complex hash function, this should be deeply considered.

buranmert:
i didn't mention hashing to keep it short but it was the actual gist of it:
```swift
struct Key: Hashable {
  let name: String
  var hash: Int { name.hash }
}
```
this ensures that there cannot be duplicate attributes in a dictionary
and we don't have to be careful with the order of operations (eg: first `userAttr` and then `internalAttr` in `LogEncoder`)
only required change would be `if case .sdk = key.kind { skip(key) }` in `LogSanitizer`

i'm leaving this comment here only not to forget later :)

PR #160, opened 2020-06-30
RUMM-486 Run all platform tests on manual trigger

ncreated:
### What and why?

📦 This PR adds a GitHub workflow to manually trigger the `all-platform-tests` CI job on Bitrise.

### How?

A 3rd party [pull-request-comment-trigger](https://github.com/Khan/pull-request-comment-trigger) action is used to [trigger the build on Bitrise](https://devcenter.bitrise.io/api/build-trigger/#triggering-a-new-build) if **`@test-all-platform`** keyword is given in PR description or in PR comment.

### Security

As this is a 3rd party code, I pin its version to a specific commit, so we can run the security audit before using it.

Before getting the security approval I don't pass the real GH nor Bitrise tokens through `secrets.` in this PR. Instead, a set of `DUMMY_*` repository secrets was defined, to only demonstrate how this is managed to work.

Although [passing secrets to command-line is discouraged](https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#using-encrypted-secrets-in-a-workflow) I found no better way to call the Bitrise API than using `curl`. I made sure the secrets are properly redacted, so we don't leak them in [the GH workflow output](https://github.com/DataDog/dd-sdk-ios/pull/160/checks?check_run_id=823068628). As an alternative, another GH action could be used to trigger the Bitrise web-hook - but that would mean using another 3rd party, or developing it in-house.

### Testing

Because of how GH workflow works - this cannot be tested before merging to `master`. I qualified it as smart as I could, but we'd have to land it on `master` to verify.

### Limitations

This is a one-way Bitrise web-hook. We tell Bitrise to start the build, but unfortunately [Bitrise Checks](https://devcenter.bitrise.io/jp/builds/bitrise-checks-on-github-checks/) are yet too limited to make the build pop up in PR's "Checks" section. We'd have to find another way to get the status back (either by Slack notification or by pulling the Bitrise build status from GH workflow in some time interval).

### Review checklist

### DIFF
--- a/.github/workflows/all-platform-tests.yml
+++ b/.github/workflows/all-platform-tests.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Look for a keyword triggering Bitrise build
         id: check-keyword-trigger
         env:
-          GITHUB_TOKEN: ${{ secrets.DUMMY_CONSTRAINED_GH_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         uses: khan/pull-request-comment-trigger@014b821
         with:
           trigger: '@test-all-platforms'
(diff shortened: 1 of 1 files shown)


ncreated:
🙏 testing comment trigger...

ncreated:
🙏 testing comment trigger 2...

ncreated:
> I was wondering if we could add a status badge to README to see which iOS versions fail?

As this job is triggered manually, where would we get this data from? Having `README` badges would imply that we run the `all-platform-tests` for every merge to `master` and this will impact the CI load in a bad way. We could run it for release tags, but that in turn, is too late in the process.

> And can't we use github.team instead of github.actor?

Yeah, I wish there's something like this, but [the `github context` available in GH workflows](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context) doesn't give the team membership information.

PR #161, opened 2020-07-01
RUMM-575 URLSessionSwizzlerTests flakiness

buranmert:
### What and why?

We use URLs such as "https://foo.bar" in those tests.
They don't resolve to a real address, so they only result in DNS queries, not actual HTTPS connections.
However, those DNS queries are actual connections anyway and they cause flakiness in unit tests.
[Origin of the issue](https://github.com/DataDog/dd-sdk-ios/pull/157#discussion_r448234067)

### How?

Now we use unsupported protocols in tests.

Those tasks go through session's `URLProtocol` chain and returns immediately with "unsupported URL" error.
That prevents actual network traffic in unit tests.

#### Alternative how?

We could also use custom `URLProtocol`s to achieve the same result (as we did in other test cases)
but that wouldn't work in case of `URLSession.shared`

#### ⚠️  Note ⚠️ 

If we notice weird behavior in swizzling tests again, we may want to verify if `TracingAutoInstrument.instance.unswizzle()` is properly called in other test cases

### Review checklist

### DIFF
--- a/Tests/DatadogTests/Datadog/Core/AutoInstrumentation/URLSessionSwizzlerTests.swift
+++ b/Tests/DatadogTests/Datadog/Core/AutoInstrumentation/URLSessionSwizzlerTests.swift
@@ -9,10 +9,13 @@ import XCTest
 
 class URLSessionSwizzlerTests: XCTestCase {
     /// URL.mockAny() is a valid URL so that it loads actual resource and exceeds expectation timeouts
-    let mockURL = URL(string: "https://foo.bar")!
-    let mockURLRequest = URLRequest(url: URL(string: "https://foo.bar")!)
-    let modifiedURLRequest = URLRequest(url: URL(string: "https://bar.foo")!)
-    let secondModifiedURLRequest = URLRequest(url: URL(string: "https://bar.foo.bar")!)
+    /// We use unsupported URLs so that the task goes through its URLProtocol chain and **immediately** returns with "unsupported URL" error
+    let mockURL = URL(string: "foo://example.com")!
+    let mockURLRequest = URLRequest(url: URL(string: "foo://example.com")!)
+    let modifiedURLRequest = URLRequest(url: URL(string: "bar://example.com")!)
+    let secondModifiedURLRequest = URLRequest(url: URL(string: "bar://foo.example.com")!)
+
+    let timeout = 1.0
 
     var session: URLSession { URLSession.shared }
     // swiftlint:disable implicitly_unwrapped_optional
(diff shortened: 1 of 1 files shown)


Tests/DatadogTests/Datadog/Core/AutoInstrumentation/URLSessionSwizzlerTests.swift
```diff
@@ -9,10 +9,13 @@ import XCTest
 
 class URLSessionSwizzlerTests: XCTestCase {
     /// URL.mockAny() is a valid URL so that it loads actual resource and exceeds expectation timeouts
-    let mockURL = URL(string: "https://foo.bar")!
-    let mockURLRequest = URLRequest(url: URL(string: "https://foo.bar")!)
-    let modifiedURLRequest = URLRequest(url: URL(string: "https://bar.foo")!)
-    let secondModifiedURLRequest = URLRequest(url: URL(string: "https://bar.foo.bar")!)
+    /// We use unsupported URLs so that the task goes through its URLProtocol chain and **immediately** returns with "unsupported URL" error
+    let mockURL = URL(string: "foo://example.com")!
```
ncreated, line 13: Is this a documented behaviour of `URLSession` or our observation?

buranmert: going through `URLProtocol` chain is documented in `URLProtocol`, the rest is observation
but "immediately" may be misunderstood: i mean "without making a connection attempt"

ncreated: Ok, sounds good enough. Just a note - if using `ServerMock` which mocks the `URLProtocol` we wouldn't have to explain anything in the comment 😉.

buranmert: yes, `URLProtocol` way could be nice but it doesn't work with `URLSession.shared`

ncreated: Yup, but we don't need to unit-test `URLSession.shared`, any other session created within the app will be a subject for interception (through swizzling), so `server.session` may also work. We cover `URLSession.shared` in integration tests, so it's safe at any front.

Anyway, that's just an idea, doesn't change the fact it works. Good to go for me 👌.

buranmert: sure, we could do that 👌 

the idea of testing `URLSession.shared/(configuration:)/(config:delegate:queue:)` came when i realized those methods return private `URLSession` subclasses, they may return different types one day (i admit this is a bit of over-caution ⚠️ )
as you said, we've been safe since we covered `.shared` in integration, now we can change if needed 👍

PR #163, opened 2020-07-03
RUMM-565 Make integration tests execute faster

nachoBonafonte:
### What and why?

 This PR avoid waiting more than necessary in integration tests, so can be run faster

### How?

Creates a method in ServerSession to actively check for a fixed number of recorded POST requests. This method waits for a specific number of recorded post requests and returns, or timeouts and returns existing recorded requests 

This method is used in some integration tests that were waiting a fixed time (higher than ecessary)

### Review checklist

### DIFF
--- a/Tests/DatadogIntegrationTests/TracingIntegrationTests.swift
+++ b/Tests/DatadogIntegrationTests/TracingIntegrationTests.swift
@@ -32,11 +32,9 @@ class TracingIntegrationTests: IntegrationTests {
         )
         app.tapSendTracesForUITests()
 
-        // Wait for delivery
-        Thread.sleep(forTimeInterval: Constants.dataDeliveryTime)
+        // Return desired count or timeout
+        let recordedTracingRequests = try tracingServerSession.pullRecordedPOSTRequests(count: 1, timeout: Constants.dataDeliveryTime)
 
-        // Assert tracing requests
-        let recordedTracingRequests = try tracingServerSession.getRecordedPOSTRequests()
         recordedTracingRequests.forEach { request in
             // Example path here: `/36882784-420B-494F-910D-CBAC5897A309/ui-tests-client-token?batch_time=1589969230153`
             let pathRegexp = #"^(.*)(/ui-tests-client-token\?batch_time=)([0-9]+)$"#
--- a/Tests/DatadogIntegrationTests/TracingIntegrationTests.swift
+++ b/Tests/DatadogIntegrationTests/TracingIntegrationTests.swift
@@ -52,7 +52,7 @@ class TracingIntegrationTests: IntegrationTests {
         let autoTracedWithRequest = spanMatchers[4]
         let autoTracedWithError = spanMatchers[5]
 
-        let recordedNetworkRequests = try dataSourceServerSession.getRecordedPOSTRequests()
+        let recordedNetworkRequests = try dataSourceServerSession.pullRecordedPOSTRequests(count: 1, timeout: Constants.dataDeliveryTime)
         XCTAssert(recordedNetworkRequests.count == 1)
         let traceID = try! autoTracedWithRequest.traceID().hexadecimalNumberToDecimal
         XCTAssert(recordedNetworkRequests.first!.httpHeaders.contains("x-datadog-trace-id: \(traceID)"), "Trace: \(traceID) Actual: \(recordedNetworkRequests.first!.httpHeaders)")
@@ -165,7 +165,7 @@ class TracingIntegrationTests: IntegrationTests {
         }
 
         // Assert logs requests
-        let recordedLoggingRequests = try loggingServerSession.getRecordedPOSTRequests()
+        let recordedLoggingRequests = try loggingServerSession.pullRecordedPOSTRequests(count: 1, timeout: Constants.dataDeliveryTime)
 
         // Assert logs
         let logMatchers = try recordedLoggingRequests
--- a/instrumented-tests/http-server-mock/Sources/HTTPServerMock/ServerSession.swift
+++ b/instrumented-tests/http-server-mock/Sources/HTTPServerMock/ServerSession.swift
@@ -44,4 +44,32 @@ public class ServerSession {
                 )
             }
     }
+
+    /// Actively fetches 'POST` requests recorded by the server until a desired count is found, or timeouts returning current recorded requests
+    public func pullRecordedPOSTRequests(count: Int, timeout: TimeInterval) throws -> [POSTRequestDetails] {
+        var currentRequests = [ServerMock.RequestInfo]()
+
+        let timeoutTimer = DispatchSource.makeTimerSource(queue: DispatchQueue.global())
+        timeoutTimer.setEventHandler { timeoutTimer.cancel() }
+        timeoutTimer.schedule(deadline: .now() + timeout, leeway: .nanoseconds(0))
+        if #available(iOS 10.0, *) {
+            timeoutTimer.activate()
+        }
+
+        while !timeoutTimer.isCancelled && currentRequests.count < count {
+            Thread.sleep(forTimeInterval: 0.2)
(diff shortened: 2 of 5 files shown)


instrumented-tests/http-server-mock/Tests/HTTPServerMockTests/HTTPServerMockTests.swift
```diff
@@ -52,6 +52,69 @@ final class HTTPServerMockTests: XCTestCase {
...
+        let recordedRequests = try session.pullRecordedPOSTRequests(count: 2, timeout: timeoutTime)
+        XCTAssertLessThan(Date(), initialTime.addingTimeInterval(timeoutTime))
+        XCTAssertEqual(recordedRequests.count, 2)
+        XCTAssertTrue(recordedRequests[0].path.hasSuffix("/resource/1"))
+        XCTAssertEqual(recordedRequests[0].httpBody, "1st request body".data(using: .utf8)!)
+        XCTAssertTrue(recordedRequests[1].path.hasSuffix("/resource/2"))
+        XCTAssertEqual(recordedRequests[1].httpBody, "2nd request body".data(using: .utf8)!)
+    }
+    
+    func testItReturnspullRecordedPOSTRequestsTimedOut() throws {
```
ncreated, line 89: How about?
```suggestion
    func testWhenPullingRecordedPOSTRequestExceedsTimeout_itThrownsAnError() throws {
```

instrumented-tests/http-server-mock/Sources/HTTPServerMock/ServerSession.swift
```diff
@@ -44,4 +44,32 @@ public class ServerSession {
...
+        }
+
+        while !timeoutTimer.isCancelled && currentRequests.count < count {
+            Thread.sleep(forTimeInterval: 0.2)
+            currentRequests = try server
+                .getRecordedPOSTRequestsInfo()
+                .filter { requestInfo in requestInfo.path.contains(sessionIdentifier) }
+        }
+        timeoutTimer.cancel()
+
```
ncreated, line 66: How about throwing an exception from this method if we don't get what we expect:
```swift
throw ServerSessionException("Exceeded \(timeout)s timeout by receiving only \(currentRequests.count) requests, where \(count) were expected.")
```
This way, on the caller side (in integration tests), the execution will be stopped due to inconsistency. Otherwise, the execution will continue and we'll be accessing requests on indexes which do not exist, resulting with tests crash, instead of failure.

WDYT @nachoBonafonte ?

Tests/DatadogIntegrationTests/TracingIntegrationTests.swift
```diff
@@ -32,11 +32,9 @@ class TracingIntegrationTests: IntegrationTests {
         )
         app.tapSendTracesForUITests()
 
-        // Wait for delivery
-        Thread.sleep(forTimeInterval: Constants.dataDeliveryTime)
+        // Return desired count or timeout
+        let recordedTracingRequests = try tracingServerSession.pullRecordedPOSTRequests(count: 1, timeout: Constants.dataDeliveryTime)
 
-        // Assert tracing requests
-        let recordedTracingRequests = try tracingServerSession.getRecordedPOSTRequests()
```
ncreated, line 39: There are also two other places in this `TracingIntegrationTests` where we use `.getRecordedPOSTRequests()` - let's change them to `.pullRecordedPOSTRequests()` 💪.

Explanation: we use 3 different server sessions in this test:
* one for recording traces,
* one for recording logs (send by `span.log()`),
* one for recording data request which should be auto-instrumented with trace propagation HTTP headers.

instrumented-tests/http-server-mock/Sources/HTTPServerMock/ServerSession.swift
```diff
@@ -44,4 +44,32 @@ public class ServerSession {
...
+    /// Actively fetches 'POST` requests recorded by the server until a desired count is found, or timeouts returning current recorded requests
+    public func pullRecordedPOSTRequests(count: Int, timeout: TimeInterval) throws -> [POSTRequestDetails] {
+        var currentRequests = [ServerMock.RequestInfo]()
+
+        let timeoutTimer = DispatchSource.makeTimerSource(queue: DispatchQueue.global())
+        timeoutTimer.setEventHandler { timeoutTimer.cancel() }
+        timeoutTimer.schedule(deadline: .now() + timeout, leeway: .nanoseconds(0))
+        if #available(iOS 10.0, *) {
+            timeoutTimer.activate()
+        }
```
buranmert, line 57: is timer automatically activated in macOS? why do we need `if #available`?

instrumented-tests/http-server-mock/Sources/HTTPServerMock/ServerSession.swift
nachoBonafonte: Its a great idea, I will change to it

Tests/DatadogIntegrationTests/TracingIntegrationTests.swift
nachoBonafonte: Yes, I saw them, but they were not waiting a fixed time before calling `.getRecordedPOSTRequests()` thats why I didn't change them, should I?

ncreated: Yeah, it works, because data send to that sessions is delivered before data send to `tracingServerSession`. But this is an implicit implementation-based assumption, so I think by using `.pullRecordedPOSTRequests()` we will make it more explicit. Anyway, it will still keep working, so both ways are good for me 👍.

instrumented-tests/http-server-mock/Sources/HTTPServerMock/ServerSession.swift
nachoBonafonte: No, the HTTP mock server should only run on macOS not on iOS, I changed the package for that and also for a minimum macOS version. But Xcode still shows iOS as a destination for the Scheme, so I added that line to avoid the Xcode error if you just try to build the library and you have an iPhone selected in Xcode.

ncreated:
Hmm, I also checked the result on CI 🤔.

This branch:
Screenshot 2020-07-03 at 12 08 07

Last build on `master`:
Screenshot 2020-07-03 at 12 08 13

Shouldn't it be more significant win? Does it run for ~30 seconds also locally?

nachoBonafonte:
Yes, locally the call is taking ~25 seconds, and we have added two tests more that take ~3 seconds both

Tests/DatadogIntegrationTests/TracingIntegrationTests.swift
nachoBonafonte: Ok, the I will change them also

instrumented-tests/http-server-mock/Tests/HTTPServerMockTests/HTTPServerMockTests.swift
```diff
@@ -111,9 +110,11 @@ final class HTTPServerMockTests: XCTestCase {
...
         
         let timeoutTime: TimeInterval = 2
-        let recordedRequests = try session.pullRecordedPOSTRequests(count: 2, timeout: timeoutTime)
-        XCTAssertLessThan(recordedRequests.count, 2)
-        XCTAssertGreaterThan(Date(), initialTime.addingTimeInterval(timeoutTime))
+        var thrownError: Error?
+        XCTAssertThrowsError(try session.pullRecordedPOSTRequests(count: 2, timeout: timeoutTime)) {
+            thrownError = $0
+        }
+        XCTAssertTrue( thrownError is ServerSession.Exception, "Unexpected error type: \(type(of: thrownError))")
```
ncreated, line 117:
```suggestion
        XCTAssertTrue(thrownError is ServerSession.Exception, "Unexpected error type: \(type(of: thrownError))")
```

PR #164, opened 2020-07-03
RUMM-266 Use setUpWithError or tearDownWithError in tests

nachoBonafonte:
### What and why?

Use `setUpWithError()` or `tearDownWithError()` in test setups that can throw, it allows properly handling of errors
Clean all `try!` uses in the test code since can be handled as error instead of crashes

### How?

Just replaced  `setUp()` with `setUpWithError()` and `tearDown()`  with `tearDownWithError()`  in those methods that were using methods that could throw

### Review checklist


nachoBonafonte:
I just changed some of the methods (those that could throw), should I have changed all the setup methods in project?

PR #167, opened 2020-07-07
RUMM-343 Automate sdk version releases

nachoBonafonte:
### What and why?

Automate changing the version of the library

### How?

Added a `make newversion` that modifies version for the library and in podspecs
After modifiying those files commits the changes to the repo with a message

### Review checklist


Makefile
```diff
@@ -30,3 +30,13 @@ test-carthage:
...
 		@cd dependency-manager-tests/cocoapods && $(MAKE)
+
+newversion:
+		@read -p "Enter version number: " version;  \
+		echo "// GENERATED FILE: Do not edit directly\n\ninternal let sdkVersion = \"$$version\"" > Sources/Datadog/Versioning.swift; \
+		sed "s/__DATADOG_VERSION__/$$version/g" DatadogSDK.podspec.meta > DatadogSDK.podspec; \
+		sed "s/__DATADOG_VERSION__/$$version/g" DatadogSDKObjc.podspec.meta > DatadogSDKObjc.podspec; \
+		git add . ; \
+		git commit -m "Bumped version to $$version"; \
+		git push -u origin master; \
```
buranmert, line 41: i'm not sure if that'll work, `master` is protected
i think we should do the push explicitly/manually

Makefile
```diff
@@ -1,5 +1,5 @@
 all: tools dependencies xcodeproj-httpservermock templates
-.PHONY : examples tools
+.PHONY : examples tools newversion
```
buranmert, line 2: this is probably not needed, this `phony` thing is needed for the folders with those names such as `examples` and `tools`

Makefile
buranmert: maybe we can split bumping and shipping into 2 tasks
the code above can be `bump`
and we can push to `pod trunk` and Github with `ship`

what do you think?

nachoBonafonte: it sounds good, I will work on it

Makefile
```diff
@@ -30,3 +30,13 @@ test-carthage:
 # Tests if current branch ships a valid Cocoapods project.
 test-cocoapods:
 		@cd dependency-manager-tests/cocoapods && $(MAKE)
+
+newversion:
+		@read -p "Enter version number: " version;  \
+		echo "// GENERATED FILE: Do not edit directly\n\ninternal let sdkVersion = \"$$version\"" > Sources/Datadog/Versioning.swift; \
+		sed "s/__DATADOG_VERSION__/$$version/g" DatadogSDK.podspec.meta > DatadogSDK.podspec; \
+		sed "s/__DATADOG_VERSION__/$$version/g" DatadogSDKObjc.podspec.meta > DatadogSDKObjc.podspec; \
```
ncreated, line 38: We use string replacement in [dependency-manager-tests](https://github.com/DataDog/dd-sdk-ios/tree/master/dependency-manager-tests), but the convention was to name meta files as `.src`. Could we do it here as well? 

* `DatadogSDK.podspec.src`
* `DatadogSDKObjc.podspec.src`

Sources/Datadog/Versioning.swift
```diff
@@ -0,0 +1,3 @@
+// GENERATED FILE: Do not edit directly
```
ncreated, line 1: The CI fails on missing license header in this file:
```
🔥 Missing the license header in files: 
./Sources/Datadog/Versioning.swift
```

Let's exclude this file in [check-license.sh](https://github.com/DataDog/dd-sdk-ios/blob/master/tools/license/check-license.sh) 👍.

Makefile
nachoBonafonte: Sure

Sources/Datadog/Versioning.swift
nachoBonafonte: yep, thanks for the tip

Makefile
nachoBonafonte: ok, i will clean also the examples one, since there is not target for examples anymore

nachoBonafonte:
I removed all the push logic, I dont really what to do about it if the branch is blocked

Makefile
```diff
@@ -30,3 +30,19 @@ test-carthage:
...
 		@cd dependency-manager-tests/cocoapods && $(MAKE)
+
+bump:
+		@read -p "Enter version number: " version;  \
+		echo "// GENERATED FILE: Do not edit directly\n\ninternal let sdkVersion = \"$$version\"" > Sources/Datadog/Versioning.swift; \
+		sed "s/__DATADOG_VERSION__/$$version/g" DatadogSDK.podspec.src > DatadogSDK.podspec; \
+		sed "s/__DATADOG_VERSION__/$$version/g" DatadogSDKObjc.podspec.src > DatadogSDKObjc.podspec; \
+		git add . ; \
+		git commit -m "Bumped version to $$version"; \
+		git tag $$version; \
```
ncreated, line 41: Let's not add `tag` 🙌 - we do releases through GH > "New Release" UI and it provides the option to add tag + mark it as pre-release (plus a nice UI for crafting release notes 🙂). If we push tag from `make bump`, but have to fix something in podspecs for `make ship`, this will lead to inconsistency. Doing pre-release + tag through GH UI, then calling `make ship` and then changing from pre-release to regular release guarantees consistency.

tools/license/check-license.sh
```diff
@@ -21,7 +21,8 @@ function files {
 		-not -name "OTFormat.swift" \
 		-not -name "OTTracer.swift" \
 		-not -name "OTReference.swift" \
-		-not -name "OTSpanContext.swift"
+		-not -name "OTSpanContext.swift" \
+        -not -name "Versioning.swift"
```
ncreated, line 25:
```suggestion
		-not -name "Versioning.swift"
```

Makefile
nachoBonafonte: ok, removed that line

tools/license/check-license.sh
nachoBonafonte: ups, it looked well tabulated on my editor

PR #168, opened 2020-07-08
RUMM-563 Generate api-surface-swift and api-surface-objc

ncreated:
### What and why?

📦 This PR adds `api-surface-swift` and `api-surface-objc` files, both listing the `public` interfaces from SDK.

### How?

I created a command-line tool, `api-surface`. It uses [SourceKitten](https://github.com/jpsim/SourceKitten) package to generate documentation for SDK and parses its intermediate output to only list interesting things from publicly available interface.

This tool is used in `Makefile`, so we can re-create api-surfaces using `make api-surface`:
```make
# Makefile

# Generate api-surface files for Datadog and DatadogObjc.
api-surface:
   @cd tools/api-surface/ && swift build --configuration release
   @echo "Generating api-surface-swift"
   ./tools/api-surface/.build/x86_64-apple-macosx/release/api-surface workspace --workspace-name Datadog.xcworkspace --scheme Datadog --path . > api-surface-swift
   @echo "Generating api-surface-objc"
   ./tools/api-surface/.build/x86_64-apple-macosx/release/api-surface workspace --workspace-name Datadog.xcworkspace --scheme DatadogObjc --path . > api-surface-objc
```

### Running `api-surface` on CI

A nice extension to api-surface workflow would be generating surfaces on CI and checking if they are up to date. However, I don't include this in this PR, as `make api-surfaces` needs almost `2 minutes` on my machine on cold run, and `29s` if `swift build` was already cached. We may think of [caching build artifacts between Bitrise CI jobs](https://devcenter.bitrise.io/builds/caching/about-caching-index/) OR shipping `api-surface` as binary to make it faster.

### Review checklist

### DIFF
--- a/LICENSE-3rdparty.csv
+++ b/LICENSE-3rdparty.csv
@@ -1,2 +1,4 @@
 Component,Origin,License,Copyright
 import,io.opentracing,MIT,Copyright 2018 LightStep
+import (tools),https://github.com/jpsim/SourceKitten,MIT,Copyright (c) 2014 JP Simard
+import (tools),https://github.com/apple/swift-argument-parser,Apache-2.0,(c) 2020 Apple Inc. and the Swift project authors
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,14 @@ test-carthage:
 test-cocoapods:
 		@cd dependency-manager-tests/cocoapods && $(MAKE)
 
+# Generate api-surface files for Datadog and DatadogObjc.
+api-surface:
+		@cd tools/api-surface/ && swift build --configuration release
+		@echo "Generating api-surface-swift"
+		./tools/api-surface/.build/x86_64-apple-macosx/release/api-surface workspace --workspace-name Datadog.xcworkspace --scheme Datadog --path . > api-surface-swift
+		@echo "Generating api-surface-objc"
+		./tools/api-surface/.build/x86_64-apple-macosx/release/api-surface workspace --workspace-name Datadog.xcworkspace --scheme DatadogObjc --path . > api-surface-objc
+
 bump:
 		@read -p "Enter version number: " version;  \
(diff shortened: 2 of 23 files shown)


buranmert:
we actually need the `api-surface` tool itself, not its source code and tests: so i agree on your point of keeping it as binary in the repo. we/you can even create its own repo of `api-surface` tool and we can use its binary in this repo.
we can do the same for `HTTPServerMock` 💡 

given Bitrise is already pretty slow, we can generate `api-surface` with a local `pre-push` hook: every time we push, hook can run the binary `./tools/api-surface --some parameters` and it can reject push operation if `api-surface.txt` is changed
that way we can ensure `api-surface` is always up-to-date & Bitrise not having more workload

wdyt?

ncreated:
> we actually need the api-surface tool itself, not its source code and tests: so i agree on your point of keeping it as binary in the repo. we/you can even create its own repo of api-surface tool and we can use its binary in this repo.
we can do the same for HTTPServerMock 💡

That's the plan 👌, I'd love to share `HTTPServerMock` and `api-surface` with the community, but it requires some work to move those things to separate repos, configure CI, link it back in `dd-sdk-ios` etc. I'll discuss availability for this work.

> given Bitrise is already pretty slow, we can generate api-surface with a local pre-push hook: every time we push, hook can run the binary ./tools/api-surface --some parameters and it can reject push operation if api-surface.txt is changed
that way we can ensure api-surface is always up-to-date & Bitrise not having more workload

Surface generation takes `~30s` on my laptop and this would be painful to have each push delayed for that amount of time. I think the first idea of moving it to a separate repo and running on CI is much better. Even if it adds this `~30s` to the CI job time, we may play around and optimise it (e.g. only running when merging to `master` with a Slack notification reporting failure OR getting additional CI runner for running sanity checks like lint, license and api-surface).

ncreated:
@test-all-platforms 🤞

ncreated:
Testing CI: @test-all-platforms ncreated/RUMM-563-generate-api-surface

PR #169, opened 2020-07-08
RUMM-486 Fix `test-all-platforms` trigger

ncreated:
### What and why?

📦 This PR fixes the `test-all-platforms.yml` CI job to pick the right git branch when triggered from a PR comment.

### How?

The GH workflow was configured to either:
* run if `@test-all-platforms` is mentioned in PR description,
* run if `@test-all-platforms` is mentioned in PR comment.

The first scenario used to work fine and is not changed 👌. The second scenario was buggy: because of how GH `issue_comment` trigger works, the workflow was always triggered on the `master` branch, resulting with CI picking the `master` for its job.

I changed the behaviour of `Trigger Bitrise build` step to get the branch name from PR comment. **From now, the expected syntax for running `test-all-platforms.yml` CI job from PR comment is**:
```
@test-all-platforms <branch-name>
```
e.g.:
```
"@test-all-platforms ncreated/RUMM-486-fix-all-platform-tests"
```

I added regexp to get the branch name from comment, so this is also allowed and should work:
```
"Please 🙏 @test-all-platforms ncreated/RUMM-486-fix-all-platform-tests thank you!"
```

### Testing

I tested the script locally as much as possible, but again, we can only verify if it works, after this is merged to `master` 🤞.

### Review checklist

~~- [] Feature or bugfix MUST have appropriate tests (unit, integration)~~

### DIFF
--- a/.github/workflows/all-platform-tests.yml
+++ b/.github/workflows/all-platform-tests.yml
@@ -14,7 +14,7 @@ jobs:
 
     runs-on: ubuntu-latest
     steps:
-      - name: Look for a keyword triggering Bitrise build
+      - name: Look for a keyword triggering the build on Bitrise
         id: check-keyword-trigger
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


ncreated:
Huh, let's @test-all-platforms

PR #173, opened 2020-07-10
RUMM-571 Let users add global tags on Tracer

nachoBonafonte:
### What and why?

It let users add global tags on Tracer

### How?

The desired tags must be set set in TracerConfiguration before initializing the tracer and will be copied to all created spans

### Review checklist

### DIFF
--- a/Sources/Datadog/TracerConfiguration.swift
+++ b/Sources/Datadog/TracerConfiguration.swift
@@ -18,15 +18,20 @@ extension Tracer {
         /// - Parameter enabled: `false` by default
         public var sendNetworkInfo: Bool
 
+        /// Tags that will be added to all new spans created by the tracer
+        public var globalTags: [String: Encodable]?
+
         /// Initializes the Datadog Tracer configuration.
         /// - Parameter serviceName: the service name that will appear in traces (if not provided or `nil`, the SDK default `serviceName` will be used).
         /// - Parameter sendNetworkInfo: adds network connection info to every span and span logs (`false` by default).
         public init(
             serviceName: String? = nil,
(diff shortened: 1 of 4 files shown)


Sources/Datadog/TracerConfiguration.swift
```diff
@@ -18,15 +18,20 @@ extension Tracer {
         /// - Parameter enabled: `false` by default
         public var sendNetworkInfo: Bool
 
+        /// Tags that will be added to all new spans created by the tracer
```
ncreated, line 21:
```suggestion
        /// Tags that will be added to all new spans created by the tracer.
```

Tests/DatadogTests/Datadog/TracerTests.swift
```diff
@@ -108,6 +108,33 @@ class TracerTests: XCTestCase {
...
+            configuration: .init(
+                serviceName: "custom-service-name",
+                globalTags: [
+                    "globaltag1": "globalValue1",
+                    "globaltag2": "globalValue2"
+                ]
+            )
+        )
+
+        let span = tracer.startSpan(operationName: .mockAny())
```
ncreated, line 129: We could also test that tags are properly overwritten with span-local tag:
```swift
span.setTag(key: "globaltag2", value: "overwritten value")

// ...

XCTAssertEqual(try spanMatcher.meta.custom(keyPath: "meta.globaltag2"), "overwritten value")
```
WDYT?

nachoBonafonte: Good idea, it covers the logic better. Changed

PR #174, opened 2020-07-10
RUMM-515 RUM data is uploaded to server

ncreated:
### What and why?

📦 This PR configures `RUMFeature` so it can upload data to RUM Intake

### How?

I added `RUMFeature` with `Storage` and `Upload` stacks and configured them to use the expected RUM HTTP message format.

RUM uses `text/plain;charset=UTF-8` content type with RUM events encoded as newline-separated JSONs:
```http
POST <rum-intake-url>/<access-token>?ddsource=ios&batch_time=<timestamp in milliseconds>&ddtags=service:<service-name>,version:<app-version>,sdk_version:<sdk-version>,env:<environment-name>
Content-Type: text/plain;charset=UTF-8

rumEventJSON1
rumEventJSON2
rumEventJSON3
```

Except unit tests, to additionally test if all works fine, I've send a fake data to RUM Explorer (I will add this debug utility in a next PR, so we can use it when iterating further on RUM feature):

Screenshot 2020-07-10 at 09 28 03

### Review checklist

### DIFF
--- a/Sources/Datadog/Datadog.swift
+++ b/Sources/Datadog/Datadog.swift
@@ -117,13 +117,6 @@ public class Datadog {
                 performance: performance,
                 mobileDevice: mobileDevice,
                 httpClient: httpClient,
-                logsUploadURLProvider: UploadURLProvider(
-                    urlWithClientToken: validConfiguration.logsUploadURLWithClientToken,
-                    queryItemProviders: [
-                        .ddsource(),
-                        .batchTime(using: dateProvider)
-                    ]
-                ),
                 dateProvider: dateProvider,
                 userInfoProvider: userInfoProvider,
                 networkConnectionInfoProvider: networkConnectionInfoProvider,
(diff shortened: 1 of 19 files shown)


Tests/DatadogTests/Datadog/Tracing/TracingFeatureTests.swift
```diff
@@ -99,6 +86,13 @@ class TracingFeatureTests: XCTestCase {
 
         let payload = server.waitAndReturnRequests(count: 1)[0].httpBody!
 
+        // Expected payload format:
```
mariusc83, line 89: Is this asserted somehow ?

ncreated: Yes, the `SpanMatcher's` expects exactly this format in: `try SpanMatcher.fromNewlineSeparatedJSONObjectsData (payload)`. It throws an error if the array of `SpanMatchers` cannot be build.

ncreated: Same will be done later for RUM payload:
```swift
let rumMatchers = try RUMEventMatcher.fromNewlineSeparatedJSONObjectsData(payload)
```
For now I build RUM payload assertions around the raw data, because we don't yet have `RUMEventMatcher` ready.

Sources/Datadog/Datadog.swift
```diff
@@ -152,8 +140,24 @@ public class Datadog {
...
+                networkConnectionInfoProvider: networkConnectionInfoProvider,
+                carrierInfoProvider: carrierInfoProvider
+            )
+        }
+
         LoggingFeature.instance = logging
         TracingFeature.instance = tracing
+        RUMFeature.instance = rum
+
         TracingAutoInstrumentation.instance = TracingAutoInstrumentation(with: configuration)
```
buranmert, line 161: can you please add `configuration.rumEndpoint.url` to `excludedURLs` within this method?
you will then need to fix a test case in `TracingAutoInstrumentationTests` i guess

ncreated: Ah, sure - done 👍. Good catch 👌.

PR #175, opened 2020-07-10
RUM-515 Add "Debug RUM" screen in Example app

ncreated:
### What and why?

🚚 This PR adds "Debug RUM" screen in Example app. It can be used to send events to RUM Explorer:

Screenshot 2020-07-10 at 11 45 30

### How?

As RUM data upload was done in #174, in this PR I just pass the `RUMViewEvent` encodable model to RUM's `fileWriter`, so it gets uploaded by the RUM's upload worker.

To not block RUM development by _"`RUM-517` RUM data models are auto generated"_, we can keep crafting `RUMDataModels` manually (as I did for `RUMViewEvent`). After `RUM-517` is done, we will just replace them.

#### `Datadog.xcconfig`

For RUM, we have to update the `.xcconfig` file. Now, next to `DATADOG_CLIENT_TOKEN` it also takes `RUM_APPLICATION_ID`:
```
RUM_APPLICATION_ID=// use your own RUM Application ID obtained on datadoghq.com
DATADOG_CLIENT_TOKEN=// use your own Client Token, generated for RUM_APPLICATION_ID
```

RUM application needs a separate `DATADOG_CLIENT_TOKEN`, but this token works also for Logging and Tracing. Both, application ID and token can be found in `UX Monitoring > RUM Applications > ios-sdk-example-app > Edit Application`.

### Review checklist

### DIFF
--- a/Sources/Datadog/RUMMonitor.swift
+++ b/Sources/Datadog/RUMMonitor.swift
@@ -1,0 +1,35 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+public class RUMMonitor {
+    private let rumApplicationID: String
+
+    public init(rumApplicationID: String) {
+        self.rumApplicationID = rumApplicationID
+    }
+
+    /// TODO: RUMM-585 Replace with real RUMMonitor public API
+    public func sendFakeViewEvent(viewURL: String) {
+        let event = RUMViewEvent(
+            date: Date(timeIntervalSinceNow: -1).timeIntervalSince1970.toMilliseconds,
+            application: .init(id: rumApplicationID),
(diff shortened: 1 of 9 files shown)


Sources/Datadog/RUMMonitor.swift
```diff
@@ -0,0 +1,35 @@
...
+
+public class RUMMonitor {
+    private let rumApplicationID: String
+
+    public init(rumApplicationID: String) {
+        self.rumApplicationID = rumApplicationID
+    }
+
+    /// TODO: RUMM-585 Replace with real RUMMonitor public API
+    public func sendFakeViewEvent(viewURL: String) {
```
buranmert, line 17: can't we `@testable import Datadog` in `Example` app instead of having this public method in `master` branch?

ncreated: This won't be merged to `master`.

`@testable import` is disabled in Example app's project - it hosts integration tests, and we want those tests to only use public APIs. IMO exposing one public method on a `rum` feature branch is better & easier than changing the project configuration for debugging purpose. Please also note - it's clearly marked with `TODO: RUMM-585` and will never get into `master` before it gets resolved.

Issue #177, opened 2020-07-10
App Store Submit Issues


hyling:
### The issue
When I submit my app to the App Store I get the following errors:

- App Store Connect Operation Error
 ERROR ITMS-90685: "CFBundleIdentifier Collision. There is more than one bundle with the CFBundleIdentifier value 'com.datadogqh.Datadog' under the iOS application 'XXX.app'."
- App Store Connect Operation Error
ERROR ITMS-90205: "Invalid Bundle. The bundle at 'XXX.app/Frameworks/DatadogObjc.framework' contains disallowed nested bundles."
- App Store Connect Operation Error
 ERROR ITMS-90206: "Invalid Bundle. The bundle at 'XXX.app/Frameworks/DatadogObjc.framework' contains disallowed file 'Frameworks'."
#### Datadog SDK version:
1.2.2
#### Last working Datadog SDK version:
NA
#### Dependency Manager:
Carthage
#### Other toolset:
NA
#### Xcode version:
11.3.0
#### Swift version:
5.0
#### Deployment Target:
iOS 11, iPhone
#### macOS version:
10.14.5

ncreated:
Hello @hyling 👋. Thank you once again for reporting us the issue and sorry for troubles. This should be fixed in #182. I will let you know once the hotfix is released.

ncreated:
Fixed in [`1.2.3`](https://github.com/DataDog/dd-sdk-ios/releases/tag/1.2.3) ✅

hyling:
Awesome!  Thanks for the quick fixes.

PR #182, opened 2020-07-15
RUMM-611 Hotfix App Store submission issues for `DatadogObjc`

ncreated:
### What and why?

🐛 This PR is addressing #177 where App Store Connect validation issues are reported when uploading the app that links `DatadogObjc` with Carthage:

Screenshot 2020-07-15 at 11 11 03

### How?

The reason of duplicated framework and colliding `com.datadoghq.Datadog` bundle identifier was a mistaken configuration of `DatadogObjc` "Frameworks and Libraries", where `Datadog.framework` was embedded and signed.

I changed it to **not embed** the `Datadog.framework` (as Carthage users do embed it through Carthage script):

Screenshot 2020-07-15 at 11 22 42

I made sure that Objective-C app linking this branch with Carthage passes the ASC validation:

Screenshot 2020-07-15 at 11 22 10

### Review checklist

### DIFF
--- a/Datadog/Datadog.xcodeproj/project.pbxproj
+++ b/Datadog/Datadog.xcodeproj/project.pbxproj
@@ -82,7 +82,6 @@
 		61133C6D2423990D00786299 /* TestsDirectory.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61133C462423990D00786299 /* TestsDirectory.swift */; };
 		61133C6E2423990D00786299 /* DatadogExtensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61133C472423990D00786299 /* DatadogExtensions.swift */; };
 		61133C702423993200786299 /* Datadog.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 61133B82242393DE00786299 /* Datadog.framework */; };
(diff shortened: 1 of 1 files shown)


buranmert:
1. even though there are 2 different `Info.plist`s and bundle identifiers, it collides if framework is embedded?
2. how can we declare `DDObjC` depends on `DD` for `carthage` users as it requires user action now? maybe a mention in `README` or docs?

ncreated:
@buranmert 

> 1. even though there are 2 different Info.plists and bundle identifiers, it collides if framework is embedded?

Yes, because `DatadogObjc.framework` was embedding the `Datadog.framework`, the `Datadog.framework` was bundled into `app.ipa` twice:
```
app.ipa
|-- Datadog.framework
|-- DatadogObjc.framework
    |-- Frameworks/Datadog.framework
```

Now it is:
```
|-- Datadog.framework
|-- DatadogObjc.framework
```

> 2. how can we declare DDObjC depends on DD for carthage users as it requires user action now? maybe a mention in README or docs?

`DatadogObjc` has the setting of "Link but do not embed `Datadog.framework`", so it won't compile if `Datadog.framework` is not provided. The [Carthage instruction](https://github.com/Carthage/Carthage#if-youre-building-for-ios-tvos-or-watchos) says: _"drag and drop each framework you want to use from the Carthage/Build folder on disk"_, so this should be obvious for the users to include all Datadog frameworks - at least the reporter of this GH Issue got it right. We don't have docs variant for `DatadogObjc`, so for now there is no good place to explain Objc-related hints.

PR #184, opened 2020-07-15
RUMM-605 - POC: UIViewControllers auto instrumentation

nachoBonafonte:
### What and why?
POC: UIViewControllers auto instrumentation

### How?

Create a `UIViewControllerAutoInstrumentation` class that can be configured to monitor `UIViewController.viewDidLoad()` it uses UIViewControllerSwizzler swizzle `viesDidLoadMethod` and log viewControllers loads depending on Configuration.
By default instrumentation is off, it can be configured in two ways:
	- `.automatic` mode, to instrument all UIViewController's ( it tries to guess  viewcontroller's name)
	-`.userdefined` mode, it only instruments those UIViewController implementing `Instrumentable` protocol, and user can set a name implementing `getInstrumentedName()` or let system guess

### Review checklist


ncreated:
Nice job 👌. IMO, main conclusions are:
* RUM views auto instrumentation will be based on swizzling,
* We can either auto-instrument all the views or give the user an opt-in mechanism to mark key views with `Instrumentable` protocol conformance (I love this idea 👌),
* Most complexity goes into finding the top-most VC to determine which view is "active".

This POC is a great input for the "Views" part of the RUM Auto Instrumentation RFC 👌.

ncreated:
@nachoBonafonte I think we can close this PR and keep while keeping the branch alive 👌.

Issue #178, opened 2020-07-12
Extremely high memory usage resulting in process termination


hyling:
### The crash

Writing lots of logs causes memory usage to spike.  On the first wave the memory usage reached 1.2GB.  On the 2nd wave memory usage reached 1.8 GB then iOS terminated process due to memory issue.  Would it be possible to process the logs in smaller sets and/or using @autoreleasepool?

With Datadog OFF:
![image](https://user-images.githubusercontent.com/577992/87252152-e22f7980-c43e-11ea-8fb8-957a208657cb.png)

Datadog ON:
first wave reaches 1.2GB
![image](https://user-images.githubusercontent.com/577992/87252156-e78cc400-c43e-11ea-8807-1f936500e285.png)

then releases memory
![image](https://user-images.githubusercontent.com/577992/87252166-fc695780-c43e-11ea-8a45-086edc18a041.png)

2nd wave : It reached 1.8GB of memory usage then iOS terminated process due to memory issue
![image](https://user-images.githubusercontent.com/577992/87252176-068b5600-c43f-11ea-8def-664097e5bcc2.png)

---

#### Datadog SDK versions:
1.2.2
#### Last stable Datadog SDK version:
NA
#### Volume:
NA not in production
#### OS version:
13.3.1
#### Deployment Target:
iOS 11, iPhone
#### Device version:
iPhone XR
#### Environment:
Memory pressure

ncreated:
Hey @hyling 👋 , thanks for reaching out. Could you give a bit more details on the way you are performing this heavy logging? Is it more a set of synchronous calls to `logger`, or rather a stream of asynchronous log messages?

hyling:
Hi @ncreated.  Thanks for the response.  This is happening on calls to the logger from a background thread.

ncreated:
@hyling I was able to reproduce the issue 👍 and fix it in https://github.com/DataDog/dd-sdk-ios/pull/180. I will let you know once this is confirmed and when we have a new version ready.

ncreated:
Fixed in [`1.2.3`](https://github.com/DataDog/dd-sdk-ios/releases/tag/1.2.3) ✅

hyling:
Awesome!  Thanks!

hyling:
@ncreated Unfortunately this doesn't fix the issue for me. Memory usage is still hitting 1.8 GB and the process is still getting terminated.  I'll profile with Instruments and provide more info soon.

hyling:
[...]

PR #185, opened 2020-07-16
RUMM-610 Hotfix memory usage when doing intensive logging

ncreated:
### What and why?

🧪 [As reported](https://github.com/DataDog/dd-sdk-ios/issues/178#issuecomment-658884672) in #178, the data upload memory leaks fixed in #181 are not solving out-of-memory issue on very intensive logging. This PR adds necessary optimisation to keep the allocations graph flat.

### How?

I added this benchmarking code in Example app to stress-test logging:
```swift
internal class SendLogsFixtureViewController: UIViewController {
   var objcLogger = DDLogger.builder().build()
   let queue1 = DispatchQueue(label: "intensive-logging-1")
   let queue2 = DispatchQueue(label: "intensive-logging-2")
   let queue3 = DispatchQueue(label: "intensive-logging-3")

   override func viewDidLoad() {
      super.viewDidLoad()

      keepLogging(delay: 0.01, on: queue1)
      keepLogging(delay: 0.02, on: queue2)
      keepLogging(delay: 0.03, on: queue3)
   }

   func keepLogging(delay: TimeInterval, on queue: DispatchQueue) {
      queue.asyncAfter(deadline: .now() + delay) { [weak self] in
         self?.objcLogger.info(createBigMessage())
         self?.keepLogging(delay: 0.01, on: queue)
      }
   }
}

private func createBigMessage() -> String {
    let seed = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789 "
    return (0..<1_024).map({ _ in seed }).joined(separator: "")
}
```

And noticed crazy high number of system allocations coming from `try purgeFilesDirectoryIfNeeded()` in `FilesOrchestrator`. This method is iterating through the list of files in data directory, doing a lot of OS-internal `_FileCache` and `_NSFastEnumerationEnumerator` allocations (out of our control).

Above benchmark results with this allocations graph, leading to out-of-memory crash when iOS process exceeds `1.8GB` limit:

Screenshot 2020-07-16 at 11 48 37

To mitigate this impact, I tuned the `FilesOrchestrator` to call `try purgeFilesDirectoryIfNeeded()` only when necessary - if it knows that a new file will be created (vs. each time when asking for a writable file). This keeps the memory graph flat for the same benchmark:

Screenshot 2020-07-16 at 11 45 57

IMHO, by doing this, we give the OS enough time to reuse `_FileCache` things in a memory-performant way.

### Review checklist

~- [ ] Feature or bugfix MUST have appropriate tests (unit, integration)~


ncreated:
@buranmert 

> when you wait idle for a while after peaking at ~1.5Gb memory, it goes down to normal, correct?

Good question. I did a test - if we stress the `_FileCache` only a bit, it goes back to normal - however, I'd expect a gradual slope, but instead it makes a significant dump:

Screenshot 2020-07-16 at 18 30 43

When we stress it too much (`~>1GB`) it results with random crashes and:
```
2020-07-16 18:32:02.729981+0200 Example[57917:40426048] [ServicesDaemonManager] interruptionHandler is called. -[FontServicesDaemonManager connection]_block_invoke
Example(57917,0x16d61f000) malloc: can't allocate region securely
Example(57917,0x16d61f000) malloc: *** set a breakpoint in malloc_error_break to debug
Example(57917,0x16d61f000) malloc: can't allocate region securely
Example(57917,0x16d61f000) malloc: *** set a breakpoint in malloc_error_break to debug
Example(57917,0x16d61f000) malloc: can't allocate region securely
Example(57917,0x16d61f000) malloc: *** set a breakpoint in malloc_error_break to debug
Example(57917,0x16d61f000) malloc: can't allocate region securely
Example(57917,0x16d61f000) malloc: *** set a breakpoint in malloc_error_break to debug
```

With the optimisation applied in this PR, all remains functional even a long while after excessive logging ends.

> and does that mean that now if we log intensively, SDK may exceed max file size limit with one single file?

SDK starts a new file if the recently used one exceeds `4MB`, so it's fine 👌.

hyling:
@ncreated If it's not too much trouble would you mind describing how you tracked it down to `try purgeFilesDirectoryIfNeeded() ` 🙏 ?   I'm trying to get better at debugging these types of issues.  Which Instruments views did you use?  Did you just use the Allocations instrument?  How did you connect the code in the background thread to your own code.  Instruments doesn't show who started the thread.  🤔   
image

ncreated:
@hyling I used Allocations instrument. When running the benchmark snippet, you can see that `_FileCache` is one of the most popular and consuming allocation on the list. When going into details it gives the stack trace. For most of this allocations it is `try purgeFilesDirectoryIfNeeded()`:

Screenshot 2020-07-17 at 13 28 41

hyling:
> @hyling I used Allocations instrument. When running the benchmark snippet, you can see that `_FileCache` is one of the most popular and consuming allocation on the list. When going into details it gives the stack trace. For most of this allocations it is `try purgeFilesDirectoryIfNeeded()`:

@ncreated Ah thanks for the explanation and the screenshot, they help me realize that debugging symbols for datadog didn't get loaded into Instruments.  After I fixed that the Allocations instrument was much more helpful. 😄

PR #186, opened 2020-07-16
RUMM-526 Add `RUMMonitor` foundations

ncreated:
### What and why?

📦 This PR grounds necessary foundation for RUM Data Collection, including:

- `RUMMonitor` class conforming to `RUMInternal`
- `RUMScope` protocol
- `RUMCommand` enum
- `RUMContext` struct
- `RUMApplicationScope`

### How?

The `RUMMonitor` interface follows the RFC with only one difference: it is not possible to identify resources by `Hashable` value as proposed:
```swift
// Compiler error: Protocol 'Hashable' can only be used as a generic constraint
func start(resource id: Hashable, name: String, ..., attributes: [String: Encodable])
```

Instead, I decided to identify resources by `String` ([as we do in `dd-sdk-android`](https://github.com/DataDog/dd-sdk-android/blob/0153e4211d457d2739e6938e0c7598e9f9f7fe79/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/monitor/DatadogRumMonitor.kt#L77)) which should bring enough flexibility.

### Review checklist

### DIFF
--- a/Sources/Datadog/RUM/RUMMonitor/RUMContext.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/RUMContext.swift
@@ -1,0 +1,21 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+internal struct RUMContext {
+    /// An ID of RUM application.
+    let rumApplicationID: String
+    /// An ID of current RUM session. May change over time.
+    var sessionID: UUID
+
+    /// An ID of currently displayed view.
+    var activeViewID: UUID?
+    /// An URI of currently displayed view.
+    var activeViewURI: String?
+    /// An ID of active user action.
+    var activeUserActionID: UUID?
+}
--- a/Tests/DatadogTests/Datadog/Mocks/RUMFeatureMocks.swift
+++ b/Tests/DatadogTests/Datadog/Mocks/RUMFeatureMocks.swift
@@ -7,6 +7,25 @@
 @testable import Datadog
 
 extension RUMFeature {
+    /// Mocks feature instance which performs no writes and no uploads.
+    static func mockNoOp(temporaryDirectory: Directory) -> RUMFeature {
+        return RUMFeature(
+            directory: temporaryDirectory,
+            configuration: .mockAny(),
+            performance: .combining(storagePerformance: .noOp, uploadPerformance: .noOp),
+            mobileDevice: .mockAny(),
+            httpClient: .mockAny(),
+            dateProvider: SystemDateProvider(),
+            userInfoProvider: .mockAny(),
+            networkConnectionInfoProvider: NetworkConnectionInfoProviderMock.mockWith(
+                networkConnectionInfo: .mockWith(
+                    reachability: .no // so it doesn't meet the upload condition
+                )
+            ),
+            carrierInfoProvider: CarrierInfoProviderMock.mockAny()
+        )
+    }
+
     /// Mocks feature instance which performs uploads to given `ServerMock` with performance optimized for fast delivery in unit tests.
     static func mockWorkingFeatureWith(
         server: ServerMock,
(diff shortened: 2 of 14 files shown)


Tests/DatadogTests/Datadog/RUMMonitorTests.swift
```diff
@@ -0,0 +1,81 @@
...
+
+        monitor.start(resource: "/resource/1", attributes: mockAttributes)
+        monitor.stop(resource: "/resource/1", attributes: mockAttributes)
+        monitor.stop(resource: "/resource/1", withError: ErrorMock(), attributes: mockAttributes)
+
+        monitor.start(userAction: .scroll, attributes: mockAttributes)
+        monitor.stop(userAction: .scroll, attributes: mockAttributes)
+        monitor.add(userAction: .tap, attributes: mockAttributes)
+
+        queue.sync {}
```
buranmert, line 34: this feels like a hack, we don't need to know this `queue`
you could do that as well
```swift
// RUMMonitor
private let queue: DispatchQueue // implementation detail, hidden

// RUMScopeMock
func commandsExpectation(for numberOfCommands: Int) -> XCTestExpectation { ... }

// In tests
wait(for: scope.commandsExpectation(for: 9))
XCTAssertEqual(scope.recordedCommands, [...])
```

Sources/Datadog/RUM/RUMMonitor/RUMContext.swift
```diff
@@ -0,0 +1,21 @@
...
+    /// An ID of current RUM session. May change over time.
+    var sessionID: UUID
+
+    /// An ID of currently displayed view.
+    var activeViewID: UUID?
+    /// An URI of currently displayed view.
+    var activeViewURI: String?
+    /// An ID of active user action.
+    var activeUserActionID: UUID?
+}
```
buranmert, line 21: the main idea is sharing the context among scopes, isn't it?
if we make this a `class`, the same context will be shared and we won't have to compute it every time as shown in RFC, wdyt?

ncreated: The idea is to propagate the context, not "share" it. Sharing (with `class`) won't work, i.e. given this RUM state:
```
RUMMonitor
| -- Application Scope
     |-- Session Scope
         |--- View A Scope
              |--- Resource A1 Scope
              |--- ...  
         |--- View B Scope
              |--- Resource B1 Scope
              |--- ...
```
The `Resource A1 Scope` and `Resource B1 Scope` can't share the same `RUMContext`, as their `activeViewID` and `activeViewURI` will be different (`viewA.id` + `viewA.uri` vs `viewB.id` + `viewB.uri`). Same for `activeUserActionID`: one view may have an active action, while the other doesn't.

Now, the question is _"when a Session Scope may have two active View scopes?"_. There example cases:
* when master - details View Controller hierarchy is presented on iPad,
* when user is instrumenting `viewDidLoad()` / `viewDidDisappear` in his view controllers (the `.startView` and `.stopView` commands will be delivered in a wrong order).

Tests/DatadogTests/Datadog/RUMMonitorTests.swift
ncreated: Done 👍. I did it in the same manner as `waitAndReturn...()`  in `ServerMock`:
```swift
let scope = RUMScopeMock()
// ...
let recordedCommands = scope.waitAndReturnProcessedCommands(count: 9, timeout: 0.5)
```

Tests/DatadogTests/Datadog/Mocks/RUMFeatureMocks.swift
```diff
@@ -105,10 +137,8 @@ class RUMScopeMock: RUMScope {
         activeUserActionID: nil
     )
 
-    var recordedCommands: [RUMCommand] = []
-
     func process(command: RUMCommand) -> Bool {
-        recordedCommands.append(command)
+        queue.async { self.commands.append(command) }
```
buranmert, line 141: i think we should do `expectation.fulfill()` here as well

ncreated: Yes! Added 👌.

PR #187, opened 2020-07-17
RUMM-612 Add active span concept to Tracer

nachoBonafonte:
### What and why?

When a Span is created without parent it assigns the active span of the current execution context if exists, so spans can now inherit from a span even when they initially don't know about it, or keep a reference to it

### How?

The implemantation uses os.activity framework to create activities when the spans are started, and keep a reference to this id in SpanContext.
ActiveSpanUtils keep a map with all the relation activity_id and Spans, where spans are added on creation and removed on finish.
When we want to know the active Span we get the current activity and locate the Span associated with it
Also changed some tests that were failing because spans were not being finished properly and next test was inheriting from previous spans creates

### Review checklist


Sources/Datadog/Tracer.swift
```diff
@@ -139,14 +139,22 @@ public class Tracer: OTTracer {
         return nil
     }
 
+    public var activeSpan: OTSpan? {
```
ncreated, line 142: This property should be also added to `OTTracer` interface, otherwise our users won't see it.

Our users do not use `Tracer` directly. When they register global tracer:
```swift
Global.sharedTracer = Tracer.initialize(configuration: .init())
```
what is available to them through `Global.sharedTracer` is `OTTracer` interface.

Sources/Datadog/Tracing/DDSpanContext.swift
```diff
@@ -5,6 +5,17 @@
...
+
+// Bridging Obj-C variabled defined as c-macroses. See `activity.h` header.
+private
+let OS_ACTIVITY_CURRENT = unsafeBitCast( dlsym( UnsafeMutableRawPointer(bitPattern: -2), "_os_activity_current"), to: os_activity_t.self)
+
+@_silgen_name("_os_activity_create") private
+func _os_activity_create(_ dso: UnsafeRawPointer?,
+                         _ description: UnsafePointer<Int8>,
+                         _ parent: Unmanaged<AnyObject>?,
+                         _ flags: os_activity_flag_t) -> AnyObject!
```
buranmert, line 18: i can't find `activity.h`, can we please add some more comments about this part? why we need it, how it works, etc.?
maybe we can comment the source of this code for more context as well

Sources/Datadog/Tracer.swift
```diff
@@ -139,14 +139,22 @@ public class Tracer: OTTracer {
...
+        return ActiveSpanUtils.getActiveSpan()
+    }
+
     // MARK: - Internal
 
     internal func createSpanContext(parentSpanContext: DDSpanContext? = nil) -> DDSpanContext {
+        var currentParentContext = parentSpanContext
+        if currentParentContext == nil {
+            currentParentContext = activeSpan?.context as? DDSpanContext
```
buranmert, line 151: if we have 2 `Tracer` objects: `Tracer1` & `Tracer2`
if the current active span is `Span1` from `Tracer1`
when we do `Tracer2.createContext(...)`, this code will take `Span1.context` as `parentContext` for `Span2`

is this the expected result?
shouldn't `Tracer1` and `Tracer2` have their own separate `activeSpan`s?

Sources/Datadog/Tracing/DDSpanContext.swift
```diff
@@ -16,6 +27,21 @@ internal struct DDSpanContext: OTSpanContext {
...
+    init(traceID: TracingUUID, spanID: TracingUUID, parentSpanID: TracingUUID?, baggageItems: BaggageItems) {
+        self.traceID = traceID
+        self.spanID = spanID
+        self.parentSpanID = parentSpanID
+        self.baggageItems = baggageItems
+
+        let dso = UnsafeMutableRawPointer(mutating: #dsohandle)
+        let activity = _os_activity_create(dso, "InitDDSpanContext", OS_ACTIVITY_CURRENT, OS_ACTIVITY_FLAG_DEFAULT)
+        activityId = os_activity_get_identifier(activity, nil)
+        os_activity_scope_enter(activity, &activity_state)
```
buranmert, line 42: why do we need `os_activity`?

Sources/Datadog/Tracing/Utils/ActiveSpanUtils.swift
```diff
@@ -0,0 +1,45 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+import os.activity
+
+/// Helper class to get the current Span
+internal struct ActiveSpanUtils {
```
ncreated, line 11: How about making it a real class and assigning this object to a `Tracer`, i.e.:
```swift
public class Tracer: OTTracer {
   // ...
   internal let activeSpansPool = ActiveSpansPool()
}
```
?

This way we won't need to hassle with `XCTAssertNil(ActiveSpanUtils.getActiveSpan())` checks in unit tests, as the pool will be local to the `Tracer` and its lifecycle will follow the tracer's lifecycle (which is already enforced in tests). I should be less error-prone and produce less flakiness.

For further `XCTests` auto instrumentation we could either use `Global.sharedTracer` to issue `XCTest` spans OR if we decide to use a separate instance of `Tracer`, we could pass it the `Global.sharedTracer.activeSpansPool` on initialization.

Tests/DatadogTests/Datadog/Tracing/Utils/ActiveSpanUtilsTest.swift
```diff
@@ -0,0 +1,94 @@
...
+class ActiveSpanUtilsTest: XCTestCase {
+    override func tearDown() {
+        //Confirm that tests have cleaned up spans
+        XCTAssertNil(ActiveSpanUtils.getActiveSpan())
+        super.tearDown()
+    }
+
+    func testsWhenSpanIsStartedIsAssignedToActiveSpan() throws {
+        let tracer = Tracer.mockAny()
+        let previousSpan = ActiveSpanUtils.getActiveSpan()
```
ncreated, line 19: We could use `tracer.activeSpan()` instead of `ActiveSpanUtils.getActiveSpan()` in all this tests. This way we're also testing the public API, instead of internals.

Tests/DatadogTests/Datadog/Tracing/Utils/ActiveSpanUtilsTest.swift
```diff
@@ -0,0 +1,94 @@
...
+            let secondSpan = tracer.startSpan(operationName: .mockAny())
+            XCTAssertEqual(ActiveSpanUtils.getActiveSpan()?.dd.ddContext.spanID, secondSpan.dd.ddContext.spanID)
+            expectation2.fulfill()
+        }
+
+        XCTAssertEqual(ActiveSpanUtils.getActiveSpan()?.dd.ddContext.spanID, oneSpan.dd.ddContext.spanID)
+        waitForExpectations(timeout: 5, handler: nil)
+        oneSpan.finish()
+    }
+}
```
ncreated, line 94: Except unit tests in `ActiveSpanUtilsTests`, how about adding following public api-level test to `TracerTests`?
```swift
func testSendingSpanWithImplicitParent() throws {
    let server = ServerMock(delivery: .success(response: .mockResponseWith(statusCode: 200)))
    TracingFeature.instance = .mockWorkingFeatureWith(
        server: server,
        directory: temporaryDirectory
    )
    defer { TracingFeature.instance = nil }

    let tracer = Tracer.initialize(configuration: .init()).dd
    let queue1 = DispatchQueue(label: "\(#function)-queue1")
    let queue2 = DispatchQueue(label: "\(#function)-queue2")

    let rootSpan = tracer.startSpan(operationName: "root operation")

    queue1.sync {
        let childSpan = tracer.startSpan(operationName: "child operation", childOf: rootSpan.context)
        queue2.sync {
            let grandchildSpan = tracer.startSpan(operationName: "grandchild operation", childOf: childSpan.context)
            grandchildSpan.finish()
        }
        childSpan.finish()
    }

    rootSpan.finish()

    let spanMatchers = try server.waitAndReturnSpanMatchers(count: 3)
    let rootMatcher = spanMatchers[2]
    let childMatcher = spanMatchers[1]
    let grandchildMatcher = spanMatchers[0]

    XCTAssertEqual(try rootMatcher.parentSpanID(), "0")
    XCTAssertEqual(try childMatcher.parentSpanID(), try rootMatcher.spanID())
    XCTAssertEqual(try grandchildMatcher.parentSpanID(), try childMatcher.spanID())
}
```
It tests this feature only using public APIs of the `Tracer` + real request captured by intercepting the `URLSession's` protocol (instead of touching internals of `ActiveSpanUtils`). This is a convention we use for all public components (`Tracer`, `Logger`, `RUMMonitor`).

Sources/Datadog/Tracing/DDSpanContext.swift
nachoBonafonte: Activity tracing is not very much documented, but it allows to keep the context trought the different threads, inheriting parent context when a thread is created, or when some part of code is executed in a separated queue, it keeps the activity that was active when the code is run. 
We map this activities with spans, so we can know which span was active when another was started, even in different queues ( there is a test `testActiveSpanIsKeptPerTask()` where this behavious is shown. In this way you can also know when a network requests return which was the active span that launched the request, not the span that was recently started.

Sources/Datadog/Tracer.swift
nachoBonafonte: Is it possible to have two tracers?, anyway as I said in the previous comment it must be handled by the system because it can internally transmit this information to threads when are internally created

ncreated: Yes, this will mix up the active spans, and it's also something I propose in https://github.com/DataDog/dd-sdk-ios/pull/187#discussion_r456356841 (+ this will bring less hassle in unit tests). But on the other hand, the Open Tracing standard enforces using only one tracer (`Global.sharedTracer`), so having this issue may mean API misuse, but IMO we can overcome all this with having tracer-specific active span pools.

nachoBonafonte:
About the backward compatibility, yes, it should probably be evaluated and check throroughly before adding to master, probably `1.4.0` is a good timeframe. The change to the spans will be related to the span parents, it could probably change results but probably would mean that will improve the information for free (if used the traces properly and finished as they should)

Sources/Datadog/Tracing/DDSpanContext.swift
```diff
@@ -5,6 +5,17 @@
  */
 
 import Foundation
+import os.activity
+
+// Bridging Obj-C variabled defined as c-macroses. See `activity.h` header.
```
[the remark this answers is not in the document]
buranmert: this comment is copy&pasted from a gist, it's not clear what it says. we should improve it

Sources/Datadog/Tracing/Utils/ActiveSpanUtils.swift
```diff
@@ -0,0 +1,42 @@
...
+    static let RTLD_DEFAULT = UnsafeMutableRawPointer(bitPattern: -2)
+    static let sym = dlsym(RTLD_DEFAULT, "_os_activity_current")
+    static let OS_ACTIVITY_CURRENT = unsafeBitCast(sym, to: os_activity_t.self)
```
buranmert, line 14: actually we can do these private API stuff in Obj-C: you can check `Sources/_Datadog_Private` folder

ncreated:
> About the backward compatibility, yes, it should probably be evaluated and check throroughly before adding to master, probably `1.4.0` is a good timeframe. The change to the spans will be related to the span parents, it could probably change results but probably would mean that will improve the information for free (if used the traces properly and finished as they should)

That's true. It will only impact the code that doesn't reference parent span explicitly 👍.

Sources/Datadog/Tracing/Utils/ActiveSpanUtils.swift
nachoBonafonte: Great idea, much better than it is now, implementing now

Tests/DatadogTests/Datadog/Tracing/Utils/ActiveSpanUtilsTest.swift
nachoBonafonte: Yes, changing

Sources/Datadog/Tracing/Utils/ActiveSpanUtils.swift
nachoBonafonte: Yes, its true, probably using directly Obj-c here and using our own wrapper  is much more clear than all this hacky code. I will work on it

Sources/Datadog/Tracer.swift
nachoBonafonte: I have added it already, not sure if this is the proper way, though

Tests/DatadogTests/Datadog/Tracing/Utils/ActiveSpanUtilsTest.swift
nachoBonafonte: Added

Sources/Datadog/Tracer.swift
nachoBonafonte: Ok, I implemented as a real class and assigning this object to the tracer

Sources/Datadog/Tracing/Utils/ActiveSpanUtils.swift
nachoBonafonte: I moved this code to _Datadog_Provate modle

Sources/_Datadog_Private/ObjcOSActivityUtils.m
```diff
@@ -0,0 +1,18 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+#import <Foundation/Foundation.h>
+#import "ObjcOSActivityUtils.h"
+
+@implementation ObjcOSActivityUtils
+
+static os_activity_t _currentActivity = OS_ACTIVITY_CURRENT;
```
buranmert, line 12: if the host app starts another activity (eg: `HostAppCustomButtonActivity`), does it affect us?
(i don't think this is gonna happen, i'm just curious)

Sources/Datadog/Tracing/DDSpanContext.swift
```diff
@@ -16,6 +18,21 @@ internal struct DDSpanContext: OTSpanContext {
     /// The baggage items of this span.
     let baggageItems: BaggageItems
 
+    let activityId: os_activity_id_t
+    var activityState = os_activity_scope_state_s()
```
buranmert, line 22: are we going to use them in the future? or can we make them private?

Sources/_Datadog_Private/ObjcOSActivityUtils.m
nachoBonafonte: Yes, its true can be returned directly, i just followed the class property standard implementation, but better returning directly.

If the app starts another activity, depending how it is created we could fail to find current Span while in the activity. We can locate a current span in the current activity or in its parent activity, but we cannot iterate more

Sources/Datadog/Tracing/DDSpanContext.swift
nachoBonafonte: I have refactored the code and moved this logic to DDSpan, and have made it private there. Also all the logic interacting with ActiveSpansPool is now done in DDSPan, so the variables can be kept private

nachoBonafonte:
@test-all-platforms

nachoBonafonte:
@test-all-platforms

ncreated:
@test-all-platforms

ncreated:
@test-all-platforms nachoBonafonte/RUMM-612-Add-active-span-concept-to-Tracer

Sources/Datadog/Tracing/DDSpanContext.swift
```diff
@@ -16,6 +16,13 @@ internal struct DDSpanContext: OTSpanContext {
     /// The baggage items of this span.
     let baggageItems: BaggageItems
 
+    init(traceID: TracingUUID, spanID: TracingUUID, parentSpanID: TracingUUID?, baggageItems: BaggageItems) {
+        self.traceID = traceID
+        self.spanID = spanID
+        self.parentSpanID = parentSpanID
+        self.baggageItems = baggageItems
+    }
```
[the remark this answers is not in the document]
nachoBonafonte: Yes true, after moving the activity logic aout this initializer is not needed anymore

Sources/Datadog/Tracing/Utils/ActiveSpansPool.swift
```diff
@@ -0,0 +1,41 @@
...
+import Foundation
+import os.activity
+import _Datadog_Private
+
+internal let OS_ACTIVITY_CURRENT = unsafeBitCast(dlsym(UnsafeMutableRawPointer(bitPattern: -2), "_os_activity_current"), to: os_activity_t.self)
+
+/// Helper class to get the current Span
+internal class ActiveSpansPool {
+    var contextMap = [os_activity_id_t: DDSpan]()
+    let rlock = NSRecursiveLock()
```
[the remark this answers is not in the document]
nachoBonafonte: Sure

Sources/Datadog/Tracing/Utils/ActiveSpansPool.swift
```diff
@@ -0,0 +1,41 @@
...
+        return returnSpan
+    }
+
+    func addSpan(span: DDSpan, activityId: os_activity_id_t) {
+        rlock.lock()
+        contextMap[activityId] = span
+        rlock.unlock()
+    }
+
+    func removeSpan(span: DDSpan, activityId: os_activity_id_t) {
```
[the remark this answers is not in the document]
nachoBonafonte: Yes, you are true

nachoBonafonte:
@test-all-platforms nachoBonafonte/RUMM-612-Add-active-span-concept-to-Tracer

buranmert:
we may try to fix that thread sanitizer issue when we have time one day 🎗️

ncreated:
We also need to fix one more thing in `SendTracesFixtureViewController` (`Example` app target). Before, it was sending 3 separate traces for `URLSession` requests:

Screenshot 2020-08-05 at 10 41 14

Now, they get linked as a single trace:

Screenshot 2020-08-05 at 10 40 42

This is because the fixture starts next requests without making sure the previous one did complete. Let's keep them separate (do one by one), so we keep the clean preview of different span configuration in APM Traces Explorer.

buranmert:
@ncreated `SendTracesFixtureViewController` makes the requests sequentially but we call completion first and then `span.finish()` in auto-instrumented requests. with `activeSpan` concept it makes more sense to `span.finish()` first and then call completion 👍 

the required change is in `URLSessionSwizzler.swift` line 73 and 117
can we make the change in this PR @nachoBonafonte or shall i raise another PR for that?

nachoBonafonte:
Yes, I can make the change as part of the PR

nachoBonafonte:
Added latest changes

> Please also squash the commits 
You mean that I should squash and merge? or any other thing?

Sources/Datadog/OpenTracing/OTTracer.swift
```diff
@@ -54,14 +54,17 @@ public protocol OTTracer {
     ///        injected into the carrier (presumably in a remote process)
     /// 
     func extract(reader: OTFormatReader) -> OTSpanContext?
+
+    /// Returns the active span in the current execution context, this span will be automatically
```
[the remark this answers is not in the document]
nachoBonafonte: added missing info

Sources/Datadog/Tracing/Utils/ActiveSpansPool.swift
```diff
@@ -0,0 +1,41 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+import os.activity
+
+//This symbol is only exported like this for objective-C since uses a macro to create it, we must reproduce it
```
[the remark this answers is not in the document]
nachoBonafonte: Reworded a bit to make it clearer

Sources/Datadog/Tracing/Utils/ActiveSpansPool.swift
```diff
@@ -0,0 +1,42 @@
...
+internal let OS_ACTIVITY_CURRENT = unsafeBitCast(dlsym(UnsafeMutableRawPointer(bitPattern: -2), "_os_activity_current"), to: os_activity_t.self)
+
+/// Helper class to get the current Span
+internal class ActiveSpansPool {
+    private var contextMap = [os_activity_id_t: DDSpan]()
+    private let rlock = NSRecursiveLock()
+
+    /// Returns the Span from the current context
+    func getActiveSpan() -> DDSpan? {
+        // We should try to traverse all hierarchy to locate the Span, but I could not find a way, just direct parent
```
markst, line 21: Guess this is related to https://github.com/DataDog/dd-sdk-ios/issues/2399 ?

PR #188, opened 2020-07-17
RUMM-518 Add user Session monitoring

ncreated:
### What and why?

📦 This PR adds RUM session monitoring.

### How?

The RUM Session monitoring has 3 requirements:
* the initial Session is started with the first View,
* if Session duration exceeds `4h`, a new one is started,
* if no interaction is observed for more than `15min`, a new Session is started.

This logic is implemented in two scopes:
* `RUMSessionScope` tracks the activity time and returns `true` from `process(command:)` on expiration or timeout.
* `RUMApplicationScope` manages creation and re-start of the `RUMSessionScope`.

When the RUM Session is started for the first time, the `application_start` RUM Action is sent. Right now, this is implemented in `RUMSessionScope`, but will be moved to `RUMViewScope` in `RUMM-519`. The test is already added and won't change: `RUMMonitorTests.testWhenFirstViewIsStarted_itSendsApplicationStartAction()`

I also added RUM integration tests fixture. When the Example app is launched and the "Send RUM events for UI Tests" button is tapped, the presented view controller starts a new RUM View. This sends the `application_start ` RUM Action, which is asserted in `RUMIntegrationTests`:

Screenshot 2020-07-17 at 10 34 39

I also added UI to debug RUM Action events:

Screenshot 2020-07-17 at 10 41 30

### Review checklist

### DIFF
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
@@ -15,16 +15,15 @@ internal struct RUMScopeDependencies {
 
 internal class RUMApplicationScope: RUMScope {
     struct Constants {
-        /// No-op session ID used shortly before the real session is initialized.
+        /// No-op session ID used before the real session is started.
         static let nullUUID = UUID(uuidString: "00000000-0000-0000-0000-000000000000") ?? UUID()
     }
 
     // MARK: - Child Scopes
 
-    private lazy var sessionScope: RUMScope = RUMSessionScope(
-        parent: self,
-        dependencies: self.dependencies
-    )
+    /// Session scope. It gets created with the first `.startView` event.
+    /// Might be re-created later according to session duration constraints.
+    private(set) var sessionScope: RUMScope?
 
     // MARK: - Initialization
 
@@ -49,7 +48,24 @@ internal class RUMApplicationScope: RUMScope {
     let context: RUMContext
 
     func process(command: RUMCommand) -> Bool {
-        _ = sessionScope.process(command: command)
+        if let currentSession = sessionScope {
+            let shouldRefreshSession = currentSession.process(command: command)
+            if shouldRefreshSession {
+                let refreshedSession = RUMSessionScope(parent: self, dependencies: dependencies)
+                sessionScope = refreshedSession
+                _ = refreshedSession.process(command: command)
+            }
+        } else {
+            switch command {
+            case .startView:
+                let newSession = RUMSessionScope(parent: self, dependencies: dependencies)
+                sessionScope = newSession
+                _ = newSession.process(command: command)
+            default:
+                break
+            }
+        }
+
         return false
     }
 }
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
@@ -6,23 +6,38 @@
 
 import Foundation
 
-internal class RUMApplicationScope: RUMScope {
-    /// No-op session ID used shortly before the real session is initialized.
-    static let nullSessionID = UUID(uuidString: "00000000-0000-0000-0000-000000000000") ?? UUID()
+/// Injection container for common dependencies used by all `RUMScopes`.
+internal struct RUMScopeDependencies {
+    let dateProvider: DateProvider
+    let eventBuilder: RUMEventBuilder
+    let eventOutput: RUMEventOutput
+}
 
-    let eventBuilder: RUMEventBuilder // TODO: RUMM-518 move to `RUMMSessionScope`
-    let eventOutput: RUMEventOutput // TODO: RUMM-518 move to `RUMMSessionScope`
+internal class RUMApplicationScope: RUMScope {
+    struct Constants {
+        /// No-op session ID used shortly before the real session is initialized.
+        static let nullUUID = UUID(uuidString: "00000000-0000-0000-0000-000000000000") ?? UUID()
+    }
+
+    // MARK: - Child Scopes
+
+    private lazy var sessionScope: RUMScope = RUMSessionScope(
+        parent: self,
+        dependencies: self.dependencies
+    )
+
+    // MARK: - Initialization
+
+    let dependencies: RUMScopeDependencies
 
     init(
         rumApplicationID: String,
-        eventBuilder: RUMEventBuilder,
-        eventOutput: RUMEventOutput
+        dependencies: RUMScopeDependencies
     ) {
-        self.eventBuilder = eventBuilder
-        self.eventOutput = eventOutput
+        self.dependencies = dependencies
         self.context = RUMContext(
             rumApplicationID: rumApplicationID,
-            sessionID: RUMApplicationScope.nullSessionID,
+            sessionID: Constants.nullUUID,
             activeViewID: nil,
             activeViewURI: nil,
             activeUserActionID: nil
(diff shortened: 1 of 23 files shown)


Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
```diff
@@ -34,6 +48,24 @@ internal class RUMApplicationScope: RUMScope {
...
+            }
+        } else {
+            switch command {
+            case .startView:
+                let newSession = RUMSessionScope(parent: self, dependencies: dependencies)
+                sessionScope = newSession
+                _ = newSession.process(command: command)
+            default:
+                break
+            }
```
buranmert, line 66: i didn't understand the logic here: if a session is expired, new one can start with any command; however if there is no expired session, new session can only start with `startView`. 
why are those 2 cases different?

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
```diff
@@ -34,6 +48,24 @@ internal class RUMApplicationScope: RUMScope {
     let context: RUMContext
 
     func process(command: RUMCommand) -> Bool {
+        if let currentSession = sessionScope {
+            let shouldRefreshSession = currentSession.process(command: command)
```
buranmert, line 52: shouldn't this be `let should... = !currentSession.process(...)` as `process` returns `false` when `session` needs to be deallocated?

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
```diff
@@ -0,0 +1,111 @@
...
+
+    var context: RUMContext {
+        var context = parent.context
+        context.sessionID = sessionUUID
+        return context
+    }
+
+    func process(command: RUMCommand) -> Bool {
+        if timedOutOrExpired() {
+            return true // end session
```
buranmert, line 50: this line should return `false` according to RFC

Tests/DatadogTests/Datadog/Mocks/ServerMock.swift
```diff
@@ -226,3 +226,20 @@ extension ServerMock {
         }
     }
 }
+
+// MARK: - RUM feature helpers
+
+extension ServerMock {
+    func waitAndReturnRUMEventMatchers(count: UInt, file: StaticString = #file, line: UInt = #line) throws -> [RUMEventMatcher] {
```
buranmert, line 233: if we had a similar method for other mathcers, is it possible to make this generic function?

Tests/DatadogTests/Datadog/RUMMonitorTests.swift
```diff
@@ -76,4 +76,45 @@ class RUMMonitorTests: XCTestCase {
 
         server.waitAndAssertNoRequestsSent()
     }
+
+    // MARK: - Usage
+
+    func testWhenCallingPublicAPI_itProcessessExpectedCommandsThrougScopes() {
```
buranmert, line 82: typo: `itProcessesExpectedCommandsThroughScopes`

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
buranmert, line 52: according to RFC, this should be `let should... = !currentSession.process(...)` as `process` returns `false` when `session` needs to be deallocated

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
```diff
@@ -34,6 +48,24 @@ internal class RUMApplicationScope: RUMScope {
...
+            if shouldRefreshSession {
+                let refreshedSession = RUMSessionScope(parent: self, dependencies: dependencies)
+                sessionScope = refreshedSession
+                _ = refreshedSession.process(command: command)
+            }
+        } else {
+            switch command {
+            case .startView:
```
buranmert, line 60: i didn't understand the logic here: if a session is expired, new one can start with any command; however if there is no expired session, new session can only start with `startView`. 
why are those 2 cases different?

Tests/DatadogTests/Datadog/Mocks/ServerMock.swift
ncreated: IMO it will be more hassle to work with generics and no gain. Instead of this usage:
```swift
let logMatchers = serverMock.waitAndReturnLogMatchers(count: 2)
let spanMatchers = serverMock.waitAndReturnSpanMatchers(count: 2)
let rumEventMatchers = serverMock.waitAndReturnRUMEventMatchers(count: 2)
```
we'd be typing sth like this in every place:
```swift
let logMatchers = serverMock.waitAndReturnMatchers(count: 2, genericKind: LogMatcher.self)
let spanMatchers = serverMock.waitAndReturnMatchers(count: 2, genericKind: LogMatcher.self)
let rumEventMatchers = serverMock.waitAndReturnMatchers(count: 2, genericKind: LogMatcher.self)
```
the `genericKind` would have to let the `ServerMock` map the request payload `Data` to a concrete matcher. And this mapping is different for Logs and Spans + RUM Events, as they use different payload format.

There's almost no code duplication between 3 existing methods.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
ncreated: It's coming from the first product requirement:

> The RUM Session monitoring has 3 requirements:
> * the initial Session is started with the first View,
> * ...
> * ...

So the session starts as soon as some view is displayed in the app. Once the session is started, RUM always has a valid session. Although, it doesn't need to be the same session, because we implement session timeout and expiration logic.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
ncreated: 👍 good catch, done ✅. The RFC says `false` in one place, but then keeps saying `true` in code commends. I will update it as well.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
ncreated: Done 👍

Tests/DatadogTests/Datadog/Mocks/ServerMock.swift
buranmert: you don't need `genericKind` parameter, you can do `let logMatchers: [LogMatcher] = serverMock.waitAndReturn...`
if it doesn't make life easier, nevermind then 👌

ncreated: The thing is that all 3 matchers offer also other factory methods:
```swift
LogMatcher.fromArrayOfJSONObjectsData(Data) -> [LogMatcher]
LogMatcher.fromJSONObjectData(Data) -> LogMatcher

SpanMatcher.fromNewlineSeparatedJSONObjectsData(Data) -> [SpanMatcher]
SpanMatcher.fromJSONObjectData(Data) -> SpanMatcher

RUMEventMatcher.fromNewlineSeparatedJSONObjectsData(Data) -> [RUMEventMatcher]
RUMEventMatcher.fromJSONObjectData(Data) -> RUMEventMatcher // not yet existing, but may come later
```

Finding a common name for all three will IMO immerse the readability:
```swift
Matcher.fromMultiObjectPayload(Data) -> [Matcher]
Matcher.fromSingleiObjectPayload(Data) -> Matcher
```
plus, it will require quite a refactoring effort with no much gain.

PR #190, opened 2020-07-22
RUMM-519 Add Views monitoring

ncreated:
### What and why?

📦 This PR adds RUM Views monitoring.

Screenshot 2020-07-22 at 08 49 14

### How?

Two public APIs are exposed on `RUMMonitor`:
```swift
/// Notifies that the View starts being presented to the user.
/// - Parameters:
///   - viewController: the instance of `UIViewController` representing this View.
///   - attributes: custom attributes to attach to the View.
public func start(viewController: UIViewController, attributes: [AttributeKey: AttributeValue]? = nil) {
    start(view: viewController, attributes: attributes)
}

/// Notifies that the View stops being presented to the user.
/// - Parameters:
///   - viewController: the instance of `UIViewController` representing this View.
///   - attributes: custom attributes to attach to the View.
public func stop(viewController: UIViewController, attributes: [AttributeKey: AttributeValue]? = nil) {
    stop(view: viewController, attributes: attributes)
}
```

New behaviours are introduced with adding `RUMViewScope`:
* when initial `View` is started, it sends the `application_start` RUM Action event,
* when View is started, it sends the RUM View event updating the View state in RUM Explorer,
* when View is stopped, it sends the RUM View event updating the View state in RUM Explorer.

The `RUMSessionScope` has one new behaviour:
* it manages `RUMViewScope` lifecycle (starting / stopping `RUMViewScopes`).

The session expiration behaviour in `RUMApplicationScope` was updated accordingly:
* when session expires, it starts the new one **and transfers active View scopes to the new session**.

The RUM Integration test fixture now navigates through 3 view controllers, issuing corresponding start/stop events:

Screenshot 2020-07-22 at 08 42 17

### Review checklist

### DIFF
--- a/Sources/Datadog/RUM/RUMMonitor/RUMScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/RUMScope.swift
@@ -20,17 +20,17 @@ extension RUMScope {
     /// Propagates given `command` to the child scope and manages its lifecycle by
     /// removing it if it gets closed.
     func propagate(command: RUMCommand, to chilScope: inout RUMScope?) {
-        if chilScope?.process(command: command) == true {
+        if chilScope?.process(command: command) == false {
             chilScope = nil
         }
     }
 
     /// Propagates given `command` through array of child scopes and manages their lifecycle by
     /// removing child scopes that get closed.
-    func propagate(command: RUMCommand, to chilScopes: inout [RUMScope]) {
-        chilScopes = chilScopes.filter { chilScope in
-            let shouldBeClosed = chilScope.process(command: command)
-            return shouldBeClosed == false
+    func propagate(command: RUMCommand, to childScopes: inout [RUMScope]) {
+        childScopes = childScopes.compactMap { childScope in
+            let shouldBeKept = childScope.process(command: command)
+            return shouldBeKept ? childScope : nil
         }
     }
 }
--- a/Sources/Datadog/RUM/RUMMonitor/RUMScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/RUMScope.swift
@@ -15,3 +15,32 @@ internal protocol RUMScope: class {
     /// * `false` if the scope should be closed.
     func process(command: RUMCommand) -> Bool
 }
+
+extension RUMScope {
+    /// Propagates given `command` to the child scope and manages its lifecycle by
+    /// removing it if it gets closed.
+    func propagate(command: RUMCommand, to chilScope: inout RUMScope?) {
+        if chilScope?.process(command: command) == true {
+            chilScope = nil
+        }
+    }
+
+    /// Propagates given `command` through array of child scopes and manages their lifecycle by
+    /// removing child scopes that get closed.
+    func propagate(command: RUMCommand, to chilScopes: inout [RUMScope]) {
+        chilScopes = chilScopes.filter { chilScope in
+            let shouldBeClosed = chilScope.process(command: command)
+            return shouldBeClosed == false
+        }
+    }
+}
+
+extension Dictionary where Key == AttributeKey, Value == AttributeValue {
+    /// Merges given `rumCommandAttributes` to current dictionary, by overwriting values.
+    mutating func merge(rumCommandAttributes: [AttributeKey: AttributeValue]?) {
+        guard let additionalAttributes = rumCommandAttributes else {
+            return
+        }
+        merge(additionalAttributes) { _, new in new }
+    }
+}
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
@@ -43,6 +43,32 @@ internal class RUMSessionScope: RUMScope {
         self.lastInteractionTime = startTime
     }
 
+    /// Creates a new Session upon expiration of the previous one.
+    convenience init(
+        from expiredSession: RUMSessionScope,
+        startTime: Date
+    ) {
+        self.init(
+            parent: expiredSession.parent,
+            dependencies: expiredSession.dependencies,
+            startTime: startTime
+        )
+
+        // Transfer active Views by creating new `RUMViewScopes` for their identity objects:
+        self.viewScopes = expiredSession.viewScopes.compactMap { expiredView in
+            guard let expiredView = expiredView as? RUMViewScope, let expiredViewIdentity = expiredView.identity else {
+                return nil // if the underlying `UIVIewController` no longer exists, skip transferring this scope
+            }
+            return RUMViewScope(
+                parent: self,
+                dependencies: dependencies,
+                identity: expiredViewIdentity,
+                attributes: expiredView.attributes,
+                startTime: startTime
+            )
+        }
+    }
+
     // MARK: - RUMScope
 
     var context: RUMContext {
@@ -68,7 +94,7 @@ internal class RUMSessionScope: RUMScope {
             break
         }
 
-        // Propagation
+        // Propagate command
         propagate(command: command, to: &viewScopes)
 
         return true
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
@@ -14,27 +14,33 @@ internal class RUMSessionScope: RUMScope {
         static let sessionMaxDuration: TimeInterval = 4 * 60 * 60 // 4 hours
     }
 
+    // MARK: - Child Scopes
+
+    /// Active View scopes. Scopes are added / removed when the View starts / stops displaying.
+    private(set) var viewScopes: [RUMScope] = []
+
     // MARK: - Initialization
 
     unowned let parent: RUMScope
     private let dependencies: RUMScopeDependencies
 
-    /// This session UUID.
+    /// This Session UUID.
     private var sessionUUID: UUID
-    /// The start time of this session.
+    /// The start time of this Session.
     private var sessionStartTime: Date
-    /// Time of the last RUM interaction noticed by this session.
+    /// Time of the last RUM interaction noticed by this Session.
     private var lastInteractionTime: Date
 
     init(
         parent: RUMScope,
-        dependencies: RUMScopeDependencies
+        dependencies: RUMScopeDependencies,
+        startTime: Date
     ) {
         self.parent = parent
         self.dependencies = dependencies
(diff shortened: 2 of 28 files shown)


Sources/Datadog/RUM/RUMMonitor/RUMScope.swift
```diff
@@ -15,3 +15,32 @@ internal protocol RUMScope: class {
...
+    }
+
+    /// Propagates given `command` through array of child scopes and manages their lifecycle by
+    /// removing child scopes that get closed.
+    func propagate(command: RUMCommand, to childScopes: inout [RUMScope]) {
+        childScopes = childScopes.compactMap { childScope in
+            let shouldBeKept = childScope.process(command: command)
+            return shouldBeKept ? childScope : nil
+        }
+    }
```
buranmert, line 35: name doesn't imply nullifying behavior, can't we achieve this explicitly instead?
```swift
// SessionScope
process(cmd) {
  self.viewScopes = propagate(cmd, viewScopes)
}
// RUMScope
func propagate(cmd, scopes) -> Scopes {
  return scopes.filter { $0.process(cmd) }
}
```

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
```diff
@@ -14,27 +14,59 @@ internal class RUMSessionScope: RUMScope {
         static let sessionMaxDuration: TimeInterval = 4 * 60 * 60 // 4 hours
     }
 
+    // MARK: - Child Scopes
+
+    /// Active View scopes. Scopes are added / removed when the View starts / stops displaying.
+    private(set) var viewScopes: [RUMScope] = []
```
buranmert, line 20: why do we need `[RUMScope]`? i don't see why not `[RUMViewScope]`.?
that would lead to simpler `process(command: OnlyCommandTypeThatConcernsMe)` methods
we won't need to test, for example, `RUMViewScope` with every `RUMCommand` type

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
```diff
@@ -14,27 +14,59 @@ internal class RUMSessionScope: RUMScope {
...
+    ) {
+        self.init(
+            parent: expiredSession.parent,
+            dependencies: expiredSession.dependencies,
+            startTime: startTime
+        )
+
+        // Transfer active Views by creating new `RUMViewScopes` for their identity objects:
+        self.viewScopes = expiredSession.viewScopes.compactMap { expiredView in
+            guard let expiredView = expiredView as? RUMViewScope, let expiredViewIdentity = expiredView.identity else {
```
buranmert, line 59: maybe we should check `(identity as UIViewController).window == nil` here? 
sometimes apps keep VCs alive although they are not part of view hierarchy, do we consider this case as "view is alive" or not?

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
ncreated: That's true, we can change it to concrete scope 👌 and this will make few things easier now and later. The idea o `RUMScope` comes from the RFC, but now I don't see any negative impact on testing if we make it a concrete scope (the child scopes creation happens internally, so we don't need / cannot mock the child scopes ✨):

Screenshot 2020-07-24 at 09 36 58

The unidirectional commands flow can't be enforced more by the architecture, but we keep a good track of it in the implementation and CR 👌.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
```diff
@@ -14,27 +14,59 @@ internal class RUMSessionScope: RUMScope {
...
     }
 
+    // MARK: - Child Scopes
+
+    /// Active View scopes. Scopes are added / removed when the View starts / stops displaying.
+    private(set) var viewScopes: [RUMScope] = []
+
     // MARK: - Initialization
 
     unowned let parent: RUMScope
```
buranmert, line 24: does `parent` need to be a `RUMScope`? if we split `RUMScope` protocol into `ContextProvider` and `CommandHandler` (i made up those names) then we can do `parent: ContextProvider`; that way we ensure commands are propagated top to down only.
does that make sense?

Sources/Datadog/RUM/RUMMonitor/RUMScope.swift
ncreated: This will lead to more code and could be more error-prone if we consider all required ways of owning the scope.

Scopes will be owned in 3 ways:
```swift
var viewScopes: [RUMScope]                 // array
var resourceScopes: [String: RUMScope]     // dictionary keyed by resource name
var actionScope: RUMScope?                 // direct reference
```

With the two `propagate(command:to:)` methods, this is simply:
```swift
propagate(command: command, to: &viewScopes)
propagate(command: command, to: &resourceScopes[resourceName])
propagate(command: command, to: &actionScope)
```

Going "explicit" adds more code code on the caller side:
```swift
viewScopes = propagate(command, to: viewScopes)

if self.resourceScopes[resourceName] != nil {
   resourceScopes = propagate(command, to: resourceScope)
}

self.actionScope = propagate(command, actionScope)
```
and requires 3 different implementations of `propagate(...)` method. Also, when we change to `[RUMViewScope]` https://github.com/DataDog/dd-sdk-ios/pull/190#discussion_r459485890, the `propagate()` method would have to become generic... so it's going even more complex for no reason.

I don't understand what is the problem with current `inout` implementation 🤔. I don't see neither how proposed `propagate(cmd, scopes) -> Scopes` implies the nullifying behaviour vs current one doesn't. 

Would renaming it to `manage(childScopes: inout [RUMScope], byPropagatingCommand command: RUMCommand)` make it better? Please keep in mind - it's just implementation detail of the `RUMScope` that child scopes are closed on propagation. I don't think we need to express it more. LTM WDYT.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
ncreated: Good idea 👍. I think it's best to have both: check if the object exists + check if it's removed from the window. This way we could be even sending "reference leak" RUM Erorrs in such case 🙌.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
ncreated: Yes, it makes sense 👌. Indeed, the `parent` reference is to only get the `parent.context`. The `RUMContextProvider` interface is planned for `RUMM-597: Current `RUMContext` is available globally`. Let's try this refactoring then - we will have the knowledge on all scopes (right now few are still missing). WDYT?

Sources/Datadog/RUM/RUMMonitor/RUMScope.swift
buranmert: yes, renaming it to that would be 👍 ✅ 

***

`Dict` case would be `dict[name] = propagate(cmd, dict[name]) // second param is (Scope?)`.
`propagate() -> Scopes` doesn't imply it neither but it doesn't nullify anything, it's the caller who sets things to nil practically and this is the difference.

it can be nice to express when we can even on `private` things as our users may need to read/understand them one day

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
buranmert: a `TODO` here would be 👌 then

Sources/Datadog/RUM/RUMMonitor/RUMScope.swift
ncreated: Renamed 👌 here: https://github.com/DataDog/dd-sdk-ios/pull/193/commits/594b5e57c2166e1d676706d028504f962a27604a (to avoid 2PRs rebasing).

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
ncreated: Added 👌 https://github.com/DataDog/dd-sdk-ios/pull/193/commits/fed1d347a78fb02e52a6af6419558371e405111c I also updated `RUMM-597` with this notice.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
ncreated: Done in https://github.com/DataDog/dd-sdk-ios/pull/193/commits/42ec45e06c30c2907746ec5873efd65a7ddaec66 for all scopes (Application, Session, View and Resource). This enforces generic `manage(childScopes:...)` anyway, but it reduces the complexity of `as?` casting in some places 👌.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
ncreated: Done in https://github.com/DataDog/dd-sdk-ios/pull/193/commits/50693ac26132de67b675929125259bafff461b23. I've also created `RUMM-634 Produce a RUM error when the UIViewController is leaked`.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -0,0 +1,145 @@
...
+    let viewUUID: RUMUUID
+    /// The URI of this View, used as the `view.url` in RUM Explorer.
+    let viewURI: String
+    /// The start time of this View.
+    private var viewStartTime: Date
+
+    /// Number of actions happened on this View.
+    private var actionsCount: UInt = 0
+    /// Current version of this View to use for RUM `documentVersion`.
+    private var version: UInt = 0
```
buranmert, line 31: this is not clear to me 🤔 what does version of view mean?

ncreated: Whenever something happens on the View (i.e. Resource finishes loading or User Action is observed), a new `RUMViewEvent` is send to the backend. The `event.dd.documentVersion` is incremented for every new event of the View. This way, backend can reduce received events (View updates) into single RUM View displayed in RUM Explorer.

This `version` property tracks the `documentVersion` of this View.

PR #191, opened 2020-07-22
RUMM-517 Model generator tool added

buranmert:
### What and why?

We need to generate Swift data models from json schemas in [`rum-events-format` repo](https://github.com/DataDog/rum-events-format)

### How?

Instead of writing our own _json-schema-to-Swift_ generator, we forked [an open-source tool called `quicktype`](https://github.com/quicktype/quicktype)

We also wrote a script to automate the process
The command below generates `RUMModels.swift` from the tip of `master` in `rum-events-format`
```zsh
./tools/generate-models/run.sh generate RUMModels.swift
```

And this command verifies if `RUMModels.swift` is original or modified
```zsh
./tools/generate-models/run.sh compare RUMModels.swift
```

It appends the SHA of its origin commit to the generated source file.
It uses this SHA to verify existing models.
It generates a new file from this SHA and then diffs with the given file.

### Review checklist

### Question?

Where should we keep generated `RUMModels.swift` file?
1. In `Sources/Datadog`?
2. In `tools/generated-models`?
3. In `Datadog/TargetSupport`?

### TODO
- [x] Merge https://github.com/DataDog/rum-events-format/pull/11 first as it causes a minor-ish problem with `quicktype` tool
- [x] Add an entry to `Makefile` to use this tool with the correct file location


tools/generate-models/run.sh
```diff
@@ -0,0 +1,64 @@
...
+if [[ $MODE == $MODE_VERIFY ]]; then
+    DIFF=$(diff RUMModels.swift.bak $TARGET_FILE)
+    if [[ ! -z $DIFF ]]; then 
+        echo "Diff detected!\n$DIFF"
+        return 1;
+    fi
+    echo "Verified!"
+    return 0;
+elif [[ $MODE == $MODE_GENERATE ]]; then
+    cp RUMModels.swift.bak $TARGET_FILE
```
ncreated, line 60: As we expose `protocol RUMDataModel`, let's name the target file `RUMDataModels.swift`, WDYT?

tools/generate-models/run.sh
```diff
@@ -0,0 +1,64 @@
...
+    else
+        echo "Target SHA could not be read in $TARGET_FILE !"
+        return 1;
+    fi
+elif [[ $MODE == $MODE_GENERATE ]]; then
+    GIT_REF="master"
+else
+    echo "Invalid command!"
+    echo "Model generation: run.sh generate ${TARGET_LOCATION_FOR_NEW_FILE}"
+    echo "Model verification: run.sh verify ${FILE_TO_BE_VERIFIED}"
```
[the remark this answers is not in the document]
buranmert: it was meant for "input your target location here" but now we hardcode these params and i remove them from echos

tools/generate-models/run.sh
```diff
@@ -0,0 +1,64 @@
...
+fi
+cd "rum-events-format"
+git fetch origin $GIT_REF
+git checkout FETCH_HEAD
+SHA=$(git rev-parse HEAD)
+cd ..
+
+# install quicktype and generate model
+if [[ ! -a $(npm bin)/quicktype ]]; then
+    npm install buranmert/quicktype
```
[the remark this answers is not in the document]
buranmert: yes but first i need to talk to opensource people about this because that fork is not for public use
i think `npm` has private packages, maybe i can publish this as a private `npm` package

tools/generate-models/run.sh
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env zsh
```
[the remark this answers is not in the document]
buranmert: `zsh` is the default shell, if we have `bin/bash` in other places we should replace them too

tools/generate-models/run.sh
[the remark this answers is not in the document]
buranmert: EDIT: private packages don't seem to be what we need. i will move my fork to DD org once opensource people approves ⏳

tools/generate-models/run.sh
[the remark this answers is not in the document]
ncreated: Oh, indeed it's a default one since Catalina. Let's keep it 👍. Other tools work fine with `bin/bash` so there's no need to update them now. Our Bitrise stacks are Catalina too.

tools/generate-models/run.sh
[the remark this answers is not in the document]
buranmert: done ✅ 
https://github.com/DataDog/quicktype

tools/generate-models/run.sh
```diff
@@ -0,0 +1,77 @@
...
+fi
+cd "rum-events-format"
+git fetch origin $GIT_REF
+git checkout FETCH_HEAD
+SHA=$(git rev-parse HEAD)
+cd ..
+
+# install quicktype and generate model
+if [[ ! -a $(npm bin)/quicktype ]]; then
+    npm install DataDog/quicktype
```
ncreated, line 49: How do we specify the version of `DataDog/quicktype`? If it always uses the latest, this won't be possible to leverage caching it between CI builds (upgrading won't be possible), no?

Sources/Datadog/RUM/DataModels/RUMDataModels.swift
```diff
@@ -0,0 +1,809 @@
...
+
+import Foundation
+
+internal protocol RUMDataModel: Codable { }
+
+/// Schema of all properties of an Action event
+///
+/// Schema of common properties of RUM events
+// MARK: - RUMAction
+internal struct RUMAction: RUMDataModel {
```
ncreated, line 17: `struct` comment must go after section `// MARK:`, otherwise the documentation is not rendered:

Screenshot 2020-07-23 at 09 58 41

Sources/Datadog/RUM/DataModels/RUMDataModels.swift
```diff
@@ -0,0 +1,809 @@
...
+    /// Size in octet of the resource response body
+    let size: Int64?
+    /// SSL phase properties
+    let ssl: RUMSSL?
+    /// HTTP status code of the resource
+    let statusCode: Int64?
+    /// Resource type
+    let type: RUMResourceType
+    /// URL of the resource
+    let url: String
```
ncreated, line 524: Is it easy doable to sort these fields, i.e. non-optional first or just follow the order they are defined in `rum-events-format`? It's a bit weird that the most important ones (`type` and `url`) are defined at the end, while less important / optional (`dns`, `firstByte`) are at the top of initializer. The `rum-events-format` lists important things first.

tools/generate-models/run.sh
buranmert: this line doesn't check version and fetches the latest commit from this repo 
this is hopefully a temporary workaround because we didn't publish this package yet
as i said in slack, if we publish our fork and cache global `npm` modules in CI then we can improve CI time

for published npm packages, the syntax is like this: `"graphql": "^0.11.7",`

ncreated: OK, I added `RUMM-631: Speed up RUM models generation on CI` to the backlog.

Sources/Datadog/RUM/DataModels/RUMDataModels.swift
buranmert: let me see if i can change this quickly

Sources/Datadog/RUM/DataModels/RUMDataModels.swift
buranmert: it follows the ordering in json-schemas... but in reversed
let me see this one too how i can reverse it

buranmert:
good news: this last build took ~16mins, this is almost our average

PR #194, opened 2020-07-27
RUMM-513 Validating JSON payload before sending to intake

buranmert:
### What and why?

We observe non-parseable logs sent to intake
In a certain org which has high volumes of logs,
%0.3-0.5 of all logs are not parsed properly in dashboard

Hypothesis #1
Data being read is malformed, because every malformed log has their malformed part at the very end
So if we can read the file later, we can have valid payload

Hypothesis #2
Write op corrupts the file and we need to watch write ops closely so that we can catch and recover errors
This requires major refactoring in File.swift and related files

### How?

This commit implements the first hypothesis by validating the payload before uploading it to intake

### How to test the fix?

```swift
let content = "some payload".data
let numberOfExtraBytes = 4
fileHandle.write(content)
// this puts 4 extra null bytes at the end of the file
fileHandle.truncate(atOffset: content.count + numberOfExtraBytes)
// now JSONSerialization throws with "Garbage at the end" error
try JSONSerialization.jsonObject(with: readData, options: [])
```

There are other cases than null bytes at the end, such as incomplete log message, this fix should also cover them

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/Persistence/FileReader.swift
+++ b/Sources/Datadog/Core/Persistence/FileReader.swift
@@ -38,11 +38,17 @@ internal final class FileReader {
         }
     }
 
+    private static let utf8NewlineData = "\n".data(using: .utf8)!.first! // swiftlint:disable:this force_unwrapping
     private func synchronizedReadNextBatch() -> Batch? {
         if let file = orchestrator.getReadableFile(excludingFilesNamed: Set(filesRead.map { $0.name })) {
             do {
                 let fileData = try file.read()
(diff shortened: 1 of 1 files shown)


buranmert:
i'd add `validator: (Data) -> Bool` property to `DataFormat` type and validate it accordingly whether it is a single json or a bunch of newline-separate-json, this is solvable ✅ 

sure it's a trade-off: validation adds 0.04 seconds more CPU time in a background thread in every 5 seconds and/but in return (_hopefully_ 🤞) %100 of the logs will be parsed successfully

i can't reproduce the issue at all: i'm closing the issue

PR #197, opened 2020-07-31
RUMM-459 Shopist master and detail pages added

buranmert:
### What and why?

We need Shopist app to generate demo data in dashboards

### How?

This PR develops the home/master page and detail page of Shopist app
The app will also have Cart and Item pages 🔜 

### Review checklist

### DIFF
--- a/Shopist/Shopist/API.swift
+++ b/Shopist/Shopist/API.swift
@@ -1,0 +1,63 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+import Alamofire
+
+struct Category: Decodable {
+    let id: String
+    let title: String
+    let cover: URL
+}
+
+struct Item: Decodable {
+    let name: String
+    let price: String
+    let cover: URL
+    let id: Int
+    let isInStock: Bool
+}
+
+final class API {
+    private static let baseURL = "https://" + (Bundle.main.object(forInfoDictionaryKey: "ShopistBaseURL") as! String)
+    typealias Completion<T: Decodable> = (Result<T, Error>) -> Void
+    private let httpClient = Alamofire.Session(configuration: .default)
+    private let jsonDecoder = JSONDecoder()
+
+    func getCategories(completion: @escaping Completion<[Category]>) {
+        get("\(Self.baseURL)/categories.json", completion: completion)
+    }
+
+    func getItems(for category: Category, completion: @escaping Completion<[Item]>) {
+        let categoryID = category.id
+        let urlString = "\(Self.baseURL)/category_\(categoryID).json"
+        get(urlString, completion: completion)
+    }
+
+    private func get<T: Decodable>(_ urlString: String, completion: @escaping Completion<T>) {
+        guard let url = URL(string: urlString) else {
+            // Datadog.log maybe?
+            fatalError("\(urlString) is not a valid URL!")
+        }
+        let request = URLRequest(url: url)
+
+        httpClient.request(request).response { response in
+            if let someError = response.error {
+                completion(.failure(someError))
+            } else if let someData = response.data {
+                do {
+                    let decoded = try self.jsonDecoder.decode(T.self, from: someData)
+                    completion(.success(decoded))
+                } catch {
+                    completion(.failure(error))
+                }
+            } else {
+                // Datadog.log maybe?
+                print("Unknown case")
+            }
+        }
+    }
+}
--- a/Shopist/Shopist/API.swift
+++ b/Shopist/Shopist/API.swift
@@ -13,7 +13,7 @@ struct Category: Decodable {
     let cover: URL
 }
 
-struct Item: Decodable {
+struct Product: Decodable {
     let name: String
     let price: String
     let cover: URL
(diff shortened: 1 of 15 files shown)


Shopist/Shopist/API.swift
```diff
@@ -0,0 +1,63 @@
...
+
+    func getItems(for category: Category, completion: @escaping Completion<[Item]>) {
+        let categoryID = category.id
+        let urlString = "\(Self.baseURL)/category_\(categoryID).json"
+        get(urlString, completion: completion)
+    }
+
+    private func get<T: Decodable>(_ urlString: String, completion: @escaping Completion<T>) {
+        guard let url = URL(string: urlString) else {
+            // Datadog.log maybe?
```
ncreated, line 42: We should be logging only data necessary for demo flows.

Shopist/Shopist/AppDelegate.swift
```diff
@@ -9,7 +9,7 @@ import Datadog
 
 fileprivate(set) var logger: Logger!
 
-let appConfig = ExampleAppConfig(serviceName: "ios-sdk-example-app")
+let appConfig = ExampleAppConfig(serviceName: "ios-sdk-shopist-app-debug")
```
ncreated, line 12: `debug` has the environment meaning. If we want it, then it should be used for: `Datadog.Configuration.builderUsing(clientToken:environment)`.

Shopist/Shopist/Scenes/ListViewController.swift
```diff
@@ -0,0 +1,82 @@
...
+        if !refreshNeeded { return }
+        fetch(with: api)
+    }
+
+    func reload() {
+        DispatchQueue.main.async {
+            self.collectionView.refreshControl?.endRefreshing()
+            self.collectionView.reloadData()
+        }
+    }
```
ncreated, line 55: Is pull to refresh feature necessary for demo flow? It brings complexity, so if it's unnecessary let's avoid adding it. I don't see it on Shopist website, but dunno if we leverage it on Android.

xcconfigs/Shopist.xcconfig
```diff
@@ -0,0 +1,4 @@
+#include "Datadog.xcconfig"
+
+SHOPIST_BASE_URL=some.url
+#include? "Shopist.local.xcconfig" // overrides SHOPIST_BASE_URL
```
ncreated, line 4: Can we 🙏 include this file in project navigator, so it's easily accessible? It will be red when not created, which is also good for understanding that it's missing. As we do in SDK project:

Screenshot 2020-08-03 at 11 31 15

Shopist/Shopist/Scenes/MasterViewController.swift
```diff
@@ -0,0 +1,39 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import UIKit
+
+final class MasterViewController: ListViewController {
```
ncreated, line 9: How about naming it `CategoriesListViewController`? "Master" doesn't describe what this screen is responsible for.

Shopist/Shopist/Scenes/DetailViewController.swift
```diff
@@ -0,0 +1,45 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import UIKit
+
+final class DetailViewController: ListViewController {
```
ncreated, line 9: How about naming it `ProductsListViewController`? "Detail" doesn't describe what this screen is made for.

Shopist/Shopist/Scenes/DetailViewController.swift
```diff
@@ -0,0 +1,45 @@
...
+            self.reload()
+        }
+    }
+
+    override func config(cell: Cell, for index: Int) {
+        let item = items[index]
+        var text = "\(item.id): \(item.name)\n\(item.price)"
+        if item.isInStock {
+            text.append(" IN STOCK!")
+        }
```
ncreated, line 37: Presenting ID to the user looks weird. Let's keep the number of info the same as Android - so only product name:

ref2

Shopist/Shopist/Scenes/MasterViewController.swift
```diff
@@ -0,0 +1,39 @@
...
+                print(error)
+            }
+            self.reload()
+        }
+    }
+
+    override func config(cell: Cell, for index: Int) {
+        let category = categories[index]
+        cell.text = "\(category.id): \(category.title)"
+        cell.setImage(url: category.cover)
```
ncreated, line 27: Presenting category ID looks weird. Let's display only the category name as on Android:

ref1

Shopist/Shopist/Views/CollectionViewCells.swift
```diff
@@ -0,0 +1,49 @@
...
+        get { imageView.image }
+        set { imageView.image = newValue }
+    }
+    func setImage(url: URL) {
+        imageView.af.setImage(withURL: url)
+    }
+
+    private let imageView: UIImageView = {
+        let imageView = UIImageView(frame: CGRect.zero)
+        imageView.contentMode = .scaleAspectFit
```
ncreated, line 25: How about using `.scaleAspectFill` and clipping it to the cell corners? Now it looks buggy:

Screenshot 2020-08-03 at 11 52 43

As we don't have design for this, we can use Shopist website and Shopist Android as the reference for UI.

buranmert: i was told nobody would see that UI, have we changed the plan? `contentMode` doesn't change generated rum events

buranmert:
screens from f619ffbf71bee2fd03f303ae66472f2c933efb6f:
categories
products

Shopist/Shopist/Views/CollectionViewCells.swift
ncreated: IMO it's fair to expect our users to launch the app and browse the instrumentation & code. And it's almost zero effort to make it display well - nothing fancy.

PR #204, opened 2020-08-07
RUMM-524 Add Resource Errors monitoring

ncreated:
### What and why?

📦 This PR adds Resource Errors monitoring.

Screenshot 2020-08-07 at 12 12 10

This aligns RUM Explorer counters (`Resource`, `User Action`, `Error`) to give valid values all the time. Here, an example of the initial `View` which downloads two `Resources` after `User Action` of "tap" on the button. The first `Resource` completed successfully, the second one failed with `Error`:

Screenshot 2020-08-07 at 12 12 27

### How?

Resources monitoring public API was changed in `RUMMonitor`. Now it exposes 4 methods:
```swift
/// Notifies that the Resource starts being loaded.
/// - Parameters:
///   - resourceName: the name representing the Resource - must be unique among all Resources being currently loaded.
///   - url: the `URL` of the Resource.
///   - httpMethod: the HTTP method used to load the Resource.
///   - attributes: custom attributes to attach to the Resource.
public func startResourceLoading(resourceName: String, url: URL, httpMethod: RUMHTTPMethod, attributes: [AttributeKey : AttributeValue]? = nil)

/// Notifies that the Resource stops being loaded succesfully.
/// - Parameters:
///   - resourceName: the name representing the Resource - must match the one used in `startResourceLoading(...)`.
///   - kind: the type of the Resource.
///   - httpStatusCode: the HTTP response status code for this Resource.
///   - size: the size of the Resource (in bytes).
///   - attributes: custom attributes to attach to the Resource.
public func stopResourceLoading(resourceName: String, kind: RUMResourceKind, httpStatusCode: Int?, size: UInt64? = nil, attributes: [AttributeKey : AttributeValue]? = nil)

/// Notifies that the Resource stops being loaded with error.
/// This should be used when `Error` object is received on Resource failure.
/// - Parameters:
///   - resourceName: the name representing the Resource - must match the one used in `startResourceLoading(...)`.
///   - error: the `Error` object received when loading the Resource.
///   - httpStatusCode: HTTP status code (optional).
///   - attributes: custom attributes to attach to the Resource.
public func stopResourceLoadingWithError(resourceName: String, error: Error, source: RUMErrorSource, httpStatusCode: Int?, attributes: [AttributeKey : AttributeValue]? = nil)

/// Notifies that the Resource stops being loaded with error.
/// If `Error` object available on Resource failure `stopResourceLoadingWithError(..., error:, ...)` should be used instead.
/// - Parameters:
///   - resourceName: the name representing the Resource - must match the one used in `startResourceLoading(...)`.
///   - errorMessage: the message explaining Resource failure.
///   - httpStatusCode: HTTP status code (optional).
///   - attributes: custom attributes to attach to the Resource.
public func stopResourceLoadingWithError(resourceName: String, errorMessage: String, source: RUMErrorSource, httpStatusCode: Int? = nil, attributes: [AttributeKey : AttributeValue]? = nil)
```

Behaviours implemented:
* when `Resource` completes successfully, it is counted on `view.resource.count` (and `action.resource.count` if there's an active UA);
* when `Resource` completes with error, it is counted on `view.error.count` (and `action.error.count` if there's an active UA).

I also replaced hand-crafted `RUMResourceEvent` with its auto generated `RUMResource` counterpart.

### Review checklist


Sources/Datadog/RUM/RUMMonitor/Scopes/RUMUserActionScope.swift
```diff
@@ -83,51 +83,29 @@ internal class RUMUserActionScope: RUMScope {
...
+                resourcesCount += 1
             } else if command is RUMStopResourceWithErrorCommand {
-                trackResourceError()
+                activeResourcesCount -= 1
+                errorsCount += 1
             }
 
         case _ as RUMAddCurrentViewErrorCommand:
-            trackViewError()
+            errorsCount += 1
```
nachoBonafonte, line 101: Are this counters thread safe?, can a user actions load some network resources that can fail asynchronously in different threads?

ncreated: Thread safety is ensured at the `RUMMonitor` level:
```swift
public class RUMMonitor {
    private let queue = DispatchQueue

    public methodExposedToUser(...) {
        process(...)
    }

    private func process(command: RUMCommand) {
        queue.async {
            _ = self.applicationScope.process(command: command)
        }
    }
}
```

Every processing after `RUMMonitor` (so in all `RUMScopes`) is single threaded.

Sources/Datadog/RUM/DataModels/RUMDataModelsMapping.swift
```diff
@@ -6,14 +6,59 @@
...
+        return Int64(exactly: self) ?? .max
+    }
+}
+
+extension UInt64 {
+    var toInt64: Int64 {
+        if self > Int64.max {
+            return .max
+        }
+        return Int64(exactly: self) ?? .max
```
buranmert, line 22: `return` line seems to cover all the cases, we can remove if clause above

Sources/Datadog/RUMMonitor.swift
```diff
@@ -7,6 +7,28 @@
...
+    case xhr
+    case beacon
+    case css
+    case document
+    case fetch
+    case font
+    case js
+    case media
+    case other
+}
```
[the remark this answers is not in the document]
ncreated: This is not needed now, as we receive `RUMResourceKind` from the user. But we may think of it in `RUMM-633 Add more Resource information` 👌 where a convenient helpers could be added to infer resource kind from the `Content-Type` header of an HTTP response.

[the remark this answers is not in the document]
buranmert: i mean our users will need to do mapping to be able to provide us `RUMResourceKind`, with `Kind: String` they can provide `url.httpMethod` value directly

Sources/Datadog/RUM/DataModels/RUMDataModelsMapping.swift
ncreated: Indeed, done 👌

Sources/Datadog/RUMMonitor.swift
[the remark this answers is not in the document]
ncreated: Ah, I thought this refers to `RUMResourceKind`. Yes, good idea, done for `RUMHTTPMethod` 👌. With `RUMResourceKind` it's not that easy as single kind may be backed by several MIME types (e.g. `.image` is `image/png`, `image/jpeg`, ...). This is a good fit for a separate helper in `RUMM-633`.

PR #205, opened 2020-08-07
RUMM-654 Use auto generated RUM models

ncreated:
### What and why?

⚙️ This PR replaces manually created RUM models with their `RUMDataModels` counterparts.

### How?

```diff
- RUMViewEvent
- RUMActionEvent
```

### Review checklist


Sources/Datadog/RUM/DataModels/RUMDataModelsMapping.swift
```diff
@@ -14,6 +14,15 @@ extension Int {
     }
 }
 
+extension UInt {
```
buranmert, line 17: we can implement `toInt64` in `extension BinaryInteger`, it should cover all our cases

ncreated: Could be 👍. I will incorporate next upcoming PR (too much rebasing to do it here).

PR #207, opened 2020-08-07
RUMM-602 Shopist: Checkout implemented

buranmert:
### What and why?

Shopist: an app to generate demo RUM data

### How?

Checkout is implemented
You can also input any discount code in Cart page

If checkout succeeds, the app goes back to homepage/categories
If checkout fails, the app presents an alert with the error
I copied `randomFail()` from android, %5 of the chance we present an error without making the request

### Visual

checkout

### Review checklist

### DIFF
--- a/Shopist/Shopist/API.swift
+++ b/Shopist/Shopist/API.swift
@@ -21,11 +21,56 @@ struct Product: Decodable, Equatable {
     let isInStock: Bool
 }
 
+struct Payment: Encodable {
+    struct Checkout: Encodable {
+        let cardNumber: String
+        let cvc: Int
+        let exp: String
+
+        init() {
+            cardNumber = "\(Self.random(length: 16))"
+            cvc = Self.random(length: 3)
+            exp = "\(Int.random(in: 1...12))/\(Int.random(in: 21...30))"
+        }
+
+        static private func random(length: UInt8) -> Int {
+            let max = Int(pow(10.0, Double(length)))
+            let min = max / 10
+            let output = Int.random(in: min..<max)
+            return output
+        }
+    }
+    let checkout = Checkout()
+
+    struct Response: Decodable {
+        let id: String?
+        let cartID: String?
+        let cardNumber: String
+        let cvc: Int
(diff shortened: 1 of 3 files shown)


Shopist/Shopist/API.swift
```diff
@@ -37,6 +82,28 @@ final class API {
...
+        request.setValue("application/json", forHTTPHeaderField: "Content-Type")
+        httpClient.request(request).response { response in
+            if let someError = response.error {
+                completion(.failure(someError))
+            } else if let someData = response.data {
+                do {
+                    let paymentResponse = try self.jsonDecoder.decode(Payment.Response.self, from: someData)
+                    completion(.success(paymentResponse))
+                } catch {
+                    completion(.failure(error))
```
ncreated, line 101: This leads to decoding error presented in UI when server responds with `500` and HTML page. How about considering the response code and improving the error message?

Shopist/Shopist/Scenes/CartViewController.swift
```diff
@@ -56,9 +81,37 @@ final class CartViewController: UITableViewController {
...
+                self.handleError(error)
+            }
+        }
+    }
+
+    private func handleSuccess() {
+        DispatchQueue.main.async {
+            cart.products.removeAll()
+            (self.presentingViewController as? UINavigationController)?.popViewController(animated: false)
+            self.dismissPage()
```
ncreated, line 102: When thinking about UITests automation (to play Shopist scenarios), maybe it would be easier to also present a success alert from here? Then we'll be just asserting the alert content, not hierarchy change.

Shopist/Shopist/API.swift
buranmert: resolved ✅ 

this line doesn't look outdated but adding `validate()` to line 93 creates a response error in case of non-2XX HTTP status code
and the error message is smth like `HTTP Status Code 500 is not an acceptable status code`

ncreated: Awesome 🚀

PR #209, opened 2020-08-10
RUMM-523 Add RUM with Logging integration

ncreated:
### What and why?

🚚 This PR Adds RUM with Logging integration, so users can see in RUM Explorer which logs were produced while given View was being displayed:

Screenshot 2020-08-10 at 14 32 00

Also, when `logger.error("message")` or `logger.critical("message")` is logged, the RUM Error event is created and associated with the active View:

Screenshot 2020-08-10 at 14 32 23

### How?

Two small components are created to keep `Logger` and `RUMMonitor` isolated:
* `LoggingWithRUMContextIntegration`,
* `LoggingWithRUMErrorsIntegration`.

I used a temporary `RUMMonitor.shared` property which will be re-visited in `RUMM-614` when adding a final initialization API for RUM.

### Review checklist


Sources/Datadog/Logger.swift
```diff
@@ -211,9 +223,16 @@ public class Logger {
...
             date: date,
-            attributes: LogAttributes(userAttributes: combinedAttributes, internalAttributes: nil),
+            attributes: LogAttributes(
+                userAttributes: combinedAttributes,
+                internalAttributes: rumContextIntegration?.currentRUMContextAttributes
+            ),
             tags: tags
         )
+
+        if level.rawValue >= LogLevel.error.rawValue {
```
buranmert, line 233: can we make `enum LogLevel: Comparable { ... }`?

Sources/Datadog/FeaturesIntegration/LoggingWithRUMIntegration.swift
```diff
@@ -0,0 +1,42 @@
...
+        return [
+            RUMLogAttributes.applicationID: rumContext.rumApplicationID,
+            RUMLogAttributes.sessionID: rumContext.sessionID.rawValue.uuidString.lowercased(),
+            RUMLogAttributes.viewID: rumContext.activeViewID?.rawValue.uuidString.lowercased(),
+        ]
+    }
+}
+
+/// Creates RUM Errors for Logs.
+internal struct LoggingWithRUMErrorsIntegration {
```
buranmert, line 37: do we have a case where we use `LogWithRUMContext` without `LogWithRUMErrors`? can't we merge them into a single type for the sake of simplicity?

Sources/Datadog/Logger.swift
ncreated: The `Comparable` methods are not automatically synthesized, so we'd still have to put the `.rawValue` somewhere. Here it's just in a single place, with no impact on the public interface.

Sources/Datadog/FeaturesIntegration/LoggingWithRUMIntegration.swift
ncreated: Yes. The `LogWithRUMContext` can be enabled/disabled with
```swift
logger.builder()
   .bundleWithRUM(...)
   .build()
```
but `LogWithRUMErrors` is always enabled. Also, I think the `LogWithRUMContext` will be needed for `span.log()` integration (when log is created through `span.log()`, the RUM context should also be injected).

PR #210, opened 2020-08-11
RUMM-656 Shopist: Manual RUM events

buranmert:
### What and why?

Until auto-instrumentation for RUM arrives, we add manual instrumentation calls to a few places in Shopist app

### How?

1. View tracking:
  * `Categories`, `Products`, `ProductDetails` and `Cart` view controllers are tracked
2. Resources
  * `GET` requests made from `Categories` and `Products` are tracked
3. User Actions
  * Cell taps in `Categories` and `Products`, pay action in `Cart` are tracked

## Known Issues

During this task, we noticed that discrete user actions are not tracked properly due to their timeout requirement.
This PR should be merged after a fix 🔜 

I'll add visuals from RUM Explorer after the fix too

### Review checklist

### DIFF
--- a/Shopist/Shopist/API.swift
+++ b/Shopist/Shopist/API.swift
@@ -6,6 +6,7 @@
 
 import Foundation
 import Alamofire
+import Datadog
 
 struct Category: Decodable {
     let id: String
@@ -73,13 +74,14 @@ final class API {
     }()
 
     func getCategories(completion: @escaping Completion<[Category]>) {
-        get("\(Self.baseURL)/categories.json", completion: completion)
+        let urlString = "\(Self.baseURL)/categories.json"
+        make(request: URLRequest(urlString), completion: completion)
     }
 
     func getItems(for category: Category, completion: @escaping Completion<[Product]>) {
         let categoryID = category.id
         let urlString = "\(Self.baseURL)/category_\(categoryID).json"
-        get(urlString, completion: completion)
+        make(request: URLRequest(urlString), completion: completion)
     }
 
     func checkout(with discountCode: String?, payment: Payment = Payment(), completion: @escaping Completion<Payment.Response>) {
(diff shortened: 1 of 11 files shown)


Shopist/Shopist/ExampleAppConfig.swift
```diff
@@ -11,6 +11,8 @@ struct ExampleAppConfig {
     let clientToken: String
     /// Service name used for logs and traces.
     let serviceName: String
```
ncreated, line 13: Let's rename this file 🙏 `ExampleAppConfig` → `ShopistAppConfig` or similar.

Shopist/Shopist/API.swift
```diff
@@ -109,11 +110,15 @@ final class API {
             fatalError("\(urlString) is not a valid URL!")
         }
         let request = URLRequest(url: url)
-
+        let httpMethod = RUMHTTPMethod(rawValue: request.httpMethod!) ?? .GET
+        rum?.startResourceLoading(resourceName: urlString, url: url, httpMethod: httpMethod)
         httpClient.request(request).response { response in
+            let statusCode = response.response?.statusCode
             if let someError = response.error {
+                rum?.stopResourceLoadingWithError(resourceName: urlString, error: someError, source: .network, httpStatusCode: statusCode)
```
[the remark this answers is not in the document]
buranmert: we have `/checkout` request which fails randomly
but it's not a resource it is a POST request, does it still qualify as a RUM Resource?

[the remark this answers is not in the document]
ncreated: Totally 👌, every request should be a RUM Resource. If you query the Android Shopist app for `@error.origin:network` it will list its Resource errors.

xcconfigs/Shopist.xcconfig
```diff
@@ -1,4 +1,6 @@
 #include "Datadog.xcconfig"
 
+// TODO: RUMM-664 temporary value, otherwise test cases fail in CI
+RUM_APPLICATION_ID=12345
```
ncreated, line 4: Like said, what about also including the `DATADOG_CLIENT_TOKEN=` here 🙂? This will be more obvious on what to fill in when creating the `Sophist.local.xcconfig`:
```suggestion
RUM_APPLICATION_ID=12345 // use your own RUM Application ID obtained on datadoghq.com
DATADOG_CLIENT_TOKEN=// use your own Client Token, generated for RUM_APPLICATION_ID
```

buranmert: yes makes sense 👍 if you don't provide it in `local.xcconfig` you will get fatalError at runtime 👌

PR #213, opened 2020-08-12
RUMM-660 Active user actions stopped with stopView

buranmert:
### What and why?

Active user actions were being discarded when their view is stopped
We should stop/finalize those user actions in that case

### How?

`UserActionScope` handles `StopView` command and sends the corresponding RUM event
This PR also limits `action.loadingTime` to hardcoded max duration values (0.1 sec for discrete actions, 10 for continuous actions)

### Review checklist

### DIFF
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMUserActionScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMUserActionScope.swift
@@ -68,39 +68,30 @@ internal class RUMUserActionScope: RUMScope, RUMContextProvider {
     // MARK: - RUMScope
 
     func process(command: RUMCommand) -> Bool {
-        if isContinuous { // e.g. "scroll"
-            if expired(currentTime: command.time) {
-                sendActionEvent(completionTime: command.time)
-                return false
-            }
-        } else { // e.g. "tap"
-            if timedOut(currentTime: command.time) && allResourcesCompletedLoading() {
-                sendActionEvent(completionTime: lastActivityTime)
-                return false
-            }
+        if let expirationTime = possibleExpirationTime(currentTime: command.time),
+           allResourcesCompletedLoading() {
+            sendActionEvent(completionTime: expirationTime)
+            return false
         }
 
         lastActivityTime = command.time
-
         switch command {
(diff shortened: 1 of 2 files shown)


Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMViewScopeTests.swift
```diff
@@ -247,6 +247,25 @@ class RUMViewScopeTests: XCTestCase {
 
     // MARK: - User Action Tracking
 
+    func test_finalizeActiveUserActionsOnStopView() throws {
```
ncreated, line 250: Let's keep the same behaviour-testing convention as used in all RUM scope tests:
```suggestion
    func testGivenActiveUserAction_whenViewIsStopped_itSendsUserActionEvent() throws {
```
It is way easier to understand the tests this way, as they map 1:1 to product requirements.

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMViewScopeTests.swift
```diff
@@ -247,6 +247,25 @@ class RUMViewScopeTests: XCTestCase {
...
+            identity: mockView,
+            attributes: [:],
+            startTime: Date()
+        )
+        XCTAssertTrue(scope.process(command: RUMStartViewCommand.mockWith(identity: mockView)))
+        let mockUserActionCmd = RUMAddUserActionCommand.mockAny()
+        XCTAssertTrue(scope.process(command: mockUserActionCmd))
+        XCTAssertFalse(scope.process(command: RUMStopViewCommand.mockWith(identity: mockView)))
+
+        let recordedActionEvents = try output.recordedEvents(ofType: RUMEvent<RUMAction>.self)
```
ncreated, line 263: Have you considered putting this test in `RUMUserActionScopeTests`? If you look closer at the assertions we do in each scope tests, the testing strategy will emerge: _"Scope X tests only assert `RUMEvents` send directly from this scope"_. This helps with tests granulation. Without a common strategy, we could be testing everything in the `RUMApplicationScopeTests`, which will lower the readability.

Issue #214, opened 2020-08-12
Compilation fails with Datadog SDK 1.3.0 using Cocoapods and Xcode 11.3.1


provTheodoreNewell:
### The issue

Compilation fails with 5 issues in a new Xcode 11.3.1 project with DataDog SDK 1.3.0:

1. `Value of type 'FileHandle' has no member 'seekToEnd'` in File.swift line 62
2. `Extraneous argument label 'contentsOf:' in call` in File.swift line 63
  1. Compiler-suggested fix: `Replace 'contentsOf: ' with ''`
3. `Value of type 'FileHandle' has no member 'readToEnd'` in File.swift line 94
4. `Ambiguous reference to member 'swizzle(using:)'` in URLSessionSwizzler.swift line 68
5. `Ambiguous reference to member 'swizzle(using:)'` in URLSessionSwizzler.swift line 109

---

buranmert:
hey @provTheodoreNewell 👋 thanks for reporting the issue!
we made a fix and are planning to ship it as a hotfix
i'll keep you posted once this is done

### The issue
we use some `FileHandle` methods which are only available in iOS 13.4+ and Xcode 11.3.1 has an inferior base SDK version
as a fix, we wrap this code with `#if compiler(>=5.2) ... #else ... #endif`; `compiler(5.2)` means Xcode 11.4 whose base SDK is iOS 13.4
in your case, Xcode 11.3.1, compiler shouldn't see those new methods and should compile the `#else` scope

provTheodoreNewell:
Thank you!

buranmert:
hi @provTheodoreNewell 👋 
`1.3.1` hotfix version is shipped to fix the issue 🚀 

please let me know if the issue persists, otherwise feel free to close the issue

provTheodoreNewell:
Resolved: `1.3.1` compiles with Xcode 11.3.1 🥳

PR #217, opened 2020-08-13
RUMM-669 Xcode 11.3.1 compilation fix

buranmert:
### What and why?

Xcode 11.3.1 doesn't have iOS 13.4 as base SDK
therefore `#available(iOS 13.4)` closures don't compile

For more: https://github.com/DataDog/dd-sdk-ios/issues/214

### How?

We added compiler version macros as workaround
```swift
#if compiler(>=5.2)
  if #available(iOS 13.4, *) { newMethod() }
  else { legacy() }
#else
legacy()
#endif
```

_Note:_ It's not trivial to run our tests in different Xcode versions with our CI provider

Screenshot 2020-08-13 at 14 19 59

### Release info

This PR is branched off from `1.3.0` tag and to be merged into `master`
We will ship this change as a hotfix once it is merged

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/Persistence/Files/File.swift
+++ b/Sources/Datadog/Core/Persistence/Files/File.swift
@@ -46,60 +46,84 @@ internal struct File: WritableFile, ReadableFile {
     func append(data: Data) throws {
         let fileHandle = try FileHandle(forWritingTo: url)
 
+        // NOTE: RUMM-669
+        // https://github.com/DataDog/dd-sdk-ios/issues/214
+        // https://en.wikipedia.org/wiki/Xcode#11.x_series
+        // compiler version needs to have iOS 13.4+ as base SDK
+        #if compiler(>=5.2)
+        /**
+         Even though the `fileHandle.seekToEnd()` should be available since iOS 13.0:
+         ```
+         @available(OSX 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+         public func seekToEnd() throws -> UInt64
+         ```
+         it crashes on iOS Simulators prior to iOS 13.4:
+         ```
+         Symbol not found: _$sSo12NSFileHandleC10FoundationE9seekToEnds6UInt64VyKF
+         ```
+         This is fixed in iOS 14/Xcode 12
+        */
         if #available(iOS 13.4, *) {
-            /**
-             Even though the `fileHandle.seekToEnd()` should be available since iOS 13.0:
-             ```
-             @available(OSX 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
-             public func seekToEnd() throws -> UInt64
-             ```
-             it crashes on iOS Simulators prior to iOS 13.4:
-             ```
-             Symbol not found: _$sSo12NSFileHandleC10FoundationE9seekToEnds6UInt64VyKF
-             ```
-            */
             defer { try? fileHandle.close() }
             try fileHandle.seekToEnd()
             try fileHandle.write(contentsOf: data)
         } else {
-            defer {
-                try? objcExceptionHandler.rethrowToSwift {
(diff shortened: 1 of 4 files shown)


Sources/Datadog/Core/Persistence/Files/File.swift
```diff
@@ -46,60 +46,60 @@ internal struct File: WritableFile, ReadableFile {
...
-             Even though the `fileHandle.seekToEnd()` should be available since iOS 13.0:
-             ```
-             @available(OSX 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
-             public func seekToEnd() throws -> UInt64
-             ```
-             it crashes on iOS Simulators prior to iOS 13.4:
-             ```
-             Symbol not found: _$sSo12NSFileHandleC10FoundationE9seekToEnds6UInt64VyKF
-             ```
-            */
```
ncreated, line 60: Why removing this comment 🤔? It's a valid explanation of the unobvious setup - we use `#available(iOS 13.4, *)` for the API which is marked as `@available(iOS 13.0, *)`. Next time we come here and we may remove it by mistake.

Sources/Datadog/Core/Persistence/Files/File.swift
```diff
@@ -46,60 +46,60 @@ internal struct File: WritableFile, ReadableFile {
...
+        return try legacyRead(from: fileHandle)
+        #endif
+    }
+
+    private func legacyRead(from fileHandle: FileHandle) throws -> Data {
+        let data = fileHandle.readDataToEndOfFile()
+        try? objcExceptionHandler.rethrowToSwift {
+            fileHandle.closeFile()
         }
+        return data
```
ncreated, line 102: nit: we can keep its original order
```suggestion
        try? objcExceptionHandler.rethrowToSwift {
            fileHandle.closeFile()
        }
        fileHandle.readDataToEndOfFile()
```

Sources/Datadog/Core/Persistence/Files/File.swift
buranmert: it doesn't compile with `#available(iOS 13)` ❌ 
maybe they fix the available versions were wrong in the previous iOS SDK and now fixed

Screenshot 2020-08-14 at 10 07 08
Screenshot 2020-08-14 at 10 07 26

Sources/Datadog/Core/Persistence/Files/File.swift
buranmert: previous code was closing `fileHandle` in `defer` block, this is the original order as `readDataToEndOfFile` doesn't throw and `closeFile` is called after reading

buranmert:
@ncreated i changed the order of parameters because apparently Xcode 11.3.1 has a problem with parameter ordering. it looks like we need to put the block parameter (instead of default value parameter) at the end

Sources/Datadog/Core/Persistence/Files/File.swift
ncreated: You're right. I updated my suggestion. Anyway, it does the same as the code in PR 👍.

Sources/Datadog/Core/Persistence/Files/File.swift
buranmert: this issue is fixed in iOS 14/Xcode 12, but Xcode 11.6 still has the issue: leaving the comment here

Sources/Datadog/Core/Persistence/Files/File.swift
buranmert: yes, it does the same thing, i go with shorter code 🩳

Sources/Datadog/Core/Persistence/Files/File.swift
ncreated: 👌

PR #219, opened 2020-08-13
RUMM-535 Resources and User Actions are linked with the active View

ncreated:
### What and why?

📦 This PR makes the Resources and User Actions being associated with the active View.

### How?

This corresponds to the case where multiple Views are started at a time, for example:
```swift
monitor.startView(viewController: master)
monitor.startView(viewController: details)
monitor.startResourceLoading(resourceName: "/resource/1")
```

In such case, prior to this PR, the `/resource/1` would appear in both, "master" and "details" Views in RUM Explorer. This PR updates this behaviour, and the Resource will be linked only to the "details" View.

This also solves the range of errors with manual (and further: auto) instrumentation, where `startView()` and `stopView()` are called from different `UIViewController` lifecycle callbacks, i.e. when calling "start" from `viewDidLoad()` and "stop" from `viewDidDisappear()`.

### Review checklist

### DIFF
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
@@ -31,6 +31,8 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
     let viewURI: String
     /// The start time of this View.
     private let viewStartTime: Date
+    /// Tells if this View is the active one. `true` for every new started View. `false` if any other View was started before this one is stopped.
+    private var isActiveView: Bool = true
 
     /// Number of Actions happened on this View.
     private var actionsCount: UInt = 0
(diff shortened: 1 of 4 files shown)


Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -85,27 +87,30 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
                 sendApplicationStartAction()
             }
             needsViewUpdate = true
+        case let command as RUMStartViewCommand where command.identity !== identity:
+            isActiveView = false
         case let command as RUMStopViewCommand where command.identity === identity:
+            isActiveView = false
```
buranmert, line 93: let's say the app pushes `VC1` and presents `VC2` and then dismisses `VC2`:
1. `startView(VC1)` (active view is VC1)
2. `startView(VC2)` (active view is VC2)
3. `stopView(VC2)` (shouldn't active view be VC1?)

if i understand well, currently in this case we are left with no active view, correct?

ncreated: This would mean the incorrect instrumentation. When a new View is started, the old one should be stopped. Transferring it to the `UIKit` language: when presenting `VC2`, the `VC1` should be stopped, and when dismissing the `VC2`, `VC1` should be started again.

We know this is not the most convenient way for `UIKit`, but what we do here is complying to the RUM product requirements, not `UIKit` specifics. The RUM requirements are loosely defined, so they leave the door open for any integration specifics (`UIKit`, `SwiftUI`, Android activities & fragments, web frameworks, ...). This sort of problems will be solved later with specific `UIKit` helpers and in the auto instrumentation components. It should not impact the general rules of the RUM.

PR #220, opened 2020-08-13
RUMM-671 DatadogObjC target is fixed for SPM

buranmert:
### What and why?

`DatadogObjC` wasn't compiling with _@objc attribute used without importing module 'Foundation'_ errors
This only happens in Swift Package Manager; other dependency managers work fine
For more info: https://github.com/DataDog/dd-sdk-ios/issues/218

### How?

I added `import Foundation` lines to source files.
I also made dep-manager-test/spm import `DatadogObjC` instead of `Datadog`, this should test both targets at once.
Note: I'm still investigating how come Xcode manages to build the project without those imports 🤔 

Release info

This PR is branched off from 1.3.0 tag and to be merged into master
We will ship this change as a hotfix once it is merged

### Review checklist

### DIFF
--- a/Sources/DatadogObjc/OpenTracing/OTGlobal+objc.swift
+++ b/Sources/DatadogObjc/OpenTracing/OTGlobal+objc.swift
@@ -4,6 +4,7 @@
  * Copyright 2019-2020 Datadog, Inc.
  */
 
+import Foundation
 import Datadog
 
 @objcMembers
--- a/Sources/DatadogObjc/OpenTracing/OTNoop.swift
+++ b/Sources/DatadogObjc/OpenTracing/OTNoop.swift
@@ -4,6 +4,8 @@
  * Copyright 2019-2020 Datadog, Inc.
  */
 
+import Foundation
+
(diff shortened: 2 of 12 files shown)


ncreated:
> Release info
>
> This PR is branched off from 1.3.0 tag and to be merged into master
> We will ship this change as a hotfix once it is merged

If we plan to include two patches in a hotfix, please create the `hotfix/1.3.1` branch and point both PRs to it. This way we'll have two merge commits, keeping the git history right. Otherwise, both patches will be included in a single, `RUMM-669` merge commit, which is not correct.

buranmert:
@ncreated `.linkedFramework("Foundation")` didn't work, it's mostly for [C++ based libraries and non-modular libraries](https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html#linkersetting) according to docs and [only very few projects use it](https://github.com/search?q=.linkedFramework%28%22Foundation%22&type=Code)
we cannot easily miss that because otherwise `dependency-manager-tests-spm` will fail ❌ 

it's weird that Xcode can compile the project, when i create new project it fails when `import Foundation` is absent 🤔

PR #223, opened 2020-08-14
RUMM-636 Resources loading can exceed the View lifespan

ncreated:
### What and why?

📦 This PR makes the View wait for all its pending Resources until it gets finished.

### How?

The `RUMViewScope` is marked `didReceiveStopCommand` on `stopView()`, but it remains in memory until its Resources count goes down to `0`.

Also, the corresponding behaviour test was added to `RUMMonitorTests`:
* `testStartingLoadingResourcesFromTheFirstView_thenStartingAnotherViewWhichAlsoLoadsResources`

### Review checklist


ncreated:
> we may also add a `if !didReceiveStopCommand` check for `startXXX` commands
> currently it's possible to start new UAs/resources on a stopped view

Those are excluded by `isActiveView` check, e.g.:
```swift
case let command as RUMStartResourceCommand where isActiveView:
   startResource(on: command)
```
The `isActiveView` is set when View receives the "stop" command or when another View is started.

PR #224, opened 2020-08-17
RUMM-590 Add RUM Sessions sampling

ncreated:
### What and why?

📦 This PR adds the option to sample RUM Sessions.

### How?

The public API was added to `Datadog.Configuration` builder:
```swift
/// Sets the sampling rate for RUM Sessions.
///
/// - Parameter rumSessionsSamplingRate: the sampling rate must be a value between `0.0` and `100.0`. A value of `0.0`
/// means no RUM events will be sent, `100.0` means all sessions will be kept (default value is `100.0`).
public func set(rumSessionsSamplingRate: Float)
```

RUM Events from sampled-out session are not send to Datadog. Such session's scope is marked as `shouldBeSampledOut` and it doesn't create any child scopes.

When RUM Session is sampled, `RUMContext` attributes are not added to Logs nor Traces.

### Review checklist

### DIFF
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
@@ -25,8 +25,12 @@ internal class RUMSessionScope: RUMScope, RUMContextProvider {
     unowned let parent: RUMContextProvider
     private let dependencies: RUMScopeDependencies
 
-    /// This Session UUID.
+    /// This Session UUID. Equals `.nullUUID` if the Session is sampled.
     let sessionUUID: RUMUUID
+    /// RUM Session sampling rate.
+    private let samplingRate: Float
+    /// Tells if events from this Session should be sampled-out (not send).
+    private let shouldBeSampledOut: Bool
     /// The start time of this Session.
     private let sessionStartTime: Date
     /// Time of the last RUM interaction noticed by this Session.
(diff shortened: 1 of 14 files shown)


Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
```diff
@@ -89,6 +94,10 @@ internal class RUMSessionScope: RUMScope, RUMContextProvider {
         }
         lastInteractionTime = command.time
 
+        if shouldBeSampledOut {
+            return true
+        }
```
buranmert, line 99: if we already know that the session is discarded (sampled-out) at init method, why do we wait until the next command?
i thought `ApplicationScope` would be responsible for sampling:
```swift
// AppScope
process(cmd) {
  case is StartSession and shouldSample(at: rate):
    sessionScope = SessionScope(...)
  // if shouldSample(at: rate) returns false, don't create sessionScope
}
```
am i missing something?

buranmert (changes requested):

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
ncreated: > if we already know that the session is discarded (sampled-out) at init method, why do we wait until the next command?

To let the session expire or time-out when any command arrives at certain time.

> ```swift
> // AppScope
> process(cmd) {
>   case is StartSession and shouldSample(at: rate):
>     sessionScope = SessionScope(...)
>   // if shouldSample(at: rate) returns false, don't create sessionScope
> }
> ```

This wouldn't be correct - if the Session scope is not created, how would we know that the RUM Session has expired or timed-out? Without this information, we wouldn't know that the RUM Session should re-created.

> i thought ApplicationScope would be responsible for sampling

This would mean that the `RUMApplicationScope` manages session creation, session sampling and session time-out/expiration. With current model, the responsibilities are separated:
* `RUMApplicationScope` - manages session creation,
* `RUMSessionScope` - manages its time-out/expiration and sampling.

buranmert (approved):

PR #225, opened 2020-08-17
RUMM-664: Shopist demo runs

buranmert:
### What and why?

We need Shopist app to generate RUM demo data

### How?

The app was implemented, this PR implements its UI tests and CI configuration to run them

Also this PR contains linter error fixes in Shopist project
⚠️ In order to make review easier, you can filter "Shopist CI setup is done" commit ⚠️ 

### Visuals
dashboard
Screenshot 2020-08-25 at 11 30 03

`CartViewController` has resources, traces, actions, errors and logs ✅ 

### Review checklist

- [x] Add more events (eg: user actions, resources, spans) to Shopist app


Shopist/ShopistUITests/ShopistUITests.swift
```diff
@@ -1,18 +1,57 @@
...
+            cardButton.tap()
+            if !app.alerts.firstMatch.waitForExistence(timeout: 5.0) {
+                fatalError("Alert should show up!")
+            }
+            app.alerts.firstMatch.buttons.firstMatch.tap()
+
+            if sessionIndex == sessionCount {
+                Thread.sleep(forTimeInterval: 5.0)
+            }
+            app.terminate()
```
ncreated, line 37: This is very hard to read as it's tightly coupled with implementation details. It will be also hard to maintain on even small UI changes and impossible to reuse. How about using the [Page Objects](https://github.com/DataDog/dd-sdk-ios/blob/rum/Tests/DatadogIntegrationTests/UITestsHelpers.swift) approach (it's kind of a standard in iOS UI Testing)? This is exactly what we use for [RUM integration tests](https://github.com/DataDog/dd-sdk-ios/blob/rum/Tests/DatadogIntegrationTests/UITestsHelpers.swift).

The PO is just a basic abstraction of the screens. With just few changes, this test will become something like this:
```swift
let app = ShopistApplication()
let categoriesListsScreen = app.launch()
let productsListScreen = categoriesListsScreen.tapTheLastCategory()
let productScreen = productsListScreen.tapTheLastProduct()
productScreen.addToCart()
let cartScreen = productScreen.openCart()
let completionAlert = cartScreen.payWithCreditCard()
app.terminate()
``` 

It's much more readable than direct calls to `XCUI` elements. When building next scenarios, we will just copy & paste the interesting parts, gaining the code reuse. And when we slightly change the UI, we won't have to modify tests scenarios each time.

Shopist/ShopistUITests/ShopistUITests.swift
```diff
@@ -1,18 +1,57 @@
...
+    override func setUp() { // swiftlint:disable:this empty_xctest_method
         continueAfterFailure = false
     }
 
-    func testAppCanBeLaunched() {
+    func testRunFullCheckoutFlow() {
+        let sessionCount = 3
         let app = XCUIApplication()
-        app.launch()
+        for sessionIndex in 0...sessionCount {
```
ncreated, line 17: Why running it 4 times?

Shopist/ShopistUITests/ShopistUITests.swift
```diff
@@ -1,18 +1,57 @@
...
+            if sessionIndex == sessionCount {
+                Thread.sleep(forTimeInterval: 5.0)
+            }
+            app.terminate()
+        }
+    }
+}
+
+private extension XCUIElement {
+    func goToOffScreenCell(_ identifier: String) -> XCUIElement? {
```
ncreated, line 43: nit: would be more elegant to call it `findCell(identifier: String)`, because we're using it with `.tap()`:
```swift
app.collectionViews.firstMatch.findCell(identifier: "lastCell")?.tap()
```
The "go to off screen cell" refers to the implementation, which may not always be true.

Shopist/Makefile
```diff
@@ -0,0 +1,11 @@
...
+
+define SECRETS
+RUM_APPLICATION_ID=${RUM_APPLICATION_ID}\n
+DATADOG_CLIENT_TOKEN=${DATADOG_CLIENT_TOKEN}\n
+SHOPIST_BASE_URL=${SHOPIST_BASE_URL}\n
+endef
+export SECRETS
+
+setup:
+	@echo $$SECRETS > ../xcconfigs/Shopist.local.xcconfig;
```
ncreated, line 11: Which branch / git refs will we use to seed the Demo data? How will we ensure that we don't break it with recent development? I.e. if the recent change to `master` breaks the data consistency, how we will notice that we're breaking the data in Demo environment?

Shopist/ShopistUITests/ShopistUITests.swift
buranmert: 4 is only for debugging, it will be 20 something when we run them regularly.
android runs the same scenarios many times to generate multiple sessions, i don't know if they do it in source code or Gradle config but this is what @xgouchet told me.

Shopist/Makefile
buranmert: Screenshot 2020-08-18 at 15 56 27

1. `shopist_demo`: Shown in Demo org dashboards, to be triggered regularly
2. `shopist_org2`: Shown in Org2 dashboards, to be triggered manually for verification

Shopist/ShopistUITests/ShopistUITests.swift
buranmert: after discussing with @xgouchet, our full checkout flow scenario seems enough for demos (i just need to add some fake resources, spans and logs to pages)
i can add comments to UI tests for readability but it's not necessary to add more complexity for now (as we don't reuse them)

Shopist/ShopistUITests/ShopistUITests.swift
xgouchet: One Bitrise job runs around 100 scenarios (each one takes around 20 secs to run on average). This allows us to "fill" 30 minutes of data for each run.

Shopist/Makefile
xgouchet: Actually on Android the Shopist and library development are done in two separate repository, to ensure we don't couple the Shopist app with the lib implementation. That way we also simulate exactly how an application would work "in the wild" : 
 - fetch the sdk library from the appropriate dependency management system
 - only use released versions
 - no possibility to use shorthand tricks or hacks

Shopist/ShopistUITests/ShopistUITests.swift
ncreated: > i can add comments to UI tests for readability but it's not necessary to add more complexity for now (as we don't reuse them)

Why adding comments instead of naming things in a better way 🤔? It's not a big deal to move the call to a separate class/method, no?

Shopist/ShopistUITests/ShopistUITests.swift
```diff
@@ -1,18 +1,57 @@
...
+            app.collectionViews.firstMatch.goToOffScreenCell("lastCell")?.tap()
+
+            let navBar = app.navigationBars.firstMatch
+            navBar.buttons["cart.badge.plus"].tap()
+            navBar.buttons["cart"].tap()
+
+            let cardButton = app.navigationBars["Cart"].buttons["creditcard"]
+            cardButton.tap()
+            if !app.alerts.firstMatch.waitForExistence(timeout: 5.0) {
+                fatalError("Alert should show up!")
```
ncreated, line 30: If it crashes for, let's say 3th iteration out of 20, we will lost data for 17 iterations, leaving the gap in time series. We shouldn't crash the tests.

Shopist/ShopistUITests/ShopistUITests.swift
ncreated: I see 👌. Then we must check how does it behave in case of failure. Currently we set:
```swift
continueAfterFailure = false
```
which means "stop UITests on the first failure". So if the 1st iteration fails, no more data from given CI run will be send to demo org (leaving the gap in demo fixture).

buranmert: as the demo data must be consistent all the time, if anything goes wrong shouldn't we discard the run? a previous batch will be used at worst case. if we continue after failure, inconsistent data will be sent to demo org

Shopist/ShopistUITests/ShopistUITests.swift
buranmert: as the demo data must be consistent all the time, if anything goes wrong shouldn't we discard the run? a previous batch will be used at worst case. if we continue after failure, inconsistent data will be sent to demo org

Shopist/Makefile
ncreated: @xgouchet the idea for iOS is to run `shopist_demo` only from the last release tag - this should ensure the same level of safety 👌.

Shopist/ShopistUITests/ShopistUITests.swift
ncreated: Data consistency will be ensured with the help of `shopist_org2` workflow run on the `master` or feature branch. What I mean here is inconsistency due to tests flakiness, i.e. slow CI run exceeding this `5.0s` timeout.

buranmert: there should not be that much flakiness in those scenarios, if it happens then there is probably an actual problem rather than just flakiness

let's say Shopist backend is down for a moment, do we prefer no new data or wrong data in RUM Explorer?
if wrong data is not a problem then `shopist_org2` workflow is only nice-to-have and not needed.
if wrong data is a problem then we shouldn't send data from unexpected scenarios.

buranmert:
as one last thing i'm trying to add fake `computeCart` spans to the app 🚧 
apart from that, every page has user actions, resources and random errors now ✅

ncreated (changes requested):
Can't review more, as this branch doesn't build:

Screenshot 2020-08-24 at 12 02 15]

Changing the PR to "draft".

Shopist/Shopist/API.swift
```diff
@@ -89,38 +103,82 @@ final class API {
...
-            } else if let someData = response.data {
-                rum?.stopResourceLoading(resourceName: resourceName, kind: .fetch, httpStatusCode: statusCode)
+            } else if let someData = result.data {
+                rum?.stopResourceLoading(
+                    resourceName: resourceName,
+                    kind: .fetch,
+                    httpStatusCode: statusCode,
+                    size: UInt64(someData.count)
+                )
+                let decodingSpan = tracer.startSpan(operationName: "decoding response data")
```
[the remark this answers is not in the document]
buranmert: it's the child of active network span
Screenshot 2020-08-24 at 14 15 49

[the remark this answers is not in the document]
ncreated: Where is this network span created? I can only see:
```swift
let span = tracer.startSpan(operationName: request.url?.path ?? "network request")
```
but it's not marked as `.setActive()`.  And auto-instrumentation is disabled:
```swift
// Currently, SDK doesn't auto-trace Alamofire requests
// .set(tracedHosts: [API.baseHost])
```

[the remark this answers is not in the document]
buranmert: pulled `setActive()` commits and rebased ✅

Shopist/Shopist/Scenes/CheckoutViewController.swift
```diff
@@ -55,38 +62,73 @@ final class CartViewController: UITableViewController {
...
+    }
+
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        viewDidAppearDate = Date()
     }
 
     override func viewWillDisappear(_ animated: Bool) {
         super.viewWillDisappear(animated)
         rum?.stopView(viewController: self)
```
ncreated, line 90: When Checkout RUM View gets closed, the underlying view is not started, resulting with no active RUM View until some navigation action is taken:

Screenshot 2020-09-02 at 11 12 59

ncreated (changes requested):
Few inline feedbacks and two questions:
* what about the IP randomization? I think we need a JIRA to address this.
* when will we configure the `shopist_demo` trigger? I think we need a JIRA to address this.

Shopist/Shopist/API.swift
```diff
@@ -75,52 +89,162 @@ final class API {
...
         let resourceName = url.pathComponents.joined()
         let httpMethod = RUMHTTPMethod(rawValue: request.httpMethod!)!
+        // swiftlint:enable force_unwrapping
         rum?.startResourceLoading(resourceName: resourceName, url: url, httpMethod: httpMethod)
-        httpClient.request(request).validate().response { response in
-            let statusCode = response.response?.statusCode
-            if let someError = response.error {
+
+        let tracer = Global.sharedTracer
+        let span = tracer.startSpan(operationName: request.url?.path ?? "network request").setActive()
```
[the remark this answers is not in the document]
buranmert: 👍  `setRootSpan` wasn't available during dev of this PR

Shopist/Shopist/AppDelegate.swift
```diff
@@ -42,6 +75,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
         // Register global `RUMMonitor`
         Global.rum = RUMMonitor.initialize()
+        rum?.addAttribute(forKey: "usr.handle", value: user.email)
```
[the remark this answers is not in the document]
buranmert: historical reasons i guess. when android did their shopist, `rum-events` wasn't around and they used `usr.handle` instead. sales people still seem to look for it (although they can get the same info from `usr.email` now)

Shopist/Shopist/AppDelegate.swift
```diff
@@ -7,29 +7,62 @@
 import UIKit
 import Datadog
 
-fileprivate(set) var logger: Logger!
+internal fileprivate(set) var logger: Logger! // swiftlint:disable:this implicitly_unwrapped_optional
+internal let appConfig = AppConfig(serviceName: "ios-sdk-shopist-app")
+internal var rum: DDRUMMonitor? { Global.rum }
```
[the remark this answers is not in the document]
buranmert: 👍  `Global.rum` wasn't around when this PR started, i used global `rum` variable first and it stayed. we can replace them now

Shopist/Shopist/AppDelegate.swift
```diff
@@ -7,29 +7,62 @@
 import UIKit
 import Datadog
 
-fileprivate(set) var logger: Logger!
+internal fileprivate(set) var logger: Logger! // swiftlint:disable:this implicitly_unwrapped_optional
+internal let appConfig = AppConfig(serviceName: "ios-sdk-shopist-app")
```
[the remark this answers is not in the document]
buranmert: 👍 now `appConfig.serviceName` (`io.shopist.ios`) is used in `DatadogConfiguration`

Shopist/Shopist/AppDelegate.swift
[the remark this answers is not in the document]
ncreated: OK, clear 👌. Maybe it's worth adding the comment above that coded, so this can be removed in the future once everyone sells it by `usr.email` 🙌.

Shopist/Shopist/AppDelegate.swift
```diff
@@ -7,29 +7,62 @@
...
-
         // Initialize Datadog SDK
         Datadog.initialize(
             appContext: .init(),
             configuration: Datadog.Configuration
                 .builderUsing(
                     rumApplicationID: appConfig.rumAppID,
-                    clientToken: appConfig.rumClientToken,
+                    clientToken: appConfig.clientToken,
                     environment: "tests"
```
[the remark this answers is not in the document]
buranmert: `shop.ist` is an illegal service name, apparently `.` is invalid char
i make it `shopist` 👍

Shopist/Shopist/Extensions/UIImageView+Network.swift
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+import UIKit
+import AlamofireImage
+
+internal extension UIImageView {
+    func setImage(with url: URL) {
+        rum?.startResourceLoading(resourceName: url.path, url: url, httpMethod: .GET)
+        // randomizing cacheKey forces to make new request
+        let cacheKey = UUID().uuidString
```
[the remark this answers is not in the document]
buranmert: AF docs says we should use `config.urlCache = nil` while init'ing `ImageDownloader`
done ✅

bitrise.yml
```diff
@@ -13,7 +13,6 @@ workflows:
     - run_linter
     - run_unit_tests
     - run_integration_tests
-    - run_ui_tests
```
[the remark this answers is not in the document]
buranmert: `shopist-sanity-check` step added: it builds shopist for testing ✅

Shopist/Shopist/AppDelegate.swift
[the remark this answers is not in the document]
ncreated: So we have a different definition of valid `env` between platforms. This must be unified. I created JIRA, so this will be addressed later:
* `RUMM-700 Investigate the `env` name validation in Android RUM and iOS RUM`

buranmert:
@ncreated 
> what about the IP randomization?

we have a ticket for that ✅ 

> when will we configure the shopist_demo trigger?

based on Android team's experience, this should be an almost-trivial task. 
i'll keep the current jira in progress until trigger is done, sounds okay?

ncreated (approved):
👍

PR #230, opened 2020-08-25
RUMM-677 Fix: Ongoing UAs stop before discarded

buranmert:
### What?

Screenshot 2020-08-25 at 17 07 12

**When** the user
1. taps on button with plus to add the item to cart
2. taps on cart button to go to cart page

**Expected:** RUM Explorer shows both actions associated with this view
**Actual:** RUM Explorer shows only the last action (_Go to cart_) and ignores the previous actions

### Why?

User actions are started with `StartUserAction` and stopped with `StopUserAction` or they timeout.
Discrete user actions rely on timeouts. 
In order to trigger a timeout, an arbitrary command must be sent to the user action scope; otherwise, they don't create any RUM events.

#### Before
In case of `StartUserAction`: 
1. create a new action scope
2. replace the existing one with it
3. propagate the command to it

#### Now
In case of `StartUserAction`: 
1. create a new action scope
2. append it to private user actions array
3. propagate the command to elements in array **until one returns true**
4. keep the first true-returning (ie: active) element, discard the rest

### Review checklist


buranmert:
closing in favor of https://github.com/DataDog/dd-sdk-ios/pull/231
it is simpler

PR #231, opened 2020-08-25
RUMM-677 Fix bug with User Action not being send

ncreated:
### What and why?

🐞 This PR fixes the bug with discrete User Action event not being send when the UA expires and the next propagated command is `RUMStartUserActionCommand` or `RUMAddUserActionCommand`.

### How?

To reproduce the bug:
```swift
monitor.startView(viewController: mockView)
// wait 1s
monitor.registerUserAction(type: .tap, name: "1st action")
// wait 1s
monitor.registerUserAction(type: .swipe, name: "2nd action")
```

I added unit test to reproduce the bug and fixed it by enforcing the command processing in UA scope before it gets replaced with the next instance.

### Review checklist

### DIFF
--- a/Tests/DatadogTests/Datadog/RUMMonitorTests.swift
+++ b/Tests/DatadogTests/Datadog/RUMMonitorTests.swift
@@ -330,6 +330,37 @@ class RUMMonitorTests: XCTestCase {
             }
     }
 
+    func testStartingView_thenTappingButton_thenTappingAnotherButton() throws {
+        RUMFeature.instance = .mockByRecordingRUMEventMatchers(
+            directory: temporaryDirectory,
+            dependencies: .mockWith(
+                dateProvider: RelativeDateProvider(startingFrom: Date(), advancingBySeconds: 1)
+            )
+        )
+        defer { RUMFeature.instance = nil }
+
+        let monitor = RUMMonitor.initialize()
+
+        monitor.startView(viewController: mockView)
+        monitor.registerUserAction(type: .tap, name: "1st action")
+        monitor.registerUserAction(type: .swipe, name: "2nd action")
+        monitor.stopView(viewController: mockView)
+
+        let rumEventMatchers = try RUMFeature.waitAndReturnRUMEventMatchers(count: 4)
+        try rumEventMatchers.lastRUMEvent(ofType: RUMAction.self) { $0.action.target?.name == "1st action" }
+            .model(ofType: RUMAction.self) { rumModel in
+                XCTAssertEqual(rumModel.action.type, .tap)
+            }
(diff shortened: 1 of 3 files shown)


buranmert (changes requested):
🎗️: also `startContinuousUserAction` and `addDiscreteUserAction` need `if userActionScope == nil { userActionScope = UserActionScope(...) }` check due to the product requirements
[Android counterpart code](https://github.com/DataDog/dd-sdk-android/blob/master/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt#L150)

Tests/DatadogTests/Datadog/RUMMonitorTests.swift
```diff
@@ -330,6 +330,37 @@ class RUMMonitorTests: XCTestCase {
             }
     }
 
+    func testStartingView_thenTappingButton_thenTappingAnotherButton() throws {
```
buranmert, line 333: shouldn't this case be in `ViewScopeTests` as the fix done in `ViewScope`?

ncreated: We have two levels of RUM unit test:
* small tests - defined per scope (e.g. `RUMViewScopeTests`, `RUMUserActionScopeTests`),
* big tests  - defined in `RUMMonitorTests`.

Small tests use the scope instance directly, pass the `RUMCommand` for processing and expect on the scope state + assert produced events.

Big tests use the `RUMMonitor` public interface and assert uploaded events.

As big tests are much wider (they cover more code), they should be considered as more valuable. All the behaviours expressed in big tests can be also written as small tests, but this will produce longer testing code, more fragile to the implementation changes (as small test are closer to the implementation).

I decided to add the big test for this bug, as the bug was reported as user scenario (tap one button, wait, tap another button). I could move this test to the `RUMUserActionScopeTests` (rather than `RUMViewScopeTests` as we want to test the behaviour of UA Scope), but it won't be as widely covering as here. It will only assert that _"when expired UA scope receives the `RUMRegisterUserActionCommand`, it sends the `RUMAction` event and gets closed"_. Or I can add one more small test in the UA scope. WDYT @buranmert ?

buranmert: so long as we have this test case, it doesn't matter much to me where :)

i was asking bc to my understanding the bug&fix is about the way `ViewScope` manages `UserActionScope`, so `ViewScope` is the unit to test here. if i understand well, you see `RUMMonitorTests` like user scenarios 👌

buranmert (approved):

PR #232, opened 2020-08-26
RUMM-638 Global RUM attributes

buranmert:
### What and why?

Users may want to use common attributes in their RUM events
We should provide them a convenient way to set those attributes

### How?

Now users can add, update or remove global attributes in their `RUMMonitor` instances
```swift
rumMonitor.setAttribute(forKey: "some key", value: "some value") // adds an attribute
rumMonitor.setAttribute(forKey: "some key", value: "some other value") // updates the value
rumMonitor.setAttribute(forKey: "some key", value: nil) // removes the key
let globalAttrs = rumMonitor.attributes // returns current global attributes to be added to events
```

Note: `attributes` is thread-safe

### Review checklist

### DIFF
--- a/Sources/Datadog/DDRUMMonitor.swift
+++ b/Sources/Datadog/DDRUMMonitor.swift
@@ -11,7 +11,7 @@ import UIKit
 /// `DDRUMMonitor` allows you to record User events that can be explored and analyzed in Datadog Dashboards.
 /// You can only have one active `RUMMonitor`, and should register/retrieve it from the `Global` object.
 public class DDRUMMonitor {
-    // MARK: - Public
+    // MARK: - Public methods
 
     /// Notifies that the View starts being presented to the user.
     /// - Parameters:
@@ -122,6 +122,23 @@ public class DDRUMMonitor {
     ///   - name: the User Action name
     ///   - attributes: custom attributes to attach to the User Action.
     public func registerUserAction(type: RUMUserActionType, name: String, attributes: [AttributeKey: AttributeValue]? = nil) {
+    }
+
+    // MARK: - Attributes
+
+    /// Adds a custom attribute to all future commands sent by this monitor.
+    /// - Parameters:
+    ///   - key: key for this attribute. See `AttributeKey` documentation for information about
+    ///   nesting attribute values using dot `.` syntax.
+    ///   - value: any value that conforms to `AttributeValue` typealias. See `AttributeValue` documentation
+    ///   for information about nested encoding containers limitation.
+    public func addAttribute(forKey key: AttributeKey, value: AttributeValue) {
+    }
+
+    /// Removes the custom attribute from all future commands sent by this monitor.
+    /// Previous commands won't lose this attribute if they were created prior to this call.
+    /// - Parameter key: key for the attribute that will be removed.
+    public func removeAttribute(forKey key: AttributeKey) {
     }
 
     // MARK: - Internal
--- a/Sources/Datadog/RUMMonitor.swift
+++ b/Sources/Datadog/RUMMonitor.swift
@@ -56,6 +56,8 @@ public class RUMMonitor: DDRUMMonitor {
     internal let contextProvider: RUMCurrentContext
     /// Time provider.
     private let dateProvider: DateProvider
+    /// Attributes associated with every command.
+    private var rumAttributes: [AttributeKey: AttributeValue] = [:]
     /// Queue for processing RUM commands off the main thread and providing current RUM context.
     private let queue = DispatchQueue(
         label: "com.datadoghq.rum-monitor",
@@ -125,7 +127,7 @@ public class RUMMonitor: DDRUMMonitor {
         process(
             command: RUMStartViewCommand(
                 time: dateProvider.currentDate(),
-                attributes: attributes ?? [:],
+                attributes: aggregate(attributes),
                 identity: viewController
             )
         )
@@ -135,7 +137,7 @@ public class RUMMonitor: DDRUMMonitor {
         process(
             command: RUMStopViewCommand(
                 time: dateProvider.currentDate(),
-                attributes: attributes ?? [:],
+                attributes: aggregate(attributes),
                 identity: viewController
             )
         )
@@ -152,7 +154,7 @@ public class RUMMonitor: DDRUMMonitor {
                 message: message,
                 source: source,
                 stack: stack,
-                attributes: attributes ?? [:]
+                attributes: aggregate(attributes)
             )
         )
     }
@@ -163,7 +165,7 @@ public class RUMMonitor: DDRUMMonitor {
                 time: dateProvider.currentDate(),
                 error: error,
                 source: source,
-                attributes: attributes ?? [:]
+                attributes: aggregate(attributes)
             )
         )
     }
@@ -173,7 +175,7 @@ public class RUMMonitor: DDRUMMonitor {
             command: RUMStartResourceCommand(
                 resourceName: resourceName,
                 time: dateProvider.currentDate(),
-                attributes: attributes ?? [:],
+                attributes: aggregate(attributes),
                 url: url.absoluteString,
                 httpMethod: httpMethod
             )
@@ -185,7 +187,7 @@ public class RUMMonitor: DDRUMMonitor {
             command: RUMStopResourceCommand(
                 resourceName: resourceName,
                 time: dateProvider.currentDate(),
-                attributes: attributes ?? [:],
+                attributes: aggregate(attributes),
                 kind: kind,
                 httpStatusCode: httpStatusCode,
                 size: size
@@ -201,7 +203,7 @@ public class RUMMonitor: DDRUMMonitor {
                 error: error,
                 source: source,
                 httpStatusCode: httpStatusCode,
-                attributes: attributes ?? [:]
+                attributes: aggregate(attributes)
             )
         )
     }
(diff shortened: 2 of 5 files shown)


Tests/DatadogTests/Datadog/RUMMonitorTests.swift
```diff
@@ -475,6 +489,7 @@ class RUMMonitorTests: XCTestCase {
             case 9: monitor.stopUserAction(type: .scroll)
             case 10: monitor.registerUserAction(type: .tap, name: .mockRandom())
             case 11: _ = monitor.dd.contextProvider.context
+            case 12: monitor.setAttribute(forKey: String.mockRandom(), value: String.mockRandom())
```
ncreated, line 492: 👌 👌

Sources/Datadog/RUMMonitor.swift
```diff
@@ -254,6 +270,13 @@ public class RUMMonitor: DDRUMMonitor {
 
     // MARK: - Private
 
+    private func aggregate(_ localAttributes: [AttributeKey: AttributeValue]?) -> [AttributeKey: AttributeValue] {
+        if let someAttributes = localAttributes {
+            return attributes.merging(someAttributes) { $1 }
+        }
+        return attributes
```
ncreated, line 277: We have this extension defined in `RUMScope.swift`:
```swift
extension Dictionary where Key == AttributeKey, Value == AttributeValue {
    /// Merges given `rumCommandAttributes` to current dictionary, by overwriting values.
    mutating func merge(rumCommandAttributes: [AttributeKey: AttributeValue]?) {
        guard let additionalAttributes = rumCommandAttributes else {
            return
        }
        merge(additionalAttributes) { _, new in new }
    }
}
```
It is also covered with tests (to ensure proper merging priority). Maybe we can reuse it here?

Tests/DatadogTests/Datadog/RUMMonitorTests.swift
```diff
@@ -551,4 +566,22 @@ class RUMMonitorTests: XCTestCase {
...
+    }
+
+    private func verifyGlobalAttributes(in matchers: [RUMEventMatcher]) {
+        for matcher in matchers {
+            expectedAttributes.forEach { attrKey, attrValue in
+                XCTAssertEqual(try? matcher.attribute(forKeyPath: attrKey), attrValue)
+            }
+        }
+    }
 }
```
ncreated, line 587: Beside all the test added to `RUMMonitorTests` (which ensure that the attributes are aggregated well in all commands 👌), we also need the behavioural test of attributes modification. After we clarify the product requirement for attributes processing, something like the following would be very handy IMO, WDYT?
```swift

// MARK: - Sending Attributes

func testSendingAttributes() throws {
    RUMFeature.instance = .mockByRecordingRUMEventMatchers(directory: temporaryDirectory)
    defer { RUMFeature.instance = nil }

    let view1 = createMockView(viewControllerClassName: "FirstViewController")
    let view2 = createMockView(viewControllerClassName: "SecondViewController")

    let monitor = RUMMonitor.initialize()

    // set global attributes:
    monitor.setAttribute(forKey: "attribute1", value: "value 1")
    monitor.setAttribute(forKey: "attribute2", value: "value 2")

    // start View 1:
    monitor.startView(viewController: view1)

    // update global attributes while the View 1 is active
    monitor.setAttribute(forKey: "attribute1", value: "changed value 1") // change the attribute value
    monitor.setAttribute(forKey: "attribute2", value: nil) // remove the attribute

    monitor.stopView(viewController: view1)

    // start View 2:
    monitor.startView(viewController: view2)
    monitor.stopView(viewController: view2)

    let rumEventMatchers = try RUMFeature.waitAndReturnRUMEventMatchers(count: 3)
    let firstViewEvent = try rumEventMatchers
        .lastRUMEvent(ofType: RUMView.self) { rumModel in rumModel.view.url == "FirstViewController" }

    XCTAssertEqual(try firstViewEvent.attribute(forKeyPath: "attribute1") as String, "changed value 1")
    XCTAssertEqual(try firstViewEvent.attribute(forKeyPath: "attribute2") as String, "value 2")

    let secondViewEvent = try rumEventMatchers
        .lastRUMEvent(ofType: RUMView.self) { rumModel in rumModel.view.url == "SecondViewController" }

    XCTAssertEqual(try secondViewEvent.attribute(forKeyPath: "attribute1") as String, "changed value 1")
    XCTAssertNil(try? secondViewEvent.attribute(forKeyPath: "attribute2") as String)
}
```

ncreated (changes requested):
Few comments added 👌. Also, before we merge it, let's discuss with the team the precise product requirements for global attributes processing.

Sources/Datadog/DDRUMMonitor.swift
```diff
@@ -11,7 +11,18 @@ import UIKit
...
+    // MARK: - Public properties
+
+    /// Global attributes to be injected into every command
+    public var attributes: [AttributeKey: AttributeValue] { [:] }
+
+    /// Sets a global attribute to be injected into every command
+    /// - Parameters:
+    ///   - key: key of the global attribute. If key already exists, new value replaces old value
+    ///   - value: attribute value. If nil, key is removed from global attributes
+    public func setAttribute(forKey key: AttributeKey, value: AttributeValue?) { }
```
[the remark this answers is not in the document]
buranmert: `var attributes` is used in `RUMMonitor.aggregate` method, it also serves as thread-safe public getter of private property

i looked at Logger API but i didn't understand 
1. why we hide user's attributes from the user, why can't they get what they set?
2. why don't we have one single `set(attribute: Value?)` to add/update/remove attributes?

a similar request was made yesterday via Slack (i've just sent you the link)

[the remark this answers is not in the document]
ncreated:
```swift
public var attributes: [AttributeKey: AttributeValue]
```
is basically extending the public interface with the interface of `Dictionary`. If we decide to change from storing attributes as dictionary we will be put into trouble.

> why we hide user's attributes from the user, why can't they get what they set?

For the sake of keeping things simple. If people don't need it, we don't add it (as we don't have it for `Logger`).

> why don't we have one single set(attribute: Value?) to add/update/remove attributes?

Because it's less explicit and people need to guess that setting `nil` will remove the attribute. This is cleaner and sets the smaller API comments in the right places:
```swift
public func addAttribute(forKey key: AttributeKey, value: AttributeValue)
public func removeAttribute(forKey key: AttributeKey)
```
It's also how people used to work with `Datadog.Logger`.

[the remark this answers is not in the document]
buranmert: ℹ️  in short, okay i'm gonna copy `Logger` API to `RUM` because i need this feature for my main ticket asap ✅ 
i'm resolving the conversation as i'll make changes and the discussion seems to be around beliefs and simplicity

---

but i'd like to leave my opinion here for the records 🎗️ 

the new type must support `init(from: Dictionary)` as we receive attributes as `Dictionary` in public methods.
if this type doesn't support conversion back to a `Dictionary`, yes then we have a problem. if you genuinely believe that may happen, then it's fine 👌 
we can do smth like `var attributes: Collection` too btw.?

if a user needs an attribute value, they need to double their sources of truth; honestly, i don't think this is simple.
also, we received a similar request about getting set values just yesterday.

`set(value: nil, for: key)` has other examples in Swift, `Dictionary.subscript` is pretty similar and common too.
if i were a user dealing with an optional value, i would need to do
```swift
if let someValue = myOptionalValue {
  rum.addAttribute(forKey: myKey, value: someValue)
} else {
  rum.removeAttribute(forKey: myKey)
}
```
vs. `rum.set(myOptionalValue, for: myKey)`
either way works to be honest 👌  i just don't understand how we can go with the former while we are for simplicity 😕

Sources/Datadog/RUMMonitor.swift
```diff
@@ -56,6 +56,8 @@ public class RUMMonitor: DDRUMMonitor {
     internal let contextProvider: RUMCurrentContext
     /// Time provider.
     private let dateProvider: DateProvider
+    /// Attributes associated with every command.
+    private var RUMAttributes: [AttributeKey: AttributeValue] = [:]
```
ncreated, line 60: Convention: variable names should start with a lowercase
```suggestion
    private var rumAttributes: [AttributeKey: AttributeValue] = [:]
```
or just `attributes` as we are in `RUM` monitor.

Sources/Datadog/RUMMonitor.swift
```diff
@@ -245,15 +247,35 @@ public class RUMMonitor: DDRUMMonitor {
...
+    override public func removeAttribute(forKey key: AttributeKey) {
+        queue.async {
+            self.RUMAttributes[key] = nil
+        }
+    }
+
     // MARK: - Private
 
+    private func aggregate(_ localAttributes: [AttributeKey: AttributeValue]?) -> [AttributeKey: AttributeValue] {
+        var mergedAttributes = queue.sync { return self.RUMAttributes }
```
ncreated, line 274: Ref, I created `RUMM-691 Add RUMMonitor benchmarks` so we can test if there's no performance drop due to this extra thread switch for every public API call.

ncreated (approved):
Just fix the variable name thing, then we can merge 👌

Sources/Datadog/RUMMonitor.swift
buranmert: i think the cost should be negligible but if we have concerns, we can do that:
```swift
protocol RUMCommand {
  var attributes { get set }
  ...
}
class RUMMonitor {
  func process(cmd) {
    queue.async {
      cmd.attributes.merge(globalAttrs)
      appScope.process(cmd)
    }
  }
```
then we won't need `queue.sync` to get global attrs
let me know if you need this

PR #233, opened 2020-08-31
RUMM-672 Add improvements noticed in RUM dogfooding

ncreated:
### What and why?

🐣 This PR fixes 3 first issues noticed when early dogfooding the RUM feature in Datadog mobile app.

While installing the RUM feature in mobile app I noticed 3 main problems:
* It wasn't possible to infer a meaningful name for the RUM Views as our Datadog mobile app heavily uses generic view controllers.
* It wasn't possible to match each `startView()` call with the corresponding `stopView()` for all view controllers because our Datadog mobile app uses `UISearchController` which follows `viewWillAppear()` / `didDisappear()` differently.
* Because of the 2nd, it wasn't easy to debug the current state of the `RUM` feature, mainly active / inactive RUM Views.

### How?

Each problem lead to a separate solution / improvement, respectively:

#### New `path: String?` parameter in `monitor.startView(...)` API

To fix the first issue, new `path: String?` parameter was added to `monitor.startView()` public API:
> - path: the View path used for RUM Explorer. If not provided, the `UIViewController` class name will be used.
```diff
- startView(viewController: UIViewController, attributes: [AttributeKey: AttributeValue]?)
+ startView(viewController: UIViewController, path: String?, attributes: [AttributeKey: AttributeValue]?)
```

#### Starting next View automatically stops the previous one

The second issue was fixed by unifying the iOS SDK behaviour with [the Android one](https://github.com/DataDog/dd-sdk-android/blob/2a9cd519fc2ac63671d64d1d43c1b7803e4540eb/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/internal/domain/scope/RumViewScope.kt#L123-L127). Whenever a new RUM View is started, the previous one gets closed (or becomes inactive until all its Resources complete). Although this is an optimisation done only for `UIKit`, I decided to make it default (instead of introducing `UIKit`-specific component) to match both SDKs behaviours.

With this improvement, some of the app scenarios (like the one with `UISearchController`) can be instrumented by only using the `rum.startView()` API.

#### New `Datadog.debugRUM` utility

To make it easy for the user to inspect the current RUM state, I introduced a new public API:
```swift
/// Utility setting to inspect the active RUM View. It only makes effect on the iOS Simulator.
/// If set, a debugging outline will be diesplayed on top of the application, describing the name of the active RUM View.
/// May be used to debug issues with RUM instrumentation in your app.
/// Default is `false`.
public static var debugRUM: Bool = false
```

It helps the user by rendering the RUM Views information in iOS Simulator. This makes it easy to spot the current or stale RUM Views (i.e. when their Resources are never completed due to wrong instrumentation):

Simulator Screen Shot - iPhone 11 - 2020-08-28 at 19 38 01

### Review checklist

### DIFF
--- a/Sources/Datadog/Datadog.swift
+++ b/Sources/Datadog/Datadog.swift
@@ -61,6 +61,16 @@ public class Datadog {
     /// Default is `nil`.
     public static var verbosityLevel: LogLevel? = nil
 
+    /// Utility setting to inspect the active RUM View. It only makes effect on the iOS Simulator.
+    /// If set, a debugging outline will be diesplayed on top of the application, describing the name of the active RUM View.
+    /// May be used to debug issues with RUM instrumentation in your app.
+    /// Default is `false`.
+    public static var debugRUM: Bool = false {
+        didSet {
+            (Global.rum as? RUMMonitor)?.enableRUMDebugging(debugRUM)
+        }
+    }
+
     public static func setUserInfo(
         id: String? = nil,
         name: String? = nil,
--- a/Sources/Datadog/Datadog.swift
+++ b/Sources/Datadog/Datadog.swift
@@ -62,7 +62,7 @@ public class Datadog {
     public static var verbosityLevel: LogLevel? = nil
 
     /// Utility setting to inspect the active RUM View. It only makes effect on the iOS Simulator.
-    /// If set, a debugging outline will be diesplayed on top of the application, describing the name of the active RUM View.
+    /// If set, a debugging outline will be displayed on top of the application, describing the name of the active RUM View.
     /// May be used to debug issues with RUM instrumentation in your app.
     /// Default is `false`.
     public static var debugRUM: Bool = false {
--- a/Sources/Datadog/Datadog.swift
+++ b/Sources/Datadog/Datadog.swift
@@ -61,7 +61,7 @@ public class Datadog {
     /// Default is `nil`.
     public static var verbosityLevel: LogLevel? = nil
 
-    /// Utility setting to inspect the active RUM View. It only makes effect on the iOS Simulator.
+    /// Utility setting to inspect the active RUM View.
     /// If set, a debugging outline will be displayed on top of the application, describing the name of the active RUM View.
     /// May be used to debug issues with RUM instrumentation in your app.
     /// Default is `false`.
--- a/Sources/Datadog/RUM/Debugging/RUMDebugging.swift
+++ b/Sources/Datadog/RUM/Debugging/RUMDebugging.swift
@@ -109,7 +109,7 @@ internal class RUMDebuggingInSimulator: RUMDebugging {
     }
 }
 
-private class RUMViewOutline: RUMDebugView {
+internal class RUMViewOutline: RUMDebugView {
     private struct Constants {
         static let activeViewColor =  #colorLiteral(red: 0.3882352941, green: 0.1725490196, blue: 0.6509803922, alpha: 1)
         static let inactiveViewColor =  #colorLiteral(red: 0.6000000238, green: 0.6000000238, blue: 0.6000000238, alpha: 1)
@@ -128,7 +128,7 @@ private class RUMViewOutline: RUMDebugView {
     private let label: UILabel
     private let stackOffset: CGFloat
 
-    init(viewInfo: RUMDebugInfo.View, stack: (index: Int, total: Int)) {
+    fileprivate init(viewInfo: RUMDebugInfo.View, stack: (index: Int, total: Int)) {
         self.label = UILabel(frame: .zero)
         self.stackOffset = CGFloat(stack.index) * Constants.labelHeight
 
@@ -170,7 +170,7 @@ private class RUMViewOutline: RUMDebugView {
     }
 }
 
-private class RUMDebugView: UIView {
+internal class RUMDebugView: UIView {
     override init(frame: CGRect) {
         super.init(frame: frame)
         self.backgroundColor = .clear
--- a/Sources/Datadog/RUM/Debugging/RUMDebugging.swift
+++ b/Sources/Datadog/RUM/Debugging/RUMDebugging.swift
@@ -12,7 +12,7 @@ internal protocol RUMDebugging {
 import UIKit
 import Foundation
 
-internal struct RUMDebugInfo {
+private struct RUMDebugInfo {
     struct View {
         let uri: String
         let isActive: Bool
@@ -35,16 +35,41 @@ internal class RUMDebuggingInSimulator: RUMDebugging {
     private lazy var canvas: UIView = {
         let window = UIApplication.shared.keyWindow
         let view = RUMDebugView(frame: window?.bounds ?? .zero)
+        view.autoresizingMask = [.flexibleWidth, .flexibleHeight]
         window?.addSubview(view)
         return view
     }()
 
+    // MARK: - Initialization
+
+    init() {
+        UIDevice.current.beginGeneratingDeviceOrientationNotifications()
+
+        NotificationCenter.default
+            .addObserver(
+                self,
+                selector: #selector(RUMDebuggingInSimulator.updateLayout),
+                name: UIDevice.orientationDidChangeNotification,
+                object: nil
+        )
+    }
+
     deinit {
+        UIDevice.current.endGeneratingDeviceOrientationNotifications()
+
+        NotificationCenter.default.removeObserver(
+            self,
+            name: UIDevice.orientationDidChangeNotification,
+            object: nil
+        )
+
         let canvas = self.canvas
         DispatchQueue.main.async {
             canvas.removeFromSuperview()
         }
     }
+
+    // MARK: - Internal
 
     func debug(applicationScope: RUMApplicationScope) {
         // `RUMDebugInfo` must be created on the caller thread.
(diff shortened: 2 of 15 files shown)


Sources/Datadog/RUM/Debugging/RUMDebugging.swift
```diff
@@ -0,0 +1,185 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+internal protocol RUMDebugging {
+    func debug(applicationScope: RUMApplicationScope)
+}
+
+#if targetEnvironment(simulator)
```
buranmert, line 11: ❔  you probably considered `#if DEBUG` too, why not `#DEBUG`?

Sources/Datadog/RUM/Debugging/RUMDebugging.swift
```diff
@@ -0,0 +1,185 @@
...
+    // MARK: - Initialization
+
+    init() {
+        UIDevice.current.beginGeneratingDeviceOrientationNotifications()
+
+        NotificationCenter.default
+            .addObserver(
+                self,
+                selector: #selector(RUMDebuggingInSimulator.updateLayout),
+                name: UIDevice.orientationDidChangeNotification,
```
buranmert, line 52: ❔  is this an `UIKit` bug that this event not triggerring `layoutSubviews` calls?

buranmert (changes requested):
debug utility is pretty cool 💯 
i left a few questions below ❓

Sources/Datadog/RUM/Debugging/RUMDebugging.swift
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    // MARK: - Private
+
+    private func renderOnMainThread(rumDebugInfo: RUMDebugInfo) {
+        canvas.subviews.forEach { view in
+            view.removeFromSuperview()
+        }
+
+        let viewOutlines: [RUMViewOutline] = zip(rumDebugInfo.views, 0..<rumDebugInfo.views.count)
```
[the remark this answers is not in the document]
ncreated: Although in many cases it's equal, using `enumerated()` for index enumeration [is wrong](https://khanlou.com/2017/03/you-probably-don%27t-want-enumerated/). `zip()` is always correct.

Sources/Datadog/RUM/RUMMonitor/RUMCommand.swift
```diff
@@ -23,10 +24,28 @@ internal struct RUMStartViewCommand: RUMCommand {
...
+        self.identity = identity
+        self.path = path ?? RUMStartViewCommand.viewPath(from: identity)
+    }
+
+    private static func viewPath(from id: AnyObject) -> String {
+        guard let viewController = id as? UIViewController else {
+            return ""
+        }
+
+        return "\(type(of: viewController))"
```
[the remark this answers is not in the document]
ncreated: Today we provide API for `UIKit`, but it will be also `SwiftUI` tomorrow. I consider `UIViewController` in public API + `Any` in internal as a good compromise to avoid huge refactoring when we start working with `SwiftUI`. We just cast the thing at the very top and at the very bottom - everywhere internally we work with `Any`.

Sources/Datadog/Datadog.swift
```diff
@@ -61,6 +61,16 @@ public class Datadog {
     /// Default is `nil`.
     public static var verbosityLevel: LogLevel? = nil
 
+    /// Utility setting to inspect the active RUM View. It only makes effect on the iOS Simulator.
+    /// If set, a debugging outline will be diesplayed on top of the application, describing the name of the active RUM View.
+    /// May be used to debug issues with RUM instrumentation in your app.
+    /// Default is `false`.
+    public static var debugRUM: Bool = false {
```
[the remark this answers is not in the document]
ncreated: This gives users more flexibility as they may want to enable/disable debugging at runtime (as we do in Datadog mobile app with a toggle on Debug Assistant screen). Also, for some apps installing `debugRUM = true` in `AppDelegate` may be too early, as the `UIAppDelegate.shared.keyWindow` is not yet ready. Thus, with runtime switch people can overcome this.

Also, this is similar utility to `Datadog.verbosityLevel`, so having `Datadog.debugRUM` next to it unifies the API usage.

Sources/Datadog/RUM/Debugging/RUMDebugging.swift
ncreated: Some apps may provide different setup in `RELEASE` configuration or even use custom `DEBUG` configuration name, so my consideration was: `#if targetEnvironment(simulator)` is more flexible.

Sources/Datadog/RUM/Debugging/RUMDebugging.swift
ncreated: I think it gets called if the `UIView` is managed by the view controller. Here we just add the view to the `UIWindow`.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -92,18 +95,23 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
         switch command {
         // View commands
         case let command as RUMStartViewCommand where command.identity === identity:
+            if didReceiveStartCommand {
+                // This is the case of duplicated "start" command. We know that the Session scope has created another instance of
+                // the `RUMViewScope` for tracking this View, so we mark this one as inactive.
+                isActiveView = false
```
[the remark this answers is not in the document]
ncreated: Not sure I understand. "Duplicated views case" is actually something we want. If one View is waiting for its Resources to complete, but another one is started, we have two running `RUMViewScope`, which is totally fine. We just mark the prior as `isActiveView == false` so it doesn't capture new Resources anymore.

Sources/Datadog/RUM/RUMMonitor/RUMCommand.swift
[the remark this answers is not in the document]
buranmert: yes, maybe i wasn't clear but that's what i'm saying: all the internals can be `Any`, why do we need to to cast `id` to `UIViewController` here?

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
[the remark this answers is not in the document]
buranmert: the case in my mind was calling `startView` twice for some reason (eg: once in `viewDidLoad` once in `viewWillAppear`) then we will have duplicate views, is this desired?

Sources/Datadog/RUM/Debugging/RUMDebugging.swift
buranmert: in order for us to be on the same page: 
we should allow debugging in non-production builds and disallow in production builds, no?

btw you can use `#if DEBUG` with different config names (as long as you don't remove default `DEBUG` active compilation condition)

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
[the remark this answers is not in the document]
ncreated: Yes, this would mean bad instrumentation. 

Please note, that it's also fine to receive:
```swift
rum.startView(vc1)
rum.stopView(vc1)
rum.startView(vc1)
rum.stopView(vc1)
rum.startView(vc2)
rum.stopView(vc2)
```
and result with 3 running `RUMViewScopes` at a time:
Simulator Screen Shot - iPhone 11 - 2020-08-28 at 19 38 01
if both instances of `vc1` are still waiting for Resources.

If we'd like the `RUMSessionScope` to mark and manage the "active" View, it would have to know how many Resources are pending on each instance of `RUMViewScope` and close that scope if it goes down to `0`. This would mean that `viewScope.pendingResourcesCount` would have to be accessed by `RUMSessionScope` and that the result of `viewScope.process(command:)` should be ignored by the Session scope. Both things violate the unidirectional architecture of scopes.

Sources/Datadog/RUM/RUMMonitor/RUMCommand.swift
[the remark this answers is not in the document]
ncreated: I see 👌. Yes, I removed the `UIViewController` casting here along with `import UIKit` - resulting with less `UIKit` coupling in RUM internals 👍.

Sources/Datadog/RUM/Debugging/RUMDebugging.swift
ncreated: Actually, after talking with @jegnux, we concluded that for our mobile app it is valid to use `debugRUM = true` even in non `DEBUG` builds, so it can be toggled in Test Flight or even App Store builds with Debug Assistant view. As this may be also the use case for other apps, WDYT about removing `#if` at all, @buranmert? Then, this would be up to the user to ensure it's not enabled in production builds when not necessary. Less safety / more control.

Sources/Datadog/RUM/Debugging/RUMDebugging.swift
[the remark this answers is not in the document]
jegnux: But this way it's still wrong. Try `zip(views, views.indices)` instead ;)

Sources/Datadog/Datadog.swift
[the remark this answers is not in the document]
buranmert: ℹ️ this discussion and the one below are converging, i'm resolving this one

Sources/Datadog/RUM/Debugging/RUMDebugging.swift
buranmert: yes, then non-devs can test the app with this 👍 

but i'm still hesitating to add it to public api? we can still provide it in a semi-public way with `NotificationCenter.post` for example?
do we want to support this feature in tvOS, SwiftUI, React Native, etc.?

ncreated: TBH I don't understand the idea with `NotificationCenter.post`, how would it work? To me those are clear enough:
```swift
/// One debug utility:
Datadog.verbosityLevel

/// Another debug utility:
Datadog.debugRUM
```

buranmert:
```swift
NotificationCenter.default.post(name: .init("datadog.enableRUMDebugging"), object: true/false)
```
this is a raw idea, we can mention this notification name in docs or readme (assuming if they are debugging the SDK then they probably look at docs and/or readme first)
the idea is if we decide to drop it from public API, we can do it silently without breaking their builds
wdyt?

btw i found out that `verbosityLevel` was a debug utility thanks to a GH issue, it may not be clear to everyone

ncreated: I don't see how the notification may be discoverable and TBH I never saw the API like that. Where we would put it in the documentation?

> btw i found out that verbosityLevel was a debug utility thanks to a GH issue, it may not be clear to everyone

The code documentation states it clear that it's for debugging purposes.
```swift
/// Verbosity level of Datadog SDK. Can be used for debugging purposes.
/// If set, internal events occuring inside SDK will be printed to debugger console if their level is equal or greater than `verbosityLevel`.
/// Default is `nil`.
public static var verbosityLevel: LogLevel? = nil
```

ncreated: As concluded on Slack - let's go with `Datadog.debugRUM` and we may consider adding `ProcessInfo`-based configuration later.

buranmert (approved):

PR #234, opened 2020-09-01
RUMM-682 Resources are associated with actions at init

buranmert:
### What and why?

Expected behavior is resource being associated with the action which caused/triggered the resource
Resource scope was getting active user action from the current context at `sendEvent` phase
This resulted in unexpected behavior

### How?

Now resource scope keeps the active user action info at its `init`
If active user action changes during the lifetime of the resource, it doesn't change resource event's action

### Review checklist

### DIFF
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
@@ -6,10 +6,10 @@
 
 import Foundation
 
-internal class RUMResourceScope: RUMScope, RUMContextProvider {
+internal class RUMResourceScope: RUMScope {
     // MARK: - Initialization
 
-    private unowned let parent: RUMContextProvider
+    let context: RUMContext
     private let dependencies: RUMScopeDependencies
 
     /// The name used to identify this Resource.
@@ -25,7 +25,7 @@ internal class RUMResourceScope: RUMScope, RUMContextProvider {
     private var resourceHTTPMethod: RUMHTTPMethod
 
     init(
-        parent: RUMContextProvider,
+        context: RUMContext,
         dependencies: RUMScopeDependencies,
         resourceName: String,
         attributes: [AttributeKey: AttributeValue],
@@ -33,19 +33,13 @@ internal class RUMResourceScope: RUMScope, RUMContextProvider {
         url: String,
         httpMethod: RUMHTTPMethod
     ) {
-        self.parent = parent
+        self.context = context
         self.dependencies = dependencies
         self.resourceName = resourceName
         self.attributes = attributes
         self.resourceURL = url
         self.resourceLoadingStartTime = startTime
         self.resourceHTTPMethod = httpMethod
-    }
-
-    // MARK: - RUMContextProvider
-
-    var context: RUMContext {
-        return parent.context
     }
 
     // MARK: - RUMScope
--- a/Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScopeTests.swift
+++ b/Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScopeTests.swift
@@ -10,19 +10,17 @@ import XCTest
 class RUMResourceScopeTests: XCTestCase {
     private let output = RUMEventOutputMock()
     private lazy var dependencies: RUMScopeDependencies = .mockWith(eventOutput: output)
-    private let parent = RUMContextProviderMock(
-        context: .mockWith(
-            rumApplicationID: "rum-123",
-            sessionID: .mockRandom(),
-            activeViewID: .mockRandom(),
-            activeViewURI: "FooViewController",
-            activeUserActionID: .mockRandom()
-        )
+    private let context = RUMContext.mockWith(
+        rumApplicationID: "rum-123",
+        sessionID: .mockRandom(),
+        activeViewID: .mockRandom(),
+        activeViewURI: "FooViewController",
+        activeUserActionID: .mockRandom()
     )
 
     func testDefaultContext() {
         let scope = RUMResourceScope(
-            parent: parent,
+            context: context,
             dependencies: .mockAny(),
             resourceName: .mockAny(),
             attributes: [:],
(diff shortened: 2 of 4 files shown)


Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
```diff
@@ -23,6 +23,8 @@ internal class RUMResourceScope: RUMScope, RUMContextProvider {
     private var resourceLoadingStartTime: Date
     /// The HTTP method used to load this Resource.
     private var resourceHTTPMethod: RUMHTTPMethod
+    /// ID of the action which triggered this Resource.
+    private let triggeringActionID: RUMUUID?
```
ncreated, line 27: The `activeUserActionID` is part of the `context`. Instead of duplicating this information, we could basically capture the context once at scope initialization time:
```diff
+ let context: RUMContext 
- var context: RUMContext {
-     return parent.context
- }
``` 
as [we fixed this bug in Android SDK](https://github.com/DataDog/dd-sdk-android/commit/b88b5b50062b59d43be437501844ae29ecef42ab#).

By introducing separate `triggeringActionID` value the `self.context` becomes inconsistent which is error-prone.

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScopeTests.swift
```diff
@@ -51,6 +51,9 @@ class RUMResourceScopeTests: XCTestCase {
         )
 
         currentTime.addTimeInterval(2)
+        let expectedActionID = parent.context.activeUserActionID
+        parent.context.activeUserActionID = .mockRandom()
+        parent.context.activeViewID = .mockRandom()
```
ncreated, line 56: It's hard to understand as this test will pass normally without this code. It only replaces the action ID so we prove that this has no impact on the `model.action.id`. IMO this should be a separate test case explaining what we want to test using clear given / when / then.

ncreated (changes requested):

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
buranmert: 🤔 but this is not duplicate info: the ticket says _"Resources are linked with the wrong Action"_, what you suggest also changes which view is linked with the resource because active view is accessed via current context 😕  
either the ticket is wrong or Android's fix is wrong.?

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScopeTests.swift
buranmert: the description of `ResourceScope` seems to be
1. keep the initial active action
2. get the final active view
this case verifies the description of component

i can remove if it's confusing.?

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
ncreated: By design, when initializing the `RUMResourceScope` the `context` defines the exact state of the RUM product during Resource start, including active View ID, active Action ID, Session ID and Application ID. None of those should change for the Resource - once we capture the context, it will always describe the right state.

CC @xgouchet ?

buranmert: then ticket is wrong 👌 
i'm changing the ticket first, then the code

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScopeTests.swift
ncreated: As we add the new behaviour: _**given** the Resource started on User Action, **when** new Action is started, **it** tracks the action which caused it_, we should have the unit test for it.

I just mean that we test it in a wrong way: by modifying the existing test which asserts totally different behaviour. Removing this code will still make this test pass.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
xgouchet: Sorry if the ticket was misleading. Indeed the goal is to capture what the active context is when the resource is started, so that when it is sent, is has the relevant context information.
A resource is always linked to the context that was active when it started, so that even if another view/action is started before the resource stops, it'll be linked with the original view/action.

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScopeTests.swift
xgouchet: Well, the definition of ResourceScope should be that it should always be linked with the initial active context (both view and action). If another view is started, it should definitely not be linked with the resource.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
buranmert: so `ResourceScope` should not be a `ContextProvider` anymore as it doesn't always have the up-to-date context, correct @ncreated ?

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScopeTests.swift
buranmert: as `ResourceScope` takes `context` in `init` now, we don't need to test context mutation during life-cycle anymore: problem solved automatically ✅

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
ncreated: > so ResourceScope should not be a ContextProvider anymore as it doesn't always have the up-to-date context

It may still be the `ContextProvider` for architectural consistency. There's nothing invalid about its context (it's never "not up-to-date"). But as `RUMResourceScope` has no children we can remove `ContextProvider` conformance if yo prefer and add it once needed.

ncreated (approved):

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
buranmert: okay, i removed as it's unused ✅

PR #236, opened 2020-09-01
RUMM-675 Add `tracer.startRootSpan()` API

ncreated:
### What and why?

📦 This PR adds the `startRootSpan()` API to `Datadog.Tracer`. 

This API enables using `setActive()` span in asynchronous jobs, e.g.:

```swift
func makeAPIRequest(on queue: DispatchQueue, completion: @escaping () -> Void) {
    let requestSpan = tracer
        .startRootSpan(operationName: "request")
        .setActive() // set active, so the asynchronous job can continue the trace without the span reference

    queue.asyncAfter(deadline: .now() + 1) {
        let responseDecodingSpan = tracer.startSpan(operationName: "response decoding")
        responseDecodingSpan.finish()
        requestSpan.finish()
        completion()
    }
}

makeAPIRequest(on: queue) {}
makeAPIRequest(on: queue) {}
```
which results with two root "request" spans, each having "response decoding" child span.

### How?

New API was introduced:
```swift
/// Start a new root span with the given operation name.
/// - Parameters:
///   - operationName: the operation name for the newly-started span
///   - tags:          a set of tag keys and values per `OTSpan#setTag:value:`, or `nil` to start with
///                    an empty tag map
///   - startTime:     an explicitly specified start timestamp for the `OTSpan`, or `nil` to use the
///                    current walltime
/// - returns:         a valid Span instance; it is the caller's responsibility to call `finish()`.
func startRootSpan(
    operationName: String,
    tags: [String: Encodable]?,
    startTime: Date?
) -> OTSpan
```

### Review checklist

### DIFF
--- a/Sources/Datadog/OpenTracing/OTTracer.swift
+++ b/Sources/Datadog/OpenTracing/OTTracer.swift
@@ -8,15 +8,31 @@ public protocol OTTracer {
     /// Start a new span with the given operation name.
     ///
     /// - parameter operationName: the operation name for the newly-started span
-    /// - parameter references:    an optional list of Reference instances to record causal relationships
-    /// - parameter tags:          a set of tag keys and values per OTSpan#setTag:value:, or nil to start with
+    /// - parameter references:    an optional list of Reference instances to record causal relationships. If no
+    ///                            reference is provided, and an active span exists in the current execution context
+    ///                            the active span will be used as the parent.
+    /// - parameter tags:          a set of tag keys and values per `OTSpan#setTag:value:`, or `nil` to start with
     ///                            an empty tag map
-    /// - parameter startTime:     an explicitly specified start timestamp for the OTSpan, or nil to use the
+    /// - parameter startTime:     an explicitly specified start timestamp for the `OTSpan`, or `nil` to use the
     ///                            current walltime
-    /// - returns:                 a valid Span instance; it is the caller's responsibility to call finish()
+    /// - returns:                 a valid Span instance; it is the caller's responsibility to call `finish()`.
     func startSpan(
         operationName: String,
         references: [OTReference]?,
+        tags: [String: Encodable]?,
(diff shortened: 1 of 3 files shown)


nachoBonafonte (approved):

buranmert (approved):
1. `setActive()` can be called both on normal and root spans, correct?
2. are we gonna add this to the `OpenTracing/Telemetry` API (the original repo) as well?

ncreated:
> * `setActive()` can be called both on normal and root spans, correct?

Yes. Regular span can become the active span in the current execution context. Same for the root span.

> * are we gonna add this to the `OpenTracing/Telemetry` API (the original repo) as well?

Not for OT, as there's [no activity in `opentracing-swift` for many months](https://github.com/opentracing/opentracing-swift) which is why we decided to put the OT code in our repo. Not sure if there's `startRootSpan()` concept in `OTel` - @nachoBonafonte ?

nachoBonafonte:
In `OTel`, there is a span builder, and it has a `setNoParent()` method, that basically makes a span a root span when is built

Sources/Datadog/OpenTracing/OTTracer.swift
```diff
@@ -87,4 +105,24 @@ public extension OTTracer {
...
+    func startRootSpan(
+        operationName: String,
+        tags: [String: Encodable]? = nil,
+        startTime: Date? = nil
+    ) -> OTSpan {
+        return self.startRootSpan(
+            operationName: operationName,
+            tags: tags,
+            startTime: startTime
+        )
```
buranmert, line 126: this code goes into infinite loop
do we need this extended method?

ncreated: As discussed, it goes infinite only in case of `NoOpTracer`. We will fix it 👍.

PR #237, opened 2020-09-03
RUMM-685: Fix - Span is made thread safe

buranmert:
### What and why?

`Span` used non-thread-safe methods for setting tags and adding logs

### How?

The issue was that setting a value via subscript is not thread-safe
We wrapped those set operations with `queue.a/sync { ... }` calls

### Review checklist


ncreated (approved):
Very clear 👌 👌 👌

PR #239, opened 2020-09-07
RUMM-679 Encode Span error attributes in RUM Error fields

ncreated:
### What and why?

📦 This PR updates the way we encode `Span` error attributes for RUM Errors in "Tracing with RUM" integration. 

This unifies the data of RUM Errors on Android and iOS when user marks the `Span` as error (either directly or through `span.log()` integration). Counterpart of https://github.com/DataDog/dd-sdk-android/pull/351

### How?

Before, the error attributes were encoded at the root level of RUM event JSON using 3 custom fields:
```diff
-        static let spanErrorMessage = "span.error_message"
-        static let spanErrorType = "span.error_type"
-        static let spanErrorStack = "span.error_stack"
```

From now, span's `errorMessage`, `errorType` and `errorStack` are encoded into `RUMError` object directly:
```swift
// RUMViewScope.swift

let eventData = RUMError(
	// ...
    error: .init(
        message: command.message,
        source: command.source.toRUMDataFormat,
        stack: command.stack,
		// ...
    ),
    // ...
)
```

### Review checklist

### DIFF
--- a/Sources/Datadog/FeaturesIntegration/TracingWithRUMIntegration.swift
+++ b/Sources/Datadog/FeaturesIntegration/TracingWithRUMIntegration.swift
@@ -24,41 +24,52 @@ internal struct TracingWithRUMContextIntegration {
 
 /// Sends given `Span` as RUM Errors.
 internal struct TracingWithRUMErrorsIntegration {
-    struct Attributes {
-        static let spanErrorMessage = "span.error_message"
-        static let spanErrorType = "span.error_type"
-        static let spanErrorStack = "span.error_stack"
-    }
-
     private let rumErrorsIntegration = RUMErrorsIntegration()
 
     func addError(for ddspan: DDSpan) {
-        let rumErrorMessage = "Span \"\(ddspan.operationName)\" reported an error"
-        let rumErrorAttributes = captureRUMErrorAttributesFromOTLogFields(in: ddspan)
+        let rumErrorAttributes = captureRUMErrorAttributes(from: ddspan)
 
-        rumErrorsIntegration.addError(with: rumErrorMessage, attributes: rumErrorAttributes)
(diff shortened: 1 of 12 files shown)


Sources/Datadog/FeaturesIntegration/TracingWithRUMIntegration.swift
```diff
@@ -24,41 +24,52 @@ internal struct TracingWithRUMContextIntegration {
...
+        let spanErrorType = spanTags[DDTags.errorType] as? String
+        let spanErrorStack = spanTags[DDTags.errorStack] as? String
 
-        if let spanErrorMessage = spanTags[DDTags.errorMessage] {
-            rumErrorAttributes[Attributes.spanErrorMessage] = spanErrorMessage
-        }
-        if let spanErrorType = spanTags[DDTags.errorType] {
-            rumErrorAttributes[Attributes.spanErrorType] = spanErrorType
+        switch (spanErrorMessage, spanErrorType) {
+        case (let message?, let type?):
```
buranmert, line 53: oh i didn't know question marks were required here, good to know 👍 
do you know why it doesn't work without question marks?

ncreated: It's a [convenience for matching `.some(let value)`](https://docs.swift.org/swift-book/ReferenceManual/Patterns.html#ID520) on the optional type. As Swift doc says, this can be also used in loops:
```swift
let arrayOfOptionalInts: [Int?] = [nil, 2, 3, nil, 5]
// Match only non-nil values.
for case let number? in arrayOfOptionalInts {
    print("Found a \(number)")
}
```

buranmert (approved):

PR #245, opened 2020-09-09
RUMM-657 Add Active Span attributes to Logs

ncreated:
### What and why?

📦 This PR adds Logging with Tracing integration. When there is an active span registered, its attributes are added to the produced `Log`.

This enables two-way integration between Logging and APM:
```swift
let span = Global.sharedTracer.startSpan("logs containing span").setActive()
logger.warn("warn message")
logger.error("error message")
logger.critical("critical message")
span.finish()
```

#### User can see traces which contain given log:

Screenshot 2020-09-09 at 15 20 21

#### User can see logs send during the trace:

Screenshot 2020-09-09 at 15 20 46

### How?

New public API was added to `Logger.Builder`:
```swift
/// Enables the logs integration with active span API from Tracing.
/// If enabled all the logs will be bundled with the `Global.sharedTracer.activeSpan` trace and
/// it will be possible to see all the logs sent during that specific trace.
/// - Parameter enabled: `true` by default
public func bundleWithTrace(_ enabled: Bool) -> Builder
```

it's a counterpart of Android's [`setBundleWithTraceEnabled()`](https://github.com/DataDog/dd-sdk-android/blob/6eb7afdd3f84aa138eec3e50024f8af558176dfd/dd-sdk-android/src/main/kotlin/com/datadog/android/log/Logger.kt#L257)

The `LoggingWithActiveSpanIntegration` was added in the same form as other features integration.

This new integration unrevealed the bug we had in printing warning messages on SDK misconfiguration:
* when RUM feature is disabled, the warnings from Logging With RUM integration should not be printed,
* when RUM feature is disabled, the warnings from Tracing With RUM integration should not be printed.

This is now fixed.

### Review checklist

### DIFF
--- a/Tests/DatadogTests/Datadog/TracerConfigurationTests.swift
+++ b/Tests/DatadogTests/Datadog/TracerConfigurationTests.swift
@@ -40,7 +40,7 @@ class TracerConfigurationTests: XCTestCase {
             configuration: .init()
         ).dd
 
-        XCTAssertNotNil(tracer.rumContextIntegration)
+        XCTAssertNil(tracer.rumContextIntegration)
 
         guard let spanBuilder = (tracer.spanOutput as? SpanFileOutput)?.spanBuilder else {
             XCTFail()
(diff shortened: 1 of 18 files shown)


Tests/DatadogTests/Datadog/TracerConfigurationTests.swift
```diff
@@ -69,6 +69,15 @@ class TracerConfigurationTests: XCTestCase {
...
     }
 
+    func testDefaultTracerWithRUMEnabled() {
+        RUMFeature.instance = .mockNoOp()
+        defer { RUMFeature.instance = nil }
+
+        let tracer = Tracer.initialize(configuration: .init()).dd
+
+        XCTAssertNotNil(tracer.rumContextIntegration)
+    }
```
ncreated, line 79: I forgot to assert `nil` integration if `TracerConfiguration(..., bundleWithRUM: false, ...)` is passed. Will do before merge.

buranmert (approved):

Tests/DatadogTests/Datadog/TracerConfigurationTests.swift
ncreated: Done https://github.com/DataDog/dd-sdk-ios/pull/245/commits/4c9ca9b538ee0ff4cac18023130a3eed85c6ac3b

PR #246, opened 2020-09-09
RUMM-700 Environment regex changed

buranmert:
### What and why?

There was a discrepancy between Android's environment validation rule and ours.
For instance, `shop.ist` was invalid for us whereas it was valid for Android and more importantly valid for RUM Explorer.

### How?

Now we use Android's regex, which allows "_./-" as non-alphanumerics.

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/FeaturesConfiguration.swift
+++ b/Sources/Datadog/Core/FeaturesConfiguration.swift
@@ -140,9 +140,12 @@ extension FeaturesConfiguration {
 }
 
 private func ifValid(environment: String) throws -> String {
-    let regex = #"^[a-zA-Z0-9_]+$"#
+    /// 1. cannot be more than 200 chars (including `env:` prefix)
+    /// 2. cannot end with `:`
+    /// 3. can contain letters, numbers and _:./-_ (other chars are converted to _ at backend)
+    let regex = #"^[a-zA-Z0-9_:./-]{0,195}[a-zA-Z0-9_./-]$"#
     if environment.range(of: regex, options: .regularExpression, range: nil, locale: nil) == nil {
-        throw ProgrammerError(description: "`environment` contains illegal characters (only alphanumerics and `_` are allowed)")
+        throw ProgrammerError(description: "`environment`: \(environment) contains illegal characters (only alphanumerics and `_` are allowed)")
     }
     return environment
 }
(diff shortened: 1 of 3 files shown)


Sources/Datadog/Core/FeaturesConfiguration.swift
```diff
@@ -140,9 +140,9 @@ extension FeaturesConfiguration {
 }
 
 private func ifValid(environment: String) throws -> String {
-    let regex = #"^[a-zA-Z0-9_]+$"#
+    let regex = #"^[a-zA-Z0-9_./-]+$"#
```
ncreated, line 143: [There are more rules](https://docs.datadoghq.com/getting_started/tagging/#defining-tags) to include in this regexp, i.e. _"must start with a letter"_, _"can be up to 200 characters long"_.

ncreated (changes requested):
The JIRA also says to update the `env` in Shopist project to match the one used in Android.

Sources/Datadog/Core/FeaturesConfiguration.swift
ncreated: Ref to Android regexp: https://github.com/DataDog/dd-sdk-android/pull/369#discussion_r486145119

Sources/Datadog/Core/FeaturesConfiguration.swift
```diff
@@ -140,9 +140,13 @@ extension FeaturesConfiguration {
 }
 
 private func ifValid(environment: String) throws -> String {
-    let regex = #"^[a-zA-Z0-9_]+$"#
+    /// 1. starts with a letter
```
ncreated, line 143: As @mariusc83 concluded in https://github.com/DataDog/dd-sdk-android/pull/369#discussion_r486145119 this first requirement doesn't apply here, as what we validate here is the value of the tag. It's name will be `env:` (added by the backend), so it will always start with the letter `e`. The regexp should be then:
```swift
[a-zA-Z0-9_:./-]{0,195}[a-zA-Z0-9_./-]
```

ncreated (changes requested):
- [ ] The JIRA also says to update the env in Shopist project to match the one used in Android.

ncreated (approved):

PR #247, opened 2020-09-10
RUMM-711 Prepare `RUMMonitor` for auto instrumentation

ncreated:
### What and why?

📦 Following the RFC, this PR prepares the `RUMMonitor` for auto instrumentation.

### How?

The `RUMMonitor` now conform to `RUMCommandSubscriber`. When initialised, it gets subscribed to the `RUMAutoInstrumentation.instance`, where it will be further (next JIRAs) passed to registered handlers. Which handlers should be registered is determined by the `FeaturesConfiguration.RUM.AutoInstrumentation`.

As auto instrumentation handlers will be producing its own `command.attributes`, I also updated the way we aggregate attributes in `RUMMonitor`. Now it happens directly in `process(command:)`, not in manual instrumentation calls. This way we also gain one less thread switch for each `RUMMonitor` method call.

### Review checklist


Sources/Datadog/RUMMonitor.swift
```diff
@@ -137,7 +139,7 @@ public class RUMMonitor: DDRUMMonitor {
                 time: dateProvider.currentDate(),
                 identity: viewController,
                 path: path,
-                attributes: aggregate(attributes)
+                attributes: attributes ?? [:]
```
buranmert, line 142: we can set default value to `[:]` instead of `nil`, then we won't need to do that in these places

buranmert (approved):

Sources/Datadog/RUMMonitor.swift
ncreated: Yup, done 👍.

PR #248, opened 2020-09-10
CIAPP-25 Add initial test instrumentation to Swift tracer

nachoBonafonte:
### What and why?

This PR add automatic test instrumentation to the tests run by any target that links with `DatadogSDK`.
Also the target must have set the following environment variables set in XCode Test action of the desired schemes:

`DD_TEST_RUNNER`, if this env variable doesnt exist no  test instrumentation will be done
`DATADOG_CLIENT_TOKEN` with the client token to initialize the Datadog instance

If running as part of a CI workflow several other env variables must be set, but are very CI dependant, documentation will need to be provided. Currently can be found in `CIEnvironmentValues.swift`file

### How?

Added private module `_Datadog_TestRunner` where all files for test instrumentation are located:
`DDTestRunner` is a test observer where most of the logic is done, initialises its own tracer and creates spans for each test
`DDTestingTags` has all the specific tags that test span can use
`CIEnvironmentValues` has the logic for CI environment variables parsing and setting the tags
`FrameworkLoadHandler` and objC files use objC class load method to run code on library loading before any other code has run, or app has started

Added some changes to the SDK also needed for it:
-`Span.type` can now be set at creation instead of using always `"custom"`
-Added a wildcard `"."` to `URLFilter` so all network requests can be intercepted

Modified podspec to include all this new code
Also added unit tests to Example app configured to use this instrumentation

### Review checklist


Sources/Datadog/Tracing/Utils/URLFilter.swift
```diff
@@ -40,6 +40,9 @@ internal struct URLFilter: URLFiltering, Equatable {
     /// matches hosts and their subdomains: example.com -> example.com, api.example.com, sub.example.com, etc.
     private static func buildRegexString(from hosts: Set<String>) -> String {
         return hosts.map {
+            if $0 == "." {
+                return "."
+            }
```
ncreated, line 45: We can't allow wildcard to be passed from the public API. Our approach for requests auto instrumentation, is to receive an exact list of hosts from the user, without generalisations, so we must find an other way around for tests instrumentation.

Sources/_Datadog_TestRunner/DDTestRunner.swift
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        Datadog.initialize(
+            appContext: .init(),
+            configuration: Datadog.Configuration
+                .builderUsing(clientToken: clientToken, environment: CIEnvironmentValues.getEnvVariable("DD_ENV") ?? "test")
+                .set(serviceName: CIEnvironmentValues.getEnvVariable("DD_SERVICE") ?? ProcessInfo.processInfo.processName)
+                .set(tracedHosts: ["."])
+                .build()
+        )
```
ncreated, line 43: Why exactly do we need to initialise the entire SDK here? Do we need all its features (Logging, Tracing and RUM) or maybe just the Tracing?

All the features inside the SDK are **fully isolated**. What happens in result of `Datadog.initialize()` is that we instantiate/or not each feature respectively (according to the configuration):
```swift
LoggingFeature.instance = LoggingFeature(...)
TracingFeature.instance = TracingFeature(...)
RUMFeature.instance = RUMFeature(...)
```

Maybe we don't need to call the entire `Datadog.initialize()` but just create the instance of `TracingFeature` to make the instrumentation `Tracer` work? If possible, that would be IMO recommended way, because `Datadog.initialize()` called here may bring some unwanted consequences (today or in the feature).

Sources/_Datadog_TestRunner/DDTestRunner.swift
```diff
@@ -0,0 +1,139 @@
...
+
+        Global.sharedTracer = Tracer.initialize(
```
ncreated, line 45: By initialising the `Global.sharedTracer` in tests instrumentation, we allow the target-under-test to overwrite it. I'm wondering if it's really needed for us to use the global tracer in tests instrumentation.

What is the "link" between the tests instrumentation and the target-under-test (TUD)? If it's only about starting the active span in DD Tests Runner (DDTR) and continuing it in TUD, maybe we only need to build a bridge between DDTR and TUD so that the first can inject the active span reference to the latter?

Sources/_Datadog_TestRunner/DDTestRunner.swift
nachoBonafonte: The idea is collecting all possible information for the system under test for the duration of a test, being it a unit or an integration test. If the user has added logs or traces or some things are autoinstrumented, collecting them as part of the test would be very useful for understanding why the test has failed or why it is working slower. With this information, it is also very useful to write and run a test just to understand what the system is doing under the hood as a developer tool.  I have some doubts with RUM, don't still know if it could add value to a test, not unit test probably but maybe to integration tests.

Sources/_Datadog_TestRunner/DDTestRunner.swift
nachoBonafonte: The problem here, is that by definition in Open Tracing (and also OTel) there should only exist one tracer, we could avoid assigning the second tracer to shared if one already exists. Having two traces we would have also two different traces being sent and the injected headers would differ depending where the request is made. Also synchronizing two span generators for upload to backend could present some challenges.

ncreated: I see. I'm only worried if this is a scalable approach. Even now, the user may configure the state of his/her tracer (i.e. by customizing the `sendNetworkInfo` or passing `globalTags`):
```swift
Global.sharedTracer = Tracer.initialize(
    configuration: Tracer.Configuration(
        sendNetworkInfo: true,
        globalTags: ["foo": "bar"]
    )
)
```
If overwriting the `Global.sharedTracer` by our own, we will lost the user configuration. Do you see any nice solution for this problem?

> Having two traces we would have also two different traces being sent and the injected headers would differ depending where the request is made. Also synchronizing two span generators for upload to backend could present some challenges.

Instantiating a separate `Tracer`:
```swift
internal let testsTracer = Tracer(...)
```
doesn't instantiate separate serializer/uploader. The `storage` (serializer) and `upload` (uploader) components are managed by the `TracingFeature.instance`, which can be shared between multiple tracers (just like logging storage and upload can be shared among several loggers). The only challenge would be to pass the `testsTracer.activeSpan.context` value to the `Global.sharedTracer` when starting spans in the app under test.

Sources/_Datadog_TestRunner/DDTestRunner.swift
ncreated: I see 👌. I'm only wondering, what can we do to avoid "double" initialization of the SDK? Let's say the customer app calls `Datadog.initialize()` in his `AppDelegate`, which is also invoked in unit tests / UI Tests. Also, what if the user initialization leads to specific SDK state, i.e. `.enableTracing(false)`?

nachoBonafonte: We should not let the SDK to be initialised twice, we should fail if already initialised in tests. Also the configuration for the tests is probaby not the same that the user may desire for production, we cannot let the user disable tracing while testing for example (it should disable al the test instrumentation instead). We can start by doing all configuration to the test tracer by env variables, and see in the future how can the user modify it in code if necessary.

Sources/_Datadog_TestRunner/DDTestRunner.swift
nachoBonafonte: As I said in previous comment, we don't want to have the production tracer running on tests, and probably the user wont either, because receiving traces and logs from tests in their production endpoints probably will pollute their output. If they are already using datadog to trace their tests in an ad hoc setting, they are probably not interested in our solution or should need some transition in their code. 
About the configuration for the test tracer,  the idea for first version is all configuration being done by env variables, and evaluate in the future other ways

Sources/_Datadog_TestRunner/DDTestRunner.swift
ncreated: Sounds good 👍. We're currently failing (with a console warning) on doubled initialisation, but we could do it silently in tests 👌.

Sources/_Datadog_TestRunner/DDTestRunner.swift
ncreated: All clear then 👌 - we'll handle doubled initialisation silently in unit tests.

Tests/DatadogTests/DatadogTestRunner/FrameworkLoadHandlerTests.swift
```diff
@@ -0,0 +1,69 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import XCTest
+@testable import Datadog
+
+class FrameworkLoadHandlerTests: XCTestCase {
+    var testEnvironment = [String: String]()
```
ncreated, line 11:
```suggestion
    private var testEnvironment = [String: String]()
```

Sources/_Datadog_TestRunner/DDTestRunner.swift
```diff
@@ -0,0 +1,138 @@
...
+        /// We need to wait for all the traces to be written to the backend before exiting
+
+        tracer.queue.sync {} // waits until data is passed to writer
+        if let writer = TracingFeature.instance?.storage.writer as? FileWriter {
+            writer.queue.sync {} // waits until data is written
+        }
+        /// <--- here the file will be existing
+
+        do {
+            let directory = try Directory(withSubdirectoryPath: "com.datadoghq.traces/v1")
```
ncreated, line 69: The `"com.datadoghq.traces/v1"` path is shared with `TracingFeature` and it's important to keep it in sync. How about decoupling it to a `TracingFeature.dataDirectoryPath` constant?

Sources/_Datadog_TestRunner/DDTestRunner.swift
```diff
@@ -0,0 +1,138 @@
...
+
+        activeTestSpan = tracer.startSpan(spanContext: tracer.createSpanContext(), operationName: "XCTest.test", spanType: DDTestingTags.typeTest, tags: tags)
+        activeTestSpan?.setActive()
+
+        guard let testSpan = activeTestSpan else {
+            return
+        }
+
+        let envValues = CIEnvironmentValues()
+        envValues.addTagsToSpan(span: testSpan)
```
ncreated, line 107: nit: would be shorter if we store it on local variable:
```suggestion
let testSpan = tracer.startSpan(spanContext: tracer.createSpanContext(), operationName: "XCTest.test", spanType: DDTestingTags.typeTest, tags: tags)
testSpan.setActive()
envValues.addTagsToSpan(span: testSpan)
activeTestSpan = testSpan
```

Sources/_Datadog_TestRunner/DDTestRunner.swift
```diff
@@ -0,0 +1,138 @@
...
+        startTracer()
+    }
+
+    func startTracer() {
+        var clientToken = ""
+        if let envToken = ProcessInfo.processInfo.environment["DATADOG_CLIENT_TOKEN"] {
+            clientToken = envToken.trimmingCharacters(in: CharacterSet.whitespacesAndNewlines)
+        } else if let bundleToken = Bundle.main.infoDictionary?["DatadogClientToken"] as? String {
+            clientToken = bundleToken.trimmingCharacters(in: CharacterSet.whitespacesAndNewlines)
+        }
```
ncreated, line 30: Some `ProcessInfo` ENVs are abstracted by the `CIEnvironmentValues` and some are not. How about moving all ENVs consideration to the `CIEnvironmentValues` and just keeping the `env` on a single class property:
```swift
internal class DDTestRunner: NSObject, XCTestObservation {
   private let env = CIEnvironmentValues()

   // ...
   // env.datadogClientToken, ...
}
```

This way we could also test `DATADOG_CLIENT_TOKEN` and `DatadogClientToken` retrieval in `CIEnvironmentValuesTests`. Right now it's not tested.

Datadog/Example/ExampleTests/ExampleTests.swift
```diff
@@ -0,0 +1,40 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import XCTest
+@testable import Example
+
+
+class ExampleTests: XCTestCase {
```
ncreated, line 11: I'm not sure if having tests target in `Example` app is the best choice. As this app is used by the SDK unit tests (as a host), integration tests (as a subject) and benchmarks (host), we may need to modify it in this way or another either for the SDK or Test Runner. This may not play well in the future.

How about creating separate `TestRunner` app with `TestRunnerTests`? If we do so, would it be additionally possible to insturment it with the `HTTPServerMock` (Python server) to add integration tests for the test runner? Right now the `DDTestRunner` has no tests coverage, but it runs pretty important business logic. I think it would be a good subject for the integration testing.

ncreated (changes requested):
Really nice work 👌. I left few comments and some questions.

Sources/_Datadog_TestRunner/objc/ObjcLoadHandler.m
```diff
@@ -0,0 +1,16 @@
...
+
+#import <Foundation/Foundation.h>
+#import "ObjcLoadHandler.h"
+#if SWIFT_PACKAGE
+#else
+#import <Datadog/Datadog-Swift.h>
+#endif
+
+@implementation ObjcLoadHandler : NSObject
++ (void)load { [FrameworkLoadHandler handleLoad]; }
```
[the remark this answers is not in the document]
xgouchet: I'd definitely go with the second option indeed. We should not ship code in production whose sole purpose is for running tests.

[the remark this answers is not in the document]
mariusc83: I'd also go with the second solution and I would actually add an extra option there to try to find a way to not include this code in the compiled production application (only for the test app)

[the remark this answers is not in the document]
nachoBonafonte: I think the "zero code" instrumentation is a product requirement, but lI will check.

Sources/Datadog/Tracing/DDSpan.swift
```diff
@@ -43,20 +43,29 @@ internal class DDSpan: OTSpan {
         set { ddTracer.queue.async { self.unsafeIsFinished = newValue } }
     }
 
+    /// Unsynchronized span type. Use `self.type` setter & getter.
+    private var unsafeType: String
+    private(set) var type: String {
+        get { ddTracer.queue.sync { unsafeType } }
+        set { ddTracer.queue.async { self.unsafeType = newValue } }
+    }
+
```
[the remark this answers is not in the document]
nachoBonafonte: Yep, much cleaner, its not going to change

Sources/Datadog/Tracing/Utils/URLFilter.swift
```diff
@@ -5,6 +5,9 @@
  */
 
 import Foundation
+#if canImport(_Datadog_TestRunner)
```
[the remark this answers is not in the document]
nachoBonafonte: Yes, will remove

Sources/_Datadog_TestRunner/DDTestRunner.swift
```diff
@@ -0,0 +1,138 @@
...
+import Foundation
+@_implementationOnly import XCTest
+
+internal class DDTestRunner: NSObject, XCTestObservation {
+    static var instance: DDTestRunner?
+
+    let testNameRegex = try? NSRegularExpression(pattern: "([\\w]+) ([\\w]+)", options: .caseInsensitive)
+    let supportsSkipping = NSClassFromString("XCTSkippedTestContext") != nil
+    var currentBundleName = ""
+    var activeTestSpan: DDSpan?
```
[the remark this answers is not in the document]
nachoBonafonte: changed

Sources/_Datadog_TestRunner/DDTestRunner.swift
nachoBonafonte: Added as constant

Sources/_Datadog_TestRunner/DDTestRunner.swift
nachoBonafonte: yes, much nicer

Sources/_Datadog_TestRunner/DDTestRunner.swift
nachoBonafonte: Modified to store also Datadog environment values (DATADOG_CLIENT_TOKEN, DD_SERVICE and DD_ENV), and not only CI values. 
Renamed to DDEnvironmentValues. 
Added test for these new values

Datadog/Example/ExampleTests/ExampleTests.swift
nachoBonafonte: Yes, that is a great idea, will work on it.

Sources/_Datadog_TestRunner/DDTestRunner.swift
```diff
@@ -22,22 +23,15 @@ internal class DDTestRunner: NSObject, XCTestObservation {
...
             return
         }
 
         Datadog.initialize(
             appContext: .init(),
             configuration: Datadog.Configuration
-                .builderUsing(clientToken: clientToken, environment: CIEnvironmentValues.getEnvVariable("DD_ENV") ?? "test")
-                .set(serviceName: CIEnvironmentValues.getEnvVariable("DD_SERVICE") ?? ProcessInfo.processInfo.processName)
+                .builderUsing(clientToken: clientToken, environment: env.ddEnvironment ?? "test")
+                .set(serviceName: env.ddService ?? ProcessInfo.processInfo.processName)
```
ncreated, line 34: If not provided, `serviceName` defaults to the app bundle identifier, then to `"ios"`:
https://github.com/DataDog/dd-sdk-ios/blob/787a9d2a5d754a71e920f82839398135e2c229d3/Sources/Datadog/Core/FeaturesConfiguration.swift#L82
Not sure if we're interested in having the same defaults in the tests runner - but why not (all data is already segmented by the span type).

nachoBonafonte:
We are definitely moving in other direction with testing instrumentation, so closing this PR

PR #249, opened 2020-09-10
RUMM-709 Upload delay increases in case of network error

buranmert:
### What and why?

SDK was decreasing upload interval as long as there is a batch to upload
That caused problems in case of network errors

Example case: the device can't resolve upload URL for some reason, yet SDK keeps trying to upload at short intervals.

### How?

Now if there is a network error or client token error, interval increases (interval used to decrease in that case)

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/Upload/DataUploadDelay.swift
+++ b/Sources/Datadog/Core/Upload/DataUploadDelay.swift
@@ -7,7 +7,7 @@
 import Foundation
 
 internal protocol Delay {
-    func nextUploadDelay() -> TimeInterval
+    var current: TimeInterval { get }
     mutating func decrease()
     mutating func increase()
 }
@@ -29,9 +29,7 @@ internal struct DataUploadDelay: Delay {
         self.delay = performance.initialUploadDelay
     }
 
-    func nextUploadDelay() -> TimeInterval {
-        return delay
-    }
+    var current: TimeInterval { delay }
 
     mutating func decrease() {
         delay = max(minDelay, delay * (1.0 - changeRate))
--- a/Sources/Datadog/Core/Upload/DataUploadDelay.swift
+++ b/Sources/Datadog/Core/Upload/DataUploadDelay.swift
@@ -6,12 +6,18 @@
 
 import Foundation
 
+internal protocol Delay {
+    func nextUploadDelay() -> TimeInterval
+    mutating func decrease()
+    mutating func increase()
+}
+
 /// Mutable interval used for periodic data uploads.
-internal struct DataUploadDelay {
+internal struct DataUploadDelay: Delay {
     private let defaultDelay: TimeInterval
     private let minDelay: TimeInterval
     private let maxDelay: TimeInterval
-    private let decreaseFactor: Double
+    private let changeRate: Double
 
     private var delay: TimeInterval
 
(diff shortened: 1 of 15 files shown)


Sources/Datadog/Core/Upload/DataUploadWorker.swift
```diff
@@ -67,21 +67,23 @@ internal class DataUploadWorker: DataUploadWorkerType {
...
-
-                self.delay.decrease()
             } else {
                 let batchLabel = nextBatch != nil ? "YES" : (isSystemReady ? "NO" : "NOT CHECKED")
                 let systemLabel = isSystemReady ? "✅" : "❌"
                 developerLogger?.info("💡 (\(self.featureName)) No upload. Batch to upload: \(batchLabel), System conditions: \(systemLabel)")
                 userLogger.debug("💡 (\(self.featureName)) No upload. Batch to upload: \(batchLabel), System conditions: \(systemLabel)")
 
-                self.delay.increaseOnce()
+                self.delay.increase()
```
buranmert, line 86: i couldn't find a way to test `increase/decrease` logic here (without sleeping the thread for a while)
i'm thinking of moving this logic to `UploadDelay` type, basically like that:
```swift
struct DataUploadDelay {
  ...
  mutating func handle(uploadStatus: DataUploadStatus?) {
    if nil -> no upload tried (increase)
    if failure -> upload failed (increase)
    if success/droppable -> upload succeeded (decrease)
  }
}
```
then i can test the logic without `DataUploadWorker`/`queue.async {...}`
otherwise i can leave it as is, we don't have tests for this logic already

wdyt @ncreated ?

Sources/Datadog/Core/Upload/DataUploadDelay.swift
```diff
@@ -6,37 +6,38 @@
 
 import Foundation
 
+internal protocol Delay {
+    func nextUploadDelay() -> TimeInterval
+    mutating func decrease()
+    mutating func increase()
+}
```
ncreated, line 13: nit: the name, `Delay` is domain-agnostic, but the `nextUploadDelay()` has "upload" meaning. This
```swift
func next() -> TimeInterval
mutating func decrease()
mutating func increase()
```
would be more coherent IMO.

Tests/DatadogTests/Datadog/Core/Upload/DataUploadWorkerTests.swift
```diff
@@ -22,68 +39,187 @@ class DataUploadWorkerTests: XCTestCase {
...
 
+            uploaderQueue.sync {} // wait until last "process upload" operation completes (to make sure "delete file" was requested)
+            fileReadWriteQueue.sync {} // wait until last scheduled "delete file" operation completed
+
+            XCTAssertEqual(try temporaryDirectory.files().count, 0)
+        }
+    }
+
+    // swiftlint:disable multiline_arguments_brackets
+    func test_whenThereIsNoBatch_thenIntervalIncreases() throws {
```
ncreated, line 76: Convention: we commonly use `testWhen`, not `test_when`

Screenshot 2020-09-11 at 17 33 40

ncreated (changes requested):
Looks good 👌, just one nit and convention. Let's fix and merge.

Sources/Datadog/Core/Upload/DataUploadDelay.swift
buranmert: i didn't pay attention before but actually `current` makes more sense than `next`, wdyt?

ncreated (approved):

PR #253, opened 2020-09-22
RUMM-734 Add Tracing auto-instrumentation tests for `NSURLSession`

ncreated:
### What and why?

📦 This PR adds integration test for `NSURLSession` (Objective-C) auto instrumentation with Tracing.

This is to add regression tests coverage for upcoming changes to Tracing auto instrumentation required by RUM.

### How?

**Before**, there was only one Tracing integration test, working this way:
* Start VC which:
   * sends bunch of `Spans` using Tracer API,
   * then sends 3 auto-instrumented `URLSession` requests.
* Assert all `Spans` delivered to the Python server.

**In this PR**, I split this into 3 separate integration tests:
1. `TracingManualInstrumentationScenarioTests`
* Start VC which sends bunch of `Spans` using Tracer API.
* Asserts `Spans` delivered to the Python server.

2. `TracingURLSessionScenarioTests` → `testTracingURLSessionScenario()`:
* Start VC which sends 3 auto-instrumented `URLSession` requests (in Swift).
* Asserts `Spans` delivered to the Python server.

3. `TracingURLSessionScenarioTests` → `testTracingNSURLSessionScenario()`:
* Start VC which sends 3 auto-instrumented `NSURLSession` requests (in Objective-C).
* Asserts `Spans` delivered to the Python server.

As `2` and `3` are very similar (they send requests to the same endpoints), I shared all the assertions code between them. The only difference is that `URLSession` requests are send from Swift VC, while `NSURLSession` requests are fired from VC written in Objective-C.

Beside the new VC written in Objective-C and 2 new Storyboards, this PR has no logic changes - only refactoring.

### Review checklist


mariusc83 (approved):
LGTM

Datadog/Example/Scenarios/Tracing/URLSessionAutoInstrumentation/TracingURLSessionViewController.swift
```diff
@@ -0,0 +1,63 @@
...
+        let task = session.dataTask(with: testScenario.customPOSTRequest) { _, _, error in
+            assert(error == nil)
+            completion()
+        }
+        task.resume()
+    }
+
+    private func callBadURL(completion: @escaping () -> Void) {
+        let task = session.dataTask(with: testScenario.badResourceURL) { _, _, error in
+            assert(error != nil)
```
mariusc83, line 49: it's weird that you are doing the assertions in the ViewController directly.

mariusc83 (approved):

Datadog/Example/Scenarios/Tracing/URLSessionAutoInstrumentation/TracingURLSessionViewController.swift
ncreated: It's just sanity. If it's not met in integration tests, this will crash the host app (and fail the test). I added it here, as we use exactly the same setup for debugging our SDK with running the `Example` app directly - in such case I want to have an instant feedback on the assumption (if it has or doesn't have the error).

PR #254, opened 2020-09-25
RUMM-717 Add RUM Actions auto instrumentation for "tap" event

ncreated:
### What and why?

🚚 This PR adds the RUM Actions auto instrumentation for "tap" event. If enabled, all `UITouch` events recognised on the `UIControl` or collection cells will be reported as RUM Action. Any events on the on-screen keyboard are ignored for privacy reason.

### How?

One public API was added to the SDK configuration:
```swift
/// Enables or disables automatic tracking of `UITouch` events as RUM Actions.
///
/// When enabled, the SDK will capture every `UIEvent` send to the application and use the in-build heuristic to recognize
/// key `UIViews` and `UIControls` that user interacts with. For every recognized element, the RUM Action is send automatically.
/// The action will be named by the class name of interacted element and will be extended with `accessibilityIdentifier` (if it's set) for more context.
///
/// ...
///
/// - Parameter enabled: `false` by default
public func trackUIKitActions(_ enabled: Bool) -> Builder
```

The internal implementation follows existing conventions on passing the instrumentation config and follows the RFC proposal for swizzler and handler.

I decided to swizzle the `UIApplication.sendEvent(_:)` as it gives broader information than `UIApplication.sendAction(...)` (which gets triggered only for interactions with `UIControls`).

As part of this PR I also introduce the test scenario UI in the `Example` app (integration test implementation will come within a next PR). I used it to debug and adjust the implementation. It uses this setup:

Screenshot 2020-09-25 at 14 26 50

for sending those example actions:

Screenshot 2020-09-25 at 14 24 32

### Review checklist


xgouchet (approved):
LGTM

Sources/Datadog/Core/FeaturesConfiguration.swift
```diff
@@ -121,9 +121,10 @@ extension FeaturesConfiguration {
         if configuration.rumEnabled {
             var autoInstrumentation: RUM.AutoInstrumentation?
 
-            if let rumUIKitViewsPredicate = configuration.rumUIKitViewsPredicate {
+            if configuration.rumUIKitViewsPredicate != nil || configuration.rumUIKitActionsTrackingEnabled {
```
buranmert, line 124: isn't this confusing for the user? maybe we can use an enum here like 
`rumUIKitActionsTrackingEnabled = .forNone/.forSome(aPredicate)/.forAll`?

ncreated: The `predicate` is for RUM Views, whereas the `enabled` flag is for RUM Actions tracking. From the user perspective, it's very simple:
```swift
configuration
   .trackUIKitRUMViews(using: predicate) // to enable RUM Views tracking
   .trackUIKitActions(true) // to enable RUM Actions tracking
```

buranmert: ah okay my bad, i missed actions vs views

buranmert (approved):

PR #258, opened 2020-09-29
RUMM-739 Xcode 12 migration

buranmert:
### What and why?

Xcode 12 is released 🚀 We cleared new warnings.
But Bitrise's Xcode 12 stack doesn't have earlier iOS versions (starting from iOS 13.3) ❌ 

### How?

We split `test-all-platforms` workflow into `test-all-platforms-xcode11` and `xcode12`
They use different stacks in Bitrise in order to cover iOS runtimes as early as iOS 12.0

`bitrise-test-all-platforms.yml` is renamed to `bitrise-test-all-platforms-xcode11.yml` ♊ 
`bitrise-test-all-platforms-xcode12.yml` is passing (with the changes from #257 ) ✅ 
Screenshot 2020-09-29 at 18 41 49

## TODO

- [ ] #257 should be merged first

### Review checklist


ncreated (changes requested):
Looks good 👌, except that:
- [ ] we also need to update the **run-all-platform-tests** trigger to start two bitrise jobs instead of just one and to use the correct workflow name. This can be configured in `.github/workflows/all-platform-tests.yml`.

buranmert:
test-all-platforms buranmert/RUMM-739-xcode-12

buranmert:
@test-all-platforms buranmert/RUMM-739-xcode-12

ncreated:
> @test-all-platforms buranmert/RUMM-739-xcode-12
> 
> @ncreated good catch, i totally forgot about GH action 👍
> well i think i can't test it here before merging it into `master` but workflow should be fine now

The action code looks OK, but for some reason there is only one build on Bitrise and it got aborted:

Screenshot 2020-10-01 at 14 06 43

When calling the trigger, it should fire up two builds. Maybe it's because we have the "abort previous build for the same branch" setting enabled on Bitrise?

buranmert:
i looked at Github action details and it picked up `yml` file from `master` branch, which doesn't have this change
that's why it fired old `trigger_all_platform_tests` instead of `trigger_all_platforms_xcode11` and `xcode12`

i aborted the build manually ✋ as it was clear that this workflow isn't what we want ❌ 
once we merge this PR, GH Action should fire both workflows ✅

ncreated (approved):

PR #260, opened 2020-09-30
RUMM-740 Fix: UI Test software keyboard issue

buranmert:
### What and why?

iOS 13 Swipe Typing feature blocks software keyboard with onboarding
We must tap Continue button in order to get the real keyboard

software-keyboard-detail

### How?

Now UI Test tries to tap on `Continue` button if exists before typing given letters
(For demonstration video check out jira ticket 🎥 )

### Review checklist

### DIFF
--- a/Tests/DatadogIntegrationTests/Scenarios/RUM/RUMTapActionScenarioTests.swift
+++ b/Tests/DatadogIntegrationTests/Scenarios/RUM/RUMTapActionScenarioTests.swift
@@ -40,6 +40,13 @@ private extension ExampleApplication {
     }
 
     func enterTextUsingKeyboard(_ text: String) {
+        // NOTE: RUMM-740 iOS 13 Swipe typing feature presents its onboarding
+        // That blocks the keyboard with a Continue button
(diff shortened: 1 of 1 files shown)


Tests/DatadogIntegrationTests/Scenarios/RUM/RUMTapActionScenarioTests.swift
```diff
@@ -89,7 +96,7 @@ class RUMTapActionScenarioTests: IntegrationTests, RUMCommonAsserts {
         app.tapShowVariousUIControls()
         app.tapTextField()
         // TODO: RUMM-740 Enable it back once we fix the Software Keyboard work on CI
```
ncreated, line 98:
```suggestion
```

ncreated (changes requested):
Nice we found it 👍. The test still fails, but it's due to:
```
TODO: RUMM-742 Make the HTTP Server Mock not depend on the requests count
```
I think we should bump the priority of this ticket.

ncreated (approved):

PR #261, opened 2020-10-02
RUMM-738 Shopist: Run is bound to time limit instead of session count

buranmert:
### What and why?

Shopist events were not aligned with its build triggers.
1. Runs are triggered once in every 30mins
2. Yet, runs usually take less than 30mins: sometimes takes 15mins sometimes 25mins
That leaves gaps in timeseries of events in RUM Explorer, which is not desired.

### How?

Now we run until the given target duration is reached.
Shopist continues running for 27.5 minutes regardless of session count.

### Review checklist

### DIFF
--- a/Shopist/ShopistUITests/ShopistUITests.swift
+++ b/Shopist/ShopistUITests/ShopistUITests.swift
@@ -6,15 +6,19 @@
 
 import XCTest
 
-let targetSessionCount = 10
-
 class ShopistUITests: XCTestCase {
+    private func shouldContinueRunning(ifStartedAt startDate: Date) -> Bool {
+        let targetRunDuration: TimeInterval = 27.5 * 60.0
+        return Date().timeIntervalSince(startDate) < targetRunDuration
+    }
+
     func testRunFullCheckoutFlow() {
+        let startDate = Date()
         let app = ShopistApp()
-        for sessionIndex in 1...targetSessionCount {
+        while shouldContinueRunning(ifStartedAt: startDate) {
             let categories = app.launchToHomepage()
 
-            let targetItemCountInCart = Int.random(in: 1...5)
+            let targetItemCountInCart = Int.random(in: 1...2)
             var itemCount = 0
             while itemCount < targetItemCountInCart {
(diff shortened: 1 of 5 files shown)


buranmert:
ℹ️ Carthage tests fail due to API rate limiting, other tests pass. I'm moving it to ready for review

Shopist/ShopistUITests/ShopistUITests.swift
```diff
@@ -6,15 +6,20 @@
 
 import XCTest
 
-let targetSessionCount = 10
+let targetSessionDuration: TimeInterval = 27.5 * 60.0 // 25mins
```
ncreated, line 9:
```suggestion
let targetSessionDuration: TimeInterval = 27.5 * 60.0 // 27.5mins
```

ncreated (commented):
👍 this should fix the "gaps" in events distribution over time. 

But how this PR relates to events variety in Shopist iOS vs Shopist Android (also part of RUMM-738)?

Shopist iOS, daily stats:
Screenshot 2020-09-25 at 18 04 17

Shopist Android, daily stats:
Screenshot 2020-09-25 at 18 04 22

e.g.:
* on iOS we have `~9.7` Errors per Session, while on Android: `~0.23`,
* on iOS we have `~0.46` Errors per View (almost every second View has an Error), while on Android: `~0.038`.

The goal of Shopist is to mimic the "regular iOS app", and I think the Android ratios are more natural for this goal.

Shopist/ShopistUITests/ShopistUITests.swift
buranmert: 👍 also i just noticed `targetSessionDuration` is misleading, it should be `targetRunDuration`

buranmert:
@ncreated it will probably look good enough if not matching android numbers exactly, hard to estimate without running some batches.
and they can change again when we switch to auto-instrumentation.
so i propose waiting for feedback from sales before adjusting the number ⏳

ncreated (approved):

PR #263, opened 2020-10-05
RUMM-759 Workaround for Carthage&Xcode 12 issue

buranmert:
### What and why?

Carthage has an known issue with Xcode 12
For more: https://github.com/Carthage/Carthage/issues/3019

### How?

This workaround excludes ARM arches from carthage build
It fixes our build failure as lipo doesn't merge x86 and ARM anymore
Source: https://github.com/Carthage/Carthage/issues/3019#issuecomment-664099506

### Review checklist

### DIFF
--- a/dependency-manager-tests/carthage/Makefile
+++ b/dependency-manager-tests/carthage/Makefile
@@ -5,6 +5,9 @@ else
 endif
 
 PWD := $(shell pwd)
+# TODO: RUMM-760 Remove this workaround once Carthage fixes their Xcode 12 issue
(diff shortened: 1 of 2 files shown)


ncreated (approved):
Nice 👌 So, it's a Carthage issue, and nothing is badly configured / can be fixed on our side, right?

buranmert:
@ncreated yes, it's not _our_ problem in that sense.
looks like Xcode 12 changed something and that broke the way carthage works - lipo'ing different slices. maybe they will need to switch to XCFramework, i don't know exactly though.
for now we can go along with this workaround

PR #269, opened 2020-10-06
RUMM-732 RUM Resources auto instrumentation

ncreated:
### What and why?

📦 This PR adds RUM Resources auto instrumentation along with first temporal metric collection - DNS lookup phase:

Screenshot 2020-10-05 at 17 18 03

### How?

I had to add few last missing pieces to have the RUM Resource values automatically computed:
* `RUMAddResourceMetricsCommand` - a new command, consumed by the `RUMResourceScope`. If received by the scope, it overwrites existing temporal metrics and adds new one (i.e. for DNS).
* Automatic inference of `RUMHTTPMethod` based on the `request`.
* Automatic inference of `RUMResourceKind` based on the `request` and `resource` MIME type.
* `URLFilter` is now split into _first party URLs filter_ + _internal URLs filter_ as both are used exclusively.

Most of the additions in this PR come from introduction this basic stuff, although about 60% of this PR are tests.

The actual work of RUM Resources monitoring is done by introducing `URLSessionRUMResourceHandler` (handler), which is used by `URLSessionInterceptor` (interceptor).

From now, when `URLSession` auto instrumentation is enabled:
* if intercepted request is first party url, both Tracing span and RUM Resource are send,
* if intercepted request is third party url, only RUM Resource is send,
* if intercepted request is an internal URL (used by the SDK), no Tracing span and no RUM Resource are send.

To keep this PR small, the integration test suite will come in a next PR.

### Review checklist


Sources/Datadog/URLSessionAutoInstrumentation/Interception/URLSessionInterceptor.swift
```diff
@@ -30,66 +30,87 @@ internal protocol URLSessionInterceptorType: class {
 }
 
 internal class URLSessionInterceptor: URLSessionInterceptorType {
-    private let firstPartyHostsFilter: URLFilter
-    private let queue: DispatchQueue
+    /// Filters first party `URLs` defined by the user.
+    private let firstPartyURLsFilter: FirstPartyURLsFilter
+    /// Filters internal `URLs` used by the SDK.
+    private let internalURLsFilter: InternalURLsFilter
```
buranmert, line 36: i think we could keep those 2 merged in 1 `URLFilter`, is there a reason for the split?

Sources/Datadog/URLSessionAutoInstrumentation/Interception/URLSessionInterceptor.swift
```diff
@@ -30,66 +30,87 @@ internal protocol URLSessionInterceptorType: class {
...
+        tracingHandler: URLSessionTracingHandlerType?,
+        rumResourceHandler: URLSessionRUMResourcesHandlerType?
     ) {
-        self.firstPartyHostsFilter = firstPartyHostsFilter
-        self.tracingInterceptionHandler = tracingInterceptionHandler
+        self.firstPartyURLsFilter = FirstPartyURLsFilter(configuration: configuration)
+        self.internalURLsFilter = InternalURLsFilter(configuration: configuration)
+        self.tracingHandler = tracingHandler
+        self.rumResourceHandler = rumResourceHandler
         self.queue = DispatchQueue(label: "com.datadoghq.URLSessionInterceptor", target: .global(qos: .utility))
```
buranmert, line 65: it's out of scope but anyway, i'm not sure if `interceptor` needs its own queue: RUM has its own queue already and starting spans in different queues may have effects on `activeSpan`.
also i guess that makes its tests async.

is there a reason for using a private queue here?

ncreated: This `queue` synchronizes the access to:
```swift
/// Maps `URLSessionTask` to its `TaskInterception` object.
private var interceptionByTask: [URLSessionTask: TaskInterception] = [:]
```
While we can expect that single `URLSession` synchronizes its calls using its own `queue`, we must assume that the user can use several `URLSessions`.

Speaking on linking network `Spans` with the `activeSpan` running in the app - we're not really focusing on it:
- it was never an objective / requirement to do support it;
- this won't be soon possible (RUMM-696) as the `Span` will be generated by backend.

Sources/Datadog/RUM/AutoInstrumentation/RUMAutoInstrumentation.swift
```diff
@@ -70,8 +69,9 @@ internal class RUMAutoInstrumentation {
                 userActions = nil
             }
         } catch {
-            userLogger.warn("🔥 RUM automatic tracking can't be set up due to error: \(error)")
-            developerLogger?.warn("🔥 RUM automatic tracking can't be set up due to error: \(error)")
+            consolePrint(
+                "🔥 Datadog SDK error: RUM automatic tracking can't be set up due to error: \(error)"
+            )
```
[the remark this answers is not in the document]
ncreated: It's not a change of the trouble-reporting approach 🙂 - this one was just incorrect. The swizzling is applied at the SDK initialization time, when the usage of `userLogger` is invalid, because this logger is flagged by the `Datadog.verbosity = .debug` setup applied by the user after the SDK is initialised. In many other places, `user` / `dev` loggers are used to report problems **after** SDK initialization, which is correct.

Sources/Datadog/URLSessionAutoInstrumentation/Interception/URLSessionInterceptor.swift
buranmert: if you prefer removing this queue/less queues less synchronization, we can use associated objects on `task`
```swift
func taskCreated(task) {
  task.interception = Interception(...)
}
func taskCompleted(task) {
  task.interception?.completed()
}
```
how does that look like?

Sources/Datadog/URLSessionAutoInstrumentation/Interception/URLSessionInterceptor.swift
ncreated: I did the split, as both work very differently and are configured with distinct information. It's more SRP to have two, instead of one `URLFilter` which would have half code dedicated for "first party hosts filtering" and second half for "internal URLs filtering", but nothing in common (the [cohesion](https://en.wikipedia.org/wiki/Cohesion_(computer_science)) score for such component would be `0`).

Sources/Datadog/URLSessionAutoInstrumentation/Interception/URLSessionInterceptor.swift
ncreated: Indeed, that would mean no need for synchronisation. However, referring to the RFC, what we have discussed in CHALLENGE 4 was preparing the solution to not depend on runtime modifications. And with its current implementation the `URLSessionInterceptor` does not depend on any objc runtime at all. This means, we could easily make it `public` and let users call it directly, having all the functionality of RUM and APM instrumentation for `0` swizzling / associated objects cost.

We don't focus on making the `URLSessionInterceptor` public now, but given two comparable approaches, I'd rather prefer the one which scales better for CHALLENGE 4. WDYT?

buranmert: i think we can make it public in both cases because `task.interception` will be private, can be used within `Interceptor.swift` only ✅ 
i guess assoc objects are different than swizzling as we add new things and not replacing existing things, it's closer to `extension`s in that sense.
we don't change the memory layout of types or anything and also i don't think the cost of associated object is heavier than queue sync'ing.
sounds good?

ncreated: Although I see associated objects as a method of last resort, I did it a try. When going this way, the `testRandomlyCallingDifferentAPIsConcurrentlyDoesNotCrash()` test (in `URLSessionInterceptorTests`) shows the synchronization issue in `TaskInterception`:

Screenshot 2020-10-08 at 08 34 03

It's because `TaskInterception.register(metrics:)`, `TaskInterception.register(completion:)` and `TaskInterception.isDone` are called from two separate interceptor methods:
```swift
// URLSessionInterceptor

func taskMetricsCollected(urlSession: URLSession, task: URLSessionTask, metrics: URLSessionTaskMetrics)
func taskCompleted(urlSession: URLSession, task: URLSessionTask, error: Error?)
```
and this test tries to run them concurrently for the same task.

The [`URLSession.init()` documentation says](https://developer.apple.com/documentation/foundation/urlsession/1411597-init) however that each session synchronizes its delegate cals and completion handlers using session `queue`, so this test should be rather improved to synchronize calls for the same `URLSession` instance or removed entirely.

However, it is fact that calls between different session instances will happen concurrently, meaning that we may need to add thread synchronization as soon as we introduce any shared resource in `URLSessionInterceptor` or any of it's child components: `URLSessionTracingHandler` or `URLSessionRUMResourcesHandler`. If we're OK with this eventual cost, I can change to use runtime utils instead of synchronization `queue`. LTMWDYT @buranmert @xgouchet @mariusc83

buranmert: to my understanding tasks are intercepted independently, that's why i can't think of a case with shared resource among interceptions. can you think of any?
if there are, then yes we can't get rid of `queue`, otherwise associated objects can simplify the code.

if `TracingHandler` or `RUMHandler` change `Tracer`'s or `RUMMonitor`'s states, those handlers can have their own queue but still `Interceptor` can stay without queue.

buranmert: conclusion: we keep the shared resource and async execution instead of using `associatedObject`s
@ncreated 's point is AO is in `objc/runtime.h` and considered dangerous, it's also slow.
i don't agree and i think it would improve the code in our case but it works either way, so convo is resolved ✅

buranmert (approved):

PR #270, opened 2020-10-07
RUMM-732 Add RUM Resources auto instrumentation integration test

ncreated:
### What and why?

🧪 This PR adds integration test to RUM Resources auto instrumentation.

### How?

I reused test setup made for Tracing auto instrumentation tests. The difference was that in RUM Resources we also want to test that resources are linked to the right RUM Views, so we need more than one VC. Because of this, I updated the Tracing tests to navigate through two view controllers:

Screenshot 2020-10-07 at 12 25 06

The first one sends requests to first party hosts, the second one targets third parties.

To make the execution faster, now we send all requests in parallel, instead of sequentially. This required passing distinct `URLs` from the UITests runner to tested app, so that we can match the `RUM Event` and `Span` to its corresponding `URL`. This is done by passing `HTTPServerMockConfiguration` struct through ENV variables instead of just one single, custom `URL`.

I also added assertions for request metrics:
* request duration,
* DNS phase resolution,

so we make sure it all works in production setup.

### Review checklist


Tests/DatadogIntegrationTests/UITestsHelpers.swift
```diff
@@ -8,23 +8,16 @@ import XCTest
...
         launchArguments = [Environment.Argument.isRunningUITests]
 
         var variables: [String: String] = [:]
         variables[Environment.Variable.testScenarioIdentifier] = testScenario.envIdentifier()
-
-        logsEndpointURL.flatMap { variables[Environment.Variable.logsEndpoint] = $0.absoluteString }
-        tracesEndpointURL.flatMap { variables[Environment.Variable.tracesEndpoint] = $0.absoluteString }
-        rumEndpointURL.flatMap { variables[Environment.Variable.rumEndpoint] = $0.absoluteString }
-        customEndpointURL.flatMap { variables[Environment.Variable.customEndpoint] = $0.absoluteString }
+        variables[Environment.Variable.serverMockConfiguration] = serverConfiguration.toEnvironmentValue
```
buranmert, line 20: i didn't understand this part: first we serialize `serverConf` into base64 and then deserialize it into a Swift object? why do we do that?

buranmert (approved):
i have a question above but LGTM

Tests/DatadogIntegrationTests/UITestsHelpers.swift
ncreated: The `ProcessInfo.environment` expects `[String : String]`, so we need to represent the configuration as JSON `String`. I used Base64 additionally to limit the characters to ASCII - depending on how Xcode passes those values to the Simulator process, I was afraid that some characters in JSON may break the input (but I didn't test if it breaks without Base64).

PR #272, opened 2020-10-08
RUMM-801 Fix tests flakiness on CI

ncreated:
### What and why?

⚙️ This PR fixes recent flakiness observed on CI.

### How?

When analysing failed job reports, I observed two main sources of the flakiness:
* too optimistic timeouts,
* failed preconditions in `ServerMock`.

The conclusion made back in #72 was that for some complex tests, CI needs the timeout of 40x more than the test execution in local. I bumped recently failing timeouts to level close to this.

With the introduction of `URLSessionDelegate`, we no longer make blocking `expectation.fulfill()` for all `URLSessionTask` completions (the one received on delegate are not blocking). This makes some of the original assumptions included in `ServerMock` invalid. I changed them to print warning, instead of failing the test.

More in description of each commit.

### Result

I ran unit tests successfully 20 times in a row on CI ✅, vs 40% of flakiness before.

### Further improvements

This PR should improve the workflow for now, but some small flakiness may still be observed. There are two JIRAs created a while a go and both are still valid:
* `RUMM-417` to look closer into `ServerMock`,
* `RUMM-742` to make integration tests not depend on the number of requests sent to the Python server.

### Review checklist

### DIFF
--- a/Tests/DatadogTests/Datadog/Mocks/ServerMock.swift
+++ b/Tests/DatadogTests/Datadog/Mocks/ServerMock.swift
@@ -54,7 +54,10 @@ private class ServerMockProtocol: URLProtocol {
     }
 
     override func stopLoading() {
-        precondition(ServerMock.activeInstance != nil, "Request was stopped while no `ServerMock` instance is active.")
+        if ServerMock.activeInstance != nil {
+            // This may happen when testing requests which deliver completion on the `URLSessionDelegate`.
+            print("⚠️ Request was stopped while no `ServerMock` instance is active.")
+        }
     }
 }
 
--- a/Tests/DatadogTests/Datadog/Mocks/ServerMock.swift
+++ b/Tests/DatadogTests/Datadog/Mocks/ServerMock.swift
@@ -32,24 +32,20 @@ private class ServerMockProtocol: URLProtocol {
     }
 
     override func startLoading() {
-        guard let serverMock = ServerMock.activeInstance else {
-            preconditionFailure("Request was started while no `ServerMock` instance is active.")
-        }
-
-        if let response = serverMock.mockedResponse {
+        if let response = ServerMock.activeInstance?.mockedResponse {
             client?.urlProtocol(self, didReceive: response, cacheStoragePolicy: .notAllowed)
         }
-        if let data = serverMock.mockedData {
+        if let data = ServerMock.activeInstance?.mockedData {
             client?.urlProtocol(self, didLoad: data)
         }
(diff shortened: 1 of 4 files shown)


Tests/DatadogTests/Datadog/Mocks/ServerMock.swift
```diff
@@ -32,29 +32,28 @@ private class ServerMockProtocol: URLProtocol {
     }
 
     override func startLoading() {
-        guard let serverMock = ServerMock.activeInstance else {
-            preconditionFailure("Request was started while no `ServerMock` instance is active.")
```
buranmert, line 36: this part is not clear: what does it mean for us if `activeInstance` is `nil`? it doesn't fail anymore but is it a success?

ncreated: There is only one global session instance: `URLSession.serverMockURLSession`, configured with the `ServerMockProtocol`. When `ServerMock` instance is created it sets itself to static `ServerMock.activeInstance` weak property. The protocol talks to the `server` through this `activeInstance`.

If the `activeInstance` is `nil` it basically means the `ServerMock` was destroyed, so the test completed and there is no other running test which created the `server`.

Now, I made this change because some tests were failing on:
```swift
// ServerMock

deinit {
    precondition(Thread.isMainThread, "`ServerMock` should be deinitialized on the main thread.")
}
```

As this `func startLoading()` is the only place where we capture strong reference to the `ServerMock` on a background thread, this is probably the suspicious place causing the flakiness. My bet is that we end up some test from one of the client completion calls below, e.g. `client?.urlProtocol(self, didLoad: data)`, which completes the test, releases the reference to `ServerMock` on the test's thread, but the captured reference here in `startLoading()` still remains, causing the deinit off the main thread.

---

Bottom line: I don't like the current setup of `ServerMock` and the shared `URLSession.serverMockSession` approach. What we should probably do instead is creating one `URLSession` for each `ServerMock`. This could be done in `RUMM-417`.

buranmert: in short `activeInstance == nil` is not an expected case and we are not sure how we end up there.?
yes let's see in 417 👌

buranmert (approved):

Tests/DatadogTests/Datadog/Mocks/ServerMock.swift
ncreated: Yes, it's not an expected state 👌.

PR #275, opened 2020-10-12
Add Zen doc

xgouchet:


ZEN.md
```diff
@@ -0,0 +1,66 @@
...
+- Keep backward compatibility on minor updates
+
+## Trust
+
+- Use sensible static analysis tools
+- Use benchmark tools to measure the performance of the library
+- All parts of the libraries must be thoroughly tested
+
+
+## Workflow
```
buranmert, line 56: extra newline
```suggestion
- All parts of the libraries must be thoroughly tested

## Workflow
```

ZEN.md
```diff
@@ -0,0 +1,66 @@
...
+
+- All code must be tested and reviewed;
+- Keep your PR small (each PR must solve one and only one issue);
+- Keep your PR as Draft until it's ready to be reviewed;
+
+## Code architecture
+
+- Favor Object Oriented design, rather than procedural programing;
```
buranmert, line 64: doesn't that sound like a false dichotomy? what about functional programming and/or others then?

buranmert (changes requested):

ZEN.md
xgouchet: The kind of functional programming we use in Swift/Kotlin is not true stateless functional programming.

ZEN.md
```diff
@@ -0,0 +1,65 @@
...
+    - Avoid unnecessary dependencies;
+- **Network load**
+    - Batch requests as much as possible;
+    - Avoid making request when the device is asleep or has low battery;
+
+## Stability
+
+- **Zero crash caused by our code!**
+    - Unless the crash is on top-level method, caused by identifiable developer mistake, and notified with an understandable error message;
+    - Even in top-level methods, prefer making the library non operating and logging the issue in the console logs, rather than throwing an exception
```
ncreated, line 27: 💅 
```suggestion
    - Unless the crash is on top-level method, caused by identifiable developer mistake, and notified with an understandable error message.
    - Even in top-level methods, prefer making the library non operating and logging the issue in the console logs, rather than throwing an exception.
```

ncreated (approved):
🏅

buranmert (approved):

Issue #274, opened 2020-10-10
Not working for macCatalyst and tvOS


Hengyu:
### The issue

The latest package won't get compiled when the project target is set to `macCatalyst` or `tvOS`.

### Detail

Since the `CoreTelephony` has not been supported for `macCatalyst` and `tvOS`. (Even the official documentation says the `CoreTelephony` can be used in `macCatalyst 13.0+`) So the compiler would generate complaints when build for mac and tv.

I know this SDK is targeted for iOS, however, after the release of macCatalyst, the same project could also be compiled to mac app. I tried to constrain the platforms to iOS only in project build settings. But the compiler is still building the Datadog module when targeting mac. Looks like this is a bug between Swift Package Manager & Xcode.

<img width="566" alt="screenshot" src="https://user-images.githubusercontent.com/4032674/95649232-11718780-0b0f-11eb-9c38-aa15532fe844.png">

### Proposed solution

Add macro in files who are using `CoreTelephony`

https://github.com/Hengyu/dd-sdk-ios/commit/c51a24fa9d2ef626b5b7e1be07a1a4c68e287034

---

#### Datadog SDK version:

`1.3.1`

#### Last working Datadog SDK version:

N.A.

#### Dependency Manager:

Swift Package Manager

#### Other toolset:

N.A.

#### Xcode version:

`Xcode 12.0 (12A7209)`

#### Swift version:

`Swift 5.1`

#### Deployment Target:

`macCatalyst 13.0+` & `tvOS 11.0+`

#### macOS version:

`macOS Catalina 10.15.7 (19H2)`

Hengyu:
Or we can change the package type from `.dynamic` to default in Swift Package Manger.

E.g., 
```
.library(
            name: "Datadog",
            targets: ["Datadog"]),
```

By doing this, the Datadog package is embedded in the project main target as a static library. And we can choose to only embed it when the build target is iOS. The Xcode compiler itself won't compile it when the target is Mac.

<img width="320" alt="1" src="https://user-images.githubusercontent.com/4032674/95773670-84d2ef00-0cf1-11eb-89d0-a790fff48860.png">

Hengyu:
The pros and cons between the first and second solutions:

#### 1st solution (add macros)

Pros: Make the Datadog can also be used in `macCatalyst` and `tvOS`

Cons: Code is not clean

#### 2nd solution (change type to static)

Pros: Minimal changes, clean code

[...]

PR #277, opened 2020-10-13
RUMM-816 SDK is made compile for Catalyst

buranmert:
## ⚠️ Important ⚠️

With this PR the SDK compiles for Catalyst, yet we do not guarantee its correctness at runtime at the moment
Official Catalyst support is in our roadmap and we will announce it once it's done

### What and why?

As #274 points out, our users may need to compile their app target for Mac Catalyst too
Since our SDK doesn't compile for Catalyst, either our users need to duplicate their app target for Catalyst without linking Datadog SDK or add the framework conditionally via linker flags (which may be tricky or impossible with Swift Package Manager though)

### How?

SDK codebase doesn't depend on iOS-only frameworks, except `CoreTelephony`
We used `#if targetEnvironment(macCatalyst)` to disable `CoreTelephony`-dependent code from compiling

### Review checklist

### DIFF
--- a/Sources/Datadog/Datadog.swift
+++ b/Sources/Datadog/Datadog.swift
@@ -47,14 +47,14 @@ public class Datadog {
     ///   - appContext: context passing information about the app.
     ///   - configuration: the SDK configuration obtained using `Datadog.Configuration.builderUsing(clientToken:)`.
     public static func initialize(appContext: AppContext, configuration: Configuration) {
+        // TODO: RUMM-511 remove this warning
+        #if targetEnvironment(macCatalyst)
+        consolePrint("⚠️ Catalyst is not officially supported by Datadog SDK: some features may NOT be functional!")
+        #endif
         do {
             try initializeOrThrow(
                 configuration: try FeaturesConfiguration(configuration: configuration, appContext: appContext)
             )
-            // TODO: RUMM-511 remove this warning
-            #if targetEnvironment(macCatalyst)
-            userLogger.warn("⚠️ Catalyst is not officially supported by Datadog SDK: use at your own risk!")
-            #endif
         } catch {
             consolePrint("\(error)")
         }
--- a/Sources/Datadog/Datadog.swift
+++ b/Sources/Datadog/Datadog.swift
@@ -51,6 +51,10 @@ public class Datadog {
             try initializeOrThrow(
                 configuration: try FeaturesConfiguration(configuration: configuration, appContext: appContext)
             )
+            // TODO: RUMM-511 remove this warning
+            #if targetEnvironment(macCatalyst)
+            userLogger.warn("⚠️ Catalyst is not officially supported by Datadog SDK: use at your own risk!")
+            #endif
         } catch {
             consolePrint("\(error)")
         }
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -137,16 +137,24 @@ workflows:
             #!/usr/bin/env bash
             set -e
             make test-spm ci=true
+    - xcodebuild:
+        title: Build SPMProject for tests - Catalyst
+        inputs:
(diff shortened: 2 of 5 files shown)


Sources/Datadog/Datadog.swift
```diff
@@ -51,6 +51,10 @@ public class Datadog {
             try initializeOrThrow(
                 configuration: try FeaturesConfiguration(configuration: configuration, appContext: appContext)
             )
+            // TODO: RUMM-511 remove this warning
+            #if targetEnvironment(macCatalyst)
+            userLogger.warn("⚠️ Catalyst is not officially supported by Datadog SDK: use at your own risk!")
```
ncreated, line 56: 1/. We can't use `userLogger` during the SDK initialization. This logger is only functional if user sets `Datadog.verbosity`, which should happen after `Datadog.initialize()` returns. Instead, we should use `consolePrint()` directly.

2/ I think _"some features will not work correctly"_ sounds better than _"use at your own risk!"_ which suggests damage 🙂, WDYT?

bitrise.yml
```diff
@@ -137,16 +137,24 @@ workflows:
...
         inputs:
         - scheme: SPMProject
+        - simulator_device: iPhone 11
         - is_clean_build: 'yes'
         - cache_level: none
         - project_path: "$BITRISE_SOURCE_DIR/dependency-manager-tests/spm/SPMProject.xcodeproj"
         - xcpretty_test_options: --color --report html --output "${BITRISE_DEPLOY_DIR}/SPMProject-tests.html"
     - script:
-        title: Test Carthage compatibility
+        title: Test Carthage compatibility (Carthage doesn't support Catalyst!)
```
ncreated, line 157: The `title` appears in CI logs, so it's IMO better to keep it as code comment:
```
# Carthage doesn't support Catalyst, so we don't test CTProject for `variant=Mac Catalyst`
- script:
   title: Test Carthage compatibility
   ...
```
WDYT?

ncreated (changes requested):
Two comments, looks good 👍.

Sources/Datadog/Datadog.swift
buranmert: i used `userLogger` so that the user could suppress warning (@iGranDav 's idea) by changing `verbosityLevel` but you are right 👍 
can we let the user do so with `consolePrint`?

changing to "Some features may not be functional"? 
eg: carrier info is disabled, some other features may not work

ncreated: > can we let the user do so with `consolePrint`?

No, and it's by design. Depending on the message severity we can either pick:
* `userLogger` - for "warning" purpose which can be suppressed,
* `consolePrint` - for "error" purpose which cannot be suppressed.

buranmert: `userLogger` seems like a better fit as this is a warning though 🤔 
how about that?
```swift
/// Default is `nil` in supported platforms and `.warn` in non-supported platforms
public static var verbosityLevel: LogLevel? = {
  #if target(Catalyst) return .warn
  #else return nil
  #endif
}
```
otherwise i'll leave it without any suppression option.?

ncreated: Current flow is plain simple: `userLogger` considers the `.verbosityLevel`, `consolePrint()` does not. I see no reason to make it more complex in the implementation and tests. Even if it's a "warning", not "error" (debatable), this PR is only a quick win o make the SDK compile on the platform which we do not support.

buranmert: that popped up in the Slack discussion (i've just cc'ed you)

okay 👌 changing to `consolePrint()` ✅

ncreated (approved):
👌 👌 👌

PR #281, opened 2020-10-14
RUMM-812 DDURLSessionDelegate is made subclassable

buranmert:
### What and why?

`DDURLSessionDelegate` is supposed to be subclassable
The users may need to change its behavior to their own needs
Yet, it was not possible to subclass it practically due to lack of public designated initilalizer

It was possible to create a new class `class MyDelegate: DDURLSessionDelegate { ... }`
It was not possible to create an instance of it:
```swift
delegate = MyDelegate() // cannot be constructed because it has no accessible initializers
```
It was also not possible to override `init` methods because
1. superclass does not have a public designated init
2. overridden init needs to call super's designated init at some point

### How?

`DDURLSessionDelegate` now has a `public` (also overrideable) designated initializer

### Review checklist

### DIFF
--- a/Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
+++ b/Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
@@ -11,11 +11,8 @@ open class DDURLSessionDelegate: NSObject, URLSessionTaskDelegate {
     private let interceptor: URLSessionInterceptorType?
 
     @objc
-    override public convenience init() {
-        self.init(interceptor: URLSessionAutoInstrumentation.instance?.interceptor)
-    }
-
-    internal init(interceptor: URLSessionInterceptorType?) {
+    override public init() {
+        self.interceptor = URLSessionAutoInstrumentation.instance?.interceptor
         if interceptor == nil {
             let error = ProgrammerError(
                 description: """
@@ -25,7 +22,19 @@ open class DDURLSessionDelegate: NSObject, URLSessionTaskDelegate {
             )
             consolePrint("\(error)")
         }
+    }
+
+    internal init(interceptor: URLSessionInterceptorType?) {
         self.interceptor = interceptor
+        if interceptor == nil {
+            let error = ProgrammerError(
+                description: """
+                To use `DDURLSessionDelegate` you must specify first party hosts in `Datadog.Configuration` -
+                use `track(firstPartyHosts:)` to define which requests should be tracked.
+                """
+            )
+            consolePrint("\(error)")
+        }
     }
 
     open func urlSession(_ session: URLSession, task: URLSessionTask, didFinishCollecting metrics: URLSessionTaskMetrics) {
--- a/Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
+++ b/Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
@@ -8,24 +8,11 @@ import Foundation
 
 @objc
 open class DDURLSessionDelegate: NSObject, URLSessionTaskDelegate {
-    private let interceptor: URLSessionInterceptorType?
+    var interceptor: URLSessionInterceptorType?
 
     @objc
     override public init() {
-        self.interceptor = URLSessionAutoInstrumentation.instance?.interceptor
-        if interceptor == nil {
-            let error = ProgrammerError(
-                description: """
-                To use `DDURLSessionDelegate` you must specify first party hosts in `Datadog.Configuration` -
-                use `track(firstPartyHosts:)` to define which requests should be tracked.
-                """
-            )
-            consolePrint("\(error)")
-        }
-    }
-
-    internal init(interceptor: URLSessionInterceptorType?) {
-        self.interceptor = interceptor
+        interceptor = URLSessionAutoInstrumentation.instance?.interceptor
         if interceptor == nil {
             let error = ProgrammerError(
                 description: """
--- a/Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegateAsSuperclassTests.swift
+++ b/Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegateAsSuperclassTests.swift
@@ -1,0 +1,26 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import XCTest
+import Datadog
+
+internal final class SubDDURLSessionDelegate: DDURLSessionDelegate {
+    let someProp: String
+    override init() {
+        someProp = "someProp"
+        super.init()
+    }
+    override func urlSession(_ session: URLSession, task: URLSessionTask, didCompleteWithError error: Error?) {
+        super.urlSession(session, task: task, didCompleteWithError: error)
+    }
+}
+
+class DDURLSessionDelegateAsSuperclassTests: XCTestCase {
+    func testSubclassability() {
+        // Success: tests compile, failure: compilation error
+        _ = SubDDURLSessionDelegate()
+    }
+}
--- a/Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegateAsSuperclassTests.swift
+++ b/Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegateAsSuperclassTests.swift
@@ -8,9 +8,9 @@ import XCTest
 import Datadog
 
 internal final class SubDDURLSessionDelegate: DDURLSessionDelegate {
-    let someProp: String
+    let property: String
     override init() {
-        someProp = "someProp"
+        property = "someProp"
         super.init()
     }
     override func urlSession(_ session: URLSession, task: URLSessionTask, didCompleteWithError error: Error?) {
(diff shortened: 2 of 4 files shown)


Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
```diff
@@ -11,11 +11,21 @@ open class DDURLSessionDelegate: NSObject, URLSessionTaskDelegate {
...
+                description: """
+                To use `DDURLSessionDelegate` you must specify first party hosts in `Datadog.Configuration` -
+                use `track(firstPartyHosts:)` to define which requests should be tracked.
+                """
+            )
+            consolePrint("\(error)")
+        }
     }
 
     internal init(interceptor: URLSessionInterceptorType?) {
```
ncreated, line 27: We don't need this initializer now as it is only used in tests. We should remove it and mock the global dependency (`URLSessionAutoInstrumentation.instance?.interceptor`) in the same way as we do it for other features:
```swift
func test...() {
   URLSessionAutoInstrumentation.instance = .mockWith(interceptor: interceptor)
   defer { URLSessionAutoInstrumentation.instance = nil }

   let delegate = DDURLSessionDelegate() // it will use the `interceptor`

   // ...
}
```

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegateAsSuperclassTests.swift
```diff
@@ -0,0 +1,26 @@
...
+    override init() {
+        someProp = "someProp"
+        super.init()
+    }
+    override func urlSession(_ session: URLSession, task: URLSessionTask, didCompleteWithError error: Error?) {
+        super.urlSession(session, task: task, didCompleteWithError: error)
+    }
+}
+
+class DDURLSessionDelegateAsSuperclassTests: XCTestCase {
```
ncreated, line 21: The `DDURLSessionDelegateTests.swift` already defines the region for this kind of tests:
```swift
// MARK: - Usage
```
we should move it there. We test usage this way in other places too, see: `LoggerTests`, `TracerTests`, `RUMMonitorTests`.

Also, for simplicity, the subclass can be defined in test function scope, to not pollute the namespace for only single usage.

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegateAsSuperclassTests.swift
```diff
@@ -0,0 +1,26 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import XCTest
+import Datadog
+
+internal final class SubDDURLSessionDelegate: DDURLSessionDelegate {
+    let someProp: String
```
ncreated, line 11:
```suggestion
    let property: String
```
or
```suggestion
    let someProperty: String
```

ncreated (changes requested):

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegateAsSuperclassTests.swift
buranmert: this must be in a separate file in which `import Datadog` without `@testable`
otherwise it would pass even before the fix as designated initializer was already `internal`

yes, the subclass was supposed to be a `private` or nested class 👌

Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
buranmert: but we don't need a `URLSessionAutoInstrumentation` instance to create a delegate neither

i think
1. we can make `DDURLSessionDelegate.interceptor` an `internal var` so that we can change internally in unit tests
2. or we can keep duplicate init's (ideally we would have those methods anyway but only due to Swift initialization rules their implementations are not ideal)

wdyt?

ncreated: > but we don't need a `URLSessionAutoInstrumentation` instance to create a delegate neither

Sorry, I don't understand 🤔. Yes, we don't need it, but how does it relate to the test I proposed?

> we can make `DDURLSessionDelegate.interceptor` an `internal var` so that we can change internally in unit tests

Yes, this could be a way to go as well. If it's `internal`, we can also test that `interceptor` is set correctly when instrumentation is enabled:
```swift
func testGivenURLSessionAutoInstrumentationEnabled_whenCreatingDelegate_itUsesInterceptor() {
   // Given
   // ...

   // When
   let delegate = DDURLSessionDelegate()
   
   // Then
   XCTAssertTrue(delegate.interceptor === URLSessionAutoInstrumentation.instance.interceptor)
}
```
Right now, this explicit test was not possible.

> or we can keep duplicate init's (ideally we would have those methods anyway but only due to Swift initialization rules their implementations are not ideal)

We should avoid doing this if necessary. We only do it when other ways are not possible (AFAIK there are only 2, carefully controlled places like this).

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegateAsSuperclassTests.swift
ncreated: If we remove the unused initializer, as suggested https://github.com/DataDog/dd-sdk-ios/pull/281#discussion_r504823965, this may go into `DDURLSessionDelegateTests`, no?

buranmert: i don't think so becuase the case definition is "can anyone subclass this class without @testable keyword?"

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegateTests.swift
```diff
@@ -176,4 +181,29 @@ class DDURLSessionDelegateTests: XCTestCase {
...
+        Datadog.initialize(
+            appContext: .mockAny(),
+            configuration: Datadog.Configuration.builderUsing(
+                clientToken: "abc-123",
+                environment: "tests"
+            )
+            .track(firstPartyHosts: ["example.com"])
+            .build()
+        )
+        XCTAssertNotNil(URLSessionAutoInstrumentation.instance)
```
ncreated, line 196: We don't need to initialize the entire SDK. This is enough:
```swift
// Given

URLSessionAutoInstrumentation.instance = URLSessionAutoInstrumentation(
   configuration: .mockAny(),
   dateProvider: SystemDateProvider()
)
defer { URLSessionAutoInstrumentation.instance = nil }
```

When initializing the full SDK like this we enable many unnecessary things:
* Logs / Traces / RUM upload workers (they start running),
* Logs / Traces / RUM storage `DispatchQueues`,
* `URLSession` swizzler is applied,

which will slow down the tests execution and may lead to more flakiness.

It's already ensured by other tests (`DatadogTests`) that the initialization leads to configuring `URLSessionAutoinstrumentation` stack properly.

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegateAsSuperclassTests.swift
ncreated: I don't see the difference. If the initializer is not defined, how can it be called if `@testable` is used 🤷.

ncreated (changes requested):

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegateTests.swift
```diff
@@ -148,10 +152,11 @@ class DDURLSessionDelegateTests: XCTestCase {
 
     // MARK: - Usage errors
 
-    func testGivenFirstPartyHostsNotSet_whenInitializingDDURLSessionDelegate_itPrintsError() throws {
+    func testGivenAutoInstrumentationInstanceIsNil_whenInitializingDDURLSessionDelegate_itPrintsError() throws {
         let printFunction = PrintFunctionMock()
+        let prevConsolePrint = consolePrint
```
ncreated, line 157:
```suggestion
        let previousConsolePrint = consolePrint
```

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegateTests.swift
```diff
@@ -148,10 +152,11 @@ class DDURLSessionDelegateTests: XCTestCase {
...
         consolePrint = printFunction.print
-        defer { consolePrint = { print($0) } }
+        defer { consolePrint = prevConsolePrint }
```
ncreated, line 159:
```suggestion
        defer { consolePrint = previousConsolePrint }
```

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegateAsSuperclassTests.swift
buranmert: i probably didn't understand 🤔 

if one day someone makes some changes and this type becomes unusable in the host app, we hope that this test case will catch that.
if this case had `@testable import Datadog` then we would miss it and this is what just happened: 
i wrote this case in `DDURLSessionDelegateTests` *before* the fix and it passed (but it shouldn't)

ncreated: This should be the part of integration test then - testability is ultimately disabled in their configuration. Anyway, good to go right now, we can improve it later.

ncreated (approved):

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegateAsSuperclassTests.swift
buranmert: but this is a test on the unit, basically it tests its public interface only. by definition this is not an integration test to my understanding

PR #282, opened 2020-10-14
RUMM-812 Shopist: RUM Autoinstrumentation

buranmert:
### What and why?

We test RUM autoinstrumentation in Shopist app

### How?

What is autoinstrumented:

1. network requests
2. `tap` user actions
3. views

What is to be autoinstrumented later

1. custom and swipe user actions

batch-results

### Review checklist

### DIFF
--- a/Shopist/Shopist/Scenes/CheckoutViewController.swift
+++ b/Shopist/Shopist/Scenes/CheckoutViewController.swift
@@ -44,7 +44,6 @@ internal final class CheckoutViewController: UITableViewController {
         }
     }
     private static let cellIdentifier = "cell"
-    private let api = API()
     private var hasOngoingComputation = false
     private var viewDidAppearDate: Date?
     private var totalAmount: Float = 0.0
@@ -67,7 +66,6 @@ internal final class CheckoutViewController: UITableViewController {
 
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
-        Global.rum.startView(viewController: self)
         if !hasOngoingComputation {
             hasOngoingComputation = true
             cart.generateBreakdown {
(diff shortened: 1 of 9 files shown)


buranmert:
Resources volume was low due to caching
Now fixed, retrying... ⏳

Shopist/Shopist/Scenes/CheckoutViewController.swift
```diff
@@ -67,7 +66,6 @@ internal final class CheckoutViewController: UITableViewController {
 
     override func viewWillAppear(_ animated: Bool) {
```
ncreated, line 67: We should use `viewDidAppear()`, to get the correct resources assignment.

Shopist/Shopist/Scenes/ListViewController.swift
```diff
@@ -27,7 +26,6 @@ internal class ListViewController: UICollectionViewController {
 
     override func viewWillAppear(_ animated: Bool) {
```
ncreated, line 27: Same: we should use `viewDidAppear()`, to get the correct resources assignment

Shopist/Shopist/Scenes/ProductViewController.swift
```diff
@@ -34,15 +34,9 @@ internal class ProductViewController: UIViewController {
 
     override func viewWillAppear(_ animated: Bool) {
```
ncreated, line 35: Same: we should use `viewDidAppear()`, to get the correct resources assignment

ncreated (changes requested):
Great that we get more with less code 👍. The reason of resources volume drop may be because of starting requests in `viewWillAppear()`. Let's switch to `viewDidAppear()` as pointed out and it should be corrected. Once we improve this in SDK, we can switch to "will" again.

buranmert:
after-v3

now looks good ✅

ncreated (approved):

PR #286, opened 2020-10-19
RUMM-742 Make integration tests not depend on number of requests to Python server

ncreated:
### What and why?

📦 This PR updates the integration tests to not depend on the number of requests send to the Python server. This dependency was a source of flakiness in some CI jobs.

### How?

For all matchers: `LogMatcher`, `SpanMatcher`, `RUMSessionMatcher`, I updated the integration tests from this model:
```swift
let requests = try server.pullRequests(count: X, timeout: ...) // wait and return X requests
let matchers = try Matchers.from(request: requests) // transform requests into matchers
XCTAssertEqual(matchers.count, Y) // assert expected number of matchers
```
to this one:
```swift
let requests = try server.pullRequests(timeout: ...) { requests in
   Matchers.from(request: requests).count == Y // wait until receiving expected number of matchers
}
let matchers = try Matchers.from(request: requests) // here, we get expected number of matchers
```

This required adding improvements to the `HTTPServerMock` - now it returns any kind of requests (vs only `POST` before) as soon as the `condition: ([Request]) -> Bool` evaluates to `true`.

To minimalize number of connections to `localhost`, I also updated the Python server, to return details of all requests from a single `GET /inspect` endpoint (vs `GET/inspect` + `GET/inspect/<rid>/body` + `GET/inspect/<rid>/headers` before).

Lasts, I simplified integration tests by reusing common assertion code for Logging and installing `RUMSessionMatcher` in the oldest tests.

### Review checklist


instrumented-tests/http-server-mock/Sources/HTTPServerMock/ServerSession.swift
```diff
@@ -35,50 +21,38 @@ public class ServerSession {
...
+        var conditionMet = false
         repeat {
-            currentRequests = try server
-                .getRecordedPOSTRequestsInfo()
-                .filter { requestInfo in requestInfo.path.contains(sessionIdentifier) }
+            pulledRequests = try getRecordedRequests()
+            conditionMet = try condition(pulledRequests)
             Thread.sleep(forTimeInterval: 0.2)
-        } while !timeoutTimer.isCancelled && currentRequests.count < count
+        } while !timeoutTimer.isCancelled && !conditionMet
```
buranmert, line 47: this can simplified: `while timeoutTimer.isCancelled || conditionMet`

instrumented-tests/http-server-mock/Tests/HTTPServerMockTests/HTTPServerMockTests.swift
```diff
@@ -77,7 +77,9 @@ final class HTTPServerMockTests: XCTestCase {
             )
         }
         let timeoutTime: TimeInterval = 2
-        let recordedRequests = try session.pullRecordedPOSTRequests(count: 2, timeout: timeoutTime)
+        let recordedRequests = try session.pullRecordedRequests(timeout: timeoutTime) { requests in
+            requests.count == 2
```
buranmert, line 81: there are many `count == N` in this PR, we can change them to `count >= N` to guard against any unexpected problem (eg: threading, double-counting, etc.)

instrumented-tests/http-server-mock/Tests/HTTPServerMockTests/HTTPServerMockTests.swift
```diff
@@ -86,7 +88,7 @@ final class HTTPServerMockTests: XCTestCase {
         XCTAssertEqual(recordedRequests[1].httpBody, "2nd request body".data(using: .utf8)!)
     }
 
-    func testWhenPullingRecordedPOSTRequestExceedsTimeout_itThrownsAnError() throws {
+    func testWhenPullingRecordedRequestExceedsTimeout_itThrownsAnError() throws {
```
buranmert, line 91: typo
```suggestion
    func testWhenPullingRecordedRequestExceedsTimeout_itThrowsAnError() throws {
```

instrumented-tests/http-server-mock/python/start_mock_server.py
```diff
@@ -77,34 +69,13 @@ def __GET_inspect(self, parameters):
         inspection_info = []
         for request in history.all_requests():
             inspection_info.append({
-                "request_method": request.http_method,
-                "request_path": request.path,
-                "body_inspection_path": "/inspect/{request_id}/body".format( request_id = request.id ),
-                "headers_inspection_path": "/inspect/{request_id}/headers".format( request_id = request.id )
+                "method": request.http_method,
+                "path": request.path,
+                "body": base64.b64encode(request.http_body), # use Base64 to not corrupt the JSON
```
buranmert, line 74: > use Base64 to not corrupt the JSON

what does it mean? does ascii corrupt it for example?

buranmert (approved):
i left a few comments but LGTM

instrumented-tests/http-server-mock/python/start_mock_server.py
ncreated: In this case, we send JSON within a JSON:
```json
{
   "path": "/foo/bar",
   "body": <json>
}
```
By default, Python escapes the `"body"` json, so it becomes: `"\{\"foo\":\"bar"\}\"`. I found it much easier and less error-prone to just encode it as a base-64 ASCII (easy in Swift) instead of un-escaping the `.utf8` JSON `String` data.

instrumented-tests/http-server-mock/Sources/HTTPServerMock/ServerSession.swift
ncreated: Actually, it becomes `!(timeoutTimer.isCancelled || conditionMet)` but it's still better, done 👍.

instrumented-tests/http-server-mock/Tests/HTTPServerMockTests/HTTPServerMockTests.swift
ncreated: Indeed - updated for Logging and Tracing 👍. In RUM I'd rather not assert `session.viewVisits` by `>=`. It is critical to ensure that we get exactly the expected number of RUM Views, not more.

Also, it's not that convenient for RUM, as we have optional chaining in: `RUMSession.from(requests)?.viewVisits` so `==` works out of the box, but `>=` would require some default.

PR #288, opened 2020-10-20
RUMM-694 Add Trace information to RUM Resources

ncreated:
### What and why?

🚚 This PR adds `trace_id`, `span_id` and `id` information to RUM Resource events. Soon (`RUMM-696`) the `Span` won't be sent from the mobile SDK, but instead the RUM backend will generate it.

### How?

Recap: session interceptor manages the `URLRequest` mutation and notifies both, tracing and RUM handlers with different states of the `TaskInterception`.

Now,  the `SpanContext` is attached to `TaskInterception`. Each handler reads the `SpanContext` and processes it accordingly (tracing handler sends the `Span`, RUM handler encodes `trace_id` and `span_id` in RUM Resource event).

This required updating data models to recent `rum-events-format`.

### Review checklist

### DIFF
--- a/Sources/Datadog/RUM/DataModels/RUMDataModels.swift
+++ b/Sources/Datadog/RUM/DataModels/RUMDataModels.swift
@@ -20,6 +20,8 @@ internal struct RUMAction: RUMDataModel {
     let date: Int64
     /// Application properties
     let application: RUMApplication
+    /// The service name for this application
+    let service: String?
     /// Session properties
     let session: RUMSession
     /// View properties
@@ -38,6 +40,7 @@ internal struct RUMAction: RUMDataModel {
     enum CodingKeys: String, CodingKey {
         case date = "date"
         case application = "application"
+        case service = "service"
         case session = "session"
         case view = "view"
         case usr = "usr"
@@ -295,6 +298,8 @@ internal struct RUMError: RUMDataModel {
     let date: Int64
     /// Application properties
     let application: RUMApplication
+    /// The service name for this application
+    let service: String?
     /// Session properties
     let session: RUMSession
     /// View properties
@@ -315,6 +320,7 @@ internal struct RUMError: RUMDataModel {
     enum CodingKeys: String, CodingKey {
         case date = "date"
         case application = "application"
+        case service = "service"
         case session = "session"
         case view = "view"
         case usr = "usr"
@@ -410,6 +416,8 @@ internal struct RUMLongTask: RUMDataModel {
     let date: Int64
     /// Application properties
     let application: RUMApplication
+    /// The service name for this application
+    let service: String?
     /// Session properties
     let session: RUMSession
     /// View properties
@@ -430,6 +438,7 @@ internal struct RUMLongTask: RUMDataModel {
     enum CodingKeys: String, CodingKey {
         case date = "date"
         case application = "application"
+        case service = "service"
         case session = "session"
         case view = "view"
         case usr = "usr"
@@ -475,6 +484,8 @@ internal struct RUMResource: RUMDataModel {
     let date: Int64
     /// Application properties
     let application: RUMApplication
+    /// The service name for this application
+    let service: String?
     /// Session properties
     let session: RUMSession
     /// View properties
@@ -484,7 +495,7 @@ internal struct RUMResource: RUMDataModel {
     /// Device connectivity properties
     let connectivity: RUMConnectivity?
     /// Internal properties
-    let dd: RUMActionDD
+    let dd: RUMResourceDD
     /// RUM event type
     let type = "resource"
     /// Resource properties
@@ -495,6 +506,7 @@ internal struct RUMResource: RUMDataModel {
     enum CodingKeys: String, CodingKey {
         case date = "date"
         case application = "application"
+        case service = "service"
         case session = "session"
         case view = "view"
         case usr = "usr"
@@ -518,10 +530,30 @@ internal struct RUMResourceAction: Codable {
     }
 }
 
+// MARK: - RUMResourceDD
+
+/// Internal properties
+internal struct RUMResourceDD: Codable {
+    /// Version of the RUM event format
+    let formatVersion = 2
+    /// span identifier in decimal format
+    let spanID: String?
+    /// trace identifier in decimal format
+    let traceID: String?
+
+    enum CodingKeys: String, CodingKey {
+        case formatVersion = "format_version"
+        case spanID = "span_id"
+        case traceID = "trace_id"
+    }
+}
+
 // MARK: - RUMResourceResource
 
 /// Resource properties
 internal struct RUMResourceResource: Codable {
+    /// UUID of the resource
+    let id: String?
     /// Resource type
     let type: RUMResourceType
     /// HTTP method of the resource
@@ -548,6 +580,7 @@ internal struct RUMResourceResource: Codable {
     let download: RUMDownload?
 
     enum CodingKeys: String, CodingKey {
+        case id = "id"
         case type = "type"
         case method = "method"
         case url = "url"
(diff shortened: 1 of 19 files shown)


Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScopeTests.swift
```diff
@@ -36,6 +37,26 @@ class RUMResourceScopeTests: XCTestCase {
...
+                url: .mockAny(),
+                httpMethod: .mockAny(),
+                spanContext: nil
+            )
+        }
+
+        let scope1 = createScope()
+        let scope2 = createScope()
+
+        XCTAssertNotEqual(scope1.resourceUUID, scope2.resourceUUID)
```
buranmert, line 57: i think we shouldn't be testing this, we need `id`s in generated `RUMResourceResource` to be different, no?
`resourceUUID` can be hidden as `private` since it is an implementation detail; the actual identifier of this type is `resourceName`

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMViewScopeTests.swift
```diff
@@ -53,6 +53,24 @@ class RUMViewScopeTests: XCTestCase {
...
+                uri: "UIViewController",
+                attributes: [:],
+                startTime: .mockAny()
+            )
+        }
+
+        let scope1 = createScope()
+        let scope2 = createScope()
+
+        XCTAssertNotEqual(scope1.viewUUID, scope2.viewUUID)
```
buranmert, line 71: the same comment i guess

Sources/Datadog/URLSessionAutoInstrumentation/Interception/URLSessionInterceptor.swift
```diff
@@ -94,6 +94,11 @@ internal class URLSessionInterceptor: URLSessionInterceptorType {
             let interception = TaskInterception(request: request)
             self.interceptionByTask[task] = interception
 
+            if let tracer = Global.sharedTracer as? Tracer,
+               let spanContext = self.extractSpanContext(from: request, using: tracer) {
```
buranmert, line 98: in `HTTPHeadersReader.extract()` we have 
```swift
return DDSpanContext(..., parentSpanID: nil, ...)
```
shouldn't we retrieve `activeSpan` here? (this might be out of this PR's scope though)

xgouchet (approved):
LGTM

Sources/Datadog/RUM/DataModels/RUMDataModels.swift
```diff
@@ -844,4 +882,4 @@ internal struct RUMViewResource: Codable {
         case count = "count"
     }
 }
-// b6eb2bf04511aa22db9b73ba0d195ee9c365bee7
+// c84571fec8cbb47cf0fb30b0f476a0d4d09d7262
```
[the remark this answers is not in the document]
buranmert: this is the commit hash from which those models are generated.
verification phase picks that certain commit, generates the models and compare.
how do you do verification in Android?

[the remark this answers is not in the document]
xgouchet: Oh ok. On our side the Model is generated from a snapshot of the Json Schema repository instead. We keep a copy of the json schema definitions and the generation is performed at compile time.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
```diff
@@ -95,8 +104,12 @@ internal class RUMResourceScope: RUMScope {
             ),
             usr: dependencies.userInfoProvider.current,
             connectivity: dependencies.connectivityInfoProvider.current,
-            dd: .init(),
+            dd: .init(
+                spanID: spanContext.flatMap { String($0.spanID) },
+                traceID: spanContext.flatMap { String($0.traceID) }
+            ),
```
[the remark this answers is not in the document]
ncreated: Indeed, the `TracingUUID` → `Int64` → `String` mapping was unnecessary. I updated it to do `TracingUUID` → `String` directly in `URLSessionRUMResourcesHandler`.

> as a bridge, RUMSpanContext can also provide `init(from: DDSpanContext)` and `var toResourceDD { ... }` helpers (otherwise obviously every caller needs to know how to convert)

`init(from: DDSpanContext)` - I haven't done it intentionally, to not introduce more coupling between Tracing and RUM features. Now, the coupling happens only in `URLSessionRUMResourcesHandler` (for mapping `Tracing.DDSpanContext` to `RUM.RUMSpanContext`). It's simple and easy embrace it now and refactor if needed - spreading it more into RUM codebase will make both harder.

`var toRUMDataFormat` - I've been thinking of it, but it doesn't fit in this particular case. What we have is `RUMSpanContext`, what we'd need from `.toRUMDataFormat` is `RUMResourceDD` which describes more than just span context. Let's wait and see how it evolves, then we can make this small refactor.

> by the way, i can't find occurences of any type with `.init` usage with Find selected text/symbol in Workspace; how do you search?

⌘ + ⇧ + Click on the symbol, and here you go:

Screenshot 2020-10-22 at 13 18 11

Sources/Datadog/URLSessionAutoInstrumentation/Interception/URLSessionInterceptor.swift
```diff
@@ -165,4 +170,9 @@ internal class URLSessionInterceptor: URLSessionInterceptorType {
         }
         return newRequest
     }
+
+    private func extractSpanContext(from request: URLRequest, using tracer: Tracer) -> DDSpanContext? {
+        let reader = HTTPHeadersReader(httpHeaderFields: request.allHTTPHeaderFields ?? [:])
+        return tracer.extract(reader: reader) as? DDSpanContext
```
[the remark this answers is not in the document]
ncreated: Indeed, done 👍

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScopeTests.swift
ncreated: Hmm, not sure if I understand. Each Resource should have unique `UUID`, why we shouldn't test it?

Sources/Datadog/URLSessionAutoInstrumentation/Interception/URLSessionInterceptor.swift
ncreated: It's a question on the requirement. Do we want auto-instrumented spans to be linked with / continue user-created spans in the app? It was never the goal, so it's not implemented.

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScopeTests.swift
buranmert: yes, i mean each `Resource` (the output of `ResourceScope`) must have unique identifiers. here we are testing identifiers of `ResourceScope` only. 
does that make sense?

buranmert (approved):
there is still a question above regarding testing unique identifiers above
but the rest LGTM

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScopeTests.swift
ncreated: It was testing uniqueness of the `scope.id`, and later that `event.id == scope.id`. Anyway, we can easily make assertions on the actual events sent, so I updated it 👍. Also, added more validation logic in `RUMSessionMatcher`, so we have this assertion for free in all integration tests.

buranmert (approved):

PR #290, opened 2020-10-26
RUMM-851 Fix: Tests flakiness

buranmert:
### What and why?

`URLSessionSwizzlerTests` started giving non-deterministic results.

Problems:
1. According to logs, some expectations are fulfilled multiple times.
2. `interceptor.tasksCompleted.count` is 3 during assertion (it must be 4) but a few lines later it becomes 4

Reasons:
1. In another, unrelated place SDK swizzles a method without unswizzling it later. Then, when this particular test case swizzles it becomes double-swizzled.
2. `InterceptorMock` fulfills its expectation before adding new item to its storage

### How?

Reasons are fixed

### Review checklist


ncreated (approved):
👌 👌 👌

PR #291, opened 2020-10-27
RUMM-726: Renaming RUM models

buranmert:
### What and why?

We had `RUMViewFromPredicate` type in our public API
This was not the ideal name for this type, it was supposed to be `RUMView` yet this name was taken by a `RUMDataModel`
We make this change in order to improve our public API

### How?

We rename related types
Prefix in `RUMDataModels.swift`: `RUM` -> `RUMData` (eg: `RUMView` -> `RUMDataView`)
Type name: `RUMViewFromPredicate` -> `RUMView`

### Review checklist

# NOTE

CI will probably fail now but this will be fixed by https://github.com/DataDog/dd-sdk-ios/pull/290

### DIFF
--- a/Datadog/Example/TestScenarios.swift
+++ b/Datadog/Example/TestScenarios.swift
@@ -177,7 +177,7 @@ struct RUMNavigationControllerScenario: TestScenario {
     static let storyboardName = "RUMNavigationControllerScenario"
 
     private class Predicate: UIKitRUMViewsPredicate {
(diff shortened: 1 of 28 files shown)


ncreated (changes requested):
Looks good, but there are other places to rename too - which is why the CI is red.

cswatt (commented):
approved for docs 👍

ncreated (approved):

PR #292, opened 2020-10-27
RUMM-719 Collect advanced temporal metrics for RUM Resources

ncreated:
### What and why?

📦 This PR adds collection of advanced temporal metrics for auto-instrumented RUM Resources. Example:

Screenshot 2020-10-26 at 15 13 55

### How?

More values are captured from `URLSessionTaskMetrics` object. If task execution required more than one transaction, the last transaction is used to compute detailed values and all leading transactions are described as "redirection" phase, e.g. in such case:
```bash
→ Transaction 1: 301 redirect
   → Transaction 2: 301 redirect
      → Transaction 3: 200 OK
```
the "redirection" phase spans from T1 start to T2 end and T3 is used to track `dns`, `connect`, ...

### Review checklist

### DIFF
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
@@ -116,17 +116,42 @@ internal class RUMResourceScope: RUMScope {
                 statusCode: command.httpStatusCode?.toInt64,
                 duration: resourceDuration.toInt64Nanoseconds,
                 size: size ?? 0,
-                redirect: nil,
-                dns: resourceMetrics?.dns.flatMap { dns in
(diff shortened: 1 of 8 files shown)


buranmert (approved):
looks good 👍 
there are also `isProxyConnection` and `isExpensive` properties in task metrics, maybe we can somehow include them to the RUM event (via attributes?), wdyt?

ncreated:
> there are also `isProxyConnection` and `isExpensive` properties in task metrics, maybe we can somehow include them to the RUM event (via attributes?), wdyt?

@buranmert I think it's a good idea for the improvement in `rum-events-format`. I wouldn't add it as generic attribute, as we don't do it anywhere in rum (everything goes through `rum-events-format`). We should discuss it with the entire team👌.

PR #294, opened 2020-10-28
RUMM-851 Missing expectation added to flaky test case

buranmert:
### What and why?

`testGivenSuccessfulTask_whenUsingSwizzledAPIs_itPassessAllValuesToTheInterceptor` case wasn't waiting for 4 task completions.
As the test continues without that expectation, `interceptor.tasksCompleted.count == 4` assertion was failing; although `tasksCompleted[3]` was not causing index out of range right after `count` assertion because it was just a matter of time for tasks to be completed.

### How?

Missing expectation added

### Review checklist


ncreated (approved):
👌

PR #299, opened 2020-11-03
RUMM-633 Simplify RUM Resources manual instrumentation API

ncreated:
### What and why?

📦 This PR simplifies the API for manual tracking of RUM Resources.

### How?

In the RUM auto instrumentation PRs we added few helpers to compute `RUMHTTPMethod` and `RUMResourceKind` from `request` and `response`. Those helpers were now used to simplify the `RUMMonitor` API:
```diff
- public func startResourceLoading(resourceKey:url:httpMethod:attributes:)
- public func stopResourceLoading(resourceKey:kind:httpStatusCode:size:attributes:)
- public func stopResourceLoadingWithError(resourceKey:error:httpStatusCode:attributes:)
- public func stopResourceLoadingWithError(resourceKey:errorMessage:httpStatusCode:attributes:)
+ public func startResourceLoading(resourceKey:request:attributes:)
+ public func stopResourceLoading(resourceKey:response:request:size:attributes:)
+ public func stopResourceLoadingWithError(resourceKey:error:response:attributes:)
+ public func stopResourceLoadingWithError(resourceKey:errorMessage:response:attributes:)
```

### Review checklist


cswatt (approved):
approved for documentation.

Sources/Datadog/RUMMonitor.swift
```diff
@@ -248,53 +248,87 @@ public class RUMMonitor: DDRUMMonitor, RUMCommandSubscriber {
         )
     }
 
-    override public func startResourceLoading(resourceKey: String, url: URL, httpMethod: RUMHTTPMethod, attributes: [AttributeKey: AttributeValue]) {
+    override public func startResourceLoading(
+        resourceKey: String,
+        request: URLRequest,
```
buranmert, line 253: `request` parameter seems confusing to me
we use it only for detecting `.xhr` kind, doesn't its `ResourceScope` already know that?
shouldn't we assume that request is immutable and the one that starts `ResourceScope` will finish it?

Sources/Datadog/RUMMonitor.swift
```diff
@@ -248,53 +248,87 @@ public class RUMMonitor: DDRUMMonitor, RUMCommandSubscriber {
         )
     }
 
-    override public func startResourceLoading(resourceKey: String, url: URL, httpMethod: RUMHTTPMethod, attributes: [AttributeKey: AttributeValue]) {
```
buranmert, line 251: i think that having an extra `startResourceLoading(key:url:...)` method (in which `httpMethod` is default to `GET`) can be useful for some manual tracing cases.
how do you see the usage with `dataTask(withURL:)`?

Sources/Datadog/RUMMonitor.swift
ncreated: I think you refer to `stopResourceLoading(...)`:
```swift
override public func stopResourceLoading(
    resourceKey: String,
    response: URLResponse,
    request: URLRequest? = nil,
    size: Int64? = nil,
    attributes: [AttributeKey: AttributeValue]
)
```
Indeed, that parameter, although optional, looks weird 👍. 

I refactored the `RUMResourceScope` a bit, so I managed to remove it. Now when the resource is started, we capture "kind by request" (optional), and when the resource is finished, we capture "kind by response" (non-optional). When sending the actual event, the "by request" takes precedence before the "by response".

With this approach, we will be compatible with upcoming bridge interface for RN, where `kind:` will be explicitly given in public API. I wanted to avoid passing `URLRequest` and `URLResponse` to scopes, as those will not exist in RN.

Sources/Datadog/RUMMonitor.swift
ncreated: Good idea, done 👍. Now we have two methods for starting RUM Resource: one with `request`, another one with `url`.

buranmert (approved):

PR #300, opened 2020-11-04
RUMM-867 Alamofire removed from Shopist

buranmert:
### What and why?

After https://github.com/DataDog/dd-sdk-ios/pull/289 fix, we lost Alamofire compatibility
As we cannot use a `DDURLSessionDelegate` kind instance as `session.delegate`, we don't trace network requests made by Alamofire session

### How?

To fix that in internal Shopist app, we drop Alamofire and switch to `URLSession`

### Review checklist


Shopist/ShopistUITests/ShopistUITests.swift
```diff
@@ -44,6 +44,7 @@ class ShopistUITests: XCTestCase {
             app.terminate()
         }
         // Wait for uploading pending batches
+        app.launchToHomepage()
```
ncreated, line 47: Is it related to AF drop, or we fix something else here? 💭

ncreated (approved):
Just one question, looks good 🚀

Shopist/ShopistUITests/ShopistUITests.swift
buranmert: no it's unrelated actually

the idea behind that `Thread.sleep` was waiting for the app to upload pending batches, but i realized we sleep the thread while there is no app running.
at worst case, we might have been losing the latest batch to upload. otherwise it shouldn't be a major issue

PR #303, opened 2020-11-08
RUMM-817 Error can be logged

### LINKED ISSUE #276: Unable to send exception via Logs
### The thing

I'm integrating Datadog SDK for Logs in both Android and iOS and I came with an issue that for android I'm able to do
Log.E(msg, throwable) but I don't see any overload method to add neither an NSError nor Error.

Is there a way to achieve send an exception? I tried to add the attribute "error.message" or "error.stack" but those are reserved.

buranmert:
### What and why?

fix #276 
There was no way to log an `Error`

### How?

1. `Error` parameter added to logging methods
2. `error.stack/kind/message` keywords are removed from `reservedAttributes` set

### Review checklist

### DIFF
--- a/Sources/Datadog/Logger.swift
+++ b/Sources/Datadog/Logger.swift
@@ -17,10 +17,10 @@ public enum LogLevel: Int, Codable {
     case critical
 }
 
-public enum ErrorAttributes: String {
-    case kind = "error.kind"
-    case message = "error.message"
-    case stack = "error.stack"
+internal enum ErrorAttributes {
+    static let kind = "error.kind"
+    static let message = "error.message"
+    static let stack = "error.stack"
 }
 
 /// Because `Logger` is a common name widely used across different projects, the `Datadog.Logger` may conflict when
@@ -76,6 +76,7 @@ public class Logger {
     /// Sends a DEBUG log message.
     /// - Parameters:
     ///   - message: the message to be logged
+    ///   - error: `Error` instance to be logged with its properties
     ///   - attributes: a dictionary of attributes to add for this message. If an attribute with
     /// the same key already exist in this logger, it will be overridden (just for this message).
     public func debug(_ message: String, error: Error? = nil, attributes: [AttributeKey: AttributeValue]? = nil) {
@@ -85,6 +86,7 @@ public class Logger {
     /// Sends an INFO log message.
     /// - Parameters:
     ///   - message: the message to be logged
+    ///   - error: `Error` instance to be logged with its properties
     ///   - attributes: a dictionary of attributes to add for this message. If an attribute with
     /// the same key already exist in this logger, it will be overridden (just for this message).
     public func info(_ message: String, error: Error? = nil, attributes: [AttributeKey: AttributeValue]? = nil) {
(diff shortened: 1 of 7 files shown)


Sources/Datadog/Logger.swift
```diff
@@ -17,6 +17,12 @@ public enum LogLevel: Int, Codable {
     case critical
 }
 
+public enum ErrorAttributes: String {
```
ncreated, line 20: Why making it `public`? The logs we create use many other attributes, but here we list only few of them. I'd rather keep it `internal`. If this is inspired by `public DDTags`, then it's not exactly the same, because tracing tags behave differently: we expect the user to give us those particular tags, but here, those error attributes are created internally.

Another thing: why do we need `enum`, instead of `struct`? I don't see how we leverage the enumeration aspect. Because of `enum`, we need to call `.rawValue` each time we reference `kind`, `message` or `stack` internally. Also, for other similar concepts (`DDTags`, `RUMContextIntegration.Attributes`, `LoggingWithActiveSpanIntegration.Attributes`) we use `struct`, so this `enum` is inconsistent.

Sources/Datadog/Logger.swift
```diff
@@ -72,53 +78,53 @@ public class Logger {
     ///   - message: the message to be logged
     ///   - attributes: a dictionary of attributes to add for this message. If an attribute with
     /// the same key already exist in this logger, it will be overridden (just for this message).
-    public func debug(_ message: String, attributes: [AttributeKey: AttributeValue]? = nil) {
-        log(level: .debug, message: message, messageAttributes: attributes)
+    public func debug(_ message: String, error: Error? = nil, attributes: [AttributeKey: AttributeValue]? = nil) {
```
ncreated, line 81: All this `public` APIs changed in `Logger` need the documentation comment update.

ncreated (changes requested):
Looks good 👌, but IMO needs two changes.

Sources/Datadog/Logger.swift
buranmert: after the discussion at daily, we decided to keep them open/customizable

Sources/Datadog/Logger.swift
```diff
@@ -17,6 +17,12 @@ public enum LogLevel: Int, Codable {
     case critical
 }
 
+public enum ErrorAttributes {
```
ncreated, line 20: If we want to keep them as part of the public API:
- [ ] `LogAttributes` (as its [Android counterpart](https://github.com/DataDog/dd-sdk-android/blob/master/dd-sdk-android/src/main/kotlin/com/datadog/android/log/LogAttributes.kt)) is a better name as it scopes those attributes - it will be more obvious that those attributes can be used with Logs, but not RUM. It will also provide better API discoverability.
- [ ] It needs public API comment.

Those are only few of all the standard attributes that we support. IMO, instead of listing just a few, we should either list all of them or none (so, make this `internal`). Currently, it may be misleading to the users who found the list of standard attributes on our website, but only few of them listed in public API.

buranmert: i make them `internal` to stay in the scope of this PR, we can add all the attributes as public in another PR later

Sources/Datadog/Logger.swift
```diff
@@ -70,55 +76,61 @@ public class Logger {
     /// Sends a DEBUG log message.
     /// - Parameters:
     ///   - message: the message to be logged
+    ///   - error: `Error` instance to be logged with its properties
```
buranmert, line 79: structs and enums are logged with their properties
class support may come with #304

ncreated (approved):
Awesome 🚀

PR #304, opened 2020-11-08
RUMM-817 DDError improvements

buranmert:
### What and why?

_This is an optional improvement; it has its cons beside its pros._

SDK converts `Error` instances into `DDError` so that it can extract information
Yet, it ignores properties of `class` types

### How?

Now `DDError` reaches the details of `Error` instance via `Mirror`
It reads `mirror.children` in order to get its properties

Before it was always `ddError.message == ddError.details`

```swift
struct MyStrError: Error { var someProp = "some value" }
class MyClsError: Error { var someProp = "some value" }
enum MyEnumError: Error { case someCase }

let dd1 = DDError(error: MyStrError())
//title "MyStrError"
//message "MyStrError(someProp: "some value")"
//details "someProp: some value"

let dd2 = DDError(error: MyClsError())
//title "MyClsError"
//message "__lldb_expr_11.MyClsError"
//details "someProp: some value"

let dd3 = DDError(error: MyEnumError.someCase)
//title "MyEnumError"
//message "someCase"
//details ""
```

### Review checklist


ncreated (commented):
This is an interesting PR - we attempt to improve the error stacks by inferring more values using reflection. However, I share your concerns about pros / cons balance. I can see some info being improved, but some other becoming less readable + not backward compatible.

I'd say:
- We know that the current implementation is limiting for anonymous `Errors`.
- We don't know if anonymous `Errors` are that popular among our user-base.
- We don't know how to improve it, as we can only guess real use cases and user expectations.
- We don't know if when users find it limiting, they will expect our SDK to be smarter in first place, or they will improve the error definition it on their end.
- We don't know how much our users depend on the current `DDError` format in their dashboards and filters, so we don't know how breaking this change would be.

There are too many "don't knows". IMHO we should keep it simple and wait for the real feedback on this topic and make `DDError` more complex only when we know which direction to go. WDYT @buranmert ?

buranmert:
@ncreated yes, actually this PR improves `class SomeError: Error` case only.
i'm closing for now, we can come back to it after the next major release with user some feedback

PR #307, opened 2020-11-13
RUMM-880 RUM attributes moved under context key

buranmert:
### What and why?

RUM v2 requires all custom attributes to be within `context` key in json

### How?

We keep the current data model as is
We add `context.*` prefix during encoding of the data model

### Review checklist

### DIFF
--- a/Tests/DatadogTests/Datadog/RUMMonitorTests.swift
+++ b/Tests/DatadogTests/Datadog/RUMMonitorTests.swift
@@ -540,14 +540,17 @@ class RUMMonitorTests: XCTestCase {
         let firstViewEvent = try rumEventMatchers
             .lastRUMEvent(ofType: RUMDataView.self) { rumModel in rumModel.view.url == "FirstViewController" }
 
-        XCTAssertEqual(try firstViewEvent.attribute(forKeyPath: "attribute1") as String, "changed value 1")
-        XCTAssertEqual(try firstViewEvent.attribute(forKeyPath: "attribute2") as String, "value 2")
+        XCTAssertNil(try? firstViewEvent.attribute(forKeyPath: "attribute1") as String)
+        XCTAssertNil(try? firstViewEvent.attribute(forKeyPath: "attribute2") as String)
+        XCTAssertEqual(try firstViewEvent.attribute(forKeyPath: "context.attribute1") as String, "changed value 1")
+        XCTAssertEqual(try firstViewEvent.attribute(forKeyPath: "context.attribute2") as String, "value 2")
 
         let secondViewEvent = try rumEventMatchers
             .lastRUMEvent(ofType: RUMDataView.self) { rumModel in rumModel.view.url == "SecondViewController" }
 
-        XCTAssertEqual(try secondViewEvent.attribute(forKeyPath: "attribute1") as String, "changed value 1")
+        XCTAssertNil(try? secondViewEvent.attribute(forKeyPath: "attribute1") as String)
         XCTAssertNil(try? secondViewEvent.attribute(forKeyPath: "attribute2") as String)
+        XCTAssertEqual(try secondViewEvent.attribute(forKeyPath: "context.attribute1") as String, "changed value 1")
     }
 
     func testWhenViewIsStarted_attributesCanBeAddedOrUpdatedButNotRemoved() throws {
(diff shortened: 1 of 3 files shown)


Tests/DatadogTests/Datadog/RUMMonitorTests.swift
```diff
@@ -540,14 +540,17 @@ class RUMMonitorTests: XCTestCase {
         let firstViewEvent = try rumEventMatchers
             .lastRUMEvent(ofType: RUMDataView.self) { rumModel in rumModel.view.url == "FirstViewController" }
 
-        XCTAssertEqual(try firstViewEvent.attribute(forKeyPath: "attribute1") as String, "changed value 1")
-        XCTAssertEqual(try firstViewEvent.attribute(forKeyPath: "attribute2") as String, "value 2")
+        XCTAssertNil(try? firstViewEvent.attribute(forKeyPath: "attribute1") as String)
+        XCTAssertNil(try? firstViewEvent.attribute(forKeyPath: "attribute2") as String)
```
ncreated, line 544: Why those assertions? Since this PR, `"attribute1"` and `"attribute2"` are no longer valid key paths, so here we test that nothing is encoded on **some** invalid key path. As there is infinite number of such invalid key-paths, this IMO makes very little sense and only lowers the readability.

Tests/DatadogTests/Datadog/RUMMonitorTests.swift
```diff
@@ -540,14 +540,17 @@ class RUMMonitorTests: XCTestCase {
...
+        XCTAssertEqual(try firstViewEvent.attribute(forKeyPath: "context.attribute1") as String, "changed value 1")
+        XCTAssertEqual(try firstViewEvent.attribute(forKeyPath: "context.attribute2") as String, "value 2")
 
         let secondViewEvent = try rumEventMatchers
             .lastRUMEvent(ofType: RUMDataView.self) { rumModel in rumModel.view.url == "SecondViewController" }
 
-        XCTAssertEqual(try secondViewEvent.attribute(forKeyPath: "attribute1") as String, "changed value 1")
+        XCTAssertNil(try? secondViewEvent.attribute(forKeyPath: "attribute1") as String)
```
ncreated, line 551: Same here, this assertion makes no sense IMO - it checks if some very weird thing didn't happen.

Tests/DatadogTests/Datadog/RUMMonitorTests.swift
```diff
@@ -540,14 +540,17 @@ class RUMMonitorTests: XCTestCase {
...
         XCTAssertNil(try? secondViewEvent.attribute(forKeyPath: "attribute2") as String)
```
ncreated, line 552: Here we should test that second attribute is not encoded, as it was removed:
```suggestion
        XCTAssertNil(try? secondViewEvent.attribute(forKeyPath: "context.attribute2") as String)
```

Tests/DatadogTests/Datadog/RUMMonitorTests.swift
```diff
@@ -570,9 +573,12 @@ class RUMMonitorTests: XCTestCase {
         let rumEventMatchers = try RUMFeature.waitAndReturnRUMEventMatchers(count: 3)
         let lastViewUpdate = try rumEventMatchers.lastRUMEvent(ofType: RUMDataView.self)
 
-        try XCTAssertEqual(lastViewUpdate.attribute(forKeyPath: "a1"), "bar1", "The value should be updated")
-        try XCTAssertEqual(lastViewUpdate.attribute(forKeyPath: "a2"), "foo2", "The attribute should not be removed")
-        try XCTAssertEqual(lastViewUpdate.attribute(forKeyPath: "a3"), "foo3", "The attribute should be added")
+        XCTAssertNil(try? lastViewUpdate.attribute(forKeyPath: "a1") as String)
+        XCTAssertNil(try? lastViewUpdate.attribute(forKeyPath: "a2") as String)
+        XCTAssertNil(try? lastViewUpdate.attribute(forKeyPath: "a3") as String)
```
ncreated, line 578: Same - no reason to test that values are not encoded on unknown key-paths.

ncreated (changes requested):

Tests/DatadogTests/Datadog/RUMMonitorTests.swift
buranmert: what you said makes sense for other keys like `foo` or `bar`
but here, those assertions make sense when you look at the line above
```swift
monitor.addAttribute(forKey: "attribute1", value: "value 1")
```
_i set some value for this key_ **but** _i expect it not to be there_

ncreated: If we want to test that the attribute is encoded on one and only one JSON key, this assertion should go to encoder tests:

```swift
// RUMEventEncoderTests

let event = RUMEvent(model: RUMDataModelMock(), attributes: ["foo": "bar"])
let jsonString = try! JSONEncoder().encode(event).utf8String
XCTAssertEqual(jsonString, #"{"context.foo":"bar"}"#)
```

That would be comprehensive, reading: _"The `foo` attribute is encoded on `context.foo` JSON key-path and no other key-path is created."_

ncreated: Just noticed, that we already have the full JSON assertion: https://github.com/DataDog/dd-sdk-ios/blob/a4fbe967a3cc641c6476a6f10023cd0b52fd30a5/Tests/DatadogTests/Datadog/RUM/RUMEventOutputs/RUMEventFileOutputTests.swift#L65-L72 which is enough to ensure that attributes are encoded only once and on the certain key-path.

buranmert: good point, now we test the whole chain in `RUMMonitorTests` and we should be testing `RUMMonitor` only.
i think you mean we shouldn't assert on JSON output in `RUMMonitorTests` since
1. output of `RUMMonitor` is not JSON (in theory, we should assert on commands received by its `AppScope`)
2. JSON is being tested in `RUMEventFileOutputTests`

i agree. but this requires refactoring in tests, beyond the scope of this PR

ncreated (approved):

PR #308, opened 2020-11-13
RUMM-674 Shopist randomizes IP addresses

buranmert:
### What and why?

We'd like to populate our dashboards with data from all over the world.
_Fake_ Shopist users of the world, unite! ✊ 

### Review checklist

- [x] Merge #307 first

### DIFF
--- a/Shopist/Shopist/User.swift
+++ b/Shopist/Shopist/User.swift
@@ -1,0 +1,70 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+private extension UInt8 {
+    static func randomByte(
+        in range: ClosedRange<UInt8> = UInt8.min...UInt8.max,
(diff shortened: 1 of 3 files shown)


Shopist/Shopist/User.swift
```diff
@@ -0,0 +1,70 @@
...
+            let byte = UInt8.random(in: range)
+            if !except.contains(byte) {
+                return byte
+            }
+        }
+    }
+}
+
+private extension String {
+    static func randomIPv4() -> Self {
```
ncreated, line 24: nit: ok as it's only a private extension, but I generally dislike the idea of extending the type with a method only because we need a value of that type. `String` and "IPv4 address" have nothing in common 🙂, private global function would work much better IMO.

ncreated (approved):
All good 👍

Shopist/Shopist/User.swift
buranmert: so long as it's private, i don't see much of a difference between extension or anything else tbh.
in this case `IPv4` is a `String`. in order to make them unrelated, we could've created a new type for `IPv4` but it has its own problems like more complexity for too few use cases.

also, maybe you can think of it like that:
```swift
typealias IPv4 = String
extension IPv4 { static func random() -> Self { ... } }
```

PR #314, opened 2020-11-20
RUMM-696 Stop sending tracing `Span` for RUM Resources

ncreated:
### What and why?

🚚 This PR changes the logic of auto-instrumented, network `Span` generation depending of which feature (RUM or Tracing) is enabled. This is because recently our RUM backend started generating spans for RUM Resources on behalf of the SDK.

New behaviour, assuming that `set(firstPartyHosts:)` is set and `DDURLSessionDelegate()` is installed:
| Tracing feature  | RUM feature | `Global.sharedTracer` | `Global.rum` | 1st party request | 3rd party request |
| :--------------: | :----------: | :--------: | :-------------: | :----------------: | :----------------: |
| ✅ | ❌ | registered | - | send `Span` + inject tracing headers | - |
| ❌ | ✅ | - | registered | send RUM Resource | send RUM Resource |
| ✅ | ✅ | registered | registered | send RUM Resource with `trace_id` and `span_id` + inject tracing headers + inject `x-datadog-origin: rum` header | send RUM Resource |

Also, appropriate warnings are printed on different misconfiguration, e.g. _when Tracing auto-instrumentation is enabled, and a first party request is sent while `Global.sharedTracer` is not registered_.

### How?

Continuing what was started in #262, the logic of `interceptor` was simplified. Now, instead of two distinct handlers (for RUM and Tracing), it takes only one interface:
```swift
internal protocol URLSessionInterceptionHandler {
    func notify_taskInterceptionStarted(interception: TaskInterception)
    func notify_taskInterceptionCompleted(interception: TaskInterception)
}
```
implemented by `URLSessionRUMResourceHandler` and `URLSessionTracingHandler`.

**Interceptor:**
* injects / extracts `SpanContext` and custom `x-datadog-origin: rum` header into/from 1st party requests,
* notifies handler on interception start and completion.

**Handler:**
* `URLSessionRUMResourceHandler` - send RUM Resources and associate `RUMSpanContext` if set for the `interception`,
* `URLSessionTracingHandler` - send `Span` only for interceptions flagged as `interception.isFirstPartyRequest`.

### Review checklist

### DIFF
--- a/Sources/Datadog/URLSessionAutoInstrumentation/Interception/URLSessionInterceptor.swift
+++ b/Sources/Datadog/URLSessionAutoInstrumentation/Interception/URLSessionInterceptor.swift
@@ -76,10 +76,10 @@ internal class URLSessionInterceptor: URLSessionInterceptorType {
         guard !internalURLsFilter.isInternal(url: request.url) else {
             return request
         }
-        if let tracer = Global.sharedTracer as? Tracer {
-            if firstPartyURLsFilter.isFirstParty(url: request.url) {
-                return injectSpanContext(into: request, using: tracer)
-            }
+        if tracingHandler != nil, firstPartyURLsFilter.isFirstParty(url: request.url) {
+            // If the tracing handler is enabled, we inject tracing context to the request.
+            // This context is later used by either tracing or RUM handler when creating the Span or RUM Resource.
+            return injectSpanContext(into: request)
         }
         return request
     }
@@ -94,8 +94,7 @@ internal class URLSessionInterceptor: URLSessionInterceptorType {
             let interception = TaskInterception(request: request)
             self.interceptionByTask[task] = interception
 
-            if let tracer = Global.sharedTracer as? Tracer,
-               let spanContext = self.extractSpanContext(from: request, using: tracer) {
+            if let spanContext = self.extractSpanContext(from: request) {
                 interception.register(spanContext: spanContext)
             }
 
@@ -148,17 +147,26 @@ internal class URLSessionInterceptor: URLSessionInterceptorType {
     private func finishInterception(task: URLSessionTask, interception: TaskInterception) {
         interceptionByTask[task] = nil
 
-        if let tracer = Global.sharedTracer as? Tracer,
-           firstPartyURLsFilter.isFirstParty(url: interception.request.url) {
-            tracingHandler?.sendSpan(for: interception, using: tracer)
+        if let rumResourceHandler = rumResourceHandler {
+            // If the RUM feature is enabled we only send RUM Resource.
+            // Tracing Span is not send, no matter of the Tracing feature state. The Span will be generated
+            // by Datadog backend if the tracing information is encoded in RUM Resource.
+            rumResourceHandler.notify_taskInterceptionCompleted(interception: interception)
+        } else if let tracingHandler = tracingHandler {
+            if firstPartyURLsFilter.isFirstParty(url: interception.request.url) {
+                // If the RUM feature is disabled we send the tracing Span only for 1st party hosts.
+                tracingHandler.sendSpan(for: interception)
+            }
         }
-
-        self.rumResourceHandler?.notify_taskInterceptionCompleted(interception: interception)
     }
 
     // MARK: - SpanContext Injection & Extraction
 
-    private func injectSpanContext(into request: URLRequest, using tracer: Tracer) -> URLRequest {
+    private func injectSpanContext(into request: URLRequest) -> URLRequest {
+        guard let tracer = Global.sharedTracer as? Tracer else {
+            return request
+        }
+
         let writer = HTTPHeadersWriter()
         let spanContext = tracer.createSpanContext()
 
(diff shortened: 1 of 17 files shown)


Sources/Datadog/URLSessionAutoInstrumentation/Interception/URLSessionInterceptor.swift
```diff
@@ -147,34 +170,41 @@ internal class URLSessionInterceptor: URLSessionInterceptorType {
...
     }
 
-    private func extractSpanContext(from request: URLRequest, using tracer: Tracer) -> DDSpanContext? {
+    private func extractSpanContext(from request: URLRequest) -> DDSpanContext? {
+        guard let tracer = Global.sharedTracer as? Tracer else {
+            return nil
+        }
         guard let headers = request.allHTTPHeaderFields else {
             return nil
         }
```
buranmert, line 206:
```suggestion
        guard let tracer = Global.sharedTracer as? Tracer,
        let headers = request.allHTTPHeaderFields else {
            return nil
        }
```

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/Interception/URLSessionInterceptorTests.swift
```diff
@@ -101,14 +94,77 @@ class URLSessionInterceptorTests: XCTestCase {
...
+        // Then
+        XCTAssertNotNil(interceptedFirstPartyRequest.allHTTPHeaderFields?[TracingHTTPHeaders.traceIDField])
+        XCTAssertNotNil(interceptedFirstPartyRequest.allHTTPHeaderFields?[TracingHTTPHeaders.parentSpanIDField])
+        XCTAssertEqual(interceptedFirstPartyRequest.allHTTPHeaderFields?[TracingHTTPHeaders.originField], TracingHTTPHeaders.rumOriginValue)
+        XCTAssertNil(interceptedThirdPartyRequest.allHTTPHeaderFields?[TracingHTTPHeaders.traceIDField])
+        XCTAssertNil(interceptedThirdPartyRequest.allHTTPHeaderFields?[TracingHTTPHeaders.parentSpanIDField])
+        XCTAssertNil(interceptedThirdPartyRequest.allHTTPHeaderFields?[TracingHTTPHeaders.originField])
+        XCTAssertNil(interceptedInternalRequest.allHTTPHeaderFields?[TracingHTTPHeaders.traceIDField])
+        XCTAssertNil(interceptedInternalRequest.allHTTPHeaderFields?[TracingHTTPHeaders.parentSpanIDField])
+        XCTAssertNil(interceptedInternalRequest.allHTTPHeaderFields?[TracingHTTPHeaders.originField])
```
buranmert, line 146: i guess we expect `request.allHTTPHeaderFields == nil/unchanged` in those 2 cases:
```swift
XCTAssertNil(interceptedThirdPartyRequest.allHTTPHeaderFields?)
XCTAssertNil(interceptedInternalRequest.allHTTPHeaderFields?)
```
?
and the next lines might already cover them:
```swift
XCTAssertEqual(thirdPartyRequest, interceptedThirdPartyRequest)
XCTAssertEqual(internalRequest, interceptedInternalRequest)
```
if `HTTPHeaderFields` are changed, those assertions fail; don't they?

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/Interception/URLSessionInterceptorTests.swift
```diff
@@ -101,14 +94,77 @@ class URLSessionInterceptorTests: XCTestCase {
...
+            "Tracing headers should be injected when both Tracing and RUM instrumentations are enabled."
+        )
+        XCTAssertEqual(
+            interceptor.additionalHeadersForFirstPartyRequests,
+            [TracingHTTPHeaders.originField: TracingHTTPHeaders.rumOriginValue],
+            "Additional `x-datadog-origin: rum` header should be injected when both Tracing and RUM instrumentations are enabled."
+        )
     }
 
     // MARK: - URLRequest Interception
```
buranmert, line 109: we can add already traced request (_original request with tracing headers_) test case here
i guess we expect to have the original headers after the interception, correct?

buranmert (approved):
looks good, we can add one more test case though (i couldn't find if the case was already there)

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/Interception/URLSessionInterceptorTests.swift
ncreated: Yes, inspecting `.allHTTPHeaderFields` is redundant if we assert the full request equality later 👍. However, due to request equality bug in iOS 14 SDK: https://openradar.appspot.com/radar?id=4988276943355904 it seemed safer to do it before. To be more verbose on this, I added extra helper: `assertRequestsEqual(_:_:)` which implements the workaround. Redundant checks are now removed 👌.

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/Interception/URLSessionInterceptorTests.swift
ncreated: Sorry, I don't understand 🙂. In quoted lines we do not test requests, but only internal state of the `interceptor` (`interceptor.additionalHeadersForFirstPartyRequests`).

There's no such case, where `interceptor.modify(request:)` is called for a request, which was previously intercepted. This contract is guaranteed in swizzler tests.

xgouchet (approved):
LGTM

PR #315, opened 2020-11-20
RUMM-886 `setUserInfo(..., extraInfo)` added

buranmert:
### What and why?

`Datadog.UserInfo` has arbitrary attributes via `extraInfo`

### How?

`UserInfoProvider` keeps `extraInfo` dictionary along with `userInfo`
Features (Logs, APM and RUM) encode `extraInfo` according to their own rules
#### Logs:

```swift
extraInfo: ["key": "value"] -> "usr.key: value"
```

#### APM:

```swift
extraInfo: ["key": "value"] -> "meta.usr.key: value"
```

#### RUM:

```swift
extraInfo: ["key": "value"] -> "context.usr.key: value"
```

### Discussion 💭

In this PR, I put `usr.` prefix to `extraInfo` within `UserInfoProvider` and pass it to features as regular attributes.
But `UserInfoProvider` passes `UserInfo` value without encoding and the features (`LogEncoder` and `SpanEncoder`) are responsible for its encoding.
This is some sort of inconsistency in `UserInfoProvider`, I didn't like that ❌ 

Can't we make `UserInfo: Encodable` to de-duplicate its encoding logic?

### Review checklist

### DIFF
--- a/Sources/Datadog/Datadog.swift
+++ b/Sources/Datadog/Datadog.swift
@@ -75,6 +75,13 @@ public class Datadog {
         }
     }
 
+    /// Sets current user information.
+    /// Those will be added to logs, traces and RUM events automatically.
+    /// - Parameters:
+    ///   - id: User ID, if any
+    ///   - name: Name representing the user, if any
+    ///   - email: User's email, if any
+    ///   - extraInfo: User's custom attributes, if any
     public static func setUserInfo(
         id: String? = nil,
         name: String? = nil,
--- a/Sources/Datadog/Datadog.swift
+++ b/Sources/Datadog/Datadog.swift
@@ -78,9 +78,15 @@ public class Datadog {
     public static func setUserInfo(
         id: String? = nil,
         name: String? = nil,
-        email: String? = nil
+        email: String? = nil,
+        extraInfo: [AttributeKey: AttributeValue] = [:]
     ) {
-        instance?.userInfoProvider.value = UserInfo(id: id, name: name, email: email)
+        instance?.userInfoProvider.value = UserInfo(
+            id: id,
+            name: name,
+            email: email,
+            extraInfo: extraInfo
+        )
     }
 
     // MARK: - Internal
--- a/Sources/Datadog/Tracing/Span/SpanEncoder.swift
+++ b/Sources/Datadog/Tracing/Span/SpanEncoder.swift
@@ -198,5 +198,8 @@ internal struct SpanEncoder {
             let metaKey = "meta.\($0.key)"
             try container.encode($0.value, forKey: DynamicCodingKey(metaKey))
         }
+        try span.userInfo.extraInfo.forEach {
+            try container.encode(EncodableValue($1), forKey: DynamicCodingKey("meta.usr.\($0)"))
+        }
     }
 }
--- a/Sources/Datadog/Tracing/Span/SpanEncoder.swift
+++ b/Sources/Datadog/Tracing/Span/SpanEncoder.swift
@@ -198,8 +198,5 @@ internal struct SpanEncoder {
             let metaKey = "meta.\($0.key)"
             try container.encode($0.value, forKey: DynamicCodingKey(metaKey))
         }
-        try span.userInfo.extraInfo.forEach {
-            try container.encode(EncodableValue($1), forKey: DynamicCodingKey("meta.usr.\($0)"))
-        }
     }
 }
--- a/Tests/DatadogTests/Datadog/LoggerTests.swift
+++ b/Tests/DatadogTests/Datadog/LoggerTests.swift
@@ -186,16 +186,16 @@ class LoggerTests: XCTestCase {
         Datadog.setUserInfo(id: "abc-123", name: "Foo")
         logger.debug("message with user `id` and `name`")
 
-        Datadog.setUserInfo(id: "abc-123", name: "Foo", email: "foo@example.com")
-        logger.debug("message with user `id`, `name` and `email`")
+        Datadog.setUserInfo(id: "abc-123", name: "Foo", email: "foo@example.com", extraInfo: ["extra_key": "extra_value"])
+        logger.debug("message with user `id`, `name`, `email` and `extraInfo`")
 
         Datadog.setUserInfo(id: nil, name: nil, email: nil)
         logger.debug("message with no user info")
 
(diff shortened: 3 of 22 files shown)


ncreated:
> In this PR, I put `usr.` prefix to `extraInfo` within `UserInfoProvider` and pass it to features as regular attributes.
But `UserInfoProvider` passes `UserInfo` value without encoding and the features (`LogEncoder` and `SpanEncoder`) are responsible for its encoding.
> This is some sort of inconsistency in `UserInfoProvider`, I didn't like that ❌

What exactly is inconsistent? 

`UserInfoProvider` provides the `UserInfo` model and each feature decides on its own, how to encode it:
* Logging feature puts them at the root level of the log JSON: `usr.id`, `usr.name`, `usr.email`,
* Tracing feature nests them under `meta` object in the span JSON: `meta.usr.id`, `meta.usr.name`, `meta.usr.email`,
* RUM feature uses auto-generated `RUMDataUSR` model to encode it according to `rum-events-format`.

Having single `struct UserInfo: Encodable` would be a wrong abstraction, as each feature needs to decide by its own.

> In this PR, I put `usr. prefix` to extraInfo within `UserInfoProvider` 

In all other places (see: `NetworkConnectionInfoProvider`, `CarrierInfoProvider`, `UserInfoProvider`) the encoding keys of custom data are decided by the encoder (`LogEncoder`, `SpanEncoder`). Creating user attribute keys at the beginning of data flow leaks this abstraction. I don't see any benefit of this approach, why do you think it's better?

Sources/Datadog/Datadog.swift
```diff
@@ -78,9 +78,15 @@ public class Datadog {
     public static func setUserInfo(
```
ncreated, line 78: This public API is missing the comment.

Sources/Datadog/Tracing/Span/SpanEncoder.swift
```diff
@@ -198,5 +198,8 @@ internal struct SpanEncoder {
             let metaKey = "meta.\($0.key)"
             try container.encode($0.value, forKey: DynamicCodingKey(metaKey))
         }
+        try span.userInfo.extraInfo.forEach {
+            try container.encode(EncodableValue($1), forKey: DynamicCodingKey("meta.usr.\($0)"))
```
ncreated, line 202: This leads to all spans being rejected by Intake if the user extra attribute is not `String`, e.g.:
```swift
Datadog.setUserInfo(
    id: "abcd-1234",
    name: "foo",
    email: "foo@example.com",
    extraInfo: [
        "is-registered": true
    ]
)
```
As stated in this method comment:
```swift
// NOTE: RUMM-299 only string values are supported for `meta.*` attributes
```
Whatever we encode on `meta.*` key must be serialised to `String` representation. We should use `JSONStringEncodableValue` in `SpanBuilder` as it is done for `span.tags`.

Tests/DatadogTests/Datadog/RUMMonitorTests.swift
```diff
@@ -445,7 +445,12 @@ class RUMMonitorTests: XCTestCase {
             directory: temporaryDirectory,
             dependencies: .mockWith(
                 userInfoProvider: .mockWith(
-                    userInfo: UserInfo(id: "abc-123", name: "Foo", email: "foo@bar.com")
+                    userInfo: UserInfo(
+                        id: "abc-123",
+                        name: "Foo",
+                        email: "foo@bar.com",
+                        extraInfo: ["extra_key": "extra_value"]
```
ncreated, line 452: Because `extraInfo` encoding depends on the feature (Logging and RUM supports multiple `Encodable` types, but Tracing encodes all values as `String`), it seems better to mix few types in this test:
```swift
extraInfo: ["extra_string": "string_value", "extra_int": 42, "extra_bool": true]
```
and document the expected encoding result in test assertions.

Tests/DatadogTests/Datadog/TracerTests.swift
```diff
@@ -335,8 +335,8 @@ class TracerTests: XCTestCase {
         Datadog.setUserInfo(id: "abc-123", name: "Foo")
         tracer.startSpan(operationName: "span with user `id` and `name`").finish()
 
-        Datadog.setUserInfo(id: "abc-123", name: "Foo", email: "foo@example.com")
-        tracer.startSpan(operationName: "span with user `id`, `name` and `email`").finish()
+        Datadog.setUserInfo(id: "abc-123", name: "Foo", email: "foo@example.com", extraInfo: ["extra_key": "extra_value"])
```
ncreated, line 338: Because `extraInfo` encoding depends on the feature (Logging and RUM supports multiple `Encodable` types, but Tracing encodes all values as `String`), it seems better to mix few types in this test:
```swift
extraInfo: ["extra_string": "string_value", "extra_int": 42, "extra_bool": true]
```
and document the expected encoding result in test assertions.

Tests/DatadogTests/Datadog/LoggerTests.swift
```diff
@@ -186,16 +186,16 @@ class LoggerTests: XCTestCase {
         Datadog.setUserInfo(id: "abc-123", name: "Foo")
         logger.debug("message with user `id` and `name`")
 
-        Datadog.setUserInfo(id: "abc-123", name: "Foo", email: "foo@example.com")
-        logger.debug("message with user `id`, `name` and `email`")
+        Datadog.setUserInfo(id: "abc-123", name: "Foo", email: "foo@example.com", extraInfo: ["extra_key": "extra_value"])
```
ncreated, line 189: Because `extraInfo` encoding depends on the feature (Logging and RUM supports multiple `Encodable` types, but Tracing encodes all values as `String`), it seems better to mix few types in this test:
```swift
extraInfo: ["extra_string": "string_value", "extra_int": 42, "extra_bool": true]
```
and document the expected encoding result in test assertions.

ncreated (changes requested):
Looks very good 👌, but needs the fix for Tracing.

Tests/DatadogTests/Datadog/LoggerTests.swift
buranmert: i can't pass different types in a dict to `assertUserInfo` method below
i will need to use `matcher.assertValue(forKey:value:)`

ncreated (approved):
All good 🚀

PR #318, opened 2020-11-24
RUMM-655 Add `lyft/Kronos` dependency for NTP time sync

ncreated:
### What and why?

📦 We want to use NTP to compute the real time of writing events (logs, spans, rum). In this PR I'm adding the [lyft/Kronos](https://github.com/lyft/Kronos) dependency which does the NTP sync. It is not yet used in SDK code, as I wanted to separate project configuration from the actual work in SDK. This PR is not targeting the `master` branch yet.

### How?

**Cocoapods** :
```ruby
# Datadog.podspec + Datadog.podspec.src
s.dependency 'Kronos', '~> 4.1'
```
**SPM**:
```swift
// Package.swift
.package(url: "https://github.com/lyft/Kronos.git", .upToNextMinor(from: "4.1.0"))
```

**Carthage** required few changes to the project (and the update in `dependency-manager-tests/carthage/CTProject`). As it was quite strenuous, I'm leaving the note on what was changed (it's also written in commit messages):

Things change fast in Carthage (their guides are outdated), but this is how it worked in Xcode 12.0 (12A7209), target by target:
* `Datadog`:
   * Add `Kronos.framework` to Datadog > Build Phases > Link Binary With Libraries.
   * Add Datadog > Build Settings > Frameworks Search Path: `$(SRCROOT)/../Carthage/Build/iOS`
* `Example`:
   * Add `Kronos.framework` to Example > Frameworks, Libraries and Embedded Content.
   * Add Example > Build Settings > Frameworks Search Path: `$(SRCROOT)/../Carthage/Build/iOS`
* `DatadogIntegrationTests`:
   * Add 'Copy Files' build phase, coping `Kronos.framework` into `Frameworks` folder
* `DatadogTests`, `DatadogBenchmarkTests`:
   * no changes

Also, I had to exclude the `arm64` architecture for iOS Simulator builds (done through `.xcconfig`) due to getting:
> error: module 'Kronos' was created for incompatible target arm64-apple-ios8.0

### Review checklist

### DIFF
--- a/Datadog/Datadog.xcodeproj/project.pbxproj
+++ b/Datadog/Datadog.xcodeproj/project.pbxproj
@@ -155,6 +155,10 @@
 		614E9EB3244719FA007EE3E1 /* BundleType.swift in Sources */ = {isa = PBXBuildFile; fileRef = 614E9EB2244719FA007EE3E1 /* BundleType.swift */; };
 		6152C83E24BE1C91006A1679 /* HTTPServerMock in Frameworks */ = {isa = PBXBuildFile; productRef = 6152C83D24BE1C91006A1679 /* HTTPServerMock */; };
 		6152C84024BE1CC8006A1679 /* DataUploaderBenchmarkTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6152C83F24BE1CC8006A1679 /* DataUploaderBenchmarkTests.swift */; };
+		61569795256CF6D400C6AADA /* Kronos.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 61569794256CF6D400C6AADA /* Kronos.framework */; };
+		615697E3256CFB4700C6AADA /* Kronos.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 61569794256CF6D400C6AADA /* Kronos.framework */; };
+		615697E4256CFB4700C6AADA /* Kronos.framework in ⚙️ Embed Framework Dependencies */ = {isa = PBXBuildFile; fileRef = 61569794256CF6D400C6AADA /* Kronos.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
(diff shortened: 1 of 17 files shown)


Datadog/Datadog.xcodeproj/project.pbxproj
```diff
@@ -2979,12 +3010,17 @@
...
 			buildSettings = {
 				CODE_SIGN_STYLE = Automatic;
 				DEFINES_MODULE = YES;
 				DYLIB_COMPATIBILITY_VERSION = 1;
 				DYLIB_CURRENT_VERSION = 1;
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
+				FRAMEWORK_SEARCH_PATHS = (
+					"$(inherited)",
+					"$(SRCROOT)/../Carthage/Build/iOS",
+				);
```
buranmert, line 3023: i don't know if that works but what if we add `Khronos` as a Swift Package via Xcode's interface?
i guess that we may not need to change `framework search paths` and `make dependencies` step in `Makefile` (also we will be less dependent on `carthage` 🙌 )

wdyt?

ncreated: IDK if Carthage supports fetching `.xcodeproj` SPM dependencies before build (probably yes, as it's `xcodebuild's` job), but even if it does, this will be problematic / incorrect if the app is using another `Carthage` dependency which depends on `Kronos`. I can see two problems:
* version resolution: `Carthage` doesn't talk to `SPM` so they won't be able to resolve common `Kronos` version,
* conflicting symbols: our SPM's `Kronos` symbols may conflict with `Kronos` installed by `Carthage`.

buranmert: hmm, yes that's true. the host project will also crash at runtime as it won't have `Khronos` embedded in itself.
okay, so we will need to stick with `carthage` then: resolved

buranmert (approved):

PR #320, opened 2020-11-25
RUMS-132 Allow data uploads if battery status is `.unknown`

ncreated:
### What and why?

🚚 This PR fixes an issue reported as data being not uploaded from iPad (6th gen) iOS 14.0.1, when it is connected to MacBook through a lightning cable.

### How?

The investigation nailed this issue down to receiving `batteryStatus: unknown` with `batteryLevel: -1`. Given existing logic in `DataUploadConditions`, this clearly evaluated to `false` (no upload).

This fix returns arbitrary `true` value if battery status is undetermined. As we can see from RUMS-132, this may lead to making assumptions on unreliable values. 

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/Upload/DataUploadConditions.swift
+++ b/Sources/Datadog/Core/Upload/DataUploadConditions.swift
@@ -30,6 +30,13 @@ internal struct DataUploadConditions {
     }
 
     private func shouldUploadFor(batteryStatus: BatteryStatus) -> Bool {
+        if batteryStatus.state == .unknown {
+            // Note: in RUMS-132 we got the report on `.unknown` battery state reporing `-1` battery level on iPad device
+            // plugged to Mac through lightning cable. As `.unkown` may lead to other unreliable values,
+            // it seems safer to arbitrary allow uploads in such case.
+            return true
+        }
+
         let batteryFullOrCharging = batteryStatus.state == .full || batteryStatus.state == .charging
         let batteryLevelIsEnough = batteryStatus.level > Constants.minBatteryLevel
         let isLowPowerModeEnabled = batteryStatus.isLowPowerModeEnabled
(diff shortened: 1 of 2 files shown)


Sources/Datadog/Core/Upload/DataUploadConditions.swift
```diff
@@ -30,6 +30,13 @@ internal struct DataUploadConditions {
     }
 
     private func shouldUploadFor(batteryStatus: BatteryStatus) -> Bool {
+        if batteryStatus.state == .unknown {
+            // Note: in RUMS-132 we got the report on `.unknown` battery state reporing `-1` battery level on iPad device
+            // plugged to Mac through lightning cable. As `.unkown` may lead to other unreliable values,
+            // it seems safer to arbitrary allow uploads in such case.
+            return true
```
buranmert, line 37: it makes more sense to return false only in certain cases and return true for all the rest. that way, we will be uploading by default

```swift
let runningOnBattery = (state == .unplugged)
let lowBattery = runningOnBattery && (level <= .minBattery)
let lowPowerEnabled = .isLowPowerEnabled
return !(lowBattery || lowPowerEnabled)
```

buranmert (changes requested):

Sources/Datadog/Core/Upload/DataUploadConditions.swift
ncreated: What's the difference 💭? It seems similar, with the conditions being just re-written, but I don't find it easier for read.

And it doesn't have the same behaviour, for `state = .unkown` + `.isLowPowerModeEnabled = true` it gives `false`. As I suggested in the PR description, IMHO we shouldn't rely on any further info if the state is `.unknown`.

buranmert: 🎗️ for the records: `UIDevice.BatteryState.{all unknown cases}` are mapped to `Datadog.BatteryState.unknown` and `.unknown` here is `Datadog.BatteryState.unknown`. basically, `Datadog.BatteryState` is a frozen enum in that case.

buranmert (approved):

PR #317, opened 2020-11-20
RUMM-894 URLProvider fixed for illegal chars

buranmert:
### What and why?

Fixes #316 

### How?

Wrong property of `URLComponents` (`percentEncodedQueryItems`) was used in `URLProvider`
It is fixed by being replaced with the correct method (`queryItems`)

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference

PR #321, opened 2020-11-26
RUMM-655 / RUMM-689 Use NTP time for Logs, Spans and RUM events

ncreated:
### What and why?

📦 This PR updates all 3 features to use the NTP server time for events (logs, span, rum events) creation time. This ensures time compatibility with other Datadog products, e.g. aligns client's portion of a distributed trace. It also help by reporting the real time value of events, instead of depending on the device clock, which may be adjusted manually by the user.

### How?

Dedicated component is introduced for applying the time correction:
```swift
internal protocol DateCorrectionType {
    /// Corrects given device time to server time using the last known time difference between the two.
    func toServerDate(deviceDate: Date) -> Date
}
```
Internally, the `DateCorrection` uses [lyft/Kronos](https://github.com/lyft/Kronos) to perform time sync over NTP.

NTP synchronisation starts with `Datadog.initialize()` and may take from several to a dozen of seconds until it computes the precise time difference. This is notified by printing either `.info` or `.warning` message to user console (if using `Datadog.debug` flag).

All events are collected using **device time**. The correction to **server time** is applied at the writing time. This is to preserve relative times within the events, i.e. in Tracing we correct only the "start" date, as the "finish" date is given by relative `span.duration`. Similar for RUM, where all events have their `date` and other metrics use relative, nanosecond durations.

Last, but not least, to not perform double-correction, the `&batch_time=` query item was removed from all products.

### Review checklist


Sources/Datadog/Core/Upload/DataUploader.swift
```diff
@@ -41,7 +33,9 @@ internal class UploadURLProvider {
 
     var url: URL {
         var urlComponents = URLComponents(url: urlWithClientToken, resolvingAgainstBaseURL: false)
-        urlComponents?.percentEncodedQueryItems = queryItemProviders.map { $0.value() }
+        if !queryItemProviders.isEmpty {
+            urlComponents?.percentEncodedQueryItems = queryItemProviders.map { $0.value() }
```
buranmert, line 37:
```suggestion
            urlComponents?.queryItems = queryItemProviders.map { $0.value() }
```
this was fixed here #317

Sources/Datadog/Logging/Log/LogBuilder.swift
```diff
@@ -22,10 +22,13 @@ internal struct LogBuilder {
     let networkConnectionInfoProvider: NetworkConnectionInfoProviderType?
     /// Shared mobile carrier info provider (or `nil` if disabled for given logger).
     let carrierInfoProvider: CarrierInfoProviderType?
+    /// Adjusts log's time (device time) to server time.
+    let dateCorrection: NTPDateCorrectionType?
 
     func createLogWith(level: LogLevel, message: String, date: Date, attributes: LogAttributes, tags: Set<String>) -> Log {
+        let dateInServerTime = dateCorrection?.toServerDate(deviceDate: date)
```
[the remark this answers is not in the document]
ncreated: This is optional because of the internal logger (used when `Datadog.debug` is set), which doesn't need time correction - it prints logs with device time.

For comparison, it's non-optional in `SpanBuilder` as we don't have any `Tracer` use case where we don't want the time correction to be applied.

buranmert (commented):
`DateCorrection.toServerDate(Date) -> Date` looks confusing to me: `DateCorrection` sounds more like a `struct` containing correction data and `toServerDate` looks more like an instance method for `Date` type

also, do all `Logger`, `SpanBuilder`, `RUMScopes`, etc. need to know `DateCorrectionType`? i think they just need the correct date/time, they don't need to know more than `DateProvider.now()`. we can correct the time within `DateProvider` by injecting `SystemTimeProvider` in it, otherwise `DateCorrection` instances are not useful by themselves.

wdyt?

Sources/Datadog/Core/Upload/DataUploader.swift
ncreated: 👍 , rebased.

ncreated:
> do all `Logger`, `SpanBuilder`, `RUMScopes`, etc. need to know `DateCorrectionType`? i think they just need the correct date/time, they don't need to know more than `DateProvider.now()`. we can correct the time within `DateProvider` by injecting `SystemTimeProvider` in it
> 
> wdyt?

As explained: 

> All events are collected using **device time**. The correction to server time is applied at the **writing time**. This is to preserve relative times within the events, i.e. in Tracing we correct only the "start" date, as the "finish" date is given by relative `span.duration`. Similar for RUM, where all events have their date and other metrics use relative, nanosecond durations.

Returning server date from `DateProvider` would mean collecting events in **server time**. This is incorrect, as the server time becomes more precise while the NTP sync is being performed. If we capture server time in `tracer.startSpan()` and then capture it again after `200ns` in `span.finish()` those two dates will vary by more than `200ns` (the duration might be even negative!).

There are also other examples where returning server date from `DateProvider` will be wrong, including:
* batch files are named and ordered by timestamp (`Clock.now` doesn't have to return ascending timestamps while sync),
* network requests are not signed by `dateProvider.currentDate()` - we take their timing from task metrics.

> `DateCorrection.toServerDate(Date) -> Date` looks confusing to me: `DateCorrection` sounds more like a struct containing correction data and `toServerDate` looks more like an instance method for `Date` type

`DateCorrection` is a basic interface with clear responsibility:

```swift
/// Adjusts device time to server time using the time difference calculated with NTP.
internal protocol DateCorrectionType {
    /// Corrects given device time to server time using the last known time difference between the two.
    func toServerDate(deviceDate: Date) -> Date
}
```

Abstracting it with `protocol` is extremely handy for unit tests, as we can run them stable with no date correction. One integration test was added for each product to test if a known correction is applied.

buranmert (approved):
i see the point of using device time until the very end 👌 

regarding the namings, of course it's very clear for you because you wrote it :)
i see a simpler implementation in [Android](https://github.com/DataDog/dd-sdk-android/blob/master/dd-sdk-android/src/main/kotlin/com/datadog/android/core/internal/time/KronosTimeProvider.kt): if you need to get time information you immediately know where to go.

i'm approving because it works ✅ 
but there seems to be some room for improvement here

PR #322, opened 2020-11-27
RUMM-895 Add single API to configure all endpoints

ncreated:
### What and why?

📦 This PR introduces new APIs to configure Logs / Traces / RUM endpoint:
```swift
.set(endpoint: DatadogEndpoint) // .us | .eu | .gov
.set(customLogsEndpoint: URL)
.set(customTracesEndpoint: URL)
.set(customRUMEndpoint: URL)
```
It also marks old APIs as deprecated:
```swift
.set(logsEndpoint:)   // .us | .eu | .gov | .custom(...)
.set(tracesEndpoint:) // .us | .eu | .gov | .custom(...)
.set(rumEndpoint:)    // .us | .eu | .gov | .custom(...)
```

Old APIs were inconsistent: it made no sense to configure one feature for `.us` and another for `.eu`, because clientToken / RUM application ID are valid in only one server instance. Because default value was `.us`, this was leading to misconfigurations for clients who only set it to `.eu` for one product. 

### How?

Because we still need to support old APIs (while they are deprecated), following rules were implemented and tested:
* if `DatadogEndpoint` is set, it overwrites values set for `LogsEndpoint`, `TracesEndpoint` and `RUMEndpoint`,
* if `custom<feature>Endpoint` is set, it overwrites the value for that feature endpoint (including the value coming from `DatadogEndpoint`).

I also updated configuration tests in the way that makes it easier to remove old logic in the future.

### Review checklist


Sources/Datadog/Core/FeaturesConfiguration.swift
```diff
@@ -82,6 +82,33 @@ extension FeaturesConfiguration {
         var rum: RUM?
         var urlSessionAutoInstrumentation: URLSessionAutoInstrumentation?
 
+        var logsEndpoint = configuration.logsEndpoint
```
mariusc83, line 85: Hmmm...I guess we should add a task also in Android to align the APIs. @xgouchet what do you think ?

mariusc83 (approved):
LGTM

buranmert (approved):

Sources/Datadog/Core/FeaturesConfiguration.swift
ncreated: FTR: It's already there in the Android SDK 👌.

PR #323, opened 2020-11-30
RUMM-776 Add custom timings API for RUM Views

ncreated:
### What and why?

📦 This PR introduces custom timings API for RUM:
```swift
Global.rum.addTiming(name: "content-ready")
```
Screenshot 2020-11-30 at 15 16 12

### How?

Timing is measured as the number of nanoseconds since the start of the current view.

Timings are encoded as `view.custom_timings.<name-of-the-timing>` values in RUM event JSON. To make it work, I had to apply workaround for backend issue: we must enforce that `view.custom_timings.*` appears **after** the `view.` object (achieved by enabling `.sortedKeys` option in JSON encoder).

To test timings with mocked `RelativeDateProvider`, I had to ensure that it is not used by the storage part of the mocked feature. This is done by injecting `SystemDateProvider` when partially mocking each feature.

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/Utils/JSONEncoder.swift
+++ b/Sources/Datadog/Core/Utils/JSONEncoder.swift
@@ -15,7 +15,9 @@ extension JSONEncoder {
             try container.encode(formatted)
         }
         if #available(iOS 13.0, OSX 10.15, *) {
-            encoder.outputFormatting = [.withoutEscapingSlashes]
+            encoder.outputFormatting = [.withoutEscapingSlashes, .sortedKeys]
+        } else {
+            encoder.outputFormatting = [.sortedKeys]
         }
         return encoder
     }
--- a/Sources/Datadog/DDRUMMonitor.swift
+++ b/Sources/Datadog/DDRUMMonitor.swift
@@ -33,6 +33,14 @@ public class DDRUMMonitor {
         attributes: [AttributeKey: AttributeValue] = [:]
     ) {}
 
+    /// Adds a specific timing in the currently presented View. The timing duration will be computed as the
+    /// number of nanoseconds between the time the View was started and the time the timing was added.
+    /// - Parameters:
+    ///   - name: the name of the custom timing attribute.
+    public func addTiming(
+        name: String
+    ) {}
+
     /// Notifies that an Error occurred in currently presented View.
     /// - Parameters:
     ///   - message: a message explaining the Error.
--- a/Sources/Datadog/RUM/RUMEvent/RUMEventEncoder.swift
+++ b/Sources/Datadog/RUM/RUMEvent/RUMEventEncoder.swift
@@ -16,7 +16,7 @@ internal struct RUMEvent<DM: RUMDataModel>: Encodable {
     let userInfoAttributes: [String: Encodable]
 
     /// Custom View timings (only available if `DM` is a RUM View model)
-    let customViewTimings: [RUMViewCustomTiming]?
+    let customViewTimings: [String: Int64]?
 
     func encode(to encoder: Encoder) throws {
         try RUMEventEncoder().encode(self, to: encoder)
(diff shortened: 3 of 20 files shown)


Sources/Datadog/RUM/RUMEvent/RUMEventEncoder.swift
```diff
@@ -15,6 +15,9 @@ internal struct RUMEvent<DM: RUMDataModel>: Encodable {
     let attributes: [String: Encodable]
     let userInfoAttributes: [String: Encodable]
 
+    /// Custom View timings (only available if `DM` is a RUM View model)
+    let customViewTimings: [String: Int64]?
```
buranmert, line 19: this comment doesn't sound like it's intended.?
IMO we can add `customViewTimings` via an `RUMView` extension
```swift
extension RUMView {
  func addCustomTimingAttribute(_ timing) { self.attributes["view.custom_timings.\(timing.name)"] = timing.value }
  // ...and if needed
  var customViewTimings { return attributes.filter { $0.startsWith("view.custom_timings") } }
}
```

Tests/DatadogTests/Datadog/Core/Utils/JSONEncoderTests.swift
```diff
@@ -29,4 +29,26 @@ class JSONEncoderTests: XCTestCase {
...
+
+    func testWhenEncoding_thenKeysFollowLexicographicOrder() throws {
+        struct Foo: Codable {
+            var one = 1
+            var two = 1
+            var three = 1
+            var four = 1
+
+            enum CodingKeys: String, CodingKey {
+                case one = "aaaaaa"
```
buranmert, line 41: i think our use case is keys with `.`, no? if we add keys with `.` here, maybe we can catch potential issues in next iOS versions

Sources/Datadog/Core/Utils/JSONEncoder.swift
```diff
@@ -15,7 +15,9 @@ extension JSONEncoder {
             try container.encode(formatted)
         }
         if #available(iOS 13.0, OSX 10.15, *) {
-            encoder.outputFormatting = [.withoutEscapingSlashes]
+            encoder.outputFormatting = [.withoutEscapingSlashes, .sortedKeys]
```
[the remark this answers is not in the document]
ncreated: Indeed, done 👌

Sources/Datadog/DDRUMMonitor.swift
```diff
@@ -33,6 +33,14 @@ public class DDRUMMonitor {
         attributes: [AttributeKey: AttributeValue] = [:]
     ) {}
 
+    /// Adds a specific timing in the currently presented View. The timing duration will be computed as the
+    /// number of nanoseconds between the time the View was started and the time the timing was added.
+    /// - Parameters:
+    ///   - name: the name of the custom timing attribute.
```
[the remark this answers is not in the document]
ncreated: Added the note that `name` must be unique for each timing 👍.

Tests/DatadogTests/Datadog/Core/Utils/JSONEncoderTests.swift
ncreated: True 👌. Added one more variant: `aaa.aaa` to cover this.

Sources/Datadog/RUM/RUMEvent/RUMEventEncoder.swift
ncreated: `RUMView` is created by the user to configure the auto-instrumentation and `UIKitRUMViewsPredicate`, I don't find it like a good place to store custom timings. Do you mean `RUMDataView: RUMDataModel`? That one in turn is auto-generated, so I didn't want to extend it with any custom functionality (even through extension in separate file).

As all other encoding logic is done in `RUMEventEncoder`, I found it quite natural to put timings also in there. Having additional property for timings (`customViewTimings`) will come handy for data scrubbing, as it adds clear context to those values and we can make them (un-)editable as we want. If merged with `attributes`, we'd have to additionally filter them out as you suggest, so I prefer the easiest way for now.

buranmert: sorry yes, i meant `RUMDataView`
ideally this was supposed to be part of `RUMDataView` but it's not so due to limitations of JSON schema
i'd stay as close as possible to the ideal solution

but i'm resolving because we just follow android here 👌

buranmert (approved):

PR #324, opened 2020-11-30
RUMM-767 prepare iOS Bridge

xgouchet:
### What and why?

Prepare the protocols (generated from `dd-mobile-bridge`) and empty implementation of the iOS Bridge.

### DIFF
--- a/DatadogSDK.podspec
+++ b/DatadogSDK.podspec
@@ -19,6 +19,7 @@ Pod::Spec.new do |s|
   s.source = { :git => "https://github.com/DataDog/dd-sdk-ios.git", :tag => s.version.to_s }
   
   s.source_files = ["Sources/Datadog/**/*.swift",
+                    "Sources/Bridge/**/*.swift",
                     "Sources/_Datadog_Private/**/*.{h,m}",
                     "Datadog/TargetSupport/Datadog/Datadog.h"]
   s.public_header_files = "Datadog/TargetSupport/Datadog/Datadog.h"
--- a/Sources/Bridge/Bridge.swift
+++ b/Sources/Bridge/Bridge.swift
@@ -1,0 +1,29 @@
+//
+//  Bridge.swift
+//  Datadog
+//
+//  Created by Xavier Gouchet on 27/11/2020.
+//  Copyright © 2020 Datadog. All rights reserved.
+//
+
+import Foundation
+
+public enum Bridge {
+    
+  public static func getDdLogs() -> DdLogs {
+     return DdLogsImplementation()
+  }
+    
+    
+  public static func getDdRum() -> DdRum {
+     return DdRumImplementation()
+  }
+    
+  public static func getDdTrace() -> DdTrace {
+     return DdTraceImpementation()
+  }
+    
+  public static func getDdSdk() -> DdSdk {
+     return DdSdkImplementation()
+  }
+}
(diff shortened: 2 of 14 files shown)


Sources/Bridge/DdSdkImplementation.swift
```diff
@@ -0,0 +1,16 @@
...
+//  DatadogSDK
+//
+//  Created by Xavier Gouchet on 30/11/2020.
+//
+
+import Foundation
+
+class DdSdkImplementation: DdSdk {
+    func initialize(configuration: DdSdkConfiguration) {
+//        let config = Datadog.Configuration(rumApplicationID: configuration.applicationId as String?, clientToken: configuration.clientToken as String, environment: configuration.env as String, loggingEnabled: true, tracingEnabled: true, rumEnabled: true, logsEndpoint: .us, tracesEndpoint: .us, rumEndpoint: .us, rumSessionsSamplingRate: 100.0, rumUIKitActionsTrackingEnabled: false)
```
mariusc83, line 12: ?

DatadogSDK.podspec
```diff
@@ -19,6 +19,7 @@ Pod::Spec.new do |s|
   s.source = { :git => "https://github.com/DataDog/dd-sdk-ios.git", :tag => s.version.to_s }
   
   s.source_files = ["Sources/Datadog/**/*.swift",
+                    "Sources/Bridge/**/*.swift",
                     "Sources/_Datadog_Private/**/*.{h,m}",
```
ncreated, line 23: If the bridge integrates with RN SDK through `podspec`, I'd rather create a separate podspec: `DatadogBridge.podspec` and laid out the module in there. There's no point of including all the bridge files for Carthage and SPM distribution. It makes the binary and dSYMs files bigger, it adds bridge symbols into public namespace making the autocompletion harder / misleading and it impacts the code coverage for the SDK (although this depends on how we're going to test bridge's implementation).

I think this is sth that we can take care of, WDYT @buranmert ?

Sources/Bridge/Bridge.swift
```diff
@@ -0,0 +1,29 @@
+//
+//  Bridge.swift
+//  Datadog
+//
+//  Created by Xavier Gouchet on 27/11/2020.
+//  Copyright © 2020 Datadog. All rights reserved.
+//
```
ncreated, line 7: All the headers must use our license template.

ncreated (changes requested):
Nice work 👌, great to finally see the bridge code auto-generated 🚀.

However, to merge this PR, it requires additional work, which rather needs the iOS expertise. This includes:
* deciding on how we're going to lay out this module (as I noted in one of the comments, I'd separate it entirely from the Datadog SDK by isolating the module);
* we don't follow many Swift conventions in this generated code (getting `53` linter warnings, where `0` is expected);
* many of the naming / structuring conventions seem weird in Swift, maybe we can improve them in code-generator;
* the use of Objective-C types seems unobvious. I think we should first validate on the RN side if we really need `NSString` / `NSDictionary` and `@objc()`, as those will be inconvenient to work with in the bridge implementation.

DatadogSDK.podspec
buranmert: yes, sounds good 👍 
we should also exclude `Bridge` folder the project to prevent Carthage artifact from having them

buranmert:
as discussed in Slack: we will move the bridge to a separate repo, this will close this PR

PR #328, opened 2020-12-02
RUMM-830 Add consent provider

ncreated:
### What and why?

📦 As the next part of tracking consent feature, this PR introduces the `ConsentProvider`.

### How?

`ConsentProvider` implements basic subscription model. The `ConsentAwareDataWriter` reads the consent value and subscribes to its updates.

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/Persistence/Privacy/ConsentAwareDataWriter.swift
+++ b/Sources/Datadog/Core/Persistence/Privacy/ConsentAwareDataWriter.swift
@@ -7,9 +7,9 @@
 import Foundation
 
 /// File writer which writes data to different folders depending on the tracking consent value.
-internal class ConsentAwareDataWriter: FileWriterType {
+internal class ConsentAwareDataWriter: FileWriterType, ConsentSubscriber {
     /// Queue used to synchronize reads and writes for the feature.
-    /// TODO: RUMM-777 will be used synchronize `activeFileWriter` swaps on consent change.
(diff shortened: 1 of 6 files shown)


Sources/Datadog/Core/Persistence/Privacy/ConsentAwareDataWriter.swift
```diff
@@ -20,23 +20,61 @@ internal class ConsentAwareDataWriter: FileWriterType {
...
+            )
+        }
+    }
+
+    // MARK: - Private
+
+    private func synchronized(block: () -> Void) {
+        objc_sync_enter(self)
+        block()
+        objc_sync_exit(self)
```
buranmert, line 65: this is totally fine 👍 i'm just wondering why we didn't use `self.queue`?

buranmert (approved):

Sources/Datadog/Core/Persistence/Privacy/ConsentAwareDataWriter.swift
ncreated: It's because in this PR the `FileWriter` is still async. It dispatches writes asynchronously to this `queue`. Making `self.queue.async {}` in this PR will lead to doubling the async call to `queue` (`queue.async { queue.async { /* write */ } }`).

But I'm aware of this and this was changed in #330 where `queue` is used instead of `objc_sync`. In #330 the `FileWriter` is changed to be synchronous, which provides simpler threading model.

PR #329, opened 2020-12-02
CIAPP-509 Support managing context through environment variables.

nachoBonafonte:
### What and why?
The idea is that the app that dd-sdk-ios can use an injected context through environment variables so all the sdk spans and logs can refer to this "parent context"

With this PR swift-testing sdk it will use this injected context to make all dd-sdk traces and logs to appear as part of the test that is currently running

### How?

This injected context is accessed through `Global.environmentContext` and if it exists it acts as the root context for all logs and traces

### Review checklist

- [X ] Feature or bugfix MUST have appropriate tests (unit, integration)

### DIFF
--- a/Sources/Datadog/FeaturesIntegration/EnvironmentSpanIntegration.swift
+++ b/Sources/Datadog/FeaturesIntegration/EnvironmentSpanIntegration.swift
@@ -1,0 +1,49 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+internal struct EnvironmentSpanIntegration {
+    /// Tracing context read from environment variables if injected
+    internal static var environmentContext: (spanID: String, traceID: String)? {
+        guard let spanIDValue = ProcessInfo.processInfo.environment[TracingHTTPHeaders.parentSpanIDField] ,
+              let traceIDValue = ProcessInfo.processInfo.environment[TracingHTTPHeaders.traceIDField] else {
+            return nil
+        }
+        return (spanIDValue, traceIDValue)
+    }
+}
+
+internal struct TracingWithEnvironmentSpanIntegration {
+    /// Span context to be attached to all root spans in order to continue the trace
+    /// injected by external process through ENV variables.
+    var environmentSpanContext: (spanID: TracingUUID, traceID: TracingUUID)? {
+        guard let environmentContext = EnvironmentSpanIntegration.environmentContext,
+              let spanID = UInt64(environmentContext.spanID).flatMap({ TracingUUID(rawValue: $0) }),
+              let traceID = UInt64(environmentContext.traceID).flatMap({ TracingUUID(rawValue: $0) }) else {
+            return nil
+        }
+        return (spanID, traceID)
+    }
+}
+
+internal struct LoggingWithEnvironmentSpanIntegration {
+    struct Attributes {
+        static let spanID = "dd.span_id"
+        static let traceID = "dd.trace_id"
+    }
+    /// Additional log attributes describing the ENV span injected by external process.
+    /// Adding those attributes to `Log` will correlate them with that span.
+    var environmentSpanAttributes: [String: Encodable]? {
+        guard let environmentContext = EnvironmentSpanIntegration.environmentContext else {
+            return nil
+        }
+        return [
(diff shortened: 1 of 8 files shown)


Sources/Datadog/FeaturesIntegration/LoggingWithActiveSpanIntegration.swift
```diff
@@ -16,17 +16,22 @@ internal struct LoggingWithActiveSpanIntegration {
...
             ]
+        } else if let environmentContext = Global.environmentContext {
+            return [
+                Attributes.traceID: "\(environmentContext.traceID.rawValue)",
+                Attributes.spanID: "\(environmentContext.spanID.rawValue)"
+            ]
         } else {
+            if tracer == nil {
+                userLogger.warn("Tracing feature is enabled, but no `Tracer` is registered. The Tracing integration with Logging will not work.")
+            }
```
ncreated, line 34: I wouldn't mix the ENV span context with `LoggingWithActiveSpanIntegration` as both mean something conceptually different. The first one is immutable (never changes) span context passed from external process, while the latter is a mutable active span generated within the process (it does change).

I think it would be way easier and more maintainable if we just expose two separate integrations (like we did for `LoggingWithRUMContextIntegration` and `TracingWithRUMContextIntegration`):
```swift
struct TracingWithEnvironmentSpanIntegration {
   /// Span context to be attached to all root spans in order to continue the trace
   /// injected by external process through ENV variables.
   let spanContext: (spanID: TracingUUID, traceID: TracingUUID)

   init?() {
      // return an instance if ENV parameters are available
      // return `nil` otherwise
   }
}

struct LoggingWithEnvironmentSpanIntegration {
   /// Additional log attributes describing the ENV span injected by external process.
   /// Adding those attributes to `Log` will correlate them with that span.
   let environmentSpanAttributes: [String: Encodable]

   init?() {
      // return an instance if ENV parameters are available
      // return `nil` otherwise
   }
}
```

Both in `Logger` and `Tracer` this can be laid off just like any other integration:
```swift
// Logger

/// Integration with Span context injected by environment. `nil` if not available.
internal let environmentSpanIntegration = LoggingWithEnvironmentSpanIntegration()

// Usage:
if let activeSpanAttributes = activeSpanIntegration?.activeSpanAttributes {
   combinedInternalAttributes.merge(activeSpanAttributes) { $1 }
} else if let environmentSpanIntegration = environmentSpanIntegration?.environmentSpanAttributes {
   combinedInternalAttributes.merge(environmentSpanAttributes) { $1 }
}
```
```swift
// Tracer

/// Integration with Span context injected by environment. `nil` if not available.
internal let environmentSpanIntegration = TracingWithEnvironmentSpanIntegration()

// Usage:
traceID: parentSpanContext?.traceID ?? (environmentSpanIntegration?.spanContext.traceID ?? tracingUUIDGenerator.generateUnique()),
spanID: tracingUUIDGenerator.generateUnique(),
parentSpanID: parentSpanContext?.spanID ?? environmentSpanIntegration?.spanContext.spanID,
```

IMO this would be more streamlined implementation, which isolates the entire thing to separate units and describes each place very well with comments. Also, it keeps equal level of abstraction in both `Logger` and `Tracer`, as both just install the integration and capture parameters from it.

We can also get rid of `Global. environmentContext`, if we capture it inside each integration (or even reuse one integration within another or decouple common thing to a third integration as we did in `RUMContextIntegration`).

ncreated (commented):
This looks like a nice feature 👌🙂. I think the integration is correct, but IMO we can improve it by aligning more with the `*Integration` concept that we already use in `Tracer` and `Logger`. I described the idea in the comment.

nachoBonafonte:
I have added the changes with the ideas indicated previously, with the exception that it takes into account that environment span and trace can change during the life of the application, so cannot be initialised only once at the beginning. The use case here is when the app is running unit tests, the app is launched only once, but the tests change (each test set its own span context).

Sources/Datadog/FeaturesIntegration/EnvironmentSpanIntegration.swift
```diff
@@ -0,0 +1,49 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+internal struct EnvironmentSpanIntegration {
```
ncreated, line 9: I think this could be `fileprivate` 💭.

ncreated (approved):
With one small comment, LGTM 👌

PR #334, opened 2020-12-08
RUMM-832 Only authorized data can be uploaded

ncreated:
### What and why?

📦 As part of tracking consent feature, this PR makes sure that only authorized data can be uploaded. This means:
* all data collected with `.granted` consent,
* all data collected with `.pending` consent if it was then changed to `.granted`.

### How?

As in `dd-sdk-android`, `DataMigrator` type is introduced. There are two migrators implemented:
* `DeleteAllDataMigrator` - deletes all data in given folder,
* `MoveDataMigrator` - moves all data from source directory do destination folder.

Migrators are resolved in `ConsentAwareDataWriter` and their `migrate()` work is dispatched to the read/write queue, which keeps all writes (on collection) and reads (on upload) synchronized on the background thread.

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/Persistence/Files/Directory.swift
+++ b/Sources/Datadog/Core/Persistence/Files/Directory.swift
@@ -55,7 +55,7 @@ internal struct Directory {
     }
 
     /// Moves all files from this directory to `destinationDirectory`.
-    func moveAllFilesTo(_ destinationDirectory: Directory) throws {
+    func moveAllFiles(to destinationDirectory: Directory) throws {
         try files().forEach { file in
             let destinationFileURL = destinationDirectory.url.appendingPathComponent(file.name)
             try? FileManager.default.moveItem(at: file.url, to: destinationFileURL)
--- a/Sources/Datadog/Core/Persistence/Files/Directory.swift
+++ b/Sources/Datadog/Core/Persistence/Files/Directory.swift
@@ -43,6 +43,24 @@ internal struct Directory {
             .contentsOfDirectory(at: url, includingPropertiesForKeys: [.isRegularFileKey, .canonicalPathKey])
             .map { url in File(url: url) }
     }
+
+    /// Deletes all files in this directory.
+    func deleteAllFiles() throws {
+        // Instead of iterating over all files and removing them one by one, we create a temporary
+        // empty directory and replace source directory content with (empty) temporary folder.
+        // This makes the deletion atomic, and is more performant in benchmarks.
+        let temporaryDirectory = try Directory(withSubdirectoryPath: "com.datadoghq/\(UUID().uuidString)")
+        _ = try FileManager.default.replaceItemAt(url, withItemAt: temporaryDirectory.url)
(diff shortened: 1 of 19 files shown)


Sources/Datadog/Core/Persistence/Files/Directory.swift
```diff
@@ -43,6 +43,24 @@ internal struct Directory {
...
+        let temporaryDirectory = try Directory(withSubdirectoryPath: "com.datadoghq/\(UUID().uuidString)")
+        _ = try FileManager.default.replaceItemAt(url, withItemAt: temporaryDirectory.url)
+        try? FileManager.default.removeItem(at: temporaryDirectory.url)
+    }
+
+    /// Moves all files from this directory to `destinationDirectory`.
+    func moveAllFiles(to destinationDirectory: Directory) throws {
+        try files().forEach { file in
+            let destinationFileURL = destinationDirectory.url.appendingPathComponent(file.name)
+            try? FileManager.default.moveItem(at: file.url, to: destinationFileURL)
```
buranmert, line 61: i guess there is no real solution to this but assuming this `try` will throw sometimes (similar to that extra `,` problem), maybe we can retry the unsuccessful move operations once again, wdyt?

buranmert (commented):
LGTM actually, i just left an open question regarding file operations

Sources/Datadog/Core/Persistence/Files/Directory.swift
mariusc83: We're actually doing this in Android. I think we retry 3 times.

Tests/DatadogBenchmarkTests/DataStorage/DataMigratorBenchmarkTests.swift
```diff
@@ -0,0 +1,57 @@
...
+
+    func testMigrating500FilesWithDeleteAllDataMigrator() throws {
+        let directory = try Directory(withSubdirectoryPath: UUID().uuidString)
+        defer { directory.delete() }
+
+        measure(metrics: benchmarkMetrics, options: benchmarkOptions) {
+            createFiles(in: directory, count: 500)
+
+            self.startMeasuring()
+            let migrator = DeleteAllDataMigrator(directory: directory)
```
mariusc83, line 28: you do not want to create this before measuring ?

Sources/Datadog/Core/Persistence/Files/Directory.swift
ncreated: Yup, done ✅. We're now retrying the entire `moveAllFiles()` up to 3 times and retrying each `moveItem(at:to:)` up to 3 times as well.

Tests/DatadogBenchmarkTests/DataStorage/DataMigratorBenchmarkTests.swift
ncreated: I've been considering this carefully - creation of migrator is part of the actual migration work, so I consider it rather correct to be included in the benchmarked scope.

mariusc83 (approved):
LGTM

PR #335, opened 2020-12-08
RUMM-833 Add public API for Data Tracking Consent

ncreated:
### What and why?

📦 This PR adds the public API for configuring tracking consent:
```swift
Datadog.initialize(
   appContext: ...,
   trackingConsent: .pending,
   configuration: ....
)

Datadog.set(trackingConsent: .granted)
```

### How?

New public APIs were introduced both for `Datadog` and `DatadogObjc`. 

The previous variant of `Datadog.initialize()` is marked deprecated.

I also added tests for `Logger`, `Tracer` and `RUMMonitor` to assert they respect the consent value when collecting data.

### Review checklist

### DIFF
--- a/Sources/Datadog/Datadog.swift
+++ b/Sources/Datadog/Datadog.swift
@@ -49,6 +49,8 @@ public class Datadog {
     @available(*, deprecated, message: """
     This method is deprecated and uses the `TrackingConsent.granted` value as a default privacy consent.
     This means that the SDK will start recording and sending data immediately after initialisation without waiting for the user's consent to be given.
+
+    Use `Datadog.initialize(appContext:trackingConsent:configuration:)` and set consent to `.granted` to preserve previous behaviour.
     """)
     public static func initialize(appContext: AppContext, configuration: Configuration) {
         initialize(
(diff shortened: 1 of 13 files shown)


Sources/Datadog/Datadog.swift
```diff
@@ -96,18 +123,29 @@ public class Datadog {
         )
     }
 
+    /// Sets the tracking consent regarding the data collection for the Datadog SDK.
+    /// - Parameter trackingConsent: new consent value, which will be applied for all data collected from now on
+    public static func set(trackingConsent: TrackingConsent) {
```
buranmert, line 128: sorry if we already discussed that, i might've forgotten
wouldn't it be handy to give the user a getter as well? it can be useful for debugging for example

ncreated: AFAIK, we've already discussed that with @xgouchet and @mariusc83 - we don't see any good use case for a getter. This setting is known on the user side and may be easily persisted there (even in memory). Adding it to the SDK costs additional threading consideration.

If `TrackingConsent` becomes problematic in debugging, we could rather include it in logs printed in `Debug.verbosity = .debug`. In any way, I'd rather wait for a real request / use case.

buranmert: if already discussed -> resolved

buranmert (approved):

PR #337, opened 2020-12-10
RUMM-922 Do not start another RUM View if `path` did not change

ncreated:
### What and why?

📦 This PR addresses the feedback learned when watching the user's RUM views predicate instrumentation.

The user was using such `UIKitRUMViewsPredicate `:

```swift
class Predicate: UIKitRUMViewsPredicate {
    func rumView(for viewController: UIViewController) -> RUMView? {
        return RUMView(path: "single-view")
    }
}
```

This leads to very bad and counter-intuitive results. For every new view controller captured by the RUM auto instrumentation, the SDK starts new RUM View:

Screenshot 2020-12-10 at 11 31 29

The user's expectation is to have only one single view for this session and group all actions and resources under it.

This PR changes the behaviour of SDK so that it does not start another RUM View, when the `view.path` retrieved from predicate does not change. It gives expected result:

Screenshot 2020-12-10 at 11 33 22

### How?

Basic check was installed in `UIKitRUMViewsHandler` to not send start view commands if `path` does not change.

#### Alternative: ignoring `commands` in `RUMViewScope`

As an alternative, we could also compare `command.path` in `RUMViewScope` (with `scope.path`) and ignore if it's equal. This would change the behaviour for both manual and auto-instrumentation. I didn't pick this solution, as IMO it is too strict. IMO this manual instrumentation code:
```swift
Global.rum.startView(viewController: vc1, path: "path1")
Global.rum.startView(viewController: vc2, path: "path1")
```
shows a clear user's intent to start another RUM View with the same `path` name.

### Review checklist

### DIFF
--- a/Sources/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsHandler.swift
+++ b/Sources/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsHandler.swift
@@ -66,9 +66,11 @@ internal class UIKitRUMViewsHandler: UIKitRUMViewsHandlerType {
 
     /// The `UIViewController` indicating the active `RUMView`.
     private weak var lastStartedViewController: UIViewController?
+    /// The last started `RUMView`.
+    private var lastStartedRUMView: RUMView?
 
     private func startIfNotStarted(rumView: RUMView, for viewController: UIViewController) {
-        if viewController === lastStartedViewController {
+        if viewController === lastStartedViewController || rumView.path == lastStartedRUMView?.path {
             return
         }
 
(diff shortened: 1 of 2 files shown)


ncreated:
Closing this PR. As we concluded on daily meeting, this solves the problem for individual user, but may be ambiguous for others. It is still possible to achieve single view auto instrumentation by implementing it explicitly in predicate (e.g. return `RUMView` only once, then continue with returning `nil`).

This particular user was trying single view, as there were dozen of view controller in the app. This problem is solved by `DefaultUIKitRUMViewsPredicate`.

buranmert:
maybe we can use default values like `trackUIKitRUMViews(using predicate: UIKitRUMViewsPredicate = DefaultUIKitRUMViewsPredicate())` in order to save our users from dealing with `Predicate`s as it seems like `Predicate` concept might be hard to understand

PR #350, opened 2020-12-17
RUMM-917 RUM for ObjC

### LINKED ISSUE #332: Objective-C interop for RUM functionality
RUM functionality is in beta, but it's only accessible using the Swift code. The Objective-C interop framework doesn't have anything relating to RUM in it.
I don't know if this is on the roadmap, but it would be great to have this added to the DatadogObjC framework, or otherwise have the Swift RUM code exposed to Objective-C.

buranmert:
### What and why?

📦 This PR adds public RUM interfaces for `DatadogObjc`. Closes #332 

### How?

New public APIs were introduced to `DatadogObjc.framework`:
```diff
+ public class DDConfiguration: NSObject
+   public static func builder(rumApplicationID: String, clientToken: String, environment: String) -> DDConfigurationBuilder
+   public func enableRUM(_ enabled: Bool)
+   public func set(customRUMEndpoint: URL)
+   public func set(rumSessionsSamplingRate: Float)
+   public func trackUIKitRUMViews()
+   public func trackUIKitRUMViews(using predicate: DDUIKitRUMViewsPredicate)
+   public func trackUIKitActions()
+ 
+ public class DDGlobal: NSObject
+   public static var sharedTracer: DatadogObjc.OTTracer = noopTracer
+   public static var rum: DatadogObjc.DDRUMMonitor = noopRUMMonitor
+ 
+ public class DDRUMView: NSObject
+   public var path: String
+   public var attributes: [String: Any]
+   public init(path: String, attributes: [String: Any])
+ public protocol DDUIKitRUMViewsPredicate: AnyObject
+   func rumView(for viewController: UIViewController) -> DDRUMView?
+ public enum DDRUMErrorSource: Int
+   case source
+   case network
+   case webview
+   case custom
+ public enum DDRUMUserActionType: Int
+   case tap
+   case scroll
+   case swipe
+   case custom
+ public class DDRUMMonitor: NSObject
+   override public convenience init()
+   public func startView(viewController: UIViewController,path: String?,attributes: [String: Any])
+   public func stopView(viewController: UIViewController,attributes: [String: Any])
+   public func addTiming(name: String)
+   public func addError(error: Error,source: DDRUMErrorSource,attributes: [String: Any])
+   public func startResourceLoading(resourceKey: String,request: URLRequest,attributes: [String: Any])
+   public func startResourceLoading(resourceKey: String,url: URL,attributes: [String: Any])
+   public func addResourceMetrics(resourceKey: String,metrics: URLSessionTaskMetrics,attributes: [String: Any])
+   public func stopResourceLoading(resourceKey: String,response: URLResponse,size: Int64?,attributes: [String: Any])
+   public func stopResourceLoadingWithError(resourceKey: String,error: Error,response: URLResponse?,attributes: [String: Any])
+   public func stopResourceLoadingWithError(resourceKey: String,errorMessage: String,response: URLResponse?,attributes: [String: Any])
+   public func startUserAction(type: DDRUMUserActionType,name: String,attributes: [String: Any])
+   public func stopUserAction(type: DDRUMUserActionType,name: String?,attributes: [String: Any])
+   public func addUserAction(type: DDRUMUserActionType,name: String,attributes: [String: Any])
+   public func addAttribute(forKey key: String,value: Any)
+   public func removeAttribute(forKey key: String)
+ public class DDTracer: NSObject, DatadogObjc.OTTracer
+   public convenience init(configuration: DDTracerConfiguration)
```

### Review checklist


Sources/DatadogObjc/RUM+objc.swift
```diff
@@ -0,0 +1,219 @@
...
+        case .swipe: return .swipe
+        case .custom: return .custom
+        default: return .custom
+        }
+    }
+}
+
+@objcMembers
+public class DDRUM: NSObject {
+    let sdkRUM: DDRUMMonitor
```
ncreated, line 87: We can keep uniformed naming in obj-c public API: `DDRUMMonitor` if we namespace the Swift one:
```swift
public class DDRUMMonitor: NSObject {
   let sdkRUMMonitor: Datadog.DDRUMMonitor
}
```

Sources/DatadogObjc/OpenTracing/OTGlobal+objc.swift
```diff
@@ -7,19 +7,38 @@
...
+    public internal(set) static var tracer: OTTracer = noopTracer {
+        didSet {
+            // We must also set the Swift `Global.tracer`
+            // as it's used internally by auto-instrumentation feature.
+            if let ddTracer = tracer.dd?.swiftTracer {
+                Global.sharedTracer = ddTracer
+            }
+        }
+    }
+    public internal(set) static var rum = DDRUM(isNoOp: true) {
```
ncreated, line 21: `DDRUM(isNoOp: true)` looks very weird. Why not just providing no-op implementation in `DatadogObjc` as we did for tracer (`let noopTracer: OTTracer = DDNoopTracer()`)?

ncreated (changes requested):

Sources/Datadog/DDRUMMonitor.swift
```diff
@@ -198,5 +198,6 @@ public class DDRUMMonitor {
 }
 
 /// The no-op variant of `DDRUMMonitor`.
-internal class DDNoopRUMMonitor: DDRUMMonitor {
+public class DDNoopRUMMonitor: DDRUMMonitor {
```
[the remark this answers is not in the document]
ncreated: IMO we don' need to expose it if we provide no-op implementation in `DatadogObjc` as suggested in [this comment](https://github.com/DataDog/dd-sdk-ios/pull/350#discussion_r546587739).

Sources/DatadogObjc/RUM+objc.swift
buranmert: there already is a `DDRUMMonitor` in Swift target, i used `DDRUM` to avoid confusion

Sources/Datadog/DDRUMMonitor.swift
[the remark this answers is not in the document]
buranmert: right 👍 we can start with default `Global.rum` without exposing `NoOpMonitor`

Sources/DatadogObjc/OpenTracing/OTGlobal+objc.swift
buranmert: (replied in the comment above ☝️ )

buranmert:
**feel free to proceed when i'm off** since i don't know what to do with unit tests

i can inject a `MockNativeRUM` into `DDRUM` and test if correct methods are called in correct places
but `DDLogger/TracerTests` are not done this way and i guess i must follow them as conventions.

when i look at those examples, they test the whole feature (until encoded output) as `Logger/TracerTests` do in Swift SDK; so it makes sense to copy all the cases `Logger/TracerTests` from Swift SDK tests as they are supposed to cover the same scenarios. but this is not what conventions do neither.

ncreated:
I continued this PR. Started with adding necessary missing APIs:
```diff
+ public class DDConfiguration: NSObject
+    public static func builder(rumApplicationID: String, clientToken: String, environment: String) -> DDConfigurationBuilder
+    public func enableRUM(_ enabled: Bool)
+    public func set(customRUMEndpoint: URL)
+    public func set(rumSessionsSamplingRate: Float)
```
then added unit tests, reaching 100% coverage for Objc RUM (~90% in total for `DatadogObjc`).

> i can inject a MockNativeRUM into DDRUM and test if correct methods are called in correct places
> but DDLogger/TracerTests are not done this way and i guess i must follow them as conventions.

> when i look at those examples, they test the whole feature (until encoded output) as Logger/TracerTests do in Swift SDK; so it makes sense to copy all the cases Logger/TracerTests from Swift SDK tests as they are supposed to cover the same scenarios. but this is not what conventions do neither.

I agree, there's a glitch. Keep in mind that the existing methodology was developed for Logging product and `DDLogger` (objc) tests are pretty straightforward. It doesn't scale well to RUM, because RUM is state-full (vs Logging being state-less), so we need to do much more effort to write the scenario for using particular APIs (e.g. when testing Resource APIs, we must also use View API to start the view). 

Shaping tests around _"When Objc X API is called, it forwards to Swift X API"_ would be much more practical 👍.

To not block this topic, I added tests using the previous approach (making sure that 100% of the public API is covered). To improve this testing methodology further, I updated the `RUMM-896` JIRA, so we can came up with unified testing methodology for `dd-sdk-ios` and do necessary refactoring.

Sources/DatadogObjc/Tracer+objc.swift
```diff
@@ -11,7 +11,8 @@ import struct Datadog.OTReference
 import class Datadog.HTTPHeadersWriter
 
 @objcMembers
-public class DDTracer: DatadogObjc.OTTracer {
+public class DDTracer: NSObject, DatadogObjc.OTTracer {
```
ncreated, line 14: @buranmert this change of adding `NSObject` subclass for `DDTracer` fixed critical issue with DatadogObjc tracing. Without `NSObject` subclassing the entire `DDTracer` was not visible to Objective-C code (⚠️).

It means that `OTGlobal+objc` API was impossible to use 😞 as `DDTracer` was never visible. Because of this, IMO we can just remove it without deprecating, WDYT?

buranmert: oh that's bad, then yes we can just remove it 👍 
as i said in Slack: we are not really testing generated ObjC interface, we should be calling our SDK from ObjC somehow in tests to ensure that, wdyt?

ncreated: This would be cool and would help us avoid such mistakes as this one. But it seems difficult at the first glance to build Objc unit tests, as the assertions would need to call non-bridged Swift code. Anyway, I think we should consider this in `RUMM-896` 👌.

Sources/DatadogObjc/Global+objc.swift
```diff
@@ -0,0 +1,30 @@
...
+    public static var sharedTracer: DatadogObjc.OTTracer = noopTracer {
+        didSet {
+            // We must also set the Swift `Global.tracer`
+            // as it's used internally by auto-instrumentation feature.
+            if let swiftTracer = sharedTracer.dd?.swiftTracer {
+                Global.sharedTracer = swiftTracer
+            }
+        }
+    }
+    public static var rum: DatadogObjc.DDRUMMonitor = noopRUMMonitor {
```
buranmert, line 21: if `DatadogObjc` is just a translation layer then we should somehow use the default Swift `RUMMonitor` here.
if the default Swift monitor changes, default Objc monitor should follow the change: this code doesn't follow, it creates Objc's own default value

Sources/DatadogObjc/RUMMonitor+objc.swift
```diff
@@ -0,0 +1,221 @@
...
+    }
+}
+
+internal let noopRUMMonitor = DatadogObjc.DDRUMMonitor(swiftRUMMonitor: nil)
+
+@objcMembers
+public class DDRUMMonitor: NSObject {
+    // MARK: - Internal
+
+    internal let swiftRUMMonitor: Datadog.DDRUMMonitor?
```
buranmert, line 97: we don't have any use case with `swiftRUMMonitor == nil`, we have `NoOpRUMMonitor` instead.

```swift
internal let noopRUMMonitor = DatadogObjc.DDRUMMonitor(swiftRUMMonitor: nil)
```
the line above doesn't need to exist, we can do `DDRUMMonitor(swiftRUMMonitor: NoOpMonitor())`

Sources/DatadogObjc/Global+objc.swift
ncreated: `DatadogObjc.DDRUMMonitor` (Objc) translates its calls to `Datadog.RUMMonitor` (Swift). Both Swift and Objc SDKs need to provide its `Global` namespace which uses no-op implementation of monitor by default.  This is exactly what this code does and how it reads. I don't see how this can be done differently 🤔.

Sources/DatadogObjc/RUMMonitor+objc.swift
ncreated: `NoOpRUMMonitor` is an internal type in Datadog (Swift) module. It can't be used here in DatadogObjc (Objc) module. To provide no-op implementation of `DatadogObjc.DDRUMMonitor` we just create it without providing Swift monitor. All calls to it result with no operation.

Sources/DatadogObjc/Global+objc.swift
buranmert: both Swift and Objc don't need to provide their own `NoOp`s; Objc can use Swift's `NoOp` without duplication.
plus, Objc doesn't need to decide what default to use (`NoOp` or `FooOp` or etc.?), it should use Swift's default.

you can do this like this:
```swift
static var rum = ObjC.DDRUMMonitor(swiftRUMMonitor: Global.rum)
```
now Objc gets whatever Swift gives by default

ncreated: Sounds good 👍. I removed the no-op RUM monitor and no-op Tracer from `DatadogObjc`. Now they use whatever Swift SDK exposes as `Global.*`.

buranmert (commented):
LGTM ✅  but Github doesn't let me approve _my own PR_ 😄 
you can approve and self-merge 👌

ncreated (approved):

PR #351, opened 2020-12-21
RUMM-914 Improve RUM models generation

ncreated:
### What and why?

📦 With the upcoming data scrubbing feature, we need more control over our `RUMDataModels`. This PR brings a command-line utility for generating RUM models the way we want. It now implements `Swift` output, but it can be easily extended with `Objc` output later.

### How?

I created a small Swift command-line tool:
```sh
$ rum-models-generator generate-swift --path <json-schemas-path>
```
it prints generated models to standard output. Later (in next PR) this output will be piped into `RUMDataModels.swift` file and necessary changes will be done in our SDK code (update to new models).

The `RUMModelsGenerator` implements `read → transform → print` architecture and uses ternary representation of data:
* `JSONSchema` - deserialisable representation of [JSON Schema](https://json-schema.org/draft/2019-09/json-schema-core.html) (limited to what we need for `rum-events-format`).
* `JSONType` - a type-safe representation of `JSONSchema`.
* `SwiftType` - a type-safe representation of Swift schema.

Data flow is done as follow:
```swift
(JSON file) → JSONSchemaReader → JSONSchema → JSONTypeReader → JSONType → SwiftTypeReader → SwiftType
```
Before the resultant `SwiftType` is passed to the `SwiftPrinter` (which prints Swift syntax), it gets transformed by `RUMSwiftTypeTransformer`:
```swift
SwiftType → RUMSwiftTypeTransformer → SwiftPrinter
```

`RUMSwiftTypeTransformer` is the only place where we apply customisations for RUM, e.g.: renaming properties to "camelCase", mapping `Int` primitives to `Int64`, adding `RUMDataModel` conformance.

### Tests

This tool was tested with few last versions of the `rum-events-format` schema. The recent schemas are included in unit tests as a fixture along with the expected output (`RUMDataModels.swift`). That's why this PR is so big.

### Coming in next PR

There is one more customisation that we may need. All models nest common definition of `Connectivity` and `USR`, e.g. `RUMErrorEvent.Connectivity`, `RUMResourceEvent.Connectivity`. The `connectivity` is entirely defined in `_common-schema.json`, so can be generated as top-level model (this will save us from duplicating mapping logic in the SDK).

Next PR will also update SDK code to use new models and will plug the tool into CI validation job.

### Review checklist


tools/rum-models-generator/Tests/rum-models-generator-coreTests/Fixtures/Output/RUMDataModels.swift
```diff
@@ -0,0 +1,1474 @@
...
+        }
+
+        /// HTTP method of the resource
+        internal enum Method: String, Codable {
+            case pOST = "POST"
+            case gET = "GET"
+            case hEAD = "HEAD"
+            case pUT = "PUT"
+            case dELETE = "DELETE"
+            case pATCH = "PATCH"
```
buranmert, line 576: the first letter should be uppercase or other letters should be lowercase

buranmert (changes requested):
> this will save us from duplicating mapping logic in the SDK

i don't understand why we don't get rid of mappings after making those models `public`?

---

currently we generate and validate models; i guess validation functionality comes in the next PR, correct?

---

we are adding 3k lines to the repo with this PR, do we have plans to extract this tool out of the repo and install it only when needed?

ruthnaebeck (approved):
👍  for docs

ncreated:
> i don't understand why we don't get rid of mappings after making those models `public`?

Our architecture marks out 4 modules: Core, Logging, Tracing and RUM. The Core module delivers `UserInfo`, `CarrierInfo` and `NetworkConnectionInfo`, used by 3 other modules:
* `Logging` uses them directly to encode in serializable `Log`,
* `Tracing` uses them directly to encode in serializable `Span`,
* `RUM` transforms `Core.UserInfo` into `RUM.UserInfo` and transforms `Core.CarrierInfo` + `Core.NetworkConnectionInfo` into `RUM.Connectivity`.

If we use `RUM.USerInfo` and `RUM.Connectivity` directly in Core module, this will violate the dependency graph in our architecture, as `Loogging`, `Tracing` and `Core` modules will depend directly on `RUM`.

Speaking on duplication:  each RUM event requires its `UserInfo` and `Connectivity`. To avoid duplicating mapping: 
* `f(Core.UserInfo) -> ViewEvent.UserInfo`,
* `f(Core.UserInfo) -> ActionEvent.UserInfo`,
* ...
* `f(Core.CarrierInfo, Core.NetworkConnectionInfo) -> ResourceEvent.Connectivity`,
* `f(Core.CarrierInfo, Core.NetworkConnectionInfo) -> ErrorEvent.Connectivity`,
* ...

we should generate single `RUM.UserInfo` and `RUM.Connectivity` and share them in all events (`ViewEvent`, `ActionEvent`, ...).

> currently we generate and validate models; i guess validation functionality comes in the next PR, correct?

Yes, that's what I mean by plugging into CI validation job.

> we are adding 3k lines to the repo with this PR, do we have plans to extract this tool out of the repo and install it only when needed?

This is only `112KB` and it adds nothing to the SDK size nor its compilation time. I don't think this is a reason to extract it and bother with separate repo versioning. Do you see any reason?

tools/rum-models-generator/Tests/rum-models-generator-coreTests/Fixtures/Output/RUMDataModels.swift
ncreated: ✅

xgouchet (changes requested):
There's no UT on the code generation script. It could be nice to add a couple of simple use case to ensure the Json Schema parser and code generation works the exact way you want it to.

ncreated:
> There's no UT on the code generation script. It could be nice to add a couple of simple use case to ensure the Json Schema parser and code generation works the exact way you want it to.

I wanted to keep this PR relatively small, so UT were pushed as part of #353 even though I did TDD. But indeed, it makes more sense to include them in this base PR, so I added the [UTs commit](https://github.com/DataDog/dd-sdk-ios/pull/351/commits/669da5aa37e61f25088043b75df2a2cd02c668f6) to this branch 👌.

buranmert (approved):

xgouchet (approved):

PR #352, opened 2020-12-22
RUMM-946 Example app target doesn't compile Xcode 12.3

buranmert:
### What and why?

Building `Example` gives

```
...../dd-sdk-ios/Datadog/Datadog.xcodeproj Building for iOS Simulator, but the linked and embedded framework 'Kronos.framework' was built for iOS + iOS Simulator.
```

Previous Xcode versions don't have this problem

### How?

Fix source: https://github.com/AFNetworking/AFNetworking/issues/4618#issuecomment-745314301

### Review checklist


ncreated (approved):
👌 Seems like this is required for all Carthage projects now, right?

buranmert:
i checked `CTProject` if the same problem happens there but it works fine without `Validate workspace`. 
honestly, i didn't understand why this problem happens but fix seems to work

PR #353, opened 2020-12-29
RUMM-914 Allow sharing types in generated RUM models

ncreated:
### What and why?

📦 This PR delivers improvement mentioned in #351. Now, particular common types can be shared in generated RUM model definitions, while other can remain event-specific.

Before, the generated code was sort of:
```swift
struct RUMResourceEvent {
   let resource: Resource
   let connectivity: Connectivity
   let usr: USR

   struct Resource { /* Resource definition */ }
   struct Connectivity { /* Resource's Connectivity definition */ }
   struct USR { /* Resource's USR definition */ }
}

struct RUMErrorEvent {
   let error: Error
   let connectivity: Connectivity
   let usr: USR

   struct Error { /* Error definition */ }
   struct Connectivity { /* Error's Connectivity definition */ }
   struct USR { /* Error's USR definition */ }
}
```

Now, it is possible to share common definitions:
```swift
struct RUMResourceEvent {
   let resource: Resource
   let connectivity: RUMConnectivity
   let usr: RUMUser

   struct Resource { /* Resource definition */ }
}

struct RUMErrorEvent {
   let error: Error
   let connectivity: RUMConnectivity
   let usr: RUMUser

   struct Error { /* Error definition */ }
}

struct RUMConnectivity { /* Shared RUM Connectivity definition */ }
struct RUMUser { /* Shared RUM User definition */ }
```

### How?

Following the `read → transform → print` architecture, I updated the transformation logic to find predefined types and detach them to the root level (`RUMConnectivity`, `RUMUser` and `RUMHTTPMethod`).

I also remove the fixture-based tests used in #351. They were asserting the entire content of generated `rum-events-format` schema, which was convenient to preview the generated code.

### Next PR

In the next PR, I will plug this tool into CI validation job and update the SDK code to use this new, generated models.

### Review checklist


tools/rum-models-generator/Sources/RUMModelsGeneratorCore/RUM/RUMSwiftTypeTransformer.swift
```diff
@@ -136,6 +147,55 @@ internal class RUMSwiftTypeTransformer: TypeTransformer<SwiftType> {
             fixedName = format(structName: parentTypeName) + fixedName
         }
 
+        if fixedName == "Connectivity" {
+            fixedName = "RUMConnectivity"
+        }
+
+        if fixedName == "USR" {
+            fixedName = "RUMUser"
+        }
```
buranmert, line 156: i don't understand why we needed these, don't we already add `RUM` prefix to top level types? if `Connectivity` and `USR` are nested types: we don't need `RUM` prefix for them, do we?

ncreated: This is because of the way `RUM*Transformer` works - it first decorates root level types (so adds `RUM` prefix), then adds additional root level types (collected in `sharedRootTypes` array):
```swift
override func transform(types: [SwiftType]) throws -> [SwiftType] {
   sharedRootTypes = []
   let transformed = try types.map { try transformAny(type: $0) }
   return transformed + sharedRootTypes
}
```

buranmert: if that was a conscious decision then it's fine 👌 
personally i would expect to manage all top level types at once, like `if topLevel then addPrefix`

ncreated: Yes, this is intentional. Although this could be improved as you propose, I'll keep it like this until we have more than just 2 shared types to decorate.

PR #355, opened 2020-12-31
RUMM-904 First party hosts are sanitized

ncreated:
### What and why?

📦 This PR adds additional check to the `firstPartyHosts` configured by the user. If the host is invalid, it may get sanitized or dropped - both with appropriate warning.

### How?

Similar to https://github.com/DataDog/dd-sdk-android/pull/447 URL, host and IP regexes were used to validate the input `host`.

Example behaviours:
```swift
"https://first-party.com", // sanitize to → "first-party.com"
"http://api.first-party.com", // sanitize to → "api.first-party.com"
"https://first-party.com/v2/api", // sanitize to → "first-party.com"
"https://192.168.0.1/api", // sanitize to → "192.168.0.1"
"https://192.168.0.2", // sanitize to → "192.168.0.2"
"invalid-host-name", // drop
"192.168.0.3:8080", // drop
"", // drop
"localhost", // accept
"192.168.0.4", // accept
"valid-host-name.com", // accept
```

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/FeaturesConfiguration.swift
+++ b/Sources/Datadog/Core/FeaturesConfiguration.swift
@@ -230,10 +230,18 @@ private func sanitized(firstPartyHosts: Set<String>) -> Set<String> {
     let hostRegex = #"^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])\.)+([A-Za-z]|[A-Za-z][A-Za-z0-9-]*[A-Za-z0-9])$"#
     let ipRegex = #"^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$"#
 
+    var warnings: [String] = []
+
     let array: [String] = firstPartyHosts.compactMap { host in
         if host.range(of: urlRegex, options: .regularExpression) != nil {
             // if an URL is given instead of the host, take its `host` part
-            return URL(string: host)?.host
+            if let sanitizedHost = URL(string: host)?.host {
+                warnings.append("'\(host)' is an url and will be sanitized to: '\(sanitizedHost)'.")
+                return sanitizedHost
+            } else {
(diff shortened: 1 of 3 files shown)


Sources/Datadog/Core/FeaturesConfiguration.swift
```diff
@@ -173,7 +173,7 @@ extension FeaturesConfiguration {
         if let firstPartyHosts = configuration.firstPartyHosts, !firstPartyHosts.isEmpty {
```
buranmert, line 173: it may make sense to not init'ing `URLSessionAutoInstrumentation` in case of empty `userDefinedFirstPartyHosts`, what do you think?

```swift
// pseudo-code
if let firstPartyHosts = sanitized(config.firstPartyHosts) { ... }
```

buranmert (approved):
lgtm, we can add one minor thing

Sources/Datadog/Core/FeaturesConfiguration.swift
ncreated: Hmm, I see 👍, but I'm not that sure 💭. 

By continuing on `URLSessionAutoInstrumentation` config we do one more validation (if RUM or Tracing are enabled):
```swift
if let firstPartyHosts = configuration.firstPartyHosts, !firstPartyHosts.isEmpty {
    if configuration.tracingEnabled || configuration.rumEnabled {
        // ... stanitize hosts ...
    } else {
        let error = ProgrammerError(
            description: """
            To use `.track(firstPartyHosts:)` either RUM or Tracing should be enabled.
            """
        )
        consolePrint("\(error)")
    }
}
```
This means that if the user has both things misconfigured, he/she will be warned in one shot, instead of two. At the same time, `firstPartyHosts` misconfiguration is not an expected state, so I wouldn't additionally optimise for it. WDYT?

buranmert: maybe it makes more sense to add this into `URLSessionAutoInstrumentation.init` method, `if userDefinedFirstPartyHosts.isEmpty { return nil }` to avoid unneeded swizzling.

as you said that shouldn't happen often though, you can merge it as it is ✅

PR #358, opened 2021-01-05
RUMM-824 Add API for customizing batch size and upload frequency

ncreated:
### What and why?

📦 This PR adds two new public APIs, so that the user can customize the size and frequency of requests made by the SDK.

### How?

The write / upload settings were already abstracted by `PerformancePreset` type, so this PR just configures its values depending on the `BatchSize` and `UploadFrequency` received from the user with two new public APIs:

```swift
public func set(batchSize: BatchSize)
public func set(uploadFrequency: UploadFrequency)
```

Because the iOS SDK provides different presets for iOS apps and iOS app extensions, the values computed for `PerformancePreset` are additionally customized to work well in short-lived app extension environment.

The `PerformancePreset` computed for iOS app is unified with the values introduced for Android SDK https://github.com/DataDog/dd-sdk-android/pull/454 and https://github.com/DataDog/dd-sdk-android/pull/453.

The new APIs are also exposed for Objc SDK.

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/PerformancePreset.swift
+++ b/Sources/Datadog/Core/PerformancePreset.swift
@@ -68,56 +68,80 @@ internal struct PerformancePreset: Equatable, StoragePerformancePreset, UploadPe
     let minUploadDelay: TimeInterval
     let maxUploadDelay: TimeInterval
     let uploadDelayChangeRate: Double
+}
 
-    // MARK: - Predefined presets
+internal extension PerformancePreset {
+    init(
+        batchSize: Datadog.Configuration.BatchSize,
+        uploadFrequency: Datadog.Configuration.UploadFrequency,
+        bundleType: BundleType
+    ) {
+        let meanFileAgeInSeconds: TimeInterval = {
+            switch (bundleType, batchSize) {
+            case (.iOSApp, .small): return 5
+            case (.iOSApp, .medium): return 15
+            case (.iOSApp, .large): return 60
+            case (.iOSAppExtension, .small): return 1
+            case (.iOSAppExtension, .medium): return 3
+            case (.iOSAppExtension, .large): return 12
+            }
+        }()
 
-    /// Default performance preset.
-    static let `default` = lowRuntimeImpact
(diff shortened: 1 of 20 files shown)


Sources/Datadog/Core/PerformancePreset.swift
```diff
@@ -68,56 +68,80 @@ internal struct PerformancePreset: Equatable, StoragePerformancePreset, UploadPe
...
-        maxObjectSize: 256 * 1_024, // 256KB
+internal extension PerformancePreset {
+    init(
+        batchSize: Datadog.Configuration.BatchSize,
+        uploadFrequency: Datadog.Configuration.UploadFrequency,
+        bundleType: BundleType
+    ) {
+        let meanFileAgeInSeconds: TimeInterval = {
+            switch (bundleType, batchSize) {
+            case (.iOSApp, .small): return 5
```
xgouchet, line 81: Out of curiosity, what's the difference between iOSApp and iOSAppExtension ?

xgouchet (approved):

buranmert (approved):

Sources/Datadog/Core/PerformancePreset.swift
buranmert: this is due to GH issue number 52
(i'm not giving the actual refererence to it on purpose, no need to alert it i guess)

ncreated: [iOS App Extensions](https://developer.apple.com/app-extensions/) are the way to add functionality beyond the app so they can interact with other apps or the system (e.g. home screen widgets). 

We differentiate `PerformancePreset` in app extension, because its process is usually designed to work only for the **very short amount of time**, when the user interacts with it, e.g. chooses the Slack channel and message to upload photo directly from the iOS photo library through Slack Extension (w/o launching the Slack app).

Such `PerformancePreset` for app extension stores and uploads data much mor frequently + its initial upload is scheduled almost right away when the app extension starts. This is to make sure the SDK has enough chance to upload all data before the extension is closed (which usually happens in few seconds after it was started).

PR #359, opened 2021-01-05
RUMM-917 Fix Objc API visibility in `RUMMonitor`

ncreated:
### What and why?

🐞 The `[DDGlobal.rum stopResourceLoadingWithResourceKey:response:attributes:]` public API was not visible for Objective-C.

### How?

This API was using `size: Int64?` parameter which can't be translated to Objective-C so it was not generated into Obj-C interface. I removed it, as we infer this information any way from the `HTTPURLResponse`.

### Review checklist


buranmert (approved):

buranmert:
ℹ️ i'm merging as @ncreated is off today

PR #360, opened 2021-01-06
RUMM-917 Non-exposed ObjC public API is fixed

buranmert:
### What and why?

@objcMembers tries to expose all members; in case of failure, fails silently.
As we don't have tests calling the API from ObjC, we don't make sure `DatadogSDKObjc` is exposed to Obj-C

### How?

@objc doesn't compile in case of failure and gives errors

`public let OTFormatHTTPHeaders = "OTFormatHTTPHeaders"` is exposed as `OT.formatTextMap`

### Review checklist

### DIFF
--- a/Sources/DatadogObjc/OpenTracing/OTTracer+objc.swift
+++ b/Sources/DatadogObjc/OpenTracing/OTTracer+objc.swift
@@ -5,11 +5,14 @@
  */
 
 import Foundation
-public let OTFormatHTTPHeaders = "OTFormatHTTPHeaders"
 
 @objc
+public class OT: NSObject {
+    @objc public static let formatTextMap = "OTFormatTextMap"
+}
+
 /// Corresponds to: https://github.com/opentracing/opentracing-objc/blob/master/Pod/Classes/OTTracer.h
-public protocol OTTracer {
+@objc public protocol OTTracer {
     func startSpan(_ operationName: String) -> OTSpan
     func startSpan(_ operationName: String, tags: NSDictionary?) -> OTSpan
     func startSpan(_ operationName: String, childOf parent: OTSpanContext?) -> OTSpan
--- a/Sources/DatadogObjc/Datadog+objc.swift
+++ b/Sources/DatadogObjc/Datadog+objc.swift
@@ -7,7 +7,7 @@
 import Foundation
 import Datadog
 
-@objcMembers
+@objc
 public class DDTrackingConsent: NSObject {
     internal let sdkConsent: TrackingConsent
 
@@ -17,27 +17,34 @@ public class DDTrackingConsent: NSObject {
 
     // MARK: - Public
 
+    @objc
     public static func granted() -> DDTrackingConsent { .init(sdkConsent: .granted) }
+
+    @objc
     public static func notGranted() -> DDTrackingConsent { .init(sdkConsent: .notGranted) }
+
+    @objc
     public static func pending() -> DDTrackingConsent { .init(sdkConsent: .pending) }
 }
 
-@objcMembers
+@objc
 public class DDAppContext: NSObject {
     internal let sdkAppContext: Datadog.AppContext
 
     // MARK: - Public
 
+    @objc
     public init(mainBundle: Bundle) {
         self.sdkAppContext = Datadog.AppContext(mainBundle: mainBundle)
     }
 
+    @objc
     override public init() {
         self.sdkAppContext = Datadog.AppContext()
     }
 }
 
-@objcMembers
+@objc
 public class DDDatadog: NSObject {
     // MARK: - Public
 
@@ -47,6 +54,7 @@ public class DDDatadog: NSObject {
 
     Use `DDDatadog.initialize(appContext:trackingConsent:configuration:)` and set consent to `granted()` to preserve previous behaviour.
     """)
+    @objc
     public static func initialize(appContext: DDAppContext, configuration: DDConfiguration) {
         Datadog.initialize(
             appContext: appContext.sdkAppContext,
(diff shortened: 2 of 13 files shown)


Sources/DatadogObjc/OpenTracing/OTTracer+objc.swift
```diff
@@ -5,11 +5,12 @@
  */
 
 import Foundation
-public let OTFormatHTTPHeaders = "OTFormatHTTPHeaders"
+// TODO: RUMM-917 this can't be exposed to ObjC but it is required in `inject(format:...)` method
+// Currently `inject` method is not usable from ObjC
+public let OTFormatHTTPHeaders = "OTFormatHTTPHeaders" // swiftlint:disable:this non_exposed_objc
```
ncreated, line 10: Constants can be exposed in the way used in `ExampleAppDelegate`: https://github.com/DataDog/dd-sdk-ios/blob/53109cfe918bc2d17ac75e95f681ce7fc82ec60b/Datadog/Example/ExampleAppDelegate.swift#L94-L98

I couldn't find a better way.

tools/lint/sources.swiftlint.yml
```diff
@@ -70,6 +68,8 @@ whitelist_rules: # we enable lint rules explicitly - only the ones listed below
   - vertical_whitespace_opening_braces
   - void_return
   - xct_specific_matcher
+  # - missing_docs # TODO: RUMM-126 enforce documentation for public APIs
+  # - attributes
```
ncreated, line 72: Why disabling this rule? It ensures the attributes format. If disabled, all following will lint, so must be code reviewed:
```swift
// The way we want:
@available(iOS 12, *)
internal class Foo {}

// The way we don't want:
@available(iOS 12, *)

internal class Foo {}

// The way we don't want:
@available(iOS 12, *) internal class Foo {}
```

xgouchet (approved):
LGTM

ncreated (changes requested):

Sources/DatadogObjc/OpenTracing/OTTracer+objc.swift
buranmert: that's what i thought too but i'm not sure of naming, what do you prefer as names?

tools/lint/sources.swiftlint.yml
buranmert: at first i disabled it becuase it doesn't allow `@objc func foo() { ... }` on the same line, otherwise it looks bad in for example `RUM+Objc.swift` where many tiny methods are stacked.

then it also comes handy for the custom rule regex, i couldn't come up with a regex which works with `@objc` on the line above. if you can find one, we can re-enable this rule 👌 

i didn't know we didn't want `@available(iOS 12, *) internal class Foo {}`. it doesn't work or is it just too long?

ncreated: Our current code style enforces the rule on attributes. We shouldn't just disable it because we want to add another custom rule that conflicts with it. We should resolve this conflict instead - this means to include newline separator in the regexp.

Sources/DatadogObjc/OpenTracing/OTTracer+objc.swift
ncreated: This particular constant is named by its counterpart in [opentracing-objc](https://github.com/opentracing/opentracing-objc/blob/9d73066311e511e6649bd24c17a5e1689eb88184/Pod/Classes/OTTracer.h#L8-L17):

```objc
/**
 * The TextMap format identifier for OTTracer#inject:format:carrier: and
 * OTTracer#extractWithFormat:carrier:
 *
 * The carrier for OTFormatTextMap must be an NSMutableDictionary* instance for
 * `inject` and an NSDictionary* instance for `extract`.
 *
 * @see http://opentracing.io/propagation/
 */
FOUNDATION_EXPORT NSString *const OTFormatTextMap;
```

Ideally, we should expose it as an Obj-c constant, but due to this Swift <> Objc bridging limitation maybe we can create an artificial namespace of `OT` → `OT.FormatTextMap`?

tools/lint/sources.swiftlint.yml
buranmert: i can't get it working by adding a newline, can you please have a look?
https://regex101.com/r/1VlTj4/3

`^\s*(?!@objc)\n(override )?public` -> matches nothing
`^\s*(?!@objc\n)(override )?public` -> matches the wrong case below
```swift
@objc
override public foo()
```
---
with `attributes` re-enabled,
```swift
@objc public static func granted() -> DDTrackingConsent { .init(sdkConsent: .granted) }
@objc public static func notGranted() -> DDTrackingConsent { .init(sdkConsent: .notGranted) }
@objc public static func pending() -> DDTrackingConsent { .init(sdkConsent: .pending) }
```
will be
```swift
@objc
public static func granted() -> DDTrackingConsent { .init(sdkConsent: .granted) }

@objc
public static func notGranted() -> DDTrackingConsent { .init(sdkConsent: .notGranted) }

@objc 
public static func pending() -> DDTrackingConsent { .init(sdkConsent: .pending) }
```
do you prefer the latter?

ncreated: > do you prefer the latter?

It's not about my preference. Currently, we lint our code using `attributes` rule and I see no reason to disable it. 

This is valid syntax and it follows our linter:
```swift
@objc
public static func granted() -> DDTrackingConsent { .init(sdkConsent: .granted) }
```

I've been playing with the regex but cannot came up with the solution quickly. I can try later.

buranmert: i removed linter rule to unblock this PR

ncreated (approved):
👌 looks good! Let's work on this linter rule in `RUMM-788` 👌.

PR #364, opened 2021-01-11
RUMM-974 Allow referencing RUM Views by String key

ncreated:
### What and why?

📦 This PR introduces two additional APIs for RUM Views manual instrumentation. Now it is possible to use `String` key to mark the view start and view end:
```swift
// New APIs:
Global.rum.startView(key: "view1", path: "NameOfTheView")
Global.rum.stopView(key: "view1")
```
This builds up an alternative for use cases where `UIViewController` cannot be referenced using existing APIs:
```swift
// Existing APIs
Global.rum.startView(viewController: self, path: "NameOfTheView")
Global.rum.stopView(viewController: self)
```

### How?

I introduced a single interface for any kind of RUM View reference:
```swift
internal protocol RUMViewIdentifiable {
    func equals(_ otherIdentifiable: RUMViewIdentifiable) -> Bool
}
```
It's implemented by `UIViewController` and `String` types and can be extended any further in the future (e.g. for `SwiftUI`).

Additional `RUMViewIdentity` wrapper-type is introduced to manage the reference lifecycle. It ensures that the `UIViewController` is captured as a weak reference (so we don't leak it) and translates `String` into plain value. It provides basic interface for comparing identities:
```swift
let vc = UIViewController()
let otherVC = UIViewController()

let identity: RUMViewIdentity = vc.asRUMViewIdentity()

print(identity.equals(vc)) // true
print(identity.equals(otherVC)) // false
```
```swift
let key = "foo"
let otherKey = "bar"

let identity: RUMViewIdentity = key.asRUMViewIdentity()
print(identity.equals(key)) // true
print(identity.equals(otherKey)) // false
```

### Another option for new public API

Instead of providing two distinct sets of public APIs:
* `Global.rum.startView(key: String)`
* `Global.rum.stopView(key: String)`
* `Global.rum.startView(viewController: UIViewController)`
* `Global.rum.stopView(viewController: UIViewController)`

we can merge them into single, generic one:
* `Global.rum.startView(identity: RUMViewIdentifiable)`
* `Global.rum.stopView(identity: RUMViewIdentifiable)`

which will work automatically for both `UIViewController` and `String`. That would require deprecating the existing API, but I don't have any strong opinion on having 2 vs 1 set of the APIs. WDYT @xgouchet @mariusc83 @buranmert ?

### Review checklist

### DIFF
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
@@ -19,8 +19,8 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
     private unowned let parent: RUMContextProvider
     private let dependencies: RUMScopeDependencies
 
-    /// Weak reference to corresponding `UIViewController`, used to identify this View.
-    private(set) weak var identity: AnyObject?
+    /// The value holding stable identity of this RUM View.
+    let identity: RUMViewIdentity
     /// View attributes.
     private(set) var attributes: [AttributeKey: AttributeValue]
     /// View custom timings, keyed by name. The value of timing is given in nanoseconds.
@@ -55,7 +55,7 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
     init(
         parent: RUMContextProvider,
         dependencies: RUMScopeDependencies,
-        identity: AnyObject,
+        identity: RUMViewIdentifiable,
         uri: String,
         attributes: [AttributeKey: AttributeValue],
         customTimings: [String: Int64],
@@ -63,7 +63,7 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
     ) {
         self.parent = parent
         self.dependencies = dependencies
-        self.identity = identity
+        self.identity = identity.asRUMViewIdentity()
         self.attributes = attributes
         self.customTimings = customTimings
         self.viewUUID = dependencies.rumUUIDGenerator.generateUnique()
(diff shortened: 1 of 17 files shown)


xgouchet (approved):
LGTM

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -113,10 +113,10 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
             }
             didReceiveStartCommand = true
             needsViewUpdate = true
-        case let command as RUMStartViewCommand where command.identity !== identity:
+        case let command as RUMStartViewCommand where !identity.equals(command.identity):
             isActiveView = false
             needsViewUpdate = true // sanity update (in case if the user forgets to end this View)
-        case let command as RUMStopViewCommand where command.identity === identity:
+        case let command as RUMStopViewCommand where identity.equals(command.identity):
```
buranmert, line 119: we can add a case or if clause to handle when identity view controller becomes nil
as we keep a weak ref to VC, its `ViewScope` will stay forever if its identity `VC` is deallocated

something like that?
```swift
if self.identity.object == nil && self.identity.value == nil { stopView() }
```

ncreated: This is already covered by closing the scope if another scope is started:
```swift
case let command as RUMStartViewCommand where !identity.equals(command.identity):
   isActiveView = false
   needsViewUpdate = true // sanity update (in case if the user forgets to end this View)
```

Also, if the session expires, before we re-create the scope for last `identity` we first check if the identity still exists:
```swift
// RUMSessionScope

guard let expiredViewIdentifiable = expiredView.identity.identifiable else {
    return nil // if the underlying identifiable (`UIVIewController`) no longer exists, skip transferring its scope
}
```
I think we already have enough measures for keeping it consistent. WDYT @buranmert?

buranmert (approved):

PR #365, opened 2021-01-12
RUMM-976 GH Issue #363

### LINKED ISSUE #363: Baggage items not propagated to span's descendants
## The problem

As I was testing out various Datadog features, I noticed that baggage items set in a span are not propagated to its descendants.

It seems that the `OTSpan.baggageItem(withKey:)` only returns the baggage item if it was set directly on the span, and does not query the span's parent.

This is also evident from the implementation of `DDSpan` and `BaggageItems`:

```swift
internal class DDSpan: OTSpan {
    //...
    func baggageItem(withKey key: String) -> String? {
        if warnIfFinished("baggageItem(withKey:)") {
            return nil
        }
        return ddContext.baggageItems.get(key: key)
    }
    //...
}

internal class BaggageItems {
    //...
    /// Baggage items of the parent `DDSpan` or`nil` for items of the root span.
    private let parent: BaggageItems?

    /// Unsynchronized baggage items dictionary. Use `queue` to synchronize the access.
    private var unsafeItems: [String: String] = [:]
	//...
    func get(key: String) -> String? {
        queue.sync { self.unsafeItems[key] }
    }
    //...
    /// Returns all baggage items for the span, including its parent items.
    /// This property is unsafe and should be accessed using `queue`.
    private var unsafeAll: [String: String] {
        let parentItems = parent?.unsafeAll ?? [:]
        let selfItems = unsafeItems

        let allItems = parentItems.merging(selfItems) { _, selfItem -> String in
            return selfItem
        }

        return allItems
    }
}
```

The only way to access the parent span's baggage items is to iterate through all of then in `DDSpanContext`:

```swift
internal class DDSpanContext: OTSpanContext {
    //...
    func forEachBaggageItem(callback: (String, String) -> Bool) {
        for (itemKey, itemValue) in baggageItems.all {
            if callback(itemKey, itemValue) {
                break
            }
        }
    }
    //...
}
```

The documentation for `OTSpan.baggageItem(forKey:)` does not specify whether this is expected behavior.

## Suggestions

**If this is expected behavior**, adding a note in the `OTSpan.baggageItem(forKey:)` would be very useful.

**If `OTSpan.baggageItem(forKey:)` is expected to return ancestor baggage items**, then `BaggaggeItems` could be modified in the following way:

```swift
internal class BaggageItems {
    //...
    func get(key: String) -> String? {
        queue.sync { self.unsafeItems[key] } ?? parent?.get(key: key)
    }
    //...
```

buranmert:
Fixes #363 

### What and why?

Baggage items were not propagated to span's children
A great explanation can be found in #363 

### How?

Now `BaggageItems.get(key)` searches in `parent.baggageItems` recursively

### Review checklist

### DIFF
--- a/Sources/Datadog/Tracing/DDSpanContext.swift
+++ b/Sources/Datadog/Tracing/DDSpanContext.swift
@@ -49,7 +49,7 @@ internal class BaggageItems {
     }
 
     func get(key: String) -> String? {
-        queue.sync { self.unsafeItems[key] }
+        queue.sync { self.unsafeAll[key] }
     }
 
     var all: [String: String] {
--- a/Tests/DatadogTests/Datadog/Tracing/DDSpanContextTests.swift
+++ b/Tests/DatadogTests/Datadog/Tracing/DDSpanContextTests.swift
@@ -49,4 +49,17 @@ class DDSpanContextTests: XCTestCase {
         XCTAssertEqual(parentBaggageItems.all["foo"], "a")
         XCTAssertEqual(childBaggageItems.all["foo"], "b")
     }
+
+    func testChildItemsGetParentItems() {
+        let parentBaggageItems = BaggageItems(targetQueue: queue, parentSpanItems: nil)
+        let childBaggageItems = BaggageItems(targetQueue: queue, parentSpanItems: parentBaggageItems)
+
+        parentBaggageItems.set(key: "foo", value: "a")
+        childBaggageItems.set(key: "bar", value: "b")
+
+        XCTAssertEqual(childBaggageItems.get(key: "foo"), "a")
+
+        XCTAssertNil(parentBaggageItems.get(key: "bar"))
+        XCTAssertEqual(childBaggageItems.get(key: "bar"), "b")
+    }
 }


Sources/Datadog/Tracing/DDSpanContext.swift
```diff
@@ -49,7 +49,7 @@ internal class BaggageItems {
     }
 
     func get(key: String) -> String? {
-        queue.sync { self.unsafeItems[key] }
+        queue.sync { self.unsafeItems[key] } ?? parent?.get(key: key)
```
ncreated, line 52: This read isn't atomic and blocks the thread several times:
```swift
queue.sync { ... } ?? queue.sync { ... } ?? queue.sync { ... } ?? ... N
```
depending on how many `N` parent contexts there are in the hierarchy.

---
We can make it atomic and ensure there's only one thread sync (`queue.sync { ... }`) executed if we just do:
```swift
queue.sync { self.unsafeAll[key] }
```

As `unsafeAll` doc comment says, it:
> Returns all baggage items for the span, including its parent items.

ncreated (changes requested):

ncreated (approved):
👌

PR #367, opened 2021-01-12
RUMM-788 Add data scrubbing API

ncreated:
### What and why?

🚚 This PR adds RUM events scrubbing feature with set of 4 new public APIs:
```swift
public func setRUMViewEventMapper(_ mapper: @escaping (RUMViewEvent) -> RUMViewEvent?)
public func setRUMResourceEventMapper(_ mapper: @escaping (RUMResourceEvent) -> RUMResourceEvent?)
public func setRUMActionEventMapper(_ mapper: @escaping (RUMActionEvent) -> RUMActionEvent?)
public func setRUMErrorEventMapper(_ mapper: @escaping (RUMErrorEvent) -> RUMErrorEvent?)
```

Each API allows the user to inspect and modify RUM events before they get send. Returning `nil` drops the event.

### How?

A basic interface of `EventMapper` is introduced:
```swift
internal protocol EventMapper {
    func map<T: Encodable>(event: T) -> T?
}
```
and plugged into `DataProcessor's` pipeline, just before writing the event to the file.

It is implemented by `RUMEventsMapper` which intercepts each value using closures configured by the user in the public interface.

In the next PR I will also add test scenario to `Example` application with the integration test for scrubbing API.

### Review checklist

### DIFF
--- a/Sources/Datadog/RUM/Scrubbing/RUMEventsMapper.swift
+++ b/Sources/Datadog/RUM/Scrubbing/RUMEventsMapper.swift
@@ -1,0 +1,65 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+internal typealias RUMViewEventMapper = (RUMViewEvent) -> RUMViewEvent?
+internal typealias RUMErrorEventMapper = (RUMErrorEvent) -> RUMErrorEvent?
+internal typealias RUMResourceEventMapper = (RUMResourceEvent) -> RUMResourceEvent?
+internal typealias RUMActionEventMapper = (RUMActionEvent) -> RUMActionEvent?
+
+/// The `EventMapper` for RUM events.
+internal struct RUMEventsMapper: EventMapper {
+    private let viewEventMapper: RUMViewEventMapper
+    private let errorEventMapper: RUMErrorEventMapper
+    private let resourceEventMapper: RUMResourceEventMapper
+    private let actionEventMapper: RUMActionEventMapper
+
+    init(
+        viewEventMapper: RUMViewEventMapper?,
+        errorEventMapper: RUMErrorEventMapper?,
+        resourceEventMapper: RUMResourceEventMapper?,
+        actionEventMapper: RUMActionEventMapper?
+    ) {
+        self.viewEventMapper = viewEventMapper ?? RUMEventsMapper.noOpMapper(_:)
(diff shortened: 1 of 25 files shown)


Sources/Datadog/RUM/Scrubbing/RUMEventsMapper.swift
```diff
@@ -0,0 +1,65 @@
...
+
+    private func map<DM: RUMDataModel>(rumEvent: RUMEvent<DM>, using mapper: (DM) -> DM?) -> RUMEvent<DM>? {
+        if let mappedModel = mapper(rumEvent.model) {
+            var mutableRUMEvent = rumEvent
+            mutableRUMEvent.model = mappedModel
+            return mutableRUMEvent
+        } else {
+            return nil
+        }
+    }
```
buranmert, line 59:
```suggestion
  // let viewEventMapper: RUMViewEventMapper?, ...
    private func map<DM: RUMDataModel>(rumEvent: RUMEvent<DM>, using mapper: ((DM) -> DM?)?) -> RUMEvent<DM>? {
    guard let someMapper = mapper else { return rumEvent }

      if let mappedModel = mapper(rumEvent.model) {
          var mutableRUMEvent = rumEvent
          mutableRUMEvent.model = mappedModel
          return mutableRUMEvent
      } else {
          return nil
      }
  }
```
that way we won't need `noOpMapper` and unnecessary copying of `RUMEvent`s ✅ 
this is already how we do it in `DataProcessor.swift`, we can follow the same pattern in both places

ncreated: Indeed, I didn't notice we're copying the event value unnecessarily. Done 👍.

Tests/DatadogTests/Datadog/Mocks/RUMDataModelMocks.swift
```diff
@@ -0,0 +1,199 @@
...
+            ),
+            usr: .mockRandom(),
+            view: .init(
+                id: .mockRandom(),
+                referrer: .mockRandom(),
+                url: .mockRandom()
+            )
+        )
+    }
+}
```
buranmert, line 199: can we use model generator to create these mocks too?

buranmert (approved):

Tests/DatadogTests/Datadog/Mocks/RUMDataModelMocks.swift
ncreated: I was thinking of this, but it took me `~10 min` to do it by hand, while adding this in `rum-models-generator` would be at least few hours. Let's see how difficult would it be to maintain this hand-made mocks, if too expensive, we can invest our time in `rum-models-generator` 👌.

PR #369, opened 2021-01-13
RUMM-952 Missing RUM API added

buranmert:
### What and why?

Those methods are needed for internal usage

#### Swift API

```swift
func addError(message: String,source: RUMErrorSource = .custom,stack: String? = nil,attributes: [AttributeKey: AttributeValue] = [:],file: StaticString? = #file,line: UInt? = #line)

func startResourceLoading(resourceKey: String,httpMethod: String,urlString: String,kind: RUMResourceKind,attributes: [AttributeKey: AttributeValue] = [:])

func stopResourceLoading(resourceKey: String,statusCode: Int?,kind: RUMResourceKind = .other,size: Int64? = nil,attributes: [AttributeKey: AttributeValue] = [:])

enum RUMResourceKind
```

#### Obj-C API
```swift
func addError(message: String,source: DDRUMErrorSource,stack: String?,attributes: [String: Any])

func startResourceLoading(resourceKey: String,httpMethod: String,urlString: String,kind: DDRUMResourceKind,attributes: [String: Any])

func stopResourceLoading(resourceKey: String,statusCode: Int,kind: DDRUMResourceKind,attributes: [String: Any])

enum DDRUMResourceKind
```
### Review checklist

### DIFF
--- a/Sources/Datadog/DDRUMMonitor.swift
+++ b/Sources/Datadog/DDRUMMonitor.swift
@@ -65,7 +65,7 @@ public class DDRUMMonitor {
     /// - Parameters:
     ///   - message: a message explaining the Error.
     ///   - source: the origin of the error.
-    ///   - stack: stack trace of the error. no specific format needed. overwrites `file` and `line` parameters below.
+    ///   - stack: stack trace of the error. No specific format needed. Overwrites `file` and `line` parameters below.
     ///   - attributes: custom attributes to attach to the Error
     ///   - file: the file in which the Error occurred (the default is the file name in which this method was called).
     ///   - line: the line number on which the Error occurred (the default is the line number on which this method was called).
@@ -111,7 +111,7 @@ public class DDRUMMonitor {
         attributes: [AttributeKey: AttributeValue] = [:]
     ) {}
 
-    /// Notifies that the Resource starts being loaded using GET request to given `url`.
+    /// Notifies that the Resource starts being loaded from given `urlString`.
     /// - Parameters:
     ///   - resourceKey: the key representing the Resource - must be unique among all Resources being currently loaded.
     ///   - httpMethod: HTTP method used to fetch resource
@@ -119,7 +119,7 @@ public class DDRUMMonitor {
     ///   - attributes: custom attributes to attach to the Resource.
     public func startResourceLoading(
         resourceKey: String,
-        httpMethod: String,
+        httpMethod: RUMMethod,
         urlString: String,
         attributes: [AttributeKey: AttributeValue] = [:]
     ) {}
--- a/Sources/Datadog/DDRUMMonitor.swift
+++ b/Sources/Datadog/DDRUMMonitor.swift
@@ -152,13 +152,13 @@ public class DDRUMMonitor {
     /// - Parameters:
     ///   - resourceKey: the key representing the Resource - must match the one used in `startResourceLoading(...)`.
     ///   - statusCode: HTTP status code of response. e.g: `response.statusCode`
-    ///   - kind: corresponding `RUMResourceKind` of the resource.
+    ///   - kind: corresponding `RUMResourceType` of the resource.
     ///   - size: an optional size of the data received for the Resource (in bytes). If not provided, the SDK will try to infer it from the "Content-Length" header of the `response`.
     ///   - attributes: custom attributes to attach to the Resource.
     public func stopResourceLoading(
         resourceKey: String,
         statusCode: Int?,
-        kind: RUMResourceKind,
+        kind: RUMResourceType,
         size: Int64? = nil,
         attributes: [AttributeKey: AttributeValue] = [:]
     ) {}
--- a/Sources/Datadog/DDRUMMonitor.swift
+++ b/Sources/Datadog/DDRUMMonitor.swift
@@ -65,12 +65,14 @@ public class DDRUMMonitor {
     /// - Parameters:
     ///   - message: a message explaining the Error.
     ///   - source: the origin of the error.
+    ///   - stack: stack trace of the error. no specific format needed. overwrites `file` and `line` parameters below.
     ///   - attributes: custom attributes to attach to the Error
     ///   - file: the file in which the Error occurred (the default is the file name in which this method was called).
     ///   - line: the line number on which the Error occurred (the default is the line number on which this method was called).
     public func addError(
         message: String,
         source: RUMErrorSource = .custom,
+        stack: String? = nil,
         attributes: [AttributeKey: AttributeValue] = [:],
         file: StaticString? = #file,
         line: UInt? = #line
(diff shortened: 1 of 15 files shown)


Sources/Datadog/DDRUMMonitor.swift
```diff
@@ -65,12 +65,14 @@ public class DDRUMMonitor {
     /// - Parameters:
     ///   - message: a message explaining the Error.
     ///   - source: the origin of the error.
+    ///   - stack: stack trace of the error. no specific format needed. overwrites `file` and `line` parameters below.
```
ncreated, line 68:
```suggestion
    ///   - stack: stack trace of the error. No specific format needed. Overwrites `file` and `line` parameters below.
```

Sources/Datadog/DDRUMMonitor.swift
```diff
@@ -109,6 +111,19 @@ public class DDRUMMonitor {
...
 
+    /// Notifies that the Resource starts being loaded using GET request to given `url`.
+    /// - Parameters:
+    ///   - resourceKey: the key representing the Resource - must be unique among all Resources being currently loaded.
+    ///   - httpMethod: HTTP method used to fetch resource
+    ///   - urlString: the `URL` for the Resource in `String` form.
+    ///   - attributes: custom attributes to attach to the Resource.
+    public func startResourceLoading(
+        resourceKey: String,
+        httpMethod: String,
```
ncreated, line 122: In #366 we merged public `RUMMethod` enumeration, which should be used here.

Sources/Datadog/DDRUMMonitor.swift
```diff
@@ -109,6 +111,19 @@ public class DDRUMMonitor {
         attributes: [AttributeKey: AttributeValue] = [:]
     ) {}
 
+    /// Notifies that the Resource starts being loaded using GET request to given `url`.
```
ncreated, line 114: This is not precise - we don't constraint this API to only `GET` if we expect `httpMethod`.

Sources/Datadog/RUMMonitor.swift
```diff
@@ -283,14 +283,19 @@ public class RUMMonitor: DDRUMMonitor, RUMCommandSubscriber {
...
-        }
+        let stack: String? = stack ?? {
+            if let file = file,
+               let fileName = "\(file)".split(separator: "/").last,
+               let line = line {
+                return "\(fileName):\(line)"
+            }
+            return nil
+        }()
         addError(message: message, stack: stack, source: RUMInternalErrorSource(source), attributes: attributes)
```
ncreated, line 299: The:
```swift
internal func addError(
   message: String,
   stack: String?,
   source: RUMInternalErrorSource,
   attributes: [AttributeKey: AttributeValue]
)
```
is not needed anymore. It was there to provide explicit internal API for passing `stack` parameter. Now we can replace it's only usage with:
```swift
override public func addError(
   message: String,
   source: RUMErrorSource,
   stack: String?,
   attributes: [AttributeKey: AttributeValue],
   file: StaticString?,
   line: UInt?
)
```
(it's only called once in `RUMErrorsIntegration`).

Sources/Datadog/RUMMonitor.swift
```diff
@@ -23,7 +23,7 @@ internal enum RUMHTTPMethod: String {
     }
 }
 
-internal enum RUMResourceKind {
+public enum RUMResourceKind {
```
ncreated, line 26: In #366 we merged a public enumeration for this: `RUMResourceEvent.Resource.ResourceType`, so we will be duplicating this information in public interface. Lets go with this `public enum RUMResourceKind` for `-alpha2`, but I will create JIRA to generate `RUMResourceEvent.Resource.ResourceType` as `RUMResourceKind` and update it before `1.5.0` GA. WDYT?

ncreated (changes requested):
Few small comments, but looks good in general 👌

Sources/Datadog/RUMMonitor.swift
buranmert: it seems it was there to receive `RUMInternalErrorSource`
i guess we should keep it as we can't make `RUMInternalErrorSource` public

Sources/Datadog/DDRUMMonitor.swift
buranmert: `internal enum RUMHTTPMethod` should be removed then
i don't remember why it was created in the first place 🤔

ncreated: It was there because the old version of rum data models used two different representations of `method` (one for `resource.method` and another one for `error.resource.method`. I think we can remove it as part of the JIRA I suggest in https://github.com/DataDog/dd-sdk-ios/pull/369#discussion_r557228073. This requires update in `rum-models-generator` to rename it from `RUMMethod` to `RUMHTTPMethod` (which sounds better).

Sources/Datadog/RUMMonitor.swift
ncreated: Ah, indeed 👍

Sources/Datadog/RUMMonitor.swift
buranmert: it should be quick to replace `RUMResourceKind` with `ResourceType` but we should first discuss how to expose `RUMDataModels` to ObjC 🤔 until they are `@objc`, we will need to duplicate them manually in ObjC (which is not ideal)

so yes, let's keep it for now as it's done already and fix it before `1.5.0` 👍 
i'll leave a `TODO` here

Sources/Datadog/DDRUMMonitor.swift
buranmert: i just removed it and replaced it with `RUMMethod` ✅ it was trivial

i think `RUMMethod` is okay because properties too have the name `method` (`Resource.method: RUMMethod`)
if we rename it, IMO we should also rename properties as well

Sources/Datadog/RUMMonitor.swift
buranmert: actually there is the same issue for `String -> RUMMethod` replacement: now i replaced `httpMethod: String` with `httpMethod: RUMMethod` but i also needed to create `DDRUMMethod` in ObjC (a duplication, basically)

i'd say let's fix both in this PR or let's leave both and see later, wdyt?

buranmert: replaced ✅

Sources/DatadogObjc/RUMMonitor+objc.swift
```diff
@@ -10,6 +10,9 @@ import class Datadog.DDRUMMonitor
 import class Datadog.RUMMonitor
 import enum Datadog.RUMErrorSource
 import enum Datadog.RUMUserActionType
+//import enum Datadog.RUM
```
ncreated, line 13:
```suggestion
```

Sources/Datadog/RUMMonitor.swift
```diff
@@ -7,50 +7,36 @@
...
-    case fetch
-    case font
-    case js
-    case media
-    case other
-
-    private static let xhrHTTPMethods: Set<String> = ["POST", "PUT", "DELETE"]
-
-    /// Determines the `RUMResourceKind` based on a given `URLRequest`.
+public typealias RUMResourceType = RUMResourceEvent.Resource.ResourceType
```
ncreated, line 21:
```suggestion
public typealias RUMResourceType = RUMResourceEvent.Resource.ResourceType

```

ncreated (approved):
Two last details💄 , and we're good to go 👌.

PR #372, opened 2021-01-15
RUMM-990 Special InternalTimestamp attribute is handled

buranmert:
### What and why?

In some special cases, `RUMMonitor` should let the caller feed event time explicitly

### How?

We use `_dd.timestamp` special attribute for such cases.
`RUMMonitor` extracts and removes it from `attributes` and use it as `command.time`.
This attribute overwrites default `command.time: Date()` behavior.

### Review checklist

### DIFF
--- a/Tests/DatadogTests/Datadog/RUMMonitorTests.swift
+++ b/Tests/DatadogTests/Datadog/RUMMonitorTests.swift
@@ -1077,6 +1077,25 @@ class RUMMonitorTests: XCTestCase {
         try Datadog.deinitializeOrThrow()
     }
 
+    // MARK: - Internal attributes
+
+    func testHandlingInternalTimestampAttribute() throws {
+        RUMFeature.instance = .mockNoOp()
+        defer { RUMFeature.instance = nil }
+
+        var mockCommand = RUMCommandMock()
+        mockCommand.attributes = [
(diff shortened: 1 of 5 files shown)


Tests/DatadogTests/Datadog/RUMMonitorTests.swift
```diff
@@ -1077,6 +1077,15 @@ class RUMMonitorTests: XCTestCase {
...
     }
 
+    // MARK: - Internal attributes
+
+    func testHandlingInternalTimestampAttribute() {
+        // TODO: RUMM-990
+        // this test should mock `ApplicationScope` and assert on its received `RUMCommands`
+        // we can discuss how to do this in this PR
+        return
+    }
```
buranmert, line 1087: either i need to duplicate almost every test case: one with default time and one with internal timestamp.
or i will mock `ApplicationScope` and assert `receivedCommand.time`: but this requires creating `ApplicationScope` protocol, is this fine with you @ncreated ?

if i remember well, we already have reviewing those end-to-end tests one day, maybe we can leave mocking `ApplicationScope` to that day. until then i can test internal timestamp in only some `RUMMonitor` methods.?

ncreated (approved):

PR #373, opened 2021-01-15
RUMM-908 Sanitize Span and RUM Event attributes

ncreated:
### What and why?

📦 This PR adds `Span` and `RUMEvent` attributes sanitization as it exists for `Logs`. The same rules were applied, unified with `dd-sdk-android` (https://github.com/DataDog/dd-sdk-android/pull/427).

### How?

Similar to `LogSanitizer`, two new sanitizers are introduced:
* `SpanSanitizer`,
* `RUMEventSanitizer`.

All three share common `AttributesSanitizer` which applies equal rules to all attributes with dynamic count:
* `Span` tags,
* `Span` meta information coming from `UserInfo.extraInfo`,
* `RUMEvent<DM>` attributes,
* `RUMEvent<DM>` attributes coming from `UserInfo.extraInfo`,
* `RUMEvent<DM>` custom view timings.

Only two rules are shared:
* attribute key may contain up to 8 nested levels (`"one.two.(...).eight"`),
* there can be up to `128` attributes in total.

### Review checklist

### DIFF
--- a/Sources/Datadog/Tracing/Span/SpanSanitizer.swift
+++ b/Sources/Datadog/Tracing/Span/SpanSanitizer.swift
@@ -8,7 +8,7 @@ import Foundation
 
 /// Sanitizes `Span` representation received from the user, so it can match Datadog APM constraints.
 internal struct SpanSanitizer {
-    private let attributesSanitizer = AttributesSanitizer()
+    private let attributesSanitizer = AttributesSanitizer(featureName: "Span")
 
     func sanitize(span: Span) -> Span {
         // Sanitize attribute names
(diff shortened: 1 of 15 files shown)


Sources/Datadog/Tracing/Span/SpanSanitizer.swift
```diff
@@ -0,0 +1,53 @@
...
+            networkConnectionInfo: span.networkConnectionInfo,
+            mobileCarrierInfo: span.mobileCarrierInfo,
+            userInfo: .init(
+                id: span.userInfo.id,
+                name: span.userInfo.name,
+                email: span.userInfo.email,
+                extraInfo: sanitizedUserExtraInfo
+            ),
+            tags: sanitizedTags
+        )
```
buranmert, line 51: we can make variable parts `var` instead of this, the same for `Log-Span-RUMEvent`

buranmert (commented):
overall looks good but we can remove those long `init` calls

Sources/Datadog/Tracing/Span/SpanSanitizer.swift
ncreated: ✅ done

buranmert (approved):

Issue #370, opened 2021-01-14
Datadog use in ios (notification) extension


SimpleApp:
### The issue

I'm trying to use datadog in both the app bundle and the notification extension bundle. I've tried two and encountered two different errors: 
1/ using spm. I had to manually embed the Datadog library for the notification extension code to compile. That worked just fine, however, doing that triggered a bug when submitting to apple : 
- ERROR ITMS-90685: "CFBundleIdentifier Collision. There is more than one bundle with the CFBundleIdentifier value 'Datadog' under the iOS application
- ERROR ITMS-90205: "Invalid Bundle. The bundle at … /Plugins/NotificationServiceExt.appex' contains disallowed nested bundles.
- ERROR ITMS-90206: "Invalid Bundle. The bundle at '…/PlugIns/NotificationServiceExt.appex' contains disallowed file 'Frameworks'."

So, i tried using Cocoapods, and wrote  "pod 'DatadogSDK' " in both the app and the extension targets. 
This doesn't compile, because of a "UIApplication.shared" dependency in RUMDebugging.swift : 
- 'shared' is unavailable in application extensions for iOS: Use view controller based solutions where appropriate instead.
At this point i feel kind of stuck.. Did i miss something ?

---

#### Datadog SDK version:

1.4.0
#### Last working Datadog SDK version:

?
#### Dependency Manager:
CocoaPods & SwiftPM

#### Xcode version:

12.2

#### Swift version:

5.x

#### Deployment Target:
ios 13.0

ncreated:
Hello @SimpleApp 👋. Thanks for reporting this issue. Indeed, there was a problem with `UIApplication.shared` API when compiling the SDK for use in some app extensions (in this case, Notifications Service Extension). This is solved in #375 and will be available soon as `1.4.1` hotfix 🙂👌.

> 1/ using spm. I had to manually embed the Datadog library for the notification extension code to compile.

What was the reason of embedding `Datadog` in extension's target? It should work fine with _"Do Not Embed"_ setting. I managed to build a simple example using Share Extension and link `Datadog` through SPM. It compiles and sends logs with no issue  (tested on iOS Simulator, Xcode 12.3):

<img width="1300" alt="Screenshot 2021-01-18 at 17 43 30" src="https://user-images.githubusercontent.com/2358722/104943058-7d3c7980-59b5-11eb-8a87-93ded3c4d831.png">

ncreated:
[...]

PR #374, opened 2021-01-18
RUMM-989 Fix compiler error in Notification Service Extension

ncreated:
### What and why?

🔧 When compiling the SDK for some app extension types (e.g. Notification Service Extension), compiler error is issued:
> `'shared'` is unavailable in application extensions for iOS: Use view controller based solutions where appropriate instead

Solves #370, qualifies for `1.4.1` hotfix.

### How?

There is no [compiler control statement](https://docs.swift.org/swift-book/ReferenceManual/Statements.html#ID538) to exclude [`UIApplication.shared`](https://developer.apple.com/documentation/uikit/uiapplication/1622975-shared) usage in iOS app extension, so I used a known workaround (https://github.com/slackhq/PanModal/pull/81) of referencing it using key-path.

Convenient extension of `UIApplication.managedShared: UIApplication?` is created and swift-lint rule is added to enforce its usage. This required bumping the swift-lint version in `bitrise.yml` to leverage its new `excluded_match_kinds` syntax.

### Review checklist


buranmert (approved):
lgtm 🚀

PR #376, opened 2021-01-19
RUMM-786 Support RUM Explorer's indicator for active views

acgh:
### What and why?

This is the iOS counter-part of RUMM-787 / #430.
As a reminder, it sets the `view.is_active` attribute for RUM Views to let the frontend know when a View is still in progress.

### How?

In `dd-sdk-ios`, the `RUMViewScope` was already able to handle `RUMStartViewCommand` and `RUMStopViewCommand` and it was tracking some `isActiveView`. This change makes sure a `RUMViewEvent` gets a `View` with its `isActive` property's value matching `RUMViewScope.isActiveView`'s value.

This PR also changes the behaviour of `RUMViewScope.context` by checking `isActiveView` before setting/overwriting `RUMContext.active*` properties.

Finally, it updates existing unit and integration tests to assert on the `view.isActive` based on a given navigation path.

### Review checklist

- ✅ Feature or bugfix MUST have appropriate tests (unit, integration)
- ✅ Make sure each commit and the PR mention the Issue number or JIRA reference

### DIFF
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
@@ -76,9 +76,11 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
 
     var context: RUMContext {
         var context = parent.context
-        context.activeViewID = viewUUID
-        context.activeViewURI = viewURI
-        context.activeUserActionID = userActionScope?.actionUUID
+        if isActiveView {
+            context.activeViewID = viewUUID
+            context.activeViewURI = viewURI
+            context.activeUserActionID = userActionScope?.actionUUID
+        }
         return context
     }
 
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
@@ -278,7 +278,7 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
                 firstContentfulPaint: nil,
                 firstInputDelay: nil,
                 id: viewUUID.toRUMDataFormat,
-                isActive: nil,
+                isActive: isActiveView,
                 largestContentfulPaint: nil,
                 loadEvent: nil,
                 loadingTime: nil,
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
@@ -76,11 +76,9 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
 
     var context: RUMContext {
         var context = parent.context
-        if isActiveView {
-            context.activeViewID = viewUUID
-            context.activeViewURI = viewURI
-            context.activeUserActionID = userActionScope?.actionUUID
-        }
+        context.activeViewID = viewUUID
+        context.activeViewURI = viewURI
+        context.activeUserActionID = userActionScope?.actionUUID
         return context
     }
 
--- a/Tests/DatadogIntegrationTests/Scenarios/RUM/RUMCommonAsserts.swift
+++ b/Tests/DatadogIntegrationTests/Scenarios/RUM/RUMCommonAsserts.swift
@@ -61,10 +61,6 @@ extension RUMSessionMatcher {
         return sessionMatchers.first
     }
 
-    class func assertViewWasInitiallyActive(_ viewVisit: ViewVisit) {
-        XCTAssertTrue(try XCTUnwrap(viewVisit.viewEvents.first?.view.isActive))
-    }
-
     class func assertViewWasEventuallyInactive(_ viewVisit: ViewVisit) {
         XCTAssertFalse(try XCTUnwrap(viewVisit.viewEvents.last?.view.isActive))
     }
--- a/Tests/DatadogIntegrationTests/Scenarios/RUM/RUMCommonAsserts.swift
+++ b/Tests/DatadogIntegrationTests/Scenarios/RUM/RUMCommonAsserts.swift
@@ -60,4 +60,12 @@ extension RUMSessionMatcher {
 
         return sessionMatchers.first
     }
+
+    class func assertViewWasInitiallyActive(_ viewVisit: ViewVisit) {
+        XCTAssertTrue(try XCTUnwrap(viewVisit.viewEvents.first?.view.isActive))
+    }
+
+    class func assertViewWasEventuallyInactive(_ viewVisit: ViewVisit) {
+        XCTAssertFalse(try XCTUnwrap(viewVisit.viewEvents.last?.view.isActive))
+    }
 }
--- a/Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMViewScopeTests.swift
+++ b/Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMViewScopeTests.swift
@@ -232,15 +232,13 @@ class RUMViewScopeTests: XCTestCase {
         )
 
         let viewEvents = try output.recordedEvents(ofType: RUMEvent<RUMViewEvent>.self)
-        XCTAssertTrue(viewEvents.count == 2)
-        if viewEvents.count == 2 {
-            let view1WasActive = try XCTUnwrap(viewEvents[0].model.view.isActive)
-            XCTAssertTrue(view1WasActive)
-            XCTAssertEqual(viewEvents[1].model.view.url, "FirstViewController")
-            let view2IsActive = try XCTUnwrap(viewEvents[1].model.view.isActive)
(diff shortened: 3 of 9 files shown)


Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -76,9 +76,11 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
 
     var context: RUMContext {
         var context = parent.context
-        context.activeViewID = viewUUID
-        context.activeViewURI = viewURI
-        context.activeUserActionID = userActionScope?.actionUUID
+        if isActiveView {
```
acgh, line 79: @ncreated @buranmert this change made sense to me, and tests are fine with it.
Could you think of bad consequences such a small change of behaviour could cause though? Maybe I'm missing some context about why `isActiveView` was not accounted for before, and maybe my change doesn't have any effect at all.

buranmert: i guess it was working as `viewScopes.last` is always the _active view_
whatever the other `viewScope`s set as `activeViewID`, it was always corrected by `viewScopes.last`
so it would probably break if `viewScopes` were let's say a `Set` or anything unordered

your change totally makes sense to me 👍

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -76,9 +76,11 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
...
+            context.activeViewID = viewUUID
+            context.activeViewURI = viewURI
+            context.activeUserActionID = userActionScope?.actionUUID
+        }
```
ncreated, line 83: I'm not so sure on this 💭. The `context` propagates to child scopes which in case of `RUMViewScope` are: `RUMResourceScope` and `RUMUserActionScope`.

The `RUMResourceScope` captures the view's `context` by value:
```swift
// RUMResourceScope.swift

let context: RUMContext
```
so it's immutable and not impacted by this change.

However, the `RUMUserActionScope` references view's `context` using `parent`:
```swift
// RUMUserActionScope.swift

private unowned let parent: RUMContextProvider // it's set to `RUMViewScope`

var context: RUMContext { parent.context }
```
so it will receive session scope's `context` if the view scope becomes inactive.

Oh the other hand, following the implementation of `RUMViewScope` and `RUMUserActionScope` the view's `context` seems to be not read by action scope after the view becomes inactive. What was the reason for this change?

Tests/DatadogIntegrationTests/Scenarios/RUM/RUMCommonAsserts.swift
```diff
@@ -60,4 +60,12 @@ extension RUMSessionMatcher {
...
         return sessionMatchers.first
     }
+
+    class func assertViewWasInitiallyActive(_ viewVisit: ViewVisit) {
+        XCTAssertTrue(try XCTUnwrap(viewVisit.viewEvents.first?.view.isActive))
+    }
+
+    class func assertViewWasEventuallyInactive(_ viewVisit: ViewVisit) {
+        XCTAssertFalse(try XCTUnwrap(viewVisit.viewEvents.last?.view.isActive))
+    }
```
ncreated, line 70: I like this assertions 👍. But instead of repeating them in all tests, how about moving it to the `RUMSessionMatcher.init()'s` implementation? We already raise few `RUMSessionConsistencyException` for common inconsistencies if spotted in RUM Session. This would keep integration tests cleaner 👌.

ncreated (changes requested):
Look very good 👌. Few questions and suggestions left 🙂.

Also, please update the PR name so it follows the convention and is more verbose on what it adds, e.g.:
- [x] "RUMM-786 Support RUM Explorer's indicator for active views"

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
ncreated: I commented on this in https://github.com/DataDog/dd-sdk-ios/pull/376#discussion_r560060267. In short - IMO it doesn't break anything nor doesn't add anything new. In such case, I'd keep the code simple and revert to the previous implementation. We can consider it if there's a reason or the requirements change.

buranmert: if `context.activeView*` is not neccessarily of current `activeView`'s, maybe it shouldn't be called `context.activeView*`, wdyt?

ncreated: This is just the naming thing to keep the lingo consistent with `dd-sdk-android`.

Indeed, `context.viewID` may work better in scopes, but it won't do well for usages in `Logger` or `Tracer`, where we retrieve `contextProvider.context.activeViewID` (which sounds IMO cleaner than `.viewID`). I'd keep it as it is for now, as `RUMM-786` is too small for architecture-wise refactoring.

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMViewScopeTests.swift
```diff
@@ -226,9 +232,15 @@ class RUMViewScopeTests: XCTestCase {
         )
 
         let viewEvents = try output.recordedEvents(ofType: RUMEvent<RUMViewEvent>.self)
-        let event = try XCTUnwrap(viewEvents.dropFirst().first)
-        XCTAssertEqual(event.model.view.url, "FirstViewController")
-        XCTAssertEqual(event.model.view.timeSpent, TimeInterval(1).toInt64Nanoseconds, "The View should last for 1 second")
+        XCTAssertTrue(viewEvents.count == 2)
```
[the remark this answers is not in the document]
acgh: Oh yes, good catch!

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMViewScopeTests.swift
```diff
@@ -226,9 +232,15 @@ class RUMViewScopeTests: XCTestCase {
...
+        if viewEvents.count == 2 {
+            let view1WasActive = try XCTUnwrap(viewEvents[0].model.view.isActive)
+            XCTAssertTrue(view1WasActive)
+            XCTAssertEqual(viewEvents[1].model.view.url, "FirstViewController")
+            let view2IsActive = try XCTUnwrap(viewEvents[1].model.view.isActive)
+            XCTAssertFalse(view2IsActive)
+            XCTAssertEqual(viewEvents[1].model.view.timeSpent, TimeInterval(1).toInt64Nanoseconds, "The View should last for 1 second")
+        }
```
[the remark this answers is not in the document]
acgh: Yes, that makes sense, and it would be consistent with other places where we assume the number of events to check.

Tests/DatadogIntegrationTests/Scenarios/RUM/RUMCommonAsserts.swift
acgh: That's interesting. I can see how this would help have 0 changes in the integration tests files. Let's try this out.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
acgh: Thanks for the context (no pun). The reason was mainly for consistency in case `RUMViewScope.context` was called somehow after the view became inactive, but I wasn't sure how/if it would happen.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
acgh: Yeah no worries, I wouldn't take on a refactoring inside such a small scope task :)
Also if this doesn't add any value right now, I'm ok reverting it. Still that would be because we can assume some pattern of calls.
That also means this might be the kind of details that could come back to bite us in a while once we start querying scopes differently for some unrelated reason, and I'm on fence about what to do. The very least would be a comment/TODO, what do you think?

ncreated: I think we can leave a comment in `RUMUserActionScope`, where it retrieves the `context` from `parent` (so the view scope). Right now, it never reads `parent.context` of an inactive view scope, but this may indeed change due to new requirements or so 👍.

Tests/DatadogIntegrationTests/Scenarios/RUM/RUMCommonAsserts.swift
ncreated: As a bonus, we will validate it also in some unit tests as we use it there as well 😉. `RUMSessionMatcher` is a fairly new construct, so it doesn't have 100% adoption in unit tests yet.

ncreated (approved):
Looks good 👌 💪

PR #378, opened 2021-01-21
RUMM-954 DDURLSessionDelegate takes firstPartyHosts in init

buranmert:
### What and why?

Auto-instrumented `firstPartyHosts` were set at global level via `DatadogConfiguration`
Every `URLSession` instance had the same set of `firstPartyHosts`

Now this set can be customized per instance via `DDURLSessionDelegate`
```swift
let delegateWithCustomHosts = DDURLSessionDelegate(firstPartyHosts: ["foo.com"])
let session = URLSession(delegate: delegateWithCustomHosts)
```
`session` tracks requests made to host `foo.com` regardless of `DatadogConfiguration`

### How?

`URLSessionInterceptor` checks if `session.delegate` has its own URL filter;
1. if so, this filter is used
2. otherwise, default URL filter of `URLSessionInterceptor` is used

### Note on Objc

Next PR will bring its Objc counterpart once Swift one is approved

### Review checklist

### DIFF
--- a/Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
+++ b/Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
@@ -13,9 +13,32 @@ import Foundation
 @objc
 open class DDURLSessionDelegate: NSObject, URLSessionTaskDelegate {
     var interceptor: URLSessionInterceptorType?
+    let firstPartyURLsFilter: FirstPartyURLsFilter?
 
     @objc
     override public init() {
+        firstPartyURLsFilter = nil
+        interceptor = URLSessionAutoInstrumentation.instance?.interceptor
+        if interceptor == nil {
+            let error = ProgrammerError(
+                description: """
+                `Datadog.initialize()` must be called before initializing the `DDURLSessionDelegate` and
+                first party hosts must be specified in `Datadog.Configuration`: `track(firstPartyHosts:)`
+                to enable network requests tracking.
+                """
+            )
+            consolePrint("\(error)")
+        }
+    }
+
+    // NOTE: RUMM-954 copy&pasting `init()` is a conscious decision.
+    // otherwise `DDURLSessionDelegateAsSuperclassTests` fails.
+    // if `init()` was made convenience and call the designated `init` below,
+    // that would result in potential breaking changes.
+    // host projects would need to change their `init()`s in subclasses.
+    // we can fix this in v2.0
+    public init(firstPartyHosts: Set<String>) {
+        firstPartyURLsFilter = FirstPartyURLsFilter(firstPartyHosts: firstPartyHosts)
         interceptor = URLSessionAutoInstrumentation.instance?.interceptor
         if interceptor == nil {
             let error = ProgrammerError(
--- a/Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
+++ b/Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
@@ -35,7 +35,7 @@ open class DDURLSessionDelegate: NSObject, URLSessionTaskDelegate {
         // host projects would need to change their `init()`s in subclasses.
         // we can fix this in v2.0
         Self.datadogInitializationCheck()
-        firstPartyURLsFilter = FirstPartyURLsFilter(firstPartyHosts: additionalFirstPartyHosts)
+        firstPartyURLsFilter = FirstPartyURLsFilter(hosts: additionalFirstPartyHosts)
         interceptor = URLSessionAutoInstrumentation.instance?.interceptor
     }
 
--- a/Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
+++ b/Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
@@ -7,7 +7,7 @@
 import Foundation
 
 /// The `URLSession` delegate object which enables network requests instrumentation. **It must be
-/// used together with** `Datadog.Configuration.track(firstPartyHosts:)`.
+/// used together with** `Datadog.Configuration.trackURLSession(firstPartyHosts:)`.
 ///
 /// All requests made with the `URLSession` instrumented with this delegate will be intercepted by the SDK.
 @objc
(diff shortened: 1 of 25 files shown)


Sources/Datadog/URLSessionAutoInstrumentation/Interception/URLFiltering/FirstPartyURLsFilter.swift
```diff
@@ -12,9 +12,9 @@ internal struct FirstPartyURLsFilter {
     /// "example.com", "api.example.com", but not "foo.com".
     private let regex: String
 
-    init(configuration: FeaturesConfiguration.URLSessionAutoInstrumentation) {
+    init(firstPartyHosts: Set<String>) {
```
acgh, line 15: Ha, interesting I have a similar change that I haven't pushed yet. I'm going to revert on my end to avoid merge-conflicts :)
Some more compact naming suggestion:
```suggestion
    init(hosts: Set<String>) {
```

Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
```diff
@@ -13,9 +13,32 @@ import Foundation
...
+            consolePrint("\(error)")
+        }
+    }
+
+    // NOTE: RUMM-954 copy&pasting `init()` is a conscious decision.
+    // otherwise `DDURLSessionDelegateAsSuperclassTests` fails.
+    // if `init()` was made convenience and call the designated `init` below,
+    // that would result in potential breaking changes.
+    // host projects would need to change their `init()`s in subclasses.
+    // we can fix this in v2.0
```
ncreated, line 39: We cannot have this comment in public API. Instead it should be documenting the feature, especially:
* what is `firstPartyHosts` made for?
* how `firstPartyHosts` should be formatted (e.g. `example.com` not `https://example.com`)?
* what happens if `firstPartyHosts` are set for `DDURLSessionDelegate` and in global `Datadog.Configuration`?

Also, it would be cool to update `track(firstPartyHosts:)` documentation comment with a note that they can be overwritten per `DDURLSessionDelegate` instance.

Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
```diff
@@ -13,9 +13,32 @@ import Foundation
...
+        if interceptor == nil {
+            let error = ProgrammerError(
+                description: """
+                `Datadog.initialize()` must be called before initializing the `DDURLSessionDelegate` and
+                first party hosts must be specified in `Datadog.Configuration`: `track(firstPartyHosts:)`
+                to enable network requests tracking.
+                """
+            )
+            consolePrint("\(error)")
+        }
```
ncreated, line 31: Instead of duplicating this error code, we can move the validation to private static function:
```swift
private static func retrieveInterceptor() -> URLSessionInterceptorType? {
   guard let interceptor = URLSessionAutoInstrumentation.instance?.interceptor else {
      let error = ProgrammerError(
         description: """
            `Datadog.initialize()` must be called before initializing the `DDURLSessionDelegate` and
            first party hosts must be specified in `Datadog.Configuration`: `track(firstPartyHosts:)`
            to enable network requests tracking.
            """
         )
         consolePrint("\(error)")
         return nil
   }
   return interceptor
}
```

Sources/Datadog/URLSessionAutoInstrumentation/Interception/URLSessionInterceptor.swift
```diff
@@ -20,7 +20,9 @@ public class URLSessionInterceptor: URLSessionInterceptorType {
     }
 
     /// Filters first party `URLs` defined by the user.
-    private let firstPartyURLsFilter: FirstPartyURLsFilter
+    /// **NOTE:** If `session.delegate` is a `DDURLSessionDelegate` initialized with its own
+    /// set of `firstPartyHosts`, then `defaultFirstPartyURLsFilter` is not used
```
ncreated, line 24: IMO this comment is unnecessary 💭 - it just explains how the following code works.

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/Interception/URLSessionInterceptorTests.swift
```diff
@@ -128,11 +130,22 @@ class URLSessionInterceptorTests: XCTestCase {
         )
         Global.sharedTracer = Tracer.mockAny()
         defer { Global.sharedTracer = DDNoopGlobals.tracer }
+        let sessionWithCustomFirstPartyHosts = URLSession(
+            configuration: .default,
+            delegate: DDURLSessionDelegate(
+                firstPartyHosts: [alternativeFirstPartyRequest.url!.host!]
+            ),
+            delegateQueue: nil
+        )
```
ncreated, line 139: If we need to mock only part of the `URLSession` interface, it would be more readable to create partial mock:
```swift
let sessionWithCustomFirstPartyHosts: URLSession = .mockWith(
   delegate: DDURLSessionDelegate(firstPartyHosts: [alternativeFirstPartyRequest.url!.host!])
)
```
as we do for many other things in there.

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/Interception/URLSessionInterceptorTests.swift
```diff
@@ -29,6 +29,8 @@ class URLSessionInterceptorTests: XCTestCase {
     private let handler = URLSessionInterceptionHandlerMock()
     /// Mock request made to a first party URL.
     private let firstPartyRequest = URLRequest(url: URL(string: "https://api.first-party.com/v1/endpoint")!)
+    /// Alternative mock request made to a first party URL.
+    private let alternativeFirstPartyRequest = URLRequest(url: URL(string: "https://api.another-first-party.com/v1/endpoint")!)
```
ncreated, line 33: We don't cover this behaviour in tests:
* _if one set of hosts is defined in global config, but another set of hosts is specified for `DDURLSessionDelegate`, the global set is not considered first party._

I think it deserves separate test function.

ncreated (changes requested):
As we're adding public API we should also include cover it in integration tests 🛡️.

Sources/Datadog/URLSessionAutoInstrumentation/Interception/URLFiltering/FirstPartyURLsFilter.swift
buranmert: yes good idea 👍 i'll shorten the name

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/Interception/URLSessionInterceptorTests.swift
buranmert: we don't cover many other things too, in theory we need to duplicate every test case we currently have with this new configuration but that would be too much.
as an attempt to keep a balance, i tested new config in very basic cases only; shall we change that?

Sources/Datadog/URLSessionAutoInstrumentation/Interception/URLSessionInterceptor.swift
```diff
@@ -95,10 +93,12 @@ public class URLSessionInterceptor: URLSessionInterceptorType {
...
             return request
         }
-        let ddDelegate = session?.delegate as? DDURLSessionDelegate
-        let firstPartyURLsFilter = ddDelegate?.firstPartyURLsFilter ?? defaultFirstPartyURLsFilter
-        if injectTracingHeadersToFirstPartyRequests,
-           firstPartyURLsFilter.isFirstParty(url: request.url) {
+        let delegateURLFilter = (session?.delegate as? DDURLSessionDelegate)?.firstPartyURLsFilter
+        let isFirstPartyForDelegate = (delegateURLFilter?.isFirstParty(url: request.url)) ?? false
+        let isFirstPartyForInterceptor = self.defaultFirstPartyURLsFilter.isFirstParty(url: request.url)
+        let isFirstParty = isFirstPartyForDelegate || isFirstPartyForInterceptor
```
ncreated, line 99: How about decoupling it to single `isFirstPartyRequest(request: Request, delegate: URLSessionDelegate?) -> Bool`? Now we repeat the same logic in two places.

ncreated (changes requested):
Two things on the public API doc 👌.

Datadog/Datadog.xcodeproj/xcshareddata/xcschemes/Example.xcscheme
```diff
@@ -99,7 +99,7 @@
          <EnvironmentVariable
             key = "DD_TEST_SCENARIO_IDENTIFIER"
             value = "RUMResourcesScenario"
-            isEnabled = "NO">
+            isEnabled = "YES">
```
ncreated, line 102: This should not be committed as it's modifying the shared scheme for everyone 😉.

ncreated (approved):
Beside one unnecessary change: looks very good 🚀 💪

PR #381, opened 2021-01-25
RUMM-1026 Use sysctl to gather a more accurate App Launch Time

acgh:
### What and why?

`ObjcAppLaunchHandler` is the existing mechanism that observes early steps of the [App Launch sequence](https://developer.apple.com/documentation/uikit/app_and_environment/responding_to_the_launch_of_your_app/about_the_app_launch_sequence) to provide a Launch Time for further reporting by `RUMViewScope` into `RUMActionEvent.Action` for the `start_application` event.

This currently relies on Objective-C's runtime initialisation which calls classes' `+(void)load` method as soon as they are loaded into the runtime. This happens before the `UIApplication.sharedInstance` is instantiated, and this provides an early point in time regarding the launch sequence.
The limitations are that the "earliness" is not guaranteed and depends on many factors, so it might lead to not accounting for some launch time.

This is why we aim to use a more accurate and stable source of information for the App Launch time.

Note that in that PR, we don't change the existing definition of App Launch Time as we still include the time to load the main UI, create/setup the app delegate. 

### How?

In that PR, we refactor the pair `ObjcAppLaunchHandler`/`AppLaunchTimer` into a single `AppLaunchHandler` class that becomes an implementation detail for a C function `AppLaunchTime()`.

The `AppLaunchTimer` class still implements and reacts to `+(void)load` to set up some observation on `UIApplication`.
Note the `AppLaunchTimer` is only kept alive by the capturing blocks until it has computed the times.

More specifically, `AppLaunchTimer` calls `sysctl()` to get a `kinfo_proc` for the current process (`getpid()`) and especially looks at `p_starttime`.

### Review checklist

- [x] Run the Example.app target in Simulator/on Device and check that the values are provided and make sense. 
- [ ] ~~Feature or bugfix MUST have appropriate tests (unit, integration)~~
  - This depends on Foundation/UIKit/System APIs that can't be mocked so this wouldn't be straightforward.

### DIFF
--- a/Sources/Datadog/Core/System/LaunchTimeProvider.swift
+++ b/Sources/Datadog/Core/System/LaunchTimeProvider.swift
@@ -16,10 +16,7 @@ internal protocol LaunchTimeProviderType {
 
 internal class LaunchTimeProvider: LaunchTimeProviderType {
     var launchTime: TimeInterval? {
-        // Following dynamic dispatch synchronisation mutes TSAN issues when running tests from CLI.
-        objc_sync_enter(self)
-        let time = ObjcAppLaunchHandler.launchTime()
-        objc_sync_exit(self)
+        let time = AppLaunchTime()
         return time > 0 ? time : nil
     }
 }
--- a/Sources/Datadog/Core/System/LaunchTimeProvider.swift
+++ b/Sources/Datadog/Core/System/LaunchTimeProvider.swift
@@ -16,7 +16,10 @@ internal protocol LaunchTimeProviderType {
 
 internal class LaunchTimeProvider: LaunchTimeProviderType {
     var launchTime: TimeInterval? {
+        // Even if AppLaunchTime() is using a lock behind the scenes, TSAN will report a data race if there are no synchronizations at this level.
+        objc_sync_enter(self)
         let time = AppLaunchTime()
+        objc_sync_exit(self)
         return time > 0 ? time : nil
     }
 }
--- a/Sources/_Datadog_Private/ObjcAppLaunchHandler.m
+++ b/Sources/_Datadog_Private/ObjcAppLaunchHandler.m
@@ -4,60 +4,94 @@
 * Copyright 2019-2020 Datadog, Inc.
 */
 
-#import <UIKit/UIKit.h>
 #import "ObjcAppLaunchHandler.h"
+#import <sys/sysctl.h>
+#import <UIKit/UIKit.h>
 
-@interface AppLaunchTimer : NSObject
-@property NSDate *frameworkLoadTime;
-// Knowing how this value is collected and that it is written only once, the use of
-// `atomic` property is enough for thread safety. It guarantees that neither partial
-// nor garbage value will be returned.
-@property (atomic) NSTimeInterval timeToApplicationBecomeActive;
+// `AppLaunchHandler` aims to track some times as part of the sequence described in Apple's "About the App Launch Sequence"
+// https://developer.apple.com/documentation/uikit/app_and_environment/responding_to_the_launch_of_your_app/about_the_app_launch_sequence
+
+// Note that the following properties will be set early enough in the App lifecycle to avoid any race conditions without any specific synchronization means.
+static NSTimeInterval TimeToApplicationFinishedLaunching = 0.0;
+static NSTimeInterval TimeToLoadApplicationRootUI = 0.0;
+
+@interface AppLaunchHandler : NSObject
 @end
 
-@implementation AppLaunchTimer
+@implementation AppLaunchHandler {
+    NSTimeInterval _processStartTime;
+    NSTimeInterval _applicationMainEndTime;
+    NSTimeInterval _frameworkLoadTime;
+}
 
-- (instancetype)init
++ (void)load {
+    // This is called at the `_Datadog_Private` load time, keep the work minimal
+    NSTimeInterval now = CFAbsoluteTimeGetCurrent();
+    // Schedule a call for the first turn of run loop before the main nib/storyboard is loaded:
+    [NSRunLoop.mainRunLoop performInModes:@[NSRunLoopCommonModes] block:^{
+        AppLaunchHandler *handler = [[self alloc] initWithFrameworkLoadTime:now];
+        [handler markEndOfUIApplicationMain];
+        [handler scheduleUIApplicationDidFinishLaunchingMarking];
+    }];
+}
+
+- (instancetype)initWithFrameworkLoadTime:(NSTimeInterval)frameworkLoadTime
 {
-    self = [super init];
-    if (self) {
-        NSDate *time = [NSDate new];
-        self.frameworkLoadTime = time;
-        self.timeToApplicationBecomeActive = -1;
-        [[NSNotificationCenter defaultCenter] addObserver:self
-                                                 selector:@selector(handleDidBecomeActiveNotification)
-                                                     name:UIApplicationDidBecomeActiveNotification
-                                                   object:nil
-         ];
+    if (self = [super init]) {
+        _frameworkLoadTime = frameworkLoadTime;
     }
     return self;
 }
 
-- (void)handleDidBecomeActiveNotification {
-    if (self.timeToApplicationBecomeActive > -1) { // sanity check & extra safety
-        return;
+- (void)markEndOfUIApplicationMain {
+    _applicationMainEndTime = CFAbsoluteTimeGetCurrent();
+}
+
+- (void)scheduleUIApplicationDidFinishLaunchingMarking {
+    id __block token = [NSNotificationCenter.defaultCenter
+                        addObserverForName:UIApplicationDidFinishLaunchingNotification
+                        object:nil
+                        queue:NSOperationQueue.mainQueue
+                        usingBlock:^(NSNotification *_){
+        [self markUIApplicationDidFinishLaunching];
+        [NSNotificationCenter.defaultCenter removeObserver:token];
+    }];
+}
+
+- (void)markUIApplicationDidFinishLaunching {
+    NSTimeInterval now = CFAbsoluteTimeGetCurrent();
+
+    [self queryProcessStartTime];
+
+    TimeToApplicationFinishedLaunching = now - _processStartTime;
+    TimeToLoadApplicationRootUI = now - _applicationMainEndTime;
+}
+
+- (void)queryProcessStartTime {
+    // Query the current process' start time:
+    // https://www.freebsd.org/cgi/man.cgi?sysctl(3)
+    // https://github.com/darwin-on-arm/xnu/blob/707bfdc4e9a46e3612e53994fffc64542d3f7e72/bsd/sys/sysctl.h#L681
+    // https://github.com/darwin-on-arm/xnu/blob/707bfdc4e9a46e3612e53994fffc64542d3f7e72/bsd/sys/proc.h#L97
+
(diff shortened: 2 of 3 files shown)


Sources/Datadog/Core/System/LaunchTimeProvider.swift
```diff
@@ -16,10 +16,7 @@ internal protocol LaunchTimeProviderType {
 
 internal class LaunchTimeProvider: LaunchTimeProviderType {
     var launchTime: TimeInterval? {
-        // Following dynamic dispatch synchronisation mutes TSAN issues when running tests from CLI.
-        objc_sync_enter(self)
```
acgh, line 20: I couldn't reproduce the TSAN issues locally, but I believe that if we don't involve any Objc objects (and call a C function) in that call paths there can't be any dynamic dispatches involved to begin with.

xgouchet (approved):
Really smart approach, we might reuse this on Android too !

Sources/_Datadog_Private/ObjcAppLaunchHandler.m
```diff
@@ -4,60 +4,94 @@
...
+
+// `AppLaunchHandler` aims to track some times as part of the sequence described in Apple's "About the App Launch Sequence"
+// https://developer.apple.com/documentation/uikit/app_and_environment/responding_to_the_launch_of_your_app/about_the_app_launch_sequence
 
-@interface AppLaunchTimer : NSObject
-@property NSDate *frameworkLoadTime;
-// Knowing how this value is collected and that it is written only once, the use of
-// `atomic` property is enough for thread safety. It guarantees that neither partial
-// nor garbage value will be returned.
-@property (atomic) NSTimeInterval timeToApplicationBecomeActive;
```
acgh, line 15: I believe `atomic` is already the default for a `@property`. That being said, I don't think this was necessary due the inherent sequencing of setting/getting this property throughout the App lifecycle.

acgh: So it looks like `LaunchTimeProviderTests.testThreadSafety()` disagrees with my assumptions 🤔

acgh:
Thanks @xgouchet for the swift review! I'll keep that PR open a little bit longer to get @buranmert and @ncreated's opinion on this change :)

Sources/_Datadog_Private/ObjcAppLaunchHandler.m
```diff
@@ -4,60 +4,94 @@
...
+    if (res == 0) {
+        // The process' start time is provided relative to 1 Jan 1970
+        struct timeval startTime = kip.kp_proc.p_starttime;
+        _processStartTime = startTime.tv_sec + startTime.tv_usec / USEC_PER_SEC;
+        // Convert to time since 1 Jan 2001 to align with CFAbsoluteTimeGetCurrent()
+        _processStartTime -= kCFAbsoluteTimeIntervalSince1970;
+    } else {
+        // Fallback to less accurate delta with DD's framework load time
+        _processStartTime = _frameworkLoadTime;
+    }
```
buranmert, line 90: do we know documented cases where `res != 0`?
this is quite a longshot but if `res` is practically always 0 and if we can find a way to execute code pre-main in Swift, we can call `sysctl` from Swift and remove this ObjC code. eventually, we can also remove `_Datadog_Private`

buranmert (approved):
i didn't know `sysctl` gives that info, TIL 👍

Sources/_Datadog_Private/ObjcAppLaunchHandler.m
acgh: As per the documentation it's only going to be 0 or -1:
```
Upon successful completion, the value 0 is	returned; otherwise the
     value -1 is returned and the global variable errno	is set to indicate the
     error.
```

In general, we should be able to call `sysctl()` at any time on the Swift side, but to get close to a relevant app launch time without the Objc `+(void)load` trick we should either ask for explicit calls from our users (which could be easily overlooked), or integrate this mechanism into the Datadog/singleton setup that is hopefully done as part of the `UIApplication.delegate` `application(...:didFinishLaunching...)` by our users.
An obvious downside of that is spreading the whole app launch time state across our SDK, along with potential discrepancies if our assumptions of Datadog's setup are wrong.

Sources/_Datadog_Private/ObjcAppLaunchHandler.m
```diff
@@ -4,60 +4,94 @@
...
+    NSTimeInterval _processStartTime;
+    NSTimeInterval _applicationMainEndTime;
+    NSTimeInterval _frameworkLoadTime;
+}
+
++ (void)load {
+    // This is called at the `_Datadog_Private` load time, keep the work minimal
+    NSTimeInterval now = CFAbsoluteTimeGetCurrent();
+    // Schedule a call for the first turn of run loop before the main nib/storyboard is loaded:
+    [NSRunLoop.mainRunLoop performInModes:@[NSRunLoopCommonModes] block:^{
```
acgh, line 31: Some piece of offline feedback @buranmert shared with me:
This run loop callback is specifically there as a mean to call `markEndOfUIApplicationMain` at the right time. This is a value that we will not use yet. So maybe it would be better to take note of this solution for later and get rid of this block (i.e. just `alloc`/`init` and call `scheduleUIApplicationDidFinishLaunchingMarking` inline).

Sources/Datadog/Core/System/LaunchTimeProvider.swift
ncreated: Sounds reasonable 👍. My suspicion behind this TSAN issue was indeed around the dynamic dispatch with this call.

Sources/_Datadog_Private/ObjcAppLaunchHandler.m
ncreated: IMO: if we don't use it → remove it as it's just a dead code. Do I miss something 🤔?

Sources/_Datadog_Private/ObjcAppLaunchHandler.m
```diff
@@ -4,60 +4,94 @@
...
+    _applicationMainEndTime = CFAbsoluteTimeGetCurrent();
+}
 
-    NSDate *time = [NSDate new];
-    NSTimeInterval duration = [time timeIntervalSinceDate:self.frameworkLoadTime];
-    self.timeToApplicationBecomeActive = duration;
-    [[NSNotificationCenter defaultCenter] removeObserver:self];
+- (void)scheduleUIApplicationDidFinishLaunchingMarking {
+    id __block token = [NSNotificationCenter.defaultCenter
+                        addObserverForName:UIApplicationDidFinishLaunchingNotification
```
ncreated, line 52: We've been using `UIApplicationDidBecomeActiveNotification` before, but now we use `UIApplicationDidFinishLaunchingNotification` - if they do not mark the same moment in time, this changes the definition of RUM "launch time" for iOS.

How is `UIApplicationDidFinishLaunchingNotification` different than the first `UIApplicationDidBecomeActiveNotification`?

Sources/_Datadog_Private/ObjcAppLaunchHandler.m
```diff
@@ -4,60 +4,94 @@
...
 
-@interface AppLaunchTimer : NSObject
-@property NSDate *frameworkLoadTime;
-// Knowing how this value is collected and that it is written only once, the use of
-// `atomic` property is enough for thread safety. It guarantees that neither partial
-// nor garbage value will be returned.
-@property (atomic) NSTimeInterval timeToApplicationBecomeActive;
+// Note that the following properties will be set early enough in the App lifecycle to avoid any race conditions without any specific synchronization means.
+static NSTimeInterval TimeToApplicationFinishedLaunching = 0.0;
+static NSTimeInterval TimeToLoadApplicationRootUI = 0.0;
```
ncreated, line 16: I don't see any usages of this value - we only set it, but never read / process. Do we need it?

ncreated (changes requested):
Good idea to use the process start time 👌. I left few questions and suggestions. Main blocker for me is the presence of code that adds complexity but is not used.

Sources/_Datadog_Private/ObjcAppLaunchHandler.m
acgh: After the `UIApplication.delegate` is done with the callbacks and `UIWindow` is made key and visible, the sequence of notifications will be `UIApplicationDidFinishLaunchingNotification` then `UIApplicationDidBecomeActiveNotification`. So I believe they happen close to each other (depending on potential observers and the work they do).

That being said `UIApplicationDidBecomeActiveNotification` is a repeating notification that is meant for the whole app lifecycle (although by removing the observer this is only fires initially) and `UIApplicationDidFinishLaunchingNotification` fits with the semantic of App Launch Time better.

I don't think it really changes the definition of RUM "launch time" but I'm ok using `UIApplicationDidBecomeActiveNotification` if that is safer as it's still related to the app first appearance anyway.

Sources/_Datadog_Private/ObjcAppLaunchHandler.m
acgh: You're right, I've sent this along for reference in case we consider this additional step eventually, but I'm going to clean this up in a new commit.

buranmert (approved):

Sources/_Datadog_Private/ObjcAppLaunchHandler.m
ncreated: Got it 👍. I also did some digging around and realised that "finished launching" relates to cold start, whereas "did become active" is for returning from background. Both should work fine and mean the same in our implementation 👌.

ncreated (approved):
LG 🎸!

Sources/_Datadog_Private/ObjcAppLaunchHandler.m
```diff
@@ -4,60 +4,81 @@
...
-// Knowing how this value is collected and that it is written only once, the use of
-// `atomic` property is enough for thread safety. It guarantees that neither partial
-// nor garbage value will be returned.
-@property (atomic) NSTimeInterval timeToApplicationBecomeActive;
-@end
+// `AppLaunchHandler` aims to track some times as part of the sequence described in Apple's "About the App Launch Sequence"
+// https://developer.apple.com/documentation/uikit/app_and_environment/responding_to_the_launch_of_your_app/about_the_app_launch_sequence
+
+// A Read-Write lock to allow concurrent reads of TimeToApplicationDidBecomeActive, unless the initial (and only) write is locking it.
+static pthread_rwlock_t rwLock;
```
buranmert, line 16: doesn't it work with `objc_sync_enter/exit` in `LaunchTimeProvider`? now we have a global lock staying around, not great :/

Sources/Datadog/Core/System/LaunchTimeProvider.swift
acgh: I actually had to come back to using `objc_sync_enter()`/`objc_sync_exit()`. I could not reproduce locally the test failure, but based on Bitrise logs, TSAN reported a potential data race on the call to `AppLaunchTime()` where a write happened according to TSAN. When reporting, TSAN crashes on a `SIGABRT`.

The only write in that C function might come from `pthread_rwlock_rdlock(&rwLock);` which in itself must be thread safe (otherwise nothing can be). So anyway, this is back here, it might be extraneous but let's trust TSAN for now.

PR #384, opened 2021-01-26
RUMM-722 Fix: firsty party url regex fixed

buranmert:
#### Issue 1
`testWhenURLHostDoesNotMatchEndingOfAnyOfUserDefinedHosts_itIsNotConsideredFirstParty` was failing randomly
if regex happens to be `^(.*\\.)*eu|firsty-party.com$` then test case `foo.eu.com` becomes false positive

#### Issue 2
now we have use cases in which `FirstPartyURLFilter` is init'ed with empty set
that results in `isFirstParty(url: any_url) == true`

#### Issue 1
regex now has `$` at the end of every first party host

#### Issue 2
`regex` is now optional and stays `nil` if `hosts` set is empty, this prevents `isFirstParty(any_url)` from returning true

### Review checklist


Sources/Datadog/URLSessionAutoInstrumentation/Interception/URLFiltering/FirstPartyURLsFilter.swift
```diff
@@ -13,9 +13,11 @@ internal struct FirstPartyURLsFilter {
     private let regex: String
 
     init(hosts: Set<String>) {
-        // pattern = "^(.*\\.)*tracedHost1|tracedHost2|...$"
-        let escapedHosts = hosts.map { NSRegularExpression.escapedPattern(for: $0) }.joined(separator: "|")
-        self.regex = "^(.*\\.)*\(escapedHosts)$"
+        // pattern = "^(.*\\.)*tracedHost1$|tracedHost2$|...$"
+        let escapedHosts = hosts
+            .map { "\(NSRegularExpression.escapedPattern(for: $0))$" }
```
acgh, line 18: Good catch!

Sources/Datadog/URLSessionAutoInstrumentation/Interception/URLFiltering/FirstPartyURLsFilter.swift
```diff
@@ -13,9 +13,11 @@ internal struct FirstPartyURLsFilter {
     private let regex: String
 
     init(hosts: Set<String>) {
```
acgh, line 15: If somehow the Set passed in here stores an order based on how it was build, especially as we build a Set from an Array in the related tests, and if you were able to spot the bug by playing with the input order in the test case. Should we also add a test case with the same input in a different order?

acgh (approved):
Thanks for catching this! 🐛

Sources/Datadog/URLSessionAutoInstrumentation/Interception/URLFiltering/FirstPartyURLsFilter.swift
ncreated: Not sure if I get it right, but as this is a `Set`, there should be no difference if running the test against `[h1, h2, h3]` or `[h3, h2, h1]` vectors. Also, providing different permutations of the same vector to initialize `Set` would be very unobvious.

Maybe, to solidify this test, we could run it multiple times? 
```swift
(0..<50).forEach { _ in
   // run the test
}
```
This could be even turned into convention for all tests working with order-agnostic inputs.

ncreated (approved):
Nice 🥇. LGTM - feel free to consider the comments on improving the test.

Sources/Datadog/URLSessionAutoInstrumentation/Interception/URLFiltering/FirstPartyURLsFilter.swift
buranmert: i too first thought something similar to @acgh 's proposal: passing an array instead of set and test different orders. but keeping it as set makes more sense to me. 
running it multiple times as @ncreated suggests sounds better as a patch (no guarantee for correctness but will probably work), i went that way ✅

ncreated (approved):

PR #382, opened 2021-01-25
RUMM-987 Prepare code generation for RUM @objc models

ncreated:
### What and why?

📦 This PR updates the `rum-models-generator` tool to generate `@objc` RUM data models. This is required to further expose scrubbing API in `DatadogObjc` as it was done for Swift (#367). 

### How?

I leveraged the existing architecture of `rum-models-generator`: 
* `read → transform → print`

by implementing another `Printer`:

* `read → transform → SwiftPrinter`
* `read → transform → ObjcInteropPrinter`

The `SwiftType` schema generated and transformed for Swift code generation is now used by `ObjcInteropReader` to generate `ObjcInteropType` schema. The schema is later passed to the `ObjcInteropPrinter` for printing the actual code.

### Swift / Obj-c interoperability for RUM

`@objc` interoperability is done by wrapping Swift `structs` in `@objc` classes and exposing Swift `enums` as `@objc enum` .

Simple example can be illustrated as:
```swift
// MARK: - Swift

public struct Foo {
    public let immutableString: String
    public let optionalImmutableString: String?
    public var mutableString: String
    public var optionalMutableString: String?
}

// MARK: - ObjcInterop

@objc
public class DDFoo: NSObject {
    internal var swiftModel: Foo
    internal var root: DDFoo { self }

    internal init(swiftModel: Foo) {
        self.swiftModel = swiftModel
    }

    @objc public var immutableString: String {
        root.swiftModel.immutableString
    }

    @objc public var optionalImmutableString: String? {
        root.swiftModel.optionalImmutableString
    }

    @objc public var mutableString: String {
        set { root.swiftModel.mutableString = newValue }
        get { root.swiftModel.mutableString }
    }

    @objc public var optionalMutableString: String? {
        set { root.swiftModel.optionalMutableString = newValue }
        get { root.swiftModel.optionalMutableString }
    }
}
```

Nested `structs` are managed by transitive `@objc classes`:
```swift
// MARK: - Swift

public struct Foo {
    public var bar: Bar

    public struct Bar {
        public let property: Int
    }
}

// MARK: - ObjcInterop

@objc
public class DDFoo: NSObject {
    internal var swiftModel: Foo
    internal var root: DDFoo { self }

    internal init(swiftModel: Foo) {
        self.swiftModel = swiftModel
    }

    @objc public var bar: DDFooBar {
        DDFooBar(root: root)
    }
}

@objc
public class DDFooBar: NSObject {
    internal let root: DDFoo

[...]

PR #386, opened 2021-01-26
RUMM-987 Validate RUM Objc models in a CI job

ncreated:
### What and why?

🚚 This PR closes the topic of RUM Scrubbing API for Objc. It updates `make rum-models-verify` and `make rum-models-generate` commands to also validate and generate `RUMDataModels+objc.swift` file.

It also brings the refactoring discussed with @acgh in #382 (I add it here instead of the original PR to keep the orignal one relatively small).

### How?

It's done in commit-by-commit fashion:
* https://github.com/DataDog/dd-sdk-ios/commit/d799fae5694027c7fb5601594223c7ae0b7f4c5e and https://github.com/DataDog/dd-sdk-ios/commit/45fd39cda768ab62b743d0de7cb143dfb6dcc3af are purely refactoring PRs to streamline the architecture of `rum-models-generator`:
```swift
let swiftModels = JSONSchemaReader().read(schemaFiles)
   → JSONSchemaToJSONTypeTransformer().transform(jsonSchemas)
   → JSONToSwiftTypeTransformer().transform(jsonTypes)
   → RUMSwiftTypeTransformer().transform(swiftTypes)

SwiftPrinter().print(swiftModels)

let objcInteropModels = SwiftToObjcInteropTypeTransformer().transform(swiftModels)

ObjcInteropPrinter().print(models)
```
* https://github.com/DataDog/dd-sdk-ios/commit/cf80c5ef928962fe8d1b0b866cca9870b514e304 updates the tool to print `@objc` file with RUM models,
* https://github.com/DataDog/dd-sdk-ios/commit/04da53680bb49bc9614aa5d967d8926e17ab8bfd updates the `tools/rum-models-generator/run.sh` to handle both Swift and Objc files.

### Review checklist


tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Output/Printing/ObjcInteropPrinter.swift
```diff
@@ -6,9 +6,9 @@
 
 import Foundation
 
-/// Generates Swift code for Obj-c interoperability for given `SwiftStructs`.
+/// Generates Swift code for Obj-c interoperability for given `ObjcInteropType` schemas.
 ///
-/// E.g. given `SwiftStruct` describing:
+/// E.g. given `ObjcInteropType` describing Swift struc:
```
acgh, line 11: Typo
```suggestion
/// E.g. given `ObjcInteropType` describing Swift struct:
```

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/Swift/SwiftType+reflection.swift
```diff
@@ -0,0 +1,32 @@
+/*
+* Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+* This product includes software developed at Datadog (https://www.datadoghq.com/).
+* Copyright 2019-2020 Datadog, Inc.
+*/
+
+import Foundation
+
+// MARK: - Reflection Helpers
```
acgh, line 9: It's a good idea to extract those out into a dedicated file!

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Output/Printing/SwiftPrinter.swift
```diff
@@ -6,20 +6,12 @@
...
 /// Generates Swift code from `SwiftTypes`.
-public class SwiftPrinter: Printer, SwiftCodePrinter {
-    // MARK: - Printing
-
+public class SwiftPrinter: BasePrinter {
     public func print(swiftTypes: [SwiftType]) throws -> String {
         reset()
 
         try swiftTypes.forEach { type in
-            precondition(indentationLevel == 0)
```
acgh, line 22: What's the reason behind removing this check?

tools/rum-models-generator/run.sh
```diff
@@ -50,30 +68,49 @@ cd -
 swift build --configuration release
 GENERATOR=".build/x86_64-apple-macosx/release/rum-models-generator"
 
-# Generate file in temporary location
+# Generate RUM models (Swift) file in temporary location
+mkdir -p ".temp"
+GENERATED_SWIFT_FILE=".temp/RUMDataModels.swift"
+$GENERATOR generate-swift --path "rum-events-format/schemas" > $GENERATED_SWIFT_FILE
+echo "// Generated from https://github.com/DataDog/rum-events-format/tree/$SHA" >> $GENERATED_SWIFT_FILE
```
acgh, line 75: Nitpick: I assume this reference is meant for us human, if that's the case maybe prepend that comment to the beginning of the file to make any manual checks faster.

acgh (approved):
I've left a few minor comments but that looks solid to me! This PR really helps understanding the model generator pipeline.

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Output/Printing/SwiftPrinter.swift
ncreated: It was quite handy for development, but as now the generated Swift syntax is asserted in `SwiftPrinterTests` I thought this `precondition()` is unnecessary.

tools/rum-models-generator/run.sh
ncreated: In `RUMModelsGenerator.swift`, when building the actual output for `*.swift` and `*+objc.swift` files, we print this comment along with the license header, `imports` etc.:
```swift
// This file was generated from JSON Schema. Do not modify it directly.
```
I think this is exactly what you mean, no?

ncreated: It's both for us and for `run.sh` to validate generated files. If running the tool with `verify` option (instead of `generate`), it reads the tail of the generated file, captures this `$SHA`, creates temporary models for that `$SHA` and compares if the `diff <original> <temporary>` is empty.

It is just easier to print it at the end of the file than in the middle. Printing in the middle would require understanding the file structure or at least printing-in some token to replace it later later (e.g. `s/<SHA-HERE>/$SHA/g`) here in `.sh`.

At the very top of each file we print this:
```swift
// This file was generated from JSON Schema. Do not modify it directly.
```
to highlight that this file is auto-generated. To find the signature, we need to scroll to the bottom.

I can generate the signature at the top of the file (after the license), but this would require some work - I'm not sure if it's worth it 🙂. WDYT?

buranmert (approved):

PR #388, opened 2021-01-27
RUMM-647 Add a Span.setError convenience API for Swift on top of Tracing

acgh:
### What and why?

Currently as part of our OpenTracing integration, we provide log fields that can be set on a Span to attach an error to that Span. It's usable but we seek to provide a more convenient way to allow passing some an `Error` conforming type, an `NSError` object or some message and details to log an error using the proper log fields as described by OpenTracing.

In this PR, we add 2 convenience APIs on `OTSpan` (in Swift) to facilitate attaching an error onto the given `OTSpan` using the right field keys.

### How?

This PR introduces an extension on `OTSpan` and piggyback on the existing `log()` as well as `OTLogFields` to properly attach the provided information. It uses `DDError` to extract information out of `NSError` or any object conforming to the `Error` protocol.
It populates the `OTLogFields.stack` field with at least a `#fileId` and `#line`, and the provided stack if any.

### Review checklist

- [x] Run `make api-surface`

### Next

- Add the Objc counterpart API
- Use the API in AutoInstrumentation (e.g. URLSession)
- Use the API in the mobile app

### DIFF
--- a/Sources/Datadog/OpenTracing/OTSpan.swift
+++ b/Sources/Datadog/OpenTracing/OTSpan.swift
@@ -74,3 +74,52 @@ public extension OTSpan {
         self.finish(at: Date())
     }
 }
+
+/// Error conveniences
+public extension OTSpan {
+    /// Set or replace the error for the given span.
+    /// This is a convenience to set the proper tags with error details. Consider the `setError(message:stacktrace:file:line:)` variant for a better control over the error details.
+    ///
+    /// - parameter error: An object conforming to the `Error` protocol.
+    /// - parameter file: A string identifying the file where the `Error` was caught. The default is `#fileID` which means `ModuleName/Filename.extension`, consider an helpful yet concise identifier when overriding the default.
+    /// - parameter line: The line number in the file where the `Error` was caught.
+    func setError(
+        _ error: Error,
+        file: StaticString = #fileID,
+        line: UInt = #line
+    ) {
+        let dderror = DDError(error: error)
+        setError(
+            kind: dderror.title,
+            message: dderror.message,
+            stacktrace: dderror.details,
+            file: file,
+            line: line
+        )
+    }
+
+    /// Set or replace the error for the given span.
+    /// This is a convenience to set the proper tags with error details.
+    ///
+    /// - parameter kind: The type of error to be logged.
+    /// - parameter message: An error message to be logged.
+    /// - parameter stacktrace: A string detailing the state of the stack when the error was caught. Note that it can also be any details that could help further triaging and investigation of the error downstream, it doesn't have to be an actual stack trace.
+    /// - parameter file: A string identifying the file where the error was caught. The default is `#fileID` which means `ModuleName/Filename.extension`, consider an helpful yet concise identifier when overriding the default.
+    /// - parameter line: The line number in the file where the error was caught.
+    func setError(
+        kind: String,
+        message: String,
+        stacktrace: String = "",
+        file: StaticString = #fileID,
+        line: UInt = #line
+    ) {
+        log(
+            fields: [
+                OTLogFields.event: "error",
+                OTLogFields.errorKind: kind,
+                OTLogFields.message: message,
+                OTLogFields.stack: stacktrace + " \(file):\(line)"
+            ]
+        )
+    }
+}
--- a/Sources/Datadog/OpenTracing/OTSpan.swift
+++ b/Sources/Datadog/OpenTracing/OTSpan.swift
@@ -92,7 +92,7 @@ public extension OTSpan {
         setError(
             kind: dderror.title,
             message: dderror.message,
-            stacktrace: dderror.details,
+            stack: dderror.details,
             file: file,
             line: line
         )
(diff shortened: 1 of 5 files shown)


api-surface-swift
```diff
@@ -75,6 +75,7 @@ public class Datadog
    public func set(tracesEndpoint: TracesEndpoint) -> Builder
    public func set(tracedHosts: Set<String>) -> Builder
    public func track(firstPartyHosts: Set<String>) -> Builder
+   public func trackURLSession(firstPartyHosts: Set<String> = []) -> Builder
```
acgh, line 78: I'm not sure why this was added here.

api-surface-swift
```diff
@@ -561,7 +565,7 @@ public class HTTPHeadersWriter: OTHTTPHeadersWriter
  public private(set) var tracePropagationHTTPHeaders: [String: String] = [:]
  public func inject(spanContext: OTSpanContext)
  override public init()
- public init(firstPartyHosts: Set<String>)
+ public init(additionalFirstPartyHosts: Set<String>)
```
acgh, line 568: Ditto

Sources/Datadog/OpenTracing/OTSpan.swift
```diff
@@ -74,3 +74,56 @@ public extension OTSpan {
...
+public extension OTSpan {
+    /// Set or replace the error for the given span.
+    /// This is a convenience to set the proper tags with error details. Consider the `setError(message:stacktrace:file:line:)` variant for a better control over the error details.
+    ///
+    /// - parameter error: An object conforming to the `Error` protocol.
+    /// - parameter file: A string identifying the file where the `Error` was caught. The default is `#fileID` which means `ModuleName/Filename.extension`, consider an helpful yet concise identifier when overriding the default.
+    /// - parameter line: The line number in the file where the `Error` was caught.
+    func setError(
+        _ error: Error,
+        file: StaticString = #fileID,
```
ncreated, line 88: Nice, I didn't know this new `#fileID`. Regarding the [Expressions doc](https://docs.swift.org/swift-book/ReferenceManual/Expressions.html) I think we may want to apply it wider in our codebase. WDYT about creating separate JIRA to incorporate this change in other places in the future?

Sources/Datadog/OpenTracing/OTSpan.swift
```diff
@@ -74,3 +74,56 @@ public extension OTSpan {
...
+        message: String,
+        stack: String = "",
+        file: StaticString = #fileID,
+        line: UInt = #line
+    ) {
+        var stackWithFile = "\(file):\(line)"
+        if stack.count > 0 {
+            stackWithFile += "\n" + stack
+        }
+        log(
```
ncreated, line 120: As this uses Logging feature under the hood, maybe we can add a note on its availability in those 2 API comments? Otherwise, if someone configures the SDK with explicitly disabling logging: `.enableLogging(false)`, this error won't be send and the warning printed to the console will be:
```
"The log for span <span> will not be send, because the Logging feature is disabled."
```
This message makes it unclear: what log if `span.setError()` API was called.

api-surface-swift
ncreated: It's because we forgot to run `make api-surface` on the PR adding this API. Good to list it now 👍.

Tests/DatadogTests/Datadog/Tracing/DDSpanTests.swift
```diff
@@ -46,6 +60,101 @@ class DDSpanTests: XCTestCase {
...
+
+        span.setError(ErrorMock())
+
+        XCTAssertEqual(span.logFields.count, 1)
+        let logFields = span.logFields.first!
+        XCTAssertNotEqual(logFields.count, 0)
+        try XCTAssertEqual(logFields.otEvent(), "error")
+        let spanErrorStack = try logFields.otStack()
+        let fileName = #fileID.components(separatedBy: "/").last!
+        XCTAssertTrue(spanErrorStack.contains(fileName))
```
ncreated, line 77: If we use `#sourceLocation`:
```swift
#sourceLocation(file: "/user/abc/Foo.swift", line: 100)
span.setError(ErrorMock())
#sourceLocation()
```
this will become:
```swift
try XCTAssertEqual(logFields.otStack(), "Foo.swift: 100")
```

ncreated (commented):
Looks very good 💪. I left few suggestions we may want to consider.

ncreated (approved):

Sources/Datadog/OpenTracing/OTSpan.swift
acgh: I think that it's a great idea to do a follow-up pass on this. I'll create a JIRA for that!

acgh: This is tracked by RUMM-1035.

Sources/Datadog/OpenTracing/OTSpan.swift
acgh: Maybe it's confusing but I think this `log()` API is part of OpenTracing (especially `OTSpan`) and only requires to enable tracing, which if the user is being consistent should already be enabled prior to using `OTSpan`/`DDSpan`.
Is my assumption correct? Happy to add the comment otherwise.

Tests/DatadogTests/Datadog/Tracing/DDSpanTests.swift
acgh: That's a great tip, thanks!

Sources/Datadog/OpenTracing/OTSpan.swift
buranmert: i'm _slightly_ for adding a note at the end of comments
and the same for `setError(_ error: Error...)` above

buranmert (approved):
this should make it much easier to deal with errors 🎉

Sources/Datadog/OpenTracing/OTSpan.swift
ncreated: > Maybe it's confusing but I think this log() API is part of OpenTracing (especially OTSpan) and only requires to enable tracing, which if the user is being consistent should already be enabled prior to using OTSpan/DDSpan.
Is my assumption correct?

Yes `span.log()` is part of Open Tracing API. However, our implementation of this API uses the `Logging` feature under the hood. When the `span.log()` is called, we send a real `Log` to Datadog Logging intake, using integration with `LoggingFeature` (this `Log` contains `spanID` and `traceID`, so we can correlate it on backend).

This means that initializing the SDK with `.enableLogging(false)` and using `span.log()` will print the warning to the user. The warning is defined here:

https://github.com/DataDog/dd-sdk-ios/blob/71b46468396adbd101adc9a4309a5707ad43e4ce/Sources/Datadog/Tracer.swift#L209-L216

and is part of the control flow executed for `span.log()`.

acgh: Got it! I'll add a comment about that.

PR #389, opened 2021-01-28
RUMM-1034 Fix running integration tests on real device

ncreated:
### What and why?

🧪 This PR fixes the runtime issue with running integration tests on physical device:
```
Library not loaded: @rpath/Datadog.framework/Datadog
```
Unfortunately it doesn't fully unlock testing on real iPhone, due to a potential bug in `UITests` explained at the end.

### How?

The original `Library not loaded: @rpath/Datadog.framework/Datadog` originated from referencing those source files in `DatadogIntegrationTests` target:
```diff
- LoggingScenarios.swift
- TracingScenarios.swift
- RUMScenarios.swift
- TrackingConsentScenarios.swift
- URLSessionScenarios.swift
```
All of them do `import Datadog`. The `Datadog.framework` doesn't need to be referenced in the `DatadogIntegrationTests` as we do nothing with it (we just obtain HTTP payloads sent from the `Example` app and run assertions).

Similar, `Kronos.framework` doesn't need to be linked to `DatadogIntegrationTests`, so I'm removing this reference as well.

De-referenced files were used in `DatadogIntegrationTests` for type-safe passing of the individual test scenario name, e.g.:
```swift
// In `XCTestCase` file:
app.launchWith(
   testScenario: TrackingConsentStartGrantedScenario.self,
   // ...
)
```
I refactored it to use class name:
```swift
app.launchWith(
   testScenarioClassName: "TrackingConsentStartGrantedScenario",
   // ...
)
```
The type safety is not important here, because the class name consistency is fundamental thing in each test execution - the test will not pass locally and on CI if the name is mistaken. Also, we used to use class names in `schema` configuration to run `Example` app for each scenario, so now this setup is more coherent.

### Remaining (although non-blocking) issue with `UITests`

It is still not possible to run integration tests on device. It appears that since iOS 14 (and introduction of [`NSLocalNetworkUsageDescription`](https://developer.apple.com/documentation/bundleresources/information_property_list/nslocalnetworkusagedescription)) it is not possible to send `HTTP` network requests to a server running in the local network (all result with `Error Domain=NSURLErrorDomain Code=-1009 "The internet connection appears to be offline." `).

I addressed this topic by contributing to this thread on Apple Forum:
* https://developer.apple.com/forums/thread/668729

and opening the radar:
* https://openradar.appspot.com/radar?id=5033806549483520

Anyway, IMO it doesn't block this PR, because it fixes all the issues we're in charge of.

### Review checklist


Tests/DatadogIntegrationTests/Scenarios/Logging/LoggingScenarioTests.swift
```diff
@@ -14,7 +14,7 @@ class LoggingScenarioTests: IntegrationTests, LoggingCommonAsserts {
 
         let app = ExampleApplication()
         app.launchWith(
-            testScenario: LoggingManualInstrumentationScenario.self,
+            testScenarioClassName: "LoggingManualInstrumentationScenario",
```
acgh, line 17: I think it's fair to go with this more dynamic approach. As you said very well in the PR description, a mismatch should be caught early by running the tests.

acgh (approved):
Nice clean-up and great PR description!

buranmert (approved):
we had a similar problem in Shopist and we can apply the same solution if one day we need more control over scenarios:
1. we share `Scenario.swift` source between Shopist and its UI Target
2. `Scenario` is `Codable`: we encode it in UI Test and pass it via launch arguments to app target where we decode it back to a `Scenario` instance

ncreated:
> 1. we share `Scenario.swift` source between Shopist and its UI Target
> 2. `Scenario` is `Codable`: we encode it in UI Test and pass it via launch arguments to app target where we decode it back to a `Scenario` instance

I've been considering this as well, but then realised how cumbersome this would be in case of `Example` app instrumentation. Adding new scenario would mean:
* 1/ defining it's class (type-safe),
* 2/ defining class name in `Scenarios` constants (type-safe),
* 3/ using the defined constant in UITests's `XCTest` (type-safe),
* 4/ listing the constant's value in ENV parameters (not type-safe).

As there's still one type-unsafe element which we cannot get rid of (4), I decided to go with simpler way by removing 2nd point from the way. And we still use the same `: Codable` approach to pass ENV variables safely from UITest runner to app under tests 👌.

PR #392, opened 2021-01-29
RUMM-647 Use OTSpan.setError for AutoInstrumentation

acgh:
### What and why?

In #388, we introduced a convenience API to attach an error on a `OTSpan`/`DDSpan`. This PR ensures that this API is used (replacing some previous logic based on tags) for AutoInstrumentation of `URLSession`.

### How?

`URLSessionTracingHandler` consumes errors caught by the Interception mechanism attached to `URLSession`, and creates Spans with errors accordingly.

In this PR we replace the previous logic that used `OTSpan.setTag()` with some `DDTags` values, by using `OTSpan.setError()` with some `Error` instance.

We introduce an extension on `HTTPURLResponse` to return an `Error` when possible, based on its `statusCode`.

This PR also updates existing unit and integration tests (some that could have been updates before that PR).

### Review checklist


Sources/Datadog/Tracing/AutoInstrumentation/URLSessionTracingHandler.swift
```diff
@@ -55,20 +55,17 @@ internal class URLSessionTracingHandler: URLSessionInterceptionHandler {
...
-
-            if httpResponseStatusCode >= 400 && httpResponseStatusCode < 500 {
-                span.setTag(key: OTTags.error, value: true)
-                if httpResponseStatusCode == 404 {
+        if let httpResponse = resourceCompletion.httpResponse {
+            let httpStatusCode = httpResponse.statusCode
+            span.setTag(key: OTTags.httpStatusCode, value: httpStatusCode)
+            if httpStatusCode >= 400 && httpStatusCode < 500 {
+                if let error = httpResponse.asError() {
+                    span.setError(error)
```
ncreated, line 66: As we're capturing the `file` and `line` automatically as a part of the `setError()` API, this will be always giving `Datadog/URLSessionTracingHandler.swift` for the customers (as on the picture below). Do we consider this as a problem? IMO this could be slightly misleading (especially when aggregating "all errors" by their "stack", all auto instrumented network errors will lead to our SDK's stack trace). Is it possible to not pass this information?

To seed this span, enable `TracingURLSessionScenario` in `Example's` app `ENV`, then run it:

Screenshot 2021-02-01 at 11 06 01

Tests/DatadogIntegrationTests/Scenarios/Tracing/TracingManualInstrumentationScenarioTests.swift
```diff
@@ -74,7 +74,7 @@ class TracingManualInstrumentationScenarioTests: IntegrationTests, TracingCommon
 
         // Assert logs requests
         let recordedLoggingRequests = try loggingServerSession.pullRecordedRequests(timeout: dataDeliveryTimeout) { requests in
-            try LogMatcher.from(requests: requests).count == 1
+            try LogMatcher.from(requests: requests).count == 2
```
ncreated, line 77: Let's assert important stuff on this new `Log` 🙌 using `LogMatcher`.

Sources/Datadog/Core/Utils/DDError.swift
```diff
@@ -42,3 +42,13 @@ private func isNSErrorOrItsSubclass(_ error: Error) -> Bool {
...
     return false
 }
+
+internal extension HTTPURLResponse {
+    func asError() -> Error? {
+        guard statusCode >= 400 && statusCode < 500 else {
+            return nil
+        }
+        let message = "\(statusCode) " + HTTPURLResponse.localizedString(forStatusCode: statusCode)
+        return NSError(domain: "HTTPURLResponse", code: statusCode, userInfo: [NSLocalizedDescriptionKey: message])
```
ncreated, line 52: I'm not sure on the use of `localizedString` and `NSLocalizedDescriptionKey` here. Using those for the error message makes it impossible to aggregate data across entire user base, allowing only for per-locale aggregation. We should use non-localized descriptions if possible.

ncreated (changes requested):
Great that we infer more error information from the `HTTPURLResponse` (I can see that the auto-instrumented span errors that we send from `Example` app's `TracingURLSessionScenario` are now richer 💪).

Left few suggestions and raised one important point about error messages localization.

Sources/Datadog/Core/Utils/DDError.swift
buranmert: `NSError` has pretty much everything localized and we use it in `DDError` too. i think it's consistent to use `localizedString` here. aggregation shouldn't/can't be done with `userInfo/message`, they should use other values like `title/domain/code`

Sources/Datadog/Tracing/AutoInstrumentation/URLSessionTracingHandler.swift
```diff
@@ -55,20 +55,17 @@ internal class URLSessionTracingHandler: URLSessionInterceptionHandler {
...
-            span.setTag(key: OTTags.httpStatusCode, value: httpResponseStatusCode)
-
-            if httpResponseStatusCode >= 400 && httpResponseStatusCode < 500 {
-                span.setTag(key: OTTags.error, value: true)
-                if httpResponseStatusCode == 404 {
+        if let httpResponse = resourceCompletion.httpResponse {
+            let httpStatusCode = httpResponse.statusCode
+            span.setTag(key: OTTags.httpStatusCode, value: httpStatusCode)
+            if httpStatusCode >= 400 && httpStatusCode < 500 {
+                if let error = httpResponse.asError() {
```
buranmert, line 65: 400 < status code < 500 logic is in `asError()` method already, so we can remove the first `if` here

buranmert (changes requested):

Sources/Datadog/Tracing/AutoInstrumentation/URLSessionTracingHandler.swift
acgh: I thought about that as well but left the outer `if` to keep it explicit what we were checking for. You're right it's extraneous.
I'll rename `asError` to something like `asClientError` to remind of the 400 <= status code < 500 logic.

Sources/Datadog/Core/Utils/DDError.swift
acgh: That's a good point @ncreated, it's definitely going to appear in different languages but I assumed it was an ok expectations from our users (I've seen this many times with Crashlytics). In general it's likely that some instances of a an error will appear in `en_*`.
@buranmert also raises a good point, `DDError` has this localization problem already here while it potentially provides helpful information in the current locale.

In our case, the description is kind of a nice-to-have and not really required. HTTP status codes are well documented and it's not hard to get the description manually.

I think that it's reasonable to not include that part. The question remains then what is relevant for the message?

Sources/Datadog/Tracing/DDSpan.swift
```diff
@@ -59,6 +59,24 @@ internal class DDSpan: OTSpan {
...
+        logFields: [String: Encodable] = [:]
+    ) {
+        self.init(
+            tracer: tracer,
+            context: context,
+            operationName: operationName,
+            startTime: startTime,
+            tags: tags
+        )
+        unsafeLogFields.append(contentsOf: [logFields])
```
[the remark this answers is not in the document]
acgh: Yes, that sounds like a cleaner approach.

Sources/Datadog/Core/Utils/DDError.swift
ncreated: Maybe this is sth we should have an overall look into in `RUMM-1042`. I left a note in this ticket to look holistically into errors analytics given the localization aspect.

Sources/Datadog/OpenTracing/OTSpan.swift
```diff
@@ -91,18 +91,21 @@ public extension OTSpan {
     /// - parameter error: An object conforming to the `Error` protocol.
     /// - parameter file: A string identifying the file where the `Error` was caught. The default is `#fileID` which means `ModuleName/Filename.extension`, consider an helpful yet concise identifier when overriding the default.
     /// - parameter line: The line number in the file where the `Error` was caught.
+    /// - parameter includeFileInStack: Whether or not the `file` and `line` should be included in the stack, the default is `true` and can be overriden if the `file`/`line` are extraneous.
     func setError(
         _ error: Error,
         file: StaticString = #fileID,
-        line: UInt = #line
+        line: UInt = #line,
+        includeFileInStack: Bool = true
```
ncreated, line 99: I'm not a huge fan of [flag arguments], especially in the public API if we add it only for our own need. I see 2 options to get rid of it:

1/ Add `includeFileInStack` in `internal` API and call it with `false` from our code and `true` from `public` API.

2/ Remove this argument, force the `file:` and `line:` values in our custom code with `#sourceLocation`:
```swift
// URLSessionTracingHandler.swift

#sourceLocation(file: "", line: 0)
span.setError(error)
#sourceLocation()
```
and skip it here in `OTSpan` implementation if the `file` is `""`.

WDYT?

acgh: Yes I agree that it's not a good path to take for a public API and I previously considered the option #2 but that means assuming a rule (i.e. `file.count == 0` means skip it), it would be implicit as far as the API is concerned and I'm not sure that this is sound either.

I like the option 1 better, it keeps things clear at the public/internal API levels while going through the same code path.

An alternative that I have considered was to expect an Optional-like enum instead of a `StaticString` that defaults to `#fileID`, but I discarded it because it was too complex... Something that looked like `line: .none` or `line: .some("...")` and default `line: .some(#filedID)`. 

I think the more sensible option is 1 at this point. Thanks for suggesting this!

acgh: Well, don't mind my previous hesitation, I already included a similar choice as far as the `stack` is concerned (skipped if empty)!
Option 2 shouldn't be a problem after all :)

ncreated (approved):
👌 ✨

buranmert (approved):

PR #393, opened 2021-01-29
RUMM-961 DDError.title is used in RUMEvents.error.type

buranmert:
### What and why?

As part of error tracking, `type: String?` is introduced to `RUMErrorEvent.Error`
This value is expected to be the same with `error.kind` in `Logger`
`Logger` uses `DDError.title` as `error.kind`

### How?

`DDError.title` was already used in `RUMErrorEvent.Error.message` property
Now we basically use the same value in `message` and `type` properties of the same instance

### Review checklist


acgh (approved):
For context this is a first step we decided to take before thinking about/implementing a better structure for the input of `DDError`s, which ideally would avoid repeating information like in that case `title` and `message`.
This PR looks good to me 🚀

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Output/Printing/SwiftPrinter.swift
```diff
@@ -87,7 +87,11 @@ public class SwiftPrinter: Printer, SwiftCodePrinter {
         }
     }
 
-    private func printCodingKeys(for properties: [SwiftStruct.Property]) throws {
+    private func printCodingKeys(for properties: [SwiftStruct.Property]) {
+        // TODO: RUMM-1000 should remove this early return clause
+        if properties.isEmpty {
+            return
+        }
```
ncreated, line 94: This is not the right place for this patch. `rum-models-generator's` architecture [is build around the stream](https://github.com/DataDog/dd-sdk-ios/pull/386):
```swift
let rumModels = [*.json] → JSONSchema → JSONType → SwiftType -> RUM SwiftType
```
`rumModels` are then passed to `SwiftPrinter()` and `ObjcInteropPrinter()` for code generation. Instead of patching the printer(s), we should **dismiss unsupported types as early as possible in the stream**, so that it is not processed further causing unexpected behaviours. 

Ideally, we should cut it off in `JSONSchemaReader` by explicitly deleting schemas which use unsupported `additionalProperties` modifier. 

As this is temporary and short-living patch, fixing it in `RUMSwiftTypeTransformer` by filtering out the `SwiftStruct` named `"CustomTimings"` might be easier.

In any way, this should not reach the code generation, as it generates potentially wrong API (we may need to support `additionalProperties` in other way than a `struct`).

ncreated (changes requested):
Few comments left. Also, this PR musts be rebased onto recent `master`, so it generated `@objc` interop code too.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
```diff
@@ -203,10 +203,11 @@ internal class RUMResourceScope: RUMScope {
                     url: resourceURL
                 ),
                 source: command.errorSource.toRUMDataFormat,
-                stack: command.stack
+                stack: command.stack,
+                type: command.errorMessage
```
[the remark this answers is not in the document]
buranmert: yes i thought about that but it makes more sense to make those changes in 1024

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/RUM/RUMSwiftTypeTransformer.swift
```diff
@@ -94,7 +94,7 @@ internal class RUMSwiftTypeTransformer: TypeTransformer<SwiftType> {
 
         var `struct` = `struct`
         `struct`.name = format(structName: `struct`.name)
-        `struct`.properties = try `struct`.properties.map { try transform(structProperty: $0) }
+        `struct`.properties = try `struct`.properties.map { try transform(structProperty: $0) }.filter { property in property.name != "customTimings" }
```
acgh, line 97: Maybe add back the `TODO: RUMM-1000` here?

buranmert: oh yes definitely, thanks 🙏

ncreated (approved):
LGT 👌

PR #394, opened 2021-01-29
RUMM-908 Fix: Unit test

buranmert:
### What and why?

`testWhenNumberOfAttributesExceedsLimit_itDropsExtraOnes` unit test case sometimes gives false negatives

False negative case example:
`sanitized.customViewTimings.count == 49`
`sanitized.userInfoAttributes.count == 58`

### How?

We want to make sure if removals are done in the right order, this is unrelated to `count` comparisons.
Now we compute the expected item counts and `assertEqual` with them.

### Review checklist

### DIFF
--- a/Tests/DatadogTests/Datadog/RUM/RUMEvent/RUMEventSanitizerTests.swift
+++ b/Tests/DatadogTests/Datadog/RUM/RUMEvent/RUMEventSanitizerTests.swift
@@ -141,18 +141,18 @@ class RUMEventSanitizerTests: XCTestCase {
             let sanitized = RUMEventSanitizer().sanitize(event: event)
 
             // Then
-            XCTAssertEqual(
-                sanitized.attributes.count + sanitized.userInfoAttributes.count + (sanitized.customViewTimings?.count ?? 0),
-                AttributesSanitizer.Constraints.maxNumberOfAttributes
-            )
-            XCTAssertTrue(
-                (sanitized.customViewTimings?.count ?? 0) >= sanitized.userInfoAttributes.count,
-                "If number of attributes needs to be limited, `userInfoAttributes` are removed prior to `customViewTimings`."
-            )
-            XCTAssertTrue(
-                sanitized.userInfoAttributes.count >= sanitized.attributes.count,
-                "If number of attributes needs to be limited, `attributes` are removed prior to `userInfoAttributes`."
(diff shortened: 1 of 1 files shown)


Tests/DatadogTests/Datadog/RUM/RUMEvent/RUMEventSanitizerTests.swift
```diff
@@ -141,18 +141,18 @@ class RUMEventSanitizerTests: XCTestCase {
...
+            var remaining = AttributesSanitizer.Constraints.maxNumberOfAttributes
+            let expectedSanitizedCustomTimings = min(sanitized.customViewTimings!.count, remaining)
+            remaining -= expectedSanitizedCustomTimings
+            let expectedSanitizedUserInfo = min(sanitized.userInfoAttributes.count, remaining)
+            remaining -= expectedSanitizedUserInfo
+            let expectedSanitizedAttrs = min(sanitized.attributes.count, remaining)
+            remaining -= expectedSanitizedAttrs
+
+            XCTAssertGreaterThanOrEqual(remaining, 0)
+            XCTAssertEqual(sanitized.customViewTimings!.count, expectedSanitizedCustomTimings, "If number of attributes needs to be limited, `customViewTimings` are removed last")
```
acgh, line 153: Nitpick: it's not a big deal but to help following along, maybe reverse the order of the asserts? Like so:
- If number of attributes needs to be limited, `attributes` are removed first.
- If number of attributes needs to be limited, `userInfoAttributes` are removed second
- If number of attributes needs to be limited, `customViewTimings` are removed last

acgh (approved):
Good catch 👍

ncreated (approved):
Nice 👌, it's much cleaner (and correct!) now 💪

Tests/DatadogTests/Datadog/RUM/RUMEvent/RUMEventSanitizerTests.swift
buranmert: i followed the same order above while computing expected values, this is more readable for me personally.
i'm merging as it is now, okay?

acgh: That sounds good 👍

PR #395, opened 2021-01-29
RUMM-892 Provide more details when upload conditions are not met in console logs

acgh:
### What and why?

This is a follow-up to RUMS-132 where some troubleshooting based on the console logs was a bit hindered by the lack of details at to why upload conditions were not met.

In this PR, we expose the underlying blockers to the upload conditions. It can be the battery state/level, low power mode, the network reachability.
The following details may appear in the console:

`🐶 → 19:00:10.949 [INFO] 💡 (logging) No upload. Batch to upload: NOT CHECKED, System conditions: ❌ → 🔋 Battery: unplugged 10%; 🔌 Low Power Mode: on; 📡 Reachability: no`

In the above example all three blockers appear, but it can be any combination.

### How?

The internal `DataUploadConditions.canPerformUpload` API is changed to return an enum `Report` that is either `.go` or `.noGo`. When `Report` is `.noGo` it will come with a list of `Blocker`s as an associated value. That list is then checked and printed by `DataUploadWorker`.

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/Upload/DataUploadConditions.swift
+++ b/Sources/Datadog/Core/Upload/DataUploadConditions.swift
@@ -23,27 +23,28 @@ internal struct DataUploadConditions {
         }
 
         if let batteryStatus = batteryStatus {
-            return shouldUploadFor(networkConnectionInfo: networkConnectionInfo) && shouldUploadFor(batteryStatus: batteryStatus)
+            return canUploadWith(networkConnectionInfo: networkConnectionInfo) && canUploadWith(batteryStatus: batteryStatus)
         } else {
-            return shouldUploadFor(networkConnectionInfo: networkConnectionInfo)
+            return canUploadWith(networkConnectionInfo: networkConnectionInfo)
         }
     }
 
-    private func shouldUploadFor(batteryStatus: BatteryStatus) -> Bool {
-        if batteryStatus.state == .unknown {
+    private func canUploadWith(batteryStatus: BatteryStatus) -> Bool {
+        let state = batteryStatus.state
+        if state == .unknown {
             // Note: in RUMS-132 we got the report on `.unknown` battery state reporing `-1` battery level on iPad device
             // plugged to Mac through lightning cable. As `.unkown` may lead to other unreliable values,
             // it seems safer to arbitrary allow uploads in such case.
             return true
         }
 
-        let batteryFullOrCharging = batteryStatus.state == .full || batteryStatus.state == .charging
+        let batteryFullOrCharging = state == .full || state == .charging
         let batteryLevelIsEnough = batteryStatus.level > Constants.minBatteryLevel
         let isLowPowerModeEnabled = batteryStatus.isLowPowerModeEnabled
         return (batteryLevelIsEnough || batteryFullOrCharging) && !isLowPowerModeEnabled
     }
 
-    private func shouldUploadFor(networkConnectionInfo: NetworkConnectionInfo) -> Bool {
+    private func canUploadWith(networkConnectionInfo: NetworkConnectionInfo) -> Bool {
         return networkConnectionInfo.reachability == .yes || networkConnectionInfo.reachability == .maybe
     }
 }
--- a/Sources/Datadog/Core/Upload/DataUploadConditions.swift
+++ b/Sources/Datadog/Core/Upload/DataUploadConditions.swift
@@ -8,6 +8,30 @@ import Foundation
 
 /// Tells if data upload can be performed based on given system conditions.
 internal struct DataUploadConditions {
+    enum Blocker {
+        case batteryLevel
+        case lowPowerModeOn
+        case networkCondition
+    }
+
+    enum Report {
+        case go
+        case noGo(blockers: Set<Blocker>)
+
+        static func && (lhs: Report, rhs: Report) -> Report {
+            switch (lhs, rhs) {
+            case (.go, .go):
+                return .go
+            case (.go, .noGo):
+                return rhs
+            case (.noGo, .go):
+                return lhs
+            case (let .noGo(blockers: lbs), let .noGo(blockers: rbs)):
+                return .noGo(blockers: lbs.union(rbs))
+            }
+        }
+    }
+
     struct Constants {
         /// Battery level above which data upload can be performed.
         static let minBatteryLevel: Float = 0.1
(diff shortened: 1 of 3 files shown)


Sources/Datadog/Core/Upload/DataUploadConditions.swift
```diff
@@ -16,34 +40,52 @@ internal struct DataUploadConditions {
     let batteryStatus: BatteryStatusProviderType?
     let networkConnectionInfo: NetworkConnectionInfoProviderType
 
-    func canPerformUpload() -> Bool {
+    func canPerformUpload() -> Report {
```
ncreated, line 43: nit: the `Report` doesn't sound well to me. I'd expect kind of `Result` or maybe even:
```swift
func currentConditions() -> Conditions

enum Conditions {
   case sufficient
   case insufficient(blockers: [Blocker])
}
```
But as it's a naming thing - feel free to keep it your way 👍.

Sources/Datadog/Core/Upload/DataUploadWorker.swift
```diff
@@ -90,3 +97,27 @@ internal class DataUploadWorker: DataUploadWorkerType {
...
+    }
+}
+
+extension DataUploadConditions.Report: CustomStringConvertible {
+    var description: String {
+        switch self {
+        case .go:
+            return "✅"
+        case let .noGo(blockers: blockers):
+            return "❌ → " + blockers.map { $0.description }.joined(separator: "; ")
```
ncreated, line 120: I think this may help more with self-diagnostic if this message is more explanatory giving "because" and "AND" context i.e.:

`🐶 → 19:00:10.949 [INFO] 💡 (logging) No upload. Batch to upload: NOT CHECKED, System conditions: ❌ [upload was skipped because: 🔋 Battery is unplugged (10%) AND 🔌 Low Power Mode is enabled AND 📡 Network reachability is: no]`

WDYT?

ncreated (approved):
Nice 👌. I left two minor suggestions.

Sources/Datadog/Core/Upload/DataUploadConditions.swift
```diff
@@ -8,6 +8,30 @@ import Foundation
...
+            case (.go, .go):
+                return .go
+            case (.go, .noGo):
+                return rhs
+            case (.noGo, .go):
+                return lhs
+            case (let .noGo(blockers: lbs), let .noGo(blockers: rbs)):
+                return .noGo(blockers: lbs + rbs)
+            }
+        }
```
buranmert, line 32: what would `.noGo(blockers: [])` mean?

i think `go` is just lack of blockers, it's not a real state and this `&&` operator is just appending blockers.
if we change to
```swift
// before
func canUploadWith(...) -> Report { ... }
// after
func blockersForUpload(with: ...) -> [Blocker] { ... }
```
then `blockers.isEmpty` means `.go` state, we can remove `&&` operator, we don't need `switch case .noGo(...)` to unwrap blockers

wdyt?

buranmert (commented):
i'm not requesting changes but i think we can simplify it a bit further

Sources/Datadog/Core/Upload/DataUploadConditions.swift
acgh: I have considered naming this `Conditions` as well but couldn't find a way to have repeating and maybe confusing pattern like `DataUploadConditions` returns some other kind of `Conditions`. That being said, your proposal still sounds clearer when used at call-sites.

Sources/Datadog/Core/Upload/DataUploadWorker.swift
acgh: This is clearer, I like it! I'll correct that.

Sources/Datadog/Core/Upload/DataUploadConditions.swift
acgh: Yeah I have considered just returning the list of blockers and letting the call-site branch on the count. In the end, I thought that it would be more explicit to provide some `.go`/`.noGo` cases, but I overlooked the potential confusion if the `blockers` array was empty (although it's a case that should not happen, it would be a bug in `DataUploadConditions`).
I think your suggestion aligns with @ncreated's naming idea, that should work and make things simpler indeed. I'm going to correct that!

Sources/Datadog/Core/Upload/DataUploadWorker.swift
```diff
@@ -90,3 +91,22 @@ internal class DataUploadWorker: DataUploadWorkerType {
...
+        case let .networkReachability(description: description):
+            return "📡 Network reachability is: " + description
+        }
+    }
+}
+
+extension Array where Element == DataUploadConditions.Blocker {
+    var description: String {
+        "❌ [upload was skipped because: " + self.map { $0.description }.joined(separator: " AND ") + "]"
+    }
```
buranmert, line 111: i think it was 💯 with
```swift
var description: String {
  if self.isEmpty { return "✅" }
  else { return "❌ [upload..." }
}
```
i was wondering why you removed ✅ case?

acgh: I did not remove the `isEmpty` case, it wasn't there, but that's a great idea :)

Sources/Datadog/Core/Upload/DataUploadWorker.swift
```diff
@@ -102,22 +94,21 @@ extension DataUploadConditions.Blocker: CustomStringConvertible {
...
+            return "🔌 Low Power Mode is: enabled"
         case let .networkReachability(description: description):
-            return "📡 Reachability: " + description
+            return "📡 Network reachability is: " + description
         }
     }
 }
 
-extension DataUploadConditions.Report: CustomStringConvertible {
+extension Array where Element == DataUploadConditions.Blocker {
```
ncreated, line 106: nit: those two `extensions` can be `private` (or `fileprivate`) to avoid indexing them in the global namespace. 

On a longer term, keeping narrow namespaces should give faster compilation time if the _whole module optimisation_ is enabled (impossible to spot on individual usage).

ncreated (approved):
Looks good 💪

Sources/Datadog/Core/Upload/DataUploadWorker.swift
acgh: Thanks for the reminder, I overlooked that aspect! I'll correct that.

buranmert (approved):

PR #397, opened 2021-02-02
RUMM-956 Set up Crash Reporting feature

ncreated:
### What and why?

📦 This PR sets up the Crash Reporting feature so that crash reports captured by `PLCrashReporter` in `DatadogCrashReporting` SDK are passed to the `Datadog` SDK.

Later (`RUMM-1050` + `RUMM-960`) those crash reports will be send as `Log` and / or `RUM Error`.

### How?

The communication between both modules is established with dependency inversion interface:
```swift
// Datadog SDK:

public protocol DDCrashReportingPluginType: class {
    func readPendingCrashReport(completion: (DDCrashReport?) -> Bool)
}

// DatadogCrashReporting SDK:

import Datadog 

public class DDCrashReportingPlugin: DDCrashReportingPluginType {
   // ...
}
```

To enable Crash Reporting feature, new public API was introduced to `Datadog.Configuration` (both for Swift and Objc):
```swift
public func enableCrashReporting(using crashReportingPlugin: DDCrashReportingPluginType) -> Builder
```

In contrast to what was proposed in _RFC: Releasing iOS Crash Reporting_ I decided to expose this ☝️ API on configuration builder, instead of RFC-way:
```swift
// Abandoned API:

Datadog.initialize(
   configuration: configuration,
   appContext: appContext,
   installedPlugins: [DDCrashReportingPlugin()]
)
````
First, `.enableCrashReporting(using:)` is much more consistent with existing `.enableLogging(_:)`, `.enableTracing(_:)` and `.enableRUM(_:)`. Second, it's way healthier for the architecture to process concrete plugin type than ambiguous array of plugins (in `dd-sdk-android` this lead [to artificial plugin-feature assignment](https://github.com/DataDog/dd-sdk-android/blob/2d92c610bf26d245f0490531a660859a07276200/dd-sdk-android/src/main/kotlin/com/datadog/android/DatadogConfig.kt#L424-L434)). 

The `CrashReportingFeature` itself is no different than other features in the SDK. It's built around a singleton pattern, abstracting its integration with `LoggingFeature` and `RUMFeature` through two shims:
* `CrashReportingWithLoggingIntegration`,
* `CrashReportingWithRUMIntegration`.

Both expose single `send(crashReport:)` API.

Updated SDK architecture diagram:

Screenshot 2021-02-03 at 11 03 07

### Manual testing

Crash Reporting can be tested manually by crashing the `Example` app with disconnected debugger and then launching it again with debugger turned-on. Example crash report's stack-trace:
```
... 

Exception Type:  SIGILL
Exception Codes: ILL_ILLOPC at 0x10f007004
Crashed Thread:  0

Thread 0 Crashed:
0   libswiftCore.dylib                  0x000000010f007004 $ss17_assertionFailure__4file4line5flagss5NeverOs12StaticStringV_SSAHSus6UInt32VtF + 532
1   Example                             0x000000010e872796 $s7Example26DebugLoggingViewControllerC16didTapSend10LogsyyypF + 182
2   Example                             0x000000010e8727f0 $s7Example26DebugLoggingViewControllerC16didTapSend10LogsyyypFTo + 80
3   UIKitCore                           0x00007fff246972b2 -[UIApplication sendAction:to:from:forEvent:] + 83
4   UIKitCore                           0x00007fff23fc08a0 -[UIControl sendAction:to:forEvent:] + 223
5   UIKitCore                           0x00007fff23fc0bc3 -[UIControl _sendActionsForEvents:withEvent:] + 332
6   UIKitCore                           0x00007fff23fbf4ac -[UIControl touchesEnded:withEvent:] + 500
7   UIKitCore                           0x00007fff246d3962 -[UIWindow _sendTouchesForEvent:] + 1287

... 
```

### Review checklist

### CODE REVIEW

File: Datadog/Example/AppConfiguration.swift [hunk 1]
```diff
@@ -39,6 +40,7 @@ struct ExampleAppConfiguration: AppConfiguration {
             .set(serviceName: serviceName)
             .set(batchSize: .small)
             .set(uploadFrequency: .frequent)
+            .enableCrashReporting(using: DDCrashReportingPlugin())
```
acgh, line 43: I like that Dependency Inversion pattern 👍

Sources/Datadog/Core/FeaturesConfiguration.swift
```diff
@@ -67,6 +72,8 @@ internal struct FeaturesConfiguration {
     let rum: RUM?
     /// `URLSession` auto instrumentation configuration, `nil` if not enabled.
     let urlSessionAutoInstrumentation: URLSessionAutoInstrumentation?
+    /// Crsah Reporting feature configuration or `nil` if the feature was not enabled.
```
acgh, line 75: Typo 👀 
```suggestion
    /// Crash Reporting feature configuration or `nil` if the feature was not enabled.
```

Sources/Datadog/CrashReporting/CrashReportingFeature.swift
```diff
@@ -0,0 +1,60 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+/// Creates and owns componetns enabling Crash Reporting feature.
```
acgh, line 9:
```suggestion
/// Creates and owns components enabling Crash Reporting feature.
```

Sources/Datadog/CrashReporting/CrashReportingFeature.swift
```diff
@@ -0,0 +1,60 @@
...
+        configuration: FeaturesConfiguration.CrashReporting,
+        loggingIntegration: CrashReportingIntegration?,
+        rumIntegration: CrashReportingIntegration?
+    ) {
+        self.plugin = configuration.crashReportingPlugin
+        self.loggingIntegration = loggingIntegration
+        self.rumIntegration = rumIntegration
+    }
+
+    func sendCrashReportIfFound() {
```
acgh, line 34: That looks like an extra responsibility for this `Feature` class. Maybe extract that responsibility out to a dedicated class that is owned by `CrashReportingFeature`.

Sources/Datadog/CrashReporting/DDCrashReportingPluginType.swift
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+import Foundation
+
+/// Crash Report format supported by Datadog SDK.
+@objc
+public class DDCrashReport: NSObject {
+    /// The date of the crash occurrence.
+    internal let crashDate: Date?
+    // TODO: RUMM-1053 - consider providing / formatting this characteristic of the crash report
```
acgh, line 14: 💪

Sources/DatadogCrashReporting/DDCrashReportingPlugin.swift
```diff
@@ -4,32 +4,61 @@
...
+            _ = completion(nil)
+            consolePrint("🔥 DatadogCrashReporting error: failed to load crash report: \(error)")
         }
     }
 }
+
+// MARK: - Utils
+
+/// Function printing `String` content to console.
+internal var consolePrint: (String) -> Void = { content in
```
acgh, line 62: What's the added value compared to a straight call to `print()`?

Sources/Datadog/CrashReporting/DDCrashReportingPluginType.swift
```diff
@@ -0,0 +1,47 @@
...
+
+/// An interface for enabling crash reporting feature in Datadog SDK.
+/// It is implemented by `DDCrashReportingPlugin` from `DatadogCrashReporting` framework.
+@objc
+public protocol DDCrashReportingPluginType: class {
+    /// Reads unprocessed crash report if available.
+    /// - Parameter completion: the completion block called with the value of `DDCrashReport` if a crash report is available
+    /// or with `nil` otherwise. The value returned by the receiver should indicate if the crash report was processed correctly (`true`)
+    /// or something went wrong (`false)`. Depending on the returned value, the crash report will be purged or perserved for future read.
+    func readPendingCrashReport(completion: (DDCrashReport?) -> Bool)
```
acgh, line 46: Are there any expectations (for callers and implementations) on whether it should be synchronous or asynchronous?

Sources/Datadog/Datadog.swift
```diff
@@ -246,6 +256,9 @@ public class Datadog {
             userInfoProvider: userInfoProvider,
             launchTimeProvider: launchTimeProvider
         )
+
+        // After everything is set up, send crash report if available
+        CrashReportingFeature.instance?.sendCrashReportIfFound()
```
acgh, line 261: Assuming the implementation is synchronous, could this become a bottleneck? And even more so if it's called when the App delegate is called back on the main thread.

acgh (commented):
This looks great, I love the architecture around that feature! Besides, this is well introduced in the PR's description 💪 
I left a few comments/questions for clarifications, I'll do another pass with those in mind.

Sources/Datadog/CrashReporting/CrashReportingFeature.swift
ncreated: Makes sense 👌, especially that with upcoming `RUMM-957` and `RUMM-1049` the crash reporting feature will learn another range of functionality: injecting SDK context into crash reports (last `RUMViewEvent`, last `UserInfo`, last `NetworkInfo` etc.).

I think it may be good to create `CrashReporter` class and put `sendCrashReportIfFound()` + upcoming ~`inject(context:)` in there. This way, the `Datadog.CrashReporter` will be abstracting the `CrashReporter` injected from `DatadogCrashReporting`. Seems good for testing, readability (single abstraction for 3rd party reporter) and will be unified with other features. WDYT?

Sources/DatadogCrashReporting/DDCrashReportingPlugin.swift
ncreated: We use this pattern in `Datadog` SDK to mock the `consolePrint` in tests and add assertions on user-facing messages.

acgh: I see:
https://github.com/DataDog/dd-sdk-ios/blob/119329d1209e97d5fbfc1c0be446f1d713f9a522/Tests/DatadogTests/Datadog/Mocks/CoreMocks.swift#L744-L757
This is clever 👍

Sources/Datadog/CrashReporting/CrashReportingFeature.swift
acgh: That sounds like a great plan!

Sources/Datadog/CrashReporting/DDCrashReportingPluginType.swift
ncreated: The note on sync/async + threading assumptions should be definitely there in this public API comment 👍.

Right now it's synchronous, but the `completion` is there because it will be async at the end. The next PR will deliver SDK context injection to crash reports, and that operation will need to be synchronised in `DDCrashReportingPlugin's` implementation (as it might be called from different threads). The same thread will be then used for dispatching `completion: (DDCrashReport?) -> Bool`asynchronously.

I'd prefer to update this comment once a real threading is introduced in next PR. Right now I can only guess how it may look like 🙂. WDYT?

Sources/Datadog/Datadog.swift
ncreated: Yes, if kept in the current form it's definitely a bottleneck as crash report deserialisation (in `PLCrashReporter`) may take a while. Both `sendLog()` and `sendRUMEvent()` will be fast in our code, as they are dispatched to background thread by the `Writer`. However, as explained in [the previous comment](https://github.com/DataDog/dd-sdk-ios/pull/397/files/9901f739f1207f3a44c1d0fc3c258c9485802f11#r569217837) the `readPendingCrashReport(completion:)` will be executed off the main thread in next PR, so the impact will be reduced to just  `queue.async {}`.

Sources/Datadog/CrashReporting/CrashReportingFeature.swift
ncreated: Done ✅ ([new commit](https://github.com/DataDog/dd-sdk-ios/pull/397/commits/71ed63cf793078ae8c44d054e8fe579224de5d48)).

I introduced the `CrashReporter` class which owns the `sendCrashReportIfFound()`. It will later expose an interface for injecting SDK context into crash reports.

It should be understood as an equivalent of `Logger` for `LoggingFeature`, `Tracer` for `TracingFeature` and `RUMMonitor` for `RUMFeature`. The only difference is that it remains `internal`, as we don't need to expose it for the user. I pinned it to `Global.crashReporter` as it has similar semantic as global `tracer` and `rum`.

The responsibility of `CrashReportingFeature` is now cleaned up. In next PRs it will be armed with necessary components for sending crash reports to Datadog.

// I also updated the diagram in the description of this PR.

acgh: Wow this is great!

Sources/Datadog/CrashReporting/DDCrashReportingPluginType.swift
acgh: That sounds clear and sensible 👍

Tests/DatadogCrashReportingTests/DDCrashReportingPluginTests.swift
```diff
@@ -0,0 +1,140 @@
...
+
+        // When
+        plugin.readPendingCrashReport { crashReport in
+            XCTAssertEqual(crashReport, crashReporter.pendingCrashReport)
+            expectation.fulfill()
+            return true // the caller succeeded in processing the crash report
+        }
+
+        // Then
+        waitForExpectations(timeout: 0.5, handler: nil)
```
acgh, line 40: I see that this should automatically work once `readPendingCrashReport()` becomes async 👍

Tests/DatadogTests/Datadog/Core/FeaturesConfigurationTests.swift
```diff
@@ -324,6 +324,59 @@ class FeaturesConfigurationTests: XCTestCase {
...
+
+    func testWhenCrashReportingIsDisabled() throws {
+        XCTAssertNil(
+            try FeaturesConfiguration(configuration: .mockWith(crashReportingPlugin: nil), appContext: .mockAny()).crashReporting,
+            "Feature configuration should not be available if the feature is disabled"
+        )
+    }
+
+    func testWhenRUMorLoggingFeaturesAreEnabled_thenCrashReportingFeatureCanBeEnabled() throws {
+        let random = [true, true, false].shuffled() // get at least one `true` on first two positions
```
acgh, line 337: This is a more of a general point (maybe addressable in RUMM-896). I'm wondering what's the impact of randomizing test inputs and if it could lead to flakiness by randomly uncovering a regression.

Sources/Datadog/CrashReporting/DDCrashReportingPluginType.swift
acgh: A question that's again probably out of scope and not a big deal now, will there a be some kind of queueing or debouncing if that method is called multiple times before the first call completes?
Especially if the implementation behind it tracks some state beginning with the first call.

acgh (approved):
Brilliant 😎

Tests/DatadogTests/Datadog/Core/FeaturesConfigurationTests.swift
ncreated: Definitely something we can discuss in `RUMM-896`! I find this sort of [property testing](https://en.wikipedia.org/wiki/Property_testing) as a practical approach to covering multiple inputs leading to a single result. By definition, this should not lead to flakiness as this randomization should be build very consiously.

PR #399, opened 2021-02-04
RUMM-1000 Add support for additional properties to the models generation pipeline

acgh:
### What and why?

We ultimately want to add support for custom timings. This was previously introduced to the schema in [rum-events-format #25](https://github.com/DataDog/rum-events-format/pull/25) relying on [JSON Schema's `additionalProperty`](https://json-schema.org/draft/2019-09/json-schema-core.html#additionalProperties).

Our Swift/Objc pipeline that generates models from a schema doesn't know how to handle `additionalProperties` at this point. To prevent from mis-generating the model for custom timings [we previously had to put a dedicated filter](https://github.com/DataDog/dd-sdk-ios/pull/393/files#diff-dd3caa4e6db960eecd95024f6d7df904948b8f53b1026c00dd19a9627317ac3e) until we add the support for `additionalProperties`.

In this PR we add the plumbing needed for a basic support of `additionalProperty`. It will allow us to remove the filter and generate the model for custom timings.

### How?

As of 8e6fbbf the approach has been simplified. To sum-up, we model `additionalProperties` as dictionaries while adding a relatively restrictive support of `SwiftDictionary` (only primitive types).
Any future gaps between an improved use of JSON Schema in the input files and the current limitations of the pipeline should be caught by exceptions at runtime, and prior to shipping the generated code inside the SDK.

#### JSON

We add an `additionalProperties` property to the `JSONSchema`. It is optional and of type `JSONSchema` itself [as required by the spec](https://json-schema.org/draft/2019-09/json-schema-core.html#additionalProperties).

We map that schema to a `JSONObject.AdditionalProperties` struct which is a subset of a `JSONObject.Property`. It holds a `type` that is a `JSONPrimitive`, meaning that we only support `bool, double, integer, string` at this point. The mapping happens in the `JSONSchemaToJSONTypeTransformer`.

#### Swift

To support additional properties in `SwiftStruct`, we introduce a basic `SwiftDictionary`. Note that keys are `String` and values must be `SwiftPrimitiveType`, i.e. `Bool, Int, Int64, String, Double`.
From that point we can declare a `SwiftStruct.Property` of type `SwiftDictionary` to model JSONSchema's `additionalProperties`. The mapping from `JSONObject` to `SwiftStruct` is done in `JSONToSwiftTypeTransformer`.

`JSONToSwiftTypeTransformer` enforces a couple of new rules:
- A provided root object cannot map to a Dictionary, especially a root object cannot have `additionalProperties` that would require a mapping to a dictionary.
- A JSON object cannot hold both regular properties and `additionalProperties`.

#### Objc Interop

To mirror additions to the Swift type system and transformer, we introduce a `ObjcInteropNSDictionary` and add the expected mapping methods inside `SwiftToObjcInteropTypeTransformer`. 

#### RUM

We just add some basic expect methods to allow traversing a `SwiftDictionary` and transforming any primitive types, e.g. `Int` -> `Int64`.

#### Printers

This has been simplified a lot compared to the initial approach (41a1ac8d and before), they only need to know how to print dictionaries while converting some core Swift <-> Objc types.

### Review checklist

### Next

- [ ] Run `tools/rum-models-generator/run.sh generate` to add custom timings
Spoiler this should generate something to resolve this:
```
... RUMDataModels.swift is out of sync with rum-events-format: 8b955a03d0fe0b2f032a02d6800c61ef3fc9fada
Difference was found when comparing it with template file:
>>>
111,113d110
<         /// User custom timings of the view
<         public let customTimings: [String: Int64]?
< 
172d168
<             case customTimings = "custom_timings"
<<<
```

### DIFF
--- a/tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/JSON/JSONType.swift
+++ b/tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/JSON/JSONType.swift
@@ -41,12 +41,18 @@ internal struct JSONObject: JSONType {
         let isReadOnly: Bool
     }
 
+    struct AdditionalProperties: JSONType {
+        let comment: String?
+        let type: JSONPrimitive
+        let isReadOnly: Bool
+    }
+
     let name: String
     let comment: String?
     let properties: [Property]
-    let additionalProperties: Property?
+    let additionalProperties: AdditionalProperties?
 
-    init(name: String, comment: String?, properties: [Property], additionalProperties: Property? = nil) {
+    init(name: String, comment: String?, properties: [Property], additionalProperties: AdditionalProperties? = nil) {
         self.name = name
         self.comment = comment
         self.properties = properties.sorted { property1, property2 in property1.name < property2.name }
--- a/tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/JSON/JSONType.swift
+++ b/tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/JSON/JSONType.swift
@@ -36,7 +36,7 @@ internal struct JSONObject: JSONType {
         let name: String
         let comment: String?
         let type: JSONType
-        let defaultVaule: DefaultValue?
+        let defaultValue: DefaultValue?
         let isRequired: Bool
         let isReadOnly: Bool
     }
--- a/tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/JSON/JSONType.swift
+++ b/tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/JSON/JSONType.swift
@@ -44,11 +44,13 @@ internal struct JSONObject: JSONType {
     let name: String
     let comment: String?
     let properties: [Property]
+    let additionalProperties: Property?
 
-    init(name: String, comment: String?, properties: [Property]) {
+    init(name: String, comment: String?, properties: [Property], additionalProperties: Property? = nil) {
         self.name = name
         self.comment = comment
         self.properties = properties.sorted { property1, property2 in property1.name < property2.name }
+        self.additionalProperties = additionalProperties
     }
 }
 
--- a/tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/ObjcInterop/SwiftToObjcInteropTypeTransformer.swift
+++ b/tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/ObjcInterop/SwiftToObjcInteropTypeTransformer.swift
@@ -66,8 +66,10 @@ internal class SwiftToObjcInteropTypeTransformer {
         case let swiftArray as SwiftArray:
             return ObjcInteropNSArray(element: try objcInteropType(for: swiftArray.element))
         case let swiftDictionary as SwiftDictionary:
-            return ObjcInteropNSDictionary(key: try objcInteropType(for: swiftDictionary.key),
-                                           value: try objcInteropType(for: swiftDictionary.value))
+            return ObjcInteropNSDictionary(
+                key: try objcInteropType(for: swiftDictionary.key),
+                value: try objcInteropType(for: swiftDictionary.value)
+            )
         default:
             throw Exception.unimplemented(
                 "Cannot create `ObjcInteropType` type for \(type(of: swiftType))."
--- a/tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/ObjcInterop/SwiftToObjcInteropTypeTransformer.swift
+++ b/tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/ObjcInterop/SwiftToObjcInteropTypeTransformer.swift
@@ -19,7 +19,7 @@ internal class SwiftToObjcInteropTypeTransformer {
 
         try takeRootSwiftStructs(from: swiftTypes)
             .forEach { rootStruct in
-                let rootClass = ObjcInteropRootClass(managedSwiftStruct: rootStruct)
+                let rootClass = ObjcInteropRootClass(bridgedSwiftStruct: rootStruct)
                 outputObjcInteropTypes.append(rootClass)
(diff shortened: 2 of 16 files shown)


tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/ObjcInterop/SwiftToObjcInteropTypeTransformer.swift
```diff
@@ -135,13 +65,105 @@ internal class SwiftToObjcInteropTypeTransformer {
...
+            )
+            return propertyWrapper
+        case let swiftArray as SwiftArray where swiftArray.element is SwiftPrimitiveType:
+            let propertyWrapper = ObjcInteropPropertyWrapperManagingSwiftStructProperty(
+                owner: objcClass,
+                swiftProperty: swiftProperty
+            )
+            propertyWrapper.objcInteropType = try objcInteropType(for: swiftArray)
+            return propertyWrapper
+        case let swiftDictionary as SwiftDictionary where swiftDictionary.value is SwiftPrimitiveType:
```
acgh, line 126: The whole function comes from the block of code previously starting at line 35. But I want to draw attention to this case that handles a `SwiftDictionary`.

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Output/Printing/SwiftPrinter.swift
```diff
@@ -89,11 +129,59 @@ public class SwiftPrinter: BasePrinter {
         writeLine("}")
     }
 
+    private func printCodableMethods(for properties: [SwiftStruct.Property], additionalProperties: SwiftStruct.Property) throws {
```
acgh, line 132: Please see the associated tests for a preview of the output:
https://github.com/DataDog/dd-sdk-ios/pull/399/files#diff-fcc08ef045de4c9d692921649cfe668e0900e3900473a1e5702383cc67c37efa

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/JSON/JSONType.swift
```diff
@@ -36,19 +36,21 @@ internal struct JSONObject: JSONType {
...
-        let defaultVaule: DefaultValue?
+        let defaultValue: DefaultValue?
         let isRequired: Bool
         let isReadOnly: Bool
     }
 
     let name: String
     let comment: String?
     let properties: [Property]
+    let additionalProperties: Property?
```
ncreated, line 47: I think that we're using the `Property` type here with a wrong purpose. Looking at how "additionalProperties" work, only `Property.type`, `Property.isReadOnly` and `Property.comment` apply to it and we force artificial / uncontrolled values for `name`, `defaultValue` and `isRequired`. Maybe it's worth introducing separate type: `JSONObject.AdditionalProperties(comment:type:isReadOnly)`, WDYT?

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/ObjcInterop/SwiftToObjcInteropTypeTransformer.swift
```diff
@@ -78,6 +78,17 @@ internal class SwiftToObjcInteropTypeTransformer {
...
+                    let propertyWrapper = ObjcInteropPropertyWrapperManagingSwiftStructProperty(
+                        owner: objcClass,
+                        swiftProperty: swiftProperty
+                    )
+                    propertyWrapper.objcInteropType = try objcInteropType(for: swiftDictionary)
+                    return propertyWrapper
+                case let swiftDictionary as SwiftDictionary where swiftDictionary.key is SwiftEnum || swiftDictionary.value is SwiftEnum:
+                    throw Exception.unimplemented(
+                        "Objc Interop for `SwiftDictionary`s with an Enum as key or value is not supported."
+                    )
```
ncreated, line 91: Maybe it's easier to change `SwiftDictionary.value` to be `SwiftPrimitiveType` to have compile-time check for this?

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/Swift/SwiftType.swift
```diff
@@ -28,6 +28,11 @@ internal struct SwiftArray: SwiftType {
     var element: SwiftType
 }
 
+internal struct SwiftDictionary: SwiftType {
+    var key: SwiftType
+    var value: SwiftType
+}
```
ncreated, line 34: IMO we can save some control flows / tests if being more tailored on that, i.e.:
```swift
    var key: SwiftPrimitive<String>
    var value: SwiftPrimitiveType
```

ncreated (commented):
Looks way simpler and much more straightforward now 🚀. I left some comments that we may consider to simplify it even further - let me know WDYT.

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/JSON/JSONType.swift
acgh: I think that it's a good point and good idea!

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/ObjcInterop/SwiftToObjcInteropTypeTransformer.swift
acgh: That's a good idea, and similarly `SwiftDictionary.key` should be a `SwiftPrimitiveType` for now.

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/Swift/SwiftType.swift
acgh: Yes you're right!

ncreated (approved):
LVG 🎸

PR #400, opened 2021-02-04
RUMM-957 Add `Example` app scenario for testing Crash Reporting feature

ncreated:
### What and why?

📦 This PR adds single test scenario in `Example` app to play with the Crash Reporting feature. Later (in `RUMM-960`) it will be used to send the crash report to Python server and run assertions in integration tests.

### How?

The `CrashReportingCollectOrSendScenario` scenario consists of a single View Controller with conditionally displayed UI:

Screenshot 2021-02-04 at 15 25 44

If the scenario is started with no crash report file on disc, it displays the _"Crash The App"_ button. Once pressed, it issues a `fatalError()` and terminates the process.

When a crash report file is found on startup, it displays the _"Sending crash report..."_ label, meaning that it found the crash report and it expects the SDK to send it.

Once the actual `RUMError` upload is implemented in `RUMM-960`, it will be sending real data to Datadog.

### Manual Testing

To test this scenario manually, enable `CrashReportingCollectOrSendScenario` in Example schema's `ENV`, then:
→ disconnect debugger → run the Example app so it presents "Crash The App" button → crash  → connect debugger → run again → notice the breakpoint hit in `CrashReportingWithRUMIntegration`.

### Review checklist

### DIFF
--- a/Datadog/Example/Scenarios/CrashReporting/CrashReporting/CrashReportingViewController.swift
+++ b/Datadog/Example/Scenarios/CrashReporting/CrashReporting/CrashReportingViewController.swift
@@ -1,0 +1,30 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import UIKit
+
+internal class CrashReportingViewController: UIViewController {
+    @IBOutlet weak var sendingCrashReportLabel: UILabel!
+    @IBOutlet weak var crashTheAppButton: UIButton!
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        let testScenario = (appConfiguration.testScenario as! CrashReportingCollectOrSendScenario)
+
+        if testScenario.hadPendingCrashReportDataOnStartup {
+            sendingCrashReportLabel.isHidden = false
+            crashTheAppButton.isHidden = true
+        } else {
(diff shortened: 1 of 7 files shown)


Datadog/Example/Scenarios/CrashReporting/CrashReporting/CrashReportingViewController.swift
```diff
@@ -0,0 +1,30 @@
...
+            sendingCrashReportLabel.isHidden = false
+            crashTheAppButton.isHidden = true
+        } else {
+            sendingCrashReportLabel.isHidden = true
+            crashTheAppButton.isHidden = false
+        }
+    }
+
+    @IBAction func didTapCrashTheApp(_ sender: Any) {
+        fatalError("The 'Crash The App' button was tapped.")
```
buranmert, line 28: does it crash the process in the same way that a real crash (eg: crashing with `array[array.count + 10]`) does?

buranmert (approved):
lgtm, i just left a question out of curiosity

Datadog/Example/Scenarios/CrashReporting/CrashReporting/CrashReportingViewController.swift
ncreated: Yes, it receives `SIGILL` in both cases 👌.
```swift
crashReport.signalCode = "ILL_ILLOPC"
crashReport.signalName = "SIGILL"
crashReport.signalDetails = "Illegal instruction"
```
I haven't yet jumped into details of each signal code to determine the correctness of this → instead `RUMM-1053` was created to investigate deeper.

PR #401, opened 2021-02-04
RUMM-957 Inject application context to crash reports

ncreated:
### What and why?

🚚 This PR implements injection of a last `RUMViewEvent` and current `TrackingConsent` values to crash reporter, so they can be retrieved when the app restarts after crash.

Those values are necessary for:
* considering if the crash report can be send regarding the user tracking policy,
* sending the crash report as a `RUMErrorEvent` associated with the RUM session identifier from the crashed session.

### How?

`CrashContext` structure was introduced for bundling SDK state values. For now it has only two properties:
```swift
struct CrashContext: Codable {
    /// Last value of `TrackingConsent`.
    var lastTrackingConsent: TrackingConsent
    /// Last RUM View event.
    var lastRUMViewEvent: RUMViewEvent?
}
```
More will be added in `RUMM-1049` (including User Info, Network Info and Carrier Info).

The `CrashContext` value is constructed in `Datadog` SDK and passed to the `DatadogCrashReporting` through `DDCrashReportingPluginType` (plugin) interface. Raw `Data` is used as the transfer object to limit the dependency on SDK.

`CrashContext` value is managed by the `CrashContextProvider` object. It's role is to receive partial updates to the context from different SDK components and notify the plugin with the new context on each change:
* updates for the `lastTrackingConsent` come from the subscription to existing `ConsentProvider`;
* updates for the `lastRUMViewEvent` come from the `RUMViewScope` (through `RUMWithCrashContextIntegration` type to not make the `RUM` module depend directly on the `CrashReporting`).

Architecture wrap-up:

Screenshot 2021-02-04 at 16 39 08

### Review checklist

### DIFF
--- a/Sources/Datadog/CrashReporting/CrashContext/CrashContext.swift
+++ b/Sources/Datadog/CrashReporting/CrashContext/CrashContext.swift
@@ -1,0 +1,47 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+/// Describes current Datadog SDK context, so the app state information can be attached to
+/// the crash report and retrieved back when the application is started again.
+///
+/// Note: as it gets saved along with the crash report during process interruption, it's good
+/// to keep this data well-packed and as small as possible.
+internal struct CrashContext: Codable {
+    /// Codable representation of the public `TrackingConsent`.
+    /// Uses `Int8` for optimized packing.
+    enum TrackingConsent: Int8, Codable {
+        case granted
+        case notGranted
+        case pending
+    }
+
+    /// Last value of `TrackingConsent`.
+    var lastTrackingConsent: TrackingConsent
+    /// Last RUM View event.
+    var lastRUMViewEvent: RUMViewEvent?
+
+    // TODO: RUMM-1049 Add Codable version of `UserInfo?`, `NetworkInfo?` and `CarrierInfo?`
+
+    enum CodingKeys: String, CodingKey {
+        case lastTrackingConsent = "ltc"
+        case lastRUMViewEvent = "lre"
+    }
+}
+
+// MARK: - Helpers
+
+extension CrashContext.TrackingConsent {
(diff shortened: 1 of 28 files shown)


Sources/Datadog/CrashReporting/CrashContext/CrashContext.swift
```diff
@@ -0,0 +1,47 @@
...
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+/// Describes current Datadog SDK context, so the app state information can be attached to
+/// the crash report and retrieved back when the application is started again.
+///
+/// Note: as it gets saved along with the crash report during process interruption, it's good
+/// to keep this data well-packed and as small as possible.
```
acgh, line 13: 💪

Tests/DatadogTests/Datadog/CrashReporting/CrashContext/CrashContextProviderTests.swift
```diff
@@ -0,0 +1,85 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import XCTest
+@testable import Datadog
+
+/// This suite tests if `CrashContextProvider` gets updated by different SDK componetns, each delivering
```
acgh, line 10:
```suggestion
/// This suite tests if `CrashContextProvider` gets updated by different SDK components, each delivering
```

Tests/DatadogTests/Datadog/RUMMonitorTests.swift
```diff
@@ -874,6 +874,36 @@ class RUMMonitorTests: XCTestCase {
         XCTAssertEqual(session.viewVisits[0].errorEvents[0].error.message, "Modified error message")
     }
 
+    // MARK: - Integration with Crash Reporting
+
+    func testGivenRegisteredCrashReporer_whenRUMViewEventIsSend_itIsUpdatedInCurrentCrashContext() throws {
```
acgh, line 879:
```suggestion
    func testGivenRegisteredCrashReporter_whenRUMViewEventIsSend_itIsUpdatedInCurrentCrashContext() throws {
```

acgh (approved):
LGTM 👍

buranmert (approved):

PR #405, opened 2021-02-08
Swift Package Manager - Add static library variants to allow use inside frameworks

bpollman:
### What and why?

Datadog cannot be added to a Framework via SPM due to the `dynamic` property being setting in `Package.swift`

When adding Datadog using SPM as a package to a framework, the `type: .dynamic` property causes Xcode to insist on embedding the Datadog inside the framework. The Embed Framework step is automatically created and removing, causes Datadog to be removed from the project. 

Note that with LaunchDarkly, the package only appears in `Link Binary with Libraries` as desired. 

Debug builds of the project will build and run fine. However, archiving and uploading to TestFlight will succeed but the archive produced is not valid and will crash on launch as the Datadog library is missing

#### Screenshot
image

#### Crash on Launch Log
```Exception Type:  EXC_CRASH (SIGABRT)
Exception Codes: 0x0000000000000000, 0x0000000000000000
Exception Note:  EXC_CORPSE_NOTIFY
Termination Description: DYLD, dyld: Using shared cache: F1161374-C3FB-358B-897A-6C2AD68BB0CC | dependent dylib '@rpath/Datadog.framework/Datadog' not found for '/private/var/containers/Bundle/Application/8F88DB89-E6FF-48D8-9F50-52270379FC0C/Deputy.app/Frameworks/DeputyKit.framework/DeputyKit', tried but didn't find: 
'/usr/lib/swift/Datadog.framework/Datadog'
'/private/var/containers/Bundle/Application/8F88DB89-E6FF-48D8-9F50-52270379FC0C/Deputy.app/Frameworks/Datadog.framework/Datadog' 
'/private/var/containers/Bundle/Application/8F88DB89-E6FF-48D8-9F50-52270379FC0C/Deputy.app/Frameworks/DeputyKit.framework/Frameworks/Datadog.framework/Datadog' 
'/Users/distiller/project/app/DerivedData/Build/Intermediates.noindex/ArchiveIntermediates/Deputy/BuildProductsPath/Release-iphoneos/Datadog.framework/Datadog'
'/Users/distiller/project/app/../Carthage/Build/iOS/Datadog.framework/Datadog'
'@rpath/Datadog.framework/Datadog' '/System/Library/Frameworks/Datadog.framework/Datadog'
Highlighted by Thread:  0
```
### How?

Removal of the `type: .dynamic` property allows Xcode to choose the right course of action for the project. 

I am not sure if removing this property will cause issues for other applications but we are currently using ~20 other dependencies via SPM and this is the only one that sets the `dynamic` properly and causes this error. 

### Review checklist

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -16,6 +16,14 @@ let package = Package(
             name: "DatadogObjc",
             type: .dynamic,
             targets: ["DatadogObjc"]),
+        .library(
+            name: "DatadogStatic",
+            type: .static,
+            targets: ["Datadog"]),
+        .library(
+            name: "DatadogStaticObjc",
+            type: .static,
+            targets: ["DatadogObjc"]),
     ],
     dependencies: [],
     targets: [


buranmert:
Hi @bpollman 👋 
Thanks for raising this issue 🙏 

When we let Swift Package Manager decide on dynamic vs static linking, it fails to choose dynamic in app extension targets.
We specify `.dynamic` in `Package.swift` in order to support app extensions (e.g: Notification Service Extension).

[We are aware of "Do not embed -> framework disappears" bug in Xcode](https://github.com/DataDog/dd-sdk-ios/issues/370#issuecomment-762800297) and we hoped it would be resolved in the next version but unfortunately it's still there in Xcode 12.4

I created a new iOS app project with 1 framework and 1 app extension target bundled to it. I used `Datadog` framework in all 3 of them by using one of the workarounds below:

### The workaround

Link `Datadog` to **all 3 targets** with `Embed & Sign` or `Embed without signing`
While **internal framework target** / **app extension target** is selected, open `Build phases -> Embed frameworks`
`Destination` should be `Frameworks` by default; change it to `Products Directory`

Screenshot 2021-02-09 at 15 21 46

1. `SomeApp` embeds `Datadog` to `Frameworks`
2. `SomeFramework` / `SomeExtension` embeds `Datadog` to `Producsts directory`

Now, at compile time the compiler should be able to find symbols in `Products Directory` and at launch/runtime the dynamic linker should be able to find `Datadog` binary in `Frameworks` folder of the host app

I hope that may help you 🤞 

### Last resort workaround

As suggested in [a discussion in Swift Forums](https://forums.swift.org/t/is-this-an-xcode-bug-or-somehow-related-to-spm/33987/3) you may remove `Embed Frameworks` for `Datadog` manually from `pbxproj`.

Please let me know if that works for you

bpollman:
Thanks for the great response! Will look into the workaround you mentioned.

Another approach that could work would be to present the library both ways, ie:

```
 products: [
        .library(
            name: "Datadog",
            type: .dynamic,
            targets: ["Datadog"]),
        .library(
            name: "DatadogObjc",
            type: .dynamic,
            targets: ["DatadogObjc"]),
        .library(
            name: "DatadogStatic",
            targets: ["Datadog"]),
        .library(
            name: "DatadogStaticObjc",
            targets: ["DatadogObjc"]),
    ],
```
Hoping that a proper solutions comes with Xcode soon 🤞

buranmert:
@bpollman that's a very good idea! can you please go ahead and make this change? then i will merge the PR ✅ 

I just want to remind you that if you are shipping your framework, your clients will not be able to use `Datadog` framework in their apps. They will have "duplication of library code" errors.

bpollman:
Thanks @buranmert! Changes have been made. 

With luck these sorts of workarounds won't be necessary by the time we need to ship a framework.

Package.swift
```diff
@@ -16,6 +16,12 @@ let package = Package(
             name: "DatadogObjc",
             type: .dynamic,
             targets: ["DatadogObjc"]),
+        .library(
+            name: "DatadogStatic",
+            targets: ["Datadog"]),
+        .library(
+            name: "DatadogStaticObjc",
+            targets: ["DatadogObjc"]),
```
buranmert, line 24: if Xcode changes its behavior one day, these may end up being linked dynamically
can we please enforce static linking by adding `type: .static` explicitly?

```suggestion
        .library(
            name: "DatadogStatic",
            type: .static,
            targets: ["Datadog"]),
        .library(
            name: "DatadogStaticObjc",
            type: .static,
            targets: ["DatadogObjc"]),
```

buranmert (changes requested):

Package.swift
bpollman: Very good point. Changes have been made.

buranmert (approved):

PR #406, opened 2021-02-09
RUMM-1000 Use View.customTimings from actual schema instead of RUMEvent

acgh:
### What and why?

[The JSON Schema for RUM Event models "custom timings" of a view](https://github.com/DataDog/rum-events-format/blob/master/schemas/view-schema.json#L108-L117) using the [`additionalProperties` keyword of the JSON Schema spec](https://json-schema.org/draft/2019-09/json-schema-core.html#additionalProperties).

This keyword was not supported until #399, and prior to that the SDK had to use a few workarounds to store the custom timings in memory until the eventual encoding into JSON.

In this PR we leverage #399 to remove the workarounds and directly encode `customTimings` from `View` models.

### How?
In this PR we:
- Remove the filter on `customTimings`423f095e
- Update the data models by running the codegen tool 8b70eabf
- Remove the few workarounds relying on RUMEvent and impact JSONDecoder d829781f

### Review checklist

- [x] Existing unit and integration tests pass
- [x] Run the Example app with the RUM Manual Instrumentation scenario and check the payload in the web UI. Like so:

Screenshot 2021-02-09 at 11 36 25

Screenshot 2021-02-09 at 11 36 47

Screenshot 2021-02-09 at 11 37 13

Screenshot 2021-02-09 at 11 37 40

### DIFF
--- a/Tests/DatadogTests/Datadog/Mocks/SystemFrameworks/FoundationMocks.swift
+++ b/Tests/DatadogTests/Datadog/Mocks/SystemFrameworks/FoundationMocks.swift
@@ -33,7 +33,15 @@ import Foundation
 
 // MARK: - Basic types
 
-extension Data {
+protocol AnyMockable {
+    static func mockAny() -> Self
+}
+
+protocol RandomMockable {
+    static func mockRandom() -> Self
+}
+
+extension Data: AnyMockable {
     static func mockAny() -> Data {
         return Data()
     }
@@ -71,7 +79,13 @@ extension Array {
     }
 }
 
-extension Date {
+extension Dictionary: AnyMockable where Key: AnyMockable, Value: AnyMockable {
+    static func mockAny() -> Dictionary {
+        return [Key.mockAny(): Value.mockAny()]
+    }
+}
+
+extension Date: AnyMockable {
     static func mockAny() -> Date {
         return Date(timeIntervalSinceReferenceDate: 1)
     }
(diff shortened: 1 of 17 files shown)


Tests/DatadogTests/Datadog/Mocks/SystemFrameworks/FoundationMocks.swift
```diff
@@ -33,7 +33,15 @@ import Foundation
...
 // MARK: - Basic types
 
-extension Data {
+protocol AnyMockable {
+    static func mockAny() -> Self
+}
+
+protocol RandomMockable {
+    static func mockRandom() -> Self
+}
```
ncreated, line 42: Nice 💪! I feel that we could even push this further with some level of code generation in the future 💡.

Tests/DatadogTests/Datadog/TracerTests.swift
```diff
@@ -583,7 +583,7 @@ class TracerTests: XCTestCase {
         )
         XCTAssertEqual(
             try spanMatcher.meta.custom(keyPath: "meta.person"),
-            #"{"age":30,"name":"Adam","nationality":"Polish"}"#
+            #"{"name":"Adam","age":30,"nationality":"Polish"}"#
```
ncreated, line 586: Is it required because of the change in `JSONEncoder`? If so - are we sure it's not flaky?

ncreated (approved):
Looks 💯👌

Tests/DatadogTests/Datadog/TracerTests.swift
acgh: Yeah this was added in #323 along with the key-sorting for `JSONEncoder`:
https://github.com/DataDog/dd-sdk-ios/pull/323/files#diff-25ffd9e630ae2d8dc218a66fa74cdb1ccf96171e92a96b13cc7270d7c9a7b77f

So after reverting the `JSONEncoder` setup, this test should fail consistently if we don't revert that line as well.

PR #413, opened 2021-02-16
RUMM-1053 Improve the `error.message` for Crash Reports

ncreated:
### What and why?

📦 This PR aims at improving `error.type` and `error.message` for Crash Reports by providing additional troubleshooting information for devs.

### How?

I've spent most of my time researching possible improvements and considering when and where they can be applied. To examine multiple variants, I updated the Crash Reporting scenario in the `Example` app with new cases:

Screenshot 2021-02-16 at 14 58 34

After all, this resultant PR improves only the case where the crash originates from unhandled Objective-C exception. In such scenario, the `PLCrashReporter` records the exception info and provides it on the `PLCrashReport`, so we can consume it and pass to RUM Error.

Tapping above buttons bottom-up results with following `error.type` and `error.message` being sent:

crash-reporting-vague

### Other improvements

This PR also disables on-device symbolication of crash reports. Applying it on-device gives more readable reports, but also makes backend-side symbolication impossible. We may consider enabling it in the future (while still sending unsymbolicated reports to backend) → `RUMM-1088`.

I also updated the `DDCrashReport` structure to contain digested information: `type`, `message`, `stackTrace`, instead of their raw representation from `DatadogCrashReporting` (`signalCode`, `signalName`, ...). This should better shape responsibilities of both modules, as now `DCR` is responsible for building and formatting the crash report, and `Datadog's` role is only to send it with RUM or Logger.

### Further iterations on error messages

Getting more from crash reports is not an one-off effort and will need to be continued with next iterations after we start symbolicating stack frames. Full data should enable further possibilities of crash diagnostics and inferring their root causes. To schedule this work, I created `RUMM-1088` where I describe more conclusions from my research.

### Review checklist

### DIFF
--- a/Sources/DatadogCrashReporting/PLCrashReporterIntegration/PLCrashReportFormatter.swift
+++ b/Sources/DatadogCrashReporting/PLCrashReporterIntegration/PLCrashReportFormatter.swift
@@ -1,0 +1,88 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+import CrashReporter
+import Datadog
+
+/// A convenience interface of the`PLCrashReport` to enable its testability.
+internal protocol PLCrashReportType {
+    var timestamp: Date? { get }
+    var signalName: String? { get }
+    var signalCode: String? { get }
+    var exceptionName: String? { get }
+    var exceptionReason: String? { get }
+    var formattedStackTrace: String? { get }
+    var contextData: Data? { get }
+}
+
+extension PLCrashReport: PLCrashReportType {
+    var timestamp: Date? { systemInfo?.timestamp }
+    var signalName: String? { signalInfo?.name }
+    var signalCode: String? { signalInfo?.code }
+    var exceptionName: String? { hasExceptionInfo ? exceptionInfo?.exceptionName : nil }
+    var exceptionReason: String? { hasExceptionInfo ? exceptionInfo?.exceptionReason : nil }
+    var formattedStackTrace: String? { PLCrashReportTextFormatter.stringValue(for: self, with: PLCrashReportTextFormatiOS) }
+    var contextData: Data? { customData }
+}
+
+/// Formats the `PLCrashReport` to Datadog format (`DDCrashReport`).
+internal struct PLCrashReportFormatter {
+    private let unknown = "<unknown>"
+    private let unavailable = "<unavailable>"
+
+    func ddCrashReport(from crashReport: PLCrashReportType) -> DDCrashReport {
+        return DDCrashReport(
+            date: crashReport.timestamp,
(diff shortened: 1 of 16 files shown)


Sources/DatadogCrashReporting/PLCrashReporterIntegration/PLCrashReportFormatter.swift
```diff
@@ -0,0 +1,88 @@
...
+
+    private func readMessage(from crashReport: PLCrashReportType) -> String {
+        if crashReport.exceptionName != nil || crashReport.exceptionReason != nil {
+            // If the crash was caused by an uncaught exception
+            let exceptionName = crashReport.exceptionName // e.g. `NSInvalidArgumentException`
+            let exceptionReason = crashReport.exceptionReason // e.g. `-[NSObject objectForKey:]: unrecognized selector sent to instance 0x...`
+            return "Terminating app due to uncaught exception '\(exceptionName ?? unknown)', reason: '\(exceptionReason ?? unknown)'."
+        } else {
+            // Use signal description available in OS
+            guard let signalName = crashReport.signalName else { // e.g. SIGILL
```
acgh, line 61: I'm wondering if for Swift runtime exceptions, we could leverage the (little) information available here:

https://developer.apple.com/documentation/xcode/diagnosing_issues_using_crash_reports_and_device_logs/identifying_the_cause_of_common_crashes/addressing_crashes_from_swift_runtime_errors

SIG TRAP 5.

acgh (approved):
I agree with the approach of deferring the complex analysis to the backend-side!
I also think that dumping uncaught `NSException`s should be very helpful. If I understand correctly, disabling the on-device symbolication should prevent any extra work (that could fail) at the time of crash reporting so it's better for stability.
I haven't much else to say about the code itself, it looks great.

Sources/DatadogCrashReporting/PLCrashReporterIntegration/PLCrashReportFormatter.swift
ncreated: I added one more variant: _"Force try! Swift Error"_, which does:
```swift
try! throwException()
```
Got this on my iPhone X:

Screenshot 2021-02-16 at 17 03 16

```
Hardware Model: iPhone10,6
Process: Example [24108]
...
Code Type: ARM-64
...
OS Version: iPhone OS 14.3 (18C66)
...
Exception Type: SIGTRAP
Exception Codes: #0 at 0x1a0eceaa8
```
It's somehow similar to what Apple doc says, but doesn't give `Termination Reason: Namespace SIGNAL, Code 0x5` 💭 ...

ncreated:
> If I understand correctly, disabling the on-device symbolication should prevent any extra work (that could fail) at the time of crash reporting so it's better for stability.

Not exactly - the on-device symbolication can happen after the app is restarted. We can symbolicate the crash report using runtime addresses from the new instance of the process.

To get the symbol we need its relative address (offset) and its library address. Relative addresses are the same, no matter of the process instance. But the library can be loaded ad different address each time, so the "trick" is to find the new address of the library and then get the symbol from offset. It's doable at runtime 👌.

PR #377, opened 2021-01-19
RUMM-924 Add Data Scrubbing and GDPR docs

ncreated:
### What and why?

🎓 This PR updates docs with recent APIs introduced since `1.4.x`: tracking consent and data scrubbing.

### How?

To align it a bit more with [`dd-sdk-android` docs](https://github.com/DataDog/dd-sdk-android/tree/master/docs):
* I created separate `gdpr.md` doc,
* added Tracking Consent API description in all feature docs (Logging / Tracing / RUM),
* added Data Scrubbing section in RUM docs,
* added picture and link to RUM docs in repository's `README.md`.

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Merge to / after `release/1.5.0` is issued

PR #415, opened 2021-02-17
RUMM-1077 Prevent discarding RUM View events

acgh:
### What and why?

My understanding is that the view event opens a parent-child relationship with the subsequent Error/Resource/Action events, and dropping a view event would cause "leaks" of Error/Resource/Action events.
Thus a solution is to forbid dropping view events while keeping the ability to redact their properties and drop "children" events.

### How?

In this PR we change the `RUMViewEventMapper` API to statically prevent from dropping `RUMViewEvent` by requiring a non-optional return type.

As this is statically enforced, we also remove a few checks in existing test cases as they would not compile nor make sense anymore.

### Review checklist

### DIFF
--- a/Sources/Datadog/DatadogConfiguration.swift
+++ b/Sources/Datadog/DatadogConfiguration.swift
@@ -456,7 +456,10 @@ extension Datadog {
             /// Sets the custom mapper for `RUMViewEvent`. This can be used to modify RUM View events before they are send to Datadog.
             /// - Parameter mapper: the closure taking `RUMViewEvent` as input and expecting `RUMViewEvent` as output.
             /// The implementation should obtain a mutable version of the `RUMViewEvent`, modify it and return it.
-            /// Note that the mapper intentionally prevents from returning a `nil` to drop the `RUMViewEvent` entirely, this ensures that all `RUMViewEvent` are sent to Datadog.
+            ///
+            /// **NOTE** The mapper intentionally prevents from returning a `nil` to drop the `RUMViewEvent` entirely, this ensures that all `RUMViewEvent` are sent to Datadog.
+            ///
+            /// Use the `UIKitRUMViewsPredicate` API to ensure upstream consideration or filtering out of `UIViewController`/`RUMView`s.
             public func setRUMViewEventMapper(_ mapper: @escaping (RUMViewEvent) -> RUMViewEvent) -> Builder {
                 configuration.rumViewEventMapper = mapper
                 return self
--- a/Sources/Datadog/DatadogConfiguration.swift
+++ b/Sources/Datadog/DatadogConfiguration.swift
@@ -454,10 +454,10 @@ extension Datadog {
             }
 
             /// Sets the custom mapper for `RUMViewEvent`. This can be used to modify RUM View events before they are send to Datadog.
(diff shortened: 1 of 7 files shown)


ncreated (approved):
Looks good 👌.

Could you also add a change suggestion [in the PR with RUM docs update](https://github.com/DataDog/dd-sdk-ios/pull/377/files#diff-6829a1f2fe259b327b2c2c9b8352da5bebc49b7001e189e0df905d6ac08b54ec)? We should add a note there, that returning `nil` is impossible for view events.

Sources/Datadog/DatadogConfiguration.swift
```diff
@@ -454,10 +454,10 @@ extension Datadog {
             }
 
             /// Sets the custom mapper for `RUMViewEvent`. This can be used to modify RUM View events before they are send to Datadog.
-            /// - Parameter mapper: the closure taking `RUMViewEvent` as input and expecting `RUMViewEvent` or `nil` as output.
-            /// The implementation should obtain a mutable version of the `RUMViewEvent`, modify it and return. Returning `nil` will result
-            /// with dropping the RUM View event entirely, so it won't be send to Datadog.
-            public func setRUMViewEventMapper(_ mapper: @escaping (RUMViewEvent) -> RUMViewEvent?) -> Builder {
+            /// - Parameter mapper: the closure taking `RUMViewEvent` as input and expecting `RUMViewEvent` as output.
+            /// The implementation should obtain a mutable version of the `RUMViewEvent`, modify it and return it.
+            /// Note that the mapper intentionally prevents from returning a `nil` to drop the `RUMViewEvent` entirely, this ensures that all `RUMViewEvent` are sent to Datadog.
```
buranmert, line 459: if i were a user, that could be a bit confusing to me
i have `ViewPredicate` where I can keep or drop views and then i also have `viewEventMapper` which intentionally prevents me from dropping them.

i have a dilemma here:
1. if we mention `ViewPredicate` here and how it's different than `viewMapper`, etc. this may be even more confusing in case the user never saw `ViewPredicate` API before
2. if we don't mention it, this may be still confusing if the user is already aware of `ViewPredicate` API

what do you think about that?

acgh: Those are good points and that can look confusing.

My understanding is that the `ViewPredicate` gives an opportunity to drop views already, and downstream the `RUMViewEventMapper` will be fed with events based on views that have been already filtered.
Before that PR, we give the opportunity (again) to drop view related data at the `RUMViewEventMapper`, I think it can create a confusing pipeline (the other way around regarding your concerns).

My point is that `ViewPredicate` let's the user code filter out views, there will be no events for already filtered out views and `RUMViewEventMapper` will only receive events for views that have been chosen already.

Does that make sense?  

That being said, maybe it would help to mention the `UIKitRUMViewsPredicate` here for context.

ncreated: I'd vote for mentioning the `UIKitRUMViewsPredicate` in there.

Using this scrubbing API to ignore particular `UIViewControllers` from being tracked is a misuse of the SDK APIs. It means the user doesn't understand the SDK auto instrumentation & scrubbing APIs well and we should guide him towards the success.

That said, this change to "view events cannot be dropped" makes the contract between auto instrumentation & scrubbing APIs even more solid and puts the user on pathway to success.

PR #418, opened 2021-02-18
RUMM-1030 Add custom "VIEW NAME" parameter

ncreated:
### What and why?

🚚 This PR adds support to "VIEW NAME" parameter as the primary attribute to differentiate views in RUM Explorer, preferred over "VIEW URL" which doesn't fit well mobile ecosystems:

Screenshot 2021-02-18 at 13 19 30

From now on, the "VIEW URL" will use canonical class name of the underlying `UIViewController` and the "VIEW NAME" can be freely customized by the user (although it will default to "VIEW URL" if not specified).

### How?

This includes two breaking changes to the public API:
```diff
- Global.rum.startView(viewController:path:attributes)
+ Global.rum.startView(viewController:name:attributes)
```
```diff
- RUMView(path:attributes)
+ RUMView(name:attributes)
```
and one non-breaking (as this API is not there in `1.4.x`):
```diff
- Global.rum.startView(key:path:attributes)
+ Global.rum.startView(key:name:attributes)
```

Under the hood, `name` property was introduced in `RUMStartViewCommand` and necessary tests were updated on different levels.

I did additional update to the RUM debugging utility, so now it renders view `name`, instead of `path`, e.g.:

change-to-debug-rum

I also replaced all usages of `viewURI` by renaming it to the `viewPath`. The therm `URI` seemed correct, when we were collecting `path` in public API and using it to describe both: technical view location and human-readable view identity. Now both can be clearly separated by `viewPath` and `viewName`.

### Review checklist


acgh (approved):
Looks good! 🚀

PR #419, opened 2021-02-18
RUMM-1097 Fix `DefaultUIKitRUMViewsPredicate` for Objc view controllers

ncreated:
### What and why?

🐛 When RUM Views auto instrumentation was enabled, no `UIViewController` defined in `Objc` was tracked. No matter if the SDK was configured in Objc:
```objc
[configuration trackUIKitRUMViews];
```
or swift:
```swift
.trackUIKitRUMViews()
```

### How?

The problem was caused by the filtering logic done in `DefaultUIKitRUMViewsPredicate`:
```swift
public func rumView(for viewController: UIViewController) -> RUMView? {
   let canonicalClassName = NSStringFromClass(type(of: viewController))
   let isCustomClass = canonicalClassName.contains(".") // custom class contains module prefix

   if isCustomClass {
      // ... capture the view
   } else {
      return nil // ignore the view
   }
}
```
`isCustomClass` was meant to indicate if the `viewController` is a custom VC provided by the user app (`true`) or is a standard VC coming from `UIKit` (`false`). This was to ignore all the "system" view controllers like `UINavigationViewController`, `UITabBarController` etc.

🐛 The implementation was based on the observation that custom Swift VCs are prefixed with module name when inspecting their name through `NSStringFromClass()` (e.g. `MyApp.LoginViewController`). This turned out to be not true for custom Objc VCs, which have no module prefix.

🌟 The fix is to change the way we recognise custom VC. Now we lookup the `Bundle` that the class is loaded from and we simply filter out all VCs coming from `UIKitCore.framework` bundle.

### Why it was not catched before?

This bug was not visible, because we didn't have any tests covering Obj-C view controllers and RUM Views predicate.

### How to prevent it in the future?

I added one unit test for covering this exact scenario and configured one of the integration tests to also work with Objective-C view controllers (to cover other potential issues of mixed Swift <> Objc codebase instrumentation).

The integration test was done by splitting the existing `RUMResourcesScenario` into two separate scenarios: `RUMURLSessionResourcesScenario` and `RUMNSURLSessionResourcesScenario`. They both have identical setup, storyboards and view controller implementations. The prior defines its VCs in Swift, the latter in Objective-C. It didn't require adding anything new in the UI, as those tests are based on the existing `URLSession` + `NSURLSession` scenarios.

Screenshot 2021-02-18 at 18 41 48

As a bonus 🎁 we gain integration coverage for RUM Resources tracked from Obj-c code.

### Review checklist

### DIFF
--- a/Sources/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsPredicate.swift
+++ b/Sources/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsPredicate.swift
@@ -59,7 +59,9 @@ public struct DefaultUIKitRUMViewsPredicate: UIKitRUMViewsPredicate {
 
     public func rumView(for viewController: UIViewController) -> RUMView? {
         guard !isUIKit(class: type(of: viewController)) else {
-            // do not track bare UIKit's view controllers (UINavigationController, UITabBarController, ...)
+            // Part of our heuristic for (auto) tracking view controllers is to ignore
+            // container view controllers coming from `UIKit` if they are not subclassed.
+            // This condition is wider and it ignores all view controllers defined in `UIKit` bundle.
             return nil
         }
 
@@ -71,7 +73,6 @@ public struct DefaultUIKitRUMViewsPredicate: UIKitRUMViewsPredicate {
 
     /// If given `class` comes from UIKit framework.
     private func isUIKit(`class`: AnyClass) -> Bool {
-        let bundle = Bundle(for: `class`)
-        return bundle.bundleURL.lastPathComponent == "UIKitCore.framework"
+        return Bundle(for: `class`).isUIKit
     }
 }
--- a/Sources/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsPredicate.swift
+++ b/Sources/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsPredicate.swift
@@ -58,15 +58,20 @@ public struct DefaultUIKitRUMViewsPredicate: UIKitRUMViewsPredicate {
     public init () {}
 
     public func rumView(for viewController: UIViewController) -> RUMView? {
-        let canonicalClassName = viewController.canonicalClassName
-        let isCustomClass = canonicalClassName.contains(".") // custom class contains module prefix
-
-        if isCustomClass {
-            var view = RUMView(name: canonicalClassName)
-            view.path = canonicalClassName
-            return view
-        } else {
+        guard !isUIKit(class: type(of: viewController)) else {
+            // do not track bare UIKit's view controllers (UINavigationController, UITabBarController, ...)
             return nil
         }
+
+        let canonicalClassName = viewController.canonicalClassName
+        var view = RUMView(name: canonicalClassName)
+        view.path = canonicalClassName
+        return view
+    }
+
+    /// If given `class` comes from UIKit framework.
+    private func isUIKit(`class`: AnyClass) -> Bool {
+        let bundle = Bundle(for: `class`)
+        return bundle.bundleURL.lastPathComponent == "UIKitCore.framework"
     }
 }
--- a/Tests/DatadogTests/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsPredicateTests.swift
+++ b/Tests/DatadogTests/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsPredicateTests.swift
@@ -17,6 +17,7 @@ class UIKitRUMViewsPredicateTests: XCTestCase {
         let rumView = predicate.rumView(for: customViewController)
 
         // Then
+        XCTAssertEqual(rumView?.name, "Module.CustomViewController")
         XCTAssertEqual(rumView?.path, "Module.CustomViewController")
         XCTAssertTrue(rumView!.attributes.isEmpty)
     }
--- a/Tests/DatadogTests/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsPredicateTests.swift
+++ b/Tests/DatadogTests/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsPredicateTests.swift
@@ -8,17 +8,31 @@ import XCTest
 import Datadog
 
 class UIKitRUMViewsPredicateTests: XCTestCase {
-    func testGivenDefaultPredicate_whenAskingForCustomViewController_itNamesTheViewByItsClassName() {
+    func testGivenDefaultPredicate_whenAskingForCustomSwiftViewController_itNamesTheViewByItsClassName() {
         // Given
         let predicate = DefaultUIKitRUMViewsPredicate()
 
         // When
-        let customViewController = createMockView(viewControllerClassName: "Module.CustomViewController")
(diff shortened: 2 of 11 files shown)


Tests/DatadogTests/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsPredicateTests.swift
```diff
@@ -8,16 +8,31 @@ import XCTest
 import Datadog
 
 class UIKitRUMViewsPredicateTests: XCTestCase {
```
acgh, line 10: We could also add test cases to check that some known UIKit classes are filtered out, to catch any regressions of  the internal `isUIKit` with future releases or iOS.

acgh (approved):
This is great 💪

Sources/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsPredicate.swift
```diff
@@ -58,15 +58,20 @@ public struct DefaultUIKitRUMViewsPredicate: UIKitRUMViewsPredicate {
...
         }
+
+        let canonicalClassName = viewController.canonicalClassName
+        var view = RUMView(name: canonicalClassName)
+        view.path = canonicalClassName
+        return view
+    }
+
+    /// If given `class` comes from UIKit framework.
+    private func isUIKit(`class`: AnyClass) -> Bool {
```
acgh, line 73: Nitpick: Put this as an internal extension on `Bundle` in the `UIKitExtensions.swift` file

Sources/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsPredicate.swift
```diff
@@ -58,15 +58,20 @@ public struct DefaultUIKitRUMViewsPredicate: UIKitRUMViewsPredicate {
...
-        let canonicalClassName = viewController.canonicalClassName
-        let isCustomClass = canonicalClassName.contains(".") // custom class contains module prefix
-
-        if isCustomClass {
-            var view = RUMView(name: canonicalClassName)
-            view.path = canonicalClassName
-            return view
-        } else {
+        guard !isUIKit(class: type(of: viewController)) else {
+            // do not track bare UIKit's view controllers (UINavigationController, UITabBarController, ...)
```
buranmert, line 62: the real goal is excluding container view controllers, exclusion of `UIKit` types is just the workaround that we could come up with.
this comment here may be misleading if we come back to this issue later on, can we correct it?

here is the same issue in PR's description as well:
> ...This was to ignore all the "system" view controllers...

Tests/DatadogTests/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsPredicateTests.swift
buranmert: +1 to @acgh , we rely on private information (`UIKitCore` is private) and that may break at any time

buranmert (commented):
did we try it in a real device? IMO it's unlikely but `bundleURL` may be different there

Tests/DatadogTests/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsPredicateTests.swift
ncreated: This is already tested on `master`:
https://github.com/DataDog/dd-sdk-ios/blob/4e952d31597cd4a1e1c774c6411e1efb7086dcf4/Tests/DatadogTests/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsPredicateTests.swift#L24

https://github.com/DataDog/dd-sdk-ios/blob/4e952d31597cd4a1e1c774c6411e1efb7086dcf4/Tests/DatadogTests/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsPredicateTests.swift#L11

ncreated: It's already there - just not visible in this diff. We had this test before:
```swift
    func testGivenDefaultPredicate_whenAskingUIKitViewController_itReturnsNoView() {
        // Given
        let predicate = DefaultUIKitRUMViewsPredicate()
        // When
        let uiKitViewController = UIViewController()
        let rumView = predicate.rumView(for: uiKitViewController)
        // Then
        XCTAssertNil(rumView)
    }
```

Sources/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsPredicate.swift
ncreated: No no, it's about plain (not subclassed / customized) VCs from `UIKit`.

Before this PR, this implementation was considering `MyCustomNavigationViewController` as the RUM View. I think we shouldn't change that default if no one has reported it being wrong. We're doing a bug fix here, I wouldn't change the previous behaviour.

Sources/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsPredicate.swift
ncreated: Do you mean sth like `Bundle.isUIKit(class:) -> Bool`? This would read weird, no? This seems to be rather a characteristic of the `AnyClass` than a `Bundle` - but as `AnyClass` is a metatype it cannot be extended.

buranmert: it's not important for me but i was thinking of smth like this:
```
// extension
extension AnyObject {
  var isUIKit: Bool { ... }
}
// calling site
if viewController.isUIKit { return }
else { ... }
```

Sources/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsPredicate.swift
buranmert: no i'm not talking about the behavior, my concern is that the comments don't tell the whole story. 
we first wanted to exclude containers and then noticed invisible view controllers in the hiearchy. if one asks "why do we exclude UIKit types?", the answer is not obvious without this story.

IMO this comment `do not track UIKit controllers` brings questions (_why?_) rather than answering

Sources/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsPredicate.swift
ncreated: Done 👌 , I added:
```swift
internal extension Bundle {
    var isUIKit: Bool
}
```
extension and bunch of fixtures in tests.

Sources/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsPredicate.swift
ncreated: I see 👍. Improved it with discussing the reasoning and optimisation.

buranmert (approved):

PR #423, opened 2021-02-24
RUMM-1113 Pass logged error to RUM integration

buranmert:
### What and why?

When `logger.error(...)` is called, `LoggingWithRUMIntegration` creates an equivalent RUM error.
However, this integration is not fully up to date: when it was written `logger.error(...)` could only take `message: String` as parameter but now it also takes `error: Error?` as parameter.

### How?

This PR passes logged `error` to RUM integration. That way, RUM error is created out of the logged `error` instead of logged `message`. This gives more accurate information in RUM Explorer as can be seen in screenshots below.

#### `log.error("an error message occurred")`

log-without-error

#### `log.error("some message", error: NSError(...))`

log-with-error

### Review checklist


Sources/Datadog/Logging/LogOutputs/LogOutput.swift
```diff
@@ -15,5 +15,5 @@ internal struct LogAttributes {
 
 /// Type writting logs to some destination.
 internal protocol LogOutput {
-    func writeLogWith(level: LogLevel, message: String, date: Date, attributes: LogAttributes, tags: Set<String>)
+    func writeLogWith(level: LogLevel, message: String, error: Error?, date: Date, attributes: LogAttributes, tags: Set<String>)
```
ncreated, line 18: If we change `error: Error?` to `dderror: DDError?` here, we won't need to play with the vague `Error` representation downstream (it will be mapped to `DDError` early in the `Logger`). This will also enable seamless integration of crash reporting (`RUMM-1050`), as the `DDError()` for crash report can be easily constructed.

Tests/DatadogTests/Datadog/Mocks/LoggingFeatureMocks.swift
```diff
@@ -189,14 +191,22 @@ class LogOutputMock: LogOutput {
...
 
     var recordedLog: RecordedLog? = nil
 
-    func writeLogWith(level: LogLevel, message: String, date: Date, attributes: LogAttributes, tags: Set<String>) {
-        recordedLog = RecordedLog(level: level, message: message, date: date, attributes: attributes, tags: tags)
+    func writeLogWith(level: LogLevel, message: String, error: Error?, date: Date, attributes: LogAttributes, tags: Set<String>) {
+        recordedLog = RecordedLog(
+            level: level,
+            message: message,
+            error: error.flatMap { DDError(error: $0) },
```
ncreated, line 206: ⚠️  Another reason for using `DDError` in `LogOutput` signature. Now we need to make this mock smarter (by using part of the real output logic), to make it work. Mapping to `DDError` as soon as possible (in `Logger`) would fix this.

ncreated (changes requested):
Looks very good 👌, but I left few notes on how we can make `DDError` definition & usage better for what's required in crash reporting.

Sources/Datadog/Core/Utils/DDError.swift
```diff
@@ -7,7 +7,7 @@
 import Foundation
 
 /// Common representation of Swift `Error` used by different features.
-internal struct DDError {
+internal struct DDError: Equatable {
     let title: String
     let message: String
     let details: String
```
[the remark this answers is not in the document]
buranmert: good idea, done ✅

Sources/Datadog/RUMMonitor.swift
```diff
@@ -307,11 +307,13 @@ public class RUMMonitor: DDRUMMonitor, RUMCommandSubscriber {
             }
             return nil
         }()
-        addError(message: message, stack: stack, source: RUMInternalErrorSource(source), attributes: attributes)
+        // Note for CR: should `type` be nil?
```
ncreated, line 310: I think this is what we concluded with Android SDK team, right?
```suggestion
```

Sources/Datadog/RUMMonitor.swift
```diff
@@ -491,6 +494,8 @@ public class RUMMonitor: DDRUMMonitor, RUMCommandSubscriber {
                 resourceKey: resourceKey,
                 time: dateProvider.currentDate(),
                 message: errorMessage,
+                // Note for CR: should we leave it nil?
```
ncreated, line 497: Same - concluded with the Android SDK team.
```suggestion
```

ncreated (approved):
Nice 🚀! Should help a lot with crash reporting as well.

Sources/Datadog/RUMMonitor.swift
buranmert: yes, removed ✅

PR #427, opened 2021-03-01
RUMM-1050 Refactor logs and spans processing in Logger and Tracer

ncreated:
### What and why?

🧰 This PR applies small refactoring to `Logs` and `Spans` processing within `Logger` and `Tracer`. This is required to continue the work on crash reporting, where we need to build the `Log` value for the crash report and pass it to the `LogFileOutput` without using `LogBuilder`.

The `LogBuilder` cannot be used in crash reporting, as its role is to decorate `Log` with a current process information (network / carrier / user info). Crash reporting passes this information through `CrashContext`.

Existing `LogFileOutput` cannot be used in crash reporting, as it takes raw log information and uses `LogBuilder` internally. Crash reporting requires `write(log: Log)` interface for `LogOutput`.

### How?

Before this PR, there was this asymmetry in the events processing for our 3 features:
```swift
// Logging:
let rawUserLog: (level: LogLevel, message: String, error: Error?, messageAttributes: [String: Encodable]?) = // ...
logOutput.writeLogWith(rawUserLog) // internally, in `LogFileOutput` the `LogBuilder` was building `Log` from `rawUserLog`

// Tracing:
let rawUserSpan: (ddspan: DDSpan, finishTime: Date) = // ...
spanOutput.writeSpan(rawUserSpan) // internally, in `SpanFileOutput` the `SpanBuilder` was building `Span` from `rawUserSpan`

// RUM:
let rawRUMEvent: (model: RUMDataModel, attributes: [String: Encodable]) = // ...
let rumEvent = rumEventBuilder.createRUMEvent(rawRUMEvent)
let rumEventOutput.write(rumEvent)
```

Logging and Tracing outputs were "smarter" comparing to RUM output. They were using `LogBuilder` and `SpanBuilder` internally to transform information received from the user into serializable `Log` or `Span`. This transformation was not visible on the stream level: `rawInformation → output`. On the other side, RUM abstraction is more streamlined and acts as a simple stream: `rawInformation → serializableInformation  → output`.

This PR uniforms the data stream for all 3 features, so now it's using the same level of abstraction. In pseudocode:
```swift
// Logging, Tracing, RUM:
let rawInformation = /* (log | span | RUM event) info */
let serializableInformation = (log|span|rumEvent)Builder.create(from: rawInformation)
output.write(serializableInformation)
```

### Benefits

This unlocks the work on crash reporting, where an `ERROR` `Log` must be created from `DDCrashReport` without using the `LogBuilder`.

This refactoring makes many interfaces and tests cleaner, as now the amount of (raw) information is reduced upstream and what's passed downstream is simply `Log`, `Span` or `RUMEvent`.

This also makes many tests more solid, as the `LogOutput` and `SpanOutput` mocks now expect a concrete `Log` / `Span` values, instead of raw non-digested pieces of information. This made me update few tests with a clear and concrete assertions on what's received by the mock outputs, instead of just asserting if they receive what's passed to the surface.

Last, this uniforms the processing of `Logs` and `Spans` with existing processing models for RUM events.

### Review checklist

### DIFF
--- a/Sources/Datadog/Logger.swift
+++ b/Sources/Datadog/Logger.swift
@@ -15,6 +15,30 @@ public enum LogLevel: Int, Codable {
     case warn
     case error
     case critical
+
+    // MARK: - `LogLevel` <> `Log.Status` conversion
+
+    internal var asLogStatus: Log.Status {
+        switch self {
+        case .debug:    return .debug
+        case .info:     return .info
+        case .notice:   return .notice
+        case .warn:     return .warn
+        case .error:    return .error
+        case .critical: return .critical
+        }
+    }
+
+    internal init(from logStatus: Log.Status) {
+        switch logStatus {
+        case .debug:    self = .debug
+        case .info:     self = .info
+        case .notice:   self = .notice
+        case .warn:     self = .warn
+        case .error:    self = .error
+        case .critical: self = .critical
+        }
+    }
 }
 
 /// Because `Logger` is a common name widely used across different projects, the `Datadog.Logger` may conflict when
@@ -33,8 +57,10 @@ public enum LogLevel: Int, Codable {
 public typealias DDLogger = Logger
 
 public class Logger {
-    /// Writes `Log` objects to output.
-    let logOutput: LogOutput
+    /// Builds the `Log` from user input; `nil` for no-op logger.
+    internal let logBuilder: LogBuilder?
+    /// Writes the `Log` to file; `nil` for no-op logger.
+    internal let logOutput: LogOutput?
     /// Provides date for log creation.
     private let dateProvider: DateProvider
     /// Attributes associated with every log.
@@ -51,12 +77,14 @@ public class Logger {
     internal let environmentSpanIntegration = LoggingWithEnvironmentSpanIntegration()
 
     init(
-        logOutput: LogOutput,
+        logBuilder: LogBuilder?,
+        logOutput: LogOutput?,
         dateProvider: DateProvider,
         identifier: String,
         rumContextIntegration: LoggingWithRUMContextIntegration?,
         activeSpanIntegration: LoggingWithActiveSpanIntegration?
     ) {
+        self.logBuilder = logBuilder
         self.logOutput = logOutput
         self.dateProvider = dateProvider
         self.queue = DispatchQueue(
@@ -217,7 +245,9 @@ public class Logger {
     // MARK: - Private
 
     private func log(level: LogLevel, message: String, error: Error?, messageAttributes: [String: Encodable]?) {
-        let date = dateProvider.currentDate()
+        guard let logBuilder = logBuilder, let logOutput = logOutput else {
+            return // ignore, as the `Logger` is no-op
+        }
 
         var combinedUserAttributes = messageAttributes ?? [:]
         combinedUserAttributes = queue.sync {
@@ -240,17 +270,19 @@ public class Logger {
             return self.loggerTags
         }
 
-        logOutput.writeLogWith(
+        let log = logBuilder.createLogWith(
             level: level,
             message: message,
             error: error.flatMap { DDError(error: $0) },
-            date: date,
+            date: dateProvider.currentDate(),
             attributes: LogAttributes(
                 userAttributes: combinedUserAttributes,
                 internalAttributes: combinedInternalAttributes
             ),
             tags: tags
         )
+
+        logOutput.write(log: log)
     }
 
     // MARK: - Logger.Builder
@@ -357,7 +389,8 @@ public class Logger {
             } catch {
                 consolePrint("\(error)")
                 return Logger(
-                    logOutput: NoOpLogOutput(),
+                    logBuilder: nil,
+                    logOutput: nil,
                     dateProvider: SystemDateProvider(),
                     identifier: "no-op",
                     rumContextIntegration: nil,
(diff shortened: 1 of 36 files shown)


acgh (approved):
This standardization of data processing/writing across Log/Traces/RUM makes a lot of sense, and the code looks great 👍

Tests/DatadogTests/Datadog/Mocks/LoggingFeatureMocks.swift
```diff
@@ -103,25 +105,51 @@ extension Log {
...
 
-extension Log.Status {
+extension Log.Status: RandomMockable {
     static func mockAny() -> Log.Status {
         return .info
     }
+
+    static func mockRandom() -> Log.Status {
+        return [.debug, .info, .notice, .warn, .error, .critical].randomElement()!
+    }
```
buranmert, line 137: you can make `Log.Status: CaseIterable` and then `Log.Status.allCases.randomElement!` instead of manually maintaining this array

Sources/Datadog/Logging/LogOutputs/LogFileOutput.swift
```diff
@@ -6,18 +6,16 @@
...
     let rumErrorsIntegration: LoggingWithRUMErrorsIntegration?
 
-    func writeLogWith(level: LogLevel, message: String, error: DDError?, date: Date, attributes: LogAttributes, tags: Set<String>) {
-        let log = logBuilder.createLogWith(level: level, message: message, error: error, date: date, attributes: attributes, tags: tags)
+    func write(log: Log) {
         fileWriter.write(value: log)
 
-        if level.rawValue >= LogLevel.error.rawValue {
+        if log.status == .error || log.status == .critical {
             rumErrorsIntegration?.addError(for: log)
```
buranmert, line 19: when i worked on this part recently, i didn't fully understand why passing logs to RUM is `LogFileOutput`s responsibility and why doesn't `LogFileOutputTests` have it as a test case?

is it a product requirement that RUM integration works only with file output and not with console output?

buranmert (commented):
LG, this makes it easier to understand and track inputs and outputs of components 👍 
but i left a couple of questions before approving

Sources/Datadog/Logger.swift
```diff
@@ -15,6 +15,30 @@ public enum LogLevel: Int, Codable {
     case warn
     case error
     case critical
+
+    // MARK: - `LogLevel` <> `Log.Status` conversion
```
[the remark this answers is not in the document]
ncreated: Yes, they are the same and we could remove `Log.Status` in favour of `LogLevel`. This would require `switch / case` in `LogEncoder` to implement encoding of each level as `String`. With `Log.Status: String` this is now automated and this mapping here keeps it consistent.

Tests/DatadogTests/Datadog/Mocks/LoggingFeatureMocks.swift
ncreated: `CaseIterable` implies SIL code generation and must be declared in the same file as `Log.Status`. I try to avoid enhancing production types with test utilities - to keep the compiled binary size small. But maybe it's too puristic way and we can do it as we do with `Equatable` in few places already 👍.

Sources/Datadog/Logging/LogOutputs/LogFileOutput.swift
ncreated: This is tested using behaviour test defined in `LoggerTests`:
```swift
func testWhenSendingErrorOrCriticalLogs_itCreatesRUMErrorForCurrentView()
```
Thanks to this, we can move this code freely without touching its test. `LogFileOutput` is just one of the places it could be defined at (it could be done in `Logger` as well).

Tests/DatadogTests/Datadog/Mocks/LoggingFeatureMocks.swift
ncreated: Done 👍.

Sources/Datadog/Logger.swift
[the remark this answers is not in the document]
ncreated: Also, because those are two separate types now, we can adjust one (e.g. make it raw `String` representable + `CaseIterable`) without touching the other (defined in `public` API). If they would become one, we'd have `public`, `CaseIterable`, raw `String` representable, `Encodable` enum in public API. Even today's raw `Int` representable + `Codable` is useless for the public API of `LogLevel` (it shouldn't be there, but as it's public now there's no easy way back).

Sources/Datadog/Logging/LogOutputs/LogFileOutput.swift
ncreated: > is it a product requirement that RUM integration works only with file output and not with console output?

The reasoning here was: if the user doesn't want logs to be send to Datadog, he/she won't want RUM Errors for that logs to be send as well. It wasn't a product requirement.

buranmert: the single case in `LogFileOutputTests` is already covered by each case in `LoggerTests` but it is still there, this is confusing to me: why do we test `testItWritesLogToFileAsJSON` but not `testItSendsLogsToRUM` in the same place?

wouldn't it be easier to manage (eg: catching missing cases) if we test all `LogFileOutput` does in `LogFileOutputTests`?

buranmert (approved):

Sources/Datadog/Logging/LogOutputs/LogFileOutput.swift
ncreated: Discussed on Slack 🙂👌.

PR #428, opened 2021-03-02
RUMM-1050 Send crash report as Log

ncreated:
### What and why?

🚚 This PR sends collected crash reports as Logs. 

When the app restarts after crash, we load the crash report, read its associated context and then:

* if the Logging feature is enabled,
* and the RUM feature is disabled,
* and the crash report was collected with tracking consent `.granted`,
* then the crash report is send as `Log` with severity set to `EMERGENCY` (as we do in `dd-sdk-android`).

Example:

Screenshot 2021-03-02 at 12 28 57

### How?

Everything was prepared in prior #422 and #427. This PR only adds the implementation in `CrashReportingWithLoggingIntegration` as it was previously done for `CrashReportingWithRUMIntegration`.

This PR also adds separate scenario in `Example` app, so now we can test integration with RUM and Logging separately:

Screenshot 2021-03-02 at 13 25 08

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/Attributes/UserInfoProvider.swift
+++ b/Sources/Datadog/Core/Attributes/UserInfoProvider.swift
@@ -42,5 +42,5 @@ internal struct UserInfo {
     let email: String?
     let extraInfo: [AttributeKey: AttributeValue]
 
-    internal static let empty = UserInfo(id: nil, name: nil, email: nil, extraInfo: [:])
+    internal static var empty: UserInfo { UserInfo(id: nil, name: nil, email: nil, extraInfo: [:]) }
 }
--- a/Sources/Datadog/Core/Attributes/UserInfoProvider.swift
+++ b/Sources/Datadog/Core/Attributes/UserInfoProvider.swift
@@ -14,8 +14,7 @@ internal class UserInfoProvider {
     private let publisher: ValuePublisher<UserInfo>
 
     init() {
-        let emptyUserInfo = UserInfo(id: nil, name: nil, email: nil, extraInfo: [:])
-        self.publisher = ValuePublisher(initialValue: emptyUserInfo)
+        self.publisher = ValuePublisher(initialValue: .empty)
     }
 
     // MARK: - `UserInfo` Value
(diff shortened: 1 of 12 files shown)


buranmert (approved):

Sources/Datadog/Core/Attributes/UserInfoProvider.swift
```diff
@@ -42,4 +41,6 @@ internal struct UserInfo {
     let name: String?
     let email: String?
     let extraInfo: [AttributeKey: AttributeValue]
+
+    internal static let empty = UserInfo(id: nil, name: nil, email: nil, extraInfo: [:])
```
acgh, line 45: Nitpick: not sure how this is optimized by the compiler, but this convenience technically leads to a leak (never deallocated at least). An alternative would be an `internal convenience init`?

Sources/Datadog/FeaturesIntegration/CrashReporting/CrashReportingWithRUMIntegration.swift
```diff
@@ -18,7 +18,7 @@ internal struct CrashReportingWithRUMIntegration: CrashReportingIntegration {
         static let viewEventAvailabilityThreshold: TimeInterval = 14_400 // 4 hours
     }
 
-    /// The output for writting RUM events. It uses the authorized data folder and is synchronized with the eventual
+    /// The output for writing RUM events. It uses the authorized data folder and is synchronized with the eventual
```
acgh, line 21: Good catch!

acgh (approved):
Nothing much to say, this is clean ✨ :)

Sources/Datadog/FeaturesIntegration/CrashReporting/CrashReportingWithRUMIntegration.swift
ncreated: I found this option on Xcode 🔨 

Screenshot 2021-03-04 at 10 49 39

Sources/Datadog/Core/Attributes/UserInfoProvider.swift
ncreated: True 👍. Replaced it with static computed `var` - same footprint as initializer, but keeps the verbosity.

PR #433, opened 2021-03-08
RUMM-1078 Move RUMEventsMapper out of DataProcessing and into RUMEventBuilder

acgh:
for synchronous mapping in RUM Scopes

### What and why?

In #425 we have attempted introducing an asynchronous communication between `RUMEventsMapper` and `RUMScope`s, this came with a lot of complexity due to asynchrony and interleaving of `RUMCommand`s and `RUMEvent`s.

In this PR, we try another approach by making the communication between `RUMEventsMapper` and `RUMScope`s synchronous, and somewhat abstracted by `RUMEventBuilder`. This makes RUM Data Scrubbing decorrelated from the DataProcessing/Writing pipeline that is running downstream on a separate thread.

Events mapping is a user input just like `RUMCommand`s, in that sense it's consistent to have them both managed inside `RUMScope`s.

This change makes solving RUMM-1078 pretty simple and safe.

### How?

We move `RUMEventsMapper` out of `DataProcessor` into `RUMEventBuilder`.
We make the `create` method of `RUMEventBuilder` return the wrapped output of `RUMEventsMapper`, this means returning a `nil` when a `RUMEvent` should be dropped.

We move DataProcessor test cases into `RUMEventBuilder` test file and leave some placeholders to solve RUMM-1078 in `RUMViewScope` et al.

This is the pre-requisite to #434.

### Review checklist


Sources/Datadog/RUM/Scrubbing/RUMEventsMapper.swift
```diff
@@ -12,14 +12,16 @@ internal typealias RUMResourceEventMapper = (RUMResourceEvent) -> RUMResourceEve
 internal typealias RUMActionEventMapper = (RUMActionEvent) -> RUMActionEvent?
 
 /// The `EventMapper` for RUM events.
-internal struct RUMEventsMapper: EventMapper {
+internal struct RUMEventsMapper {
```
ncreated, line 15: 👍  This actually makes a lot of sense when I think of upcoming data scrubbing for other features (logging and tracing). Constraining all features to `f(T) -> T?` would not work well, as in APM scrubbing we may want to give users a `Span` along with the `URLRequest` / `HTTPURLResponse` for additional context decoration (e.g. logging the response body in case of error).

ncreated (approved):
I think it's a very good approach 👌. First: it simplifies `RUMM-1078`, second: it makes the event mapping interface less generic, so we will be able to incorporate more feature-specific things in mapping functions for other features.

buranmert (approved):

PR #436, opened 2021-03-08
RUMM-1064 AppStateListener added

buranmert:
### The problem

We are seeing spans in APM that have duration longer than hours, even days.
We realized that those are caused by the app going background and staying there for long time.
The span finishes only once the app goes back to foreground and this can result in spans longer than hours.

### What and why?

The problem above happens since the app gets suspended/paused at background and picks up where it left off when it comes back to foreground.
We cannot change this behavior from our SDK, therefore we decided to provide more information so that our users can tell when their network requests take too long from when their apps stay at background for too long.

### How?

We will add `foreground_duration: TimeInterval` and `isBackground: Bool` tags/metrics to `Span`s.
1. `TaskInterception` will be responsible for adding these to the `Span`
2. `AppStateListener` will be responsible for listening to app state along the lifetime of interception and providing `AppStateHistory` data structure
3. `foregroundDuration` and `isBackground` are properties of `AppStateHistory` data structure

~In this PR, we implemented 2nd and 3rd steps. In the next PR, we will implement 1st step by plugging `AppStateListener` into `TaskInterception`~ It's done in this PR as it was not that major.

### Review checklist

### TODO

- [x] `TaskInterception`/`URLSessionTracingHandler` needs new test cases
- [ ] Verify `foreground_duration` is sent as numerical value and can be queried

### DIFF
--- a/Sources/Datadog/Core/System/AppStateListener.swift
+++ b/Sources/Datadog/Core/System/AppStateListener.swift
@@ -1,0 +1,138 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+import class UIKit.UIApplication
+
+/// A data structure to represent recorded app states in a given period of time
+internal struct AppStateHistory: Equatable {
+    /// Snapshot of the app state at `date`
+    struct Snapshot: Equatable {
+        let isActive: Bool
+        let date: Date
+    }
+
+    var initialState: Snapshot
+    var changes = [Snapshot]()
+    var finalDate: Date
+    var finalState: Snapshot {
+        return Snapshot(
+            isActive: (changes.last ?? initialState).isActive,
+            date: finalDate
+        )
+    }
+
+    /// Limits or extrapolates app state history to the given range
+    /// This is useful when you record between 0...3t but you are concerned of t...2t only
+    /// - Parameter range: if outside of `initialState` and `finalState`, it extrapolates; otherwise it limits
+    /// - Returns: a history instance spanning the given range
+    func take(between range: ClosedRange<Date>) -> AppStateHistory {
+        var taken = self
+        // move initial state to lowerBound
+        taken.initialState = Snapshot(
+            isActive: isActive(at: range.lowerBound),
+            date: range.lowerBound
+        )
+        // move final state to upperBound
+        taken.finalDate = range.upperBound
+        // filter changes outside of the range
+        taken.changes = taken.changes.filter { range.contains($0.date) }
+        return taken
+    }
+
+    var foregroundDuration: TimeInterval {
+        var duration: TimeInterval = 0.0
+        var lastActiveStartDate: Date?
+        let allEvents = [initialState] + changes + [finalState]
+        for event in allEvents {
+            if let startDate = lastActiveStartDate {
+                duration += event.date.timeIntervalSince(startDate)
+            }
+            if event.isActive {
+                lastActiveStartDate = event.date
+            } else {
+                lastActiveStartDate = nil
+            }
+        }
+        return duration
+    }
+
+    var didRunInBackground: Bool {
+        return !initialState.isActive || !finalState.isActive
+    }
+
+    private func isActive(at date: Date) -> Bool {
+        if date <= initialState.date {
+            // we assume there was no change before initial state
+            return initialState.isActive
+        } else if finalState.date <= date {
+            // and no change after final state
+            return finalState.isActive
+        }
+        var active = initialState
+        for change in changes {
+            if date < change.date {
+                break
+            }
+            active = change
+        }
+        return active.isActive
+    }
+}
+
+internal protocol AppStateListening: class {
+    var history: AppStateHistory { get }
+}
+
+internal class AppStateListener: AppStateListening {
+    typealias Snapshot = AppStateHistory.Snapshot
+
+    private let dateProvider: DateProvider
+    private let publisher: ValuePublisher<AppStateHistory>
+
+    var history: AppStateHistory {
+        var current = publisher.currentValue
+        current.finalDate = dateProvider.currentDate()
+        return current
+    }
+
+    private static var isAppActive: Bool {
+        return UIApplication.managedShared?.applicationState == .active
+    }
+
+    init(dateProvider: DateProvider) {
+        self.dateProvider = dateProvider
+        let currentState = Snapshot(
+            isActive: AppStateListener.isAppActive,
+            date: dateProvider.currentDate()
+        )
+        self.publisher = ValuePublisher(
+            initialValue: AppStateHistory(
+                initialState: currentState,
+                finalDate: currentState.date
+            )
+        )
+
+        let nc = NotificationCenter.default
+        nc.addObserver(self, selector: #selector(appWillResignActive), name: UIApplication.willResignActiveNotification, object: nil)
+        nc.addObserver(self, selector: #selector(appDidBecomeActive), name: UIApplication.didBecomeActiveNotification, object: nil)
+    }
+
+    @objc
+    private func appWillResignActive() {
+        let now = dateProvider.currentDate()
+        var value = publisher.currentValue
+        value.changes.append(Snapshot(isActive: false, date: now))
+        publisher.publishAsync(value)
+    }
+    @objc
+    private func appDidBecomeActive() {
+        let now = dateProvider.currentDate()
+        var value = publisher.currentValue
+        value.changes.append(Snapshot(isActive: true, date: now))
+        publisher.publishAsync(value)
+    }
+}
--- a/Sources/Datadog/Tracing/AutoInstrumentation/URLSessionTracingHandler.swift
+++ b/Sources/Datadog/Tracing/AutoInstrumentation/URLSessionTracingHandler.swift
@@ -7,6 +7,13 @@
 import Foundation
 
 internal class URLSessionTracingHandler: URLSessionInterceptionHandler {
+    /// Listening to app state changes and use it to report `foreground_duration`
+    let appStateListener: AppStateListening
+
+    init(appStateListener: AppStateListening) {
+        self.appStateListener = appStateListener
+    }
+
     // MARK: - URLSessionInterceptionHandler
 
     func notify_taskInterceptionStarted(interception: TaskInterception) {
@@ -68,6 +75,11 @@ internal class URLSessionTracingHandler: URLSessionInterceptionHandler {
                 }
             }
         }
+        let appStateHistory = appStateListener.history.take(
+            between: resourceMetrics.fetch.start...resourceMetrics.fetch.end
+        )
+        span.setTag(key: DDTags.foregroundDuration, value: appStateHistory.foregroundDuration.toNanoseconds)
+        span.setTag(key: DDTags.isBackground, value: appStateHistory.didRunInBackground)
 
         span.finish(at: resourceMetrics.fetch.end)
     }
--- a/Sources/Datadog/URLSessionAutoInstrumentation/Interception/TaskInterception.swift
+++ b/Sources/Datadog/URLSessionAutoInstrumentation/Interception/TaskInterception.swift
@@ -22,7 +22,10 @@ internal class TaskInterception {
     /// or when the task was created through `URLSession.dataTask(with:url)` on some iOS13+.
     private(set) var spanContext: DDSpanContext?
 
-    init(request: URLRequest, isFirstParty: Bool) {
+    init(
+        request: URLRequest,
+        isFirstParty: Bool
+    ) {
         self.identifier = UUID()
         self.request = request
         self.isFirstPartyRequest = isFirstParty
--- a/Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/Interception/AppStateListenerTests.swift
+++ b/Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/Interception/AppStateListenerTests.swift
@@ -1,0 +1,195 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import XCTest
+@testable import Datadog
+
+class AppStateHistoryTests: XCTestCase {
+    func testForegroundDurationWithoutChanges() {
+        let startDate = Date.mockDecember15th2019At10AMUTC()
+        let history = AppStateHistory(
+            initialState: .init(isActive: true, date: startDate),
+            changes: [],
+            finalDate: startDate + 1.0
+        )
+
+        XCTAssertEqual(history.foregroundDuration, 1.0)
+    }
+
+    func testForegroundDuration() {
+        let startDate = Date.mockDecember15th2019At10AMUTC()
+        let history = AppStateHistory(
+            initialState: .init(isActive: true, date: startDate),
+            changes: [
+                .init(isActive: false, date: startDate + 1.0),
+                .init(isActive: true, date: startDate + 2.0),
+                .init(isActive: false, date: startDate + 3.0),
+                .init(isActive: true, date: startDate + 4.0)
+            ],
+            finalDate: startDate + 5.0
+        )
+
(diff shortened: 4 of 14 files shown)


Sources/Datadog/URLSessionAutoInstrumentation/Interception/AppStateListener.swift
```diff
@@ -0,0 +1,167 @@
...
+        #if canImport(UIKit)
+        return AppStateListener()
+        #else
+        // keep #else block at minimum since it's hard to test
+        return NonUIKitAppStateListener()
+        #endif
+    }
+}
+
+#if canImport(UIKit)
```
ncreated, line 101: Why do we need this? In all other places we use unconditioned `import UIKit`.

Sources/Datadog/URLSessionAutoInstrumentation/Interception/TaskInterception.swift
```diff
@@ -26,6 +26,10 @@ internal class TaskInterception {
         self.identifier = UUID()
         self.request = request
         self.isFirstPartyRequest = isFirstParty
+        // TODO: RUMM-1064
+        // create AppStateListening.build()
```
ncreated, line 30: This would create `NotificationCenter` listener for every single request sent from the app. Isn't it better to share single listener among all requests? The global listener may also be needed for other components in Background Events Monitoring project.

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/Interception/AppStateListenerTests.swift
```diff
@@ -0,0 +1,201 @@
...
+            changes: [
+                .init(isInBackground: false, date: startDate.addingTimeInterval(1.0)),
+                .init(isInBackground: true, date: startDate.addingTimeInterval(2.0))
+            ],
+            finalState: .init(isInBackground: false, date: startDate.addingTimeInterval(3.0))
+        )
+        XCTAssertEqual(history, expected)
+    }
+
+    func testChangingFinalState() {
```
ncreated, line 190: The `AppStateListener` has its lifecycle and behaviours, let's use `Given` / `When` / `Then` or `When` / `Then` verbosity for this tests. It's hard to understand what "changing final state" means.

ncreated (changes requested):
I get the idea, looks good 👍. Left important questions on threading and listener's life-cycle.

Sources/Datadog/URLSessionAutoInstrumentation/Interception/AppStateListener.swift
```diff
@@ -0,0 +1,167 @@
...
+/// A data structure to represent recorded app states in a given period of time
+internal struct AppStateHistory: Equatable {
+    /// Snapshot of the app state at `date`
+    struct Snapshot: Equatable, Comparable {
+        let isInBackground: Bool
+        let date: Date
+
+        static func < (lhs: Snapshot, rhs: Snapshot) -> Bool {
+            return lhs.date < rhs.date
+        }
```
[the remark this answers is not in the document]
buranmert: `Comparable` conformance lets me do `changes.sorted()` without calling site knowing how to sort them.
the problem you described is real (2 snapshots with the same date is ambiguous) but your solution too has the same problem (we still have 2 snapshots with the same date and still ambiguous).

let me know if i'm missing something

Sources/Datadog/URLSessionAutoInstrumentation/Interception/AppStateListener.swift
```diff
@@ -0,0 +1,167 @@
...
+    private static func currentState(with date: Date) -> Snapshot {
+        return Snapshot(isInBackground: AppStateListener.isBackground, date: date)
+    }
+
+    private let dateProvider: DateProvider
+    private var _appStateHistory: AppStateHistory
+    var appStateHistory: AppStateHistory {
+        _appStateHistory.finalState = AppStateListener.currentState(with: dateProvider.currentDate())
+        return _appStateHistory
+    }
```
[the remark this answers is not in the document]
buranmert: good point 👍  
`appStateHistory` shouldn't be called from non-main thread as it calls `UIApplication.shared`, otherwise `Main Thread Checker` gives warnings. i need to ensure that.

i thought `UIApplication.back/foregroundNotifications` are always posted in main thread but i can't find any mention in docs. i need to ensure main thread there too.

Sources/Datadog/URLSessionAutoInstrumentation/Interception/AppStateListener.swift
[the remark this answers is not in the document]
ncreated: I mean: `changes.sorted(by: { $0.date < $1.date })` is idiomatic and reads in-line. With `Comparable` you assume that the reader knows how `Snapshots` can be compared and `changes.sorted()` can't be understood in-line. 

This seems minor anyway.

Sources/Datadog/URLSessionAutoInstrumentation/Interception/TaskInterception.swift
buranmert: the worst case scenario with non-singleton can be ~100 simultaneous listeners and i added such test case, my measurement gives reasonable results (`100 -> 0.046sec`) even in this case.

i didn't prefer a singleton here as it will stay around even while not needed but singleton would work too.
if background monitoring requires a singleton, we can add `static var sharedHistory` and use it in `TaskInterception` and that will be it, shouldn't require major refactoring.

Sources/Datadog/URLSessionAutoInstrumentation/Interception/AppStateListener.swift
```diff
@@ -0,0 +1,167 @@
...
+        }
+        var isInBackground = initialState.isInBackground
+        for change in changes {
+            if date < change.date {
+                break
+            }
+            isInBackground = change.isInBackground
+        }
+        return isInBackground
+    }
```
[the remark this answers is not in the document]
buranmert: your suggestion finds the first change after the date ❌ 
this `for` loop finds the last change before the date ✅  

(i had made the change but then hopefully a unit test failed 😅 )

Sources/Datadog/Core/System/AppStateListener.swift
```diff
@@ -0,0 +1,140 @@
...
+        self.dateProvider = dateProvider
+        let currentState = AppStateListener.currentState(with: dateProvider.currentDate())
+        _appStateHistory = AppStateHistory(
+            initialState: currentState,
+            changes: [],
+            finalDate: currentState.date
+        )
+        let nc = NotificationCenter.default
+        nc.addObserver(self, selector: #selector(appWillResignActive), name: UIApplication.willResignActiveNotification, object: nil)
+        nc.addObserver(self, selector: #selector(appDidBecomeActive), name: UIApplication.didBecomeActiveNotification, object: nil)
```
buranmert, line 125: `didEnterBackground/willEnterForeground` pair wasn't covering all the states, leaving `.inactive` state out.
now i use `willResignActive/didBecomeActive` instead, this gives more accurate `foreground/active duration` information as it properly measures `active` state.

Sources/Datadog/URLSessionAutoInstrumentation/Interception/TaskInterception.swift
buranmert: i ensured main thread for getting current app state due to Main Thread Checker warning but now every `AppStateListener(...)` call leads to `main.sync { ... }`.
a singleton can solve this issue. i'm making the change.

Sources/Datadog/Tracing/AutoInstrumentation/URLSessionTracingHandler.swift
```diff
@@ -68,6 +68,12 @@ internal class URLSessionTracingHandler: URLSessionInterceptionHandler {
                 }
             }
         }
+        let appStateHistory = interception.appStateListener.history.take(
+            between: resourceMetrics.fetch.start...resourceMetrics.fetch.end
+        )
+        // TODO: RUMM-1064 tag doesn't seem like the best fit but can't find anything better
+        span.setTag(key: "foreground_duration", value: "\(appStateHistory.foregroundDuration)")
+        span.setTag(key: "is_background", value: "\(appStateHistory.didRunInBackground)")
```
buranmert, line 76: 1. should they be declared in `DDTags`? (unlike other `DDTags` these are not officially supported by Datadog)
2. should they be added as `tag`s? (i couldn't think of any other place for these)
cc @ncreated

Sources/Datadog/Core/System/AppStateListener.swift
```diff
@@ -0,0 +1,150 @@
...
+    private func appWillResignActive() {
+        objc_sync_enter(self)
+        _appStateHistory.add(change: Snapshot(isActive: false, date: dateProvider.currentDate()))
+        objc_sync_exit(self)
+    }
+    @objc
+    private func appDidBecomeActive() {
+        objc_sync_enter(self)
+        _appStateHistory.add(change: Snapshot(isActive: true, date: dateProvider.currentDate()))
+        objc_sync_exit(self)
```
ncreated, line 148: This will block the main thread (the _writer thread_) if any other thread (any _reader thread_) is accessing the `_appStateHistory`. If there are 100 simultaneous reads, this write operation will block the app UI until all of them complete 😞.

Sources/Datadog/Core/System/AppStateListener.swift
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    private static var isAppActive: Bool {
+        if Thread.isMainThread {
+            return UIApplication.managedShared?.applicationState == .active
+        } else {
+            return DispatchQueue.main.sync {
+                UIApplication.managedShared?.applicationState == .active
+            }
+        }
```
ncreated, line 121: This if-check against the current thread looks odd. We can get rid of this if controlling the `AppStateListener` initialization and submitting this read asynchronously to the main thread.

ncreated (changes requested):
The `AppStateHistory` looks good 👍, but the threading model proposed for `AppStateListener` is incorrect.

W need to solve a classic readers-writers problem with:
* occasional writes (when the app transitions between background and foreground)
* intensive reads (when a network request completes).

The model implemented in this PR doesn't leverage those conditions and it treats writes and reads equally. Because of writes being always performed from the main thread, **the main thread receives a penalty** from reads happening on background thread.

Given above conditions, the optimal threading model should work this way:
* occasional writes should be not blocking (no locks executed on the main thread),
* intensive reads should be concurrent (allowing simultaneous reads).

This can be achieved with a concurrent `DispatchQueue` (simultaneous reads) and `.barrier` API (for exclusive writes).

This problem was solved in https://github.com/DataDog/dd-sdk-ios/pull/422 where we introduced `ValuePublisher` type. We can leverage it this way:
```swift
let appStateHistory = ValuePublisher<AppStateHistory>()

// From the main thread, non-blocking async write:
appStateHistory.publishAsync(newAppStateHistory)

// From any background thread, concurrent read:
_ = appStateHistory.currentValue
```

Sources/Datadog/Core/System/AppStateListener.swift
```diff
@@ -0,0 +1,148 @@
...
+        nc.addObserver(self, selector: #selector(appWillResignActive), name: UIApplication.willResignActiveNotification, object: nil)
+        nc.addObserver(self, selector: #selector(appDidBecomeActive), name: UIApplication.didBecomeActiveNotification, object: nil)
+    }
+
+    @objc
+    private func appWillResignActive() {
+        let now = dateProvider.currentDate()
+        publisher.mutateAsync {
+            $0.add(change: Snapshot(isActive: false, date: now))
+        }
```
ncreated, line 139: Do we need to introduce `mutateAsync` and use `inout`? It seems over-optimizing it by using complex machinery (mutating value type by reference between threads). I'm not that familiar with consequences of using `inout` across threads and can't see if we synchronize it well.

We could keep it plain simple:
```swift
var value = publisher.currentValue // has negligible overhead as reads are concurrent
value.add(change: Snapshot(isActive: false, date: now))
publisher.publishAsync(value)
```
or even remove the read overhead at all by leveraging the fact that notifications are delivered on a single thread:
```swift
private lastHistory: AppStateHistory {
   didSet { publisher.publishAsync(lastHistory) }
}

private let publisher: ValuePublisher<AppStateHistory>

private func appWillResignActive() {
   var newHistory = self.lastHistory
   newHistory.add(change: Snapshot(isActive: false, date: dateProvider.currentDate()))
   self.lastHistory = newHistory
}
```

Sources/Datadog/Core/System/AppStateListener.swift
```diff
@@ -0,0 +1,148 @@
...
+    var finalState: Snapshot {
+        return Snapshot(
+            isActive: (changes.last ?? initialState).isActive,
+            date: finalDate
+        )
+    }
+
+    mutating func add(change: Snapshot) {
+        changes.append(change)
+        changes.sort()
```
ncreated, line 35: Why does it require sorting? Is it possible that changes will be added in the random-time order?

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/Interception/AppStateListenerTests.swift
```diff
@@ -0,0 +1,197 @@
...
+                let nc = NotificationCenter.default
+                nc.post(
+                    name: (Bool.random() ?
+                            UIApplication.willResignActiveNotification :
+                            UIApplication.didBecomeActiveNotification),
+                    object: nil
+                )
+            }
+            // read
+            XCTAssertFalse(listener.history.changes.isEmpty)
```
ncreated, line 194: We can use existing [`callConcurrently {}` helper](https://github.com/DataDog/dd-sdk-ios/blob/36c57c0ea5e9323be92b4a87d903350aa33536cb/Tests/DatadogTests/Helpers/XCTestCase.swift#L27-L37) to mix concurrent writes and reads uniformly:

```swift
// swiftlint:disable opening_brace
callConcurrently(
   closures: [
      { /* write (post notification) */ }
      { /* read (access value) */ }
   ],
   iterations: 1_000
)
// swiftlint:enable opening_brace
```

This should give us approximately 50% of overlap between writes and reads. Current solution is only overlapping 10% (1K out of 10K).

buranmert: as you said before, the real use case is write-rarely&read-often.
this test case is closer to the real use case.

Sources/Datadog/Tracing/AutoInstrumentation/URLSessionTracingHandler.swift
buranmert: i removed string conversion, now it's
```swift
span.setTag(key: "foreground_duration", value: appStateHistory.foregroundDuration.toNanoseconds)
```
with this, you can create a `measure` (open the trace, click on `foreground_duration` tag, select add facet, select `measure` tab at the top) and filter traces longer/shorter than your threshold

NOTE: once you do that, it takes some time (a few minutes maybe) before you can use it in your query. at first i thought it wasn't working but it actually works 👌 

example-query

Sources/Datadog/Tracing/AutoInstrumentation/URLSessionTracingHandler.swift
```diff
@@ -68,6 +75,12 @@ internal class URLSessionTracingHandler: URLSessionInterceptionHandler {
                 }
             }
         }
+        let appStateHistory = appStateListener.history.take(
+            between: resourceMetrics.fetch.start...resourceMetrics.fetch.end
+        )
+        // TODO: RUMM-1064 tag doesn't seem like the best fit but can't find anything better
```
ncreated, line 81: Seems solved, so:
```suggestion
```

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/Interception/TaskInterceptionTests.swift
```diff
@@ -30,6 +30,14 @@ class TaskInterceptionTests: XCTestCase {
...
         XCTAssertTrue(interception.isDone)
     }
+
+//    func testWhenInterceptionIsCreated_itHasSharedAppStateListener() {
+//        // When
+//        let interception = TaskInterception(request: .mockAny(), isFirstParty: true)
+//
+//        // Then
+//        XCTAssert(interception.appStateListener === AppStateListener.shared)
+//    }
```
ncreated, line 40: Some leftovers, 🧽:
```suggestion
```

ncreated (approved):
Looks good 👌, except few comments to clean up.

PR #439, opened 2021-03-10
RUMM-1145 Carthage XCFrameworks support

buranmert:
### Problem

The main thing of `carthage` is building projects for every possible arch-platform so that they don't need to be rebuilt when the developer changes the target arch-platform.
It was achieving this by `lipo`ing `BuildProducts` for different arch-platforms and giving one single `.framework` containing multiple slices.

This model stopped working as `arm64-simulator` arch-platform came along with Apple Silicon; it duplicated `arm64-ios` and `lipo` doesn't merge duplicate slices.

### Apple's solution

Apple introduced `XCFramework`s to solve this problem: this format can contain multiple `.framework`s for different arch-platforms. 
```
Example.xcframework:
- ios-arm64_armv7/Example.framework
- ios-arm64_i386_x86_64-simulator/Example.framework
- macos-arm64_x86_64/Example.framework
- tvos-arm64/Example.framework
- tvos-arm64_x86_64-simulator/Example.framework
Info.plist
```

### Carthage's solution

`carthage` started supporting this new format starting from `0.37`.
`carthage` also maintains backward compatibility by not breaking projects that are configured for `.framework` dependencies.
For some reason (this looks like a bug in `carthage` to me), `dd-sdk-ios` works only in this backward-compatibility mode although it is configured with `.xcframework`

#### Backward compatibility

If the target has `FRAMEWORK_SEARCH_PATH` set to `Carthage/Build/`, `carthage` assumes that it searches for `Example.framework` whereas there is `Example.xcframework` instead.
[In that case, `carthage` extracts the content of `Example.xcframework` and inject its location as an extra `FRAMEWORK_SEARCH_PATH` to `xcodebuild`](https://github.com/Carthage/Carthage/blob/0668de43eb5d323d2e816eaab83677f50a8eeb24/Source/CarthageKit/Xcode.swift#L840)

```swift
guard let platformTripleOS = settings.platformTripleOS.value,
  let frameworkSearchPaths = settings.frameworkSearchPaths.value,
  frameworkSearchPaths.contains(where: isRelativeToBuildDirectory) else {
  // don't extract
  return
}
extractXCFramework()
//
```

Setting such a `FRAMEWORK_SEARCH_PATH` value triggers XCFramework extraction and compilation succeeds ✅ 

ℹ️ Note that this build setting is not supposed to be needed for compilation. 
ℹ️ Based on my trials, always the second target fails at building for `iphonesimulator`

#### Build settings of `Kronos`

`Kronos` had `VALID_ARCHS` in its base xcconfig file and it was not updated after the introduction of `arm64-simulator`.
I fixed it with https://github.com/MobileNativeFoundation/Kronos/pull/76
I also asked for a hotfix version and I'm waiting for a response.

#### If `Kronos` ships a hotfix...

We can update to this version and `dd-sdk-ios` can be used by Apple Silicon owner `carthage` users ✅ 
Sooner or later, we will need to support this arch-platform.

### Review checklist

- [ ] If `Kronos` ships a hotfix, remove the content of [`Base.xcconfig`](https://github.com/DataDog/dd-sdk-ios/blob/master/xcconfigs/Base.xcconfig#L5)

### DIFF
--- a/Datadog/Datadog.xcodeproj/project.pbxproj
+++ b/Datadog/Datadog.xcodeproj/project.pbxproj
@@ -183,9 +183,6 @@
 		614E9EB3244719FA007EE3E1 /* BundleType.swift in Sources */ = {isa = PBXBuildFile; fileRef = 614E9EB2244719FA007EE3E1 /* BundleType.swift */; };
 		6152C83E24BE1C91006A1679 /* HTTPServerMock in Frameworks */ = {isa = PBXBuildFile; productRef = 6152C83D24BE1C91006A1679 /* HTTPServerMock */; };
 		6152C84024BE1CC8006A1679 /* DataUploaderBenchmarkTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6152C83F24BE1CC8006A1679 /* DataUploaderBenchmarkTests.swift */; };
-		61569795256CF6D400C6AADA /* Kronos.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 61569794256CF6D400C6AADA /* Kronos.framework */; };
-		615697E3256CFB4700C6AADA /* Kronos.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 61569794256CF6D400C6AADA /* Kronos.framework */; };
-		615697E4256CFB4700C6AADA /* Kronos.framework in ⚙️ Embed Framework Dependencies */ = {isa = PBXBuildFile; fileRef = 61569794256CF6D400C6AADA /* Kronos.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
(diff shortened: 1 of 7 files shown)


buranmert:
there is still something that i don't understand:
`Datadog` target links against `Kronos.xcframework` (not `Kronos.framework`)
i don't understand why it still requires `extractXCFrameworks`
once there is `Carthage/Build/Kronos.xcframework` in place, it builds without problem from Xcode

ncreated (approved):
Looks good 👌. I'm assuming something extra would have to be done for the `crash-reporting` branch once it's merged, but that's another work. Let's ship this as a hotfix?

buranmert:
for the records:
this PR is a workaround, `carthage` is supposed to succeed without `FRAMEWORK_SEARCH_PATHS`
we can revert this workaround once there is news from https://github.com/Carthage/Carthage/issues/3145

PR #441, opened 2021-03-11
RUMM-1128 δ Add Internal Monitoring feature

ncreated:
### What and why?

📺 🌟  This PR enables internal monitoring of the SDK performance on end-devices. It adds an opt-in feature sending internal logs to our Datadog org under the `service:dd-sdk-ios`. The feature is hidden behind the compiler flag and requires `clientToken` configured for our org to be set in:
```swift
#if DD_SDK_ENABLE_INTERNAL_MONITORING

/// Enables the internal monitoring feature.
///
/// This feature provides observability for internal events happening in the SDK. All data collected by this feature
/// is sent to Datadog org, not to the customer's org. This feature is opt-in and requires specific configuration to be enabled.
/// It is never enabled by default. We do not collect any internal monitoring data for those who didn't explicitly opt-in for that by contacting Datadog.
///
/// - Parameter clientToken: the client token authorizing monitoring data to be uploaded to Datadog's org.
public func enableInternalMonitoring(clientToken: String) -> Builder

#endif
```

Internal logs can be queried using `service:dd-sdk-ios` and filtered to `logger.version:1.5.0`:

Screenshot 2021-03-11 at 20 21 58

### How?

I added a new feature: `InternalMonitoringFeature`. It is no different than other features (`LoggingFeature`, `TracingFeature`, `RUMFeature` and `CrashReportingFeature`). It has its `FeatureStorage` and `FeatureUpload` configured to batch data and upload it using the client token configured in public API `enableInternalMonitoring(clientToken:)`.

Internal logs can be collected using:
```swift
InternalMonitoringFeature.instance?.monitor.sdkLogger
```
which is just a regular `Logger` from the logging feature.

By reusing core components, we get thread safety, tracking consent conformance and data consistency for free 💪.

### Current instrumentation

Current instrumentation includes few errors and warnings tracked from different core components (`FileWriter`, `FilesOrchestrator`, `DataUploader`, etc.). This will be expanded to more places in the future.

### `InternalMonitor` injection

Instead of referencing global `InternalMonitoringFeature.instance?.monitor.sdkLogger` (directly or through helper), its reference is passed as optional dependency to all core components. This is to avoid self-instrumentation of the `InternalMonitoringFeature` which also uses the instrumented components to upload data.

### `DD_SDK_ENABLE_INTERNAL_MONITORING ` flag

To make this new API visible to `Example` app and other targets (unit, benchmark and integration tests), the `DD_SDK_ENABLE_INTERNAL_MONITORING` compiler flag must be provided. To avoid setting it in `Datadog.framework's` build setting (and exposing it in Carthage builds), I created local, git-ignored `.xcconfig` file: `xcconfigs/Base.local.xcconfig`. It can be generated by running `make dependencies` and is it from now required to work on the SDK.

### Review checklist


Sources/Datadog/InternalMonitoring/InternalMonitoringFeature.swift
```diff
@@ -0,0 +1,147 @@
...
+internal func obtainInternalMonitoringFeatureLogDirectories() throws -> FeatureDirectories {
+    let version = "v1"
+    return FeatureDirectories(
+        unauthorized: try Directory(withSubdirectoryPath: "com.datadoghq.im-logs/intermediate-\(version)"),
+        authorized: try Directory(withSubdirectoryPath: "com.datadoghq.im-logs/\(version)")
+    )
+}
+
+///// Internal Monitoring `Logger`, sending SDK monitoring data to Datadog org.
+//internal var developerLogger: Logger? { InternalMonitoringFeature.instance?.monitor.sdkLogger }
```
ncreated, line 19: 🧹
```suggestion
```

Tests/DatadogTests/Datadog/Core/FeaturesConfigurationTests.swift
```diff
@@ -467,6 +467,39 @@ class FeaturesConfigurationTests: XCTestCase {
...
+                bundleName: "AppName"
+            )
+        )
+
+        // Then
+        let configuration = try XCTUnwrap(featuresConfiguration.internalMonitoring)
+        XCTAssertEqual(configuration.common, featuresConfiguration.common)
+        XCTAssertEqual(configuration.sdkServiceName, "dd-sdk-ios", "Internal monitoring data should be available under \"service:dd-sdk-ios\"")
+        XCTAssertEqual(configuration.sdkEnvironment, "prod", "Internal monitoring data should be available under \"env:prod\"")
+        XCTAssertEqual(configuration.sdkServiceName, "dd-sdk-ios", "Internal monitoring data should be available under \"service:dd-sdk-ios\"")
```
ncreated, line 496: 🧹
```suggestion
```

buranmert (commented):
this is going to be super helpful for us 💯 

code looks good overall but it's not clear to me how to use it in a project.
i think it should be documented (at least for our internal support) how to use this via cocoapods, carthage and spm.

ncreated:
> this is going to be super helpful for us 💯
> 
> code looks good overall but it's not clear to me how to use it in a project.
> i think it should be documented (at least for our internal support) how to use this via cocoapods, carthage and spm.

👍 I've enhanced the API doc with giving more info on the compiler setting that enables this feature.

Issue #438, opened 2021-03-09
[Question] Manually tracing resources


sdejesusF:
Hi 👋 , 

We are using Xamarin.iOS and we don't have direct access to NSURLSession therefore, we have to manually intercept requests and add them as resources using StartResource and StopResource (or WithError). My question would be if it's possible to add in between the start and stop other timings like for example TTFB, or any other custom time that would be digested as timings for that resource.

ncreated:
Hello @sdejesusF 👋, thanks for reaching us. Resource timings (TTFB, SSL, connection etc.) can be collected by using:
```swift
Global.rum.addResourceMetrics(resourceKey:metrics:)
```
This methods accepts `URLSessionTaskMetrics` for the `metrics` parameter. This method should be called after resource is started (`startResource()`) and between it gets completed (`stopResource()` or with error). Datadog SDK will do the job of computing timing metrics for you.

We don't yet support Xamarin.iOS, so I'm not sure if you'll be able to capture `URLSessionTaskMetrics` from [`URLSessionTaskDelegate`](https://developer.apple.com/documentation/foundation/urlsessiontaskdelegate/1643148-urlsession). You may want to contact our support and raise the feature request for Xamarin support, which will help us prioritising our next supported platforms.

sdejesusF:
Hi  👋  @ncreated, 

Yes, sadly we don't have access to URLSessionTaskMetrics :( so for now we will continue to manually track the resources without having TTFB.

And sorry to bother you again with another related topic to this question. On dd-android, we are able to pass the size of the request on the StopResource method. But I don't see that possibility on iOS. Can we use something else to add the size of the resource or is not possible?

sdejesusF:
btw, just realize that its in here https://github.com/DataDog/dd-sdk-ios/blob/c2b4e6445f87f3c908782881ef253352e0bbc248/Sources/Datadog/DDRUMMonitor.swift#L173 but its not exposed to objective c

ncreated:
Hey @sdejesusF 👋. Thanks for spotting this! We will expose this API to Objc shortly and this should be available in the next release.

sdejesusF:
thank you @ncreated 🙏

ncreated:
Available in [`1.5.2`](https://github.com/DataDog/dd-sdk-ios/releases/tag/1.5.2) 👌

PR #447, opened 2021-03-22
RUMM-1203 Expose missing RUM Monitor APIs to Objc

ncreated:
### What and why?

📦 This PR exposes missing resource `size` attribute to RUM Objc. 

Solves #438 

### How?

I've reviewed all 20 `RUMMonitor` public APIs, making sure they are aligned between Swift and Objc:
* added `size` parameter to two resource APIs,
* made `statusCode` optional parameter in one API.

Ran `make api-surface` after all.

### Review checklist


xgouchet (approved):

Sources/DatadogObjc/RUMMonitor+objc.swift
```diff
@@ -257,19 +257,27 @@ public class DDRUMMonitor: NSObject {
...
     ) {
-        swiftRUMMonitor.stopResourceLoading(resourceKey: resourceKey, response: response, attributes: castAttributesToSwift(attributes))
+        swiftRUMMonitor.stopResourceLoading(resourceKey: resourceKey, response: response, size: size?.int64Value, attributes: castAttributesToSwift(attributes))
     }
 
     @objc
     public func stopResourceLoading(
         resourceKey: String,
-        statusCode: Int,
+        statusCode: NSNumber?,
```
buranmert, line 269: why do we need this change?

buranmert (approved):

0xnm:
It seems that this PR has wrong RUMM ticket reference.

Sources/DatadogObjc/RUMMonitor+objc.swift
ncreated: `statusCode` is an optional parameter to our `RUMMonitor` API and because Objc doesn't have in-build optional type, this must be mapped to a null-able reference type (`NSNumber`).

PR #451, opened 2021-03-26
RUMM-1160 Static pod support

buranmert:
### What and why?

`DatadogSDK.podspec` requires `!use_frameworks` in client's `Podfile`
Lack of it results in error at `pod install`

### How?: A proposal

For cocoapods and carthage, I can mix Swift and ObjC by importing ObjC headers in framework's umbrella header [as Apple suggests](https://developer.apple.com/documentation/swift/imported_c_and_objective-c_apis/importing_objective-c_into_swift)

For Swift Package Manager, this is not possible as mixed targets are not supported. 
To solve this, I pass `SPM_BUILD` flag from `Package.swift` where we create `_Datadog_Private` target.
Now I can `#if SPM_BUILD import _Datadog_Private #endif` in the source code.

### Review checklist

- [x] Rename ObjC methods so that clients will notice that they are not for public use


buranmert:
ℹ️ CI passed with `use_frameworks`
🗣️ now i'm pushing a commit to remove `use_frameworks`

Package.resolved
```diff
@@ -0,0 +1,25 @@
+{
```
ncreated, line 1: This file should be git-ignored 🧽

ncreated (changes requested):
Looks good 👌 💪 , let's just remove the `Package.resolved`.

ncreated (approved):

PR #456, opened 2021-03-31
RUMM-1213 Allow additional configuration when initializing the SDK

xgouchet:
### What and why?

Let customer set additional configuration as a Dictionary. This will be used to configure internal settings of the SDK

### DIFF
--- a/Sources/DatadogObjc/DatadogConfiguration+objc.swift
+++ b/Sources/DatadogObjc/DatadogConfiguration+objc.swift
@@ -278,6 +278,11 @@ public class DDConfigurationBuilder: NSObject {
     }
 
     @objc
+    public func set(additionalConfiguration: [String: Any]) {
+        _ = sdkBuilder.set(additionalConfiguration: castAttributesToSwift(additionalConfiguration))
+    }
+
+    @objc
     public func build() -> DDConfiguration {
         return DDConfiguration(sdkConfiguration: sdkBuilder.build())
     }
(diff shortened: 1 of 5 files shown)


Sources/DatadogObjc/DatadogConfiguration+objc.swift
```diff
@@ -277,6 +277,11 @@ public class DDConfigurationBuilder: NSObject {
         _ = sdkBuilder.set(uploadFrequency: uploadFrequency.swiftType)
     }
 
+    @objc
+    public func set(additionalConfiguration: [String: Any]) {
+        _ = sdkBuilder.set(additionalConfiguration: castAttributesToSwift(additionalConfiguration))
```
ncreated, line 282: We shouldn't and we can't use `castAttributesToSwift()` in this case. The `castAttributesToSwift()` wraps `Any` value into `AnyEncodable` wrapper - so it changes its identity. That said, if we pass `"_dd.foo": "bar"` from Obj-c, the downstream assumption in Swift SDK will fail:
```swift
// This will fail, as `additionalConfigurations["_dd.foo"]` is `AnyEncodable`, not `String`:
let bar = additionalConfigurations["_dd.foo"] as? String
```

Instead, we should just pass the `additionalConfiguration` as it is:
```swift
_ = sdkBuilder.set(additionalConfiguration: additionalConfiguration)
```
so the above assumption can pass.

The `castAttributesToSwift()` is only used for capturing serializable values (log / span / rum attributes). In their case, we don't make the `as?` assumption on their identity - we just encode them using encoding implementation from `AnyEncodable`.

ncreated (changes requested):
Nice 👌. I left one important change request.

Additionally, we could pass this value to the `FeaturesConfiguration.Common` so it will become visible to all features. Or we can do it later, when implementing the first additional attribute.

ncreated (approved):
👌

PR #457, opened 2021-03-31
RUMM-1229 δ Collect internal telemetry for Crash Reporting while dogfooding

ncreated:
### What and why?

📦 This PR improves internal monitoring data send to Datadog while dogfooding.

### How?

There are two things improved:
* **Got rid of _"Failed to read previously used writable file"_ false-positive warning.** It was reported each time when the last writable file is accessed after successful upload - in such case the file is missing (as it's deleted by the upload worker) which is expected behaviour and should not be reported as warning.

Screenshot 2021-03-31 at 13 58 38

* **Collect Crash Reports telemetry for further analysis.** I've added some error logs in crash reporting feature and included `DEBUG` logs for shipping useful diagnostic info about the crash reports:

Screenshot 2021-03-31 at 13 56 48

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/Persistence/Files/Directory.swift
+++ b/Sources/Datadog/Core/Persistence/Files/Directory.swift
@@ -28,10 +28,16 @@ internal struct Directory {
         return File(url: fileURL)
     }
 
-    /// Returns file with given name.
+    /// Checks if a file with given `fileName` exists in this directory.
+    func hasFile(named fileName: String) -> Bool {
+        let fileURL = url.appendingPathComponent(fileName, isDirectory: false)
+        return FileManager.default.fileExists(atPath: fileURL.path)
+    }
+
+    /// Returns file with given name or throws an error if file does not exist.
     func file(named fileName: String) throws -> File {
         let fileURL = url.appendingPathComponent(fileName, isDirectory: false)
-        guard FileManager.default.fileExists(atPath: fileURL.path) else {
+        guard hasFile(named: fileName) else {
             throw InternalError(description: "File does not exist at path: \(fileURL.path)")
         }
         return File(url: fileURL)
--- a/Sources/Datadog/Core/Persistence/FilesOrchestrator.swift
+++ b/Sources/Datadog/Core/Persistence/FilesOrchestrator.swift
@@ -61,6 +61,10 @@ internal class FilesOrchestrator {
 
     private func reuseLastWritableFileIfPossible(writeSize: UInt64) -> WritableFile? {
         if let lastFileName = lastWritableFileName {
+            if !directory.hasFile(named: lastFileName) {
+                return nil // this is expected if the last writable file was deleted after upload
+            }
+
             do {
                 let lastFile = try directory.file(named: lastFileName)
                 let lastFileCreationDate = fileCreationDateFrom(fileName: lastFile.name)
@@ -74,7 +78,7 @@ internal class FilesOrchestrator {
                     return lastFile
                 }
             } catch {
-                internalMonitor?.sdkLogger.warn("Failed to read previously used writable file", error: error)
+                internalMonitor?.sdkLogger.warn("Failed to reuse last writable file", error: error)
             }
         }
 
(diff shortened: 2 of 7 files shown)


xgouchet (approved):

Sources/Datadog/Core/Persistence/Files/Directory.swift
```diff
@@ -28,10 +28,16 @@ internal struct Directory {
...
+        let fileURL = url.appendingPathComponent(fileName, isDirectory: false)
+        return FileManager.default.fileExists(atPath: fileURL.path)
+    }
+
+    /// Returns file with given name or throws an error if file does not exist.
     func file(named fileName: String) throws -> File {
         let fileURL = url.appendingPathComponent(fileName, isDirectory: false)
-        guard FileManager.default.fileExists(atPath: fileURL.path) else {
+        guard hasFile(named: fileName) else {
             throw InternalError(description: "File does not exist at path: \(fileURL.path)")
```
buranmert, line 41: maybe this is not an error state.? does it make sense if this method returns an optional in case that file doesn't exist?
i'll leave another comment at calling site below

Sources/Datadog/Core/Persistence/FilesOrchestrator.swift
```diff
@@ -61,6 +61,10 @@ internal class FilesOrchestrator {
 
     private func reuseLastWritableFileIfPossible(writeSize: UInt64) -> WritableFile? {
         if let lastFileName = lastWritableFileName {
+            if !directory.hasFile(named: lastFileName) {
+                return nil // this is expected if the last writable file was deleted after upload
+            }
+
             do {
                 let lastFile = try directory.file(named: lastFileName)
```
buranmert, line 69: we can merge those 2 into `guard let lastFile = directory.file(named:) else { return nil }`
that's not a major issue but currently we catch `try directory.file(named:)` and `try lastFile.size()` in the same block, so we can't know which error happened when

does that make sense?

Sources/DatadogCrashReporting/PLCrashReporterIntegration/PLCrashReporterIntegration.swift
```diff
@@ -42,3 +48,53 @@ internal final class PLCrashReporterIntegration: ThirdPartyCrashReporter {
         try crashReporter.purgePendingCrashReportAndReturnError()
     }
 }
+
+#if DD_SDK_ENABLE_INTERNAL_MONITORING
+/// Diagnositc information about the crash report, collected for `DatadogCrashReporting` observability.
```
buranmert, line 53: `Diagnositc` -> `Diagnostic`

Sources/Datadog/Core/Persistence/Files/Directory.swift
ncreated: This is a core logic in our data persistence and IMO the strickier we are, the better. We know exactly when the file is supposed to exist and when it should / could be gone. By throwing the error here, we can control data consistency more and base less on assumptions. How does it sound?

Sources/Datadog/CrashReporting/DDCrashReportingPluginType.swift
```diff
@@ -20,6 +20,13 @@ public class DDCrashReport: NSObject {
     /// The last context injected through `inject(context:)`
     internal let context: Data?
 
+#if DD_SDK_ENABLE_INTERNAL_MONITORING
+    /// Additional diagnositc information about the crash report, collected for `DatadogCrashReporting` observability.
```
ncreated, line 24:
```suggestion
    /// Additional diagnostic information about the crash report, collected for `DatadogCrashReporting` observability.
```

Sources/DatadogCrashReporting/PLCrashReporterIntegration/PLCrashReporterIntegration.swift
ncreated:
```suggestion
/// Diagnostic information about the crash report, collected for `DatadogCrashReporting` observability.
```

buranmert (approved):

PR #459, opened 2021-03-31
RUMM-1038 add option to customize the `source` tag

xgouchet:
### What and why?

This is based on #456 , and let all feature use a customized `source` tag instead of the default value (`ios`).

### How?

We read the `_dd.source` tag from the `additionalConfiguration` dictionary and inject it:
 - in the `ddsource()` method when uploading log-based batches (Logs, RUM, InternalLogs)
 - in the `SpanBuilder` class when converting `DDSpan` to `Span` before the encoding step (APM)

### DIFF
--- a/Sources/Datadog/Core/FeaturesConfiguration.swift
+++ b/Sources/Datadog/Core/FeaturesConfiguration.swift
@@ -18,6 +18,7 @@ internal struct FeaturesConfiguration {
         let serviceName: String
         let environment: String
         let performance: PerformancePreset
+        let source: String
     }
 
     struct Logging {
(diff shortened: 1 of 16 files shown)


Sources/DatadogObjc/DatadogConfiguration+objc.swift
```diff
@@ -277,6 +277,11 @@ public class DDConfigurationBuilder: NSObject {
         _ = sdkBuilder.set(uploadFrequency: uploadFrequency.swiftType)
     }
 
+    @objc
+    public func set(additionalConfiguration: [String: Any]) {
+        _ = sdkBuilder.set(additionalConfiguration: castAttributesToSwift(additionalConfiguration))
```
ncreated, line 282: Mind the note on this in #456 ⚠️

ncreated (approved):
💯 correct, LGTM.

ncreated:
To fix the CI: compile the `DatadogBenchmarkTests` scheme and it will guide you on the change.

Sources/DatadogObjc/DatadogConfiguration+objc.swift
xgouchet: Yes it'll be solved after my rebase

xgouchet:
@buranmert 
> do we plan to add other additional configurations in the future?

Yes that's the plan to have something that's scalable for the future. 
Also although the `UserDefault` option does work in iOS (afaik), there's no equivalent that we might use safely on Android, and we need to make this a global feature across platforms.

ncreated (approved):

buranmert (approved):

PR #461, opened 2021-04-01
RUMM-1234 DDNSURLSessionDelegate build errors fixed

buranmert:
### What and why?

In order to ship standalone binaries of the SDK, we need to build our targets with `BUILD_LIBRARY_FOR_DISTRIBUTION` build flag. 
Yet, `@objc open DDNSURLSessionDelegate: DDURLSessionDelegate` line was giving compiler errors: 
either we needed to remove `@objc` or superclass `DDURLSessionDelegate`

### How?

In this PR, we removed superclass of `DDNSURLSessionDelegate` and needed to make some other changes to make auto-network-tracing work.

### Review checklist

### DIFF
--- a/Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
+++ b/Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
@@ -6,15 +6,14 @@
 
 import Foundation
 
-/// In case that you are not able to use `DDURLSessionDelegate` or its subclasses in your `URLSession` instance,
-/// you can use `DDURLSessionDelegateProviding` protocol.
-/// Implement this protocol in your `URLSessionDelegate` class and forward required API calls to `DDURLSessionDelegate`.
+/// An interface for forwarding `URLSessionDelegate` calls to `DDURLSessionDelegate`.
+/// The implementation must ensure that required methods are called on the `ddURLSessionDelegate`.
 @objc
-public protocol URLSessionDelegateProviding: URLSessionDelegate {
+public protocol __URLSessionDelegateProviding: URLSessionDelegate {
     /// Datadog delegate object.
     /// The class implementing `DDURLSessionDelegateProviding` must ensure that following method calls are forwarded to `ddURLSessionDelegate`:
     // - `func urlSession(_:task:didFinishCollecting:)`
-    // - `func urlSession(_:task: didCompleteWithError:)`
+    // - `func urlSession(_:task:didCompleteWithError:)`
     var ddURLSessionDelegate: DDURLSessionDelegate { get }
 }
 
@@ -23,7 +22,7 @@ public protocol URLSessionDelegateProviding: URLSessionDelegate {
 ///
 /// All requests made with the `URLSession` instrumented with this delegate will be intercepted by the SDK.
 @objc
-open class DDURLSessionDelegate: NSObject, URLSessionTaskDelegate, URLSessionDelegateProviding {
+open class DDURLSessionDelegate: NSObject, URLSessionTaskDelegate, __URLSessionDelegateProviding {
     public var ddURLSessionDelegate: DDURLSessionDelegate {
         return self
     }
--- a/Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
+++ b/Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
@@ -6,12 +6,28 @@
 
 import Foundation
 
+/// In case that you are not able to use `DDURLSessionDelegate` or its subclasses in your `URLSession` instance,
+/// you can use `DDURLSessionDelegateProviding` protocol.
+/// Implement this protocol in your `URLSessionDelegate` class and forward required API calls to `DDURLSessionDelegate`.
+@objc
+public protocol URLSessionDelegateProviding: URLSessionDelegate {
+    /// Datadog delegate object.
+    /// The class implementing `DDURLSessionDelegateProviding` must ensure that following method calls are forwarded to `ddURLSessionDelegate`:
+    // - `func urlSession(_:task:didFinishCollecting:)`
+    // - `func urlSession(_:task: didCompleteWithError:)`
+    var ddURLSessionDelegate: DDURLSessionDelegate { get }
+}
+
 /// The `URLSession` delegate object which enables network requests instrumentation. **It must be
 /// used together with** `Datadog.Configuration.trackURLSession(firstPartyHosts:)`.
 ///
 /// All requests made with the `URLSession` instrumented with this delegate will be intercepted by the SDK.
 @objc
-open class DDURLSessionDelegate: NSObject, URLSessionTaskDelegate {
+open class DDURLSessionDelegate: NSObject, URLSessionTaskDelegate, URLSessionDelegateProviding {
+    public var ddURLSessionDelegate: DDURLSessionDelegate {
+        return self
+    }
+
     var interceptor: URLSessionInterceptorType?
     let firstPartyURLsFilter: FirstPartyURLsFilter?
 
--- a/Sources/DatadogObjc/DDURLSessionDelegate+objc.swift
+++ b/Sources/DatadogObjc/DDURLSessionDelegate+objc.swift
@@ -8,4 +8,27 @@ import Foundation
 import Datadog
 
 @objc
-public class DDNSURLSessionDelegate: DDURLSessionDelegate {}
+open class DDNSURLSessionDelegate: NSObject, URLSessionTaskDelegate, URLSessionDelegateProviding {
+    let swiftDelegate: DDURLSessionDelegate
+    public var ddURLSessionDelegate: DDURLSessionDelegate {
+        return swiftDelegate
+    }
+
+    @objc
+    override public init() {
+        swiftDelegate = DDURLSessionDelegate()
+    }
+
+    @objc
+    public init(additionalFirstPartyHosts: Set<String>) {
+        swiftDelegate = DDURLSessionDelegate(additionalFirstPartyHosts: additionalFirstPartyHosts)
+    }
+
+    open func urlSession(_ session: URLSession, task: URLSessionTask, didCompleteWithError error: Error?) {
+        swiftDelegate.urlSession(session, task: task, didCompleteWithError: error)
+    }
+
+    open func urlSession(_ session: URLSession, task: URLSessionTask, didFinishCollecting metrics: URLSessionTaskMetrics) {
+        swiftDelegate.urlSession(session, task: task, didFinishCollecting: metrics)
+    }
+}
--- a/Sources/DatadogObjc/DDURLSessionDelegate+objc.swift
+++ b/Sources/DatadogObjc/DDURLSessionDelegate+objc.swift
@@ -8,8 +8,8 @@ import Foundation
 import Datadog
 
 @objc
-open class DDNSURLSessionDelegate: NSObject, URLSessionTaskDelegate, URLSessionDelegateProviding {
-    let swiftDelegate: DDURLSessionDelegate
+open class DDNSURLSessionDelegate: NSObject, URLSessionTaskDelegate, __URLSessionDelegateProviding {
+    var swiftDelegate: DDURLSessionDelegate
     public var ddURLSessionDelegate: DDURLSessionDelegate {
         return swiftDelegate
     }
--- a/Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegateAsSuperclassTests.swift
+++ b/Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegateAsSuperclassTests.swift
@@ -6,6 +6,18 @@
 
 import XCTest
 import Datadog
+import DatadogObjc
+
+@objc
+public class SubDDNSURLSessionDelegate: DDNSURLSessionDelegate {
+    override public func urlSession(_ session: URLSession, task: URLSessionTask, didCompleteWithError error: Error?) {
+        super.urlSession(session, task: task, didCompleteWithError: error)
+    }
+
+    override public func urlSession(_ session: URLSession, task: URLSessionTask, didFinishCollecting metrics: URLSessionTaskMetrics) {
+        super.urlSession(session, task: task, didFinishCollecting: metrics)
+    }
+}
 
 internal final class SubDDURLSessionDelegate: DDURLSessionDelegate {
     let property: String
--- a/Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegateAsSuperclassTests.swift
+++ b/Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegateAsSuperclassTests.swift
@@ -34,5 +34,6 @@ class DDURLSessionDelegateAsSuperclassTests: XCTestCase {
     func testSubclassability() {
         // Success: tests compile, failure: compilation error
         _ = SubDDURLSessionDelegate()
+        _ = SubDDNSURLSessionDelegate()
     }
 }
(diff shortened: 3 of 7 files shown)


Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
```diff
@@ -6,12 +6,25 @@
 
 import Foundation
 
+/// In case that you are not able to use `DDURLSessionDelegate` or its subclasses in your `URLSession` instance,
+/// you can use `DDURLSessionDelegateProviding` protocol.
+/// Make sure that your type implements the same `URLSessionTaskDelegate` methods as `DDURLSessionDelegate` does.
+/// Otherwise your `URLSession` will not be instrumented automatically.
```
ncreated, line 12: This is not enough to _"implement the same methods"_, the calls must be forwarded to to our delegate. How about:
```suggestion
/// In case that you are not able to use `DDURLSessionDelegate` or its subclasses in your `URLSession` instance,
/// you can use `DDURLSessionDelegateProviding` protocol.
/// Implement this protocol in your `URLSessionDelegate` class and forward required API calls to `DDURLSessionDelegate`.
```

ncreated (commented):
All looks good 👌. But I left few notes - what you think on them?

Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
```diff
@@ -6,12 +6,25 @@
...
+@objc
+public protocol DDURLSessionDelegateProviding {
+    var delegate: DDURLSessionDelegate { get }
```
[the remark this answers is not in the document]
buranmert: tbh i still have doubts about this protocol

1. we can kind of hide it with discouraging comments while keeping it `public`
2. we don't have to introduce this protocol. i can still go with the other (`sharedInterceptor`) solution, its only problem was non-user-facing testability issues.
3. if we go with adding more `var ***delegate` then every new addition will be a breaking change -> major version update. if this protocol wasn't public, we could make those changes in minor/patch version updates.
4. if we go with specialised `xxxDelegateProviding` protocols then in the future we will have to make sure that existing protocols keep working, this may cause problems for our development

in any case, we communicate via comments which is not the ideal way. our users can end up with wrong implementation while the code still compiles successfully.

wdyt?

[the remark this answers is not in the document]
ncreated: My point is to also make something usable to this small number of users who currently report _"I can't subclass `DDURLSessionDelegate`"_ problem. The composition-based approach proposed in this PR may work as a replacement of inheritance.

> in any case, we communicate via comments which is not the ideal way. our users can end up with wrong implementation while the code still compiles successfully.

Yes, but the alternative is to provide no instrumentation for users who can't subclass `DDURLSessionDelegate` or do it through swizzling, resulting with tracking every single request made from the app or its 3rd party frameworks.

Anyway, this discussion is far from the core topic - adding `XCFrameworks`. What I proposed is solving another problem as a bonus, but we can do it separate.

Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
```diff
@@ -6,12 +6,25 @@
...
 /// The `URLSession` delegate object which enables network requests instrumentation. **It must be
 /// used together with** `Datadog.Configuration.trackURLSession(firstPartyHosts:)`.
 ///
 /// All requests made with the `URLSession` instrumented with this delegate will be intercepted by the SDK.
 @objc
-open class DDURLSessionDelegate: NSObject, URLSessionTaskDelegate {
+open class DDURLSessionDelegate: NSObject, URLSessionTaskDelegate, DDURLSessionDelegateProviding {
+    public var delegate: DDURLSessionDelegate {
+        return self
+    }
```
ncreated, line 26: It is IMO strange to see generic `delegate` property on the `delegate` object. Naming it `ddDelegate` / `ddForwardingDelegate` may sound better.

buranmert:
I changed the content of this PR, now we solve the problem without public API changes

Tests/DatadogTests/Datadog/Mocks/URLSessionAutoInstrumentationMocks.swift
```diff
@@ -25,27 +29,60 @@ class URLSessionInterceptorMock: URLSessionInterceptorType {
...
+//        tasksCompleted.append((task: task, error: error))
+//        onTaskCompleted?(task, error)
+//    }
+//
+//    func taskMetricsCollected(task: URLSessionTask, metrics: URLSessionTaskMetrics) {
+//        taskMetrics.append((task: task, metrics: metrics))
+//        onTaskMetricsCollected?(task, metrics)
+//    }
+//}
+
```
ncreated, line 85:
```suggestion
```

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegateTests.swift
```diff
@@ -8,12 +8,21 @@ import XCTest
...
+
+    private var currentAutoInstrumentation: URLSessionAutoInstrumentation?
+    override func setUpWithError() throws {
+        try super.setUpWithError()
+        currentAutoInstrumentation = URLSessionAutoInstrumentation.instance
+        URLSessionAutoInstrumentation.instance = URLSessionAutoInstrumentation(interceptor: interceptor)
+    }
+
+    override func tearDownWithError() throws {
+        URLSessionAutoInstrumentation.instance = currentAutoInstrumentation
```
ncreated, line 22: We shouldn't need to store the previous instance (`currentAutoInstrumentation`). Each test should start with a clean state and reset it after completion, otherwise it will introduce flakiness. This should be enough:
```swift
URLSessionAutoInstrumentation.instance = nil
```

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/URLSessionSwizzlerTests.swift
```diff
@@ -17,24 +17,30 @@ extension URLSessionSwizzler {
...
         swizzler.swizzle()
+
+        currentAutoInstrumentation = URLSessionAutoInstrumentation.instance
+        URLSessionAutoInstrumentation.instance = URLSessionAutoInstrumentation(interceptor: interceptor)
     }
 
     override func tearDown() {
         swizzler.unswizzle()
+
+        URLSessionAutoInstrumentation.instance = currentAutoInstrumentation
```
ncreated, line 36: Same: we shouldn't need to store the previous instance (`currentAutoInstrumentation`). Each test should start with a clean state and reset it after completion, otherwise it will introduce flakiness. This should be enough:
```swift
URLSessionAutoInstrumentation.instance = nil
```

Sources/DatadogObjc/DDURLSessionDelegate+objc.swift
```diff
@@ -8,4 +8,42 @@ import Foundation
 import Datadog
 
 @objc
-public class DDNSURLSessionDelegate: DDURLSessionDelegate {}
+open class DDNSURLSessionDelegate: NSObject, URLSessionTaskDelegate {
```
ncreated, line 11: There's a bunch of critical and not very obvious implementations in this class - this needs tests to document and assert expected behaviours.

ncreated (changes requested):

Sources/DatadogObjc/DDURLSessionDelegate+objc.swift
buranmert: `TracingNSURLSessionScenario` and `RUMNSURLSessionResourcesScenario` test `DDNSURLSessionDelegate`, what other test cases may we need?

what about that @ncreated ?
```
/// NOTE: RUMM-1234
/// `DDNSURLSessionDelegate` imitates `DDURLSessionDelegate` class
/// it responds to whatever `DDURLSessionDelegate` responds to and
/// it forwards every method call to its private `DDURLSessionDelegate` instance.
/// That way, `DDNSURLSessionDelegate` inherits `DDURLSessionDelegate` functionality
/// without actual inheritance/superclass.
```

ncreated: There are no `DDNSURLSessionDelegate` unit tests - there was no point having it as this class was only subclassing the swift delegate. Now that it executes its own logic we need to add tests as usual.

Tests you referred to are integration tests covering basic scenarios. We can't leverage them for documenting the code, but we can do it easily with few unit tests.

Tests/DatadogTests/DatadogObjc/DDNSURLSessionDelegateTests.swift
```diff
@@ -0,0 +1,237 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import XCTest
+@testable import Datadog
+@testable import DatadogObjc
+
+class DDNSURLSessionDelegateTests: XCTestCase {
```
ncreated, line 11: Those tests are copy & pasted [from `DDURLSessionDelegate` tests](https://github.com/DataDog/dd-sdk-ios/blob/master/Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegateTests.swift) and do not cover anything new than what's covered in swift delegate tests. We should write tests for the new logic we added.

The new behaviours added in `DDNSURLSessionDelegate` are:
* 1/ objc delegate must forward all calls to the wrapped swift delegate,
* 2/ objc delegate's type must appear as the wrapped swift delegate type at runtime.

Additionally, as objc delegate is `open class` and can be subclassed, we should assert that:
* 3/ when subclassing objc delegate it doesn't change the 1st and 2nd behaviour.

Sources/DatadogObjc/DDURLSessionDelegate+objc.swift
```diff
@@ -8,4 +8,49 @@ import Foundation
...
+        return DDURLSessionDelegate.self
+    }
+
+    override open func responds(to aSelector: Selector!) -> Bool {
+        return swiftDelegate.responds(to: aSelector)
+    }
+
+    override open class func responds(to aSelector: Selector!) -> Bool {
+        return DDURLSessionDelegate.responds(to: aSelector)
+    }
```
ncreated, line 46: This way of forwarding calls to wrapped swift delegate doesn't work if subclassing the `DDNSURLSessionDelegate`. The `super` methods are not available as they are not explicitly defined in our delegate, so the code doesn't even compile. This can be seen when defining unit test for objc → swift calls forwarding:

Screenshot 2021-04-08 at 10 54 47

Screenshot 2021-04-08 at 10 56 49

Sources/DatadogObjc/DDURLSessionDelegate+objc.swift
```diff
@@ -8,4 +8,49 @@ import Foundation
...
+    // swiftlint:enable implicitly_unwrapped_optional
+
+    override open func isKind(of aClass: AnyClass) -> Bool {
+        return swiftDelegate.isKind(of: aClass)
+    }
+
+    override open class func isKind(of aClass: AnyClass) -> Bool {
+        return DDURLSessionDelegate.isKind(of: aClass)
+    }
```
ncreated, line 55: This changes not only the kind of our delegate, but also its user-defined superclass. This can be seen when defining test for delegate's type appearing at runtime:

Screenshot 2021-04-08 at 11 03 37

Changing the runtime identity of user object is dangerous, as it will break any user logic which depends on the delegate kind.

ncreated (changes requested):

Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
```diff
@@ -6,12 +6,28 @@
 
 import Foundation
 
+/// In case that you are not able to use `DDURLSessionDelegate` or its subclasses in your `URLSession` instance,
+/// you can use `DDURLSessionDelegateProviding` protocol.
+/// Implement this protocol in your `URLSessionDelegate` class and forward required API calls to `DDURLSessionDelegate`.
+@objc
+public protocol URLSessionDelegateProviding: URLSessionDelegate {
```
ncreated, line 13: We could additionally prefix this name with `__` (`__URLSessionDelegateProviding`), which would emphasize its internal use. And the protocol comment could be less-encouraging for the use - IMO sth less direct may work better:
```swift
/// An interface for forwarding `URLSessionDelegate` calls to `DDURLSessionDelegate`.
/// The implementation must ensure that required methods are called on the `ddURLSessionDelegate`.
```

Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
```diff
@@ -6,12 +6,28 @@
...
+    /// Datadog delegate object.
+    /// The class implementing `DDURLSessionDelegateProviding` must ensure that following method calls are forwarded to `ddURLSessionDelegate`:
+    // - `func urlSession(_:task:didFinishCollecting:)`
+    // - `func urlSession(_:task: didCompleteWithError:)`
```
ncreated, line 17:
```suggestion
    // - `func urlSession(_:task:didCompleteWithError:)`
```

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegateAsSuperclassTests.swift
```diff
@@ -6,6 +6,18 @@
...
+@objc
+public class SubDDNSURLSessionDelegate: DDNSURLSessionDelegate {
+    override public func urlSession(_ session: URLSession, task: URLSessionTask, didCompleteWithError error: Error?) {
+        super.urlSession(session, task: task, didCompleteWithError: error)
+    }
+
+    override public func urlSession(_ session: URLSession, task: URLSessionTask, didFinishCollecting metrics: URLSessionTaskMetrics) {
+        super.urlSession(session, task: task, didFinishCollecting: metrics)
+    }
+}
```
ncreated, line 20: I think we may also want to instantiate this class as we do for `DDURLSessionDelegate` subclass:
```swift
_ = SubDDURLSessionDelegate()
```

Sources/DatadogObjc/DDURLSessionDelegate+objc.swift
```diff
@@ -8,4 +8,27 @@ import Foundation
...
+
+    open func urlSession(_ session: URLSession, task: URLSessionTask, didCompleteWithError error: Error?) {
+        swiftDelegate.urlSession(session, task: task, didCompleteWithError: error)
+    }
+
+    open func urlSession(_ session: URLSession, task: URLSessionTask, didFinishCollecting metrics: URLSessionTaskMetrics) {
+        swiftDelegate.urlSession(session, task: task, didFinishCollecting: metrics)
+    }
+}
```
ncreated, line 34: Even though it's a simple wrapper, we should add unit tests (like we do for other Objc <> Swift wrappers):
```swift
func testItForwardsCallsToSwiftDelegate() {
    let objcDelegate = DDNSURLSessionDelegate()
    let swiftDelegate = DDURLSessionDelegateMock()
    objcDelegate.swiftDelegate = swiftDelegate

    let sessionMock = URLSession.shared
    let taskMock = URLSessionTask.mockWith(request: .mockAny(), response: .mockAny())
    let metricsMock = URLSessionTaskMetrics.mockAny()
    let errorMock = ErrorMock()

    objcDelegate.urlSession(sessionMock, task: taskMock, didFinishCollecting: metricsMock)
    objcDelegate.urlSession(sessionMock, task: taskMock, didCompleteWithError: errorMock)

    XCTAssertTrue(swiftDelegate.calledDidFinishCollecting)
    XCTAssertTrue(swiftDelegate.calledDidCompleteWithError)
}
```
```swift
class DDURLSessionDelegateMock: DDURLSessionDelegate {
    var calledDidFinishCollecting = false
    var calledDidCompleteWithError = false

    override func urlSession(_ session: URLSession, task: URLSessionTask, didFinishCollecting metrics: URLSessionTaskMetrics) {
        calledDidFinishCollecting = true
    }

    override func urlSession(_ session: URLSession, task: URLSessionTask, didCompleteWithError error: Error?) {
        calledDidCompleteWithError = true
    }
}
```
Two tests for initialisation looks easy as well.

ncreated (changes requested):

ncreated (approved):

PR #462, opened 2021-04-02
RUMM-1087 Standalone binary distribution

buranmert:
### What and why?

We need to provide our SDK as pre-build binaries to the users who don't use any package managers.

### How?

We build our SDK with `carthage --use-xcframeworks --no-skip-current` and then will upload the zipped `Carthage/Build/` folder to Github release

I tried the script in my personal fork first, you can see the output [here](https://github.com/buranmert/dd-sdk-ios/releases/tag/0.0.1)

### TODO

- [ ] https://github.com/DataDog/dd-sdk-ios/pull/461 should be merged first

### Review checklist

### DIFF
--- a/tools/standalone-binary-distro/make_distro_builds.sh
+++ b/tools/standalone-binary-distro/make_distro_builds.sh
@@ -1,0 +1,46 @@
+#!/bin/zsh
+
+gh auth status;
+if [[ $? != 0 ]]; then
+  exit 1
+fi
+
+SDK_VERSION=$(git describe --exact-match --tags)
+if [[ $? != 0 ]]; then
+  echo "❌ Aborting: HEAD doesn't have a tag!"  
+  exit 2
+elif [[ `git status --porcelain` ]]; then
+  echo "❌ Aborting: Working directory has changes!"
+  exit 3
+elif [ ! -f "Cartfile" ]; then
+	echo "❌ Aborting: \`make_distro_builds.sh\` must be run from the same place with `Cartfile`"
+  exit 4
+fi
+OUT_FILENAME="Datadog-$SDK_VERSION.zip"
+
+# create temporary xcconfig for carthage build
+set -euo pipefail
+xcconfig=$(mktemp /tmp/static.xcconfig.XXXXXX)
+trap 'rm -f "$xcconfig" ;' INT TERM HUP EXIT
+echo "BUILD_LIBRARY_FOR_DISTRIBUTION = YES" >> $xcconfig
+export XCODE_XCCONFIG_FILE="$xcconfig"
+
+# clear existing carthage artifacts
+rm -rf Carthage/
+rm Cartfile.resolved
+
+# fetch 3rd party deps via carthage
+echo "carthage bootstrap with no build..."
+carthage bootstrap --no-build
+
+echo "carthage build..."
+carthage build --platform iOS --use-xcframeworks --no-skip-current
+
+# zip artifacts
+cd Carthage/Build/
+zip --symlinks -r $OUT_FILENAME *.xcframework
+mv $OUT_FILENAME ../../$OUT_FILENAME
+cd ../../
+
+# upload zip
+gh release upload $SDK_VERSION $OUT_FILENAME -R DataDog/dd-sdk-ios
--- a/tools/standalone-binary-distro/make_distro_builds.sh
+++ b/tools/standalone-binary-distro/make_distro_builds.sh
@@ -27,7 +27,6 @@ export XCODE_XCCONFIG_FILE="$xcconfig"
 
 # clear existing carthage artifacts
 rm -rf Carthage/
-rm Cartfile.resolved
 
 # fetch 3rd party deps via carthage
 echo "carthage bootstrap with no build..."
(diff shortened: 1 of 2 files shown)


tools/standalone-binary-distro/make_distro_builds.sh
```diff
@@ -0,0 +1,46 @@
...
+# create temporary xcconfig for carthage build
+set -euo pipefail
+xcconfig=$(mktemp /tmp/static.xcconfig.XXXXXX)
+trap 'rm -f "$xcconfig" ;' INT TERM HUP EXIT
+echo "BUILD_LIBRARY_FOR_DISTRIBUTION = YES" >> $xcconfig
+export XCODE_XCCONFIG_FILE="$xcconfig"
+
+# clear existing carthage artifacts
+rm -rf Carthage/
+rm Cartfile.resolved
```
ncreated, line 30: Why deleting [`Cartfile.resolved`](https://github.com/DataDog/dd-sdk-ios/blob/dogfooding/Cartfile.resolved)? If removed the `carthage bootstrap` will use up-to-date versions of our dependencies, not the one that we work against.

tools/standalone-binary-distro/make_distro_builds.sh
```diff
@@ -0,0 +1,46 @@
...
+carthage build --platform iOS --use-xcframeworks --no-skip-current
+
+# zip artifacts
+cd Carthage/Build/
+zip --symlinks -r $OUT_FILENAME *.xcframework
+mv $OUT_FILENAME ../../$OUT_FILENAME
+cd ../../
+
+# upload zip
+gh release upload $SDK_VERSION $OUT_FILENAME -R DataDog/dd-sdk-ios
```
ncreated, line 46: `-R DataDog/dd-sdk-ios` is redundant - we'll be running this tool from our repo.
```suggestion
gh release upload $SDK_VERSION $OUT_FILENAME
```

tools/standalone-binary-distro/make_distro_builds.sh
```diff
@@ -0,0 +1,46 @@
...
+echo "carthage bootstrap with no build..."
+carthage bootstrap --no-build
+
+echo "carthage build..."
+carthage build --platform iOS --use-xcframeworks --no-skip-current
+
+# zip artifacts
+cd Carthage/Build/
+zip --symlinks -r $OUT_FILENAME *.xcframework
+mv $OUT_FILENAME ../../$OUT_FILENAME
```
ncreated, line 42: This will leave the repo dirty - quite annoying to always delete this file after `make ship`. Can't we upload it directly from `Carthage/Build/$OUT_FILENAME`? The `Carthage` folder is already git-ignored.

tools/standalone-binary-distro/make_distro_builds.sh
```diff
@@ -0,0 +1,46 @@
...
+# clear existing carthage artifacts
+rm -rf Carthage/
+rm Cartfile.resolved
+
+# fetch 3rd party deps via carthage
+echo "carthage bootstrap with no build..."
+carthage bootstrap --no-build
+
+echo "carthage build..."
+carthage build --platform iOS --use-xcframeworks --no-skip-current
```
ncreated, line 37: How does it know to also build our `Datadog` frameworks? Does the `--no-skip-current` make it lookup `.xcschemes` in current hierarchy?

ncreated (changes requested):

tools/standalone-binary-distro/make_distro_builds.sh
```diff
@@ -0,0 +1,46 @@
...
+
+gh auth status;
+if [[ $? != 0 ]]; then
+  exit 1
+fi
+
+SDK_VERSION=$(git describe --exact-match --tags)
+if [[ $? != 0 ]]; then
+  echo "❌ Aborting: HEAD doesn't have a tag!"  
+  exit 2
```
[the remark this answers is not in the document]
buranmert: i didn't understand what you tried to solve here 🤔 can you explain a bit more?

this script can be run outside of `make ship` command too and you must be in a commit that is tagged while running this script, otherwise you may create binaries from `commit1` and upload it to `release2`

tools/standalone-binary-distro/make_distro_builds.sh
buranmert: yes

tools/standalone-binary-distro/make_distro_builds.sh
buranmert: i'll keep it because if the repo has other remotes, `gh` doesn't tell you which repo it tries when it fails.
do you think it may cause problems?

tools/standalone-binary-distro/make_distro_builds.sh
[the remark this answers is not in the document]
ncreated: Ah I see. So this indeed requires fetching the tag from remote - makes sense for the consistency 👍.

tools/standalone-binary-distro/make_distro_builds.sh
ncreated: No problem at all 👌, just noting this for the solution simplicity (less code). I don't think we'd ever have a scenario for adding remote (IMO equally possible as changing the repo / org name 😉).

tools/standalone-binary-distro/make_distro_builds.sh
buranmert: this may make `carthage bootstrap` unneeded, unresolving the discussion for a moment ⏳

tools/standalone-binary-distro/make_distro_builds.sh
buranmert: to be clear, in my local i have 2 remotes (my fork and upstream)
actually, i added my fork to test this PR and will probably remove it after but that's why i add `-R` in the first place.

ncreated (approved):

PR #464, opened 2021-04-09
RUMM-1087 Test schemes are hidden from make_distro_builds

buranmert:
### What and why?

`dependency-manager-tests` schemes can be tried by `carthage build`

### How?

We hide this folder temporarily

### Review checklist

### DIFF
--- a/tools/standalone-binary-distro/make_distro_builds.sh
+++ b/tools/standalone-binary-distro/make_distro_builds.sh
@@ -21,7 +21,9 @@ OUT_FILENAME="Datadog-$SDK_VERSION.zip"
 # create temporary xcconfig for carthage build
 set -euo pipefail
 xcconfig=$(mktemp /tmp/static.xcconfig.XXXXXX)
-trap 'rm -f "$xcconfig" ;' INT TERM HUP EXIT
+# carthage may try to build schemes in dependency-manager-tests folder, so we hide it temporarily
+mv dependency-manager-tests/ .dependency-manager-tests/
+trap 'rm -f "$xcconfig" ; mv .dependency-manager-tests/ dependency-manager-tests/ ;' INT TERM HUP EXIT
 echo "BUILD_LIBRARY_FOR_DISTRIBUTION = YES" >> $xcconfig
 export XCODE_XCCONFIG_FILE="$xcconfig"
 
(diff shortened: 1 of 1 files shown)


tools/standalone-binary-distro/make_distro_builds.sh
```diff
@@ -21,7 +21,9 @@ OUT_FILENAME="Datadog-$SDK_VERSION.zip"
 # create temporary xcconfig for carthage build
 set -euo pipefail
 xcconfig=$(mktemp /tmp/static.xcconfig.XXXXXX)
-trap 'rm -f "$xcconfig" ;' INT TERM HUP EXIT
+# carthage may try to build schemes in dependency-manager-tests folder, so we hide it temporarily
+mv dependency-manager-tests/ .dependency-manager-tests/
+trap 'rm -f "$xcconfig" ; mv .dependency-manager-tests/ dependency-manager-tests/ ;' INT TERM HUP EXIT
```
ncreated, line 26: This is getting quite messy as we're adding more patching. It will be hard to ensure clean state before / after running this script if we don't control it clearly. How about ensuring clean state before / after running this script by introducing simple functions, e.g.:
```shell
$xcconfig = "binary_distribution.xcconfig"

# Prepares the repository folder for building xcframeworks.
function prepare() {
   echo "BUILD_LIBRARY_FOR_DISTRIBUTION = YES" >> $xcconfig
   mv dependency-manager-tests/ .dependency-manager-tests/ 
   mv Carthage .Carthage
}

# Resets repository changes made in 'prepare()'
function cleanup() {
   rm $xcconfig
   mv .dependency-manager-tests/ dependency-manager-tests/
   mv .Carthage Carthage
}

trap cleanup INT TERM HUP EXIT
prepare

# ... produce xcframeworks
```

tools/standalone-binary-distro/make_distro_builds.sh
```diff
@@ -34,6 +36,9 @@ carthage bootstrap --no-build
 
 echo "carthage build..."
 carthage build --platform iOS --use-xcframeworks --no-skip-current
+# reset trap
+trap - INT TERM HUP EXIT
+rm -f "$xcconfig"; mv .dependency-manager-tests/ dependency-manager-tests/ ;
```
ncreated, line 41: Why resetting the trap? It makes it unclear on how to maintain this script for next changes - should we let the trap undo state changes, or do it by hand after all? What if the script fails after trap is reset, but not all changes are reset?

tools/standalone-binary-distro/make_distro_builds.sh
buranmert: i made changes along those lines 👍

ncreated (approved):

PR #473, opened 2021-04-22
RUMM-1154 User can add attributes to auto instrumented RUM Resources

ncreated:
### What and why?

📦 This PR adds `rumResourceAttributesProvider` API, which makes it possible to add extra attributes to RUM Resources collected automatically:

```swift
public func setRUMResourceAttributesProvider(
   _ provider: @escaping (URLRequest, URLResponse?, Data?, Error?) -> [AttributeKey: AttributeValue]?
)
```

The `provider` block is called for each `URLSessionTask` intercepted by the SDK. The implementation should look into task information and return custom attributes for RUM Resource, e.g. send redacted response headers or decode response body and send its additional context. This block is also called for RUM Errors created for resources failed on transport.

This is a counterpart of https://github.com/DataDog/dd-sdk-android/pull/520 introduced in `dd-sdk-android`.

### How?

The implementation starts with passing the `provider` block through configuration object to `URLSessionRUMResourcesHandler` as we do for other stuff. `URLSessionInterceptor` is extended with additional method: `taskReceivedData(task:data:)` which is called from two places:
* from `URLSesionSwizzler` - for requests sent with completion block,
* from `DDURLSessionDelegate` - for requests sent with no completion block.

About the public API, I considered few other choices:
* adding the `provider` parameter to `DDURLSessionDelegate.init()` - this would require also exposing it in `URLSessionInterceptor` (for no-delegate usage including Alamofire extension),
* adding the `provider` parameter to existing `.trackURLSession(firstPartyHosts:)` as both are mutually related - this would make the discovery of this new feature harder (the doc comment for `.trackURLSession` is already huge).

The proposed API should keep things simple: the feature can announced by its API name and the mutual relationship with `trackURLSession` is solved by installing misconfiguration warning.

All related unit tests pass on iOS 14.3, 13.4, 12.4 and 11.1.

### Review checklist

### DIFF
--- a/Datadog/Example/Scenarios/RUM/RUMScenarios.swift
+++ b/Datadog/Example/Scenarios/RUM/RUMScenarios.swift
@@ -129,6 +129,7 @@ final class RUMURLSessionResourcesScenario: URLSessionBaseScenario, TestScenario
     override func configureSDK(builder: Datadog.Configuration.Builder) {
         _ = builder
             .trackUIKitRUMViews()
+            .setRUMResourceAttributesProvider(rumResourceAttributesProvider(request:response:data:error:))
 
         super.configureSDK(builder: builder) // applies the `trackURLSession(firstPartyHosts:)`
     }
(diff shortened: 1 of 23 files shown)


xgouchet (approved):
LGTM 👍

buranmert (approved):

PR #476, opened 2021-04-30
make createSpanText accessible to implementing modules

tshedor:
### What and why?

Libraries that extend dd-sdk-ios do not have access to generating span or trace IDs. This is very likely by design, however, it excludes community SDKs like [datadog_flutter](https://github.com/greenbits/datadog_flutter). This PR allows access of `createSpanContext` in modules (such as a Flutter plugin). 

This may be an [overly simplistic implementation](https://github.com/greenbits/datadog_flutter/compare/unsound-null-safety...add-trace?expand=1), and I'm likely not understanding the large complexity behind trace IDs. But I figured that, either way, a PR would be some answer. 

### How?

`internal` access is converted to `public` access.

### Review checklist

^ Not sure on the JIRA or Issue? Is this a public JIRA?

### DIFF
--- a/Sources/Datadog/Tracer.swift
+++ b/Sources/Datadog/Tracer.swift
@@ -183,9 +183,7 @@ public class Tracer: OTTracer {
         return activeSpansPool.getActiveSpan()
     }
 
-    // MARK: - Internal
-
-    internal func createSpanContext(parentSpanContext: DDSpanContext? = nil) -> DDSpanContext {
+    public func createSpanContext(parentSpanContext: DDSpanContext? = nil) -> DDSpanContext {
         return DDSpanContext(
             traceID: parentSpanContext?.traceID ?? (environmentSpanIntegration.environmentSpanContext?.traceID ?? tracingUUIDGenerator.generateUnique()),
             spanID: tracingUUIDGenerator.generateUnique(),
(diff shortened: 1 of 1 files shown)


ncreated:
Hello @tshedor 👋. We're extremely happy to see community efforts on bridging our native SDKs to Flutter 🚀🌟. To help you more on this, we need to better understand the context of this PR, especially:
* why do you need to generate the span context and / or know the trace / span ID?
* cannot this be achieved with just using [`Global.sharedTracer.startSpan(operationName:)` API](https://docs.datadoghq.com/tracing/setup/ios/?tab=us)?

On a wider picture - span context is something internal to our SDK. The user should only work with `OTSpan` and `OTSpanContext` interfaces, which follow the [Open Tracing spec](https://github.com/opentracing/specification/blob/master/specification.md). That said we rather avoid exposing concrete `DDSpanContext` in public API.

tshedor:
@ncreated Thanks for getting back to me on this PR. 

I'm trying to add trace and span IDs to Flutter HTTP calls. Originally, I was reverse engineering the code, trying to figure out where the headers were generated and then extracting that code. 

However, I completely overlooked that documentation. Thank you for linking it. That's exactly what I need to generate the headers and to acknowledge that the span finished.

Thanks for your attention and patience, I really appreciate it.

ncreated:
You're welcome @tshedor 👌, happy to help. We're super excited with your community efforts 💪, feel free to ask any further questions 🙂. You may want to use our public Slack to reach us directly (https://www.datadoghq.com/support/) and get more help.

PR #478, opened 2021-04-30
RUMM-1104 Add `SpanEvent` scrubbing API

ncreated:
### What and why?

📦 This PR brings the first part of data scrubbing for `SpanEvents`. It applies necessary refactoring to ensure we can pass clear and understandable values to the user. The scrubbing API will be added in next PR, to keep the size small.

### How?

To allow span events scrubbing, the the [`Span` declaration](https://github.com/DataDog/dd-sdk-ios/blob/1cd1a46c61e6a792e138b3aac00a29f5a12c97e7/Sources/Datadog/Tracing/Span/SpanEncoder.swift#L34) must be made `public`. It is easy for all properties, except two:
```swift
struct Span {
   // ...
   struct UserInfo {
      // ...
      var extraInfo: [String: JSONStringEncodableValue]
   }
   // ...
   var tags: [String: JSONStringEncodableValue]
}
```
Exposing `JSONStringEncodableValue` in public API would require educating users how to work with this wrapper.

To mitigate this, the refactoring done in this PR replaces types of `extraInfo` and `tags` with `[String: String]`, which is the real type used to transport those values to Datadog (it also has its counterpart in [span-schema.json](https://github.com/DataDog/dd-sdk-android/blob/master/dd-sdk-android/src/main/json/trace/span-schema.json) in `dd-sdk-android`).

As a consequence of replacing `JSONStringEncodableValue` with `String`, I had to update the threading model around building the `Span` event. `JSONStringEncodableValue` was doing its JSON string encoding work **lazily** on the writer's thread (changing `Encodable` to `String` is expensive and grows with the number of tags / extra info attributes). Now this work must be done in `SpanBuilder`, which used to be run on the user thread. 

This PR invokes the `SpanBuilder` on background queue, making sure it doesn't delay the user thread with `Encodable` <> `String` conversion. This means that spans collection and encoding is now unified with RUM events, where corresponding things happen off the main thread.

### Review checklist


Sources/Datadog/Tracing/Span/SpanBuilder.swift
```diff
@@ -23,50 +23,129 @@ internal struct SpanBuilder {
...
+
+    /// Encodes `Span` attributes to JSON strings
+    private let attributesJSONEncoder: JSONEncoder = .default()
+
+    /// Converts `Encodable` attributes to its lossless JSON string representation, e.g.:
+    /// * it will convert `"abc"` string value to `"abc"` JSON string value
+    /// * it will convert `1` integer value to `"1"` JSON string value
+    /// * it will convert `true` boolean value to `"true"` JSON string value
+    /// * it will convert `Person(name: "foo")` encodable struct to `"{\"name\": \"foo\"}"` JSON string value
+    private func castValuesToString(_ dictionary: [String: Encodable]) -> [String: String] {
```
ncreated, line 89: 💡  This is the old code from `JSONStringEncodableValue` it's just moved to `SpanBuilder` as now we convert `[String: Encodable]` to `[String: String]` in there, instead of doing this lazy on writer's thread.

Tests/DatadogTests/Datadog/TracerTests.swift
```diff
@@ -333,9 +333,11 @@ class TracerTests: XCTestCase {
         let tracer = Tracer.initialize(configuration: .init()).dd
 
         tracer.startSpan(operationName: "span with no user info").finish()
+        tracer.queue.sync {} // wait for processing the span event in `DDSpan`
```
ncreated, line 336: 💡  We need to synchronize test execution like this, otherwise the `Span` creation asynchronously dispatched from may fall into race condition `.setUserInfo()` called in the next line. It's a common trick, already used widely [in `RUMMonitorTests`](https://github.com/DataDog/dd-sdk-ios/blob/1cd1a46c61e6a792e138b3aac00a29f5a12c97e7/Tests/DatadogTests/Datadog/RUMMonitorTests.swift#L804).

Sources/Datadog/Tracing/Span/SpanBuilder.swift
```diff
@@ -23,50 +23,129 @@ internal struct SpanBuilder {
...
+        traceID: TracingUUID,
+        spanID: TracingUUID,
+        parentSpanID: TracingUUID?,
+        operationName: String,
+        startTime: Date,
+        finishTime: Date,
+        tags: [String: Encodable],
+        baggageItems: [String: String],
+        logFields: [[String: Encodable]]
+    ) -> Span {
```
ncreated, line 36: 💡 By passing these immutable values we avoid the thread sync penalty when reading them from `DDSpan`. If reading one by one, we'd need `queue.sync {}` on each value. If passing in bulk, we only sync once.

buranmert (approved):

PR #479, opened 2021-05-03
RUMM-1218 App going background stops active RUM view

buranmert:
### What and why?

When the app goes to background, technically there is no active view. Previous behavior didn't match that.

### How?

When the app goes to background (by resigning active) `UIKitRUMViewsHandler` stops the last started RUM view and keeps a weak reference to it.
When the app comes back to foreground (IOW when it does become active) `UIKitRUMViewsHandler` re-starts that view.

### Review checklist

### DIFF
--- a/Sources/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsHandler.swift
+++ b/Sources/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsHandler.swift
@@ -33,25 +33,6 @@ internal class UIKitRUMViewsHandler: UIKitRUMViewsHandlerType {
         nc.addObserver(self, selector: #selector(appDidBecomeActive), name: UIApplication.didBecomeActiveNotification, object: nil)
     }
 
-    private weak var lastActiveAppVC: UIViewController?
-    @objc
-    private func appWillResignActive() {
-        if lastActiveAppVC != nil {
-            return
-        }
-        lastActiveAppVC = lastStartedViewController
-        stop(viewController: lastActiveAppVC)
-    }
-
-    @objc
-    private func appDidBecomeActive() {
-        if let vc = lastActiveAppVC,
-           let rumView = predicate.rumView(for: vc) {
-            start(viewController: vc, rumView: rumView)
-        }
-        lastActiveAppVC = nil
-    }
-
     // MARK: - UIKitRUMViewsHandlerType
 
     weak var subscriber: RUMCommandSubscriber?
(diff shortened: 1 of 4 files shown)


Sources/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsHandler.swift
```diff
@@ -27,6 +27,29 @@ internal class UIKitRUMViewsHandler: UIKitRUMViewsHandlerType {
...
+        nc.addObserver(self, selector: #selector(appWillResignActive), name: UIApplication.willResignActiveNotification, object: nil)
+        nc.addObserver(self, selector: #selector(appDidBecomeActive), name: UIApplication.didBecomeActiveNotification, object: nil)
+    }
+
+    private weak var lastActiveAppVC: UIViewController?
+    @objc
+    private func appWillResignActive() {
+        if lastActiveAppVC != nil {
+            return
+        }
```
xgouchet, line 41: What if the `lastActiveAppVC` is not `nil` but doesn't match the `lastStartedViewController` ? When the app comes back in background, won't you be starting the wrong RUM View ?

Tests/DatadogTests/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsHandlerTests.swift
```diff
@@ -193,4 +196,57 @@ class UIKitRUMViewsHandlerTests: XCTestCase {
...
+        // Given
+        predicate.result = .init(name: viewName, attributes: ["foo": "bar"])
+        handler.notify_viewDidAppear(viewController: view, animated: .mockAny())
+
+        // When
+        for _ in 0..<3 {
+            NotificationCenter.default.post(name: UIApplication.willResignActiveNotification, object: nil)
+        }
+        dateProvider.advance(bySeconds: 1)
+        for _ in 0..<3 {
```
ncreated, line 214: It's quite unclear why we call it 3 times - I guess it's just for sanity, no? Maybe it's worth expressing somehow (either with an inline comment or by labelling `3` with a named variable).

Sources/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsHandler.swift
ncreated: Also, I'm wondering if we need the `lastActiveAppVC` property at all. As far as I can see, we either set it to `lastStartedViewController` or to `nil`, so can't we simply do:
* if the app goes to background: stop `lastStartedViewController`, 
* when it goes to foreground: start `lastStartedViewController`
?

If there's sth special about `lastActiveAppVC` that I miss, I think it's worth adding it in the property comment.

Tests/DatadogTests/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsHandlerTests.swift
buranmert: removed ✅

Sources/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsHandler.swift
buranmert: @xgouchet good point. 
now i remove `lastActiveAppVC` along with @ncreated comments below. it was introduced to guard against multiple notifications received but probably not really needed.

ncreated (approved):
LG 👌

PR #481, opened 2021-05-04
RUMM-1104 Add `SpanEvent` scrubbing API - implement event mapper

ncreated:
### What and why?

📦 This PR adds the actual implementation of `SpanEvents` scrubbing - the `SpanEvents` mapper API:
```swift
public func setSpanEventMapper(_ mapper: @escaping (SpanEvent) -> SpanEvent)
```

### How?

Similar to how RUM events mapper works, the mapper is set up from public interface and passed to the `SpanBuilder` through `FeaturesConfiguration` structure. If the mapper is configured, the builder asks it for redacted version of the `SpanEvent`.

This required making `SpanEvent` public along with some of its dependencies (notably `NetworkConnectionInfo` and `CarrierInfo`).

### Review checklist

### DIFF
--- a/Tests/DatadogTests/Helpers/EquatableInTests.swift
+++ b/Tests/DatadogTests/Helpers/EquatableInTests.swift
@@ -1,0 +1,80 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+/// Utility protocol adding `Equatable` conformance to any arbitrary type.
+/// The equatabiliity is determined based on comparing type mirrors and values.
+protocol EquatableInTests: Equatable {}
+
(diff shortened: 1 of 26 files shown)


Tests/DatadogTests/Helpers/EquatableInTests.swift
```diff
@@ -0,0 +1,80 @@
...
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+/// Utility protocol adding `Equatable` conformance to any arbitrary type.
+/// The equatabiliity is determined based on comparing type mirrors and values.
+protocol EquatableInTests: Equatable {}
+
+extension EquatableInTests {
```
ncreated, line 13: 💡 This new implementation of `EquatabeInTests` is way smarter than the previous one. The old one was just comparing `String(describing: lhs)` vs `String(describing: rhs)`, which was not working with unordered collections like `[String: Encodable]` or `Set<AnyHashable>`. Now, the comparison looks into each value's `Mirror` and decides on the equality using reflection 👌.

buranmert (approved):

PR #482, opened 2021-05-04
RUMM-1104 Add `SpanEvent` scrubbing API - cleanup

ncreated:
### What and why?

📦 This is the final PR for `SpanEvent` scrubbing feature. It cleans things up in `DDSpan` implementation, mainly reducing the number of queue synchronisations that we were doing there.

This cleanup is possible due to refactoring done in #478 - where entire `SpanEvent` creation was moved to secondary thread.

### How?

Before #478, the `SpanBuilder` was reading span information from `DDSpan` partially:
```swift
_ = ddspan.operationName
_ = ddspan.tags
...
```
and each read [required thread synchronisation on the property getter](https://github.com/DataDog/dd-sdk-ios/blob/658dee7b6515aa296f09e43016759c4f6fa85164/Sources/Datadog/Tracing/DDSpan.swift#L22-L46).

After #478 all properties are read in bulk and no longer require thread sync. This PR removes the old property getters that are no longer used. It also cleans up tests for `DDSpan` by separating `OTSpan.setError(_:)` tests to separate file.

### Review checklist

### DIFF
--- a/Sources/Datadog/Tracing/DDSpan.swift
+++ b/Sources/Datadog/Tracing/DDSpan.swift
@@ -24,32 +24,16 @@ internal class DDSpan: OTSpan {
     /// Writes span logs to output. `nil` if Logging feature is disabled.
     private let logOutput: LoggingForTracingAdapter.AdaptedLogOutput?
 
-    /// Unsynchronized span operation name. Use `self.operationName` setter & getter.
+    /// Queue used for synchronizing mutable properties access.
+    private let queue: DispatchQueue
+    /// Unsynchronized span operation name. Must be accessed on `queue`.
     private var unsafeOperationName: String
-    private(set) var operationName: String {
-        get { ddTracer.queue.sync { unsafeOperationName } }
-        set { ddTracer.queue.async { self.unsafeOperationName = newValue } }
-    }
-
-    /// Unsynchronized span tags. Use `self.tags` setter & getter.
+    /// Unsynchronized span tags.  Must be accessed on `queue`.
     private var unsafeTags: [String: Encodable]
-    var tags: [String: Encodable] {
-        ddTracer.queue.sync { unsafeTags }
-    }
-
-    /// Unsychronized span log fields. Use `self.logFields` setter & getter.
+    /// Unsychronized span log fields.  Must be accessed on `queue`.
     private var unsafeLogFields: [[String: Encodable]]
-    /// A collection of all log fields send for this span.
-    var logFields: [[String: Encodable]] {
-        ddTracer.queue.sync { unsafeLogFields }
-    }
-
-    /// Unsychronized span completion. Use `self.isFinished` setter & getter.
+    /// Unsychronized span completion.  Must be accessed on `queue`.
     private var unsafeIsFinished: Bool
-    private(set) var isFinished: Bool {
-        get { ddTracer.queue.sync { unsafeIsFinished } }
-        set { ddTracer.queue.async { self.unsafeIsFinished = newValue } }
-    }
 
     private var activityReference: ActivityReference?
 
@@ -58,8 +42,7 @@ internal class DDSpan: OTSpan {
         context: DDSpanContext,
         operationName: String,
         startTime: Date,
-        tags: [String: Encodable],
-        logFields: [[String: Encodable]] = []
+        tags: [String: Encodable]
     ) {
         self.ddTracer = tracer
         self.ddContext = context
@@ -67,10 +50,11 @@ internal class DDSpan: OTSpan {
         self.spanBuilder = tracer.spanBuilder
         self.spanOutput = tracer.spanOutput
         self.logOutput = tracer.logOutput
+        self.queue = ddTracer.queue // share the queue among all spans
         self.unsafeOperationName = operationName
         self.unsafeTags = tags
+        self.unsafeLogFields = []
         self.unsafeIsFinished = false
-        self.unsafeLogFields = logFields
     }
 
     // MARK: - Open Tracing interface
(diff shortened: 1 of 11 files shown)


Sources/Datadog/Tracing/DDSpan.swift
```diff
@@ -84,33 +68,35 @@ internal class DDSpan: OTSpan {
...
         }
     }
 
     func setBaggageItem(key: String, value: String) {
-        if warnIfFinished("setBaggageItem(key:value:)") {
-            return
+        let isFinished = queue.sync { self.warnIfFinished("setBaggageItem(key:value:)") }
+        if !isFinished {
+            // Baggage items must accessed outside `tracer.queue` as it uses that queue for internal sync.
+            ddContext.baggageItems.set(key: key, value: value)
```
ncreated, line 92: 💡 It is possible to synchronise `BaggageItems` access using `tracer.queue`, which would additionally clean up this code, leading to single `queue.async {}`. That would require small refactoring but I'm not sure if sharing single `tracer.queue` for all spans and baggage items is the best strategy. Alternatively, each `DDSpan` could work on its own `queue` and use it for baggage items synchronisation. The path to take highly depends on what do we optimise for: huge number of `DDSpans` (many queues) vs intensive use of `BaggageItems` (many accesses to single queue).

I stopped here, without going deeper in my refactoring, as the point of this PR is to clean up the interface of `DDSpan`. The solution in this PR already does less synchronization than before, so IMHO we can call it good enough and visit baggage items improvement on next occasion.

xgouchet (approved):
LGTM

Sources/Datadog/Tracing/DDSpan.swift
```diff
@@ -84,33 +68,35 @@ internal class DDSpan: OTSpan {
     }
 
     func setOperationName(_ operationName: String) {
-        if warnIfFinished("setOperationName(_:)") {
-            return
+        queue.async {
```
mariusc83, line 71: I wonder why do you need to run these operations in an Async Queue anyway ? Is it because it is an expensive operation under the hood ?

Sources/Datadog/Tracing/DDSpan.swift
```diff
@@ -123,38 +109,42 @@ internal class DDSpan: OTSpan {
...
         }
-        sendSpan(finishTime: time)
     }
 
     // MARK: - Writting SpanEvent
 
     /// Sends span event for given `DDSpan`.
     private func sendSpan(finishTime: Date) {
-        // Baggage items must be read before entering the `tracer.queue` as it uses that queue for internal sync.
+        // Baggage items must accessed outside `tracer.queue` as it uses that queue for internal sync.
```
mariusc83, line 140: `must be accessed` ?

buranmert (approved):

mariusc83 (approved):
LGTM but I have one question mark about all these `async` operations. Is this a best practice in Swift ?

Sources/Datadog/Tracing/DDSpan.swift
ncreated: It's because we access the shared state: `unsafeOperationName` property. As this is a public API and can be called from any thread, we need to synchronize it. When it comes to running operation on `queue`, there are two choices: `queue.sync {}` or `queue.async {}`. The prior is blocking the caller thread and the latter does not, so `async` will perform better for the user and do what needs to be done for us.

Speaking Android, I think this is use of `queue` is quite similar to Java `@synchronised` annotation - we use it as a locking mechanism. Although there are [few other choices available for iOS](https://dmytro-anokhin.medium.com/concurrency-in-swift-reader-writer-lock-4f255ae73422) (including raw POSIX locks), `queues` are well optimised and can be equally performant as other locking mechanisms.

Sources/Datadog/Tracing/DDSpan.swift
```diff
@@ -84,33 +68,35 @@ internal class DDSpan: OTSpan {
...
             self.unsafeTags[key] = value
         }
     }
 
     func setBaggageItem(key: String, value: String) {
-        if warnIfFinished("setBaggageItem(key:value:)") {
-            return
+        let isFinished = queue.sync { self.warnIfFinished("setBaggageItem(key:value:)") }
+        if !isFinished {
+            // Baggage items must accessed outside `tracer.queue` as it uses that queue for internal sync.
```
ncreated, line 91:
```suggestion
            // Baggage items must be accessed outside the `tracer.queue` as it uses that queue for internal sync.
```

Sources/Datadog/Tracing/DDSpan.swift
```diff
@@ -84,33 +68,35 @@ internal class DDSpan: OTSpan {
...
-        ddContext.baggageItems.set(key: key, value: value)
     }
 
     func baggageItem(withKey key: String) -> String? {
-        if warnIfFinished("baggageItem(withKey:)") {
-            return nil
-        }
-        return ddContext.baggageItems.get(key: key)
+        let isFinished = queue.sync { self.warnIfFinished("baggageItem(withKey:)") }
+        // Baggage items must accessed outside `tracer.queue` as it uses that queue for internal sync.
```
ncreated, line 98:
```suggestion
            // Baggage items must be accessed outside the `tracer.queue` as it uses that queue for internal sync.
```

Sources/Datadog/Tracing/DDSpan.swift
ncreated: Fixing also in few other places, thx 🙂 👌 
```suggestion
        // Baggage items must be accessed outside the `tracer.queue` as it uses that queue for internal sync.
```

PR #484, opened 2021-05-06
RUMM-1318 tagged Bitrise workflow is added

buranmert:
### What and why?

We'd like to make sure that every tagged commit is automatically pushed to Cocoapods trunk and whose pre-built binaries can be uploaded from its Github release.

### How?

This PR is part of automating the process mentioned above.

The whole process consists of
1. `Tag creation` webhook from Github
2. `Tag` triggered build in CI
3. Shipping via `make ship`

### Review checklist


bitrise.yml
```diff
@@ -237,3 +247,14 @@ workflows:
...
+    description: |-
+        Ships the tagged version to Cocoapods trunk and uploads binaries to Github.
+    steps:
+    - script:
+        title: Ship tagged version
+        inputs:
+        - content: |-
+            #!/usr/bin/env zsh
+            set -e
+            make ship
```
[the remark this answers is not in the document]
buranmert: Once you mentioned making `DatadogSDKObjc/AlamofireExtension` subspecs of `DatadogSDK` so that only one `podspec` will need to be pushed, i think that makes the most sense in this case as it solves the root cause.

we can arrange `DatadogSDK.podspec` in the following way:
```
s.default_subspec = 'Swift`

s.subspec 'Swift' do ... end

s.subspec 'Objc' do ... end
s.subspec 'AlamofireExtension' do ... end
```
this shouldn't be a breaking change for Objc users.
also, in the next `DatadogSDKObjc` version we can add warnings to the console to direct users to use `pod DatadogSDK/Objc` instead of `pod DatadogSDKObjc`.

does it make sense until here?
if so, i will verify that this won't be a breaking change for Swift users neither (they may need to import `DatadogSDKSwift` in their source code but i haven't verified yet)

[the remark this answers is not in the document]
ncreated: Sure, our `.podspec` layout is not ideal and it forces us to submit 3 specs, instead of one. This is where the issue originates - submitting 2nd and 3rd podspec requires waiting until CP processes the 1st one. It would be ideal to fix the root cause, agree.

However, changing to `subspecs` model will require refactoring in numerous Objc SDK files and taking additional decision for `AlamofireExtension` layout. Note that with CP subspecs the codedbase is not modeled as modules, but rather a compilation units with different sets of source files. The `import Datadog` will not work neither in `Objc` spec, nor in `AlamofireExtension`. We would need to make each import conditional with `#if !COCOAPODS` to make it work. In result, the modules layout for CP will be different than the one for SPM and Carthage.

Makefile
```diff
@@ -102,11 +102,28 @@ bump:
...
 		pod spec lint --allow-warnings DatadogSDK.podspec
 		pod trunk push --allow-warnings --synchronous DatadogSDK.podspec
-		pod repo update
 		./tools/standalone-binary-distro/make_distro_builds.sh
-		@echo "⚠️ DatadogSDKObjc.podspec needs to be tried after ~1 hour:"
-		@echo "pod spec lint --allow-warnings DatadogSDKObjc.podspec"
-		@echo "pod trunk push --allow-warnings DatadogSDKObjc.podspec"
+ifeq (${ci}, true)
+		curl https://app.bitrise.io/app/$$BITRISE_APP_SLUG/build/start.json \
+		--data '{"hook_info":{"type":"bitrise","build_trigger_token":"$$BITRISE_PERSONAL_ACCESS_TOKEN"},\
```
buranmert, line 108: 🎗️  this line may require some discussion in private

bitrise.yml
```diff
@@ -237,3 +247,14 @@ workflows:
...
 
+  ship:
+    description: |-
+        Ships the tagged version to Cocoapods trunk and uploads binaries to Github.
+    steps:
+    - script:
+        title: Ship tagged version
+        inputs:
+        - content: |-
+            #!/usr/bin/env zsh
```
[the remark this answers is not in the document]
buranmert: it's added to `make ship_part_2` 👍

Makefile
```diff
@@ -102,11 +102,28 @@ bump:
 ship:
 		pod spec lint --allow-warnings DatadogSDK.podspec
 		pod trunk push --allow-warnings --synchronous DatadogSDK.podspec
-		pod repo update
 		./tools/standalone-binary-distro/make_distro_builds.sh
-		@echo "⚠️ DatadogSDKObjc.podspec needs to be tried after ~1 hour:"
-		@echo "pod spec lint --allow-warnings DatadogSDKObjc.podspec"
-		@echo "pod trunk push --allow-warnings DatadogSDKObjc.podspec"
+ifeq (${ci}, true)
+		curl https://app.bitrise.io/app/$$BITRISE_APP_SLUG/build/start.json \
```
ncreated, line 107: These API is [now deprecated](https://discuss.bitrise.io/t/deprecation-of-build-trigger-api-on-www-subdomain-use-app-subdomain-or-the-public-api-instead/4721/2) and we should use [`POST /apps/{app-slug}/builds` instead](https://api-docs.bitrise.io/#/builds/build-trigger).

Makefile
```diff
@@ -102,11 +102,28 @@ bump:
...
+		--data '{"hook_info":{"type":"bitrise","build_trigger_token":"$$BITRISE_PERSONAL_ACCESS_TOKEN"},\
+		"build_params":{"workflow_id":"tagged_commit_part_2","commit_hash":"$$BITRISE_GIT_COMMIT"},\
+		"triggered_by":"curl"}'
+endif
+
+ship_part_2:
```
ncreated, line 113: I'm getting syntax error when trying to run this on my local:
```
❯ make ship_part_2
Makefile:112: *** missing separator.  Stop.
```

Makefile
```diff
@@ -102,11 +102,28 @@ bump:
...
+		pod trunk me
+		pod spec lint --allow-warnings DatadogSDKObjc.podspec
+		pod spec lint --allow-warnings DatadogSDKAlamofireExtension.podspec
+		objcRetVal=1 ; \
+		while [[ $$objcRetVal -ne 0 ]] ; do \
+			pod trunk push --allow-warnings DatadogSDKObjc.podspec ; \
+        	((objcRetVal = $$?)) ; \
```
ncreated, line 120: * I'm afraid that this may end up badly if `pod trunk` considers us DDOS-ing their endpoint. Adding `~5min` sleep for each attempt sounds reasonable IMO.
* Don't we need to `pod repo update` somewhere on the way?

Makefile
```diff
@@ -102,11 +102,28 @@ bump:
...
+    	done
+		alamofireRetVal=1 ; \
+		while [[ $$alamofireRetVal -ne 0 ]] ; do \
+			pod trunk push --allow-warnings DatadogSDKAlamofireExtension.podspec ; \
+        	((alamofireRetVal = $$?)) ; \
+    	done
```
ncreated, line 126: Does this unusual `done` indentation and line breaks (`\`) come from running the shell script inside `Makefile`? 

If yes, then it will be very hard and error-prone to maintain this script even for a small changes as no dry run can be performed for this automation. Let's at least move it to separate shell script, where the syntax will simpler.

It might be a clear candidate for `distribution` toolset, along with the `make_distro_builds.sh` tool:
```shell
/tools/distribution
    - make_distro_builds.sh
    - push_podspecs.sh
```

dependency-manager-tests/carthage/Makefile
```diff
@@ -1,7 +1,9 @@
-ifeq (${ci}, true)
-		GIT_BRANCH := ${BITRISE_GIT_BRANCH}
-else
-		GIT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD)
+GIT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD)
+ifneq (${BITRISE_GIT_BRANCH},)
+	GIT_BRANCH := ${BITRISE_GIT_BRANCH}
+endif
+ifneq (${BITRISE_GIT_TAG},)
+	GIT_BRANCH := ${BITRISE_GIT_TAG}
```
buranmert, line 6: `GIT_BRANCH` name is not totally accurate anymore but i couldn't find a better name

bitrise.yml
```diff
@@ -27,6 +27,21 @@ workflows:
...
 
+  tagged_commit:
+    after_run:
+    - _make_dependencies
+    - run_linter
+    - run_unit_tests
+    - run_integration_tests
+    - check_dependency_managers
+    - ship
+    - _deploy_artifacts
```
ncreated, line 38: Does this workflow report to GH Checks somehow? I'm wondering how we can add observability to it, ensuring we never overlook its failure. Calling the `_notify_failure_on_slack` in this workflow can be an idea.

dependency-manager-tests/carthage/Makefile
ncreated: nit: a "`GIT_REFERENCE`" could be more accurate.

ncreated (changes requested):
Looks good 👌, but we're missing an observability mechanism - I suggested a simple solution in comments.

bitrise.yml
```diff
@@ -27,6 +27,21 @@ workflows:
...
+
+  tagged_commit_part_2:
+    after_run:
+    - ship_part_2
+    - _deploy_artifacts
```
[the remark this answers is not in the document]
buranmert: added 👌 
build failures are reported via email by bitrise but it doesn't hurt to have one more notification 👍

ncreated (approved):

PR #487, opened 2021-05-10
RUMM-1353 Disable Crash Reporting feature in upcoming releases

ncreated:
### What and why?

⚙️ This PR removes crash reporting API from public interface and makes `DatadogCrashReporting.framework` artifact not available for upcoming release(s). We will make it available again once crash reporting is ready for pre-release.

### How?

My goal was to create setup which:
* would make the crash reporting feature not available in upcoming release(s),
* would be easy to revert once the crash reporting is ready,
* would let us still work on crash reporting using the main branch and existing CI setup,
* would let us keep dogfooding crash reporting in our own products,
* might scale for other experimental APIs in the future.

It was achieved in 2 steps:

**1/ Adding `DD_SDK_ENABLE_EXPERIMENTAL_APIS` swift compiler flag**

This compiler flag is used to conditionally remove crash reporting APIs from `Datadog` and `DatadogObjc` artifacts. As client applications won't define it, the public entry point to the CR feature won't be available.

This flag needs to be defined in our `Base.local.xcconfig` so that we can keep working on the feature as usual. This config is git-ignored, so I made addition of this flag a part of the `make dependencies` command. This way, the CI runner is provisioned with the flag and can still build unit and integration tests for crash reporting.

**2/ Removing `DatadogCrashReporting` product from dependency managers**

* CocoaPods - no change was required as `DatadogCrashReporting.podspec` is not created,
* SPM - I removed the `DatadogCrashReporting` product from `Package.swift` descriptor,
* Carthage - I removed `DatadogCrashReporting.xcscheme` from Xcode shared schemes (so Carthage won't produce its artifact). I also added `enable-crash-reporting.sh` patch to `make dependencies`, so that this scheme is installed as non-shared Xcode scheme on our local machines and on CI.

With this setup we can keep dogfooding crash reporting in our own products. Once this is merged, I will revert the change to `Package.swift` and define `DD_SDK_ENABLE_EXPERIMENTAL_APIS` flag on `dogfooding` branch.

### Review checklist


tools/standalone-binary-distro/make_distro_builds.sh
```diff
@@ -28,17 +28,21 @@ prepare() {
    echo "BUILD_LIBRARY_FOR_DISTRIBUTION = YES" >> $xcconfig
    # carthage may try to build schemes in dependency-manager-tests folder, so we hide it temporarily
    mv dependency-manager-tests/ .dependency-manager-tests/ 
+   # hide '*.local.xcconfigs' so they are not applied to produced artifacts (they are renamed to '*_bak'):
+   find . -type f -name '*.local.xcconfig' -exec mv '{}' '{}_bak' ';'
```
ncreated, line 32: 💡 This change should be originally done to the `make_distro_builds.sh` to not apply local & git-ignored configuration to produced XCFrameworks. I'm adding it in this PR, as now it's even more necessary.

IMO it's good for now, but in general I see this patching approach very inconvenient - we need to patch repository before exporting XCFrameworks and then revert patches again to reset it to the original state. IMO it would be way easier and safer to export XCFrameworks from new clone of the repository, instead of current working directory. It's easy to imagine that some setup might be overlooked (like these `*.local.xcconfig`) leading to exporting bad artifacts.

buranmert (approved):

PR #489, opened 2021-05-10
RUMM-1320 Let user extra attributes use the usr prefix

xgouchet:
### What and why?

To align the data with the browser SDK, and make the infos more consistent, we keep all the usr custom attributes in the root `usr` object.


buranmert (commented):
i think now `AttributesSanitizer.Constraints.maxNestedLevelsInAttributeName` for `userInfo.attributes` should be 9 instead of 8, wdyt @ncreated ?

buranmert (approved):

PR #495, opened 2021-05-21
RUMM-1274 Implement Memory reader

xgouchet:
### What and why?

Implement a class to read the Resident Memory used by the current application

### How?

This uses the [`task_vm_info`](https://opensource.apple.com/source/xnu/xnu-4570.41.2/osfmk/mach/task_info.h.auto.html) struct to read the relevant memory information

### Review checklist

### DIFF
--- a/Datadog/Datadog.xcodeproj/project.pbxproj
+++ b/Datadog/Datadog.xcodeproj/project.pbxproj
@@ -439,6 +439,9 @@
 		9EEA4871258B76A100EBDA9D /* Global+objc.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9EEA4870258B76A100EBDA9D /* Global+objc.swift */; };
 		9EF963E82537556300235F98 /* DDURLSessionDelegateAsSuperclassTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9EF963E72537556300235F98 /* DDURLSessionDelegateAsSuperclassTests.swift */; };
 		9EFD112C24B32D29003A1A2B /* FirstPartyURLsFilter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9EFD112B24B32D29003A1A2B /* FirstPartyURLsFilter.swift */; };
+		B3BBBCB2265E71C700943419 /* VitalMemoryReader.swift in Sources */ = {isa = PBXBuildFile; fileRef = B3BBBCB0265E71C600943419 /* VitalMemoryReader.swift */; };
+		B3BBBCB3265E71C700943419 /* VitalReader.swift in Sources */ = {isa = PBXBuildFile; fileRef = B3BBBCB1265E71C600943419 /* VitalReader.swift */; };
+		B3BBBCBC265E71D100943419 /* VitalMemoryReaderTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = B3BBBCBB265E71D100943419 /* VitalMemoryReaderTest.swift */; };
 		B3FC3C0926526F0000DEED9E /* VitalInfo.swift in Sources */ = {isa = PBXBuildFile; fileRef = B3FC3C0626526EFF00DEED9E /* VitalInfo.swift */; };
 		B3FC3C0A26526F0000DEED9E /* VitalObserver.swift in Sources */ = {isa = PBXBuildFile; fileRef = B3FC3C0726526F0000DEED9E /* VitalObserver.swift */; };
 		B3FC3C332652AE1400DEED9E /* VitalListener.swift in Sources */ = {isa = PBXBuildFile; fileRef = B3FC3C322652AE1400DEED9E /* VitalListener.swift */; };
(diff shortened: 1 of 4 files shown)


Sources/Datadog/RUM/RUMVitals/VitalMemoryReader.swift
```diff
@@ -0,0 +1,33 @@
...
+                    mach_task_self_,
+                    task_flavor_t(TASK_VM_INFO),
+                    $0,
+                    &vmInfoSize
+                )
+            }
+        }
+
+        if kern == KERN_SUCCESS {
+            return Double(vmInfo.resident_size)
```
buranmert, line 28: does that give how much memory is being used only? does that provide value to the user? shouldn't we provide more context such as max memory/memory utilization % ?

Datadog/Datadog.xcodeproj/project.pbxproj
```diff
@@ -2858,6 +2864,21 @@
...
+		};
+		B3FC3C1226526F4100DEED9E /* RUMVitals */ = {
+			isa = PBXGroup;
+			children = (
+				B3FC3C3B2653A97700DEED9E /* VitalObserverTest.swift */,
+				B3ACF8B92656644000738CEE /* VitalMemoryReaderTest.swift */,
+			);
+			path = RUMVitals;
+			sourceTree = "<group>";
+		};
```
ncreated, line 2881: There is sth wrong around this lines and it malforms the `pbxproj` file:
```
--- xcodebuild: WARNING: Unable to open project file '/Users/vagrant/git/Datadog/Datadog.xcodeproj' in workspace '/Users/vagrant/git/Datadog.xcworkspace'.
```
You may want to reset changes in `pbxproj`, do rebase and then add `Vital*` files once again.

Sources/Datadog/RUM/RUMVitals/VitalMemoryReader.swift
```diff
@@ -0,0 +1,33 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+internal class VitalMemoryReader: VitalReader {
+    static let TASK_VM_INFO_COUNT = MemoryLayout<task_vm_info>.size / MemoryLayout<natural_t>.size
```
ncreated, line 10: We don't use uppercase for defining constants - they are reserved for compiler directives. Also, seems that this one can be private:
```swift
private static let task_vm_info_count = MemoryLayout<task_vm_info>.size / MemoryLayout<natural_t>.size
```

Sources/Datadog/RUM/RUMVitals/VitalMemoryReader.swift
```diff
@@ -0,0 +1,33 @@
...
+
+    func readVitalData() -> Double? {
```
ncreated, line 12: The return value seems to be "used memory in bytes", so having `UInt64` would be a cleaner type for it, no? Also, I'm missing some comment either for this method or the entire class, that would explain which memory metric do we actually capture - it's quite hard to guess it from the low-level APIs below 🙂.

Tests/DatadogTests/Datadog/RUM/RUMVitals/VitalMemoryReaderTest.swift
```diff
@@ -0,0 +1,19 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import XCTest
+@testable import Datadog
+
+internal class VitalMemoryReaderTest: XCTestCase {
+    /// This is definitely not an ideal test, but I'm out of my depth on how to mock the `task_info` method
```
ncreated, line 11: How about testing its property?
```swift
func testWhenMemoryConsumptionGrows() {
   // Given
   let reader = VitalMemoryReader()
   threshold = reader.readVitalData()

   // When
   allocateHeapMemory(size: 100 * 1_024)

   // Then
   measure = reader.readVitalData()
   XCTAssertGreatherThan(measure, threshold)
}

func testWhenMemoryConsumptionDecreases() {
   // ...
   XCTAssertGreatherThan(threshold, measure)
}
```

ncreated (changes requested):

Sources/Datadog/RUM/RUMVitals/VitalMemoryReader.swift
ncreated: Sounds fair 👍

ncreated (approved):
Looks good to go 🚀👌.

Tests/DatadogTests/Datadog/RUM/RUMVitals/VitalMemoryReaderTest.swift
```diff
@@ -0,0 +1,60 @@
...
+        let reader = VitalMemoryReader()
+        let threshold = reader.readVitalData()
+        let allocatedSize = 128 * 1_024
+
+        // When
+        let intPointer = UnsafeMutablePointer<Int>.allocate(capacity: allocatedSize)
+        for i in 0..<allocatedSize {
+            (intPointer + i).initialize(to: i)
+        }
+        let measure = reader.readVitalData()
```
mariusc83, line 30: The way I read from the docs about the `UnsafeMutablePointer` the `capacity` should be the number of memory blocks specific for that type. The way I understand this is that the memory consumed should be `allocatedSize * 32`, isn't this correct ?

Sources/Datadog/RUM/RUMVitals/VitalMemoryReader.swift
xgouchet: We can look into knowing how much the device is capable in a future version, but for now just the size used should already give some info

buranmert (approved):

PR #498, opened 2021-05-26
RUMM-1381 Add internal API to send buffered data before deinitializing SDK

ncreated:
### What and why?

⚙️ This PR adds an internal API to flush all authorised data and deinitialize the SDK:
```swift
#if DD_SDK_COMPILED_FOR_TESTING

public static func flushAndDeinitialize() { /* ... */ }

#endif
```

It executes synchronously, so upon return it guarantees that:
* the SDK instance is destroyed,
* all asynchronous work is completed in `storage` and `upload` stacks,
* all authorised data is uploaded arbitrarily (without retrying on failure).

This API is required for E2E δ project (#496) and will be used in `E2ETests` module for tearing down the SDK in each unit test.

### How?

This is highly experimental API, so it is conditioned by `DD_SDK_COMPILED_FOR_TESTING` compiler flag and is not planned for visibility in public releases. This flag needs to be defined in `Base.local.xcconfig` - requires `make dependencies`.

The implementation of `flushAndDeinitialize()` executes (synchronously) a chain of commands for deinitializing all active components in each feature. This chain is mirroring the components' initialization chain:

**Initialization chain - on `Datadog.initialize()`:**
* for each feature (Logging, Tracing, RUM):
   * initialize `FeatureStorage`
   * initialize `FeatureUpload`
      * start upload worker
* for each instrumentation (`UIViewController`, `UIApplication`, `URLSession`):
   * swizzle instrumented APIs

**Deinitialization chain - on `Datadog. flushAndDeinitialize()`:**
* for each feature (Logging, Tracing, RUM):
   * deinitialize `FeatureStorage`
      * wait for pending async writes completion
   * deinitialize `FeatureUpload`
      * cancel upload worker
      * upload unsent authorised data
* for each instrumentation (`UIViewController`, `UIApplication`, `URLSession`):
   * unswizzle instrumented APIs

---

Given this more strict control over features deinitialization, I also replaced all arbitrary deallocations (`.instance = nil`) with the new explicit APIs (`.instance.deinitialize()`). This revealed the source of recent tests flakiness that we observe on CI and in local. To further fix this flakiness, I improved the implementation of `ServerMock`, so it's also more strict and explicit.

The `ServerMock` instance which we create in unit tests intercepts the `URLSession` by installing `URLProtocol`. The problem is that we don't control the initialization of `URLProtocol` subclass - it is performed behind the scene. This is why the `ServerMockProtocol` implementation needs to obtain `ServerMock` reference through static global `.activeServer`.

We had an issue in few tests, where the `.activeServer` reference was leaked from one test to another, causing some asynchronous request completions to be reported to the new instance. I've fixed that by:
* repairing that tests,
* ensuring that `ServerMockProtocol` always captures the right  instance of `ServerMock` (tests will fail otherwise).

This should solve some of the existing flakiness and help us adding new tests by controlling their consistency. For verification, I've run unit tests 40 times on CI without observing any failure 🏆.

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/Feature.swift
+++ b/Sources/Datadog/Core/Feature.swift
@@ -40,7 +40,7 @@ internal struct FeatureStorage {
 
     /// An arbitrary `Writer` which always writes data to authorized folder.
     /// Should be only used by components which implement their own consideration of the `TrackingConsent` value
-    /// associated with data written.
+    /// associated with data written (e.g. crash reporting integration which saves the consent value along with the crash report).
     let arbitraryAuthorizedWriter: Writer
 
     init(
@@ -120,6 +120,19 @@ internal struct FeatureStorage {
         self.reader = reader
         self.arbitraryAuthorizedWriter = arbitraryAuthorizedWriter
     }
+
+#if DD_SDK_COMPILED_FOR_TESTING
+    /// Flushes all async write operations.
+    ///
+    /// This method is executed synchronously. After return, the storage feature has no more
+    /// pending asynchronous write operations so all its data is ready for upload.
+    func flush() {
+        let consentAwareDataWriter = writer as? ConsentAwareDataWriter
+        let arbitraryDataWriter = arbitraryAuthorizedWriter as? ArbitraryDataWriter
+        let readWriteQueue = consentAwareDataWriter?.readWriteQueue ?? arbitraryDataWriter?.readWriteQueue
+        readWriteQueue?.sync {}
+    }
+#endif
 }
 
 internal struct FeatureUpload {
@@ -166,4 +179,18 @@ internal struct FeatureUpload {
     init(uploader: DataUploadWorkerType) {
         self.uploader = uploader
     }
+
+#if DD_SDK_COMPILED_FOR_TESTING
+    /// Flushes all authorised data and tears down the upload stack.
+    /// - It flushes all data stored in authorized files by performing their arbitrary upload (without retrying).
+    /// - It completes all pending asynchronous work in upload worker and cancels its next schedules.
+    ///
+    /// This method is executed synchronously. After return, the upload feature has no more
+    /// pending asynchronous operations and all its authorized data should be considered uploaded.
+    func flushAndTearDown() {
+        let dataUploadWorker = uploader as? DataUploadWorker
+        dataUploadWorker?.cancelSynchronously()
+        dataUploadWorker?.flushSynchronously()
+    }
+#endif
 }
--- a/Sources/Datadog/Core/Feature.swift
+++ b/Sources/Datadog/Core/Feature.swift
@@ -34,14 +34,14 @@ internal struct FeaturesCommonDependencies {
 internal struct FeatureStorage {
     /// Writes data to files. This `Writer` takes current value of the `TrackingConsent` into consideration
     /// to decided if the data should be written to authorized or unauthorized folder.
-    let writer: Writer
+    let writer: AsyncWriter
     /// Reads data from files in authorized folder.
-    let reader: Reader
+    let reader: SyncReader
 
     /// An arbitrary `Writer` which always writes data to authorized folder.
     /// Should be only used by components which implement their own consideration of the `TrackingConsent` value
     /// associated with data written (e.g. crash reporting integration which saves the consent value along with the crash report).
-    let arbitraryAuthorizedWriter: Writer
+    let arbitraryAuthorizedWriter: AsyncWriter
 
     init(
         featureName: String,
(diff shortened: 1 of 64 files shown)


Tests/DatadogTests/Datadog/Core/Upload/DataUploadWorkerTests.swift
```diff
@@ -35,159 +35,219 @@ class DataUploadWorkerTests: XCTestCase {
...
+        let server = ServerMock(delivery: .success(response: .mockResponseWith(statusCode: 200)))
         let dataUploader = DataUploader(
             urlProvider: .mockAny(),
-            httpClient: HTTPClient(session: .serverMockURLSession),
+            httpClient: HTTPClient(session: server.getInterceptedURLSession()),
             httpHeaders: .mockAny()
         )
-        // Start logs uploader
-        withExtendedLifetime([
-            ServerMock(delivery: .success(response: .mockResponseWith(statusCode: 200))),
```
ncreated, line 88: This was an example of broken test - the `ServerMock` was created, the global `.serverMockURLSession` was used in `DataUploader` (diff above), but no `_ = server.waitAndReturnRequests()` method was called. As this test was not awaiting for async requests, their completion was sometimes arriving in another test that was awaiting for its requests.

Sources/Datadog/Core/Feature.swift
```diff
@@ -120,6 +120,19 @@ internal struct FeatureStorage {
...
+#if DD_SDK_COMPILED_FOR_TESTING
+    /// Flushes all async write operations.
+    ///
+    /// This method is executed synchronously. After return, the storage feature has no more
+    /// pending asynchronous write operations so all its data is ready for upload.
+    func flush() {
+        let consentAwareDataWriter = writer as? ConsentAwareDataWriter
+        let arbitraryDataWriter = arbitraryAuthorizedWriter as? ArbitraryDataWriter
+        let readWriteQueue = consentAwareDataWriter?.readWriteQueue ?? arbitraryDataWriter?.readWriteQueue
+        readWriteQueue?.sync {}
```
mariusc83, line 133: I don't see where you are stopping this, I guess you are actually stopping the feature. This means the feature needs to be stopped first to avoid having a new event in the pipeline that could happen to bot be stored by the time your process exits.

Sources/Datadog/Core/Feature.swift
```diff
@@ -120,6 +120,19 @@ internal struct FeatureStorage {
         self.reader = reader
         self.arbitraryAuthorizedWriter = arbitraryAuthorizedWriter
     }
+
+#if DD_SDK_COMPILED_FOR_TESTING
+    /// Flushes all async write operations.
+    ///
+    /// This method is executed synchronously. After return, the storage feature has no more
+    /// pending asynchronous write operations so all its data is ready for upload.
+    func flush() {
```
buranmert, line 129: `flush()` is not `public`, why do we wrap it with `#if DD_... #endif`?

Sources/Datadog/Datadog.swift
```diff
@@ -288,31 +288,36 @@ public class Datadog {
...
-        }
-
-        // First, reset internal loggers:
+#if DD_SDK_COMPILED_FOR_TESTING
+    /// Flushes all authorised data for each feature, tears down and deinitializes the SDK.
+    /// - It flushes all data authorised for each feature by performing its arbitrary upload (without retrying).
+    /// - It completes all pending asynchronous work in each feature.
+    ///
+    /// This is highly experimental API and only supported in tests.
+    public static func flushAndDeinitialize() {
```
buranmert, line 297: alternatively, you can keep this method internal and call it from outside via notifications
```swift
init() {
  NotificationCenter.default.addObserver("DD.flush", self, #selector(flushAndDeinit))
}
// in tests
NotificationCenter.postNotif("DD.flush")
```

Sources/Datadog/Core/Feature.swift
ncreated: You're right 👍. I made the assumption that no data will be written (through logger, tracer or RUM monitor) after deinitializing the SDK. It doesn't need to be true, so I took more time on this PR and updated it in https://github.com/DataDog/dd-sdk-ios/pull/498/commits/9e0d0e4fdf3ad67069da04f916e8b780caad64a7 (👀 🙏).

Now the "storage" is deinitialized by:
```swift
writer.flushAndCancelSynchronously()
arbitraryAuthorizedWriter.flushAndCancelSynchronously()
reader.markAllFilesAsReadable()
```
which:
- sets the `isCancelled` flag in both async writers, making them act as no-op,
- sets the `ignoreFilesAgeWhenReading` flag in files orchestrator to skip file age condition for reading.

This way on `.deinitialize()` no more data can be saved, but everything can be uploaded. The consistency is additionally checked with `asserts()` for strict control in tests and no flakiness. Everything is conditionally compiled only for tests - no logic change in production code.

Sources/Datadog/Core/Feature.swift
ncreated: To avoid using `flush()` and `deinitialize()` APIs in production code. These features are highly experimental and work with few arbitrary assumptions, so using them for serving any other logic in production code would be dangerous.

Sources/Datadog/Datadog.swift
ncreated: Notifications are made for broadcasting. Direct method call is IMO way cleaner and easier to read. 

In both cases, we would need to use conditional compilation for this experimental `deinit()` code - not only from the the public interface, but also for internals sanity.

buranmert (approved):

mariusc83 (approved):

PR #504, opened 2021-05-28
RUMM-1265 Correctly handle RUM events when app in background

mariusc83:
### What and why?

Currently if the application is not in foreground (there is no active View scope), any resource or action event will be ignored as there is no active `View` on which to attach it.
In this **PR** we are adding a change in the way we handle the received commands in the `RUMSessionScope` in order to be able to record RUM events while the application is in background.

### How?

We are introducing the concept of `RUM background View scope` which by definition is a normal View scope with predefined `name` and `url`. Once the `RUMSessionScope` needs to handle a `RUMStartAction` or `RUMStartResource` command and there is no active `RUMViewScope`, instead of ignoring the command as before it will automatically create a special `RUMViewScope` on which the received resource or action will be attached. For more information please check: [RUM Background events RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/1385267440/RFC+-+Background+Mobile+Events)

### Review checklist

### DIFF
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
@@ -7,6 +7,11 @@
 import Foundation
 
 internal class RUMViewScope: RUMScope, RUMContextProvider {
+    struct Constants {
+        static let backgroundViewURL = "com/datadog/background/view"
+        static let backgroundViewName = "Background"
+    }
+
     // MARK: - Child Scopes
 
     /// Active Resource scopes, keyed by .resourceKey.
@@ -33,7 +38,7 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
     /// The name of this View, used as the `VIEW NAME` in RUM Explorer.
     let viewName: String
     /// The start time of this View.
-    private let viewStartTime: Date
+    let viewStartTime: Date
     /// Date correction to server time.
     private let dateCorrection: DateCorrection
     /// Tells if this View is the active one.
@@ -129,7 +134,6 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
         case let command as RUMStopViewCommand where identity.equals(command.identity):
             isActiveView = false
             needsViewUpdate = true
-
         case let command as RUMAddViewTimingCommand where isActiveView:
             customTimings[command.timingName] = command.time.timeIntervalSince(viewStartTime).toInt64Nanoseconds
             needsViewUpdate = true
--- a/Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScopeTests.swift
+++ b/Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScopeTests.swift
@@ -8,6 +8,8 @@ import XCTest
 @testable import Datadog
 
 class RUMSessionScopeTests: XCTestCase {
+    // MARK: - Unit Tests
+
     func testDefaultContext() {
         let parent: RUMApplicationScope = .mockWith(rumApplicationID: "rum-123")
         let scope = RUMSessionScope(parent: parent, dependencies: .mockAny(), samplingRate: 100, startTime: .mockAny())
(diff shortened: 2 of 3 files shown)


Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -7,6 +7,9 @@
 import Foundation
 
 internal class RUMViewScope: RUMScope, RUMContextProvider {
+    // MARK: - Background Scope
+    static let RUM_BACKGROUND_VIEW_URL = "com/datadog/background/view"
+    static let RUM_BACKGROUND_VIEW_NAME = "Background"
```
ncreated, line 12: ✏️ We don't use capitals for defining constants in Swift. Also, in `dd-sdk-ios` we adopt the convention of grouping constants in `Constants` struct acting as a namespace:
```swift
internal class RUMViewScope: RUMScope, RUMContextProvider {
   struct Constants {
      static let backgroundViewURL = "com/datadog/background/view"
      static let backgroundViewName = "Background"
   }
   
   // ...
}
```

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScopeTests.swift
```diff
@@ -77,6 +79,66 @@ class RUMSessionScopeTests: XCTestCase {
...
+
+        _ = scope.process(command: RUMAddUserActionCommand.mockWith(time: currentTime))
+
+        XCTAssertEqual(scope.viewScopes.count,1)
+        XCTAssertEqual(scope.viewScopes[0].viewStartTime, currentTime)
+        XCTAssertEqual(scope.viewScopes[0].viewName, RUMViewScope.RUM_BACKGROUND_VIEW_NAME)
+        XCTAssertEqual(scope.viewScopes[0].viewPath, RUMViewScope.RUM_BACKGROUND_VIEW_URL)
+    }
+
+    func testWhenActiveViewScope_andReceivingStartCommand_doesNotCreateNewViewScope() {
```
ncreated, line 124: ✏️ missing "it":
```suggestion
    func testWhenActiveViewScope_andReceivingStartCommand_itDoesNotCreateNewViewScope() {
```

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScopeTests.swift
```diff
@@ -77,6 +79,66 @@ class RUMSessionScopeTests: XCTestCase {
...
+        let parent = RUMContextProviderMock()
+        let currentTime = Date()
+
+        let scope = RUMSessionScope(parent: parent, dependencies: .mockAny(), samplingRate: 100, startTime: Date())
+        _ = scope.process(command: RUMStartViewCommand.mockAny())
+        _ = scope.process(command: RUMAddUserActionCommand.mockWith(time: currentTime))
+        XCTAssertEqual(scope.viewScopes.count, 1)
+    }
+
+    func testWhenNoActiveViewScope_andReceivingNotValidStartCommand_doesNotCreateNewViewScope() {
```
ncreated, line 134: ✏️ missing "it":
```suggestion
    func testWhenNoActiveViewScope_andReceivingNotValidStartCommand_itDoesNotCreateNewViewScope() {
```

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScopeTests.swift
```diff
@@ -77,6 +79,66 @@ class RUMSessionScopeTests: XCTestCase {
...
+        XCTAssertEqual(scope.viewScopes[0].viewPath, RUMViewScope.RUM_BACKGROUND_VIEW_URL)
+    }
+
+    func testWhenActiveViewScope_andReceivingStartCommand_doesNotCreateNewViewScope() {
+        let parent = RUMContextProviderMock()
+        let currentTime = Date()
+
+        let scope = RUMSessionScope(parent: parent, dependencies: .mockAny(), samplingRate: 100, startTime: Date())
+        _ = scope.process(command: RUMStartViewCommand.mockAny())
+        _ = scope.process(command: RUMAddUserActionCommand.mockWith(time: currentTime))
```
ncreated, line 130: I think we want to call `generateRandomValidStartCommand()` here, no?

ncreated (approved):
Looks very good 💪! I left few convention feedbacks. 

- [ ] Also, please update the PR description 🙏, especially on "how" - as we implement a very custom way of determining if the app is in background (notably, without the [app's life cycle events](https://developer.apple.com/documentation/uikit/app_and_environment/managing_your_app_s_life_cycle)) I think it's worth explaining it for future references.

ncreated:
💯 👌

PR #509, opened 2021-06-02
RUMM-1339: Keep custom user action even if there is active action already

0xnm:
### What and why?

iOS implementation of the change done in https://github.com/DataDog/dd-sdk-android/pull/614. This allows to send custom instant (discrete) user actions even if there is an active one. Solves the issue only for the case of `addUserAction` usage, case of `startUserAction/stopUserAction` is still unsolved.

### Review checklist

### DIFF
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
@@ -150,6 +150,9 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
         case let command as RUMAddUserActionCommand where isActiveView:
             if userActionScope == nil {
                 addDiscreteUserAction(on: command)
+            } else if command.actionType == .custom {
+                // still let it go, just instantly without any dependencies
+                sendDiscreteCustomUserAction(on: command)
             }
 
         // Error command
@@ -227,8 +230,8 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
         )
     }
 
-    private func addDiscreteUserAction(on command: RUMAddUserActionCommand) {
-        userActionScope = RUMUserActionScope(
+    private func createDiscreteUserActionScope(on command: RUMAddUserActionCommand) -> RUMUserActionScope {
+        return RUMUserActionScope(
             parent: self,
             dependencies: dependencies,
             name: command.name,
@@ -241,6 +244,22 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
                 self?.actionsCount += 1
                 self?.needsViewUpdate = true
             }
+        )
+    }
+
+    private func addDiscreteUserAction(on command: RUMAddUserActionCommand) {
+        userActionScope = createDiscreteUserActionScope(on: command)
+    }
+
+    private func sendDiscreteCustomUserAction(on command: RUMAddUserActionCommand) {
+        let customActionScope = createDiscreteUserActionScope(on: command)
+        _ = customActionScope.process(
+            command: RUMStopUserActionCommand(
+                                    time: command.time,
+                                    attributes: [:],
+                                    actionType: .custom,
+                                    name: nil
+            )
         )
     }
 
--- a/Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMViewScopeTests.swift
+++ b/Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMViewScopeTests.swift
@@ -448,7 +448,12 @@ class RUMViewScopeTests: XCTestCase {
         XCTAssertTrue(
             scope.process(command: RUMStartUserActionCommand.mockWith(actionType: .swipe, name: .mockRandom()))
         )
-        XCTAssertEqual(scope.userActionScope?.name, actionName, "View should ignore the next UA if one is pending.")
+        XCTAssertEqual(scope.userActionScope?.name, actionName, "View should ignore the next (only non-custom) UA if one is pending.")
+
+        XCTAssertTrue(
+            scope.process(command: RUMAddUserActionCommand.mockWith(actionType: .custom, name: .mockRandom()))
+        )
+        XCTAssertEqual(scope.userActionScope?.name, actionName, "View should not change existing pending action when adding custom UA (but this custom action should be recorded anyway).")
 
         XCTAssertTrue(
             scope.process(command: RUMStopUserActionCommand.mockWith(actionType: .swipe))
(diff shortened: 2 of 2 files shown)


Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -244,6 +247,22 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
         )
     }
 
+    private func addDiscreteUserAction(on command: RUMAddUserActionCommand) {
+        userActionScope = createDiscreteUserActionScope(on: command)
+    }
+
+    private func sendDiscreteCustomUserAction(on command: RUMAddUserActionCommand) {
```
0xnm, line 254: not sure about method name though considering codebase terminology

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -150,6 +150,9 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
         case let command as RUMAddUserActionCommand where isActiveView:
             if userActionScope == nil {
                 addDiscreteUserAction(on: command)
+            } else if command.actionType == .custom {
+                // still let it go, just instantly without any dependencies
+                sendDiscreteCustomUserAction(on: command)
             }
```
ncreated, line 156: I see this is also how we do it in `dd-sdk-android` 👌, but I'm wondering if this shouldn't be:
```swift
if command.actionType == .custom {
   sendDiscreteCustomUserAction(on: command)
} else if userActionScope == nil {
   addDiscreteUserAction(on: command)
}
```
?

This will uniformly treat all `.custom` actions as instant and send them right away. With the current implementation some `.custom` actions are managed on `userActionScope` (so have some duration) and some other do not (so have `0ns` duration). WDYT?

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -244,6 +247,22 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
         )
     }
 
+    private func addDiscreteUserAction(on command: RUMAddUserActionCommand) {
+        userActionScope = createDiscreteUserActionScope(on: command)
+    }
+
+    private func sendDiscreteCustomUserAction(on command: RUMAddUserActionCommand) {
+        let customActionScope = createDiscreteUserActionScope(on: command)
+        _ = customActionScope.process(
```
ncreated, line 256: This is hacky solution  but super simple 👍  - we create an unmanaged scope just for the sake of sending it a known command because we know what will happen internally (and if the internals change, this may break). It is also the very first place where we create a `RUMCommand` internally in scopes hierarchy - all other use cases that I know base on the command being consistently delivered from `RUMMonitor`. OK to me for now 👌 but let's keep an eye on this so it doesn't leak in the future.

More consistent approach would be to pin this scope on `self.customActionScope` and manage it in `self.process()`. That would require changing the `RUMUserActionScope` implementation so it can expire with `0ns` duration.

ncreated (commented):
Looks good 👌, but I left few questions.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -244,6 +247,22 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
...
+            command: RUMStopUserActionCommand(
+                                    time: command.time,
```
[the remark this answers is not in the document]
0xnm: I would say yes, but on Android it will be `1ns` (not possible to have 0), so I have to check this in web UI to confirm that `0` works fine, otherwise we use `1ns`. Anyway, the idea is to have instant action, without any duration (hence no dependencies possible).

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
0xnm: On Android things are a bit different, there is no dedicated `AddCustomUserAction` command - instead there is `StartAction` with `waitForStop: Boolean` flag. In this case action will be sent only if in 100ms there was no activity. But I guess it is possible to have resource call in this 100ms, which makes an action added via `addAction` to have a dependency, so we cannot send it instantly.

So I'm not sure if we want to change this logic right now, instead current change is a bit safer and has less impact and just kind of `Plan B` if main branch didn't work.

I would say from the user point of view they would like to see resources attached to their custom actions, so it is better to keep it like that in my opinion.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
0xnm: Indeed it is a hack with minimal impact. Also it is possible to send build `ActionEvent` directly from `ViewScope` and send it to `writer`, but it is also a leak/duplication of the logic from `ActionScope` which is not nice.

Do you suggest to implement pinning to the `self.customActionScope` as part of this PR?

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMViewScopeTests.swift
```diff
@@ -445,11 +445,18 @@ class RUMViewScopeTests: XCTestCase {
         XCTAssertNotNil(scope.userActionScope)
         XCTAssertEqual(scope.userActionScope?.name, actionName)
 
+        // add non-custom action while there is another action pending
         XCTAssertTrue(
             scope.process(command: RUMStartUserActionCommand.mockWith(actionType: .swipe, name: .mockRandom()))
         )
         XCTAssertEqual(scope.userActionScope?.name, actionName, "View should ignore the next UA if one is pending.")
 
+        // add custom instant action while there is another action is pending - this should be handled without any change for active action scope
```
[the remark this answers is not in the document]
0xnm: good suggestion. yes, probably having it in the assert message is better.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
ncreated: Not at all, as you say this is minimal impact - good to go IMO. We may want to use sth more complex only when we have a good reason.

ncreated (approved):
Thanks @0xnm !

buranmert (approved):

PR #514, opened 2021-06-04
RUMM-1276 VitalCPUReader is implemented

buranmert:
### What and why?

`VitalCPUReader` is implemented in order to measure CPU utilization of the app.

### How?

`host_statistics(...)` method with `HOST_CPU_LOAD_INFO` parameter returns CPU utilization data from the kernel.
This low-level API requires working with pointers and type conversions; however
1. there is no higher level API to get such information
2. as this is the kernel, hopefully it's likely that they don't change often.

### TODO

- [x] ~`testWhenCPUUnderHeavyLoad` test case fails quite often, this must be fixed first~ kernel fails to report too small intervals, for loop counts increased.

### Review checklist

### DIFF
--- a/Datadog/Datadog.xcodeproj/project.pbxproj
+++ b/Datadog/Datadog.xcodeproj/project.pbxproj
@@ -457,6 +457,8 @@
 		9E68FB55244707FD0013A8AA /* ObjcExceptionHandler.m in Sources */ = {isa = PBXBuildFile; fileRef = 9E68FB53244707FD0013A8AA /* ObjcExceptionHandler.m */; };
 		9E68FB56244707FD0013A8AA /* ObjcExceptionHandler.h in Headers */ = {isa = PBXBuildFile; fileRef = 9E68FB54244707FD0013A8AA /* ObjcExceptionHandler.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		9E989A4225F640D100235FC3 /* AppStateListenerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9E989A4125F640D100235FC3 /* AppStateListenerTests.swift */; };
+		9EC8B5DA2668197B000F7529 /* VitalCPUReader.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9EC8B5D92668197B000F7529 /* VitalCPUReader.swift */; };
+		9EC8B5EE2668E4DB000F7529 /* VitalCPUReaderTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9EC8B5ED2668E4DB000F7529 /* VitalCPUReaderTest.swift */; };
 		9ED6A6B425F2901800CB2E29 /* AppStateListener.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9ED6A6B325F2901800CB2E29 /* AppStateListener.swift */; };
 		9EE5AD8226205B82001E699E /* DDNSURLSessionDelegateTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9EE5AD8126205B82001E699E /* DDNSURLSessionDelegateTests.swift */; };
 		9EEA4871258B76A100EBDA9D /* Global+objc.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9EEA4870258B76A100EBDA9D /* Global+objc.swift */; };
@@ -1068,6 +1070,8 @@
 		9E68FB54244707FD0013A8AA /* ObjcExceptionHandler.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = ObjcExceptionHandler.h; sourceTree = "<group>"; };
 		9E989A4125F640D100235FC3 /* AppStateListenerTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AppStateListenerTests.swift; sourceTree = "<group>"; };
 		9E9EB37624468CE90002C80B /* Datadog.modulemap */ = {isa = PBXFileReference; lastKnownFileType = "sourcecode.module-map"; path = Datadog.modulemap; sourceTree = "<group>"; };
+		9EC8B5D92668197B000F7529 /* VitalCPUReader.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = VitalCPUReader.swift; sourceTree = "<group>"; };
+		9EC8B5ED2668E4DB000F7529 /* VitalCPUReaderTest.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = VitalCPUReaderTest.swift; sourceTree = "<group>"; };
 		9ED6A6B325F2901800CB2E29 /* AppStateListener.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AppStateListener.swift; sourceTree = "<group>"; };
 		9EE5AD8126205B82001E699E /* DDNSURLSessionDelegateTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = DDNSURLSessionDelegateTests.swift; sourceTree = "<group>"; };
 		9EEA4870258B76A100EBDA9D /* Global+objc.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "Global+objc.swift"; sourceTree = "<group>"; };
@@ -3059,6 +3063,7 @@
 		B3FC3C0426526EE900DEED9E /* RUMVitals */ = {
 			isa = PBXGroup;
 			children = (
+				9EC8B5D92668197B000F7529 /* VitalCPUReader.swift */,
 				B3BBBCB0265E71C600943419 /* VitalMemoryReader.swift */,
 				B3BBBCB1265E71C600943419 /* VitalReader.swift */,
 				B3FC3C0626526EFF00DEED9E /* VitalInfo.swift */,
@@ -3071,6 +3076,7 @@
 		B3FC3C1226526F4100DEED9E /* RUMVitals */ = {
 			isa = PBXGroup;
 			children = (
+				9EC8B5ED2668E4DB000F7529 /* VitalCPUReaderTest.swift */,
 				B3BBBCBB265E71D100943419 /* VitalMemoryReaderTest.swift */,
 				B3FC3C3B2653A97700DEED9E /* VitalObserverTest.swift */,
 			);
@@ -3685,6 +3691,7 @@
 				6161247925CA9CA6009901BE /* CrashReporter.swift in Sources */,
 				61133BE62423979B00786299 /* LogSanitizer.swift in Sources */,
 				615F197C25B5A64B00BE14B5 /* UIKitExtensions.swift in Sources */,
+				9EC8B5DA2668197B000F7529 /* VitalCPUReader.swift in Sources */,
 				614D812C24E3EA15004C9C5D /* Feature.swift in Sources */,
 				613E792F2577B0F900DFCC17 /* Reader.swift in Sources */,
 				61B0385A2527247000518F3C /* DDURLSessionDelegate.swift in Sources */,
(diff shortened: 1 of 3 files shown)


Sources/Datadog/RUM/RUMVitals/VitalCPUReader.swift
```diff
@@ -0,0 +1,55 @@
...
+         https://github.com/opensource-apple/xnu/blob/master/osfmk/mach/machine.h#L76
+         // machine.h (tested in iOS 14.4)
+         #define CPU_STATE_USER          0
+         #define CPU_STATE_SYSTEM        1
+         #define CPU_STATE_IDLE          2
+         #define CPU_STATE_NICE          3
+         */
+        let userTicks = cpuLoadInfo.cpu_ticks.0
+        // systemTicks is always 0 (tested in iOS 14.4)
+        let systemTicks = cpuLoadInfo.cpu_ticks.1
```
mariusc83, line 46: Why do you want to take system ticks into account ?

buranmert: at first i thought `systemTicks` might be meaningful in case of background network tasks where `URLSession` spawns another process to load the URL.
but then we decided to ignore cpu ticks at background state as they are unreliable, **so yes i should remove `systemTicks`** 👍 

it's always `0` in an iOS device and under normal conditions (foreground state) anyway

mariusc83 (approved):

Sources/Datadog/RUM/RUMVitals/VitalCPUReader.swift
```diff
@@ -0,0 +1,82 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+import UIKit.UIApplication
+
+/// A class reading the CPU ticks (_1 second = 600 ticks_) since the start of the process.
```
ncreated, line 10: Where does _600 ticks_ come from? Won't it differ per CPU?

ncreated (changes requested):
Looks good 👍, but we need to reset unwanted changes to `pbxproj`.

Datadog/Datadog.xcodeproj/project.pbxproj
```diff
@@ -4494,14 +4504,17 @@
...
+				DEVELOPMENT_TEAM = 2ZP8A5AU33;
 				INFOPLIST_FILE = TargetSupport/Example/Info.plist;
 				LD_RUNPATH_SEARCH_PATHS = (
 					"$(inherited)",
 					"@executable_path/Frameworks",
 				);
-				PRODUCT_BUNDLE_IDENTIFIER = com.datadogqh.Example;
+				PRODUCT_BUNDLE_IDENTIFIER = com.mert.Example;
 				PRODUCT_NAME = "$(TARGET_NAME)";
+				PROVISIONING_PROFILE_SPECIFIER = "";
```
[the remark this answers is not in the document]
buranmert: when i select `Automatically manage signing` and choose my personal team, it made those changes.
yes, i should keep these in my `.local.xcconfig` 👍

Sources/Datadog/RUM/RUMVitals/VitalCPUReader.swift
```diff
@@ -0,0 +1,82 @@
...
+                host_statistics(
+                    mach_host_self(),
+                    HOST_CPU_LOAD_INFO,
+                    integerPtr,
+                    &host_cpu_load_info_count_size
+                )
+            }
+        }
+        if result != KERN_SUCCESS {
+            // TODO: RUMM-1276 use sdkLogger to log errors?
```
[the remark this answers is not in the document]
buranmert: in the end we will attach vitals to RUM views, then we will be calling this method at RUM events (i'm sure of RUM View start/stop, maybe we will also read CPU vitals for RUM Action/Resource as well.?)
and this method will be called many times. 

but i couldn't simulate an error during my tests*, internal logger seems to be our only way to catch&fix bugs here.
i'd say let's go with this for dogfooding/alpha/beta testing and after we can change to fewer logs** for production version.

*: i can only simulate a `KERN_FAILURE` case when i pass `host_cpu_load_info_count_size` with less than 4
**: for example, we can log a certain error code only once: 
```
var loggedCodes: [kern_return_t]; 
if !loggedCodes.contains(result) { log(result); loggedCodes.add(result); }
```

Sources/Datadog/RUM/RUMVitals/VitalCPUReader.swift
buranmert: https://github.com/opensource-apple/xnu/blob/master/osfmk/kern/timer.h#L65
this should be the explanation why macOS reports both user and sys ticks while iOS reports user ticks only

https://github.com/opensource-apple/xnu/blob/master/osfmk/kern/host.c#L455
and this is probably where iOS returns `false` and puts all ticks into `CPU_STATE_USER` in the else clause

Sources/Datadog/RUM/RUMVitals/VitalCPUReader.swift
buranmert: TL;DR: i looked for `counter`s to get executed instructions count or anything like that. however, kernel returns us `timer`s even at the lowest point and tells us _how long_ CPU spent in user/sys/idle mode.

i couldn't debug the kernel code but as far as i could follow [the code](https://github.com/opensource-apple/xnu/blob/master/osfmk/kern/host.c#L436):
here is the macro that computes `cpu_ticks`
```
#define GET_TICKS_VALUE_FROM_TIMER(processor, state, timer) {
  GET_TICKS_VALUE(state, timer_grab(&PROCESSOR_DATA(processor, timer)));
}
```
`GET_TICKS_VALUE` doesn't do much, divides `ticks` _probably_ by 10^7 (coming from [here](https://github.com/opensource-apple/xnu/blob/master/osfmk/kern/clock.c#L367)) and stores it in `cpu_load_info`
```
#define GET_TICKS_VALUE(state, ticks) {
  cpu_load_info->cpu_ticks[(state)] += (uint32_t)(ticks / hz_tick_interval);
}
```
[`timer_grab`](https://github.com/opensource-apple/xnu/blob/master/osfmk/kern/timer.h#L148) takes a `timer_t` and returns its `all_bits`, that should mean something like timer's timestamp i guess
```
static inline uint64_t timer_grab(timer_t timer) { return timer->all_bits; }
```
[`PROCESSOR_DATA`](https://github.com/opensource-apple/xnu/blob/master/osfmk/kern/processor_data.h#L116) macro takes a `processor` and returns the `timer_t` corresponding to given `member`
`member` can be `user_state`/`system_state`/`idle_state`
```
#define PROCESSOR_DATA(processor, member) { (processor)->processor_data.member }
```
and processor_data struct has a bunch of timers
```
struct processor_data {
  timer_data_t idle_state;
  ...
}
```
and that's the lowest i could get.

Sources/Datadog/RUM/RUMVitals/VitalCPUReader.swift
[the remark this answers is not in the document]
buranmert: actually, when i looked at the [`host_statistics`](https://github.com/opensource-apple/xnu/blob/master/osfmk/kern/host.c#L425) function once again it seems like it never fails (except `count < HOST_CPU_LOAD_INFO_COUNT` case but this is trivial)

if only they changed the kernel code one day, we will be flooded with errors here and our feature will break too!
we shouldn't overlook any errors coming from this line, do you think `.errorOnce(...)` can do the job? 
can we ensure the log won't get lost among all the others in the web UI?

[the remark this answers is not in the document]
ncreated: > can we ensure the log won't get lost among all the others in the web UI?

We can triage it as `CRITICAL` log (`criticalOnce()`). There are not much internal logs like that - they only indicate serious misalignments in _our assumptions vs reality_ or some other **very** important stuff.

Our current adoption of internal monitoring is quite strict and there's not much data sent from user device. But looking at `service:dd-sdk-ios` logs and muting non-important stuff should be our continuous routine.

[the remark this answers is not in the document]
buranmert: added a note for the future ✅

Tests/DatadogTests/Datadog/RUM/RUMVitals/VitalCPUReaderTest.swift
```diff
@@ -0,0 +1,64 @@
...
+
+import XCTest
+import UIKit.UIApplication
+@testable import Datadog
+
+class VitalCPUReaderTest: XCTestCase {
+    let testNotificationCenter = NotificationCenter()
+    lazy var cpuReader = VitalCPUReader(notificationCenter: testNotificationCenter)
+
+    func test_whenCPUUnderHeavyLoad_itMeasuresHigherCPUTicks() throws {
```
ncreated, line 15: naming convention:
```suggestion
    func testWhenCPUUnderHeavyLoadItMeasuresHigherCPUTicks() throws {
```

Tests/DatadogTests/Datadog/RUM/RUMVitals/VitalCPUReaderTest.swift
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        let lowLoadAverage = try averageCPUTicks {
+            Thread.sleep(forTimeInterval: 1.0)
+        }
+
+        XCTAssertGreaterThan(highLoadAverage, lowLoadAverage)
+    }
+
+    func test_whenInactiveAppState_itIggnoresCPUTicks() throws {
```
ncreated, line 30: typo & naming convention:
```suggestion
    func testWhenInactiveAppStateItIgnoresCPUTicks() throws {
```

Sources/Datadog/RUM/RUMVitals/VitalCPUReader.swift
```diff
@@ -0,0 +1,88 @@
...
+            let ongoingInactiveTicks = ticks - (utilizedTicksWhenResigningActive ?? ticks)
+            let inactiveTicks = totalInactiveTicks + ongoingInactiveTicks
+            return Double(ticks - inactiveTicks)
+        }
+        return nil
+    }
+
+    @objc
+    private func appWillResignActive() {
+        utilizedTicksWhenResigningActive = readUtilizedTicks()
```
ncreated, line 33: When I looked further, I can see that this will lead to threading issue. The `readVitalData()` will be accessed from background thread (by `RUMScope`), but the `appWillResignActive()` will be called on the main thread. Both access the `utilizedTicksWhenResigningActive` state by either reading or writing.

We can solve it later, when actually plugging `VitalReaders` to RUM (this will pop up in threading tests), but it's worth leaving this note.

Tests/DatadogTests/Datadog/RUM/RUMVitals/VitalCPUReaderTest.swift
```diff
@@ -0,0 +1,64 @@
...
+                let random = Double.random(in: Double.leastNonzeroMagnitude...Double.greatestFiniteMagnitude)
+                _ = tan(random).squareRoot()
+            }
+        }
+
+        let lowLoadAverage = try averageCPUTicks {
+            Thread.sleep(forTimeInterval: 1.0)
+        }
+
+        XCTAssertGreaterThan(highLoadAverage, lowLoadAverage)
```
ncreated, line 27: This test seems stable when running on a Mac, but when running on an iPhone, I receive ~13% failures. And it's coming from a reason that the `highLoadAverage` is sometimes `0`. While we can call the last 3 failures a "flakiness" factor, the first 10 do rather look like some measurement issue, WDYT?

Screenshot 2021-06-11 at 08 39 30

Sources/Datadog/RUM/RUMVitals/VitalCPUReader.swift
ncreated: I see 👌 , this is pretty complex to me as I don't know into these details. I made few more tests on my iPhone X, receiving some failures - explained in comment below.

Sources/Datadog/RUM/RUMVitals/VitalCPUReader.swift
buranmert: yes, that's a good point and i'm still not sure where the `VitalReader`s will live.
if they live in `RUMMonitor` we can read data in main thread within `rum.start/stopView(...)` methods (even before `process(command:)`) and that probably gives the most accurate measurement.
if they live in the scopes then we will need to add extra code as you said.

i'm leaving a comment 👍

Tests/DatadogTests/Datadog/RUM/RUMVitals/VitalCPUReaderTest.swift
buranmert: now i'm running `heavyLoad()` for 1.0 second. that should improve test results.

although usually `host_statistics` updates its values after 0.01 second (which makes sense when we look at its source code), it sometimes keeps reporting the same values for 1.0 second (which doesn't make sense to me).
never 2.0 seconds or 0.5 second. i still can't figure out why.

i just noticed that when i do `Thread.sleep(...)`, `host_statistics` returns updated values after 0.01 second most of the time. 
without `Thread.sleep(...)` it usually doesn't update for 1.0 second.

ncreated (approved):

PR #516, opened 2021-06-09
RUMM-1401 Update RUM models generator for recent event schemas

ncreated:
### What and why?

📦 This PR updates `rum-events-generator` tool to support recent updates to [rum-events-format](https://github.com/DataDog/rum-events-format) JSON schemas.

This is to stay up-to-date with [rum-events-format](https://github.com/DataDog/rum-events-format) and unblock adding support to new features, notably metrics that we will be collecting in Mobile Vitals project.

### How?

Since last update, there were 2 new concepts introduced to our JSON schemas which required additional implementation:
* `"additionalProperties": true` schema element for adding _any_ properties to JSON object,
* `"items": {"type": "object", ...}` schema variant for generating an array of JSON objects.

#### `"additionalProperties": true`

The initial support for `additionalProperties` was already added in https://github.com/DataDog/dd-sdk-ios/pull/399, but we decided to constraint it only to the case where `type` is provided (and generate `[String: T]` property for it). Now, we need to push it further and allow anonymous properties of any type.

This new schema is used in [`usr` object](https://github.com/DataDog/rum-events-format/blob/d40d93a607a2d4483c95bd000a4e8ebd96fdf1fd/schemas/_common-schema.json#L91) to encode _user extra info_ attributes:
```js
"usr": {
  "type": "object",
  "additionalProperties": true,
  "properties": { ... }
}
```
and in [event's `context` object](https://github.com/DataDog/rum-events-format/blob/d40d93a607a2d4483c95bd000a4e8ebd96fdf1fd/schemas/_common-schema.json#L177) to encode additional _attributes_ for each RUM event:
```js
"context": {
  "type": "object",
  "additionalProperties": true
}
```

To generate RUM models code for it, I used the [`DynamicCodingKey` concept which we already use](https://github.com/DataDog/dd-sdk-ios/blob/71f39a91006129f4bffcf06caa9b672b337afdf0/Sources/Datadog/RUM/RUMEvent/RUMEventEncoder.swift#L28-L48) for encoding these attributes in free-form format.

In this PR I'm not updating our code to leverage new `usrInfo` and `contextInfo` properties. Instead, this work is detached to `RUMM-1420`. I made sure there is no regression in keeping those properties `nil` and empty and we can do refactoring later.

#### `"items": {"type": "object", ...}`

This construct is used to encode `InForegroundPeriods` for `RUMViewEvent`:
```swift
/// List of the periods of time the user had the view in foreground (focused in the browser)
public let inForegroundPeriods: [InForegroundPeriods]?
```

This didn't require any change in `Swift` code printer, but the `ObjcInterop` printer had to be updated as similar semantic was not used before. For simplicity, I constrained the generator to only support immutable arrays of structures. Mutability can be added later in the places where generator throws `unimplemented()` exception.

### Review checklist


tools/rum-models-generator/Tests/rum-models-generator-coreTests/RUMModelsGeneratorTests.swift
```diff
@@ -21,12 +21,12 @@ final class RUMModelsGeneratorTests: XCTestCase {
 //
 //        let schemas = try RUMJSONSchemaFiles(folder: schemasFolderURL)
 //        let generator = RUMModelsGenerator()
-//
+
 //        print(">>>>>>>>>>>>>>>>>> Swift >>>>>>>>>>>>>>>>>>>>>>")
-//        print(try generator.printRUMModels(for: schemas, using: SwiftPrinter()))
+//        print(try generator.printRUMModels(for: schemas, using: .swift))
```
ncreated, line 26: 💡 This test is commented out intentionally. It is used to run generator against real schemas from `rum-events-format` and do step-by-step analysis with debugger. This is annotated with test comment:
```swift
/// Test made for debugging purpose.
/// Uncomment it to run code generation for `../../../rum-events-format/schemas`.
```

buranmert (approved):

PR #520, opened 2021-06-14
RUMM-1138 Generate monitors for E2E tests

ncreated:
### What and why?

🧰 This PR adds a tool to generate terraform definitions for Datadog monitors. These monitors will which assert data and performance metrics from  E2E tests nightly runs.

### How?

Terraforming concept is simple: the `terraform apply` applies all definitions from `main.tf` to our _Mobile - Integration_ org.

The `main.tf` file is generated from `*.tf.src` templates and variables defined above each test method, e.g.:
```swift
/// - data monitor:
/// ```logs
/// $monitor_id = logs_logger_debug_log_data
/// $monitor_name = "[RUM] [iOS] Nightly - logs_logger_debug_log: number of logs is below expected value"
/// $monitor_query = "logs(\"service:com.datadog.ios.nightly @test_method_name:logs_logger_debug_log status:debug\").index(\"*\").rollup(\"count\").last(\"1d\") < 1"
/// ```
func test_logs_logger_DEBUG_log() {
   // ...
}
``` 

↑ This will generate monitor code using `templates/monitor-logs.tf.src` template. The tool will substitute `$monitor_id`, `$monitor_name` and `$monitor_query` variables in template with their values defined for each monitor.

Generation can be launched with `make e2e-monitors-generate` - it will parse all E2E test files, find test methods, read monitors declaration and generate `monitors-gen/main.tf`. By running `terraform plan` in `monitors-gen` it is possible to preview all changes that will be applied to MI org.

#### `*.tf.src` templates

As for now, there are 3 templates:
* `templates/main.tf.src` - with a definition of Datadog provider. It uses secrets from git-ignored `secrets.tf`.
* `templates/monitor-logs.tf.src` - with a definition of "log alert" (Logs) monitor.
* `templates/monitor-apm.tf.src` - with a definition of "query alert" (APM) monitor.

Each template defines its own variables: `${{var_name}}`. Some variables use default value: `${{var_name:-default}}`.

The generation will succeed only if all required variables are defined in monitor's declaration.

#### Linter extension

To make defining monitors easier and less error-prone, the generator is armed with linter capability. It can spot simple but impactful mistakes and should guarantee consistency of defined monitors.

It runs as part of `E2ETests` target build action, so linter issues can be seen instantly, e.g.:

Screenshot 2021-06-15 at 15 04 37

Screenshot 2021-06-15 at 15 05 31

#### Code review

This PR is split into 3 commits:
* first - updates Xcode 12.4 compiler issue in `E2ETests` target,
* second - adds the actual Python tool,
* third - adds definition for first 4 monitors in Logging product.

Definitions for all other Logging tests will be added in separate PR, to not flood this one.

### Review checklist


tools/nightly-e2e-tests/monitors-gen/monitor.tf.src
```diff
@@ -0,0 +1,25 @@
...
+@maciek.grzybowski@datadoghq.com
+@mert.buran@datadoghq.com
+@marius@datadoghq.com
+EOT
+  escalation_message = <<EOT
+<nil>
+EOT
+
+  query             = ${{monitor_query}}
+  notify_no_data    = false
```
0xnm, line 15: I guess this should be behind variable as well, because we want to notify if there is no data for 24h (with exemption of monitors when we want to check that we don't get particular value)

tools/nightly-e2e-tests/monitors-gen/monitor.tf.src
```diff
@@ -0,0 +1,25 @@
...
+  renotify_interval = 0
+  notify_audit      = false
+  timeout_h         = 0
+  include_tags      = true
+  new_host_delay    = 300
+  monitor_thresholds {
+    critical = ${{monitor_threshold:-1.0}}
+  }
+
+}
```
0xnm, line 25: very minor: missing new line at the EOF.

tools/nightly-e2e-tests/monitors-gen/monitor.tf.src
ncreated: Indeed, added a variable with default value set to `false`.

tools/nightly-e2e-tests/monitors-gen/terraform.tfstate
```diff
@@ -0,0 +1,225 @@
+{
```
ncreated, line 1: 💡 I'm not yet sure what to do with this file. Ideally, we should put it on [TF remote backend](https://www.terraform.io/docs/language/settings/backends/remote.html) so that the state is always managed from single machine. I will come with the best answer before merging this PR.

ncreated: Got my opinion on that - proposed simplified Terraform workflow in #521.

tools/nightly-e2e-tests/monitors-gen/templates/monitor-apm.tf.src
```diff
@@ -0,0 +1,25 @@
...
+  message            = <<EOT
+@maciek.grzybowski@datadoghq.com
+@mert.buran@datadoghq.com
+EOT
+  escalation_message = <<EOT
+<nil>
+EOT
+
+  query               = ${{monitor_query}}
+  notify_no_data      = ${{notify_no_data:-false}}
```
0xnm, line 14: shouldn't we make it `true` by default? because majority of the tests expect data

tools/nightly-e2e-tests/nightly_e2e.py
```diff
@@ -0,0 +1,147 @@
...
+
+        if cli_args.command == 'lint':
+            lint(cli_args)
+        elif cli_args.command == 'generate-tf':
+            generate_terraform_file(cli_args)
+
+    except Exception as error:
+        print(f'❌ nightly_e2e.py failed: {error}')
+        print('-' * 60)
+        traceback.print_exc(file=sys.stdout)
```
0xnm, line 143: I'm not familiar with python practices, but just wondering: should it be `stderr` here instead?

tools/nightly-e2e-tests/monitors-gen/.terraform.lock.hcl
```diff
@@ -0,0 +1,22 @@
+# This file is maintained automatically by "terraform init".
+# Manual edits may be lost in future updates.
+
+provider "registry.terraform.io/datadog/datadog" {
```
0xnm, line 4: just to confirm: this file should be commited?

tools/nightly-e2e-tests/monitors-gen/terraform.tfstate
```diff
@@ -0,0 +1,225 @@
...
+              "service:com.datadog.ios.nightly",
+              "source:ios",
+              "team:rumm"
+            ],
+            "timeout_h": 0,
+            "type": "log alert",
+            "validate": null
+          },
+          "sensitive_attributes": [],
+          "private": "bnVsbA=="
```
0xnm, line 58: what `private` means in the `tfstate` schema? can it leak sensitive info somehow? We need to be careful with this file by looking at this page https://www.terraform.io/docs/language/state/sensitive-data.html

tools/nightly-e2e-tests/monitors-gen/terraform.tfstate
```diff
@@ -0,0 +1,225 @@
...
+              "env:instrumentation",
+              "service:com.datadog.ios.nightly",
+              "source:ios",
+              "team:rumm"
+            ],
+            "timeout_h": 0,
+            "type": "log alert",
+            "validate": null
+          },
+          "sensitive_attributes": [],
```
0xnm, line 57: we need to be careful with that as well, by looking on the name.

0xnm (approved):
looks great! I just left few minor comments, my biggest concern is state file from the perspective of security.

tools/nightly-e2e-tests/monitors-gen/templates/monitor-apm.tf.src
ncreated: TBH I didn't put too much attention to this value initially. Among all 210 Android monitors that I initially exported none was defining `notify_no_data = true` 🤔, so I didn't customize this value for any iOS monitor definition.

The [doc says](https://registry.terraform.io/providers/DataDog/datadog/latest/docs/resources/monitor#optional):
> * `notify_no_data` (Boolean) A boolean indicating whether this monitor will notify when data stops reporting. Defaults to `false`.

so indeed, sounds like this will raise an alert if monitor has no data. On the other side, if there is no data the monitor will trigger anyway - as its assertion won't be met. That said, I think that this value is redundant to what we already observe, but I will try with `true` for the sake of correctness 👌.

tools/nightly-e2e-tests/monitors-gen/.terraform.lock.hcl
ncreated: Like noticed in #521 this file will be removed. I will first merge #521 to this PR, so this file won't be committed to the main branch. I also made sure that there are no secrets inside before pushing it to remote 👍.

tools/nightly-e2e-tests/monitors-gen/terraform.tfstate
ncreated: n/a: https://github.com/DataDog/dd-sdk-ios/pull/520#discussion_r653454702

tools/nightly-e2e-tests/monitors-gen/terraform.tfstate
ncreated: n/a: https://github.com/DataDog/dd-sdk-ios/pull/520#discussion_r653454702

tools/nightly-e2e-tests/monitors-gen/templates/monitor-apm.tf.src
ncreated: Done in https://github.com/DataDog/dd-sdk-ios/pull/521/commits/2c571d18f7ac7f91c0db91ecbf1a54f881120952 👍

0xnm: I didn't check it in the export, but majority of the monitors I've created have this:

image

Maybe `dd2tf` is missing something in the export template? I'm not sure if monitor will be triggered if there is no data, but I didn't check if it happens or not tbh.

tools/nightly-e2e-tests/nightly_e2e.py
ncreated: It could 👍. It's more Unix than Pythonic discussion on whether we should use `STDOUT` or `STDERR`. I used `stdout` for this tool (similar to few other that we have in `dd-sdk-ios`) as I know that this output is well displayed in Bitrise build log 🙂 and I'm not sure if it will appear there if it's `stderr`. But TBH I never tried as it doesn't seem important.

0xnm: indeed, not that important.

tools/nightly-e2e-tests/monitors-gen/templates/monitor-apm.tf.src
ncreated: Which made me realise that we also have to use `no_data_timerframe` set to `24h` - done in https://github.com/DataDog/dd-sdk-ios/pull/521/commits/b217d6ea50e78202d5904d84d91f7b8cdd891f00 thanks 🙂.

About `dd2tf` - it is not very up-to-date and, so maybe that's why (I can't find `no_data_timerframe` there neither).

tools/nightly-e2e-tests/src/main_tf_renderer.py
```diff
@@ -0,0 +1,133 @@
...
+
+class MonitorTemplate:
+    """
+    Renders `monitor-logs.tf.src` file by replacing variables with values read from `MonitorConfiguration`.
+    If a given `$variable` is not defined in monitor configuration, then its default from template will be used.
+    """
+    def __init__(self, template_content: str):
+        self.template_content = template_content
+
+    def render(self, monitor: MonitorConfiguration):
```
mariusc83, line 72: are you missing the return type on all these functions or is this intentionally ?

tools/nightly-e2e-tests/src/linter.py
```diff
@@ -0,0 +1,93 @@
...
+
+@contextlib.contextmanager
+def linter_context(code_reference: CodeReference):
+    """
+    Context manager for convenient work with linter's code reference.
+    After context returns, the code reference is reseted to its previous value.
+    """
+    Linter.shared.push_context(code_reference=code_reference)
+    try:
+        yield
```
mariusc83, line 27: this keyword is mostly used in Generators ...I am not sure I understand its usage here, can you please explain?

tools/nightly-e2e-tests/src/linter.py
```diff
@@ -0,0 +1,93 @@
...
+        pass
+
+    def emit_warning(self, message: str):
+        pass
+
+    def emit_error(self, message: str):
+        pass
+
+
+class Linter(NoOpLinter):
```
mariusc83, line 52: Is this `XCode` specific ? Are we going to have issues using this script from Android ?

ncreated: The `Linter's` interface is platform agnostic, so I hope this could be _somehow_ integrated with Gradle / Android linter:
```
Linter
- emitWarning(message, file_path, line_number)
- emitError(message, file_path, line_number)
```

The way we integrate it with Xcode build is by printing this to `STDOUT`:
```
<file_path>:<line_number> (warning|error): message
```
Xcode understands this format and renders warnings or errors in appropriate source code locations.

tools/nightly-e2e-tests/src/linter.py
ncreated: The main goal of this code is to define `linter_context()` context manager. There are [two ways](https://book.pythontips.com/en/latest/context_managers.html#context-managers) of defining context managers and one of them is [by using generator](https://book.pythontips.com/en/latest/context_managers.html#implementing-a-context-manager-as-a-generator) (hence the `yield` keyword that you notice is indeed coming from generator). I used generator-way as it is shorter than class-way with implementing `_enter_` and `_exit_`.

And about context manager itself, I use it as a way to manage the linter context:
```python
with linter_context(code_reference1):
   Linter.shared.emitWarning('foo')  # 'foo' warning will be emitted in file-path and line-no from code_reference1 

   with linter_context(code_reference2): 
      Linter.shared.emitWarning('bar') # 'bar' warning will be emitted in code_reference2 

   Linter.shared.emitWarning('fizz') # 'fizz' warning will be emitted in code_reference1 
```

tools/nightly-e2e-tests/src/main_tf_renderer.py
ncreated: Indeed, it was missing in a few methods - I added return types in https://github.com/DataDog/dd-sdk-ios/pull/521/commits/3e16ca074809f31eb7e426c8bd36e1bca2136171. Python uses [Duck typing](https://en.wikipedia.org/wiki/Duck_typing) so it's type control is not that strict as Swift or Kotlin, thus Pycharm doesn't even warn you if you ommit the type declaration. This can be perhaps enforced with some additional plugins or configuration.

ncreated:
All monitors are green 👌, merging.

Screenshot 2021-06-21 at 14 28 20

PR #521, opened 2021-06-16
RUMM-1138 Define E2E monitors for all Logging APIs

ncreated:
### What and why?

⚙️ This PR adds E2E monitor definitions for test methods in Logging product. It also proposes simplified workflow for managing Terraform state of iOS monitors in _Mobile - Integration_ org until further improvement is done in `RUMM-1440`

### How?

Leveraging the work done in #520, I added monitor definitions to all public APIs in Logging product.

There was one case not considered in #520 - generating monitors for common APIs used in multiple test methods:
* `Datadog.initialize(appContext:trackingConsent:configuration:)` - measured with `sdk_initialize` span,
* `Logger.builder.build()` - measured with `logs_logger_initialize` span,
* `Datadog.set(trackingConsent:)` - measured with `sdk_set_tracking_consent` span.

Unlike all others, these monitors are not attached to any test method, so I added additional logic in `test_file_parser.py` to read `independent_monitors` and store them in `TestFile` dataclass:
```python
@dataclass()
class TestFile:
    test_methods: [TestMethod]
    independent_monitors: [MonitorConfiguration]  # monitors defined in test file, but not attached to any `TestMethod`
```

#### Simplified workflow for iOS monitors management in _Mobile - Integration_ org

In this PR I remove `main.tf` and `terraform.tfstate` files from version control.
* The `main.tf` doesn't require versioning because it's generated using `make e2e-monitors-generate` (think: running it for particular commit will always produce the same `main.tf`).
* Versioning `terraform.tfstate` is discouraged and using [Terraform Backend](https://www.terraform.io/docs/language/settings/backends/index.html) solution is preferred (so the state source-of-truth is centralised and managed from single machine with no need for `terraform import`). We could leverage Datadog Terraform Configs repository to manage this state and only submit partial updates to iOS monitors, but this requires additional work and will not be practical during E2E tests development for other features. We could think of this improvement in `RUMM-1440` after we reach full coverage for existing public APIs.

Until we migrate to Terraform Backend, **the workflow is this**:
* 1/ Add new monitor definition(s).
* 2/ Call `make e2e-monitors-generate` to generate `main.tf` (it will also delete any existing `terraform.tfstate`)
* 3/ Manually delete all "*iOS*" monitors in _Mobile - Integration_ org.
* 4/ Call `terraform apply` to submit all monitors (this will include deleted monitors and new ones).

I've tested this workflow few times when iterating on monitors added in this PR - it works good. The 3rd step requires just two clicks in _Mobile - Integration_ org:

Screenshot 2021-06-16 at 21 02 37

### Review checklist


0xnm:
> * Versioning `terraform.tfstate` is discouraged and using [Terraform Backend](https://www.terraform.io/docs/language/settings/backends/index.html) solution is preferred (so the state source-of-truth is centralised and managed from single machine with no need for `terraform import`). We could leverage Datadog Terraform Configs repository to manage this state and only submit partial updates to iOS monitors, but this requires additional work and will not be practical during E2E tests development for other features. We could think of this improvement in `RUMM-1440` after we reach full coverage for existing public APIs.

Ok, then it answers my comment about `tfstate` in #520. We probably need to refresh #520 by removing this file ASAP, just in case.

> Until we migrate to Terraform Backend, **the workflow is this**:
> 
> * 1/ Add new monitor definition(s).
> * 2/ Call `make e2e-monitors-generate` to generate `main.tf` (it will also delete any existing `terraform.tfstate`)
> * 3/ Manually delete all "_iOS_" monitors in _Mobile - Integration_ org.
> * 4/ Call `terraform apply` to submit all monitors (this will include deleted monitors and new ones).

Actually we can still leverage `terraform import`. What if before `terraform apply` we export existing monitors, extract `name` + `id` pairs and then just link existing ones to the ones in `main.tf` by name and then use `terraform import` with mapped ID?

Datadog/E2ETests/Logging/LoggerBuilderE2ETests.swift
```diff
@@ -121,7 +204,14 @@ class LoggerBuilderE2ETests: E2ETests {
     }
 
     /// - api-surface: Logger.Builder.bundleWithTrace(_ enabled: Bool) -> Builder
-    func test_logs_logger_builder_BUNDLE_WITH_TRACE_enabled() { // E2E:wip
+    ///
+    /// - data monitor - unfortunately we can't assert any APM trait in this monitor, so we just check if the data comes in:
```
0xnm, line 208: it seems when enabled it adds `trace_id` and `span_id` to the log attributes, given there is active span. we cannot assert these attributes?

ncreated:
> Ok, then it answers my comment about `tfstate` in #520. We probably need to refresh #520 by removing this file ASAP, just in case.

This PR will be merged to #520 so `tfstate` will be removed before getting to the main branch. I had a very close look to this file initially, and it has no sensitive information as we don't leverage much of TF capabilities 🙂.

> > Until we migrate to Terraform Backend, **the workflow is this**:
> > 
> > * 1/ Add new monitor definition(s).
> > * 2/ Call `make e2e-monitors-generate` to generate `main.tf` (it will also delete any existing `terraform.tfstate`)
> > * 3/ Manually delete all "_iOS_" monitors in _Mobile - Integration_ org.
> > * 4/ Call `terraform apply` to submit all monitors (this will include deleted monitors and new ones).
> 
> Actually we can still leverage `terraform import`. What if before `terraform apply` we export existing monitors, extract `name` + `id` pairs and then just link existing ones to the ones in `main.tf` by name and then use `terraform import` with mapped ID?

Could you explain it a bit more 💭? I'm not sure if I understand how this will work for adding / destroying / updating monitors. Can this be done with just terraform commands or would require writing updates to generated `main.tf`?

Datadog/E2ETests/Logging/LoggerBuilderE2ETests.swift
ncreated: Yes, when enabled we add two extra attributes to each `Log`:
```swift
static let traceID = "dd.trace_id"
static let spanID = "dd.span_id"
```
so it can create the link in Datadog. The `span` can be seen from Log UI, but these attributes are not preserved in the UI thus cannot be seen / asserted and the facet cannot be created for further querying. We have the same issue in Android monitors for `logs_logger_builder_bundle_with_trace_enabled` 🙁. @mariusc83 am I correct / do you see any other way around?

0xnm:
> > Actually we can still leverage `terraform import`. What if before `terraform apply` we export existing monitors, extract `name` + `id` pairs and then just link existing ones to the ones in `main.tf` by name and then use `terraform import` with mapped ID?
> 
> Could you explain it a bit more 💭? I'm not sure if I understand how this will work for adding / destroying / updating monitors. Can this be done with just terraform commands or would require writing updates to generated `main.tf`?

So let's say we have monitors created from the code in the `main.tf` file and there is monitor called `test_foo` there. Then we can do export of existing monitors with `dd2tf` (or some other tool, TF format is not mandatory there) and extract pairs `(monitor_name, monitor_id)`. Then for the `monitor_name` which can be found in `main.tf` we can simply run `terraform import %monitor_id_from_main_tf% %monitor_id_from_export%`, binding the state. then `terraform apply`, but this still leaves us with the need to delete monitors which exist in the export, but were not bound. So in the nutshell, we will do simplified diffing manually.

ncreated:
> So let's say we have monitors created from the code in the `main.tf` file and there is monitor called `test_foo` there. Then we can do export of existing monitors with `dd2tf` (or some other tool, TF format is not mandatory there) and extract pairs `(monitor_name, monitor_id)`. Then for the `monitor_name` which can be found in `main.tf` (say `test_foo` exists in both) we can simply run `terraform import %monitor_id_from_main_tf% %monitor_id_from_export%`, binding the state. Then `terraform apply`, but this still leaves us with the need to delete monitors which exist in the export, but were not bound. So in the nutshell, we will do simplified diffing manually.

I see 👍. Indeed, this will work for adding and updating monitors. With more effort we could even handle deleting existing monitors basing on the assumption that all iOS monitors are managed through TF: `to destroy` = `remote monitors containing *iOS*` - `generated monitors`.

In any way, I don't think this is worth implementing - it seems pretty complex and sounds like re-doing what we will get for free when switching to remote backend in `RUMM-1440`. WDYT?

0xnm:
> In any way, I don't think this is worth implementing - it seems pretty complex and sounds like re-doing what we will get for free when switching to remote backend in `RUMM-1440`. WDYT?

yes, indeed, remote backend should answer our needs.

0xnm (approved):

PR #522, opened 2021-06-17
RUMM-1278 Screen Refresh Rate vitals

buranmert:
### What and why?

We will be reading the refresh rate during app's lifetime in order to detect UI hitches.

### How?

iOS renders UI in the main run loop, this is a `RunLoop` instance. We want to be notified when a new frame is drawn.
`CADisplayLink` is a special timer type which listens to frame paints in a given run loop.
We add a `CADisplayLink` instance to `RunLoop.main` so that it calls our method every time a new frame is drawn.

### Review checklist

### DIFF
--- a/Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
+++ b/Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
@@ -9,7 +9,8 @@ import UIKit
 
 /// A class reading the refresh rate (frames per second) of the main screen
 internal class VitalRefreshRateReader {
-    private var observers = [VitalObserver]()
+    private let valuePublisher = ValuePublisher(initialValue: 0.0)
+
     private var displayLink: CADisplayLink?
     private var lastFrameTimestamp: CFTimeInterval?
 
@@ -37,19 +38,13 @@ internal class VitalRefreshRateReader {
     /// `VitalRefreshRateReader` keeps pushing data to its `observers` at every new frame.
     /// - Parameter observer: receiver of refresh rate per frame.
     func register(_ observer: VitalObserver) {
-        DispatchQueue.main.async {
-            self.observers.append(observer)
-        }
+        valuePublisher.subscribe(observer)
     }
 
     /// `VitalRefreshRateReader` stops pushing data to `observer` once unregistered.
     /// - Parameter observer: already added observer; otherwise nothing happens.
     func unregister(_ observer: VitalObserver) {
-        DispatchQueue.main.async {
-            self.observers.removeAll { existingObserver in
-                return existingObserver === observer
-            }
-        }
+        valuePublisher.unsubscribe(observer)
     }
 
     // MARK: - Private
@@ -59,10 +54,7 @@ internal class VitalRefreshRateReader {
         if let lastTimestamp = self.lastFrameTimestamp {
             let frameDuration = link.timestamp - lastTimestamp
             let currentFPS = 1.0 / frameDuration
-            // NOTE: RUMM-1278 `oldValue` is not used
-            observers.forEach {
-                $0.onValueChanged(oldValue: 0.0, newValue: currentFPS)
-            }
+            valuePublisher.publishAsync(currentFPS)
         }
         lastFrameTimestamp = link.timestamp
     }
--- a/Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
+++ b/Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
@@ -1,0 +1,112 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+import UIKit
+
+/// A class reading the refresh rate (frames per second) of the main screen
+internal class VitalRefreshRateReader {
+    private var observers = [VitalObserver]()
+    private var displayLink: CADisplayLink?
+    private var lastFrameTimestamp: CFTimeInterval?
+    private(set) var isRunning = false
+
+    init(notificationCenter: NotificationCenter = .default) {
+        notificationCenter.addObserver(
+            self,
+            selector: #selector(appWillResignActive),
+            name: UIApplication.willResignActiveNotification,
+            object: nil
+        )
+        notificationCenter.addObserver(
+            self,
+            selector: #selector(appDidBecomeActive),
+            name: UIApplication.didBecomeActiveNotification,
+            object: nil
+        )
+    }
+
+    deinit {
+        stop()
+    }
+
+    /// `VitalRefreshRateReader` keeps pushing data to its `observers` at every new frame.
+    /// - Parameter observer: receiver of refresh rate per frame.
+    func register(_ observer: VitalObserver) {
+        DispatchQueue.main.async {
+            self.observers.append(observer)
+        }
+    }
+
+    /// `VitalRefreshRateReader` stops pushing data to `observer` once unregistered.
+    /// - Parameter observer: already added observer; otherwise nothing happens.
+    func unregister(_ observer: VitalObserver) {
+        DispatchQueue.main.async {
+            self.observers.removeAll { existingObserver in
+                return existingObserver === observer
+            }
+        }
+    }
+
+    /// Starts listening to frame paints.
+    /// - Throws: only if `UIScreen.main` cannot generate its `CADisplayLink`
+    func start() throws {
+        try private_start()
+        isRunning = true
+    }
+
+    /// Stops listening frame paints. Automatically called at `deinit()`.
+    func stop() {
+        private_stop()
+        isRunning = false
+    }
+
+    // MARK: - Private
+
+    @objc
+    private func displayTick(link: CADisplayLink) {
+        if let lastTimestamp = self.lastFrameTimestamp {
+            let frameDuration = link.timestamp - lastTimestamp
+            let currentFPS = 1.0 / frameDuration
+            // NOTE: RUMM-1278 `oldValue` is not used
+            observers.forEach {
+                $0.onValueChanged(oldValue: 0.0, newValue: currentFPS)
+            }
+        }
+        lastFrameTimestamp = link.timestamp
+    }
+
+    @objc
+    private func appWillResignActive() {
+        private_stop()
+    }
+
+    @objc
+    private func appDidBecomeActive() {
+        if isRunning {
+            try? private_start()
+        }
+    }
+
+    private func private_start() throws {
+        stop()
+
+        guard let link = UIScreen.main.displayLink(
+            withTarget: self,
+            selector: #selector(displayTick(link:))
+        ) else {
+            throw InternalError(description: "CADisplayLink could not be created!")
+        }
+        link.add(to: .main, forMode: .default)
+        self.displayLink = link
+    }
+
+    private func private_stop() {
+        displayLink?.invalidate()
+        displayLink = nil
+        lastFrameTimestamp = nil
+    }
+}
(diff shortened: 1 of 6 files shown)


Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
```diff
@@ -0,0 +1,35 @@
...
+    private var lastRefreshRate: Double?
+
+    init() throws {
+        guard let link = UIScreen.main.displayLink(
+            withTarget: self,
+            selector: #selector(displayTick(link:))
+        ) else {
+            throw InternalError(description: "CADisplayLink could not be created!")
+        }
+        link.add(to: .main, forMode: .default)
```
ncreated, line 20: [The doc](https://developer.apple.com/documentation/quartzcore/cadisplaylink/1621323-add) says:
> The run loop retains the display link. To remove the display link from all run loops, send an `invalidate()` message to the display link.

We need to take this into account as it might be causing flakiness (in tests) and might leak memory (in tests and in production code) depending on how this component will be used from the caller.

Will we instantiate several `VitalRefreshRateReader` instances or use a single one?

buranmert: good point 👍 made a change: added `start()/stop()` which adds `link` to run loop and invalidates it later.

i think we should have an always-running singleton (`RefreshRateReader`) and multiple observers (`[VitalObserver]`) so that the usage will be like that:
```
// RUMView starts
let newObserver = createObserverForNewRUMView(
reader.shared.subscribe(newObserver)
...
// RUMView ends
let refreshRateForView = newObserver.meanValue
reader.shared.unsubscribe(newObserver)
```
i think a singleton is a good fit here because, like in `AppStateListener` case, we hit main thread when we start `reader`. that's because [Apple](https://developer.apple.com/documentation/foundation/runloop) says `RunLoop`s cannot be called from different threads.

how does that sound to you?

currently i tried to port `RefreshRateReader` logic in https://github.com/DataDog/dd-sdk-android/pull/640/ to iOS. 
i haven't figured out yet how they read refresh rate per `RUMView`. i'm still reading Android code.

ncreated: Having single instance of "refresh rate reader" makes sense 👍. It doesn't need to be singleton - IMO it sounds reasonable to store it in `RUMFeature.instance` as vitals monitoring is purely RUM-related stuff (also it doesn't need to be instantiated if RUM is disabled).

For managing the subscription in a thread-safe way you may want to have a look at existing [ValuePublisher](https://github.com/DataDog/dd-sdk-ios/blob/master/Sources/Datadog/Core/Utils/ValuePublisher.swift). It is made for performant reads and non-blocking writes using concurrent queue.

Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
```diff
@@ -0,0 +1,112 @@
...
+                return existingObserver === observer
+            }
+        }
+    }
+
+    /// Starts listening to frame paints.
+    /// - Throws: only if `UIScreen.main` cannot generate its `CADisplayLink`
+    func start() throws {
+        try private_start()
+        isRunning = true
```
ncreated, line 58: This `start()` method calls `private_start()`, which calls `stop()`, which sets `isRunning` to `false`, but then here we set it back to `true`. This logic is too much coupled and it has a hidden bug:

- 1/ `refreshRateReader.start()` → `isRunning = true`
- 2/ `appDidBecomeActive()` → it will call `private_start()` → `stop()` and leave `isRunning = false`
- 3/ `appWillResignActive()` → no change to `isRunning`
- 4/ `appDidBecomeActive()`- as `isRunning == false` it won't restart receiving updates.

This scenario corresponds to initializing SDK early in App Delegate. The refresh reader will be started and there will be one "did become active" notification delivered right away. I did this test in https://github.com/DataDog/dd-sdk-ios/commit/543b9a0288bd349e5d4945a8672d9d0a87c069ec

Tests/DatadogTests/Datadog/RUM/RUMVitals/VitalRefreshRateReaderTests.swift
```diff
@@ -0,0 +1,104 @@
...
+import XCTest
+import UIKit
+@testable import Datadog
+
+class VitalRefreshRateReaderTests: XCTestCase {
+    private let mockNotificationCenter = NotificationCenter()
+
+    func testRefreshRateReader() throws {
+        let reader = VitalRefreshRateReader(notificationCenter: mockNotificationCenter)
+        XCTAssertFalse(reader.isRunning)
```
ncreated, line 16: The `isRunning` property seems very internal to the component we test. IMO it should be made `private` and instead, we should be testing the properties / behaviours of "refresh rate monitor":
* when UI is clunky, then measured mean FPS is lower than `UIScreen.main.maximumFramesPerSecond`;
* when UI is smooth, then measured mean FPS is about `UIScreen.main.maximumFramesPerSecond`;
* when app goes to background and then back to foreground, then measured mean FPS is not impacted.

ncreated (changes requested):

Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
```diff
@@ -0,0 +1,112 @@
...
+        DispatchQueue.main.async {
+            self.observers.removeAll { existingObserver in
+                return existingObserver === observer
+            }
+        }
+    }
+
+    /// Starts listening to frame paints.
+    /// - Throws: only if `UIScreen.main` cannot generate its `CADisplayLink`
+    func start() throws {
```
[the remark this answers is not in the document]
buranmert: > IDK who will be the caller of that method
i add `start()` to `init()` so that `Reader` will be always running ✅ 

> this method should be kept simple and marked non-throwing
`UIScreen.main.displayLink()` throws but `CADisplayLink()` constructor doesn't throw.
i changed to non-throwing method ✅

Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
```diff
@@ -0,0 +1,112 @@
...
+    private func appWillResignActive() {
+        private_stop()
+    }
+
+    @objc
+    private func appDidBecomeActive() {
+        if isRunning {
+            try? private_start()
+        }
+    }
```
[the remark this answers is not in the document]
buranmert: addressing the first comment, i made `Reader` always running. that eliminates `private_start/stop()` methods and simplifies the logic ✅ 

`displayTick` is going to be called 60 times in 1 second and app state will change rarely. `start/stop` sounds like a cheaper choice than adding an early return to `displayTick`.
either way works though, if you insist i can add `isInBackground` too

Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
buranmert: i added `private_start/stop()` later and `start()` was supposed to call `private_stop()` but i forgot to change 👍 
now this logic is removed as `Reader` is made always running ✅

Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
```diff
@@ -0,0 +1,94 @@
...
+        if let lastTimestamp = self.lastFrameTimestamp {
+            let frameDuration = link.timestamp - lastTimestamp
+            let currentFPS = 1.0 / frameDuration
+            // NOTE: RUMM-1278 `oldValue` is not used
+            observers.forEach {
+                $0.onValueChanged(oldValue: 0.0, newValue: currentFPS)
+            }
+        }
+        lastFrameTimestamp = link.timestamp
+    }
```
ncreated, line 68: [We should avoid doing anything on the main thread](https://github.com/DataDog/dd-sdk-ios/blob/master/ZEN.md#small-footprint). Here, I can see two main thread penalties:
* 1/ In managing the array of observers (`observers.append()` and `observers.removeAll()`) on the `main` queue.
* 2/ In calling `onValueChanged()` on each observer, which runs [this `VitalInfo` computation](https://github.com/DataDog/dd-sdk-ios/blob/01e4c856971a8e530b1cd3d823e467895e9544c6/Sources/Datadog/RUM/RUMVitals/VitalObserver.swift#L25-L42) and [notifies the `VitalListener`](https://github.com/DataDog/dd-sdk-ios/blob/01e4c856971a8e530b1cd3d823e467895e9544c6/Sources/Datadog/RUM/RUMVitals/VitalObserver.swift#L41).

I assume that the `RUMViewScope` will be implementing the `VitalListener` interface, so it will contain:
```swift
func onVitalInfo(info: VitalInfo) {
   // ...
}
```
which will be run on the main thread. The rest of logic in `RUMViewScope` is executed [on the `background` queue](https://github.com/DataDog/dd-sdk-ios/blob/01e4c856971a8e530b1cd3d823e467895e9544c6/Sources/Datadog/RUMMonitor.swift#L130-L134). 

I don't see any clear link nor plan on how this could be synchronized later without refactoring this `VitalRefreshRateReader`. The `info` would have to be accessed synchronously [when creating the `RUMViewEvent`](https://github.com/DataDog/dd-sdk-ios/blob/01e4c856971a8e530b1cd3d823e467895e9544c6/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift#L304), but we can't perform `DispatchQueue.main.sync {}` as it will produce a hitch for each view update.

What is the idea for plugging this reader into `RUMViewScope`? 

IMO we should change the synchronisation in this class and dispatch things to background thread as early as possible to simplify further considerations and make it "thread safe" for the rest of architecture. These problems are already solved in [`NetworkInfoProvider`](https://github.com/DataDog/dd-sdk-ios/blob/01e4c856971a8e530b1cd3d823e467895e9544c6/Sources/Datadog/Core/System/NetworkConnectionInfoProvider.swift#L61-L93), [`CarrierInfoProvider`](https://github.com/DataDog/dd-sdk-ios/blob/01e4c856971a8e530b1cd3d823e467895e9544c6/Sources/Datadog/Core/System/CarrierInfoProvider.swift#L74-L110) and [`UserInfoProvider`](https://github.com/DataDog/dd-sdk-ios/blob/01e4c856971a8e530b1cd3d823e467895e9544c6/Sources/Datadog/Core/Attributes/UserInfoProvider.swift#L13-L36) with the Publish / Subscribe model and utilities introduced in https://github.com/DataDog/dd-sdk-ios/pull/422. Could we reuse the same concepts?

ncreated (commented):
I left thread synchronization question as IMO there is an unnecessary penalty on the main thread and I don't clearly see how we will further read these value form secondary thread in RUM scope. 

@xgouchet how do you synchronize publishing this vital from `main` thread and reading it from `RUM scope` thread on Android?

xgouchet:
> I left thread synchronization question as IMO there is an unnecessary penalty on the main thread and I don't clearly see how we will further read these value form secondary thread in RUM scope.
> 
> @xgouchet how do you synchronize publishing this vital from `main` thread and reading it from `RUM scope` thread on Android?

On Android the RUM View Scope doesn't read from it. What happens is the view gets notified when a new aggregation is available and stores its value locally, and when it sends the event it sends it with the latest local value it knows

Tests/DatadogTests/Datadog/RUM/RUMVitals/VitalRefreshRateReaderTests.swift
```diff
@@ -0,0 +1,78 @@
...
+        DispatchQueue.global().asyncAfter(deadline: .now() + 0.5) {
+            expectation2.fulfill()
+        }
+        waitForExpectations(timeout: 1.0) { _ in }
+
+        // Block UI thread
+        Thread.sleep(forTimeInterval: 1.5)
+
+        XCTAssertGreaterThan(observer_view2.vitalInfo.sampleCount, 0)
+        XCTAssertGreaterThan(observer_view1.vitalInfo.meanValue, observer_view2.vitalInfo.meanValue)
```
buranmert, line 49: this line is flaky. i'm looking into it...

buranmert: fixed: now we first block UI thread (lowering FPS) and then wait in background thread so that reader can read FPS **BEFORE** `XCTAssert`s

ncreated (approved):
Looks very good 👌

PR #523, opened 2021-06-21
RUMM-1397 Update Endpoints

xgouchet:
### What and why?

Update the endpoints available in the SDK and mark previous ones as deprecated.

### Review checklist

### DIFF
--- a/Sources/DatadogObjc/DatadogConfiguration+objc.swift
+++ b/Sources/DatadogObjc/DatadogConfiguration+objc.swift
@@ -18,6 +18,18 @@ public class DDEndpoint: NSObject {
     // MARK: - Public
 
     @objc
+    public static func us1() -> DDEndpoint { .init(sdkEndpoint: .us1) }
+
+    @objc
+    public static func us3() -> DDEndpoint { .init(sdkEndpoint: .us3) }
+
+    @objc
+    public static func eu1() -> DDEndpoint { .init(sdkEndpoint: .eu1) }
+
+    @objc
+    public static func us1_fed() -> DDEndpoint { .init(sdkEndpoint: .us1_fed) }
+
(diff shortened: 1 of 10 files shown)


Sources/DatadogObjc/DatadogConfiguration+objc.swift
```diff
@@ -16,6 +16,18 @@ public class DDEndpoint: NSObject {
...
+    public static func us1() -> DDEndpoint { .init(sdkEndpoint: .us1) }
+    
+    @objc
+    public static func us3() -> DDEndpoint { .init(sdkEndpoint: .us3) }
+    
+    @objc
+    public static func eu1() -> DDEndpoint { .init(sdkEndpoint: .eu1) }
+    
+    @objc
+    public static func us1_fed() -> DDEndpoint { .init(sdkEndpoint: .us1_fed) }
```
buranmert, line 30: we have `DDLogsEndpoint` / `DDTracesEndpoint` but we don't have `DDRUMEndpoint`, did we forget it or should we deprecate `DDLogsEndpoint` / `DDTracesEndpoint` as well?

buranmert (approved):
lgtm but i left a question regarding `DDRUMEndpoint`

Sources/DatadogObjc/DatadogConfiguration+objc.swift
ncreated: Both `set(logsEndpoint: DDLogsEndpoint)` and `public func set(tracessEndpoint: DDLogsEndpoint)` were deprecated when introducing common `func set(endpoint: DDEndpoint)`. Enum cases were not deprecated, but they could be.

ncreated (approved):
LG 👍

PR #525, opened 2021-06-24
RUMM-1057: Sanitize custom timings

0xnm:
### What and why?

Follow-up on https://github.com/DataDog/dd-sdk-android/pull/504 - the same sanitization of custom timings, but in iOS SDK.

### How?

This change adds sanitization of names in `custom_timings` per defined characters set.

### Review checklist


xgouchet (approved):

ncreated (approved):
Looks great 👌! Maybe we could reference this [line of `rum-events-format`](https://github.com/DataDog/rum-events-format/blob/fcd5662c4007a3225ce9abb443bbb898b513b8b1/schemas/view-schema.json#L115) in the PR description:

> As timing name is used as facet path, it must contain only letters, digits, or the characters - _ . @ $

PR #527, opened 2021-06-29
RUMM-1434 read resource span_id and trace_id from attributes

xgouchet:
### What and why?

Cross platform sdks will generate the resource's span_id and trace_id on their side and forward the attributes as `_dd.span_id` and `_dd.trace_id` respectively. This PR reads those value and store them in the relevant place.

### Review checklist


ncreated (approved):
✨ 👌

PR #530, opened 2021-07-01
RUMM-1280 Attaching Vitals to Views

buranmert:
### What and why?

Vitals (CPU, memory and screen refresh rate) info is added to RUM View events

### How?

`VitalInfoSampler` takes readers for CPU, memory and screne refresh rate; samples them at given frequency and returns vital info in thread-safe manner when requested.

### TODO

- [x] Normalize `refreshRate` with `UIScreen.main.maximumFramesPerSecond` in `RUMViewScope`

### Review checklist


Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -88,6 +90,18 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
         self.viewStartTime = startTime
         self.dateCorrection = dependencies.dateCorrector.currentCorrection
         self.crashContextIntegration = RUMWithCrashContextIntegration()
+
+        if let cpuReader = RUMFeature.instance?.vitalCPUReader,
+           let memoryReader = RUMFeature.instance?.vitalMemoryReader,
+           let refreshRateReader = RUMFeature.instance?.vitalRefreshRateReader {
```
ncreated, line 96: We should rather inject it through `dependencies: RUMScopeDependencies` so it is easier to mock in tests and we do not depend on a global state that much.

buranmert:
Screenshot from Datadog web UI:
Screenshot 2021-07-07 at 18 53 20

it seems to work ✅

Sources/Datadog/RUM/RUMVitals/VitalInfo.swift
```diff
@@ -6,16 +6,53 @@
...
+        // Assuming M(n) is the mean value of the first n samples
+        // M(n) = ∑ sample(n) / n
+        // n⨉M(n) = ∑ sample(n)
+        // M(n+1) = ∑ sample(n+1) / (n+1)
+        //        = [ sample(n+1) + ∑ sample(n) ] / (n+1)
+        //        = (sample(n+1) + n⨉M(n)) / (n+1)
+        meanValue = (sample + (sampleCountDouble * (meanValue ?? 0.0))) / (sampleCountDouble + 1.0)
+        // swiftlint:disable force_unwrapping
+        minValue = minValue == nil ? sample : min(minValue!, sample)
+        maxValue = maxValue == nil ? sample : max(maxValue!, sample)
```
ncreated, line 40: Any reason for force unwrapping instead of `if let` 💭 ?

Sources/Datadog/RUM/RUMVitals/VitalInfoSampler.swift
```diff
@@ -0,0 +1,92 @@
...
+        )
+        // NOTE: RUMM-1280 based on my running Example app
+        // non-main run loops don't fire the timer.
+        // Although i can't catch this in unit tests
+        RunLoop.main.add(timer, forMode: .default)
+        self.timer = timer
+    }
+
+    deinit {
+        self.timer?.invalidate()
```
ncreated, line 76: We have a `VitalInfoSampler` leak 🚰 - it's `deinit()` is never called. 

I profiled `Example` app ("tap action scenario") for a while and it has 19 instances remaining:

Screenshot 2021-07-08 at 12 32 29


Here init / deinit logs for `RUMViewScope` and `VitalInfoSampler`:
```
⚡️ RUMViewScope.init() [view.name = MenuView)]
  →⚡️ VitalInfoSampler.init()
⚡️ RUMViewScope.deinit(view.name = MenuView))
⚡️ RUMViewScope.init() [view.name = TableView)]
  →⚡️ VitalInfoSampler.init()
⚡️ RUMViewScope.deinit(view.name = TableView))
⚡️ RUMViewScope.init() [view.name = MenuView)]
  →⚡️ VitalInfoSampler.init()
⚡️ RUMViewScope.deinit(view.name = MenuView))
⚡️ RUMViewScope.init() [view.name = UIControlsView)]
  →⚡️ VitalInfoSampler.init()
⚡️ RUMViewScope.deinit(view.name = UIControlsView))
⚡️ RUMViewScope.init() [view.name = MenuView)]
  →⚡️ VitalInfoSampler.init()
```

Sources/Datadog/RUM/RUMVitals/VitalInfoSampler.swift
```diff
@@ -0,0 +1,92 @@
...
+        let timer = Timer(
+            timeInterval: frequency,
+            target: self,
+            selector: #selector(takeSample),
+            userInfo: nil,
+            repeats: true
+        )
+        // NOTE: RUMM-1280 based on my running Example app
+        // non-main run loops don't fire the timer.
+        // Although i can't catch this in unit tests
```
ncreated, line 70: What can't be observed in unit tests? I'm not sure if I understand this comment - do we want to keep it?

Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
```diff
@@ -54,7 +60,12 @@ internal class VitalRefreshRateReader {
         if let lastTimestamp = self.lastFrameTimestamp {
             let frameDuration = link.timestamp - lastTimestamp
             let currentFPS = 1.0 / frameDuration
-            valuePublisher.publishAsync(currentFPS)
+
+            for publisher in valuePublishers {
+                publisher.mutateAsync { currentInfo in
+                    currentInfo.addSample(currentFPS)
+                }
+            }
```
ncreated, line 68: Why not having one single publisher for the `currentFPS` value? This way we won't need to run blocks on `DispatchQueue.main` for every subscription / unsubscription (for every new view started) and the overall threading model could be simpler:

```swift
// In a single instance of refresh rate reader; calling this non-blocking "publish" on the main thread:
publisher.publishAsync(currentFPS)

// In each sampler (one sampler per view); every 1 second; calling this blocking read from `RUMViewScope` thread:
let lastFPS = publisher.value
```

ncreated (changes requested):
I get the idea with `mutateAsync` - it looks good 👍. However, this PR needs some more work as it introduces memory leak.

Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
ncreated: For what I see - this subscribe / unsubscribe introduces [the leak](https://github.com/DataDog/dd-sdk-ios/pull/530#discussion_r666059664) which makes it worth having a simpler threading model around the `FPS` metric.

buranmert: `VitalInfoSampler` also needs to `refreshRateReader.unregister(refreshRatePublisher)` in `deinit()`, this was another bug. fixing this now 👍

buranmert: memory leak comes from the `timer`. [Apple docs](https://developer.apple.com/documentation/foundation/timer/1408356-init) says
> `target`: The timer maintains a strong reference to this object until it (the timer) is invalidated.

`VitalInfoSampler` retains `timer` as instance variable and `timer` retains `VitalInfoSampler` as `target`. i'm looking into it 👀

buranmert: here `ValuePublisher` doesn't actually _publish_ anything reactively.
it is used as a wrapper for one single `value: VitalInfo` so that every `RUMViewScope` will create its own subscriber and will be getting its own value.
if there are 5 different `RUMViewScope`s at once, each needs different `VitalInfo` values; so each needs to have its own `ValuePublisher`.

if we want to use a single publisher it will be `reader -> reader's publisher -> publisher's publishers -> receiver`.
i can't think of a case where `DispatchQueue.main.async { array.append(element) }` can cause any issues, do you have a case in your mind?

buranmert: ✅ retain cycle is fixed here: https://github.com/DataDog/dd-sdk-ios/pull/530/files#diff-f15ba741c3c811beb0029e017dd16bd5271a7d25c1d0d819a2f0de509fb271d9R61-R66

```swift
let timer = Timer(...) { [weak self] _ in
  self?.takeSample()
}
```

Sources/Datadog/RUM/RUMVitals/VitalInfoSampler.swift
buranmert: i added [this unit test](https://github.com/DataDog/dd-sdk-ios/pull/530/files#diff-c783872e2fe3a78e3fddd755e1d35685db2474c6df444818b0d4456eab6e4fcbR49) to catch cases where sampler is created in background thread so its timer wouldn't fire.
however this test never fails, regardless using `RunLoop.main` or `RunLoop.current` 🤔 

although when i run Example app with `RunLoop.current`, i can observe that the timer isn't fired

i think we should keep a comment here, the wording can be different

Sources/Datadog/RUM/RUMVitals/VitalInfoSampler.swift
buranmert: fixed here: https://github.com/DataDog/dd-sdk-ios/pull/530#discussion_r666157508

Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
ncreated: >  can't think of a case where `DispatchQueue.main.async { array.append(element) }` can cause any issues, do you have a case in your mind?

Its current impact is negligible (one sync for each view transition), but this model would need to be changed if we want to add more things there in the future. It's OK for now if we consider this not harmful 👍.

ncreated (commented):
The memory leak💧is gone  - nice fix 👍

Sources/Datadog/RUM/RUMVitals/VitalInfo.swift
buranmert: just for the sake of brevity, it becomes longer otherwise
```swift
if let someMin = minValue {
  minValue = min(someMin, sample)
} else {
  minValue = sample
}
// x2, same for max
```

can you think of any potential problems with force unwrap here?

ncreated: Force unwrapping is generally considered a bad practise hence we control it with linter (explicitly disabled here). It won't cause any issue in this case 👌.

ncreated (approved):

PR #534, opened 2021-07-06
RUMM-1454 fix RUM view path

xgouchet:
### What and why?

Fix discrepancy with Android view path/url.

### How?

We keep the `key` used to start a RUM View as a view url/path. This way if the user provides a `name`, we still have a technical `url` which won't change based on localisation, app version etc…


buranmert (approved):

buranmert:
`DatadogIntegrationTests/Scenarios/RUM/RUMManualInstrumentationScenarioTests.swift:102` failed ❌ 
```swift
XCTAssertEqual(view3.path, "SendRUMFixture3View")
```

ncreated (approved):
👌

PR #536, opened 2021-07-07
RUMM-1386 Implement new format for `DDCrashReports`

ncreated:
### What and why?

📦  This PR updates the way we format crash reports for RUM and Logging integrations.

This new format gives us more flexibility over crash report elements - instead of using free-form text format, now we send each thread information separately along with binary images. Detailed discussion is available in internal _"RFC - iOS Crash Reports Minimization"_, but the gist of it are these 6 fields:
* `error.type` - the [process exception type](https://developer.apple.com/documentation/xcode/diagnosing_issues_using_crash_reports_and_device_logs/understanding_the_exception_types_in_a_crash_report),
* `error.message` - uncaught exception message (if available) or  termination signal details,
* `error.stack` - the stack trace of uncaught exception (if available) or backtrace of crashed thread,
* `error.threads[]` - an array of JSON objects describing each thread in the process (along with its `.stack`),
* `error.binary_images[]` - an array of JSON objects each describing a binary image loaded by the process,
* `error.meta.*` - selected fields from [crash report header](https://developer.apple.com/documentation/xcode/examining-the-fields-in-a-crash-report).

### How?

Raw information about the crash is read from [`PLCrashReport`](https://github.com/microsoft/plcrashreporter/blob/master/Source/PLCrashReport.h) object produced by [PLCR](https://github.com/microsoft/plcrashreporter). It is first transformed into internal representation. The intermediate `CrashReport` is passed to `DDCrashReportExporter` which produces final `DDCrashReport`:
```
PLCrashReport → CrashReport → DDCrashReport
```

The intermediate `CrashReport` structure solves few problems:
* Bridges dynamic types from Obj-c to their explicit types in Swift:
```swift
plCrashReport.threads: [Any]! // obj-c
crashReport.threads: [Thread] // swift
```
* Bridges implicitly unwrapped optionals from Objc-interop to explicit optionals in Swift:
```swift
plCrashReport.exceptionInfo: PLCrashReportExceptionInfo! // obj-c
crashReport.exceptionInfo: ExceptionInfo? // swift
```
* Infers additional information requires by our new format:
```swift
plBinaryImage.imagePath // objc
binaryImage.isSystemLibrary // additional swift value inferred from `imagePath`
```

Pointers arithmetic operations implemented in `DDCrashReportExporter` are based on the [`PLCrashReportTextFormatter` implementation](https://github.com/microsoft/plcrashreporter/blob/master/Source/PLCrashReportTextFormatter.m) from PLCR. Stack traces exported in our format are equal to the ones exported from PLCR, e.g.:

```
0    DatadogCrashReporting               0x00000001086e3830 0x1086a9000 + 239664
1    DatadogCrashReportingTests          0x000000010bf68b6e 0x10bf60000 + 35694
2    DatadogCrashReportingTests          0x000000010bf6b078 0x10bf60000 + 45176
3    CoreFoundation                      0x00007fff20429a3c 0x7fff20311000 + 1149500
4    CoreFoundation                      0x00007fff20426c6f 0x7fff20311000 + 1137775
5    XCTest                              0x0000000109dff37b 0x109dd6000 + 168827
6    XCTest                              0x0000000109ea2e9c 0x109dd6000 + 839324
```

### Review checklist


Sources/Datadog/CrashReporting/DDCrashReportingPluginType.swift
```diff
@@ -10,14 +10,140 @@ import Foundation
...
+    /// Ref.: https://developer.apple.com/documentation/xcode/examining-the-fields-in-a-crash-report
+    public struct Meta: Encodable {
+        /// A client-generated 16-byte UUID of the incident.
+        internal let incidentIdentifier: String?
+        /// The name of the crashed process.
+        internal let processName: String?
+        /// Parent process information.
+        internal let parentProcess: String?
+        /// The location of the executable on disk.
+        internal let path: String?
```
mariusc83, line 91: Just FMI what is exactly this executable ?

Sources/DatadogCrashReporting/PLCrashReporterIntegration/DDCrashReportExporter.swift
```diff
@@ -0,0 +1,190 @@
...
+            )
+        }
+    }
+
+    private func binaryImages(from crashReport: CrashReport) -> [DDCrashReport.BinaryImage] {
+        return crashReport.binaryImages.map { image in
+            // Ref. for this computation:
+            // https://github.com/microsoft/plcrashreporter/blob/dbb05c0bc883bde1cfcad83e7add25862c95d11f/Source/PLCrashReportTextFormatter.m#L447
+            let loadAddressHex = "0x\(image.imageBaseAddress.toHex)"
+            var maxAddressOffset = image.imageSize.subtractIfNoOverflow(1) ?? image.imageSize
```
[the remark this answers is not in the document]
ncreated: `imageSize` is an unsigned integer and here we check for reaching overflow below `0` (by subtracting `1` from `UInt`). Although images cant have negative size and (perhaps) won't have the size of `0`, this is a sanity check to avoid overflow crash if sth goes wrong. We retrieve this `image.imageSize` value from `PLCrashReporter` (3rd party), which reads it from the system (it's read from `C` to `Obj-C` to `Swift`), so IMO extra safety is a good idea here 🙂.

Sources/Datadog/CrashReporting/DDCrashReportingPluginType.swift
ncreated: This is the location of the app's executable - it's main binary. An example value of this for `SimpleApp` would be:
```
/private/var/containers/Bundle/Application/10C61F35-5173-4AD9-8E3A-1C6C73478448/SimpleApp.app/SimpleApp
```

Sources/DatadogCrashReporting/PLCrashReporterIntegration/DDCrashReportExporter.swift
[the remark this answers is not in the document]
mariusc83: Tks for the explanation, yes I do agree that extra safety will not harm.

Sources/Datadog/CrashReporting/DDCrashReportingPluginType.swift
```diff
@@ -10,14 +10,140 @@ import Foundation
...
+    /// In Datadog Error Tracking this corresponds to `error.stack`.
+    internal let stack: String
+    /// All threads running in the process.
+    internal let threads: [Thread]
+    /// List of binary images referenced from all stack traces.
+    internal let binaryImages: [BinaryImage]
+    /// Meta information about the crash and process.
+    internal let meta: Meta
+    /// If any stack trace information was truncated due to crash report minimization.
+    internal let wasTruncated: Bool
```
mariusc83, line 146: What happens if it gets truncated ? Where you will send the `stacktrace` continuation or you will not send it at all ?

ncreated: Truncation is added in https://github.com/DataDog/dd-sdk-ios/pull/538, we remove less meaningful frames and replace them with `...`, for example:

124959128-3140c800-e01b-11eb-86df-03053de89f71

Sources/DatadogCrashReporting/PLCrashReporterIntegration/DDCrashReportExporter.swift
```diff
@@ -0,0 +1,190 @@
...
+
+            let knownSignalNames = Mirror(reflecting: sys_signame)
+                .children
+                .compactMap { $0.value as? UnsafePointer<Int8> }
+                .map { String(cString: $0).uppercased() } // [HUP, INT, QUIT, ILL, TRAP, ABRT, ...]
+
+            let knownSignalDescriptions = Mirror(reflecting: sys_siglist)
+                .children
+                .compactMap { $0.value as? UnsafePointer<Int8> }
+                .map { String(cString: $0) } // [Hangup, Interrupt, Quit, Illegal instruction, ...]
```
mariusc83, line 70: Shouldn't you keep this in a static or class dictionary  instead of computing them all the time ? Are they depending in any way on this method arguments or am I missing something ?

mariusc83 (commented):
LGTM just some minor comments here and there and one suggestion regarding some optimisations

buranmert (approved):
lgtm

Sources/DatadogCrashReporting/PLCrashReporterIntegration/DDCrashReportExporter.swift
ncreated: Good idea 👍. We can consider these values constant as it's hard to imagine that this will change in the OS internals. I changed to use static dictionary in implementation and added OS-value inference in tests (this way we will spot that it changed eventually). Done in https://github.com/DataDog/dd-sdk-ios/pull/536/commits/2d68a62623d101c4ae75f83bd8ad16e5aca64590.

mariusc83 (approved):

PR #537, opened 2021-07-08
RUMM-1269 Add configuration to enable/disable background RUM events

xgouchet:
### What and why?

Because tracking background event can lead to more sessions being tracked, we need to disable this feature by default and let customers enable/disable it in the configuration


mariusc83 (commented):
LGTM

ncreated (approved):
All good 💪👌

buranmert (approved):

PR #538, opened 2021-07-08
RUMM-1432 Minify the crash report

ncreated:
### What and why?

📦 This PR implements minification algorithm for crash reports as discussed in the internal _"RFC - iOS Crash Reports Minimization"_. The goal is to reduce the amount of information in crash reports to make sure it fits ingestion limits for Datadog.

Two things are reduced before exporting the `DDCrashReport`:
* We remove **binary images** which are not referenced from any stack frame. Such images are not important for symbolication process. A typical iOS app loads ~300-600 images, but only a fraction of them (approximately few %) is referenced among all stack traces.
* We enforce each **stack trace** to be under 200 frames limit. For stack traces which are bigger than that, middle frames are removed and replaced with `...` mark. This shouldn't impact much reports, as the limit of `200` is pretty high when looking at "typical" iOS crash reports.

### How?

New component was added to the architecture introduced in #536 : `CrashReportReducer`. It takes the intermediate `CrashReport` and produces its minified version. Report processing now looks this:
```
PLCrashReport → (original) CrashReport → (minified) CrashReport → DDCrashReport
```

Truncated stack trace will look this (this screenshot was taken when debugging the limit of `5`):
Screenshot 2021-07-08 at 18 34 17

### Review checklist

### DIFF
--- a/Sources/DatadogCrashReporting/PLCrashReporterIntegration/CrashReport.swift
+++ b/Sources/DatadogCrashReporting/PLCrashReporterIntegration/CrashReport.swift
@@ -28,6 +28,8 @@ internal struct CrashReport {
     var binaryImages: [BinaryImageInfo]
     /// Custom user data injected before the crash occured.
     var contextData: Data?
+    /// Additional flag (for telemetry) meaning if any of the stack traces was truncated due to minification.
+    var wasTruncated: Bool
 }
 
 /// Intermediate representation of `PLCrashReportSystemInfo`.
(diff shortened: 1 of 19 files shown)


Sources/DatadogCrashReporting/PLCrashReporterIntegration/CrashReport.swift
```diff
@@ -104,6 +106,9 @@ internal struct BinaryImageInfo {
 
 /// Intermediate representation of `PLCrashReportStackFrameInfo`.
 internal struct StackFrame {
+    /// The number of this frame in the stack trace.
+    /// This must be recorded as less meaningful stack frames might be removed when minifying the `CrashReport`.
+    var number: Int
```
mariusc83, line 111: how is this being used...the way I see it you remove from the middle of the set not from the end...

mariusc83 (approved):
LGTM

buranmert (approved):
looks good ✅ 
the only minor issue is we used the term "minification" so far, now we have a component named `Reducer`. that is fine but it could be named `Minifier` as well in order to keep the existing lingo, wdyt?

ncreated:
> the only minor issue is we used the term "minification" so far, now we have a component named `Reducer`. that is fine but it could be named `Minifier` as well in order to keep the existing lingo, wdyt?

This makes sense 👌 - I made the renaming so we can have unified lingo 👍.

buranmert (approved):

PR #539, opened 2021-07-09
RUMM-1340 & RUMM-1341: Add a way to attach user attributes for automatic actions and allow users to rename them

0xnm:
### What and why?

This change introduces a protocol which users can use to attach custom attributes to the automatic actions and also to rename them.

### How?

The possibility to attach custom attributes and rename actions which are collected automatically is achieved by introducing the following predicate:

```
func rumAction(targetView: UIView) -> RUMAction?
```

It should return the following `RUMAction` with fields: `name: String` and `attributes: [AttributeKey: AttributeValue`. These fields are then consumed by the `.addUserAction` command. If `nil` is returned, then action is dropped and won't be recorded.

The change is done with keeping the consistency with existing `UIKitRUMViewsPredicate` protocol and `trackUIKitViews(predicate)` method. That is why (for the consistency) old method `trackUIKitActions(boolean)` was deprecated and new method `trackUIKitRUMActions(predicate)` is introduced (both name and argument changed).

### Review checklist


mariusc83 (approved):
LGTM

docs/rum_collection.md
```diff
@@ -164,18 +164,27 @@ This will make the SDK track requests sent from this instance of the `URLSession
...
-   .trackUIKitActions()
+   .trackUIKitRUMActions()
    .build()
 
 Global.rum = RUMMonitor.initialize()
 ```
 
 This makes the SDK track all significant taps occurring in the app. For privacy reasons, all interactions with the on-screen keyboard are ignored.
 
+To customize RUM views tracking, use `.trackUIKitRUMActions(using: predicate)` and provide your own implementation of the `predicate` which conforms to `UIKitRUMUserActionsPredicate` protocol:
```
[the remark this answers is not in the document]
0xnm: yes. I've updated the doc - now it only contains method name update (because `trackUIKitActions` is deprecated now), it doesn't mention predicate.

Sources/Datadog/RUM/AutoInstrumentation/Actions/UIKitRUMUserActionsHandler.swift
```diff
@@ -46,14 +48,16 @@ internal class UIKitRUMUserActionsHandler: UIKitRUMUserActionsHandlerType {
...
         }
 
-        subscriber?.process(
-            command: RUMAddUserActionCommand(
-                time: dateProvider.currentDate(),
-                attributes: [:],
-                actionType: .tap,
-                name: targetName(for: actionTargetView)
+        if let rumAction = userActionsPredicate.rumAction(targetView: actionTargetView) {
+            subscriber?.process(
```
buranmert, line 52: nit-picking: if we first nil-check `subscriber`, we can avoid unnecessary calls to `userActionsPredicate`
```suggestion
        if let subscriber = subscriber,
           let rumAction = userActionsPredicate.rumAction(targetView: actionTargetView) {
            subscriber.process(
```

buranmert (approved):
lgtm

ruthnaebeck (approved):
👍  for docs

Sources/Datadog/RUM/AutoInstrumentation/Actions/UIKitRUMUserActionsHandler.swift
0xnm: Indeed, thanks for the suggestion. I've added early return if `subscriber` is not set.

PR #540, opened 2021-07-12
Add support for DatadogSDKTesting to DatadogCrashReportingTests

nachoBonafonte:
### What and why?
`PLCrashReporter` symbols are duplicated both in `DatadogCrashReporting` and `DatadogSDKTesting`, this was producing, apart from warnings, that `PLCR`symbols were found in the `DatadogSDKTesting` framework which made the test fail. The reason for that is because `DatadogSDKTesting` framework was being linked first, as its liking was configured in the `.xcconfig` file and the `DatadogCrashReporting` linking was added later in the Build phase.

### How?

We add the linking with the `DatadogCrashReporting` framework before the linking with `DatadogSDKTesting`, which makes the linker find its symbols as expected

### Review checklist

### DIFF
--- a/Datadog/TargetSupport/DatadogCrashReportingTests/DatadogCrashReportingTests.xcconfig
+++ b/Datadog/TargetSupport/DatadogCrashReportingTests/DatadogCrashReportingTests.xcconfig
@@ -1,11 +1,7 @@
 // Add common settings from Datadog.xcconfig
 #include "../xcconfigs/Datadog.xcconfig"
 
-// Add DatadogSDKTesting instrumentation (if available in current environment)
-//#include "../xcconfigs/DatadogSDKTesting.local.xcconfig"
-//
-// Note: The 'DatadogSDKTesting' instrumentation is temporarily disabled as it causes issues in 'DatadogCrashReportingTests'.
-//       Both 'DatadogSDKTesting' and 'DatadogCrashReporting' link PLCR dependency statically, resulting with these runtime warnings:
+// Both 'DatadogSDKTesting' and 'DatadogCrashReporting' link PLCR dependency statically, resulting with these runtime warnings:
 //
 //       """
 //       objc[50310]: Class PLCrashReportThreadInfo is implemented in both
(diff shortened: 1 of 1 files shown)


Datadog/TargetSupport/DatadogCrashReportingTests/DatadogCrashReportingTests.xcconfig
```diff
@@ -15,5 +11,11 @@
...
 //       """
 //
-//       Because some mocks in 'DatadogCrashReportingTests' use PLCR type's subclassing, their runtime cast with 'as? <PLCR_TYPE>' fails
-//       resulting with tests failures. Until a workaround or a better solution is found, this instrumentation must be disabled.
+//       Because some mocks in 'DatadogCrashReportingTests' use PLCR type's subclassing, their runtime cast with 'as? <PLCR_TYPE>' could fail depending in which library is found first
+// We force 'DatadogCrashReporting' to be linked before linking 'DatadogSDKTesting' so PLCR symbols are found in that library.
+OTHER_LDFLAGS= $(inherited) -framework DatadogCrashReporting
+
+// Add DatadogSDKTesting instrumentation (if available in current environment)
+ #include "../xcconfigs/DatadogSDKTesting.local.xcconfig"
```
buranmert, line 19: we can have `#include?` here if `DatadogSDKTesting.local.xcconfig` is not essential

buranmert (approved):
lgtm apart from a minor improvement

Datadog/TargetSupport/DatadogCrashReportingTests/DatadogCrashReportingTests.xcconfig
nachoBonafonte: But it is essential, right? I mean, including it is the only reason for the existence of this file (DatadogCrashReportingTests.xcconfig), it not Datadog.xcconfig could be used instead

buranmert: we have it optional in [other places](https://github.com/DataDog/dd-sdk-ios/blob/master/Datadog/TargetSupport/DatadogTests/DatadogTests.xcconfig#L5), but i don't know what is different with this file.

if everything can still work without `local.xcconfig`, let's make it optional. does that make sense?

nachoBonafonte: Ok, I will change

buranmert (approved):

PR #543, opened 2021-07-15
RUMM-1492 isSystemImage property was wrong in simulator

buranmert:
### What and why?

`isSystemImage` property was set by searching for `/Bundle/Application/` in the image paths.
This works well when running the app in devices where image paths can be like the followings:
```
// binaries in the App bundle:
/private/var/containers/Bundle/Application/00...00/Example.app/Example
/private/var/containers/Bundle/Application/00...00/Example.app/Frameworks/DatadogCrashReporting.framework/DatadogCrashReporting

// binaries coming from iOS
/System/Library/PrivateFrameworks/UIKitCore.framework/UIKitCore
/usr/lib/system/libdyld.dylib
/usr/lib/system/libsystem_kernel.dylib
/usr/lib/system/libsystem_c.dylib
```

However, this check gives wrong results when running in iOS simulator where non-system images do not have to be in the App bundle:
```
// non-system binary
~/Library/Developer/Xcode/DerivedData/Datadog-abcd/Build/Products/Release-iphonesimulator/DatadogCrashReporting.framework/DatadogCrashReporting

// binaries coming from iOS
'~/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS.simruntime/Contents/Resources/RuntimeRoot/System/Library/Frameworks/CoreFoundation.framework/CoreFoundation
```

### How?

As seen in the examples above, non-system binaries come from `Derived Data` path which is customizable and system binaries come from `{Xcode.app path}/Contents/Developer/Platforms` which is at least fixed.

Therefore I added `Contents/Developer/Platforms` sub-path to our `isSystemImage` detection logic.

### Review checklist

### DIFF
--- a/Sources/DatadogCrashReporting/PLCrashReporterIntegration/CrashReport.swift
+++ b/Sources/DatadogCrashReporting/PLCrashReporterIntegration/CrashReport.swift
@@ -241,7 +241,8 @@ extension BinaryImageInfo {
 
         self.uuid = imageUUID
         self.imageName = URL(string: imagePath)?.lastPathComponent
-        self.isSystemImage = !imagePath.contains("/Bundle/Application/")
+        // NOTE: RUMM-1492 refer to JIRA ticket or `CrashReportTests.swift` to see imagePath examples
+        self.isSystemImage = !imagePath.contains("/Bundle/Application/") || imagePath.contains("/Contents/Developer/Platforms/")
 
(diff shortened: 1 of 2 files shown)


mariusc83 (approved):
LGTM

ncreated (approved):
Seems good 👌

PR #547, opened 2021-07-27
RUMM-728: Add a warning when event is received, but there is no active view scope

0xnm:
### What and why?

Mirror of the change https://github.com/DataDog/dd-sdk-android/pull/375 done in Android SDK before: it adds a warning if there is no view scope active when any event other than events which create view scopes is received.

### Review checklist

### DIFF
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
@@ -118,7 +118,17 @@ internal class RUMSessionScope: RUMScope, RUMContextProvider {
         }
 
         // Propagate command
-        viewScopes = manage(childScopes: viewScopes, byPropagatingCommand: command)
+        if !viewScopes.isEmpty {
+            viewScopes = manage(childScopes: viewScopes, byPropagatingCommand: command)
+        } else {
+            userLogger.warn(
+                """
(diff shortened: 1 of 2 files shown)


Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
```diff
@@ -118,7 +118,17 @@ internal class RUMSessionScope: RUMScope, RUMContextProvider {
         }
 
         // Propagate command
-        viewScopes = manage(childScopes: viewScopes, byPropagatingCommand: command)
+        if !viewScopes.isEmpty {
+            viewScopes = manage(childScopes: viewScopes, byPropagatingCommand: command)
+        } else {
+            userLogger.warn(
+                """
+                A RUM event was detected, but no view is active. To track views automatically, try calling the
```
buranmert, line 126: we can also log `command` in the message, this may help the developer with debugging the problem
```swift
\(String(describing: command)) is detected but no view is active...
```

xgouchet (approved):

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
0xnm: Indeed, this can help even more. Thanks for the suggestion, I've updated the code.

buranmert (approved):

PR #548, opened 2021-07-27
Release `1.7.0-beta2`

buranmert:
### What and why?

This PR prepares 1.7.0-beta2 submission.

### How?

Besides regular make bump I re-enabled crash reporting for Swift Package Manager only.

Once this PR is approved, I will add a pre-release tag.

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference

PR #549, opened 2021-07-28
Merge 1.7.0-beta2

buranmert:
1.7.0-beta2 is released [here](https://github.com/DataDog/dd-sdk-ios/pull/548#issuecomment-887621266)
This PR resolves conflicts and merges that release into `master`


0xnm (approved):

dependency-manager-tests/spm/SPMProject/ViewController.swift
```diff
@@ -18,6 +19,7 @@ internal class ViewController: UIViewController {
             trackingConsent: .granted,
             configuration: Datadog.Configuration
                 .builderUsing(clientToken: "abc", environment: "tests")
+                .enableCrashReporting(using: DDCrashReportingPlugin())
```
mariusc83, line 22: shouldn't this be enable by default ?

buranmert: is it so in Android?
as a principle, nothing is enabled by default.

Issue #131, opened 2020-06-10
Cannot install with cocoapods without use_frameworks


amloelxer:
### The issue

Hey all, first thanks so much working on this, I'm excited to give it a whirl in our app. When trying to install with cocoapods, I get the following error: `Using Swift static libraries with custom module maps is currently not supported. Please build DatadogSDK as a framework or remove the custom module map`. Unfortunately for my team, we can't move to `use_frameworks` and need to keep modules. When looking at your custom Datadog module map, I see it is eerily similar to an autogenerated one with the exception of the sudo private module you have set exposing `_Datadog_Private`
```
framework module _Datadog_Private {
  export *
  header "ObjcExceptionHandler.h"
}
```
Any reason this can't just be exposed in DatadogObjc and let the autogenerated module map be used (other than the fact you want to discourage 3d party devs from consuming `ObjcExceptionHandler.m`)? Or any easy workarounds short of manually integrating into our project? Thank you for your time. 

- Alex 

---

#### Datadog SDK version:

DatadogSDK (1.2.1)

#### Last working Datadog SDK version:

First time using

#### Dependency Manager:

Cocoapods 1.9.1

#### Other toolset:

N / A

#### Xcode version:

Version 11.5 (11E608c)

#### Swift version:

Swift 5 

#### Deployment Target:

iOS 11

#### macOS version:

Mac OS Catalina v 10.15.2

buranmert:
hi @amloelxer 👋  thanks for reporting and looking into the issue!

you are right, [we use a custom `modulemap` in order to use ObjC in Swift while hiding it from our users](https://github.com/DataDog/dd-sdk-ios/pull/80). 
unfortunately, [we can't remove it at the moment](https://github.com/DataDog/dd-sdk-ios/pull/64).

you can use Carthage or Swift Package Manager without this problem, but i assume you use only Cocoapods in your projects

you can try [the snippet below in your `Podfile`s](https://stackoverflow.com/a/60914505/1515075):
```ruby
target 'Your-Target-Name' do
  use_frameworks!
  pod 'DatadogSDK'
  pod 'YourOtherPod'
  ...
  target 'Your-TestTarget-Name' do
    inherit! :search_paths
    pod 'YourTestPod'
    ...
  end
end

dynamic_frameworks = ['DatadogSDK','DatadogSDKObjc']
pre_install do |installer|
  installer.pod_targets.each do |pod|
    if !dynamic_frameworks.include?(pod.name)
[...]

PR #553, opened 2021-08-05
DatadogSDKCrashReporting.podspec added

buranmert:
### What and why?

`DatadogSDK` was using a custom `modulemap` and therefore Cocoapods was enforcing `use_frameworks!` when you have `DatadogSDK` pod.
On the other hand, `PLCrashReporter` needs to be statically linked and these 2 limitations were contradicting.
As a result, `DatadogCrashReporting` didn't support Cocoapods in the first place.

### How?

Since `DatadogSDK` doesn't require `use_frameworks!` anymore, we simply added `DatadogSDKCrashReporting.podspec` and it works.

⚠️  Please note that `PLCrashReporter` still needs to linked statically, so you can't use `use_frameworks!` with `DatadogSDKCrashReporting` pod.

### Review checklist

### DIFF
--- a/DatadogSDKCrashReporting.podspec
+++ b/DatadogSDKCrashReporting.podspec
@@ -10,8 +10,7 @@ Pod::Spec.new do |s|
   s.license            = { :type => "Apache", :file => 'LICENSE' }
   s.authors            = { 
     "Maciek Grzybowski" => "maciek.grzybowski@datadoghq.com",
-    "Mert Buran" => "mert.buran@datadoghq.com",
-    "Alexandre Costanza" => "alexandre.costanza@datadoghq.com"
+    "Mert Buran" => "mert.buran@datadoghq.com"
   }
 
   s.swift_version      = '5.1'
--- a/DatadogSDKCrashReporting.podspec
+++ b/DatadogSDKCrashReporting.podspec
@@ -1,0 +1,25 @@
+Pod::Spec.new do |s|
+  s.name         = "DatadogSDKCrashReporting"
+  s.module_name  = "DatadogCrashReporting"
+  s.version      = "1.7.0-beta2"
+  s.summary      = "Official Datadog Crash Reporting SDK for iOS."
+  
+  s.homepage     = "https://www.datadoghq.com"
+  s.social_media_url   = "https://twitter.com/datadoghq"
+
+  s.license            = { :type => "Apache", :file => 'LICENSE' }
+  s.authors            = { 
+    "Maciek Grzybowski" => "maciek.grzybowski@datadoghq.com",
+    "Mert Buran" => "mert.buran@datadoghq.com",
+    "Alexandre Costanza" => "alexandre.costanza@datadoghq.com"
+  }
+
+  s.swift_version      = '5.1'
+  s.ios.deployment_target = '11.0'
+
+  s.source = { :git => 'https://github.com/DataDog/dd-sdk-ios.git', :tag => s.version.to_s }
+
+  s.source_files = "Sources/DatadogCrashReporting/**/*.swift"
+  s.dependency 'DatadogSDK', '1.7.0-beta2'
+  s.dependency 'PLCrashReporter'
+end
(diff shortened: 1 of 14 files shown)


ncreated (commented):
> ⚠️ Please note that PLCrashReporter still needs to linked statically, so you can't use `use_frameworks!` with DatadogSDKCrashReporting pod.

Lack of [dynamic linking support is a known won't fix issue in PLCR](https://github.com/microsoft/plcrashreporter/issues/128) and it was the reason why we decided to go with SPM and Carthage first until we find a way to fully support CP.

I'm not fully convinced of "no `use_framework!`" requirement. On one side, it brings `DatadogCrashReporting` to CP users who do not use this option, but on the other it forces static linking for those who don't - otherwise `pod install` fails with error. It will very likely cause confusion and require support.

Hence, I'm not sure which is better:
* starting with partial support and explaining this limitation in docs,
* working on the full support (for both with and without `use_frameworks!`).

WDYT @buranmert  @xgouchet ?

xgouchet:
I think that for now we can provide a partial support while explaining the situation and then work on getting a full support. Also there are CP plugins to mitigate this issue and selectively apply static/dynamic linking for each dependency

buranmert:
that can solve the problem for the users who needs `use_frameworks!`:
```ruby
static_frameworks = ['DatadogSDKCrashReporting']
pre_install do |installer|
  installer.pod_targets.each do |pod|
    if static_frameworks.include?(pod.name)
      def pod.static_framework?;
        true
      end
      def pod.build_type;
        Pod::BuildType.static_library
      end
    end
  end
end
```
we can mention it in Readme and pin it in GH issues.
i took [our workaround for static linking users](https://github.com/DataDog/dd-sdk-ios/issues/131#issuecomment-642602206) and changed it

DatadogSDKCrashReporting.podspec
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  s.swift_version      = '5.1'
+  s.ios.deployment_target = '11.0'
+
+  s.source = { :git => 'https://github.com/DataDog/dd-sdk-ios.git', :tag => s.version.to_s }
+
+  s.source_files = "Sources/DatadogCrashReporting/**/*.swift"
+  s.dependency 'DatadogSDK', '1.7.0-beta2'
+  s.dependency 'PLCrashReporter'
```
ncreated, line 24: This needs a version requirement. `1.8.1` is what we use for Carthage and SPM.

dependency-manager-tests/cocoapods/Makefile
```diff
@@ -10,5 +10,5 @@ test:
 		@echo "⚙️  Configuring CPProject with remote branch: '${GIT_REFERENCE}'..."
 		@sed "s|REMOTE_GIT_REFERENCE|${GIT_REFERENCE}|g" Podfile.src > Podfile
 		@rm -rf Pods/
-		pod update
+		pod install
```
ncreated, line 13: Why changing to `pod install`? If I have a `Podfile.lock` in my local (it's gitignored) and want to try out my new branch, `make test-cocoapods` will install the previous setup for me instead of updating to the new one.

tools/distribution/push_podspecs.sh
```diff
@@ -16,6 +16,15 @@ while [[ $alamofireRetVal -ne 0 ]] ; do
    	alamofireRetVal=$? ;
 done
 
+crashReportingRetVal=1
+while [[ $crashReportingRetVal -ne 0 ]] ; do
+	sleep 30s ;
+	pod repo update ;
+    pod spec lint --allow-warnings DatadogSDKCrashReporting.podspec ;
+   	crashReportingRetVal=$? ;
```
ncreated, line 24:
```suggestion
   sleep 30s ;
   pod repo update ;
   pod spec lint --allow-warnings DatadogSDKCrashReporting.podspec ;
   crashReportingRetVal=$? ;
```

ncreated (changes requested):
> that can solve the problem for the users who needs `use_frameworks!`:
> ```ruby
> static_frameworks = ['DatadogSDKCrashReporting']
> pre_install do |installer|
>   installer.pod_targets.each do |pod|
>     if static_frameworks.include?(pod.name)
>       def pod.static_framework?;
>         true
>       end
>       def pod.build_type;
>         Pod::BuildType.static_library
>       end
>     end
>   end
> end
> ```

Awesome 👌. How about using this in our `dependency-manager-tests/cocoapod` fixture? We can link `DatadogSDK` dynamically and `DatadogCrashReporting` statically in there to ensure that we never break this setup.

ncreated (approved):
Awesome 🚀

PR #555, opened 2021-08-06
RUMM-1243 Fix crash context deserialization if it contains `null` value

ncreated:
### What and why?

🐞 This PR fixes a bug where `CrashContext` decoding was failing if a `null` value was written in the JSON.

Because `CrashContext` is necessary for building the `RUMError` (it must be associated with `RUMView` which is encoded as part fo the `CrashContext`), the Crash Reporting feature was failing to upload crashes that include `null` value set in last view event attributes.

### How?

Swift uses `null` to represent absent optional value in JSON payload:
```swift
let absentString: String? = nil

let jsonData = try JSONEncoder().encode(absentString)
print(String(data: jsonData, encoding: .utf8)!)

// null
```

This PR improves our `Encodable` type erasure (now called `CodableValue`) to handle `null` value. When decoding `null` it stores internal `CodableNull()` value which is later used to encode `null` back into serialized JSON. This way we guarantee that user data representation is not alerted by our read / write operations.

### Review checklist


buranmert (approved):
👍

PR #556, opened 2021-08-09
RUMM-1548 Fix compiler warnings in unit tests + fix deprecated endpoints

ncreated:
### What and why?

🧽 This PR fixes compiler warnings emitted when building unit tests and changes deprecated Logs + Trace URLs to recommended values. It also elevates warnings policy to **treat them as errors on CI**.

This is to ensure that we always have `0` build-time issues and that nothing breaking can slip into project unnoticed:

Screenshot 2021-08-09 at 12 05 49

### How?

First, I fixed existing warnings around using deprecated APIs in tests. This was done by referencing those APIs more dynamically - either with deprecation interface or through generic shim. Alternatively, we could remove those APIs from tests as they are meant deprecated, but I think we should still have them tested although deprecated (in fact, I spotted one missing improvement for Logs + Traces deprecated URLs thanks to having those tests).

Second, I've run recommended project update for Xcode 12.5 without selecting any recommended options:

Screenshot 2021-08-09 at 11 17 24

as both seem irrelevant. We can't do firsts as we're targeting iOS 11 on purpose. Second seems to [cause some issues in not-that-old Cocoapods versions](https://github.com/CocoaPods/CocoaPods/issues/9902).

Last, as it's another time when we clean up compiler warnings, I propose an improvement: to treat them as errors on CI. This way we can still keep pace in local development, but must ensure warnings are resolved before merging to the main branch.

### Review checklist

### DIFF
--- a/Sources/Datadog/DatadogConfiguration.swift
+++ b/Sources/Datadog/DatadogConfiguration.swift
@@ -570,6 +570,7 @@ extension Datadog {
             /// Until this option is enabled, automatic tracking of `UIEvents` is disabled and no swizzling is installed on the `UIApplication` class.
             ///
             /// - Parameter predicate: predicate deciding if a given action should be recorded and which allows to give custom name and to add custom attributes to the RUM Action.
+            /// Defaults to `DefaultUIKitRUMUserActionsPredicate` instance.
             public func trackUIKitRUMActions(using predicate: UIKitRUMUserActionsPredicate = DefaultUIKitRUMUserActionsPredicate()) -> Builder {
                 configuration.rumUIKitUserActionsPredicate = predicate
                 return self
--- a/Sources/Datadog/DatadogConfiguration.swift
+++ b/Sources/Datadog/DatadogConfiguration.swift
@@ -51,15 +51,15 @@ extension Datadog {
             /// US based servers.
             /// Sends data to [app.datadoghq.com](https://app.datadoghq.com/).
             @available(*, deprecated, message: "Renamed to us1")
-            case us
+            public static let us: DatadogEndpoint = .us1
             /// Europe based servers.
             /// Sends data to [app.datadoghq.eu](https://app.datadoghq.eu/).
             @available(*, deprecated, message: "Renamed to eu1")
-            case eu
+            public static let eu: DatadogEndpoint = .eu1
             /// Gov servers.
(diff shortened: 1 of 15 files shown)


Sources/Datadog/DatadogConfiguration.swift
```diff
@@ -126,13 +117,10 @@ extension Datadog {
 
             internal var url: String {
                 switch self {
-                case .us1: return "https://logs.browser-intake-datadoghq.com/v1/input/"
+                case .us1, .us: return "https://logs.browser-intake-datadoghq.com/v1/input/"
                 case .us3: return "https://logs.browser-intake-us3-datadoghq.com/v1/input/"
-                case .eu1: return "https://mobile-http-intake.logs.datadoghq.eu/v1/input/"
-                case .us1_fed: return "https://logs.browser-intake-ddog-gov.com/v1/input/"
-                case .us: return "https://mobile-http-intake.logs.datadoghq.com/v1/input/"
```
ncreated, line 133: This new value of `.us` is different after this change, but this was inspired by this https://github.com/DataDog/dd-sdk-android/pull/632#discussion_r649731331 discussion in https://github.com/DataDog/dd-sdk-android/pull/632, where the new value is meant to be recommended one. @xgouchet would be good to have your look on this.

Sources/Datadog/DatadogConfiguration.swift
```diff
@@ -166,13 +154,10 @@ extension Datadog {
 
             internal var url: String {
                 switch self {
-                case .us1: return "https://trace.browser-intake-datadoghq.com/v1/input/"
+                case .us1, .us: return "https://trace.browser-intake-datadoghq.com/v1/input/"
                 case .us3: return "https://trace.browser-intake-us3-datadoghq.com/v1/input/"
-                case .eu1: return "https:/public-trace-http-intake.logs.datadoghq.eu/v1/input/"
-                case .us1_fed: return "https://trace.browser-intake-ddog-gov.com/v1/input/"
-                case .us: return "https://public-trace-http-intake.logs.datadoghq.com/v1/input/"
```
ncreated, line 173: Similar to [this change](https://github.com/DataDog/dd-sdk-ios/pull/556/files#r685072152) - new value for `.us` is changed on purpose.

buranmert (approved):

xgouchet (approved):

PR #559, opened 2021-08-11
RUMM-1553: Add console error type to the public API

0xnm:
### What and why?

This change adds additional `.console` type to the public API. It exists internally, but is not exposed publicly. This type is needed for the errors reported from the bridges, to be consistent with Android SDK.

### How?

Just additional field in the `RUMErrorSource` enum. Also this change updates `api-surface` files, this wasn't done in a while.

### Review checklist


ncreated (approved):
All good 🚀

buranmert (approved):

PR #562, opened 2021-08-12
RUMM-1490 Migrate to Intake API v2

ncreated:
### What and why?

🚚 This PR updates the SDK's upload layer for Intake API v2, by:
* changing the `path` from `/v1/input/<token>` to `/api/v2/<track-type>` where `track-type` is `rum`, `logs` or `spans`;
* adding new HTTP headers:
  * `DD-API-KEY` - the client token,
  * `DD-EVP-ORIGIN` - the origin sdk tag (e.g. `ios` or `react-native`),
  * `DD-EVP-ORIGIN-VERSION`: the SDK version,
  * `DD-REQUEST-ID`: a debug Intake request ID;
* adding support to v2 HTTP response codes: `202`, `400`, `401`, `403`, `408`, `413`, `429`, `500` and `503`.

It also adds more verbosity to the console logs produced by `DataUploadWorker`, examples:
```
[DATADOG SDK] 🐶 → 10:13:05.933 [DEBUG]    → (rum) accepted, won't be retransmitted: [response code: 202 (accepted), request ID: 74293D75-FFEE-43A9-B328-70A71966AA69]
[DATADOG SDK] 🐶 → 10:15:17.138 [DEBUG]    → (logging) not delivered, will be retransmitted: [error: The Internet connection appears to be offline.]
[DATADOG SDK] 🐶 → 10:17:02.941 [DEBUG]    → (logging) accepted, won't be retransmitted: [response code: 403 (forbidden), request ID: 4DC9965B-357E-426D-B21C-4B61450C7634]
```

Last, it adds few improvements to the internal monitoring we do around the upload stack:
* certain `NSURLError` codes are now ignored to not send false positives (e.g. when device goes offline it's not an error);
* it handles new HTTP response codes to send errors on certain ones (e.g. `413 Payload Too Large`).

### How?

The upload architecture remains the same, but I made few changes to make testing easier with new HTTP status codes:
* The _"upload needs retry"_ consideration was moved from `DataUploadWorker` to the `DataUploadStatus` type. With that change, I was able to write simple tests for `DataUploadStatus` and only test two scenarios in `DataUploadWorker`: needs retry or does not.
* As console log output highly depends on the HTTP status code (or network error), it is now part of the `DataUploadStatus`, which enables the same easy testing as for `.needsRetry`.

### Review checklist


Sources/Datadog/Core/Upload/DataUploadStatus.swift
```diff
@@ -0,0 +1,164 @@
...
+private let ignoredNSURLErrorCodes = Set([
+    NSURLErrorNetworkConnectionLost, // -1005
+    NSURLErrorTimedOut, // -1001
+    NSURLErrorCannotParseResponse, // - 1017
+    NSURLErrorNotConnectedToInternet, // -1009
+    NSURLErrorCannotFindHost, // -1003
+    NSURLErrorSecureConnectionFailed, // -1200
+    NSURLErrorDataNotAllowed, // -1020
+    NSURLErrorCannotConnectToHost, // -1004
+])
```
ncreated, line 139: I made this list by looking at Datadog iOS app telemetry. None of these codes looks like SDK fault:
Screenshot 2021-08-12 at 14 35 40

Sources/Datadog/Core/Upload/DataUploadStatus.swift
```diff
@@ -0,0 +1,164 @@
...
+    case .tooManyRequests, .internalServerError, .serviceUnavailable:
+        // These codes mean Datadog service issue - do not produce SDK error as this is already monitored by other means.
+        return nil
+    case .badRequest, .payloadTooLarge, .requestTimeout:
+        // These codes mean that something wrong is happening either in the SDK or on the server - produce an error.
+        return (
+            message: "Data upload finished with status code: \(statusCode)",
+            error: nil,
+            attributes: ["dd_request_id": requestID ?? "(???)"]
+        )
```
ncreated, line 119: Here, I'm not sure if `408 Request Timeout` is worth sending error for Internal Monitoring. On one side this could indicate a slow response from Intake (monitored by other means), but on the other the SDK might be lagging on uploads. WDYT?

Sources/Datadog/Logging/LoggingFeature.swift
```diff
@@ -72,19 +72,23 @@ internal final class LoggingFeature {
...
                     .userAgentHeader(
                         appName: configuration.common.applicationName,
                         appVersion: configuration.common.applicationVersion,
                         device: commonDependencies.mobileDevice
-                    )
+                    ),
+                    .ddAPIKeyHeader(clientToken: configuration.clientToken),
+                    .ddEVPOriginHeader(source: configuration.common.source),
+                    .ddEVPOriginVersionHeader(),
+                    .ddRequestIDHeader(),
```
ncreated, line 86: There is [an open question](https://github.com/DataDog/dd-sdk-android/pull/671#discussion_r685768247) from @0xnm (in https://github.com/DataDog/dd-sdk-android/pull/671) if the `DD-REQUEST-ID` should be included all the time or only conditionally with a user setting.

Sources/Datadog/Core/Feature.swift
```diff
@@ -143,8 +143,8 @@ internal struct FeatureUpload {
     init(
         featureName: String,
         storage: FeatureStorage,
-        uploadHTTPHeaders: HTTPHeaders,
-        uploadURLProvider: UploadURLProvider,
+        uploadHTTPHeadersProvider: HTTPHeadersProvider,
+        uploadURL: UploadURL,
```
buranmert, line 147: i think these 2 are not useful individually and they always go together.
so we can merge them into one single `UploadRequestProvider` component?

rough sketch of `UploadRequestProvider`:
```swift
class UploadRequestProvider {
  let baseURL: URL
  let queries: [Query]
  let headers: [HTTPHeader]
  ...
  func buildRequest() -> URLRequest { ... }
}
```
that way we can pass one less parameter to `Feature` init methods and have one single component who is responsible for creating `URLRequest`.

as far as i see, currently we pass `HTTPHeadersProvider` and `uploadURL` separately down until `DataUploader` and then it creates `URLRequest` by using both. am i missing a point?

Sources/Datadog/Core/Upload/HTTPHeadersProvider.swift
```diff
@@ -0,0 +1,96 @@
...
+            return HTTPHeader(
+                field: userAgentHeaderField,
+                value: .constant("\(appName)/\(appVersion) CFNetwork (\(device.model); \(device.osName)/\(device.osVersion))")
+            )
+        }
+
+        // MARK: - Datadog Headers
+
+        /// Datadog request authentication header.
+        static func ddAPIKeyHeader(clientToken: String) -> HTTPHeader {
```
buranmert, line 52: this might be confusing for the outsiders/newcomers as we have `client token`s and [`API key`s](https://app.datadoghq.com/organization-settings/api-keys)

what do you think of renaming this function to `ddClientTokenHeader`? (of course we will keep `"DD-API-KEY"` as value)

buranmert (commented):
i left a few comments, not blocking but i wanted to discuss before approving

Sources/Datadog/Core/Feature.swift
ncreated: Yes, totally - merging both makes sense now as the `URL` logic is pretty straightforward. Done in https://github.com/DataDog/dd-sdk-ios/pull/562/commits/3b0f69032a817ab4e0369e2fe29f12db2af318a8 with introduction of `RequestBuilder`.

Sources/Datadog/Core/Upload/DataUploader.swift
```diff
@@ -6,132 +6,91 @@
...
+                uploadStatus = DataUploadStatus(networkError: error)
             }
 
             semaphore.signal()
         }
 
         _ = semaphore.wait(timeout: .distantFuture)
 
-        return uploadStatus ?? .unknown
+        return uploadStatus ?? DataUploader.unreachableUploadStatus
```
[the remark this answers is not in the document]
ncreated: The `?? DataUploader.unreachableUploadStatus` here is to only satisfy the compiler - the right side of `??` operator will never be evaluated as enforced by the previous logic in this function. That's why the name and comment:
```swift
/// An unreachable upload status - only meant to satisfy the compiler.
private static let unreachableUploadStatus = DataUploadStatus(...)
```
Alternative to this would be:
```swift
guard let uploadStatus = uploadStatus else {
   fatalError()
}
return uploadStatus
```
but as we avoid using `fatalError()` in the SDK codebase, I adopted the solution which is already used in other places: returning known default as fallback.

Hence I don't think we should improve the error value here - this would give a false sense of reachability / usability.

Sources/Datadog/Core/Upload/HTTPHeadersProvider.swift
ncreated: Here I used consistent convention for 3 new headers:
```swift
static func ddAPIKeyHeader(clientToken: String) -> HTTPHeader
static func ddEVPOriginHeader(source: String) -> HTTPHeader
static func ddEVPOriginVersionHeader() -> HTTPHeader
```
to make it clearly visible that we follow the spec for all features:
```swift
// in LoggingFeature, TracingFeature, RUMFeature and InternalMonitoringFeature:
requestBuilder: RequestBuilder(
   // ...
   headers: [
      // ...
      .ddAPIKeyHeader(clientToken: configuration.clientToken),
      .ddEVPOriginHeader(source: configuration.common.source),
      .ddEVPOriginVersionHeader(),
   ]
)
```
The spec says we must send `DD-API-KEY`, `DD-EVP-ORIGIN` and `DD-EVP-ORIGIN-VERSION` headers - and I shaped the above code to tell exactly that story.

IMO if we reverse it by naming the function `ddClientTokenHeader()`, the `DD-API-KEY` value used internally would be equally puzzling to the reader.

xgouchet (approved):
LGTM

Sources/Datadog/Core/Upload/DataUploadStatus.swift
buranmert: we can merge once this discussion is resolved

ncreated: There's distinction between client (`URLSession`) timeout and server timeout - `429` means the server.

> maybe we can internal-monitor tooManyRequests too.

Good idea, I added `.tooManyRequests` to the list for alerting 👌.

buranmert (approved):

PR #566, opened 2021-08-18
RUMM-1530 Add Datadog.isInitialized variable

maxep:
### What and why?

The public interface lacks a way of checking the SDK initialized state.

### How?

A static variable `isInitialized` of the `Datadog` configuration class checks the internal `instance` availability.

### Review checklist

### DIFF
--- a/Sources/Datadog/Datadog.swift
+++ b/Sources/Datadog/Datadog.swift
@@ -104,7 +104,7 @@ public class Datadog {
 
     /// Returns `true` if the Datadog SDK is already initialized, `false` otherwise.
     public static var isInitialized: Bool {
-        instance != nil
+        return instance != nil
     }
 
     /// Sets current user information.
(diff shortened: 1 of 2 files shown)


Sources/Datadog/Datadog.swift
```diff
@@ -102,6 +102,11 @@ public class Datadog {
         }
     }
 
+    /// Returns `true` if the Datadog SDK is already initialized, `false` otherwise.
+    public static var isInitialized: Bool {
+        instance != nil
```
buranmert, line 107: iirc some previous Xcode/Swift versions don't compile when `return` is omitted.
better safe than sorry: can you please add `return` to this line?

buranmert (changes requested):
this might break old Xcodes, but otherwise looks good 👍

buranmert (approved):

ncreated (approved):
Looks good 🚀

xgouchet (commented):
LGTM

PR #568, opened 2021-08-20
Use preferred isEmpty checks

michalsrutek:
### What and why?

Use preferred way of checking emptiness with `Collection`'s `isEmpty` property, as it has a complexity of O(1).

### How?

By replacing `.count == 0` checks with `.isEmpty`.


ncreated (approved):
Thank you @michalsrutek , it looks good 🙂 👍

PR #578, opened 2021-08-31
RUMM-1566 Enable using `DatadogSDKCrashReporting` pod with and without `use_frameworks!`

ncreated:
### What and why?

📦 This PR removes the need for adding following `pre_install` workaround to the `Podfile` (introduced in https://github.com/DataDog/dd-sdk-ios/pull/553):
```ruby
pod 'DatadogSDK'
pod 'DatadogSDKCrashReporting'

# ...

static_frameworks = ['DatadogSDKCrashReporting','PLCrashReporter']
pre_install do |installer|
  installer.pod_targets.each do |pod|
    if static_frameworks.include?(pod.name)
      def pod.static_framework?;
        true
      end
      def pod.build_type;
        Pod::BuildType.static_library
      end
    end
  end
end
```

This workaround was necessary as `DatadogSDKCrashReporting` has a transitive dependency on the static binary in `PLCrashReporter` and without this patch the `pod install` was failing with _"target has transitive dependencies that include static binaries"_.

### How?

The fix is this one-line change in `DatadogSDKCrashReporting.podspec`:
```diff
+  s.static_framework = true
```
By using this option, the pod is configured _on our side_ (in `.podspec`) to be static framework, so we don't need to ask the user for patch it _on their end_ (in `Podfile`). [Podspec Syntax Reference](https://guides.cocoapods.org/syntax/podspec.html#static_framework):
> `static_framework`
> Indicates, that if use_frameworks! is specified, the pod should include a static library framework.

#### CP compatibility tests enhancement

All other changes in this PR are related to enhancing our Cocoapods compatibility tests. Before, we were testing `CPProject.xcworkspace` only with using `use_frameworks!` + the above patch, but the setup with no `use_frameworks!` was never covered.

I made an update and introduced two distinct set of targets:
Screenshot 2021-09-02 at 13 49 47

One is configured to link our SDK **with** `use_frameworks!`:
Screenshot 2021-09-02 at 13 49 58

and the other links it **without** `use_frameworks!`:
Screenshot 2021-09-02 at 13 50 08

CI works as usual - it performs `pod update` and runs tests against the SDK fetched from the current branch.

### Review checklist


bitrise.yml
```diff
@@ -253,22 +252,24 @@ workflows:
             #!/usr/bin/env bash
             set -e
             make test-cocoapods ci=${CI}
-    - xcodebuild:
-        title: Build CPProject for tests - Catalyst
```
ncreated, line 257: Here I remove the Mac Catalyst build for Cocoapods from `bitrise.yml`. Mac Catalyst compatibility is already tested in the previous step, when building with Swift Package Manager. IMO testing it once again for Cocoapods is redundant as if it worked for SPM, it must for for CP too (nothing had to be configured in `Package.swift` nor `podspec` when enabling it in https://github.com/DataDog/dd-sdk-ios/pull/277).

Removing this step gives us additional CI time, now used for testing _no `use_frameworks!` compatibility for CP.

maxep (commented):
Would it be possible to just duplicate targets in xcodeproj and then use a single `Podfile`?  with:
``` ruby
target 'CPProject-dynamic' do
  use_frameworks!
  ...
end

target 'CPProject-static' do
  # use_modular_headers!
  ...

end
```

Cocoapods also support static linkng with `use_frameworks! :linkage => :static` since 1.9.0, might be something we would like to test as well!

ncreated:
> Would it be possible to just duplicate targets in xcodeproj and then use a single `Podfile`?

Great idea 👍, I did it and made an update to this PR description.

> Cocoapods also support static linkng with `use_frameworks! :linkage => :static` since 1.9.0, might be something we would like to test as well!

I've played with this for a while, but seems that CP has issues when using `:static` in one target and `:dynamic` in another target. If both are used in separate projects, everything works. If mixing them in one project, it fails on finding umbrella headers. I haven't dig into details of this issue, as the difference between `:static` and `:dynamic` linking is minimal and only relates to adding `[CP] Embed Pods` build phase which runs `sh` script. With the current setup, that build phase is added, so I think this is good enough. WDYT @maxep ?

maxep (approved):
👌 Yes, the linkage shouldn't make much of a difference in the scope of this integration test.
It is so cool to get ride of the podfile workaround, nice work 👏

PR #580, opened 2021-08-31
RUMM-1420 Set internal(set) accessor to readonly additionalProperties

maxep:
### What and why?

As part of applying new RUM event schema, we need to add internal mutation capabilities to rum event models in order to sanitise custom attributes. The `rum-models-generator` tool was creating immutable models which prevented from sanitising the `usr` and `context` additional properties. 

### How?

The `rum-models-generator` now support 3 levels of property mutability:
```swift
enum Mutability: Int {
    case immutable
    case mutableInternally
    case mutable
 }
```
The `mutableInternally`  will be applied to read-only models with `additionalProperties` set to `true`. The generated models will then have an accessor level of `public internal(set) var` in a transitive way.

### Review checklist

### DIFF
--- a/tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/Swift/JSONToSwiftTypeTransformer.swift
+++ b/tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/Swift/JSONToSwiftTypeTransformer.swift
@@ -78,6 +78,7 @@ internal class JSONToSwiftTypeTransformer {
                 // In generated encoding code, this dictionary is erased but its keys and values are used as dynamic
                 // properties encoded in JSON.
                 let additionalPropertyName = jsonObject.name + "Info"
+                let mutability: SwiftStruct.Property.Mutability = additionalProperties.isReadOnly ? .mutableInternally : .mutable
                 var `struct` = try transformJSONToStruct(jsonObject)
                 `struct`.properties.append(
                     SwiftStruct.Property(
@@ -87,7 +88,7 @@ internal class JSONToSwiftTypeTransformer {
                             value: SwiftPrimitive<SwiftCodable>()
                         ),
                         isOptional: false,
-                        isMutable: !additionalProperties.isReadOnly,
+                        mutability: mutability,
                         defaultValue: nil,
                         codingKey: .dynamic
                     )
@@ -125,12 +126,13 @@ internal class JSONToSwiftTypeTransformer {
             }
 
             return try objectProperties.map { jsonProperty in
+                let mutability: SwiftStruct.Property.Mutability = jsonProperty.isReadOnly ? .immutable : .mutable
                 return SwiftStruct.Property(
                     name: jsonProperty.name,
                     comment: jsonProperty.comment,
                     type: try transformJSONToAnyType(jsonProperty.type),
                     isOptional: !jsonProperty.isRequired,
-                    isMutable: !jsonProperty.isReadOnly,
+                    mutability: mutability,
                     defaultValue: try readDefaultValue(for: jsonProperty),
                     codingKey: .static(value: jsonProperty.name)
                 )
@@ -175,7 +177,7 @@ internal class JSONToSwiftTypeTransformer {
 
         `struct`.properties = `struct`.properties.map { property in
             var property = property
-            property.isMutable = property.isMutable || hasTransitiveMutableProperty(type: property.type)
+            property.mutability = transitiveMutability(property: property)
 
             if let nestedStruct = property.type as? SwiftStruct {
                 property.type = resolveTransitiveMutableProperties(in: nestedStruct)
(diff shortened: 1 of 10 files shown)


tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/Swift/SwiftType.swift
```diff
@@ -54,6 +54,12 @@ internal struct SwiftEnum: SwiftType {
 
 internal struct SwiftStruct: SwiftType {
     struct Property: SwiftType {
+        enum Mutability: Int {
+           case immutable
+           case mutableInternally
+           case mutable
+        }
```
ncreated, line 61: As the order here matters a lot (changing it would be destructive), how about adding simple unit test for ensuring it or at least adding a comment to watch it?

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Output/Printing/ObjcInteropPrinter.swift
```diff
@@ -241,7 +241,7 @@ internal class ObjcInteropPrinter: BasePrinter {
         let objcPropertyName = swiftProperty.name
         let objcEnumName = objcTypeNamesPrefix + nestedObjcEnum.objcTypeName
 
-        if swiftProperty.isMutable {
+        if swiftProperty.mutability == .mutable {
```
ncreated, line 244: How about using `switch` instead of `if` here and in other places in this class? Doing that is a bit more repetitive, but has the benefit of compiler explicitly guiding us when adding new cases to `Mutability` enum in the future:
```swift
switch swiftProperty.mutability {
   case .mutable:
      // the `if {}` code
   case .immutable, .mutableInternally:
      // the `else {}` code
}
```

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/Swift/JSONToSwiftTypeTransformer.swift
```diff
@@ -187,17 +189,27 @@ internal class JSONToSwiftTypeTransformer {
...
         switch type {
         case let array as SwiftArray:
-            return hasTransitiveMutableProperty(type: array.element)
+            return transitiveMutableProperty(of: array.element)
         case let `struct` as SwiftStruct:
-            return `struct`.properties.contains { property in
-                property.isMutable || hasTransitiveMutableProperty(type: property.type)
+            return `struct`.properties.reduce(.immutable) { max, property in
+                let innerMutability = transitiveMutability(of: property)
+                return innerMutability.rawValue > max.rawValue ? innerMutability : max
```
ncreated, line 209: I'm afraid this `.rawValue` comparison won't be so clear when visiting this code in the future - the problem we solve in `rum-models-generator` is quite abstract and examples and context are priceless 😉.

How about extracting it to a separate method with dedicated naming and more context given in its comment:
```swift
/// Returns new `Mutability` for parent type given its child `Mutability`.
///
/// This stands for: _if the child is mutable, its parent must be mutable too_;
/// e.g.: in `foo.bar.property = 2` expression, not only `property` must be mutable, but also its parent `bar` accessor.
private func resolve(parentMutability: Mutability, childMutability: Mutability) -> Mutability {
    return parentMutability.rawValue > childMutability.rawValue ? parentMutability : childMutability
}
```
This could be then reused from `transitiveMutableProperty(of type:)` and `transitiveMutability(of property:)`.

ncreated (commented):
All looks good 👍. I left few comments regarding code documentation and codebase resilience for future changes - let me know what you think 🙌.

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/Swift/JSONToSwiftTypeTransformer.swift
maxep: That's a good idea! What do you thing of adding:
```swift
internal func max<T>(_ lhs: T, _ rhs: T) -> T where T: RawRepresentable, T.RawValue: Comparable {
    rhs.rawValue > lhs.rawValue ? rhs : lhs
}
``` 
to the `Utilities.swift`?

This way I could reuse that function for testing the `Mutability` level order.

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/Swift/SwiftType.swift
maxep: Totally!

ncreated (approved):
Looks great 💪

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/Swift/JSONToSwiftTypeTransformer.swift
ncreated: I think this wouldn't address the main concern on documenting what we do: generic `func max<T>()` is same vague as `>` operator, whereas plain `func` gives us the clear point for documenting the context in its comment.

PR #586, opened 2021-09-03
RUMM-1420 Remove attributes and userInfoAttributes property from RUMEvent

maxep:
### What and why?

The RUM events schema now support `additionalProperties` in `usr` and `context` fields. Therefor, `RUMEvent`'s properties `attributes` and `userInfoAttributes` are no longer required.

#### RUM Event Sanitization

The `RUMSanitizableEvent` protocol is a type constraint on events that require sanitisation by enforcing mutation on `usr` and `context` properties.

#### RUM Event

Error attributes were previously included in the  `RUMEvent`'s `attributes` and then separated during encoding. A new `errorAttributes` property has been added until error are added to the schema.

The `userInfoProvider` property has been removed from the `RUMEventBuilder` as the `usr` is now part of the model and will be populated by the `RUMViewScope`.

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/Utils/ValuePublisher.swift
+++ b/Sources/Datadog/Core/Utils/ValuePublisher.swift
@@ -19,7 +19,7 @@ internal protocol ValueObserver: AnyObject {
 }
 
 /// Manages the `Value` in a thread safe manner and notifies subscribed `ValueObservers` on its change.
-internal class ValuePublisher<Value> {
+internal final class ValuePublisher<Value> {
     /// Type erasure for `ValueObserver` type.
     private struct AnyObserver<ObservedValue> {
         let notifyValueChanged: (ObservedValue, ObservedValue) -> Void
--- a/Sources/Datadog/Core/Utils/ValuePublisher.swift
+++ b/Sources/Datadog/Core/Utils/ValuePublisher.swift
@@ -48,7 +48,7 @@ internal class ValuePublisher<Value> {
         }
     }
 
-    init(initialValue: Value) {
+    required init(initialValue: Value) {
         self.unsafeValue = initialValue
         self.unsafeObservers = []
     }
--- a/Sources/Datadog/CrashReporting/CrashContext/CrashContext.swift
+++ b/Sources/Datadog/CrashReporting/CrashContext/CrashContext.swift
@@ -150,47 +150,35 @@ private struct CodableUserInfo: Codable {
 
 private struct CodableRUMViewEvent: Codable {
     private let model: RUMViewEvent
-    private let attributes: [String: Encodable]
-    private let userInfoAttributes: [String: Encodable]
+    private let errorAttributes: [AttributeKey: AttributeValue]?
 
     init(from managedValue: RUMEvent<RUMViewEvent>) {
         self.model = managedValue.model
-        self.attributes = managedValue.attributes
-        self.userInfoAttributes = managedValue.userInfoAttributes
+        self.errorAttributes = managedValue.errorAttributes
     }
 
     var managedValue: RUMEvent<RUMViewEvent> {
-        return .init(
-            model: model,
-            attributes: attributes,
-            userInfoAttributes: userInfoAttributes
-        )
+        return .init(model: model, errorAttributes: errorAttributes)
     }
 
     // MARK: - Codable
 
     enum CodingKeys: String, CodingKey {
         case model = "mdl"
-        case attributes = "att"
-        case userInfoAttributes = "uia"
+        case errorAttributes = "ea"
     }
 
     init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: CodingKeys.self)
-
         self.model = try container.decode(RUMViewEvent.self, forKey: .model)
-        self.attributes = try container.decode([String: CodableValue].self, forKey: .attributes)
-        self.userInfoAttributes = try container.decode([String: CodableValue].self, forKey: .userInfoAttributes)
+        self.errorAttributes = try container.decodeIfPresent([String: CodableValue].self, forKey: .errorAttributes)
     }
 
     func encode(to encoder: Encoder) throws {
         var container = encoder.container(keyedBy: CodingKeys.self)
-        let encodedAttributes = attributes.mapValues { CodableValue($0) }
-        let encodedUserInfoAttributes = userInfoAttributes.mapValues { CodableValue($0) }
-
         try container.encode(model, forKey: .model)
-        try container.encode(encodedAttributes, forKey: .attributes)
-        try container.encode(encodedUserInfoAttributes, forKey: .userInfoAttributes)
+        let error = errorAttributes?.mapValues { CodableValue($0) }
+        try container.encode(error, forKey: .errorAttributes)
     }
 }
 
--- a/Sources/Datadog/CrashReporting/CrashContext/CrashContext.swift
+++ b/Sources/Datadog/CrashReporting/CrashContext/CrashContext.swift
@@ -150,35 +150,29 @@ private struct CodableUserInfo: Codable {
 
 private struct CodableRUMViewEvent: Codable {
     private let model: RUMViewEvent
-    private let errorAttributes: [AttributeKey: AttributeValue]?
 
     init(from managedValue: RUMEvent<RUMViewEvent>) {
         self.model = managedValue.model
-        self.errorAttributes = managedValue.errorAttributes
     }
 
     var managedValue: RUMEvent<RUMViewEvent> {
-        return .init(model: model, errorAttributes: errorAttributes)
+        return .init(model: model)
     }
 
     // MARK: - Codable
 
     enum CodingKeys: String, CodingKey {
         case model = "mdl"
-        case errorAttributes = "ea"
     }
 
     init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: CodingKeys.self)
         self.model = try container.decode(RUMViewEvent.self, forKey: .model)
-        self.errorAttributes = try container.decodeIfPresent([String: CodableValue].self, forKey: .errorAttributes)
     }
 
     func encode(to encoder: Encoder) throws {
         var container = encoder.container(keyedBy: CodingKeys.self)
         try container.encode(model, forKey: .model)
-        let error = errorAttributes?.mapValues { CodableValue($0) }
-        try container.encode(error, forKey: .errorAttributes)
     }
 }
 
--- a/Sources/Datadog/RUM/RUMEvent/RUMEventBuilder.swift
+++ b/Sources/Datadog/RUM/RUMEvent/RUMEventBuilder.swift
@@ -7,14 +7,9 @@
 import Foundation
 
 internal class RUMEventBuilder {
-    let userInfoProvider: UserInfoProvider
     let eventsMapper: RUMEventsMapper
 
-    init(
-        userInfoProvider: UserInfoProvider,
-        eventsMapper: RUMEventsMapper
-    ) {
-        self.userInfoProvider = userInfoProvider
+    init(eventsMapper: RUMEventsMapper) {
         self.eventsMapper = eventsMapper
     }
 
@@ -26,10 +21,6 @@ internal class RUMEventBuilder {
 
         if !attributes.isEmpty {
             model.context = RUMEventAttributes(contextInfo: attributes)
-        }
-
-        if !userInfoProvider.value.extraInfo.isEmpty {
-            model.usr = RUMUser(email: model.usr?.email, id: model.usr?.id, name: model.usr?.name, usrInfo: userInfoProvider.value.extraInfo)
         }
 
         let event = RUMEvent(model: model)
--- a/Sources/Datadog/RUM/RUMEvent/RUMEventBuilder.swift
+++ b/Sources/Datadog/RUM/RUMEvent/RUMEventBuilder.swift
@@ -13,18 +13,8 @@ internal class RUMEventBuilder {
         self.eventsMapper = eventsMapper
     }
 
-    func createRUMEvent<DM: RUMDataModel>(
-        with model: DM,
-        attributes: [String: Encodable]
-    ) -> RUMEvent<DM>? {
-        var model = model
-
-        if !attributes.isEmpty {
-            model.context = RUMEventAttributes(contextInfo: attributes)
-        }
-
+    func createRUMEvent<DM: RUMDataModel>(with model: DM) -> RUMEvent<DM>? {
         let event = RUMEvent(model: model)
-        let mappedEvent = eventsMapper.map(event: event)
-        return mappedEvent
+        return eventsMapper.map(event: event)
     }
 }
--- a/Sources/Datadog/RUM/RUMEvent/RUMEventBuilder.swift
+++ b/Sources/Datadog/RUM/RUMEvent/RUMEventBuilder.swift
@@ -22,11 +22,17 @@ internal class RUMEventBuilder {
         with model: DM,
         attributes: [String: Encodable]
     ) -> RUMEvent<DM>? {
-        let event = RUMEvent(
-            model: model,
-            attributes: attributes,
-            userInfoAttributes: userInfoProvider.value.extraInfo
-        )
+        var model = model
+
+        if !attributes.isEmpty {
+            model.context = RUMEventAttributes(contextInfo: attributes)
+        }
+
+        if !userInfoProvider.value.extraInfo.isEmpty {
+            model.usr = RUMUser(email: model.usr?.email, id: model.usr?.id, name: model.usr?.name, usrInfo: userInfoProvider.value.extraInfo)
+        }
+
+        let event = RUMEvent(model: model)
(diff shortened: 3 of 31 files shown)


Sources/Datadog/RUM/DataModels/RUMDataModels.swift
```diff
@@ -1241,7 +1241,7 @@ public struct RUMConnectivity: Codable {
 
 /// User provided context
 public struct RUMEventAttributes: Codable {
-    public internal(set) var contextInfo: [String: Codable]
+    public internal(set) var contextInfo: [String: Encodable]
```
maxep, line 1244: The `rum-models-generator` needs to be modified to use `Encodable` here.

Sources/Datadog/RUM/RUMEvent/RUMEventBuilder.swift
```diff
@@ -22,11 +22,17 @@ internal class RUMEventBuilder {
...
+
+        if !attributes.isEmpty {
+            model.context = RUMEventAttributes(contextInfo: attributes)
+        }
+
+        if !userInfoProvider.value.extraInfo.isEmpty {
+            model.usr = RUMUser(email: model.usr?.email, id: model.usr?.id, name: model.usr?.name, usrInfo: userInfoProvider.value.extraInfo)
+        }
+
+        let event = RUMEvent(model: model)
```
maxep, line 35: I'm really not sure this is the right place to do such formatting!

Tests/DatadogTests/Datadog/RUM/RUMEventOutputs/RUMEventFileOutputTests.swift
```diff
@@ -62,7 +64,9 @@ class RUMEventFileOutputTests: XCTestCase {
             jsonString: """
             {
                 "attribute": "foo",
-                "context.custom.attribute": "value"
+                "context": {
+                    "custom.attribute": "value"
+                }
```
maxep, line 69: I guess this is fine?

Sources/Datadog/RUM/RUMEvent/RUMEventBuilder.swift
```diff
@@ -22,11 +22,17 @@ internal class RUMEventBuilder {
...
-        )
+        var model = model
+
+        if !attributes.isEmpty {
+            model.context = RUMEventAttributes(contextInfo: attributes)
+        }
+
+        if !userInfoProvider.value.extraInfo.isEmpty {
+            model.usr = RUMUser(email: model.usr?.email, id: model.usr?.id, name: model.usr?.name, usrInfo: userInfoProvider.value.extraInfo)
+        }
```
ncreated, line 33: With the change we made to `usr`:
```swift
public internal(set) var usr: RUMUser?
```
we can simplify this to:
```swift
model.usr?.usrInf = userInfoProvider.value.extraInfo
```
no?

Sources/Datadog/RUM/RUMEvent/RUMEventBuilder.swift
ncreated: Why not? Looks good IMO 👌 - we implement the previous behaviour in a new way, good enough for refactoring 🙂.

Sources/Datadog/RUM/RUMEvent/RUMEventEncoder.swift
```diff
@@ -8,22 +8,41 @@ import Foundation
 
 /// `Encodable` representation of RUM event.
 /// Mutable properties are subject of sanitization or data scrubbing.
-internal struct RUMEvent<DM: RUMDataModel>: Encodable {
+/// /// TODO: RUMM-1463 Remove `errorAttributes` property once new error format is managed through `RUMDataModels`
```
ncreated, line 11: This is already TODO-ed in `func encode(to:)`
```suggestion
```

Sources/Datadog/RUM/RUMEvent/RUMEventEncoder.swift
```diff
@@ -8,22 +8,41 @@ import Foundation
...
+/// TODO: RUMM-1584 - Remove `RUMEvent` container.
+internal struct RUMEvent<DM>: Encodable where DM: RUMDataModel, DM: RUMSanitizableEvent {
     /// The actual RUM event model created by `RUMMonitor`
     var model: DM
 
-    /// Custom attributes set by the user
-    var attributes: [String: Encodable]
-    var userInfoAttributes: [String: Encodable]
+    /// Error attributes.
+    let errorAttributes: [String: Encodable]?
```
ncreated, line 18: A handy note to future us could be this:
```suggestion
    /// Error attributes. Only set when `DM == RUMErrorEvent` and error describes a crash.
    /// Can be entirely removed when RUMM-1463 is resolved and error values are part of the `RUMErrorEvent`.
    let errorAttributes: [String: Encodable]?
```

Tests/DatadogTests/Datadog/Mocks/RUMFeatureMocks.swift
```diff
@@ -88,43 +88,38 @@ extension RUMResourceType {
...
-        userInfoAttributes: [String: Encodable] = [:]
-    ) -> RUMEvent<DM> {
-        return RUMEvent<DM>(
-            model: model,
-            attributes: attributes,
-            userInfoAttributes: userInfoAttributes
-        )
-    }
-
     static func mockRandomWith<DM: RUMDataModel>(model: DM) -> RUMEvent<DM> {
```
ncreated, line 106: I think now this function is not needed at all. All `RUMDataModels` already implement the `mockRandom()` function in `RUMDataModelMocks.swift` and it produces random `usr` and `context`. WDYT?

Tests/DatadogTests/Datadog/RUM/RUMEvent/RUMEventBuilderTests.swift
```diff
@@ -18,8 +18,8 @@ class RUMEventBuilderTests: XCTestCase {
         )
 
         XCTAssertEqual(event.model.attribute, "foo")
-        XCTAssertEqual((event.attributes as? [String: String])?["foo"], "bar")
-        XCTAssertEqual((event.attributes as? [String: String])?["fizz"], "buzz")
+        XCTAssertEqual((event.model.context?.contextInfo as? [String: String])?["foo"], "bar")
+        XCTAssertEqual((event.model.context?.contextInfo as? [String: String])?["fizz"], "buzz")
     }
```
ncreated, line 23: I'm surprised we don't test if `RUMEventBuilder` sets the user attributes here. We could do it easily by mocking the value in the provider:
```diff
- userInfoProvider: .mockAny(),
+ userInfoProvider: .mockWith(
+   userInfo: .init(id: nil, name: nil, email: nil, extraInfo: ["bazz": "buzz"])
+ ),
```
WDYT?

Tests/DatadogTests/Datadog/RUM/RUMEvent/RUMEventSanitizerTests.swift
```diff
@@ -99,31 +98,29 @@ class RUMEventSanitizerTests: XCTestCase {
...
             remaining -= expectedSanitizedUserInfo
-            let expectedSanitizedAttrs = min(sanitized.attributes.count, remaining)
+            let expectedSanitizedAttrs = min(sanitized.context!.contextInfo.count, remaining)
             remaining -= expectedSanitizedAttrs
 
             XCTAssertGreaterThanOrEqual(remaining, 0)
-            XCTAssertEqual(sanitized.userInfoAttributes.count, expectedSanitizedUserInfo, "If number of attributes needs to be limited, `userInfoAttributes` are removed second")
-            XCTAssertEqual(sanitized.attributes.count, expectedSanitizedAttrs, "If number of attributes needs to be limited, `attributes` are removed first.")
+            XCTAssertEqual(sanitized.usr?.usrInfo.count, expectedSanitizedUserInfo, "If number of attributes needs to be limited, `userInfoAttributes` are removed second")
+            XCTAssertEqual(sanitized.context?.contextInfo.count, expectedSanitizedAttrs, "If number of attributes needs to be limited, `attributes` are removed first.")
```
ncreated, line 117:
```suggestion
            XCTAssertEqual(sanitized.usr?.usrInfo.count, expectedSanitizedUserInfo, "If number of attributes needs to be limited, `usrInfo` are removed second")
            XCTAssertEqual(sanitized.context?.contextInfo.count, expectedSanitizedAttrs, "If number of attributes needs to be limited, `contextInfo` are removed first.")
```

Tests/DatadogTests/Datadog/RUM/RUMEventOutputs/RUMEventFileOutputTests.swift
ncreated: Totally 👍. For our backend it's the same.

ncreated (commented):
Excellent contribution 👌. I left few comments - it's still draft, but definitely the right direction.

Sources/Datadog/RUM/RUMEvent/RUMEventBuilder.swift
maxep: If `model.usr` is nil, then we would loose the extraInfo. Now that makes me wondering if we should get all user Info from the `userInfoProvider`, or event use the `RUMUserInfoProvider` instead?

Tests/DatadogTests/Datadog/Mocks/RUMFeatureMocks.swift
maxep: I agree, good catch!

Sources/Datadog/RUM/RUMEvent/RUMEventBuilder.swift
ncreated: > If model.usr is nil, then we would loose the extraInfo.

Ah, right 👍.

> Now that makes me wondering if we should get all user Info from the `UserInfoProvider`, or event use the `RUMUserInfoProvider` instead?

Indeed, but we can do even better. We're using `UserInfoProvider` here, as this code is older than `RUMUser` added to `RUMDataModels` (through schema). Now, as we use `RUMUserInfoProvider` directly in RUM scopes (e.g. [here in `RUMViewScope`](https://github.com/DataDog/dd-sdk-ios/blob/16a55a14b422fe4cf5dff72fe14a61150d54cf1b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift#L309)), the user model is already emitted to all events and there is no need override it here.

However, to remove `userInfoProvider` from here, we'd need to pay off one more piece of technical debt: [the `RUMUserInfoProvider` seems to ignore `usrInfo` attributes](https://github.com/DataDog/dd-sdk-ios/blob/16a55a14b422fe4cf5dff72fe14a61150d54cf1b/Sources/Datadog/RUM/RUMEvent/RUMUserInfoProvider.swift#L17-L27). This seems aligned with the scope of this RUMM-1420 - @maxep do you think this is something we improve now?

maxep: Ah yes, totally! I will change that!

Tests/DatadogTests/Datadog/RUM/RUMEvent/RUMEventBuilderTests.swift
maxep: By applying [this change](https://github.com/DataDog/dd-sdk-ios/pull/586/files#r702641511), we won't be setting the `userInfo` in here.
This an be covered by [this test](https://github.com/DataDog/dd-sdk-ios/blob/02963df9beca1734bbabf3bd6730281c00d5664b/Tests/DatadogTests/Datadog/RUM/RUMEvent/RUMUserInfoProviderTests.swift#L34) tho!

Sources/Datadog/Core/Utils/ValuePublisher.swift
```diff
@@ -48,7 +48,7 @@ internal class ValuePublisher<Value> {
         }
     }
 
-    init(initialValue: Value) {
+    required init(initialValue: Value) {
```
maxep, line 51: This is necessary to use this init in protocol extension with `Self` requirement, such as `RandomMock`. This could also be fixed by turning `ValuePublisher` into a `final class`.

ncreated: It could be `final` indeed, but it's OK for now I guess 👍

Tests/DatadogTests/Datadog/RUM/RUMEvent/RUMUserInfoProviderTests.swift
```diff
@@ -29,8 +29,9 @@ class RUMUserInfoProviderTests: XCTestCase {
         userInfoProvider.value = UserInfo(id: "abc-123", name: "Foo", email: "foo@bar.com", extraInfo: [:])
         XCTAssertEqual(rumUserInfoProvider.current, RUMUser(email: "foo@bar.com", id: "abc-123", name: "Foo", usrInfo: [:]))
 
-        userInfoProvider.value = UserInfo(id: "abc-123", name: "Foo", email: "foo@bar.com", extraInfo: mockRandomAttributes())
+        let extraInfo: [String: Encodable] = mockRandomAttributes()
+        userInfoProvider.value = UserInfo(id: "abc-123", name: "Foo", email: "foo@bar.com", extraInfo: extraInfo)
         // TODO: RUMM-1420 Encode user `extraInfo` info as RUMUser `usrInfo`
```
ncreated, line 34: 😉 
```suggestion
```

Sources/Datadog/CrashReporting/CrashContext/CrashContext.swift
```diff
@@ -150,47 +150,35 @@ private struct CodableUserInfo: Codable {
 
 private struct CodableRUMViewEvent: Codable {
     private let model: RUMViewEvent
-    private let attributes: [String: Encodable]
-    private let userInfoAttributes: [String: Encodable]
+    private let errorAttributes: [AttributeKey: AttributeValue]?
```
ncreated, line 153: I guess it's only for sanity now, as this will be removed in `RUMM-1463` as we're saying in this new comment:
```swift
// RUMEvent.swift

/// Error attributes. Only set when `DM == RUMErrorEvent` and error describes a crash.
/// Can be entirely removed when RUMM-1463 is resolved and error values are part of the `RUMErrorEvent`.
let errorAttributes: [String: Encodable]?
```
Right? Meaning, `RUMM-1463` will make this crash-context-coding thing even cleaner ✨.

Sources/Datadog/RUM/RUMEvent/RUMEventBuilder.swift
```diff
@@ -7,26 +7,23 @@
...
     ) -> RUMEvent<DM>? {
-        let event = RUMEvent(
-            model: model,
-            attributes: attributes,
-            userInfoAttributes: userInfoProvider.value.extraInfo
-        )
+        var model = model
+
+        if !attributes.isEmpty {
+            model.context = RUMEventAttributes(contextInfo: attributes)
```
ncreated, line 23: There's one last thing to cleanup. Similar to how we emit `RUMUser` in RUM scopes, we can now emit `RUMEventAttributes` right in scopes:
```swift
// RUMViewScope.swift
// RUMResourceScope.swift
// RUMUserActionScope.swift

let eventData = RUM####Event(
   // ...
   usr: dependencies.userInfoProvider.current
   // ...
   context: .init(contextInfo: attributes),
   // ...
)

let event = dependencies.eventBuilder.createRUMEvent(with: eventData)
```
That will polish the architecture by (finally 🚀) building RUM events only in RUM scopes ✨.

This change should be relatively easy and safe as all RUM scope tests use the actual models and we've already updated assertions in this PR.

ncreated (changes requested):

Sources/Datadog/RUM/RUMEvent/RUMEventBuilder.swift
maxep: Awesome, will do! 🧹

Sources/Datadog/CrashReporting/CrashContext/CrashContext.swift
maxep: Ah yes, I forgot about that one 👍

maxep: So, I cannot remove `CodableRUMViewEvent` as I originally thought because `RUMViewEvent` is not Decodable. And if  we make it Decodable, it means that we need to encode the error attributes...

Sources/Datadog/Core/Utils/ValuePublisher.swift
maxep: I kind of prefer `final` as it is stricter.

Sources/Datadog/CrashReporting/CrashContext/CrashContext.swift
ncreated: Right 👍. We will do it in `RUMM-1584` (_Remove `RUMEvent` container._) 👌.

ncreated (approved):
💯

PR #587, opened 2021-09-03
RUMM-1420  `additionalProperties` as [String: Encodable]

maxep:
### What and why?

RUM models additional properties needs to be set as `[String: Encodable]` to match the public API definition.

### How?

The codegen `rum-models-generator` now adds addition properties with type `[String: Encodable]`.

A new protocol `SwiftPrimitiveNoObjcInteropType` has been added to prevent the objc interface from being generated, both `SwiftCodable` and `SwiftEncodable` now complies to the protocol.

### Review checklist


ncreated (approved):
🎯

PR #589, opened 2021-09-06
RUMM-1597 Add tests for matching our crash report format with PLCR text format

ncreated:
### What and why?

🛡️ All is good, but this PR adds two more unit tests to `DatadogCrashReporting` to ensure that the arithmetic behind our unsymbolicated stack trace computation produces same results as in [`PLCrashReportTextFormatter`](https://github.com/microsoft/plcrashreporter/blob/b6a0696d2add340999c6a03fcd10cbffdfc2a73d/Source/PLCrashReportTextFormatter.m#L47-L50).

This is to further ensure reliability of our crash reports and prevent future regressions.

### How?

Two unit tests are added. Both compare values from `DDCrashReport` to values appearing in raw crash report exported from `PLCR` (to its free-form text format). I leveraged `PLCR's` `.generateLiveReportAndReturnError()` API which produces a snapshot of running process (like it would crash, but without actual crashing).

### Review checklist


maxep (approved):
👌

PR #593, opened 2021-09-14
RUMM-1614 Upgrade PLCR to `1.10.0`

ncreated:
### What and why?

📦 This PR upgrades PLCR dependency version to `1.10.0`.

The `1.10.0` version was very much awaited, as it is meant to solve our problem with `DDCR.podspec` distribution.

### How?

Upgraded PLCR version in `Package.swift` (for SPM), `Cartfile` (for Carthage) and `DDCR.podspec` (for Cocoapods).

I also tested that `DDCR.podspec` passes `pod spec lint`:
```
$ pod spec lint --allow-warnings --verbose DatadogSDKCrashReporting.podspec

DatadogSDKCrashReporting.podspec passed validation.
```

Later I will check if `DDCR` compiles on M1 laptop.

### Review checklist

- [x] Check if `dd-sdk-ios` compiles on M1 laptop

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,2 +1,2 @@
 github "lyft/Kronos" ~> 4.2
-github "microsoft/plcrashreporter" == 1.8.1
+github "microsoft/plcrashreporter" == 1.10.0
--- a/Cartfile
+++ b/Cartfile
@@ -1,2 +1,2 @@
 github "lyft/Kronos" ~> 4.2
-github "microsoft/plcrashreporter" == 1.10.0
+github "microsoft/plcrashreporter" ~> 1.10.0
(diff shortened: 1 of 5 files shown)


Cartfile
```diff
@@ -1,2 +1,2 @@
 github "lyft/Kronos" ~> 4.2
-github "microsoft/plcrashreporter" == 1.8.1
+github "microsoft/plcrashreporter" == 1.10.0
```
buranmert, line 2: this can/should be `~> 1.10.0`.
it was `== 1.8.1` just to avoid the problematic `1.9` version.

buranmert (approved):

Cartfile
ncreated: Indeed - fixed. Now it's consistent with SPM and CP 💪

xgouchet (approved):
LGTM

ncreated:
Tested on M1 laptop against this branch - `pod install` works fine, project compiles and runs with no runtime issue ✨ .

PR #598, opened 2021-09-16
RUMM-1599 If crash occurs before starting RUM session send it within dummy session

ncreated:
### What and why?

📦 This PR adds an improvement to our Crash Reporting. Now it will also capture crashes that happened before starting any RUM view, which corresponds to issues during application launch (between calling `Datadog.initialize()` and displaying the very first view controller).

### How?

The problem was that `CrashReportingWithRUMIntegration` was requiring a `CrashContext` with associated `lastRUMViewEvent` (holding the UUID of previous app session). If app crashed before starting first view, that information was not available, and the integration was not sending RUM error.

The improvement is to handle this case by sending dummy RUM session: a dummy RUM view is created and the RUM error is attached to it, making up the session with arbitrary view name (`AppLaunch`) and duration (`1ns`):

Screenshot 2021-09-16 at 16 55 04

The new logic is added to `CrashReportingWithRUMIntegration`. This required small refactoring around passing RUM session sampling rate value - to share it, I introduced `RUMSessionSampler`.

### Review checklist


Sources/Datadog/FeaturesIntegration/CrashReporting/CrashReportingWithRUMIntegration.swift
```diff
@@ -53,16 +61,28 @@ internal struct CrashReportingWithRUMIntegration: CrashReportingIntegration {
...
+        // to the server time before processing. Following use of the current correction is not ideal (it's not the correction
+        // from the moment of crash), but this is the best approximation we can get.
         let currentTimeCorrection = dateCorrector.currentCorrection
 
+        if let lastRUMViewEvent = crashContext.lastRUMViewEvent {
+            sendCrashReportToPreviousSession(crashReport, rumViewEventFromPreviousSession: lastRUMViewEvent, using: currentTimeCorrection)
+        } else {
+            // As this will create a new RUM session, we must respect the user sampling configuration
+            if rumFeatureConfiguration.sessionSampler.isSampled() {
+                sendCrashReportToNewDummySession(crashReport, using: currentTimeCorrection)
```
maxep, line 74:
```suggestion
        } else if rumFeatureConfiguration.sessionSampler.isSampled() {
                // As this will create a new RUM session, we must respect the user sampling configuration
                sendCrashReportToNewDummySession(crashReport, using: currentTimeCorrection)
```

Sources/Datadog/FeaturesIntegration/CrashReporting/CrashReportingWithRUMIntegration.swift
```diff
@@ -78,6 +98,21 @@ internal struct CrashReportingWithRUMIntegration: CrashReportingIntegration {
         }
     }
 
+    /// If the crash occurred before starting RUM session (after initializing SDK, but before starting the first view) we don't have any session UUID to associate the error with.
+    /// In that situation, we create a dummy view to associate the error with. In result, a dummy RUM session will be started and it will contain one view and one error (crash).
+    private func sendCrashReportToNewDummySession(
```
maxep, line 103: Would it be possible that this method and `sendCrashReportToPreviousSession` calls the same implementation so we don't duplicate the format? It could involve renaming the methods. WDYT?

maxep (approved):
Looks great! I've some minor suggestions

ncreated:
Adding this feature in current shape would introduce discrepancy between iOS and Android "app launch" crashes tracking.

I'll hold with this PR ✋ until we resolve "Collecting App Launch Crashes in Mobile SDKs" RFC.

Tests/DatadogTests/Datadog/Mocks/RUMFeatureMocks.swift
```diff
@@ -150,6 +150,24 @@ extension RUMEventsMapper {
     }
 }
 
+extension RUMSessionSampler {
+    static func mockAny() -> RUMSessionSampler {
```
Asavarkhul, line 154: Quick `codestyle` question, why not using computed properties? Isn't it more elegant to use sth like
`.mockAny` instead of `.mockAny()` 🙌

Sources/Datadog/FeaturesIntegration/CrashReporting/CrashReportingWithRUMIntegration.swift
```diff
@@ -16,8 +16,13 @@ internal struct CrashReportingWithRUMIntegration: CrashReportingIntegration {
         /// This condition originates from RUM backend constraints on processing `RUMViewEvents` in stale sessions. If the session does not
         /// receive any updates for a long time, then sending some significantly later may lead to inconsistency.
         static let viewEventAvailabilityThreshold: TimeInterval = 14_400 // 4 hours
+        /// The name of a view sent in dummy RUM session created when crash occurred before starting a real session.
+        static let dummyViewName = "ApplictionLaunch"
```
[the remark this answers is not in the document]
ncreated: Woah, small but impactful - this is a user facing value that will appear in Datadog web-ui 🙂. Good catch, thanks!

Tests/DatadogTests/Datadog/Mocks/RUMFeatureMocks.swift
ncreated: Next to `.mockAny()` (with no parameters) we also have `.mockWith(param1:param2:...)` convention. With static function semantic we can achieve similarity in both 🙂.

ncreated:
I'm closing this PR, as it will be shipped as part of separate topic of "collecting app launch events" as we concluded in internal RFC. It's way easier to me to first implement "app launch events" and then cherry pick stuff from this PR.

PR #600, opened 2021-09-21
RUMM-1609 Invoke Kronos from a single thread

### LINKED ISSUE #588: ThreadSanitizer data race
### The thing

Some of our developers are getting errors from the ThreadSanitizer. It seems `ServerDateProvider.synchronize` and `ServerDateProvider.currentDate` are calling `Kronos.Clock.now` and `Kronos.Clock.sync` from different threads (main and the com.datadoghq.rum-monitor queue), which apparently is discouraged by the Kronos developers. See this issue: https://github.com/MobileNativeFoundation/Kronos/issues/85

Attached is a debug log with the errors.

[debuglog.txt](https://github.com/DataDog/dd-sdk-ios/files/7116185/debuglog.txt)

# Environment

XCode 12.5
Simulator with iOS 14.5
DatadogSDK 1.6.0

maxep:
### What and why?

As per [this issue comment](https://github.com/MobileNativeFoundation/Kronos/issues/85#issuecomment-822664766), `Clock.now` and `.sync` from Kronos are expected to be called from the same thread.

Fix #588 

### How?

Invoke Kronos from a dedicated `DispatchQueue`.

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/System/Time/ServerDateProvider.swift
+++ b/Sources/Datadog/Core/System/Time/ServerDateProvider.swift
@@ -11,20 +11,28 @@ import Kronos
 internal protocol ServerDateProvider {
     /// Start the clock synchronisation with NTP server.
     /// Calls the `completion` by passing it the server time when the synchronization succeeds or`nil` if it fails.
-    func synchronize(with ntpPool: String, completion: @escaping (Date?) -> Void)
+    func synchronize(with pool: String, completion: @escaping (Date?) -> Void)
     /// Returns the server time or `nil` if not yet determined.
     /// This time gets more precise while synchronization is pending.
     func currentDate() -> Date?
 }
 
 internal class NTPServerDateProvider: ServerDateProvider {
-    func synchronize(with ntpPool: String, completion: @escaping (Date?) -> Void) {
-        Clock.sync(from: ntpPool, completion: { serverTime, _ in
-            completion(serverTime)
-        })
+    /// Queue used for invoking Kronos clock.
+    private let queue = DispatchQueue(
+        label: "com.datadoghq.date-provider",
+        target: .global(qos: .userInteractive)
+    )
+
+    func synchronize(with pool: String, completion: @escaping (Date?) -> Void) {
+        queue.async {
+            Clock.sync(from: pool, completion: { serverTime, _ in
+                completion(serverTime)
+            })
+        }
     }
 
     func currentDate() -> Date? {
-        return Clock.now
+        queue.sync { Clock.now }
     }
 }


maxep:
@ncreated @buranmert we could also make `currentDate` async but this would require some refactoring.
I'm also not sure how to test this since `ServerDateProvider` is mocked in our test suite 🤔

Sources/Datadog/Core/System/Time/ServerDateProvider.swift
```diff
@@ -11,20 +11,28 @@ import Kronos
...
+        queue.async {
+            Clock.sync(from: pool, completion: { serverTime, _ in
+                completion(serverTime)
+            })
+        }
     }
 
     func currentDate() -> Date? {
-        return Clock.now
+        queue.sync { Clock.now }
```
ncreated, line 36: Have we managed to reproduce this issue and are we sure this is the right fix 🤔💭?

This TSAN issue in [Kronos](https://github.com/MobileNativeFoundation/Kronos) is raised by two simultaneous operations accessing the same memory:
* [`self.stableTime` write](https://github.com/MobileNativeFoundation/Kronos/blob/3b2783e4d7b13fa9934692522b2c011ee0a12231/Sources/Clock.swift#L76) - coming from `Clock.sync` completion (called multiple times while sync is pending!);
* [`self.stableTime` read](https://github.com/MobileNativeFoundation/Kronos/blob/3b2783e4d7b13fa9934692522b2c011ee0a12231/Sources/Clock.swift#L50) - coming from accessing `Clock.now`.

Digging deeper into the completion block execution in Kronos, it [leads to this](https://github.com/MobileNativeFoundation/Kronos/blob/3b2783e4d7b13fa9934692522b2c011ee0a12231/Sources/DNSResolver.swift#L59):
```swift
CFHostSetClient(hostReference, callback, &clientContext)
CFHostScheduleWithRunLoop(hostReference, CFRunLoopGetMain(), CFRunLoopMode.commonModes.rawValue)
```
which looks like always running the completion callback on the main thread. This is in fact what I can observe when putting two basic `print()` statements in our code and running `Example` project:

Screenshot 2021-09-22 at 12 08 42

To add further, [Kronos maintainers say they call both APIs from the main thread](https://github.com/MobileNativeFoundation/Kronos/issues/85#issuecomment-822664766):
> In our app we only ever call this from the main thread. I think we'd be happy to support this if you wanted to submit a change for it though

So, by using the `queue`, we only synchronize calls to `Clock.sync` and `Clock.now` (which is not the issue), but we do not force the `completion` block to be called on the same thread as we later use `Clock.now`. **The completion is always called on the `main` thread**.

Before going further - does my observation make sense @maxep @buranmert ?

ncreated (changes requested):

Sources/Datadog/Core/System/Time/ServerDateProvider.swift
maxep: That's a good point! This wasn't my understanding of [this comment](https://github.com/MobileNativeFoundation/Kronos/issues/85#issuecomment-822664766): `So technically right now we expect you to call .now and .sync from the same thread.`

If actually the `.now ` should be called from the same thread as the completion block of `.sync`, then yes it should be called on the `.main`. Seeing [this](https://github.com/MobileNativeFoundation/Kronos/blob/master/Sources/Clock.swift#L84), your observation makes total sense! Now, having to call `NTPServerDateProvider.currentData` on the main thread will require `async` for sure. I will also look at submitting a change to Kronos.

> I'm not able to reproduce but I will do more investigation.

ncreated: > Now, having to call NTPServerDateProvider.currentData on the main thread will require async for sure. I will also look at submitting a change to Kronos.

Yes, I think these are valid choices.

IMO fixing the issue in Kronos seems more correct - right now it seems to enforce main thread access to `Clock.now`, `Clock.annotatedNow` and `Clock.timestamp` to all users. As this is not stated in their public doc this might be a bug and we could contribute.

However, if we want to fix it on our side, IMO `queue.async {}` is not an easy way to go. That might enforce quite refactoring to data collection system, as we read `currentDate()` synchronously in many places.

A better idea would be to use our [`ValuePublisher`](https://github.com/DataDog/dd-sdk-ios/blob/master/Sources/Datadog/Core/Utils/ValuePublisher.swift) which is meant for building async / sync bridge in a performant way. It uses [concurrent queue optimised for simultaneous reads and exclusive writes](https://github.com/DataDog/dd-sdk-ios/blob/6fc6d5d8ed8b60794ca0e7bea7741dfc7bd991c3/Sources/Datadog/Core/Utils/ValuePublisher.swift#L35), which fits perfectly our usage of `Clock.sync` (write) and `Clock.now` (read):
* `Clock.sync` is used **only once**, in `Datadog.initialize()`, to receive the real NTP time;
* `Clock.now` is called from different threads **for every event** we create (log / span / rum).

With above constraints, and `ValuePublisher`, we could try something this in `NTPServerDateProvider`:
```swift
internal class NTPServerDateProvider: ServerDateProvider {
    private let serverTimePublisher: ValuePublisher<Date?> = ValuePublisher(initialValue: .empty)

    func synchronize(with ntpPool: String, completion: @escaping (Date?) -> Void) {
        Clock.sync(from: ntpPool, completion: { serverTime, _ in
            self.serverTimePublisher.publishAsync(serverTime)
            completion(serverTime)
        })
    }

    func currentDate() -> Date? {
        return serverTimePublisher.currentValue
    }
}
```
This way, the real (NTP) will be available after sync completes and all should be thread-safe. This would be regressive to what we have now, where `Clock.now` is available earlier than completion and it becomes more precise while NTP sync continues (to become stable when sync ends). To fill this gap further, we could have a look at Kronos [`first` completion API](https://github.com/MobileNativeFoundation/Kronos/blob/3b2783e4d7b13fa9934692522b2c011ee0a12231/Sources/Clock.swift#L67) and even look closer at their [`UserDefaults` capability](https://github.com/MobileNativeFoundation/Kronos/blob/3b2783e4d7b13fa9934692522b2c011ee0a12231/Sources/Clock.swift#L95). By using both to additionally feed `serverTimePublisher`, we should achieve similar behaviour to what we have now, in a thread-safe manner.

maxep: Nice! Thanks for pointing out the `ValuePublisher`, this would be a great approach indeed. I will first update this PR with a proposal and then look at modifying Kronos 👍

maxep:
Closing in favor of #607

PR #605, opened 2021-09-23
RUMM-1605 Use Native resource type instead of XHR

xgouchet:
### What and why?

Use `native` resource type instead of `xhr`. The latter was used as a legacy based on the Browser RUM SDK, but we need something more relevant for mobile apps.

### DIFF
--- a/Sources/Datadog/RUMMonitor.swift
+++ b/Sources/Datadog/RUMMonitor.swift
@@ -27,11 +27,11 @@ internal extension RUMResourceType {
     /// - Parameters:
     ///   - request: the `URLRequest` for the resource.
     init?(request: URLRequest) {
-        let xhrHTTPMethods: Set<String> = ["POST", "PUT", "DELETE"]
+        let nativeHTTPMethods: Set<String> = ["POST", "PUT", "DELETE"]
 
         if let requestMethod = request.httpMethod?.uppercased(),
-           xhrHTTPMethods.contains(requestMethod) {
-            self = .xhr
+            nativeHTTPMethods.contains(requestMethod) {
+            self = .native
         } else {
             return nil
         }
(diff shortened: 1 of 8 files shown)


Sources/Datadog/RUM/AutoInstrumentation/Resources/URLSessionRUMResourcesHandler.swift
```diff
@@ -41,7 +41,7 @@ internal class URLSessionRUMResourcesHandler: URLSessionInterceptionHandler {
                 attributes: [:],
                 url: url,
                 httpMethod: RUMMethod(httpMethod: interception.request.httpMethod),
-                kind: RUMResourceType(request: interception.request),
+                kind: nil,
```
buranmert, line 44: i guess you forgot `nil` here.?

buranmert (changes requested):

Sources/Datadog/RUM/AutoInstrumentation/Resources/URLSessionRUMResourcesHandler.swift
ncreated: And I'm wondering, why do we remove `RUMResourceType(request: interception.request)` here 🤔💭?

**The way it works:** 
We first try to deduce `kind` from request, and then read it from `response`. At the end, `request` has priority:
https://github.com/DataDog/dd-sdk-ios/blob/6fc6d5d8ed8b60794ca0e7bea7741dfc7bd991c3/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift#L129

xgouchet: I tried something and forgot to change this back

buranmert (approved):

ncreated (approved):

PR #607, opened 2021-09-24
RUMM-1609 Use NTP server offset

### LINKED ISSUE #588: ThreadSanitizer data race

maxep:
### What and why?

The Kronos `Clock.now` property is not thread safe and should always be invoked from the main thread, this change would have a large impact on our codebase. As an alternative, we can rely on the server time offset we get from `Clock.sync`: The `Clock.now` is in fact used by `DateCorrector` to actually compute the server offset.

Below a quick benchmark of the computed offset previously used and the server offset applied by this PR:
|Previous Computed Offset|Server Offset|
|---|---|
|**0.0537**11891174316406|**0.0537**5981330871582|
|**0.0476**3531684875488|**0.0476**7334461212158|
|**0.0457**2319984436035|**0.0457**66234397888184|
|**0.081**27260208129883|**0.081**31551742553711|
|**0.051**30136013031006|**0.051**299452781677246|
|**0.0388**5924816131592|**0.0388**74030113220215|

> Note that `Kronos.Clock.now` has a [sub-second precision](https://github.com/MobileNativeFoundation/Kronos/blob/master/Sources/TimeFreeze.swift#L69) which is reflected in the above tests.

Fix #588

### How?

The `ServerDateProvider` will now provide an `offset` time interval and its concrete implementation `NTPServerDateProvider` will retrieve that value from `Kronos.Clock.sync`.

`Clock.sync` callback is invoked on the main thread, so we  are now using a `ValuePublisher` to safely access the server time offset. 

⚠️  `Clock.sync` only notifies the first and last samples, so we are losing the intermediate offsets!

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/System/Time/ServerDateProvider.swift
+++ b/Sources/Datadog/Core/System/Time/ServerDateProvider.swift
@@ -10,21 +10,41 @@ import Kronos
 /// Abstract the monotonic clock synchronized with the server using NTP.
 internal protocol ServerDateProvider {
     /// Start the clock synchronisation with NTP server.
-    /// Calls the `completion` by passing it the server time when the synchronization succeeds or`nil` if it fails.
-    func synchronize(with ntpPool: String, completion: @escaping (Date?) -> Void)
-    /// Returns the server time or `nil` if not yet determined.
-    /// This time gets more precise while synchronization is pending.
-    func currentDate() -> Date?
+    /// Calls the `completion` by passing it the server time offset when the synchronization succeeds or`nil` if it fails.
+    func synchronize(with pool: String, completion: @escaping (TimeInterval?) -> Void)
+    /// Returns the server time offset or `nil` if not yet determined.
+    /// This offset gets more precise while synchronization is pending.
+    var offset: TimeInterval? { get }
 }
 
 internal class NTPServerDateProvider: ServerDateProvider {
-    func synchronize(with ntpPool: String, completion: @escaping (Date?) -> Void) {
-        Clock.sync(from: ntpPool, completion: { serverTime, _ in
-            completion(serverTime)
-        })
+    /// Server offset publisher.
+    private let publisher: ValuePublisher<TimeInterval?> = ValuePublisher(initialValue: nil)
+
+    /// Returns the server time offset or `nil` if not yet determined.
+    /// This offset gets more precise while synchronization is pending.
+    var offset: TimeInterval? {
+        return publisher.currentValue
     }
 
-    func currentDate() -> Date? {
-        return Clock.now
+    func synchronize(with pool: String, completion: @escaping (TimeInterval?) -> Void) {
+        // Kronos only notifies for the first and last samples.
+        // In case, the last sample does not return an offset, we calculate the offset
+        // from `Clock.now`.
+        Clock.sync(
+            from: pool,
+            first: { _, offset in
+                self.publisher.publishAsync(offset)
+            },
+            completion: { now, offset in
+                if let offset = offset {
+                    self.publisher.publishAsync(offset)
+                } else if let now = now {
+                    self.publisher.publishAsync(now.timeIntervalSinceNow)
+                }
+
+                completion(self.publisher.currentValue)
+            }
+        )
     }
 }
--- a/Sources/Datadog/Core/System/Time/ServerDateProvider.swift
+++ b/Sources/Datadog/Core/System/Time/ServerDateProvider.swift
@@ -47,5 +47,9 @@ internal class NTPServerDateProvider: ServerDateProvider {
                 completion(self?.publisher.currentValue)
             }
         )
+
+        // `Kronos.sync` first loads the previous state from the `UserDefaults` if any.
+        // We can invoke `Clock.now` to retrieve the stored offset.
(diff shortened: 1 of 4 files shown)


Sources/Datadog/Core/System/Time/ServerDateProvider.swift
```diff
@@ -10,21 +10,41 @@ import Kronos
...
+    var offset: TimeInterval? {
+        return publisher.currentValue
     }
 
-    func currentDate() -> Date? {
-        return Clock.now
+    func synchronize(with pool: String, completion: @escaping (TimeInterval?) -> Void) {
+        // Kronos only notifies for the first and last samples.
+        // In case, the last sample does not return an offset, we calculate the offset
+        // from `Clock.now`.
```
ncreated, line 33: This seems to be a leftover - I can't see `Clock.now` anywhere, no?

Sources/Datadog/Core/System/Time/ServerDateProvider.swift
```diff
@@ -10,21 +10,41 @@ import Kronos
...
+        Clock.sync(
+            from: pool,
+            first: { _, offset in
+                self.publisher.publishAsync(offset)
+            },
+            completion: { now, offset in
```
ncreated, line 39: I think it's safer / no cost to use `[weak self]` capturing in these closures. Although it won't happen in production apps (the SDK can't be deinitialized), we deinitialize SDK quite often in some tests (including E2E tests, where NTP time is not mocked).

Sources/Datadog/Core/System/Time/ServerDateProvider.swift
```diff
@@ -10,21 +10,41 @@ import Kronos
...
+                if let offset = offset {
+                    self.publisher.publishAsync(offset)
+                } else if let now = now {
+                    self.publisher.publishAsync(now.timeIntervalSinceNow)
```
ncreated, line 43: Is it possible to get `now` but not `offset`, or this is just a sanity fallback? In any case, it would be good to have inline comment on this, otherwise it seems unclear.

ncreated (changes requested):
> ⚠️ `Clock.sync` only notifies the first and last samples, so we are losing the intermediate offsets!

Is it true that previously (with using `Clock.now`) we were receiving some initial approximation even before `Clock.sync's` `first {}` callback?

I'm wondering if we can cover this gap by using `Clock.now` right after `Clock.sync`:
```swift
func synchronize(with pool: ...) {
   Clock.sync(
      // ...
      first: { publisher.publishAsync(...) } // first approximation
      completion: { publisher.publishAsync(...) } // stable result
   )

   publisher.publishAsync(Clock.now - currentDeviceTime()) // ← this
}
```
Kronos [seems to integrate with UserDefaults](https://github.com/MobileNativeFoundation/Kronos/blob/3b2783e4d7b13fa9934692522b2c011ee0a12231/Sources/Clock.swift#L72) and stores the last-known NTP time in there. With this change we should read it back as soon as possible. This will be thread safe and will be as precise as before, no?

Sources/Datadog/Core/System/Time/ServerDateProvider.swift
maxep: Yes, my comment above is misplaced. This is indeed a fallback in case the last sample returns a `nil` offset. 
> the `now` parameter is `Clock.now`.

maxep: I've updated my comment, LMKWYT!

maxep:
Yes, Kronos first loads the previous offset from the UserDefaults. Their interface is not public, but we can indeed do this workaround 👍 It is thread-safe since their load from the user defaults on the current thread, but they could change that behavior any time.

ncreated (approved):

ncreated:
> Yes, Kronos first loads the previous offset from the UserDefaults. Their interface is not public, but we can indeed do this workaround 👍 It is thread-safe since they load from the user defaults on the current thread, but they could change that behavior any time.

It's IMO fine 👍. Kronos now requires using `Clock.sync` and `Clock.now` on the same thread (which we do now 🚀), and it's hard to to imagine this changing to "must be used on different threads".

maxep:
> Kronos now requires using `Clock.sync` and `Clock.now` on the same thread

Actually, their requirements should be to 'invoke `Clock.now` on the main thread'. **If**, at some point, they decide to dispatch [`loadFromDefaults`](https://github.com/MobileNativeFoundation/Kronos/blob/3b2783e4d7b13fa9934692522b2c011ee0a12231/Sources/Clock.swift#L72) on the main thread, then we would still risk a data race.

PR #608, opened 2021-09-24
RUMM-1278 RefreshRate is normalized to 0...60

buranmert:
### What and why?

`RefreshRate` is shown as a mobile vital in Datadog UI.
It needs to be normalized as different devices can have different maximum FPS rates.

### How?

It was normalized to 0...1, this was not consistent with other platforms.
Now it is normalized to 0...60 range, as browser and Android SDKs do.

### Review checklist

### DIFF
--- a/Sources/Datadog/RUM/RUMVitals/VitalInfoSampler.swift
+++ b/Sources/Datadog/RUM/RUMVitals/VitalInfoSampler.swift
@@ -39,6 +39,8 @@ internal final class VitalInfoSampler {
 
     var refreshRate: VitalInfo {
         let info = refreshRatePublisher.currentValue
+        // NOTE: RUMM-1278 refreshRate is normalized to 0...60 range
+        // assuming 60 is the industry standard.
         return info.scaledDown(by: maximumRefreshRate / 60.0)
     }
 
--- a/Sources/Datadog/RUM/RUMVitals/VitalInfoSampler.swift
+++ b/Sources/Datadog/RUM/RUMVitals/VitalInfoSampler.swift
@@ -17,6 +17,12 @@ internal protocol ContinuousVitalReader {
 }
 
 internal final class VitalInfoSampler {
+    struct Constants {
+        // We use normalized 0...60 range for refresh rate in Mobile Vitals,
+        // assuming 60 is the industry standard.
+        static let normalizedRefreshRate = 60.0
+    }
+
     private static let frequency: TimeInterval = 1.0
 
     let cpuReader: SamplingBasedVitalReader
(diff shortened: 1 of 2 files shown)


Tests/DatadogTests/Datadog/RUM/RUMVitals/VitalInfoSamplerTests.swift
```diff
@@ -39,7 +39,7 @@ class VitalInfoSamplerTests: XCTestCase {
             XCTAssertEqual(sampler.memory.meanValue, 321.0)
             XCTAssertGreaterThan(sampler.memory.sampleCount, 1)
             let maxFPS = Double(UIScreen.main.maximumFramesPerSecond)
-            XCTAssertEqual(sampler.refreshRate.meanValue, 666.0 / maxFPS)
```
xgouchet, line 42: You could have left the 666 value and then used:
```swift
XCTAssertEqual(sampler.refreshRate.meanValue, (666.0 / maxFPS) * 60.0)
```

xgouchet (approved):
LGTM

Sources/Datadog/RUM/RUMVitals/VitalInfoSampler.swift
```diff
@@ -39,7 +39,7 @@ internal final class VitalInfoSampler {
 
     var refreshRate: VitalInfo {
         let info = refreshRatePublisher.currentValue
-        return info.scaledDown(by: maximumRefreshRate)
+        return info.scaledDown(by: maximumRefreshRate / 60.0)
```
ncreated, line 42: How about defining a constant and adding a comment on normalization? I'm afraid that this might become forgotten and unclear.

buranmert: comment added to the line above
> NOTE: `refreshRate` is normalized to 0...60 range, assuming 60 is the industry standard.

maxep: I agree with using a constant value for this!

buranmert: i can't come up with a local variable name which doesn't make things more confusing 🤔 
any suggestions?

ncreated: ’’’swift
struct Constants {
    // We use normalized 0...60 range for refresh rate in Mobile Vitals,
    // assuming 60 is the industry standard.
    static let normalizedRefreshRate = 60
}
’’’
Is +/- what we do in other places.

maxep (approved):
🙌

Issue #609, opened 2021-09-26
iOS 15 crash on MobileDevice.swift


arnauddorgans:
### The crash

Random crash on iOS 15 `MobileDevice.swift` / `isLowPowerModeEnabled`

<img width="1264" alt="Capture d’écran 2021-09-27 à 00 05 11" src="https://user-images.githubusercontent.com/10015557/134825733-170a825d-6cfd-44f3-b7ea-4039a8f3aac4.png">

```
crash_info_entry_0
BUG IN CLIENT OF LIBPLATFORM: Trying to recursively lock an os_unfair_lock

Crashed: com.datadoghq.ios-sdk-rum-upload
0  libsystem_platform.dylib       0x1f3c620c0 _os_unfair_lock_recursive_abort + 36
1  libsystem_platform.dylib       0x1f3c5ca10 _os_unfair_lock_lock_slow + 304
2  Foundation                     0x184fea730 -[NSProcessInfo(NSProcessInfoHardwareState) isLowPowerModeEnabled] + 68
3  WebCore                        0x19329d004 <redacted> + 56
4  CoreFoundation                 0x1837c1ee8 __CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ + 28
5  CoreFoundation                 0x18385db9c ___CFXRegistrationPost_block_invoke + 52
6  CoreFoundation                 0x183830f54 _CFXRegistrationPost + 456
7  CoreFoundation                 0x1837d7d54 _CFXNotificationPost + 716
8  Foundation                     0x184fdc028 -[NSNotificationCenter postNotificationName:object:userInfo:] + 96
9  Foundation                     0x1850549d4 NSProcessInfoNotifyPowerState + 188
10 Foundation                     0x184fea768 -[NSProcessInfo(NSProcessInfoHardwareState) isLowPowerModeEnabled] + 124
11 Datadog                        0x105285390 partial apply for closure #3 in MobileDevice.init(uiDevice:processInfo:) + 67 (MobileDevice.swift:67)
12 Datadog                        0x105287d50 DataUploadConditions.blockersForUpload() + 31 (BatteryStatusProvider.swift:31)
13 Datadog                        0x105288838 closure #1 in DataUploadWorker.scheduleNextUpload(after:) + 56 (DataUploadWorker.swift:56)
[...]

PR #613, opened 2021-09-28
Prevent iOS 15 crashes by not polling isLowPowerModeEnabled

earltedly:
### What and why?

As described in #609, iOS 15 has introduced a race condition around `ProcessInfo.isLowPowerModeEnabled`:

The conditions under which this can occur are:

1. Register to observer: `NSProcessInfoPowerStateDidChangeNotification`
2. In the observation block, query `ProcessInfo.isLowPowerModeEnabled`
3. Later on another piece of code calls `ProcessInfo.isLowPowerModeEnabled`
4. _Sometimes_ that will trigger a change in the power state which will be broadcast to observers before returning
5. The code in (2) is called which re-enters `isLowPowerModeEnabled`
6. Thread deadlock over an `os_unfair_lock` occurs and the app crashes

### How?

We introduce a `LowPowerModeMonitor` which:

1. Listens to `NSProcessInfoPowerStateDidChangeNotification`
2. On any updates it queries `ProcessInfo.isLowPowerModeEnabled` and caches the return value

`MobileDevice` retains an instance of this in the local scope and queries it instead of `ProcessInfo` locally.

This prevents us from constantly polling for new values and thus takes us out of the call stack for re-entry.

### Review checklist

### Notes

This is a change in the implementation details rather than behaviour, so I've not added any additional tests

### DIFF
--- a/Sources/Datadog/Core/System/MobileDevice.swift
+++ b/Sources/Datadog/Core/System/MobileDevice.swift
@@ -54,6 +54,7 @@ internal class MobileDevice {
 
     convenience init(uiDevice: UIDevice, processInfo: ProcessInfo) {
         let wasBatteryMonitoringEnabled = uiDevice.isBatteryMonitoringEnabled
+        let lowPowerModeMonitor = LowPowerModeMonitor(processInfo)
         self.init(
             model: uiDevice.model,
             osName: uiDevice.systemName,
@@ -64,12 +65,12 @@ internal class MobileDevice {
                 return BatteryStatus(
                     state: MobileDevice.toBatteryState(uiDevice.batteryState),
                     level: uiDevice.batteryLevel,
-                    isLowPowerModeEnabled: processInfo.isLowPowerModeEnabled
+                    isLowPowerModeEnabled: lowPowerModeMonitor.isLowPowerModeEnabled
                 )
             }
         )
     }
-
+    
     /// Returns current mobile device  if `UIDevice` is available on this platform.
     /// On other platforms returns `nil`.
     static var current: MobileDevice {
@@ -96,6 +97,32 @@ internal class MobileDevice {
         case .charging:     return .charging
         case .full:         return .full
         @unknown default:   return.unknown
+        }
+    }
+}
+
+private class LowPowerModeMonitor {
+    
+    var isLowPowerModeEnabled = false
+    private var powerStateDidChangeObserver: Any?
+    
+    init(_ processInfo: ProcessInfo) {
+        isLowPowerModeEnabled = processInfo.isLowPowerModeEnabled
+        powerStateDidChangeObserver = NotificationCenter
+            .default
+            .addObserver(forName: .NSProcessInfoPowerStateDidChange,
+                         object: nil,
+                         queue: .main) { [weak self] notification in
+                guard let processInfo = notification.object as? ProcessInfo else {
+                    return
+                }
+                self?.isLowPowerModeEnabled = processInfo.isLowPowerModeEnabled
+            }
+    }
+    
+    deinit {
+        if let observer = powerStateDidChangeObserver {
+            NotificationCenter.default.removeObserver(observer)
         }
     }
 }
--- a/Sources/Datadog/Core/System/MobileDevice.swift
+++ b/Sources/Datadog/Core/System/MobileDevice.swift
@@ -70,7 +70,6 @@ internal class MobileDevice {
             }
         )
     }
-    
     /// Returns current mobile device  if `UIDevice` is available on this platform.
     /// On other platforms returns `nil`.
     static var current: MobileDevice {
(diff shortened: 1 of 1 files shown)


Sources/Datadog/Core/System/MobileDevice.swift
```diff
@@ -99,3 +100,29 @@ internal class MobileDevice {
...
+        isLowPowerModeEnabled = processInfo.isLowPowerModeEnabled
+        powerStateDidChangeObserver = NotificationCenter
+            .default
+            .addObserver(forName: .NSProcessInfoPowerStateDidChange,
+                         object: nil,
+                         queue: .main) { [weak self] notification in
+                guard let processInfo = notification.object as? ProcessInfo else {
+                    return
+                }
+                self?.isLowPowerModeEnabled = processInfo.isLowPowerModeEnabled
```
buranmert, line 119: iOS fires this notification (so we write to it) in the main thread whereas we read it in background threads ,`UploadWorker`s have their own `DispatchQueue`s in which they read `MobileDevice.current`.

if read and write happen at the same time the app may crash. can we do read/write in the same `DispatchQueue`?

we have [a class called `ValuePublisher`](https://github.com/DataDog/dd-sdk-ios/blob/master/Sources/Datadog/Core/Utils/ValuePublisher.swift#L22) which can be used for this, but it doesn't exist in your branch.
can you rebase onto `master`? or you'd like to have this fix as a hotfix to `v1.5.2`?

buranmert (commented):
hey @earltedly 👋 
thanks so much for the PR, we are really happy to have your contribution! 🥳 

switching from polling to listening to notifications seems to be a step in the right direction, but i guess there is still a problem with `isLowPowerModeEnabled` property. i left a comment below 👇

Sources/Datadog/Core/System/MobileDevice.swift
```diff
@@ -99,3 +100,29 @@ internal class MobileDevice {
         }
     }
 }
+
+private class LowPowerModeMonitor {
```
Asavarkhul, line 104: Hey guys 👋

Quick question, does this class needs to be open to subclassing? Why not putting it as `final`?

Sources/Datadog/Core/System/MobileDevice.swift
earltedly: I've rebased onto `master` and updated to use a ValuePublisher - thanks for the prompts on the usual practices in this codebase 👍

Sources/Datadog/Core/System/MobileDevice.swift
earltedly: I'm fairly sure this is implicit for a `private` class, however always best to be certain so I've added `final` too 👍

earltedly:
Thanks for the review 🎉 . I think I've addressed everything, but please let me know if anything else is needed.

buranmert (changes requested):
once again thanks @earltedly 🙏 
unfortunately CI failed due to linter errors, i left suggestions to fix them.
once these trivial errors are fixed, CI will probably pass and we can merge the PR 🚀 

_note: you can see linter output as warnings in Xcode and/or by running `./tools/lint/run_linter.sh` in Terminal_

Sources/Datadog/Core/System/MobileDevice.swift
```diff
@@ -64,12 +65,12 @@ internal class MobileDevice {
...
                     state: MobileDevice.toBatteryState(uiDevice.batteryState),
                     level: uiDevice.batteryLevel,
-                    isLowPowerModeEnabled: processInfo.isLowPowerModeEnabled
+                    isLowPowerModeEnabled: lowPowerModeMonitor.isLowPowerModeEnabled
                 )
             }
         )
     }
-
+
```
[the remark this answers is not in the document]
buranmert:
```suggestion

```

Sources/Datadog/Core/System/MobileDevice.swift
```diff
@@ -99,3 +100,34 @@ internal class MobileDevice {
...
+        
+        powerStateDidChangeObserver = NotificationCenter
+            .default
+            .addObserver(forName: .NSProcessInfoPowerStateDidChange,
+                         object: nil,
+                         queue: .main) { [weak self] notification in
+                guard let processInfo = notification.object as? ProcessInfo else {
+                    return
+                }
+                self?.publisher.publishAsync(processInfo.isLowPowerModeEnabled)
```
yousefhamza, line 124: Hi @earltedly 👋 

Can you help me understand how accessing `processInfo` like this prevent against calling the deadlock?

I'm confused because the object that posts the notification will be the same object that we query for `isLowPowerModeEnabled` which cause the same issue.

earltedly:
@buranmert Thanks for the the CR and fix it suggestions, I've applied them all so 🤞 it will pass now :)

Sources/Datadog/Core/System/MobileDevice.swift
earltedly: Hi @yousefhamza 👋 

It's just as @buranmert, at a _worst case_ this will greatly reduce the number of times DatadogSDK access `processInfo` and thus reduces the potential for hitting this bug.

We can also hope that `processInfo` itself only notifies when the state has recently changed and that immediately querying it will be unlikely to cause `processInfo` to immediately find it has changed again (and thus re-broadcast within the same callstack).

Also, it would be great to hear if you're experiencing this too and in what kind of numbers.

buranmert (approved):

Sources/Datadog/Core/System/MobileDevice.swift
earltedly: Here's the radar: FB9661108. Filing a duplicate would be great as it increases the priority within Apple's triage system.

ncreated (approved):

Sources/Datadog/Core/System/MobileDevice.swift
yousefhamza: @buranmert @earltedly @ncreated  Thanks for the very thorough responses 🙏 

What I understand is it will help with this fix because:
1. It reduces how many times `isLowPowerModeenabled` is queried
1. Querying it after it has changed will give a much lower chance that it will change again while querying
1. Reading it from a single thread will reduce the chances of multiple reads (which can cause firing a notification) at the same time, while making it thread safe.

I'm working on [Instabug SDK](https://instabug.com) and [this issue came into my radar](https://twitter.com/steipete/status/1443506043733032962?s=20) because we use `ProcessInfo.processInfo.isLowPowerModeEnabled`, we are also prone to this crash too. however, we didn't get any reports of it yet.

ncreated: @yousefhamza yes, your reasoning about this change is correct 👍 with the additional note that our `MobileDevice` is a shared component, instantiated only once from SDK init - so we will have only one active subscriber to LPM-change notification.

Thanks for more details, and I hope you will manage to prevent this crash in Instabug SDK 🙂🤞!

maxep (approved):
👌

PR #616, opened 2021-09-29
RUMM-1377 Add E2E tests for Tracing APIs

ncreated:
### What and why?

📦 This PR adds E2E tests and their monitor definitions for Tracing product.

### How?

Added tests with their monitors (`terraform plan` works fine). I haven't yet updated monitors in Mobile Integration org - that needs to be done after merge otherwise they will have no data.

I also added two changes to E2E tests linter:
* it now checks if `$feature` variable is one of allowed values (`core`, `logs`, `trace`, `rum`, `crash`),
* it no longer checks if `$monitor_query` contains the test method name (the way we set up some monitors was not compatible with this rule).

### Review checklist

### DIFF
--- a/Datadog/E2ETests/Tracing/SpanE2ETests.swift
+++ b/Datadog/E2ETests/Tracing/SpanE2ETests.swift
@@ -1,0 +1,181 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Datadog
+
+class SpanE2ETests: E2ETests {
+    /// - api-surface: OTSpan.setOperationName(_ operationName: String)
+    ///
+    /// - data monitor:
+    /// ```apm
+    /// $feature = trace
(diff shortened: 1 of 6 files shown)


Datadog/E2ETests/Tracing/SpanE2ETests.swift
```diff
@@ -0,0 +1,181 @@
...
+    /// - api-surface: OTSpan.setBaggageItem(key: String, value: String)
+    ///
+    /// - data monitor: (it uses `ios_trace_span_set_baggage_item` metric defined in "APM > Generate Metrics > Custom Span Metrics")
+    /// ```apm
+    /// $feature = trace
+    /// $monitor_id = trace_span_set_baggage_item_data
+    /// $monitor_name = "[RUM] [iOS] Nightly - trace_span_set_baggage_item: number of hits is below expected value"
+    /// $monitor_query = "sum(last_1d):avg:ios_trace_span_set_baggage_item.hits_with_proper_payload{*}.as_count() < 1"
+    /// $monitor_threshold = 1
+    /// ```
```
ncreated, line 77: This kind of monitors require manual management of custom span metric generated at ingestion
Screenshot 2021-09-29 at 13 51 53

This can't be automated and this is the only way possible if we want to make monitor assertions on span tags.

PR #620, opened 2021-10-01
RUMM-1394 Add more verbosity to flaky test errors

ncreated:
### What and why?

📦 This PR adds a bit more verbosity to errors which we capture in flaky tests. 

Looking at last `7d`, we got 11 failures in nightly tests:
Screenshot 2021-10-01 at 17 42 37

Unfortunately, the error messages are not handy in troubleshooting:

Screenshot 2021-10-01 at 17 44 45

As I couldn't reproduce these failures locally (in exactly the same OS + simulator setup and with `dd-swift-testing` instrumentation enabled), I propose to first add more verbosity in this PR, and collect more data in CI App.

### How?

I improved the way we compare two task instances. Instead of:
```swift
XCTAssertTrue(actualTask === expectedTask)
```
now it uses a helper method:
```swift
AssertURLSessionTasksIdentical(actualTask, expectedTask)
```
which dumps additional information about the `lhs` and `rhs` tasks if they are not identical, e.g:

Screenshot 2021-10-01 at 17 48 41

### Review checklist

### DIFF
--- a/Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegateTests.swift
+++ b/Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegateTests.swift
@@ -94,9 +94,11 @@ class DDURLSessionDelegateTests: XCTestCase {
 
         // When
         let taskWithURL = session.dataTask(with: URL.mockAny())
+        taskWithURL.taskDescription = "taskWithURL"
         taskWithURL.resume()
 
         let taskWithURLRequest = session.dataTask(with: URLRequest(url: .mockAny()))
+        taskWithURLRequest.taskDescription = "taskWithURLRequest"
         taskWithURLRequest.resume()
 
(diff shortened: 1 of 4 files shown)


Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegateTests.swift
```diff
@@ -94,26 +94,28 @@ class DDURLSessionDelegateTests: XCTestCase {
 
         // When
         let taskWithURL = session.dataTask(with: URL.mockAny())
+        taskWithURL.taskDescription = "taskWithURL"
```
ncreated, line 97: As (in flaky scenario) we receive some other task than we expect, here I tag each task with a custom description, so we can see it in the error (expected vs actual `taskDescription`).

buranmert (approved):

maxep (approved):

PR #621, opened 2021-10-04
RUMM-1688 Add tests to "Low Power Mode" setting changes

ncreated:
### What and why?

📦 This PR adds tests to changing "Low Power Mode" setting.

This is to add coverage and prevent regression in fix merged in #613.

### How?

I changed existing test to update `BatteryStatus.isLowPowerModeEnabled` on notification, instead of pulling it from `ProcessInfo`.

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/System/MobileDevice.swift
+++ b/Sources/Datadog/Core/System/MobileDevice.swift
@@ -54,7 +54,11 @@ internal class MobileDevice {
 
     convenience init(uiDevice: UIDevice, processInfo: ProcessInfo) {
         let wasBatteryMonitoringEnabled = uiDevice.isBatteryMonitoringEnabled
-        let lowPowerModeMonitor = LowPowerModeMonitor(processInfo)
+
+        // We capture this `lowPowerModeMonitor` in `currentBatteryStatus` closure so its lifecycle
+        // is owned and controlled by `MobileDevice` object.
+        let lowPowerModeMonitor = LowPowerModeMonitor(initialProcessInfo: processInfo)
+
         self.init(
             model: uiDevice.model,
             osName: uiDevice.systemName,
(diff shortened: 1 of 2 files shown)


maxep (approved):

Sources/Datadog/Core/System/MobileDevice.swift
```diff
@@ -100,14 +104,24 @@ internal class MobileDevice {
     }
 }
 
+/// Observes "Low Power Mode" setting changes and provides `isLowPowerModeEnabled` value in a thread-safe manner.
+///
+/// Note: this was added in https://github.com/DataDog/dd-sdk-ios/issues/609 where `ProcessInfo.isLowPowerModeEnabled` was considered
+/// not thread-safe on iOS 15. With this monitor, we change from pulling to push model for reading this property. Now, it will never be read simultaneously
+/// by multiple SDK threads - instead it will be read only once after LMP setting change and bridged to other threads through thread-safe `ValuePublisher`.
```
buranmert, line 111:
```suggestion
/// by multiple SDK threads - instead it will be read only once after LPM setting change and bridged to other threads through thread-safe `ValuePublisher`.
```

Tests/DatadogTests/Datadog/Core/System/MobileDeviceTests.swift
```diff
@@ -52,13 +52,25 @@ class MobileDeviceTests: XCTestCase {
...
+            uiDevice: UIDeviceMock(),
+            processInfo: ProcessInfoMock(isLowPowerModeEnabled: isLowPowerModeEnabled)
         )
-        XCTAssertFalse(
-            MobileDevice(uiDevice: UIDeviceMock(), processInfo: ProcessInfoMock(isLowPowerModeEnabled: false)).currentBatteryStatus().isLowPowerModeEnabled
+
+        XCTAssertEqual(mobileDevice.currentBatteryStatus().isLowPowerModeEnabled, isLowPowerModeEnabled)
+
+        // When
+        NotificationCenter.default.post(
```
buranmert, line 67: if you think we might add more test cases to `MobileDeviceTests` later, you may want to use different `NotificationCenter` instances like [this](https://github.com/DataDog/dd-sdk-ios/blob/master/Sources/Datadog/Core/System/AppStateListener.swift#L108) to avoid interference between cases.

buranmert (approved):

Tests/DatadogTests/Datadog/Core/System/MobileDeviceTests.swift
ncreated: Good idea 👌 - added.

Issue #623, opened 2021-10-05
CarrierInfoProvider.swift line 129


safa-ads:
### The crash

Crashed: com.datadoghq.ios-sdk-tracing-upload
EXC_BAD_ACCESS KERN_INVALID_ADDRESS 0x0000000f42a79040

```
Crashed: com.datadoghq.ios-sdk-tracing-upload
0  libswiftCore.dylib             0x189cba440 swift_unknownObjectRetain + 16
1  libswiftFoundation.dylib       0x189884818 closure #2 in static Dictionary._forceBridgeFromObjectiveC(_:result:) + 644
2  libswiftFoundation.dylib       0x1898848c4 partial apply for closure #2 in static Dictionary._forceBridgeFromObjectiveC(_:result:) + 32
3  libswiftFoundation.dylib       0x189888f30 _NativeDictionary.init(_unsafeUninitializedCapacity:allowingDuplicates:initializingWith:) + 252
4  libswiftFoundation.dylib       0x189883528 static Dictionary._unconditionallyBridgeFromObjectiveC(_:) + 1240
5  Datadog                        0x104283608 iOSCarrierInfoProvider.current.getter + 129 (CarrierInfoProvider.swift:129)
6  Datadog                        0x1042c739c LogBuilder.createLogWith(level:message:error:date:attributes:tags:) + 304888 (<compiler-generated>:304888)
7  Datadog                        0x1042cf18c Logger.log(level:message:error:messageAttributes:) + 284 (Logger.swift:284)
8  Datadog                        0x1042a70b4 closure #1 in DataUploadWorker.scheduleNextUpload(after:) + 173072 (DataUploadWorker.swift:173072)
9  Datadog                        0x1042ce118 thunk for @escaping @callee_guaranteed () -> () + 332916 (<compiler-generated>:332916)
10 libdispatch.dylib              0x185c67298 _dispatch_client_callout + 16
11 libdispatch.dylib              0x185c0c028 _dispatch_continuation_pop$VARIANT$mp + 412
12 libdispatch.dylib              0x185c1cd64 _dispatch_source_invoke$VARIANT$mp + 1304
13 libdispatch.dylib              0x185c0f908 _dispatch_lane_serial_drain$VARIANT$mp + 300
14 libdispatch.dylib              0x185c10518 _dispatch_lane_invoke$VARIANT$mp + 420
15 libdispatch.dylib              0x185c19fac _dispatch_workloop_worker_thread + 712
[...]

Issue #619, opened 2021-10-01
Crash in CarrierInfoProvider.swift


matcartmill:
As a note, I do know that we're on v1.6.0 and the current appears to be v.1.7.0. We will be updating but given the infrequency of this crash it may be quite some time before we understand if the upgrade fixed the issue or not.

Thanks.

### The crash

Here is the stack trace (the crash happens on line 5 of the trace):

```
Crashed: com.datadoghq.ios-sdk-logging-read-write
0  libswiftCore.dylib             0x18ed94b70 swift_unknownObjectRetain + 20
1  libswiftFoundation.dylib       0x18e8b7f48 closure #2 in static Dictionary._forceBridgeFromObjectiveC(_:result:) + 736
2  libswiftFoundation.dylib       0x18e8b8000 partial apply for closure #2 in static Dictionary._forceBridgeFromObjectiveC(_:result:) + 36
3  libswiftFoundation.dylib       0x18e8bd32c _NativeDictionary.init(_unsafeUninitializedCapacity:allowingDuplicates:initializingWith:) + 340
4  libswiftFoundation.dylib       0x18e8b6ad0 static Dictionary._unconditionallyBridgeFromObjectiveC(_:) + 1368
5  Datadog                        0x103c948b8 iOSCarrierInfoProvider.current.getter + 132 (CarrierInfoProvider.swift:132)
6  Datadog                        0x103cd9298 LogBuilder.createLogWith(level:message:error:date:attributes:tags:) + 308728 (<compiler-generated>:308728)
7  Datadog                        0x103ce1120 Logger.log(level:message:error:messageAttributes:) + 284 (Logger.swift:284)
8  Datadog                        0x103cdcfb0 specialized Dictionary.filter(_:) + 324368 (<compiler-generated>:324368)
9  Datadog                        0x103cdbc80 Log.encode(to:) + 43 (LogSanitizer.swift:43)
10 Datadog                        0x103cdc04c protocol witness for Encodable.encode(to:) in conformance Log + 320428 (<compiler-generated>:320428)
11 libswiftCore.dylib             0x18ed1d010 dispatch thunk of Encodable.encode(to:) + 28
12 libswiftFoundation.dylib       0x18e9224fc __JSONEncoder.box_(_:) + 2200
13 libswiftFoundation.dylib       0x18e921824 JSONEncoder.encode<A>(_:) + 684
[...]

PR #627, opened 2021-10-07
RUMM-1691 Reduce utilization of `CTTelephonyNetworkInfo` to mitigate `CarrierInfoProvider` crash

ncreated:
### What and why?

📦 This PR updates the way we retrieve `CarrierInfo` information on iOS 12+. It's an educated guess to mitigate infrequent (2 impacted users) crashes reported in #623 and #619. Instead of pulling `CTTelephonyNetworkInfo` for each collected event to create `CarrierInfo`, now we only listen to `CTTelephonyNetworkInfo` changes to update `CarrierInfo` when necessary.

We use `CTTelephonyNetworkInfo` (from `CoreTelephony` framework) to read carrier information:

Screenshot 2021-10-07 at 12 46 30

A loop of investigation, research and debugging, lead me to following conclusions:
* C1. Reported crashes are very infrequent (one report on iOS 14.8 and one on iOS 14.6),
* C2. I observed is that `CTTelephonyNetworkInfo` only changes when swapping SIM card to another one (it does not change when removing SIM card - the last known info is still available). This is the only example listed in [Apple doc](https://developer.apple.com/documentation/coretelephony/cttelephonynetworkinfo).
* C3. We already had similar issues with utilization of Apple networking frameworks:
   * [recent crash on polling `ProcessInfo.isLowPowerModeEnabled`](https://github.com/DataDog/dd-sdk-ios/pull/613),
   * [old issue with pulling `NWPath.currentPath `](https://github.com/DataDog/dd-sdk-ios/pull/109).

Both issues in C3 were fixed by changing from pulling model, to subscription model (it was also a [recommendation I received](https://developer.apple.com/forums/thread/133112) for `NWPath` issue from Apple's engineer). Given that `CTTelephonyNetworkInfo` changes are very infrequent (C2) it makes no sense to pull this value and compute `Datadog.CarrierInfo` for each event we collect. Instead, we can listen to `CTTelephonyNetworkInfo` changes and update `Datadog.CarrierInfo` only when necessary.

I can't confirm if this fixes the crash, because I didn't manage to reproduce it even when stress testing our SDK with `~500logs/s` sent from iPhone X and swapping between different SIM cards during the test.

Screenshot 2021-10-06 at 18 27 04

### How?

Now, we use [`serviceSubscriberCellularProvidersDidUpdateNotifier`](https://developer.apple.com/documentation/coretelephony/cttelephonynetworkinfo/3024512-servicesubscribercellularprovide) available on iOS 12+ to update `CarrierInfo`. Initial value is read when SDK starts.

The logic for iOS 11 remains the same with the difference that now we also read initial value for `CarrierInfo` when starting the SDK. Although there is an API for listening to `CTTelephonyNetworkInfo` updates on iOS 11, it is deprecated (ref.: [`CTRadioAccessTechnologyDidChange `](https://developer.apple.com/documentation/foundation/nsnotification/name/1616908-ctradioaccesstechnologydidchange)) and because we have no reports for iOS 11, I decided to not change existing logic.

Last, I added simple utility to stress test Logging feature in Example app:

Screenshot 2021-10-07 at 12 42 58

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/System/CarrierInfoProvider.swift
+++ b/Sources/Datadog/Core/System/CarrierInfoProvider.swift
@@ -66,87 +66,144 @@ extension CarrierInfo.RadioAccessTechnology {
     }
 }
 
-/// An interface for the target-specific carrier info provider.
-internal protocol WrappedCarrierInfoProvider {
-    var current: CarrierInfo? { get }
-}
-
+/// Platform-agnostic carrier info provider. It wraps the platform-specific provider inside.
 internal class CarrierInfoProvider: CarrierInfoProviderType {
     /// The `CarrierInfo` provider for the current platform.
-    private let wrappedProvider: WrappedCarrierInfoProvider
-    /// Publisher for notifying observers on `CarrierInfo` change.
-    private let publisher: ValuePublisher<CarrierInfo?>
+    private let wrappedProvider: CarrierInfoProviderType
 
     convenience init() {
         #if targetEnvironment(macCatalyst)
-            self.init(
-                wrappedProvider: MacCatalystCarrierInfoProvider()
-            )
+        self.init(
+            wrappedProvider: MacCatalystCarrierInfoProvider()
+        )
         #else
-            self.init(
-                wrappedProvider: iOSCarrierInfoProvider(
-                    networkInfo: CTTelephonyNetworkInfo()
-                )
-            )
+        if #available(iOS 12.0, *) {
+            self.init(wrappedProvider: iOS12CarrierInfoProvider(networkInfo: CTTelephonyNetworkInfo()))
+        } else {
+            self.init(wrappedProvider: iOS11CarrierInfoProvider(networkInfo: CTTelephonyNetworkInfo()))
+        }
         #endif
     }
 
-    init(wrappedProvider: WrappedCarrierInfoProvider) {
+    init(wrappedProvider: CarrierInfoProviderType) {
         self.wrappedProvider = wrappedProvider
-        self.publisher = ValuePublisher(initialValue: nil)
     }
 
     var current: CarrierInfo? {
-        let nextValue = wrappedProvider.current
-        // `CarrierInfo` subscribers are notified as a side-effect of retrieving the
-        // current `CarrierInfo` value.
-        publisher.publishAsync(nextValue)
-        return nextValue
+        wrappedProvider.current
     }
 
     func subscribe<Observer: CarrierInfoObserver>(_ subscriber: Observer) where Observer.ObservedValue == CarrierInfo? {
-        publisher.subscribe(subscriber)
+        wrappedProvider.subscribe(subscriber)
     }
 }
 
 #if targetEnvironment(macCatalyst)
 
-internal struct MacCatalystCarrierInfoProvider: WrappedCarrierInfoProvider {
-    /// Carrier info is not supported on macCatalyst
+/// Dummy provider for Mac Catalyst which doesn't support carrier info.
+internal struct MacCatalystCarrierInfoProvider: CarrierInfoProviderType {
     var current: CarrierInfo? { return nil }
+    func subscribe<Observer>(_ subscriber: Observer) where Observer: ValueObserver, Observer.ObservedValue == CarrierInfo? {}
 }
 
 #else
 
-internal struct iOSCarrierInfoProvider: WrappedCarrierInfoProvider {
-    let networkInfo: CTTelephonyNetworkInfo
+/// Carrier info provider for iOS 12 and above.
+/// It reads `CarrierInfo?` from `CTTelephonyNetworkInfo` only when `CTCarrier` has changed (e.g. when the SIM card was swapped).
+@available(iOS 12, *)
+internal class iOS12CarrierInfoProvider: CarrierInfoProviderType {
+    private let networkInfo: CTTelephonyNetworkInfo
+    /// Publisher for notifying observers on `CarrierInfo` change.
+    private let publisher: ValuePublisher<CarrierInfo?>
+
+    init(networkInfo: CTTelephonyNetworkInfo) {
+        self.networkInfo = networkInfo
+        self.publisher = ValuePublisher(
+            initialValue: iOS12CarrierInfoProvider.readCarrierInfo(
+                from: networkInfo,
+                cellularProviderKey: networkInfo.serviceCurrentRadioAccessTechnology?.keys.first
+            )
+        )
+
+        // The `serviceSubscriberCellularProvidersDidUpdateNotifier` block object executes on the default priority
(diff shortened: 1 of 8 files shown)


Sources/Datadog/Core/System/CarrierInfoProvider.swift
```diff
@@ -66,88 +66,145 @@ extension CarrierInfo.RadioAccessTechnology {
...
+
+        // The `serviceSubscriberCellularProvidersDidUpdateNotifier` block object executes on the default priority
+        // global dispatch queue when the user’s cellular provider information changes.
+        // This occurs, for example, if a user swaps the device’s SIM card with one from another provider, while the app is running.
+        // ref.: https://developer.apple.com/documentation/coretelephony/cttelephonynetworkinfo/3024512-servicesubscribercellularprovide
+        networkInfo.serviceSubscriberCellularProvidersDidUpdateNotifier = { [weak self] cellularProviderKey in
+            let carrierInfo = iOS12CarrierInfoProvider.readCarrierInfo(
+                from: networkInfo,
+                cellularProviderKey: networkInfo.serviceCurrentRadioAccessTechnology?.keys.first
+            )
```
ncreated, line 136: I've been debugging what happens to `networkInfo: CTTelephonyNetworkInfo` when removing and changing SIM card. The object itself is not deallocated, only its values are updated:
Screenshot 2021-10-06 at 18 43 46

What changes though is the `CTCarrier` retrieved in `readCarrierInfo(from:cellularProviderKey:)`:
```swift
let carrier: CTCarrier = networkInfo.serviceSubscriberCellularProviders?[cellularProviderKey]
```
Screenshot 2021-10-07 at 13 40 44


The `swift_unknownObjectRetain` in reported crash:
```
0  libswiftCore.dylib             0x18ed94b70 swift_unknownObjectRetain + 20
1  libswiftFoundation.dylib       0x18e8b7f48 closure #2 in static Dictionary._forceBridgeFromObjectiveC(_:result:) + 736
2  libswiftFoundation.dylib       0x18e8b8000 partial apply for closure #2 in static Dictionary._forceBridgeFromObjectiveC(_:result:) + 36
3  libswiftFoundation.dylib       0x18e8bd32c _NativeDictionary.init(_unsafeUninitializedCapacity:allowingDuplicates:initializingWith:) + 340
4  libswiftFoundation.dylib       0x18e8b6ad0 static Dictionary._unconditionallyBridgeFromObjectiveC(_:) + 1368
5  Datadog                        0x103c948b8 iOSCarrierInfoProvider.current.getter + 132 (CarrierInfoProvider.swift:132)
```
seems to correspond to extending the `CTCarrier's` lifespan in our code after it was released in `CoreTelephony`. This is one of the scenarios I've been trying to reproduce (extending it explicitly), but it didn't lead to any crash.

maxep: This creates a retain cycle of `networkInfo`.

maxep: Why not using the `cellularProviderKey` from the `serviceSubscriberCellularProvidersDidUpdateNotifier` callback?

Sources/Datadog/Core/System/CarrierInfoProvider.swift
```diff
@@ -66,88 +66,145 @@ extension CarrierInfo.RadioAccessTechnology {
...
+
+    init(networkInfo: CTTelephonyNetworkInfo) {
+        self.networkInfo = networkInfo
+        self.publisher = ValuePublisher(
+            initialValue: iOS11CarrierInfoProvider.readCarrierInfo(from: networkInfo)
         )
     }
+
+    private static func readCarrierInfo(from networkInfo: CTTelephonyNetworkInfo) -> CarrierInfo? {
+        if let radioTechnology = networkInfo.currentRadioAccessTechnology,
```
maxep, line 182: /nit using a `guard` would make it more readable IMHO 👓

Sources/Datadog/Core/System/CarrierInfoProvider.swift
ncreated: > I see a retain cycle of networkInfo here.

Oh yeah, my bad. It doesn't make any bad side effects, as `networkInfo` is a singleton (once obtained from `CoreTelephony` it never changes). Anyway, will fix that, good catch 👍 .

> Also, why not using the `cellularProviderKey` returned by the `serviceSubscriberCellularProvidersDidUpdateNotifier` closure?

My bad, again, good catch 👍. In all my tests, it has the same value as `serviceCurrentRadioAccessTechnology's` `.keys.first`, but the entire point for having this callback is to indeed use `cellularProviderKey` instead. Will fix.

ncreated: Fixed 👍. I did one more test, it works.

Screenshot 2021-10-08 at 16 05 05

Sources/Datadog/Core/System/CarrierInfoProvider.swift
```diff
@@ -66,88 +66,147 @@ extension CarrierInfo.RadioAccessTechnology {
...
+           let carrier = networkInfo.serviceSubscriberCellularProviders?[cellularProviderKey] else {
+            return nil // the service is not registered on any network
         }
+        return CarrierInfo(
+            carrierName: carrier.carrierName,
+            carrierISOCountryCode: carrier.isoCountryCode,
+            carrierAllowsVOIP: carrier.allowsVOIP,
+            radioAccessTechnology: .init(ctRadioAccessTechnologyConstant: radioTechnology)
+        )
+    }
```
maxep, line 159: /nit I think you can simplify that by using an instance method to read the carrier, something like:

```swift
init(networkInfo: CTTelephonyNetworkInfo) {
    ...
    networkInfo.serviceSubscriberCellularProvidersDidUpdateNotifier = { [weak self] cellularProviderKey in
        let carrierInfo = self?.readCarrierInfo(provider: cellularProviderKey)
        self?.publisher.publishAsync(carrierInfo)
    }
}

private func readCarrierInfo(provider key: String) -> CarrierInfo? {
    guard 
       let radioTechnology = networkInfo.serviceCurrentRadioAccessTechnology?[key],
       let carrier = networkInfo.serviceSubscriberCellularProviders?[key] 
    else {
       return nil // the service is not registered on any network
    }
    return CarrierInfo(
          carrierName: carrier.carrierName,
          carrierISOCountryCode: carrier.isoCountryCode,
          carrierAllowsVOIP: carrier.allowsVOIP,
          radioAccessTechnology: .init(ctRadioAccessTechnologyConstant: radioTechnology)
    )
}

maxep (approved):
LGTM 👌 
I've just a minor suggestion

buranmert (approved):

PR #628, opened 2021-10-08
Add "Debug webview" screen to Example app

ncreated:
### What and why?

🐶 This PR adds debug screen to the Example app, where we can test webviews. Native integration is yet to be added, but this will unlock other teams on testing `dd-sdk-ios` and webviews.

Screenshot 2021-10-08 at 13 04 43Screenshot 2021-10-08 at 12 56 02Screenshot 2021-10-08 at 12 55 59

### How?

Added menu option and basic "webview launch" screen with minimal setup:
* starting a native RUM Session before opening webview (this can be later used for testing session propagation),
* starting a webview with custom URL (Datadog website by default).

### Review checklist

### DIFF
--- a/Datadog/Example/Debugging/DebugWebviewViewController.swift
+++ b/Datadog/Example/Debugging/DebugWebviewViewController.swift
@@ -46,7 +46,10 @@ class DebugWebviewViewController: UIViewController {
     @IBOutlet weak var webviewURLTextField: UITextField!
 
     private var webviewURL: String {
-        webviewURLTextField.text!.isEmpty ? "https://www.datadoghq.com" : webviewURLTextField.text!
+        guard let text = webviewURLTextField.text, !text.isEmpty else {
+            return  "https://www.datadoghq.com"
+        }
+        return text
     }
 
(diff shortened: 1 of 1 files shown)


Datadog/Example/Debugging/DebugWebviewViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+        Global.rum.startView(viewController: self)
+    }
+
+    // MARK: - Starting webview
+
+    @IBOutlet weak var webviewURLTextField: UITextField!
+
+    private var webviewURL: String {
+        webviewURLTextField.text!.isEmpty ? "https://www.datadoghq.com" : webviewURLTextField.text!
+    }
```
maxep, line 50: /nit to avoid any crash

```suggestion
    private var webviewURL: String {
    guard
        let text = webviewURLTextField.text,
        !text.isEmpty 
    else { return  "https://www.datadoghq.com" }
    
    return text
    }
```

maxep (approved):

PR #629, opened 2021-10-08
RUMM-1394 Fix flaky tests for iOS 11.x

ncreated:
### What and why?

📦 This PR fixes issues with tests flakiness on iOS 11.x, nailed down after recent update to their verbosity in https://github.com/DataDog/dd-sdk-ios/pull/620. This should make our nightly CI jobs more stable for low iOS versions.

### How?

There were 18 nightly failures in last week:

Screenshot 2021-10-08 at 14 58 15

14 of which correspond to 4 tests failures in `URLSessionSwizzlerTests` and `DDURLSessionDelegateTests` with a common root cause. In all these tests, we send several requests through `URLSession` and we assert certain aspects of their interception. One thing that we unconsciously assumed was that all requests will complete in the same order as they were started. Nothing in the test code (e.g. asynchronous expectation) is asserting the order, so this assumption was causing flakiness due to missorder, e.g.:
```
Actual task:
- taskIdentifier: '1'
- taskDescription: 'taskWithURLRequestAndCompletion'

Expected task:
- taskIdentifier: '2'
- taskDescription: 'taskWithURLAndCompletion'
```

Another 4 failures correspond to a single test in `DataUploadWorkerTests`. I didn't manage to reproduce it (although I repeated this test 1K times with Xcode 13), so in this PR I add additional verbosity to collect more data and solve it in another PR.

### Review checklist


Tests/DatadogTests/Datadog/Core/Upload/DataUploadWorkerTests.swift
```diff
@@ -252,7 +252,7 @@ class DataUploadWorkerTests: XCTestCase {
             fileReader: reader,
             dataUploader: mockDataUploader,
             uploadConditions: .alwaysUpload(),
-            delay: DataUploadDelay(performance: UploadPerformanceMock.veryQuick),
+            delay: DataUploadDelay(performance: UploadPerformanceMock.veryQuickInitialUpload),
```
ncreated, line 255: This value fits better what we do in this test. We cancel upload worker after the first upload. To eliminate eventual race condition, `.veryQuickInitialUpload` should be used here in the first place (it does 1st upload very quick, and schedules next ones in a distant future).

PR #631, opened 2021-10-11
RUMM-1090 Make `LogEvent` and `UserInfo` public

maxep:
### What and why?

Make `LogEvent` and `UserInfo` public.

### Review checklist

### DIFF
--- a/api-surface-swift
+++ b/api-surface-swift
@@ -227,9 +227,14 @@ public struct LogEvent: Encodable
   public let name: String?
   public let email: String?
   public var extraInfo: [String: Encodable]
+ public struct Error
+  public let kind: String?
+  public let message: String?
+  public let stack: String?
  public let date: Date
  public let status: Status
  public var message: String
+ public let error: Error?
  public let serviceName: String
  public let environment: String
  public let loggerName: String


Sources/Datadog/Core/Attributes/UserInfoProvider.swift
```diff
@@ -35,12 +35,24 @@ internal class UserInfoProvider {
...
 
-/// Information about the user.
-internal struct UserInfo {
-    let id: String?
-    let name: String?
-    let email: String?
-    let extraInfo: [AttributeKey: AttributeValue]
-
-    internal static var empty: UserInfo { UserInfo(id: nil, name: nil, email: nil, extraInfo: [:]) }
+public struct UserInfo {
```
ncreated, line 38: This `UserInfo` is part of `Datadog/Core`. Both Tracing and RUM use their own representation of user info:
* [`SpanEvent.UserInfo` for Tracing](https://github.com/DataDog/dd-sdk-ios/blob/181359bd655b894041702332c26fa05b2d451990/Sources/Datadog/Tracing/Span/SpanEventEncoder.swift#L68) - it has slightly different signature (`[String: String]` for extra info),
* [`RUMUser` for RUM](https://github.com/DataDog/dd-sdk-ios/blob/181359bd655b894041702332c26fa05b2d451990/Sources/Datadog/RUM/DataModels/RUMDataModels.swift#L1557) -it is auto-generated model from RUM schema.

I'm in between:
* 1/ exposing this `Core.UserInfo` only for `LogEvent`,
* 2/ creating `LogEvent.UserInfo` and having mapper.

Going with the 2nd approach would make the architecture aligned with other products for `UserInfo`, but would require duplicating its definition. I personally prefer small duplication over bad abstraction and that might be the case given that we're planning to modularize our SDK soon into separate libraries (Tracing / RUM / Logging).

On the other hand, same thing applies to `NetworkConnectionInfo` and `CarrierInfo` which are both `public` and used from `SpanEvent` and (now) `LogEvent`, but not `RUMEvent` (RUM maps both into a single [`RUMConnectivity`](https://github.com/DataDog/dd-sdk-ios/blob/181359bd655b894041702332c26fa05b2d451990/Sources/Datadog/RUM/DataModels/RUMDataModels.swift#L1469) value). That would need to be tackled somehow later in mentioned modularization.

What's your take on that?

Sources/Datadog/Logging/Log/LogEventEncoder.swift
```diff
@@ -18,24 +19,45 @@ internal struct LogEvent: Encodable {
...
-
-    func encode(to encoder: Encoder) throws {
+    /// The log's timestamp
+    public let date: Date
+    /// The log status
+    public let status: Status
+    /// The log message
+    public var message: String
+    /// The associated log error
+    internal let error: DDError?
```
ncreated, line 36: There are two motivations behind the scrubbing API, uniformly met in Tracing and RUM scrubbing:
* 1/ the user can see all user-controlled values that will be sent to Datadog,
* 2/ the user can redact some of these values (we decide what is mutable and what is not).

Here, this internal `error` does clearly hide some information (1), because [it is serialized](https://github.com/DataDog/dd-sdk-ios/blob/181359bd655b894041702332c26fa05b2d451990/Sources/Datadog/Logging/Log/LogEncoder.swift#L109-L114) into 3 distinct fields:
```swift
// Encode log.error properties
if let someError = log.error {
    try container.encode(someError.type, forKey: .errorKind)
    try container.encode(someError.message, forKey: .errorMessage)
    try container.encode(someError.stack, forKey: .errorStack)
}
```
How about defining these 3 values as explicit properties on `LogEvent`:
```swift
public let error: LogEvent.Error?

public struct LogEvent.Error {
   public let kind: String?
   public let message: String?
   public let stack: String?
}
```
and extracting their values from `DDError` before creating the `LogEvent` (so in `LogEventBuilder`)? Then, we can decide (2) if any of these values should be mutable `var` - in RUM we allow redaction of [`error.message`](https://github.com/DataDog/dd-sdk-ios/blob/181359bd655b894041702332c26fa05b2d451990/Sources/Datadog/RUM/DataModels/RUMDataModels.swift#L1115) and [`error.stack`](https://github.com/DataDog/dd-sdk-ios/blob/181359bd655b894041702332c26fa05b2d451990/Sources/Datadog/RUM/DataModels/RUMDataModels.swift#L1124) but not [`error.type`](https://github.com/DataDog/dd-sdk-ios/blob/181359bd655b894041702332c26fa05b2d451990/Sources/Datadog/RUM/DataModels/RUMDataModels.swift#L1127). WDYT?

Sources/Datadog/Core/Attributes/UserInfoProvider.swift
maxep: I had the exact same thought process, I started by creating an inner `LogEvent.UserInfo` but then I noticed both `NetworkConnectionInfo` and `CarrierInfo` were shared.

For modularization prospect, using contained model definition per feature and use abstractions for shared data would be a good approach IMHO. Using an inner UserInfo would be a step in that direction and facilitate the transition. I will do the change!

PR #633, opened 2021-10-12
RUMM-1090 Log Event Scrubbing

maxep:
### What and why?

Allow logs data scrubbing.

### How?

Provide a public interface to mutate ` LogEvent` before serialisation.

### Review checklist

### DIFF
--- a/Sources/Datadog/DatadogConfiguration.swift
+++ b/Sources/Datadog/DatadogConfiguration.swift
@@ -406,8 +406,6 @@ extension Datadog {
             /// - Parameter mapper: the closure taking `LogEvent` as input and expecting `LogEvent` as output.
             /// The implementation should obtain a mutable version of the `LogEvent`, modify it and return it. Returning `nil` will result
             /// with dropping the Log event entirely, so it won't be send to Datadog.
-            ///
-            /// Use the `trackURLSession(firstPartyHosts:)` API to configure tracing only the hosts that you are interested in.
             public func setLogEventMapper(_ mapper: @escaping (LogEvent) -> LogEvent) -> Builder {
                 configuration.logEventMapper = mapper
                 return self


Sources/Datadog/DatadogConfiguration.swift
```diff
@@ -401,6 +402,18 @@ extension Datadog {
                 return self
             }
 
+            /// Sets the custom mapper for `logEvent`. This can be used to modify logs before they are send to Datadog.
+            /// - Parameter mapper: the closure taking `logEvent` as input and expecting `logEvent` as output.
+            /// The implementation should obtain a mutable version of the `logEvent`, modify it and return it.
```
ncreated, line 407:
```suggestion
            /// Sets the custom mapper for `LogEvent`. This can be used to modify logs before they are send to Datadog.
            /// - Parameter mapper: the closure taking `LogEvent` as input and expecting `LogEvent` as output.
            /// The implementation should obtain a mutable version of the `LogEvent`, modify it and return it.
```

Sources/Datadog/DatadogConfiguration.swift
```diff
@@ -401,6 +402,18 @@ extension Datadog {
...
+            ///
+            /// **NOTE** The mapper intentionally prevents from returning a `nil` to drop the `logEvent` entirely, this ensures that all logs are sent to Datadog.
```
ncreated, line 409: We should check with Android SDK (cc @xgouchet ) but I think we don't need this limitation for `LogEvents` - any `LogEvent` can be safely dropped. 

This was only required for `RUMViewEvents` and `SpanEvents` which have hierarchical structure and dropping a particular event could break it (e.g. dropping a root `SpanEvent` would leave its child spans inconsistent; similar for dropping the `RUMViewEvent` which is referenced by some action / resource / error events).

That would mean, the mapper signature should be:
```swift
(LogEvent) -> LogEvent?
```
and we need to properly handle it downstream.

ncreated (changes requested):
The implementation looks 👌 with a note that IMO we should change the signature of mapper to allow `nil` for dropping `LogEvents`. Awesome to have integration tests updated 🙌 !

Tests/DatadogTests/Datadog/Logging/Log/LogBuilderTests.swift
```diff
@@ -74,10 +74,52 @@ class LogBuilderTests: XCTestCase {
...
         waitForExpectations(timeout: 1, handler: nil)
     }
+
+    func testGivenBuilderWithEventMapper_whenEventIsModified_itBuildsModifiedEvent() throws {
+        let builder: LogEventBuilder = .mockWith(
+            logEventMapper: { log in
+                var mutableSpan = log
+                mutableSpan.message = "modified log message"
+                mutableSpan.tags = .mockRandom()
+                return mutableSpan
```
ncreated, line 90:
```suggestion
                var mutableLog = log
                mutableLog.message = "modified log message"
                mutableLog.tags = .mockRandom()
                return mutableLog
```

Sources/Datadog/DatadogConfiguration.swift
```diff
@@ -401,6 +402,17 @@ extension Datadog {
                 return self
             }
 
+            /// Sets the custom mapper for `LogEvent`. This can be used to modify logs before they are send to Datadog.
+            /// - Parameter mapper: the closure taking `LogEvent` as input and expecting `LogEvent` as output.
+            /// The implementation should obtain a mutable version of the `LogEvent`, modify it and return it. Returning `nil` will result
+            /// with dropping the Log event entirely, so it won't be send to Datadog.
+            ///
+            /// Use the `trackURLSession(firstPartyHosts:)` API to configure tracing only the hosts that you are interested in.
```
buranmert, line 410:
```
Use the `trackURLSession(firstPartyHosts:)` API to configure tracing only the hosts that you are interested in.
```
is this comment intentional here?

PR #641, opened 2021-10-15
RUMM-1596 Fix issue with truncated stack traces not being displayed in Crash Reporting

ncreated:
### What and why?

🐞 This PR fixes a bug with the crash reports containing a very long stack traces not being rendered properly by our frontend. This situation may correspond to crashes due to segmentation fault on infinite recursive call. In such case, we truncate the stack trace SDK-side by inserting `...` mark. This is to ensure it doesn't exceed our ingestion limits.

### How?

Problem was that our SCI BE was failing to parse the stack trace due to `...` truncation mark. In result, RUM Error was displayed, but it had no associated stack:

Screenshot 2021-10-15 at 12 50 20

The fix was discussed with ET team and they will (`ET-1602`) insert this mark on the frontend-side when displaying the error. A counterpart work on our side is to remove the `...` mark from the stack, which is done in this PR.

By removing truncation mark, the stack trace is rendered properly:
Screenshot 2021-10-15 at 12 54 35

As an enhancement, I added "Infinite recursive call" button to our crash debugging tool:

Screenshot 2021-10-15 at 12 13 20

### Review checklist

### DIFF
--- a/Sources/DatadogCrashReporting/PLCrashReporterIntegration/DDCrashReportExporter.swift
+++ b/Sources/DatadogCrashReporting/PLCrashReporterIntegration/DDCrashReportExporter.swift
@@ -20,8 +20,6 @@ import Datadog
 internal struct DDCrashReportExporter {
     private let unknown = "<unknown>"
     private let unavailable = "???"
-    /// Truncation mark printed in a stack trace in the place of stack frames that were removed.
-    private let stackTraceTruncationMark = "..."
 
     /// Different signals and their descriptions available in OS.
     private let knownSignalDescriptionByName = [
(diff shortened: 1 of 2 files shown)


Datadog/Example/Scenarios/CrashReporting/CrashReporting/CrashReportingViewController.swift
```diff
@@ -82,4 +84,22 @@ internal class CrashReportingViewController: UIViewController {
...
+    func recursiveCall(parameter: Int) -> Int {
+        if parameter > 1_000_000 {
+            heap.append("abcdefghijkl")
+            return recursiveCall(parameter: parameter - 1)
+        } else if parameter <= 1_000_000 {
+            heap.append("abcdefghijkl")
+            return recursiveCall(parameter: parameter + 1)
+        } else {
+            return 0 // not reachable, but mutes compiler warning
+        }
```
buranmert, line 103: trivial:
```suggestion
        } else {
            heap.append("abcdefghijkl")
            return recursiveCall(parameter: parameter + 1)
        }
```

ncreated: Indeed, my mind was too much into recursion and I haven't noticed that, thx!

PR #642, opened 2021-10-18
RUMM-1615 Allow launch time earlier than `UIApplicationDidBecomeActiveNotification`

maxep:
### What and why?

This PR is a proposal on allowing sending the `application_start` action with a `loading_time` value **before** the application has become active.

The SDK measures the Application launch time from the time the application process starts until when it is responsive and accepts touch events. This is performed by responding to `UIApplicationDidBecomeActiveNotification`.

In case the RUM monitor receives the first `startView` command before the application has become active, the `application_start` action event is recorded **without** a `loading_time`. This will be the case when invoking `startView` manually from the `UIViewController.viewWillAppear` or form `SwiftUI.View.onAppear` for instance.

In this proposal, we make sure to always return a Launch Time when requested, by either returning the time to become active (as before) or the time to receive the first `startView` command.

### How?

The `__dd_private_AppLaunchTime` function now returns the time from start until 'now' if the `UIApplicationDidBecomeActiveNotification` has not yet been received. 
To reflect this change, the `LaunchTimeProviderType` now expect a **non-optional** `launchTime`.

### Review checklist


Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -311,7 +311,7 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
                 crash: nil,
                 error: nil,
                 id: dependencies.rumUUIDGenerator.generateUnique().toRUMDataFormat,
-                loadingTime: dependencies.launchTimeProvider.launchTime?.toInt64Nanoseconds,
+                loadingTime: dependencies.launchTimeProvider.launchTime.toInt64Nanoseconds,
```
ncreated, line 314: There are two things I'm concerned about:
* The new "lazy" aspect of `.launchTime` API makes it slightly dangerous. It's easy to imagine introducing a bug by calling it somewhere earlier in the stack which will skew the measurement. Seems hard to cover this in tests.
* Using `launchTimeProvider.launchTime` to stop measurement in `RUMViewScope` skews the real measurement by the time of [`queue.async {}` used to run RUM "start view" command](https://github.com/DataDog/dd-sdk-ios/blob/a88103e71ebd79262af6723c095eb0dd1e0d83f7/Sources/Datadog/RUMMonitor.swift#L605).

What's our take on the above? Is there any alternative we can consider 🤔 💭?

ncreated: > Another approach would be to respond to the following notifications (...)

Or we could add our own notification, sent from `RUMMonitor.startView()`, but this sounds like adding cruft...

> The `launchTime` property is not lazy but computed (...). It's applying a fallback 

I totally missed that `.launchTime` is computed now. This mitigates the first point on my list 👍.

Maybe the `queue.async {}` impact is not that significant as it will be uniformly added to all measurements for the same scenario. @maxep @buranmert ?

PR #644, opened 2021-10-19
RUMM-640 Add API for clearing out all data

ncreated:
### What and why?

📦 This PR adds `Datadog.clearAllData()` API.
```swift
/// Clears all data that has not already been sent to Datadog servers.
public static func clearAllData()
```
It deletes all authorised (`.granted`) and unauthorised (`.pending`) data (not yet uploaded) buffered on device. This includes data for Logging, Tracing, RUM and Internal Monitoring features.

### How?

Because all our features use the same abstraction, I simply added `DataOrchestrator` type to `FeatureStorage`. Unlike `FilesOrchestrator`, `DataOrchestrator` manages data in multiple folders (we use it for managing authorised and unauthorised directories). The public API calls `DataOrchestrator.deleteAllFiles()` for each active feature.

### Review checklist

### DIFF
--- a/api-surface-objc
+++ b/api-surface-objc
@@ -15,10 +15,12 @@ public class DDDatadog: NSObject
  public static func verbosityLevel() -> DDSDKVerbosityLevel
  public static func setUserInfo(id: String? = nil, name: String? = nil, email: String? = nil, extraInfo: [String: Any] = [:])
  public static func setTrackingConsent(consent: DDTrackingConsent)
+ public static func clearAllData()
  public static func flushAndDeinitialize()
 public class DDEndpoint: NSObject
  public static func us1() -> DDEndpoint
  public static func us3() -> DDEndpoint
+ public static func us5() -> DDEndpoint
  public static func eu1() -> DDEndpoint
  public static func us1_fed() -> DDEndpoint
  public static func eu() -> DDEndpoint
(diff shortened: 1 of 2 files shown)


Sources/Datadog/Core/Persistence/DataOrchestrator.swift
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+#if DD_SDK_COMPILED_FOR_TESTING
+    func markAllFilesAsReadable() {
+        queue.sync {
+            authorizedFilesOrchestrator.ignoreFilesAgeWhenReading = true
+            unauthorizedFilesOrchestrator.ignoreFilesAgeWhenReading = true
+        }
+    }
+#endif
```
ncreated, line 36: Previously, this was part of `FileReader` as it had the access to `filesOrchestrator` (not accessible anywhere else). As now, `DataOrchestrator` manages files orchestrators, it seems better to move this logic here. This allowed cleaning up `FileReader` so it no longer exposes `filesOrchestrator` internally.

Sources/Datadog/Datadog.swift
```diff
@@ -134,6 +134,11 @@ public class Datadog {
         instance?.consentProvider.changeConsent(to: trackingConsent)
     }
 
+    /// Clears all data that has not already been sent to Datadog servers.
```
buranmert, line 137: maybe it might worth mentioning that this method is async and the files are not deleted once it returns.?

buranmert (approved):

Sources/Datadog/Datadog.swift
ncreated: Good question 👌. Actually, initially  I made this API `sync` to guarantee "no data" after it returns (by sacrificing the performance ofc), but then I realised that it makes no difference from the user standpoint. The SDK synchronizes all data collection internally so any event collected before `clearAllData()` will be deleted and any event collected after will be stored. That said, adding the information on `async` execution would be purely informative and could be considered a breaking change if we need to use `sync` for some reason in the future. WDYT @buranmert @maxep ?

maxep: Good question indeed 🤔  Hypothetically, if I do:
```
log(something) // <- will be deleted
clearAllData()
log(something) // <- will be stored
```
Because we are using serial `readWriteQueue` queues in each feature, am I correct?
If so, I don't think we have to specify it's async, the `clearAllData` will be doing what it's expected to do in a optimised manner.

PR #645, opened 2021-10-19
RUMM-1615 SwiftUI View Instrumentation

maxep:
### What and why?

Add `SwiftUI.View` instrumentation by providing a `SwiftUI.View` extension to monitor a view with RUM.

```swift
import SwiftUI
import Datadog

struct FooView: View {

    var body: some View {
        FooContent {
            ...
        }
        .trackRUMView(name: "Foo")
    }
}
```

#### UIKit Default Predicate

`DefaultUIKitRUMViewsPredicate` now exclude any `UIViewController` defined in `SwiftUI` framework. It will exclude the public [`UIHostingController`](https://developer.apple.com/documentation/swiftui/uihostingcontroller) but also any private container that `SwiftUI` creates under the hood.

#### SwiftUI View Modifier

A `SwiftUI.View` extension provides a simple interface to start monitoring the view. The extension applies a [`ViewModifier`](https://developer.apple.com/documentation/swiftui/viewmodifier) that responds to the [`.onAppear`](https://developer.apple.com/documentation/swiftui/view/onappear(perform:)) and [`.onDisappear`](https://developer.apple.com/documentation/swiftui/view/ondisappear(perform:)) to respectively start and stop monitoring the view with the internal `RUMInstrumentation` instance.

A [`ViewModifier`](https://developer.apple.com/documentation/swiftui/viewmodifier) returns a different version of the view which allow applying other modifiers or other handlers for `.onAppear` or `.onDisappear` without interfering with the RUM instrumentation.

#### SwiftUI RUM View Handler

The introduced `SwiftUIRUMViewsHandler` will keep track of the appeared views published by the `RUMViewModifier` to send Start and Stop RUM view commands in a consistent way. This handler uses a stack of appeared views:
- When a view appears, it stops the previous view and start the new one.
- When a view disappears, it removes it from the stack. If the view is the last one added, it will send a Stop rum command for that view and look for any remaining view in the stack to re-start.

This behaviour allow to re-start views when cancelling the back interaction in a navigation stack, and when dismissing a page sheet in modal navigation.

### Review checklist


Sources/Datadog/RUM/AutoInstrumentation/Views/SwiftUIViewModifier.swift
```diff
@@ -0,0 +1,52 @@
...
+    /// - Parameters:
+    ///   - key: a `String` value identifying this View.
+    ///   - name: the View name used for RUM Explorer. If not provided, the `key` name will be used.
+    ///   - attributes: custom attributes to attach to the View.
+    /// - Returns: This view after applying a `ViewModifier` for monitoring the view.
+    func monitorView(
+        key: String,
+        name: String? = nil,
+        attributes: [AttributeKey: AttributeValue] = [:]
+    ) -> some View {
```
ncreated, line 47: I'm not sure on this signature. TO me, seeing `.monitorView(key: "SwiftUI View")` is quite puzzling at the first glance. It doesn't mention Datadog or RUM so it's hard to figure out where does this modifier come from. Also, with "monitor" being both a verb and noun this is a bit misleading. I know it was discussed in the RFC, but I intentionally skipped commenting there as it's an implementation detail to me.

Maybe `monitorRUMView(...)` or `trackRUMView(...)` (a counterpart to [`trackUIKitRUMViews(...)`](https://github.com/DataDog/dd-sdk-ios/blob/0d1fbd1cf2819576ba4a797d5184ede94d8e6555/Sources/Datadog/DatadogConfiguration.swift#L564)) could work better? We uniformly add `*RUM*` phrase to our APIs, IMO this could be preserved here as well.

Further, `key` parameter is just a technical reference for matching "start" and "stop" commands added for React-Native (where we can't identify RUM Views by `UIViewController` object reference). I don't think this makes sense to expose it in this manual API, as that would be forcing users to always give it unique value to avoid clashing views with common `name`. IMO this:
```swift
name: String,
```
can be enough. The name would be mandatory (makes sense for manual API) and we can manage `key` complexity inside `RUMViewModifier` implementation, e.g. by using `UUID()`:
```diff
- let key: String
+ private let key: String = UUID().uuidString
```

`String` is already a [supported `RUMViewIdentifiable` type](https://github.com/DataDog/dd-sdk-ios/blob/0d1fbd1cf2819576ba4a797d5184ede94d8e6555/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewIdentity.swift#L49). To my knowledge, there is no Swift UI specific we could leverage for additional conformance, but if you have any idea, that could be an option as well.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
```diff
@@ -5,7 +5,6 @@
  */
 
 import Foundation
-import class UIKit.UIViewController
```
ncreated, line 8: Good catch 👌

Tests/DatadogIntegrationTests/UITestsHelpers.swift
```diff
@@ -78,3 +78,14 @@ extension String {
...
 }
+
+extension XCUIElement {
+    @discardableResult
+    func safeTap(within timeout: TimeInterval = 0) -> Bool {
+        if waitForExistence(timeout: timeout) && isHittable {
+            tap()
+            return true
+        }
+        return false
```
ncreated, line 89: I don't see how can we use these return values in our UITests - as we're not using them now, let's keep it simple:
```swift
func safeTap(within timeout: TimeInterval = 0)
```
💭?

Tests/DatadogTests/Datadog/Utils/SwiftUIExtensionsTests.swift
```diff
@@ -0,0 +1,43 @@
...
+@available(iOS 13, *)
+class CustomHostingController: UIHostingController<AnyView> {}
+
+@available(iOS 13, *)
+final class TestView: View {
+    var body = EmptyView()
+}
+
+class SwiftUIExtensionsTests: XCTestCase {
+    func testBundleIsUIKit() {
```
ncreated, line 22:
```suggestion
    func testBundleIsSwiftUI() {
```

Sources/Datadog/RUM/AutoInstrumentation/Views/SwiftUIViewModifier.swift
ncreated: I see 👍. I'm sceptical to changing the `RUMMonitor` APIs signature, especially adding the `path` value back (it was removed in https://github.com/DataDog/dd-sdk-ios/pull/418). The `path` attribute is here for historical reason and was made for compatibility with the part of RUM schema coming form Browser SDK (it renders as `view.url` in web-ui - rather puzzling for mobile devs 📱). There was also an update to `path` inference from `key` in https://github.com/DataDog/dd-sdk-ios/pull/534.

Trying to distill the problem makes me came up with this constraints:
* (1) to make the API convenient, we want to ask as little values as possible (idealy: only `name`);
* (2) single value cannot solve clashing views problem (starting distinct `"Foo"` view after previous `"Foo"`);
* (3) referring to https://github.com/DataDog/dd-sdk-ios/pull/534 , the `path` must be a technical reference (independent from translations and `name` changes).

(1) and (2) can be fulfilled by making the `name` a non-optional parameter and adding optional `key` with default `nil` value. Under the hood we could use `name + (key ?? "")` for [`RUMViewIdentity`](https://github.com/DataDog/dd-sdk-ios/blob/0d1fbd1cf2819576ba4a797d5184ede94d8e6555/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewIdentity.swift#L13). That might require using [`RUMCommandSubscriber`](https://github.com/DataDog/dd-sdk-ios/blob/0d1fbd1cf2819576ba4a797d5184ede94d8e6555/Sources/Datadog/RUM/AutoInstrumentation/RUMCommandSubscriber.swift#L9) instead of `RUMMonitor` to manage the RUM view.

The difficulty of (3) is defined by SwiftUI paradigm, where views have no identity (like `UIViewController` class in UIKit). If only we get something meaningful from `SwiftUI.View` (e.g. a `struct` name), that could work well for `path`. Is this sth that you mean?

>  I also thought about using the `SwiftUI.View's` composed body type

I think this might work 👌 - if we can indeed inspect the `content: Content` to pick some of its traits and call it a `path` (`view.url`) that would be similar to `path` being the `UIViewController's` class name in our `UIKit` monitoring.

Tests/DatadogTests/Datadog/Utils/SwiftUIExtensionsTests.swift
```diff
@@ -0,0 +1,52 @@
...
+    var body = EmptyView()
+}
+
+class SwiftUIExtensionsTests: XCTestCase {
+    func testSwiftUIViewTypeDescription() {
+        guard #available(iOS 13, *) else {
+            return
+        }
+
+        let view = TestView().trackRUMView(name: "test")
```
ncreated, line 27: This is test for agnostic `SwiftUI` extension, so IDK if we want to mix it with RUM specifics. If we want to test `typeDescription` after applying modifier, it seems better to use a default one (e.g. apply background color or sth). WDYT?

ncreated (commented):
Really good and thoughtful work 👌. I left few open questions on some elements that puzzle me, wondering on your take on these.

I've also ran unit tests on iOS 11, 12 and 13 - seems good 👍.
Screenshot 2021-11-03 at 11 07 35

Sources/Datadog/RUM/Instrumentation/RUMInstrumentation.swift
```diff
@@ -107,3 +111,13 @@ internal final class RUMInstrumentation: RUMCommandPublisher {
...
+
+extension RUMInstrumentation: SwiftUIViewHandler {
+    func onAppear(identity: String, name: String, path: String, attributes: [AttributeKey: AttributeValue]) {
+        swiftUIViewInstrumentation.onAppear(identity: identity, name: name, path: path, attributes: attributes)
+    }
+
+    func onDisappear(identity: String) {
+        swiftUIViewInstrumentation.onDisappear(identity: identity)
+    }
+}
```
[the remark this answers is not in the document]
maxep: Agreed, it was just a shortcut for the `ViewModifier` but it doesn't have much value.

Sources/Datadog/RUM/Instrumentation/Views/SwiftUI/SwiftUIViewModifier.swift
```diff
@@ -0,0 +1,60 @@
...
+    ///   - name: the View name used for RUM Explorer.
+    ///   - attributes: custom attributes to attach to the View.
+    /// - Returns: This view after applying a `ViewModifier` for monitoring the view.
+    func trackRUMView(
+        name: String,
+        attributes: [AttributeKey: AttributeValue] = [:]
+    ) -> some View {
+        let path = "\(name)/\(typeDescription.hashValue)"
+        return modifier(RUMViewModifier(name: name, path: path, attributes: attributes))
+    }
```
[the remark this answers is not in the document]
maxep: We cannot call .`onAppear()/ .onDisappear()` form a `SwiftUI.View`, a view only declare its body. That's why I left it out of unit tests and cover this in integration tests. I could try to workaround that by embedding the view in a `UIHostController` and see if I can trigger the events from it.

Tests/DatadogTests/Datadog/Utils/SwiftUIExtensionsTests.swift
maxep: Yes, makes sense 👍

Sources/Datadog/RUM/Instrumentation/Views/SwiftUI/SwiftUIViewModifier.swift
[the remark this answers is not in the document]
ncreated: If it's not too complex, IMO we can give it a shot. Otherwise, integration tests coverage is fine 👌 and definitely better than tricky or flaky units.

Sources/Datadog/RUM/Instrumentation/Views/SwiftUI/SwiftUIRUMViewsHandler.swift
```diff
@@ -0,0 +1,195 @@
...
+    private weak var subscriber: RUMCommandSubscriber?
+
+    /// The appearing views stack.
+    ///
+    /// This stack allows to track appearing and disappearing views to consistently
+    /// publish start and stop commands to the subscriber. The last item of the
+    /// stack is the visible one, any items below it have appeared before but not yet
+    /// disappeared. Therefore, they are considered not visible but can be revealed
+    /// if the last item disappears.
+    private var stack: [View] = []
```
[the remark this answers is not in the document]
ncreated: > With UIKit auto-instrumentation, this is done with the help of `UIKitHierarchyInspector` and a [simple state management](https://github.com/DataDog/dd-sdk-ios/blob/master/Sources/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsHandler.swift#L88L104).

Indeed, it's all true and very good observation 👌. Thanks for more insights - clearly, we can't handle value-based `SwiftUI.Views` in the same way as we do with object-based `UIKit.UIViewControllers`. The `stack` seems like a good, minimal approach 👍.

Speaking of `RUMM-1745`, I don't have any clear idea ATM. Like you say, having a single handler that will receive both `notify_viewDidAppear()` and `onAppear()` callbacks could be a solution. This way we could be considering both `UIKit` and `SwiftUI` navigation in one place, possibly leading to more control. Another approach could be to introduce next handler that combines `UIKit` and `SwiftUI` handlers into single stream - in theory that would mean less refactoring, but in practise it can be not doable if the hybrid support is mainly about details.

Sources/Datadog/RUM/Instrumentation/Views/SwiftUI/SwiftUIViewModifier.swift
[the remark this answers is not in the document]
maxep: I've added a unit test using `UIHostingController`. It also involve a mocked `SwiftUIViewHandler` and a new `RUMInstrumentation.init`.
`onAppear` and `onDisappear` have to be tested in a single unit because it has to wait for appearance before disappearing.

ncreated (approved):
🚀

PR #650, opened 2021-10-28
RUMM-1718 `error.sourceType` added

buranmert:
### What and why?

Errors may come from different sources, this new property helps specifying the origin platform of the error.

### How?

`error.sourceType` is added to JSON schemas and iOS SDK sets it to `.ios` by default.

### Review checklist

### DIFF
--- a/Tests/DatadogTests/Datadog/Mocks/RUMDataModelMocks.swift
+++ b/Tests/DatadogTests/Datadog/Mocks/RUMDataModelMocks.swift
@@ -229,7 +229,7 @@ extension RUMErrorEvent: RandomMockable {
                     url: .mockRandom()
                 ),
                 source: [.source, .network, .custom].randomElement()!,
-                sourceType: .ios,
+                sourceType: [.android, .browser, .ios, .reactNative].randomElement()!,
                 stack: .mockRandom(),
                 type: .mockRandom()
             ),
--- a/Tests/DatadogTests/Datadog/RUM/RUMMonitor/RUMCommandTests.swift
+++ b/Tests/DatadogTests/Datadog/RUM/RUMMonitor/RUMCommandTests.swift
@@ -57,6 +57,10 @@ class RUMCommandTests: XCTestCase {
 
         XCTAssertEqual(command2.errorSourceType, .reactNative)
         XCTAssertTrue(command2.attributes.isEmpty)
+
+        let defaultCommand = RUMAddCurrentViewErrorCommand(time: .mockAny(), message: .mockAny(), type: .mockAny(), stack: .mockAny(), source: .source, attributes: [:])
+
+        XCTAssertEqual(defaultCommand.errorSourceType, .ios)
     }
 
     func testWhenRUMStopResourceWithErrorCommand_isPassedErrorSourceTypeAttribute() {
@@ -69,5 +73,9 @@ class RUMCommandTests: XCTestCase {
 
         XCTAssertEqual(command2.errorSourceType, .reactNative)
         XCTAssertTrue(command2.attributes.isEmpty)
+
+        let defaultCommand = RUMStopResourceWithErrorCommand(resourceKey: .mockAny(), time: .mockAny(), message: .mockAny(), type: .mockAny(), source: .source, httpStatusCode: .mockAny(), attributes: [:])
+
+        XCTAssertEqual(defaultCommand.errorSourceType, .ios)
     }
 }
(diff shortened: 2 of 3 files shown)


Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
```diff
@@ -208,6 +208,10 @@ internal class RUMResourceScope: RUMScope {
 
     private func sendErrorEvent(on command: RUMStopResourceWithErrorCommand) -> Bool {
         attributes.merge(rumCommandAttributes: command.attributes)
+        let sourceType = (attributes.removeValue(forKey: RUMAttribute.internalErrorSourceType) as? String)
+            .flatMap {
+                return RUMErrorEvent.Error.SourceType(rawValue: $0)
+            } ?? .ios
```
ncreated, line 214: We're missing unit tests for this logic. However, that would require similar complex test to the one added in `RUMViewScopeTests`. Wouldn't it be be better to resolve this attribute in [`RUMStopResourceWithErrorCommand`](https://github.com/DataDog/dd-sdk-ios/blob/3f98760beffeffc9f261a5dc9344a3d60ea6fb6b/Sources/Datadog/RUM/RUMMonitor/RUMCommand.swift#L168) and [`RUMAddCurrentViewErrorCommand`](https://github.com/DataDog/dd-sdk-ios/blob/3f98760beffeffc9f261a5dc9344a3d60ea6fb6b/Sources/Datadog/RUM/RUMMonitor/RUMCommand.swift#L60)? 

Note that other properties describing resource error and view error are transported as `RUMCommand` fields, produced upstream and only read downstream. If adding `sourceType` to `RUMStopResourceWithErrorCommand` and `RUMAddCurrentViewErrorCommand` we can easily test all cases in `RUMCommandTests`. If further making RUM scope tests fuzzy on `sourceType: .mockRandom()` that would be minimal effort for max safety. WDYT @buranmert @maxep ?

buranmert: done ✅

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
```diff
@@ -233,6 +233,7 @@ internal class RUMResourceScope: RUMScope {
                     url: resourceURL
                 ),
                 source: command.errorSource.toRUMDataFormat,
+                sourceType: command.errorSourceType,
```
ncreated, line 236: This is not covered in `RUMResourceScopeTests.swift` - if we put an arbitrary value here, no test will trigger.

Tests/DatadogTests/Datadog/Mocks/RUMDataModelMocks.swift
```diff
@@ -229,6 +229,7 @@ extension RUMErrorEvent: RandomMockable {
                     url: .mockRandom()
                 ),
                 source: [.source, .network, .custom].randomElement()!,
+                sourceType: .ios,
```
ncreated, line 232: This is randomized mock (`mockRandom()`) for fuzzy testing, so it should use random value (like `source:`).

Tests/DatadogTests/Datadog/RUM/RUMMonitor/RUMCommandTests.swift
```diff
@@ -43,4 +46,28 @@ class RUMCommandTests: XCTestCase {
...
+        let command1 = RUMStopResourceWithErrorCommand(resourceKey: .mockAny(), time: .mockAny(), error: SwiftError(), source: .source, httpStatusCode: .mockAny(), attributes: [RUMAttribute.internalErrorSourceType: "react-native"])
+
+        XCTAssertEqual(command1.errorSourceType, .reactNative)
+        XCTAssertTrue(command1.attributes.isEmpty)
+
+        let command2 = RUMStopResourceWithErrorCommand(resourceKey: .mockAny(), time: .mockAny(), message: .mockAny(), type: .mockAny(), source: .source, httpStatusCode: .mockAny(), attributes: [RUMAttribute.internalErrorSourceType: "react-native"])
+
+        XCTAssertEqual(command2.errorSourceType, .reactNative)
+        XCTAssertTrue(command2.attributes.isEmpty)
+    }
```
ncreated, line 72: We're missing the case of using `.ios` default value when `RUMAttribute.internalErrorSourceType` is not specified.

ncreated (approved):
👌 ✨

PR #654, opened 2021-11-04
RUMM-1583 Fix attributes reading in Objective-C

ncreated:
### What and why?

📦 This PR fixes `DatadogObjc` issue with retrieving back attribute values in public APIs.

It was not possible to obtain the value of `Any` attribute, i.e. this would fail:
```swift
let view = DDRUMView(name: "name", attributes: ["foo": "bar"])
XCTAssertEqual(view.attributes["foo"] as? String, "bar") // 🐞 fail
```
It would fail because `"bar"` is wrapped into `AnyEncodable(value: "bar")` type erasure as part of our Objective-C <> Swift interoperability (Swift SDK represents attributes as `Encodable`). `AnyEncodable` is `internal` type to `DatadogObjc` so it was not even possible to read the attribute value back in any other way.

### How?

I introduced `func castAttributesToObjectiveC(_ attributes: [String: Encodable]) -> [String: Any]` function to unpack `AnyEncodable` values before they are passed to the user.

This had to be used in `DDRUMView`, `DDRUMAction` and all generated `RUMDataModels` for Objective-C.

Now, users will get exactly what they set through our Objective-C API.

### Review checklist


maxep (approved):
👍

buranmert (approved):

PR #655, opened 2021-11-05
Release 1.7.2

ncreated:
### What and why?

📦 This PR ships second hotfix for #609 by adding a workaround to potential iOS 15 issue.

### How?

We suspect that accessing `processInfo.isLowPowerModeEnabled` from pending `NSProcessInfoPowerStateDidChange` notification block leads to a deadlock on the main thread. As a workaround, we read iLPME value from the next run loop, where all underlying locks should already be released.

I reported this issue to Apple: https://openradar.appspot.com/FB9741207

💡 This branch includes version bump to `1.7.2`, but it's not visible in diff because I had to resolve conflict with the main branch (it conflicts becase we made `1.8.0-beta1` release after `1.7.0`). I also cherry picked recent fixes to release automation, so we can deploy this hotfix smoothly. If approved, I will put the release tag on 950b087437cf78e96fd2fca9fafc5607b7a7ce17.

### Review checklist


Sources/Datadog/Core/System/MobileDevice.swift
```diff
@@ -111,33 +111,41 @@ internal class MobileDevice {
...
                 guard let processInfo = notification.object as? ProcessInfo else {
                     return
                 }
-                self?.publisher.publishAsync(processInfo.isLowPowerModeEnabled)
+
+                // We suspect an iOS 15 bug (ref.: https://openradar.appspot.com/FB9741207) which leads to rare
+                // `_os_unfair_lock_recursive_abort` crash when `processInfo.isLowPowerModeEnabled` is accessed
+                // directly in the notification handler. As a workaround, we defer its access to the next run loop
+                // where underlying lock should be already released.
+                self?.notificationQueue.addOperation {
```
ncreated, line 145: Because `WebCore` seems to run its own notification handler in reported backtraces, I was considering deferring this call even more (e.g. to 3 run loops later), just for sanity. WDYT @buranmert @maxep?

e.g.:
```
13   WebCore                                  0x190a48118     -[WebLowPowerModeObserver initWithNotifier:] + 95
14   WebCore                                  0x1914c416c     WebCore::LowPowerModeNotifier::LowPowerModeNotifier(WTF::Function<void (bool)>&&) + 51
15   WebCore                                  0x1923e9c4c     WebCore::Page::Page(WebCore::PageConfiguration&&) + 1847
```

Sources/Datadog/Core/System/MobileDevice.swift
```diff
@@ -111,33 +111,41 @@ internal class MobileDevice {
...
+/// `.NSProcessInfoPowerStateDidChange` completion handler can sometimes lead to `_os_unfair_lock_recursive_abort` crash. The issue
+/// was reported to Apple, ref.: https://openradar.appspot.com/FB9741207
 private final class LowPowerModeMonitor {
     var isLowPowerModeEnabled: Bool {
         publisher.currentValue
     }
 
     private let publisher: ValuePublisher<Bool>
     private let notificationCenter: NotificationCenter
+    private let notificationQueue: OperationQueue
```
maxep, line 124: Since it uses the `.main` queue, do we have to keep it has a property?

maxep (approved):
Looks good 🤞 I've a couple of minor comments

Sources/Datadog/Core/System/MobileDevice.swift
ncreated: No reason, updated for minimal code 👍

Sources/Datadog/Core/System/MobileDevice.swift
maxep: 🤔 The idea would be to make sure that we don't read the value in the same loop as the `WebLowPowerModeObserver`? Deferring one loop is enough IMO, by reading [this](https://opensource.apple.com/source/WebCore/WebCore-7607.1.40.1.4/platform/ios/LowPowerModeNotifierIOS.mm.auto.html):
```objective-c
- (WebLowPowerModeObserver *)initWithNotifier:(WebCore::LowPowerModeNotifier&)notifier
{
    self = [super init];
    if (!self)
        return nil;

    _notifier = &notifier;
    _isLowPowerModeEnabled = [NSProcessInfo processInfo].lowPowerModeEnabled;
    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(_didReceiveLowPowerModeChange) name:NSProcessInfoPowerStateDidChangeNotification object:nil];
    return self;
}

...

- (void)_didReceiveLowPowerModeChange
{
    _isLowPowerModeEnabled = [NSProcessInfo processInfo].lowPowerModeEnabled;
    // We need to make sure we notify the client on the main thread.
    callOnMainThread([self, protectedSelf = RetainPtr<WebLowPowerModeObserver>(self)] {
        if (_notifier)
            notifyLowPowerModeChanged(*_notifier, _isLowPowerModeEnabled);
    });
}
```

It looks like `[NSProcessInfo processInfo].lowPowerModeEnabled` is invoked on the same loop as the observer. That being said we have no way of knowing if that's the actual code running on iOS, but I would be surprised!

ncreated: Sounds reasonable 👍👌, let's stick to only one runloop delay.

buranmert (approved):

PR #271, opened 2020-10-07
RUMM-715 RUM Views - modal presentation instrumentation - part 1

ncreated:
### What and why?

📦 This is the initial PR for modal view controllers auto instrumentation with RUM Views. It adds the test scenario fixture for developing, debugging and testing (with integration test) this feature.

### How?

I added a simple storyboard composed of 2 VCs. The first one presents the second modally. The integration test (yet with no assertions) runs 6 behaviours:
* presenting modal VC using storyboard segue,
* presenting modal VC using `present(_:animated:)`,
* dismissing modal VC using `self.dismiss(animated:)`,
* dismissing modal VC using `presentingViewController.dismiss(animated:)`,
* dismissing modal VC using interactive "pull down" gesture,
* cancelling modal VC dismissal by cancelling the interactive "pull down" gesture.

Oct-07-2020 16-22-43

In the next PRs I'll be adding necessary implementation for instrumenting this flow with RUM Views.

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference

PR #657, opened 2021-11-05
RUMM-1745 Use single handler for UIKit and SwiftUI views instrumentation

maxep:
### What and why?

[`UIHostingController`](https://developer.apple.com/documentation/swiftui/uihostingcontroller) and [`UIViewControllerRepresentable`](https://developer.apple.com/documentation/swiftui/uiviewcontrollerrepresentable) provide bridges between `SwiftUI` and `UIKit` allowing navigation between the two UI frameworks.

Therefore, we need to perform a consistent view tracking when instrumenting both `SwiftUI.View` and `UIKit.UIViewController`.

#### Single View Handler

#645 has introduced a new handler for instrumenting `SwiftUI.View` only: [`SwiftUIRUMViewsHandler`](https://github.com/DataDog/dd-sdk-ios/blob/feature/swiftui/Sources/Datadog/RUM/Instrumentation/Views/SwiftUI/SwiftUIViewModifier.swift).
Both `UIKitRUMViewsHandler` and `SwiftUIRUMViewsHandler` work similarly for forward navigation but adopt different solutions for tracking backward navigation.

Backward navigation tracking of `UIKit.UIViewController` uses a view-controller hierarchy inspector to retrieve the currently visible view, this approach is incompatible with `SwiftUI` since we cannot retrieve the declared `SwiftUI.View` from the view-controllers hierarchy.
 
The `SwiftUIRUMViewsHandler`, on the other hand, uses a stack of views to keep track of the appeared views and re-start them when needed during backward navigation. This approach is ui-framework agnostic and can easily be applied to `UIKit.UIViewController`.

The goal of the PR is to showcase the usage of a stack to track both `SwiftUI.View` and `UIKit.UIViewController` in a single handler: `RUMViewsHandler`

#### Tests

The same unit tests applied to `UIKitRUMViewsHandler` now run on `RUMViewsHandler`.
The integration tests `RUMSwiftUIScenarioTests` now perform hybrid navigation between `SwiftUI` and `UIKit`. Any other integration test stays untouched.

### Review checklist


Sources/Datadog/RUM/Instrumentation/Views/RUMViewsHandler.swift
```diff
@@ -193,3 +171,56 @@ internal final class SwiftUIRUMViewsHandler: SwiftUIViewHandler {
...
+extension RUMViewsHandler: UIViewControllerHandler {
+    func notify_viewDidAppear(viewController: UIViewController, animated: Bool) {
+        if let view = stack.first(where: { $0.identity.equals(viewController) }) {
+            // If the stack already contains the view controller, just restarts the view.
+            // This prevent from calling the predicate when unnecessary.
+            add(view: view)
+        } else if let rumView = predicate?.rumView(for: viewController) {
+            add(
+                view: .init(
+                    identity: viewController,
```
maxep, line 184: This will keep a strong ref to the `viewController`, even if `notify_viewDidDisappear` will be called to release it from the stack, we should keep a `weak` ref to the vc.
This should be addressed before merging.

Tests/DatadogTests/Datadog/RUM/Instrumentation/Views/RUMViewsHandlerTests.swift
```diff
@@ -7,21 +7,238 @@
...
+
+    // MARK: - Interacting with predicate
+
+    func testGivenAppearedView_whenTransitioningBackAndForthFromThisViewToAnother_thenPredicateIsCalledOnlyTwice() {
+        class Predicate: UIKitRUMViewsPredicate {
+            var numberOfCalls = 0
+
+            func rumView(for viewController: UIViewController) -> RUMView? {
+                numberOfCalls += 1
+                return .init(name: .mockRandom())
```
maxep, line 219: It differs from the [original test implementation](https://github.com/DataDog/dd-sdk-ios/blob/master/Tests/DatadogTests/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsHandlerTests.swift#L178): If the predicate rejects the vc, it is not added to the stack.

Sources/Datadog/RUM/Instrumentation/RUMInstrumentation.swift
```diff
@@ -46,10 +25,9 @@ internal final class RUMInstrumentation: RUMCommandPublisher {
         }
     }
 
+    let viewsInstrumentation: RUMViewsHandler
     /// RUM Views auto instrumentation, `nil` if not enabled.
-    let viewsAutoInstrumentation: ViewsAutoInstrumentation?
-    /// `SwiftUI.View` RUM instrumentation
-    let swiftUIViewInstrumentation: SwiftUIViewHandler
+    let viewsAutoInstrumentation: UIViewControllerSwizzler?
```
maxep, line 30: I'm not very satisfied with the naming here

Tests/DatadogIntegrationTests/Scenarios/RUM/RUMModalViewsScenarioTests.swift
```diff
@@ -9,7 +9,7 @@ import XCTest
 
 private extension ExampleApplication {
     func tapButton(titled buttonTitle: String) {
-        buttons[buttonTitle].tap()
+        buttons[buttonTitle].safeTap(within: 5)
```
maxep, line 12: Mitigating flakiness

Sources/Datadog/RUM/Instrumentation/Views/SwiftUI/SwiftUIViewHandler.swift
```diff
@@ -0,0 +1,17 @@
...
+import Foundation
+import UIKit
+
+/// Publisher generating RUM Commands on `SwiftUI.View` events.
+internal protocol SwiftUIViewHandler: RUMCommandPublisher {
+    /// Respond to a `SwiftUI.View.onAppear` event.
+    func onAppear(identity: String, name: String, path: String, attributes: [AttributeKey: AttributeValue])
+
+    /// Respond to a `SwiftUI.View.onDisappear` event.
+    func onDisappear(identity: String)
```
ncreated, line 16: nit: for consistency with other "handler" interfaces we could prefix these methods with `notify_` (or remove this prefix in other handlers): see `UIKit` views + actions handlers and `URLSession` tracing + RUM handlers.

Sources/Datadog/RUM/Instrumentation/Views/UIKit/UIKitHierarchyInspector.swift
```diff
@@ -1,66 +0,0 @@
...
-            guard let topVC = navigationVC.topViewController else {
-                return navigationVC // when `UINavigationController` displays no view controllers
-            }
-            // When `UINavigationController` has top VC, continue searching, as the top VC
-            // may be containing another hierarchy (e.g. `UITabBarController`).
-            return findTopViewControllerInHierarchy(of: topVC)
-        } else {
-            return viewController
-        }
-    }
```
ncreated, line 65: All these are smoothly supported with the new approach of tracking the `stack`? Or do you think there are still some edge cases we need to look closer at?

ncreated (commented):
Lookw like a very good approach 👌. I left few notes, but to me this is a good direction 👍.

Sources/Datadog/RUM/Instrumentation/Views/UIKit/UIKitHierarchyInspector.swift
maxep: `UITabBarController` and `UINavigationController` are both supported by the stack, it send events in the same manner. I did some manual monkey testing to validate it, and integration tests are both green. @ncreated do you see any other checks I should run? (don't know if e2e tests applies here).

I can think of one edge case which I need to verify: When presenting multiple modal page sheet and one of them is rejected by the predicate. Then the back navigation can be inconsistent:
1. start view 1
2. reject view 2
3. start view 3
4. Dismiss view 3 -> will restart view 1 ❌
 > This is due to the fact that the vc below a page sheet does not disappear, in this case view 1 will still be in the stack

Sources/Datadog/RUM/Instrumentation/RUMInstrumentation.swift
```diff
@@ -46,10 +25,9 @@ internal final class RUMInstrumentation: RUMCommandPublisher {
         }
     }
 
+    let viewsInstrumentation: RUMViewsHandler
```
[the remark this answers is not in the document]
maxep: Agreed 👍

Sources/Datadog/RUM/Instrumentation/RUMInstrumentation.swift
maxep: I'm not sure yet if `UIKitRUMUserActionsHandler` will support all SwiftUI actions. But I agree with clarifying by using `uiViewControllerSwizzler`, will see if I need to revisit in RUMM-1658.

Sources/Datadog/RUM/Instrumentation/Views/UIKit/UIKitHierarchyInspector.swift
ncreated: I see 👍. Yes, this new approach will require quite manual testing, as we're entirely changing the model. The initial reason for introducing `UIKitHierarchyInspector` was to deal with the complexity of back and forth transitions between `UIViewControllers`, notably modal view controllers. Test fixtures were added to `Example` app in #271 and the implementation of `UIKitHierarchyInspector` was done in #278.

Except having integration tests pass, we must do manual tests for the new implementation as well and I'll be happy to help on that later 🙌. Our integration tests do not cover all complexity (that would be tedious), and there might be still scenarios worth manual testing, e.g.:
* changing the entire navigation tree, by `window.rootViewController = ...`;
* dismissing first VC in the modal hierarchy of `N` VCs (it should dismiss the entire hierarchy);
* changing `navigationController.viewControllers` stack programmatically;
* jumping by few steps at once in the VCs hierarchy (e.g. imagine app receiving deeplink and handling it by pushing one VC to the stack + opening modal with another navigation VC).

I also remember that our `UIKitHierarchyInspector` was not well compatible with `UISearchController` and handling it required additional logic in the "views predicate". This compatibility could be best tested when dogfooding a branch in our Datadog iOS App, as it uses `UISearchController` in many places. We should have a look into this at the end, before final merge.

maxep: I have performed and validate the following manual testing by modifying the Example scenarios:
- changing the entire navigation tree, by `window.rootViewController = ...` ✅ 
- dismissing first VC in the modal hierarchy of `N` VCs, it should dismiss the entire hierarchy ✅ 
- changing `navigationController.viewControllers` stack programmatically ✅ 
- jumping by few steps at once in the VCs hierarchy (e.g. imagine app receiving deeplink and handling it by pushing one VC to the stack + opening modal with another navigation VC). ✅ 

I still need to test it with a `UISearchController`!

Sources/Datadog/RUM/Instrumentation/RUMInstrumentation.swift
```diff
@@ -61,13 +39,15 @@ internal final class RUMInstrumentation: RUMCommandPublisher {
...
-        var viewsAutoInstrumentation: ViewsAutoInstrumentation?
+        viewsHandler = RUMViewsHandler(dateProvider: dateProvider)
+        var viewsAutoInstrumentation: UIViewControllerSwizzler?
         var userActionsAutoInstrumentation: UserActionsAutoInstrumentation?
         var longTasks: LongTaskObserver?
 
         do {
             if let predicate = configuration.uiKitRUMViewsPredicate {
-                viewsAutoInstrumentation = try ViewsAutoInstrumentation(predicate: predicate, dateProvider: dateProvider)
+                viewsHandler.predicate = predicate
```
ncreated, line 49: I think it was much safer when `predicate` was injected in initializer and its access was scoped to `private let` instead of `internal var`. This was guaranteeing state consistency in `RUMViewsHandler` and was easier to follow the logic. What's the reason to assign the `predicate` after init?

Sources/Datadog/RUM/Instrumentation/Views/RUMViewsHandler.swift
```diff
@@ -35,14 +26,18 @@ internal final class SwiftUIRUMViewsHandler: SwiftUIViewHandler {
...
+
     /// The notification center where this handler observe the following notifications:
     /// - `UIApplicationDidEnterBackgroundNotification`
     /// - `UIApplicationWillEnterForegroundNotification`
     private weak var notificationCenter: NotificationCenter?
 
     /// The RUM Command subscriber responsible for processing
     /// this publisher's commands.
-    private weak var subscriber: RUMCommandSubscriber?
+    internal weak var subscriber: RUMCommandSubscriber?
```
ncreated, line 40: Does it need to be `internal`? Now `RUMViewsHandler` state can be altered from outside.

Sources/Datadog/RUM/Instrumentation/Views/RUMViewsHandler.swift
```diff
@@ -193,3 +179,56 @@ internal final class SwiftUIRUMViewsHandler: SwiftUIViewHandler {
         }
     }
 }
+
+extension RUMViewsHandler: UIViewControllerHandler {
+    func notify_viewDidAppear(viewController: UIViewController, animated: Bool) {
+        if let view = stack.first(where: { $0.identity.equals(viewController) }) {
+            // If the stack already contains the view controller, just restarts the view.
+            // This prevent from calling the predicate when unnecessary.
```
ncreated, line 187:
```suggestion
            // This prevents from calling the predicate when unnecessary.
```

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMViewIdentityTests.swift
```diff
@@ -39,8 +41,10 @@ class RUMViewIdentityTests: XCTestCase {
         // Then
         XCTAssertTrue(identity1.equals(key1))
         XCTAssertTrue(identity2.equals(key2))
+        XCTAssertTrue(identity1.equals(identity1))
         XCTAssertFalse(identity1.equals(key2))
         XCTAssertFalse(identity2.equals(key1))
+        XCTAssertFalse(identity1.equals(identity2))
     }
```
ncreated, line 48: As now `equals()` accepts `nil`, I think we need to cover more cases in this file.

Sources/Datadog/RUM/Instrumentation/Views/UIKit/UIKitHierarchyInspector.swift
ncreated: > I still need to test it with a UISearchController!

Keep in mind we don't yet support `UISearchController` very well - it still requires manual tweaking in the implementation of `predicate` to support search bar transitions to embedded VCs. I think it will be enough to just do regression test on `UISearchController` and check if it gets recognised by `UIKit` instrumentation as a whole. Solving the problem with search bar transitions to embedded VCs is beyond the scope of this PR.

Sources/Datadog/RUM/Instrumentation/Views/RUMViewsHandler.swift
maxep: I think this was for testing, let me double check!

Sources/Datadog/RUM/Instrumentation/Views/RUMViewsHandler.swift
```diff
@@ -193,3 +179,56 @@ internal final class SwiftUIRUMViewsHandler: SwiftUIViewHandler {
...
+            )
+        }
+    }
+
+    func notify_viewDidDisappear(viewController: UIViewController, animated: Bool) {
+        guard stack.contains(where: { $0.identity.equals(viewController) }) else {
+            return
+        }
+
+        remove(identity: viewController.asRUMViewIdentity())
```
buranmert, line 206: `remove(identity:)` already does the `contains` check implicitly

buranmert (commented):
pr lgtm but i was wondering if we could steer our users away from using `RUMViewModifier` with small views?
for example there can be many fragments in the screen at once in Android, i guess now it's potentially the same in iOS with SwiftUI. is it somehow possible to disallow using `RUMViewModifier` with list cells or small views like that?

Sources/Datadog/RUM/Instrumentation/Views/RUMViewsHandler.swift
maxep: So the SwiftUI view handler wasn't exposing the `subscriber`, but the [`UIKitRUMViewsHandler` did](https://github.com/DataDog/dd-sdk-ios/blob/master/Sources/Datadog/RUM/AutoInstrumentation/Views/UIKitRUMViewsHandler.swift#L38). This is required to be able to run [this test](https://github.com/DataDog/dd-sdk-ios/blob/master/Tests/DatadogTests/Datadog/RUM/AutoInstrumentation/RUMAutoInstrumentationTests.swift#L23L45).

I agree it would be best to keep it private but this would prevent from checking it in tests. WDYT?

Sources/Datadog/RUM/Instrumentation/Views/RUMViewsHandler.swift
maxep: Ah yes. good catch!

maxep:
@buranmert That is true, but we have no way of knowing the rendered view ratio or presentation context sadly. We could use something like [`GeometryReader`](https://developer.apple.com/documentation/swiftui/geometryreader), but this would add much more complexity in terms of integration for our users.

Sources/Datadog/RUM/Instrumentation/RUMInstrumentation.swift
maxep: The `viewsHandler` property is now non-optional, this makes the implementation of the `RUMInstrumentation.init` quite complex given conditions in `FeaturesConfiguration.RUM.Instrumentation` and throwable underlying init. I will see if I can improve that, I agree it should be best to keep the `predicate` immutable.

ncreated: I think just following could be safe enough and not that complex, no?
```swift
viewsHandler = RUMViewsHandler(
   dateProvider: dateProvider, 
   predicate: configuration.uiKitRUMViewsPredicate
)

do {
   if configuration.uiKitRUMViewsPredicate != nil {
      // UIKit auto instrumentation is enabled, so we need to install swizzler:
      viewsAutoInstrumentation = try UIViewControllerSwizzler(handler: viewsHandler)
   }
} ...
```

maxep: Yep, that works :)

Sources/Datadog/RUM/Instrumentation/Views/RUMViewsHandler.swift
ncreated: Ah I see 👍. I think it's all good until:
* it's `private` and has integration tests coverage;
* it's `internal` and has unit tests coverage.

We have both unit and integration coverage, so all options are fine. We can keep it `internal` 👍.

Tests/DatadogTests/Datadog/RUM/Instrumentation/Views/RUMViewsHandlerTests.swift
ncreated: Nice 👌, this clearly shows the improvement over previous approach, where UIKit handler would be invoked 5 times for this test 👍.

ncreated (approved):
Great work 🏅, it looks good to me 👍.

buranmert (approved):

PR #658, opened 2021-11-09
RUMM-1709 Allow to specify custom SDK version

ncreated:
### What and why?

📦 This PR adds an option to set custom `sdkVersion` with existing `set(additionalConfiguration:)` used by `bridge-ios`. The goal is to enable React Native and other cross-platform SDKs to send data using their own SDK version.

Counterpart of https://github.com/DataDog/dd-sdk-android/pull/747

### How?

Custom version can be passed through `_dd.sdk_version` attribute. When specified, it is resolved against the default `sdkVersion` from `Versioning.swift`. Version resolution happens on SDK startup and resolved value is consistently passed to downstream components leveraging existing dependency injection of `FeaturesConfiguration`.

**Alternatively**, we could naively override the `sdkVersion` (in `Versioning.swift`) when starting the SDK, but that would come with the classic price of introducing side effects (potential inconsistencies; state alternation in unit testing; refactoring need once we switch to supporting multiple instances of the SDK in one process). I rejected this approach.

I also did refactoring of all cross-platform attributes in the first commit (https://github.com/DataDog/dd-sdk-ios/commit/56af3d7e7da60890551b20fc187c4daa891bbbe8), by gathering all of them in a single place with describing their context and giving documentation comments.

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/FeaturesConfiguration.swift
+++ b/Sources/Datadog/Core/FeaturesConfiguration.swift
@@ -125,7 +125,7 @@ extension FeaturesConfiguration {
         var tracesEndpoint = configuration.tracesEndpoint
         var rumEndpoint = configuration.rumEndpoint
 
-        let source = (configuration.additionalConfiguration["_dd.source"] as? String) ?? Datadog.Constants.ddsource
+        let source = (configuration.additionalConfiguration[CrossPlatformAttributes.ddsource] as? String) ?? Datadog.Constants.ddsource
 
         if let datadogEndpoint = configuration.datadogEndpoint {
             // If `.set(endpoint:)` API was used, it should override the values
--- a/Sources/Datadog/Core/FeaturesConfiguration.swift
+++ b/Sources/Datadog/Core/FeaturesConfiguration.swift
@@ -19,6 +19,7 @@ internal struct FeaturesConfiguration {
         let environment: String
         let performance: PerformancePreset
         let source: String
+        let sdkVersion: String
         let proxyConfiguration: [AnyHashable: Any]?
     }
 
@@ -125,8 +126,6 @@ extension FeaturesConfiguration {
         var tracesEndpoint = configuration.tracesEndpoint
         var rumEndpoint = configuration.rumEndpoint
 
-        let source = (configuration.additionalConfiguration[CrossPlatformAttributes.ddsource] as? String) ?? Datadog.Constants.ddsource
-
         if let datadogEndpoint = configuration.datadogEndpoint {
             // If `.set(endpoint:)` API was used, it should override the values
             // set by deprecated `.set(<feature>Endpoint:)` APIs.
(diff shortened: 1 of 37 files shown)


Sources/Datadog/Versioning.swift
```diff
@@ -1,3 +1,3 @@
 // GENERATED FILE: Do not edit directly
 
-internal let sdkVersion = "1.8.0-beta1"
+internal let __sdkVersion = "1.8.0-beta1"
```
buranmert, line 3: alternatively that could be `defaultSDKVersion` too

Sources/Datadog/Core/FeaturesConfiguration.swift
```diff
@@ -150,6 +149,9 @@ extension FeaturesConfiguration {
             rumEndpoint = .custom(url: customRUMEndpoint.absoluteString)
         }
 
+        let source = (configuration.additionalConfiguration[CrossPlatformAttributes.ddsource] as? String) ?? Datadog.Constants.ddsource
+        let sdkVersion = (configuration.additionalConfiguration[CrossPlatformAttributes.sdkVersion] as? String) ?? __sdkVersion
```
buranmert, line 153: maybe we can use both cross platform SDK version and native SDK version as `sdkVersion` in order not to lose version information, wdyt?

for example if RN passes `RN-1.2.3` to native SDK version `v4.5.6` we can merge them into one somehow: like `RN-1.2.3/iOS-4.5.6`?

buranmert (approved):
lgtm, i just left 2 open questions

Sources/Datadog/Core/FeaturesConfiguration.swift
maxep: Could be confusing for the user, no? 
Maybe it would be best to remove the `~>` from [here](https://github.com/DataDog/dd-bridge-ios/blob/main/DatadogSDKBridge.podspec#L20) and [here](https://github.com/DataDog/dd-sdk-reactnative/blob/main/packages/core/DatadogSDKReactNative.podspec#L20) so we know with certainty which version `RN-1.2.3` depends on.

maxep (approved):

Sources/Datadog/Versioning.swift
ncreated: My reasoning was: this value should **be never used** directly in our code - we should always use resolved `configuration.sdkVersion`. 

To avoid potential mistakes, I came up with this underscore notation `__sdkVersion`, hoping it will raise a CR flag when used in the code. The `defaultSDKVersion` looks like any other code we write, hence it could be misused and overlooked. 

If you think relying on the `__` prefix is too weak protection, I can add a linter rule for preventing direct usage of `__sdkVersion` in other files (like [we do with `UIApplication.shared`](https://github.com/DataDog/dd-sdk-ios/blob/6b735fb9ac5d704fd732842a59d5e7602870768d/tools/lint/sources.swiftlint.yml#L81-L89)).

Sources/Datadog/Core/FeaturesConfiguration.swift
ncreated: Interesting point 👍. I think that knowing the version of underlying native SDK might be important for us in troubleshooting, but not especially for our users.

As we leverage `~> x.y.z` for version pinning, it means that given version of RN SDK will use `x.y.*` version of the native SDK - so _major_ and _minor_ are pinned, and only _patch_ component is moving. The question is - **is it critical for us to know the `patch` version of underlying native SDK?** This however, is a separate discussion - I'll bring it to Slack.

PR #659, opened 2021-11-09
RUMM-640 Fix flaky test in `DatadogTests`

ncreated:
### What and why?

⚙️ This PR fixes flaky test added with `Datadog.clearAllData()` in #644.

It caused 64 failures among all branches during last 1 month:

Screenshot 2021-11-09 at 17 29 08

### How?

In this test:
* we initialise the SDK;
* then we write 1 file to both directories (`.pending` and `.granted`) for each of 4 features (so 8 files in total);
* then we call `Datadog.clearAllData()`;
* and then we expect no data in all directories.

This initial assertion in `// Given` block was failing:
```swift
// Given
XCTAssertEqual(numberOfFiles, 8, "Each feature stores 2 files - one authorised and one unauthorised")
```
The problem was caused by data migration performed on SDK start-up. When SDK starts, [it instantiates `ConsentAwareDataWriter` for each feature](https://github.com/DataDog/dd-sdk-ios/blob/eef82d8cd5bbb0764e2e1f4d13a324ff17f4509c/Sources/Datadog/Core/Feature.swift#L91), which [triggers asynchronous data migration](https://github.com/DataDog/dd-sdk-ios/blob/eef82d8cd5bbb0764e2e1f4d13a324ff17f4509c/Sources/Datadog/Core/Persistence/Writing/ConsentAwareDataWriter.swift#L37). Initial migration includes [removal of all files in `unauthorised` (`.pending`) directory](https://github.com/DataDog/dd-sdk-ios/blob/eef82d8cd5bbb0764e2e1f4d13a324ff17f4509c/Sources/Datadog/Core/Persistence/Migrating/DataMigrator.swift#L19). Because the test creates mock files right after instantiating the SDK, it led to classic race condition and some files being wiped out before assertion.

The solution is to perform synchronous wait (`sync {}`) on the writer's queue. This guarantees the order of events and ensures that test runs correctly. Confirmed by running 1K repetitions with 100% success rate (vs ~98% before):
```
Executed 1000 tests, with 0 failures (0 unexpected) in 47.845 (49.510) seconds
```

### Review checklist

### DIFF
--- a/Tests/DatadogTests/Datadog/DatadogTests.swift
+++ b/Tests/DatadogTests/Datadog/DatadogTests.swift
@@ -413,6 +413,18 @@ class DatadogTests: XCTestCase {
                 .build()
         )
 
+        // On SDK init, underlying `ConsentAwareDataWriter` performs data migration for each feature, which includes
+        // data removal in `unauthorised` (`.pending`) directory. To not cause test flakiness, we must ensure that
+        // mock data is written only after this operation completes - otherwise, migration may delete mocked files.
+        let loggingWriter = try XCTUnwrap(LoggingFeature.instance?.storage.writer as? ConsentAwareDataWriter)
+        let tracingWriter = try XCTUnwrap(TracingFeature.instance?.storage.writer as? ConsentAwareDataWriter)
+        let rumWriter = try XCTUnwrap(RUMFeature.instance?.storage.writer as? ConsentAwareDataWriter)
+        let internalMonitoringWriter = try XCTUnwrap(InternalMonitoringFeature.instance?.logsStorage.writer as? ConsentAwareDataWriter)
+        loggingWriter.queue.sync {}
+        tracingWriter.queue.sync {}
+        rumWriter.queue.sync {}
+        internalMonitoringWriter.queue.sync {}
+
         let featureDirectories: [FeatureDirectories] = [
             try obtainLoggingFeatureDirectories(),
             try obtainTracingFeatureDirectories(),


Tests/DatadogTests/Datadog/DatadogTests.swift
```diff
@@ -413,6 +413,18 @@ class DatadogTests: XCTestCase {
...
+        // data removal in `unauthorised` (`.pending`) directory. To not cause test flakiness, we must ensure that
+        // mock data is written only after this operation completes - otherwise, migration may delete mocked files.
+        let loggingWriter = try XCTUnwrap(LoggingFeature.instance?.storage.writer as? ConsentAwareDataWriter)
+        let tracingWriter = try XCTUnwrap(TracingFeature.instance?.storage.writer as? ConsentAwareDataWriter)
+        let rumWriter = try XCTUnwrap(RUMFeature.instance?.storage.writer as? ConsentAwareDataWriter)
+        let internalMonitoringWriter = try XCTUnwrap(InternalMonitoringFeature.instance?.logsStorage.writer as? ConsentAwareDataWriter)
+        loggingWriter.queue.sync {}
+        tracingWriter.queue.sync {}
+        rumWriter.queue.sync {}
+        internalMonitoringWriter.queue.sync {}
```
ncreated, line 426: I wish we can do it smarter than forcing into internals, but in this test we operate with public API and depend on deeper behaviour, so I can't see a nicer way for now. Classic `Thread.sleep()` would work as well, but I don't find it a better alternative (unlike this, it would be flaky by definition).

Easier testability of thread-dependent logic could a good candidate for SDK V2 constraint.

buranmert (approved):

maxep (approved):

PR #661, opened 2021-11-15
RUMM-1658 Manual instrumentation of SwiftUI actions

maxep:
### What and why?

Actionable `SwiftUI` components such as `Button` or `NavigationLink` are rendered in an opaque `UIView` subclass: `CGDrawingView`*. The SDK's auto-instrumentation of `UIEvent`s is not able to capture information from the view to accurately log an action event.

Given the `SwiftUI` api, we have no way of providing an auto-instrumentation of actions so we have to provide a simple way for instrumentation User Actions manually.

> *This is not true for all actionable components, Bar buttons or List items are respectively rendered in `UIControl` and `UITableViewCell` and are auto-instrumented by the SDK out of the box.

### How?

We will provide a `SwiftUI.View` extension that allows tracking of  Tap action by adding `simultaneousGesture` using a `ViewModifier`. The `Gesture` will then send User Actions events through the `Global.rum`.

### Review checklist


Sources/Datadog/RUM/Instrumentation/Actions/SwiftUI/SwiftUIActionModifier.swift
```diff
@@ -0,0 +1,103 @@
...
+                    }
+                }
+                .onEnded { _ in
+                    if started {
+                        Global.rum.stopUserAction(type: .swipe, name: name, attributes: attributes)
+                        started.toggle()
+                    }
+                }
+        )
+    }
```
maxep, line 64: This is a proposal for instrumentation `swipe` and/or `scroll` actions. SwiftUI only provide a `DragGesture`.

maxep: This is not covered by tests. If we adopt this, we should add tests before merging

Datadog/Example/Scenarios/RUM/SwiftUIInstrumentation/SwiftUIRootViewController.swift
```diff
@@ -44,12 +44,12 @@ struct RootView: View {
...
         // appears` and once when the Tab item itself appears. This
-        // lead to RUM views being reported twice. This issue was fixed
-        // in iOS 14.5, see https://developer.apple.com/documentation/ios-ipados-release-notes/ios-ipados-14_5-release-notes
+        // lead to RUM views being reported twice. This issue seems to
+        // have been fixed in iOS 15.
         // As a workaround, the tab view items can be embedded in a
         // `LazyVStack` or `LazyHStack` to lazily load its content when
         // it needs to render them onscreen.
-        if #available(iOS 14.5, *) {
+        if #available(iOS 15, *) {
```
ncreated, line 52: Why exactly we need this change in this PR 🤔💭?

Sources/Datadog/RUM/Instrumentation/Actions/SwiftUI/SwiftUIActionModifier.swift
```diff
@@ -0,0 +1,103 @@
...
+    let count: Int
+
+    /// Action Name used for RUM Explorer.
+    let name: String
+
+    /// Custom attributes to attach to the Action.
+    let attributes: [AttributeKey: AttributeValue]
+
+    func body(content: Content) -> some View {
+        content.simultaneousGesture(
```
ncreated, line 24: `SwiftUI` newbie question - it is not possible to use generics / protocol constraints to know that `Content` is for example `NavigationLink` and read its `Text` then `label` to have the `name` inferred automatically (e.g. `"Push to Next View"`)?

ncreated (commented):
I think this looks like a good way 👌. Left one question, but to my `SwiftUI` understanding it is not possible to make it more "automatic".

Datadog/Example/Scenarios/RUM/SwiftUIInstrumentation/SwiftUIRootViewController.swift
maxep: Arg, this is not related to this PR scope. The new CI stack has highlighted that the bug was not actually fixed in iOS 14.5 but in 15 (Or maybe in between but we have no simulator to test, xcode provide simulators for iOS 14.5 and the iOS 15). The iOS 14.5 changelog might be referencing another issue..

This change is also in #657, I will rebase when it get merged!

Sources/Datadog/RUM/Instrumentation/Actions/SwiftUI/SwiftUIActionModifier.swift
maxep: Indeed, I would have liked that very much... Definition of `NavigationLink` or `Button`, or any other components I encountered does not expose anything of their state or content. I also tried to use accessibility info but I couldn't infer anything from the `ViewModifier.Content` or the `SwiftUI.View` itself.
I might be missing something tho, but I haven't been able to infer any valuable information from the view so far.

maxep:
Actually, I will have to remove the `.swipe` instrumentation using `SwiftUI.DragGesture`. During the implementation of integration tests, I've noticed that the `simultaneousGesture(_:)` is not actually working and act as `gesture(_:)`, so it will block any other interaction in the view.

Sources/Datadog/RUM/Instrumentation/Actions/SwiftUI/SwiftUIActionModifier.swift
```diff
@@ -0,0 +1,51 @@
...
+                Global.rum.addUserAction(type: .tap, name: name, attributes: attributes)
+            }
+        )
+    }
+}
+
+@available(iOS 13, *)
+public extension SwiftUI.View {
+    /// Monitor this view tap actions with Datadog RUM. An Action event will be logged after a number
+    /// of required tap.
```
ncreated, line 35:
```suggestion
    /// of required taps.
```

ncreated (approved):
👌 LG! Only wondering if `RUMM-1764` is not a problem here - I don't remember the conclusion and if it impacts iOS or only Android.

maxep:
>  I don't remember the conclusion and if it impacts iOS or only Android.
IIRC It was concluded that Android will align to iOS by removing side-effect detection on manual instrumentation.

PR #665, opened 2021-11-16
RUMM-1763 Read `_dd.error.is_crash` crossplatform attribute

maxep:
### What and why?

Fatal errors from RN (and future cross plateform sdks)  are reported with a custom attribute `_dd.error.is_crash`. 

### How?

The value of `_dd.error.is_crash` will map to RUM Error’s `error.is_crash`.

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference

PR #667, opened 2021-11-18
RUMM-1763 Run test on view scope instead of resource

maxep:
### What and why?

#665 is running fuzzy test on the resource scope instead of the view. A resource scope error is not fatal, so `RUMResourceScopeTests` has been reverted and the test now applies to `RUMViewScopeTests`

### Review checklist


Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMViewScopeTests.swift
```diff
@@ -596,6 +596,38 @@ class RUMViewScopeTests: XCTestCase {
         XCTAssertEqual(viewUpdate.model.view.error.count, 1)
     }
 
+    func testGivenStartedView_whenCrossPlatformErrorIsAdded_itSendsCorrectErrorEvent() throws {
+        var currentTime: Date = .mockDecember15th2019At10AMUTC()
+        let isCrash: Bool = .random()
```
buranmert, line 601: i think we can make this `true`, it should be good enough if `true -> true` works (no need to test `false -> false` too after that)

ncreated (approved):
👌

buranmert (approved):

PR #668, opened 2021-11-18
RUMM-1386 Fix `BinaryImage.imageName` resolution on iOS Simulator with Xcode 13.1

ncreated:
### What and why?

⚙️🐞 This PR fixes bug discovered by recent upgrade to Xcode 13.1 on CI.

### How?

With Xcode 13.1 and iOS Simulator some paths to binary images may include illegal characters (e.g. spaces) which return `nil` for `URL(string:)`:

```swift
let path = "/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS 13.4.simruntime/Contents/Resources/RuntimeRoot/System/Library/Frameworks/CoreFoundation.framework/CoreFoundation"
//                                                                ^ 🤨
print(URL(string: path)) // nil
```

We shouldn't use `URL(string:)` for path processing, so I'm replacing it with `URL(fileURLWithPath:)` to fix the issue. The improvement is that now it returns `URL`, not `URL?` so I also managed to simplify the code a little bit.

🛡 Tested on real device as well.

### Review checklist

### DIFF
--- a/Sources/DatadogCrashReporting/PLCrashReporterIntegration/CrashReport.swift
+++ b/Sources/DatadogCrashReporting/PLCrashReporterIntegration/CrashReport.swift
@@ -93,7 +93,7 @@ internal struct BinaryImageInfo {
     /// The UUID of this image.
     var uuid: String
     /// The name of this image (referenced by "library name" in the stack frame).
-    var imageName: String?
+    var imageName: String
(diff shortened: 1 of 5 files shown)


maxep (approved):
LGTM 👍

buranmert (approved):

PR #669, opened 2021-11-19
RUMM-1758 Add RUM Session debug utility

ncreated:
### What and why?

📦 As part of `RUMM-1758` work (proposing RFC to reduce number of view updates), I made a simple tool to debug RUM session. I think we can leverage it on a longer run, so I'm adding it to our existing debug toolset.

### How?

Unlike existing "Debug RUM" tool, this new "Debug RUM session" utility allows for sending RUM session composed from different events (actions, errors, resources). Additionally, resources can be controlled by their "start" and "stop", so it is possible to debug sequences like:
* start view A
* start resource R1
* start resource R2
* stop resource R1
* start view B
* stop resource R2

It's also all made in SwiftUI, so adding new features / utils could be easier.

Screenshot 2021-11-19 at 14 29 24

### Review checklist

~- [] Feature or bugfix MUST have appropriate tests (unit, integration)~

### DIFF
--- a/Datadog/Example/Debugging/Helpers/SwiftUI.swift
+++ b/Datadog/Example/Debugging/Helpers/SwiftUI.swift
@@ -1,0 +1,43 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import SwiftUI
+
+@available(iOS 13.0, *)
+extension Color {
+    /// Datadog purple.
+    static var datadogPurple: Color {
+        return Color(UIColor(red: 99/256, green: 44/256, blue: 166/256, alpha: 1))
+    }
+
+    static var rumViewColor: Color {
+        return Color(UIColor(red: 0/256, green: 107/256, blue: 194/256, alpha: 1))
+    }
+
+    static var rumResourceColor: Color {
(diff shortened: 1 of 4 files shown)


Datadog/Example/Debugging/Helpers/SwiftUI.swift
```diff
@@ -0,0 +1,43 @@
...
+
+    static var rumErrorColor: Color {
+        return Color(UIColor(red: 235/256, green: 54/256, blue: 7/256, alpha: 1))
+    }
+}
+
+@available(iOS 13.0, *)
+internal struct DatadogButtonStyle: ButtonStyle {
+    func makeBody(configuration: DatadogButtonStyle.Configuration) -> some View {
+        return configuration.label
```
ncreated, line 36: @maxep I'm wondering if this kind of styling is sth we could leverage in `SwiftUI` instrumentation? The way it works is by inserting:
```swift
.buttonStyle(DatadogButtonStyle())
```
in view declaration **globally**, which then runs this piece of code for all `SwiftUI` buttons. There are also distinct styles for lists and other elements.

maxep: Interesting! This could be something usable for populating the name, But from what I see, the `PrimitiveButtonStyle` and `PrimitiveButtonStyleConfiguration` do not expose much. The only thing we could leverage is the `ButtonRole`, which can be cancel or destructive. Worth exploring tho 👍

maxep (commented):
Very nice, it will be useful 👌

maxep (approved):

buranmert (approved):

Issue #653, opened 2021-11-01
`try?` causes Swift Error breakpoint to break on launch


iliaskarim:
### The thing

E.g. in https://github.com/DataDog/dd-sdk-ios/blob/master/Sources/Datadog/Core/Persistence/Files/Directory.swift

```swift
try? FileManager.default.removeItem(at: temporaryDirectory.url)
```
causes Xcode to break for "Swift Error."

This can prolong a debugging session when relying on a Swift Error breakpoint.

Here's an alternative that would address the issue at hand.
```swift
do {
  try FileManager.default.removeItem(at: temporaryDirectory.url)
} catch {
  // no-op
}
```

buranmert:
hi @iliaskarim 👋 

thanks for reporting the issue.
i was able to reproduce it: `try? FileManager.default.removeItem(at: temporaryDirectory.url)` always throws 3 times at app launch. i made a fix above ☝️ 
other `throw` statements don't need fix as they don't always throw and are useful to catch errors, so i left them unchanged.

PR #670, opened 2021-11-19
RUMM-1780 Prevent FileManager from throwing

buranmert:
Closes https://github.com/DataDog/dd-sdk-ios/issues/653

### What and why?

`try? FileManager.default.removeItem(at: tempDirectory.url)` is called as an extra safety measure.
It throws most of the time, for example at the app launch it typically throws 3 times: once per feature.
When `Swift Error Breakpoint` is enabled, that causes debugger to halt multiple times.

### How?

`try? FileManager.default.removeItem(at: tempDirectory.url)` should only be called if `FileManager.default.replaceItemAt(url, withItemAt: tempDirectory.url)` fails.
Now this method call is wrapped with a `do-catch` and `removeItem` is called only if an error is caught.

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/Persistence/Files/Directory.swift
+++ b/Sources/Datadog/Core/Persistence/Files/Directory.swift
@@ -59,7 +59,9 @@ internal struct Directory {
         try retry(times: 3, delay: 0.001) {
             _ = try FileManager.default.replaceItemAt(url, withItemAt: temporaryDirectory.url)
         }
-        try? FileManager.default.removeItem(at: temporaryDirectory.url)
+        if FileManager.default.fileExists(atPath: temporaryDirectory.url.path) {
+            try FileManager.default.removeItem(at: temporaryDirectory.url)
+        }
     }
 
     /// Moves all files from this directory to `destinationDirectory`.


Sources/Datadog/Core/Persistence/Files/Directory.swift
```diff
@@ -55,11 +55,14 @@ internal struct Directory {
         // Instead of iterating over all files and removing them one by one, we create a temporary
         // empty directory and replace source directory content with (empty) temporary folder.
         // This makes the deletion atomic, and is more performant in benchmarks.
-        let temporaryDirectory = try Directory(withSubdirectoryPath: "com.datadoghq/\(UUID().uuidString)")
-        try retry(times: 3, delay: 0.001) {
-            _ = try FileManager.default.replaceItemAt(url, withItemAt: temporaryDirectory.url)
+        let tempDirectory = try Directory(withSubdirectoryPath: "com.datadoghq/\(UUID().uuidString)")
+        do {
```
ncreated, line 59: Doesn't it change the existing logic entirely? Errors thrown from here are catched in a layer above (in data migrator) and get send to our Internal Monitoring system, for observability. Now it seems we just mute them here with no further consideration and measure, no?

Sources/Datadog/Core/Persistence/Files/Directory.swift
```diff
@@ -55,11 +55,14 @@ internal struct Directory {
...
+            try retry(times: 3, delay: 0.001) {
+                _ = try FileManager.default.replaceItemAt(url, withItemAt: tempDirectory.url)
+            }
+        } catch {
+            try? FileManager.default.removeItem(at: tempDirectory.url)
```
ncreated, line 64: I think the real issue is in this line. Why does it throw multiple times after SDK init? There must be clearly something we're missing and I'd rather vote for fixing the problem, not muting it more. It is already muted once by `try?` (we're not interested in handling this error as it doesn't seem actionable). 

Maybe this temp location doesn't exist when the runtime gets there. In such case, we can just if-check it and cleanup if necessary. WDYT?

ncreated (changes requested):
I'm not sure if this is the right fix. It seems to disable observability of issues in data migration.

Sources/Datadog/Core/Persistence/Files/Directory.swift
buranmert: tl;dr: we can keep propagating the error after `removeItem(at: tempDirectory.url)`
```swift
catch {
  try? FileManager.default.removeItem(at: tempDirectory.url) // try to cleanup, we can ignore this error
  throw error // this error comes from `replaceItemAt`, this can be useful for IM
}
```
wdyt?

---
it is called once per feature (logging, tracing, rum) and 3 (or 4 if internal monitoring is enabled) times in total.
it seems like `replaceItemAt` removes `tempDirectory`:
```swift
removeItem(tempDir) // succeeds
replaceItemAt(url, tempDir)
---
replaceItemAt(url, tempDir)
removeItem(tempDir) // throws
```

my first idea was:
```swift
if fm.fileExists(at: tempDir.url) {
  fm.removeItem(at: tempDir.url)
}
```
but then i found [this in Apple docs](https://developer.apple.com/documentation/foundation/filemanager/1415645-fileexists):

> It’s far better to attempt an operation, check for errors, and handle those errors gracefully than it is to try to figure out ahead of time whether the operation will succeed.

Sources/Datadog/Core/Persistence/Files/Directory.swift
buranmert: true, but i think there is already a problem with the current code:
```
try replaceItemAt(...) // throws and stops execution in case of error...
try? removeItem(...) // cleanup. but this line never executes in case of error as it stops above already
```

i proposed something in the other comment to keep propating the error and doing the cleanup at the same time 👇

Sources/Datadog/Core/Persistence/Files/Directory.swift
ncreated:
```swift
if FileManager.default.fileExists(atPath: temporaryDirectory.url.path) {
    try FileManager.default.removeItem(at: temporaryDirectory.url)
}
```
I don't see anything wrong with such approach. From the previous logic, we don't expect `temporaryDirectory.url` to exist (it should be removed in `mv` operation). However, if it's still there (due to unexpected / random issue in `mv`) we wan't to delete it explicitly. If that deletion causes another error, we don't have any more choices. In such case, we track it through Internal Monitoring to see how frequent it is. I don't expect we will see many reports from it.

buranmert: done

ncreated (approved):

PR #674, opened 2021-11-26
RUMM-1785 DatadogEventBridge implemented

buranmert:
### What and why?

As part of `WKWebView` tracking, Datadog iOS SDK needs to communicate with Datadog Browser SDK.

### How?

`DatadogEventBridge` is implemented as a mean of communication between those two platforms.
It receives messages from Browser SDK, parses them into JSON and forwards to corresponding `WebEventConsumer` (`Log` or `RUM`).

### NOTE

Also, an extension of `WKUserContentController` is added in order to inject `DatadogEventBridge` into the `WKWebView`. However, I can't find a way to test this extension in unit tests. I tried `webView.evaluateJavaScript(...)` API to call JS methods within the webview, yet that didn't work in unit tests (although it works in `Example` app).
We may need to test this extension in integration tests.

### Review checklist


Sources/Datadog/RUM/WebView/WKUserContentController+Datadog.swift
```diff
@@ -0,0 +1,71 @@
...
+
+    func userContentController(
+        _ userContentController: WKUserContentController,
+        didReceive message: WKScriptMessage
+    ) {
+        print(message.name + String(describing: message.body))
+        do {
+            try eventBridge.consume(message.body)
+        } catch {
+            userLogger.error("Web Event Error", error: error)
```
ncreated, line 68: This won't be actionable nor verbose to the user, because `userLogger` prints logs in a short format, so only message is displayed in the console (`"Web Event Error"`). We should rather [print error information inline](https://github.com/DataDog/dd-sdk-ios/blob/f7c7a3a79d416f836235cfdb4a9ee9b9faa4e95b/Sources/Datadog/Core/Persistence/Writing/FileWriter.swift#L45).

Sources/Datadog/RUM/WebView/DatadogEventBridge.swift
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        if eventType == Constants.eventTypeLog {
+            logEventConsumer.consume(event: wrappedEvent, eventType: eventType)
+        } else {
+            rumEventConsumer.consume(event: wrappedEvent, eventType: eventType)
+        }
+    }
+
+    static func parse(_ message: String) throws -> JSON {
```
ncreated, line 56: Why `static`? Can't it be `internal`? I don't think we want to call it from outside, no?

Sources/Datadog/RUM/WebView/DatadogEventBridge.swift
```diff
@@ -0,0 +1,66 @@
...
+        guard let data = message.data(using: .utf8) else {
+            throw WebEventError.dataSerialization(message: message)
+        }
+        let rawJSON = try JSONSerialization.jsonObject(with: data, options: [])
```
ncreated, line 60: Which thread does it execute on? Seems that we will be calling it from the WebKit's callback thread - perhaps the main thread, and we should avoid deserialising things on the UI queue.

Sources/Datadog/RUM/WebView/DatadogEventBridge.swift
```diff
@@ -0,0 +1,66 @@
...
+        guard let json = rawJSON as? JSON else {
+            throw WebEventError.JSONSerialization(rawJSONDescription: String(describing: rawJSON))
```
ncreated, line 62: It's rather "de-" serialization what we're doing here ("serialization" would be JSON → String).
```suggestion
            throw WebEventError.JSONDeserialization(rawJSONDescription: String(describing: rawJSON))
```

Tests/DatadogTests/Datadog/RUM/WebView/DatadogEventBridgeTests.swift
```diff
@@ -0,0 +1,85 @@
...
+    private let mockRUMEventConsumer = MockEventConsumer()
+    lazy var eventBridge = DatadogEventBridge(
+        logEventConsumer: mockLogEventConsumer,
+        rumEventConsumer: mockRUMEventConsumer
+    )
+
+    // MARK: - Parsing
+
+    func testWhenLogMessageIsValid_itSucceedsParsing() throws {
+        let json = try DatadogEventBridge.parse(messageLog)
```
ncreated, line 36: Seems we're testing private method of `DatadogEventBridge` - do we need need `parse()` in internal interface? Can't this can be tested through `consume()`?

Tests/DatadogTests/Datadog/RUM/WebView/DatadogEventBridgeTests.swift
```diff
@@ -0,0 +1,85 @@
...
+        XCTAssertEqual(json[DatadogEventBridge.Constants.eventTypeKey] as? String, "log")
```
ncreated, line 37: What about asserting other attributes of `json`? `"log"` event type is just one of many things we may want to ensure.

ncreated (changes requested):
I left few questions, about the architecture plan (vs upcoming V2 efforts), threading and testing. 

IMO this can't be single-threaded feature, as Browser events are equally complex to mobile events and we should process them on a background queue.

Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
```diff
@@ -0,0 +1,76 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+import WebKit
+
+// TODO: RUMM-1794 rename the method
+public extension WKUserContentController {
+    func addDatadogMessageHandler(allowedWebViewHosts: [String]) {
```
ncreated, line 12: How about unit testing this by subclassing `WKUserContentController` and spying calls to `add(messageHandler, name: bridgeName)` and `addUserScript()`? This way we will test basic, but fundamental aspect of the bridge configuration.

Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
```diff
@@ -0,0 +1,76 @@
...
+            .map { return "\"\($0)\"" }
+            .joined(separator: ",")
+
+        let js = """
+        window.\(bridgeName) = {
+          send(msg) {
+            \(webkitMethodName)(msg)
+          },
+          getAllowedWebViewHosts() {
+            return '[\(allowedWebViewHostsString)]'
```
ncreated, line 38: 🐞 Here we pass user input into JS code without sanitization or escaping and it can clearly break the syntax. Seems like a good candidate for unit test 👌.

Sources/Datadog/FeaturesIntegration/WebView/DatadogEventBridge.swift
```diff
@@ -0,0 +1,66 @@
...
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+internal typealias JSON = [String: Any]
+
+internal protocol WebEventConsumer {
+    func consume(event: JSON, eventType: String)
+}
```
maxep, line 13: How the `eventType` will be used by the consumer? From the bridge implementation, the event type is used to choose which consumer to invoke. So I guess the consumer will only process to event json.

Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
```diff
@@ -0,0 +1,76 @@
...
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+import WebKit
+
+// TODO: RUMM-1794 rename the method
+public extension WKUserContentController {
+    func addDatadogMessageHandler(allowedWebViewHosts: [String]) {
+        let bridgeName = DatadogMessageHandler.name
```
maxep, line 13: Why using a static `DatadogMessageHandler.name`? I don't see it being used outside of this method.

Sources/Datadog/FeaturesIntegration/WebView/DatadogEventBridge.swift
buranmert: you can see it [here in Android](https://github.com/DataDog/dd-sdk-android/blob/feature/hybrid-application/dd-sdk-android/src/main/kotlin/com/datadog/android/webview/internal/WebRumEventConsumer.kt#L50)

RUM events have different `eventType` values although Logs only have `eventType: "log"`

Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
buranmert: i thought `WKUserContentController` should know as little as possible about `DatadogMessageHandler` internals, but this is a trivial change and doesn't change much as you noticed.

should i change anyway?

maxep: Yea not a big deal, makes it less easy to read maybe. Up to you!

ncreated (approved):

maxep (approved):

PR #677, opened 2021-12-01
RUMM-1765 Collect RUM events during application launch - part 1

ncreated:
### What and why?

📦 This is the first PR on _application launch events tracking_. It implements the base logic of starting "ApplicationLaunch" view if any RUM event is received before tracking the first view:

Screenshot 2021-12-01 at 14 55 22

### How?

It was known that this feature will collide with existing [Background Events Tracking](https://github.com/DataDog/dd-sdk-ios/blob/b28016175691665c9c2e1d3d867e90d6f4437ab2/Sources/Datadog/DatadogConfiguration.swift#L678-L690) option, which creates "Background" view for capturing off-view events. Most of the effort in this PR is focused on decoupling both and implementing a clear separation of following cases:
* when event is received and existing session didn't yet track any view → start "ApplicationLaunch" view;
* when event is received and existing session has tracked some view → start "Background" view (if [BET](https://github.com/DataDog/dd-sdk-ios/blob/b28016175691665c9c2e1d3d867e90d6f4437ab2/Sources/Datadog/DatadogConfiguration.swift#L678-L690) is enabled);

This required introducing a notion of "initial RUM session" - the very first session in current app process. If session times out, next RUM session is not marked as "initial".

To make code more readable and the implementation more explicit, I added two values to the `RUMCommand` interface, so we can clearly know if a given event should result with a side effect in off-view tracking:
```swift
/// Whether or not receiving this command should start the "Background" view if no view is active
/// and ``Datadog.Configuration.Builder.trackBackgroundEvents(_:)`` is enabled.
var canStartBackgroundView: Bool { get }

/// Whether or not receiving this command should start the "ApplicationLaunch" view if no view was yet started in current app process.
var canStartApplicationLaunchView: Bool { get }
```

#### Next PRs

There are two key elements missing in this PR, which will be delivered in separate PRs:
* When the app is launched in background (e.g. due to location change), we should not create "ApplicationLaunch" view - instead, "Background" view should be created if user opted-in for [BET](https://github.com/DataDog/dd-sdk-ios/blob/b28016175691665c9c2e1d3d867e90d6f4437ab2/Sources/Datadog/DatadogConfiguration.swift#L678-L690).
* When the app crashes while there is no active RUM view, we need to handle this crash respectively from restarted session. We need to consider if it should be sent within "ApplicationLaunch" view, "Background" view or sampled.

More on these edge cases can be read in _"RFC - Collecting App Launch Crashes in Mobile SDKs"_.

### Review checklist

### DIFF
--- a/Sources/Datadog/DatadogConfiguration.swift
+++ b/Sources/Datadog/DatadogConfiguration.swift
@@ -675,7 +675,7 @@ extension Datadog {
                 return self
             }
 
-            /// Enables or disables automatic tracking of background events (events hapenning when no UIViewController is active).
+            /// Enables or disables automatic tracking of background events (events hapenning when no `UIViewController` is active).
             ///
             /// When enabled, the SDK will track RUM Events into an automatically created Background RUM View (named `Background`)
             ///
--- a/Sources/Datadog/RUM/RUMMonitor/RUMCommand.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/RUMCommand.swift
@@ -36,11 +36,6 @@ internal struct RUMStartViewCommand: RUMCommand {
     /// The path of this View, rendered in RUM Explorer as `VIEW URL`.
     let path: String
 
-    /// Used to indicate if this command starts the very first View in the app.
-    /// * default `false` means _it's not yet known_,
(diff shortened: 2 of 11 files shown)


buranmert (approved):

maxep (approved):
LGTM 👍

PR #679, opened 2021-12-02
🚩RUM-1755 Add config overrides for debug launch arguments

fuzzybinary:
Adds two launch arguments:
DD_DEBUG sets sampling to 100%, verbosity to debug, and upload frequency to frequent
DD_DEBUG_RUM enables Datadog.debugRUM

All of these settings are changed during initialization to override whatever configuration the user set

### DIFF
--- a/Sources/Datadog/Core/System/MobileDevice.swift
+++ b/Sources/Datadog/Core/System/MobileDevice.swift
@@ -79,50 +79,29 @@ internal class MobileDevice {
         )
     }
 
-    // MARK: - Singleton
-    private static var _instance: MobileDevice?
-
-    /// Returns current mobile device  if `UIDevice` is available on this platform.
-    /// On other platforms returns `nil`.
-    static var current: MobileDevice {
-        get {
-            if let instance = _instance {
-                return instance
-            }
-
-            #if !targetEnvironment(simulator)
-            // Real device
-            _instance = MobileDevice(
-                uiDevice: UIDevice.current,
-                processInfo: ProcessInfo.processInfo,
-                notificationCenter: .default
-            )
-            #else
-            // iOS Simulator - battery monitoring doesn't work on Simulator, so return "always OK" value
-            _instance = MobileDevice(
-                model: UIDevice.current.model,
-                osName: UIDevice.current.systemName,
-                osVersion: UIDevice.current.systemVersion,
-                processInfo: ProcessInfo.processInfo,
-                enableBatteryStatusMonitoring: {},
-                resetBatteryStatusMonitoring: {},
-                currentBatteryStatus: { BatteryStatus(state: .full, level: 1, isLowPowerModeEnabled: false) }
-            )
-            #endif
-
-            // swiftlint:disable:next force_unwrapping
-            return _instance!
-        }
-        set(newInstance) {
-            _instance = newInstance
-        }
+    convenience init() {
+        #if !targetEnvironment(simulator)
+        // Real device
+        self.init(
+            uiDevice: UIDevice.current,
+            processInfo: ProcessInfo.processInfo,
+            notificationCenter: .default
+        )
+        #else
+        // iOS Simulator - battery monitoring doesn't work on Simulator, so return "always OK" value
+        self.init(
+            model: UIDevice.current.model,
+            osName: UIDevice.current.systemName,
+            osVersion: UIDevice.current.systemVersion,
+            processInfo: ProcessInfo.processInfo,
+            enableBatteryStatusMonitoring: {},
+            resetBatteryStatusMonitoring: {},
+            currentBatteryStatus: { BatteryStatus(state: .full, level: 1, isLowPowerModeEnabled: false) }
+        )
+        #endif
     }
 
-    #if DD_SDK_COMPILED_FOR_TESTING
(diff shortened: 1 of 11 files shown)


Sources/Datadog/Core/System/MobileDevice.swift
```diff
@@ -74,28 +78,51 @@ internal class MobileDevice {
...
+        set(newInstance) {
+            _instance = newInstance
+        }
+    }
+
+    #if DD_SDK_COMPILED_FOR_TESTING
+    static func clearForTesting() {
+        _instance = nil
     }
+    #endif
```
maxep, line 125: I think we can simplify this and avoid having to use a second singleton by creating another `convenience init`, like:
```swift
    convenience init() {
        #if !targetEnvironment(simulator)
        // Real device
        self.init(
            uiDevice: UIDevice.current,
            processInfo: ProcessInfo.processInfo,
            notificationCenter: .default
        )
        #else
        // iOS Simulator - battery monitoring doesn't work on Simulator, so return "always OK" value
        self.init(
            model: UIDevice.current.model,
            osName: UIDevice.current.systemName,
            osVersion: UIDevice.current.systemVersion,
            processInfo: ProcessInfo.processInfo,
            enableBatteryStatusMonitoring: {},
            resetBatteryStatusMonitoring: {},
            currentBatteryStatus: { BatteryStatus(state: .full, level: 1, isLowPowerModeEnabled: false) }
        )
        #endif
    }

    /// Returns current mobile device.
    static var current = MobileDevice()

    #if DD_SDK_COMPILED_FOR_TESTING
    static func clearForTesting() {
        current = MobileDevice()
    }
    #endif
```

Also the `clearForTesting` might not be even necessary if we do:
```swift
func testSomething() {
    MobileDevice.current = .mock()
    defer { MobileDevice.current = .init() }

    ...
}
```

maxep (commented):
This looks great 🚀  well done with the configuration mutation and the console warning!
I have a suggestion re. the `MobileDevice` singleton, LMKWYT!

Sources/Datadog/Core/System/MobileDevice.swift
fuzzybinary: That's a good call. I was originally thinking ok making `set` only visible to testing, but that wasn't really possible so your solution ends up being a lot cleaner given what's possible.

fuzzybinary:
@maxep Updated with your suggestions.

ncreated (commented):
Good contribution 👍, although I wonder if we couldn't achieve this in a simpler way. Here we propose a few new concepts like `MobileDevice` singleton or `Datadog.Configuration` mutability and I'm not sure if adding a debug utility is enough reason for this. I left separate comments, I wonder what's your take on these topics 🙂🙌.

Also, please leverage our [GH PR template](https://github.com/DataDog/dd-sdk-ios/blob/master/.github/PULL_REQUEST_TEMPLATE.md) for next PRs 🚀 (what/how sections), as it's part of our contribution guide and really helps visiting past topics in the future 🙏.

Sources/Datadog/Core/System/MobileDevice.swift
```diff
@@ -74,29 +78,31 @@ internal class MobileDevice {
...
             osVersion: UIDevice.current.systemVersion,
+            processInfo: ProcessInfo.processInfo,
             enableBatteryStatusMonitoring: {},
             resetBatteryStatusMonitoring: {},
             currentBatteryStatus: { BatteryStatus(state: .full, level: 1, isLowPowerModeEnabled: false) }
         )
         #endif
     }
 
+    static var current = MobileDevice()
```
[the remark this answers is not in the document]
fuzzybinary: 👍 I'm largely against singletons as well. I'll look for a way to remove it cleanly in this PR if we want. I was mostly looking to disturb as little as possible for this particular PR.

fuzzybinary:
@ncreated Thinking about the best way to approach this, I'm thinking putting ProcessInfo into AppContext might actually be the simplest / best approach here that allows us use the system supplied default or inject a fake ProcessInfo without cluttering the  public API.

Thoughts?

ncreated:
> @ncreated Thinking about the best way to approach this, I'm thinking putting ProcessInfo into AppContext might actually be the simplest / best approach here that allows us use the system supplied default or inject a fake ProcessInfo without cluttering the public API.
> 
> Thoughts?

@fuzzybinary I think this is really good idea 👌! Extending "app context" with "launch arguments" sounds conceptually correct, and if thinking of multiple SDK instances - each would have it's own context, so that will be easy to scope this feature to the right instance of `Datadog` 👍.

fuzzybinary:
@ncreated Made the changes, and removed `MobileDevice.current` -- as you pointed out `MobileDevice.init()` essentially did the same thing. I also moved the flag overrides into `FeaturesConfiguration` for most of the places that was possible.

I did leave `ProcessInfo` as the argument to AppContext, as there might be other items in ProcessInfo we could leverage in the future, like environment. Let me know what you think.

ncreated (approved):
Looks 💯🚀! And really well done in testing 👍. 

I'm adding `needs-doc` label, so we can track this for future documentation work (usually we update docs later, after release).

PR #680, opened 2021-12-03
Allow passing in a type for errors sent with a message

AvdLee:
### What and why?

The current error reporting of Datadog didn't result in the best results for us:
image

As you can see, these 3 errors are the same, but they aren't seen as the same since they differ due to the associated value.
We've implemented our own custom error parsing locally which uses mirroring to get the enum case and associated values separately:

```swift
// A wrapper around tracked errors to convert into Datadog suitable error messages.
struct DatadogError {
    let underlyingError: Error
    let type: String
    let message: String
    let attributes: [String: Encodable]
    let file: StaticString
    let line: UInt

    init(trackedError: TrackedError) {
        let caseName = trackedError.error.caseName
        let associatedValues = trackedError.error.associatedValues
        let errorDescription = (trackedError.error as? LocalizedError)?.errorDescription ?? caseName

        message = errorDescription
        type = String(describing: Swift.type(of: trackedError.error))
        underlyingError = trackedError.error
        file = trackedError.file
        line = trackedError.line
        attributes = [
            "description": trackedError.description,
            "file": "\(URL(fileURLWithPath: String(describing: trackedError.file)).lastPathComponent)#L\(trackedError.line)",
            "function": String(describing: trackedError.function),
            "associated_values": associatedValues
        ]
    }
}
```

This results in an improved error message:

image

And the following custom attributes:

image

### How?

Though, for this to work we had to be able to set the `type` when adding an error using a custom message. This PR opens up that initialiser due to adjusting the public method.

### Review checklist

- [X] Feature or bugfix MUST have appropriate tests (unit, integration) (note: tests didn't run locally, so I'll see what CI reports)
- [X] Make sure each commit and the PR mention the Issue number or JIRA reference (There was no issue for this)

### DIFF
--- a/Sources/Datadog/DDRUMMonitor.swift
+++ b/Sources/Datadog/DDRUMMonitor.swift
@@ -223,11 +223,13 @@ public class DDRUMMonitor {
     /// - Parameters:
     ///   - resourceKey: the key representing the Resource - must match the one used in `startResourceLoading(...)`.
     ///   - errorMessage: the message explaining the Resource failure.
+    ///   - type: the type of Error. Defaults to `nil`.
     ///   - response: an optional `URLResepone` received for the Resource.
     ///   - attributes: custom attributes to attach to the Resource.
     public func stopResourceLoadingWithError(
         resourceKey: String,
         errorMessage: String,
+        type: String? = nil,
         response: URLResponse? = nil,
         attributes: [AttributeKey: AttributeValue] = [:]
     ) {}
--- a/Sources/Datadog/DDRUMMonitor.swift
+++ b/Sources/Datadog/DDRUMMonitor.swift
@@ -79,7 +79,7 @@ public class DDRUMMonitor {
     /// Notifies that an Error occurred in currently presented View.
     /// - Parameters:
     ///   - message: a message explaining the Error.
-    ///   - type: the type of Error.
+    ///   - type: the type of Error. Defaults to `nil`.
     ///   - source: the origin of the error.
     ///   - stack: stack trace of the error. No specific format needed. Overwrites `file` and `line` parameters below.
     ///   - attributes: custom attributes to attach to the Error
@@ -87,7 +87,7 @@ public class DDRUMMonitor {
     ///   - line: the line number on which the Error occurred (the default is the line number on which this method was called).
     public func addError(
         message: String,
-        type: String?,
+        type: String? = nil,
         source: RUMErrorSource = .custom,
         stack: String? = nil,
         attributes: [AttributeKey: AttributeValue] = [:],
--- a/Sources/Datadog/DDRUMMonitor.swift
+++ b/Sources/Datadog/DDRUMMonitor.swift
@@ -79,6 +79,7 @@ public class DDRUMMonitor {
     /// Notifies that an Error occurred in currently presented View.
     /// - Parameters:
     ///   - message: a message explaining the Error.
+    ///   - type: the type of Error.
     ///   - source: the origin of the error.
     ///   - stack: stack trace of the error. No specific format needed. Overwrites `file` and `line` parameters below.
     ///   - attributes: custom attributes to attach to the Error
@@ -86,6 +87,7 @@ public class DDRUMMonitor {
     ///   - line: the line number on which the Error occurred (the default is the line number on which this method was called).
     public func addError(
         message: String,
+        type: String?,
         source: RUMErrorSource = .custom,
         stack: String? = nil,
         attributes: [AttributeKey: AttributeValue] = [:],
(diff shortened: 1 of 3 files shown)


Sources/Datadog/DDRUMMonitor.swift
```diff
@@ -79,13 +79,15 @@ public class DDRUMMonitor {
     /// Notifies that an Error occurred in currently presented View.
     /// - Parameters:
     ///   - message: a message explaining the Error.
+    ///   - type: the type of Error.
```
BasThomas, line 82: It's not super clear to me what a "type of Error" is, or how one would be expected to use it. I think it'd help to add least add an example for context here. :)

Sources/Datadog/DDRUMMonitor.swift
```diff
@@ -79,13 +79,15 @@ public class DDRUMMonitor {
...
     ///   - source: the origin of the error.
     ///   - stack: stack trace of the error. No specific format needed. Overwrites `file` and `line` parameters below.
     ///   - attributes: custom attributes to attach to the Error
     ///   - file: the file in which the Error occurred (the default is the file name in which this method was called).
     ///   - line: the line number on which the Error occurred (the default is the line number on which this method was called).
     public func addError(
         message: String,
+        type: String?,
```
maxep, line 90: Thank you for your contribution @AvdLee 🙏 
For backward compatibility, it would be best to add `nil` as a default value here.

Sources/Datadog/DDRUMMonitor.swift
AvdLee: I kept the documentation in line with other parts of the SDK for now!

ncreated (changes requested):
Thanks for this contribution @AvdLee 👌! We have discussed this internally and we'd be happy to extend our public API in the proposed way 👍 - indeed we should allow for customizing `error.type` as it's not always inferred automatically.

Diff looks fine, although I think it would be more consistent to also update resource error API in the same way, as it clearly misses the `type` as well:
```swift
public func stopResourceLoadingWithError(
   resourceKey: String,
   errorMessage: String,
   response: URLResponse? = nil,
) {}
```

Second, we need to test this feature ⚔️. In case of `RUMMonitor` and its public APIs we use integration tests defined in [`RUMMonitorTests`](https://github.com/DataDog/dd-sdk-ios/blob/master/Tests/DatadogTests/Datadog/RUMMonitorTests.swift). I think we could simply update this test case to pass random `type` in and expect it in assertion:
https://github.com/DataDog/dd-sdk-ios/blob/8e7c19e90ed3d53ec61e40ddb1b202a660303dc4/Tests/DatadogTests/Datadog/RUMMonitorTests.swift#L379-L381

We have similar behaviour [test defined for resource error as well](https://github.com/DataDog/dd-sdk-ios/blob/8e7c19e90ed3d53ec61e40ddb1b202a660303dc4/Tests/DatadogTests/Datadog/RUMMonitorTests.swift#L503).

AvdLee:
@ncreated I tried running tests locally, but those fail:

image

I've also run make, but there seems to be something wrong in the script:

```bash
avanderlee@Antoines-MBP dd-sdk-ios % make
⚙️  Installing dependencies...
Warning: carthage 0.38.0 already installed
Please update to the latest Carthage version: 0.38.0. You currently are on 0.36.0
Unrecognized arguments: --use-xcframeworks
make: *** [dependencies] Error 1
```

I'll try to update your requested code changes and write the tests w/o running them locally

AvdLee:
@ncreated added the tests. If they fail, I'm happy to fix them, but it would be great if I can run the tests locally. Could you look into the `make` file and see what's wrong? It would help for other contributors too.

ncreated:
> @ncreated added the tests. If they fail, I'm happy to fix them, but it would be great if I can run the tests locally. Could you look into the `make` file and see what's wrong? It would help for other contributors too.

@AvdLee I think your Carthage version is behind what we require - I can see this in your log:
```
Please update to the latest Carthage version: 0.38.0. You currently are on 0.36.0
```
`dd-sdk-ios` requires `0.38.0`, which comes with XCFrameworks support. I tested it on my side, it works fine with just `make`, but we will update `CONTRIBUTING.md` with this requirement 👌.

---

And thanks for adding tests! There was few things missing, not easy to spot by external contributor 🙂. When sending new resource / error events, the SDK might send some more intermediate events, which wasn't asserted correctly. I updated tests in this PR by using our new and simplified assertion mechanism that reduces intermediate events.

It should be all good now 👍. I will merge it when it completes on CI (we're struggling with some flakiness today ☔). Once again, thanks a lot for this contribution 🚀🏅, we really appreciate your help 💪!

ncreated (approved):

maxep (approved):
LGTM 🚀  Thanks again @AvdLee !

AvdLee:
@ncreated I was already on `0.38.0` as the logs state as well:

> Warning: carthage 0.38.0 already installed

So I think there's something wrong in the script still 🤔 

> It should be all good now 👍. I will merge it when it completes on CI (we're struggling with some flakiness today ☔). Once again, thanks a lot for this contribution 🚀🏅, we really appreciate your help 💪!

All good, thanks for helping out on those tests! 💪

PR #681, opened 2021-12-06
RUMM-1765 Add debug utility to test background events

ncreated:
### What and why?

🧰🐞 This PR adds debug utility to Example app for testing [Background Events Tracking](https://github.com/DataDog/dd-sdk-ios/blob/8e7c19e90ed3d53ec61e40ddb1b202a660303dc4/Sources/Datadog/DatadogConfiguration.swift#L678-L690) feature. This will also come extremely handy in my current work on collecting application launch events (`RUMM-1765`).

So far we didn't have an easy way of debugging scenario where the app is launched by the OS in background. Thanks to this utility I located a bug 🐞 in our BET implementation, where "Background" view was not created automatically if there was no RUM session (starting initial RUM session [required an explicit "start view" command](https://github.com/DataDog/dd-sdk-ios/blob/8e7c19e90ed3d53ec61e40ddb1b202a660303dc4/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift#L80-L81)). This problem was fixed in #677.

### How?

This utility leverages [iOS background location tracking](https://developer.apple.com/documentation/corelocation/getting_the_user_s_location/handling_location_events_in_the_background). Location tracking can be enabled and the app will keep sending RUM actions for each location change even if the app is sent to background or terminated.

**To use it:**

1. Open "Background events" debug screen.
2. Tap "START" - it will request location tracking.
3. Authorize it for "While Using App" - since now, the app will be tracking location changes in foreground.
4. Kill the app - it will request authorizing location tracking in background, answer "Change To Always Allow".

From now on, it will start tracking _significant location changes_ even if the app is not running. To stop tracking, open the app again and tap "STOP". To reset authorization status, uninstall the app from Simulator or device.

Screenshot 2021-12-06 at 16 58 01

It can be easily tested when simulating location changes, e.g. in "Freeway Drive" mode:

Screenshot 2021-12-06 at 16 59 08

For each location update it sends both logs and RUM actions:

Screenshot 2021-12-06 at 15 58 56

### Review checklist

### DIFF
--- a/Datadog/Example/Debugging/BackgroundEvents/BackgroundLocationMonitor.swift
+++ b/Datadog/Example/Debugging/BackgroundEvents/BackgroundLocationMonitor.swift
@@ -23,7 +23,7 @@ internal class BackgroundLocationMonitor: NSObject, CLLocationManagerDelegate {
     ///
     /// Note: `BackgroundLocationMonitor` can be started independently from receiving location monitoring authorization status.
     /// Even if this value is `true`, location updates might not be delivered due to restricted or denied status.
-    var isStarted: Bool {
+    private(set) var isStarted: Bool {
         get { UserDefaults.standard.bool(forKey: Constants.locationMonitoringUserDefaultsKey) }
         set { UserDefaults.standard.set(newValue, forKey: Constants.locationMonitoringUserDefaultsKey) }
     }
--- a/Datadog/Example/Debugging/BackgroundEvents/BackgroundLocationMonitor.swift
+++ b/Datadog/Example/Debugging/BackgroundEvents/BackgroundLocationMonitor.swift
@@ -1,0 +1,129 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+import Datadog
+import CoreLocation
+
+internal var backgroundLocationMonitor: BackgroundLocationMonitor?
+
+/// Location monitor used in "Example" app for debugging and testing iOS SDK features in background.
+internal class BackgroundLocationMonitor: NSObject, CLLocationManagerDelegate {
+    private struct Constants {
+        static let locationMonitoringUserDefaultsKey = "is-location-monitoring-started"
+    }
+
+    private let locationManager = CLLocationManager()
+
+    /// Tells if location monitoring is started.
+    /// This setting is preserved between application launches. Defaults to `false`.
+    ///
+    /// Note: `BackgroundLocationMonitor` can be started independently from receiving location monitoring authorization status.
+    /// Even if this value is `true`, location updates might not be delivered due to restricted or denied status.
+    var isStarted: Bool {
+        get { UserDefaults.standard.bool(forKey: Constants.locationMonitoringUserDefaultsKey) }
+        set { UserDefaults.standard.set(newValue, forKey: Constants.locationMonitoringUserDefaultsKey) }
+    }
+
+    /// Current authorization status for location monitoring.
+    var currentAuthorizationStatus: String { authorizationStatusDescription(for: locationManager) }
+
+    /// Notifies change of authorization status for location monitoring.
+    var onAuthorizationStatusChange: ((String) -> Void)? = nil
+
+    override init() {
+        super.init()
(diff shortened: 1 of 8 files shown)


Datadog/Example/ExampleAppDelegate.swift
```diff
@@ -72,6 +72,13 @@ class ExampleAppDelegate: UIResponder, UIApplicationDelegate {
             launch(storyboard: storyboard)
         }
 
+        // Instantiate location monitor if the Example app is run in interactive mode. This will
+        // enable background location tracking if it was started in previous session.
+        let isRunningExampleApp = !Environment.isRunningUnitTests() && !Environment.isRunningUITests()
+        if isRunningExampleApp {
+            backgroundLocationMonitor = BackgroundLocationMonitor()
+        }
+
```
maxep, line 81: Could be handy to have an `Environment.isInteractive` var.

Datadog/Example/Debugging/BackgroundEvents/BackgroundLocationMonitor.swift
```diff
@@ -0,0 +1,129 @@
...
+
+    /// Tells if location monitoring is started.
+    /// This setting is preserved between application launches. Defaults to `false`.
+    ///
+    /// Note: `BackgroundLocationMonitor` can be started independently from receiving location monitoring authorization status.
+    /// Even if this value is `true`, location updates might not be delivered due to restricted or denied status.
+    var isStarted: Bool {
+        get { UserDefaults.standard.bool(forKey: Constants.locationMonitoringUserDefaultsKey) }
+        set { UserDefaults.standard.set(newValue, forKey: Constants.locationMonitoringUserDefaultsKey) }
+    }
```
maxep, line 29: Looks like an instance var but it's actually a `static` one :) could we make this more explicit?

maxep (approved):
Nice addition, and in SwiftUI :) I a couple of minor/non-blocking suggestions!

buranmert (approved):
can we/do we need to test background fetch in a similar way?

Datadog/Example/Debugging/BackgroundEvents/BackgroundLocationMonitor.swift
ncreated: I think there was one detail missing. I changed this property to `private(set)`:
```swift
private(set) var isStarted: Bool {
    get { UserDefaults.standard.bool(forKey: Constants.locationMonitoringUserDefaultsKey) }
    set { UserDefaults.standard.set(newValue, forKey: Constants.locationMonitoringUserDefaultsKey) }
}
```

Now, when looking solely on this property's code then yes, it can be static, but using instance semantic in this case gives us third option for its value: `backgroundLocationManager?.isStarted` can be `true`, `false` or `nil`.

Note that this value is **read** in `init()` to start the underlying `CLLocationManager` and it is **written** from `startMonitoring()` and `stopMonitoring()`. Instance semantic and `private(set)` ensure we have maximum cohesion in BLM state:
* if `blm?.isStarted` is `true`, we know that `CLLocationManager` is started,
* if `blm?.isStarted` is `false`, we know that `CLLocationManager` is not started,
* if `blm?.isStarted` evaluates to `nil`, we don't know the state of location monitoring.

If this is changed to `static`, then `BLM.isStarted == true` would not necessarily mean that `CLLocationManager` is running, as this thing is managed by `blm` instance.

Datadog/Example/ExampleAppDelegate.swift
ncreated: Good point, changed 👍

ncreated:
> can we/do we need to test background fetch in a similar way?

We could, but I found [Background Fetch API](https://developer.apple.com/documentation/uikit/app_and_environment/scenes/preparing_your_ui_to_run_in_the_background/using_background_tasks_to_update_your_app) way more complex than background location monitoring. From debugging perspective - it's easy to trigger BF on Simulator, but not on physical device (iOS BF heuristic is unpredictable). 

On the other hand, Background Location Monitoring is quite reliable - same easy way to trigger it on Simulator and it is possible to debug it on real device by moving `~500m` away.

Datadog/Example/Debugging/BackgroundEvents/BackgroundLocationMonitor.swift
maxep: Yes, it makes more sense with `private(set)` 👍

PR #682, opened 2021-12-08
RUMM-1821 Fix flakiness in `RUMResourcesScenarioTests`

ncreated:
### What and why?

⚙️🐞 This PR fixes two issues in `RUMResourcesScenarioTests` which causes recent flakiness.

Frequent problem (> 10 CI failures in last `15d`):
Screenshot 2021-12-09 at 14 00 00

Rare problem (2 CI failures in last `15d`):
Screenshot 2021-12-09 at 14 00 47

### How?

The **frequent problem** was caused by sending requests to these 3rd party URLs:
```diff
-let thirdPartyGETResourceURL = URL(string: "https://bitrise.io")!
-let thirdPartyPOSTResourceURL = URL(string: "https://bitrise.io/about")!
```
For some reason, when running this in Bitrise container the SSL and DNS phases were sometimes skipped. I don't know the exact reason (it might be related to Bitrise infra), but changing these to proper `GET` and `POST` endpoints (external) seems to improve situation. I ran these tests successfully 20 times in a CI loop.
```diff
+let thirdPartyGETResourceURL = URL(string: "https://httpbin.org/get")!
+let thirdPartyPOSTResourceURL = URL(string: "https://httpbin.org/post")!
```

The **rare problem** was caused by using too weak condition for pulling requests from Python server:
```diff
// Get RUM Sessions with expected number of View visits and Resources
let rumRequests = try rumServerSession.pullRecordedRequests(timeout: dataDeliveryTimeout) { requests in
-    try RUMSessionMatcher.singleSession(from: requests)?.viewVisits.count == 2
}
```
Depending on SDK batching, 2 views could be sent within one batch, whereas their resources and errors would be transported in a separate one. Pulling condition was fulfilled too early leading to incomplete `RUMSessionMatcher` being used in further assertions.

I changed pulling condition to be more strict:
```diff
// Get RUM Sessions with expected number of View visits and Resources
let rumRequests = try rumServerSession.pullRecordedRequests(timeout: dataDeliveryTimeout) { requests in
+    let session = try RUMSessionMatcher.singleSession(from: requests)
+    let hasAllViews = session?.viewVisits.count == 2
+    let hasAllResources = session?.resourceEventMatchers.count == 4
+    let hasAllErrors = session?.errorEventMatchers.count == 1
+    return hasAllViews && hasAllResources && hasAllErrors
}
```

To avoid this sort of problems, I also updated debug description for `RUMSessionMatcher` and now we log it for each CI run. In result, we have a nice description of RUM session attached to each test execution in CI app, which will help debugging similar issues next time.

### Review checklist

### DIFF
--- a/Tests/DatadogIntegrationTests/Scenarios/RUM/RUMResourcesScenarioTests.swift
+++ b/Tests/DatadogIntegrationTests/Scenarios/RUM/RUMResourcesScenarioTests.swift
@@ -61,9 +61,9 @@ class RUMResourcesScenarioTests: IntegrationTests, RUMCommonAsserts {
         let firstPartyBadResourceURL = URL(string: "https://foo.bar")!
 
         // Requesting this third party by the app should create the RUM Resource.
-        let thirdPartyGETResourceURL = URL(string: "https://bitrise.io")!
+        let thirdPartyGETResourceURL = URL(string: "https://httpbin.org/get")!
         // Requesting this third party by the app should create the RUM Resource.
-        let thirdPartyPOSTResourceURL = URL(string: "https://bitrise.io/about")!
+        let thirdPartyPOSTResourceURL = URL(string: "https://httpbin.org/post")!
 
         let app = ExampleApplication()
         app.launchWith(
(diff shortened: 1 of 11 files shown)


Tests/DatadogIntegrationTests/Scenarios/RUM/RUMResourcesScenarioTests.swift
```diff
@@ -113,13 +113,17 @@ class RUMResourcesScenarioTests: IntegrationTests, RUMCommonAsserts {
...
+            let hasAllResources = session?.resourceEventMatchers.count == 4
+            let hasAllErrors = session?.errorEventMatchers.count == 1
+            return hasAllViews && hasAllResources && hasAllErrors
         }
 
         assertRUM(requests: rumRequests)
 
         let session = try XCTUnwrap(try RUMSessionMatcher.singleSession(from: rumRequests))
-        XCTAssertEqual(session.viewVisits.count, 2)
+        sendCIAppLog(session)
```
ncreated, line 126: It sends the log that is associated to CI App's test run:
Screenshot 2021-12-09 at 14 16 56


Sample description looks this:
Screenshot 2021-12-09 at 13 49 35

maxep (approved):
LGTM 👍

PR #683, opened 2021-12-08
RUMM-1791 WebEventConsumers added

buranmert:
### What and why?

Datadog `browser-sdk` integration will send log and RUM events to the native SDK and these events will be decorated with the contextual information of the native app.
That way any app with web view can display its users' session in the same timeline, native + webview.

### How?

Once the message is passed from `browser-sdk`, a JS script will catch it and pass it to either `WebLogEventConsumer` or `WebRUMEventConsumer`.
Then they will add information to events and persist them in the filesystem where they wait to be uploaded to Datadog.

### Review checklist

### DIFF
--- a/Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
+++ b/Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
@@ -18,9 +18,9 @@ public extension WKUserContentController {
 
         let contextProvider = (Global.rum as? RUMMonitor)?.contextProvider
 
-        var logEventConsumer: WebLogEventConsumer? = nil
+        var logEventConsumer: DefaultWebLogEventConsumer? = nil
         if let loggingFeature = LoggingFeature.instance {
-            logEventConsumer = WebLogEventConsumer(
+            logEventConsumer = DefaultWebLogEventConsumer(
                 userLogsWriter: loggingFeature.storage.writer,
                 internalLogsWriter: InternalMonitoringFeature.instance?.logsStorage.writer,
                 dateCorrector: loggingFeature.dateCorrector,
@@ -30,9 +30,9 @@ public extension WKUserContentController {
             )
         }
 
-        var rumEventConsumer: WebRUMEventConsumer? = nil
+        var rumEventConsumer: DefaultWebRUMEventConsumer? = nil
         if let rumFeature = RUMFeature.instance {
-            rumEventConsumer = WebRUMEventConsumer(
+            rumEventConsumer = DefaultWebRUMEventConsumer(
                 dataWriter: rumFeature.storage.writer,
                 dateCorrector: rumFeature.dateCorrector,
                 contextProvider: contextProvider
--- a/Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
+++ b/Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
@@ -79,10 +79,10 @@ public extension WKUserContentController {
     }
 }
 
-private class DatadogMessageHandler: NSObject, WKScriptMessageHandler {
+internal class DatadogMessageHandler: NSObject, WKScriptMessageHandler {
     static let name = "DatadogEventBridge"
     private let eventBridge: WebEventBridge
-    private let queue = DispatchQueue(
+    let queue = DispatchQueue(
         label: "com.datadoghq.JSEventBridge",
         target: .global(qos: .userInteractive)
     )
--- a/Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
+++ b/Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
@@ -33,7 +33,6 @@ public extension WKUserContentController {
             rumEventConsumer = WebRUMEventConsumer(
                 dataWriter: rumFeature.storage.writer,
                 dateCorrector: rumFeature.dateCorrector,
-                webRUMEventMapper: WebRUMEventMapper(),
                 contextProvider: contextProvider
             )
         }
--- a/Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
+++ b/Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
@@ -24,7 +24,9 @@ public extension WKUserContentController {
                 userLogsWriter: loggingFeature.storage.writer,
                 internalLogsWriter: InternalMonitoringFeature.instance?.logsStorage.writer,
                 dateCorrector: loggingFeature.dateCorrector,
-                rumContextProvider: contextProvider
+                rumContextProvider: contextProvider,
+                applicationVersion: loggingFeature.configuration.common.applicationVersion,
+                environment: loggingFeature.configuration.common.environment
             )
         }
 
--- a/Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
+++ b/Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
@@ -17,18 +17,26 @@ public extension WKUserContentController {
         let bridgeName = DatadogMessageHandler.name
 
         let contextProvider = (Global.rum as? RUMMonitor)?.contextProvider
-        let logEventConsumer = WebLogEventConsumer(
-            userLogsWriter: LoggingFeature.instance?.storage.writer,
-            internalLogsWriter: InternalMonitoringFeature.instance?.logsStorage.writer,
-            dateCorrector: LoggingFeature.instance?.dateCorrector,
-            rumContextProvider: contextProvider
-        )
-        let rumEventConsumer = WebRUMEventConsumer(
-            dataWriter: RUMFeature.instance?.storage.writer,
-            dateCorrector: RUMFeature.instance?.dateCorrector,
-            webRUMEventMapper: WebRUMEventMapper(),
-            contextProvider: contextProvider
-        )
+
+        var logEventConsumer: WebLogEventConsumer? = nil
+        if let loggingFeature = LoggingFeature.instance {
+            logEventConsumer = WebLogEventConsumer(
+                userLogsWriter: loggingFeature.storage.writer,
+                internalLogsWriter: InternalMonitoringFeature.instance?.logsStorage.writer,
+                dateCorrector: loggingFeature.dateCorrector,
+                rumContextProvider: contextProvider
+            )
+        }
+
+        var rumEventConsumer: WebRUMEventConsumer? = nil
+        if let rumFeature = RUMFeature.instance {
+            rumEventConsumer = WebRUMEventConsumer(
+                dataWriter: rumFeature.storage.writer,
+                dateCorrector: rumFeature.dateCorrector,
+                webRUMEventMapper: WebRUMEventMapper(),
+                contextProvider: contextProvider
+            )
+        }
 
         let messageHandler = DatadogMessageHandler(
             eventBridge: WebEventBridge(
--- a/Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
+++ b/Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
@@ -7,17 +7,6 @@
 import Foundation
 import WebKit
 
-internal protocol WebRUMEventContextProviding {
-    var context: RUMContext? { get }
-}
-
-internal class WebRUMEventContextProvider: WebRUMEventContextProviding {
-    var context: RUMContext? {
-        // TODO: RUMM-1786 implement web event context provider
-        return nil
-    }
-}
-
 // TODO: RUMM-1794 rename the methods
 public extension WKUserContentController {
     func addDatadogMessageHandler(allowedWebViewHosts: Set<String>) {
@@ -27,8 +16,13 @@ public extension WKUserContentController {
     internal func __addDatadogMessageHandler(allowedWebViewHosts: Set<String>, hostsSanitizer: HostsSanitizing) {
         let bridgeName = DatadogMessageHandler.name
 
-        let contextProvider = WebRUMEventContextProvider()
-        let logEventConsumer = WebLogEventConsumer()
+        let contextProvider = (Global.rum as? RUMMonitor)?.contextProvider
+        let logEventConsumer = WebLogEventConsumer(
+            userLogsWriter: LoggingFeature.instance?.storage.writer,
+            internalLogsWriter: InternalMonitoringFeature.instance?.logsStorage.writer,
+            dateCorrector: LoggingFeature.instance?.dateCorrector,
+            rumContextProvider: contextProvider
+        )
         let rumEventConsumer = WebRUMEventConsumer(
             dataWriter: RUMFeature.instance?.storage.writer,
             dateCorrector: RUMFeature.instance?.dateCorrector,
--- a/Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
+++ b/Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
@@ -7,6 +7,17 @@
 import Foundation
 import WebKit
 
+internal protocol WebRUMEventContextProviding {
+    var context: RUMContext? { get }
+}
+
+internal class WebRUMEventContextProvider: WebRUMEventContextProviding {
+    var context: RUMContext? {
+        // TODO: RUMM-1786 implement web event context provider
+        return nil
+    }
+}
+
 // TODO: RUMM-1794 rename the methods
 public extension WKUserContentController {
     func addDatadogMessageHandler(allowedWebViewHosts: Set<String>) {
@@ -16,10 +27,19 @@ public extension WKUserContentController {
     internal func __addDatadogMessageHandler(allowedWebViewHosts: Set<String>, hostsSanitizer: HostsSanitizing) {
         let bridgeName = DatadogMessageHandler.name
 
+        let contextProvider = WebRUMEventContextProvider()
+        let logEventConsumer = WebLogEventConsumer()
+        let rumEventConsumer = WebRUMEventConsumer(
+            dataWriter: RUMFeature.instance?.storage.writer,
+            dateCorrector: RUMFeature.instance?.dateCorrector,
+            webRUMEventMapper: WebRUMEventMapper(),
+            contextProvider: contextProvider
+        )
+
         let messageHandler = DatadogMessageHandler(
             eventBridge: WebEventBridge(
-                logEventConsumer: WebLogEventConsumer(),
-                rumEventConsumer: WebRUMEventConsumer()
+                logEventConsumer: logEventConsumer,
+                rumEventConsumer: rumEventConsumer
             )
         )
         add(messageHandler, name: bridgeName)
@@ -73,9 +93,11 @@ private class DatadogMessageHandler: NSObject, WKScriptMessageHandler {
         _ userContentController: WKUserContentController,
         didReceive message: WKScriptMessage
     ) {
+        // message.body must be called within UI thread
+        let messageBody = message.body
         queue.async {
             do {
-                try self.eventBridge.consume(message.body)
+                try self.eventBridge.consume(messageBody)
             } catch {
                 userLogger.error("🔥 Web Event Error: \(error)")
             }
--- a/Sources/Datadog/FeaturesIntegration/WebView/WebEventBridge.swift
+++ b/Sources/Datadog/FeaturesIntegration/WebView/WebEventBridge.swift
@@ -9,7 +9,7 @@ import Foundation
 internal typealias JSON = [String: Any]
 
 internal protocol WebEventConsumer {
-    func consume(event: JSON, eventType: String)
+    func consume(event: JSON, eventType: String) throws
 }
 
 internal enum WebEventError: Error, Equatable {
@@ -47,9 +47,9 @@ internal class WebEventBridge {
         }
 
         if eventType == Constants.eventTypeLog {
-            logEventConsumer.consume(event: wrappedEvent, eventType: eventType)
+            try logEventConsumer.consume(event: wrappedEvent, eventType: eventType)
         } else {
-            rumEventConsumer.consume(event: wrappedEvent, eventType: eventType)
+            try rumEventConsumer.consume(event: wrappedEvent, eventType: eventType)
         }
     }
 
--- a/Sources/Datadog/FeaturesIntegration/WebView/WebEventBridge.swift
+++ b/Sources/Datadog/FeaturesIntegration/WebView/WebEventBridge.swift
@@ -24,6 +24,7 @@ internal class WebEventBridge {
         static let eventTypeKey = "eventType"
         static let eventKey = "event"
         static let eventTypeLog = "log"
+        static let eventTypeInternalLog = "internal_log"
     }
 
     private let logEventConsumer: WebEventConsumer?
@@ -46,10 +47,19 @@ internal class WebEventBridge {
             throw WebEventError.missingKey(key: Constants.eventKey)
         }
 
-        if eventType == Constants.eventTypeLog {
-            try logEventConsumer?.consume(event: wrappedEvent, eventType: eventType)
+        if eventType == Constants.eventTypeLog ||
+            eventType == Constants.eventTypeInternalLog {
+            if let consumer = logEventConsumer {
+                try consumer.consume(event: wrappedEvent, eventType: eventType)
+            } else {
+                userLogger.warn("A WebView log is lost because Logging is disabled in iOS SDK")
+            }
         } else {
-            try rumEventConsumer?.consume(event: wrappedEvent, eventType: eventType)
+            if let consumer = rumEventConsumer {
+                try consumer.consume(event: wrappedEvent, eventType: eventType)
+            } else {
+               userLogger.warn("A WebView RUM event is lost because RUM is disabled in iOS SDK")
+           }
         }
     }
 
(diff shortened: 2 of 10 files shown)


Sources/Datadog/FeaturesIntegration/WebView/WebRUMEventConsumer.swift
```diff
@@ -6,8 +6,91 @@
...
+    private func getOffset(viewID: String) -> Offset {
+        var offset = offsets[viewID]
+        if offset == nil {
+            offset = dateCorrector?.currentCorrection.serverTimeOffset
+            offsets[viewID] = offset
+        }
+
+        purgeOffsets()
+        return offset ?? 0.0
+    }
```
buranmert, line 90: @mariusc83 is `offsets` an optimization or an obligation? android uses an ordered hashmap data structure which Swift doesn't have; if this is not necessary, i will just read `serverTimeOffset` whenever it's needed instead of caching it. 
is this okay?

Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
```diff
@@ -7,6 +7,17 @@
...
+internal protocol WebRUMEventContextProviding {
+    var context: RUMContext? { get }
+}
+
+internal class WebRUMEventContextProvider: WebRUMEventContextProviding {
+    var context: RUMContext? {
+        // TODO: RUMM-1786 implement web event context provider
+        return nil
+    }
+}
```
buranmert, line 19: here is what android has: [WebRumEventContextProvider](https://github.com/DataDog/dd-sdk-android/blob/feature/hybrid-application/dd-sdk-android/src/main/kotlin/com/datadog/android/rum/webview/WebRumEventContextProvider.kt)

this new type is probably not needed, i guess we can achieve this by injecting `Global.rum.contextProvider` if RUM is enabled.

ncreated: Yes, we already have this thing in `dd-sdk-ios` and it is abstracted by `RUMContextProvider` interface. I don't see reason for duplicating it by another `WebRUMEventContextProvider` or `WebRUMEventContextProviding` constructs, do we?

Sources/Datadog/FeaturesIntegration/WebView/WebRUMEventConsumer.swift
ncreated: Native SDK time will always be different than Browser SDK time, hence we need to correct event times using the same source of time. There are two particular cases:
* browser views time must be corrected, to avoid time skew in RUM session timeline;
* browser actions / resources / errors time must be corrected accordingly to the correction in their parent view, to avoid time skew in RUM view timeline.

If we don't correct these times, both RUM session timeline and RUM view timeline in web-ui will be skewed and events will be misaligned. This cannot be ignored, as we work with nanosecond precision.

buranmert: @ncreated the question here is if we need to cache/store `serverTimeOffset` values in `offsets` dictionary?
i suggest, instead of keeping the recently read values in `offsets` dict, reading it directly from `dateCorrector?.currentCorrection.serverTimeOffset` whenever needed.

ncreated: Problem is `dateCorrector?.currentCorrection.serverTimeOffset` might return different offset each time ⚠️. And like I described, we need to apply view's offset to its child events. Hence, we need to cache it somehow 👍. Dictionary is one one way - we could key it by `view.id` which should already be `Hashable`. Question is if/how we will remove things from this cache.

buranmert: conclusion: `offsets` dictionary is essential. resolved.

Sources/Datadog/FeaturesIntegration/WebView/WebRUMEventConsumer.swift
```diff
@@ -6,8 +6,92 @@
...
+        }
+        let offset = dateCorrector?.currentCorrection.serverTimeOffset ?? 0.0
+        viewIDOffsetPairs.insert((viewID: viewID, offset: offset), at: 0)
+        return offset
+    }
+
+    private func purgeOffsets() {
+        while viewIDOffsetPairs.count > 3 {
+            _ = viewIDOffsetPairs.popLast()
+        }
```
maxep, line 95: What are you solving with this cache of offsets? 🤔 The server time offset can be considered as constant, and if it's not (e.g. during sync phase) the last offset should be use.

Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
```diff
@@ -16,10 +16,24 @@ public extension WKUserContentController {
...
+            internalLogsWriter: InternalMonitoringFeature.instance?.logsStorage.writer,
+            dateCorrector: LoggingFeature.instance?.dateCorrector,
+            rumContextProvider: contextProvider
+        )
+        let rumEventConsumer = WebRUMEventConsumer(
+            dataWriter: RUMFeature.instance?.storage.writer,
+            dateCorrector: RUMFeature.instance?.dateCorrector,
+            webRUMEventMapper: WebRUMEventMapper(),
+            contextProvider: contextProvider
+        )
```
ncreated, line 31: Same here, all these components must be injected, so we can mock it in tests. Also, there are no tests for `WKUserContentController+Datadog` that would guarantee its all consistent.

Sources/Datadog/FeaturesIntegration/WebView/WebRUMEventConsumer.swift
```diff
@@ -6,8 +6,92 @@
...
+    private let dataWriter: Writer?
+    private let dateCorrector: DateCorrectorType?
+    private let webRUMEventMapper: WebRUMEventMapper?
+    private let contextProvider: RUMContextProvider?
+
+    init(
+        dataWriter: Writer?,
+        dateCorrector: DateCorrectorType?,
+        webRUMEventMapper: WebRUMEventMapper?,
+        contextProvider: RUMContextProvider?
```
ncreated, line 22: Because of these optionals, we must struggle with optionality in this class implementation. I think it doesn't make sense to instantiate `WebRUMEventConsumer` if we don't have any of these values. Also, each optional adds to the number of possible code execution branches, making it more error prone.

Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
```diff
@@ -16,10 +16,24 @@ public extension WKUserContentController {
     internal func __addDatadogMessageHandler(allowedWebViewHosts: Set<String>, hostsSanitizer: HostsSanitizing) {
         let bridgeName = DatadogMessageHandler.name
 
+        let contextProvider = (Global.rum as? RUMMonitor)?.contextProvider
```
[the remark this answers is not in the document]
buranmert: we can inject it but as we can't unit test this method, i'm not sure what we gain by injecting it

do you mean the user should provide `contextProvider` in public API? (if so, `contextProvider` is an internal type so this shouldn't be possible)

Sources/Datadog/FeaturesIntegration/WebView/WebRUMEventConsumer.swift
buranmert: i did ask almost the same question :) 
this was discussed here: https://github.com/DataDog/dd-sdk-ios/pull/683#discussion_r765099800

Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
buranmert: what do you mean by consistent?
as discussed before (maybe in Slack though) i couldn't find a way to unit test the behavior of this component, so i can't see where we can make use of mocks.

Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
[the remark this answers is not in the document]
maxep: No, not from the user but from another source which is less transverse. This might not be possible, but assuming that a global instance is of some type could easily break.

Sources/Datadog/FeaturesIntegration/WebView/WebRUMEventConsumer.swift
maxep: Where is the discussion?

maxep: Ok, I get it! we need all events to be in synced with the parent view event.
Note: the sync phase of Kronos is too long IMO, that's something we should revisit!

Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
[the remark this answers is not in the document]
ncreated: In other places we solve this problem by [injecting `RUMMonitor` from its `initialize()` method](https://github.com/DataDog/dd-sdk-ios/blob/0ea2b5f6307b6005ed0cc1980754ec6d2dd3cc41/Sources/Datadog/RUMMonitor.swift#L167-L168) to all integrated components. Later, this is tested respectively in each component's tests: [`URLSessionAutoInstrumentationTests`](https://github.com/DataDog/dd-sdk-ios/blob/0ea2b5f6307b6005ed0cc1980754ec6d2dd3cc41/Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/URLSessionAutoInstrumentationTests.swift#L39-L61) and [`RUMInstrumentationTests`](https://github.com/DataDog/dd-sdk-ios/blob/0ea2b5f6307b6005ed0cc1980754ec6d2dd3cc41/Tests/DatadogTests/Datadog/RUM/Instrumentation/RUMInstrumentationTests.swift#L23-L92).

I think the key element for enabling `WKUserContentController+Datadog` testability is to create separate component for locating involved services, e.g.:
```swift
internal final class WebviewsFeature {
   /// Single, shared instance of `WebviewsFeature`.
   internal static var instance: WebviewsFeature?

   let messageHandler: WKScriptMessageHandler

   func createJSBridge(for allowedWebViewHosts: Set<String>) -> WKUserScript { /*...*/ }
}
```
Going that way the the implementation of `WKUserContentController` extension will be reduced to only:
```swift
public extension WKUserContentController {
   func addDatadogMessageHandler(allowedWebViewHosts: Set<String>) {
      guard let webviewsFeature = WebviewsFeature.instance else {
         // What does it mean to have no WebviewsFeature.instance?
         // e.g. If both Logging and RUM features are disabled, will this integration be
         // still functional? We may want to log user configuration error from here.
         return
      }
      add(webviewsFeature.messageHandler, name: webviewsFeature.bridgeName)
      addUserScript(webviewsFeature.createJSBridge(for: allowedWebViewHosts))
   }
}
```
and we will be able to test everything as part of `WebviewsFeature` configuration.

Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
ncreated: By consistency I mean all the things that may go wrong either by regression on our end or by wrong configuration from the user. Some examples:
* misconfiguration:
   * enabling this integration while RUM and Logging features are disabled
   * enabling this while no `Global.rum` is registered

* regression:
   * given Logging/RUM enabled, when receiving event from Browser SDK, it gets passed to the right writer;

Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
```diff
@@ -16,10 +16,32 @@ public extension WKUserContentController {
...
+
+        var logEventConsumer: WebLogEventConsumer? = nil
+        if let loggingFeature = LoggingFeature.instance {
+            logEventConsumer = WebLogEventConsumer(
+                userLogsWriter: loggingFeature.storage.writer,
+                internalLogsWriter: InternalMonitoringFeature.instance?.logsStorage.writer,
```
ncreated, line 25: `logsStorage.writer` is a [consent-aware data writer](https://github.com/DataDog/dd-sdk-ios/blob/fc693d3089d93ba3bc8e5d3ef21e841b6e973125/Sources/Datadog/Core/Feature.swift#L91). Do we need to consider tracking consent for events received from Browser SDK? Shouldn't we assume that all these were already authorized in Browser SDK and we can bypass consent consideration in native SDK?

Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
```diff
@@ -16,10 +16,32 @@ public extension WKUserContentController {
     internal func __addDatadogMessageHandler(allowedWebViewHosts: Set<String>, hostsSanitizer: HostsSanitizing) {
         let bridgeName = DatadogMessageHandler.name
 
+        let contextProvider = (Global.rum as? RUMMonitor)?.contextProvider
+
+        var logEventConsumer: WebLogEventConsumer? = nil
+        if let loggingFeature = LoggingFeature.instance {
+            logEventConsumer = WebLogEventConsumer(
+                userLogsWriter: loggingFeature.storage.writer,
```
ncreated, line 24: `loggingFeature.storage.writer` is a [consent-aware data writer](https://github.com/DataDog/dd-sdk-ios/blob/fc693d3089d93ba3bc8e5d3ef21e841b6e973125/Sources/Datadog/Core/Feature.swift#L91). Do we need to consider tracking consent for events received from Browser SDK? Shouldn't we assume that all these were already authorized in Browser SDK and we can bypass consent consideration in native SDK?

buranmert: you may want to discuss with @mariusc83 

as we discussed today, the only tricky situation is consent is given in browser-sdk but not in native sdk as then the browser events will be lost. but i personally see this scenario as an edge case or one in which the developer should take consent in browser-sdk and forward it to native sdk.
in short, i think it makes sense to use `context-aware data writer` here

Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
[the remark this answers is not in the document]
buranmert: i leave `Global.rum` usage as there may be no Event Bridge installed at the time of `RUMMonitor.initialize`, `Global.rum` seems like the best option for now.

the tests requested by @ncreated are also added, in another way.

Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
buranmert: - enabling this integration while RUM and Logging features are disabled
- enabling this while no Global.rum is registered

these can be legit use cases, i don't know how to handle these cases. wdyt @mariusc83 ?

Sources/Datadog/FeaturesIntegration/WebView/WebLogEventConsumer.swift
```diff
@@ -7,7 +7,70 @@
...
+    private let internalLogsWriter: Writer?
+    private let dateCorrector: DateCorrectorType
+    private let rumContextProvider: RUMContextProvider?
+    private let jsonDecoder = JSONDecoder()
+
+    private lazy var ddTags: String = {
+        let versionKey = LogEventEncoder.StaticCodingKeys.applicationVersion.rawValue
+        let versionValue = LoggingFeature.instance?.configuration.common.applicationVersion ?? ""
+        let envKey = LogEventEncoder.StaticCodingKeys.environment.rawValue
+        let envValue = LoggingFeature.instance?.configuration.common.environment ?? ""
```
ncreated, line 30: No reason to consider `envValue` and `versionValue` if `LoggingFeature.instance` is `nil` (if the feature is disabled). Can't we use dependency injection to make the solution depend explicitly on `LoggingFeature` instance? It will be simpler (less control flows), less error prone (no optionals) and will result with better tests coverage.

Sources/Datadog/FeaturesIntegration/WebView/WebLogEventConsumer.swift
```diff
@@ -7,7 +7,70 @@
...
+        }
+
+        let jsonData = try JSONSerialization.data(withJSONObject: mutableEvent, options: [])
+        let encodableEvent = try jsonDecoder.decode(CodableValue.self, from: jsonData)
+
+        if eventType == Constants.logEventType {
+            userLogsWriter.write(value: encodableEvent)
+        } else if eventType == Constants.internalLogEventType {
+            internalLogsWriter?.write(value: encodableEvent)
+        }
```
ncreated, line 74: What if the event is not recognised? We already captured it in the native SDK, but here we resign from processing it so it will be lost. We should at least inform the user on dropping an event and consider adding internal monitoring to spot these cases if they might come from our fault.

Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
ncreated: @mariusc83 what's our take on this? In any case, we should avoid inconsistency here - either by relying only on Browser SDK consent (when we receive an event, we know it was already authorized) or by adding additional logs to inform the user on inconsistency (that Browser SDK events will be lost).

ncreated (changes requested):
Needs answers on tracking consent for webview events and more logs on eventual inconsistency of the solution. Currently we only addresses the happy scenario (both Logs and RUM enabled, tracking consent granted and receiving only known events).

Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
buranmert: @bcaudan told me `browser-sdk` doesn't have tracking consent feature and it makes sense to go with the consent from native SDK as the events will eventually be shown in the native RUM session

Sources/Datadog/FeaturesIntegration/WebView/WebLogEventConsumer.swift
buranmert: ℹ️ idk why github doesn't label this line as `outdated` but it is `outdated`

Sources/Datadog/FeaturesIntegration/WebView/WebEventBridge.swift
```diff
@@ -47,9 +47,9 @@ internal class WebEventBridge {
         }
 
         if eventType == Constants.eventTypeLog {
-            logEventConsumer.consume(event: wrappedEvent, eventType: eventType)
+            try logEventConsumer?.consume(event: wrappedEvent, eventType: eventType)
         } else {
-            rumEventConsumer.consume(event: wrappedEvent, eventType: eventType)
+            try rumEventConsumer?.consume(event: wrappedEvent, eventType: eventType)
         }
```
ncreated, line 53: Because we pass Browser events to Logging / RUM feature storage, in case of Logging or RUM being disabled we will be sending Browser SDK events to void. This will result with loosing observability and entire set of telemetry from webviews, depending on native SDK configuration. This requires additional handling - @mariusc83 @buranmert @xgouchet what's our strategy for this?

I see 3 options:
* we can send events to their own storage - independent from Logging / RUM feature;
* we can inform the user on dropping event due to misconfiguration:
```swift
if eventType == Constants.eventTypeLog {
   if let logEventConsumer = logEventConsumer {
      try logEventConsumer.consume(event: wrappedEvent, eventType: eventType)
   } else {
      userLogger.warn(/* warn that event will be lost because Logging is disabled in native SDK */)
   }
} else {
   if let rumEventConsumer = rumEventConsumer {
      try rumEventConsumer(event: wrappedEvent, eventType: eventType)
   } else {
      userLogger.warn(/* warn that event will be lost because RUM is disabled in native SDK */)
   }
}
```
* we can register selectively for receiving only Logs / only RUM / both Logs and RUM events from Browser SDK.

Sources/Datadog/FeaturesIntegration/WebView/WebLogEventConsumer.swift
ncreated:
```swift
userLogger.error("🔥 Invalid Web Event Type: \(eventType)")
```
How will we know if there is an issue in this system? Say, some Browser SDK version sends unrecognised event type, don't we want to be informed that we drop entire range of events, potentially destroying user's observability in critical flow? I think this case deserves Internal Monitoring observability as well.

Sources/Datadog/FeaturesIntegration/WebView/WebRUMEventConsumer.swift
```diff
@@ -7,7 +7,81 @@
...
+            let errorEvent = try jsonDecoder.decode(RUMErrorEvent.self, from: eventData)
+            let mappedErrorEvent = mapIfNeeded(dataModel: errorEvent, context: rumContext, offset: getOffset(viewID: errorEvent.view.id))
+            write(mappedErrorEvent)
+        case "long_task":
+            let longTaskEvent = try jsonDecoder.decode(RUMLongTaskEvent.self, from: eventData)
+            let mappedLongTaskEvent = mapIfNeeded(dataModel: longTaskEvent, context: rumContext, offset: getOffset(viewID: longTaskEvent.view.id))
+            write(mappedLongTaskEvent)
+        default:
+            userLogger.error("🔥 Web RUM Event Error - Unknown event type: \(eventType)")
+        }
```
ncreated, line 52: What if we add a new RUM event type in Browser SDK version `X` (loaded by the webview), but mobile app is using old version of iOS SDK with this outdated `switch eventType` selection? It seems that we will be dropping all new events. // cc @mariusc83 @xgouchet do we mitigate this on Android?

ncreated (changes requested):

Sources/Datadog/FeaturesIntegration/WebView/WebRUMEventConsumer.swift
ncreated: Maybe instead of decoding events to concrete types (`RUMViewEvent`, `RUMActionEvent`, ...) we should only decode their mutable parts to single type, e.g.:
```swift
struct RUMBrowserEvent: Codable {
   // MARK: - Mutable, known properties
   var applicationID: String
   var sessionID: String

   // MARK: - Immutable, unknown properties
   let otherProperties: [String: Codable]
}
```
This way we can overwrite only `sessionID` and `applicationID` without changing anything else to the event:
```swift
/*

Say, `eventData` is:
{
   "application": { "id": "browser-application-id" },
   "session": { "id": "browser-session-id" },
   "other_stuff" {
      "foo": 100,
      "bar": "bizz",
   },
   "more_stuff": "fuzz"
}

*/

// Then mutating it:
var browserRUMEvent = try jsonDecoder.decode(RUMBrowserEvent.self, from: eventData)
browserRUMEvent.applicationID = rumContext.applicationID
browserRUMEvent.sessionID = rumContext.sessionID

// ... and serializing it again:
write(browserRUMEvent)

/*

... will alter only 'application.id' and 'session.id':
{
   "application": { "id": "native-application-id" },
   "session": { "id": "native-session-id" },
   "other_stuff" {
      "foo": 100,
      "bar": "bizz",
   },
   "more_stuff": "fuzz"
}

*/
```

With this approach we **won't be depending on concrete event** types and only on the minimal part of schema shared by **all events**.

buranmert: good idea, let's do this.

we don't even need a new type for this, we can manipulate Web RUM event JSON and convert it to `CodableValue` and then write to the disk.

ncreated: > we don't even need a new type for this, we can manipulate Web RUM event JSON and convert it to `CodableValue` and then write to the disk like Web Logs


Sounds good 👍

Sources/Datadog/FeaturesIntegration/WebView/WebEventBridge.swift
buranmert: ℹ️ we decided to go with `userLogger.warn` option at today's daily

buranmert: @ncreated should we warn the user every time a new event is received (that might spam their console)?
or only once when `WebEventConsumer` is created?

Sources/Datadog/FeaturesIntegration/WebView/WebLogEventConsumer.swift
buranmert: `WebLogEventConsumer` protocol changed in a way that eliminates this case

Sources/Datadog/FeaturesIntegration/WebView/WebRUMEventConsumer.swift
```diff
@@ -6,8 +6,93 @@
...
+        let encodableEvent = try jsonDecoder.decode(CodableValue.self, from: jsonData)
+
+        dataWriter.write(value: encodableEvent)
+    }
+
+    private func map(event: JSON, with context: RUMContext?) -> JSON {
+        guard let context = context,
+              context.sessionID != .nullUUID else {
+            return event
+        }
```
ncreated, line 40: _"Given native RUM session rejected by sampler, when receiving event from WebView, it is sent anyway"_ seems to be critical (it impacts user billing) and not very obvious behaviour (in case of RUM disabled we drop events, but we send events no matter of RUM sampling). I don't see it covered in unit tests, meaning we can easily overlook it and change unawarely.

ncreated (approved):

Sources/Datadog/FeaturesIntegration/WebView/WebRUMEventConsumer.swift
```diff
@@ -6,8 +6,93 @@
...
+
+        let jsonData = try JSONSerialization.data(withJSONObject: mappedEvent, options: [])
+        let encodableEvent = try jsonDecoder.decode(CodableValue.self, from: jsonData)
+
+        dataWriter.write(value: encodableEvent)
+    }
+
+    private func map(event: JSON, with context: RUMContext?) -> JSON {
+        guard let context = context,
+              context.sessionID != .nullUUID else {
```
buranmert, line 38: `context.sessionID == .nullID` when the session is sampled-out

PR #685, opened 2021-12-10
RUMM-1765 Do not start "ApplicationLaunch" view when app launches in background - part 2

ncreated:
### What and why?

📦 This PR limits _application launch events tracking_ feature to only consider events received while app is in foreground. It adds necessary implementation to not start "ApplicationLaunch" view if app is not in `active` state.

When app launches in background, its events can still be tracked if [Background Events Tracking](https://github.com/DataDog/dd-sdk-ios/blob/b28016175691665c9c2e1d3d867e90d6f4437ab2/Sources/Datadog/DatadogConfiguration.swift#L678-L690) feature is enabled. This makes a clear separation of "ApplicationLaunch" and "Background" activities.

### How?

Determining app state is done using existing `AppStateListener` - it is now injected to RUM feature as well. It is used to check the app state and consider starting either:
* "ApplicationLaunch" view - if app receives off-view events in `active` state;
* "Background" view - if app receives off-view events in not `active` state and [BET](https://github.com/DataDog/dd-sdk-ios/blob/b28016175691665c9c2e1d3d867e90d6f4437ab2/Sources/Datadog/DatadogConfiguration.swift#L678-L690) is enabled.

Below, list of cases covered with new and existing (#677) tests.

---

#### ✅ (1) Receive off-view events when app launches in foreground and BET is enabled.

The "ApplicationLaunch" view spans from SDK init to starting another view.

SC1

#### ✅ (2) Receive off-view events when app launches in background and BET is enabled.

There is no "ApplicationLaunch" view. Instead, "Background" view is created and it starts at the time of first event.

SC2

#### ✅ (3) Receive events in background when BET is enabled.

No change - receiving off-view event after app goes to background starts the "Background" view at the time of the event.

SC3

#### ✅ (4) Receive off-view events when app launches in foreground and BET is disabled.

Same as scenario 1, the "ApplicationLaunch" view tracks off-view event and the view starts on SDK init.

SC4

#### ✅ (5) Receive events in background when BET is disabled.

Nothing is tracked in this case. If BET is disabled, we do not create any view for tracking off-view background events.

#### ✅ (6) Receive events in background when BET is disabled.

Only foreground events are tracked in user-defined view. "Background" view is not started as BET is disabled.

SC6

### Review checklist

### DIFF
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
@@ -37,6 +37,9 @@ internal class RUMApplicationScope: RUMScope, RUMContextProvider {
     /// RUM Sessions sampling rate.
     internal let samplingRate: Float
 
+    /// Time of SDK initialization, measured in device date.
+    internal let sdkInitDate: Date
+
     /// Automatically detect background events
     internal let backgroundEventTrackingEnabled: Bool
 
@@ -48,10 +51,12 @@ internal class RUMApplicationScope: RUMScope, RUMContextProvider {
         rumApplicationID: String,
         dependencies: RUMScopeDependencies,
         samplingRate: Float,
+        sdkInitDate: Date,
         backgroundEventTrackingEnabled: Bool
     ) {
         self.dependencies = dependencies
         self.samplingRate = samplingRate
+        self.sdkInitDate = sdkInitDate
         self.backgroundEventTrackingEnabled = backgroundEventTrackingEnabled
         self.context = RUMContext(
             rumApplicationID: rumApplicationID,
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
@@ -10,6 +10,7 @@ internal typealias RUMSessionListener = (String, Bool) -> Void
 
 /// Injection container for common dependencies used by all `RUMScopes`.
 internal struct RUMScopeDependencies {
+    let appStateListener: AppStateListening
     let userInfoProvider: RUMUserInfoProvider
     let launchTimeProvider: LaunchTimeProviderType
     let connectivityInfoProvider: RUMConnectivityInfoProvider
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
@@ -37,8 +37,8 @@ internal class RUMApplicationScope: RUMScope, RUMContextProvider {
     /// RUM Sessions sampling rate.
     internal let samplingRate: Float
 
-    /// Time of SDK initialization, measured in device date.
-    internal let sdkInitDate: Date
+    /// The start time of the application, measured in device date. It equals the time of SDK init.
+    private let applicationStartTime: Date
 
     /// Automatically detect background events
     internal let backgroundEventTrackingEnabled: Bool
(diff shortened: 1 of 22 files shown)


Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -171,9 +171,14 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
...
+                if command.actionType == .custom {
+                    // send it instantly without waiting for child events (e.g. resource associated to this action)
+                    sendDiscreteCustomUserAction(on: command)
+                } else {
+                    addDiscreteUserAction(on: command)
+                }
             } else if command.actionType == .custom {
-                // still let it go, just instantly without any dependencies
+                // still let it go, just instantly without waiting for child events (e.g. resource associated to this action)
                 sendDiscreteCustomUserAction(on: command)
```
ncreated, line 182: While testing ALT feature, I found this bug in `.custom` actions tracking. Here, I fix it - now we treat `.custom` actions the same, no matter if added while any other action scope exists or not. I also added test for this case.

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMViewScopeTests.swift
```diff
@@ -548,6 +548,74 @@ class RUMViewScopeTests: XCTestCase {
         XCTAssertEqual(event.model.view.action.count, 2, "View should record 2 actions: non-custom + instant custom")
     }
 
+    func testGivenViewWithPendingAction_whenCustomActionIsAdded_itSendsItInstantly() throws {
```
ncreated, line 551: These two tests ensure we track `.custom` actions properly, no matter if added while any action scope exists or not. In both cases, `.custom` action is sent instantly w/o waiting for their side effects.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -171,9 +171,14 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
...
-                addDiscreteUserAction(on: command)
+                if command.actionType == .custom {
+                    // send it instantly without waiting for child events (e.g. resource associated to this action)
+                    sendDiscreteCustomUserAction(on: command)
+                } else {
+                    addDiscreteUserAction(on: command)
+                }
             } else if command.actionType == .custom {
-                // still let it go, just instantly without any dependencies
+                // still let it go, just instantly without waiting for child events (e.g. resource associated to this action)
```
maxep, line 181: /nit this can be:
```swift
if command.actionType == .custom {
    // send it instantly without waiting for child events (e.g. resource associated to this action)
    sendDiscreteCustomUserAction(on: command)
} else if userActionScope == nil {
    addDiscreteUserAction(on: command)
} else {
    reportActionDropped(type: command.actionType, name: command.name)
}
```

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
```diff
@@ -36,6 +37,9 @@ internal class RUMApplicationScope: RUMScope, RUMContextProvider {
     /// RUM Sessions sampling rate.
     internal let samplingRate: Float
 
+    /// The start time of the application, measured in device date. It equals the time of SDK init.
+    private let applicationStartTime: Date
```
maxep, line 41: I think the name can be confused with application launch time. Can we rename it to something like `scopeStartTime` or `sdkInitTime`?

maxep (approved):
LGTM 👍 I've a couple of non-blocking suggestions

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
ncreated: Indeed, it's simpler 👍. Applied in https://github.com/DataDog/dd-sdk-ios/pull/697/commits/99ffde23d08c9bb0aa681e94fb12e2ce18ff29cf to avoid rebasing.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
ncreated: Yes, makes sense - for that reason I called it `sdkInitTime` in upstream, but in `RUMMonitor` this information was translated to `applicationStartTime`. What proposed could be simpler 👍 - applied in https://github.com/DataDog/dd-sdk-ios/pull/697/commits/99ffde23d08c9bb0aa681e94fb12e2ce18ff29cf to avoid rebasing.

PR #690, opened 2021-12-16
RUMM-1765 Update `CrashContext` with necessary info for handling App Launch crashes - part 3

ncreated:
### What and why?

📦 This PR prepares `CrashContext` to handle crashes during application launch. This is necessary to decide:
* if the crash should be send in "ApplicationLaunch" view (when app crashed during foreground launch);
* if the crash should be send in "Background" view (when app crashed during background launch);
* if the crash should be dropped (e.g. when app crashed during background launch, but Background Events Tracking is disabled OR when crashed session was rejected by sampler).

### How?

First, I isolated the code deciding on "ApplicationLaunch" x "Background" x "no" view to `RUMOffViewEventsHandlingRule`. This logic is quite complex but very important (for customer billing 💸) so it's definitely worth having it separate as it will be used by both RUM and Crash Reporting to take similar decision on handling events or crashes while no active view:

application-launch-events

Second, I updated `CrashContext` with all values necessary for making the decision ☝️ in `RUMOffViewEventsHandlingRule`. This way, when the app restarts upon the crash, we will be able to run exactly the same logic and decide if the crash should be sent in "ApplicationLaunch" view, "Background" view or if it should be dropped (e.g. due to sampling).

Crash handling upon app restart will be delivered in separate, final PR to this topic.

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/System/AppStateListener.swift
+++ b/Sources/Datadog/Core/System/AppStateListener.swift
@@ -89,9 +89,16 @@ internal struct AppStateHistory: Equatable {
     }
 }
 
+/// An observer of `AppStateHistory` value.
+internal typealias AppStateHistoryObserver = ValueObserver
+
 /// Provides history of app foreground / background states.
 internal protocol AppStateListening: AnyObject {
+    /// Last published `AppStateHistory`.
     var history: AppStateHistory { get }
+
+    /// Subscribers observer to be notified on `AppStateHistory` changes.
+    func subscribe<Observer: AppStateHistoryObserver>(_ subscriber: Observer) where Observer.ObservedValue == AppStateHistory
 }
 
 internal class AppStateListener: AppStateListening {
@@ -143,5 +150,11 @@ internal class AppStateListener: AppStateListening {
         var value = publisher.currentValue
         value.changes.append(Snapshot(isActive: true, date: now))
         publisher.publishAsync(value)
+    }
+
+    // MARK: - Managing Subscribers
+
+    func subscribe<Observer: AppStateHistoryObserver>(_ subscriber: Observer) where Observer.ObservedValue == AppStateHistory {
+        publisher.subscribe(subscriber)
     }
 }
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMOffViewEventsHandlingRule.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMOffViewEventsHandlingRule.swift
@@ -1,0 +1,88 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+/// Lightweight representation of current RUM session state, used to compute `RUMOffViewEventsHandlingRule`.
+/// It gets serialized into `CrashContext` for computing the rule upon app process restart after crash.
+internal struct RUMSessionState: Equatable, Codable {
+    /// The session ID. Can be `.nullUUID` if the session was rejected by sampler.
+    let sessionUUID: UUID
+    /// If this is the very first session in the app process (`true`) or was re-created upon timeout (`false`).
+    let isInitialSession: Bool
+    /// If this session has ever tracked any view (used to reason about "application launch" events).
+    let hasTrackedAnyView: Bool
+}
+
+/// The rule for handling RUM events which are tracked while there is no active view.
+///
+/// It isolates the logic behind starting artificial views like "ApplicationLaunch" or "Background". It is used by both RUM and Crash Reporting
+/// to decide on how to track off-view events and crashes.
+internal enum RUMOffViewEventsHandlingRule: Equatable {
+    struct Constants {
+        /// The name of the view created when receiving an event while there is no active view and Background Events Tracking is enabled.
+        static let backgroundViewName = "Background"
+        /// The url of the view created when receiving an event while there is no active view and Background Events Tracking is enabled.
+        static let backgroundViewURL = "com/datadog/background/view"
+        /// The name of the view created when receiving an event before any view was started in the initial session.
+        static let applicationLaunchViewName = "ApplicationLaunch"
+        /// The url of the view created when receiving an event before any view was started in the initial session.
+        static let applicationLaunchViewURL = "com/datadog/application-launch/view"
+    }
+
+    /// Start "ApplicationLaunch" view to track the event.
+    case handleInApplicationLaunchView
+    /// Start "Background" view to track the event.
+    case handleInBackgroundView
+    /// Do not start any view (drop the event).
+    case doNotHandle
+
+    // MARK: - Init
+
+    /// - Parameters:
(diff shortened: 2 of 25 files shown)


Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScopeTests.swift
```diff
@@ -75,91 +75,109 @@ class RUMSessionScopeTests: XCTestCase {
```
ncreated, line 75: Here, I updated `RUMSessionScope` tests to transitively cover all cases in `RUMOffViewEventsHandlingRule` used internally. I took this approach instead of abstracting `RUMOffViewEventsHandlingRule` through interface and mocking it for tests. I wanted to optimise code readability in `RUMSessionScope`, even if the cost was more effort in fuzzy testing.

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMViewScopeTests.swift
```diff
@@ -1139,4 +1139,38 @@ class RUMViewScopeTests: XCTestCase {
         XCTAssertEqual(event.model.view.action.count, 0, "All actions, including ApplicationStart action should be dropped")
         XCTAssertEqual(event.model.dd.documentVersion, 1, "It should record only one view update")
     }
+
+    // MARK: Integration with Crash Context
+
+    func testWhenViewIsStarted_thenItUpdatesLastRUMViewEventInCrashContext() throws {
```
ncreated, line 1145: I added this missing test, because now `RUMWithCrashContextIntegration` is injected through DI and we're able to configure the mock.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMOffViewEventsHandlingRule.swift
```diff
@@ -0,0 +1,88 @@
...
+    case doNotHandle
+
+    // MARK: - Init
+
+    /// - Parameters:
+    ///   - sessionState: RUM session state or `nil` if no session is started
+    ///   - isAppInForeground: if the app is in foreground
+    ///   - isBETEnabled: if Background Events Tracking feature is enabled in SDK configuration
+    init(
+        sessionState: RUMSessionState?,
```
maxep, line 50: Since a event cannot be sent outside of a session, shouldn't we apply this rule only if we have a session state? this would mean making the state non-optional.
And if it's more convenient to keep it optional, why not dropping the event if the session state is `nil`?

Sources/Datadog/Core/System/AppStateListener.swift
```diff
@@ -89,9 +89,16 @@ internal struct AppStateHistory: Equatable {
     }
 }
 
+/// An observer of `AppStateHistory` value.
+internal typealias AppStateHistoryObserver = ValueObserver
```
maxep, line 93: Not sure this has much value 🤔 
The `where` clause in the protocol method is enough to describe the constraint:
```swift
func subscribe<Observer: ValueObserver>(_ subscriber: Observer) where Observer.ObservedValue == AppStateHistory
```

maxep (commented):
Very clear! The doc and the separation of concern applied in `RUMOffViewEventsHandlingRule` help understanding the complexity of this feature.

I'm wondering about the rules applied when session state is nil, I wonder if we shouldn't drop the event instead.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMOffViewEventsHandlingRule.swift
ncreated: This (`sessionState: nil`) is used later in #695 to decide on handling crashes which occurred as the very first event in the app. For example, in case of:
```swift
Datadog.initialize(
   ...
)

fatalError("crash")
```
we don't have `RUMSession` yet created, so its persisted (through `CrashContext`) state will be `nil`.

I've spent a while trying to model the minimal information we need to persist in `CrashContext` for differentiating crashes during app launch in foreground vs background and resulted with this `RUMSessionState`. If you can find a better model, I'd be happy to adopt it and simplify things 🙂 🙌! I'm all for consistency here, making things optional just for convenience might be error-prone.

Sources/Datadog/Core/System/AppStateListener.swift
ncreated: This is very small, so I'll keep it for the sake of consistency with other providers like:
https://github.com/DataDog/dd-sdk-ios/blob/bdccc5e652b8f22dd0c18d0d3657fa53c5451898/Sources/Datadog/Core/Attributes/UserInfoProvider.swift#L10
or
https://github.com/DataDog/dd-sdk-ios/blob/bdccc5e652b8f22dd0c18d0d3657fa53c5451898/Sources/Datadog/Core/System/CarrierInfoProvider.swift#L39
but in general I agree - not much value. This will be surely addressed in V2 architecture by getting rid of this by design.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMOffViewEventsHandlingRule.swift
maxep: I think the model is great and simplifies the rule definition!

In the next PRs, will the `RUMSessionState.sessionUUID` will be leveraged somehow for other purposes than checking if it's sampled ou?
I'm wondering if we could do something like:
```swift
internal struct RUMSessionState: Equatable, Codable {
    let isSampled: Bool
    let isInitialSession: Bool
    let hasTrackedAnyView: Bool

    static let `default` = RUMSessionState(isSampled: false, isInitialSession: true, hasTrackedAnyView: false)
}
```

This way, we could make the state non-optional and clarify the statements in this `init`?

ncreated: > In the next PRs, will the `RUMSessionState.sessionUUID` will be leveraged somehow for other purposes than checking if it's sampled out?

Now I think that in #695 I forgot to handle the case which I spotted when designing `RUMSessionState` 😰. Indeed, it may look now that `sessionUUID` is not necessary and we don't use it in #695 now.

The plan was to use `sessionUUID` in this scenario:
* 1/ SDK has Background Events Tracking **enabled** and views instrumentation **turned ON**;
* 2/ It tracks some views, leading to updating `lastRUMViewEvent` in `CrashContext`;
* 3/ Then the app is sent to background (so it stops the recent view);
* 4/ No other event occurs in the background except the crash.

In such case, because there is no active view (3) and the app crashed in background (4) we should send the crash in "Background" view (1). No additional sampling should be considered and both crash and its "Background" view should be sent to the `sessionUUID` used for tracking other views in this session (2).

That said: no, `isSampled: Bool` wouldn't be enough for `RUMSessionState` - we also need to know its `uuid` for above case ☝️. Knowing that `.nullUUID` means "not sampled", it also simplifies few more edge cases.

I will update next PR (#695) with necessary logic for handling this case 👍. Otherwise the `RUMSessionState` persistence in `CrashContext` doesn't make sense at all.

maxep: Makes sense then 👍

maxep (approved):

PR #691, opened 2021-12-16
Update DateFormatting

winstondu:
### What and why?

Both the `ISO8601Formatter` and the `DateFormatter` type also has the ability to get a date from the string. There's no reason not to include this as well to the protocol

### How?

Adds the method signature for the conforming protocol

### Review checklist


maxep:
Hello @winstondu,
Thank you for your contribution. Indeed, both formatters support both ways of conversion but the SDK is only leveraging the date to string. The requirements might change in the futur but let's keep it simple for now.

PR #692, opened 2021-12-17
RUMM-1805 Set `PLCrashReporter` custom path

maxep:
### What and why?

Configure `PLCrashReporter` with a custom storage location to `/Library/Caches/com.datadoghq.crash-reporting/v1`.

By setting a custom path, we allow other instances of `PLCrashReporter` to work alongside the SDK without altering the reports.

### How?

Sets the `basePath` property of the [PLCrashReporterConfig](https://github.com/microsoft/plcrashreporter/blob/master/Source/PLCrashReporterConfig.h#L185).

The `PLCrashReporter` version has been incremented to `1.10.1` to use the `[PLCrashReporter crashReportPath]` in tests.

### Review checklist

- [X] Make sure each commit and the PR mention the Issue number or JIRA reference

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,2 +1,2 @@
 github "lyft/Kronos" ~> 4.2
-github "microsoft/plcrashreporter" ~> 1.10.0
+github "microsoft/plcrashreporter" ~> 1.10.1
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,2 +1,2 @@
 github "lyft/Kronos" "4.2.1"
-github "microsoft/plcrashreporter" "1.10.0"
+github "microsoft/plcrashreporter" "1.10.1"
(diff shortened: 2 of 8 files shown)


fuzzybinary (approved):

ncreated (changes requested):
Looks all good 👍.

Integration tests might be failing as we also need to update this path [in our persistence helper](https://github.com/DataDog/dd-sdk-ios/blob/9cdb8fe7615949554e8a6af2ddbe2857050fed35/Datadog/Example/Utils/PersistenceHelper.swift#L48-L50) used to ensure clear state for some integration tests.

maxep:
@ncreated I need further investigation... It doesnt seem to be writing any report in the custom dir 🤔

maxep:
I actually let PLCR create the subdirectories by just providing the base path (not absolute string as before)

ncreated (approved):
🚀

PR #696, opened 2021-12-20
RUMM-1765 Fix foreground app state reporting in `AppStateListener` - part 5 🏁

ncreated:
### What and why?

📦 This is the final PR closing topic of _tracking app launch events (and crashes)_. While testing it (with new options added to debug UI 👇), I found a bug 🐞 in `AppStateListener`, which I'm solving in this PR.

Screenshot 2021-12-20 at 16 41 10

### How?

Problem was that [`AppStateListener` was only considering `UIApplication.State.active` as "foreground" application state](https://github.com/DataDog/dd-sdk-ios/blob/ad16182bba3f757406834044bf0d9d1fdcdd3dea/Sources/Datadog/Core/System/AppStateListener.swift#L103), whereas app lifecycle is more complex than that and it can be also in "foreground" while in `.inactive` state:

> [**`applicationState`**](https://developer.apple.com/documentation/uikit/uiapplication/1623003-applicationstate)
> The app is **inactive at launch**, and then is generally in either an active or background state. The app **may become inactive for a short period** — for example, when transitioning between active and background states, when the system presents an alert in front of it, or when the system displays the application switcher.

Because `.inactive` was not considered "foreground", any event tracked during application launch (e.g. from `application(_:didFinishLaunching:)`) was starting "Background" instead of "ApplicationLaunch" view. I fixed this issue by exposing `AppState` directly from `AppStateListener` and letting other components decide by their own on handling particular states.

#### 🏁 Final Testing

With all logic in place, I made last tests, ensuring everything works as expected. Here are some explored scenarios:
* Tracking app launch event in foreground:
foreground-launch-event

* Tracking app launch crash in foreground:
foreground-launch-crash

* Tracking app launch event in background:
background-launch-event

* Tracking app launch crash in background:
background-launch-crash

* Tracking background event:
background-event

* Tracking background crash:
background-crash

### Review checklist

### DIFF
--- a/Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/Interception/AppStateListenerTests.swift
+++ b/Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/Interception/AppStateListenerTests.swift
@@ -181,7 +181,7 @@ class AppStateListenerTests: XCTestCase {
         )
 
         // When
-        (0..<1).forEach { _ in
+        (0..<10).forEach { _ in
             let randomNotification = supportedNotifications.randomElement()!
             notificationCenter.post(name: randomNotification.name, object: nil)
             expectedHistoryStates.append(randomNotification.expectedState)
--- a/Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/Interception/AppStateListenerTests.swift
+++ b/Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/Interception/AppStateListenerTests.swift
@@ -8,52 +8,48 @@ import XCTest
 @testable import Datadog
 
 class AppStateHistoryTests: XCTestCase {
-    func testForegroundDurationWithoutChanges() {
-        let startDate = Date.mockDecember15th2019At10AMUTC()
-        let history = AppStateHistory(
-            initialState: .init(isActive: true, date: startDate),
-            changes: [],
-            recentDate: startDate + 1.0
-        )
-
-        XCTAssertEqual(history.foregroundDuration, 1.0)
+    func testItBuildsAppStateFromUIApplicationState() {
+        XCTAssertEqual(AppState(uiApplicationState: .active), .active)
+        XCTAssertEqual(AppState(uiApplicationState: .inactive), .inactive)
+        XCTAssertEqual(AppState(uiApplicationState: .background), .background)
     }
 
-    func testForegroundDuration() {
-        let startDate = Date.mockDecember15th2019At10AMUTC()
+    func testWhenAppTransitionsBetweenForegroundAndBackground_itComputesTotalForegroundDuration() {
+        let numberOfForegroundSnapshots: Int = .mockRandom(min: 0, max: 20)
+        let numberOfBackgroundSnapshots: Int = .mockRandom(min: numberOfForegroundSnapshots == 0 ? 1 : 0, max: 20) // must include at least one snapshot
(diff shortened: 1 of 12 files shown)


Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/Interception/AppStateListenerTests.swift
```diff
@@ -116,87 +116,98 @@ class AppStateHistoryTests: XCTestCase {
...
+            dateProvider: RelativeDateProvider(startingFrom: .mockRandomInThePast(), advancingBySeconds: 1.0),
+            initialAppState: .mockRandom(),
             notificationCenter: notificationCenter
         )
-        let history1 = listener.history
-        let history2 = listener.history
 
-        XCTAssertEqual(history2.currentState.date.timeIntervalSince(history1.currentState.date), 1.0)
+        // When
+        (0..<1).forEach { _ in
```
maxep, line 184: This will do a single iteration :)
```suggestion
        (0..<10).forEach { _ in
```

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/Interception/AppStateListenerTests.swift
```diff
@@ -8,106 +8,106 @@ import XCTest
...
-                .init(isActive: false, date: startDate + 1.0),
-                .init(isActive: true, date: startDate + 2.0),
-                .init(isActive: false, date: startDate + 3.0),
-                .init(isActive: true, date: startDate + 4.0)
-            ],
-            recentDate: startDate + 5.0
-        )
+    func testWhenAppTransitionsBetweenForegroundAndBackground_itComputesTotalForegroundDuration() {
+        let numberOfForegroundSnapshots: Int = .mockRandom(min: 0, max: 20)
+        let numberOfBackgroundSnapshots: Int = .mockRandom(min: numberOfForegroundSnapshots == 0 ? 1 : 0, max: 20) // must include at least one snapshot
```
maxep, line 19: Why not just
```suggestion
        let numberOfBackgroundSnapshots: Int = .mockRandom(min: 1, max: 20) // must include at least one snapshot
```

maxep (approved):
Good catch 👌

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/Interception/AppStateListenerTests.swift
maxep: This one need to be addressed before merging. All good otherwise!

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/Interception/AppStateListenerTests.swift
ncreated: I wanted to also cover scenario where `numberOfBackgroundSnapshots` is `0`. And this couldn't be done if `numberOfForegroundSnapshots` is `0` too, because we need at least one snapshot for:
```swift
initialSnapshot: allSnapshots[0],
```
In general, in this fuzzy test I wanted to cover all possibilities:
* app only in foreground,
* app only in background,
* app in both foreground and background.

Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/Interception/AppStateListenerTests.swift
ncreated: Ops! Fixed, good catch 💪

PR #700, opened 2021-12-24
RUMM-1786 Mutable RUMDataModels

buranmert:
### What and why?

Web events coming from `browser-sdk` need to be mutated and decorated with the contextual info of native RUM session (e.g: application ID, active session ID, etc.)

### How?

Such properties are made to be `internally mutable` so that we can mutate them within SDK while outside of the SDK they are immutable as they were.

### TODO

- [ ] https://github.com/DataDog/dd-sdk-ios/pull/683 needs to be merged first

### Review checklist

### DIFF
--- a/Sources/Datadog/FeaturesIntegration/WebView/WebRUMEventConsumer.swift
+++ b/Sources/Datadog/FeaturesIntegration/WebView/WebRUMEventConsumer.swift
@@ -29,32 +29,92 @@ internal class WebRUMEventConsumer: WebEventConsumer {
         switch eventType {
         case "view":
             let viewEvent = try jsonDecoder.decode(RUMViewEvent.self, from: eventData)
-            let mappedViewEvent = mapIfNeeded(dataModel: viewEvent, context: rumContext, offset: getOffset(viewID: viewEvent.view.id))
+            let mappedViewEvent = mapIfNeeded_RUMViewEvent(dataModel: viewEvent, context: rumContext, offset: getOffset(viewID: viewEvent.view.id))
             write(mappedViewEvent)
         case "action":
             let actionEvent = try jsonDecoder.decode(RUMActionEvent.self, from: eventData)
-            let mappedActionEvent = mapIfNeeded(dataModel: actionEvent, context: rumContext, offset: getOffset(viewID: actionEvent.view.id))
+            let mappedActionEvent = mapIfNeeded_RUMActionEvent(dataModel: actionEvent, context: rumContext, offset: getOffset(viewID: actionEvent.view.id))
             write(mappedActionEvent)
         case "resource":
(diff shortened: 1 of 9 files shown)


Sources/Datadog/FeaturesIntegration/WebView/WebRUMEventConsumer.swift
```diff
@@ -29,32 +29,87 @@ internal class WebRUMEventConsumer: WebEventConsumer {
...
+    private func mapIfNeeded_RUMViewEvent(dataModel: RUMViewEvent, context: RUMContext?, offset: Offset) -> RUMViewEvent {
+        guard let context = context else {
+            return dataModel
+        }
+        var mappedDataModel = dataModel
+        mappedDataModel.application.id = context.rumApplicationID
+        mappedDataModel.session.id = context.sessionID.toRUMDataFormat
+        mappedDataModel.date = mappedDataModel.date + getOffset(viewID: mappedDataModel.view.id)
+        mappedDataModel.dd.session?.plan = .plan1
+        return mappedDataModel
```
buranmert, line 64: Android also mutates `service` property as can be seen [here](https://github.com/DataDog/dd-sdk-android/blob/feature/hybrid-application/dd-sdk-android/src/main/kotlin/com/datadog/android/webview/internal/rum/WebViewRumEventMapper.kt#L28)

in iOS we send `service: nil` in `RUMViewScope`, so i don't think we should mutate it. wdyt @ncreated ?

Sources/Datadog/FeaturesIntegration/WebView/WebRUMEventConsumer.swift
```diff
@@ -29,32 +29,92 @@ internal class WebRUMEventConsumer: WebEventConsumer {
...
             userLogger.error("🔥 Web RUM Event Error - Unknown event type: \(eventType)")
         }
     }
 
-    private func mapIfNeeded<T: RUMDataModel>(dataModel: T, context: RUMContext?, offset: Offset) -> T {
-        // TODO: RUMM-1786 implement mutating session_id & application_id
-        return dataModel
+    private func mapIfNeeded_RUMViewEvent(dataModel: RUMViewEvent, context: RUMContext?, offset: Offset) -> RUMViewEvent {
+        guard let context = context,
+              context.sessionID != .nullUUID else {
```
buranmert, line 57: when the session is sampled-out, its `sessionID` is set to `.nullUUID`.
with this condition, we send web events with their browser SDK context if the native session is sampled-out.

buranmert:
We simplify the way of mutating Web RUM Events in https://github.com/DataDog/dd-sdk-ios/pull/683 so that this PR is no longer needed

Issue #647, opened 2021-10-22
Kronos dependency triggers a Local Network Permission alert on iOS >= 14


thibauddavid:
Hi,

We are experiencing an issue with your SDK.
It relies on Kronos, which has an opened issue here https://github.com/MobileNativeFoundation/Kronos/issues/94 about a local network permission popup being asked.

Any plan on your side to mitigate this ?

Thanks

buranmert:
hi again @thibauddavid 👋 
it's been a while and i was wondering if you had a chance to reproduce the issue in a newly created project?

sergiocampama:
I'm seeing a similar thing after adding the DataDog SDK to my app. It's not reproducible though, we've just gotten some random reports from some users asking why we need to scan the local network when the app shouldn't.

Would it be possible to add a config to disable clock syncing? We don't really care that much about the logs being on an exact time. We segment them by user anyways, so they should be relative to each other.

buranmert:
@sergiocampama is it possible to ask those users for more info? 
for example: did that happen on WiFi or mobile data? do they use specific DNS servers? were they in a private network, like a workplace network, etc?

i still can't reproduce the issue although i tested scenarios such as:
• mobile data and WiFi
• using my router's local IP (192.168.X.X) as DNS server (in order to make a DNS request to a local IP)
• bad connection
• inaccessible NTP server (e.g: `foo.bar.ntp`)
i can only reproduce the issue if i use a local IP or an address with `.local` extension as NTP server.

please let me know if you suspect of anything else.
syncing time is essential for most of our users and we'd like to find the root cause and fix it if possible.

sergiocampama:
I'm asking a few users for this data, but they might not be savvy enough to know this. Still, an API to disable clock sync would be ideal for us to avoid this, since we don't care as our logs are user relative, so timing the logs to all users is not really worth it for us. Is this something that could be done? More than happy to send a patch if yes.

buranmert:
that would be great if we could have more info from those users @sergiocampama 🙏 

meanwhile, 
• we will be filing a radar to Apple, although there's quite a bit of unknowns in this case.
• we might try a different NTP implementation, yet again as we can't reproduce the issue we can't validate any new implementation solves the problem.

if that's okay for you too, you can fork our repo and patch it according to your needs.
it should be straightforward to use your fork with Swift Package Manager/Carthage.

please let me know if that helps

AJ9:
[...]

PR #701, opened 2021-12-29
RUMM-1744 Copy Kronos code to SDK

ncreated:
### What and why?

📦 This PR copies https://github.com/MobileNativeFoundation/Kronos/commit/52592508bc046a96364081db640e3ee467511e87 code directly to our SDK and removes the link to [Kronos](https://github.com/MobileNativeFoundation/Kronos) in all dependency managers.

There are few motivations to use our own version of Kronos:
* We want to collect diagnostic and telemetry data from Kronos code (in our dogfood projects) to analyse and act on https://github.com/DataDog/dd-sdk-ios/issues/647 issue.
* Own version gives us more control over this code - this especially important if we want to support platforms x dependency managers which are not supported in upstream [Kronos](https://github.com/MobileNativeFoundation/Kronos).
* Embedded code means less dependencies to our SDK users - faster installation (e.g. `carthage bootstrap`) and simpler linking for their app runtimes.

On the other hand, we shouldn't lose much by dropping external dependency on [Kronos](https://github.com/MobileNativeFoundation/Kronos). It is well sustained project, solving only one problem (NTP sync) and not requiring significant updates nor bugfixes (ref.: [release history](https://github.com/MobileNativeFoundation/Kronos/releases)). We will track all updates to the original repo with applying critical patches. Also, if we manage to find & fix issues in original implementation, we should contribute to the upstream project.

### How?

* Recent Kronos code and tests (https://github.com/MobileNativeFoundation/Kronos/commit/52592508bc046a96364081db640e3ee467511e87) were copied to our repository - everything was put into separate folder. 
* I removed Kronos dependency from `Datadog.podspec`, `Cartfile` and `Package.swift`.
* For clarity, I prefixed all imported types and files with `Kronos*`.
* Changed `public` access control to `internal` for respective types.
* Adjusted code style to pass our linter.
* Added licensing header in each file, referencing original authors.

Screenshot 2021-12-29 at 11 21 39

### Next PR(s):

I found two problems, that will be addressed in separate PR(s), to keep this one small and clear:
- [ ] Some of imported tests are very flaky as they do real UDP. I want to look closer into this and see what we can do.
- [ ] I will [update our distribution tool to assert Kronos.xcframework presence](https://github.com/DataDog/dd-sdk-ios/blob/e3468ed8dfb31a93c0f7b0831bfe9197ec15a53a/tools/distribution/src/assets/gh_asset.py#L19) only in SDK versions below `1.8.x`.

### Review checklist


maxep (approved):
All good 👍

dependency-manager-tests/carthage/CTProject.xcodeproj/project.pbxproj
```diff
@@ -14,8 +14,6 @@
 		61C36425243752A600C4D4E6 /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 61C36423243752A600C4D4E6 /* LaunchScreen.storyboard */; };
 		61C36430243752A600C4D4E6 /* CTProjectTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61C3642F243752A600C4D4E6 /* CTProjectTests.swift */; };
 		61C3643B243752A600C4D4E6 /* CTProjectUITests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61C3643A243752A600C4D4E6 /* CTProjectUITests.swift */; };
-		9E9D5E8825F90FC6002F12A0 /* Kronos.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = 9E9D5E8525F90FC6002F12A0 /* Kronos.xcframework */; };
```
buranmert, line 17: this is actually a breaking change for our carthage users.
ideally this would require a major version update.
we should at least communicate this in release notes.

buranmert (approved):

dependency-manager-tests/carthage/CTProject.xcodeproj/project.pbxproj
ncreated: Indeed 👍, Carthage users will have to remove `Kronos.xcframework` from their list of linked libraries, otherwise it will fail on first clean build. Although, I don't think it's a breaking change requiring major update (well, it's Carthage limitation), but it's definitely something we need to mention under "Migration Steps (Carthage)" in Release Notes for `1.9.0` 👌.

ncreated: Added `needs-migration-steps-explained` label.

PR #702, opened 2021-12-29
RUMM-1779 Keep view active until all resources are consumed

maxep:
### What and why?

Backend stop processing view updates when one view reports `is_active: false`. 

### How?

Only send the attribute `is_active: false` after ongoing resources are consumed.

### Review checklist

### DIFF
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
@@ -345,8 +345,10 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
         version += 1
         attributes.merge(rumCommandAttributes: command.attributes)
 
-        let timeSpent = command.time.timeIntervalSince(viewStartTime)
+        // RUMM-1779 Keep view active as long as we have ongoing resources
+        let isActive = isActiveView || !resourceScopes.isEmpty
 
+        let timeSpent = command.time.timeIntervalSince(viewStartTime)
         let cpuInfo = vitalInfoSampler.cpu
         let memoryInfo = vitalInfoSampler.memory
         let refreshRateInfo = vitalInfoSampler.refreshRate
@@ -384,7 +386,7 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
                 frozenFrame: .init(count: frozenFramesCount),
                 id: viewUUID.toRUMDataFormat,
                 inForegroundPeriods: nil,
-                isActive: isActiveView,
+                isActive: isActive,
                 isSlowRendered: isSlowRendered,
                 largestContentfulPaint: nil,
                 loadEvent: nil,
--- a/Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMViewScopeTests.swift
+++ b/Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMViewScopeTests.swift
@@ -416,14 +416,19 @@ class RUMViewScopeTests: XCTestCase {
             scope.process(command: RUMStopResourceCommand.mockWith(resourceKey: "/resource/1")),
             "The View should be kept alive as all its Resources haven't yet finished loading"
         )
+
+        var event = try XCTUnwrap(output.recordedEvents(ofType: RUMEvent<RUMViewEvent>.self).last)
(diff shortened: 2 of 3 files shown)


Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMViewScopeTests.swift
```diff
@@ -416,14 +416,19 @@ class RUMViewScopeTests: XCTestCase {
...
             scope.process(command: RUMStopResourceWithErrorCommand.mockWithErrorMessage(resourceKey: "/resource/2")),
             "The View should stop as all its Resources finished loading"
         )
 
-        let event = try XCTUnwrap(output.recordedEvents(ofType: RUMEvent<RUMViewEvent>.self).last)
+        event = try XCTUnwrap(output.recordedEvents(ofType: RUMEvent<RUMViewEvent>.self).last)
         XCTAssertEqual(event.model.view.resource.count, 1, "View should record 1 successful Resource")
         XCTAssertEqual(event.model.view.error.count, 1, "View should record 1 error due to second Resource failure")
+        XCTAssertFalse(event.model.view.isActive ?? true, "View should be inactive")
     }
```
ncreated, line 432: This test looks good 👍, although we could also test that _"SDK doesn't send view updates to completed views (to ones which received their terminal, `view.isActive == false` update)"_. How about enhancing session-level validators in `RUMSessionMatcher` to also integrate this assertion? It already has the notion of `ViewVisit` aggregating all view updates in their time order and [even applying some preliminary `isActive` checks](https://github.com/DataDog/dd-sdk-ios/blob/e3468ed8dfb31a93c0f7b0831bfe9197ec15a53a/Tests/DatadogTests/Matchers/RUMSessionMatcher.swift#L214-L234).

We use `RUMSessionMatcher` in all integration and many unit tests, so this validation would be heavily applied.

ncreated (commented):
Looks good 👍, but I left one suggestion for spreading validation of this logic in tests - LMTWDYT.

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMViewScopeTests.swift
maxep: Yes, I can have a look!
So we would basically stop sending view updates once the view is inactive? 
What is the actual definition of an [`active`](https://github.com/DataDog/rum-events-format/blob/master/schemas/view-schema.json#L125) view?
I worry if BE start processing inactive view events in the future, we will loose data.

ncreated: > So we would basically stop sending view updates once the view is inactive?

This is exactly how I understand it 👍, but maybe @xgouchet can confirm this.

> What is the actual definition of an active view?

I will share more context on Slack.

maxep: I need further modification to actually prevent from sending update events after a view is inactive, here it just sets `is_active: false`

maxep:
Looks like Bitrise pipeline is green, so don't know why it's reported as failing here 🤔

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -345,8 +345,10 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
         version += 1
         attributes.merge(rumCommandAttributes: command.attributes)
 
-        let timeSpent = command.time.timeIntervalSince(viewStartTime)
+        // RUMM-1779 Keep view active as long as we have ongoing resources
+        let isActive = isActiveView || !resourceScopes.isEmpty
```
buranmert, line 349: why do we wait for resources only? 
in theory, there can be ongoing actions as well.?

maxep: It was my question from our previous daily and only resources will keep the view active, it's the same logic to [complete a view scope](https://github.com/DataDog/dd-sdk-ios/blob/master/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift#L213L216)

buranmert (approved):

ncreated (approved):

PR #704, opened 2022-01-03
RUMM-1744 Update release tool to conditionally validate `Kronos.xcframework`

ncreated:
### What and why?

📦 This PR updates our release automation so it can conditionally include/exclude certain `XCFrameworks` from GitHub release asset validation (`.*zip`). This is required as starting from `1.9.0` SDK will no longer depend on `Kronos.xcframework` (#701), but when (re)building historical tags, we want it to be included.

### How?

I extracted the code for validating each `XCFramework` and used OCP to make it configurable for a certain range of `Version`. 

### Review checklist

### DIFF
--- a/tools/distribution/release.py
+++ b/tools/distribution/release.py
@@ -16,6 +16,7 @@ from tempfile import TemporaryDirectory
 from src.git import clone_repo
 from src.assets.gh_asset import GHAsset
 from src.assets.podspec import CPPodspec
+from src.semver import Version
 
 DD_SDK_IOS_REPO_SSH = 'git@github.com:DataDog/dd-sdk-ios.git'
 DD_SDK_IOS_REPO_NAME = 'dd-sdk-ios'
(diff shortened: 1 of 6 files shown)


buranmert (approved):
i'm not sure if this is the simplest solution but as long as it works it's fine for me

maxep (approved):
LGTM!

ncreated:
> i'm not sure if this is the simplest solution but as long as it works it's fine for me

@buranmert do you see anything in particular that we could simplify? My goal was to have separate validation for each `XCFramework` and this way we can assert more critical aspects: swift compatibility, architecture slices and dSYMs presence. Given that soon we will produce more slices (V2, next platforms), I think this approach should be reliable and easy to scale.

buranmert:
it's just seeing `+468 −37` made me wonder if that was the simplest solution.
i mean as far as i can see we do basically
```
if version < SOME_VERSION
  kronos_should_not_exist()
else
  kronos_should_exist_with_its_content()
```

ncreated:
> it's just seeing `+468 −37` made me wonder if that was the simplest solution. i mean as far as i can see we do basically
> 
> ```
> if version < SOME_VERSION
>   kronos_should_not_exist()
> else
>   kronos_should_exist_with_its_content()
> ```

It would need another `if/else` for crash reporting (`1.7.0+`) and several more for next platforms (tvOS / watchOS) + eventual modules in V2, ending up with `if/else` pyramid. Given no integration tests for this tool, introducing complex and not tested control flows sounds error prone which is where basic and declarative OCP should help.

PR #706, opened 2022-01-04
RUMM-1793 WebView events fire RUMWebViewCommand

buranmert:
### What and why?

With `browser-sdk` integration, iOS SDK can receive RUM events from the web views. Yet, these events don't pass through the same pipeline that regular RUM events pass. This may result in expired sessions while receiving events from the web views.

### How?

`RUMWebViewCommand` is an empty command that keeps the session alive and it is sent every time `WebRUMEventConsumer` receives an event.

### Review checklist

### DIFF
--- a/Sources/Datadog/RUM/RUMMonitor/RUMCommand.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/RUMCommand.swift
@@ -281,3 +281,10 @@ internal struct RUMAddLongTaskCommand: RUMCommand {
 
     let duration: TimeInterval
 }
+
+// MARK: - RUM Web Events related commands
+
+internal struct RUMWebViewCommand: RUMCommand {
+    var time: Date
+    var attributes: [AttributeKey: AttributeValue]
+}
--- a/Sources/Datadog/RUM/RUMMonitor/RUMCommand.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/RUMCommand.swift
@@ -284,7 +284,8 @@ internal struct RUMAddLongTaskCommand: RUMCommand {
 
 // MARK: - RUM Web Events related commands
 
-internal struct RUMWebViewCommand: RUMCommand {
+/// RUM Events received from WebView should keep the active session alive, therefore they fire this command to do so. (ref: RUMM-1793)
+internal struct RUMKeepSessionAliveCommand: RUMCommand {
     var time: Date
     var attributes: [AttributeKey: AttributeValue]
 }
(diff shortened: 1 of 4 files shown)


Sources/Datadog/RUM/RUMMonitor/RUMCommand.swift
```diff
@@ -281,3 +281,10 @@ internal struct RUMAddLongTaskCommand: RUMCommand {
 
     let duration: TimeInterval
 }
+
+// MARK: - RUM Web Events related commands
+
+internal struct RUMWebViewCommand: RUMCommand {
```
ncreated, line 287: We should call it with _verb + noun_, like other commands. Now `RUMWebViewCommand` doesn't tell what it does and it has no comment indicating it. As I understand, we may want to name it as `RUMKeepSessionAliveCommand` or similar. Additional comment on why this command was introduced can be handy for next readers, especially that we don't handle this command in any scope.

Tests/DatadogTests/Datadog/RUM/WebView/WebRUMEventConsumerTests.swift
```diff
@@ -122,5 +142,7 @@ class WebRUMEventConsumerTests: XCTestCase {
         let writtenJSON = try XCTUnwrap(try JSONSerialization.jsonObject(with: data, options: []) as? JSON)
 
         AssertDictionariesEqual(writtenJSON, unknownWebRUMEvent)
+        let webViewCommand = try XCTUnwrap(mockCommandSubscriber.lastReceivedCommand)
+        XCTAssertEqual(webViewCommand.time, .mockDecember15th2019At10AMUTC())
     }
```
ncreated, line 147: nit: instead of updating each test case with a new side effect we could simply add one new unit test with fuzzy input:
* Given: (random RUM context `x` random Web RUM event)
* When: `try eventConsumer.consume(event: webRUMEvent)`
* Then: `mockCommandSubscriber receives` "keep alive" command

This has classic benefit of fuzzy testing: covers all possible inputs for new requirement within a single test (vs now we cover new requirement in all other tests). Also, the test will be named in a way that describes new behaviour solely (vs now we don't call new behaviour from any of existing tests - the reader needs to guess it from assertions). Last, adding next behaviours to this suite will require adding only one test (vs now the convention is to test new behaviour in all existing tests).

ncreated (approved):
Looks good 👍. I left comments on conventions.

Tests/DatadogTests/Datadog/RUM/WebView/WebRUMEventConsumerTests.swift
buranmert: if that random input fails, it's very likely that we will re-run CI given that our tests are quite flaky.
i prefer keeping it deterministic in this case.

maxep (approved):
👍

PR #709, opened 2022-01-07
RUMM-1744 Collect Kronos telemetry in Internal Monitoring

ncreated:
### What and why?

📦🔬 This PR adds `KronosMonitor` for collecting internal telemetry on Kronos execution (NTP sync). This is to gather more data and have more observability on #647.

**Note:** this additional telemetry will only  be collected in apps that enable our Internal Monitoring feature, which means our own dogfood 🐶 projects. It won't be collected from customer apps.

### How?

`KronosMonitor` traces 3 phases of Kronos execution: 
* overall clock sync, 
* DNS resolution,
* UDP connection to each resolved IP.
```swift
internal protocol KronosMonitor {
    // MARK: - Clock sync
    func notifySyncStart(from pool: String)
    func notifySyncEnd(serverOffset: TimeInterval?)

    // MARK: - DNS resolution
    func notifyResolveDNS(to addresses: [KronosInternetAddress])

    // MARK: - IP querying
    func notifyStartQuerying(ip address: KronosInternetAddress, numberOfSamples: Int)
    func notifyReceivePacket(from address: KronosInternetAddress, isValidSample: Bool)
    func notifyEndQuerying(ip address: KronosInternetAddress)
}
```

Additionally, it uses `NWConnection` to check `UDP` connection reachability for each resolved IP on port `123`. It then aggregates all results (3 Kronos phases + connection checks to all IPs) and reports single telemetry event (`Log`), e.g.:

Screenshot 2022-01-07 at 16 19 49

### Review checklist

### DIFF
--- a/Datadog/E2ETests/NTP/KronosE2ETests.swift
+++ b/Datadog/E2ETests/NTP/KronosE2ETests.swift
@@ -7,8 +7,11 @@
 @testable import Datadog
 
 class KronosE2ETests: E2ETests {
-    /// The logger sending additional telemetry on Kronos execution. These logs are available in Mobile Integrations org.
+    /// The logger sending logs on Kronos execution. These logs are available in Mobile Integrations org.
     private var logger: Logger! // swiftlint:disable:this implicitly_unwrapped_optional
+    /// The logger sending telemetry on internal Kronos execution. These logs are available in Mobile Integrations org.
+    private var telemetryLogger: Logger! // swiftlint:disable:this implicitly_unwrapped_optional
+    private let queue = DispatchQueue(label: "kronos-monitor-queue")
 
     override func setUp() {
         super.setUp()
@@ -16,11 +19,31 @@ class KronosE2ETests: E2ETests {
             .builder
             .set(loggerName: "kronos-e2e")
             .build()
+        telemetryLogger = Logger.builder
+            .set(loggerName: "kronos-e2e-internal-telemetry")
+            .sendNetworkInfo(true)
+            .build()
     }
 
     override func tearDown() {
         logger = nil
+        telemetryLogger = nil
         super.tearDown()
+    }
+
+    /// Creates kronos monitor for checking connections to all IPs resolved from NTP pool and sending additional telemetry on their statuses.
+    private func createKronosMonitor() -> KronosMonitor? {
+        if #available(iOS 14.2, *) {
+            let monitor = KronosInternalMonitor(
+                queue: queue,
+                connectionMonitor: IPConnectionMonitor(queue: queue)
+            )
+            // Here we redirect IM's logger to E2E Kronos logger (`telemetryLogger`) to send data to Mobile Integrations org, not IM's org
+            monitor.export(to: InternalMonitor(sdkLogger: telemetryLogger))
+            return monitor
+        } else {
+            return nil
+        }
     }
 
     ///  TODO: RUMM-1859: Add E2E tests for monitoring Kronos in nightly tests
@@ -58,6 +81,7 @@ class KronosE2ETests: E2ETests {
             KronosClock.sync(
                 from: pool,
                 samples: numberOfSamplesForEachIP,
+                monitor: createKronosMonitor(),
                 first: { date, offset in // this closure could not be called if all samples to all servers resulted with failure
                     result.firstReceivedDate = date
                     result.firstReceivedOffset = offset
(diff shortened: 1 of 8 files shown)


Datadog/E2ETests/NTP/KronosE2ETests.swift
```diff
@@ -7,22 +7,45 @@
 @testable import Datadog
 
 class KronosE2ETests: E2ETests {
```
ncreated, line 9: It is to also collect this telemetry from our E2E test runs - more data, always good 💭.

Sources/Datadog/InternalMonitoring/Kronos/KronosMonitor.swift
```diff
@@ -0,0 +1,294 @@
...
+                self.statsByIP[ip] = IPStats()
+
+                if let connectionMonitor = self.connectionMonitor {
+                    self.dispatchGroup.enter() // await connection check result
+                    connectionMonitor.checkConnection(to: ip) { [weak self] result in // completion is called on `queue`
+                        self?.statsByIP[ip]?.checkResult = result
+                        self?.dispatchGroup.leave() // notify connection check result
+                    }
+                }
+            }
```
buranmert, line 146:
```swift
guard let connMonitor = self.connectionMonitor else { return }
addresses.forEach { _ in self.dispatchGroup.enter() }
addresses.forEach { ip in
  self.statsByIP[ip] = IPStats()
  connMonitor.checkConnection(to: ip) { [weak self] result in
    self?.statsByIP[ip]?.checkResult = result
    self?.dispatchGroup.leave()
  }
}
```
that way if `checkConnection` calls its `completion` too early somehow, we can avoid `dispatchGroup` being left too early as well

buranmert (commented):
lgtm but i left a comment regarding usage of `DispatchGroup`

Sources/Datadog/InternalMonitoring/Kronos/KronosMonitor.swift
ncreated: I see the point 👍 and this would make sense if completion is ran on an arbitrary thread:

```swift
queue.async {
    addresses.forEach { _ in
        if let _ = self.connectionMonitor { // ⚠️ nil on iOS < 14.2
            self.dispatchGroup.enter() // await connection check result
        }
    }

    addresses.forEach { ip in
        self.statsByIP[ip] = IPStats()

        if let connectionMonitor = self.connectionMonitor { // ⚠️ nil on iOS < 14.2
            connectionMonitor.checkConnection(to: ip) { [weak self] result in // completion is called on `queue`
                self?.statsByIP[ip]?.checkResult = result
                self?.dispatchGroup.leave() // notify connection check result
            }
        }
    }
}

```

However in current model it is impossible to receive `completion` simultaneously. Both `checkConnection(to:completion:)` and its `completion:` are called on the same queue, meaning that both "start all checks" and "receive single completion" are part of the same critical section and can't happen in parallel as stated:

```swift
connectionMonitor.checkConnection(to: ip) { [weak self] result in // completion is called on `queue`
```

In fact, if that wouldn't been assured, we'd also have problem with thread safe access to `self.statsByIP` from `completion`, whereas it works fine and is covered in tests.

sergiocampama:
I know this is not the scope of this PR, but could there also be a way to disable clock sync in the API/build settings? This might help mitigate this issue in the mean time for our users. The alternative for me would be to fork the repo and do this on my own, but I'd much prefer to stay on the official Cocoapod dependency.

buranmert (approved):

ncreated:
> I know this is not the scope of this PR, but could there also be a way to disable clock sync in the API/build settings? This might help mitigate this issue in the mean time for our users.

Hello @sergiocampama 🙂👋. Unfortunately, while such API / build setting might work for you it won't work for all our users. If an app uses Datadog Distributed Tracing or tracks 1st party RUM resources, disabling NTP will lead to serious and immediate misalignments in _client to server_ spans causing serious issues in data consistency. We totally understand that some apps might be not collecting distributed telemetry but the set of APIs and configuration options we provide must be equally functional for everyone. 

This is why we're not considering it as viable solution to mitigate the problem. Instead, the last sequence of actions we took in `dd-sdk-ios` should move us closed towards finding the root cause of #647 and fixing it. Meantime we stay open to any help from the OSS community on reproducing the issue.

> The alternative for me would be to fork the repo and do this on my own, but I'd much prefer to stay on the official Cocoapod dependency.

I wish I could have better answer for you and I clearly understand why this is important for your users. Unfortunately, a tailor-made fork is currently the only option to disable NTP in `dd-sdk-ios`. It can be easily achieved by [not invoking `KronosClock.sync {}` in our `ServerDateProvider`](https://github.com/DataDog/dd-sdk-ios/blob/0c96d0ac74259b97c007d3bc95e112e42623e811/Sources/Datadog/Core/System/Time/ServerDateProvider.swift#L30).

PR #711, opened 2022-01-10
RUMM-1890 Fix tests flakiness

ncreated:
### What and why?

🔬 This PR addresses tests flakiness from last month - collected from 189K test runs on the main branch:

Screenshot 2022-01-10 at 13 41 11

* green - reproduced and fixed;
* orange - cannot reproduce, added more verbosity;
* other - too little data;

### How?

Each case explained in PR comments. There's no general conclusion.

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/System/Time/DateCorrector.swift
+++ b/Sources/Datadog/Core/System/Time/DateCorrector.swift
@@ -38,7 +38,11 @@ internal class DateCorrector: DateCorrectorType {
         self.serverDateProvider = serverDateProvider
         serverDateProvider.synchronize(
             with: DateCorrector.datadogNTPServers.randomElement()!, // swiftlint:disable:this force_unwrapping
-            completion: { offset in
+            completion: { [weak self] offset in
+                guard let _ = self else {
+                    return
+                }
+
                 if let offset = offset {
                     let difference = (offset * 1_000).rounded() / 1_000
                     userLogger.info(
--- a/Sources/Datadog/Core/System/Time/DateCorrector.swift
+++ b/Sources/Datadog/Core/System/Time/DateCorrector.swift
@@ -38,11 +38,7 @@ internal class DateCorrector: DateCorrectorType {
         self.serverDateProvider = serverDateProvider
         serverDateProvider.synchronize(
             with: DateCorrector.datadogNTPServers.randomElement()!, // swiftlint:disable:this force_unwrapping
-            completion: { [weak self] offset in
-                guard let _ = self else {
-                    return
-                }
-
+            completion: { offset in
                 if let offset = offset {
                     let difference = (offset * 1_000).rounded() / 1_000
                     userLogger.info(
--- a/Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScopeTests.swift
+++ b/Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScopeTests.swift
@@ -151,7 +151,7 @@ class RUMApplicationScopeTests: XCTestCase {
             backgroundEventTrackingEnabled: .mockAny()
         )
 
-        let simulatedSessionsCount = 200
+        let simulatedSessionsCount = 400
         (0..<simulatedSessionsCount).forEach { _ in
             _ = scope.process(command: RUMStartViewCommand.mockWith(time: currentTime, identity: mockView))
             _ = scope.process(command: RUMStopViewCommand.mockWith(time: currentTime, identity: mockView))
@@ -161,7 +161,8 @@ class RUMApplicationScopeTests: XCTestCase {
         let viewEventsCount = try output.recordedEvents(ofType: RUMEvent<RUMViewEvent>.self).count
         let trackedSessionsCount = Double(viewEventsCount) / 2 // each Session should send 2 View updates
 
-        XCTAssertGreaterThan(trackedSessionsCount, 100 * 0.8) // -20%
-        XCTAssertLessThan(trackedSessionsCount, 100 * 1.2) // +20%
+        let halfSessionsCount = 0.5 * Double(simulatedSessionsCount)
+        XCTAssertGreaterThan(trackedSessionsCount, halfSessionsCount * 0.8) // -20%
+        XCTAssertLessThan(trackedSessionsCount, halfSessionsCount * 1.2) // +20%
     }
 }
--- a/Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMViewIdentityTests.swift
+++ b/Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMViewIdentityTests.swift
@@ -89,12 +89,15 @@ class RUMViewIdentityTests: XCTestCase {
     // MARK: - Memory management
 
     func testItStoresWeakReferenceToUIViewController() throws {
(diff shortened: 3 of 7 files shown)


Sources/Datadog/Core/System/Time/DateCorrector.swift
```diff
@@ -38,7 +38,11 @@ internal class DateCorrector: DateCorrectorType {
         self.serverDateProvider = serverDateProvider
         serverDateProvider.synchronize(
             with: DateCorrector.datadogNTPServers.randomElement()!, // swiftlint:disable:this force_unwrapping
-            completion: { offset in
+            completion: { [weak self] offset in
+                guard let _ = self else {
+                    return
+                }
+
```
ncreated, line 45: This change is to address flaky execution of

https://github.com/DataDog/dd-sdk-ios/blob/181359bd655b894041702332c26fa05b2d451990/Tests/DatadogTests/Datadog/Core/Upload/DataUploadWorkerTests.swift#L230-L235

Problem was that `DateCorrector` is expected to perform real NTP sync in many high-level tests (e.g. most tests in [`DatadogTests`](https://github.com/DataDog/dd-sdk-ios/blob/181359bd655b894041702332c26fa05b2d451990/Tests/DatadogTests/Datadog/DatadogTests.swift#L10)). This closure might leak from test execution and send NTP sync log to global `userLogger`, causing bad value being recorded in other tests that depend on global `userLogger` mock, e.g.:

Screenshot 2022-01-10 at 10 19 07

The fix is to not leak this closure - and return it early when SDK was deallocated.

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScopeTests.swift
```diff
@@ -151,7 +151,7 @@ class RUMApplicationScopeTests: XCTestCase {
             backgroundEventTrackingEnabled: .mockAny()
         )
 
-        let simulatedSessionsCount = 200
+        let simulatedSessionsCount = 400
```
ncreated, line 154: Using tests repetition in local:
* with 200 samples - 4 failures in 500 runs
* with 300 samples - 1 failures in 500 runs
* with 400 samples - 0 failures in 500 runs

With 400 samples it runs in `~180ms`.

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMViewIdentityTests.swift
```diff
@@ -89,12 +89,15 @@ class RUMViewIdentityTests: XCTestCase {
...
+        try autoreleasepool {
+            var vc: UIViewController? = UIViewController()
+            identity = try XCTUnwrap(vc?.asRUMViewIdentity())
+            XCTAssertNotNil(identity.identifiable, "Reference should be available while `vc` is alive.")
+            vc = nil
+        }
 
-        XCTAssertNotNil(identity.identifiable, "Reference should be available while `vc` is alive.")
-        vc = nil
         XCTAssertNil(identity.identifiable, "Reference should not be available after `vc` was deallocated.")
```
ncreated, line 101: 2 failures in last 1 month. It was failing on:
```swift
XCTAssertNotNil(identity.identifiable, "Reference should be available while `vc` is alive.")
```
I assume that `autoreleasepool {}` should help as VC is autorelease Objective-C object and now we clean it up in different scope than the assertion.

Tests/DatadogTests/Datadog/RUM/RUMVitals/VitalRefreshRateReaderTests.swift
```diff
@@ -11,43 +11,54 @@ import UIKit
 class VitalRefreshRateReaderTests: XCTestCase {
     private let mockNotificationCenter = NotificationCenter()
 
-    func testHighAndLowRefreshRates() {
+    func testWhenMainThreadOverheadGoesUp_itMeasuresLowerRefreshRate() throws {
```
ncreated, line 14: 3 failures in last 1 month. This is quite old issue, I changed the approach for this test:
* Instead of observing main thread for certain amount of time...
* ... now it observes it until `VitalRefreshRateReader` records certain number of samples.

It passes 500x repetition in local. From my observation, time-based constraint was very flaky (sometimes recording just 1 sample, sometimes many). Current approach of recording 30 samples seems stable and still executes under `2s` in local (like before).

Sources/Datadog/Core/System/Time/DateCorrector.swift
maxep: Would this work? 
```swift
completion: { [weak userLogger] offset in
```
Not a change request, I'm not sure about lifecycle of the logger vs. DateCorrector 🤔

Tests/DatadogTests/Helpers/XCTestCase.swift
```diff
@@ -146,12 +146,23 @@ extension XCTestCase {
...
+        do {
+            let encodedValue1: Data
+            let encodedValue2: Data
+            if #available(iOS 13.0, *) {
+                encodedValue1 = try prettyEncoder.encode(value1)
+                encodedValue2 = try prettyEncoder.encode(value2)
+            } else {
+                encodedValue1 = try prettyEncoder.encode(EncodingContainer(value1))
+                encodedValue2 = try prettyEncoder.encode(EncodingContainer(value2))
+            }
```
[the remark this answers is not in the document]
maxep: 👌

maxep (approved):

Sources/Datadog/Core/System/Time/DateCorrector.swift
ncreated: `DateCorrector` is instantiated [only once for a given SDK instance](https://github.com/DataDog/dd-sdk-ios/blob/b4d285b0820fb83d561c492a8c740d6ba87307cf/Sources/Datadog/Datadog.swift#L177-L180) (from `Datadog.initialize()`).

I think more correct solution is to avoid doing **any work** in this `DateCorrector` closure if `self == nil`. From `self == nil` we know that `DateCorrector` was deallocated → so the parent `Datadog` object was deallocated → so no-one is interested in result of this work.

Capturing weak reference to `userLogger` *might* work too but it will be ignoring only a portion of the completion work related to `userLogger`. As no-one will use the result of `completion` work, I find this odd.

How does it sound?

maxep: I guess the root cause is that `Kronos.Clock` is static as well as `userLogger`. I think it's best to fix it in [`ServerDateProvider`](https://github.com/DataDog/dd-sdk-ios/blob/master/Sources/Datadog/Core/System/Time/ServerDateProvider.swift#L35L46) and call the completion only if `self` exist, WDYT?

ncreated: > I think it's best to fix it in [`ServerDateProvider`](https://github.com/DataDog/dd-sdk-ios/blob/master/Sources/Datadog/Core/System/Time/ServerDateProvider.swift#L35L46) and call the completion only if `self` exist, WDYT?

Indeed! It depends on `self` already and clearly it runs its completion inconsitently when `self` is `nil`. Will do 👍

ncreated: Done ✅

buranmert (approved):

PR #712, opened 2022-01-10
RUMM-1796 Crash Reporting Logging

maxep:
### What and why?

Add logs to crash reporting.

### How?

Inject user and internal logger to crash reporting plugin.

### Review checklist

### DIFF
--- a/Sources/Datadog/CrashReporting/CrashReporter.swift
+++ b/Sources/Datadog/CrashReporting/CrashReporter.swift
@@ -89,9 +89,11 @@ internal class CrashReporter {
         queue.async {
             self.plugin.readPendingCrashReport { [weak self] crashReport in
                 guard let self = self, let availableCrashReport = crashReport else {
+                    userLogger.debug("No pending crash available")
                     return false
                 }
 
+                userLogger.debug("Loaded pending crash report")
 #if DD_SDK_ENABLE_INTERNAL_MONITORING
                 InternalMonitoringFeature.instance?.monitor.sdkLogger
                     .debug("Loaded pending crash report", attributes: availableCrashReport.diagnosticInfo)
@@ -166,4 +168,11 @@ internal class CrashReporter {
             return nil
         }
     }
+
+#if DD_SDK_COMPILED_FOR_TESTING
+    func deinitialize() {
+        // Await asynchronous operations completion to safely sink all pending tasks.
+        queue.sync {}
+    }
+#endif
 }
--- a/Sources/Datadog/CrashReporting/CrashReporter.swift
+++ b/Sources/Datadog/CrashReporting/CrashReporter.swift
@@ -89,11 +89,9 @@ internal class CrashReporter {
         queue.async {
             self.plugin.readPendingCrashReport { [weak self] crashReport in
                 guard let self = self, let availableCrashReport = crashReport else {
-                    userLogger.debug("No pending crash available")
                     return false
                 }
 
-                userLogger.debug("Loaded pending crash report")
 #if DD_SDK_ENABLE_INTERNAL_MONITORING
                 InternalMonitoringFeature.instance?.monitor.sdkLogger
                     .debug("Loaded pending crash report", attributes: availableCrashReport.diagnosticInfo)
--- a/Sources/Datadog/CrashReporting/CrashReporter.swift
+++ b/Sources/Datadog/CrashReporting/CrashReporter.swift
@@ -89,9 +89,11 @@ internal class CrashReporter {
         queue.async {
             self.plugin.readPendingCrashReport { [weak self] crashReport in
                 guard let self = self, let availableCrashReport = crashReport else {
+                    userLogger.debug("No pending crash available")
                     return false
                 }
 
+                userLogger.debug("Loaded pending crash report")
 #if DD_SDK_ENABLE_INTERNAL_MONITORING
                 InternalMonitoringFeature.instance?.monitor.sdkLogger
                     .debug("Loaded pending crash report", attributes: availableCrashReport.diagnosticInfo)
--- a/Sources/Datadog/FeaturesIntegration/CrashReporting/CrashReportingWithRUMIntegration.swift
+++ b/Sources/Datadog/FeaturesIntegration/CrashReporting/CrashReportingWithRUMIntegration.swift
@@ -101,6 +101,7 @@ internal struct CrashReportingWithRUMIntegration: CrashReportingIntegration {
         } else {
             // We know it is too late for sending RUM view to previous RUM session as it is now stale on backend.
             // To avoid inconsistency, we only send the RUM error.
+            userLogger.debug("Sending crash as RUM error.")
             let rumError = createRUMError(from: crashReport, and: lastRUMViewEvent, crashDate: crashTimings.realCrashDate)
             rumEventOutput.write(rumEvent: rumError)
         }
@@ -146,7 +147,7 @@ internal struct CrashReportingWithRUMIntegration: CrashReportingIntegration {
                 crashContext: crashContext
             )
         case .doNotHandle:
-            // This can mean that the crash happened in background and BET is disabled OR that the previous session was sampled.
+            userLogger.debug("There was a crash in background, but it is ignored due to Background Event Tracking disabled or sampling.")
             newRUMView = nil
         }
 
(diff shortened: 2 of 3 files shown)


Sources/DatadogCrashReporting/DDCrashReportingPlugin.swift
```diff
@@ -51,18 +54,27 @@ public class DDCrashReportingPlugin: NSObject, DDCrashReportingPluginType {
...
+        internalLogger = logger
+    }
+#endif
 }
+
+// MARK: - Plugin Logs
+
+internal var consoleLogger: Logger? = nil
+
+internal var internalLogger: Logger? = nil
```
maxep, line 80: Can't find a better way without refactoring the plugin. This pattern is already used in the SDK.

Sources/Datadog/Logger.swift
```diff
@@ -57,7 +57,7 @@ public enum LogLevel: Int, Codable {
 ///
 public typealias DDLogger = Logger
 
-public class Logger {
+public class Logger: NSObject {
```
maxep, line 60: This is needed to inject the logger in the crash reporting objc plugin protocol..

ncreated: IMO it's definitely not a way to go - this will add the entire [`NSObject` boilerplate](https://developer.apple.com/documentation/objectivec/nsobject) to `Logger` object in user's code. I believe it will not only sink its APIs discoverability in auto-completion but also change the performance of `Logger`.

I'd consider different approaches, e.g.:
* having public `DDCrashReportingLogger` interface that `Logger` will conform to (internally);
* having public `DDCrashReportingLogger` type that will wrap the `Logger`.

Also, mind that this:
```swift
@objc
public protocol DDCrashReportingPluginType: AnyObject
```
is just a convenience in `Datadog` (swift) so we don't have to provide separate interface from `DatadogObjc`. What we struggle with here is just a result of this convenience shortcut but issues like this are commonly solved in `DatadogObjc` by providing `@objc` wrappers for all public Swift APIs. I think it's not worth looking for a solution in `DatadogObjc` as the problem is relatively small 😉. But such small problem should not lead to bypassing the `DatadogObjc` → `Datadog` separation and introducing `NSObject` cost to `Logger`.

maxep: Indeed, making `Logger` a `NSObject` subclass is not the right way. But The first proposed approach won't work, the `DDCrashReportingLogger` will need to be objc and we will end up with the same issue with `Logger` being pure Swift. We can create a objc `DDCrashReportingLogger` wrapper for `Logger`, but this seems to be quite excessive for just adding some logs.
I'm also wondering if we could just make the `DDCrashReportingPluginType` a pure Swift protocol, I don't see this abstraction being used by an objc integration, it is mostly for our own use. WDYT?

ncreated: > I'm also wondering if we could just make the `DDCrashReportingPluginType` a pure Swift protocol, I don't see this abstraction being used by an objc integration, it is mostly for our own use. WDYT?

Yes, this could work and it sounds good 👍. There is no use case for providing implementation to this protocol in Objc and seems that objc constraint was added to it for no reason.

maxep: Damn.. it's objc because of [this](https://github.com/DataDog/dd-sdk-ios/blob/master/Sources/DatadogObjc/DatadogConfiguration%2Bobjc.swift#L197).
To follow the current pattern, we will need to create a bridge to `DDCrashReportingPluginType` in `DatadogObjc`. This is a fairly large refactoring with impact on our public interface.. I will rather add logs to the integration layer instead of within the plugin.

ncreated: Indeed 😞. And definitely agree - tracing the plugin is separate challenge and the goal of `RUMM-1796` was to trace the integration logic itself. Maybe we can create separate task for exporting logs from the plugin later?

Sources/Datadog/CrashReporting/CrashReporter.swift
```diff
@@ -135,9 +137,8 @@ internal class CrashReporter {
                 """
                 Failed to encode crash report context. The app state information associated with eventual crash
                 report may be not in sync with the current state of the application.
-
-                Error details: \(error)
-                """
+                """,
+                error: error
```
ncreated, line 141: `userLogger` [is configured with `.short` format](https://github.com/DataDog/dd-sdk-ios/blob/79c63be50b4e5f1ec3e6432e84da34cac30efd07/Sources/Datadog/Utils/InternalLoggers.swift#L58) meaning that it [only prints the `message` to the console](https://github.com/DataDog/dd-sdk-ios/blob/79c63be50b4e5f1ec3e6432e84da34cac30efd07/Sources/Datadog/Logging/LogOutputs/LogConsoleOutput.swift#L79), hence emitting `error` to the `message` has more sense than setting `error: error` which won't be displayed to the user.

Sources/Datadog/FeaturesIntegration/CrashReporting/CrashReportingWithRUMIntegration.swift
```diff
@@ -101,6 +101,7 @@ internal struct CrashReportingWithRUMIntegration: CrashReportingIntegration {
         } else {
             // We know it is too late for sending RUM view to previous RUM session as it is now stale on backend.
             // To avoid inconsistency, we only send the RUM error.
+            userLogger.debug("Report the crash as error.")
```
ncreated, line 104: How about using present continuous tense? I think this will be more unified with other messages we add here. 
```suggestion
            userLogger.debug("Sending crash as RUM error.")
```

Sources/Datadog/FeaturesIntegration/CrashReporting/CrashReportingWithRUMIntegration.swift
```diff
@@ -201,6 +203,7 @@ internal struct CrashReportingWithRUMIntegration: CrashReportingIntegration {
 
     /// Sends given `CrashReport` by linking it to given `rumView` and updating view counts accordingly.
     private func send(crashReport: DDCrashReport, to rumView: RUMEvent<RUMViewEvent>, using realCrashDate: Date) {
+        userLogger.debug("Update view with crash report.")
```
ncreated, line 206:
```suggestion
        userLogger.debug("Updating RUM view with crash report.")
```

ncreated (changes requested):
Looks good 👍. I left wording suggestions and change request on `error: error` (see the comment).

Sources/Datadog/CrashReporting/CrashReporter.swift
ncreated: BTW, we could add an improvement (to the backlog) to make `.short` format also include `error.message` (if available). That would make more sense IMO 👌.

maxep: Ah that's why, I thought this was a mistake 😆 I will revert then.
> Makes sense to include the error in short format, I will log something!

Sources/DatadogCrashReporting/PLCrashReporterIntegration/CrashReport.swift
```diff
@@ -283,7 +283,7 @@ extension BinaryImageInfo.CodeType {
         let subtypeMask = UInt64(CPU_SUBTYPE_MASK)
 
         // Ref. for this check:
-        // https://github.com/microsoft/plcrashreporter/blob/dbb05c0bc883bde1cfcad83e7add25862c95d11f/Source/PLCrashReportTextFormatter.m#L371
+        // https://github.com/microsoft/plcrashreporter/blob/master/Source/PLCrashReportTextFormatter.m#L371
```
maxep, line 286: dd git commit hook is alerting about this hash

ncreated (commented):
LG 👍 , though what about this from `RUMM-1796`:

> * print `INFO` log determining if the crash report is found after app launch, or if it's missing;

[`sendCrashReportIfFound()`](https://github.com/DataDog/dd-sdk-ios/blob/79c63be50b4e5f1ec3e6432e84da34cac30efd07/Sources/Datadog/CrashReporting/CrashReporter.swift#L91-L94) seems a good place for it, no?

ncreated (approved):

PR #714, opened 2022-01-11
RUMM-1649 WebViewTracking integration test added

buranmert:
### What and why?

Adding integration test to WebView tracking feature.

### How?

In this new test, the app opens a WebView which loads Browser SDK instrumented content. The iOS SDK should transport all RUM events and Logs received from Browser SDK.

### Review checklist

### DIFF
--- a/Datadog/TargetSupport/DatadogIntegrationTests/DatadogCrashReportingIntegrationTests.xctestplan
+++ b/Datadog/TargetSupport/DatadogIntegrationTests/DatadogCrashReportingIntegrationTests.xctestplan
@@ -33,21 +33,9 @@
   },
   "testTargets" : [
     {
-      "skippedTests" : [
-        "IntegrationTests",
-        "LoggingScenarioTests",
-        "RUMManualInstrumentationScenarioTests",
-        "RUMMobileVitalsScenarioTests",
-        "RUMModalViewsScenarioTests",
-        "RUMNavigationControllerScenarioTests",
-        "RUMResourcesScenarioTests",
-        "RUMScrubbingScenarioTests",
-        "RUMSwiftUIScenarioTests",
-        "RUMTabBarControllerScenarioTests",
-        "RUMTapActionScenarioTests",
-        "TracingManualInstrumentationScenarioTests",
-        "TracingURLSessionScenarioTests",
-        "TrackingConsentScenarioTests"
+      "selectedTests" : [
+        "CrashReportingWithLoggingScenarioTests\/testCrashReportingCollectOrSendWithLoggingScenario()",
+        "CrashReportingWithRUMScenarioTests\/testCrashReportingCollectOrSendWithRUMScenario()"
       ],
       "target" : {
         "containerPath" : "container:Datadog.xcodeproj",
--- a/Tests/DatadogTests/Matchers/RUMEventMatcher.swift
+++ b/Tests/DatadogTests/Matchers/RUMEventMatcher.swift
@@ -40,8 +40,19 @@ internal class RUMEventMatcher {
     private let jsonDataDecoder = JSONDecoder()
 
     private init(with jsonData: Data) throws {
-        self.jsonMatcher = JSONDataMatcher(from: try jsonData.toJSONObject())
-        self.jsonData = jsonData
+        var json = try jsonData.toJSONObject()
+
+        // NOTE: RUMM-1649 WebViewScenarioTest receives lowercase `method` values
+        if let eventType = json["type"] as? String,
+           eventType == "resource",
+           var resource = json["resource"] as? [String: Any],
+           let method = resource["method"] as? String {
+            resource["method"] = method.uppercased()
+            json["resource"] = resource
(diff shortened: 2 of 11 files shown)


buranmert:
Most of the changes coming from the merge from `master` in order to get the new `source` property

ncreated:
> Most of the changes coming from the merge from `master` in order to get the new `source` property

I isolated updating `hybrid-application` with `master` to #722 to keep this PR minimal.

Datadog/TargetSupport/DatadogIntegrationTests/DatadogCrashReportingIntegrationTests.xctestplan
```diff
@@ -33,21 +33,9 @@
...
-        "RUMScrubbingScenarioTests",
-        "RUMSwiftUIScenarioTests",
-        "RUMTabBarControllerScenarioTests",
-        "RUMTapActionScenarioTests",
-        "TracingManualInstrumentationScenarioTests",
-        "TracingURLSessionScenarioTests",
-        "TrackingConsentScenarioTests"
+      "selectedTests" : [
+        "CrashReportingWithLoggingScenarioTests\/testCrashReportingCollectOrSendWithLoggingScenario()",
+        "CrashReportingWithRUMScenarioTests\/testCrashReportingCollectOrSendWithRUMScenario()"
```
ncreated, line 38: This change is on purpose. By default, all new integration tests are added to both XCTest plans (_integration_ and _crash reporting_), meaning we will run it twice on CI. To avoid this, I disabled "add new tests automatically" in _crash reporting_ XCTest plan. Apparently, this means changing from opt-out to opt-in in `.xctestplan` 👍.

ncreated:
I added more assertions on data received from Browser SDK, this includes checks like:
* if Browser events are sent to native iOS RUM session (using native `application_id`  and `session_id`);
* if Browser logs are transported through iOS SDK;
* if Browser logs are correlated with native iOS RUM session.

Tests/DatadogTests/Matchers/RUMEventMatcher.swift
```diff
@@ -40,8 +40,19 @@ internal class RUMEventMatcher {
...
+        var json = try jsonData.toJSONObject()
+
+        // NOTE: RUMM-1649 WebViewScenarioTest receives lowercase `method` values
+        if let eventType = json["type"] as? String,
+           eventType == "resource",
+           var resource = json["resource"] as? [String: Any],
+           let method = resource["method"] as? String {
+            resource["method"] = method.uppercased()
+            json["resource"] = resource
+        }
```
maxep, line 52: Is it possible to do case-insensitive comparison instead of transforming the payload?

ncreated: Not if we want to keep using `RUMEventMatcher`. This matcher decodes certain RUM event model from `Data`:
```swift
func model<DM: Decodable>(file: StaticString = #file, line: UInt = #line) throws -> DM
```
RUM events schema [defines](https://github.com/DataDog/rum-events-format/blob/114c173caac5ea15446a157b666acbab05431361/schemas/resource-schema.json#L44-L49) uppercased values for `resource.method`:
```js
"method": {
   "type": "string",
   "description": "HTTP method of the resource",
   "enum": ["POST", "GET", "HEAD", "PUT", "DELETE", "PATCH"],
   "readOnly": true
},
```
Because our `RUMDataModels` definition is strict about this and Browser SDK seems not, receiving lowercased method name from WebView (e.g. `"get"` instead of `"GET"`) leads to decoding error. IMO this is an inconsistency in Browser SDK and we need to patch it in order to leverage strict validation with `RUMEventMatcher` and `RUMSessionMatcher` in our tests.

However, I think we can move this patch to a better place. Residing here in `RUMEventMatcher`, it is applied globally to all our test data, but the only problematic test is `WebViewScenarioTest.swift` where we consume real Browser SDK events. I'd vote for transforming data directly in `WebViewScenarioTest.swift` before passing it to the matcher, as this will minimalize the impact of this patch and be more explanatory. WDYT on this @maxep ?

Alternatively, we could resign from using strict matchers and assert all values as `[String: Any]` on events deserialized with `JSONSerialization`. This way we can perform case-insensitive comparison, but we lose all other strict validation from `RUMSessionMatcher`. Adding this validation back would require bloating `WebViewScenarioTest.swift` with manual deserialization code and duplication of all validation concepts.

maxep (approved):
👌

Tests/DatadogTests/Matchers/RUMEventMatcher.swift
ncreated: Moved to `WebViewScenarioTest.swift` ✅. I also talked with `browser-sdk` team - they will fix it in future release. 

Such fix is required, as some queries in hybrid apps (e.g. _count all resources by `resource.method`_) might be giving odd results (i.e. considering "post" and "POST" as separate methods).

PR #720, opened 2022-01-14
Update to version 1.1.0 of testing framework

nachoBonafonte:
### What and why?

Update to version 1.1.0 of testing framework 

### How?

Update the version in Makefile
Update test targets so they keep the previous behaviour (Stdout and Stderr capturing must be explicitly activated now)

### Review checklist


ncreated (approved):
LG 👍👌, I gave it `env:local` run and all seems fine.

PR #723, opened 2022-01-17
RUMM-1924 Fix bug with WebView logs not appearing in Logs Explorer

ncreated:
### What and why?

🐞 This PR fixes bug with WebView logs processing. Due to wrong date correction, WebView logs were rejected by BE.

### How?

Correction logic was using `ns`, instead of `ms`.

### Review checklist


maxep (approved):
👍

PR #725, opened 2022-01-18
RUMM-1925 Add `@service` attribute to all RUM events

ncreated:
### What and why?

📦 This PR adds `@service` attribute to all RUM events. This is to unify iOS events with Browser SDK and Android SDK.

### How?

`service` is already resolved by `FeaturesConfiguration.Common`. This PR only passes it all the way down to RUM scopes thru existing DI container.

### Review checklist


maxep (approved):
LGTM 👍

PR #728, opened 2022-01-19
RUMM-1910 Strip query parameters from span resource

xgouchet:
### What and why?

Spans have a limited number of unique resource values. Removing query parameters ensure we don't use too many unique slots, and aligns the iOS SDK with android and other APM implementations.

### DIFF
--- a/Tests/DatadogTests/Datadog/Mocks/SystemFrameworks/FoundationMocks.swift
+++ b/Tests/DatadogTests/Datadog/Mocks/SystemFrameworks/FoundationMocks.swift
@@ -175,6 +175,12 @@ extension URL: AnyMockable, RandomMockable {
         return URL(string: "https://www.foo.com/")!.appendingPathComponent(pathComponent)
     }
 
+    static func mockWith(url: String, queryParams: [URLQueryItem]) -> URL {
+        var urlComponents = URLComponents(string: url)
+        urlComponents!.queryItems = queryParams
+        return urlComponents!.url!
+    }
+
     static func mockRandom() -> URL {
         return URL(string: "https://www.foo.com/")!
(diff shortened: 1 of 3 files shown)


Tests/DatadogTests/Datadog/Mocks/SystemFrameworks/FoundationMocks.swift
```diff
@@ -412,6 +418,13 @@ extension URLRequest: AnyMockable {
         request.httpMethod = httpMethod
         return request
     }
+
+    static func mockWith(url: String, queryParams: [URLQueryItem], httpMethod: String) -> URLRequest {
+        let url: URL = .mockWith(url: url, queryParams: queryParams)
+        var request = URLRequest(url: url)
+        request.httpMethod = httpMethod
+        return request
+    }
```
maxep, line 427: We could simplify mocking by getting rid of `func mockWith(url: String, queryParams: [URLQueryItem]) -> URL`

and do
```swift
static func mockWith(url string: String, httpMethod: String = "GET") -> URLRequest {
    let url: URL = URL(string: string)!
    var request = URLRequest(url: url)
    request.httpMethod = httpMethod
    return request
}

let request: URLRequest = .mockWith(url: "https://www.example.com?foo=42&lang=en")
```
WDYT?

xgouchet: I'm always more inclined to have more control. Also, assuming we could use some fuzzy testing in iOS, it'd be easier to generate a random base url and a random list of query param to build the full url than writing all like you propose

maxep (approved):

ncreated (approved):

PR #730, opened 2022-01-19
RUMM-1803 Consider UIApplication init as warm start measurement

maxep:
🚧  do not merge

### What and why?

Starting in iOS 15, the [pre-warm optimisation](https://developer.apple.com/documentation/uikit/app_and_environment/responding_to_the_launch_of_your_app/about_the_app_launch_sequence) can start the application process hours before an actual launch. This new feature results in very long launch time measurement as we consider Application Launch Time as the interval between Process Start Time and `UIApplicationDidBecomeActiveNotification`.

Prewarming executes an app’s launch sequence up until, but not including, when `main()` calls `UIApplicationMain(_:_:_:_:)`. We need to define a new application starting point in time. In this proposal we measure the warm start time as when the `UIApplication` is initialised.

### How?

To measure the point time when the `UIApplication` is initialised by `UIApplicationMain`, this proposal implement swizzling on `UIApplication.init` to record the time it get invoked.

To differentiate between a cold and a warm start we can either use a time interval threshold, or use this - not documented - yet to be validated - process variable:
```objective-c
#define IS_ACTIVE_PREWARM [NSProcessInfo.processInfo.environment[@"ActivePrewarm"] isEqual:@"1"]
```
This variable is currently being evaluated in dogfooding.

### Review checklist

### DIFF
--- a/Sources/_Datadog_Private/ObjcAppLaunchHandler.m
+++ b/Sources/_Datadog_Private/ObjcAppLaunchHandler.m
@@ -6,21 +6,28 @@
 
 #import "ObjcAppLaunchHandler.h"
 #import <sys/sysctl.h>
+#import <objc/runtime.h>
 #import <UIKit/UIKit.h>
 #import <pthread.h>
 
 // `AppLaunchHandler` aims to track some times as part of the sequence described in Apple's "About the App Launch Sequence"
 // https://developer.apple.com/documentation/uikit/app_and_environment/responding_to_the_launch_of_your_app/about_the_app_launch_sequence
 
+#define IS_ACTIVE_PREWARM [NSProcessInfo.processInfo.environment[@"ActivePrewarm"] isEqual:@"1"]
+
 // A Read-Write lock to allow concurrent reads of TimeToApplicationDidBecomeActive, unless the initial (and only) write is locking it.
 static pthread_rwlock_t rwLock;
 // The framework load time  in seconds relative to the absolute reference date of Jan 1 2001 00:00:00 GMT.
 static NSTimeInterval FrameworkLoadTime = 0.0;
-// The time interval between the application starts and it's responsive and accepts touch events.
-static NSTimeInterval TimeToApplicationDidBecomeActive = 0.0;
+// The time when UIApplicationMain instantiante UIApplication.
+static NSTimeInterval UIApplicationInstantiateSingletonTime = 0.0;
+// The time when receiving UIApplicationDidBecomeActiveNotification.
+static NSTimeInterval UIApplicationDidBecomeActiveNotificationTime = 0.0;
+// The cold start threshold in seconds. Launch Time greater than this threshold should be considered as
+// pre-warmed.
+static NSTimeInterval ColdStartThreshold = 20;
 
-NS_INLINE NSTimeInterval QueryProcessStartTimeWithFallback(NSTimeInterval fallbackTime) {
-    NSTimeInterval processStartTime;
+NS_INLINE NSTimeInterval ProcessStartTime() {
     // Query the current process' start time:
     // https://www.freebsd.org/cgi/man.cgi?sysctl(3)
     // https://github.com/darwin-on-arm/xnu/blob/707bfdc4e9a46e3612e53994fffc64542d3f7e72/bsd/sys/sysctl.h#L681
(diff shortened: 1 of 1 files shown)


Sources/_Datadog_Private/ObjcAppLaunchHandler.m
```diff
@@ -56,32 +59,60 @@ @interface AppLaunchHandler : NSObject
...
+            class_getInstanceMethod(self, @selector(init)),
+            class_getInstanceMethod(self, @selector(dd_init))
+        );
+    });
+}
+
+- (instancetype)dd_init {
+    pthread_rwlock_wrlock(&rwLock);
+    UIApplicationInstantiateSingletonTime = CFAbsoluteTimeGetCurrent();
+    pthread_rwlock_unlock(&rwLock);
```
maxep, line 96: I don't think we actually need to acquire the lock here, this should get invoke long before we need to read `UIApplicationInstantiateSingletonTime`

Sources/_Datadog_Private/ObjcAppLaunchHandler.m
```diff
@@ -6,21 +6,30 @@
...
 #import "ObjcAppLaunchHandler.h"
 #import <sys/sysctl.h>
+#import <objc/runtime.h>
 #import <UIKit/UIKit.h>
 #import <pthread.h>
 
 // `AppLaunchHandler` aims to track some times as part of the sequence described in Apple's "About the App Launch Sequence"
 // https://developer.apple.com/documentation/uikit/app_and_environment/responding_to_the_launch_of_your_app/about_the_app_launch_sequence
 
+#define IS_ACTIVE_PREWARM [NSProcessInfo.processInfo.environment[@"ActivePrewarm"] isEqual:@"1"]
```
maxep, line 16: this variable is currently being evaluated in dogfooding.

Sources/_Datadog_Private/ObjcAppLaunchHandler.m
```diff
@@ -56,32 +57,60 @@ @interface AppLaunchHandler : NSObject
...
 
+@implementation UIApplication (Tracking)
+
++ (void)load {
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        method_exchangeImplementations(
+            class_getInstanceMethod(self, @selector(init)),
+            class_getInstanceMethod(self, @selector(dd_init))
+        );
```
ncreated, line 87: AFAIK, this way of swizzling [is dangerous/incorrect](https://newrelic.com/blog/best-practices/right-way-to-swizzle) as it doesn't manage the original  implementation right. In all our swizzlings we [use get/set implementation approach](https://github.com/DataDog/dd-sdk-ios/blob/2c9811e0967cc17da080f88534cf8dfdaaa68590/Sources/Datadog/Core/Swizzling/MethodSwizzler.swift#L58-L74). How are we sure this won't cause any wrong behaviours?

Swizzling the entry point to the program doesn't sound safe IMO. Are we sure this will always perform right, on all platforms and in all setups? E.g. what if someone else already swizzles `UIApplication.init` or the app overrides `UIApplication` class with custom `main` implementation? What about purely SwiftUI apps where `@main` is synthesized in struct?
```swift
@main
struct MyApp: App {
   // ... w/ or w/o init()
}
``` 
I think we need more proofs / tests before considering this safe. Among all our tests we don't really focus on how app is being launched (all are build with classic Storyboard + `AppDelegate` approach).

Sources/_Datadog_Private/ObjcAppLaunchHandler.m
```diff
@@ -56,32 +57,60 @@ @interface AppLaunchHandler : NSObject
...
+    });
+}
+
+- (instancetype)dd_init {
+    pthread_rwlock_wrlock(&rwLock);
+    UIApplicationInstantiateSingletonTime = CFAbsoluteTimeGetCurrent();
```
ncreated, line 93: How this time is different from "will finish launching" and "did finish launching" in app delegate? Let's say the app has **very long** "`dd_init` to _did become active_" measure, then:
* what can be the root cause?
* how actionable it would be (e.g. how it can be optimised)?
* will "`Datadog.initialize()` to _did become active_" also be very long in that case?

Sources/_Datadog_Private/ObjcAppLaunchHandler.m
maxep: [Very interesting read](https://newrelic.com/blog/best-practices/right-way-to-swizzle), better to sizzle with c function then 👍 

I intuitively don't find it safe either TBO, but I don't want to rely on intuition only and have concrete red flags. This PR is made for discussing those.
> E.g. what if someone else already swizzles UIApplication.init or the app overrides UIApplication class

Theoretically, it shouldn't behave differently than other method swizzling. I will validate that 👍 

> What about purely SwiftUI apps where `@main` is synthesized in struct?

`SwiftUI.App.init()` is called before `UIApplicationMain` from `main()`, so this measure will come after custom code in SwiftUI launch flow! But, from Apple documentation:
> Prewarming executes an app’s launch sequence up until, but not including, when main() calls UIApplicationMain(_:_:_:_:).

The goal here is to get a timestamp at an early stage of `UIApplicationMain()`. Custom code called during `SwiftUI.App.init()` should be called during cold start or during pre-warming. Therefore, they are out of a warm start measure.

> I think we need more proofs / tests before considering this safe.

💯 ! The problem is to get a time as close as possible as when `main()` invoke `UIApplicationMain` and I couldn't get there without swizzling. I'm noting the following validation:
- what if someone else already swizzles `UIApplication.init`
- app overrides UIApplication class ✅ -> SwiftUI uses it's own subclass and `dd_init` is properly called.
- custom implementation of `main()` (objc)

Sources/_Datadog_Private/ObjcAppLaunchHandler.m
```diff
@@ -56,32 +57,60 @@ @interface AppLaunchHandler : NSObject
...
 
     NSNotificationCenter * __weak center = NSNotificationCenter.defaultCenter;
-    id __block token = [center
-                        addObserverForName:UIApplicationDidBecomeActiveNotification
-                        object:nil
-                        queue:NSOperationQueue.mainQueue
-                        usingBlock:^(NSNotification *_){
-
-        pthread_rwlock_init(&rwLock, NULL);
+    id __weak __block token = [center addObserverForName:UIApplicationDidBecomeActiveNotification
```
buranmert, line 66: if this is `__weak`, who retains `token`? i guess this may crash

maxep: It would crash systematically [here](https://github.com/DataDog/dd-sdk-ios/pull/730/files#diff-12e60ffaca99dbb26bf5cb6c41b8b78575a811012b4eea1574d9fb07c7254279R73). It's retained by the notification center, from apple doc: 
> Notification center strongly holds this return value until you remove the observer registration.

PR #734, opened 2022-01-21
RUMM-1584 Remove RUMEvent model container

maxep:
### What and why?

Remove `RUMEvent` model container that is no longer needed given new rum schema.

> This PR keeps a remaining `RUMCrashEvent` container to add additional properties to `RUMErrorEvent` for describing crashes. This container will be removed in a second iteration of RUMM-1584 that adds a private schema for `rum-model-generator` to consume.

### Review checklist


Sources/Datadog/CrashReporting/CrashReportingFeature.swift
```diff
@@ -30,7 +30,7 @@ internal final class CrashReportingFeature {
     let carrierInfoProvider: CarrierInfoProviderType
     /// Publishes recent `RUMEvent<RUMViewEvent>` value so it can be persisted in `CrashContext`.
```
ncreated, line 31:
```suggestion
    /// Publishes recent `RUMViewEvent` value so it can be persisted in `CrashContext`.
```

ncreated (approved):
Looks all good 👍. Great idea to split it into separate PRs 👌. And happy to see how all this simplifies our tests 🏅.

I'd only advice doing string search on "`RUMEvent`" to see if we have no leftovers in code comments.

maxep:
@ncreated I've added [this change](https://github.com/DataDog/dd-sdk-ios/pull/734/commits/821d87363d94356b5297ffdcb3128c10a652971f) since your review to scrub crash error correctly.

ncreated:
> @ncreated I've added [this change](https://github.com/DataDog/dd-sdk-ios/pull/734/commits/821d87363d94356b5297ffdcb3128c10a652971f) since your review to scrub crash error correctly.

👍 was it captured by test failure? If no, I think we should add coverage for it. Or if this will be unreachable by removing crash event container in next PR, we can skip it.

maxep:
It won't be reachable in next PR 👍

PR #739, opened 2022-01-27
RUMM-1955 Multiple calls to trackDatadogEvents handled

buranmert:
### What and why?

If `trackDatadogEvents` method is called multiple times, it crashes the app.
The reason is that calling `WKUserContentController.add(messageHandler:, name:)` multiple times with the same `name` crashes the app.

### How?

`trackDatadogEvents` method removes already-added components before adding them.

### Review checklist

### DIFF
--- a/Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
+++ b/Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
@@ -53,6 +53,7 @@ public extension WKUserContentController {
                 rumEventConsumer: rumEventConsumer
             )
         )
+        removeScriptMessageHandler(forName: bridgeName)
         add(messageHandler, name: bridgeName)
 
         // WebKit installs message handlers with the given name format below
@@ -68,7 +69,9 @@ public extension WKUserContentController {
             .map { return "\"\($0)\"" }
             .joined(separator: ",")
 
+        let jsPrefix = "/* DatadogEventBridge */"
         let js = """
+        \(jsPrefix)
         window.\(bridgeName) = {
           send(msg) {
             \(webkitMethodName)(msg)
@@ -78,6 +81,14 @@ public extension WKUserContentController {
           }
         }
         """
+
+        let userScriptsWithoutDatadog = userScripts.filter { script in
+            return !script.source.starts(with: jsPrefix)
+        }
+        if userScriptsWithoutDatadog.count != userScripts.count {
+            removeAllUserScripts()
+            userScriptsWithoutDatadog.forEach { addUserScript($0) }
+        }
         addUserScript(
             WKUserScript(
                 source: js,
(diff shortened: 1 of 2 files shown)


Tests/DatadogTests/Datadog/RUM/WebView/WKUserContentController+DatadogTests.swift
```diff
@@ -50,12 +56,14 @@ class WKUserContentController_DatadogTests: XCTestCase {
 
         let initialUserScriptCount = controller.userScripts.count
 
-        controller.addDatadogMessageHandler(allowedWebViewHosts: ["datadoghq.com"], hostsSanitizer: mockSanitizer)
+        (0..<5).forEach { _ in
+            controller.addDatadogMessageHandler(allowedWebViewHosts: ["datadoghq.com"], hostsSanitizer: mockSanitizer)
+        }
```
ncreated, line 61: It's very unclear why we add this handler 5 times. If we want to test that _"given handler already registered, when registering new handler, it removes the previous one"_, this should be a separate test case.

Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
```diff
@@ -78,6 +81,14 @@ public extension WKUserContentController {
...
         }
         """
+
+        let userScriptsWithoutDatadog = userScripts.filter { script in
+            return !script.source.starts(with: jsPrefix)
+        }
+        if userScriptsWithoutDatadog.count != userScripts.count {
+            removeAllUserScripts()
+            userScriptsWithoutDatadog.forEach { addUserScript($0) }
+        }
```
ncreated, line 91: The problem we're solving is that calling `trackDatadogEvents(in:)` more than once results with a crash:
```swift
controller.trackDatadogEvents(in hosts: ["foo.com"])
controller.trackDatadogEvents(in hosts: ["foo.com"]) // 🔥 crash
```

But this solution looks very odd and unobvious: each time the API is called we're removing **all user scripts** and then re-adding all but Datadog one.

Can't we just early return in this API if it was already called?
```swift
func trackDatadogEvents(in hosts: Set<String>) {
   guard !isAlreadyTracked(self) else {
      userLogger.warn("`trackDatadogEvents(in:)` was called more than once for the same WebView. Second call will be ignored. Make sure you call it only once.")
      return
   }

   // ...
}
```

ncreated (changes requested):

Tests/DatadogTests/Datadog/RUM/WebView/WKUserContentController+DatadogTests.swift
```diff
@@ -61,6 +67,35 @@ class WKUserContentController_DatadogTests: XCTestCase {
         XCTAssertEqual(sanitization.warningMessage, "The allowed WebView host configured for Datadog SDK is not valid")
     }
 
+    func testItAddsUserScriptAndMessageHandlerMultipleTimes() throws {
```
ncreated, line 70: The implementation of this test is fine, but I don't understand its name - sounds like we expect the controller to add handler multiple times.
```suggestion
    func testWhenAddingMessageHandlerMultipleTimes_itIgnoresExtraOnesAndPrintsWarning() throws {
```

ncreated (approved):
👌

maxep (approved):

PR #742, opened 2022-02-01
RUMM-1769 Fix logs doc rendering

maxep:
### What and why?

This section is rendered as code when using tabs.

### Review checklist


alai97 (approved):
Looks good for docs!

PR #745, opened 2022-02-03
RUMM-1931 stopTrackingDatadogEvents added to avoid memory leaks

buranmert:
### What and why?

Without removing `WKScriptMessageHandlers` explicitly, they stay alive in memory even after deallocation of the webview

### How?

By calling `stopTrackingDatadogEvents`, our users can remove Datadog components from their `WKWebViewConfiguration` instances so that `DatadogMessageHandler` can be deallocated.

### Review checklist

### DIFF
--- a/Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
+++ b/Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
@@ -30,7 +30,7 @@ public extension WKUserContentController {
     /// Disables Datadog iOS SDK and Datadog Browser SDK integration.
     ///
     /// Removes Datadog's ScriptMessageHandler and UserScript from the caller.
-    /// _NOTE:_ This method **must** be called when the webview can be deinitialized.
+    /// - Note: This method **must** be called when the webview can be deinitialized.
     func stopTrackingDatadogEvents() {
         removeScriptMessageHandler(forName: DatadogMessageHandler.name)
 
--- a/Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
+++ b/Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
@@ -27,6 +27,22 @@ public extension WKUserContentController {
         addDatadogMessageHandler(allowedWebViewHosts: hosts, hostsSanitizer: HostsSanitizer())
     }
 
+    /// Disables Datadog iOS SDK and Datadog Browser SDK integration.
+    ///
+    /// Removes Datadog's ScriptMessageHandler and UserScript from the caller.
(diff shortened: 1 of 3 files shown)


Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
```diff
@@ -27,6 +27,22 @@ public extension WKUserContentController {
         addDatadogMessageHandler(allowedWebViewHosts: hosts, hostsSanitizer: HostsSanitizer())
     }
 
+    /// Disables Datadog iOS SDK and Datadog Browser SDK integration.
+    ///
+    /// Removes Datadog's ScriptMessageHandler and UserScript from the caller.
+    /// _NOTE:_ This method **must** be called when the webview can be deinitialized.
```
maxep, line 33: This won't render properly, better to use
```suggestion
    /// - Note: This method **must** be called when the webview can be deinitialized.
```

Tests/DatadogTests/Datadog/RUM/WebView/WKUserContentController+DatadogTests.swift
```diff
@@ -23,6 +23,12 @@ final class DDUserContentController: WKUserContentController {
     }
 }
 
+final class MockMessageHandler: NSObject, WKScriptMessageHandler {
+    func userContentController(_ userContentController: WKUserContentController, didReceive message: WKScriptMessage) {
+        return
+    }
```
maxep, line 29: /nit
```suggestion
    func userContentController(_ userContentController: WKUserContentController, didReceive message: WKScriptMessage) { }
```

Tests/DatadogTests/Datadog/RUM/WebView/WKUserContentController+DatadogTests.swift
```diff
@@ -96,6 +102,31 @@ class WKUserContentController_DatadogTests: XCTestCase {
         XCTAssertEqual(recordedLogMessages, Array(repeating: "`trackDatadogEvents(in:)` was called more than once for the same WebView. Second call will be ignored. Make sure you call it only once.", count: multipleTimes - 1))
     }
 
+    func testWhenStoppingTracking_itKeepsNonDatadogComponents() throws {
+        let controller = DDUserContentController()
+
+        controller.trackDatadogEvents(in: [])
+
+        let componentCount = 10
```
maxep, line 110: why make it 10? I guess adding a single or a couple of other scripts will cover this change

maxep (approved):
LGTM! I've minor suggestions tho

ncreated (approved):
👍 It would be also good to add `stopTrackingDatadogEvents()` in "Webviews integration" screen in Example app. Otherwise we don't have any use of this API outside unit tests.

PR #755, opened 2022-02-10
RUMM-1935 Make `consolePrint` public

maxep:
### What and why?

Allow to mutate global console print. 

### How?

Make `var consolePrint` public.

Mutating this var is already covered in [tests](https://github.com/DataDog/dd-sdk-ios/blob/master/Tests/DatadogTests/Datadog/DatadogTests.swift#L19L32). Still, this PR adds a test case to `LogConsoleOutputTests` to test default `printingFunction`.

### Review checklist

### DIFF
--- a/Sources/Datadog/Utils/Globals.swift
+++ b/Sources/Datadog/Utils/Globals.swift
@@ -11,9 +11,7 @@ import _Datadog_Private
 #endif
 
 /// Function printing `String` content to console.
-internal var consolePrint: (String) -> Void = { content in
-    print(content)
-}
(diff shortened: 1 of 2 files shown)


buranmert (approved):
as an alternative, in order to make sure `consolePrint` always prints to the console, we can do this:
```
public var onConsolePrint: (String) -> Void
internal var consolePrint: (String) -> Void = { onConsolePrint($0); print($0) }
```

but current PR is ✅ too as this is intended only for a very niche audience

maxep:
Yes, this is for really niche usage. We can definitely make the API more obvious if we plan on documenting it 👍

PR #761, opened 2022-02-22
Integration with CI Visibility tests

nachoBonafonte:
### What and why?

This PR enables a much deeper integration of RUM with CI Visibility product ([dd-sdk-swift-testing](https://github.com/DataDog/dd-sdk-swift-testing)). If a UI test instrumented with CI Visibility is run and a RUM sessions is configured, the session will be identified as a created by CI App and will be shown as part of the test.

There is a related [PR](https://github.com/DataDog/dd-sdk-swift-testing/pull/60) with the needed changes in the **dd-sdk-swift-testing** framework

### How?
RUM now detects on loading if is running under an instrumented test by checking if the environment variable `CI_VISIBILITY_TEST_EXECUTION_ID` is available, which contains the test identifier. If so it makes the following changes:
* Sets `session.type` to `ci-test`
* Injects `ci_visibility.test_execution_id` to every event
* Set `_dd.origin` to `ciapp-test` so that APM traces from RUM in CI Visibility mode can be distinguished.
* Send a message to a Mach Port initialized in the CI Visibility framework (TestMachPort) to notify RUM is enabled
* Creates a new Mach Port (RUMMachPort) that will receive a message when the test ends. RUM will flush the RUM events when the message is received.

### Review checklist

### DIFF
--- a/Sources/Datadog/FeaturesIntegration/CITestIntegration.swift
+++ b/Sources/Datadog/FeaturesIntegration/CITestIntegration.swift
@@ -7,39 +7,40 @@
 import Foundation
 
 private enum DDCFMessageID {
-    static let customTags: Int32 = 0x1111
+    static let setCustomTags: Int32 = 0x1111
     static let enableRUM: Int32 = 0x2222
     static let forceFlush: Int32 = 0x3333
 }
 
-internal enum CITestIntegration {
-    /// CI context to be attached to RUM events that identifies it was started by a CI test,
-    static var ciTestExecutionID: String? {
-        return ProcessInfo.processInfo.environment["CI_VISIBILITY_TEST_EXECUTION_ID"]
-    }
+internal class CITestIntegration {
+    /// Current and active integration with CIApp.
+    /// `nil` if the integration is not enabled.
+    static let active: CITestIntegration? = CITestIntegration()
 
-    /// Convinence function to check if RUM was started in a CIApp test
-    static func isEnabled() -> Bool {
-        return ciTestExecutionID != nil
-    }
+    /// CI context to be attached to RUM events that identifies that they were created in a CIApp test,
+    let ciTestExecutionID: String
+    /// RUMCITest model to be attached to events
+    let rumCITest: RUMCITest
+    /// Tag that must be added to spans and headers when running inside a CIApp test
+    let origin = "ciapp-test"
 
-    /// Origin value that must be set to Spans and headers to indicate that the trace was initiated by a CIApp test
-    static var origin: String? {
-        if isEnabled() {
-            return "ciapp-test"
+    private init?(processInfo: ProcessInfo = .processInfo) {
+        guard let testID = processInfo.environment["CI_VISIBILITY_TEST_EXECUTION_ID"] else {
+            return nil
         }
-        return nil
+        self.ciTestExecutionID = testID
+        self.rumCITest = RUMCITest(testExecutionId: ciTestExecutionID)
     }
 
     /// Entry point for running all the tasks needed for CIApp integration
-    static func startIntegration() {
+    func startIntegration() {
         startMessageListener()
         notifyRUMSession()
     }
 
     /// Notifies the CIApp framework that a RUM session is being started. It sends a message to a CFMessagePort that is
     /// created in the CIApp framework
-    private static func notifyRUMSession() {
+    private func notifyRUMSession() {
         let timeout: CFTimeInterval = 1.0
         guard let remotePort = CFMessagePortCreateRemote(nil, "DatadogTestingPort" as CFString) else {
             return
@@ -59,7 +60,7 @@ internal enum CITestIntegration {
 
     /// Creates a CFMessagePort that is used by the CIApp framework to notify that a test is going to finish, so all
     /// information must be flushed to the backend. It uses a non public API `internalFlushAndDeinitialize()`
-    private static func startMessageListener() {
+    private func startMessageListener() {
         func attributeCallback(port: CFMessagePort?, msgid: Int32, data: CFData?, info: UnsafeMutableRawPointer?) -> Unmanaged<CFData>? {
             switch msgid {
             case DDCFMessageID.forceFlush:
--- a/Sources/Datadog/FeaturesIntegration/CITestIntegration.swift
+++ b/Sources/Datadog/FeaturesIntegration/CITestIntegration.swift
@@ -1,0 +1,81 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+private enum DDCFMessageID {
+    static let customTags: Int32 = 0x1111
+    static let enableRUM: Int32 = 0x2222
+    static let forceFlush: Int32 = 0x3333
+}
+
+internal enum CITestIntegration {
+    /// CI context to be attached to RUM events that identifies it was started by a CI test,
+    static var ciTestExecutionID: String? {
+        return ProcessInfo.processInfo.environment["CI_VISIBILITY_TEST_EXECUTION_ID"]
+    }
+
+    /// Convinence function to check if RUM was started in a CIApp test
+    static func isEnabled() -> Bool {
+        return ciTestExecutionID != nil
+    }
+
+    /// Origin value that must be set to Spans and headers to indicate that the trace was initiated by a CIApp test
+    static var origin: String? {
+        if isEnabled() {
+            return "ciapp-test"
+        }
+        return nil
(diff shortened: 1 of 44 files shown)


maxep:
Very nice use of `CFMessagePortSendRequest`, and I like that it is implemented as a FeatureIntegration.
Tho, I'm a bit concern of making it a transverse dependency by using `CITestIntegration.ciTestExecutionID`, it will surely conflict with our effort in v2. Could it be injected somehow? by making it part of `RUMScopeDependencies` for instance 🤔 

Also, I'm fine with exposing flushing methods internally 👍

Sources/Datadog/FeaturesIntegration/CITestIntegration.swift
```diff
@@ -0,0 +1,81 @@
...
+        return ciTestExecutionID != nil
+    }
+
+    /// Origin value that must be set to Spans and headers to indicate that the trace was initiated by a CIApp test
+    static var origin: String? {
+        if isEnabled() {
+            return "ciapp-test"
+        }
+        return nil
+    }
```
ncreated, line 32: This static APIs model is slightly inconvenient and can be easily overlooked. Now it's a bit unbalanced:
* `isEnabled()` is considered in callee when computing `origin`,
* `isEnabled()` is considered in caller when calling `startIntegration()` in `RUMMonitor`.

How about changing it to object-oriented way, where "integration is not enabled" is indicated by `nil` value, e.g.:
```swift
internal class CITestIntegration {
   /// Current and active integration with CIApp. 
   /// `nil` if the integration is not enabled.
   static let active: CITestIntegration? = CITestIntegration()

   let ciTestExecutionID: String
   let origin = "ciapp-test"

   private init?(processInfo: ProcessInfo = .processInfo) {
      guard let testID = processInfo.environment["CI_VISIBILITY_TEST_EXECUTION_ID"] else {
         return nil
      }
      self.ciTestExecutionID = testID
   }

   func startIntegration() { /* ... */ }
}
```

This way the caller is given streamlined API and cannot make mistake on calling sth when this integration is disabled:
```swift
CITestIntegration.active?.startIntegration()
let origin = CITestIntegration.active?.origin
let testID = CITestIntegration.active?.ciTestExecutionID
// ...
```

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
```diff
@@ -128,6 +128,13 @@ internal class RUMResourceScope: RUMScope {
         }
         let resourceType: RUMResourceType = resourceKindBasedOnRequest ?? command.kind
 
+        let ciTest: RUMResourceEvent.CiTest? = {
+            if let testID = CITestIntegration.ciTestExecutionID {
+                return RUMResourceEvent.CiTest(testExecutionId: testID)
+            }
+            return nil
+        }()
```
ncreated, line 136: We can improve on this repetitive snippet in this PR if we make `CiTest` model "shared" in `RUMDataModels` definition. This way, we will have single model for all events, instead of separate nested model in each event.

Once we have a single model, we can create its single value and pass it through `dependencies` container available in all RUM scopes:
```swift
ciTest: dependencies.ciTest,
```

To make `CiTest` model shared, we need to configure it in `tools/rum-models-generator` and call `make make rum-models-generate`. I did it [here](https://github.com/DataDog/dd-sdk-ios/commit/ad51785d084c8ecc4ff70fd09c20d4efcbf615e6#diff-dd3caa4e6db960eecd95024f6d7df904948b8f53b1026c00dd19a9627317ac3e) - we might consider just merging it.

Sources/Datadog/FeaturesIntegration/CITestIntegration.swift
```diff
@@ -0,0 +1,81 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+private enum DDCFMessageID {
+    static let customTags: Int32 = 0x1111
```
ncreated, line 10: Do we use it anywhere 🤔💭? If we need it, I think the name could be a bit more descriptive - maybe we can add some verb 🙂?

ncreated (commented):
Looks really good 👍. I left few suggestions on making this integration fit our codebase a bit simpler.

Sources/Datadog/FeaturesIntegration/CITestIntegration.swift
nachoBonafonte: Good idea, looks much better. Will change to it

Sources/Datadog/RUM/RUMFeature.swift
```diff
@@ -105,7 +105,7 @@ internal final class RUMFeature {
                         device: commonDependencies.mobileDevice
                     ),
                     .ddAPIKeyHeader(clientToken: configuration.clientToken),
-                    .ddEVPOriginHeader(source: configuration.common.source),
+                    .ddEVPOriginHeader(source: configuration.common.origin ?? configuration.common.source),
```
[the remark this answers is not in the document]
nachoBonafonte: The reason for that is that we don't want to charge users by the usage of APM (and potentially also logs on the fututre) because of a test run, the way those charges are removed are by using the `origin` tag which must be transmitted in the header in case of end to end tests. This header is also used by synthetics for the same goal I think.

Sources/Datadog/FeaturesIntegration/CITestIntegration.swift
nachoBonafonte: The reason to keep the same enum used in the testing framework, but I can reduce it to only the needed values if you prefer. I will improve the description if we keep.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
nachoBonafonte: Yes please, merge it. That was the part of the code that looked worse for me, but I didn't know how to improve it.

ncreated: @nachoBonafonte feel free to merge my branch to yours 🙌. I don't want to mess up in your workflow 🙂.

Sources/Datadog/FeaturesIntegration/CITestIntegration.swift
ncreated: We can keep it, no problem 👍. Good for parity between definitions in both SDKs 👌.

Sources/Datadog/FeaturesIntegration/CITestIntegration.swift
nachoBonafonte: I have updated it with the last commit, code is much cleaner now

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
nachoBonafonte: Merged your branch and it looks much better now

Sources/Datadog/FeaturesIntegration/CITestIntegration.swift
nachoBonafonte: I kept but changed the name to `setCustomTags`

docs/trace_collection.md
```diff
@@ -216,206 +216,206 @@ Before data is uploaded to Datadog, it is stored in cleartext in the cache direc
...
+                                                    delegateQueue:nil];
+```
+{{% /tab %}}
+{{< /tabs >}}
 
-    This will trace all requests made with this `session` to `example.com` and `api.yourdomain.com` hosts (for example, `https://api.yourdomain.com/v2/users` or `https://subdomain.example.com/image.png`).
+This traces all requests made with this `session` to `example.com` and `api.yourdomain.com` hosts (for example, `https://api.yourdomain.com/v2/users` or `https://subdomain.example.com/image.png`).
 
-    **Note**: Tracing auto instrumentation uses `URLSession` swizzling, but it is opt-in: if you do not specify `firstPartyHosts`, no swizzling is applied.
+**Note**: Tracing auto-instrumentation uses `URLSession` swizzling and is opt-in. If you do not specify `firstPartyHosts`, swizzling is not applied.
```
urseberry, line 418: Is "swizzling" defined elsewhere, or is it a commonly understood term? It is a bit confusing to me, but this is outside my normal subject area.

urseberry (approved):

docs/trace_collection.md
nachoBonafonte: Yes, is a commonly understood term in iOS programming. Anyway this change was not supposed to be here so I reverted it

maxep (approved):
👍

ncreated (approved):
👌👏

PR #763, opened 2022-02-22
RUMM-2033 Add tvOS targets

maxep:
### What and why?

Build SDKs for tvOS.

### How?

Large portion of this PR changes are in `Datadog.xcodeproj`, I suggest to pull this branch in order to review the changes.

- Add `iOS` suffix to iOS targets
- Add `Datadog tvOS` target
- Add `DatadogObjc tvOS` target
- Add `DatadogCrashReporting tvOS` target
- Add `DatadogTests tvOS` target
- Add `DatadogCrashReportingTests tvOS` target

### Known Issue

Some test needs a host application to succeed, those tests now runs behind a complication flag until RUMM-2034.

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference

PR #764, opened 2022-02-23
RUMM-2033 tvOS Compatibility

maxep:
### What and why?

Add compilation and runtime flags for tvOS.

### How?

This PR only focuses on making codebase compatible with tvOS. Targets definition will come next.

### Review checklist

### DIFF
--- a/Sources/Datadog/Core/System/CarrierInfoProvider.swift
+++ b/Sources/Datadog/Core/System/CarrierInfoProvider.swift
@@ -4,7 +4,9 @@
  * Copyright 2019-2020 Datadog, Inc.
  */
 
+#if os(iOS)
 import CoreTelephony
+#endif
 
 /// Network connection details specific to cellular radio access.
 public struct CarrierInfo: Equatable {
@@ -48,7 +50,7 @@ internal protocol CarrierInfoProviderType {
 extension CarrierInfo.RadioAccessTechnology {
     init(ctRadioAccessTechnologyConstant: String) {
         switch ctRadioAccessTechnologyConstant {
-        #if !targetEnvironment(macCatalyst)
+        #if os(iOS)
         case CTRadioAccessTechnologyGPRS: self = .GPRS
         case CTRadioAccessTechnologyEdge: self = .Edge
         case CTRadioAccessTechnologyWCDMA: self = .WCDMA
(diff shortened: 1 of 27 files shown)


Sources/Datadog/Core/System/CarrierInfoProvider.swift
```diff
@@ -48,7 +50,7 @@ internal protocol CarrierInfoProviderType {
 extension CarrierInfo.RadioAccessTechnology {
     init(ctRadioAccessTechnologyConstant: String) {
         switch ctRadioAccessTechnologyConstant {
-        #if !targetEnvironment(macCatalyst)
+        #if os(iOS) && !targetEnvironment(macCatalyst)
```
fuzzybinary, line 53: Do we need that `!targetEnvironment(macCatalyst)` if we're checking specifically for iOS?

fuzzybinary (approved):

Sources/Datadog/Core/System/CarrierInfoProvider.swift
```diff
@@ -98,16 +100,13 @@ internal class CarrierInfoProvider: CarrierInfoProviderType {
     }
 }
 
-#if targetEnvironment(macCatalyst)
-
-/// Dummy provider for Mac Catalyst which doesn't support carrier info.
-internal struct MacCatalystCarrierInfoProvider: CarrierInfoProviderType {
+/// Dummy provider for platforms which doesn't support carrier info.
+internal struct NOPCarrierInfoProvider: CarrierInfoProviderType {
```
ncreated, line 104: 👍 for this renaming

Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
```diff
@@ -4,7 +4,7 @@
  * Copyright 2019-2020 Datadog, Inc.
  */
 
-#if DD_SDK_ENABLE_EXPERIMENTAL_APIS
+#if !os(tvOS) && DD_SDK_ENABLE_EXPERIMENTAL_APIS
```
ncreated, line 7: This basically means "no WebView tracking" / "sub apps" for tvOS. Do I see it right in the docs that [`WebKit` is unavailable on tvOS](https://developer.apple.com/documentation/webkit)?

ncreated (approved):
Looks good 👌.

The number of compiler conditions and platform/env branches now becomes worrying. To ensure that everything compiles and works fine on all platforms (iOS + tvOS), it would be great to consider adding smoke + unit tests for tvOS in `RUMM-1651`'s scope.

Sources/Datadog/Core/System/CarrierInfoProvider.swift
maxep: Indeed, good catch 👍

Sources/Datadog/FeaturesIntegration/WebView/WKUserContentController+Datadog.swift
maxep: Yes, WebKit is not supported on tvOS so hybrid apps are not possible on this platform.

maxep:
@ncreated Yes, unit tests are coming next in #766.
This PR is a chunk of #763 which successfully builds and tests (units) for tvOS. Tho I need to add tasks for integration and dependency managers tests!

PR #765, opened 2022-02-23
RUMM-2033 Rename iOS targets

maxep:
### What and why?

Add `iOS` suffix to iOS targets to prepare addition of `tvOS` targets.

### Review checklist

### DIFF
--- a/Datadog/Datadog.xcodeproj/project.pbxproj
+++ b/Datadog/Datadog.xcodeproj/project.pbxproj
@@ -3591,9 +3591,9 @@
 /* End PBXHeadersBuildPhase section */
 
 /* Begin PBXNativeTarget section */
-		61133B81242393DE00786299 /* Datadog */ = {
+		61133B81242393DE00786299 /* Datadog iOS */ = {
 			isa = PBXNativeTarget;
-			buildConfigurationList = 61133B96242393DE00786299 /* Build configuration list for PBXNativeTarget "Datadog" */;
+			buildConfigurationList = 61133B96242393DE00786299 /* Build configuration list for PBXNativeTarget "Datadog iOS" */;
 			buildPhases = (
 				61133B7D242393DE00786299 /* Headers */,
 				61133B7E242393DE00786299 /* Sources */,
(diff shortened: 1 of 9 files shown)


bitrise.yml
```diff
@@ -129,22 +129,22 @@ workflows:
     - xcode-test:
         title: Run unit tests for Datadog - iOS Simulator
         inputs:
-        - scheme: Datadog
-        - simulator_device: iPhone 11
+        - scheme: Datadog iOS
+        - destination: platform=iOS Simulator,name=iPhone 11,OS=latest
```
maxep, line 133: `simulator_device` has been replaced with `destination`, see [bitrise-steplib/steps-xcode-test/#185](https://github.com/bitrise-steplib/steps-xcode-test/pull/185)

fuzzybinary (approved):

Datadog/Datadog.xcodeproj/project.pbxproj
```diff
@@ -3591,9 +3591,9 @@
 /* End PBXHeadersBuildPhase section */
 
 /* Begin PBXNativeTarget section */
-		61133B81242393DE00786299 /* Datadog */ = {
+		61133B81242393DE00786299 /* Datadog iOS */ = {
```
ncreated, line 3594: I'm wondering if we should use whitespaces in target names. As long as we properly `""` handle it in our `xcconfigs`, scripts and tools, there should be no problem. Problems may arise if we miss it somewhere or one of our 3rd-party integration tools does (Cocoapods / Carthage). Calling it `Datadog-iOS` could be the same good, but save us some trouble (in theory). WDYT @maxep ?

ncreated (changes requested):
Other places that I know which need update are:
- [x] `make api-surface` call which [depends on the scheme name](https://github.com/DataDog/dd-sdk-ios/blob/598d3f02cc73672c24ae2612bc1fcecfe08382e4/Makefile#L102-L107)
- [ ] [`PROJECT_SCHEME` in `bitrise.yml` template](https://github.com/DataDog/dd-sdk-ios/blob/598d3f02cc73672c24ae2612bc1fcecfe08382e4/tools/nightly-unit-tests/bitrise.yml.src#L9) for nightly tests

Datadog/Datadog.xcodeproj/project.pbxproj
```diff
@@ -4973,7 +4973,7 @@
 					"@loader_path/Frameworks",
 				);
 				PRODUCT_BUNDLE_IDENTIFIER = com.datadogqh.Datadog;
-				PRODUCT_NAME = "$(TARGET_NAME:c99extidentifier)";
+				PRODUCT_NAME = Datadog;
```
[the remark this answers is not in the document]
maxep: Indeed, this will centralise product names and avoid duplication 👍

Datadog/Datadog.xcodeproj/project.pbxproj
maxep: 🤔 It's pretty common to use spaces in targets (see PLCR), I find it more human readable in Xcode. I guess it will be easy to catch if we miss `""` somewhere in our automation.
CP will create its own targets based on the podspec, and Carthage is based on [shared schemes](https://github.com/Carthage/Carthage#share-your-xcode-schemes), so I think we are good. Tho I'm open to reconsidering if we have some friction in maintenance, now that the product name does not depend on the target, it will be trivial to change. How does it sounds?

ncreated: Totally fine 👌. I was wondering if you see this dangerous as well or I'm just over cautious. Good to use spaces and see how it goes 👌.

ncreated (approved):
💪 all good

PR #766, opened 2022-02-23
RUMM-2033 Add tvOS Targets

maxep:
### What and why?

Adds tvOS framework and unit tests targets and schemes.

image

### Review checklist


ncreated (approved):
Excellent 💪. I checked in my local and everything that I can think of looks OK. The minor thing we need to adjust are
- [ ] targets for code coverage in tests:

Screenshot 2022-02-25 at 11 05 52

(please check other schemes too)

maxep:
Ah good catch, thank you 🙏

PR #767, opened 2022-02-25
RUMM-1939 add link to Apple's sandbox documentation

xgouchet:
### What and why?

Add a link to Apple's app sandbox documentation


maxep (approved):
👍

ncreated (approved):
👍 As this is only doc update, I think we can consider merging it directly to `master` (it will be available faster in our online doc). Otherwise, we will merge it automatically in next release 👌.

PR #769, opened 2022-02-25
RUMM-2010 Sanitize app name in User Agent

xgouchet:
### What and why?

Prevent any non alpha numeric character from app name when building the user agent


maxep (approved):
LGTM 👍

PR #772, opened 2022-02-28
RUMM-2037 SPM tvOS Support

maxep:
### What and why?

Support tvOS with Swift Package Manager.

### How?

Add `.tvOS(.v11)` to `Package.swift` platforms list.

### Tests

To run SPM integration tests on tvOS and simplify the implementation, [SPMProject.xcodeproj.src](dependency-manager-tests/spm/SPMProject.xcodeproj.src) now create its view programmatically. All targets have been renamed to better describe the build context.

image

### Review checklist


ncreated (approved):
All good 👌. Nice idea with replacing storyboard with shared UI definition from code 👍.

PR #773, opened 2022-02-28
RUMM-2045 Carthage tvOS Support

maxep:
### What and why?

Support tvOS with Carthage.

### How?

XCframework now includes tvos slices

### Tests

To run Carthage integration tests on tvOS and simplify the implementation, [CTProject.xcodeproj](dependency-manager-tests/carthage/CTProject.xcodeproj) now create its view programmatically. All targets have been renamed to better describe the build context.

image

### Review checklist


ncreated (approved):
I did some tests on my end, looks good 👍.

As we now add more slices to our `XCFrameworks`, we also need to:
- [ ] update the [`XCFrameworks` validation logic in the release tool](https://github.com/DataDog/dd-sdk-ios/blob/develop/tools/distribution/src/release/assets/gh_asset.py) (I think this can be a separate JIRA)

ncreated:
Actually, I can see the same issue of running `CTProject` on real device as I reported for `Example iOS` [here](https://github.com/DataDog/dd-sdk-ios/pull/774#pullrequestreview-901404329). Running `CTProject` on iPhone is not that important, although if it's a small issue, would be great to have it solved here as well 👍.

maxep:
Ah thanks, I will have a look 👍

maxep:
@ncreated it should be fixed now :)

I've also added RUMM-2054 for validating `xcframework`.

PR #775, opened 2022-03-04
RUMM-2035 Support tvOS Click Action

maxep:
⚠️ **Not Tested**: previous PRs need merging before we can add integration tests for this.
 
### What and why?

Instrument tvOS remote controls by reporting RUM click action on button press.

#### `UIKit`

When build for tvOS, `UIKitRUMUserActionsHandler` will automatically log `UIPressesEvent` for some specific button types. Other types of button can still be reported by implementing a custom predicate. In case of tvOS the predicate interface now takes the `UIPress` type as parameter:
```swift
public protocol UIKitRUMUserActionsPredicate {

    /// The predicate deciding if the RUM Action should be recorded.
    ///
    /// - Parameters:
    ///   - type: the `UIPress.PressType` which received the action.
    ///   - targetView: an instance of the `UIView` which received the action.
    /// - Returns: RUM Action if it should be recorded, `nil` otherwise.
    func rumAction(press type: UIPress.PressType, targetView: UIView) -> RUMAction?
}
```

#### `SwiftUI`

On tvOS, the `SwiftUI.View` extension defines the `trackRUMClickAction` which add the only available gesture `LongPressGesture` to the current view:

```swift
public extension SwiftUI.View {

    /// Monitor this view tap actions with Datadog RUM. An Action event will be logged after the minimum duration of presss.
    ///
    /// - Parameters:
    ///   - name: The action name.
    ///   - attributes: custom attributes to attach to the View.
    ///   - minimumDuration: The minimum duration of the long press that must elapse to complete the click action.
    /// - Returns: This view after applying a `ViewModifier` for monitoring the view.
    func trackRUMClickAction(
        name: String,
        attributes: [AttributeKey: AttributeValue] = [:],
        minimumDuration: Double = 0.01
    ) -> some View
}
```
### Review checklist


Sources/Datadog/RUM/Instrumentation/Actions/UIKit/UIKitRUMUserActionsHandler.swift
```diff
@@ -48,33 +42,63 @@ internal class UIKitRUMUserActionsHandler: UIEventHandler {
...
+        guard event.allPresses.count == 1, let press = event.allPresses.first else {
+            return nil // not a single press event
+        }
+        guard press.phase == .ended else {
+            return nil // not in `.ended` phase
+        }
+        guard let view = press.responder as? UIView, isSafeForPrivacy(view) else {
+            return nil // no valid view
+        }
+        guard let action = predicate.rumAction(press: press.type, targetView: view) else {
```
maxep, line 91: We don't need to find the best target on tvOS because all focused views are interactive.

Sources/Datadog/RUM/Instrumentation/Actions/SwiftUI/SwiftUIActionModifier.swift
```diff
@@ -51,6 +51,48 @@ public extension SwiftUI.View {
...
+    ///
+    /// - Parameters:
+    ///   - name: The action name.
+    ///   - attributes: custom attributes to attach to the View.
+    ///   - minimumDuration: The minimum duration of the long press that must elapse to complete the click action.
+    /// - Returns: This view after applying a `ViewModifier` for monitoring the view.
+    func trackRUMClickAction(
+        name: String,
+        attributes: [AttributeKey: AttributeValue] = [:],
+        minimumDuration: Double = 0.01
```
ncreated, line 90: `TimeInterval` (which is a typealias to `Double`) could be more relevant here, WDYT? Also, adding "(in seconds)" to the `minimumDuration` comment could be handy.

Sources/Datadog/RUM/Instrumentation/Actions/UIKit/UIKitRUMUserActionsHandler.swift
```diff
@@ -48,33 +42,63 @@ internal class UIKitRUMUserActionsHandler: UIEventHandler {
...
+        subscriber.process(command: command)
     }
 
-    // MARK: - Events Filtering
+    // MARK: - Event Processing
 
-    /// Returns the `UITouch` for given `event` only if the event describes the "single tap ended" interaction.
-    private func captureSingleTouch(event: UIEvent) -> UITouch? {
+    #if !os(tvOS)
+    private func command(from event: UIEvent) -> RUMAddUserActionCommand? {
```
ncreated, line 51: It won't be possible to run unit tests for both implementation on single Simulator, no? How about using OCP and passing this logic as configuration 🤔💭?
```swift
let tvOSCommandFactory: (UIEvent) -> RUMAddUserActionCommand = ...
let ioOSCommandFactory: (UIEvent) -> RUMAddUserActionCommand = ...

let handler = UIKitRUMUserActionsHandler(..., commandFactory:)
```
This way we'll only need to `#if`-compile the initialization of this handler, but both platforms logic can be tested with single Simulator.

ncreated (commented):
Looks good, waiting for tests 👍.

Sources/Datadog/RUM/Instrumentation/Actions/SwiftUI/SwiftUIActionModifier.swift
maxep: Indeed, I need to specify that in the doc 👍 
For the duration type, I agree it would be more relevant! I've based this signature on the type used in [LongPressGesture](https://developer.apple.com/documentation/swiftui/longpressgesture) which use `Double`.

Sources/Datadog/RUM/Instrumentation/Actions/UIKit/UIKitRUMUserActionsHandler.swift
maxep: We can now test both logic on a single simulator. This also imply that both predicates, for `UITouch` and `UIPress` events are accessible on all platforms.

maxep:
I've added unit tests for `UIKit` instrumentation, tvOS integration test for `SwiftUI` action will be added in another to keep this one small.

Sources/Datadog/RUM/Instrumentation/Actions/SwiftUI/SwiftUIActionModifier.swift
```diff
@@ -51,6 +51,50 @@ public extension SwiftUI.View {
...
+    /// Custom attributes to attach to the Action.
+    let attributes: [AttributeKey: AttributeValue]
+
+    func body(content: Content) -> some View {
+        content.simultaneousGesture(
+            LongPressGesture(minimumDuration: minimumDuration).onEnded { _ in
+                Global.rum.addUserAction(type: .click, name: name, attributes: attributes)
+            }
+        )
+    }
```
maxep, line 75: Ok, this **does not work** actually. `simultaneousGesture` does not seem to work properly!

ncreated (approved):
LG! 👌 I left a suggestion on mitigating the breaking change in public API. Other than that - all good 👌.

Sources/Datadog/RUM/Instrumentation/Actions/UIKit/UIKitRUMUserActionsPredicate.swift
```diff
@@ -24,29 +24,41 @@ public struct RUMAction {
...
+public typealias UIKitRUMUserActionsPredicate = UITouchRUMUserActionsPredicate
+#endif
+
 /// The predicate deciding if a given RUM Action should be recorded.
 ///
-/// When the app is running, the SDK will ask the implementation of `UIKitRUMUserActionsPredicate` if any noticed user action on the target view should
+/// When the app is running, the SDK will ask the implementation of `UITouchRUMUserActionsPredicate` if any noticed user action on the target view should
 /// be considered as the RUM Action. The predicate implementation should return RUM Action parameters if it should be recorded or `nil` otherwise.
-public protocol UIKitRUMUserActionsPredicate {
+public protocol UITouchRUMUserActionsPredicate {
```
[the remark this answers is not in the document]
maxep: This will be transparent for the user: `UIKitRUMUserActionsPredicate` is still defined but as a [typealias](https://github.com/DataDog/dd-sdk-ios/pull/775/files#diff-b43a7e3806ceed834ba95f26ea02aaf2d241e6575bea97ca81d619d46b68e1f7R27-R31)

PR #779, opened 2022-03-15
RUMM-2010 Fix User-Agent tests flakiness

ncreated:
### What and why?

⚙️ This PR fixes user agent header tests flakiness.

### How?

In #769 we added sanitisation for the "app name" component in UA header. This change wasn't reflected in dependant HTTP message tests, which assert the value of UA header. The random input generated for these tests was allowing whitespaces, now sanitised in upstream logic.

Because we test random input sanitisation separately (in UA sanitisation tests), to fix the problem here, I only adjusted characters range for the input used in HTTP message tests.

### Review checklist


maxep (approved):
🙏

PR #780, opened 2022-03-15
RUMM-2020 Generate models from root schema file

maxep:
### What and why?

The `rum-models-generator` cli is too opinionated on the schema files tree, with recent changes in [rum-events-format](https://github.com/DataDog/rum-events-format) and the addition of telemetry events, the generator needs to be updated to digest the events formats from the root definition defined in [rum-events-format](https://github.com/DataDog/rum-events-format/blob/master/rum-events-format.json).

### How?

The `rum-models-generator` now takes the `rum-events-format.json` file path as an argument and is able to transverse schemas by navigating in the directories based on the `oneOf` references. This is possible because `$ref` value is always the relative path to the referred schema.

[`RUMDataModels.swift`](https://github.com/DataDog/dd-sdk-ios/compare/maxep/RUMM-2020/telemetry-schema?expand=1#diff-f36112341c95ac45cd547c02b7b296229ee812ad7d7b5eb90639b0823465bca0) and [`RUMDataModels+objc.swift`](https://github.com/DataDog/dd-sdk-ios/compare/maxep/RUMM-2020/telemetry-schema?expand=1#diff-32a755d205eba305d26630975910b99e01d740b29ebbd6dbc14c234c07b1fd6b) reflect changes in [rum-events-format](https://github.com/DataDog/rum-events-format) `master` branch, and the new generator implementation has an impact on the order of models definitions.

### Tests

Test fixtures are now defined in files and access as resources from the test bundle to allow testing against a real folder structure.

### Review checklist

### DIFF
--- a/Sources/Datadog/RUM/DataModels/RUMDataModels.swift
+++ b/Sources/Datadog/RUM/DataModels/RUMDataModels.swift
@@ -8,783 +8,6 @@
 
 internal protocol RUMDataModel: Codable {}
 
-/// Schema of all properties of a View event
-public struct RUMViewEvent: RUMDataModel {
-    /// Internal properties
-    public let dd: DD
-
-    /// Application properties
-    public let application: Application
-
-    /// CI Visibility properties
-    public let ciTest: RUMCITest?
-
-    /// Device connectivity properties
-    public let connectivity: RUMConnectivity?
-
-    /// User provided context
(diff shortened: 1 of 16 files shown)


ncreated (approved):
Great work 🏅! It's way more flexible now 💪. And good idea with moving fixtures to separate files 👌.

PR #783, opened 2022-03-18
RUMM-2054 Build xcframework with tvos slices

maxep:
### What and why?

Build and validate XCFrameworks with tvOS slices for sdk versions greater or equal to `1.10.0`.

### How?

`carthage build --platform iOS,tvOS --use-xcframeworks`

### Review checklist

### DIFF
--- a/tools/distribution/src/release/assets/gh_asset.py
+++ b/tools/distribution/src/release/assets/gh_asset.py
@@ -14,25 +14,25 @@ from src.utils import remember_cwd, shell, read_sdk_version, read_xcode_version
 from src.release.directory_matcher import DirectoryMatcher
 from src.release.semver import Version
 
+min_cr_version = Version.parse('1.7.0')
+min_tvos_version = Version.parse('1.10.0')
 
 class XCFrameworkValidator:
     name: str
 
-    def should_be_included(self, in_version: Version) -> bool:
-        pass
-
-    def validate(self, zip_directory: DirectoryMatcher):
+    def validate(self, zip_directory: DirectoryMatcher, in_version: Version) -> bool:
         pass
 
 
 class DatadogXCFrameworkValidator(XCFrameworkValidator):
     name = 'Datadog.xcframework'
 
-    def should_be_included(self, in_version: Version):
-        return True  # always expect `Datadog.xcframework`
+    def validate(self, zip_directory: DirectoryMatcher, in_version: Version) -> bool:
+        # always expect `Datadog.xcframework`
 
-    def validate(self, zip_directory: DirectoryMatcher):
-        zip_directory.get('Datadog.xcframework').assert_it_has_files([
+        dir = zip_directory.get('Datadog.xcframework')
+
+        dir.assert_it_has_files([
             'ios-arm64',
             'ios-arm64/BCSymbolMaps/*.bcsymbolmap',
             'ios-arm64/dSYMs/*.dSYM',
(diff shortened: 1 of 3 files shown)


tools/distribution/src/release/assets/gh_asset.py
```diff
@@ -138,7 +148,6 @@ def validate(self, zip_directory: DirectoryMatcher):
     DatadogObjcXCFrameworkValidator(),
     DatadogCrashReportingXCFrameworkValidator(),
     CrashReporterXCFrameworkValidator(),
-    KronosXCFrameworkValidator(),
```
ncreated, line 141: By removing Kronos validator we won't be able to re-build binaries prior to `1.9.0`. It will fail because of finding unexpected `.xcframework` inside `.zip` archive:
```python
dm.assert_number_of_files(expected_count=validated_count)  # assert there are no other files
```

tools/distribution/src/release/assets/gh_asset.py
```diff
@@ -45,6 +45,19 @@ def validate(self, zip_directory: DirectoryMatcher):
...
+            'tvos-arm64/dSYMs/*.dSYM',
+            'tvos-arm64/**/arm64.swiftinterface',
+            'tvos-arm64/**/arm64-apple-ios.swiftinterface',
+
+            'tvos-arm64_x86_64-simulator',
+            'tvos-arm64_x86_64-simulator/dSYMs/*.dSYM',
+            'tvos-arm64_x86_64-simulator/**/arm64.swiftinterface',
+            'tvos-arm64_x86_64-simulator/**/arm64-apple-ios-simulator.swiftinterface',
+            'tvos-arm64_x86_64-simulator/**/x86_64.swiftinterface',
+            'tvos-arm64_x86_64-simulator/**/x86_64-apple-ios-simulator.swiftinterface',
```
ncreated, line 60: These slices won't be available when running against `1.9.0` and earlier. I think we need to leverage / extend the logic around `should_be_included` somehow:
```python
def should_be_included(self, in_version: Version):
    return True  # always expect `Datadog.xcframework`
```

ncreated, line 60: These slices won't be available when running against `1.9.0` and earlier. I think we need to leverage / extend the logic around `should_be_included` somehow:
```python
def should_be_included(self, in_version: Version):
    return True  # always expect `Datadog.xcframework`
```

ncreated (changes requested):
These changes are not well compatible with the way we use `distribution` tool 💡.

Please mind that Module Stability is broken in `dd-sdk-ios` (due to https://bugs.swift.org/browse/SR-14195) and we can't ship portable XCFrameworks (more on this in our Confluence).

Due to lack of Module Stability we must be able to re-build XCFrameworks for arbitrary Xcode versions and sometimes do it on-demand (e.g. see [`1.9.0` release](https://github.com/DataDog/dd-sdk-ios/releases/tag/1.9.0)). This is the reason why Kronos XCFramework validator is still a part of the `distribution` tool (for the sake of re-building `1.8.x` or earlier. For the same reason, `tvOS` slices must be rather validated conditionally in `1.10.0` and above.

Ensuring this compatibility is a bit inconvenient and adds more work to configuring the tool, but this is the only "smooth" way we found so far 😞. In the past we tried using the tool version from tag instead of sourcing it from `master`, but this required even more hacks and required manual interventions for almost every release.

maxep:
Thanks for the clarification @ncreated 👍  I've made the change to be able to test tvos slices based on the sdk version. It involves changes in `XCFrameworkValidator` and `GHAsset` interfaces.

ncreated (approved):
👌✨

PR #794, opened 2022-03-29
RUMM-2025 Filter out unrecognized trailing stack frame in `error.stack`

ncreated:
### What and why?

📦 This PR aims at improving the symbolicated `error.stack` by removing trailing `???` frame (if present). 

Going from this:
Screenshot 2022-03-29 at 14 09 37

to this:
Screenshot 2022-03-29 at 14 35 32

The root cause of `???` is lack of library base address resolved for the frame. This makes its symbolication impossible. This line describes the very entry point to the code execution (even before `main`), so nothing actionable. To keep the symbolicated stack clean and avoid false positive symbolication errors, we truncate it.

### How?

Only the last frame is dropped and only if it couldn't be symbolicated.

### Review checklist


maxep (approved):
LGTM 👍

PR #801, opened 2022-04-05
RUMM-2022 Add method to send telemetry events

xgouchet:
### What and why?

Provide a method to send telemetry events


maxep (approved):
LGTM 👍

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
```diff
@@ -85,7 +87,11 @@ internal class RUMApplicationScope: RUMScope, RUMContextProvider {
             startInitialSession()
         }
```
ncreated, line 88: Note, that because of:
```swift
   if sessionScope == nil {
      startInitialSession()
   }
```
sending telemetry event as the very first thing in the app will be starting RUM session for the user. It can either mean starting the session earlier than it would start normally, or even starting it in the app where it won't be normally started. I wonder if this is ok 🤔?

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
```diff
@@ -123,4 +129,42 @@ internal class RUMApplicationScope: RUMScope, RUMContextProvider {
...
+        let dateCorrection = dependencies.dateCorrector.currentCorrection
+        let actionId = context.activeUserActionID?.toRUMDataFormat
+        let viewId = context.activeViewID?.toRUMDataFormat
+        let event = TelemetryDebugEvent(
+            dd: TelemetryDebugEvent.DD(),
+            action: actionId.flatMap { .init(id: $0) },
+            application: .init(id: context.rumApplicationID),
+            date: dateCorrection.applying(to: command.time).timeIntervalSince1970.toInt64Milliseconds,
+            service: "dd-sdk-ios",
+            session: .init(id: context.sessionID.toRUMDataFormat),
```
ncreated, line 143: The `context.sessionID` will be `00000000-0000-0000-0000-000000000000` if the session is not sampled. Also, this event will be sent no matter of session sampling as we write it to the output unconditionally. Is this what we expect and won't this zero-ed session ID be causing any problems on BE? 🤔

ncreated (approved):
Looks good 👍, although I asked few questions on the possible impact.

IMO it would be healthy to separate telemetry APIs from `RUMMonitor` at some point. Otherwise, it would mean using `Global.rum` directly across all our codebase and introducing tight coupling between all components and RUM.

Separating "telemetry monitor" and still writing and interleaving events in RUM batches should be easy - we don't need `RUMMonitor` / `RUMScope` to use RUM events output and to read RUM context. These things are already isolated.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
xgouchet: Indeed, in such a case we should send a null session ID, I'll update the PR. Sending telemetry events when the session is sampled out is indeed something we want to keep (in order to detect issues even for unkept sessions)

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
xgouchet: Yes I'll try and avoid this as well

PR #809, opened 2022-04-07
Mitigate refresh rate test flakiness

maxep:
### What and why?

Main thread overhead test is flaky when measuring refresh rate.

### How?

Allow sample count to be higher than target.
Improve test failure description for future investigation.

### Review checklist

### DIFF
--- a/Tests/DatadogTests/Datadog/RUM/RUMVitals/VitalRefreshRateReaderTests.swift
+++ b/Tests/DatadogTests/Datadog/RUM/RUMVitals/VitalRefreshRateReaderTests.swift
@@ -34,21 +34,31 @@ class VitalRefreshRateReaderTests: XCTestCase {
 
             run(
                 mainThreadWork: mainThreadWork,
-                until: { measure.currentValue.sampleCount == targetSamplesCount },
+                until: { measure.currentValue.sampleCount >= targetSamplesCount },
                 completion: {
                     reader.unregister(measure)
                     completion.fulfill()
                 }
             )
 
-            wait(for: [completion], timeout: 10)
+            let result = XCTWaiter().wait(for: [completion], timeout: 10)
+
+            switch result {
+            case .completed:
+                break // all good
+            case .timedOut:
+                XCTFail("VitalRefreshRateReader exceededed timeout with \(measure.currentValue.sampleCount)/\(targetSamplesCount) recorded samples")
+            default:
+                XCTFail("XCTWaiter unexpected failure: \(result)")
+            }
         }
 
         // Given
         let lowOverhead = { /* no-op */ } // no overhead in succeeding runloop runs
         let lowOverheadMeasure = VitalPublisher(initialValue: VitalInfo())
 
-        let highOverhead = { Thread.sleep(forTimeInterval: 0.02) } // 0.02 overhead in succeeding runloop runs
+        var highOverheadRunCount = 0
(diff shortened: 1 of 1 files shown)


Tests/DatadogTests/Datadog/RUM/RUMVitals/VitalRefreshRateReaderTests.swift
```diff
@@ -41,7 +41,7 @@ class VitalRefreshRateReaderTests: XCTestCase {
                 }
             )
 
-            wait(for: [completion], timeout: 10)
+            wait(for: [completion], timeout: 20)
```
ncreated, line 44: When looking into CI app data for this test, I'm not sure if bumping the timeout will have any effect. In average, this test is passing with `~1s` duration, so the problem is clearly not "running close to the timeout":

Screenshot 2022-04-08 at 10 15 32

Also, the timeout is responsible for 9 of 11 failures during last month:

Screenshot 2022-04-08 at 10 41 26

The other two failures executed in `1s` and `2s` with recording unexpected measurement.

---

This leads me to the conclusion that we have 2 separate problems in:
```swift
record(lowOverheadMeasure, mainThreadWork: lowOverhead)
record(highOverheadMeasure, mainThreadWork: highOverhead)
```
1. Sometimes recording measurement takes 10x more than usual, so exceeds the timeout before fulfilling `until: { measure.currentValue.sampleCount == targetSamplesCount }` condition.
2. Sometimes it completes as expected, but recorded measurement is wrong in `XCTAssertGreaterThan(expectedHighFPS, expectedLowFPS)`.

I think that we need to answer following when moving forward:
* **What does it mean when the timeout is exceeded, but the condition is not met?** 

Maybe the `measure.currentValue.sampleCount` jumps above `targetSamplesCount` for some reason while the condition is `==`? 💡 I think it would be very explaining if we report the `measure.currentValue.sampleCount` as part of the failure message, e.g.:
```
Exceeded timeout of 10 seconds, with unfulfilled expectations: "Complete measurement" and receiving 20 out of 30 expected samples
```
This can be done if we use `XCTWaiter()` and `XCTFail` APIs [like we do here](https://github.com/DataDog/dd-sdk-ios/blob/34748605ea313798674b062d98e8cdf4d9394b72/Tests/DatadogTests/Datadog/Mocks/ServerMock.swift#L200).

* **Why does it measure `59.99(9)` in `expectedLowFPS`?**

This means we're receiving almost max FPS measurement although we clearly block the main thread with `Thread.sleep(forTimeInterval: 0.02)` 🤨. IMO, to make it cleaner we could record the number of times the `highOverhead` block was called. If it turns out that we're calling it, but still reporting `~60FPS`, that might indicate a problem with the implementation and not test flakiness.

WDYT @maxep ?

ncreated (commented):
IMO it's OK to give it a chance with higher timeout 👍. Although, I'm not entirely sure if this fixes the problem and if we have only one problem in this test (see comment).

Tests/DatadogTests/Datadog/RUM/RUMVitals/VitalRefreshRateReaderTests.swift
maxep: Indeed, it makes total sense! Thanks for collecting the data 👍 
I will revert and add more details to test failures. And I will see if we can fall into `measure.currentValue.sampleCount` greater than `targetSamplesCount`

ncreated (approved):
🙌 Looks great, now we should have more insight on this problem 🕵️.

PR #811, opened 2022-04-08
RUMM-2029 Make integration tests not depend on the number of view updates

ncreated:
### What and why?

📦 This PR updates our integration tests so they no longer depend on the number of `RUMViewEvents` sent during each RUM scenario. This is a preliminary enhancement for _"RUMM-2029 Reduce number of view updates"_ coming in next PR.

The old approach of awaiting for certain `view.resource.count` or `view.action.count` in integration tests was too fragile as it strictly depends on the number of view events we send. In upcoming `RUMM-2029` work we can't guarantee presence of every "intermediate" update, so in case of last view being not stopped, we will miss some events in integration tests:

Screenshot 2022-04-11 at 09 35 25

### How?

In upcoming `RUMM-2029` work we will reduce the number of "intermediate" updates, but keep the "initial" and "terminal" ones. With that in mind, and knowing the RUM principle of _"starting a new view stops the previous one"_, integration tests are now updated to always start and stop an "END VIEW" as the final step of a test scenario:

Screenshot 2022-04-11 at 09 38 35

This guarantees that "terminal" update for the last important view is always sent and all its counts are set correctly.

To avoid implementing "END VIEW" at the end of each RUM test scenario, I added its invocation logic in UI tests definition:
```swift
// in DatadogIntegrationTests:

let app = ExampleApplication()
app.launchWith(...)

// ...

try app.endRUMSession() // <-- new
```

Because UITests runner (`DatadogIntegrationTests`) and app-under-tests (`Example`) are two separate processes, I use [`CFMessagePort`](https://developer.apple.com/documentation/corefoundation/cfmessageport-rs2) API for doing IPC. It sends a basic message from one process and receives it in another. Upon receiving `endRUMSession = 0x1111` message, the "END VIEW" is started & stopped in the `Example` app.

### Review checklist

- [ ] ~Add CHANGELOG entry for user facing change.~


maxep (approved):
Thank you for the level of documentation in this PR description and in code 👌 it's crystal clear

PR #814, opened 2022-04-12
Merge 1.10.0 to `develop`

ncreated:
### What and why?

Merging `release/1.10.0` into `develop`.

⚠️ The CI failure in this PR is expected and not dangerous - it fails in package manager tests, where we assert `tvOS` compatibility for current branch. This compatibility is implemented on `develop` but not yet provided on `release/1.10.0`, so _"does `release/1.10.0` compile fine for `tvOS`"_ assertion must fail. There is no easy workaround. This won't happen in next and further releases.

⚠️⚠️ Actually, there seems to be a bug in Bitrise Checks for GH. If one branch (`release/1.10.0`) has 2 open PRs, then both receive the same status. I restarted the `release/1.10.0` → `master` job and it overwrited the status in this PR 😞.

Screenshot 2022-04-12 at 17 46 42

### Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+# 1.10.0 / 04-12-2022
+
+### Changes
+
 * [FEATURE] Web-view tracking. See [#729][]
 * [BUGFIX] Strip query parameters from span resource. See [#728][]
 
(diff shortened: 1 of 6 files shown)


maxep (approved):
🚀

PR #816, opened 2022-04-13
Fix `CHANGELOG.md` after `1.10.0` release

ncreated:
### What and why?

⚙️🧰 This PR fixes the `CHANGELOG.md` after:
* #814 

It was altered by the git conflict resolution strategy we define in [`.gitattributes`](https://github.com/DataDog/dd-sdk-ios/blob/develop/%20.gitattributes):
```
CHANGELOG.md merge=union
```
### How?

Reviewed all entries manually and fixed them to reflect released / unreleased items.

### Review checklist

- [ ] ~Feature or bugfix MUST have appropriate tests (unit, integration)~
- [ ] ~Make sure each commit and the PR mention the Issue number or JIRA reference~
- [ ] ~Add CHANGELOG entry for user facing change.~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,17 +1,18 @@
 # Unreleased
 
+* [FEATURE] Integration with CI Visibility Tests. See[#761][]
+* [FEATURE] Add tvOS Support. See [#793][]
+* [FEATURE] Add data encryption interface on-disk data storage. See [#797][]
+* [BUGFIX] Stop reporting pre-warmed application launch time. See [#789][]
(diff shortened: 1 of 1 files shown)


ncreated:
@maxep I think we should be tracking `-alpha` and `-beta` releases in the `CHANGELOG.md`. Otherwise, we will be running into this manual fixes each time we do non-public release and it is very easy to make mistake.

maxep (approved):
It's either that, or we keep a single `release/*` branch opened where we tag `-alpha` `-beta` and `-rc`, then we fix changelog when merging to `develop`.

PR #817, opened 2022-04-15
RUMM-2110 Improve CI performance

ncreated:
### What and why?

⚙️🔨  This PR proposes an improvement to our CI pipelines configuration. The goal is to reduce the duration of CI builds for different triggers. Recently this duration crossed the bar of 60 minutes, which clearly becomes a bottleneck in our workflow.

Average build duration (p95) for last 60 days (by "phases"):

Screenshot 2022-04-14 at 13 24 54

Significant phases are:
* "setup" (~3 minutes) - preparing CI environment (installing PLCR, `dd-swift-testing` and CLIs),
* "unit tests" (~12.5 minutes) - running unit tests for iOS and tvOS (all products),
* "integration tests" (~18 minutes) - running integration tests for all products,
* "smoke tests" (~30 minutes) - testing SDK compatibility with Cocoapods, Carthage and SPM.

**Before**, these phases were always ran one after another for every pull request. Even if the PR was only about docs update, its overall duration had to be ~60 minutes.

**Now**, these phases are selectively enabled/disabled, depending on the PR configuration and content. For example, if a pull request includes only docs update, none of tests will be ran, limiting its CI duration to ~5 minutes. On merges to `develop` and `master` it will always run unit and integration tests (~30 minutes). The heaviest phase (smoke tests) will be ran nightly. For all PRs the automation will enable phases based on the _"Custom CI job"_ checklist in the PR description and / or infer it from the list of files changed in the PR.

### How?

When it comes to enabling phases based on the change context, we need:
* wisely chosen **default phases for each trigger** (pull requests / commits on `develop` / release tags / ...);
* ability to **override default phases for pull requests** (e.g. for maintenance of given phase);
* a mechanism to **enable phases automatically** on certain changes (to spot issues early in PRs, not in nightly builds).

Unfortunately, Bitrise CI is yet limited when it comes to last two points. Although it offers the [Selective Builds feature](https://devcenter.bitrise.io/en/builds/configuring-build-settings/selective-builds.html#ot-lst-cnt) that could be used to enable workflows if matching a pattern in the file path, it needs to be configured globally on bitrise.io, not in `bitrise.yml`. Even if using Selective Builds, it won't solve the problem of running particular phase for maintenance purpose or bypassing file patterns for other reason. [Unlike GitLab](https://docs.gitlab.com/ee/ci/pipelines/#add-manual-interaction-to-your-pipeline) and some other providers, Bitrise doesn't support manual triggers and we need to [start builds manually](https://devcenter.bitrise.io/en/builds/starting-builds/starting-builds-manually.html#ot-lst-cnt) or use 3rd party integrations (Slack/CLI) to customise build phases.

Given these limitations, I had to add a custom script for enabling `bitrise.yml` workflows conditionally.

I introduced a new workflow (named: `run_conditioned_workflows`) which first resolves conditions for enabling further build phases. After ENV flags are set, it runs child workflows sequentially and each workflow has the ability to skip its steps:

Screenshot 2022-04-20 at 18 34 39

The `choose_workflows.py` script toggles `DD_RUN_(phase)_TESTS` ENVs by resolving conditions in following order:
* 1/ It checks if the `DD_RUN_(phase)_TESTS` ENV value defined in `bitrise.yml` is `1`.
* 2/ If not, it checks if the `DD_OVERRIDE_RUN_(phase)_TESTS` ENV value passed from outside is `1`.
* 3/ If not, only if the build is started for a pull request:
   * 3a/ it checks if corresponding `[ ] Run (phase) tests` checkbox is selected in PR description,
   * 3b/ last, it checks if PR modifies files which match the pattern defined for given phase.

This should give us enough flexibility and safety when running CI builds for different changes, e.g.:
* When starting a new PR we don't need to do anything - we can leverage the auto configuration from 3a/ and 3b/.
* If the PR requires running a phase which is not enabled by 3/ we can do it by selecting the checkbox in PR description.
* Any PR build can be restarted with selecting different phases than originally.
* If we need to run a phase for arbitrary commit (without PR), we can [start build manually](https://devcenter.bitrise.io/en/builds/starting-builds/starting-builds-manually.html#ot-lst-cnt) and pass override ENVs (2/).

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -1,5 +1,4 @@
----
-format_version: '8'
+format_version: 11
 default_step_lib_source: https://github.com/bitrise-io/bitrise-steplib.git
 project_type: other
 
@@ -7,14 +6,61 @@ project_type: other
 # go to Workflow Editor on Bitrise.io.
 
 workflows:
-  push_to_any_branch:
+  push_to_pull_request:
+    description: |-
+        This workflow is triggered on starting new PR or pushing new changes to existing PRs.
+        By default, it doesn't run any test phases, but this behaviour is overwritten in `choose_workflows.py` when:
+        - one or more `DD_OVERWRITE_RUN_(phase)_TESTS` ENVs are passed to the current CI job:
+            - DD_OVERRIDE_RUN_UNIT_TESTS='1' to run unit tests phase
+            - DD_OVERRIDE_RUN_INTEGRATION_TESTS='1' to run integration tests phase
+            - DD_OVERRIDE_RUN_SMOKE_TESTS='1' to run smoke tests phase
+            - DD_OVERRIDE_TOOLS_TESTS='1' to run tools tests phase
+        - a phase is selected on the checklist in the PR description,
+        - the PR changes a file which matches phase filter (e.g. changing a file in `Sources/*` will trigger unit tests phase)
+    envs:
+      - DD_RUN_UNIT_TESTS: '0'
+      - DD_RUN_INTEGRATION_TESTS: '0'
+      - DD_RUN_SMOKE_TESTS: '0'
+      - DD_RUN_TOOLS_TESTS: '0'
     after_run:
     - _make_dependencies
-    - run_linter
-    - run_unit_tests
-    - run_integration_tests
-    - check_dependency_managers
+    - run_conditioned_workflows
     - _deploy_artifacts
+
+  push_to_develop_or_master:
+    description: |-
+        This workflow is triggered for each new commit pushed to `develop` or `master` branch.
+    envs:
+      - DD_RUN_UNIT_TESTS: '1'
+      - DD_RUN_INTEGRATION_TESTS: '1'
+      - DD_RUN_SMOKE_TESTS: '0'
+      - DD_RUN_TOOLS_TESTS: '0'
+    after_run:
+    - _make_dependencies
+    - run_conditioned_workflows
+    - _deploy_artifacts
+    - _notify_failure_on_slack
+
+  run_nightly_smoke_and_tools_tests:
+    description: |-
+        This workflow is triggered every night.
+    envs:
+      - DD_RUN_UNIT_TESTS: '0'
+      - DD_RUN_INTEGRATION_TESTS: '0'
+      - DD_RUN_SMOKE_TESTS: '1'
+      - DD_RUN_TOOLS_TESTS: '1'
+    after_run:
+    - _make_dependencies
+    - run_conditioned_workflows
+    - _deploy_artifacts
+    - _notify_failure_on_slack
+
+
+  # Temporary workflow for enabling backward compatibility for pre- RUMM-2110 triggers.
+  # TODO: remove this workflow and clean-up triggers when RUMM-2110 is merged into `master` with `1.11.0`.
+  push_to_any_branch:
+    after_run:
+    - push_to_pull_request
 
   push_to_dogfooding:
     after_run:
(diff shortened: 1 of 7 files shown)


bitrise.yml
```diff
@@ -1,443 +1,108 @@
----
-format_version: '8'
+format_version: 11
```
ncreated, line 1: 💡 Testing inline comment.

ncreated (commented):
💡 Testing review comment.

bitrise.yml
```diff
@@ -1,20 +1,68 @@
...
+    - run_conditioned_workflows
     - _deploy_artifacts
+    - _notify_failure_on_slack
+
+
+  # Temporary workflow for enabling backward compatibility for pre- RUMM-2110 triggers.
+  # TODO: remove this workflow and clean-up triggers when RUMM-2110 is merged into `master` with `1.11.0`.
+  push_to_any_branch:
+    after_run:
+    - push_to_pull_request
```
ncreated, line 65: I still need to figure out how to redirect bitrise.io triggers into these new workflows. Ideally, both new and old triggers need to work together (for the sake of doing work on `master` or raising a hotfix).

tools/ci/choose_workflows.py
```diff
@@ -0,0 +1,166 @@
...
+
+        if should_run_integration_tests(context):
+            export_env('DD_RUN_INTEGRATION_TESTS', '1', dry_run=dry_run)
+
+        if should_run_smoke_tests(context):
+            export_env('DD_RUN_SMOKE_TESTS', '1', dry_run=dry_run)
+
+        if should_run_tools_tests(context):
+            export_env('DD_RUN_TOOLS_TESTS', '1', dry_run=dry_run)
+
```
ncreated, line 158: I also wonder on reporting enabled phases back to PR (if running for pull request). It can be done easily with GH CLI and sending a PR comment, e.g.:

Screenshot 2022-04-21 at 08 40 27

This would be only a sanity measure - reporting which phases are going to be ran, without reporting their status (this is done by Bitrise syncing back to GH Checks). Not sure if this would be helpful / annoying - WDYT @maxep?

tools/ci/choose_workflows.py
```diff
@@ -0,0 +1,166 @@
...
+        build_env=context.build_env.DD_OVERRIDE_RUN_UNIT_TESTS,
+        pr_keyword='[x] Run unit tests',
+        pr_path_prefixes=[
+            'Sources/',
+            'Tests/DatadogTests',
+            'Tests/DatadogCrashReportingTests',
+            'Datadog/Datadog.xcodeproj/',
+        ],
+        pr_file_extensions=[]
+    )
```
ncreated, line 32: Other ideas I had for enabling a phases automatically:
- if the branch name matches some regexp (won't require passing ENVs manually for commits without PR),
- if there is a comment with given keyword in the PR (this way we don't need a checklist in the PR, but this will require restarting job manually after adding such comment),
- if a keyword appears in PR description, without the checklist (easy to break with a typo and miss necessary phases).

tools/ci/choose_workflows.py
maxep: I don't think that's necessary, we have GH checks that already gives us that info 👍

ncreated: Makes sense 👍.

maxep (approved):
This is perfect given bitrise limitations 👌 The `tools/ci/*.py` are very clear, especially in defining workflow conditions!

I wonder if we shouldn't rely only `pr_path_prefixes` and `pr_file_extensions` to trigger workflows, we could only rely on automation and remove `pr_keyword`. This would simplify any external contribution, contributors won't have to understand our CI pipeline before opening a PR. Also, it might be safer that only us as maintainer are able to trigger workflows. We could think of achieving this using bot or hooks in the future, but for now this is doable from Bitrise console.

ncreated:
> I wonder if we shouldn't rely only `pr_path_prefixes` and `pr_file_extensions` to trigger workflows, we could only rely on automation and remove `pr_keyword`. This would simplify any external contribution, contributors won't have to understand our CI pipeline before opening a PR. Also, it might be safer that only us as maintainer are able to trigger workflows. We could think of achieving this using bot or hooks in the future, but for now this is doable from Bitrise console.

As discussed on Slack, let's keep the checklist and `pr_keyword`. It is required to sync back the CI status to GH Checks. If we find the checklist misleading or vague, we can work on it.

PR #818, opened 2022-04-19
RUMM-2024 Isolate telemetry apis

maxep:
### What and why?

Isolate RUM Telemetry apis for cross feature usage.

### How?

Creates a `Telemetry` protocol to be used across features and in core layer of the SDK. RUM then complies to the protocol through `RUMTelemetry` to send debug and error messages as RUM Telemetry events.

### Review checklist

- [ ] ~~Add CHANGELOG entry for user facing change.~~

### DIFF
--- a/Sources/Datadog/Core/Telemetry.swift
+++ b/Sources/Datadog/Core/Telemetry.swift
@@ -1,0 +1,44 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2022 Datadog, Inc.
+ */
+
+import Foundation
+
+/// The `Telemetry` protocol defines methods to collect debug information
+/// and detect execution errors of the Datadog SDK.
+internal protocol Telemetry {
+    /// Collects debug information.
+    ///
+    /// - Parameter message: The debug message.
+    func debug(_ message: String)
+
+    /// Collect execution error.
+    /// 
+    /// - Parameters:
+    ///   - message: The error message.
+    ///   - kind: The kind of error.
+    ///   - stack: The stack trace.
+    func error(_ message: String, kind: String?, stack: String?)
+}
+
+extension Telemetry {
+    /// Collect execution error.
+    ///
+    /// - Parameters:
+    ///   - message: The error message.
+    ///   - kind: The kind of error.
+    func error(_ message: String, kind: String) {
+        error(message, kind: kind, stack: nil)
+    }
+
+    /// Collect execution error.
+    ///
+    /// - Parameters:
+    ///   - message: The error message.
+    ///   - stack: The stack trace.
+    func error(_ message: String, stack: String? = nil) {
+        error(message, kind: nil, stack: stack)
+    }
+}
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
@@ -83,14 +83,6 @@ internal class RUMApplicationScope: RUMScope, RUMContextProvider {
     // MARK: - RUMScope
 
     func process(command: RUMCommand) -> Bool {
-        if let teleDebugCommand = command as? RUMTelemetryDebugCommand {
-            sendTelemetryDebugEvent(command: teleDebugCommand)
-            return true
-        } else if let teleErrorCommand = command as? RUMTelemetryErrorCommand {
-            sendTelemetryErrorEvent(command: teleErrorCommand)
-            return true
-        }
-
         if sessionScope == nil {
             startInitialSession()
         }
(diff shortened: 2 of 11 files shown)


Sources/Datadog/Core/Telemetry.swift
```diff
@@ -0,0 +1,44 @@
...
+    func error(_ message: String, kind: String) {
+        error(message, kind: kind, stack: nil)
+    }
+
+    /// Collect execution error.
+    ///
+    /// - Parameters:
+    ///   - message: The error message.
+    ///   - stack: The stack trace.
+    func error(_ message: String, stack: String? = nil) {
```
maxep, line 41: This 2 extensions apply default nullability of `kind` and `stack`.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
```diff
@@ -135,47 +135,56 @@ internal class RUMApplicationScope: RUMScope, RUMContextProvider {
...
-        let session: TelemetryDebugEvent.Session?
-        if context.sessionID == RUMUUID.nullUUID {
-            session = nil
-        } else {
-            session = .init(id: context.sessionID.toRUMDataFormat)
+        let date = dependencies.dateCorrector.currentCorrection.applying(to: command.time)
+        let actionId = sessionScope?.viewScopes.last?.context.activeUserActionID?.toRUMDataFormat
+        let viewId = sessionScope?.viewScopes.last?.context.activeViewID?.toRUMDataFormat
+        let sessionId = sessionScope.flatMap {
+            $0.context.sessionID == RUMUUID.nullUUID ? nil : $0.context.sessionID.toRUMDataFormat
```
maxep, line 142: The application scope context does not contain the right data, better to use children scopes.

ncreated: We already have [`RUMCurrentContext` provider](https://github.com/DataDog/dd-sdk-ios/blob/develop/Sources/Datadog/RUM/RUMContext/RUMCurrentContext.swift) (`RUMContextProvider`) running the logic of traversing scopes hierarchy and providing `RUMContext` to other components. It is available on `Global.rum.contextProvider` and we use it [to correlate web view events](https://github.com/DataDog/dd-sdk-ios/blob/72e5025086527a7569d3e47124db8605ed73b621/Sources/Datadog/FeaturesIntegration/WebView/WebRUMEventConsumer.swift#L12) with native RUM session and to [inject standardized RUM context attributes](https://github.com/DataDog/dd-sdk-ios/blob/72e5025086527a7569d3e47124db8605ed73b621/Sources/Datadog/FeaturesIntegration/RUMIntegrations.swift#L23) to logs and spans. The way we read `actionId`, `viewId` and `sessionId` here seems to duplicate the logic from `RUMCurrentContext` - maybe we can reuse existing implementation?

Given that `RUMContext` can be easily read from any place (through `RUMCurrentContext`), **I wonder if there is any benefit of creating telemetry events in `RUMApplicationScope` and not directly in `RUMTelemetry`?** We don't leverage telemetry commands (`RUMTelemetryDebugCommand` and `RUMTelemetryErrorCommand`) in this scope, and we even early return on receiving these in `process(command:)`. These commands are not propagated to child scopes and do not alter any RUM state.

Sending telemetry events from `RUMApplicationScope` introduces unnecessary coupling between RUM scopes (user app monitoring) and RUM telemetry concept (SDK monitoring). It requires defining two artificial `RUMCommands` which are clearly different from existing ones (they do not propagate at all). Both need to implement `RUMCommand` interface which enforces [fields that seem irrelevant to RUM telemetry](https://github.com/DataDog/dd-sdk-ios/blob/72e5025086527a7569d3e47124db8605ed73b621/Sources/Datadog/RUM/RUMMonitor/RUMCommand.swift#L13-L19).


IMHO, everything related to RUM telemetry could happen solely in the `RUMTelemetry` unit added in this PR. It could read `RUMContext` from [`RUMCurrentContext` provider](https://github.com/DataDog/dd-sdk-ios/blob/develop/Sources/Datadog/RUM/RUMContext/RUMCurrentContext.swift) and receive necessary dependencies (`eventOutput`, `dateCorrector`, ...) through `dependencies` container. It can then create telemetry events directly, without using any commands.

This way we can remove coupling between RUM scopes and RUM telemetry. The resultant implementation could be simpler (we don't need `RUMTelemetryDebugCommand` and `RUMTelemetryErrorCommand`) and would promote reusability (`RUMCurrentContext`). WDYT @maxep ?

ncreated (commented):
I really like the `RUMTelemetry` thing 👍. I left a feedback on isolating RUM telemetry even more from RUM scopes. Let me know WDYT 🙂.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
maxep: This was my initial direction :) See the [fist commit](https://github.com/DataDog/dd-sdk-ios/pull/818/commits/02a1d9d5b06f39b56cbff7bb99feb3fe31014e0c)

I totally prefer this approach as well, but I see 2 problems:
1. `FeatureStorage` need to be created first to get `eventOutput` dependency, but `FeatureStorage` will need a telemetry arg for instantiation.
2. `monitor.contextProvider.context` is doing `sync` on the monitor queue leading to potential deadlock..

So coupling telemetry to application scope seems this only (soft) option. But I think we get still get rid of the `RUMTelemetry*Command` by using `monitor?.applicationScope` directly in `RUMTelemetry`. How does it sounds?

ncreated: Oh, got it 👍. Yes, calling `applicationScope` could be an option, however it will be complicated if we remove telemetry commands (they guarantee thread safe execution on single `queue` created in `RUMMonitor`). And indeed, there is also a 🐥🥚 problem of injecting `Telemetry` into core components and we need lazy binding.

I wonder if extending [`RUMCurrentContext`](https://github.com/DataDog/dd-sdk-ios/blob/develop/Sources/Datadog/RUM/RUMContext/RUMCurrentContext.swift) with async context API could be an option?
```swift
func withAsyncContext(_ callback: (context: RUMContext) -> Void)) {
   queue.async { 
      callback(activeViewContext ?? sessionContext ?? applicationContext)
   }
}
```
This way `RUMTelemetry` could safely (no deadlock) use context and write events to the output:
```swift
// RUMTelemetry

private var rumMonitor: RUMMonitor? { Global.rum as? RUMMonitor }
private var rumFeature: RUMFeature? { RUMFeature.instance }

func debug(_ message: String) {
   guard let rumFeature = rumFeature else { return }

   rumMonitor?.contextProvider.withAsyncContext { rumContext in
      let telemetryEvent = ... // create with `rumContext`, `rumFeature.dateCorrector` etc.
      rumFeature.storage.writer.write(telemetryEvent)
   }
}
```

This way we could fully decouple `RUMMonitor` from `RUMTelemetry` and leave RUM scopes unchanged. It will be also conceptually closer to what we plan for V2 (async context). The only problem is shifting the moment of probing rum context (to next runloop), but IMO this is less important for telemetry. WDYT?

maxep: I like that!

ncreated (approved):
🎯 💯

PR #821, opened 2022-04-25
RUMM-2027 Add RUM Telemetry sampling

maxep:
### What and why?

Add RUM telemetry event sampling, discard duplicates, and report up to 100 events per sessions.

### How?

`RUMTelemetry` class now takes a sampler parameter and implement a log validation before recording the telemetry event.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/Datadog.swift
+++ b/Sources/Datadog/Datadog.swift
@@ -228,7 +228,8 @@ public class Datadog {
                 applicationID: rumConfiguration.applicationID,
                 source: rumConfiguration.common.source,
                 dateProvider: dateProvider,
-                dateCorrector: dateCorrector
+                dateCorrector: dateCorrector,
+                sampler: rumConfiguration.telemetrySampler
             )
 
             rum = RUMFeature(
--- a/Sources/Datadog/RUM/RUMTelemetry.swift
+++ b/Sources/Datadog/RUM/RUMTelemetry.swift
@@ -28,7 +28,7 @@ internal final class RUMTelemetry: Telemetry {
     private var currentSessionID: RUMUUID = .nullUUID
 
     /// Keeps track of event's ids recorded during a user session.
-    private var eventIDs: [String] = []
+    private var eventIDs: Set<String> = []
 
     /// Creates a RUM Telemetry instance.
     ///
@@ -142,7 +142,7 @@ internal final class RUMTelemetry: Telemetry {
 
             // record up de `MaxEventsPerSessions`, discard duplicates
             if self.eventIDs.count < RUMTelemetry.MaxEventsPerSessions, !self.eventIDs.contains(id) {
-                self.eventIDs.append(id)
+                self.eventIDs.insert(id)
                 operation(context, writer)
             }
         }
--- a/Sources/Datadog/RUM/RUMTelemetry.swift
+++ b/Sources/Datadog/RUM/RUMTelemetry.swift
@@ -8,14 +8,27 @@ import Foundation
 
 /// Sends Telemetry events to RUM.
 ///
-/// `RUMTelemetry` complies to `Telemetry` protocol allowing
-/// sending telemetry events accross features.
+/// `RUMTelemetry` complies to `Telemetry` protocol allowing sending telemetry
+/// events accross features.
+///
+/// Events are reported up to 100 per sessions with a sampling mechanism that is
+/// configured at initialisation. Duplicates are discared.
 internal final class RUMTelemetry: Telemetry {
+    /// Maximium number of telemetry events allowed per user sessions.
+    static let MaxEventsPerSessions: Int = 100
+
     let sdkVersion: String
     let applicationID: String
     let source: String
     let dateProvider: DateProvider
     let dateCorrector: DateCorrectorType
+    let sampler: Sampler
+
+    /// Keeps track of current session
+    private var currentSessionID: RUMUUID = .nullUUID
+
+    /// Keeps track of event's ids recorded during a user session.
+    private var eventIDs: [String] = []
 
     /// Creates a RUM Telemetry instance.
     ///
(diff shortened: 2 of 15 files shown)


Sources/Datadog/RUM/RUMMonitor/Scopes/RUMScopeDependencies.swift
```diff
@@ -73,7 +71,51 @@ internal extension RUMScopeDependencies {
...
+    ///   - id: The Session UUID
+    ///   - isDiscarded: `true` if the session is discarded by sampling.
+    func onSessionStart(session id: String, isDiscarded: Bool)
+}
+
+/// `RUMSessionListeners` support registration of session start listeners.
+internal final class RUMSessionListeners {
+    typealias Listener = (String, Bool) -> Void
+
+    private var listeners: [Listener] = []
```
ncreated, line 94: We access `listeners` array from multiple threads, including:
* `sessionsListeners.register()` in main `Datadog` class - called on the user thread,
* `sessionsListeners.notify()` from `RUMApplicationScope` - called on RUM's thread,

so it should be synchronised 💥. For what I see, now all `register()` calls happen from `Datadog.initialize()` and running into `notify()` flow requires the SDK to be initialized, so this can't crash for now. However, this assumption can be easily broken if calling one more `register()` in after SDK init.

Alternatively, we could make `listeners` immutable by requiring it in `RUMSessionListeners.init()`, but this option doesn't look easier.

Sources/Datadog/RUM/RUMTelemetry.swift
```diff
@@ -8,13 +8,22 @@ import Foundation
...
+    /// Maximium number of telemetry events allowed per user sessions.
+    static let MaxEventsPerSessions: Int = 100
+
     let sdkVersion: String
     let applicationID: String
     let dateProvider: DateProvider
     let dateCorrector: DateCorrectorType
+    let sampler: Sampler
+
+    private var events: [String] = []
```
ncreated, line 26: We access (rw) this array on each event and we collect events from different threads, so it needs to be synchronised to avoid crash 💥.

Also, I think we should add tests for ensuring thread safety. We have a convention for thread safety testing [using the `callConcurrently` helper](https://github.com/DataDog/dd-sdk-ios/blob/3ae90a54513dd05fc053ceb7ad84859489cae604/Tests/DatadogTests/Datadog/Core/Utils/ValuePublisherTests.swift#L133-L141), WDYT?

Sources/Datadog/RUM/RUMTelemetry.swift
```diff
@@ -81,18 +97,21 @@ internal final class RUMTelemetry: Telemetry {
...
         guard
+            accept(id: id),
             let monitor = Global.rum as? RUMMonitor,
             let writer = RUMFeature.instance?.storage.writer
         else {
             return
         }
 
         let date = dateCorrector.currentCorrection.applying(to: dateProvider.currentDate())
+        events.append(id)
```
ncreated, line 114: If we store event IDs, wouldn't it be better to call it `eventIDs` instead? Alternatively, the `event's` comment could explain the actual content.

ncreated (changes requested):
I like the idea 👍. I left some feedback on thread safety.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMScopeDependencies.swift
maxep: Oh indeed! I would like to avoid using a queue for this. I will see if I can achieve immutability of `RUMSessionListeners`, otherwise we could so `objc_sync_enter/objc_sync_exit`.

Sources/Datadog/RUM/RUMTelemetry.swift
maxep: Yes 👍 better to do r/w operation in the `monitor.contextProvider.async` scope.
Very cool tool this `callConcurrently` check!!

Sources/Datadog/Datadog.swift
```diff
@@ -223,17 +223,28 @@ public class Datadog {
...
 
+            telemetry = rumTelemetry
+
+            var sessionListeners: [RUMSessionListeners.Listener] = []
+            sessionListeners.append(rumTelemetry.onSessionStart)
+
+            if let sessionListener = rumConfiguration.sessionsListener {
+                sessionListeners.append(sessionListener)
+            }
+
```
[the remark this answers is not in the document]
ncreated: Indeed, it looks a bit clumsy. Although I like this reactive approach as it reads very well in downstream logic (_"on session change, clear telemetry"_), it appears clumsy here when it comes to configuration. It's easy to imagine that requiring a listener somewhere else (e.g. lazily) will force us back into `register()` API.

> I'm open to suggestions!

How about using the `RUMContext` which we already read in `RUMTelemetry`? It brings the current `sessionID`, so we can compare it against previous `sessionID` to determine "session change":
```swift
monitor.contextProvider.async { [weak self] context
   if context.sessionID != self.previousSessionID {
      self.previousSessionID = context.sessionID
      reset()
   }

   // ...
}
```

Downside is, it would duplicate the logic of recognising session change. Not a big deal IMO, as `sessionID` is a core element in RUM domain - it is hard to imagine that session change would ever be indicated in another way. Also, if we ever want to reuse that logic, it can be done with a higher-level API of comparing two `contexts`.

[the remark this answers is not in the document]
maxep: 💡 Thank you! Easier, self-contained, I like that!

Sources/Datadog/RUM/RUMTelemetry.swift
```diff
@@ -8,14 +8,27 @@ import Foundation
...
     let source: String
     let dateProvider: DateProvider
     let dateCorrector: DateCorrectorType
+    let sampler: Sampler
+
+    /// Keeps track of current session
+    private var currentSessionID: RUMUUID = .nullUUID
+
+    /// Keeps track of event's ids recorded during a user session.
+    private var eventIDs: [String] = []
```
ncreated, line 31: `Set<String>` would be better - both in terms of performance and to express the purpose of `eventIDs`.

Tests/DatadogTests/Datadog/RUM/RUMTelemetryTests.swift
```diff
@@ -107,9 +120,9 @@ class RUMTelemetryTests: XCTestCase {
         class TelemetryTest: Telemetry {
             var record: (message: String, kind: String?, stack: String?)?
 
-            func debug(_ message: String) { }
+            func debug(id: String, message: String) { }
```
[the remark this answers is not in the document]
maxep: TIL :)

PR #830, opened 2022-04-27
RUMM-2153 Prevent Kronos logic from querying private IPs

ncreated:
### What and why?

🐞 This PR aims at fixing https://github.com/DataDog/dd-sdk-ios/issues/647. As explained in https://github.com/DataDog/dd-sdk-ios/issues/647#issuecomment-1108489161 we found an evidence (in our internal telemetry) that in some circumstances Kronos (NTP) logic can try to query private IP and hence bring up the _"Local Network Permission"_ alert. 

This happens very rarely, and except a single hit in our logs we couldn't manage to reproduce it.

### How?

Our telemetry shows a clear attempt on querying `192.168.1.250`. To avoid this, in this PR we apply additional filtering in `KronosDNSResolver` to ensure that private IPs are never used for NTP sync.

165082340-1c4f5a8a-219e-4ea7-81ae-a07fb5d5f976

If all resolved IPs are filtered out, Kronos will fail silently and fallback to previous known time correction (stored in user defaults). In case of fresh start, the SDK falls back to using device time:
```
[DATADOG SDK] 🐶 → 11:45:15.385 [WARN] NTP time synchronization failed.
Device time will be used for signing events (current device time is 2022-04-27 09:45:15 +0000).
```

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/Kronos/KronosInternetAddress.swift
+++ b/Sources/Datadog/Kronos/KronosInternetAddress.swift
@@ -43,6 +43,52 @@ internal enum KronosInternetAddress: Hashable {
         }
     }
 
+    /// If the address is reserved for private internets (local / private IP).
+    var isPrivate: Bool {
+        guard let host = host else {
+            return false
+        }
+
+        switch self {
+        case .ipv6:
+            // Ref.:  https://datatracker.ietf.org/doc/html/rfc4193#section-3
+            // +--------+-+------------+-----------+----------------------------+
+            // | 7 bits |1|  40 bits   |  16 bits  |          64 bits           |
+            // +--------+-+------------+-----------+----------------------------+
+            // | Prefix |L| Global ID  | Subnet ID |        Interface ID        |
+            // +--------+-+------------+-----------+----------------------------+
+            //
+            // Local IP is expected to have FC00::/7 prefix (7 bits) and L byte set to 1,
+            // which effectively means `fd` prefix for local IPs.
+            let localPrefix = "fd"
+
+            // Ref.: https://datatracker.ietf.org/doc/html/rfc4291#section-2.4
+            let multicastPrefix = "ff"
+
+            let hostLowercased = host.lowercased()
+            return hostLowercased.starts(with: localPrefix)
+                || hostLowercased.starts(with: multicastPrefix)
+        case .ipv4:
+            // Ref.: https://datatracker.ietf.org/doc/html/rfc1918#section-3
+            // Local IPs have predefined ranges:
+            // - class A: 10.0.0.0 — 10.255.255.255
+            // - class B: 172.16.0.0 — 172.31.255.255
+            // - class C: 192.168.0.0 — 192.168.255.255
+            let classABCregex = #"^((10\.)|(172\.1[6-9]\.)|(172\.2[0-9]\.)|(172\.3[0-1]\.)|(192\.168\.))"#
+
+            // Ref.: https://datatracker.ietf.org/doc/html/rfc5771#section-3
+            // Multicast address (range 224.0.0.0 - 239.255.255.255) are considered to be local network
+            // addresses too (https://developer.apple.com/forums/thread/663848)
+            //
+            let multicastRegex = #"^((22[4-9]\.)|(23[0-9]\.))"#
+            let broadcastIP = "255.255.255.255"
+
+            return host.range(of: classABCregex, options: .regularExpression) != nil
+                || host.range(of: multicastRegex, options: .regularExpression) != nil
+                || host == broadcastIP
+        }
+    }
+
     func hash(into hasher: inout Hasher) {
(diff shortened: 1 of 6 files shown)


Sources/Datadog/Kronos/KronosInternetAddress.swift
```diff
@@ -43,6 +43,37 @@ internal enum KronosInternetAddress: Hashable {
...
+            return host.starts(with: "fd")
+        case .ipv4:
+            // Ref.: https://datatracker.ietf.org/doc/html/rfc1918#section-3
+            //
+            // Local IPs have predefined ranges:
+            // - class A: 10.0.0.0 — 10.255.255.255
+            // - class B: 172.16.0.0 — 172.31.255.255
+            // - class C: 192.168.0.0 — 192.168.255.255
+            let regex = #"^((10\.)|(172\.1[6-9]\.)|(172\.2[0-9]\.)|(172\.3[0-1]\.)|(192\.168\.))"#
+            return host.range(of: regex, options: .regularExpression) != nil
```
ncreated, line 73: As an alternative, we could use bit masks instead of regex. That would require reading IP bytes from [`sockaddr_in`](https://developer.apple.com/documentation/kernel/sockaddr_in). It could be easy for IPv4, but seems trickier for IPv6 ([`sockaddr_in6`](https://developer.apple.com/documentation/kernel/sockaddr_in6)). To avoid implementing raw memory binds, I just used existing `host` value and regex.

Effectively, the SDK resolves 3-5 IPs on startup, so performance is not a problem.

Tests/DatadogTests/Datadog/Kronos/KronosInternetAddressTests.swift
```diff
@@ -0,0 +1,111 @@
...
+        precondition(bytes.count == 16, "Expected 16 bytes")
+        let groups: [String] = (0..<8).map { idx in
+            let hexA = String(bytes[idx * 2], radix: 16, uppercase: false)
+            let hexB = String(bytes[idx * 2 + 1], radix: 16, uppercase: false)
+            return hexA + hexB
+        }
+        let ipv6String = groups.joined(separator: ":") // e.g. 'ab:ab:ab:ab:ab:ab:ab:ab'
+        let address: KronosInternetAddress? = .mockWith(ipv6String: ipv6String)
+        return try XCTUnwrap(address, "\(ipv6String) is not a valid IPv6 string")
+    }
```
ncreated, line 84: The way we mock `KronosInternetAddress` values for IPv4 and IPv6 is quite tricky, but couldn't find a way to make it simpler. Alternatively, we could implement filtering as stand-alone function (`isPrivate(host:)`), but I found computed property (`ip.isPrivate`) more idiomatic.

Sources/Datadog/Kronos/KronosInternetAddress.swift
buranmert: i think we should cover [all the cases that Eskimo mentions here](https://developer.apple.com/forums/thread/663848) 
(`224.0.0.0/4` and `255.255.255.255` are currently uncovered)

buranmert (changes requested):
i left a somewhat minor comment but given the importance of this issue i'm requsting a change

Tests/DatadogTests/Datadog/Kronos/KronosInternetAddressTests.swift
```diff
@@ -0,0 +1,111 @@
...
+
+        privateIPs.forEach { ip in
+            XCTAssertTrue(ip.isPrivate, "\(ip.host ?? "nil") should be private IP")
+        }
+        publicIPs.forEach { ip in
+            XCTAssertFalse(ip.isPrivate, "\(ip.host ?? "nil") should not be private IP")
+        }
+    }
+
+    func testIfIPv6AddressIsPrivarte() throws {
```
maxep, line 39:
```suggestion
    func testIfIPv6AddressIsPrivate() throws {
```

Tests/DatadogTests/Datadog/Kronos/KronosInternetAddressTests.swift
maxep: I'm not sure of the fuzzy tests value here 🤔 especially since the generated IPs is valid/expected we could simply test against edges and some pre-determined values in between, no?

maxep (approved):

buranmert (approved):

buranmert:
we can mention our fix in Kronos repo (https://github.com/MobileNativeFoundation/Kronos/issues/94) so that other users can apply a similar fix

pepas-everly:
We have recently been bitten by this.  Is there an easy way to disable NTP altogether?  The device clocks are already more than accurate enough for our needs.

PR #837, opened 2022-05-02
Allow manually tracked resources to detect first party hosts

fuzzybinary:
### What and why?

First party hosts set by `trackUrlSession` only works for automatically instrumented URLSessions. This allows manual calls to startResourceLoading and stopResourceLoading to detect first party host calls as well.

### How?

Not sure if this is the approach we want to take, but it made sense to me. I'm pulling the configuration onto the RUM feature configuration, then using that in the dependency configuration for the application scope to create the same Filter object that the URLSessionDelegate uses.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


maxep (commented):
I like the idea of adding url filter to `RUMScopeDependencies`, with that I don't think we need the `isFirstPartyRequest ` property in `RUMStartResourceCommand` anymore. We can simply use the filter while creating the rum event, WDYT?

fuzzybinary:
Yeah that would simplify things. I'll give it a shot.

fuzzybinary:
@maxep Updated, let me know if this is what you meant.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -234,7 +234,7 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
             dateCorrection: dateCorrection,
             url: command.url,
             httpMethod: command.httpMethod,
-            isFirstPartyResource: command.isFirstPartyRequest,
+            isFirstPartyResource: dependencies.firstPartyURLsFilter.isFirstParty(string: command.url),
```
maxep, line 237: Do we still need `isFirstPartyResource` here? 

The `RUMResourceScope` also keeps `dependencies` property, so we could use `dependencies.firstPartyURLsFilter` [here](https://github.com/DataDog/dd-sdk-ios/blob/develop/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift#L279) and [here](https://github.com/DataDog/dd-sdk-ios/blob/develop/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift#L291)

maxep: Or directly in [here](https://github.com/DataDog/dd-sdk-ios/blob/develop/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift#L72)

maxep:
Yes, exactly! I left another /nit

Also, could you rebase on `develop`? I've merge the release branch since :)

maxep (approved):

fuzzybinary:
@maxep Moved it up one more level.

Tests/DatadogTests/Datadog/Mocks/RUMFeatureMocks.swift
```diff
@@ -736,6 +738,7 @@ extension RUMScopeDependencies {
             applicationVersion: applicationVersion ?? self.applicationVersion,
             sdkVersion: sdkVersion ?? self.sdkVersion,
             source: source ?? self.source,
+            firstPartyURLsFilter: firstPartyUrls != nil ? FirstPartyURLsFilter(hosts: firstPartyUrls!) : self.firstPartyURLsFilter,
```
maxep, line 741: /nit 
```suggestion
            firstPartyURLsFilter: firstPartyUrls.map { .init(hosts: $0) } ?? self.firstPartyURLsFilter,
```

maxep (approved):
Thanks 👌

PR #841, opened 2022-05-04
RUMM-2134 Write events to files in TLV format

maxep:
### What and why?

Write event to files in TLV format. This change prepares the introduction of event metadata to batches.

### How?

All events will be serialised in TLV format using the following byte alignment:
```
+-  2 bytes -+-   4 bytes   -+- n bytes -|
| block type | data size (n) |    data   |
+------------+---------------+-----------+
```
This `block type` is a 2 bytes value describing how to decode the data. In this PR, only code `0x00` identifying an event is used.

The `data size` is written in 4 bytes so we are more flexible in term of size limit.

### Note

By using TLV, the format differs from the RFC so we are more resilient to header (metadata) decoding failure.
Batches will be written in `v2` folders, this PR does not remove any `v1` folder, it will be dealt in another PR.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/Core/Feature.swift
+++ b/Sources/Datadog/Core/Feature.swift
@@ -81,14 +81,12 @@ internal struct FeatureStorage {
         )
 
         let unauthorizedFileWriter = FileWriter(
-            dataFormat: dataFormat,
             orchestrator: unauthorizedFilesOrchestrator,
             encryption: commonDependencies.encryption,
             telemetry: telemetry
         )
 
         let authorizedFileWriter = FileWriter(
-            dataFormat: dataFormat,
             orchestrator: authorizedFilesOrchestrator,
             encryption: commonDependencies.encryption,
             telemetry: telemetry
--- a/Sources/Datadog/Core/Persistence/DataBlock.swift
+++ b/Sources/Datadog/Core/Persistence/DataBlock.swift
@@ -1,0 +1,193 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+/// Block size binary type
+internal typealias BlockSize = UInt32
+
+/// Block type supported in data stream
+internal enum BlockType: UInt16 {
+    case event = 0x00
+    case headerV1 = 0x01
+}
+
+/// Reported errors while manipulateing data blocks.
+internal enum DataBlockError: Error {
+    case readOperationFailed(streamError: Error?)
+    case invalidByteSequence
+}
+
+/// A data block in defined by its type and a byte sequence.
+///
+/// A block can be serialized in data stream by following TLV format.
+internal struct DataBlock {
+    /// Type describing the data block.
+    let type: BlockType
+
+    /// The data.
+    var data: Data
+
+    /// Returns a Data block in Type-Lenght-Value format.
+    ///
+    /// A block follow TLV with bytes aligned such as:
+    ///
+    ///     +-  2 bytes -+-   4 bytes   -+- n bytes -|
+    ///     | block type | data size (n) |    data   |
+    ///     +------------+---------------+-----------+
+    ///
+    /// - Returns: a data block in TLV.
+    func serialize() -> Data {
+        var buffer = Data()
+        // T
+        withUnsafeBytes(of: type.rawValue) { buffer.append(contentsOf: $0) }
+        // L
+        withUnsafeBytes(of: BlockSize(data.count)) { buffer.append(contentsOf: $0) }
+        // V
+        buffer += data
+        return buffer
+    }
+}
+
+internal extension Data {
+    /// Returns a Data block in Type-Lenght-Value format.
+    ///
+    /// A block follow TLV with bytes aligned such as:
+    ///
+    ///     +-  2 bytes -+-   4 bytes    -+- n bytes  -|
+    ///     | block type | block size (n) | block data |
+    ///     +------------+----------------+------------+
+    ///
+    /// - Parameters:
+    ///   - type: The data type
+    ///   - data: The data
+    /// - Returns: a byte sequence in TLV format.
+    static func block(_ type: BlockType, data: Data) -> Data {
+        return DataBlock(type: type, data: data).serialize()
+    }
+}
+
+/// A block reader can read TLV formatted blocks from a data input.
+///
+///
+internal final class DataBlockReader {
+    /// The input data stream.
+    private let stream: InputStream
+
+    /// Reads block from data input.
+    ///
+    /// - Parameter data: The data input
+    init(data: Data) {
+        stream = InputStream(data: data)
+        stream.open()
+    }
+
+    /// Reads block from url input.
+    ///
+    /// At initilization, the reader will open a stream targeting the input
+    /// url. The stream will be closed when the reader instance is deallocated.
+    ///
+    /// - Parameter url: Data url.
+    init?(url: URL) {
+        guard let stream = InputStream(url: url) else {
+            return nil
+        }
+
+        stream.open()
+        self.stream = stream
+    }
+
+    deinit {
+        stream.close()
+    }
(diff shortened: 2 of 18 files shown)


Sources/Datadog/Core/Persistence/DataBlock.swift
```diff
@@ -0,0 +1,193 @@
...
+
+import Foundation
+
+/// Block size binary type
+internal typealias BlockSize = UInt32
+
+/// Block type supported in data stream
+internal enum BlockType: UInt16 {
+    case event = 0x00
+    case headerV1 = 0x01
```
maxep, line 15: This is an example, but should actually be removed

Sources/Datadog/Core/Persistence/DataBlock.swift
```diff
@@ -0,0 +1,193 @@
...
+        return DataBlock(type: type, data: data).serialize()
+    }
+}
+
+/// A block reader can read TLV formatted blocks from a data input.
+///
+///
+internal final class DataBlockReader {
+    /// The input data stream.
+    private let stream: InputStream
```
ncreated, line 78: I'm not against using streams instead of file handlers, but I'm missing a good reason for doing this change now, as part of the data format enhancement.

Since SDK `1.0.0` we went through a bunch of unobvious problems and crashes in `FileHandle` APIs. Both our [file writing](https://github.com/DataDog/dd-sdk-ios/blob/63729c909cad2438695adc86f9f2c61103ffe102/Sources/Datadog/Core/Persistence/Files/File.swift#L50-L78) and [file reading](https://github.com/DataDog/dd-sdk-ios/blob/63729c909cad2438695adc86f9f2c61103ffe102/Sources/Datadog/Core/Persistence/Files/File.swift#L94-L121) code is full of patches for different OS and Swift versions, including [Obj-c errors handling in Swift](https://github.com/DataDog/dd-sdk-ios/blob/63729c909cad2438695adc86f9f2c61103ffe102/Sources/Datadog/Core/Persistence/Files/File.swift#L126) added due to ([`FB7743153`](https://openradar.appspot.com/radar?id=4998297672482816)).

I can't say that `InputStream` APIs are the same buggy as `FileHandle` interface, but thinking of all these hard lessons with went through with `FileHandle`, I'm sceptical on changing this core element right now. How about adding it to the backlog and tackling another time? This way we will avoid shipping both, crucial refactoring and important Foundation APIs switch in a single SDK version.

Sources/Datadog/Core/Feature.swift
```diff
@@ -81,14 +81,12 @@ internal struct FeatureStorage {
         )
 
         let unauthorizedFileWriter = FileWriter(
-            dataFormat: dataFormat,
             orchestrator: unauthorizedFilesOrchestrator,
             encryption: commonDependencies.encryption,
             telemetry: telemetry
         )
```
ncreated, line 87: Great to see this change 👍. It means that writer stack is less dependant on the feature specifics and its ultimate decoupling is a key goal for V2.

Sources/Datadog/Core/Persistence/Reading/FileReader.swift
```diff
@@ -47,52 +47,66 @@ internal final class FileReader: Reader {
...
+        // get event blocks only
+        return try reader.all()
             .compactMap {
-                if let data = Data(base64Encoded: $0) {
-                    return data
+                switch $0.type {
+                case .event:
+                    return $0.data
+                default:
+                    return nil
```
ncreated, line 73: Better to have explicit `case` list - it's a core piece of our logic, so let's leverage type safety.

Sources/Datadog/Core/Persistence/DataBlock.swift
```diff
@@ -0,0 +1,193 @@
...
+    ///     +------------+----------------+------------+
+    ///
+    /// - Parameters:
+    ///   - type: The data type
+    ///   - data: The data
+    /// - Returns: a byte sequence in TLV format.
+    static func block(_ type: BlockType, data: Data) -> Data {
+        return DataBlock(type: type, data: data).serialize()
+    }
+}
```
ncreated, line 71: Is this extension needed 💭🤔? I can only see its usage in tests, not in the production code. Also, as we're using Foundation type (`Data`) to only attach a domain-specific (core) function to, so it doesn't seem to bring much improvement over using `DataBlock(type:data:)` directly, WDYT?

ncreated (commented):
I really like the idea of using TLV, it's a nice enhancement to the originally proposed format 👍. It should be also simpler to reason about - we will start with only one `BlockType` and add more when iterating on V2 ✨.

My biggest worry is on introducing `InputStream` as part of this refactoring. I left deeper explanation in the comment. LMKWDYT 🙌.

Sources/Datadog/Core/Persistence/DataBlock.swift
maxep: Yes, I saw the patches and I was worried of doing any changes in the `File` implementation. The mechanism here is that `FileHandle` reads the entire batch and `InputStream` navigate through the data in memory.
I agree that it's risky tho, it will be best to tackle that separately 👍
Note that it will require changes in `File` to be able to navigate the data

Sources/Datadog/Core/Persistence/Reading/FileReader.swift
maxep: I added another case `BlockType` as an example, it will contain only `event` and the `default` here will be gone :)

Sources/Datadog/Core/Persistence/DataBlock.swift
maxep: Yea, not really needed, it was more for convenience for future change. I will remove it!
Could be cool to leverage a pattern like [AlamofireExtended](https://github.com/Alamofire/Alamofire/blob/master/Source/AlamofireExtended.swift) in the future, so we can hide any extension behind a `dd` property!

Sources/Datadog/Core/Persistence/DataBlock.swift
ncreated: Oh, I haven't realised that we still use `FileHandle` to manage file and only use `InputStream` to read data. That said, it's not a revamp, it's an incremental change, so maybe worth doing it. 🤔💭

ncreated: The `DataBlockReader(url:)` initializer is misleading - seem that we don't use it and it can be removed. Then files will be clearly managed through `FH`.

Sources/Datadog/Core/Persistence/DataBlock.swift
```diff
@@ -0,0 +1,147 @@
...
+    ///     | block type | data size (n) |    data   |
+    ///     +------------+---------------+-----------+
+    ///
+    /// - Returns: a data block in TLV.
+    func serialize() -> Data {
+        var buffer = Data()
+        // T
+        withUnsafeBytes(of: type.rawValue) { buffer.append(contentsOf: $0) }
+        // L
+        withUnsafeBytes(of: BlockSize(data.count)) { buffer.append(contentsOf: $0) }
```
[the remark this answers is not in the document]
maxep: Nop not safe! I should be using `init(exactly:)`

Tests/DatadogTests/Datadog/Core/Persistence/DataBlockTests.swift
```diff
@@ -0,0 +1,42 @@
...
+        .reduce(Data(), +)
+
+        let reader = DataBlockReader(data: data)
+        let blocks = try reader.all()
+
+        XCTAssertEqual(blocks.count, 100)
+        XCTAssertEqual(blocks.first?.data.count, 0)
+        XCTAssertEqual(blocks.last?.data.count, 99)
+    }
+}
```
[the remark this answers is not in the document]
maxep: As discussed, we won't be able to test maximum data block sizes (4GB) without redesigning TLV implementation and using generics to mock smaller block size footprint in bytes. Tho, I've added test for 0 bytes data and for large (1MB) data block.

ncreated (approved):
Looks good 👌🚀

PR #848, opened 2022-05-11
RUMM-2132 Remove `Datadog.instance` singleton

maxep:
### What and why?

Remove `Datadog.instance` singleton.

### How?

This is a step toward architecture v2 by introducing building blocks of feature orchestration. Here is a diagram presenting those blocks:

image

`class Datadog` from v1 creates a `DatadogCore` that complies with `DatadogCoreProtocol` to use as the `default` core implementation. `DatadogCore` does not include any logic for registration yet, this will come as we remove feature singletons.

__Note:__ naming is still being discussed and is subject to changes.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/Core/DatadogCoreProtocol.swift
+++ b/Sources/Datadog/Core/DatadogCoreProtocol.swift
@@ -6,25 +6,24 @@
 
 import Foundation
 
-internal var DefaultDatadogCore: DatadogCoreProtocol = NOOPDatadogCore()
+internal var defaultDatadogCore: DatadogCoreProtocol = NOOPDatadogCore()
 
-/// A Datadog Core holds a set of features and is responsible of
-/// managing their storage and upload mechanism. It also provides a thread-safe
-/// scope for writing events.
+/// A Datadog Core holds a set of features and is responsible of managing their storage
+/// and upload mechanism. It also provides a thread-safe scope for writing events.
 public protocol DatadogCoreProtocol {
     /// Registers a feature by its name and configuration.
     ///
     /// - Parameters:
-    ///   - named: The feature name.
+    ///   - featureName: The feature name.
     ///   - storage: The feature's storage configuration.
     ///   - upload: The feature's upload configuration.
-    func registerFeature(named: String, storage: FeatureStorageConfiguration, upload: FeatureUploadConfiguration)
+    func registerFeature(named featureName: String, storage: FeatureStorageConfiguration, upload: FeatureUploadConfiguration)
 
     /// Returns a Feature scope by its name.
     ///
-    /// - Parameter named: The feature's name.
+    /// - Parameter featureName: The feature's name.
     /// - Returns: The feature if it was previously registered, `nil` otherwise.
-    func scope(forFeature named: String) -> FeatureScope?
+    func scope(forFeature featureName: String) -> FeatureScope?
 }
 
 /// Provide feature specific storage configuration.
@@ -45,10 +44,10 @@ public protocol FeatureScope {
 /// No-op implementation of `DatadogFeatureRegistry`.
 internal struct NOOPDatadogCore: DatadogCoreProtocol {
     /// no-op
-    func registerFeature(named: String, storage: FeatureStorageConfiguration, upload: FeatureUploadConfiguration) {}
+    func registerFeature(named featureName: String, storage: FeatureStorageConfiguration, upload: FeatureUploadConfiguration) {}
 
     /// no-op
-    func scope(forFeature named: String) -> FeatureScope? {
+    func scope(forFeature featureName: String) -> FeatureScope? {
         return nil
     }
 }
--- a/Sources/Datadog/Core/DatadogCoreProtocol.swift
+++ b/Sources/Datadog/Core/DatadogCoreProtocol.swift
@@ -8,9 +8,9 @@ import Foundation
 
 internal var DefaultDatadogCore: DatadogCoreProtocol = NOOPDatadogCore()
 
-/// A Datadog Feature registry hold a set of features and is responsible of
-/// managing their storage and upload mechanism. It also provide thread-safe
-/// scope for features.
+/// A Datadog Core holds a set of features and is responsible of
+/// managing their storage and upload mechanism. It also provides a thread-safe
+/// scope for writing events.
 public protocol DatadogCoreProtocol {
     /// Registers a feature by its name and configuration.
     ///
--- a/Sources/Datadog/Core/DatadogCoreProtocol.swift
+++ b/Sources/Datadog/Core/DatadogCoreProtocol.swift
@@ -6,20 +6,12 @@
 
 import Foundation
 
-/// Defines a shared feature registry for global usage.
-internal enum DatadogRegistry {
-    /// The default feature registry for global usage.
-    ///
-    /// Instances complying to `DatadogFeatureRegistry` can be used
-    /// as a global feature registry. By default, `.default` returns a no-op
-    /// registery.
-    internal static var `default`: DatadogFeatureRegistry = NOOPDatadogRegistry()
-}
+internal var DefaultDatadogCore: DatadogCoreProtocol = NOOPDatadogCore()
 
 /// A Datadog Feature registry hold a set of features and is responsible of
 /// managing their storage and upload mechanism. It also provide thread-safe
 /// scope for features.
-public protocol DatadogFeatureRegistry {
+public protocol DatadogCoreProtocol {
     /// Registers a feature by its name and configuration.
     ///
     /// - Parameters:
@@ -51,7 +43,7 @@ public protocol FeatureScope {
 }
 
 /// No-op implementation of `DatadogFeatureRegistry`.
-internal struct NOOPDatadogRegistry: DatadogFeatureRegistry {
+internal struct NOOPDatadogCore: DatadogCoreProtocol {
     /// no-op
     func registerFeature(named: String, storage: FeatureStorageConfiguration, upload: FeatureUploadConfiguration) {}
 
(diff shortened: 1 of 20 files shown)


Sources/Datadog/Core/DatadogFeatureRegistry.swift
```diff
@@ -0,0 +1,62 @@
...
+
+/// Defines a shared feature registry for global usage.
+internal enum DatadogRegistry {
+    /// The default feature registry for global usage.
+    ///
+    /// Instances complying to `DatadogFeatureRegistry` can be used
+    /// as a global feature registry. By default, `.default` returns a no-op
+    /// registery.
+    internal static var `default`: DatadogFeatureRegistry = NOOPDatadogRegistry()
+}
```
maxep, line 17: This could also be
```swift
internal var DatadogDefaultRegistry: DatadogFeatureRegistry = NOOPDatadogRegistry()
```
`enum DatadogRegistry` is only a namespace here.

Sources/Datadog/Core/DatadogCoreProtocol.swift
```diff
@@ -0,0 +1,54 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+internal var DefaultDatadogCore: DatadogCoreProtocol = NOOPDatadogCore()
```
ncreated, line 9: Is there any reasoning behind the capital `D` letter? `defaultDatadogCore` sounds more idiomatic for Swift 🙂.

Sources/Datadog/Core/DatadogCoreProtocol.swift
```diff
@@ -0,0 +1,54 @@
...
+
+/// A Datadog Feature registry hold a set of features and is responsible of
+/// managing their storage and upload mechanism. It also provide thread-safe
+/// scope for features.
```
ncreated, line 13: Small typos and it's a bit outdated after last renaming, how about:
```suggestion
/// A Datadog Core holds a set of features and is responsible of
/// managing their storage and upload mechanism. It also provides a thread-safe
/// scope for writing events.
```

Sources/Datadog/Core/DatadogCoreProtocol.swift
```diff
@@ -0,0 +1,54 @@
...
+public protocol DatadogCoreProtocol {
+    /// Registers a feature by its name and configuration.
+    ///
+    /// - Parameters:
+    ///   - named: The feature name.
+    ///   - storage: The feature's storage configuration.
+    ///   - upload: The feature's upload configuration.
+    func registerFeature(named: String, storage: FeatureStorageConfiguration, upload: FeatureUploadConfiguration)
```
ncreated, line 21: nit: lets leverage Swift named parameters for more readability:
```swift
registerFeature(named featureName: String, storage: FeatureStorageConfiguration, upload: FeatureUploadConfiguration)
```

Sources/Datadog/Core/DatadogCoreProtocol.swift
```diff
@@ -0,0 +1,54 @@
...
+
+    /// Returns a Feature scope by its name.
+    ///
+    /// - Parameter named: The feature's name.
+    /// - Returns: The feature if it was previously registered, `nil` otherwise.
+    func scope(forFeature named: String) -> FeatureScope?
```
ncreated, line 27: nit: `named` → `featureName`

Sources/Datadog/Core/DatadogCoreProtocol.swift
```diff
@@ -0,0 +1,54 @@
...
+/// Provide feature specific storage configuration.
+public struct FeatureStorageConfiguration {
+    // TODO: RUMM-2133
+}
+
+/// Provide feature specific upload configuration.
+public struct FeatureUploadConfiguration {
+    // TODO: RUMM-2133
+}
+
```
ncreated, line 39: As we discussed in RFC comment, let's make sure we can resign from having `SDKFeatureConfiguration` protocol:
```swift
public protocol SDKFeatureConfiguration {
    func register(to sdkCore: SDKCore)
}
```
without resigning from type safety in features configuration (generic `FeatureStorageConfiguration<Metadata>` and `FeatureUploadConfiguration<Metadata>`). From my experiments in POC, this interface appears to be required, but I might be missing something. Not a blocker for this PR, but something we should discuss 🙌.

Sources/Datadog/Core/DatadogCoreProtocol.swift
```diff
@@ -0,0 +1,54 @@
+/*
```
ncreated, line 1: How about putting these classes in separate `Datadog/DatadogCore` (or whatever) folder? Even though this won't yet be a standalone module, it would establish a conventional module. Just like `Datadog/Core` is a conventional core today, with no dependencies to other folders (other conventional modules).

Sources/Datadog/DatadogCore.swift
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+import Foundation
+
+/// Core implementation of Datadog SDK.
+///
+/// The core provides a storage and upload mechanism for each registered
+/// feature based on their respective configuration.
+///
+/// By complying with `DatadogFeatureRegistry`, the core can
```
ncreated, line 14:
```suggestion
/// By complying with `DatadogCoreProtocol`, the core can
```

Sources/Datadog/Datadog.swift
```diff
@@ -158,8 +162,6 @@ public class Datadog {
         static let DebugRUM = "DD_DEBUG_RUM"
     }
 
-    internal static var instance: Datadog?
-
     internal let consentProvider: ConsentProvider
     internal let userInfoProvider: UserInfoProvider
```
ncreated, line 166: We can now remove these two, no?

Sources/Datadog/Datadog.swift
```diff
@@ -158,8 +162,6 @@ public class Datadog {
...
     internal let launchTimeProvider: LaunchTimeProviderType
```
ncreated, line 167: Why we moved `consentProvider` and `userInforProvider` to `DatadogCore`, but not yet `launchTimeProvider`?

Sources/Datadog/DatadogCore.swift
```diff
@@ -0,0 +1,46 @@
...
+/// provide context and writing scopes to features for event recording.
+internal final class DatadogCore {
+    /// The user consent provider for collecting PII.
+    let consentProvider: ConsentProvider
```
ncreated, line 18: This comment is IMO misleading - tracking consent has nothing to do with PIIs, how about simply:
```suggestion
    /// The user tracking consent provider.
    let consentProvider: ConsentProvider
```

Sources/Datadog/DatadogCore.swift
```diff
@@ -0,0 +1,46 @@
...
+
+    /// User PII.
+    let userInfoProvider: UserInfoProvider
+
+    /// Creates a core instance.
+    ///
+    /// - Parameters:
+    ///   - consentProvider: The user consent provider for collecting PII.
```
ncreated, line 26:
```suggestion
    ///   - consentProvider: The user tracking consent provider.
```

Sources/Datadog/Logger.swift
```diff
@@ -401,9 +401,9 @@ public class Logger {
         private func buildOrThrow() throws -> Logger {
             guard let loggingFeature = LoggingFeature.instance else {
                 throw ProgrammerError(
-                    description: Datadog.instance == nil
-                        ? "`Datadog.initialize()` must be called prior to `Logger.builder.build()`."
-                        : "`Logger.builder.build()` produces a non-functional logger, as the logging feature is disabled."
+                    description: Datadog.isInitialized
```
ncreated, line 404: Sooner or later we'll need to also remove `Datadog` references like this and change them to use `DatadogCore`. Now we could replace this with the use of `DefaultDatadogCore`. Is the plan to change things like this in separate PR(s) or do you think it's better to wait and tackle it at once when isolating real modules?

ncreated (changes requested):
Really well done 🏅💯 . I left some feedbacks, mainly on typos and few conventions.

Sources/Datadog/Core/DatadogCoreProtocol.swift
maxep: That's a mistake :)

Sources/Datadog/Datadog.swift
maxep: We should, yes 👍

Sources/Datadog/Datadog.swift
maxep: I've noticed that `Datadog.instance?.launchTimeProvider` was never in used, it is injected as part of the `FeaturesCommonDependencies`.

Sources/Datadog/Logger.swift
maxep: Indeed, this will be removed when we remove each feature singleton. Once all features are migrated, we will rely on injected core.

Sources/Datadog/Core/DatadogCoreProtocol.swift
maxep: Yes, this strongly depends on how we will expect users to initialize features and SDK instances. We can discuss first and see how we move forward 👍

ncreated (approved):
All good 💪👌

maxep:
I've fixed the Jira ids in git commit.

PR #854, opened 2022-05-16
RUMM-2133 Add v1 feature registry

maxep:
### What and why?

Creates a v1 feature registry in `DatadogCore` to use while migrating to v2 architecture.

### How?

`DatadogCoreProtocol` now expect 2 methods specific to v1 features registering.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/DatadogCore/DatadogCore.swift
+++ b/Sources/Datadog/DatadogCore/DatadogCore.swift
@@ -20,6 +20,8 @@ internal final class DatadogCore {
     /// User PII.
     let userInfoProvider: UserInfoProvider
 
+    private var v1Features: [String: Any] = [:]
+
     /// Creates a core instance.
     ///
     /// - Parameters:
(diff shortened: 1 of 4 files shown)


Sources/Datadog/DatadogCore/DatadogCore.swift
```diff
@@ -43,4 +45,14 @@ extension DatadogCore: DatadogCoreProtocol {
...
+
+    // MARK: V1 interface
+
+    func registerFeature(named featureName: String, instance: Any?) {
+        v1Features[featureName] = instance
+    }
+
+    func feature<T>(_ type: T.Type, named featureName: String) -> T? {
+        return v1Features[featureName] as? T
+    }
```
ncreated, line 57: Just making sure - this interface is tailored for V1 features, right? Asking, because what we need for V2 is very different 🙂. In V2 features must be instantiated from `registerFeature(named:...)` and we'll only have a single return type in `feature(named:)`.

maxep: Exactly, I've kept the v2 methods above the `// MARK: V1 interface`. These 2 methods are here while we migrate v1 features, so we can start relying on core for feature registry but we don't have yet storage and upload mechanism in core.

ncreated (approved):
All good 👌👌

PR #857, opened 2022-05-17
RUMM-2133 Register Tracing v1 in `DatadogCore`

maxep:
### What and why?

Register Tracing v1 in `DatadogCore`.

### How?

Remove `TracingFeature.instance` singleton and other static access and expect a `DatadogCoreProtocol` complying instance when creating a `Tracer`, `defaultDatadogCore` is used by default.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/Datadog.swift
+++ b/Sources/Datadog/Datadog.swift
@@ -155,7 +155,8 @@ public class Datadog {
     public static func clearAllData() {
         let logging = defaultDatadogCore.feature(LoggingFeature.self, named: LoggingFeature.featureName)
         logging?.storage.clearAllData()
-        TracingFeature.instance?.storage.clearAllData()
+        let tracing = defaultDatadogCore.feature(TracingFeature.self, named: TracingFeature.featureName)
+        tracing?.storage.clearAllData()
         RUMFeature.instance?.storage.clearAllData()
     }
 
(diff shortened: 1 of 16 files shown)


Sources/Datadog/Datadog.swift
```diff
@@ -270,10 +271,12 @@ public class Datadog {
                 directories: try obtainTracingFeatureDirectories(),
                 configuration: tracingConfiguration,
                 commonDependencies: commonDependencies,
-                loggingFeatureAdapter: logging.flatMap { LoggingForTracingAdapter(loggingFeature: $0) },
+                loggingFeatureAdapter: logging.map { LoggingForTracingAdapter(loggingFeature: $0) },
```
ncreated, line 274: I just wonder, was `map` always possible in such context or has it changed when `compactMap` was introduced?

ncreated, line 274: I just wonder, was `map` always possible in such context or has it changed when `compactMap` was introduced?

ncreated (approved):
Looks good 👌👌

Sources/Datadog/Datadog.swift
maxep: `map` was always possible here as we return a non-optional value. `map` is just syntactically more correct!

PR #861, opened 2022-05-19
RUMM-2133 Register RUM v1 in `DatadogCore`

maxep:
### What and why?

Register RUM v1 in `DatadogCore`.

### How?

Remove `RUMFeature.instance` singleton and other static access and expect a `DatadogCoreProtocol` complying instance when creating a `RUMMonitor`, `defaultDatadogCore` is used by default.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/Datadog.swift
+++ b/Sources/Datadog/Datadog.swift
@@ -254,6 +254,8 @@ public class Datadog {
                     configuration: instrumentationConfiguration,
                     dateProvider: dateProvider
                 )
+
+                core.register(feature: rumInstrumentation)
             }
         }
 
@@ -298,8 +300,7 @@ public class Datadog {
 
         CrashReportingFeature.instance = crashReporting
 
-        RUMInstrumentation.instance = rumInstrumentation
-        RUMInstrumentation.instance?.enable()
+        core.feature(RUMInstrumentation.self)?.enable()
 
         URLSessionAutoInstrumentation.instance = urlSessionAutoInstrumentation
         URLSessionAutoInstrumentation.instance?.enable()
@@ -339,12 +340,13 @@ public class Datadog {
         let logging = defaultDatadogCore.feature(LoggingFeature.self, named: LoggingFeature.featureName)
         let tracing = defaultDatadogCore.feature(TracingFeature.self, named: TracingFeature.featureName)
         let rum = defaultDatadogCore.feature(RUMFeature.self, named: RUMFeature.featureName)
+        let rumInstrumentation = defaultDatadogCore.feature(RUMInstrumentation.self)
         logging?.deinitialize()
         tracing?.deinitialize()
         rum?.deinitialize()
+        rumInstrumentation?.deinitialize()
 
         CrashReportingFeature.instance?.deinitialize()
-        RUMInstrumentation.instance?.deinitialize()
         URLSessionAutoInstrumentation.instance?.deinitialize()
 
         // Reset Globals:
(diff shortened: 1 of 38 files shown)


Sources/Datadog/DatadogInternal/DatadogCoreProtocol.swift
```diff
@@ -46,6 +46,22 @@ public protocol DatadogCoreProtocol {
 }
 
 extension DatadogCoreProtocol {
+    /// Registers a feature instance by its type description.
+    ///
+    /// - Parameter instance: The feaure instance to register
+    func register<T>(feature instance: T?) {
+        registerFeature(named: String(describing: T.self), instance: instance)
+    }
```
maxep, line 54: Features' name in v1 does not make much sense, I'm thinking of using this method for all v1 features!

maxep, line 54: Features' name in v1 does not make much sense, I'm thinking of using this method for all v1 features!

Sources/Datadog/Datadog.swift
```diff
@@ -95,10 +95,8 @@ public class Datadog {
                 consolePrint("⚠️ Overriding RUM debugging due to \(LaunchArguments.DebugRUM) launch argument")
                 Datadog.debugRUM = true
             }
-        } catch let error as ProgrammerError {
-            consolePrint("\(error)")
         } catch {
-            defaultDatadogCore = NOOPDatadogCore()
+            consolePrint("\(error)")
```
maxep, line 99: I've revert this so we have a console print on every failure. The `defaultDatadogCore` is set at the very end of the initialisation.

Sources/Datadog/Datadog.swift
```diff
@@ -247,11 +247,15 @@ public class Datadog {
...
 
+            core.registerFeature(named: RUMFeature.featureName, instance: rum)
+
             if let instrumentationConfiguration = rumConfiguration.instrumentation {
                 rumInstrumentation = RUMInstrumentation(
                     configuration: instrumentationConfiguration,
                     dateProvider: dateProvider
                 )
+
+                core.register(feature: rumInstrumentation)
```
ncreated, line 258: Totally fine for now 👍, although I think `rumInstrumentation` shouldn't be a separate feature in V2. It's rather a component of RUM feature.

Sources/Datadog/DatadogInternal/DatadogCoreProtocol.swift
ncreated: Agree 👍, although shouldn't it replace:
```
func registerFeature(named featureName: String, instance: Any?)
```
in `DatadogCoreProtocol`, instead of forwarding? Now we have some features registered with `register(feature:)` and some with `registerFeature(named:instance:)`. It is very dangerous as both are not compatible and we need to know their registration method to use matching variant when obtaining the instance. Even if this is temporary shim between V1 and V2, I'd keep it straight as it can get messy.

ncreated (changes requested):
Looks ok 👍, especially the update on registering V1 features. However, I think we should update everything else to have a single way, otherwise it can get messy. WDYT?

Sources/Datadog/Datadog.swift
maxep: Agreed, this will change when we move to v2 configuration/scope. What about `URLSession` instrumentation? it is used by both Tracing and RUM.
Given how the instrumentation works with custom `URLSessionDelegate`, it could make sense to have a separate config.. WDYT?

ncreated: Yes, agree - `URLSessoinInstrumentation` needs separate config 👌. I think we may need to put its code inside API module so it can be shared between RUM and Tracing. It also won't be a "feature" if compared to Logging, RUM and Tracing, so we may need to find a good shape for it.

PR #865, opened 2022-05-23
RUMM-2133 Register `URLSession` Instrumentation v1 in `DatadogCore`

maxep:
### What and why?

Register `URLSession` instrumentation v1 in `DatadogCore`.

### How?

Remove `URLSessionAutoInstrumentation.instance` singleton and other static access and expect a `DatadogCoreProtocol` complying instance when creating a `DDURLSessionDelegate`, `defaultDatadogCore` is used by default.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Tests/DatadogTests/Datadog/RUMMonitorTests.swift
+++ b/Tests/DatadogTests/Datadog/RUMMonitorTests.swift
@@ -1371,7 +1371,8 @@ class RUMMonitorTests: XCTestCase {
         userLogger = .mockWith(logOutput: output)
 
         // Given
-        let resourcesHandler = try XCTUnwrap(URLSessionAutoInstrumentation.instance?.interceptor.handler)
+        let urlInstrumentation = try XCTUnwrap(defaultDatadogCore.feature(URLSessionAutoInstrumentation.self))
+        let resourcesHandler = urlInstrumentation.interceptor.handler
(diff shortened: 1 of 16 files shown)


Tests/DatadogTests/Datadog/RUMMonitorTests.swift
```diff
@@ -1422,10 +1423,6 @@ class RUMMonitorTests: XCTestCase {
             Make sure `Global.rum = RUMMonitor.initialize()` is called before any action happens.
             """
         )
-
-        URLSessionAutoInstrumentation.instance?.swizzler.unswizzle()
-        rumInstrumentation.viewControllerSwizzler?.unswizzle()
-        rumInstrumentation.userActionsAutoInstrumentation?.swizzler.unswizzle()
```
ncreated, line 1428: We don't need it because it is part of deinitialization, right?
```swift
Datadog.flushAndDeinitialize()
```

maxep: Yes exactly!

maxep:
> Swizzlings are static / global by nature and we may need some extra effort to make them know the core instance for redirection.

We technically can apply multiple swizzling to a same method and from our usage of swizzling in RUM and URLSession, it looks like we always use a reference to a `Handler` for RUM and to `session.delegate`, so we should be redirecting to the right feature instance in core 👍 Worth testing tho, but it should work :)

PR #866, opened 2022-05-25
RUMM-2169 Initialise V1 Logging with V2 configuration

ncreated:
### What and why?

⚙️🧰 This PR makes necessary refactoring to create V1 `LoggingFeature` from V2's storage and upload configurations.

### How?

Feature initialisation is now generic and can happen in `DatadogCore` with:
```swift
func create<Feature: V1Feature>(
   storageConfiguration: FeatureStorageConfiguration,
   uploadConfiguration: FeatureUploadConfiguration,
   featureSpecificConfiguration: Feature.Configuration
) throws -> Feature 
```
The `featureSpecificConfiguration` is used to avoid impactful refactoring. It holds the feature-specific part of the config ([`FeaturesConfiguration.Logging`](https://github.com/DataDog/dd-sdk-ios/blob/b1c10dceeca16ab6e8f6a0f0314cedcf0b13dd61/Sources/Datadog/Core/FeaturesConfiguration.swift#L28-L32)). It will be removed as the last step after migrating other features to this setup.

I had to do 2 extra V2 migration steps to make this PR simpler:
- 🎁 `DatadogCore` is now configured with `rootDirectory`. This made tests refactoring way easier and it is very much aligned with what we want in V2 (each SDK instance has its own root folder).
- 💡 I had to introduce `DatadogV1Context`, a counterpart of what we need in V2. This allowed me to pass `RequestBuilder` attributes from Core to Feature.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/DatadogCore/DatadogCore.swift
+++ b/Sources/Datadog/DatadogCore/DatadogCore.swift
@@ -12,6 +12,20 @@ internal typealias CoreConfiguration = FeaturesConfiguration.Common
 /// Feature-agnostic set of dependencies powering Features storage, upload and event recording.
 internal typealias CoreDependencies = FeaturesCommonDependencies
 
+/// A shim interface for uniforming V1 Features and letting their generic initialization in `DatadogCore`.
+internal protocol V1Feature {
+    /// The configuration specific to this Feature.
+    /// In V2 this will likely become a part of the public interface for the Feature module.
+    associatedtype Configuration
+
+    init(
+        storage: FeatureStorage,
+        upload: FeatureUpload,
+        configuration: Configuration,
+        commonDependencies: FeaturesCommonDependencies
+    )
+}
+
 /// Core implementation of Datadog SDK.
 ///
 /// The core provides a storage and upload mechanism for each registered Feature
@@ -24,6 +38,8 @@ internal final class DatadogCore {
     let configuration: CoreConfiguration
     /// A set of dependencies used by SDK core for powering Features.
     let dependencies: CoreDependencies
+    /// Telemetry monitor, if configured.
+    var telemetry: Telemetry?
 
     private var v1Features: [String: Any] = [:]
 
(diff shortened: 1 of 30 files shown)


Sources/Datadog/DatadogCore/DatadogCore.swift
```diff
@@ -6,33 +6,60 @@
...
+        commonDependencies: FeaturesCommonDependencies
+    )
+}
+
 /// Core implementation of Datadog SDK.
 ///
-/// The core provides a storage and upload mechanism for each registered
-/// feature based on their respective configuration.
+/// The core provides a storage and upload mechanism for each registered Feature
+/// based on their respective configuration.
```
ncreated, line 32: Here, I started using "**F**eature" (Logging / RUM / Tracing) instead of "**f**eature" (e.g. sampling or mapping) to distinguish it in doc comments. Does it make sense @maxep?

Sources/Datadog/DatadogInternal/DatadogCoreProtocol.swift
```diff
@@ -11,21 +11,7 @@ public internal(set) var defaultDatadogCore: DatadogCoreProtocol = NOOPDatadogCo
...
-    ///   - upload: The feature's upload configuration.
-    func registerFeature(named featureName: String, storage: FeatureStorageConfiguration, upload: FeatureUploadConfiguration)
-
-    /// Returns a Feature scope by its name.
-    ///
-    /// - Parameter featureName: The feature's name.
-    /// - Returns: The scope for feature that previously registered, `nil` otherwise.
-    func scope(forFeature featureName: String) -> FeatureScope?
-
-    // MARK: V1 interface
```
ncreated, line 28: As discussed, we will bring it back when `FeatureStorageConfiguration` and `FeatureUploadConfiguration` can become `public` APIs. It's not possible now, as they depend on things that will drag a lot of other types if made public. I should be able to do something with this when both RUM and Tracing are also migrated to V2 config.

Tests/DatadogBenchmarkTests/DataStorage/LoggingStorageBenchmarkTests.swift
```diff
@@ -19,10 +19,12 @@ class LoggingStorageBenchmarkTests: XCTestCase {
...
-            directories: FeatureDirectories(
+        let storage = FeatureStorage(
+            featureName: "logging",
+            dataFormat: DataFormat(prefix: "[", suffix: "]", separator: ","),
+            directories: .init(
                 deprecated: [],
-                unauthorized: obtainUniqueTemporaryDirectory(),
+                unauthorized: directory,
                 authorized: directory
             ),
```
ncreated, line 29: It now becomes pointless to have separate storage benchmarks for Logs, Traces and RUM as all use the same logic under the hood. To make CI pass, I'm just updating it for now, but we can re-consider Benchmarks module as part of V2 migration. It will be tedious in maintenance.

Sources/Datadog/DatadogCore/DatadogCore.swift
```diff
@@ -6,33 +6,60 @@
...
+    /// The configuration specific to this Feature.
+    /// In V2 this will likely become a part of the public interface for the Feature module.
+    associatedtype Configuration
+
+    init(
+        storage: FeatureStorage,
+        upload: FeatureUpload,
+        configuration: Configuration,
+        commonDependencies: FeaturesCommonDependencies
+    )
```
maxep, line 26: I like the idea of type constraint on v1 Features, but I'm wondering if we could get rid of the `init` and `associatedtype` 🤔 We could simply use convention to initialise v1 features and be able to register `V1Feature` conforming instances:
```swift
public protocol DatadogCoreProtocol {
    // MARK: V1 interface

    func register<T>(feature instance: T?) where T: V1Feature

    func feature<T>(_ type: T.Type) -> T? where T: V1Feature
}
```
It would then be easier to add other properties to `V1Feature`, such as storage and upload. WDYT?

maxep (approved):
Very nice work 👌 I've left one question!

Sources/Datadog/DatadogCore/DatadogCore.swift
ncreated: Indeed 👍. I tackled this by renaming `V1Feature` to `V1FeatureInitializable` in https://github.com/DataDog/dd-sdk-ios/pull/871/commits/feb5766cd7554bacbb3e3ab747793b545319b96d. This is only conventional change, but will allow for eventually having non-constrained `V1Feature` protocol:

```swift
internal protocol: V1FeatureInitializable: V1Feature {
   // constrained interface of `V1FeatureInitializable`
}

protocol V1Feature {
   // not constrained V1 feature interface
}
```
that can appear in non-generic context (e.g. `v1Features: [String: V1Feature]`). I leave that door open without actually adding `V1Feature` as it requires more refactoring (e.g. if we want it to appear in `public` interface due to `DatadogCoreProtocol` being `public`).

PR #867, opened 2022-05-25
RUMM-2173 Add log level threshold option

xgouchet:
### What and why?

Add a way to filter logs per status before they're sent to Datadog.

### How?

Added an option in the logger builder, that's passed to the LogFileOutput.

### Review checklist

### DIFF
--- a/Sources/Datadog/Logger.swift
+++ b/Sources/Datadog/Logger.swift
@@ -359,7 +359,7 @@ public class Logger {
             return self
         }
 
-        /// set the minim log level reported to Datadog servers.
+        /// Set the minim log level reported to Datadog servers.
         /// Any log with a level equal or above to the threshold will be sent
         /// - Parameter datadogReportingThreshold: `LogLevel.debug` by default
         public func set(datadogReportingThreshold: LogLevel) -> Builder {
@@ -446,10 +446,12 @@ public class Logger {
             case (true, let format?):
                 let logOutput = CombinedLogOutput(
                     combine: [
-                        LogFileOutput(
-                            fileWriter: loggingFeature.storage.writer,
-                            rumErrorsIntegration: LoggingWithRUMErrorsIntegration(),
-                            reportingThreshold: self.datadogReportingThreshold
+                        ConditionalLogOutput(
+                            conditionedOutput: LogFileOutput(
+                                fileWriter: loggingFeature.storage.writer,
+                                rumErrorsIntegration: LoggingWithRUMErrorsIntegration()
+                            ),
+                            condition: reportLogsAbove(threshold: datadogReportingThreshold)
                         ),
                         LogConsoleOutput(
                             format: format,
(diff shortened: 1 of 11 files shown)


docs/log_collection.md
```diff
@@ -239,13 +239,15 @@ DDDatadog.verbosityLevel = DDSDKVerbosityLevelDebug;
...
     .printLogsToConsole(true, usingFormat: .shortWith(prefix: "[iOS App] "))
+    .set(datadogReportingThreshold: .info)
     .build()
 ```
 {{% /tab %}}
 {{% tab "Objective-C" %}}
 ```objective-c
 DDLoggerBuilder *builder = [DDLogger builder];
 [builder sendNetworkInfo:YES];
+[builder setWithDatadogReportingThreshold:.info]
```
buraizu, line 250: It looks like there may be a missing semicolon at the end of this line. Could you confirm and update if needed?

buraizu (changes requested):

Sources/Datadog/Logger.swift
```diff
@@ -451,7 +461,8 @@ public class Logger {
             case (true, nil):
                 let logOutput = LogFileOutput(
                     fileWriter: loggingFeature.storage.writer,
-                    rumErrorsIntegration: LoggingWithRUMErrorsIntegration()
+                    rumErrorsIntegration: LoggingWithRUMErrorsIntegration(),
+                    reportingThreshold: self.datadogReportingThreshold
                 )
                 return (logBuilder, logOutput)
```
ncreated, line 467: I wonder if we shouldn't consider using [`ConditionalLogOutput`](https://github.com/DataDog/dd-sdk-ios/blob/a63954b0ca095d99205dd8000651527b447f03e6/Sources/Datadog/Logging/LogOutputs/LogUtilityOutputs.swift#L21-L22) instead of adding threshold information directly to `LogFileOutput`? It can be simpler approach - not requiring modifications to every instance of `LogFileOutput` in the SDK and only limited to this file.
```swift
let logOutput = ConditionalLogOutput(
   conditionedOutput: LogFileOutput(...), 
   condition: reportLogsAbove(threshold: datadogReportingThreshold)
)

func reportLogsAbove(threshold: LogLevel) -> (LogEvent) -> Bool {
    return { log in
        let logLevel = LogLevel(from: log.status)?.rawValue ?? 0
        return logLevel >= threshold.rawValue
    }
}
```
As a bonus, we could share `reportLogsAbove(threshold:)` logic with [internal logger](https://github.com/DataDog/dd-sdk-ios/blob/a63954b0ca095d99205dd8000651527b447f03e6/Sources/Datadog/Utils/InternalLoggers.swift#L65-L69) which does exactly the same conditioning.

ncreated (approved):
Looks fine, thanks! 👌

tools/lint/run-linter.sh
```diff
@@ -11,6 +11,6 @@ if [[ -z "${XCODE_VERSION_ACTUAL}" ]]; then
 	swiftlint lint --config ./tools/lint/tests.swiftlint.yml --reporter "emoji" --strict
 else
 	# when run by Xcode in Build Phase
-	swiftlint lint --config ./tools/lint/sources.swiftlint.yml --reporter "xcode"
-	swiftlint lint --config ./tools/lint/tests.swiftlint.yml --reporter "xcode"
+	swiftlint lint --config ./tools/lint/sources.swiftlint.yml --reporter "xcode" --fix
+	swiftlint lint --config ./tools/lint/tests.swiftlint.yml --reporter "xcode" --fix
```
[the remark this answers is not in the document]
ncreated: We're good w/ or w/o it 👍. Now it's removed - we can discuss it later when working on linter updates after V2.

buraizu (approved):

PR #873, opened 2022-05-30
Patch Browser Action Events with array of ids

maxep:
### What and why?

Browser SDK support action id as string or as array of strings as defined in schema:
https://github.com/DataDog/rum-events-format/blob/master/schemas/rum/_action-child-schema.json#L15L31

### How?

This is not yet supported by the Swift schema generator, as a workaround we pick the first id when
we encounter an array.

### Review checklist

- [ ] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


ncreated (approved):
🩹⛑✅

PR #874, opened 2022-05-31
RUMM-2171 Single Storage Queue

maxep:
### What and why?

Use single storage queue.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Tests/DatadogTests/Datadog/DatadogTests.swift
+++ b/Tests/DatadogTests/Datadog/DatadogTests.swift
@@ -438,19 +438,11 @@ class DatadogTests: XCTestCase {
         )
 
         let core = try XCTUnwrap(defaultDatadogCore as? DatadogCore)
-        let logging = core.feature(LoggingFeature.self)
-        let tracing = core.feature(TracingFeature.self)
-        let rum = core.feature(RUMFeature.self)
 
         // On SDK init, underlying `ConsentAwareDataWriter` performs data migration for each feature, which includes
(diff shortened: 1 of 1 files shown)


Sources/Datadog/DatadogCore/DatadogCore.swift
```diff
@@ -45,6 +45,11 @@ internal final class DatadogCore {
     /// Telemetry monitor, if configured.
     var telemetry: Telemetry?
 
+    let readWriteQueue = DispatchQueue(
+        label: "com.datadoghq.ios-sdk-read-write",
+        target: .global(qos: .utility)
+    )
```
ncreated, line 51: There is one potential step we could look into in this PR. Now, by having an explicit access to this r/w queue we can simplify this test:

https://github.com/DataDog/dd-sdk-ios/blob/94819c05807fcd9b77bc85b8da0de59bc29ec82b/Tests/DatadogTests/Datadog/DatadogTests.swift#L445-L453

We `.sync {}` on each Feature's queue, but now it can be single `core.readWriteQueue.sync {}`.

ncreated (approved):
Looks good 👌💯. I left one feedback on potential small improvement in tests.

Sources/Datadog/DatadogCore/DatadogCore.swift
maxep: Good cartch!

PR #875, opened 2022-06-01
RUMM-2171 V1 Interface Segregation

maxep:
### What and why?

SDK V1 implementation defines multiple `internal` objects that need to stay hidden from the public interface while we migrate to v2 core module and configuration. The PR is a proposal to segregate v1 from v2 interfaces to be able to keep all v1 apis `internal`.

### How?

The `DatadogCoreProtocol` will internally expose a `v1` extension to access all v1 apis needed to register and retrieve features.

``` swift
extension DatadogCoreProtocol {
    // Upcast `DatadogCoreProtocol` instance to `DatadogV1CoreProtocol`
    // to access v1 related implementation.
    // If upcasting fails, a `NOOPDatadogCore` instance is returned.
    var v1: DatadogV1CoreProtocol {
        self as? DatadogV1CoreProtocol ?? NOOPDatadogCore()
    }
}
```

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/DatadogCore/DatadogCore.swift
+++ b/Sources/Datadog/DatadogCore/DatadogCore.swift
@@ -99,7 +99,7 @@ internal final class DatadogCore {
     }
 }
 
-extension DatadogCore: DatadogCoreProtocol {
+extension DatadogCore: DatadogV1CoreProtocol {
     // MARK: - V1 interface
 
     /// Creates V1 Feature using its V2 configuration.
@@ -158,4 +158,8 @@ extension DatadogCore: DatadogCoreProtocol {
         let key = String(describing: T.self)
         return v1Features[key] as? T
     }
+}
+
+internal protocol V1Feature {
+    var storage: FeatureStorage { get }
 }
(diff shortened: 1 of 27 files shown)


Sources/Datadog/DatadogCore/DatadogCore.swift
```diff
@@ -12,21 +12,6 @@ internal typealias CoreConfiguration = FeaturesConfiguration.Common
...
-
-    init(
-        storage: FeatureStorage,
-        upload: FeatureUpload,
-        configuration: Configuration,
-        commonDependencies: FeaturesCommonDependencies,
-        telemetry: Telemetry?
-    )
-}
-
```
maxep, line 29: My issue with this protocol is that it creates a dependency of Features to the `DatadogCore`, we should avoid coupling those two layers.
I'm aware that `FeaturesConfiguration.Common` and `FeaturesCommonDependencies` are also coupling those two, I'm considering moving those configuration to the new `DatadogV1CoreProtocol`, WDYT?

maxep: Reverted 👍

ncreated (approved):
Awesome 🎯. Simple and convenient for V1 → V2 transitioning 👍.

PR #851, opened 2022-05-12
RUMM-2166 Add tracing sampling rate

maxep:
### What and why?

Offer a configuration option to set the tracing sampling rate. Based on this sampling rate, new network requests will have the `x-datadog-sampling-priority` header set to 0 (or 1 accordingly), and the `trace_id` and `span_id` won’t (or will accordingly) be added to the created RUM Resource.

__Note__: The `x-datadog-sampled` is now removed.

### How?

The `DDSpanContext` now include a boolean `samplingPriority` property. If it's `false`, the trace id and span id are **not** propagated to APM, RUM, nor Logs integrations. The `x-datadog-sampling-priority` is set accordingly when a request is intercepted

The `samplingPriority` is set by the `Tracer`'s newly added `sampler`.

### Review checklist
- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes

### Custom CI job configuration (optional)
- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

PR #876, opened 2022-06-02
RUMM-2077 Vitals Frequency Configuration

maxep:
### What and why?

Customers can set the desired delay between Mobile Vitals update (CPU/RAM) as follows:

- FREQUENT - every 100ms 
- AVERAGE - 500 ms 
- RARE - 1000ms
- NEVER - don’t provide vitals data

### How?

`VitalInfoSampler` will start the sample timer only if `frequency` parameter is available.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 ### Changes
+* [IMPROVEMENT] Add mobile vitals frequency configuration. See [#876][]
 
 # 1.11.0-rc1 / 18-05-2022
 
@@ -367,6 +368,7 @@
 [#837]: https://github.com/DataDog/dd-sdk-ios/issues/837
 [#832]: https://github.com/DataDog/dd-sdk-ios/issues/832
 [#851]: https://github.com/DataDog/dd-sdk-ios/issues/851
+[#876]: https://github.com/DataDog/dd-sdk-ios/issues/876
 [@00FA9A]: https://github.com/00FA9A
 [@Britton-Earnin]: https://github.com/Britton-Earnin
 [@Hengyu]: https://github.com/Hengyu
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMScopeDependencies.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMScopeDependencies.swift
@@ -10,6 +10,13 @@ internal typealias RUMSessionListener = (String, Bool) -> Void
 
 /// Dependency container for injecting components to `RUMScopes` hierarchy.
 internal struct RUMScopeDependencies {
+    struct VitalsReaders {
+        let frequency: TimeInterval
+        let cpu: SamplingBasedVitalReader
+        let memory: SamplingBasedVitalReader
+        let refreshRate: ContinuousVitalReader
+    }
+
     let rumApplicationID: String
     let sessionSampler: Sampler
     /// The start time of the application, indicated as SDK init. Measured in device time (without NTP correction).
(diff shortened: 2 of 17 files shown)


CHANGELOG.md
```diff
@@ -1,13 +1,14 @@
...
 ### Changes
+* [IMPROVEMENT] Add mobile vitals frequency configuration. See [#851][]
 
 # 1.11.0-rc1 / 18-05-2022
 
 ### Changes
 
 * [IMPROVEMENT] Allow manually tracked resources in RUM Sessions to detect first party hosts. See [#837][]
-* [IMPROVEMENT] Add tracing sampling rate. See [#851][]
+* [IMPROVEMENT] Add tracing sampling rate. See [#876][]
```
ncreated, line 11: This seems wrong, the current PR is #876 and tracing sampling was #851.

Sources/DatadogObjc/DatadogConfiguration+objc.swift
```diff
@@ -379,6 +396,11 @@ public class DDConfigurationBuilder: NSObject {
         }
     }
 
+    @objc
+    public func set(mobileVitalsFrequency: DDVitalsFrequency) {
```
ncreated, line 400: nit: Let's add it to [`DDConfiguration+apiTests.m`](https://github.com/DataDog/dd-sdk-ios/blob/e2a961ddc9439e523cf234829bd9a27a4d79451d/Tests/DatadogTests/DatadogObjc/ObjcAPITests/DDConfiguration%2BapiTests.m#L92) 🙌.

Sources/Datadog/RUM/RUMVitals/VitalInfoSampler.swift
```diff
@@ -54,7 +52,7 @@ internal final class VitalInfoSampler {
         cpuReader: SamplingBasedVitalReader,
         memoryReader: SamplingBasedVitalReader,
         refreshRateReader: ContinuousVitalReader,
-        frequency: TimeInterval = VitalInfoSampler.frequency,
+        frequency: TimeInterval?,
```
ncreated, line 55: How about requiring `frequency` and having optional `VitalInfoSampler` in `RUMViewScope`? That could be more consistent IMO, as `nil` would mean that Mobile Vitals are disabled. Now, if passing `frequency: nil` we make the sampler no-op.

This is not blocking, I'm ok with either way 👍, just wondering what's your take on this @maxep

ncreated (changes requested):
Looks good 👍, but there's a blocking issue in `CHANGELOG.md`.

CHANGELOG.md
maxep: oops!

Sources/Datadog/RUM/RUMVitals/VitalInfoSampler.swift
maxep: The `refreshRateReader` which is of type `ContinuousVitalReader` is not impacted by the `frequency` parameter, so should we disable the refresh vital as well if the frequency is `.never`?

ncreated: Ah, it makes sense then. What I suggested would require more refactoring to have `nil` instead of no-ops (e.g. having multiple samplers, one for each `reader`). Definitely not sth worth doing now if no other reason 👍.

> so should we disable the refresh vital as well if the frequency is .never?

This is a very good question though 🤔💭. AFAIK on Android we disable Refresh Rate vitals if `.never` is provided (cc @xgouchet @mariusc83). Also, "disable" is what we state in the doc:
```swift
/// Don't provide mobile vitals.
case never
```

maxep: So we confirmed to disable **all** vitals ✅

alai97 (approved):

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMScopeDependencies.swift
```diff
@@ -10,6 +10,13 @@ internal typealias RUMSessionListener = (String, Bool) -> Void
 
 /// Dependency container for injecting components to `RUMScopes` hierarchy.
 internal struct RUMScopeDependencies {
+    struct VitalsReaders {
+        let frequency: TimeInterval
+        let cpu: SamplingBasedVitalReader
+        let memory: SamplingBasedVitalReader
+        let refreshRate: ContinuousVitalReader
+    }
```
maxep, line 18: So dependency help instantiating all readers only if `frequency` is available.

ncreated (approved):
All good 👌👌

PR #877, opened 2022-06-02
RUMM-2169 Isolate Logging feature from Core's configuration and dependencies

ncreated:
### What and why?

🔨🧰 This PR makes `LoggingFeature` no longer depend on `CoreConfiguration` and `CoreDependencies`.

This refactoring is required to better isolate Feature modules from Core implementation. It prepares the codebase for further introduction of SDK context in V2.

### How?

There are two main changes and everything else is mainly their consequence:
* Common configuration is now removed from `FeaturesConfiguration.Logging`
```diff
struct Logging {
-   let common: Common
   let uploadURL: URL
   let logEventMapper: LogEventMapper?
}
```
* Common dependencies are now removed from `LoggingFeature` interface
```diff
internal final class LoggingFeature: V1FeatureInitializable {
   // ...

-   // MARK: - Dependencies
- 
-   let dateProvider: DateProvider
-   let dateCorrector: DateCorrectorType
-   let userInfoProvider: UserInfoProvider
-   let networkConnectionInfoProvider: NetworkConnectionInfoProviderType
-   let carrierInfoProvider: CarrierInfoProviderType

   // ...
}
```

In V2 both, "common configuration" and "common dependencies" will belong to the Core module and their values will be passed through SDK context.

🎁 I made a bonus change, which should help us in transitioning to V2-like Features testing. The way we now create `LoggingFeature` in tests does no longer depend on dummy `DatadogCore` instance. Instead, the feature is instantiated with mocked `FeatureStorage` and `FeatureUpload`:
* `FeatureUpload` is now 100% no-op as we won't need any uploads in V2 Features tests;
* `FeatureStorage` uses new `InMemoryWriter`, which records events in-memory and returns them back for assertions.

This bonus change is only scoped to configuration update in `LoggingFeatureMocks` and addition of `InMemoryWriter`. It's transparent for everything else.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Tests/DatadogTests/Datadog/LoggerTests.swift
+++ b/Tests/DatadogTests/Datadog/LoggerTests.swift
@@ -9,15 +9,12 @@ import XCTest
 
 // swiftlint:disable multiline_arguments_brackets
 class LoggerTests: XCTestCase {
-    let core = DatadogCoreMock()
-
     override func setUp() {
         super.setUp()
         temporaryDirectory.create()
     }
 
     override func tearDown() {
-        core.flush()
         temporaryDirectory.delete()
         super.tearDown()
     }
@@ -33,18 +30,15 @@ class LoggerTests: XCTestCase {
                     serviceName: "default-service-name",
                     environment: "tests",
                     sdkVersion: "1.2.3"
+                ),
+                dependencies: .mockWith(
+                    dateProvider: RelativeDateProvider(using: .mockDecember15th2019At10AMUTC())
                 )
             )
         )
+        defer { core.flush() }
 
-        let feature: LoggingFeature = .mockByRecordingLogMatchers(
-            directory: temporaryDirectory,
-            dependencies: .mockWith(
-                dateProvider: RelativeDateProvider(using: .mockDecember15th2019At10AMUTC())
-            )
-        )
-        defer { feature.deinitialize() }
-
+        let feature: LoggingFeature = .mockByRecordingLogMatchers(directory: temporaryDirectory)
         core.register(feature: feature)
 
         let logger = Logger.builder.build(in: core)
@@ -67,8 +61,10 @@ class LoggerTests: XCTestCase {
     }
 
     func testSendingLogWithCustomizedLogger() throws {
+        let core = DatadogCoreMock(v1Context: .mockAny())
+        defer { core.flush() }
+
         let feature: LoggingFeature = .mockByRecordingLogMatchers(directory: temporaryDirectory)
-        defer { feature.deinitialize() }
         core.register(feature: feature)
 
         let logger = Logger.builder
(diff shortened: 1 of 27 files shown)


Tests/DatadogTests/Datadog/LoggerTests.swift
```diff
@@ -724,49 +709,21 @@ class LoggerTests: XCTestCase {
         }
     }
 
-    // MARK: - Tracking Consent
-
-    func testWhenChangingConsentValues_itUploadsOnlyAuthorizedLogs() throws {
-        let consentProvider = ConsentProvider(initialConsent: .pending)
```
ncreated, line 730: I'm removing this tests, because in V2 tracking consent will be fully managed by the Core module and transparent to Feature modules. We already have it covered in [`FeatureStorage` tests](https://github.com/DataDog/dd-sdk-ios/blob/e2a961ddc9439e523cf234829bd9a27a4d79451d/Tests/DatadogTests/Datadog/Core/FeatureTests.swift#L21).

I'm doing it now, because since this PR we're no longer running full Core backend in part of Logger tests. Both `FeatureStorage` and `FeatureUpload` are mocked, so we don't have their consent logic available.

Tests/DatadogTests/Datadog/LoggerTests.swift
```diff
@@ -466,55 +483,13 @@ class LoggerTests: XCTestCase {
         logMatchers[2].assertTags(equal: ["env:tests"])
     }
 
-    // MARK: - Sending logs with different network and battery conditions
-
-    func testGivenBadBatteryConditions_itDoesNotTryToSendLogs() throws {
```
ncreated, line 471: I'm removing these 2 tests, because we're no longer running full Core backend in Logger tests and the behaviour of conditional uploads is part of the Core. This needs to be tested separately as `FeatureUpload` tests. I plan to add such tests after migrating both Tracing and RUM in similar fashion.

Tests/DatadogTests/Datadog/Mocks/CoreMocks.swift
```diff
@@ -610,6 +609,73 @@ class NoOpFileReader: SyncReader {
...
+        queue.sync {}
+    }
+
+    // MARK: - Retrieving Recorded Data
+
+    private var waitAndReturnEventsDataExpectation: XCTestExpectation?
+
+    /// Waits until given number of events is written and returns data for these events.
+    /// Passing no `timeout` will result with picking the recommended timeout for unit tests.
+    func waitAndReturnEventsData(count: UInt, timeout: TimeInterval? = nil, file: StaticString = #file, line: UInt = #line) -> [Data] {
```
ncreated, line 641: This is copied from [`DataUploadWorkerMock`](https://github.com/DataDog/dd-sdk-ios/blob/e2a961ddc9439e523cf234829bd9a27a4d79451d/Tests/DatadogTests/Datadog/Mocks/DataUploadWorkerMock.swift#L82-L119) as `InMemoryWriter` is aimed to replace it entirely after Tracing and RUM are migrated in next PRs.

Tests/DatadogTests/Datadog/Mocks/DatadogCoreMock.swift
```diff
@@ -9,6 +9,11 @@ import Foundation
 
 internal final class DatadogCoreMock: Flushable {
     private var v1Features: [String: Any] = [:]
+    private var v1Context: DatadogV1Context?
```
maxep, line 12: Can this be `internal` so we can mutate the context during a test case? it could then be reset to nil or `.mockAny()` on `flush`, WDYT?

Tests/DatadogTests/Datadog/LoggerTests.swift
```diff
@@ -9,39 +9,36 @@ import XCTest
 
 // swiftlint:disable multiline_arguments_brackets
 class LoggerTests: XCTestCase {
-    let core = DatadogCoreMock()
```
maxep, line 12: If we make the context mutable, we could keep the `core` variable and flush everything on `tearDown`.

maxep (approved):
Awesome!! The benefits in tests are already significant and this `interceptedStorage` with `InMemoryWriter` is perfect 👌

Tests/DatadogTests/Datadog/Mocks/DatadogCoreMock.swift
ncreated: Good idea 👌, it helps a lot in some repetitive tests like `LoggerTests` where we need to dance with the `core` each time. 

Changed ✅, however, I left the `init(v1Context:)` too, as in some other tests the inline initialisation comes handy.

PR #885, opened 2022-06-07
RUMM-2197 Update RUM Models to use `device` and `os` info

ncreated:
### What and why?

📦 This PR updates RUM models to https://github.com/DataDog/rum-events-format/commit/d5a943ad55117da57d0b87f9af417ccbcbd49daf so we can set `device` and `os` information.

This PR only introduces new fields that will be set in the next PR.

### How?

* Upgraded RUM models to https://github.com/DataDog/rum-events-format/commit/d5a943ad55117da57d0b87f9af417ccbcbd49daf.
* I enhanced the patch from https://github.com/DataDog/dd-sdk-ios/pull/801 to be more strict and only applied for `action.id`.
* I did smaller adjustments to the RUM models generator to make model definitions looks good in public APIs.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/JSON/JSONSchemaToJSONTypeTransformer.swift
+++ b/tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/JSON/JSONSchemaToJSONTypeTransformer.swift
@@ -7,7 +7,7 @@
 import Foundation
 
 /// Transforms ambiguous `JSONSchema` into type-safe `JSONObject` schema.
-internal class JSONSchemaToJSONTypeTransformer {
+internal class JSONSchemaToJSONTypeTransformer: TypeTransformer<(name: String, schema: JSONSchema)> {
     struct Defaults {
         /// Properties are not required by default.
         static let isRequired = false
@@ -16,7 +16,11 @@ internal class JSONSchemaToJSONTypeTransformer {
     }
 
     func transform(jsonSchemas: [JSONSchema]) throws -> [JSONObject] {
-        return try jsonSchemas.map { try transform(jsonSchema: $0) }
+        precondition(context.current == nil)
+        let transformed = try jsonSchemas.map { try transform(jsonSchema: $0) }
+        precondition(context.current == nil)
+
+        return transformed
     }
 
     private func transform(jsonSchema: JSONSchema) throws -> JSONObject {
@@ -33,12 +37,17 @@ internal class JSONSchemaToJSONTypeTransformer {
     // MARK: - Transforming ambiguous types
 
     private func transformSchemaToAnyType(_ schema: JSONSchema, named name: String) throws -> JSONType {
-        // RUMM-2022: Pick first schema of OneOf to workaround change introduced in
+        context.enter((name, schema))
+        defer { context.leave() }
+
+        // RUMM-2022: Pick first schema of OneOf to workaround `action.id` change introduced in
         // https://github.com/DataDog/rum-events-format/pull/57
         //
         // Supporting multiple types for single property need to be addressed
-        if let oneOf = schema.oneOf?.first {
-            return try transformSchemaToAnyType(oneOf, named: name)
+        if context.parent?.name == "action" && context.current?.name == "id" {
+            if let oneOf = schema.oneOf?.first {
+                return try transformSchemaToAnyType(oneOf, named: name)
+            }
         }
 
         let schemaType = try schema.type
--- a/tools/rum-models-generator/Sources/RUMModelsGeneratorCore/RUM/RUMSwiftTypeTransformer.swift
+++ b/tools/rum-models-generator/Sources/RUMModelsGeneratorCore/RUM/RUMSwiftTypeTransformer.swift
@@ -10,7 +10,16 @@ import Foundation
 internal class RUMSwiftTypeTransformer: TypeTransformer<SwiftType> {
     /// Types which will shared between all input `types`. Sharing means detaching those types from nested declaration
     /// and putting them at the root level of the resultant `types` array, so the type can be printed without being nested.
-    private let sharedTypeNames = ["RUMConnectivity", "RUMUser", "RUMMethod", "RUMEventAttributes", "RUMCITest"]
+    private let sharedTypeNames = [
+        "RUMConnectivity",
+        "RUMUser",
+        "RUMMethod",
+        "RUMEventAttributes",
+        "RUMCITest",
+        "RUMDevice",
+        "RUMOS",
+    ]
+
     /// `RUMDataModel` protocol, implemented by all RUM models.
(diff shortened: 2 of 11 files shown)


tools/rum-models-generator/Sources/RUMModelsGeneratorCore/RUM/RUMSwiftTypeTransformer.swift
```diff
@@ -10,7 +10,16 @@ import Foundation
...
     /// and putting them at the root level of the resultant `types` array, so the type can be printed without being nested.
-    private let sharedTypeNames = ["RUMConnectivity", "RUMUser", "RUMMethod", "RUMEventAttributes", "RUMCITest"]
+    private let sharedTypeNames = [
+        "RUMConnectivity",
+        "RUMUser",
+        "RUMMethod",
+        "RUMEventAttributes",
+        "RUMCITest",
+        "RUMDevice",
+        "RUMOS",
```
ncreated, line 20: I'm adding these two, to not repeat `Device` and `OS` in each event. Instead of `RUMViewEvent.Device`, `RUMErrorEvent.Device`, `RUMActionEvent.Device`, ... we will now have `RUMDevice` shared by event models.

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/RUM/RUMSwiftTypeTransformer.swift
```diff
@@ -152,7 +161,8 @@ internal class RUMSwiftTypeTransformer: TypeTransformer<SwiftType> {
 
         // If the type name collides with Swift `Type` keyword, prefix it with parent type name.
         if fixedName == "Type" {
-            let parentTypeName = (context.parent as? SwiftStruct)?.name ?? ""
+            let parentStruct = context.predecessor(matching: { $0 is SwiftStruct }) as? SwiftStruct
+            let parentTypeName = parentStruct?.name ?? ""
```
ncreated, line 165: This is to fix the problem with [`[Frustration.Type]`](https://github.com/DataDog/rum-events-format/blob/d5a943ad55117da57d0b87f9af417ccbcbd49daf/schemas/rum/action-schema.json#L85) where `Type` cannot be an `enum` name (it conflicts with `Swift.Type`). This transformer already implements the logic of prefixing `Type` with parent `struct's` name, but in case of `[Frustration.Type]` the parent is an array. To fix this, I added `predecessor(matching:)` method which finds the transitive parent matching given predicate.

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/RUM/RUMSwiftTypeTransformer.swift
```diff
@@ -176,6 +186,14 @@ internal class RUMSwiftTypeTransformer: TypeTransformer<SwiftType> {
             fixedName = "RUMCITest"
         }
 
+        if fixedName == "Device" {
+            fixedName = "RUMDevice"
+        }
+
+        if fixedName == "OS" {
+            fixedName = "RUMOS"
+        }
```
ncreated, line 195: Not sure on this new name - `RUMOS` looks odd. I was considering `RUMOperatingSystem` as an alternative. Happy to hear some other ideas 🙌.

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/JSON/JSONSchemaToJSONTypeTransformer.swift
```diff
@@ -33,12 +37,17 @@ internal class JSONSchemaToJSONTypeTransformer {
...
         // https://github.com/DataDog/rum-events-format/pull/57
         //
         // Supporting multiple types for single property need to be addressed
-        if let oneOf = schema.oneOf?.first {
-            return try transformSchemaToAnyType(oneOf, named: name)
+        if context.parent?.name == "action" && context.current?.name == "id" {
+            if let oneOf = schema.oneOf?.first {
+                return try transformSchemaToAnyType(oneOf, named: name)
+            }
         }
```
ncreated, line 51: Here, I make the `action.id` patch to be more strict and only applied where we know it should be applied. This is to avoid overlooking some other changes that might went through unnoticed.

ncreated, line 51: Here, I make the `action.id` patch to be more strict and only applied where we know it should be applied. This is to avoid overlooking some other changes that might went through unnoticed.

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/RUM/RUMSwiftTypeTransformer.swift
maxep: `RUMOperatingSystem` looks better 😄

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/JSON/JSONSchemaToJSONTypeTransformer.swift
```diff
@@ -7,7 +7,7 @@
 import Foundation
 
 /// Transforms ambiguous `JSONSchema` into type-safe `JSONObject` schema.
-internal class JSONSchemaToJSONTypeTransformer {
+internal class JSONSchemaToJSONTypeTransformer: TypeTransformer<(name: String, schema: JSONSchema)> {
```
maxep, line 10: Now that we inherit from `TypeTransformer`, we should implement [this method](https://github.com/DataDog/dd-sdk-ios/blob/develop/tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Output/Preparing/TypeTransformer.swift#L29) as well.

Since the `context` is only used internally, it might be easier to just add a property `let context: TransformationContext<(name: String, schema: JSONSchema)>`. And if this property is `private`, we could get rid of the `precondition`! WDYT?

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/RUM/RUMSwiftTypeTransformer.swift
ncreated: 👌🙃✅

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/JSON/JSONSchemaToJSONTypeTransformer.swift
maxep: We can definitely keep it 👍  I was referencing to the fact that by using a `private` context, we manage the cohesion within the class: the context cannot be altered externally.

maxep (approved):
LGTM 👍

PR #886, opened 2022-06-09
RUMM-2171 Use core feature scope in Tracer 

maxep:
### What and why?

`Tracer` instances now build and record events using `V1FeatureScope` provided by `DatadogV1CoreProtocol`.

### How?

The `Tracer` now keeps a reference to the provided `core` to retrieve the scope when a `DDSpan` needs to be recorded. The `SpanEventBuilder` is created using the context provided with the scope, the event is then written to the `writer` provided by core.

### Testing

This PR introduce a `PassthroughCoreMock` which always provide feature scopes without having to register any features. This allow easy testing of event recording.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


Tests/DatadogTests/Datadog/Mocks/DatadogInternal/DatadogCoreMock.swift
```diff
@@ -5,15 +5,16 @@
...
 
 import Foundation
+
 @testable import Datadog
 
 internal final class DatadogCoreMock: Flushable {
     private var v1Features: [String: Any] = [:]
 
-    var v1Context: DatadogV1Context?
+    var context: DatadogV1Context?
```
maxep, line 14: We don't need two variables definitions in this mock.

Sources/Datadog/Tracer.swift
```diff
@@ -226,7 +220,7 @@ public class Tracer: OTTracer {
             tracer: self,
             context: spanContext,
             operationName: operationName,
-            startTime: startTime ?? dateProvider.currentDate(),
+            startTime: startTime ?? core.v1.context?.dateProvider.currentDate() ?? Date(),
```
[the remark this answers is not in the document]
ncreated: We should be able to easily get non-optional context at this point. The `core.v1.context` should be captured at `Tracer.init`. If it resolves to `nil` it means that the core is not set, so the user makes a mistake of creating `Tracer` without initialising the SDK first (should be alerted with printing configuration error). Otherwise, the context will resolve to a value and we will be able to get rid of this implicit unwrapping here.

Sources/Datadog/Tracer.swift
```diff
@@ -45,20 +45,25 @@ public struct DDTags {
...
-    private let dateProvider: DateProvider
     private let tracingUUIDGenerator: TracingUUIDGenerator
 
+    /// The service name that will appear in traces (if not provided or `nil`, the SDK default `serviceName` will be used).
+    internal let serviceName: String?
+    /// Enriches traces with network connection info.
+    /// This means: reachability status, connection type, mobile carrier name and many more will be added to every span and span logs.
+    /// For full list of network info attributes see `NetworkConnectionInfo` and `CarrierInfo`.
+    /// - Parameter enabled: `false` by default
+    internal let sendNetworkInfo: Bool
```
ncreated, line 66: All these (`spanEventMapper`, `serviceName` and `sendNetworkInfo`) are passed through `tracerConfiguration`. How about capturing the entire immutable `tracerConfiguration` on a property instead of copying its fields into `Tracer`? The main benefit would be maintainability - adding a new field will require a change only in a single place. This can be even a convention to both `Logger`, `Tracer` and `RUMMonitor`, making things easier - both in code, tests and mocks.

Tests/DatadogTests/Datadog/Mocks/DatadogInternal/PassthroughCoreMock.swift
```diff
@@ -0,0 +1,84 @@
...
+
+    /// Execute `block` with the current context and a `writer` to record events.
+    ///
+    /// - Parameter block: The block to execute.
+    func eventWriteContext(_ block: (DatadogV1Context, Writer) throws -> Void) {
+        guard let context = context else {
+            return XCTFail("PassthroughCoreMock missing context")
+        }
+
+        try? block(context, self)
```
ncreated, line 73: Ignoring errors will make debugging harder in cases where the actual context raises an error for some reason. How about failing the test in a verbose way?
```swift
do {
   try block(context, self)
} catch let error {
   XCTFail("Encountered an error when executing `eventWriteContext`: \(error)")
}
```

ncreated (approved):
Looks good 👌👍. I left few enhancements suggestions.

Tests/DatadogTests/Datadog/Mocks/DatadogInternal/PassthroughCoreMock.swift
maxep: Good catch!

PR #888, opened 2022-06-09
RUMM-2197 Improve `device` and `os` info in RUM

ncreated:
### What and why?

📦 This PR updates the SDK to use new `device` and `os` attributes [added recently to RUM schema](https://github.com/DataDog/rum-events-format/pull/62).

With this change, `device` and `os` will no longer be inferred by backend from SDK UA header. Instead, BE can now read these values directly from RUM events.

As a user-facing change, we now enhance the `device.model` information. Previously it was either `iPhone` | `iPad` | `iPod` with no further context. Now it names the model along with its version, e.g. `iPhone13,2` or `iPad13,4`.

### How?

New spec for `device` and `os`:
* [`device.name`](https://github.com/DataDog/rum-events-format/blob/8b32d51779b83d2c7c1c3e41a68c8ff7a452594b/schemas/rum/_common-schema.json#L278-L282) can be `iPhone` | `iPad` | `Apple TV` | ... (same as before),
* [`device.model`](https://github.com/DataDog/rum-events-format/blob/8b32d51779b83d2c7c1c3e41a68c8ff7a452594b/schemas/rum/_common-schema.json#L283-L287) now includes version, e.g. `AppleTV11,1` (can be a breaking change in some filters and dashboards ⚠️),
* [`device.brand`](https://github.com/DataDog/rum-events-format/blob/8b32d51779b83d2c7c1c3e41a68c8ff7a452594b/schemas/rum/_common-schema.json#L288-L292) is always `Apple`,
* [`os.name`, `os.version` and `os.version_major`](https://github.com/DataDog/rum-events-format/blob/8b32d51779b83d2c7c1c3e41a68c8ff7a452594b/schemas/rum/_common-schema.json#L240-L254) remain unchanged - their values will be just like before.

I used a simplified version of [CwlUtils/CwlSysctl.swift](https://github.com/mattgallagher/CwlUtils/blob/master/Sources/CwlUtils/CwlSysctl.swift) to read device information from system attributes (`HW_MACHINE`). Nice discussion on this approach and its `macOS` vs `iOS` nuances can be found [here](https://www.cocoawithlove.com/blog/2016/03/08/swift-wrapper-for-sysctl.html).

Everything happens in `MobileDevice` class, which isn't perfect design, but seems pragmatic choice considering that we will make it more straightforward in with V2's SDK context.

I added 2 layers of tests:
* unit tests assert if the values are set,
* integration tests apply additional validation for all RUM events (through `RUMSessionMatcher`) to ensure that values are correct for current platform (e.g. iOS Simulator).

### Tests

I tested all values on 4 simulator and 2 real devices:

#### iPhone Simulator:
- brand: Apple
- name: iPhone
- model: iPhone13,2 Simulator
- OS name: iOS
- OS version: 15.2

#### iPod touch Simulator:
- brand: Apple
- name: iPhone (which looks like a bug 🐞 in [`UIDevice.model`](https://developer.apple.com/documentation/uikit/uidevice/1620044-model) for iPod Simulator)
- model: iPod9,1 Simulator
- OS name: iOS
- OS version: 15.2

#### iPad Simulator:
- brand: Apple
- name: iPad
- model: iPad13,5 Simulator
- OS name: iPadOS
- OS version: 15.2

#### Apple TV Simulator:
- brand: Apple
- name: Apple TV
- model: AppleTV11,1
- OS name: tvOS
- OS version: 15.2

#### Real iPhone:
- brand: Apple
- name: iPhone
- model: iPhone10,6
- OS name: iOS
- OS version: 15.2

#### Real iPad:
- brand: Apple
- name: iPad
- model: iPad13,4
- OS name: iPadOS
- OS version: 15.3.1

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [x] Run smoke tests


Sources/Datadog/Core/System/MobileDevice.swift
```diff
@@ -107,22 +133,48 @@ internal class MobileDevice {
...
+        model = "\(simulatorModel) Simulator"
+        #endif
+
+        self.init(
+            model: model,
+            uiDevice: device,
+            processInfo: processInfo,
+            notificationCenter: .default
+        )
+    }
```
ncreated, line 176: These new initialisers in `MobileDevice` are added for init consistency on both `iOS` and `tvOS` (both define matching `init()`). This and the entire `MobileDevice` class should be however a subject to change in V2. Since we added `tvOS` support, it should no longer be "mobile device" and we should really split its responsibilities. For example, the cohesion of this class is very low - battery reasoning and management is only implemented for `iOS` and we use mock values for `tvOS` just to "make it work".

Sources/Datadog/Core/Upload/RequestBuilder.swift
```diff
@@ -67,7 +67,7 @@ internal struct RequestBuilder {
 
             return HTTPHeader(
                 field: userAgentHeaderField,
-                value: .constant("\(sanitizedAppName)/\(appVersion) CFNetwork (\(device.model); \(device.osName)/\(device.osVersion))")
+                value: .constant("\(sanitizedAppName)/\(appVersion) CFNetwork (\(device.name); \(device.osName)/\(device.osVersion))")
```
ncreated, line 70: These kind of changes in this PR are OK. What used to work as `device.model` before, now is called `device.name` (`iPhone` | `iPad` | ...). The `device.model` is now reserved for exact info, e.g. `iPhone10,2`.

Sources/Datadog/FeaturesIntegration/CrashReporting/CrashReportingWithRUMIntegration.swift
```diff
@@ -358,8 +362,12 @@ internal struct CrashReportingWithRUMIntegration: CrashReportingIntegration {
...
             context: nil,
             date: startDate.timeIntervalSince1970.toInt64Milliseconds,
-            device: nil, // TODO: RUMM-2197 add device and OS info
-            os: nil, // TODO: RUMM-2197 add device and OS info
+            device: .init(from: deviceInfoProvider, telemetry: nil),
+            // RUMM-2197: In very rare cases, the OS info computed below might not be exactly the one
+            // that the app crashed on. This would correspond to a scenario when the device OS was upgraded
+            // before restarting the app after crash. To solve this, the OS information would have to be
+            // persisted in `crashContext` the same way as we do for other dynamic information.
+            os: .init(from: deviceInfoProvider),
```
ncreated, line 370: I haven't decided for this implementation as it will only address a very rare case with quite a work required. It is however a good driver for more robust SDK context management in V2.

Sources/Datadog/Core/System/MobileDevice.swift
maxep: Sounds good!

maxep (approved):
Looks great! 👍 

Once this and v2 feature scope are merged, we should apply this to the `feature/v2` and its context, so we keep v2 in sync.

PR #902, opened 2022-06-20
RUMM-2171 Use core feature scope in RUM

maxep:
### What and why?

`RUMMonitor` and scope instances now build and record events using `V1FeatureScope` provided by `DatadogV1CoreProtocol`.

### How?

The `RUMMonitor` now keeps a reference to the provided `core` to retrieve the scope when a `RUMCommand` needs to be processed by `RUMScope`s. The `RUMScope` now expect a `DatadogV1Context` and a `Writer` to record events and apply side-effects.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
```diff
@@ -127,44 +119,44 @@ internal class RUMResourceScope: RUMScope {
...
                 .init(id: rumUUID.toRUMDataFormat)
             },
-            application: .init(id: context.rumApplicationID),
+            application: .init(id: self.context.rumApplicationID),
             ciTest: dependencies.ciTest,
-            connectivity: dependencies.connectivityInfoProvider.current,
+            connectivity: .init(context: context),
             context: .init(contextInfo: attributes),
-            date: dateCorrection.applying(to: resourceStartTime).timeIntervalSince1970.toInt64Milliseconds,
+            date: resourceStartTime.timeIntervalSince1970.toInt64Milliseconds,
```
ncreated, line 136: The `resourceStartTime` will not include NTP correction if it's taken from [`URLSessionTaskTransactionMetrics`](https://developer.apple.com/documentation/foundation/urlsessiontasktransactionmetrics) (it means auto-instrumented resources):
```swift
/// Metrics values take precedence over other values.
if let metrics = resourceMetrics {
   resourceStartTime = metrics.fetch.start
   resourceDuration = metrics.fetch.end.timeIntervalSince(metrics.fetch.start)
   size = metrics.responseSize ?? command.size
}
```
Hence, using `resourceStartTime` without `dateCorrection.applying(to:)` will lead to skewing "resource to view" temporal relationship (🐞) and eventually even starting a resource before the view is started. Mind that RUM uses nanoseconds precision and NTP correction is typically milliseconds.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -43,8 +43,6 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
     let viewName: String
     /// The start time of this View.
     let viewStartTime: Date
-    /// Date correction to server time.
-    private let dateCorrection: DateCorrection
```
ncreated, line 47: I think this property wasn't enough described in this inline comment. We cannot simply remove local `dateCorrection` from view scope and use its global value from SDK context like proposed. This will lead to skewing events in RUM (🐞).

Note that the result of NTP sync done with Kronos arrives over time. When the app starts, we perform the sync and it takes up to several seconds until the final result is received. Meantime the SDK receives intermediate results - each next with better precision. Because `DateCorrection` abstracts the Kronos sync, its values cannot be considered stable or monotonic.

The idea of capturing `dateCorrection` in view scope's property is:
* to "freeze" the (NTP) date correction value when the view scope is created;
* to apply the same, "frozen" correction to all events sent from view scope;
* to pass the same ("frozen") date correction value to all child scopes (e.g. resource scope) for applying it to events sent from nested scopes.

If we don't "freeze" date correction and instead always use the one from SDK context, it means we will be applying different deltas to view scope events and its child scope events, leading to noticeable skews.

Sources/Datadog/RUM/RUMMonitor/RUMScope.swift
```diff
@@ -10,30 +10,26 @@ internal protocol RUMScope: AnyObject {
...
     ///
-    /// Returns the `childScope` requested to be kept open, `nil` if it requests to close.
-    func manage<S: RUMScope>(childScope: S?, byPropagatingCommand command: RUMCommand) -> S? {
-        if childScope?.process(command: command) == false {
-            return nil
-        } else {
-            return childScope
-        }
+    /// Returns `self`  to be kept open, `nil` if it requests to close.
+    func scope(byPropagating command: RUMCommand, context: DatadogV1Context, writer: Writer) -> Self? {
```
maxep, line 20: I've simplified the propagation methods signatures to better reflect its behaviour.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMScopeDependencies.swift
```diff
@@ -25,19 +25,11 @@ internal struct RUMScopeDependencies {
...
-    let source: String
     let firstPartyURLsFilter: FirstPartyURLsFilter
     let eventBuilder: RUMEventBuilder
     let eventOutput: RUMEventOutput
     let rumUUIDGenerator: RUMUUIDGenerator
-    /// Adjusts RUM events time (device time) to server time.
-    let dateCorrector: DateCorrectorType
     /// Integration with Crash Reporting. It updates the crash context with RUM info.
     /// `nil` if Crash Reporting feature is not enabled.
     let crashContextIntegration: RUMWithCrashContextIntegration?
```
ncreated, line 35: Nice to see it shrink 💪. I think that ultimately we should only keep RUM-specific properties in there and everything else should be passed through `SDKContext`, WDYT @maxep? Some of it is even possible now - e.g. I can see `sdkInitDate` and `MobileDevice` (transitive input for `RUMDevice` and `RUMOperatingSystem`) already defined in `DatadogV1Context`. If we agree, I think it's best to tackle them in separate PR.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -462,20 +453,21 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
...
             ),
-            action: context.activeUserActionID.flatMap { rumUUID in
+            action: self.context.activeUserActionID.map { rumUUID in
                 .init(id: rumUUID.toRUMDataFormat)
             },
-            application: .init(id: context.rumApplicationID),
+            application: .init(id: self.context.rumApplicationID),
             ciTest: dependencies.ciTest,
-            connectivity: dependencies.connectivityInfoProvider.current,
+            connectivity: .init(context: context),
```
ncreated, line 470: Nice 👍. I think we should do the same for initializing other RUM data types from `context`, e.g.: `RUMOperatingSystem`, `RUMDevice`. Can be a separate PR IMO.

Sources/Datadog/RUMMonitor.swift
```diff
@@ -593,9 +596,16 @@ public class RUMMonitor: DDRUMMonitor, RUMCommandSubscriber {
     // MARK: - RUMCommandSubscriber
 
     func process(command: RUMCommand) {
+        guard let scope = core.v1.scope(for: RUMFeature.self) else {
+            return
```
ncreated, line 600: (just checking) This `return` is unreachable, right? We do early-return in `RUMMonitor.initialize()` when `RUMFeature` is not registered, so it's not possible to run here.

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMViewScopeTests.swift
```diff
@@ -56,10 +66,15 @@ class RUMViewScopeTests: XCTestCase {
...
+            startTime: .mockAny(),
+            dateCorrection: .zero
         )
 
-        _ = scope.process(command: RUMStartUserActionCommand.mockAny())
+        _ = scope.process(
+            command: RUMStartUserActionCommand.mockAny(),
+            context: context,
+            writer: writer
+        )
```
ncreated, line 77: After seeing this all over, I wonder if it wouldn't make sense to anticipate future needs by introducing container type bundling `context` and `writer` in RUMScope APIs. For example:
```swift
scope.process(command:rumScopeContext:)
```
where initially we will have:
```swift
internal struct RUMScopeContext {
   let sdkContext: DatadogV1Context
   let writer: Writer
}
```
I will keep door open for bundling other aspects (if necessary) without huge refactoring pain. WDYT @maxep ?

ncreated (approved):
I can see a huge effort here and it looks awesome 💪🏅.

All is good for me in this PR and we can merge. I left some suggestions trying to anticipate the future - please read all holistically as some might make more sense together.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMScopeDependencies.swift
maxep: Ah indeed, that's a good point! Also makes me realise that I forgot to remove the `eventBuilder` and `eventOutput`.
I've created a dedicated story for managing the context in a single queue, [RUMM-2299](https://datadoghq.atlassian.net/browse/RUMM-2299), I can add that in the requirements!

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
```diff
@@ -190,45 +186,44 @@ internal class RUMResourceScope: RUMScope {
                 type: resourceType,
                 url: resourceURL
             ),
-            service: dependencies.serviceName,
+            service: context.service,
             session: .init(
                 hasReplay: nil,
-                id: context.sessionID.toRUMDataFormat,
+                id: self.context.sessionID.toRUMDataFormat,
                 type: dependencies.ciTest != nil ? .ciTest : .user
```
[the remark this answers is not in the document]
maxep: I agree, it's confusing. I will rename that 👍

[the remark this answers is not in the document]
maxep: Actually `self.context` is defined by `RUMContextProvider` protocol. So changing the definition will have a large impact. I will rename the local variable to `sdkContext` only. `RUMContext` will be merge in Core Context anyway!

Sources/Datadog/RUMMonitor.swift
maxep: Yes, we do an early return in `RUMMonitor.initialize()` with appropriate console prints 👍

PR #905, opened 2022-06-21
RUMM-2233 Fix session management rule

xgouchet:
### What and why?

This ensures that the timeout for sessions take into account the interactive vs non interactive events.


ncreated (approved):
Looks good 👌🚀

PR #906, opened 2022-06-21
RUMM-2203 Use unique data directory for each instance of the SDK

ncreated:
### What and why?

📦 With this PR, each instance of `DatadogCore` will use a different directory for managing its data.

This is a step forward towards V2 architecture, where we're going to support multiple SDK instances. Each instance of the SDK needs to persist data in different locations.

This is the content of `/Library/Caches/` when using two instances of `DatadogCore`:

Screenshot 2022-06-22 at 17 13 47

### How?

Each `DatadogCore` now accepts a "core directory". I introduced a separate `CoreDirectory` type that wraps [`Directory`](https://github.com/DataDog/dd-sdk-ios/blob/ef343b1781218392790bd7085fad4b27628ea44b/Sources/Datadog/Core/Persistence/Files/Directory.swift#L10) - this is to enforce type safety at configuration level and to avoid passing wrong directory by mistake.

The `CoreDirectory` creates the folder structure for SDK and its Features. It uses a hash function (sha256) for computing the folder name that is unique for given instance of the SDK:
```swift
let directoryName = sha256("\(clientToken)\(site)")
```

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/Core/Utils/Cryptography.swift
+++ b/Sources/Datadog/Core/Utils/Cryptography.swift
@@ -1,0 +1,32 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+import CryptoKit
+import CommonCrypto
+
+/// Computes SHA256 for given `string`.
+///
+/// The implementation is portable between iOS versions - it uses `CryptoKit` on iOS13+ and fallbacks
+/// to `CommonCrypto` on older devices.
+internal func sha256(_ string: String) -> String {
+    guard let data = string.data(using: .utf8) else {
+        return string
+    }
+
+    if #available(iOS 13.0, tvOS 13.0, *) { // Compute SHA256 with `CryptoKit`
+        let digest = SHA256.hash(data: data)
+        return digest
+            .map({ byte in String(format: "%02x", byte) })
+            .joined(separator: "")
+    } else { // Fallback to SHA256 with `CommonCrypto`
+        var digest: [UInt8] = Array(repeating: 0, count: Int(CC_SHA256_DIGEST_LENGTH))
+        _ = data.withUnsafeBytes { CC_SHA256($0.baseAddress, UInt32(data.count), &digest) }
+        return digest
+            .map({ byte in String(format: "%02x", byte) })
+            .joined(separator: "")
+    }
+}
--- a/Sources/Datadog/Core/Utils/Cryptography.swift
+++ b/Sources/Datadog/Core/Utils/Cryptography.swift
@@ -5,28 +5,17 @@
  */
 
 import Foundation
-import CryptoKit
 import CommonCrypto
 
 /// Computes SHA256 for given `string`.
-///
-/// The implementation is portable between iOS versions - it uses `CryptoKit` on iOS13+ and fallbacks
-/// to `CommonCrypto` on older devices.
 internal func sha256(_ string: String) -> String {
     guard let data = string.data(using: .utf8) else {
         return string
     }
 
-    if #available(iOS 13.0, tvOS 13.0, *) { // Compute SHA256 with `CryptoKit`
-        let digest = SHA256.hash(data: data)
-        return digest
-            .map({ byte in String(format: "%02x", byte) })
-            .joined(separator: "")
-    } else { // Fallback to SHA256 with `CommonCrypto`
-        var digest: [UInt8] = Array(repeating: 0, count: Int(CC_SHA256_DIGEST_LENGTH))
-        _ = data.withUnsafeBytes { CC_SHA256($0.baseAddress, UInt32(data.count), &digest) }
-        return digest
-            .map({ byte in String(format: "%02x", byte) })
-            .joined(separator: "")
-    }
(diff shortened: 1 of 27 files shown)


Tests/DatadogTests/Datadog/Mocks/SystemFrameworks/FoundationMocks.swift
```diff
@@ -224,30 +224,76 @@ extension String: AnyMockable, RandomMockable {
...
+        // > initializer succeeds if `v` is a valid Unicode scalar value---that is,
+        // > if `v` is in the range `0...0xD7FF` or `0xE000...0x10FFFF`. If `v` is
+        // > an invalid Unicode scalar value, the result is `nil`.
+        var codePoint: UInt32 = .mockRandom(min: 0x0, max: 0x10FFFF)
+        while codePoint > 0xD7FF && codePoint < 0xE000 {
+            codePoint = .mockRandom(min: 0x0, max: 0x10FFFF)
+        }
+        return Unicode.Scalar(codePoint)! // force-unwrap as we guarantee `randomCodePoint` is correct
     }
 }
```
ncreated, line 298: This is to produce truly random `String` for fuzzy testing. I found this important as the `sha256()` implementation for iOS 11 - 12 uses unsafe pointers, so I wanted to run its test on a wide input.

Sources/Datadog/DatadogInternal/DatadogCoreProtocol.swift
```diff
@@ -19,10 +19,13 @@ internal struct FeatureStorageConfiguration {
...
     struct Directories {
         /// The path for writing authorized data (when tracking consent is granted).
+        /// This path must be relative to the core directory created for given instance of the SDK.
         let authorized: String
         /// The path for writing unauthorized data (when tracking consent is pending).
+        /// This path must be relative to the core directory created for given instance of the SDK.
         let unauthorized: String
         /// The list of deprecated paths from previous versions of this feature. It is used to perform cleanup.
+        /// These paths must be relative to the known OS location (`/Library/Caches/`) containing core directories for different instances of the SDK.
         let deprecated: [String]
```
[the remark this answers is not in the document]
maxep: > a vendor SDK installed in an app where both use Datadog SDK but vendor uses V2 and app is on V1

If both versions mismatch, the app won't be able to install the vendor framework as a dependency, except if the vendor is manually linking the Datadog SDK as a static lib, which should be discouraged IMO. The app would have to upgrade to v2 or use a previous version of the vendor SDK to match the version. So I wouldn't consider this use case.

Now, multiple instances of the v2 SDK will perform the same operation on the `deprecated` v1 directories, so I wonder if this is the right place for v1 depreciation 🤔 We can consider this as a temporary measure and eventually remove this `deprecated` when we consider good adoption for v2 (which can take time..)

Sources/Datadog/Core/Utils/Cryptography.swift
```diff
@@ -0,0 +1,32 @@
...
+        return digest
+            .map({ byte in String(format: "%02x", byte) })
+            .joined(separator: "")
+    } else { // Fallback to SHA256 with `CommonCrypto`
+        var digest: [UInt8] = Array(repeating: 0, count: Int(CC_SHA256_DIGEST_LENGTH))
+        _ = data.withUnsafeBytes { CC_SHA256($0.baseAddress, UInt32(data.count), &digest) }
+        return digest
+            .map({ byte in String(format: "%02x", byte) })
+            .joined(separator: "")
+    }
```
maxep, line 31: why do both? which means linking both frameworks. I believe they both perform the same, the difference is that `CommonCrypto` is a lower-level framework, and more verbose.
`CommonCrypto` is ported to iOS13+, so we don't need `CryptoKit` for that!

Sources/Datadog/Datadog.swift
```diff
@@ -204,7 +204,7 @@ public class Datadog {
 
         // Set default `DatadogCore`:
         let core = DatadogCore(
-            rootDirectory: try Directory.cache(),
+            directory: try CoreDirectory(in: try Directory.cache(), from: configuration.common),
```
maxep, line 207: /nit 
```suggestion
            directory: try CoreDirectory(in: Directory.cache(), from: configuration.common),
```

Sources/Datadog/Core/Utils/Cryptography.swift
ncreated: I was thinking that in few years from now, when dropping iOS 12 support we will do smooth transition to `CryptoKit` as it will be already battle tested. I might be looking to far tho. And the argument on linking 2 framweorks vs 1 is good 👍. Let's stick to `CommonCrypto`.

Sources/Datadog/DatadogInternal/DatadogCoreProtocol.swift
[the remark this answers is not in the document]
ncreated: 👍 Right, I think it's fair to assume there will be a single version of the SDK even if multiple instances are enabled.

Indeed, there's a room for improvement around deprecation logic. Because it requires more refactoring in different parts of the codebase, I created `RUMM-2312` to focus solely on data deprecation logic. For now, I'll keep it as it is - capable of removing global SDK folders, without ensuring the logic is ran only once.

maxep (approved):
Nice work with the hash for identifying core instances 👍 
LGTM ✅

PR #913, opened 2022-06-30
RUMM-2299 Refactor date corrector

maxep:
### What and why?

In the process of having core context on a single queue, we need to have a better separation of concerns between `ServerDateCorrector` (previously named `DateCorrector`)and `NTPServerDateProvider`: by moving the `ValuePublisher` to `ServerDateCorrector` we make `NTPServerDateProvider` reusable on another queue.

The end goal is to hold a `NTPServerDateProvider` and be able to update the v2 Core Context on its queue.

### How?

The NTP logic has been removed from `ServerDateCorrector` and it now holds the atomic server offset.

`DateCorrection` has been removed as it does not add value, neither in logic nor in tests. `ServerDateCorrector` now provides the correction offset as `TimeInterval`.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


Sources/Datadog/Core/System/Time/ServerDateCorrector.swift
```diff
@@ -0,0 +1,48 @@
...
+internal protocol DateCorrector {
+    /// Returns recent date correction for adjusting device time to server time.
+    var offset: TimeInterval { get }
+}
+
+internal class ServerDateCorrector: DateCorrector {
+    /// Server offset publisher.
+    private let publisher: ValuePublisher<TimeInterval?> = ValuePublisher(initialValue: nil)
+
+    init(serverDateProvider: ServerDateProvider) {
```
maxep, line 19: `deviceDateProvider` was only used for the console to print the current time. It's best to remove that dependency.

Tests/DatadogTests/Datadog/Core/System/Time/DateCorrectionTests.swift
```diff
@@ -8,44 +8,30 @@ import XCTest
...
-        var randomlyChosenServers: Set<String> = []
-
-        (0..<100).forEach { _ in
-            _ = DateCorrector(deviceDateProvider: deviceDateProvider, serverDateProvider: serverDateProvider)
-            randomlyChosenServers.insert(serverDateProvider.synchronizedNTPPool!)
-        }
-
-        let allAvailableServers = Set(DateCorrector.datadogNTPServers)
-        XCTAssertEqual(randomlyChosenServers, allAvailableServers, "Each time Datadog NTP server should be picked randomly.")
-    }
```
maxep, line 38: This is no longer testable as the ntp pools have been moved to `NTPServerDateProvider`. We were only testing `Swift.Array.randomElement()`, so I think it is safe to remove. WDYT?

Sources/Datadog/DatadogInternal/Context/DateCorrector.swift
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+/// Calculates the date correction for adjusting device time to server time.
+internal protocol DateCorrector {
+    /// Returns recent date correction for adjusting device time to server time.
+    var offset: TimeInterval { get }
+}
```
ncreated, line 13: It is no longer a "date corrector" IMO if we make it return `offset`. The (previous) idea of having `DateCorrector` was to return a `DateCorrection` type that makes it impossible to make mistake:
```swift
internal struct DateCorrection {
   // ...

   func applying(to deviceDate: Date) -> Date {
      return deviceDate.addingTimeInterval(serverTimeOffset)
   }
}
```
Now if we want to work with `offset`, an instant question is _"should we add it or subtract it?"_. This IMO sounds very error-prone. I'd prefer having type-safety on date correction, instead of passing arbitrary `TimeInterval` around, like in RUM scopes:

```diff
- date: dateCorrection.applying(to: resourceStartTime).timeIntervalSince1970.toInt64Milliseconds,
+ date: resourceStartTime.addingTimeInterval(serverTimeOffset).timeIntervalSince1970.toInt64Milliseconds,
```

@maxep What is the motivation behind dropping `DateCorrection` 🤔💭?

Sources/Datadog/Core/System/Time/ServerDateCorrector.swift
ncreated: Sounds good 👍. So the V2 Logging feature will use its own device time provider for console logs and leverage NTP time for remote logging, right?

Sources/Datadog/DatadogInternal/Context/DateCorrector.swift
maxep: Arrg my previous comment didn't go through! sorry about that:

This is arguable 🙂 I don't see the value of this actually, there is no sign manipulation. IMO it is more error-prone to not apply the offset than to wrongly apply it by inverting the sign, it is also super easy to use the wrong reference date (`timeIntervalSince1970` vs. `timeIntervalSinceReferenceDate`) or to use the wrong conversion (`toInt64Milliseconds` vs. `toInt64Nanoseconds`). So even now, we would always rely on either doc (which I should add) or previous usage to build the correct timestamp for an event.

The motivation is to reduce the number of types, we have many custom types for date manipulation that it gets confusing. I would definitely agree if this type was public and used externally, but this is for internal usage.

Now, we could have a `Date` extension for internal usage. By leveraging [this pattern](https://github.com/Alamofire/Alamofire/blob/master/Source/AlamofireExtended.swift), we could do something like:
```swift
// Date extension method:
func timestamp(byApplying offset: TimeInterval) -> Int64 {
    return self.addingTimeInterval(serverTimeOffset).timeIntervalSince1970.toInt64Milliseconds
}

date.dd.timestamp(byApplying: offset)
```

IMO, the properties need to be documented, either if it's a `DateCorrection` or a `TimeInterval`. In both cases, it requires the same effort to build the timestamp.

Sources/Datadog/Core/System/Time/ServerDateCorrector.swift
maxep: No, I mean the previous `DateCorrector` was expecting a `DeviceDateProvider` at init, see [here](https://github.com/DataDog/dd-sdk-ios/pull/913/files#diff-69ca19ff72791841d0de0246d6cc7ed0551e2ca932d142cc4a0ca84374cdd82bL36-L38)

But this parameter was only used to print:
```
DD.logger.error(
    """
    NTP time synchronization failed.
    Device time will be used for signing events (current device time is \(deviceTime)).
    """
)
```

Sources/Datadog/DatadogInternal/Context/DateCorrector.swift
ncreated: Indeed there are other transformations we do (e.g. `timeIntervalSince1970` and `toInt64Milliseconds`) worth solidifying - I didn't look that far. Single extension is debatable as we use different units in RUM (milliseconds) and Tracing (nanoseconds). Maybe a separate extension in each Feature?

Anyway, I'm ok with `serverTimeOffset` to move forward and let's keep this further refactoring in backlog.

Sources/Datadog/Core/System/Time/ServerDateCorrector.swift
ncreated: Ah, got it 👍. Looks good, printing device time is not a big deal 👌 (it will appear in `DD.logger.error()` anyway).

Tests/DatadogTests/Datadog/Core/System/Time/DateCorrectionTests.swift
ncreated: Well, I disagree that it's not worth testing which pool we synchronise too (we need it to be DD pool for new skews in Distributed Tracing), but I agree that asserting on `DateCorrector.datadogNTPServers` didn't make much sense as it would be still green even after introducing a bug in the configuration (e.g. a typo).

Would be ideal to improve this test, not to remove it, WDYT @maxep?

Sources/Datadog/Core/System/Time/ServerDateCorrector.swift
```diff
@@ -0,0 +1,48 @@
...
+                        Server time will be used for signing events (\(difference)s difference with device time).
+                        """
+                    )
+                } else {
+                    DD.logger.error(
+                        """
+                        NTP time synchronization failed.
+                        Device time will be used for signing events.
+                        """
+                    )
```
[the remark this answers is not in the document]
maxep: We can indeed move forward as is 👍  My initial goal was to move the entire NPN logic to `NTPServerDateProvider` and to have an even simpler `ServerDateProvider` protocol:
```swift
protocol ServerDateProvider {
    func synchronize(update: @escaping (TimeInterval) -> Void)
}
```
That would simplify the up-comping `DatadogContextProvider`, but it is not a blocker and I prefer keeping the test as well.

Sources/Datadog/DatadogInternal/Context/DateCorrector.swift
maxep: > RUM (milliseconds) and Tracing (nanoseconds)

I didn't even noticed 😄

Tests/DatadogTests/Datadog/Core/System/Time/DateCorrectionTests.swift
maxep: Yes, let me see what I can do 👍

Sources/Datadog/Core/System/Time/ServerDateCorrector.swift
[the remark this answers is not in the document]
maxep: I was able to solve this by introducing a `KronosClockProtocol` as mentioned below 👍

ncreated (approved):
Looks good 👌🙂

PR #916, opened 2022-07-04
Maxep/rumm 2299/core context provider

maxep:
⚠️ Needs rebase on #913 

### What and why?

Provide Datadog Context on single queue.

### How?

Introduce the `DatadogContextProvider` that implement thread-safe methods read/write to the context.

```swift
// concurrent read
provider.read { context in
    // read value from the current context
}

// barrier block for write operation
provider.write { context in
    // set mutable values of the context
}
```

The provider performs reads concurrently but uses barrier block for write operations, same behaviour as `ValuePublisher`.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


maxep:
@ncreated sorry for the large PR.

It's not finalised yet, but I would like to first discuss the `ContextValuePublisher` and `ContextValueReader` that I'm introducing here. As we discussed, the goal is to be able to listen to value change or to pull values for the core context, and sometime do both for a single value when we need to support different APIs for different OS versions.

The main goals are:
1. Have generic constraints on publisher and reader so we can abstract multiple implementations and make it easy for testing.
3. Unify the way we apply mutation to the context.
4. Make `DatadogContextProvider` the single owner of the queue.

Sources/Datadog/Kronos/KronosClock.swift
```diff
@@ -125,7 +125,11 @@ internal final class KronosClock: KronosClockProtocol {
     ) {
         self.loadFromDefaults()
 
-        KronosNTPClient().query(pool: pool, numberOfSamples: samples) { offset, done, total in
+        KronosNTPClient().query(pool: pool, numberOfSamples: samples) { [weak self] offset, done, total in
+            guard let self = self else {
+                return
+            }
```
maxep, line 131: Prevent from invoking callbacks if `self` is released.

Sources/Datadog/DatadogCore/Context/DatadogContextProvider.swift
```diff
@@ -0,0 +1,95 @@
...
+            self.networkConnectionInfoReader.read { context.networkConnectionInfo = $0 }
+            self.carrierInfoReader.read { context.carrierInfo = $0 }
+            block(context)
+        }
+    }
+
+    /// Writes to the `context` asynchronously, without blocking the caller thread.
+    func write(block: @escaping (inout DatadogContext) -> Void) {
+        queue.async(flags: .barrier) { block(&self.context) }
+    }
```
ncreated, line 94: I cannot foresee the usage yet 🤔💭. Will `read(block:)` be used from _"get event write context"_ call? Will `write(block:)` be used when Features inject their portions of _SDK context_?

ncreated (commented):
Looks really good 🏅👍. Having common abstraction around publishing and reading `ContextValue` is definitely what we need! The code looks very well crafted and documented 👍. 

If possible, I'd suggest to not add more abstractions than what we introduce here - feeling we're close to the limits of readability. Let's continue with other providers and see if this scales up 🙌.

Sources/Datadog/Kronos/KronosClock.swift
ncreated: Interestingly, I'm somehow sure it will fix [this unobvious flakiness](https://app.datadoghq.com/ci/test/AQAAAYHXYfyfwZBI4wAAAABBWUhYWWZ5ZkFBQ2lRZkZmYktVNExqaUo?colorBy=service&env=ci&index=citest&spanID=12466522618453451672&spanViewType=metadata).

Sources/Datadog/DatadogCore/Context/NetworkConnectionInfoPublisher.swift
```diff
@@ -0,0 +1,202 @@
...
+        if #available(iOS 12, tvOS 12, *) {
+            self.init(NWPathMonitorPublisher())
+        } else {
+            self.init(SCNetworkReachabilityReader())
+        }
+    }
+
+    func set(queue: DispatchQueue) -> Self {
+        .init(setQueue(queue))
+    }
```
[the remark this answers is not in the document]
maxep: This one is though, and it does not read well I admit.

> the `AnyNetworkConnectionInfoPublisher` now requires the caller to follow a convention of calling set(queue:) in order to make things work as expected. It's lacking type safety (that is typically guaranteed by accepting queue in init).

This issue is that the publisher is instantiated before the context provider queue is created, so we need a way of injecting the queue. I made it so `set(queue:)` is **not required**, and it will work on the `.global(qos: .utility)` queue if no queue is specified.

> Maybe if choosing `final class` over `struct` we can mutate the state in-place and replace queue from `init()`?

The difficulty here is that only `NWPathMonitorPublisher` needs a queue, not `SCNetworkReachabilityReader`, so we need to make it work for both. `AnyNetworkConnectionInfoPublisher` does not mutate itself, but it's underlying `NWPathMonitorPublisher` does.

I will see if I can improve this! Maybe `NWPathMonitorPublisher` can keep working on a the `.global(qos: .utility)` queue and does not need injection 🤔

Sources/Datadog/DatadogCore/Context/DatadogContextProvider.swift
ncreated: Then I wonder what will be pattern of assembling context when we introduce other publishers (_user info_, _tracking consent_, _launch time_, _app state_, ...)?

In general, we have **two kinds** of publishers (in V1 called "providers"):
* **passive** ones, that require **pulling** the value (like _carrier info provider_ for iOS 11);
* **active** ones, that **publish** an update to the value (like _network info provider_ for iOS 12 or _user info provider_).

Does this pattern in `read(block:)`:
```swift
self.networkConnectionInfoReader.read { context.networkConnectionInfo = $0 }
self.carrierInfoReader.read { context.carrierInfo = $0 }
```
express that both _network_ and _carrier info_ publishers are passive, so we need to pull their value? Is this because the kind of these publishers depends on their implementation (iOS 11 / 12+)?

When it comes to active publishers, we won't need to call them in `read(block:)`, because they will update the `context` from `write(block:)` (managed through subscription). Is this correct 🤔💭?

maxep: Yes, that is correct!
As we discussed, I will split reader from publisher for _network_ and _carrier info_, and other that require both behaviours. It will make it clear from `DatadogContextProvider` what is active and what is passive 👍 

I will also update the `ContextValueReader` interface so it differs more from the `ContextValuePublisher`.

Sources/Datadog/DatadogCore/Context/NetworkConnectionInfoPublisher.swift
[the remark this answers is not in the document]
maxep: Ok, so I made the `DatadogContextProvider` agnostic to publisher and reader. We can now easily subscribe to a publisher, or assign a reader to a context value.

Also, publishers can run on different queue that the context's queue, we don't need to inject it anymore.

Sources/Datadog/DatadogCore/Context/ContextValueReader.swift
```diff
@@ -0,0 +1,129 @@
...
+    ///   - keyPath: The value's writable `keyPath`.
+    mutating func append<Reader>(reader: Reader, receiver keyPath: WritableKeyPath<Value, Reader.Value>) where Reader: ContextValueReader {
+        let previousRead = readBlock
+        // perform recursive read
+        readBlock = { value in
+            // first perform the previous read operation
+            previousRead(&value)
+            // then read the property
+            reader.read(to: &value[keyPath: keyPath])
+        }
```
maxep, line 79: Here, I'm using recursion to chain reads for value properties. Another option would be to create another `ContextValueReader` concrete type that keeps an array of read operations, such as:
```swift
internal struct KeyPathContextValueReader<Value>: ContextValueReader {
    let initialValue: Value

    private var reads: [(inout Value) -> Void] = []

    init(initialValue: Value) {
        self.initialValue = initialValue
    }

    mutating func append<Reader>(reader: Reader, receiver keyPath: WritableKeyPath<Value, Reader.Value>) where Reader: ContextValueReader {
        reads.append({ value in
            reader.read(to: &value[keyPath: keyPath])
        })
    }

    func read(to receiver: inout Value) {
        reads.forEach { read in
            read(&receiver)
        }
    }
}
```

maxep: I went for the `KeyPathContextValueReader`.

Sources/Datadog/Core/System/Time/ServerDateCorrector.swift
```diff
@@ -10,8 +10,10 @@ import Foundation
 internal class ServerDateCorrector: DateCorrector {
     /// Server offset publisher.
     private let publisher: ValuePublisher<TimeInterval> = ValuePublisher(initialValue: 0)
+    private let provider: ServerDateProvider
 
     init(serverDateProvider: ServerDateProvider) {
+        self.provider = serverDateProvider
```
maxep, line 16: Now that Kronos no longer retains `self` in completion, we need to keep a strong ref to the provider.

ncreated (changes requested):
Really well done 👍. Although I'm not a fan of so many abstract layers, what is proposed here for combining the context makes total sense, is logically structured and very natural to Swift - great job 🏅!

I left one feedback that I consider blocking as it leads to TSAN crash.

Tests/DatadogTests/Datadog/DatadogCore/DatadogContextProviderTests.swift
```diff
@@ -0,0 +1,154 @@
...
+
+        _ = provider
+    }
+
+    func testConcurrentReadWrite() {
+        let provider = DatadogContextProvider(context: context)
+
+        let serverOffsetPublisher = ContextValuePublisherMock<TimeInterval>(initialValue: 0)
+        let networkConnectionInfoPublisher = ContextValuePublisherMock<NetworkConnectionInfo?>()
+        let carrierInfoPublisher = ContextValuePublisherMock<CarrierInfo?>()
```
ncreated, line 66: This is almost a copy of the above test - isn't this a leftover?

Tests/DatadogTests/Datadog/DatadogCore/KronosClockPublisherTests.swift
```diff
@@ -0,0 +1,67 @@
...
+
+        // Then
+        XCTAssertEqual(
+            dd.logger.debugLog?.message,
+            """
+            NTP time synchronization completed.
+            Server time will be used for signing events (-1.0s difference with device time).
+            """
+        )
+    }
```
ncreated, line 46: I think the most critical behaviour in this test is _"it publishes new value"_, not _"it prints info message"_. That said, maybe it's worth updating the test name (e.g. _"it publishes new value and prints message"_)? Also, to reliably assert the update, it will be better to use `expectation` in publisher's block (otherwise this side effect can be easily missed):
```swift
publisher.publish { updatedOffset in
   XCTAssertEqual(updatedOffset, -1)
   expectation.fulfill()
}
```

Tests/DatadogTests/Datadog/DatadogCore/KronosClockPublisherTests.swift
```diff
@@ -0,0 +1,67 @@
...
+        try (0..<100).forEach { _ in
+            publisher.publish { _ in }
+            let pool = try XCTUnwrap(kronos.currentPool)
+            XCTAssertTrue(pool.hasSuffix(".datadog.pool.ntp.org"))
+            pools.insert(pool)
+        }
+
+        XCTAssertEqual(pools, Set(DatadogNTPServers), "Each time Datadog NTP server should be picked randomly.")
+    }
+
```
ncreated, line 26: We're missing coverage for one more logic branch in `KronosClock` - receiving an update on first calculated time value. In `KronosClockProtocol`, this corresponds to `first` parameter:
```swift
    /// - parameter first:      A closure that will be called after the first valid date is calculated.
    /// - parameter completion: A closure that will be called after _all_ the NTP calls are finished.
```

Sources/Datadog/DatadogCore/Context/DatadogContextProvider.swift
```diff
@@ -0,0 +1,121 @@
...
+    ///
+    ///     let reader = ServerOffsetReader<TimeInterval>(initialValue: 0)
+    ///     provider.assign(reader: reader, to: \.serverTimeOffset)
+    ///
+    /// - Parameters:
+    ///   - reader: The value reader.
+    ///   - keyPath: A context's key path that supports reading from and writing to the resulting value.
+    func assign<Reader>(reader: Reader, to keyPath: WritableKeyPath<DatadogContext, Reader.Value>) where Reader: ContextValueReader {
+        self.reader.append(reader: reader, receiver: keyPath)
+    }
```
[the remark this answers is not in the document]
ncreated: 👍 Got it. Not a blocker now - we will solve this somehow when all providers are ready for use.

ncreated (approved):
Looks great 🚀! Only non-blocking feedbacks in new tests added.

Tests/DatadogTests/Datadog/DatadogCore/DatadogContextProviderTests.swift
```diff
@@ -0,0 +1,81 @@
...
+    // MARK: - Thread Safety
+
+    func testConcurrentReadWrite() {
+        let provider = DatadogContextProvider(context: context)
+
+        DispatchQueue.concurrentPerform(iterations: 50) { iteration in
+            provider.read { _ in }
+            provider.write { $0 = .mockAny() }
+        }
+    }
```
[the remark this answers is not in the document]
ncreated: Awesome 💪. Yeah, 1K is more than enough for CI runs 👌 (I usually set it crazy high for local debugging, but then lower the bar for CI that runs it more often than me 🧑‍💻).

Sources/Datadog/DatadogCore/Context/NetworkConnectionInfoPublisher.swift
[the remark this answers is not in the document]
ncreated: New approach is much more straightforward, well done 👍.

Tests/DatadogTests/Datadog/DatadogCore/DatadogContextProviderTests.swift
maxep: Oops, yes it should be removed

Tests/DatadogTests/Datadog/DatadogCore/KronosClockPublisherTests.swift
maxep: Indeed, good catch!

maxep:
Thanks @ncreated for your reviews on this large and incremental change 🙏

PR #923, opened 2022-07-11
RUMM-2239 Fix Logger regressions and split `Logger` interface

ncreated:
### What and why?

📦 This PR does several changes to `Logger`:
* It separates `RemoteLogger` (sending logs to Datadog) and `ConsoleLogger` (printing logs to console). This is to isolate different implementations and enable getting rid of `LogOutput` interfaces later (they will be fully replaced by V2's _event write context_).
* Fixes the problem of `log.threadName` which would being incorrectly set to the SDK context thread (in V2), not the caller thread. It was a miss in #881.
* ~Improves the performance of `Logger` by removing the need of blocking user thread with `queue.sync {}` for synchronizing tags and attributes.~ (not yet possible without full V2)
* Introduces _event write context_ to `TracingWithLoggingIntegration`. This is a first step towards using event bus approach for cross-feature integrations in V2.

### How?

The public `Logger` now uses `RemoteLogger` or `ConsoleLogger` (or both through `CombinedLogger`) internally, depending on how the logger is configured.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


Sources/Datadog/Logging/RemoteLogger.swift
```diff
@@ -0,0 +1,144 @@
...
+    // MARK: - Logging
+
+    func log(level: LogLevel, message: String, error: Error?, attributes: [String: Encodable]?) {
+        guard level.rawValue >= configuration.threshold.rawValue else {
+            return
+        }
+
+        // on public API caller thread:
+        let date = dateProvider.now
+        let threadName = getCurrentThreadName()
```
ncreated, line 107: We record `date` and `threadName` before jumping off from the caller thread. Alternatively, the `date` could be read from SDK `context`, but then it would require `queue.sync {}` to manage logger tags and attributes ([like we do in V1](https://github.com/DataDog/dd-sdk-ios/blob/f330ac00db328e110ca5287818c58cc9aaa706ec/Sources/Datadog/Logger.swift#L266-L270)). I made an extra annotation on `date` in `LogEventBuilder.createLogEvent()` to make it documented and clear.

Sources/Datadog/FeaturesIntegration/TracingWithLoggingIntegration.swift
```diff
@@ -80,20 +73,26 @@ internal struct TracingWithLoggingIntegration {
...
+                context: context,
+                threadName: threadName
+            )
 
-        if let event = log {
-            loggingOutput.write(log: event)
+            if let log = log {
+                writer.write(value: log)
+            }
         }
```
ncreated, line 96: Here we switch from writing to `LogOutput` to using V2's _event write context_. The way I see it is that soon "tracing with logging" integration will be done through event bus and the final `LogEvent` write will use _event write context_. In that sense it's a step forward, WDYT @maxep ?

The reason of this is the change in `logBuilder.createLogWith()` which makes it easier to use in V2, not V1.

Sources/Datadog/Logger.swift
```diff
@@ -57,271 +45,187 @@ public enum LogLevel: Int, Codable {
...
+
+/// Datadog logger.
+public class Logger: LoggerProtocol {
+    /// An internal shim to V2 Logger.
+    /// Needed for backward compatibility of V1 `Logger` APIs.
+    internal let v2Logger: LoggerProtocol
+
+    init(v2Logger: LoggerProtocol) {
+        self.v2Logger = v2Logger
     }
```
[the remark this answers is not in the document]
maxep: ACK ✅

Sources/Datadog/Logger.swift
```diff
@@ -460,49 +364,104 @@ public class Logger {
...
+    func add(tag: String) {
+        combinedLoggers.forEach { $0.add(tag: tag) }
+    }
+
+    func remove(tag: String) {
+        combinedLoggers.forEach { $0.remove(tag: tag) }
+    }
+}
+
+internal struct NoOpLogger: LoggerProtocol {
```
maxep, line 459:
```suggestion
internal struct NOPLogger: LoggerProtocol {
```

cf. https://github.com/DataDog/dd-sdk-ios/pull/921#discussion_r917723968

Sources/Datadog/FeaturesIntegration/TracingWithLoggingIntegration.swift
maxep: Sounds great 👍

Sources/Datadog/Logging/RemoteLogger.swift
maxep: Yes, lets keep the date outside of the context thread. It will be the same behaviour with v2 context 👍

maxep (commented):
This looks great and it's very well documented 🙏 

I approve for the design review ✅

Sources/Datadog/Logging/RemoteLogger.swift
```diff
@@ -0,0 +1,158 @@
...
+            return
+        }
+
+        // on user thread:
+        let date = dateProvider.now
+        let threadName = getCurrentThreadName()
+
+        // SDK context must be requested on the user thread to ensure that it provides values
+        // that are up-to-date for the caller.
+        self.core.v1.scope(for: LoggingFeature.self)?.eventWriteContext { context, writer in
```
ncreated, line 118: @maxep Given what we discussed (that `core.scope(for:).eventWriteContext { }` will be always called on the user thread), I now wonder if it would make sense to include both `date` and `threadName` in `DatadogContext` (not now, but later). This way, the `context` might provide:
```swift
context.threadName    // the name of the user thread
context.date          // current device time
context.serverOffset  // correction to server time
context.serverDate    // device time corrected by server time
```

maxep: Yes, we could do that before jumping on the context thread [here](https://github.com/DataDog/dd-sdk-ios/pull/916/files#diff-8ea90fa9b289cf85fe53811d1d1a3f08d4a33cc7879ac74ff3ea71b9257a5a8dR73-R79) 👍

maxep (approved):
Looks perfect!!

PR #774, opened 2022-03-01
RUMM-2034 Add tvOS Example target

maxep:
### What and why?

Add tvOS Example target to host tests.

### How?

The `Example tvOS` application target only showcase SDK configuration and serve as a host for the test-suite. The app does not present any UI.

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing change.

PR #926, opened 2022-07-12
RUMM-2299 Core Context Launch Time Reader

maxep:
### What and why?

Provide Launch Time through Datadog Context.

### How?

Implement a `LaunchTimeReader` that complies to `ContextValueReader` and provide `LaunchTime` to the context.
The reader uses the objc `AppLaunchHandler` like the previous `LaunchTimeProvider`

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/DatadogCore/Context/LaunchTimeReader.swift
+++ b/Sources/Datadog/DatadogCore/Context/LaunchTimeReader.swift
@@ -1,0 +1,41 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+#if SPM_BUILD
+import _Datadog_Private
+#endif
+
+internal struct LaunchTimeReader: ContextValueReader {
+    let initialValue: LaunchTime
+
+    /// Lock object to sync launch time read.
+    let lock: Any
+
+    init() {
+        let lock = NSObject()
+        initialValue = __dd_private_objc_sync_LaunchTime(lock)
+        self.lock = lock
+    }
+
+    func read(to receiver: inout LaunchTime) {
+        receiver = __dd_private_objc_sync_LaunchTime(lock)
+    }
+}
+
+private func __dd_private_objc_sync_LaunchTime(_ lock: Any) -> LaunchTime {
+    // Even if __dd_private_AppLaunchTime() is using a lock behind the
+    // scenes, TSAN will report a data race if there are no
+    // synchronizations at this level.
+    objc_sync_enter(lock)
+    let time = LaunchTime(
+        launchTime: __dd_private_AppLaunchTime(),
+        isActivePrewarm: __dd_private_isActivePrewarm()
+    )
+    objc_sync_exit(lock)
+    return time
+}
--- a/Sources/Datadog/DatadogInternal/Context/DatadogContext.swift
+++ b/Sources/Datadog/DatadogInternal/Context/DatadogContext.swift
@@ -59,6 +59,9 @@ internal struct DatadogContext {
     /// Current user information.
     var userInfo: UserInfo
 
+    /// Application launch time..
+    var launchTime: LaunchTime
+
     // MARK: - Device Specific
 
     /// Network information.
--- a/Sources/Datadog/DatadogInternal/Context/LaunchTime.swift
+++ b/Sources/Datadog/DatadogInternal/Context/LaunchTime.swift
@@ -12,8 +12,7 @@ internal struct LaunchTime {
     /// to receiving `UIApplication.didBecomeActiveNotification` notification.
     ///
     /// If the `UIApplication.didBecomeActiveNotification` has not yet been received by the
-    /// time this variable is requested, the value should represent the time interval between now and the
-    /// process start time.
+    /// time this value is provided, it will represent the time interval between now and the process start time.
     let launchTime: TimeInterval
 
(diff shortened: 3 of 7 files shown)


Sources/Datadog/DatadogCore/Context/LaunchTimeReader.swift
```diff
@@ -0,0 +1,41 @@
...
+    let initialValue: LaunchTime
+
+    /// Lock object to sync launch time read.
+    let lock: Any
+
+    init() {
+        let lock = NSObject()
+        initialValue = __dd_private_objc_sync_LaunchTime(lock)
+        self.lock = lock
+    }
```
ncreated, line 23: Shouldn't we define single, static `lock`?
```swift
internal struct LaunchTimeReader: ContextValueReader {
   private static lock = NSObject()

   // ...
}
```
Otherwise, we will be creating distinct `lock` object for each "instance" of `LaunchTimeReader`. This will become an issue when multiple instances of the SDK run in a single process - the critical section behind this `lock` will be accessed by multiple readers at once.

Sources/Datadog/DatadogInternal/Context/DatadogContext.swift
```diff
@@ -56,6 +56,9 @@ internal struct DatadogContext {
     /// Current device information.
     let device: DeviceInfo
 
+    /// Application launch time..
```
ncreated, line 59:
```suggestion
    /// Application launch time.
```

Sources/Datadog/DatadogInternal/Context/LaunchTime.swift
```diff
@@ -0,0 +1,21 @@
...
+import Foundation
+
+/// Provides the application launch time.
+internal struct LaunchTime {
+    /// The app process launch duration (in seconds) measured as the time from process start time
+    /// to receiving `UIApplication.didBecomeActiveNotification` notification.
+    ///
+    /// If the `UIApplication.didBecomeActiveNotification` has not yet been received by the
+    /// time this variable is requested, the value should represent the time interval between now and the
+    /// process start time.
```
ncreated, line 16: Needs a wording change, as now this value is not "requested" - it is "provided".
```suggestion
    /// If the `UIApplication.didBecomeActiveNotification` has not yet been received by the
    /// time this value is provided, it will represent the time interval between now and the
    /// process start time.
```

Tests/DatadogTests/Datadog/DatadogCore/LaunchTimeReaderTests.swift
```diff
@@ -0,0 +1,83 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import XCTest
+@testable import Datadog
+
+// TODO: RUMM-2034 Remove this flag once we have a host application for tests
+#if !os(tvOS)
```
ncreated, line 11: Just checking: is this `TODO` still relevant? The https://github.com/DataDog/dd-sdk-ios/pull/774 was already done.

Tests/DatadogTests/Datadog/DatadogCore/LaunchTimeReaderTests.swift
```diff
@@ -0,0 +1,83 @@
...
+
+class LaunchTimeReaderTests: XCTestCase {
+    override func tearDown() {
+        super.tearDown()
+        setenv("ActivePrewarm", "", 1)
+    }
+
+    func testGivenStartedApplication_whenRequestingLaunchTimeAtAnyTime_itReturnsTheSameValue() {
+        // Given
+        let publisher = LaunchTimeReader()
```
ncreated, line 21: We mean `reader`, not `publisher`, no? Same in below tests 👇.

ncreated (approved):
Looks great 👌! I left small feedbacks - LMKWDYT 🙌

Sources/Datadog/DatadogCore/Context/LaunchTimeReader.swift
maxep: We already use a static lock in [ObjcAppLaunchHandler](https://github.com/DataDog/dd-sdk-ios/blob/develop/Sources/_Datadog_Private/ObjcAppLaunchHandler.m#L89L91), so we are safe in case of multiple readers. This instance lock is coming from the previous implementation of [LaunchTimeProvider](https://github.com/DataDog/dd-sdk-ios/blob/develop/Sources/Datadog/Core/System/LaunchTimeProvider.swift#L28L36), which says:
```
// Even if __dd_private_AppLaunchTime() is using a lock behind the
// scenes, TSAN will report a data race if there are no
// synchronizations at this level.
```
I've tried removing this lock and I didn't see any TSAN error.. so I'm not sure where it's coming from, do you know more?
The static lock in `ObjcAppLaunchHandler` is enough to secure concurrent access, I was cautious to keep previous behaviour but we might not need it!

Tests/DatadogTests/Datadog/DatadogCore/LaunchTimeReaderTests.swift
maxep: Ah indeed, good catch :)

Sources/Datadog/DatadogCore/Context/LaunchTimeReader.swift
ncreated: Sounds good 👍. AFAIK we only observed the TSAN issue happening on CI (when running tests through CLI).

PR #927, opened 2022-07-12
RUMM-2299 Core Context User Info Publisher

maxep:
### What and why?

Publish user info to Datadog Context.

### How?

Implement a `UserInfoPublisher` that complies to `ContextValuePublisher` and provide `UserInfo` to the context. Setting the `current` user info property of the publisher will publish the value to the subscriber.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/DatadogInternal/Context/DatadogContext.swift
+++ b/Sources/Datadog/DatadogInternal/Context/DatadogContext.swift
@@ -56,6 +56,9 @@ internal struct DatadogContext {
     /// Current device information.
     let device: DeviceInfo
 
+    /// Current user information.
+    var userInfo: UserInfo
+
     // MARK: - Device Specific
 
     /// Network information.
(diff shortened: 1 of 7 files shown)


Sources/Datadog/DatadogInternal/Context/DatadogContext.swift
```diff
@@ -56,6 +56,9 @@ internal struct DatadogContext {
     /// Current device information.
     let device: DeviceInfo
 
+    /// Current user information.
+    var userInfo: UserInfo
+
```
ncreated, line 61: Question: why do we need this and some other`DatadogContext` properties to be mutable (`var`) 🤔💭?

ncreated (approved):
🚀 Awesome 🚀! One question left (❔).

Sources/Datadog/DatadogInternal/Context/DatadogContext.swift
maxep: So they can be mutated by the publishers in the [context provider](https://github.com/DataDog/dd-sdk-ios/blob/feature/v2/Sources/Datadog/DatadogCore/Context/DatadogContextProvider.swift#L99). But since the `DatadogContext` is a struct, there is no way of mutating those values outside of the provider.

ncreated: Ah, got it 👍. So it will be `public internal(set) var` at the end - means still immutable for Features.

PR #928, opened 2022-07-12
RUMM-2299 Core Context Application State Publisher

maxep:
### What and why?

Publish application state to Datadog Context.

### How?

Implement a `AppStatePublisher` that complies to `ContextValuePublisher` and provide `AppStateHistory` to the context. 

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


Sources/Datadog/DatadogCore/Context/AppStatePublisher.swift
```diff
@@ -0,0 +1,145 @@
...
+    typealias Snapshot = AppStateHistory.Snapshot
+
+    /// The default publisher queue.
+    private static let defaultQueue = DispatchQueue(
+        label: "com.datadoghq.app-state-publisher",
+        target: .global(qos: .utility)
+    )
+
+    /// The initial history value.
+    var initialValue: AppStateHistory
```
ncreated, line 19: nit/ Seems it can be immutable, no?
```suggestion
    /// The initial history value.
    let initialValue: AppStateHistory
```

Sources/Datadog/DatadogCore/Context/AppStatePublisher.swift
```diff
@@ -0,0 +1,145 @@
...
+        )
+
+        self.initialValue = initialValue
+        self.history = initialValue
+        self.queue = queue
+        self.dateProvider = dateProvider
+        self.notificationCenter = notificationCenter
+    }
+
+    func publish(to receiver: @escaping ContextValueReceiver<AppStateHistory>) {
```
ncreated, line 76: nit/ The more often I see this in different publishers, the more I think it should be called `startPublishing(to:)` to better express what it should do 💭. Just a note, I'm totally fine with keeping existing name (after all, it is well explained in the protocol comment).

Sources/Datadog/DatadogCore/Context/AppStatePublisher.swift
```diff
@@ -0,0 +1,145 @@
...
+    ///   - initialState: The initial application state.
+    ///   - queue: The queue for publishing the history.
+    ///   - dateProvider: The date provider for the Application state snapshot timestamp.
+    ///   - notificationCenter: The notification center where this publisher observes `UIApplication` notifications.
+    init(
+        initialState: AppState,
+        queue: DispatchQueue = AppStatePublisher.defaultQueue,
+        dateProvider: DateProvider = SystemDateProvider(),
+        notificationCenter: NotificationCenter = .default
+    ) {
```
ncreated, line 58: nit/ As we have a nice doc string now (🏅), I think it's worth adding in that it needs to be called on the main thread, because it calls `application. applicationState` transitively:
```swift
private static var currentApplicationState: UIApplication.State {
   UIApplication.managedShared?.applicationState ?? .active // fallback to most expected state
}
```
WDYT?

```suggestion
    /// Creates a Application state publisher for publishing application state
    /// history.
    ///
    /// **Note**: It must be called on the main thread.
    ///
    /// - Parameters:
    ///   - initialState: The initial application state.
    ///   - queue: The queue for publishing the history.
    ///   - dateProvider: The date provider for the Application state snapshot timestamp.
    ///   - notificationCenter: The notification center where this publisher observes `UIApplication` notifications.
    init(
        initialState: AppState,
        queue: DispatchQueue = AppStatePublisher.defaultQueue,
        dateProvider: DateProvider = SystemDateProvider(),
        notificationCenter: NotificationCenter = .default
    ) {
```

Sources/Datadog/DatadogCore/Context/NetworkConnectionInfoPublisher.swift
```diff
@@ -21,12 +21,19 @@ import Network
...
+
     let initialValue: NetworkConnectionInfo
 
     private let monitor: NWPathMonitor
     private let queue: DispatchQueue
 
-    init(monitor: NWPathMonitor, queue: DispatchQueue) {
+    init(
+        monitor: NWPathMonitor = .init(),
+        queue: DispatchQueue = NWPathMonitorPublisher.defaultQueue) {
```
ncreated, line 36: nit/ weird it didn't raise linter issue:
```suggestion
        queue: DispatchQueue = NWPathMonitorPublisher.defaultQueue
    ) {
```

ncreated (approved):
Awesome 🚀. I left few non-blocking feedbacks 🙌.

Sources/Datadog/DatadogCore/Context/AppStatePublisher.swift
maxep: Oh indeed, but this should be documented in the extension below, the one that uses `UIKit`.
I didn't know about threading constraint on `UIApplication/applicationState`. This is a good argument to make the subscription/assignment thread-safe as you mentioned [here](https://github.com/DataDog/dd-sdk-ios/pull/916#discussion_r919788572) so we can dispatch the `ApplicationStatePublisher` init and subscription on the main thread!

Sources/Datadog/DatadogCore/Context/AppStatePublisher.swift
maxep: That would work, current name also correctly reflect what it should do, at the protocol level, IMO. `publisher.publish(to receiver:)` Lets keep it as is for now 👍

PR #929, opened 2022-07-12
Expose telemetry and long tasks Api through _InternalProxy

fuzzybinary:
### What and why?

We need to expose certain internal classes and functions for use by cross platform APIs. This starts that process by exposing Telemetry and addLongTask through the _InternalProxy class.

### How?

We expose internal methods through the _InternalProxy class, which is documented for internal use, and the `_internal` static member on the Datadog object. We keep Telemetry and RUM as separate proxies, in case we have to split them as v2 takes shape.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


ncreated (approved):
Looks good 👍. I left non-blocking feedback that could help simplifying tests.

Sources/Datadog/Datadog.swift
```diff
@@ -340,6 +340,13 @@ public class Datadog {
         // Deinitialize `Datadog`:
         defaultDatadogCore = NOOPDatadogCore()
     }
+
+    // MARK: - Internal Proxy - exposure of internal classes (Mostly used for cross platform libraries)
+
+    static var _internal: _InternalProxy = {
+        let monitor = Global.rum as? RUMMonitor
+        return _InternalProxy(monitor: monitor)
+    }()
```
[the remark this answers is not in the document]
fuzzybinary: Yeah I ran into this with Unit Tests, and you're right it's a weird side affect -- I'll see if I can replace it with a non-lazy property.

maxep (changes requested):
Looks good! tho, the RUM part won't be working if we set the `_InternalProxy` this way in `Datadog. initialize()`. I've left some suggestions.

Sources/Datadog/DatadogInternal/_InternalProxy.swift
```diff
@@ -0,0 +1,54 @@
...
+public class _InternalProxy {
+    let _telemtry = _TelemetryProxy()
+    let _rum: _RumProxy?
+
+    init(rumSubscriber: RUMCommandSubscriber?) {
+        if let rumSubscriber = rumSubscriber {
+            _rum = _RumProxy(rumSubscriber: rumSubscriber)
+        } else {
+            _rum = nil
+        }
```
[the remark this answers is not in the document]
fuzzybinary: I knew there had to be a better way to do this. Thanks!

Sources/Datadog/Datadog.swift
```diff
@@ -287,6 +287,8 @@ public class Datadog {
 
         defaultDatadogCore = core
 
+        _internal = _InternalProxy(rumSubscriber: Global.rum as? RUMCommandSubscriber)
```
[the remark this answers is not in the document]
maxep: Then you can do:
```swift
static let _internal = _InternalProxy(monitor: { Global.rum as? RUMMonitor })
```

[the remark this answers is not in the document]
fuzzybinary: Yeah, I have to rethink this.... lazy worked but definitely was going to be error prone, but I can't rely on that global it seems....

It's possible the answer here is not a Datadog._internal except for telemetry. I think I need Datadog._internal and RUMMonitor._internal as separate objects, which was going to be the plan eventually anyway.

fuzzybinary:
Okay, I've rethought this and put the RUMInternalProxy on the RUMMonitor where it belongs. This should have been the approach from the beginning.

Datadog/Datadog.xcodeproj/project.pbxproj
```diff
@@ -2359,27 +2371,28 @@
...
-				61C5A89524509BF600DA608C /* TracerTests.swift */,
-				61E917D2246546BF00E6C631 /* TracerConfigurationTests.swift */,
-				61B558CE2469561C001460D3 /* LoggerBuilderTests.swift */,
-				617B953C24BF4D8F00E6F443 /* RUMMonitorTests.swift */,
-				617B954124BF4E7600E6F443 /* RUMMonitorConfigurationTests.swift */,
 				61133C212423990D00786299 /* Core */,
+				61F2724725C9437C00D54BF8 /* CrashReporting */,
+				D2956CAE2869D516007D5462 /* DatadogInternal */,
+				61216278247D20D500AC5D67 /* FeaturesIntegration */,
+				61D3E0DD277B3D6E008BE766 /* Kronos */,
```
ncreated, line 2378: I just noticed that the `pbxproj` diff looks weird and seems it can bring serious git conflicts with `feature/v2`. Could we limit it (manually) to only include added files?

Sources/Datadog/Datadog.swift
```diff
@@ -340,6 +340,10 @@ public class Datadog {
         // Deinitialize `Datadog`:
         defaultDatadogCore = NOOPDatadogCore()
     }
+
+    // MARK: - Internal Proxy - exposure of internal classes (Mostly used for cross platform libraries)
+
+    public private(set) static var _internal = _InternalProxy()
```
maxep, line 346: This could be:
```suggestion
    public internal(set) static var _internal = _InternalProxy()
```

So you won't need to override it in `RUMMonitor`, WDYT?

fuzzybinary: I'm not sure I understand what you're suggesting? The variable in RUMMonitor is a completely different proxy.

maxep: oops sorry, that's the wrong line!

Sources/Datadog/DDRUMMonitor.swift
```diff
@@ -286,6 +286,10 @@ public class DDRUMMonitor {
     /// - Parameter key: key for the attribute that will be removed.
     public func removeAttribute(forKey key: AttributeKey) {}
 
+    public var _internal: _RUMInternalProxy? {
+        get { return nil }
+    }
```
maxep, line 291: Here, this could be:
```suggestion
    public internal(set) var _internal: _RUMInternalProxy? = nil
```
Then, you don't need to override it in `RUMMonitor`

fuzzybinary: Ah yeah, okay -- I think you left the initial  comment on the wrong variable. I think initially I wanted _internal to be non-optional which is why I used the getter, but that didn't work well. I'll modify it to do this instead.

PR #933, opened 2022-07-19
RUMM-2323 Support `oneOf` in RUM schema

ncreated:
### What and why?

📦 This PR updates `rum-models-generator` with ability of handling `oneOf` directive in JSON Schema.

With the first-class support for `oneOf`, this PR achieves few goals:
* it gets rid of [the workaround](https://github.com/DataDog/dd-sdk-ios/blob/b5e2d2187202d3562f084378ddea4b6ac0294561/tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/JSON/JSONSchemaToJSONTypeTransformer.swift#L36-L42) added for dealing with non-homogenous `action.id` value introduced in https://github.com/DataDog/rum-events-format/pull/57;
* it further simplifies the logic of handling root schemas - continuing the work started in https://github.com/DataDog/dd-sdk-ios/pull/780, it now handles root `oneOf` uniformly with other directives;
* last, it prepares foundation for generating Mobile Session Replay models, which heavily use `oneOf` in their schema.

This PR doesn't yet provide:
* ~Objc-interop code generation for `oneOf`. To keep this PR small and enable pre-validation of core logic, updates to `ObjcInteropPrinter` will be added in next PR (targeting this branch).~ done through #936 
* Code generation for Mobile SR models. Although `oneOf` support brings us much closer to this, we still need to cover few edge cases to handle MSR schemas. This will come in `RUMM-2266`.

### How?

This required changes in all stages of the models generation pipeline:
* "`*.json` schema → `JSONType` transformer" - to introduce `JSONOneOfs: JSONType`;
* "`JSONType` → `SwiftType` transformer" - to transform `JSONOneOfs` into `SwiftAssociatedTypeEnum`;
* `SwiftPrinter` - to print code for `SwiftAssociatedTypeEnum`.

After all, `oneOf` handling differs depending on where `oneOf` is defined:
* If it is declared at the root level of the schema, a distinct root model is generated for each one-of branch. This is the case of `rum-events-format.json` schema, which only lists different subschemas which are part of the system.
* If it is declared as nested part of another schema, we use Swift enum with associated type to generate code for each one-of branch. This is the case of non-homogenous `action.id` which can be either `String` (single UUID) or `[String]` (multiple UUIDs).

The code generated for `action.id` looks this:
```swift
/// UUID of the action
public enum RUMActionID: Codable {
    case string(value: String)
    case stringsArray(value: [String])

    // MARK: - Codable

    public func encode(to encoder: Encoder) throws {
        // Encode only the associated value, without encoding enum case
        var container = encoder.singleValueContainer()

        switch self {
        case .string(let value):
            try container.encode(value)
        case .stringsArray(let value):
            try container.encode(value)
        }
    }

    public init(from decoder: Decoder) throws {
        // Decode enum case from associated value
        let container = try decoder.singleValueContainer()

        if let value = try? container.decode(String.self) {
            self = .string(value: value)
            return
        }
        if let value = try? container.decode([String].self) {
            self = .stringsArray(value: value)
            return
        }
        let error = DecodingError.Context(
            codingPath: container.codingPath,
            debugDescription: """
            Failed to decode `RUMActionID`.
            Ran out of possibilities when trying to decode the value of associated type.
            """
        )
        throw DecodingError.typeMismatch(RUMActionID.self, error)
    }
}
```

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


tools/rum-models-generator/Sources/RUMModelsGeneratorCore/RUM/RUMSwiftTypeTransformer.swift
```diff
@@ -198,6 +217,18 @@ internal class RUMSwiftTypeTransformer {
...
+            fixedName = "RUMDisplay"
+        }
+
+        // Since https://github.com/DataDog/rum-events-format/pull/57 `action.id` can be either
+        // single `String` or an array of `[String]`. This is handled by generating Swift enum with
+        // two cases and different associated types. To not duplicate generated code in each nested
+        // context we generate single root type: `RUMActionID`.
+        if fixedName == "ID", let parentStructName = (context.parent as? SwiftStruct)?.name, parentStructName == "action" {
+            fixedName = "RUMActionID"
+        }
```
maxep, line 230: /question
It would make sense that all `$ref` are not nested, while other types are, no? instead of hard coding this mapping in the generator.

maxep (approved):
Looks great 👍  I've left a general question which does not imply any change in this PR.

tools/rum-models-generator/Sources/RUMModelsGeneratorCore/RUM/RUMSwiftTypeTransformer.swift
ncreated: I see what you mean. It more relates to this definition:
```swift
/// Types which will shared between all input `types`. Sharing means detaching those types from nested declaration
/// and putting them at the root level of the resultant `types` array, so the type can be printed without being nested.
private let sharedTypeNames = [
   "RUMConnectivity",
   "RUMUser",
   "RUMMethod",
   "RUMEventAttributes",
   "RUMCITest",
   "RUMDevice",
   "RUMOperatingSystem",
   "RUMDisplay",
   "RUMActionID",
]
```

Now, when comparing this to [`_common-schema.json`](https://github.com/DataDog/rum-events-format/blob/master/schemas/rum/_common-schema.json) (a common `$ref` in multiple event schemas), indeed most of these objects are part of it and could be inferred automatically. But some other are not (e.g. `RUMEventAttributes` or `RUMMethod`), similar to few primitive values (e.g. `service` or `version`), so it would still require manual patching with opt-ins or opt-outs declarations.

I will look closer at this when working on Session Replay models. Compared to RUM events, these use way more `$refs` and automated inferring could work better.

Last, the `RUMSwiftTypeTransformer` is only an optional add-on to the entire JSON schema <> Swift generation system implemented by `rum-models-generator`. It only decorates intermediate Swift schema before it is passed to `SwiftPrinter`, but everything would work perfectly well without it (just the names would be weird and structure not so convenient). Because it is well decoupled, implementing few decorative hacks in there, like `sharedTypeNames` or `fixedName` is not a big problem IMO.

PR #938, opened 2022-07-22
RUMM-2266 Enable Session Replay models generation

ncreated:
### What and why?

📦 This PR fixes 2 problems discovered when running models generation for Session Replay schemas. With this fix, the generator makes successful run on [session-replay-mobile-events-format.json](https://github.com/DataDog/rum-events-format/blob/master/session-replay-mobile-events-format.json) and prints Swift code.

This PR **doesn't yet**:
* Decorate SR models code appropriately. Similar to RUM, there are several things we may want to adjust in produced code (e.g. naming, or reusing common types) to make it more convenient to work with.
* Restructure `rum-models-generator` package to include separate generation tracks for RUM and SR.

Both elements will be delivered in next PRs.

### How?

First, I enhanced JSON schema reader with inferring `type` if it is not explicitly defined. This is the case of [`mutation-data-schema.json`](https://github.com/DataDog/rum-events-format/blob/1a3eef2e1422106dc4dc1ad72dcca50cdefd3284/schemas/session-replay/mobile/mutation-data-schema.json#L16-L35) which defines following array items without declaring `type: object`:
```json
"items": {
   "required": [
       "wireframe"
   ],
   "properties": {
       "previousId": {
           "type": "integer",
           "description": "The previous wireframe id next or after which this new wireframe is drawn or attached to, respectively.",
           "readOnly": true
       },
       "wireframe": {
           "$ref": "wireframe-schema.json",
           "description": "The new wireframe that needs to be added to the screen.",
           "readOnly": true
       }
   }
}
```
In such case ☝️, we can infer that item is of `type: object`, because it defines `properties`. 

Alternatively, we could consider updating `mutation-data-schema.json` with `type` information. I don't do it for two reasons:
* couldn't find anywhere in JSON schema doc if `type` is a mandatory field (seems not);
* other generators (Kotlin + TS) seem to handle this case, so it sounds reasonable to include it in ours.

Second, there was one generation branch in `oneOf` resolution marked as _"not yet supported"_. In this PR I'm adding support to mixing `JSONObject` and `JSONOneOf` definitions in a single `oneOf`.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/JSON/JSONSchema.swift
+++ b/tools/rum-models-generator/Sources/RUMModelsGeneratorCore/Input/JSON/JSONSchema.swift
@@ -86,7 +86,6 @@ internal class JSONSchema: Decodable {
             if self.type == nil && self.properties != nil {
                 self.type = .object
             }
-
         } catch let keyedContainerError as DecodingError {
             // If data in this `decoder` cannot be represented as keyed container, perhaps it encodes
             // a single value. Check known schema values:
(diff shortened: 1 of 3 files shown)


maxep (approved):
👌

PR #940, opened 2022-07-25
RUMM-2266 Reorganise `rum-models-generator` package for Session Replay code generation

ncreated:
### What and why?

📦 This PR reorganises `rum-models-generator` package so it can be later easily extended with Session Replay code generation pipeline. This is to clean up the package code and keep both generation tracks separate.

It doesn't yet generate Session Replay models - this requires abstracting some configuration  (e.g. the access control in generated code) and will be added in next PR.

### How?

The package is now composed of 3 targets:
* `CodeGeneration`, which provides generation pipeline:
```swift
import CodeGeneration

let code: String = try ModelsGenerator()
   .generateCode(from: schemaURL)
   .decorate(using: codeDecorator) // optional phase
   .print(
      using: OutputTemplate(...), 
      and: SwiftPrinter() // or `ObjcInteropPrinter()`
   )
```
* `CodeDecoration`, providing 2 distinct implementations of `codeDecorator`:
   * `RUMCodeDecorator: CodeDecorator` - applies RUM conventions to generated code
   * `SRCodeDecorator: CodeDecorator` - will apply SR conventions to generated code
* `rum-models-generator` - small CLI wrapper

🎁 With this change, the `CodeGeneration` module becomes fully RUM- and SR-agnostic. It can be understood as general-purpose generator of Swift and Objc-interop code from JSON Schema.

Conventions required by RUM and SR products are only implemented in `CodeDecoration` module, by conformance to `CodeGeneration.CodeDecorator` interface:
```swift
/// A type decorating generated code.
/// Decoration can be used to adjust naming and structure in generated code before it is printed.
public protocol CodeDecorator {
    func decorate(code: GeneratedCode) throws -> GeneratedCode
}

/// Schema describing generated code.
public struct GeneratedCode {
    /// An array of Swift schemas describing root constructs in generated code.
    public let swiftTypes: [SwiftType]

    // ...
}
```

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


tools/rum-models-generator/Sources/CodeGeneration/Generate/Transformers/Swift/SwiftType.swift
```diff
@@ -0,0 +1,168 @@
...
+
+/// Swift primitive type.
+public protocol SwiftPrimitiveType: SwiftType {}
+/// An allowed value of Swift primitive type.
+public protocol SwiftPrimitiveValue {}
+/// An allowed default value of Swift property.
+public protocol SwiftPropertyDefaultValue {}
+/// An allowed value of Swift with no obj-c interoperability.
+public protocol SwiftPrimitiveNoObjcInteropType: SwiftPrimitiveType {}
+
```
ncreated, line 20: The only changes in this file are `internal` → `public` access modifiers for all pieces of Swift schema. This is to make it visible for external module which performs code decoration.

tools/rum-models-generator/Package.swift
```diff
@@ -11,20 +11,45 @@ let package = Package(
...
         .target(
             name: "rum-models-generator",
             dependencies: [
-                "RUMModelsGeneratorCore",
+                "CodeGeneration",
+                "CodeDecoration",
                 .product(name: "ArgumentParser", package: "swift-argument-parser")
-            ]),
+            ]
+        ),
```
ncreated, line 22: As it is no longer `rum-` models generator, we may want to rename the package. I don't do this, because it has consequences on other files in the repo (all places where we call it from). I didn't want to spend time on it right now, but it's IMO something we might want consider in the future.

maxep: Yes, it's fine keep it for now 👍

maxep (approved):
Perfect 👏 👏

PR #941, opened 2022-07-25
Add processor architecture to RUMDeviceInfo

fuzzybinary:
### What and why?

Flutter / Dart needs architecture in order to select the proper symbol file to process.  Currently, only add architecture if UIKit is imported

### How?

Add architecture when creating the DeviceInfo. I've added a few checks on the Integration tests to make sure that the architecture is present and is an expected value.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [x] Run smoke tests

### DIFF
--- a/Sources/Datadog/DatadogInternal/Context/DeviceInfo.swift
+++ b/Sources/Datadog/DatadogInternal/Context/DeviceInfo.swift
@@ -25,22 +25,28 @@ internal struct DeviceInfo {
     /// The version of the operating system, e.g. "15.4.1".
     let osVersion: String
 
+    /// The architecture of the device
+    let architecture: String
+
     init(
         name: String,
         model: String,
         osName: String,
-        osVersion: String
+        osVersion: String,
+        architecture: String
     ) {
         self.name = name
         self.model = model
         self.osName = osName
         self.osVersion = osVersion
+        self.architecture = architecture
     }
 }
 
 #if canImport(UIKit)
 
 import UIKit
+import MachO
 
 extension DeviceInfo {
     /// Creates device info based on UIKit description.
@@ -50,13 +56,15 @@ extension DeviceInfo {
     ///   - device: The `UIDevice` description.
     init(
         model: String,
-        device: UIDevice
+        device: UIDevice,
+        architecture: String
     ) {
         self.init(
             name: device.model,
             model: model,
             osName: device.systemName,
-            osVersion: device.systemVersion
+            osVersion: device.systemVersion,
+            architecture: architecture
         )
     }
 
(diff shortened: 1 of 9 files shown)


Sources/Datadog/DatadogInternal/Context/DeviceInfo.swift
```diff
@@ -69,13 +77,19 @@ extension DeviceInfo {
         processInfo: ProcessInfo = .processInfo,
         device: UIDevice = .current
     ) {
+        var architecture = "unknown"
+        if let archInfo = NXGetLocalArchInfo()?.pointee {
+            architecture = String(utf8String: archInfo.name) ?? "unknown"
+        }
```
ncreated, line 83: I wonder if these problems do still exist (these are dated back to 2020):
* https://github.com/getsentry/sentry-cocoa/issues/635
* https://github.com/firebase/firebase-ios-sdk/issues/6128

TL;DR; This was causing app rejection in App Store submission. Any thoughts @fuzzybinary @maxep ? Is there any alternative way to get this information?

Tests/DatadogIntegrationTests/Scenarios/RUM/RUMManualInstrumentationScenarioTests.swift
```diff
@@ -58,6 +58,15 @@ class RUMManualInstrumentationScenarioTests: IntegrationTests, RUMCommonAsserts
         let view1 = session.viewVisits[0]
         XCTAssertEqual(view1.name, "SendRUMFixture1View")
         XCTAssertEqual(view1.path, "Example.SendRUMFixture1ViewController")
+        XCTAssertNotNil(view1.viewEvents.last?.device)
+        XCTAssertNotNil(view1.viewEvents.last?.device?.architecture)
+        if let architecture = view1.viewEvents.last?.device?.architecture {
+            // i486 is the architecture of the VMs on bitrise
```
ncreated, line 64: 🥴👍

ncreated (commented):
Looks very good 👌. I left feedback on potential use of private API - it needs some more thoughts ⚠️.

Sources/Datadog/DatadogInternal/Context/DeviceInfo.swift
fuzzybinary: I actually saw those, but if I read things right, it looks like this only affected the Mac app store / Catalyst apps (which we don't support at the moment anyway?)

But even according to the Firebase issue, they didn't actually remove the call and haven't gotten any reports of being rejected since September.

maxep: :/ this worries me a bit, but it looks like Firebase is still using this API without any report. So we can try with dogfooding and see if we got a rejection from the app store, WDYT? We would surely make sure the app team is ok with it!
My concern is that we aim to support multiplatform, so this would add debt for macos. We are not there yet and I prefer unblocking flutter and tackle that later 👍

fuzzybinary: My guess is that API was flagged as private by accident.

maxep (approved):
👍

PR #942, opened 2022-07-25
RUMM-2207 Battery Status Reader and Low Battery Mode Publisher

maxep:
### What and why?

Provide battery status and low battery mode through core context.

### How?

Introduce `BatteryStatusReader` and `LowPowerModePublisher` for providing the core context. The values have been split to reflect their respective data source and providing methods: reader vs. publisher.

Previous battery status is renamed to `BatteryStatusV1` but will be removed once the uploader is based on the v2 context.

### Review checklist

- [x] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/DatadogInternal/Context/DatadogContext.swift
+++ b/Sources/Datadog/DatadogInternal/Context/DatadogContext.swift
@@ -79,7 +79,9 @@ internal struct DatadogContext {
     /// not support telephony services.
     var carrierInfo: CarrierInfo?
 
-    /// `true` if the Low Power Mode is enabled.
+    /// The current mobile device battery status.
+    ///
+    /// This value can be `nil` of the current device battery interface is not available.
     var batteryStatus: BatteryStatus?
 
     /// `true` if the Low Power Mode is enabled.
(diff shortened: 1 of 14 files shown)


Sources/Datadog/DatadogInternal/Context/DatadogContext.swift
```diff
@@ -78,4 +78,10 @@ internal struct DatadogContext {
     /// This value can be `nil` of no service is currently registered, or if the device does
     /// not support telephony services.
     var carrierInfo: CarrierInfo?
+
+    /// `true` if the Low Power Mode is enabled.
+    var batteryStatus: BatteryStatus?
```
ncreated, line 83: The comment is wrong. It's worth explaining in the comment when to expect `nil` value, like we did for `networkConnectionInfo` and `carrierInfo` 👌.

ncreated (approved):
Looks good 👌, with one feedback on the `batteryStatus` property comment.

I really like how the publisher / reader architecture pays off here, great job 🏅.

Sources/Datadog/DatadogInternal/Context/DatadogContext.swift
maxep: Good catch! I've improved this doc and added some for the introduced reader and publisher.

PR #946, opened 2022-07-27
RUMM-2266 Add `make` commands for Session Replay models generation

ncreated:
### What and why?

📦 This is the last PR to SR models generation topic. It adds two `make` commands for working with SR models:

```
$ make sr-models-generate
$ make sr-models-verify
```

as a counterpart to:

```
$ make rum-models-generate
$ make rum-models-verify
```

### How?

Both commands interact with underlying `rum-models-generator` Swift CLI added earlier. The interaction is done through `run.py` Python script. 

I decided to replace existing Bash script with Python, because the logic become too complex for Bash and it should be more readable and maintainable in Python. Also, truth is that I'm not a Bash expert (neither Python TBH 🙃).

No changes to Swift CLI were required - the SPM tool is configurable enough.

I'm not yet plugging `make` commands into any CI workflow, as SR models have no reason for existence on `develop` branch. That will be done later on `feature/session-replay`.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


maxep (approved):
🚀

PR #963, opened 2022-08-09
RUMM-2350 Detect error tap furstration

maxep:
### What and why?

Add the `error_tap` frustration signal when a Tap action leads to one or more errors

### How?

Update models and add frustration type `error_tap` to action event when the action scope has one or more error.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMUserActionScopeTests.swift
+++ b/Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMUserActionScopeTests.swift
@@ -85,6 +85,7 @@ class RUMUserActionScopeTests: XCTestCase {
         XCTAssertEqual(recordedAction.service, "test-service")
         XCTAssertEqual(recordedAction.device?.name, "device-name")
         XCTAssertEqual(recordedAction.os?.name, "device-os")
+        XCTAssertNil(recordedAction.action.frustration)
     }
 
     func testGivenCustomSource_whenActionIsSent_itSendsCustomSource() throws {
@@ -263,6 +264,7 @@ class RUMUserActionScopeTests: XCTestCase {
         let event = try XCTUnwrap(writer.events(ofType: RUMActionEvent.self).last)
         XCTAssertEqual(event.action.resource?.count, 1, "User Action should track first successful Resource")
         XCTAssertEqual(event.action.error?.count, 1, "User Action should track second Resource failure as Error")
+        XCTAssertNil(event.action.frustration)
     }
 
     func testWhileContinuousUserActionIsActive_itCountsViewErrors() throws {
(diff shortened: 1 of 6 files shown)


Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMUserActionScopeTests.swift
```diff
@@ -577,4 +579,37 @@ class RUMUserActionScopeTests: XCTestCase {
         XCTAssertNil(writer.events(ofType: RUMActionEvent.self).first)
         XCTAssertFalse(callbackCalled)
     }
+
+    // MARK: - Actions with Frustrations
+
+    func testGivenTapUserActionWithError_itWritesErrorTapFrustration() throws {
```
ncreated, line 585: I think it might make sense to have 2 distinct tests:
* _**when** error occurs within the lifespan of an action, **then** it adds error tap frustration_
* _**when** error occurs later than the lifespan of an action, **then** it DOES NOT add error tap frustration_

This way we will assert both new behaviours explicitly and exclusively, instead of asserting "not" as a side-effect in other tests. Otherwise, IMHO the convention proposed in this PR isn't well sustainable - why to assert "not" in the 2 tests proposed, but not in other? There will be always more than one tests to assert "not".

ncreated (changes requested):
New logic looks good 👍, but I have some feedback on the convention we used to test it. I can see some cumbersome pattern starting in there 🙌.

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMUserActionScopeTests.swift
```diff
@@ -577,4 +579,37 @@ class RUMUserActionScopeTests: XCTestCase {
...
+        var currentTime = Date()
+        let scope = RUMUserActionScope.mockWith(
+            parent: parent,
+            actionType: .tap,
+            startTime: currentTime,
+            isContinuous: false
+        )
+
+        currentTime.addTimeInterval(0.05)
```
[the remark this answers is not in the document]
maxep: I intended to apply a small interval of 50ms between the action and the error, this is copied from a [test above](https://github.com/DataDog/dd-sdk-ios/blob/maxep/RUMM-2350/detect-frustration-error-tap/Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMUserActionScopeTests.swift#L483). I can do a percent of the timeout to make it clear we are in the action lifespan 👍

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMUserActionScopeTests.swift
maxep: Indeed, I've improve the coverage here. I'm also adding this one
- when error occurs on a non discrete action, then it DOES NOT add error tap frustration

ncreated (approved):
🚀 💪

PR #967, opened 2022-08-10
RUMM-2312 Remove v1 folder using default core

maxep:
### What and why?

Improve v1 data depreciation by running removal once.

### How?

`deprecated` directories are removed from the features configuration and are now deleted by the `Datadog` class at initialisation. This allow to perform deletion once by using the r/w queue from `defaultDatadogCore`, it also removes complexity of deleting v1 files as part of the `ConsentAwareDataWriter` migration process.

I took the opportunity to move files related to persistency under `DatadogCore/Storage`.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


Sources/Datadog/Datadog.swift
```diff
@@ -300,11 +300,24 @@ public class Datadog {
...
+    private static func deleteV1Folders(in core: DatadogCore) {
+        let deprecated = ["com.datadoghq.logs", "com.datadoghq.traces", "com.datadoghq.rum"].compactMap {
+            try? Directory.cache().subdirectory(path: $0) // ignore errors - deprecated paths likely do not exist
+        }
+
+        core.readWriteQueue.async {
+            // ignore errors
+            deprecated.forEach { try? FileManager.default.removeItem(at: $0.url) }
+        }
+    }
```
ncreated, line 319: question/ Just making sure - would this mean that V2 core will be aware of V1 feature folders, right? I think it's reasonable and trying to delete each folder from corresponding V2 feature would be too much hassle 👍.

ncreated (approved):
Looks good 👌, nice cleanup 👍.

Sources/Datadog/Datadog.swift
maxep: That's right! `Datadog` class is already aware of the v1 feature types, so I think it's fine. I'm not a big fan of hardcoded values out of nowhere like this, but it's a migration step that will ultimately be removed  🤷‍♂️ 

One thing to note is that it now deletes the folders, previously it was only removing the content of the folders.

PR #975, opened 2022-08-19
RUMM-2332 Span to Log through message bus

maxep:
### What and why?

Send Span log to Logs through the message bus.

### How?

`TracingWithLoggingIntegration` no longer write a `LogEvent` on the Logs event write, but build and send a message instead.

### Pattern Proposal

This PR introduces `DatadogExtension` and `DatadogExtended` which is a pattern from [Alamofire](https://github.com/Alamofire/Alamofire/blob/master/Source/AlamofireExtended.swift). This pattern allows to expose external libraries extensions hided behind a `dd` property (instance or static). It can be useful when sharing type extension through `DatadogInternal`. Here, it expose the extended property `name` of type `Foundation.Thread`
```swift
Thread.current.dd.name // returns the current thread name extended by Datadog
```

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/Tracing/Integrations/TracingWithLoggingIntegration.swift
+++ b/Sources/Datadog/Tracing/Integrations/TracingWithLoggingIntegration.swift
@@ -18,34 +18,41 @@ internal struct TracingWithLoggingIntegration {
         static let defaultErrorProperty = "Unknown"
     }
 
+    struct Configuration {
+        /// The `service` value for logs.
+        /// See: [Unified Service Tagging](https://docs.datadoghq.com/getting_started/tagging/unified_service_tagging).
+        let service: String?
+        /// The `logger.name` value for logs.
+        let loggerName: String
+        /// Whether to send the network info in `network.client.*` log attributes.
+        let sendNetworkInfo: Bool
+    }
+
     /// `DatadogCore` instance managing this integration.
     let core: DatadogCoreProtocol
     /// Builds log events.
-    let logBuilder: LogEventBuilder
+    let configuration: Configuration
 
     init(
         core: DatadogCoreProtocol,
-        context: DatadogV1Context,
-        tracerConfiguration: Tracer.Configuration,
-        loggingFeature: LoggingFeature
+        tracerConfiguration: Tracer.Configuration
     ) {
         self.init(
             core: core,
-            logBuilder: LogEventBuilder(
-                service: tracerConfiguration.serviceName ?? context.service,
+            configuration: .init(
+                service: tracerConfiguration.serviceName,
                 loggerName: "trace",
-                sendNetworkInfo: tracerConfiguration.sendNetworkInfo,
-                eventMapper: loggingFeature.configuration.logEventMapper
+                sendNetworkInfo: tracerConfiguration.sendNetworkInfo
             )
         )
     }
 
-    init(core: DatadogCoreProtocol, logBuilder: LogEventBuilder) {
+    init(core: DatadogCoreProtocol, configuration: Configuration) {
         self.core = core
-        self.logBuilder = logBuilder
+        self.configuration = configuration
     }
 
-    func writeLog(withSpanContext spanContext: DDSpanContext, fields: [String: Encodable], date: Date) {
+    func writeLog(withSpanContext spanContext: DDSpanContext, fields: [String: Encodable], date: Date, else fallback: @escaping () -> Void) {
         var userAttributes = fields
 
         // get the log message and optional error kind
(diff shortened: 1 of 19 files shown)


Sources/Datadog/Tracing/Integrations/TracingWithLoggingIntegration.swift
```diff
@@ -73,26 +80,23 @@ internal struct TracingWithLoggingIntegration {
...
+        core.send(
+            message: .custom(
+                key: "log",
+                attributes: [
+                    "date": date,
+                    "loggerName": configuration.loggerName,
+                    "service": configuration.service,
+                    "threadName": Thread.current.dd.name,
+                    "message": message,
+                    "level": level,
```
maxep, line 92: We are still sharing types such as `LogLevel` here, this dependency will eventually be removed.

Sources/Datadog/Tracing/Integrations/TracingWithLoggingIntegration.swift
```diff
@@ -73,26 +80,23 @@ internal struct TracingWithLoggingIntegration {
...
+                    "error": extractedError,
+                    "userAttributes": userAttributes,
+                    "internalAttributes": internalAttributes,
+                    "sendNetworkInfo": configuration.sendNetworkInfo
+                ]
+            ),
```
ncreated, line 98: Unlike in #974, the type we pass here through `message` is quite verbose and full of `String` keys written and read on both sides. I wonder if it wouldn't be practical from testing POV to introduce sort of DTO object to handle it in both ends, e.g.:
```swift
// in Tracing

struct TracingLog: ToFeatureMessageConvertible {
   let date: Date
   let loggerName: String
   let service: String
   // ...

   var toFeatureMessage: FeatureMessage { /* ... */ }
}

// in Logging:

struct TracingLog: FromFeatureMessageConvertible {
   let date: Date
   let loggerName: String
   let service: String
   // ...

   init(from featureMessage: FeatureMessage) throws { 
      guard ... else { // here, validation happens (e.g. check for particular keys existence)
         throw InternalError("The message received through bus doesn't match Logging contract")
      }
      /// ...
   }
}
```

This could give us the ability of easy integration testing and documenting message specs as separate types, all watched in tests. Somewhere in V2 integration tests we might then do:
```swift
@testable import DatadogTracing
@testable import DatadogLogging

func testWhenProducingTracingLogInTracing_thenItCanBeReadInLogging() {
   let messageProduced: DatadogTracing.TracingLog = .mockRandom()
   let messageRead = DatadogLogging.TracingLog(from: messageProduced.toFeatureMessage)

   XCTAssertEqual(messageProduced.date, messageRead.date)
   XCTAssertEqual(messageProduced.loggerName, messageRead.loggerName)
   // ...
}
```
We might highly leverage this kind of testing when message passing requires some data transformation. The throwing initializer in `FromFeatureMessageConvertible` would give us additional opportunity of using `DD.telemetry` and watching if contracts are met everywhere and all the time.

This is somehow addressed by the _incomplete_, _partial_, _complete_ and _rejected_ message tests we do in this PR, with a difference that current design is asymmetric (we test received message, without producing the real one).

Even if not this way but some other, I think the "produce" and "receive" message would strongly benefit from more testing flavours, consistent error throwing flow and telemetry coverage that could be maintained easily. Any thoughts @maxep?

ncreated (approved):
Looks very good - and I'm happy to see `dd` extension finally landing there ✈️ 🙌. I left one general feedback / suggestion on handling & testing EB messages on both sides (producer + receiver). I'm wondering what's your take on this @maxep - not a blocker, given that I don't yet have the full picture of V2 EB in all products.

Sources/Datadog/Tracing/Integrations/TracingWithLoggingIntegration.swift
maxep: That's a good point. I like the idea of being able to serialise/deserialise models on the bus, it would allow having data models defined in features with a simpler type check. Tho, I think we could leverage `Codable` for that instead of introducing new `Convertible` conformances. It would involve creating specific Encoder/Decoder for the bus, Features would then create their `Codable` objects and send/receive it on the bus. It's something I've already explored in my latest [oss (draft) project](https://github.com/maxep/swift-dynamic-object/tree/develop), and I think we could use a simple version of that with a pair of `AnyEncoder/AnyDecoder`.

About testing, the message 'producer' part is tested in `DDSpanTests` and `TracerTests. But indeed, it is lacking flavours for message failure on the bus: The bus is not able to say if the message failed because it's lacking a receiver, or because the message is malformed. This should be addressed so we can have valuable telemetry 👍

I will add both on dedicated PRs.

ncreated: My main concern is not that much about convenience, but rather clarity and tests of the contract between Features communication over the bus. I think this can be achieved in many ways, but at the highest level each may leverage most from introducing a type (_"When F1 sends `M`, then `F2` receives `M`"_), sort of DDD-approach.

But I like the idea of bringing convenience as well and what you propose with custom `Coder` is really interesting 👌. As long as we don't apply unnecessary perf penalty (like JSON-serialization would do), it's a valid option IMO 👍. Alternatively, we could think of code-generation to generate mirrored types for each message.

Maybe this topic is sth we should revisit when all bus communication is established (I still need to have a look at other PRs 👀).

PR #819, opened 2022-04-21
RUMM-2024 Replace internal monitoring with rum telemetry

maxep:
### What and why?

Replace Internal Monitoring with RUM Telemetry.

### How?

- `InternalMonitoring` feature is completely removed and replaced by the new `Telemetry` api.
- The `telemetry` instance is injected at initialisation of the SDK, no more access from singleton.
- RUM telemetry events do not allow custom attributes. Small attributes are now part of the events `message`, while large payloads have been removed (open to discussion).
- `KronosMonitor` has been removed: We have conclusive data to move forward. Also, adding telemetry to Kronos introduce a _chicken or egg dilemma_ (RUM Telemetry needs server sync)

### Note

This PR does not include a RUM Telemetry configuration such as sampling.

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] ~~Add CHANGELOG entry for user facing change.~~

PR #976, opened 2022-08-22
RUMM-2332 WebView tracking through message bus

maxep:
### What and why?

Send WebView events through the message bus.

### How?

Use the `FeatureMessageAttributes.AnyEncodable` type-erasure wrapper for passing `Encodable` objects through the bus. This wrapper is still inner to `FeatureMessageAttributes` but could be generalised if needed.

Logs and RUM are now able to read events from the bus and forward it to their events writers.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


Tests/DatadogTests/Datadog/RUM/WebView/WebRUMEventConsumerTests.swift
```diff
@@ -65,7 +65,7 @@ class WebRUMEventConsumerTests: XCTestCase {
 
     func testWhenValidWebRUMEventPassedWithoutRUMContext_itPassesToWriter() throws {
```
ncreated, line 66: convention/ All tests for this (and Logs) consumer are named with _"itPassesToWriter"_, which isn't now true. It should now say that respective message is passed to Core Event Bus.

Sources/Datadog/FeaturesIntegration/WebView/WebEventBridge.swift
```diff
@@ -51,22 +51,13 @@ internal class WebEventBridge {
...
-                    event: wrappedEvent,
-                    internalLog: (eventType == Constants.eventTypeInternalLog)
-                )
-            } else {
-                DD.logger.warn("A WebView log is lost because Logging is disabled in the SDK")
-            }
+        if eventType == Constants.eventTypeLog || eventType == Constants.eventTypeInternalLog {
+            try logEventConsumer.consume(
+                event: wrappedEvent,
+                internalLog: (eventType == Constants.eventTypeInternalLog)
```
ncreated, line 57: This makes me realise that `internalLog` is a leftover after https://github.com/DataDog/dd-sdk-ios/pull/819 and it leaves a dead path in the code, no? There is no use of `internalLog` bool inside this `.consume(event:internalLog:)` method as far as I can see. How about cleaning it up 🏕️?

ncreated (commented):
Looks good 👍. Although we don't yet know how WV Tracking will be designed in V2, this is a great step forward in isolating it from `LoggingFeature` and `RUMFeature` 🏅.

I left some feedback on convention and potential cleanup, LTM!

Sources/Datadog/FeaturesIntegration/WebView/WebEventBridge.swift
maxep: Ah yes, I noticed it but forgot to mention it. Browser SDK also migrated to RUM Telemetry, so we might no longer receive `internal_log` event type. I will double check with them and remove this 👍

ncreated (approved):
👌🙌

PR #977, opened 2022-08-22
RUMM-2332 Crash Reporting through message bus

maxep:
### What and why?

Send Crash Report events through the message bus.

### How?

The `FeatureV1Scope` now allows to bypass the current consent for writing events to the authorised writer. This can be used when the Feature  is already aware of the user's consent for the event it is about to write. This is true for Crash Reporting which keeps track of the user's consent in its crash-context

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


Sources/Datadog/CrashReporting/CrashReporter.swift
```diff
@@ -20,18 +20,24 @@ internal class CrashReporter {
...
 
         // If RUM rum is enabled prefer it for sending crash reports, otherwise use Logging feature.
-        if let rumFeature = rumFeature {
-            loggingOrRUMIntegration = CrashReportingWithRUMIntegration(rumFeature: rumFeature, context: context)
-        } else if let loggingFeature = loggingFeature {
-            loggingOrRUMIntegration = CrashReportingWithLoggingIntegration(loggingFeature: loggingFeature, context: context)
+        if let rumFeature = core.v1.feature(RUMFeature.self) {
+            loggingOrRUMIntegration = CrashReportingWithRUMIntegration(
+                core: core,
+                rumConfiguration: rumFeature.configuration,
```
ncreated, line 36: This will be still preventing us from calling `DatadogCrashReporting` a standalone module, right? Do we have any thoughts on removing this `rumConfiguration` dependency?

For what I see, the real dependency in `CrashReportingWithRUMIntegration` is on:
* `rumConfiguration.sessionSampler`
* `rumConfiguration.backgroundEventTrackingEnabled`
* `rumConfiguration.uuidGenerator`
* `rumConfiguration.applicationID`

ncreated (approved):
Looks great 👌. I really like the way `bypassConsent` is annotated for more clarity 👍.

Sources/Datadog/CrashReporting/CrashReporter.swift
maxep: Good catch, I can use parameters instead 👍 

We will still have feature coupling with the events types: `LogEvent`, `RUMErrorEvent`, and `RUMViewEvent` are still being shared.

ncreated: For events it's IMO fine - it's one category of code (and it's auto-generated). I'm wondering though how we will share these other things (except `applicationID` passed thru context, others don't have clear answer).

PR #980, opened 2022-08-25
Create core logic for rage tap identification

plousada:
### What and why?

Introduces core logic to transform taps into tap chains and from there identify rage taps according to certain constrains. 
It's intended to support chains on multiple objects at the same time, but only 1 chain per object

There's a few things that are still missing:
- Handling the current expiration restrictions around the age of events in `RumUserActionScope`
- Support for tvOS
- Unit Tests
- Code clean up
- Clarification around exposing API to disable feature or not

### How?

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/RUM/FrustrationSignals/RageTapManager.swift
+++ b/Sources/Datadog/RUM/FrustrationSignals/RageTapManager.swift
@@ -1,0 +1,154 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+import UIKit
+
+
+internal class RageTapManager{
+    struct Constants {
+        static let rageTapWindowTimeout : TimeInterval = 1 // 1 second
+        static let maximumTouchDistance : Double = 48*2  // ~9mm
+        static let minimumRageClicksPerSecond = 3
+    }
+    
+    class RageTapChain {
+        var lastTapPosition = CGPoint()
+        var actionCommands = [RUMAddUserActionCommand]()
+        var chainTimeout : DispatchWorkItem?
+        var workQueue : DispatchQueue?
+        
+    }
+    
+    var rageTapChainPerUIElement = [ObjectIdentifier:RageTapChain]()
+    var identifyRageTaps = true //TODO - Add API/property to change this
+    
+    
+    func canAcceptTapIntoChain(chain: RageTapChain, tapLocation: CGPoint) -> Bool{
+        if chain.actionCommands.count == 0 {
+            return true
+        }
+        
+        let sqDistanceX = (chain.lastTapPosition.x - tapLocation.x) * (chain.lastTapPosition.x - tapLocation.x)
+        let sqDistanceY = (chain.lastTapPosition.y - tapLocation.y) * (chain.lastTapPosition.y - tapLocation.y)
+
+        if (sqDistanceX + sqDistanceY) <= (RageTapManager.Constants.maximumTouchDistance * RageTapManager.Constants.maximumTouchDistance){
+            return true;
+        }
+        
+        return false
+    }
+    
+    func tryToFinalizeChain(identifier: ObjectIdentifier){
+        guard let chain = rageTapChainPerUIElement[identifier] else {
+            return
+        }
+        
+        guard let subscriber = subscriber else {
+            return
+        }
+        
+        var tapsWithinTimeout = [TimeInterval]()
+        
+        for tap in chain.actionCommands
+        {
+            tapsWithinTimeout.append(tap.time.timeIntervalSince1970)
+
+            if(tapsWithinTimeout.count == RageTapManager.Constants.minimumRageClicksPerSecond){
+                guard let firstTapTime = tapsWithinTimeout.first?.toInt64Milliseconds else {
+                    break
+                }
+                guard let lastTapTime = tapsWithinTimeout.last?.toInt64Milliseconds else {
+                    break
+                }
+                
+                let timeDiff = lastTapTime - firstTapTime
+                if (timeDiff < RageTapManager.Constants.rageTapWindowTimeout.toInt64Milliseconds)
+                {
+                    // Create rage tap at the last tap of the chain,
+                    // even though we already had matching conditions before that
+                    guard var lastCommand = chain.actionCommands.last else {
+                        break
+                    }
+                    lastCommand.isRage = true
+
+                    subscriber.process(command: lastCommand)
+                    return
+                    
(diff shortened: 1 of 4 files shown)


Sources/Datadog/RUM/FrustrationSignals/RageTapManager.swift
```diff
@@ -0,0 +1,154 @@
...
+    
+    func addCommandIntoChain(command:RUMAddUserActionCommand, chain: RageTapChain, touch: UITouch){
+        chain.actionCommands.append(command)
+        chain.lastTapPosition = touch.location(in:nil)
+    }
+    
+    func startTimeoutTimerForChain(chain: RageTapChain, identifier:ObjectIdentifier){
+        let workItem = DispatchWorkItem { self.tryToFinalizeChain(identifier: identifier) }
+        chain.chainTimeout = workItem
+        DispatchQueue.global().asyncAfter(deadline: .now() + RageTapManager.Constants.rageTapWindowTimeout, execute: workItem)
```
ncreated, line 102: Not really a fan of timer here as it comes with some problems:
* We dispatch the `workItem` to another queue (`.global()`), meaning that `tryToFinalizeChain(identifier:)` will run on another thread. That method relies a lot on the internal state of `RageTapManager` (e.g. mutable `RageTapChain`), so it will eventually crash because the same state is accessed on the main thread in `processActionCommand(command:)`. To make it thread safe, we'd need to add dedicated `queue` for synchronising `RageTapManager`. That would mean introducing at worst 2 more threads to the SDK.
* Timers are painful in tests. Rage taps unit tests would have to await enough time to fulfil the expectation, eventually introducing flakiness.

ncreated (changes requested):
I think the `UIKitRUMUserActionsHandler` is the right place to hook in for rage taps detection 👌. However, I wonder if we're not reinventing the wheel too much by defining _"tap window"_, _"touch distance"_ and _"min. rage clicks per second"_ to then compute it by our own.

Knowing how iOS manages [`UITouch`](https://developer.apple.com/documentation/uikit/uitouch) objects, I wonder if we could rely more on its [`.tapCount`](https://developer.apple.com/documentation/uikit/uitouch/1618132-tapcount) property:

> `UITouch.tapCount`
> (...) Use this property to evaluate whether the user single-tapped, double-tapped, or even triple-tapped a particular view or window.

Wouldn't this be an option check `.tapCount` in our [`UITouchCommandFactory`](https://github.com/DataDog/dd-sdk-ios/blob/927e1f014f8c729c1fc70bc7578038c5e33f28ff/Sources/Datadog/RUM/Instrumentation/Actions/UIKit/UIKitRUMUserActionsHandler.swift#L93-L118) and then set up `RUMAddUserActionCommand` accordingly if `.tapCount >= 3` 🤔💭?

Rationale is that iOS already does the heavy lifting for us:
- if succeeding touches happen to the same `UIControl` (e.g. a button) and they occur close to each other within short time window, it abstracts them as a single `UITouch` object with multiple `.tapCount`;
- the way OS does it is already natural to iOS devices, meaning it takes screen density vs touch radius, time relationship and (perhaps) some other elements into account;

Going that way, we couldn't use our custom definition of _"tap window"_ or _"touch distance"_ and we'd need to follow the way Apple does it. However, using platform-specific defaults is only for good IMO - especially if it can save us work, maintenance and bring more natural experience. WDYT @plousada @maxep ?

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMUserActionScope.swift
```diff
@@ -93,16 +97,21 @@ internal class RUMUserActionScope: RUMScope, RUMContextProvider {
...
 
         lastActivityTime = command.time
         switch command {
         case is RUMStopViewCommand:
             sendActionEvent(completionTime: command.time, on: command, context: context, writer: writer)
             return false
+        case let command as RUMAddUserActionCommand:
+            isRage = command.isRage
+            sendActionEvent(completionTime: command.time, on: command, context: context, writer: writer)
+            return false
```
maxep, line 114: The `RUMAddUserActionCommand` shouldn't be processed by an already opened action-scope. Instead, the `isRage` can be set when initialising the discrete action scope from the `RUMViewScope`.

Sources/Datadog/RUM/Instrumentation/Actions/UIKit/UIKitRUMUserActionsHandler.swift
```diff
@@ -37,6 +38,7 @@ internal class UIKitRUMUserActionsHandler: UIEventHandler {
 
     func publish(to subscriber: RUMCommandSubscriber) {
         self.subscriber = subscriber
+        rageActionManager.subscriber = subscriber
     }
```
maxep, line 42: We can remove the `subscriber` and let the `RageTapManager` handle all commands.

Sources/Datadog/RUM/RUMMonitor/RUMCommand.swift
```diff
@@ -298,6 +299,8 @@ internal struct RUMStopUserActionCommand: RUMUserActionCommand {
 internal struct RUMAddUserActionCommand: RUMUserActionCommand {
     var time: Date
     var attributes: [AttributeKey: AttributeValue]
+    var isRage = false
+    var targetView : UIView? = nil // Most likely target for this action
```
maxep, line 303: Is `targetView` necessary? I would prefer not to add dependency to `UIKit` here.

Sources/Datadog/RUM/FrustrationSignals/RageTapManager.swift
```diff
@@ -0,0 +1,154 @@
...
+        static let rageTapWindowTimeout : TimeInterval = 1 // 1 second
+        static let maximumTouchDistance : Double = 48*2  // ~9mm
+        static let minimumRageClicksPerSecond = 3
+    }
+    
+    class RageTapChain {
+        var lastTapPosition = CGPoint()
+        var actionCommands = [RUMAddUserActionCommand]()
+        var chainTimeout : DispatchWorkItem?
+        var workQueue : DispatchQueue?
```
maxep, line 22: This can be removed? I don't see it being used.

maxep (changes requested):
I agree with @ncreated about leveraging `.tapCount` instead of custom time frame and touch frame. `UITouch` is already abstracting that for us, and it worth testing 👍

Sources/Datadog/RUM/RUMMonitor/RUMCommand.swift
maxep: @ncreated gave a more detailed comment on this one :)

maciejburda:
It's been open for a while now, seems like we're not going to proceed with the change. Shall we close it?

@ncreated @maxep @plousada

PR #981, opened 2022-08-29
RUMM-2334 Use DatadogContext in V1 scopes

maxep:
### What and why?

While working on Feature communication through Core Context, I realised that we cannot apply synchronisation while maintaining 2 contexts. It is time to fade away `DatadogV1Context` in favor of `DatadogContext`.

#### Date Provider

The `DateProvider` is no longer part of context, it is now injected as part of the configuration, the goal is to let feature create a timestamp outside of the Event Write Context. The configuration will be revisited when doing the split.

#### Persisting `DatadogV1Context` Usage

`DatadogV1Context` still exists for Crash-Reporting and WebView-Tracking integrations which do not use the Event Write Context.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [x] Run smoke tests


Sources/Datadog/DatadogInternal/Context/DatadogContext.swift
```diff
@@ -74,7 +74,7 @@ internal struct DatadogContext {
     /// Represents the current state of the device network connectivity and interface.
     /// The value can be `unknown` if the network interface is not available or if it has not
     /// yet been evaluated.
-    var networkConnectionInfo: NetworkConnectionInfo = .unknown
+    var networkConnectionInfo: NetworkConnectionInfo?
```
maxep, line 77: Making this optional to keep previous behavior.

Sources/Datadog/RUMMonitor.swift
```diff
@@ -597,19 +596,15 @@ public class RUMMonitor: DDRUMMonitor, RUMCommandSubscriber {
...
 
     func process(command: RUMCommand) {
-        guard let scope = core.v1.scope(for: RUMFeature.self) else {
-            return
-        }
-
-        queue.async {
-            let transformedCommand = self.transform(command: command)
+        core.v1.scope(for: RUMFeature.self)?.eventWriteContext { context, writer in
+            self.queue.sync {
```
maxep, line 600: All 3 Features now perform `sync` on the event write context. This is not set in stone, and I'm exploring ways of synchronising Features with the shared context.

ncreated (commented):
Looks great in overall 👌. I feel that the `V1Context` layer we added now pays off in this migration 🙌.

The main concern I have is about performing `queue.sync {}` in Features, but I believe this is exactly the topic we're now exploring, right?

Sources/Datadog/RUMMonitor.swift
ncreated: 👍 Just submitted my review on that 🙂. Definitely, we need to explore the impact of this in high load situations. I'll think of possible solutions as well.

Sources/Datadog/RUM/RUMEvent/RUMDeviceInfo.swift
```diff
@@ -7,6 +7,10 @@
 import Foundation
 
 extension RUMDevice {
+    init(context: DatadogContext) {
+        self.init(device: context.device)
+    }
+
     init(context: DatadogV1Context) {
         self.init(device: context.device)
     }
```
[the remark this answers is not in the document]
maxep: Yes.. crash reporting and web view tracking. Both integration are still using v1 context

Sources/Datadog/Logging/RemoteLogger.swift
```diff
@@ -115,7 +108,7 @@ internal final class RemoteLogger: LoggerProtocol {
             var internalAttributes: [String: Encodable] = [:]
             var userTags: Set<String> = []
 
-            queue.sync {
+            self.queue.sync {
```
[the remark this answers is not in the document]
maxep: Yes! still exploring solution. It will be change in future PR.

Sources/Datadog/DatadogCore/Context/DatadogContextProvider.swift
```diff
@@ -47,7 +47,7 @@ internal final class DatadogContextProvider {
     /// to make writes exclusive.
     private let queue = DispatchQueue(
         label: "com.datadoghq.core-context",
-        attributes: .concurrent
+        qos: .utility
```
ncreated, line 50: Given it is serial now, it no longer makes sense to do `.barrier` in following code, no?
```swift
func write(block: @escaping (inout DatadogContext) -> Void) {
   queue.async(flags: .barrier)
}
```

maxep: Indeed, `.barrier` will have no effect on the serial queue, but I would like to keep it for now and move back to concurrency with a better threading design later.

ncreated (approved):

PR #986, opened 2022-09-02
RUMM-2429 [SR] Add TextField, ImageView and UISwitch semantics

ncreated:
### What and why?

📦 Supplementing #985, this PR adds 3 more semantics to session replay:
* `UITextField`
* `UIImageView`
* `UISwitch`

### How?

I leverated the `NodeRecorder` + `NodeSemantics` framework introduced in #985 by adding 3 specialised node recorders: `UITextFieldRecorder`, `UIImageViewRecorder` and `UISwitchRecorder`.

It required some adjustments to `NodeSemantics.importance` to ensure that new node semantics are resolved consistently by `ViewTreeSnapshotBuilder`. More explanation in inline comment.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/session-replay/Sources/DatadogSessionReplay/Processor/Flattening/NodesFlattener.swift
+++ b/session-replay/Sources/DatadogSessionReplay/Processor/Flattening/NodesFlattener.swift
@@ -18,8 +18,8 @@ internal struct NodesFlattener {
         var flattened: [Node] = []
 
         dfsVisit(startingFrom: snapshot.root) { nextNode in
-            // Skip nodes with no semantics:
-            if nextNode.semantics.importance > InvisibleElement.importance {
+            // Skip invisible nodes:
+            if !(nextNode.semantics is InvisibleElement) {
                 // When accepting nodes, remove ones that are covered by another opaque node:
                 flattened = flattened.compactMap { previousNode in
                     let isPreviousNodeCovered = nextNode.viewAttributes.frame.contains(previousNode.viewAttributes.frame)
--- a/session-replay/Sources/DatadogSessionReplay/Recorder/SnapshotBuilder/NodeRecorders/UITextFieldRecorder.swift
+++ b/session-replay/Sources/DatadogSessionReplay/Recorder/SnapshotBuilder/NodeRecorders/UITextFieldRecorder.swift
@@ -1,0 +1,106 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import UIKit
+
+internal struct UITextFieldRecorder: NodeRecorder {
+    func semantics(of view: UIView, with attributes: ViewAttributes, in context: ViewTreeSnapshotBuilder.Context) -> NodeSemantics? {
+        guard let textField = view as? UITextField else {
+            return nil
+        }
+
+        guard attributes.isVisible else {
+            return InvisibleElement.constant
+        }
+
+        // TODO: RUMM-2459
+        // Explore other (better) ways of infering text field appearance:
+
+        var editorProperties: UITextFieldWireframesBuilder.EditorFieldProperties? = nil
+        // Lookup the actual editor field's view in `textField` hierarchy to infer its appearance.
+        // Perhaps this can be do better by infering it from `UITextField` object in RUMM-2459:
+        dfsVisitSubviews(of: textField) { subview in
+            if subview.bounds == textField.bounds {
+                editorProperties = .init(
+                    backgroundColor: subview.backgroundColor?.cgColor,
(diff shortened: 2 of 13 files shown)


session-replay/Sources/DatadogSessionReplay/Recorder/SnapshotBuilder/ViewTreeSnapshot.swift
```diff
@@ -176,14 +176,14 @@ internal struct InvisibleElement: NodeSemantics {
 /// A semantics of an UI element that is of `UIView` type. This semantics mean that the element has visual appearance in SR, but
 /// it will only utilize its base `UIView` attributes. The full identity of the node will remain ambiguous if not overwritten with `SpecificElement`.
 internal struct AmbiguousElement: NodeSemantics {
-    static let importance: Int = 1
+    static let importance: Int = 0
     let wireframesBuilder: NodeWireframesBuilder?
 }
```
ncreated, line 181: This makes `InvisibleElement.importance` and `AmbiguousElement.importance` equal (`0`).

**Rationale:** 
* The `AmbiguousElement` semantics is recorded by `UIViewRecorder`, which doesn't know the specific `UIView` subclass (that one is recorded later by specialised node recorders). 
* The `InvisibleElement` semantics can be recognised by any node recorder. In `UIViewRecorder` it is determined by `view.isHidden` and `view.alpha`, whereas in specialised recorders this is paired with high level aspects of specialised view, e.g. empty text in `UILabel` (which makes the node effectively invisible, even if `.alpha > 0` and `.isHidden == false`).

Because `UIViewRecorder` is queried first, the importance of `InvisibleElement` cannot be lower than importance of `AmbiguousElement`. Otherwise, specialised recorders wouldn't be able to override `AmbiguousElement` with `InvisibleElement`.

ncreated: 💡 Perhaps we could go with no `importance` and record an array of semantics that will be further resolved in `Processor`:
```
struct Node {
   /// Different semantics recorded for this node.
   let semantics: [NodeSemantics]
}
```
I'd leave this idea for later - might come handy if we find `.importance` limited for more specialised views and complex hierarchies (e.g. `NodeRecorders` clashing with each other).

maxep: I think it is fine to have same importance, it's like priority value 👍

maxep: Or maybe the `InvisibleElement.importance` can be `-1`?

session-replay/Sources/DatadogSessionReplay/Recorder/SnapshotBuilder/NodeRecorders/UITextFieldRecorder.swift
```diff
@@ -0,0 +1,106 @@
...
+                )
+            }
+        }
+
+        let text: String = {
+            guard let textFieldText = textField.text, !textFieldText.isEmpty else {
+                return textField.placeholder ?? ""
+            }
+            return textFieldText
+        }()
```
maxep, line 41: /nit I kind of like the optional mapping thing, not a change request, up to you :)
```suggestion
        let text = textField.text.flatMap { $0.isEmpty ? textField.placeholder : $0 } ?? ""
```

maxep (approved):
It looks great 👏 👏 
I left a couple of suggestions but it's good to merge as is, well done!!

session-replay/Sources/DatadogSessionReplay/Processor/Flattening/NodesFlattener.swift
```diff
@@ -18,8 +18,8 @@ internal struct NodesFlattener {
         var flattened: [Node] = []
 
         dfsVisit(startingFrom: snapshot.root) { nextNode in
-            // Skip nodes with no semantics:
-            if nextNode.semantics.importance > InvisibleElement.importance {
+            // Skip invisible nodes:
+            if !(nextNode.semantics is InvisibleElement) {
```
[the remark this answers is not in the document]
ncreated: Could be an option 👍, but I will rather wait for more semantics to be added (handling other UI elements), as things might change in other direction too.

session-replay/Sources/DatadogSessionReplay/Recorder/SnapshotBuilder/NodeRecorders/UITextFieldRecorder.swift
ncreated: 👌 Noted - but if it's not a blocker, I'll keep existing code 🙂. I'd be in favour of syntax sugar if there is more similar code around and wrapping it into simple & repetitive pattern would improve readability.

PR #988, opened 2022-09-06
RUMM-2334 Logging with RUM and Tracing Through Core Context

maxep:
### What and why?

Log reads RUM and Tracing context from core.

### How?

The `DatadogCoreProtocol` now expect a method with the following signature:
```swift
func set(feature: String, attributes: @escaping () -> FeatureMessageAttributes)
```

The method will update a Feature attributes on the Core Context. The attributes autoclosure will be invoked on the context thread.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


Sources/Datadog/RUM/RUMTelemetry.swift
```diff
@@ -18,43 +18,34 @@ internal final class RUMTelemetry: Telemetry {
...
+    private var currentSessionID: String?
 
     /// Keeps track of event's ids recorded during a user session.
     private var eventIDs: Set<String> = []
 
+    /// Queue for processing RUM Telemetry
+    private let queue = DispatchQueue(
+        label: "com.datadoghq.rum-telemetry",
+        target: .global(qos: .utility)
+    )
```
maxep, line 34: `RUMTelemetry` no longer depend on the global `RUMMonitor` and its queue. This new queue sync and serialise telemetry updates.

Sources/Datadog/RUMMonitor.swift
```diff
@@ -607,6 +608,25 @@ public class RUMMonitor: DDRUMMonitor, RUMCommandSubscriber {
...
+                return [:]
+            }
+
+            return [
+                "application_id": applicationContext.rumApplicationID,
+                "session_id": sessionContext.sessionID.rawValue.uuidString.lowercased(),
+                "view.id": activeViewContext?.activeViewID?.rawValue.uuidString.lowercased(),
+                "user_action.id": activeViewContext?.activeUserActionID?.rawValue.uuidString.lowercased()
+            ]
+        })
```
maxep, line 629: After processing the `eventWriteContext`, we can set the context attributes for RUM. `self.queue` ensure serialisation of events.

Tests/DatadogTests/Datadog/LoggerTests.swift
```diff
@@ -505,38 +505,6 @@ class LoggerTests: XCTestCase {
         )
     }
 
-    func testGivenBundlingWithRUMEnabledButRUMMonitorNotRegistered_whenSendingLog_itPrintsWarning() throws {
```
maxep, line 508: Logger does not depend on `RUMMonitor` anymore and we cannot check for its initialisation through integration.

Tests/DatadogTests/Datadog/LoggerTests.swift
```diff
@@ -607,45 +575,15 @@ class LoggerTests: XCTestCase {
...
         logMatchers[0].assertValue(
-            forKeyPath: LoggingWithActiveSpanIntegration.Attributes.spanID,
+            forKeyPath: "dd.span_id",
             equals: "\(span.context.dd.spanID.rawValue)"
         )
-        logMatchers[1].assertNoValue(forKey: LoggingWithActiveSpanIntegration.Attributes.traceID)
-        logMatchers[1].assertNoValue(forKey: LoggingWithActiveSpanIntegration.Attributes.spanID)
-    }
-
-    func testGivenBundlingWithTraceEnabledButTracerNotRegistered_whenSendingLog_itPrintsWarning() throws {
```
maxep, line 621: Same with `Tracer`

Sources/Datadog/Logging/Log/LogEventSanitizer.swift
```diff
@@ -13,13 +13,8 @@ internal struct LogEventSanitizer {
...
             "host", "message", "status", "service", "source", "ddtags",
-            TracingWithLoggingIntegration.TracingAttributes.traceID,
-            TracingWithLoggingIntegration.TracingAttributes.spanID,
-            RUMContextIntegration.Attributes.applicationID,
-            RUMContextIntegration.Attributes.sessionID,
-            RUMContextIntegration.Attributes.viewID,
-            LoggingWithActiveSpanIntegration.Attributes.traceID,
-            LoggingWithActiveSpanIntegration.Attributes.spanID,
+            "dd.trace_id", "dd.span_id",
+            "application_id", "session_id", "view.id", "user_action.id",
```
ncreated, line 17: I'm a bit afraid of switching from compiled constants to string literals. This way we lose compiler check on these values and it is easy to get out of sync and cause problems 🐞 (e.g. changing their value in definition, but not here). Also, we're losing self-documenting aspect of referencing constants in other types.

On the other hand, I understand that in modular setup, such references aren't possible. Do you have any thoughts on this @maxep ? By design, such things should be tested as integration between features, however this is a good example where such tests will be very hard to establish.

Sources/Datadog/Logging/RemoteLogger.swift
```diff
@@ -37,18 +37,18 @@ internal final class RemoteLogger: LoggerProtocol {
...
     /// Integration with RUM. It is used to correlate Logs with RUM events by injecting RUM context to `LogEvent`.
-    /// Can be `nil` if the integration is disabled for this logger or if RUM feature is disabled.
-    internal let rumContextIntegration: LoggingWithRUMContextIntegration?
+    /// Can be `false` if the integration is disabled for this logger or if RUM feature is disabled.
+    internal let rumContextIntegration: Bool
     /// Integration with Tracing. It is used to correlate Logs with Spans by injecting `Span` context to `LogEvent`.
-    /// Can be `nil` if the integration is disabled for this logger or if Tracing feature is disabled.
-    internal let activeSpanIntegration: LoggingWithActiveSpanIntegration?
+    /// Can be `false` if the integration is disabled for this logger or if Tracing feature is disabled.
+    internal let activeSpanIntegration: Bool
```
ncreated, line 44: Feels like the _"or if (...) feature is disabled_" part of these comments are not true anymore, no?

Sources/Datadog/RUM/RUMTelemetry.swift
```diff
@@ -105,53 +99,58 @@ internal final class RUMTelemetry: Telemetry {
...
-            }
-
-            // record up de `MaxEventsPerSessions`, discard duplicates
-            if self.eventIDs.count < RUMTelemetry.MaxEventsPerSessions, !self.eventIDs.contains(id) {
-                self.eventIDs.insert(id)
-                operation(context, writer)
+            let attributes = context.featuresAttributes["rum"]
+            let sessionId = attributes?["session_id", type: String.self]
+
+            self.queue.sync {
```
ncreated, line 143: Do we need `.sync {}` here 🤔💭? I can't see why it can be `.async {}` in `RemoteLogger`, but `.sync {}` here. Hope it's a leftover 🙏🙂, no?

Sources/Datadog/RUMMonitor.swift
```diff
@@ -607,6 +608,25 @@ public class RUMMonitor: DDRUMMonitor, RUMCommandSubscriber {
...
+        // update the core context with rum context
+        core.set(feature: "rum", attributes: self.queue.sync {
+            let applicationContext = self.applicationScope.context
+            let sessionContext = self.applicationScope.sessionScope?.context
+            let activeViewContext = self.applicationScope.sessionScope?.viewScopes.last?.context
+
+            guard let sessionContext = sessionContext, sessionContext.sessionID != .nullUUID else {
+                // if Session was sampled or not yet started
+                return [:]
+            }
```
ncreated, line 621: Would it be possible to reuse [`RUMCurrentContext`](https://github.com/DataDog/dd-sdk-ios/blob/21fa77710f6fb6b0d8dfd17b24eec817163c57ae/Sources/Datadog/RUM/RUMContext/RUMCurrentContext.swift#L10) somehow in there? It does similar computation, with slight differences (it leverages `??`) and is already fully tested.

Sources/Datadog/RUMMonitor.swift
```diff
@@ -607,6 +608,25 @@ public class RUMMonitor: DDRUMMonitor, RUMCommandSubscriber {
                 }
             }
         }
+
+        // update the core context with rum context
+        core.set(feature: "rum", attributes: self.queue.sync {
```
ncreated, line 613: Because `process(command:)` is called from the user thread, `queue.sync {}` here will be blocking it. This is due to race conditions we have discovered in SDK context and that we have an open RFC on, right 🙂?

ncreated (commented):
Looks very good 👌. I left some questions and smaller feedbacks.

Sources/Datadog/RUMMonitor.swift
maxep: The `attributes` from `core.set(feature:.attributes:)` is an `@escaping @autoclosure` that is called from the context queue. This syntax is very confusing I admit... 2 options:
1. explicit `@autoclosure`:
```swift
core.set(feature: "rum", attributes: { 
   return self.queue.sync {
        ...
    }
}())
```
2. Remove `@autoclosure`, making `@escaping` explicit.
```swift
core.set(feature: "rum", attributes: { 
    return self.queue.sync {
        ...
    }
})
```
Closure will have always be explicit.

That second will read better, WDYT?

Sources/Datadog/RUM/RUMTelemetry.swift
maxep: Good catch!

Sources/Datadog/Logging/Log/LogEventSanitizer.swift
maxep: Yes, I tend to remove any dependency between modules. But you are right, any change in the emitted context won't be catch by tests, the reserved attributes sanitisation is not currently tested.

But we can still test this as an integration test and check against constant defined in the emitting Feature, let me draft something!

Sources/Datadog/RUMMonitor.swift
maxep: I was planning on removing `RUMCurrentContext`, so the `queue.sync` always appears in the queue's owner (= RUMMonitor) and not down the stream. But I could move this logic to a dedicated private method, WDYT?

maxep: Also, `RUMContextProvider` is still there because of WebView integration. But I plan on removing it once it's connected to the message bus.

Sources/Datadog/Logging/Log/LogEventSanitizer.swift
maxep: @ncreated I've covered this in tests, by creating constants in each Features and then test sanitisation of logs against those constants. It will catch any change made on the attributes.

Sources/Datadog/RUMMonitor.swift
ncreated: > I was planning on removing RUMCurrentContext, so the queue.sync always appears in the queue's owner

Got it, sounds good as long as we're aware of it and will revamp it later 👍. And we won't miss this change as compiler will spot it when extracting WVs integration later. All good 👌.

Sources/Datadog/RUMMonitor.swift
ncreated: Oh, definitely it was too misleading, so I was trapped. Yes, `@autoclosure` removal we did looks good 👍.

ncreated (approved):
Well done 👍. Most importantly, the `set(feature:attributes:)` API is better to use w/o `@autoclosure`.

PR #989, opened 2022-09-06
RUMM-2334 Tracing with RUM Through Core Context

maxep:
### What and why?

Tracing gets RUM context from the message bus.

### How?

We can now add `DatadogContextObserver` to the `DatadogContextProvider` to get notified when the context changes. The `DatadogCore` is an observer and send a context message on the bus on any context change.

**Note:** The context still have passive values that are read when the context is requested. To get the most up-to-date values we notify the observers on each read operation as well. This was already the case in V1 providers, but this needs to be addressed so any value change generate a push on the message-bus.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


Sources/Datadog/DatadogInternal/MessageBus/FeatureMessage.swift
```diff
@@ -27,4 +27,6 @@ public enum FeatureMessage {
         key: String,
         baggage: FeatureBaggage
     )
+
+    case context(DatadogContext)
```
ncreated, line 31: nit: it's missing a doc comment like we do for other cases

Sources/Datadog/Tracing/TracingV2Configuration.swift
```diff
@@ -49,3 +49,32 @@ internal struct TracingRequestBuilder: FeatureRequestBuilder {
...
+        guard
+            let tracer = Global.sharedTracer as? Tracer,
+            let integration = tracer.rumIntegration
+        else {
+            return false
+        }
+
+        integration.attribues = context.featuresAttributes["rum"]?.all()
+        return true
+    }
```
ncreated, line 79: The method comment here is not clear IMO. It tells about RUM error and RUM view, but nothing in the implementation seems to address this. What do we mean here?

Tests/DatadogTests/Datadog/TracerTests.swift
```diff
@@ -663,53 +663,43 @@ class TracerTests: XCTestCase {
...
-        let span = tracer.startSpan(operationName: "operation", tags: [:], startTime: Date())
+        let span = Global.sharedTracer.startSpan(operationName: "operation", tags: [:], startTime: Date())
         span.finish()
 
         // then
-        try XCTAssertTrue(
-            XCTUnwrap(dd.logger.warnLog?.message)
-                .contains("RUM feature is enabled, but no `RUMMonitor` is registered. The RUM integration with Tracing will not work.")
-        )
-
```
ncreated, line 708: As we no longer assert on this, the test name needs to be updated because it still refers to printing warning:
```swift
testGivenBundlingWithRUMEnabledButRUMMonitorNotRegistered_whenSendingSpan_itPrintsWarning
```

Sources/Datadog/Tracing/Integrations/TracingWithRUMIntegration.swift
```diff
@@ -0,0 +1,30 @@
...
+    private let lock = NSRecursiveLock()
+
+    /// Unsafe attributes.
+    private var _attribues: [String: Encodable]?
+
+    private func synchronize<T>(_ block: () -> T) -> T {
+        lock.lock()
+        defer { lock.unlock() }
+        return block()
+    }
```
ncreated, line 29: Do we know how `NSLock` perf compares to `DispatchQueue`? Do we have plans on wider use of _locks_ vs _queues_ for other products in similar use case (decorating one Feature events with other Feature context)?

Here we will be running this critical section for each `tracer.startSpan()` (on **user thread**), so overhead is important. If we follow this pattern, similar thing will happen in other Features.

Sources/Datadog/DatadogCore/Context/DatadogContextProvider.swift
```diff
@@ -71,11 +80,17 @@ internal final class DatadogContextProvider {
...
+        reader.read(to: &self.context)
         return context
     }
 
+    /// Publishes context changes to tge given receiver.
+    ///
+    /// - Parameter receiver: The receiver closure.
+    func publish(to receiver: @escaping ContextValueReceiver<DatadogContext>) {
+        queue.async(flags: .barrier) { self.receivers.append(receiver) }
+    }
```
ncreated, line 92: It doesn't seem to be tested in [`DatadogContextProviderTests`](https://github.com/DataDog/dd-sdk-ios/blob/maxep/RUMM-2334/tracing-with-rum-context/Tests/DatadogTests/Datadog/DatadogCore/DatadogContextProviderTests.swift).

ncreated (changes requested):
In overall it looks good 👍. I left few feedbacks on remaining leftovers and missing tests coverage for context provider.

Sources/Datadog/Tracing/TracingV2Configuration.swift
maxep: Sorry, c/p here :)

Tests/DatadogTests/Datadog/TracerTests.swift
maxep: The entire test should be removed actually, we do the same assertion above.

Sources/Datadog/Tracing/Integrations/TracingWithRUMIntegration.swift
maxep: Yes, here a benchmark: https://www.vadimbulavin.com/benchmarking-locking-apis/

For read operation: `DispatchQueue` is 7-8 times slower than locks.


For write operation: `DispatchQueue` is 3-4 times slower than locks


`DispatchQueue` performs better when using concurrent queue tho. But prior to this change, the `Tracer` was doing a `sync` on the (serial) queue of the `RUMMonitor` from the **user thread** ([here](https://github.com/DataDog/dd-sdk-ios/blob/develop/Sources/Datadog/Tracer.swift#L203)).

> Do we have plans on wider use of locks vs queues for other products in similar use case (decorating one Feature events with other Feature context)?

No, it's on a case by case. Here I'm using a lock to avoid syncing the `Tracer\queue` from the user-thread. But it really depends on the Feature pipeline.

PR #990, opened 2022-09-06
Add "variant" to cross platform attributes

fuzzybinary:
### What and why?

Flutter has the concept of 'flavors', similar to Android flavors. Since users of the Flutter SDK may take advantage of this to make different variants of application, we need to be able to have the iOS SDK send the variant as one of the `ddtags`, if present.

### How?

I've added '_dd.variant' to the Cross Platform Attributes, which populates a value on the core configuration, which is then fed to the request builder's 'ddtags' query parameter if it is present.

'_dd.variant' will be populated by a 'variant' property in Flutter, but will default to nil. datadog-ci defaults to 'release', but SCI knows to defaults the variant to 'release' during deobfuscation if none is present.

### Review checklist

- [ ] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


xgouchet (approved):

maxep (approved):
Thanks 👍  Do we have a ticket for this? If so, can you assign it to me so I can apply this change to the V2 context. I'm currently fading away the `DatadofV1Context` on the `feature/v2` branch.

PR #1005, opened 2022-09-19
RUMM-2464 [SR] Support incremental updates

ncreated:
### What and why?

📦 This PR adds support to incremental updates in SR segments.

This will vastly reduce the size of SR payloads. Instead of sending all wireframes each time ("full snapshot"), now we only send their differences from previous frame ("incremental snapshot").

### How?

This work is composed of two main parts:
* wireframe IDs generation,
* diff algorithm.

#### 📷 Wireframe IDs generation

I introduced `NodeIDGenerator` which generates unique and stable IDs (`NodeID`) for each `UIView` instance:
* **`NodeIDs` are unique**, so each instance of `UIView` has different ID(s);
* **`NodeIDs` are stable**, meaning that one instance of `UIView` has the same ID for its entire lifespan.

`NodeIDs` are generated by `Recorded` (UI thread) and used as `WireframeID` in `Processor` (background thread). They are keys to determine incremental changes from one array of wireframes to another.

#### 🔃 Diff algorithm

I added generic diff function:
```swift
internal protocol Diffable {
    var id: DiffableID { get }
    func isDifferent(than otherElement: Self) -> Bool
}

internal func computeDiff<E: Diffable>(oldArray: [E], newArray: [E]) throws -> Diff<E> {
```
which is further specialised by `SRWireframe: Diffable` conformance. Making it generic vastly helped for code isolation and testability (I managed to cover most cases + added property-based and fuzzy tests on complex sequences).

The algorithm itself is based on [Paul Heckel's paper: _"A technique for isolating differences between files" (1978)_](https://dl.acm.org/citation.cfm?id=359467). It is fast and performant (`O(n)` in time and memory) way of computing changes in text files. It's quite popular and there are many implementations out there (e.g. [DifferenceKit](https://github.com/ra1028/DifferenceKit), or `hdiff` in Rust). I managed to simplify it a bit for our problem - unlike lines in text files, `IDs` in our wireframes are guaranteed to be unique.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/session-replay/Sources/DatadogSessionReplay/Processor/Diffing/Diff.swift
+++ b/session-replay/Sources/DatadogSessionReplay/Processor/Diffing/Diff.swift
@@ -74,13 +74,11 @@ internal struct DiffError: Error {}
 /// An item in symbols table (`[DiffableID: Symbol]`) in Heckel's algorithm.
 ///
 /// Note: Unlike original algorithm, our implementation doesn't use `OC`, `NC` and `OLNO` counters. In our case, elements within each array are unique,
-/// so instead of repetition counters, we define basic `Bool` flags to track occurence of certain `id` in one or both files.
+/// so instead of repetition counters, we define basic `inNew` flag and optional `indexInOld` to track occurence of certain `id` in one or both files.
 private struct Symbol {
-    /// If element with certain `id` occurs in `oldArray`.
-    var inOld: Bool
     /// If element with certain `id` occurs in `newArray`.
     var inNew: Bool
-    /// The index of element in `oldArray`.
+    /// The index of element in `oldArray` (or `nil` if the element is not there).
     var indexInOld: Int?
 }
 
@@ -120,7 +118,7 @@ internal func computeDiff<E: Diffable>(oldArray: [E], newArray: [E]) throws -> D
     // 1st pass
     // Read `newArray` and store info on each element in symbols `table`:
     for element in newArray {
-        table[element.id] = Symbol(inOld: false, inNew: true, indexInOld: nil)
+        table[element.id] = Symbol(inNew: true, indexInOld: nil)
         na.append(.reference(element.id))
     }
 
@@ -129,9 +127,8 @@ internal func computeDiff<E: Diffable>(oldArray: [E], newArray: [E]) throws -> D
     // exists, update its information (otherwise create new entry):
     for (index, element) in oldArray.enumerated() {
         if table[element.id] == nil {
-            table[element.id] = Symbol(inOld: true, inNew: false, indexInOld: index)
+            table[element.id] = Symbol(inNew: false, indexInOld: index)
         } else {
-            table[element.id]?.inOld = true
             table[element.id]?.indexInOld = index
         }
         oa.append(.reference(element.id))
@@ -146,10 +143,7 @@ internal func computeDiff<E: Diffable>(oldArray: [E], newArray: [E]) throws -> D
             guard let symbol = table[id] else {
                 throw DiffError()
             }
-            if symbol.inOld && symbol.inNew {
-                guard let indexInOld = symbol.indexInOld else {
-                    throw DiffError()
-                }
+            if symbol.inNew, let indexInOld = symbol.indexInOld {
                 na[index] = .index(indexInOld)
                 oa[indexInOld] = .index(index)
             }
@@ -160,7 +154,7 @@ internal func computeDiff<E: Diffable>(oldArray: [E], newArray: [E]) throws -> D
     // > If a line has been found to be unaltered, and the lines immediately adjacent to it in both files are identical,
     // > then these lines must be the same line. This information can be used to find blocks of unchanged lines.
     if na.count > 1 {
-        for i in (1..<(na.count - 1)) {
+        for i in (0..<(na.count - 1)) {
             if case let .index(j) = na[i], j + 1 < oa.count {
                 if case let .reference(id1) = na[i + 1], case let .reference(id2) = oa[j + 1], id1 == id2 {
                     na[i + 1] = .index(j + 1)
@@ -173,7 +167,7 @@ internal func computeDiff<E: Diffable>(oldArray: [E], newArray: [E]) throws -> D
     // 5th pass
     // Similar to 4th pass, except it processes entries in descending order.
     if na.count > 1 {
-        for i in (1..<(na.count - 1)).reversed() {
+        for i in (1..<na.count).reversed() {
             if case let .index(j) = na[i], j - 1 >= 0 {
                 if case let .reference(id1) = na[i - 1], case let .reference(id2) = oa[j - 1], id1 == id2 {
                     na[i - 1] = .index(j - 1)
--- a/session-replay/Sources/DatadogSessionReplay/Processor/Diffing/Diff.swift
+++ b/session-replay/Sources/DatadogSessionReplay/Processor/Diffing/Diff.swift
@@ -1,0 +1,231 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+internal typealias DiffableID = Int64
+
+/// A base interface of array elements compared in `computeDiff(oldArray:newArray:)`.
+internal protocol Diffable {
+    /// Unique identifier of elements in `oldArray` and `newArray`.
+    ///
+    /// It is used to determinen the type of mutation applied to the element:
+    /// - "add" - if certain `id` exists in `newArray`, but not in `oldArray`,
+    /// - "remove" - if certain `id` exists in `oldArray`, but not in `newArray`,
+    /// - "update" - if certain `id` exists in both arrays and `isDifferent(than:)` returns `true`.
+    var id: DiffableID { get }
+
+    /// Determines if this element has any changes if compared to other element.
+    /// It is used from `computeDiff(oldArray:newArray:)`. Both elements are guaranteed to have the same `id`.
+    /// - Parameter otherElement: the element to compare against
+    /// - Returns: `true` if data in both elements is different; `false` otherwise
+    func isDifferent(than otherElement: Self) -> Bool
+}
+
+/// Describes differences between both arrays compared in `computeDiff(oldArray:newArray:)`. It lists minimal
+/// set of changes that needs to be applied to `oldArray` to result with `newArray`.
+///
+/// Note:
+/// - when reconstructing `newArray` from `oldArray` and `Diff` **removals must be applied before additions**;
+/// - if certain element changes its position in `newArray`, `Diff` includes its **removal and addition** (on new position).
+internal struct Diff<Element: Diffable> {
+    /// Dictates that new element was added in `newArray`.
+    struct Add {
+        /// The `id` of preceding element in `newArray` or `nil` if new element should be inserted at index `0`.
+        let previousID: DiffableID?
+        /// The element to add.
+        let new: Element
+    }
+
+    /// Dictates that given element exists in both arrays but its data was changed.
+    struct Update {
+        /// Previous version of element (it has the same `id` as `to`).
+        let from: Element
+        /// New version of element (it has the same `id` as `from`).
+        let to: Element
+    }
+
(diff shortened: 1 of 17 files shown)


session-replay/Sources/DatadogSessionReplay/Processor/Diffing/Diff.swift
```diff
@@ -0,0 +1,231 @@
...
+/// - _"A technique for isolating differences between files"_ Paul Heckel (1978) - https://dl.acm.org/citation.cfm?id=359467
+///
+/// - Parameters:
+///   - oldArray: original array
+///   - newArray: new array
+/// - Returns: `Diff` describing changes from `oldArray` to `newArray`.
+internal func computeDiff<E: Diffable>(oldArray: [E], newArray: [E]) throws -> Diff<E> {
+    var table: [DiffableID: Symbol] = [:]
+    var oa: [Entry] = [] // old array entries
+    var na: [Entry] = [] // new array entries
```
ncreated, line 118: Both `table`, `oa` and `na` names refer to the Heckel's paper and are commonly used among different implementations that I found around. I'd keep it for clarity even if it's not our convention.

session-replay/Sources/DatadogSessionReplay/Processor/Diffing/Diff.swift
```diff
@@ -0,0 +1,231 @@
...
+/// - when reconstructing `newArray` from `oldArray` and `Diff` **removals must be applied before additions**;
+/// - if certain element changes its position in `newArray`, `Diff` includes its **removal and addition** (on new position).
+internal struct Diff<Element: Diffable> {
+    /// Dictates that new element was added in `newArray`.
+    struct Add {
+        /// The `id` of preceding element in `newArray` or `nil` if new element should be inserted at index `0`.
+        let previousID: DiffableID?
+        /// The element to add.
+        let new: Element
+    }
```
ncreated, line 41: As discussed, it's gonna be `nextID: DiffableID?` in place of `previousID`. It's not a big change (and no change to the algorithm itself). It can be updated in another PR.

session-replay/Sources/DatadogSessionReplay/Processor/Diffing/Diff.swift
maxep: For algorithm like this, short var name helps for readability 👍

session-replay/Sources/DatadogSessionReplay/Processor/Diffing/Diff.swift
```diff
@@ -0,0 +1,231 @@
...
+/// Note: Unlike original algorithm, our implementation doesn't use `OC`, `NC` and `OLNO` counters. In our case, elements within each array are unique,
+/// so instead of repetition counters, we define basic `Bool` flags to track occurence of certain `id` in one or both files.
+private struct Symbol {
+    /// If element with certain `id` occurs in `oldArray`.
+    var inOld: Bool
+    /// If element with certain `id` occurs in `newArray`.
+    var inNew: Bool
+    /// The index of element in `oldArray`.
+    var indexInOld: Int?
+}
```
maxep, line 85: If `indexInOld` is nil it means that symbol is not in old. Then, this could be:
```swift
private struct Symbol {
    /// If element with certain `id` occurs in `newArray`.
    var inNew: Bool
    /// The index of element in `oldArray`.
    var indexInOld: Int?
}
```

It would simplify the condition below in `computeDiff`:
```swift
if symbol.inOld && symbol.inNew {
    guard let indexInOld = symbol.indexInOld else {
        throw DiffError()
    }
    na[index] = .index(indexInOld)
    oa[indexInOld] = .index(index)
}
```
to 
```swift
if symbol.inNew, let indexInOld = symbol.indexInOld {
    na[index] = .index(indexInOld)
    oa[indexInOld] = .index(index)
}
```

session-replay/Sources/DatadogSessionReplay/Processor/SRDataModelsBuilder/RecordsBuilder.swift
```diff
@@ -36,14 +36,62 @@ internal class RecordsBuilder {
...
+        if let lastWireframes = lastWireframes {
+            do {
+                return try createIncrementalRecord(from: snapshot, newWireframes: wireframes, lastWireframes: lastWireframes)
+            } catch { // TODO: RUMM-2410 Use `DD.logger` and / or `DD.telemetry` to report ISR errors
+                // In case of any trouble, fallback to FSR which is always possible:
+                return createFullSnapshotRecord(from: snapshot, wireframes: wireframes)
+            }
+        } else {
+            return createFullSnapshotRecord(from: snapshot, wireframes: wireframes)
+        }
```
maxep, line 56: `guard` would be best here :)

session-replay/Sources/DatadogSessionReplay/Processor/SRDataModelsBuilder/RecordsBuilder.swift
```diff
@@ -36,14 +36,62 @@ internal class RecordsBuilder {
...
+
+        let record = SRIncrementalSnapshotRecord(
+            data: .mutationData(
+                value: .init(
+                    adds: diff.adds.map { addition in .init(previousId: addition.previousID, wireframe: addition.new) },
+                    removes: diff.removes.map { removal in .init(id: removal.id) },
+                    updates: try diff.updates.compactMap { update in
+                        let newWireframe = update.to
+                        let oldWireframe = update.from
+                        return try newWireframe.mutations(from: oldWireframe)
```
maxep, line 86: lets save copying elements
```suggestion
                        try update.to.mutations(from: update.from)
```

session-replay/Sources/DatadogSessionReplay/Recorder/SnapshotBuilder/NodeIDGenerator.swift
```diff
@@ -0,0 +1,85 @@
...
+            let id = getNextID()
+            view.nodeID = id
+            return id
+        }
+    }
+
+    /// Returns two `NodeIDs` for given instance of `UIView`.
+    /// - Parameter view: the `UIView` object
+    /// - Returns: `NodeIDs` of queried instance
+    func nodeID2(for view: UIView) -> NodeID2 {
```
maxep, line 52: IMO, It would be clearer if the method returns tuples explicitly.
`func nodeID2(for view: UIView) -> (NodeID, NodeID)`

maxep: Also, this won't scale well if we have components that have other requirements.
We could have a method for requesting `n` IDs.
```func nodeIDs(for view: UIView, count: Int) -> [NodeID]```

**Review by maxep — COMMENTED** [review]
Using this algorithm is simply brillant, well done on translating this to our use case.
I've left some suggestion but it looks great!

File: session-replay/Sources/DatadogSessionReplay/Processor/Diffing/Diff.swift [hunk 7]
```diff
@@ -0,0 +1,231 @@
...
+                oa[indexInOld] = .index(index)
+            }
+        }
+    }
+
+    // 4th pass
+    // > If a line has been found to be unaltered, and the lines immediately adjacent to it in both files are identical,
+    // > then these lines must be the same line. This information can be used to find blocks of unchanged lines.
+    if na.count > 1 {
+        for i in (1..<(na.count - 1)) {
```
mariusc83, line 163: Why are you starting from 1 here ?

session-replay/Sources/DatadogSessionReplay/Processor/Diffing/Diff.swift
```diff
@@ -0,0 +1,231 @@
...
+                    oa[j + 1] = .index(i + 1)
+                }
+            }
+        }
+    }
+
+    // 5th pass
+    // Similar to 4th pass, except it processes entries in descending order.
+    if na.count > 1 {
+        for i in (1..<(na.count - 1)).reversed() {
```
mariusc83, line 176: here you are missing the last element of na. `na[count-1]`. Is this intended ?

session-replay/Sources/DatadogSessionReplay/Processor/Diffing/Diff.swift
maxep: Indeed, it looks like the [4th](https://gist.github.com/ndarville/3166060#fourth-pass) and [5th](https://gist.github.com/ndarville/3166060#fifth-pass) pass of the algorithm should process all values of `NA` and `OA`.

session-replay/Sources/DatadogSessionReplay/Processor/Diffing/Diff.swift
ncreated: Good idea, done 👌!

session-replay/Sources/DatadogSessionReplay/Processor/SRDataModelsBuilder/RecordsBuilder.swift
ncreated: Done 👍👌!

Although, I prefer using `guard x() else { y() }` to express that _"we **expect** `x`, but if it's not there, **fallback** to `y()`"_. In this case `createFSR()` is not a fallback but a regular control flow (_"if `lastWireframes` is not yet there, produce FSR"_). However, it is just my habit and not a shared convention, so I changed it to `guard` which we use more often for similar things.

session-replay/Sources/DatadogSessionReplay/Processor/SRDataModelsBuilder/RecordsBuilder.swift
ncreated: It wasn't copied 🖐️ - with Swift's copy-on-write optimisation that will use the same memory (unless it would be `var` and modified later). Anyway, changed it for more clarity 👍.

session-replay/Sources/DatadogSessionReplay/Recorder/SnapshotBuilder/NodeIDGenerator.swift
ncreated: I'd prefer to keep it as it gives me good documentation point:
```swift
/// Single unique ID of a view in view-tree hierarchy.
/// It is used to mark `UIViews` which correspond to single wireframe in the replay.
internal typealias NodeID = Int64

/// Pair of unique IDs of a single view in view-tree hierarchy.
/// It is used to mark `UIViews` which correspond to two wireframes in the replay.
internal typealias NodeID2 = (NodeID, NodeID)
```
About scalability: it is true, but I'm not afraid of it - it's extremely unlikely that we'll need more than a few IDs per `UIView`. Even this tuple is quite experimental for now and used only to represent `UISwitch` as two wireframes (this may change if we decide to render OS-specific elements on the player side by introducing OS-specific wireframes instead of mimicking its look&feel with 2 shape WFs).

How does it sound @maxep ?

session-replay/Sources/DatadogSessionReplay/Processor/Diffing/Diff.swift
ncreated: Yeah, I think it's a leftover after safeguarding boundaries of this iteration. It should start from `0` - it's fixed now 👌.

Interestingly, it wasn't causing any trouble even after I added more fuzzy tests on short and long sequences. For the sake of safety, I pushed enhancement to tests as well 👍.

session-replay/Sources/DatadogSessionReplay/Processor/Diffing/Diff.swift
ncreated: Indeed, good catch - it should loop through `(1..<na.count)` range - fixed 👍👌. Similar to https://github.com/DataDog/dd-sdk-ios/pull/1005#discussion_r995935265, I couldn't figure out if it causes any problem even after enhancing fuzzy tests.

session-replay/Sources/DatadogSessionReplay/Recorder/SnapshotBuilder/NodeIDGenerator.swift
maxep: Yes sounds good, we just need to be careful not to introduce too many `typealias`es at the module level (inner to object is fine), it can get difficult to track and be confusing.

maxep (approved):
Well done 👏

Issue #978, opened 2022-08-23
Feature Request: Make logging to DDNoopRUMMonitor louder


dfed:
## Background

Today, accidentally logging user actions or errors to the [DDNoopRUMMonitor](https://github.com/DataDog/dd-sdk-ios/blob/927e1f014f8c729c1fc70bc7578038c5e33f28ff/Sources/Datadog/RUM/RUMMonitor/DDNoopRUMMonitor.swift) does not alert developers of API misuse – developers are only [warned when `startView` events are propagated to it](https://github.com/DataDog/dd-sdk-ios/blob/927e1f014f8c729c1fc70bc7578038c5e33f28ff/Sources/Datadog/RUM/RUMMonitor/DDNoopRUMMonitor.swift#L11-L19).

When an engineer accidentally retains an instance of `Global.rum` prior to running `Global.rum = RUMMonitor.initialize()`, they capture a `DDNoopRUMMonitor` instance. If an engineer then logs a user action or error to this `rum` instance, **no warnings appear in the console**, and of course no logs appear in DataDog dashboards either.

## The baseline request

Update `DDNoopRUMMonitor` to alert a developer that they are misusing the DataDog SDK when  _any_ event (and not just `startView`) is sent to the `DDNoopRUMMonitor`. I'd be happy to put up a PR to accomplish this if the proposal is well received by maintainers.

## The stretch request

Given that console logs might get lost in apps that log a lot to the console, and that errors routed to a `Global.rum` instance are explicitly high-priority events, ideally `DDNoopRUMMonitor`'s `addError` implementation would trigger an `assertionFailure(...)`. Again, I'd be happy to put up a PR to accomplish this if the proposal is well received by maintainers.

dfed:
Sounds great! I've opened #1007 to cover the baseline request. I have no qualms if you would like to close this issue after #1007 lands, though I am hoping that `2.x` gets this stretch request 🙂

maxep:
Thank you again @dfed for your contribution. This will land in `2.x` when we migrate the RUM monitor 👍

PR #1007, opened 2022-09-20
Warn when DDNoopRUMMonitor is sent any DDRUMMonitor methods

dfed:
### What and why?

This PR resolves the baseline request in #978 by calling `warn()` when any method vended by `DDRUMMonitor` is called on an instance of `DDNoopRUMMonitor`.

This change will make it more obvious when a `DDNoopRUMMonitor` has been captured and is being utilized accidentally.

### How?

I copied over all methods from `DDRUMMonitor` into `DDNoopRUMMonitor`, replaced `public` with `override`, and called `warn()` within each method body.

### Review checklist
- [x] Add CHANGELOG entry for user facing changes – N/A

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/RUM/RUMMonitor/DDNoopRUMMonitor.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/DDNoopRUMMonitor.swift
@@ -26,6 +26,13 @@ internal class DDNoopRUMMonitor: DDRUMMonitor {
         warn()
     }
 
+    override func stopView(
+        viewController: UIViewController,
+        attributes: [AttributeKey: AttributeValue] = [:]
+    ) {
+        warn()
+    }
+
     override func startView(
         key: String,
         name: String? = nil,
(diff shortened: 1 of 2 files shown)


dfed:
I could use help understanding what about the Bitrise job failed – I don't have permissions to view the failure message, and the unit test I updated succeeds on my machine.

Sources/Datadog/RUM/RUMMonitor/DDNoopRUMMonitor.swift
```diff
@@ -26,11 +26,170 @@ internal class DDNoopRUMMonitor: DDRUMMonitor {
...
+    }
+
+    override func addAttribute(forKey key: AttributeKey, value: AttributeValue) {
+        warn()
+    }
+
+    override func removeAttribute(forKey key: AttributeKey) {
+        warn()
+    }
+
```
maxep, line 194:
```suggestion
```

maxep (commented):
Hey @dfed ! Thank you so much for this contribution 👍 

Indeed, our CI requires a manual trigger before running external PRs. It looks great, only the linter is failing because of that extra line break. After that, it would be good for merging.
Thanks again!

maxep (approved):

PR #1017, opened 2022-10-13
Add cross platform perf metrics

louiszawadzki:
### What and why?

Enable cross platform performance metrics to be reported

### How?

I'm reusing VitalInfo to handle the logic of getting the average.

### Review checklist


Sources/Datadog/RUM/RUMVitals/PerformanceMetric.swift
```diff
@@ -0,0 +1,18 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+public enum PerformanceMetric {
+    // The amount of time Flutter spent in its `build` method for this view.
+    case FLUTTER_BUILD_TIME
```
fuzzybinary, line 11: Swift convention is to have enums be lowerCamelCase.

Sources/Datadog/DDRUMMonitor.swift
```diff
@@ -286,6 +286,16 @@ public class DDRUMMonitor {
     /// - Parameter key: key for the attribute that will be removed.
     public func removeAttribute(forKey key: AttributeKey) {}
 
+    /// Adds a record for a cross-platform performance metric
+    /// - Parameters:
+    ///   - metric: the name of the performance metric. One of `flutter_build_time`, `flutter_raster_time` or `js_refresh_rate`
```
fuzzybinary, line 291: Since this is now an enum, you probably don't need the "One of comment here"

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -435,15 +440,15 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
                 firstContentfulPaint: nil,
                 firstInputDelay: nil,
                 firstInputTime: nil,
-                flutterBuildTime: nil,
-                flutterRasterTime: nil,
+                flutterBuildTime: viewPerformanceMetrics[PerformanceMetric.FLUTTER_BUILD_TIME]?.asFlutterBuildTime(),
+                flutterRasterTime: viewPerformanceMetrics[PerformanceMetric.FLUTTER_RASTER_TIME]?.asFlutterRasterTime(),
```
fuzzybinary, line 444: You should be able to omit the `PerformanceMetric` portion of this:
```swift
flutterBuildTime: viewPerformanceMetrics[.flutterBuildTime]?.asFlutterBuildTime(),
```

Tests/DatadogTests/Datadog/RUMMonitorTests.swift
```diff
@@ -1387,6 +1387,34 @@ class RUMMonitorTests: XCTestCase {
...
+        let rum: RUMFeature = .mockByRecordingRUMEventMatchers()
+        core.register(feature: rum)
+
+        let monitor = try createTestableRUMMonitor()
+
+        monitor.startView(viewController: mockView)
+        monitor.updatePerformanceMetric(metric: PerformanceMetric.JS_REFRESH_RATE, value: 40.0)
+        monitor.updatePerformanceMetric(metric: PerformanceMetric.JS_REFRESH_RATE, value: 20.0)
+        monitor.updatePerformanceMetric(metric: PerformanceMetric.FLUTTER_BUILD_TIME, value: 32.0)
+        monitor.updatePerformanceMetric(metric: PerformanceMetric.FLUTTER_RASTER_TIME, value: 42.0)
```
fuzzybinary, line 1402: Same here, you don't need to qualify the enums in Swift. These can be simplified to:
```swift
monitor.updatePerformanceMetric(metric: .jsRefreshRate, value: 40.0)
```

Sources/Datadog/RUM/RUMVitals/VitalInfo.swift
```diff
@@ -56,4 +56,31 @@ internal struct VitalInfo {
...
+    }
+
+    func asJsRefreshRate() -> RUMViewEvent.View.JsRefreshRate {
+        return RUMViewEvent.View.JsRefreshRate(
+            average: meanValue ?? 0.0,
+            max: maxValue ?? 0.0,
+            metricMax: 60.0,
+            min: minValue ?? 0.0
+        )
+    }
```
ncreated, line 85: I like these convenience methods, but adding them in `VitalInfo` doesn't seem right. The `VitalInfo` is agnostic vital accumulator and here we extend it with vital-specific (even: platform-specific) getters. It will not only add boilerplate for `VitalInfo` type lookup in autocompletion but also require tests coverage as we define new `internal` behaviours.

A better choice to introduce such convenience would be to add private extension in `RUMViewScope.swift` (so the only place where we actually need it):
```swift
// RUMViewScope.swift

// ...

private extension VitalInfo {
   func asFlutterBuildTime() -> RUMViewEvent.View.FlutterBuildTime { /* ... */ }
   func asFlutterRasterTime() -> RUMViewEvent.View.FlutterBuildTime { /* ... */ }
   func asJsRefreshRate() -> RUMViewEvent.View.FlutterBuildTime { /* ... */ }
}
```
This way we only introduce this getters for `VitalInfo` usage in `RUMViewScope.swift` without blowing up the `VitalInfo` namespace in other files. Also, we only introduce `private` behaviours, so we don't need to test it directly (that's our convention).

Eventually, for wider `internal` use, we isolate RUM-related mappings in [`RUMDataModelsMapping`](https://github.com/DataDog/dd-sdk-ios/blob/louiszawadzki/rumm-2514/add-cross-platform-perf-metrics/Sources/Datadog/RUM/DataModels/RUMDataModelsMapping.swift), but I think this is not required for this PR as these getters are scoped only to `RUMViewScope`.

ncreated (changes requested):
🎯 looks very good! I left a feedback on convention for adding helpers. I also agree with all the points reported by Jeff 👌

ncreated (approved):
Nice work 💪!

PR #1024, opened 2022-10-18
RUMM-2503 [SR] Add "mask all" privacy rule

ncreated:
### What and why?

🚚🔒 This PR adds "mask all" and "allow all" privacy rules to Session Replay.

With "mask all" (default setting), no text will be recorded in traversed `UILabels` and `UITextFields`. Instead, visible characters will be replaced by `"x"` (except whitespaces and newlines).

### How?

The public API added in this PR is still a draft (we will work on the actual one when integrating with V2 SDK).

The actual masking is done by basic iteration over `String` with replacing certain characters. To make it performant:
* It uses [`String.UnicodeScalarView.Iterator`](https://developer.apple.com/documentation/swift/string/unicodescalarview/iterator) which gives good performance and encapsulates Unicode complexity. I measured this to be 10x faster than RegExp (with 40K characters-long string and 1K repeats, it was `20ms` vs `2ms`). More on this approach and benchmarks can be also read [here](https://medium.com/@tonyallevato/strings-characters-and-performance-in-swift-a-deep-dive-b7b5bde58d53).
* The `String` is captured on the main queue, but masking is performed on the background thread.

In proposed implementation I'm trying to foresee next rules that we will introduce for parity with web SR [privacy options](https://docs.datadoghq.com/real_user_monitoring/session_replay/privacy_options):
* The "mask user input" option will likely target interactive elements (`UITextFields`). To anticipate this need, I have decentralised the logic of text obfuscation, so each node recorder (`UILabelRecorder` and `UITextFieldRecorder`) decides on its own if masking should be applied.
* Very likely we will also add ways to override this setting for certain instances of `UIView` (perhaps in similar fashion as our RUM views predicate works). To make sure the solution scales to this use case, each node recorder resolves `privacy` information on the main thread (even though the actual masking is performed on background queue). This way we will be able to inspect the element or even pass it to sort of predicate in public API.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/session-replay/Sources/DatadogSessionReplay/Processor/Privacy/TextObfuscator.swift
+++ b/session-replay/Sources/DatadogSessionReplay/Processor/Privacy/TextObfuscator.swift
@@ -6,7 +6,15 @@
 
 import Foundation
 
-internal struct TextObfuscator {
+internal protocol TextObfuscating {
+    /// Obfuscates given `text`.
+    /// - Parameter text: the text to be obfuscated
+    /// - Returns: obfuscated text
+    func mask(text: String) -> String
+}
+
+/// Text obfuscator which replaces all readable characters with `"x"`.
+internal struct TextObfuscator: TextObfuscating {
     /// The character to mask text with.
     let maskCharacter: UnicodeScalar = "x"
 
@@ -30,3 +38,13 @@ internal struct TextObfuscator {
         return masked
     }
 }
+
+/// Text obfuscator which only returns the original text.
+internal struct NOPTextObfuscator: TextObfuscating {
+    func mask(text: String) -> String {
+        return text
+    }
+}
+
+/// Text obfuscator which only returns the original text.
+internal let nopTextObfuscator: TextObfuscating = NOPTextObfuscator()
--- a/session-replay/Sources/DatadogSessionReplay/Processor/Privacy/TextObfuscator.swift
+++ b/session-replay/Sources/DatadogSessionReplay/Processor/Privacy/TextObfuscator.swift
@@ -1,0 +1,32 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+internal struct TextObfuscator {
+    /// The character to mask text with.
+    let maskCharacter: UnicodeScalar = "x"
+
+    /// Masks given `text` by replacing all not whitespace characters with `"x"`.
+    /// - Parameter text: the text to be masked
+    /// - Returns: masked text
+    func mask(text: String) -> String {
+        var masked = ""
+
+        var iterator = text.unicodeScalars.makeIterator()
+
+        while let nextScalar = iterator.next() {
+            switch nextScalar {
+            case " ", "\n", "\r", "\t":
+                masked.unicodeScalars.append(nextScalar)
+            default:
+                masked.unicodeScalars.append(maskCharacter)
+            }
+        }
+
+        return masked
+    }
+}
--- a/session-replay/Sources/DatadogSessionReplay/Recorder/SnapshotBuilder/NodeRecorders/UILabelRecorder.swift
+++ b/session-replay/Sources/DatadogSessionReplay/Recorder/SnapshotBuilder/NodeRecorders/UILabelRecorder.swift
@@ -23,7 +23,8 @@ internal struct UILabelRecorder: NodeRecorder {
             attributes: attributes,
             text: label.text ?? "",
             textColor: label.textColor?.cgColor,
-            font: label.font
+            font: label.font,
+            textObfuscator: context.options.privacy == .maskAll ? context.textObfuscator : nil
         )
         return SpecificElement(wireframesBuilder: builder)
     }
@@ -39,6 +40,8 @@ internal struct UILabelWireframesBuilder: NodeWireframesBuilder {
     let textColor: CGColor?
     /// The font used by the label.
     let font: UIFont?
+    /// Text obfuscator for masking text (`nil` if text should not be obfuscated for this element).
+    let textObfuscator: TextObfuscator?
 
     func buildWireframes(with builder: WireframesBuilder) -> [SRWireframe] {
         // The actual frame of the text, which is smaller than the frame of the label:
@@ -53,7 +56,7 @@ internal struct UILabelWireframesBuilder: NodeWireframesBuilder {
             builder.createTextWireframe(
                 id: wireframeID,
                 frame: attributes.frame,
-                text: text,
+                text: textObfuscator.map { $0.mask(text: text) } ?? text,
                 textFrame: textFrame,
                 textColor: textColor,
                 font: font,
--- a/session-replay/Sources/DatadogSessionReplay/Recorder/SnapshotBuilder/NodeRecorders/UILabelRecorder.swift
+++ b/session-replay/Sources/DatadogSessionReplay/Recorder/SnapshotBuilder/NodeRecorders/UILabelRecorder.swift
@@ -24,7 +24,7 @@ internal struct UILabelRecorder: NodeRecorder {
             text: label.text ?? "",
             textColor: label.textColor?.cgColor,
             font: label.font,
-            textObfuscator: context.options.privacy == .maskAll ? context.textObfuscator : nil
+            textObfuscator: context.options.privacy == .maskAll ? context.textObfuscator : nopTextObfuscator
         )
         return SpecificElement(wireframesBuilder: builder)
     }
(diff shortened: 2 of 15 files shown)


session-replay/Sources/DatadogSessionReplay/Processor/Privacy/TextObfuscator.swift
```diff
@@ -0,0 +1,32 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+internal struct TextObfuscator {
```
ncreated, line 9: I'm making it a separate type consciously (alternatively, it could be just a global function):
* By passing `TextObfuscator?` to wireframe builders we indicate if masking is necessary (although it could be optional closure too).
* If we have to, we can explore more options for improving the masking performance with implementing basic LRU cache (using `text.hashValue` as a key). For this (or other) perf tuning reasons, having an ability to manage internal state (in `class`) seems worthy.

maxep: The caching would be a great optimisation indeed 👍  lets keep that in mind.

session-replay/Sources/DatadogSessionReplay/Processor/Privacy/TextObfuscator.swift
```diff
@@ -0,0 +1,32 @@
...
+            switch nextScalar {
+            case " ", "\n", "\r", "\t":
+                masked.unicodeScalars.append(nextScalar)
+            default:
+                masked.unicodeScalars.append(maskCharacter)
+            }
+        }
+
+        return masked
+    }
```
maxep, line 31: I see 2 things that could be improved in this method:

1. The [Unicode.Scalar.Properties/isWhitespace](https://developer.apple.com/documentation/swift/unicode/scalar/properties-swift.struct/iswhitespace) can tell you if the character should be treated as whitespace.

2. The `String.UnicodeScalarView.Iterator` is required by the `Sequence` protocol that enable the `for in` statement. It also comes with a bunch of extensions such as `map`. It's all based on the iterator, so the perf should be the same.

This could become something like:
```swift
func mask(text: String) -> String {
    let unicodeScalars = text.unicodeScalars.map { scalar in
        scalar.properties.isWhitespace ? scalar : maskCharacter
    }

    let unicodeScalarView = String.UnicodeScalarView(unicodeScalars)
    return String(unicodeScalarView)
}
```

session-replay/Sources/DatadogSessionReplay/Recorder/SnapshotBuilder/NodeRecorders/UILabelRecorder.swift
```diff
@@ -39,6 +40,8 @@ internal struct UILabelWireframesBuilder: NodeWireframesBuilder {
     let textColor: CGColor?
     /// The font used by the label.
     let font: UIFont?
+    /// Text obfuscator for masking text (`nil` if text should not be obfuscated for this element).
+    let textObfuscator: TextObfuscator?
```
maxep, line 44: We could have a `TextObfuscator` protocol with a `NOPTextObfuscator` and make this non-optional, WDYT?
Not a change request, it's totally fine with struct+optional.

maxep (changes requested):
It looks great 🚀  I've left a couple of feedback, the main one being using [Unicode.Scalar.Properties/isWhitespace](https://developer.apple.com/documentation/swift/unicode/scalar/properties-swift.struct/iswhitespace) instead of a list of strings.

session-replay/Sources/DatadogSessionReplay/Processor/Privacy/TextObfuscator.swift
ncreated: It's interesting feedback, so I gave it a try with some benchmarks 🚤. Unfortunately the time performance is ~4x worse than current, iterator-based implementation. For that reason, I'd rather stick to existing implementation, even if `isWhitespace` looks more portable (rationale: `" ", "\n", "\r", "\t"` should cover majority of use cases and if not, it scales up pretty easy).

**More explanation:**

I tried 2 variants with `mask()` (original) and `mask2()` (new) implementations:
- big: 40K characters-long text, 1K measures
- small: 40 characters-long text, 1K measures

Getting:
- big:
   - `mask()` - 2.28 ms
   - `mask2()` - 10.04 ms
- small:
   - `mask()` - 0.0047 ms
   - `mask2()` - 0.0145 ms

When tracking it down in profiler, I found that both new elements contribute to this slower performance, so neither using `.map {}` nor `.isWhitespace` will improve speed.
* `mask()` trace:
Screenshot 2022-10-19 at 11 07 00

* `mask2()` trace:
Screenshot 2022-10-19 at 11 07 29

I didn't spent much time in digging more, but at the high level:
* `.map {}` seems to be adding quite a few routines related to generic sequences
* `.isWhitespace` leads to [`_swift_stdlib_getBinaryProperties`](https://github.com/apple/swift/blob/5265ae757e9a8fce8b48a4c8e2b37dce7c0ddbcc/stdlib/public/core/UnicodeScalarProperties.swift#L125) which is quite expensive but we lose track of it (hard to say what's inside) - in contrast, the `case "\n"` is O(1) because scalars are `ExpressibleByUnicodeScalarLiteral` so it's done with basic memory arithmetics

Let me know what you think @maxep 🙂

Benchmark code for reference:
```swift
func testBenchmark1() {
    let text: String = .mockRandom(among: .alphanumericsAndWhitespace, length: 40_000)

    benchmark("mask()") {
        _ = obfuscator.mask(text: text)
    }
}

func testBenchmark2() {
    let text: String = .mockRandom(among: .alphanumericsAndWhitespace, length: 40_000)

    benchmark("mask2()") {
        _ = obfuscator.mask2(text: text)
    }
}

func benchmark(_ label: String, block: () -> Void) {
    var measures: [TimeInterval] = []

    for i in (0..<1_010) {
        let start = Date()
        block()

        if i >= 10 { // only measure after it is warmed-up
            let stop = Date()
            measures.append(stop.timeIntervalSince(start))
        }
    }

    let sum = measures.reduce(0, +)
    let avg = sum / Double(measures.count)

    print("⭐️ AVG in \(label): \(avg * 1_000) ms")
}
```

session-replay/Sources/DatadogSessionReplay/Recorder/SnapshotBuilder/NodeRecorders/UILabelRecorder.swift
ncreated: Good idea 👍, it's less optionals in downstream elements (wireframe builder), so I like it. Done 👌.

session-replay/Sources/DatadogSessionReplay/Processor/Privacy/TextObfuscator.swift
ncreated: 👌 I created low-priority `RUMM-2671` to track perf optimisation ideas for text masking.

session-replay/Sources/DatadogSessionReplay/Processor/Privacy/TextObfuscator.swift
maxep: Oh nice!! thanks for doing the benchmark 👍 

I'm surprised how heavy is the `Unicode.Scalar.Properties/isWhitespace` 😳
Looking at the profiles, it seems that the `masked.unicodeScalars.append(...)` is also expensive compared to the `map` (its few extra routines doesn't seem heavy).

Would you be able to test a `mask3`?
```swift
func mask(text: String) -> String {
    let unicodeScalars = text.unicodeScalars.map { scalar in
        switch scalar {
            case " ", "\n", "\r", "\t":
                return scalar
            default:
                return maskCharacter
            }
    }

    let unicodeScalarView = String.UnicodeScalarView(unicodeScalars)
    return String(unicodeScalarView)
}
```

ncreated: I tried, it's in between and closer to "bad" performance:
* big:
   * mask() - 2.28 ms
   * mask2() - 10.04 ms
   * **mask3() - 8.38 ms**
* small:
   * mask() - 0.0047 ms
   * mask2() - 0.0145 ms
   * **mask3() - 0.0133 ms**

Difference is that iterator-based approach leads to `String.UnicodeScalarView.append(_:)` in the call stack, whereas map-based calls `String.UnicodeScalarView.append(contentsOf:)` that is visibly more complex:

- iterator-based profile (`String.UnicodeScalarView.append(_:)`):
Screenshot 2022-10-20 at 10 40 25


- map-based profile (`String.UnicodeScalarView.append(contentsOf:)`):
Screenshot 2022-10-20 at 10 40 15

Again, I didn't dive too deep, but even a glance at expanded call tree suggests it's coming from `.map {}` acting too generic and bringing some underlying allocation cost:

Screenshot 2022-10-20 at 10 48 49

I also tried using `.isWhitespace` + iterator in `mask4()` resulting with also "worse" performance:
* big:
   * `mask4()` - 4.52 ms
* small
   * `mask4()` - 0.0065ms

All that said, as we're discussing implementation detail of performance-critical piece (it will be called **every frame for all texts**) we should go with the most performant approach. Would be a different choice if it's about architecture-level API where convenience is key. WDYT @maxep?

maxep: Totally, we are aiming for the most performant one. I was expecting the `isWhitespace` to be a pre-filled boolean (or at least is simple binary check) and `map` to perform the same as using the iterator. But that's clearly not the case!
Lets stick with your original implementation then 👍 And thanks for testing that!!

maxep (approved):

PR #1031, opened 2022-10-21
RUMM-2670 Add a method to add user info properties

fuzzybinary:
### What and why?

Currently, the only way to add or modify the `extraInfo` for a user is to call `setUserInfo` with all the same information but with new properties.  The new method allows you to add, overwrite, or remove user properties without having the rest of the user info (id, email, etc).

### How?

The supplied info dictionary is merged with the existing dictionary after first scanning for any `nil` values, which indicates that an existing in key in the dictionary should be removed.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


Sources/Datadog/DatadogCore/DatadogCore.swift
```diff
@@ -97,6 +97,21 @@ internal final class DatadogCore {
         )
     }
 
+    /// Add or override the properties of the current user
+    ///
+    ///  - Parameters:
+    ///    - parameters: The user's custom attibutes to add over override
+    func addUserProperties(properties: [AttributeKey: AttributeValue?]) {
+        var extraInfo = dependencies.userInfoProvider.value.extraInfo
+        for property in properties {
```
Kaspik, line 106: There is much cleaner (and faster) way in swift via `compactMap` than `for` cycle (and you are kind of using it below, not sure why not using here too):

```swift
extension Dictionary {
   func removeOptionalValues<T>() -> [Key: T] where Value == T? {
      return compactMapValues { $0 }
   }
}
```
You could name it `removeNilValues` or `removeOptionalValues` or anything :)

Sources/Datadog/Datadog.swift
```diff
@@ -142,6 +142,18 @@ public class Datadog {
         )
     }
 
+    /// Add properties to the current user information
+    ///
+    /// There properties will be added to logs, traces and RUM events automatically.
+    /// - Parameters:
+    ///   - properties:
+    public static func addUserProperties(
```
plousada, line 150: We're adding a new terminology here with the wording "UserProperties". 
Why not maintain the existing one, "ExtraInfo"?

Sources/Datadog/DatadogCore/DatadogCore.swift
fuzzybinary: This isn't removing nil values in extraInfo. It's removing values from extraInfo who's values are `nil` in the properties map.

Kaspik: My bad, sorry! :)

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- [IMPROVEMENT] Add a method to add properties to user info without needing to reset other user info..
```
maxep, line 3:
```suggestion
- [IMPROVEMENT] Add a method to add properties to user info without needing to reset other user info.
```

Sources/Datadog/DatadogCore/DatadogCore.swift
```diff
@@ -97,6 +97,21 @@ internal final class DatadogCore {
...
+    ///  - Parameters:
+    ///    - extraInfo: The user's custom attibutes to add or override
+    func addUserExtraInfo(_ newExtraInfo: [AttributeKey: AttributeValue?]) {
+        var extraInfo = dependencies.userInfoProvider.value.extraInfo
+        for property in newExtraInfo {
+            if property.value == nil {
+                extraInfo.removeValue(forKey: property.key)
+            }
+        }
+        extraInfo.merge(newExtraInfo.compactMapValues { $0 }, uniquingKeysWith: { _, new in new })
```
maxep, line 111: I think this can be done a in single line:
```suggestion
        newExtraInfo.forEach { extraInfo[$0.key] = $0.value }
```

maxep (changes requested):
Looks great!! I think we can simplify the dict merging 👍

plousada (approved):

Sources/Datadog/DatadogCore/DatadogCore.swift
fuzzybinary: Nevermind... apparently you're right that adding nil to a dictionary that doesn't support it removes the key. TIL.

maxep: if `$0.value` is nil it will remove it from `extraInfo`. I did test it on your branch, no compiler error and tests are passing :)

maxep (approved):
🚀

PR #1032, opened 2022-10-25
RUMM-2687 Invert frame time to get js refresh rate

louiszawadzki:
### What and why?

Get a more sensible for the average js refresh rate by getting the harmonic mean instead of the arithmetic mean.
See https://gamedev.net/forums/topic/628085-math-fail-adding-frames-per-second/4960225/ and https://en.wikipedia.org/wiki/Harmonic_mean#Average_speed to understand why this makes more sense.

### How?

Let the RN SDK report the frame time, then invert it when reporting to backend

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
@@ -448,7 +448,7 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
                 inForegroundPeriods: nil,
                 isActive: isActive,
                 isSlowRendered: isSlowRendered,
-                jsRefreshRate: viewPerformanceMetrics[.jsRefreshRate]?.asJsRefreshRate(),
+                jsRefreshRate: viewPerformanceMetrics[.jsFrameTimeMilliseconds]?.asJsRefreshRate(),
                 largestContentfulPaint: nil,
                 loadEvent: nil,
                 loadingTime: nil,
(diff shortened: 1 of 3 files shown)


Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -637,10 +637,17 @@ private extension VitalInfo {
...
         )
     }
 }
+
+internal func invertValue(value: Double?) -> Double {
+    if value == 0.0 || value == nil {
+        return 0.0
+    }
+    return 1 / (value ?? 1)
+}
```
maxep, line 653: This can be moved to a `var` in our extension in https://github.com/DataDog/dd-sdk-ios/blob/develop/Sources/Datadog/Utils/SwiftExtensions.swift#L19L26

it can be:
```
var inverted: Double {
    return self == 0 ? 0 : 1 / self
}
```

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -637,10 +637,17 @@ private extension VitalInfo {
 
     func asJsRefreshRate() -> RUMViewEvent.View.JsRefreshRate {
         return RUMViewEvent.View.JsRefreshRate(
-            average: meanValue ?? 0.0,
-            max: maxValue ?? 0.0,
+            average: invertValue(value: meanValue) * 1_000,
+            max: invertValue(value: minValue) * 1_000,
             metricMax: 60.0,
-            min: minValue ?? 0.0
+            min: invertValue(value: maxValue) * 1_000
```
maxep, line 643: With the extension mentioned, you can do:
```swift
average: meanValue.map { $0.inverted * 1_000 } ?? 0,
max: minValue.map { $0.inverted * 1_000 } ?? 0,
metricMax: 60.0,
min: maxValue.map { $0.inverted * 1_000 } ?? 0
```

maxep (commented):
Nice!! I've a suggestion for replacing the `invertValue` function with a computed var 👍

Sources/Datadog/Utils/SwiftExtensions.swift
```diff
@@ -23,6 +23,10 @@ extension Double {
         }
         return self / divider
     }
+
+    var inverted: Double {
+        return self == 0 ? 0 : 1 / self
+    }
```
ncreated, line 29: Although it seems trivial, would be nice to have it tested in [SwiftExtensionsTests.swift](https://github.com/DataDog/dd-sdk-ios/blob/develop/Tests/DatadogTests/Datadog/Utils/SwiftExtensionsTests.swift). Similar for `divideIfNotZero(by:)` - it's missing tests.

ncreated (approved):
Looks good 👍💯. Would be great to add tests for the new extension on `Double`.

louiszawadzki:
Changed the frame time to be in seconds as it's the unit it's reported in, and that removes computations on both ends.

PR #1034, opened 2022-10-26
RUMM-2661 [SR] Inject RUM context to Session Replay

ncreated:
### What and why?

🚚 This PR injects RUM context to Session Replay. From now on, the SR module receives up-to-date information on RUM _app ID_, _session ID_ and _view ID_ (or lack of them if RUM session is not sampled).

This is a necessary step to establish the link between SR and RUM.

### How?

`Datadog` module (exposing V2 APIs) is declared as dependency in SR's `Package.swift`.

At the entry level, SR module defines conformances to V2 Feature interfaces:
* `SessionReplayFeature: DatadogFeature` - to register V2 Feature,
* `RUMContextReceiver: FeatureMessageReceiver` - to receive `DatadogContext` updates through V2 message bus.

Once received, the `RUMContext` is passed to `Recorder`. It is then bundled in recorded `ViewTreeSnapshot` sent to `Processor`. In RUMM-2679 the context will be also used to pause and resume `Recorder` accordingly to RUM session sampling (to avoid consuming device resources when RUM session is rejected).

Based on the `snapshot.rumContext` value, `Processor` will take decision on creating appropriate SR records. This work will be implemented in _RUMM-2662: Session replay data is automatically uploaded_.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Datadog.xcworkspace/contents.xcworkspacedata
+++ b/Datadog.xcworkspace/contents.xcworkspacedata
@@ -7,7 +7,4 @@
    <FileRef
       location = "group:instrumented-tests/http-server-mock">
    </FileRef>
-   <FileRef
-      location = "group:session-replay">
-   </FileRef>
 </Workspace>
--- a/Sources/Datadog/DatadogInternal/DatadogCoreProtocol.swift
+++ b/Sources/Datadog/DatadogInternal/DatadogCoreProtocol.swift
@@ -14,17 +14,17 @@ public protocol DatadogCoreProtocol {
     /// Registers a Feature instance.
     ///
     /// Feature collects and transfers data to a Datadog Product (e.g. Logs, RUM, ...). Upon registration, the Feature can
-    /// retrieve a `FeatureScope` interface for writing events. The core will store and upload events in an efficient way
-    /// with performance presets defined on initialization.
+    /// retrieve a `FeatureScope` interface for writing events to the core. The core will store and upload events efficiently
+    /// according to the performance presets defined on initialization.
     ///
-    /// The Feature can also communicate to other Features by sending messages on the message bus managed by core.
+    /// A Feature can also communicate to other Features by sending messages on the message bus managed by core.
     ///
     /// - Parameter feature: The Feature instance - it will be retained and held by core.
     func register(feature: DatadogFeature) throws
 
     /// Retrieves previously registered Feature by its name and type.
     ///
-    /// Feature's type can be specified as parameter or inferred from the return type:
+    /// A Feature type can be specified as parameter or inferred from the return type:
     ///
     ///     let feature = core.feature(named: "foo", type: Foo.self)
     ///     let feature: Foo? = core.feature(named: "foo")
--- a/Sources/Datadog/DatadogInternal/DatadogCoreProtocol.swift
+++ b/Sources/Datadog/DatadogInternal/DatadogCoreProtocol.swift
@@ -8,23 +8,23 @@ import Foundation
 
 public internal(set) var defaultDatadogCore: DatadogCoreProtocol = NOPDatadogCore()
 
-/// A Datadog Core holds a set of features and is responsible of managing their storage
+/// A Datadog Core holds a set of Features and is responsible for managing their storage
 /// and upload mechanism. It also provides a thread-safe scope for writing events.
 public protocol DatadogCoreProtocol {
     /// Registers a Feature instance.
     ///
-    /// A Feature collects and transfers data to a Datadog Product (e.g. Logs, RUM, ...). A registered Feature can
-    /// open a `FeatureScope` to write events, the core will then be responsible for storing and uploading events
-    /// in a efficient manner. Performance presets for storage and upload are define when instanciating the core instance.
+    /// Feature collects and transfers data to a Datadog Product (e.g. Logs, RUM, ...). Upon registration, the Feature can
+    /// retrieve a `FeatureScope` interface for writing events. The core will store and upload events in an efficient way
+    /// with performance presets defined on initialization.
     ///
-    /// A Feature can also communicate to other Features by sending message on the bus that is managed by the core.
+    /// The Feature can also communicate to other Features by sending messages on the message bus managed by core.
     ///
-    /// - Parameter feature: The Feature instance.
+    /// - Parameter feature: The Feature instance - it will be retained and held by core.
     func register(feature: DatadogFeature) throws
 
-    /// Retrieves a Feature by its name and type.
+    /// Retrieves previously registered Feature by its name and type.
     ///
-    /// A Feature type can be specified as parameter or inferred from the return type:
+    /// Feature's type can be specified as parameter or inferred from the return type:
     ///
     ///     let feature = core.feature(named: "foo", type: Foo.self)
     ///     let feature: Foo? = core.feature(named: "foo")
--- a/Sources/Datadog/RUMMonitor.swift
+++ b/Sources/Datadog/RUMMonitor.swift
@@ -148,9 +148,14 @@ public class RUMMonitor: DDRUMMonitor, RUMCommandSubscriber {
 
     /// RUM Attributes shared with other Feature registered in core.
     internal struct Attributes {
+        /// The ID of RUM application (`String`).
         internal static let applicationID = "application_id"
+        /// The ID of current RUM session (standard UUID `String`, lowercased).
+        /// In case the session is rejected (not sampled), RUM context is set to empty (`[:]`) in core.
         internal static let sessionID = "session_id"
+        /// The ID of current RUM view (standard UUID `String`, lowercased).
         internal static let viewID = "view.id"
+        /// The ID of current RUM action (standard UUID `String`, lowercased).
         internal static let userActionID = "user_action.id"
     }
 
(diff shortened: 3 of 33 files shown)


Datadog.xcworkspace/contents.xcworkspacedata
```diff
@@ -7,7 +7,4 @@
    <FileRef
       location = "group:instrumented-tests/http-server-mock">
    </FileRef>
-   <FileRef
-      location = "group:session-replay">
-   </FileRef>
```
ncreated, line 12: I had to disintegrate `session-replay` module from the main workspace and `Example` app target. This is because Xcode fails to resolve the dependencies within the workspace:

Screenshot 2022-10-26 at 14 08 13

Screenshot 2022-10-26 at 14 07 59

I think it's because `Datadog.pbxproj` already defines `Datadog*` modules and the same modules are shipped through local dependency on `dd-sdk-ios` package. I considedr it a "workspace" factor, because the same setup (depending on local `session-replay` package that links local `dd-sdk-ios`) works in other local projects I have.

Losing presence of `session-replay` in `Example` app's workspace is not a problem for me, as I rarely test SR in this app (I use better ones - with more challenging UIs). Also, solving this problem could be difficult if not impossible without the new project structure (RFC is in progress).

Sources/Datadog/DatadogInternal/DatadogCoreProtocol.swift
```diff
@@ -8,23 +8,23 @@ import Foundation
...
 /// and upload mechanism. It also provides a thread-safe scope for writing events.
 public protocol DatadogCoreProtocol {
     /// Registers a Feature instance.
     ///
-    /// A Feature collects and transfers data to a Datadog Product (e.g. Logs, RUM, ...). A registered Feature can
-    /// open a `FeatureScope` to write events, the core will then be responsible for storing and uploading events
-    /// in a efficient manner. Performance presets for storage and upload are define when instanciating the core instance.
+    /// Feature collects and transfers data to a Datadog Product (e.g. Logs, RUM, ...). Upon registration, the Feature can
+    /// retrieve a `FeatureScope` interface for writing events. The core will store and upload events in an efficient way
+    /// with performance presets defined on initialization.
```
fuzzybinary, line 18:
```suggestion
    /// retrieve a `FeatureScope` interface for writing events to the core. The core will store and upload events in an efficiently
    /// according to the performance presets defined on initialization.
```

Sources/Datadog/DatadogInternal/DatadogCoreProtocol.swift
```diff
@@ -8,23 +8,23 @@ import Foundation
...
     ///
-    /// - Parameter feature: The Feature instance.
+    /// - Parameter feature: The Feature instance - it will be retained and held by core.
     func register(feature: DatadogFeature) throws
 
-    /// Retrieves a Feature by its name and type.
+    /// Retrieves previously registered Feature by its name and type.
     ///
-    /// A Feature type can be specified as parameter or inferred from the return type:
+    /// Feature's type can be specified as parameter or inferred from the return type:
```
fuzzybinary, line 27: I think I prefer 'A Feature' over 'Feature' in a lot of the places you removed it, but here especially.

Sources/Datadog/RUMMonitor.swift
```diff
@@ -148,9 +148,14 @@ public class RUMMonitor: DDRUMMonitor, RUMCommandSubscriber {
 
     /// RUM Attributes shared with other Feature registered in core.
     internal struct Attributes {
+        /// The ID of RUM application (`String`).
         internal static let applicationID = "application_id"
+        /// The ID of current RUM session (standard UUID `String`, lowercased).
+        /// In case the session is rejected (not sampled), RUM context is set to empty (`[:]`) in core.
```
fuzzybinary, line 154: I'm not sure how this comment relates to this property? Is this session_id set to '[:]' if it was not sampled?

session-replay/Sources/DatadogSessionReplay/DatadogCoreIntegration/RUMDependency.swift
```diff
@@ -0,0 +1,33 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+/// Defines dependency between Session Replay and RUM modules.
```
fuzzybinary, line 9: Just for clarity so people can know for sure that SR is referring to Session Replay:
```suggestion
/// Defines dependency between Session Replay (SR) and RUM modules.
```

fuzzybinary (approved):
👍  Just a few minor comments

Sources/Datadog/DatadogInternal/DatadogCoreProtocol.swift
ncreated:
```suggestion
    /// retrieve a `FeatureScope` interface for writing events to the core. The core will store and upload events efficiently
    /// according to the performance presets defined on initialization.
```

Sources/Datadog/DatadogInternal/DatadogCoreProtocol.swift
ncreated:
```suggestion
    /// A Feature type can be specified as parameter or inferred from the return type:
```

Sources/Datadog/DatadogInternal/DatadogCoreProtocol.swift
```diff
@@ -8,23 +8,23 @@ import Foundation
...
     ///
-    /// A Feature collects and transfers data to a Datadog Product (e.g. Logs, RUM, ...). A registered Feature can
-    /// open a `FeatureScope` to write events, the core will then be responsible for storing and uploading events
-    /// in a efficient manner. Performance presets for storage and upload are define when instanciating the core instance.
+    /// Feature collects and transfers data to a Datadog Product (e.g. Logs, RUM, ...). Upon registration, the Feature can
+    /// retrieve a `FeatureScope` interface for writing events. The core will store and upload events in an efficient way
+    /// with performance presets defined on initialization.
     ///
-    /// A Feature can also communicate to other Features by sending message on the bus that is managed by the core.
+    /// The Feature can also communicate to other Features by sending messages on the message bus managed by core.
```
ncreated, line 20:
```suggestion
    /// A Feature can also communicate to other Features by sending messages on the message bus managed by core.
```

Sources/Datadog/RUMMonitor.swift
ncreated: This is interesting problem 👍. You're right and this comment doesn't directly relate to `sessionID` field. It relates to the behaviour that all `applicationID`, `sessionID`, `viewID` and `userActionID` have in common: if session is not sampled, none of these will appear in RUM Feature's context notified by core.

In other words, if session is sampled, core context will be this:
```swift
// ...
featureAttributes: [
   "rum": [
      Attributes.applicationID: "<...>",
      Attributes. sessionID: "<...>",
      Attributes.viewID: "<...>",
      Attributes.userActionID: "<...>",
   ]
]
```
and if it was rejected (not sampled), this:
```
// ...
featureAttributes: [
   "rum": [:]
]
```

Now, because these attributes describe how RUM context will be provided to other Features (Logs, Traces, Session Replay, ...) - it's IMO quite critical to have this behaviour well described and documented. I agree that single `Attributes.sessionID` is not the best place, but it's the only one that I found so far.

Maybe what we need is `RUMContextAttributes` instead of `Attributes` and having this behaviour described at the `struct` level. WDYT @fuzzybinary ?

session-replay/Sources/DatadogSessionReplay/DatadogCoreIntegration/RUMContextReceiver.swift
```diff
@@ -0,0 +1,96 @@
...
+        guard let rumBaggage = message.contextMessage?.rumBaggage else {
+            // No RUM baggage in the message
+            return false
+        }
+
+        // Extract the `RUMContext` or `nil` if RUM session is not sampled:
+        let new = rumBaggage.rumContext
+
+        // Notify only if it has changed:
+        if new != previous {
```
maxep, line 48: I think this should be handle by core: send a message only if the feature attributes changed. Not a change request, I will add a ticket for enhancing `FeatureBaggage`:

1. Enable serialisation/deserialisation of `Codable` in `FeatureBaggage`. It will be **very (very)** convenient for Features to have a validation mechanism such as `Codable` out-of-the-box. It will prevent from having to repeat properties checks in Features.
2. Make `FeatureBaggage` compliant to `Equitable` and on send a message if the baggage changes.

This will greatly simplify your message processing!

maxep (approved):
Very clear and well testing, good job! Thanks for updating the doc 👍

session-replay/Sources/DatadogSessionReplay/DatadogCoreIntegration/RUMContextReceiver.swift
ncreated: Could be - having this done from Core will work for SR as well 👍 and yeah it could be a way to avoid this (☝️) code duplication in other products. As it's not a blocker I will keep it like this for now, as for the "uploader" topic I will **perhaps** need to change this logic anyway (receive entire context instead of just the RUM part).

PR #1038, opened 2022-11-04
RUMM-2662 [SR] Implement data compression

ncreated:
### What and why?

📦 This PR implements data compression that will be later used in Session Replay payloads.

We need a distinct implementation for SR because unlike [`Deflate.encode(_:)`](https://github.com/DataDog/dd-sdk-ios/blob/d2230acdc9ba9defded481f653c33f850bef9a95/Sources/Datadog/Core/Upload/DataCompression.swift#L30) used for RUM, Logs and Traces, SR requires compressing data with [`Z_SYNC_FLUSH`](https://github.com/madler/zlib/blob/cacf7f1d4e3d44d871b605da3b647f07d718623f/zlib.h#L170) + [`Z_FINISH`](https://github.com/madler/zlib/blob/cacf7f1d4e3d44d871b605da3b647f07d718623f/zlib.h#L172) flags passed to zlib's [`deflate()`](https://github.com/madler/zlib/blob/cacf7f1d4e3d44d871b605da3b647f07d718623f/zlib.h#L250). For best results SR team recommends compression level `6`, which is also not available in `Deflate.encode(_:)` (it hardcodes level `5`).

Compressing with `Z_SYNC_FLUSH` and `Z_FINISH` enables SR to efficiently decode succeding payloads with doing `inflate()` only once (in practice it is done automatically by the web browser when streaming the whole replay to SR player). More on this approach can be read in [this SO answer authored by Mark Adler](https://stackoverflow.com/questions/14744692/concatenate-multiple-zlib-compressed-data-streams-into-a-single-stream-efficient#comment51865187_14744792).

### How?

Supporting `Z_SYNC_FLUSH` couldn't be achieved as an enhancement to `Deflate.encode(_:)` - e.g. insertion of arbitrary `0x00 0x00 0xFF 0xFF` bytes at the end. The internal zlib's routine behind `Z_SYNC_FLUSH` is more complex, i.e.:
> If the parameter flush is set to Z_SYNC_FLUSH, all pending output is
  flushed to the output buffer and the output is aligned on a byte boundary, so
  that the decompressor can get all input data available so far.
>
> Ref.: https://github.com/madler/zlib/blob/cacf7f1d4e3d44d871b605da3b647f07d718623f/zlib.h#L286-L288

In order to use it, I had to call zlib APIs directly (unlike `Deflate.encode(_:)` that uses Apple's `Compression` framework as zlib abstraction). Unfortunately, to work with `zlib.h` APIs the solution needs to be implemented in Objective-C as some have no Swift interoperability. To add Objective-C code to SR, I followed the same convention we did for `_DatadogPrivate` module.

A bunch of links that I found useful:
* [Understanding zlib](https://www.euccas.me/zlib/)
* [Usage Example on zlib.net](https://zlib.net/zlib_how.html)
* [zlib.h headers](https://github.com/madler/zlib/blob/cacf7f1d4e3d44d871b605da3b647f07d718623f/zlib.h)
* [A failed attempt to use necessary `zlib.h` APIs in Swift](https://github.com/sjmadsen/SwiftInterop/blob/263545194300e79fb5b1a3321f5d37ea92461f61/SwiftInterop/zlib.swift#L37-L38)

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


maxep:
That's a great finding and I think we can leverage this compression level flexibility in the shared `DataCompression`. We could even remove the use of the `Compression` framework, it's using the following init under the hood: https://developer.apple.com/documentation/compression/compression_zlib

Using c interface of zlib in Swift is also possible, I've tried thew following and it seems to work fine, the trick is to replace `Z_NULL` by `nil`:

```swift
/* public */ internal static func zcompress(_ data: Data, level: Int = 6, chunk: Int = 16_384) throws -> Data {
    // z_streamp->next_in requires mutable bytes
    var m_data = data

    return try m_data.withUnsafeMutableBytes {
        guard let ptr = $0.bindMemory(to: Bytef.self).baseAddress else {
            throw InternalError(description: "zlib compress: failed to bind data memory")
        }

        let buffer = UnsafeMutablePointer<Bytef>.allocate(capacity: chunk)
        defer { buffer.deallocate() }

        let stream = z_streamp.allocate(capacity: 1)
        defer { stream.deallocate() }

        stream.pointee.next_in = ptr
        stream.pointee.avail_in = uInt(data.count)
        stream.pointee.total_out = 0
        stream.pointee.zalloc = nil // Z_NULL: nil points to addr 0x0, equivalent of Z_NULL
        stream.pointee.zfree = nil // Z_NULL
        stream.pointee.opaque = nil // Z_NULL

        var result = deflateInit_(stream, Int32(level), ZLIB_VERSION, Int32(MemoryLayout<z_stream>.size))
        defer { deflateEnd(stream) } // free the allocated zlib state

        guard result == Z_OK else {
            throw InternalError(description: "zlib error: \(zerror(result))")
        }

        var c_data = Data()
        repeat {
            stream.pointee.next_out = buffer
            stream.pointee.avail_out = uInt(chunk)

            let flush = stream.pointee.avail_in == 0 ? Z_FINISH : Z_SYNC_FLUSH // if no more data to compress
            result = deflate(stream, flush)

            let count = chunk - Int(stream.pointee.avail_out)
            c_data.append(buffer, count: count)
        } while result == Z_OK

        guard result == Z_STREAM_END else {
            throw InternalError(description: "zlib error: \(zerror(result))")
        }

        return c_data
    }
}

private static func zerror(_ code: Int32) -> String {
    switch code {
    case Z_ERRNO:
        return "Z_ERRNO"
    case Z_STREAM_ERROR:
        return "Z_STREAM_ERROR"
    case Z_DATA_ERROR:
        return "Z_DATA_ERROR"
    case Z_MEM_ERROR:
        return "Z_MEM_ERROR"
    case Z_BUF_ERROR:
        return "Z_BUF_ERROR"
    case Z_VERSION_ERROR:
        return "Z_VERSION_ERROR"
    default:
        return "unknown error (\(code))"
    }
}
```

It would be best to not rely on another objc module, LMKWYT!

ncreated:
> Using c interface of zlib in Swift is also possible, I've tried thew following and it seems to work fine, the trick is to replace `Z_NULL` by `nil`

Nice try, thanks for porting my efforts 🏋️‍♀️ 🤜🤛 🚀! I totally agree with this part:

> It would be best to not rely on another objc module, LMKWYT!

so if above works ☝️, I'll be more than okay with it. Give me some time to test it end-to-end with compressing actual uploads. For that, I need to first complete the uploader part (before I was testing by hacking the uploader, but ATM I'm almost done with the final implementation). We can keep this PR open until then and merge everything at once.

session-replay/Sources/DatadogSessionReplay/Recorder/Recorder.swift
```diff
@@ -45,7 +45,7 @@ internal class Recorder {
         self.init(
             configuration: configuration,
             rumContextObserver: rumContextObserver,
-            scheduler: MainThreadScheduler(interval: 0.25),
+            scheduler: MainThreadScheduler(interval: 0.1),
```
ncreated, line 48: This change is on purpose - `0.1s` frequency just looks better in the player produces more data for compression. We will work on finding the best value in dogfooding - for now consider this one arbitrary and temporary.

ncreated (commented):
@maxep Thank you again for porting this to Swift. It works great and gives the same performance 👍.

One additional thing I discovered when testing different recording frequencies (so different size of payloads to compress) is that the `chunk` size must be a dynamic value. I applied the fix and now it seems to work stable (player reads segments no matter of the freq.).

maxep (approved):
Awesome work 🚀

PR #1039, opened 2022-11-04
RUMM-2628 Add configuration events to telemetry

fuzzybinary:
### What and why?

This reports the current of the SDK back to Datadog after the app has been running for 5 seconds. This delay allows us to gather information about configuration items that are setup after the call to "initialize" (this is common in the Cross Platform frameworks).

There is additionally an internal mapper to allow modification of the configuration event to add / modify items specific to Cross Platform frameworks.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


fuzzybinary:
Created as Draft until the schema changes are finalized.

Tests/DatadogTests/Datadog/RUM/RUMTelemetryTests.swift
```diff
@@ -260,6 +262,60 @@ class RUMTelemetryTests: XCTestCase {
...
+        // When
+        telemetry.configuration(configuration: .mockAny())
+
+        // Then immediately
+        var eventMatchers = try RUMFeature.waitAndReturnRUMEventMatchers(in: core, count: 0)
+        var events = try eventMatchers.compactMap(TelemetryConfigurationEvent.self)
+        XCTAssertEqual(events.count, 0)
+
+        // Then later
+        sleep(6)
```
ncreated, line 280: Here we add `6s` to our unit tests execution time (`12s` in total). We rather aim at keeping our tests fast (they all run under `50s` on `develop` so we're adding ~25% here). 

How about abstracting the async schedule and providing instant call here? Instead of:
```swift
DispatchQueue.main.asyncAfter(deadline: .now() + 5) { job() }
```
we could do:
```swift
class RUMTelemetry {
   /// Triggers given block asynchronously after 5s
   private let dispatchLater: (@escaping () -> Void) -> Void

   // ...
   
   func configuration(...) {
      dispatchLater { self.record(...) }
   }
}
```
And configure it async / sync respectively to the use case to avoid any wait in tests:
```swift
// async in code:
let telemetry = RUMTelemetry(
   dispatchLater: { block in DispatchQueue.main.asyncAfter(deadline: .now() + 5) { block() }
}

// sync in tests:
let telemetry = RUMTelemetry(
   dispatchLater: { block in block() }
}
```

ncreated (changes requested):
Looks really good: `FeaturesConfiguration` is the great place to hook-in and makes the solution look elegant 💼👌.

I left few feedbacks with the note on `sleep()` in tests being considered a blocker. LTM what you think on my proposed idea @fuzzybinary 

Also, keep in mind that the mapper API won't be much usable w/o making relevant fields in `RUMDataModels` (in JSON schema) mutable. This can be done separately tho 👍.

Sources/Datadog/RUM/RUMTelemetry.swift
```diff
@@ -5,6 +5,9 @@
  */
 
 import Foundation
+import UIKit
```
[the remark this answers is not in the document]
fuzzybinary: I'll double check. I was curious about that too but just assumed it was because of something in the automatic view detection... but it's possible a typo caused it to auto import.

Tests/DatadogTests/Datadog/Mocks/CoreMocks.swift
```diff
@@ -709,7 +709,10 @@ internal class InMemoryWriter: AsyncWriter {
             fatalError()
         }
 
-        return queue.sync { events }
+        return queue.sync {
+            waitAndReturnEventsDataExpectation = nil
+            return events
+        }
```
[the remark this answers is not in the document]
fuzzybinary: Basically you can't call this method twice unless you clear this out.

Tests/DatadogTests/Datadog/RUM/RUMTelemetryTests.swift
fuzzybinary: Yeah, I like that idea. I'll get that in shortly.

ncreated:
> > Also, keep in mind that the mapper API won't be much usable w/o making relevant fields in `RUMDataModels` (in JSON schema) mutable. This can be done separately tho 👍.
> 
> Is there a way to have the generator create mutable fields? I was just going to create a "copyWith" function in Flutter to return a new copy with only specific fields modified...

This is configured thru schema, not generator. Changing certain properties to `"readOnly": false` will do the trick, e.g.:

https://github.com/DataDog/rum-events-format/blob/ba3b9c6132c9cdecdfe1f846fe8a03c18ce091d5/schemas/rum/_common-schema.json#L91-L95

Sources/Datadog/RUM/RUMTelemetry.swift
```diff
@@ -156,3 +214,57 @@ internal final class RUMTelemetry: Telemetry {
...
+            viewTrackingStrategy: nil
+        )
+    }
+}
+
+private extension Float {
+    func toInt64() -> Int64 {
+        return Int64(self)
+    }
+}
```
ncreated, line 270: nit/ We use it to convert user input (sampling rate), so I wonder if this can ever overflow causing a crash `5s` after app starts. Seems safe, [as `Sampler` does sanitization and caps the value](https://github.com/DataDog/dd-sdk-ios/blob/e34d9e69f713e209ce4af89a92a059a7b9810b74/Sources/Datadog/Core/Utils/Sampler.swift#L15).

For sanity, we could use [this helper](https://github.com/DataDog/dd-sdk-ios/blob/e34d9e69f713e209ce4af89a92a059a7b9810b74/Sources/Datadog/Utils/SwiftExtensions.swift#L89) and do:
```swift
func toInt64() -> Int64? {
   return try? Int64(withReportingOverflow: self)
}
```
Better to not have the value than crash.

Tests/DatadogTests/Datadog/Mocks/CoreMocks.swift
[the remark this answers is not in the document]
ncreated: Ah, I see - makes sense 👍

ncreated (approved):
Looks great 👌, thank you! I left a minor suggestion but it's good to go anyway.

Sources/Datadog/RUM/RUMTelemetry.swift
fuzzybinary: Yeah I'll add in that change.

PR #1044, opened 2022-11-08
RUMM-2662 [SR] Write Session Replay records to `DatadogCore`

ncreated:
### What and why?

📦 This PR makes the Session Replay module write its events (records) to `DatadogCore`.

Later (next PR) events will be read back and mapped into a sequence of `URLRequests` requested by `DatadogCore` before performing SR data upload. 

### How?

Referring to the architecture discussed in SR RFC, this PR introduces the last component of the pipeline: the `Writer`:
```swift
/// A type writing Session Replay records to `DatadogCore`.
internal protocol Writing {
    /// Connects writer to the `FeatureScope` managed by `DatadogCore`.
    func startWriting(to featureScope: FeatureScope)

    /// Writes next records to `DatadogCore`.
    func write(nextRecord: EnrichedRecord)
}
```

Zooming out to sum up all the work done so far in SR, the overall architecture is now complete:
```
Recorder (`ViewTreeSnapshot`) → Processor (`[SRRecord]`) → Writer (write to V2 `DatadogCore`)
```

Before being written to `DatadogCore`, `SRRecords` must be enriched with additional RUM context and other meta information necessary to (later) combine them into distinct `SRSegments` (one per RUM context). This is achieved with `EnrichedRecord` type (a counterpart of [`EnrichedRecord` in `dd-sdk-android`](https://github.com/DataDog/dd-sdk-android/blob/feature/session-replay-vo/library/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/processor/EnrichedRecord.kt)):
```swift
internal struct EnrichedRecord: Encodable {
   // .. RUM context and other meta information ...

   // The actual records
   let records: [SRRecord]
}
```

💡 As part of this task, I had to expose necessary `FeatureScope` interface in V2. It is now made `public`.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


session-replay/Tests/DatadogSessionReplayTests/Writer/WriterTests.swift
```diff
@@ -0,0 +1,24 @@
...
+import XCTest
+@testable import DatadogSessionReplay
+
+// swiftlint:disable empty_xctest_method
+class WriterTests: XCTestCase {
+    func testWhenFeatureScopeIsConnected_itWritesRecordsToCore() {
+        // TODO: RUMM-2690
+        // Implementing this test requires creating mocks for `DatadogContext` (passed in `FeatureScope`),
+        // which is yet not possible as we lack separate, shared module to facilitate tests.
+    }
```
ncreated, line 16: What I mean here is that `Writer` depends on V2 `FeatureScope` which exposes the _event write context_:
```swift
func eventWriteContext(bypassConsent: Bool, _ block: @escaping (DatadogContext, Writer) throws -> Void)
```
In order to create `FeatureScopeMock: FeatureScope` that records what is being written, we have to mock `DatadogContext` (e.g. `.mockAny()`). Mocking it isn't yet straightforward as `DatadogContext` depends on several structures internal to `Datadog`. I don't want to make it all public as part of this PR - I believe it's best to be done through separate module that facilitates tests.

maxep: Agreed! That will be the role of the `TestUtilities` module, but it's not yet available.

session-replay/Sources/DatadogSessionReplay/Drafts/SessionReplay.swift
```diff
@@ -14,11 +14,22 @@ public struct SessionReplay {
...
+        do {
+            let feature = SessionReplayFeature(configuration: configuration)
+            try datadogInstance.register(feature: feature)
 
-        return sessionReplay
+            guard let scope = datadogInstance.scope(for: feature.name) else {
+                throw FatalError(description: "Failed to obtain `FeatureScope` for \(feature.name)")
+            }
+
+            feature.register(sessionReplayScope: scope)
```
maxep, line 26: 🥚 or 🐔 ? You need the scope in the Feature, but you need to register the Feature to get a scope. I think this should be solved by the core, but I don't have a simple solution 🤔 
I was expecting to keep ref to the core instance in Features, but then we would have a retain circle. Anyway, your usage is valid 👍

session-replay/Sources/DatadogSessionReplay/Utilities/Queue.swift
```diff
@@ -17,3 +17,15 @@ internal struct MainAsyncQueue: Queue {
         queue.async { block() }
     }
 }
+
+internal struct BackgroundAsyncQueue: Queue {
+    private let queue: DispatchQueue
+
+    init(named queueName: String) {
+        self.queue = DispatchQueue(label: "com.datadoghq.session-replay.processor", qos: .utility)
+    }
```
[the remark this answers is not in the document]
ncreated: Ops, good catch 👍. ✅.

session-replay/Sources/DatadogSessionReplay/Writer/Models/EnrichedRecord.swift
```diff
@@ -0,0 +1,86 @@
...
+
+// MARK: - Convenience
+
+private extension SRRecord {
+    var isFullSnapshot: Bool {
+        switch self {
+        case .fullSnapshotRecord: return true
+        default: return false
+        }
+    }
```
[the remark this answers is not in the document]
ncreated: Unfortunately, records are enum cases with associated type, hence we need pattern matching.

Screenshot 2022-11-10 at 14 01 52

session-replay/Sources/DatadogSessionReplay/Drafts/SessionReplay.swift
ncreated: Yes, it's exactly like you describe. From my point of view, this is okay for now 👍 - it doesn't produce much boilerplate nor extreme inconvenience. The main thing I dislike is that it enforces optionals in Feature implementation because we can't set things on initialization. This however is IMO the price we have to pay for having Features decoupled at the public API level:
```swift
Core.initialize()

Feature1.initialize()
Feature2.initialize() 
```
Coupling them together:
```swift
Core.initialize(with: [
   Feature1.initialize()
   Feature2.initialize() 
])
```
could be a way of injecting `featureScope` back to the Feature further in Core's `initialize()` implementation.

I like no of copuling more than I dislike optionals ⚖️ 🙂.

session-replay/Sources/DatadogSessionReplay/Drafts/SessionReplayFeature.swift
```diff
@@ -22,30 +22,35 @@ internal class SessionReplayFeature: DatadogFeature, SessionReplayController {
...
-        let rumContextReceiver = RUMContextReceiver()
-        let recorder = Recorder(
-            configuration: configuration,
-            rumContextObserver: rumContextReceiver,
-            processor: Processor()
+    init(configuration: SessionReplayConfiguration) {
+        let writer = Writer()
+
+        let processor = Processor(
+            queue: BackgroundAsyncQueue(named: "com.datadoghq.session-replay.processor"),
```
maciejburda, line 35: Just curious if there's a huge performance difference between sync and asynch queue here?

session-replay/Sources/DatadogSessionReplay/Writer/Models/EnrichedRecord.swift
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// MARK: - Convenience
+
+private extension SRRecord {
+    var isFullSnapshot: Bool {
+        switch self {
+        case .fullSnapshotRecord: return true
+        default: return false
+        }
```
maciejburda, line 73:
```suggestion
        if case let .fullSnapshotRecord = self {
            return true
        } else {
            return false
        }
```

After writing it - I prefer switch :D

session-replay/Sources/DatadogSessionReplay/Writer/Models/SRDataModels.swift
```diff
@@ -20,7 +20,7 @@ internal struct SRSegment: SRDataModel {
     internal let hasFullSnapshot: Bool?
 
     /// The index of this Segment in the segments list that was recorded for this view ID. Starts from 0.
-    internal let indexInView: Int64
+    internal let indexInView: Int64?
```
maciejburda, line 23: maybe we could add documentation why it could be nil?

maciejburda (approved):
Not much to add. Just some nits.

LGTM!

session-replay/Sources/DatadogSessionReplay/Writer/Models/SRDataModels.swift
ncreated: It's part of the `SRDataModels`, so everything (including the comment) comes from JSON schema and is code-generated. This field was [made optional for Mobile SR purpose](https://github.com/DataDog/rum-events-format/pull/100) - but AFAIK the goal is to provide value for it (_RUMM-2518: Correctly resolve or find different alternative for the `indexInView` segment property_). I'll keep it as it is for now - I'm not sure on the direction we'll take on, but the work is scheduled and tracked in JIRA.

session-replay/Sources/DatadogSessionReplay/Writer/Models/EnrichedRecord.swift
ncreated: So do I 😄

Issue #922, opened 2022-07-08
Setting sample rate?


dweiner13:
### The issue

The android version of the SDK allows us to set the sample rate for logs via a `setSampleRate` function on the logger builder (https://docs.datadoghq.com/logs/log_collection/android/?tabs=java#logger-initialization).

Is there similar functionality for the iOS SDK? I can't find it anywhere.

maxep:
Hi @dweiner13 ! Thank you for the report and sorry for the late reply..

Indeed, unlike Android we do not provide sampling for logs on iOS, we should work on aligning those. I will keep you posted 👍

dweiner13:
@maxep Thank you! For now it's easy enough to manually write that functionality, so no huge rush from my end.

maciejburda:
Hey @dweiner13 

Logging sampling is now available in the SDK.
https://github.com/DataDog/dd-sdk-ios/pull/1045

PR #583, opened 2021-09-01
RUMM-1582 Fix `DDRUMView` and `DDRUMAction` APIs for Objective-C

ncreated:
### What and why?

🐞 As noticed in https://github.com/DataDog/dd-sdk-ios/issues/577#issuecomment-908500889, `DDRUMView` and `DDRUMAction` were not properly exposed to Objective-C interface.

This PR targets the `1.7.0-beta4` branch, as IMO this is no-risk but high impact fix that can be shipped in `1.7.0`.

### How?

Swift definitions of `DDRUMView` and `DDRUMAction` were missing `@objc` annotation in their `public init()`. Definition for `DDRUMView` was there, but it was accidentally removed in [this refactoring](https://github.com/DataDog/dd-sdk-ios/commit/d705147efa91415009cd57f97017b154958b9f43#diff-9cfc39ecf2272240473cb67d32dd88a47848e360c5a6aea3cf8ee0a6a3fce8d6).

#### Objective-C API tests

Because unexposed or broken Objective-C APIs are recurring problem (https://github.com/DataDog/dd-sdk-ios/pull/359, https://github.com/DataDog/dd-sdk-ios/pull/360, https://github.com/DataDog/dd-sdk-ios/pull/461) I did an improvement to our `DatadogObjc` unit tests by adding smoke tests covering the public interface in Objective-C, e.g.:

```objc
// DDConfiguration+apiTests.m

- (void)testDDConfigurationBuilderAPI {
    DDConfigurationBuilder *builder = [DDConfiguration builderWithClientToken:@"" environment:@""];
    // ...
    [builder trackUIKitRUMViews];
    [builder trackUIKitRUMViewsUsing:[CustomDDUIKitRUMViewsPredicate new]];
    [builder setRUMViewEventMapper:^DDRUMViewEvent * _Nonnull(DDRUMViewEvent * _Nonnull viewEvent) {
        viewEvent.view.url = @"";
        return viewEvent;
    }];
    // ...
    [builder setWithBatchSize:DDBatchSizeMedium];
    [builder build];
}
```

From now on, when we expose a new API in `DatadogObjc`, we can preview its Objective-C representation right in the test file. Adding test would prevent regression on eventual refactoring and ensure that the API is exposed properly.

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference

PR #1045, opened 2022-11-08
RUMM-303 Logging sampling rate

maciejburda:
### What and why?

Adds capability of sampling logs. It was pointed out in the [Github issue](https://github.com/DataDog/dd-sdk-ios/issues/922) that this feature already [exists in Android](https://docs.datadoghq.com/logs/log_collection/android/?tab=java#logger-initialization).

### How?

Added implementation allows configuring sampling rate for logging both globally from configuration and locally for logger using builder.

### Review checklist

- [x] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


Sources/Datadog/Core/FeaturesConfiguration.swift
```diff
@@ -181,9 +182,16 @@ extension FeaturesConfiguration {
         )
 
         if configuration.loggingEnabled {
+            let loggingSampler: Sampler?
+            if let loggingSamplingRate = configuration.loggingSamplingRate {
+                loggingSampler = Sampler(samplingRate: loggingSamplingRate)
+            } else {
+                loggingSampler = nil
+            }
```
maxep, line 190: You can do a one-liner here :)
```suggestion
            let loggingSampler = configuration.loggingSamplingRate.map { Sampler(samplingRate: $0) }
```

But it would be even easier if `configuration.loggingSamplingRate` is not optional.

Sources/Datadog/DatadogConfiguration.swift
```diff
@@ -265,6 +265,7 @@ extension Datadog {
         private(set) var firstPartyHosts: Set<String>?
         private(set) var logEventMapper: LogEventMapper?
         private(set) var spanEventMapper: SpanEventMapper?
+        private(set) var loggingSamplingRate: Float?
```
maxep, line 268: We can make this non-optional, to align with other sampling rates in the config.

Sources/Datadog/Logger.swift
```diff
@@ -398,24 +407,70 @@ public class Logger: LoggerProtocol {
...
+            } else {
+                return logger
+            }
         }
     }
 }
 
-/// Combines multiple loggers together into single `Logger` interface.
+/// Wraps a `LoggerProtocol` and enables sampling using the injected instance of `Sampler`.
+internal struct SamplingLogger: LoggerProtocol {
```
maxep, line 437: I don't think you need this wrapper, the sampling should be applied when using the `RemoteLogger`. The `ConsoleLogger` shouldn't be sampled so all logs are printed in the console.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 - [IMPROVEMENT] Add a method to add properties to user info without needing to reset other user info..
 - [IMPROVEMENT] Send trace sample rate (`dd.rulePsr`) for APM's traffic ingestion control page. See [#1029][]
+- [IMPROVEMENT] Add logging sampling. See [#1045][]
```
maxep, line 5: We also needs to ref the PR url at the bottom of the changelog 👍

maxep (changes requested):
Thanks, @maciejburda 👏 

It's great to be able to have a global sampling rate in `Datadog.Configuration` and be able to override that in any `Logger` instance 👍 Also, the tests and objc interface looks great!

I see 2 things:
1. We can align with other sampling rate and make it non-optional, it will simplify things.
2. The sampling rate should be applied to `RemoteLogger` only, but all log entries to the `ConsoleLogger` should print.

Sources/DatadogObjc/DatadogConfiguration+objc.swift
```diff
@@ -331,6 +331,11 @@ public class DDConfigurationBuilder: NSObject {
         _ = sdkBuilder.set(serviceName: serviceName)
     }
 
+    @objc
+    public func set(loggingSamplingRate: Float) {
+        _ = sdkBuilder.set(loggingSamplingRate: loggingSamplingRate)
+    }
```
ncreated, line 337: We have some bad memories (https://github.com/DataDog/dd-sdk-ios/pull/583, https://github.com/DataDog/dd-sdk-ios/pull/359, https://github.com/DataDog/dd-sdk-ios/pull/360, https://github.com/DataDog/dd-sdk-ios/pull/461) with adding new APIs through `@objc` and not having them available in Objective-C. To anticipate problems, we added smoke tests in [DDConfiguration+apiTests.m](https://github.com/DataDog/dd-sdk-ios/blob/feature/session-replay/Tests/DatadogTests/DatadogObjc/ObjcAPITests/DDConfiguration%2BapiTests.m) - where each new Obj-c API can be called in `.m` file. Let's include this new `set(loggingSamplingRate` there, for 🧠 sanity 🙌.

ncreated (commented):
Really well done 🏅👍 - all changes make sense and tests coverage is nice 👌.

I left few minor feedbacks, but I don't see major blockers. Good to go after addressing @maxep 's feedback 🚀.

Sources/Datadog/Logger.swift
maxep: Yes, I think we should continue to print everything to the console.

That make me think that we would also need to force `sample rate = 100` when `debugOverride` is true. Like we do with other Features, e,g: https://github.com/DataDog/dd-sdk-ios/blob/develop/Sources/Datadog/Core/FeaturesConfiguration.swift#L217

maciejburda:
@ncreated @maxep I addressed all the comments. Please give it another round 🙏

ncreated (approved):
Awesome 🚀 - looks good and reads easy 👍.

maxep (approved):
Looks great, thank you 👍

PR #1046, opened 2022-11-09
RUMM-2662 [SR] Upload Session Replay records (batched into segments)

ncreated:
### What and why?

📦 This PR implements Session Replay data uploads. SR records previously (#1044) written to `DatadogCore` are now read back and mapped into `URLRequests` sent to SR Intake.

#### Challenge 1
✋ Because `DatadogCore` only supports `[Data] -> URLRequest` mapping (not yet `[Data] -> [URLRequest]`), I had to implement a temporary workaround. The first request is sent through `DatadogCore` (🏅), but other are passed to arbitrary (and temporary) uploader managed by SR.

#### Challenge 2
Another problem I had to solve in this PR relates to decoding [`[SRRecord]`](https://github.com/DataDog/dd-sdk-ios/blob/253deba65b9c5da89875e2fce7a698930e5910df/session-replay/Sources/DatadogSessionReplay/Writer/SRDataModels.swift#L775) from events `[Data]` received from core. Even though our [`SRDataModels`](https://github.com/DataDog/dd-sdk-ios/blob/feature/session-replay/session-replay/Sources/DatadogSessionReplay/Writer/SRDataModels.swift) are declared `Codable`, their decoding through `Swift.Decodable` cannot be utilised for SR uploads. Our SR JSON schemas turn out to be too complex for the [naive `Decodable` conformance we generate](https://github.com/DataDog/dd-sdk-ios/blob/86a72050d02b4b3bc9e3f0ac5b3c8af1e747f25e/tools/rum-models-generator/Tests/CodeGenerationTests/Print/SwiftPrinterTests.swift#L525-L549). This leads to ambiguity when deserializing `Data` into our Swift types - a single JSON value can be decoded to more than one SR data model, which in turn makes _"records to segments"_ mapping unreliable.

By discussing this with @maxep we concluded two possible approaches:
* We could improve [rum-models-generator](https://github.com/DataDog/dd-sdk-ios/tree/86a72050d02b4b3bc9e3f0ac5b3c8af1e747f25e/tools/rum-models-generator) to generate `Decodable` conformance that relies on `type` constants declared consistently in each schema.
* Rely either on Apple's `JSONSerialization` or our [`CodableValue: Codable`](https://github.com/DataDog/dd-sdk-ios/blob/253deba65b9c5da89875e2fce7a698930e5910df/Sources/Datadog/Core/Utils/CodableValue.swift#L9-L10) ereasure to read `[SRRecord]` data back.

Discussed enhancement to `rum-models-generator` is not a simple task, so I followed the second approach instead. Because working with `CodableValue` is not very convenient (it wasn't invented for inspecting data, but only to preserve its JSON representation) and because [`JSONSerialization` is slightly faster than `Codable`](https://flight.school/articles/benchmarking-codable/) I chosed to go with this old fashioned way of coding objects in iOS.

This PR introduces 2 data models:
* `EnrichedRecordJSON` - to decode `EnrichedRecord: Encodable` and encode it back
* `SegmentJSON` - to encode [`SRSegment`](https://github.com/DataDog/dd-sdk-ios/blob/253deba65b9c5da89875e2fce7a698930e5910df/session-replay/Sources/DatadogSessionReplay/Writer/SRDataModels.swift#L12) information through `JSONSerialization`.

Even if staying away from code-generated `SRSegment`, the `SegmentJSON` uses its `CodingKeys` for type safety. I also added strong fuzzy tests to ensure that both `SRSegment` and `SegmentJSON` can be coded back and forth with the same `Data`.

---

After solving both problems, the final workflow of SR records and its integration with `DatadogCore` can be explained this:

Screenshot 2022-11-09 at 12 50 41

On the `Processor` side, we write enriched (with RUM context) records to `DatadogCore`. Later, upon requesting upload records are retrieved back, grouped (by RUM context) into segments and uploaded to Session Replay Intake. 🏁 

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/DatadogCore/Upload/DataUploadWorker.swift
+++ b/Sources/Datadog/DatadogCore/Upload/DataUploadWorker.swift
@@ -62,32 +62,38 @@ internal class DataUploadWorker: DataUploadWorkerType {
             if let batch = nextBatch {
                 DD.logger.debug("⏳ (\(self.featureName)) Uploading batch...")
 
-                // Upload batch
-                let uploadStatus = self.dataUploader.upload(
-                    events: batch.events,
-                    context: context
-                )
+                do {
+                    // Upload batch
+                    let uploadStatus = try self.dataUploader.upload(
+                        events: batch.events,
+                        context: context
+                    )
 
-                // Delete or keep batch depending on the upload status
-                if uploadStatus.needsRetry {
-                    self.delay.increase()
+                    // Delete or keep batch depending on the upload status
+                    if uploadStatus.needsRetry {
+                        self.delay.increase()
 
-                    DD.logger.debug("   → (\(self.featureName)) not delivered, will be retransmitted: \(uploadStatus.userDebugDescription)")
-                } else {
+                        DD.logger.debug("   → (\(self.featureName)) not delivered, will be retransmitted: \(uploadStatus.userDebugDescription)")
+                    } else {
+                        self.fileReader.markBatchAsRead(batch)
+                        self.delay.decrease()
+
+                        DD.logger.debug("   → (\(self.featureName)) accepted, won't be retransmitted: \(uploadStatus.userDebugDescription)")
+                    }
+
+                    switch uploadStatus.error {
+                    case .unauthorized:
+                        DD.logger.error("⚠️ Make sure that the provided token still exists and you're targeting the relevant Datadog site.")
+                    case let .httpError(statusCode: statusCode):
+                        DD.telemetry.error("Data upload finished with status code: \(statusCode)")
+                    case let .networkError(error: error):
+                        DD.telemetry.error("Data upload finished with error", error: error)
+                    case .none: break
+                    }
+                } catch let error {
+                    // If upload can't be initiated do not retry, so drop the batch:
                     self.fileReader.markBatchAsRead(batch)
-                    self.delay.decrease()
-
-                    DD.logger.debug("   → (\(self.featureName)) accepted, won't be retransmitted: \(uploadStatus.userDebugDescription)")
-                }
-
-                switch uploadStatus.error {
-                case .unauthorized:
-                    DD.logger.error("⚠️ Make sure that the provided token still exists and you're targeting the relevant Datadog site.")
-                case let .httpError(statusCode: statusCode):
-                    DD.telemetry.error("Data upload finished with status code: \(statusCode)")
-                case let .networkError(error: error):
-                    DD.telemetry.error("Data upload finished with error", error: error)
-                case .none: break
+                    DD.telemetry.error("Failed to initiate '\(self.featureName)' data upload", error: error)
                 }
             } else {
                 let batchLabel = nextBatch != nil ? "YES" : (isSystemReady ? "NO" : "NOT CHECKED")
@@ -117,7 +123,7 @@ internal class DataUploadWorker: DataUploadWorkerType {
     internal func flushSynchronously() {
         queue.sync {
             while let nextBatch = self.fileReader.readNextBatch() {
-                _ = self.dataUploader.upload(events: nextBatch.events, context: contextProvider.read())
+                _ = try? self.dataUploader.upload(events: nextBatch.events, context: contextProvider.read())
                 self.fileReader.markBatchAsRead(nextBatch)
             }
         }
--- a/Sources/Datadog/DatadogInternal/Upload/URLRequestBuilder.swift
+++ b/Sources/Datadog/DatadogInternal/Upload/URLRequestBuilder.swift
@@ -6,16 +6,18 @@
 
 import Foundation
 
+internal typealias URLRequestBuilder = DDURLRequestBuilder
+
 /// Builds `URLRequest` for sending data to Datadog.
-/* public */ internal struct URLRequestBuilder {
-    enum QueryItem {
+public struct DDURLRequestBuilder {
+    public enum QueryItem {
         /// `ddsource={source}` query item
         case ddsource(source: String)
         /// `ddtags={tag1},{tag2},...` query item
         case ddtags(tags: [String])
     }
 
-    struct HTTPHeader {
+    public struct HTTPHeader {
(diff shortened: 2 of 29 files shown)


Sources/Datadog/DatadogCore/Upload/DataUploadWorker.swift
```diff
@@ -62,32 +62,38 @@ internal class DataUploadWorker: DataUploadWorkerType {
...
-                    self.delay.increase()
-
-                    DD.logger.debug("   → (\(self.featureName)) not delivered, will be retransmitted: \(uploadStatus.userDebugDescription)")
-                } else {
+                do {
+                    // Upload batch
+                    let uploadStatus = try self.dataUploader.upload(
+                        events: batch.events,
+                        context: context
+                    )
```
ncreated, line 70: All this diff is because I had to make `self.dataUploader.upload()` a `throwing` method..

Sources/Datadog/DatadogInternal/Upload/URLRequestBuilder.swift
```diff
@@ -6,16 +6,18 @@
 
 import Foundation
 
+internal typealias URLRequestBuilder = DDURLRequestBuilder
+
 /// Builds `URLRequest` for sending data to Datadog.
-/* public */ internal struct URLRequestBuilder {
-    enum QueryItem {
+public struct DDURLRequestBuilder {
+    public enum QueryItem {
```
ncreated, line 13: I'm doing the trick with `typealias` here to avoid renaming surgery in core. The SDK still uses `URLRequestBuilder` name, but in public API it's `DDURLRequestBuilder` to avoid ambiguity.

Sources/Datadog/DatadogInternal/Upload/URLRequestBuilder.swift
```diff
@@ -100,27 +111,31 @@ import Foundation
...
-    /// Creates `URLRequest` for uploading given `data` to Datadog.
-    /// - Parameter data: data to be uploaded
-    /// - Returns: Returns: the `URLRequest` object.
-    func uploadRequest(with data: Data) -> URLRequest {
+    /// Creates `URLRequest` for uploading given `body` to Datadog.
+    ///
+    /// - Parameter body: HTTP body to be attached to request
+    /// - Parameter compress: if `body` should be compressed into ZLIB Compressed Data Format (IETF RFC 1950)
+    /// - Returns: the `URLRequest` object.
+    public func uploadRequest(with body: Data, compress: Bool = true) -> URLRequest {
```
ncreated, line 119: Adding `Bool` argument is not a perfect way, but unlocks me to move forward with introducing limited complexity to the SDK. We could think of refactoring this later, e.g. making `RequestBuilder` only prepare pre-configured `URLRequest` that Features set `.httpBody` on in their own routines.

session-replay/Sources/DatadogSessionReplay/DatadogCoreIntegration/RequestBuilder/JSON/SegmentJSONBuilder.swift
```diff
@@ -0,0 +1,108 @@
...
+
+        // Current segment state:
+        var startIndex = 0
+        var hasFullSnapshot = current.hasFullSnapshot
+        var startTime = current.earliestTimestamp
+        var endTime = current.latestTimestamp
+
+        // Loop through the array to find continuous ranges of records that have
+        // the same RUM context. For each range create a segment:
+        for index in (1..<records.count) {
```
ncreated, line 36: This could be done fancier with functional operators, however `createSegmentJSONs(from)` is part of the critical path in SR uploads and needs to loop through all batched events, so I wanted to keep it fast and performant.

maciejburda (approved):
Gave it glance. LGTM 🙌

Lemme me know if you need more thorough review :)

Sources/Datadog/DatadogCore/Upload/DataUploadWorker.swift
maxep: Good move 👍

Sources/Datadog/DatadogInternal/Upload/URLRequestBuilder.swift
maxep: Even tho this is marked public, it will be exposed by `DatadogInternal` which shouldn't be imported by users. Importing this module results in using private APIs, so we don't have to care much about ambiguity.

Sources/Datadog/DatadogInternal/Upload/URLRequestBuilder.swift
maxep: I agree with that, tho we might need custom headers sometimes. This could return a custom object instead of a `URLRequest`. Will see when we refactor the upload 👍 

In which condition SR can decide to not compress the data?

maxep (approved):
looks great!! I've left some minor feedback/question, but it's good to merge.

The custom decoding is not ideal, but it's super clear and well isolated 👍

Sources/Datadog/DatadogInternal/Upload/URLRequestBuilder.swift
ncreated: Indeed 👍. IMO sounds good to revert this rename once we have `DatadogInternal` module separated (not the case right now, as we will be exposing `DDURLRequestBuilder` in SDK's public API).

maxep: Agreed 👍

Sources/Datadog/DatadogInternal/Upload/URLRequestBuilder.swift
ncreated: > In which condition SR can decide to not compress the data?

All the time - SR does its own compression before making this call. It always sets `compress: false` and passes _"zlib w/ flush sync"_ compressed `data` as `body` to `uploadRequest(with:compress:)`.

PR #1047, opened 2022-11-10
RUMM-2662 [SR] Set `hasReplay` flag on RUM events

ncreated:
### What and why?

📦 This PR delivers the last 🧩 element of Session Replay upload pipeline. It integrates SR data with RUM: If any RUM event is collected while SR module is enabled, it will have `hasReplay` flag set accordingly to the recording state (`true` | `false`).

This flag is crucial for RUM <> SR correlation, i.a.: thanks to this value we can display ▶️ button next to events that can be referenced from SR replays. Clicking on it will jump directly to this event within recording:

Screenshot 2022-11-10 at 08 10 28

### How?

When recorder's state changes, SR updates its `FeatureBaggage` in `DatadogContext`. This baggage is read by RUM when events are created.

For clarity, and to continue convention started with [`RUMDependency`](https://github.com/DataDog/dd-sdk-ios/blob/feature/session-replay/session-replay/Sources/DatadogSessionReplay/DatadogCoreIntegration/RUMDependency.swift), I added `SessionReplayDependency` structure to RUM. Both structures describe the contract from one module to another.

For having max correlation between RUM evens and SR, the `hasReplay` information propagates also to crash reporting flow. This requires persisting this flag within the "last RUM session state" that gets encoded along with the crash report. Later, when creating the actual RUM error, we make our best effort to deduce if there is a replay that could correspond to this crash.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/CrashReporting/Integrations/CrashReportingWithRUMIntegration.swift
+++ b/Sources/Datadog/CrashReporting/Integrations/CrashReportingWithRUMIntegration.swift
@@ -186,7 +186,10 @@ internal struct CrashReportingWithRUMIntegration: CrashReportingIntegration {
                 startDate: crashTimings.realCrashDate,
                 sessionUUID: uuidGenerator.generateUnique(), // create new RUM session
                 crashContext: crashContext,
-                hasReplay: nil // as the crash occured after initializing SDK but before starting the first view, we can't know this
+                // As the crash occured after initializing SDK but before starting the first view,
+                // we can't know if Session Replay was configured. However, lack of view implies
+                // that there must be no replay collected:
+                hasReplay: false
             )
         case .handleInBackgroundView:
             newRUMView = createNewRUMViewEvent(
@@ -195,7 +198,10 @@ internal struct CrashReportingWithRUMIntegration: CrashReportingIntegration {
                 startDate: crashTimings.realCrashDate,
                 sessionUUID: uuidGenerator.generateUnique(), // create new RUM session
                 crashContext: crashContext,
-                hasReplay: nil // as the crash occured after initializing SDK but before starting the first view, we can't know this
+                // As the crash occured after initializing SDK but before starting the first view,
+                // we can't know if Session Replay was configured. However, lack of view implies
+                // that there must be no replay collected:
+                hasReplay: false
             )
         case .doNotHandle:
             DD.logger.debug("There was a crash in background, but it is ignored due to Background Event Tracking disabled.")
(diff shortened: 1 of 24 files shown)


Sources/Datadog/CrashReporting/Integrations/CrashReportingWithRUMIntegration.swift
```diff
@@ -183,15 +185,17 @@ internal struct CrashReportingWithRUMIntegration: CrashReportingIntegration {
                 url: RUMOffViewEventsHandlingRule.Constants.applicationLaunchViewURL,
                 startDate: crashTimings.realCrashDate,
                 sessionUUID: uuidGenerator.generateUnique(), // create new RUM session
-                crashContext: crashContext
+                crashContext: crashContext,
+                hasReplay: nil // as the crash occured after initializing SDK but before starting the first view, we can't know this
```
maciejburda, line 189: Relates to our discussion today. This actually sounds like a valid optional flag.

I'm wondering if we can improve this and how harmful for the UX would be defaulting this value to false? 🤔

maciejburda (approved):
LGTM

Sources/Datadog/CrashReporting/Integrations/CrashReportingWithRUMIntegration.swift
ncreated: Indeed good catch 👍. I think we should follow the convention in this PR and send `false` instead (+ annotate it in the comment). No chance we could ever figure out anything from `nil`, so `false`  should work good and will make it unified.

maxep (approved):
Very nice!! Happy to see FeatureBaggage in action, I have plans to improve this container 👍

Sources/Datadog/DatadogInternal/Context/DatadogContext.swift
```diff
@@ -92,5 +92,5 @@ public struct DatadogContext {
     var isLowPowerModeEnabled = false
 
     /// Feature attributes provider.
-    var featuresAttributes: [String: FeatureBaggage] = [:]
+    public internal(set) var featuresAttributes: [String: FeatureBaggage] = [:]
```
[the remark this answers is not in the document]
ncreated: I see, you're right 👍. I will keep it for the sake of consistency (no other context attribute is mutable in `public` API), but I'll be ok with relaxing this once we make everything `public` in `DatadogContext` (it's still in the middle of migration, so in this PR I minimise the impact and only expose what's necessary for SR).

PR #1052, opened 2022-11-15
Add architecture to logs

fuzzybinary:
### What and why?

Add an `architecture` attribute to logs, under a new `_dd` structure for reserved attributes.

### How?

The method for encoding this is a bit different than the rest of the logs, and closer to what Android is doing. Instead of encoding nested properties with `.` keys, it encodes with the full JSON object.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


maxep (approved):
👌

ncreated (commented):
Looks good 👍

PR #1056, opened 2022-11-16
Update RUM events schema

fuzzybinary:
### What and why?

This update contains mostly changes to make properties in Telemetry.Configuration read / write. Because the memberwise initialization of structs is `internal` in Swift, I can't return a copy of the object in the mapper.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


ncreated (approved):
Looks good 👍. Do you think telemetry events could be responsible for the tracking consent tests flakiness that we've been struggling recently? 🤔💭 

I wonder if we should expect it to be stable after
```swift
.filter { event in try event.eventType() != "telemetry" }
```
added in this PR.

PR #1057, opened 2022-11-17
RUMM-2619 Manual otel b3 headers injection

maciejburda:
### What and why?

Adds `OpenTelemetryHTTPHeadersWriter` and `OpenTelemetryHTTPHeadersReader`. That allow manual injection of the Open Telemetry b3 header (both single and multiple).

Currently we only support 64-bit version of the tracing header due to backward compatibility issues. Main issue is related to setup of current Open Tracing headers in RUM which rely on decimal instead of hex and no out of the box solution for 128-bit.

I plan to reiterate on this while implementing automatic injection.

### How?

This implementation follows specification from https://github.com/openzipkin/b3-propagation while navigating around internal implementation details. This PR is initial take on the problem that will be expanded in the automatic injection PR.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


Sources/Datadog/Tracing/Propagation/OpenTelemetry/OpenTelemetryHTTPHeadersReader.swift
```diff
@@ -0,0 +1,61 @@
...
+    func extract() -> OTSpanContext? {
+        guard let baggageItemQueue = baggageItemQueue else {
+            return nil
+        }
+
+        if let traceIDValue = httpHeaderFields[OpenTelemetryHTTPHeaders.Multiple.traceIDField],
+            let spanIDValue = httpHeaderFields[OpenTelemetryHTTPHeaders.Multiple.spanIDField],
+            let traceID = UInt64(traceIDValue).flatMap({ TracingUUID(rawValue: $0) }),
+            let spanID = UInt64(spanIDValue).flatMap({ TracingUUID(rawValue: $0) }) {
+            // TODO Open Telemetry span context
```
maciejburda, line 30: For now we can live with DDSpanContext for the time being. Once we introduce 128 bit it will probably require separate context.

Sources/Datadog/Tracing/Propagation/OpenTracing/TracingHTTPHeaders.swift
```diff
@@ -8,7 +8,7 @@ import Foundation
 
 /// Trace propagation headers as explained in
 /// https://docs.datadoghq.com/real_user_monitoring/connect_rum_and_traces/?tab=browserrum#how-are-rum-resources-linked-to-traces
-internal struct TracingHTTPHeaders {
+internal enum TracingHTTPHeaders {
```
maciejburda, line 11: removes ablity to `.init()` as it acts as a static namespace. Other solution is making the `.init()` private.

Tests/DatadogTests/Datadog/TracerTests.swift
```diff
@@ -716,6 +716,66 @@ class TracerTests: XCTestCase {
...
+        let expectedHTTPHeaders2 = [
+            "X-B3-TraceId": "4",
+            "X-B3-SpanId": "5",
+            "X-B3-Sampled": "1",
+            "X-B3-ParentSpanId": "6"
+        ]
+        XCTAssertEqual(httpHeadersWriter.tracePropagationHTTPHeaders, expectedHTTPHeaders2)
+    }
+
+    func testItInjectsSpanContextWithOpenTelemetryHTTPHeadersWriter_usingSingleHeader() {
```
maxep, line 752: I think we should add a test case where `parentSpanID` is `nil` in both
- `testItInjectsSpanContextWithOpenTelemetryHTTPHeadersWriter_usingSingleHeader`
- `testItInjectsSpanContextWithOpenTelemetryHTTPHeadersWriter_usingMultipleHeaders`

Sources/Datadog/Tracing/Propagation/OpenTelemetry/OpenTelemetryHTTPHeadersWriter.swift
```diff
@@ -0,0 +1,135 @@
...
+            tracePropagationHTTPHeaders = [
+                OpenTelemetryHTTPHeaders.Multiple.sampledField: samplingPriority ? Constants.sampledValue : Constants.unsampledValue
+            ]
+
+            if samplingPriority {
+                tracePropagationHTTPHeaders[OpenTelemetryHTTPHeaders.Multiple.traceIDField] = spanContext.traceID.toHexadecimalString
+                tracePropagationHTTPHeaders[OpenTelemetryHTTPHeaders.Multiple.spanIDField] = spanContext.spanID.toHexadecimalString
+                if let parentSpanID = spanContext.parentSpanID {
+                    tracePropagationHTTPHeaders[OpenTelemetryHTTPHeaders.Multiple.parentSpanIDField] = parentSpanID.toHexadecimalString
+                }
```
maxep, line 106: /nit
```suggestion
                tracePropagationHTTPHeaders[OpenTelemetryHTTPHeaders.Multiple.parentSpanIDField] = spanContext.parentSpanID?.toHexadecimalString
```

Sources/Datadog/Tracing/Propagation/OpenTelemetry/OpenTelemetryHTTPHeadersWriter.swift
```diff
@@ -0,0 +1,135 @@
...
+                    parentSpanIdHexadecimalString = parentSpanID.toHexadecimalString
+                } else {
+                    parentSpanIdHexadecimalString = nil
+                }
+                tracePropagationHTTPHeaders[OpenTelemetryHTTPHeaders.Single.b3Field] = [
+                    spanContext.traceID.toHexadecimalString,
+                    spanContext.spanID.toHexadecimalString,
+                    Constants.sampledValue,
+                    parentSpanIdHexadecimalString
+                ]
```
maxep, line 121: /nit
```suggestion
                tracePropagationHTTPHeaders[OpenTelemetryHTTPHeaders.Single.b3Field] = [
                    spanContext.traceID.toHexadecimalString,
                    spanContext.spanID.toHexadecimalString,
                    Constants.sampledValue,
                    spanContext.parentSpanID?. toHexadecimalString
                ]
```

Sources/Datadog/Tracing/Propagation/OpenTelemetry/OpenTelemetryHTTPHeadersWriter.swift
```diff
@@ -0,0 +1,135 @@
...
+    ///
+    /// - Parameter samplingRate: Tracing sampling rate. 20% by default.
+    /// - Parameter openTelemetryHeaderType: Determines the type of telemetry header type used by the writer.
+    public init(
+        samplingRate: Float = 20,
+        openTelemetryHeaderType: OpenTelemetryHeaderType
+    ) {
+        self.sampler = Sampler(samplingRate: samplingRate)
+        self.openTelemetryHeaderType = openTelemetryHeaderType
+    }
```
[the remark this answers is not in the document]
maxep: I think we should do the same in `HTTPHeadersWriter` actually!

Sources/Datadog/Tracing/Propagation/OpenTelemetry/OpenTelemetryHTTPHeadersReader.swift
```diff
@@ -0,0 +1,61 @@
...
+        }
+        return nil
+    }
+}
+
+private extension Array {
+    subscript (safe index: Index) -> Element? {
+        0 <= index && index < count ? self[index] : nil
+    }
+}
```
maxep, line 55: Nice, we could benefit from this elsewhere by moving it to `SwiftExtensions.swift` 👍

Sources/Datadog/Tracing/Propagation/OpenTelemetry/OpenTelemetryHTTPHeadersWriter.swift
```diff
@@ -0,0 +1,135 @@
...
+                tracePropagationHTTPHeaders[OpenTelemetryHTTPHeaders.Single.b3Field] = Constants.unsampledValue
+            }
+        }
+    }
+
+    private enum Constants {
+        static let sampledValue = "1"
+        static let unsampledValue = "0"
+        static let b3Separator = "-"
+    }
```
maxep, line 134: Not sure this is necessary, especially since it's not leveraged in `OpenTelemetryHTTPHeadersReader`.

Sources/Datadog/Tracing/Propagation/OpenTelemetry/OpenTelemetryHTTPHeadersWriter.swift
```diff
@@ -0,0 +1,135 @@
...
+    /// There are two encodings of B3:
+    /// [Single Header](https://github.com/openzipkin/b3-propagation#single-header)
+    /// and [Multiple Header](https://github.com/openzipkin/b3-propagation#multiple-headers).
+    ///
+    /// Multiple header encoding uses an `X-B3-` prefixed header per item in the trace context.
+    /// Single header delimits the context into into a single entry named b3.
+    /// The single-header variant takes precedence over the multiple header one when extracting fields.
+    public enum OpenTelemetryHeaderType {
+        case multiple, single
+    }
```
maxep, line 40: [@opentelemetry/propagator-b3](https://www.npmjs.com/package/@opentelemetry/propagator-b3) uses `injectEncoding` instead of type. Also, it is already inner to `OpenTelemetryHTTPHeadersWriter`, so no need to add a prefix :)

maxep (commented):
Very nice, the documentation and tests looks great 👌

I wonder if we shouldn't reuse naming from OpenTelemetry, cf: https://www.npmjs.com/package/@opentelemetry/propagator-b3

e.g.: `B3PropagatorWriter`/`B3PropagatorReader`

Sources/Datadog/Tracing/Propagation/OpenTelemetry/OpenTelemetryHTTPHeadersWriter.swift
maciejburda: Ah right. I had different plan but forgot about it. I moved these to header.

Sources/Datadog/Tracing/Propagation/OpenTelemetry/OpenTelemetryHTTPHeadersWriter.swift
maciejburda: You suggest we change the name or move away from enum to some consts as well? 🤔

maxep: No no, just about naming. We could reuse `InjectEncoding`. But it's just a suggestion ;)

maciejburda:
@maxep Thanks for a great review! I addressed all the suggestions. Ready for the second pass 🙌

maciejburda:
Note: I talked to @ncreated and for now we won't support 128-bit encoding. Values of tracing headers were hidden from the consumer, so we decided to keep it this way. Difficult part for automatic switch was hex/decimal encoding.

I incorporated code related to it in this PR - it will ease the work on the automatic Open Telemetry support (from DD configuration).

maxep (approved):
Awesome, well done @maciejburda 👏

Sources/Datadog/Tracer.swift
```diff
@@ -175,11 +175,15 @@ public class Tracer: OTTracer {
...
     public func extract(reader: OTFormatReader) -> OTSpanContext? {
         // TODO: RUMM-385 - make `HTTPHeadersReader` available in public API
-        guard let reader = reader as? HTTPHeadersReader else {
+        if let reader = reader as? HTTPHeadersReader {
+            reader.use(baggageItemQueue: queue)
+            return reader.extract()
+        } else if let reader = reader as? OpenTelemetryHTTPHeadersReader {
+            reader.use(baggageItemQueue: queue)
+            return reader.extract()
+        } else {
```
xgouchet, line 184: At a glance wouldn't having a shared protocol for both `HTTPHeadersReader` and `OpenTelemetryHTTPHeadersReader` let us have a single `if` block instead of duplicating code?

maciejburda: definitely! I already started introducing this in the follow up PR that will enable automatic selection of encoding. I wanted to avoid touching Open Tracing interfaces in this PR.

Tests/DatadogTests/Datadog/Tracing/Propagation/OpenTelemetryHTTPHeadersWriterTests.swift
```diff
@@ -0,0 +1,141 @@
...
+    // swiftlint:enable implicitly_unwrapped_optional
+
+    override func setUp() {
+        super.setUp()
+    }
+
+    override func tearDown() {
+        super.tearDown()
+    }
+
```
ncreated, line 24: Same, as `setUp()` and `tearDown()` are empty, it doesn't seem we need internal state.
```suggestion
```

ncreated (approved):
Excellent work! 💪 Everything seems to be on the right place and having a pair of (public) writer + (internal) reader is a good choice - it should make it much easier for automatic injection (where [both are in use](https://github.com/DataDog/dd-sdk-ios/blob/564d0f9576b50b211bc948612033cd76b8a94db4/Sources/Datadog/URLSessionAutoInstrumentation/Interception/URLSessionInterceptor.swift#L237-L267)).

I few feedbacks - none is blocking, but I would like to hear your opinion.

Sources/Datadog/Tracing/Propagation/OpenTelemetry/OpenTelemetryHTTPHeadersWriter.swift
```diff
@@ -0,0 +1,124 @@
...
+    /// Open Telemetry header encoding.
+    ///
+    /// There are two encodings of B3:
+    /// [Single Header](https://github.com/openzipkin/b3-propagation#single-header)
+    /// and [Multiple Header](https://github.com/openzipkin/b3-propagation#multiple-headers).
+    ///
+    /// Multiple header encoding uses an `X-B3-` prefixed header per item in the trace context.
+    /// Single header delimits the context into into a single entry named b3.
+    /// The single-header variant takes precedence over the multiple header one when extracting fields.
+    @objc
```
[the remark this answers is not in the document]
maciejburda: Exactly! I ended up mapping enum.

I have this change in a follow up PR.

maciejburda:
@ncreated Great feedbacks! I will incorporate all of them in a separate PR if you don't mind. I want to leverage the fact that this one is already approved on CI.

Tests/DatadogTests/Datadog/TracerTests.swift
```diff
@@ -750,6 +832,40 @@ class TracerTests: XCTestCase {
         XCTAssertNil(extractedSpanContext?.dd.parentSpanID)
     }
 
+    func testItExtractsSpanContextWithOpenTelemetryHTTPHeadersReader_forMultipleHeaders() {
+        let tracer: Tracer = .mockAny(in: PassthroughCoreMock())
+        let injectedSpanContext = DDSpanContext(traceID: 1, spanID: 2, parentSpanID: 3, baggageItems: .mockAny())
```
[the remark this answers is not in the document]
maciejburda: I followed existing patterns in these tests, but we can definitely update that.

PR #1063, opened 2022-11-24
Improve internals for cross-platform SDKS

louiszawadzki:
### What and why?

This PR includes 2 improvements for cross-platform SDKs:

- move `updatePerformanceMetric` to internal proxy
- make the `VitalsFrequency` enum extend String for easier initialization

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/DDRUMMonitor+Internal.swift
+++ b/Sources/Datadog/DDRUMMonitor+Internal.swift
@@ -25,4 +25,19 @@ extension DatadogExtension where ExtendedType: DDRUMMonitor {
         let longTaskCommand = RUMAddLongTaskCommand(time: at, attributes: attributes, duration: duration)
         type.process(command: longTaskCommand)
     }
+    
+    public func updatePerformanceMetric(
+        metric: PerformanceMetric,
+        value: Double,
+        attributes: [AttributeKey: AttributeValue] = [:]
+    ) {
+        let performanceMetric = RUMUpdatePerformanceMetric(
+            metric: metric,
+            value: value,
+            time: dateProvider.now,
+            attributes: attributes
+        )
+
+        subscriber?.process(command: performanceMetric)
+    }
 }
--- a/Sources/Datadog/DDRUMMonitor+Internal.swift
+++ b/Sources/Datadog/DDRUMMonitor+Internal.swift
@@ -31,13 +31,17 @@ extension DatadogExtension where ExtendedType: DDRUMMonitor {
         value: Double,
         attributes: [AttributeKey: AttributeValue] = [:]
     ) {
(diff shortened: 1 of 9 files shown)


xgouchet (approved):
LGTM

Sources/Datadog/RUM/_RUMInternalProxy.swift
```diff
@@ -16,14 +16,31 @@ import Foundation
...
-    init(subscriber: RUMCommandSubscriber) {
+    init(subscriber: RUMCommandSubscriber, dateProvider: DateProvider) {
         self.subscriber = subscriber
+        self.dateProvider = dateProvider
     }
 
-    public func addLongTask(at: Date, duration: TimeInterval, attributes: [AttributeKey: AttributeValue] = [:]) {
-        let longTaskCommand = RUMAddLongTaskCommand(time: at, attributes: attributes, duration: duration)
+    public func addLongTask(duration: TimeInterval, attributes: [AttributeKey: AttributeValue] = [:]) {
+        let longTaskCommand = RUMAddLongTaskCommand(time: dateProvider.now, attributes: attributes, duration: duration)
```
ncreated, line 27: Won't this (lack of `Date`) cause a skew in LTs reporting due to JS <> Swift dispatch time? I wonder on the motivation for dropping `Date` parameter 🤔💭.

ncreated (approved):
LG 👍. I left one question, non-blocking if there's a good reason for  dropping `Date` parameter.

Sources/Datadog/RUM/_RUMInternalProxy.swift
louiszawadzki: If we relied on the date from the RN SDK we would not be able to have the time sync with Kronos

maxep (approved):
LGTM 👍

Sources/Datadog/RUM/_RUMInternalProxy.swift
ncreated: I know it's outdated diff, but just to give more context on NTP correction: 
* the correction is not applied automatically for `dateProvider.now`;
* instead, each product shifts dates individually when creating events;
* in case of LT and RUM, it happens in `RUMViewScope`, [here](https://github.com/DataDog/dd-sdk-ios/blob/b1838f7371a9c6ea2a073c8941e61c7f082c8fbe/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift#L573).

By design, RUM ensures correct dates for all events originated in `*Command` signals. Because we used `RUMAddLongTaskCommand` here, that means the date would be corrected properly 👍.

PR #1065, opened 2022-11-24
RUMM-2745 Remove v1 context from Crash Reporting

maxep:
### What and why?

Remove usage of `DatadogV1Context` from Crash Reporting.
This change has a large impact on the `CrashContext` and its encoding for PLCR.

#### CrashReporting
The `CrashContextProvider` is now capable of receiving messaging from the bus to update the `CrashContext`. The `CrashContext` will hold core-context values as well as rum-view-event and rum-session-state.

#### RUM
RUM now sends the view-event and sessions-state as a custom message on the bus.

#### DatadogCore
To properly initialise the Features and Integrations, the core now sends the current context right after registration.
The mocks also reflect this behaviour.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


Sources/Datadog/DatadogInternal/Codable/AnyEncodable.swift
```diff
@@ -0,0 +1,28 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+/// Type-erasure wrapper for any `Encodable` type to be transmitted through
+/// the message bus.
+public struct AnyEncodable: Encodable {
```
ncreated, line 11: Shouldn't this be yet `internal`? If not, then I believe "`AnyEncodable`" name might clash in some apps. How about calling it `DDAnyEncodable` in public API and having internal `typealias` to `AnyEncodable` for simplicity?

Tests/DatadogTests/Datadog/CrashReporting/CrashContext/CrashContextProviderTests.swift
```diff
@@ -14,355 +14,128 @@ import CoreTelephony
...
-        trackingConsentProvider.changeConsent(to: randomTrackingConsent)
+
+        core.send(message: .context(context))
 
         // Then
-        waitForExpectations(timeout: 1, handler: nil)
-        XCTAssertEqual(initialContext.lastTrackingConsent, initialTrackingConsent)
-        XCTAssertEqual(updatedContext?.lastTrackingConsent, randomTrackingConsent)
+        waitForExpectations(timeout: 0.5, handler: nil)
     }
```
ncreated, line 48: The name of this test does no longer reflect what it does. I think it should be rather saying: _**when** SDK context changes, **then** then it notifies new crash context_.

Tests/DatadogTests/Datadog/CrashReporting/CrashContext/CrashContextProviderTests.swift
```diff
@@ -14,355 +14,128 @@ import CoreTelephony
...
 
     // MARK: - `RUMViewEvent` Integration
 
     func testWhenRUMWithCrashContextIntegrationIsUpdatedWithRUMViewEvent_thenCrashContextProviderNotifiesNewContext() {
```
ncreated, line 52: Here, we also have mismatched test name, how about: _**when** new RUM view is sent through message bus, **then** it notifies new crash context_?

Tests/DatadogTests/Datadog/CrashReporting/CrashContext/CrashContextProviderTests.swift
```diff
@@ -14,355 +14,128 @@ import CoreTelephony
...
-        waitForExpectations(timeout: 1, handler: nil)
-        XCTAssertEqual(initialContext.lastRUMSessionState, initialRUMSessionState)
-        XCTAssertEqual(updatedContext?.lastRUMSessionState, randomRUMSessionState)
+        waitForExpectations(timeout: 0.5, handler: nil)
     }
 
-    // MARK: - `UserInfo` Integration
-
-    func testWhenUserInfoValueChangesInUserInfoProvider_thenCrashContextProviderNotifiesNewContext() {
+    func testWhenRUMWithCrashContextIntegrationViewReset_thenCrashContextProviderNotifiesNewContext() {
```
ncreated, line 73: Sounds very odd 🤔💭. How about simply: _**when** RUM view is resent through message bus, **then** it notifies new crash context_?

Tests/DatadogTests/Datadog/CrashReporting/CrashContext/CrashContextProviderTests.swift
```diff
@@ -14,355 +14,128 @@ import CoreTelephony
...
-    #if !os(tvOS)
-
-    private let ctTelephonyNetworkInfoMock = CTTelephonyNetworkInfoMock(
-        serviceCurrentRadioAccessTechnology: ["000001": CTRadioAccessTechnologyLTE],
-        serviceSubscriberCellularProviders: ["000001": CTCarrierMock(carrierName: "Carrier", isoCountryCode: "US", allowsVOIP: true)]
-    )
+    // MARK: - RUM Session State Integration
 
-    func testGivenRunningOniOS11_whenCurrentValueIsObtainedFromCarrierInfoProvider_thenCrashContextProviderNotifiesNewContext() throws {
+    func testWhenRUMWithCrashContextIntegrationIsUpdatedWithRUMSessionState_thenCrashContextProviderNotifiesNewContext() {
```
ncreated, line 99: _"**when** new RUM new session state is sent through message bus, **then** it notifies new crash context_ 😉?

Tests/DatadogTests/Datadog/Mocks/DatadogInternal/DatadogContextMock.swift
```diff
@@ -60,6 +62,34 @@ extension DatadogContext: AnyMockable {
             featuresAttributes: featuresAttributes
         )
     }
+
+    static func mockRandom() -> DatadogContext {
```
ncreated, line 66: 🏅 Can't wait to have this available in Session Replay tests 💪.

Tests/DatadogTests/Datadog/Mocks/DatadogInternal/DatadogContextMock.swift
```diff
@@ -60,6 +62,34 @@ extension DatadogContext: AnyMockable {
...
+            site: .mockRandom(),
+            clientToken: .mockRandom(),
+            service: .mockRandom(),
+            env: .mockRandom(),
+            version: .mockRandom(),
+            variant: .mockRandom(),
+            source: .mockRandom(),
+            sdkVersion: .mockRandom(),
+            ciAppOrigin: .mockRandom(),
+            serverTimeOffset: .zero,
```
ncreated, line 77: `.zero` might be very unfortunate value for fuzzy testing. Could we make it random 🙌? I think there is `.mockRandomInThePast()` defined on `TimeInterval` - IMO better this than `0`.

Tests/DatadogTests/Datadog/Mocks/DatadogInternal/PassthroughCoreMock.swift
```diff
@@ -68,6 +68,8 @@ internal final class PassthroughCoreMock: DatadogV1CoreProtocol, FeatureScope {
         self.expectation = expectation
         self.bypassConsentExpectation = bypassConsentExpectation
         self.messageReceiver = messageReceiver
+
+        messageReceiver.receive(message: .context(context), from: self)
```
ncreated, line 72: It feels a bit that we're re-implementing `DatadogCore` in `PassthroughCoreMock`, no? Step by step, it gains more behaviours similar to real core. I wonder if this trend will continue as we grow V2 adoption - any take on this @maxep ?

TBC: I love the idea of `PassthroughCoreMock` that isolates everything from real storage - just trying to anticipate the future.

ncreated (changes requested):
Excellent job and really strong effort 🏋️‍♂️🏅! Glad to see `CrashReportingFeature` is gone 👌, V1 shrinks and V2 expands.

I left some feedbacks. There is quite a few of them, so I request changes to have a final look later.

Sources/Datadog/CrashReporting/CrashContext/CrashContextProvider.swift
```diff
@@ -9,111 +9,94 @@ import Foundation
...
+        label: "com.datadoghq.crash-context",
+        target: .global(qos: .utility)
+    )
 
-    /// Observes changes to a particular `Value` in the `CrashContext` and manages its updates.
-    private class ContextValueUpdater<Value>: ValueObserver {
-        let queue: DispatchQueue
-        let update: (Value) -> Void
+    /// Unsafe callback instance.
+    var _callback: (CrashContext) -> Void = { _ in }
```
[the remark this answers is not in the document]
maxep: indeed 👍

Sources/Datadog/CrashReporting/CrashContext/CrashContextProvider.swift
```diff
@@ -9,111 +9,94 @@ import Foundation
...
-    }
+        if let event = attributes["view-event", type: RUMViewEvent.self] {
+            queue.async { self.viewEvent = event }
+            return true
+        }
+
+        if let state = attributes["session-state", type: RUMSessionState.self] {
+            queue.async { self.sessionState = state }
+            return true
+        }
```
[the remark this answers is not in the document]
maxep: Yes, I'm definitely not a fan of these strings either. I can make them as constant 👍

Sources/Datadog/CrashReporting/CrashContext/CrashContext.swift
```diff
@@ -11,221 +11,129 @@ import Foundation
...
     }
 
-    // MARK: - Codable
-
-    enum CodingKeys: String, CodingKey {
-        case carrierName = "crn"
-        case carrierISOCountryCode = "cri"
-        case carrierAllowsVOIP = "cra"
-        case radioAccessTechnology = "rdt"
+    static func == (lhs: CrashContext, rhs: CrashContext) -> Bool {
```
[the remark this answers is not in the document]
maxep: Yes :( 
It's a flaw in this solution where we compare the context before storing it in CR. The user info attributes are not `Equatable`, so we could be missing changes in the crash report. I don't see a simple solution for now, but I'm open to suggestion :)
Maybe we can tackle that in a dedicated ticket?

Tests/DatadogTests/Datadog/Mocks/DatadogInternal/PassthroughCoreMock.swift
maxep: I agree! Sending the context through the bus at registration shouldn't be a requirement/convention of the core. I think that this should be enforced by the API.
Another option would be to add a `core.context { _ in }` async method to the core so we can get the context at anytime. But then it kindof questioned the concept of `FeatureScope` where the context is given with the writer.

We could also split `eventWriteContext` with the following API:
```swift
protocol DatadogCoreProtocol {

    func register(feature: DatadogFeature) throws -> Writer

    func register(integration: DatadogFeatureIntegration) throws

    func context(_ block: (DatadogContext) -> Void)
```

I won't be doing that in this PR, but that something we can discuss!

Sources/Datadog/CrashReporting/CrashContext/CrashContextProvider.swift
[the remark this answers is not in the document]
maxep: Note: I have to define constants in both RUM and CR

Sources/Datadog/CrashReporting/CrashContext/CrashContext.swift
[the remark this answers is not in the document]
ncreated: This is fine for now 👍. I don't have any alternative in mind, other than previous logic (updating _crash context_ for each _user info_ change) which won't work in current model of SDK context sharing. Totally ok to visit it later, maybe we'll spot similar use cases on the way.

Sources/Datadog/DatadogInternal/Codable/AnyEncodable.swift
ncreated: Totally fine for having `public AnyEncodable` in `DatadogInternal`, but becase we don't yet have this module, we should be careful on naming public types. Using `DDAnyEncodable` + a `typealias` could work for now and be ultra cheap to replace in V2.

Tests/DatadogTests/Datadog/Mocks/DatadogInternal/PassthroughCoreMock.swift
ncreated: > Another option would be to add a `core.context { _ in }` async method to the core so we can get the context at anytime. But then it kindof questioned the concept of `FeatureScope` where the context is given with the writer.

To me both are not mutually exclusive. There is a difference between "get context" and "write event" behaviours (where the latter is supplied with `context` for write convenience). 

Anyway, I'm just putting a mark on the topic of core mocks - happy to know we're both aware of its evolution 👍. And yes, we can discuss it at some point to see what can be improved.

Sources/Datadog/CrashReporting/CrashContext/CrashContextProvider.swift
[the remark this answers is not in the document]
ncreated: > Note: I have to define constants in both RUM and CR

Which is fine 👍. After all, both modules will need to know these constants definition. Basic integration test (unit) would then assert both are equal. More fancier integration test (still unit) would verify that both modules can talk to each other 👌. All good 👍.

ncreated (approved):
🏅👍

PR #1066, opened 2022-11-24
RUM-2580 Log event async mapper

maxep:
### What and why?

Allow async mapper for log event scrubbing.

#### New Internal Exposure

As part of v2, we have introduced the `DatadogExtension` that allow objects extension behind a single entrypoint. I've moved all internal interface behind an `_internal` property. Here a list of changes:
```swift
Datadog._internal.set(customVersion:)
Datadog._internal.telemetry.setConfigurationMapper(mapper:)
Datadog._internal.telemetry.debug(id:, message:)
Datadog._internal.telemetry.error(id:, message:, kind:, stack:)
RUMMonitor._internal.addLongTask(at:, duration:, attributes:)
```

#### Async Log Event Mapper

Following the same pattern, an internal configuration of an async event mapper is exposed for logs:
```swift
Datadog.initialize(
    ...
        .builderUsing(clientToken: "<client_token>", environment: "<environment_name>")
        ...
        ._internal.setLogEventMapper(<async_mapper>)
        .build()
)
```

The mapper has to comply with the following protocol:
```swift
public protocol LogEventMapper {
    func map(event: LogEvent, callback: @escaping (LogEvent) -> Void)
}
```

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


Sources/Datadog/Logger+Internal.swift
```diff
@@ -0,0 +1,11 @@
+//
+//  Logger+Internal.swift
+//  Datadog
+//
+//  Created by Maxime Epain on 24/11/2022.
+//  Copyright © 2022 Datadog. All rights reserved.
```
plousada, line 6: Should have the usual Datadog copyright notice.

Sources/Datadog/DDRUMMonitor.swift
```diff
@@ -296,7 +296,7 @@ public class DDRUMMonitor {
         attributes: [AttributeKey: AttributeValue] = [:]
     ) {}
 
-    public internal(set) var _internal: _RUMInternalProxy? = nil
+//    public internal(set) var _internal: _RUMInternalProxy? = nil
```
plousada, line 299: Should be deleted?

Sources/Datadog/Datadog+Internal.swift
```diff
@@ -6,26 +6,38 @@
...
     public func setTelemetryConfigurationMapper(mapper: @escaping (TelemetryConfigurationEvent) -> TelemetryConfigurationEvent) {
         if let rumTelemetry = DD.telemetry as? RUMTelemetry {
             rumTelemetry.configurationEventMapper = mapper
         }
     }
 
+    /// Changes the `version` used for [Unified Service Tagging](https://docs.datadoghq.com/getting_started/tagging/unified_service_tagging).
+    public static func set(customVersion: String) {
+        guard let core = defaultDatadogCore as? DatadogCore else {
+            return
```
plousada, line 33: Shouldn't we add a warning or telemetry event in this case?

Sources/Datadog/Datadog+Internal.swift
```diff
@@ -6,26 +6,38 @@
...
 /// or one of the cross platform frameworks.
 ///
 /// Methods, members, and functionality of this class  are subject to change without notice, as they
 /// are not considered part of the public interface of the Datadog SDK.
-public class _InternalProxy {
-    public let _configuration = _ConfigurationProxy()
-    public let _telemetry = _TelemetryProxy()
-}
+extension DatadogExtension where ExtendedType: Datadog {
+    /// Internal telemtry proxy.
```
plousada, line 21:
```suggestion
    /// Internal telemetry proxy.
```

Sources/Datadog/Datadog+Internal.swift
```diff
@@ -37,13 +49,12 @@ public class _TelemetryProxy {
...
-public class _ConfigurationProxy {
-    /// Changes the `version` used for [Unified Service Tagging](https://docs.datadoghq.com/getting_started/tagging/unified_service_tagging).
-    public func set(customVersion: String) {
-        guard let core = defaultDatadogCore as? DatadogCore else {
-            return
-        }
-
-        core.applicationVersionPublisher.version = customVersion
+extension DatadogExtension where ExtendedType: Datadog.Configuration.Builder {
+    /// Sets the custom mapper for `LogEvent`. This can be used to modify logs before they are send to Datadog.
```
plousada, line 53:
```suggestion
    /// Sets the custom mapper for `LogEvent`. This can be used to modify logs before they are sent to Datadog.
```

Sources/Datadog/RUMMonitor.swift
```diff
@@ -213,7 +213,7 @@ public class RUMMonitor: DDRUMMonitor, RUMCommandSubscriber {
 
         super.init()
 
-        _internal = _RUMInternalProxy(subscriber: self)
+//        _internal = _RUMInternalProxy(subscriber: self)
```
plousada, line 216: Delete this?

Sources/Datadog/Logger+Internal.swift
maxep: This file shouldn't been commited.

Sources/Datadog/Datadog+Internal.swift
maxep: This code was just moved. Currently, if `defaultDatadogCore` is not `DatadogCore`, it means that the SDK is not initialised. This interface is made for RN to update the version at runtime, a warning won't be actionable by the user, and a telemetry won't get through since the sdk is not initialized.

plousada (approved):

Sources/Datadog/Logging/Log/LogEventBuilder.swift
```diff
@@ -48,8 +48,9 @@ internal struct LogEventBuilder {
         attributes: LogEvent.Attributes,
         tags: Set<String>,
         context: DatadogContext,
-        threadName: String
-    ) -> LogEvent? {
+        threadName: String,
+        callback: @escaping (LogEvent) -> Void
+    ) {
```
ncreated, line 53: We need to update the method comment too, notably this:
```swift
///
/// - Returns: the `LogEvent` or `nil` if the log was dropped by the user (from event mapper API).
///
```
is now replaced with asynchronous callback. Took me a while to understand that when the event is dropped, the callback is not called - worth putting it in the comment 👍.

Tests/DatadogTests/Datadog/Logging/Log/LogEventBuilderTests.swift
```diff
@@ -169,26 +183,31 @@ class LogEventBuilderTests: XCTestCase {
...
+            XCTAssertEqual(log.message, "modified message")
+            expectation.fulfill()
+        }
 
-        // Then
-        XCTAssertEqual(log.message, "modified message")
+        wait(for: [expectation], timeout: 0)
     }
 
     func testGivenBuilderWithEventMapper_whenEventIsDropped_itReturnsNil() throws {
```
ncreated, line 195:
```suggestion
    func testGivenBuilderWithEventMapper_whenEventIsDropped_thenCallbackIsNotCalled() throws {
```

ncreated (approved):
Looks good 👍. I like the extension idea 👌. I left feedback on updating comment and test names - please have a look.

Mind eventual conflicts with https://github.com/DataDog/dd-sdk-ios/pull/1063/

PR #1075, opened 2022-12-07
Change updatePerformanceMetrics to take a time

fuzzybinary:
### What and why?

Change `updatePerformanceMetrics` to take a time as its first parameter, so it can cast to `RUMCommandSubscriber` instead of `RUMMonitor` (because we eliminate the dependency on dateProvider)

This is mostly to make updatePerformanceMetrics more testable in cross-platform libraries.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


maxep (approved):
👍

PR #1083, opened 2022-12-09
RUMM-2455 [SR] Image placeholder rendering

maciejburda:
### What and why?

Adds strategy for calculating the proper frame of `UIImage` rendered in the `UIImageView`.
It's needed due to possibility of having different size depending on the `contentMode` and `clipsToBounds` properties.

### How?

There's no automatic way of getting the exact position of the UIImage after rendering. To bypass this limitation we wrote extension function that can manually calculate the desired position based on cartesian math. 

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


session-replay/Tests/DatadogSessionReplayTests/Recorder/ViewTreeSnapshotProducer/Utilties/UIImageView+RecordingTests.swift
```diff
@@ -0,0 +1,172 @@
...
+
+class UIImageViewRecordingTests: XCTestCase {
+    var frame: CGRect!
+    var imageView: UIImageView!
+
+    override func setUp() {
+        super.setUp()
+        frame = CGRect(x: 10, y: 10, width: 100, height: 100)
+        imageView = UIImageView(frame: frame)
+        if #available(iOS 13.0, *) {
```
maciejburda, line 19: do we ever run tests against lower targets?

maciejburda:
Useful for testing in playgrounds:

```
func drawRectOnView(rect: CGRect, view: UIView) {
    let path = UIBezierPath(rect: rect)
    let shapeLayer = CAShapeLayer()
    shapeLayer.path = path.cgPath
    shapeLayer.lineWidth = 2
    shapeLayer.strokeColor = UIColor.systemPink.cgColor
    shapeLayer.fillColor = UIColor.clear.cgColor
    view.layer.addSublayer(shapeLayer)
}

func drawView(_ systemName: String) -> UIView {
    let view = UIView(frame: CGRect(x: 10, y: 10, width: 120, height: 120))
    view.backgroundColor = .gray
    let imageView = UIImageView(frame: CGRect(x: 10, y: 10, width: 100, height: 100))
    imageView.contentMode = .center
    imageView.backgroundColor = .green
    imageView.clipsToBounds = true
    let theImage = UIImage(systemName: systemName)
    imageView.image = theImage

    view.addSubview(imageView)
    drawRectOnView(rect: imageView.imageFrame(in: imageView.frame), view: view)
    return view
}
drawView("testtube.2")
```

session-replay/Tests/DatadogSessionReplayTests/Recorder/ViewTreeSnapshotProducer/Utilties/UIImageView+RecordingTests.swift
ncreated: We do, through [bitrise-nightly-unit-tests.yml](https://github.com/DataDog/dd-sdk-ios/blob/develop/tools/nightly-unit-tests/bitrise-nightly-unit-tests.yml) but this is not yet configured for SR, so not a blocker here IMO.

session-replay/Tests/DatadogSessionReplayTests/Recorder/ViewTreeSnapshotProducer/Utilties/UIImageView+RecordingTests.swift
```diff
@@ -0,0 +1,173 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import XCTest
+import UIKit
+@testable import DatadogSessionReplay
+
+/// Make sure to run on iPhone 11
```
ncreated, line 11: This won't be sustainable - we must keep the freedom of changing CI stacks / simulators without requiring updates in tests. Also, tests should pass no matter of the environment we use for the sake of simulator x OS version randomization we do in [bitrise-nightly-unit-tests.yml](https://github.com/DataDog/dd-sdk-ios/blob/develop/tools/nightly-unit-tests/bitrise-nightly-unit-tests.yml) (that will be eventually enabled for SR).
```suggestion
```

session-replay/Tests/DatadogSessionReplayTests/Recorder/ViewTreeSnapshotProducer/Utilties/UIImageView+RecordingTests.swift
```diff
@@ -0,0 +1,173 @@
...
+        imageView = .init()
+        frame = .zero
+        super.tearDown()
+    }
+
+    func testImageFrameForScaleAspectFit() {
+        imageView.contentMode = .scaleAspectFit
+
+        let imageFrame = imageView.imageFrame(in: frame)
+        XCTAssertEqual(imageFrame, CGRect(x: 10.0, y: 13.57142857142857, width: 100.0, height: 92.85714285714286))
```
ncreated, line 35: When running on iPhone 14 Pro simulator (iOS 16.2) I get this:
```
XCTAssertEqual failed: ("(10.0, 13.174603174603178, 100.0, 93.65079365079364)") is not equal to ("(10.0, 13.57142857142857, 100.0, 92.85714285714286)")
```
which is obviously because I didn't follow the `/// Make sure to run on iPhone 11` comment. However, as annotated in another comment, we cannot introduce such strong dependency on certain environment.

On one side, the `13.174603174603178` vs `13.57142857142857` difference could be solved if using `accuracy`:
```swift
XCTAssertEqual(imageFrame.y, 13.57, accurracy: 0.01)
```
but this won't work reliably for `93.65` vs `92.85`.

That makes me think if we should depend on the actual `UIImageView` in tests. I could imagine scaling logic being implemented as type-agnostic, pure math function similar to [AVMakeRect(aspectRatio:insideRect:)](https://developer.apple.com/documentation/avfoundation/1390116-avmakerect):
```swift
func rect(aspectRatio: CGSize, insideRect: CGRect, contentMode: UIView.ContentMode) -> CGRect
```
If such is defined independently (not as an `UIImageView` extension), we will be able to test it on deterministic values, not influenced by any simulator or runtime environment. The `UIImageView.imageFrame(in:)` could then become a convenience to use it for images, but it won't be that critical to assert its precise values in tests.

How does it sound @maciejburda?

ncreated (changes requested):
It works perfectly well 👍💪, but I have strong concerns on sustainability of tests we're adding here (that only work on iPhone 11, iOS 16.x). Left some idea, LTMWDYT 🙌

session-replay/Tests/DatadogSessionReplayTests/Recorder/ViewTreeSnapshotProducer/Utilties/UIImageView+RecordingTests.swift
maciejburda: Interesting thoughts. I was thinking about this problem when I realised it gives different values for different simulators, although - I'm not sure why it's happening. In the test I'm creating a square UIImageView and provide explicit frame. My guess is that it's related to different pixel density of screens.

One thought I had - is that maybe we could leverage snapshot tests for things like this, but this would for sure need a lock on the simulator type and version. These unit tests don't really help much in terms of understanding what really broke, whereas snapshots would show clear visual representation of what has changed.

My initial implementation was relying on [AVMakeRect(aspectRatio:insideRect:)](https://developer.apple.com/documentation/avfoundation/1390116-avmakerect): but as far as I could tell it was producing the same results as the function written. I will explore our options here a little bit more and let you know.

maciejburda:
@ncreated What do you think about dropping the tests and revisit when we have SR testing strategy in place?

ncreated:
> @ncreated What do you think about dropping the tests and revisit when we have SR testing strategy in place?

The testing strategy we have in mind is based on snapshots and if we'd want to use it for `.imageFrame(in:)`, that would require 26 snapshots (13 content modes x two values of `clipsToBounds`). IMO we should only test a few high-level integrations with expensive strategies like snapshots and not use it for unit-like coverage (similar to _pyramid of tests_).

To me, it seems doable to add sustainable unit tests for this problematic if we make the core piece of logic `UIImageView`-agnostic and eventually leverage other types of tests. Referring to the idea of stand-alone function:
```swift
func rect(aspectRatio: CGSize, insideRect: CGRect, contentMode: UIView.ContentMode) -> CGRect
```
it won't use the (problematic) `UIImageView` so we will be able to test all 13 content modes. Later, to ensure it integrates well with `UIImageView` and `clipsToBounds`, we could leverage fuzzy & property-based testing, e.g.:
* Given a fixed-frame `UIImageView` and a random-size `UIImage`,
* when content mode is one of: `top`, `left`, `bottom`, `right` **and** `clipsToBounds == true`,
* then at least one edge of resultant frame is equal to an edge of original frame.

Similar for "aspect fill" and "aspect fit" we could test their "is not smaller than" and "is not bigger than" properties. I think that both sets of tests might come practical to have full coverage and prevent regression. IMO only then (and on top of that) we should think of adding one or two snapshot tests to ensure it works well with wireframes. How does it sound @maciejburda ? I'd be up for pairing on this topic to provide more context 🙂.

ncreated (changes requested):
The idea is good and implementation looks nice 👍. I found few smaller issues and some critical ones (mainly on allowing division by `0`).

session-replay/Sources/DatadogSessionReplay/Recorder/Utilities/CGRect+ContentFrame.swift
```diff
@@ -0,0 +1,152 @@
...
+        }
+        return contentFrame
+    }
+}
+
+fileprivate extension CGSize {
+    func scaleAspectFillRect(for contentSize: CGSize) -> CGRect {
+        guard contentSize.width > 0 && contentSize.height > 0 else {
+            return .zero
+        }
```
ncreated, line 114: nit/ We ensure this in upstream logic (`CGRect.contentFrame(for:using:)`) and this is `fileprivate` routine, so this `guard` will never be effected. Given good tests coverage we could skip this entirely here and in (`scaleAspectFitRect(for:)`)

session-replay/Tests/DatadogSessionReplayTests/Recorder/Utilties/CGRect+ContentFrameTests.swift
```diff
@@ -0,0 +1,199 @@
...
+            .zero,
+            accuracy: accuracy
+        )
+
+        XCTAssertRectsEqual(
+            CGRect.zero.contentFrame(for: CGSize.mockAny(), using: .mockRandom()),
+            .zero,
+            accuracy: accuracy
+        )
+
```
ncreated, line 26: Duplicate:
```suggestion
        XCTAssertRectsEqual(
            CGRect.zero.contentFrame(for: CGSize.mockAny(), using: .mockRandom()),
            .zero,
            accuracy: accuracy
        )

```

ncreated (approved):
Well done 🚀! I left two minor comments, but leaving approval anyway.

session-replay/Sources/DatadogSessionReplay/Recorder/Utilities/CGRect+ContentFrame.swift
maciejburda: Yeah - I thought about that and wanted to be on the safe side when expand this extension in the future and reuse private methods.

PR #1087, opened 2022-12-12
RUMM-2784 [SR] Migrate to `PointerInteractionData` for sending touch information

ncreated:
### What and why?

📦 This PR switches from using deprecated `TouchData` to a new `PointerInteractionData` for sending touches in SR.

The new `PointerInteractionData` includes information of the touch "phase" and uses "one-to-one" relationship with "incremental records" (one record = one touch position). Both make it easier for the player to seek the replay to certain position - the player doesn't need to lookup previous or next records to know if there is a touch information to render. 

Visually, new touches are rendered with a fancy animation because we know the touch phase.

### How?

This change was relatively small in current architecture. It only required recording of the touch phase and changing the events emitted to the batch.

**Before**, the `Processor` was writing one "incremental record" for a batch of touch positions.
**Now**, the `Processor` writes one "incremental record" for each touch position.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/session-replay/Sources/DatadogSessionReplay/Recorder/TouchSnapshotProducer/TouchSnapshot/TouchSnapshot.swift
+++ b/session-replay/Sources/DatadogSessionReplay/Recorder/TouchSnapshotProducer/TouchSnapshot/TouchSnapshot.swift
@@ -14,10 +14,18 @@ internal struct TouchSnapshot {
         /// An unique identifier of the touch. It persists throughout a multi-touch sequence (it is created on "touch down",
         /// continues thru "touch move" and ends in "touch up").
         let id: TouchIdentifier
+        /// Phase of the touch as distinguished in session replay.
+        let phase: TouchPhase
         /// A time of recording this touch
         let date: Date
         /// The position of this touch in application window.
         let position: CGPoint
+    }
+
+    enum TouchPhase {
+        case down
+        case move
+        case up
     }
 
     /// The time of the earliest touch.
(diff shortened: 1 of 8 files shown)


session-replay/Sources/DatadogSessionReplay/Recorder/TouchSnapshotProducer/TouchSnapshot/TouchSnapshot.swift
```diff
@@ -14,12 +14,20 @@ internal struct TouchSnapshot {
...
         let date: Date
         /// The position of this touch in application window.
         let position: CGPoint
     }
 
+    enum TouchPhase {
+        case down
+        case move
+        case up
+    }
```
maxep, line 29: Why not using the `PointerEventType` instead?

maxep (approved):
LGTM 👍  I've left a question!

session-replay/Sources/DatadogSessionReplay/Recorder/TouchSnapshotProducer/TouchSnapshot/TouchSnapshot.swift
ncreated: I prefer to have a clear separation between layers, hence:
* `TouchSnapshot ` and `TouchSnapshot .TouchPhase` - are part of the Recorder,
* `PointerEventType` - is part of the Writer (especially data models we generate).

Using `PointerEventType` in `Recorder` would work and prevent model code duplication, but might be IMO a wrong abstraction. "Touch phase" that we record could be different from what we send at the end (e.g. iOS differentiates more than "up" phase: `.ended`, `.cancelled`, `.regionExited` and if for any reason we need to handle it differently, separate model will be more convenient).

PR #1092, opened 2022-12-16
RUMM-2795 Fix Logs Tags & Attributes Consistency

maxep:
### What and why?

With the asynchronous aspect of the event write context, logs entries' attributes and tags can become inconsistent. Here an example:
```swift
logger.add(tag: "tag1")
logger.info("info message 1") // This log entry won't include tag1 because the EWC is async
logger.remove(tag: "tag1")
```

We need to capture tags and attributes before opening the EWC.

### How?

In this proposal, I replace the remote logger queue with read-write locks around the attributes and tags. Managing the lock is abstracted behind a `@_pthread_rwlock` property-wrapper. The tags and attributes property are captured on the caller thread before jumping on the EWC.

The POSIX lock has been preferred over `DispatchQueue/barrier` as it's slightly faster on read operation.

### Tests

This issue wasn't catch by unit tests because we rely on the `DatadogCoreMock` which was performing EWC in sync. I'm adding the ability to add a `queue` for async Feature scopes.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/DatadogInternal/Concurrency/Concurrency.swift
+++ b/Sources/Datadog/DatadogInternal/Concurrency/Concurrency.swift
@@ -7,19 +7,20 @@
 import Foundation
 
 /// A property wrapper using a fair, POSIX conforming reader-writer lock for atomic
-/// access to the value.
+/// access to the value.  It is optimised for concurrent reads and exclusive writes.
 ///
 /// The wrapper is a class to prevent copying the lock, it creates and initilaizes a `pthread_rwlock_t`.
-/// An additional method `mutate` allow to safely mutate the value within a closure.
+/// An additional method `mutate` allow to safely mutate the value in-place (to read it
+/// and write it while obtaining the lock only once).
 @propertyWrapper
-public final class _pthread_rwlock<Value> {
+/* public */ internal final class _pthread_rwlock<Value> {
     /// The wrapped value.
     private var value: Value
 
     /// The lock object.
     private var rwlock = pthread_rwlock_t()
 
-    public init(wrappedValue value: Value) {
+    /* public */ init(wrappedValue value: Value) {
         pthread_rwlock_init(&rwlock, nil)
         self.value = value
     }
@@ -32,7 +33,7 @@ public final class _pthread_rwlock<Value> {
     ///
     /// The `get` will acquire the lock for reading while the `set` will acquire for
     /// writing.
-    public var wrappedValue: Value {
+    /* public */ var wrappedValue: Value {
         get {
             pthread_rwlock_rdlock(&rwlock)
             defer { pthread_rwlock_unlock(&rwlock) }
@@ -45,10 +46,11 @@ public final class _pthread_rwlock<Value> {
         }
     }
 
-    /// Provides a closure for mutation that will be called synchronously.
+    /// Provides a non-escaping closure for mutation.
+    /// The lock will be acquired once for writing before invoking the closure.
     ///
     /// - Parameter closure: The closure with the mutable value.
-    public func mutate(_ closure: (inout Value) -> Void) {
+    /* public */ func mutate(_ closure: (inout Value) -> Void) {
         pthread_rwlock_wrlock(&rwlock)
         closure(&value)
         pthread_rwlock_unlock(&rwlock)
--- a/Sources/Datadog/DatadogInternal/Concurrency/Concurrency.swift
+++ b/Sources/Datadog/DatadogInternal/Concurrency/Concurrency.swift
@@ -1,0 +1,56 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+/// A property wrapper using a fair, POSIX conforming reader-writer lock for atomic
+/// access to the value.
+///
+/// The wrapper is a class to prevent copying the lock, it creates and initilaizes a `pthread_rwlock_t`.
+/// An additional method `mutate` allow to safely mutate the value within a closure.
+@propertyWrapper
+public final class _pthread_rwlock<Value> {
+    /// The wrapped value.
+    private var value: Value
+
+    /// The lock object.
+    private var rwlock = pthread_rwlock_t()
+
+    public init(wrappedValue value: Value) {
+        pthread_rwlock_init(&rwlock, nil)
+        self.value = value
+    }
+
+    deinit {
+        pthread_rwlock_destroy(&rwlock)
+    }
+
+    /// The wrapped value.
+    ///
+    /// The `get` will acquire the lock for reading while the `set` will acquire for
+    /// writing.
+    public var wrappedValue: Value {
+        get {
+            pthread_rwlock_rdlock(&rwlock)
+            defer { pthread_rwlock_unlock(&rwlock) }
+            return value
+        }
+        set {
+            pthread_rwlock_wrlock(&rwlock)
+            value = newValue
+            pthread_rwlock_unlock(&rwlock)
+        }
+    }
+
+    /// Provides a closure for mutation that will be called synchronously.
+    ///
+    /// - Parameter closure: The closure with the mutable value.
+    public func mutate(_ closure: (inout Value) -> Void) {
+        pthread_rwlock_wrlock(&rwlock)
+        closure(&value)
+        pthread_rwlock_unlock(&rwlock)
+    }
+}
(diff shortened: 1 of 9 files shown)


fuzzybinary (approved):

Sources/Datadog/Logging/RemoteLogger.swift
```diff
@@ -55,41 +53,36 @@ internal final class RemoteLogger: LoggerProtocol {
...
-        )
-
         self.rumContextIntegration = rumContextIntegration
         self.activeSpanIntegration = activeSpanIntegration
     }
 
     // MARK: - Attributes
 
     func addAttribute(forKey key: AttributeKey, value: AttributeValue) {
-        queue.async { self.unsafeAttributes[key] = value }
```
maciejburda, line 70: wouldn't the `queue.sync` solve the issue described in the description?

Sources/Datadog/DatadogInternal/Concurrency/Concurrency.swift
```diff
@@ -0,0 +1,56 @@
...
+///
+/// The wrapper is a class to prevent copying the lock, it creates and initilaizes a `pthread_rwlock_t`.
+/// An additional method `mutate` allow to safely mutate the value within a closure.
+@propertyWrapper
+public final class _pthread_rwlock<Value> {
+    /// The wrapped value.
+    private var value: Value
+
+    /// The lock object.
+    private var rwlock = pthread_rwlock_t()
```
maciejburda, line 20: This read-write lock which allows many reader threads and a single writer thread at a time. Should be used when reading is done more often than writing. Seems like we are dealing with an opposite? 🤔

maxep: Yes exactly, That's the case here: we will have few writes of tags and attributes, while we need available reads because we capture them on the caller thread when the user adds log entries.

Sources/Datadog/Logging/RemoteLogger.swift
maxep: The issue comes from the fact that we capture attributes and tags on the context queue when writing a log entry, which introduce the inconsistency like in this sequence:
```swift
logger.add(tag: "tag1")
logger.info("info message 1") // This log entry won't include tag1 because the EWC is async
logger.remove(tag: "tag1")
```

Using `sync` when writing attributes or tags won't solve the issue. We need to get the tags and attributes before invoking the `. eventWriteContext` and we need to invoke `.eventWriteContext` on the caller thread so the Core Context is also consistent.

Sources/Datadog/DatadogInternal/Concurrency/Concurrency.swift
```diff
@@ -0,0 +1,56 @@
...
+
+import Foundation
+
+/// A property wrapper using a fair, POSIX conforming reader-writer lock for atomic
+/// access to the value.
+///
+/// The wrapper is a class to prevent copying the lock, it creates and initilaizes a `pthread_rwlock_t`.
+/// An additional method `mutate` allow to safely mutate the value within a closure.
+@propertyWrapper
+public final class _pthread_rwlock<Value> {
```
ncreated, line 15: Does it need to be `public` right now? It seems that we can have it `internal` and wait for "DD internal / api" module to be introduced. No reason to pollute users namespace now, no?

Sources/Datadog/DatadogInternal/Concurrency/Concurrency.swift
```diff
@@ -0,0 +1,56 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+/// A property wrapper using a fair, POSIX conforming reader-writer lock for atomic
+/// access to the value.
```
ncreated, line 10: I'm really missing this important context, WDYT?
```suggestion
/// A property wrapper using a fair, POSIX conforming reader-writer lock for atomic
/// access to the value. It is optimised for concurrent reads and exclusive writes.
```

Sources/Datadog/DatadogInternal/Concurrency/Concurrency.swift
```diff
@@ -0,0 +1,56 @@
...
+///
+/// The wrapper is a class to prevent copying the lock, it creates and initilaizes a `pthread_rwlock_t`.
+/// An additional method `mutate` allow to safely mutate the value within a closure.
```
ncreated, line 13: Struggled a bit to understand the purpose of `mutate`. Wouldn't that make it cleaner?
```suggestion
/// An additional method `mutate` allow to safely mutate the value in-place (to read it 
/// and write it while obtaining the lock only once).
```

Sources/Datadog/DatadogInternal/Concurrency/Concurrency.swift
```diff
@@ -0,0 +1,56 @@
...
+            return value
+        }
+        set {
+            pthread_rwlock_wrlock(&rwlock)
+            value = newValue
+            pthread_rwlock_unlock(&rwlock)
+        }
+    }
+
+    /// Provides a closure for mutation that will be called synchronously.
```
ncreated, line 48: Does the word "synchronously" fit this context? We're doing locks here, hence sync/async execution doesn't seem to relate here. What I'm missing however, is the explanation that the lock is applied in-place, hence this method is a perfect fit for the use case of updating the value. WDYT?
```suggestion
    /// Provides a closure for mutation that will obtain the lock only once.
```

Sources/Datadog/Logging/RemoteLogger.swift
```diff
@@ -44,6 +36,12 @@ internal final class RemoteLogger: LoggerProtocol {
     /// Integration with Tracing. It is used to correlate Logs with Spans by injecting `Span` context to `LogEvent`.
     /// Can be `false` if the integration is disabled for this logger.
     internal let activeSpanIntegration: Bool
+    /// Logger-specific attributes.
+    @_pthread_rwlock
+    private var attributes: [String: Encodable] = [:]
```
ncreated, line 41: nit/ Wouldn't it be cleaner if we make up more descriptive name for `@_pthread_rwlock`? Now it reflects the underlying implementation, but for someone who's an occasional user of `pthread` APIs (like I am 😄), this is always confusing. Each time I had to visit docs to ensure myself that this is _"concurrent reads, exclusive writes"_ lock.

Maybe this?
```
@AtomicConcurrentReads
```

WDYT @maxep @maciejburda ?

Tests/DatadogTests/Datadog/Mocks/DatadogInternal/DatadogCoreMock.swift
```diff
@@ -25,25 +25,26 @@ internal final class DatadogCoreMock: Flushable {
...
-    /// ordered/non-recursive lock on the context.
-    private let lock = NSLock()
-    private var _context: DatadogContext {
-        didSet { send(message: .context(_context)) }
-    }
-
-    init(context: DatadogContext = .mockAny()) {
-        _context = context
-    }
+    /// This queue used for invoking Feature scopes.
```
ncreated, line 33:
```suggestion
    /// This queue is used for invoking Feature scopes.
```

ncreated (changes requested):
Everything is clear 👌. Yes, it makes sense to use lock instead of GCD barriers. While I agree the property wrapper should be part of the common module, I don't see a reason to make it `public` before `2.x`.

Also left few other feedbacks. LMKWDYT 🙂

Sources/Datadog/DatadogInternal/Concurrency/Concurrency.swift
maxep: Yes indeed 👍  better to keep internal for now!

Sources/Datadog/Logging/RemoteLogger.swift
maxep: This type of lock is usually referred as [read-write-lock](https://pubs.opengroup.org/onlinepubs/009604599/functions/pthread_rwlock_rdlock.html), we could name this wrapper `ReadWriteLock` instead (that's the name used in java)?

I agree `@_pthread_rwlock` is not clear, but it looks cool 😄

Sources/Datadog/DatadogInternal/Concurrency/Concurrency.swift
maxep: Yes indeed, how about this?
```suggestion
    /// Provides a non-escaping closure for mutation.
    /// The lock will be acquired once for writing before invoking the closure.
```

ncreated: Works good 👍

Sources/Datadog/Logging/RemoteLogger.swift
ncreated: 👍 for `ReadWriteLock` then 🙂

ncreated (approved):
🧵💜🔒 🚀

PR #1093, opened 2022-12-19
RUMM-2782 Flush Context Queue

maxep:
### What and why?

When running E2E tests, we invoke `flushAndDeinitialize` to send all events to the intake. But after introducing the Core Context Provider, some work items could still be waiting on its queue after flushing. We should for wait all execution when flushing.

### How?

Perform a `sync` on the context provider queue.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/Datadog.swift
+++ b/Sources/Datadog/Datadog.swift
@@ -331,6 +331,10 @@ public class Datadog {
     internal static func internalFlushAndDeinitialize() {
         assert(Datadog.isInitialized, "SDK must be first initialized.")
 
+        // Flush the context provider's work items
+        let core = defaultDatadogCore as? DatadogCore
+        core?.contextProvider.queue.sync(flags: .barrier) { }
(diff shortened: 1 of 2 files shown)


maciejburda (approved):

Sources/Datadog/Datadog.swift
```diff
@@ -331,6 +331,10 @@ public class Datadog {
     internal static func internalFlushAndDeinitialize() {
         assert(Datadog.isInitialized, "SDK must be first initialized.")
 
+        // Flush the context provider's work items
+        let core = defaultDatadogCore as? DatadogCore
+        core?.contextProvider.queue.sync(flags: .barrier) { }
```
maxep, line 336: The queue is actually serial, so the `.barrier` will have no effect. It's a safeguard if we ever change it to concurrent.

PR #1095, opened 2022-12-19
RUMM-2681 [SR] Add support for UITextView

maciejburda:
### What and why?

Adds support for UITextView

### How?

It's done by using clipping and intersection of actual view placement and it's content view.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/session-replay/Sources/DatadogSessionReplay/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UITextViewRecorder.swift
+++ b/session-replay/Sources/DatadogSessionReplay/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UITextViewRecorder.swift
@@ -1,7 +1,7 @@
 /*
  * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
  * This product includes software developed at Datadog (https://www.datadoghq.com/).
- * Copyright 2019-2020 Datadog, Inc.
+ * Copyright 2019-Present Datadog, Inc.
  */
 
 import UIKit
--- a/session-replay/Sources/DatadogSessionReplay/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UITextViewRecorder.swift
+++ b/session-replay/Sources/DatadogSessionReplay/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UITextViewRecorder.swift
@@ -21,7 +21,7 @@ internal struct UITextViewRecorder: NodeRecorder {
             textColor: textView.textColor?.cgColor ?? UIColor.black.cgColor,
             font: textView.font,
             textObfuscator: context.recorder.privacy == .maskAll ? context.textObfuscator : nopTextObfuscator,
-            wireframeRect: CGRect(origin: textView.contentOffset, size: textView.contentSize)
+            contentRect: CGRect(origin: textView.contentOffset, size: textView.contentSize)
         )
         return SpecificElement(wireframesBuilder: builder, recordSubtree: true)
     }
@@ -40,13 +40,17 @@ internal struct UITextViewWireframesBuilder: NodeWireframesBuilder {
     /// Text obfuscator for masking text.
     let textObfuscator: TextObfuscating
     /// The frame of the text content
-    var wireframeRect: CGRect
+    let contentRect: CGRect
+
+    var wireframeRect: CGRect {
+        attributes.frame
+    }
 
     private var clip: SRContentClip {
-        let top = abs(wireframeRect.origin.y)
-        let left = abs(wireframeRect.origin.x)
-        let bottom = max(wireframeRect.height - attributes.frame.height - top, 0)
-        let right = max(wireframeRect.width - attributes.frame.width - left, 0)
+        let top = abs(contentRect.origin.y)
+        let left = abs(contentRect.origin.x)
+        let bottom = max(contentRect.height - attributes.frame.height - top, 0)
+        let right = max(contentRect.width - attributes.frame.width - left, 0)
         return SRContentClip(
             bottom: Int64(withNoOverflow: bottom),
             left: Int64(withNoOverflow: left),
@@ -57,10 +61,10 @@ internal struct UITextViewWireframesBuilder: NodeWireframesBuilder {
 
     private var relativeIntersectedRect: CGRect {
         CGRect(
-            x: attributes.frame.origin.x - wireframeRect.origin.x,
-            y: attributes.frame.origin.y - wireframeRect.origin.y,
-            width: max(wireframeRect.width, attributes.frame.width),
-            height: max(wireframeRect.height, attributes.frame.height)
+            x: attributes.frame.origin.x - contentRect.origin.x,
+            y: attributes.frame.origin.y - contentRect.origin.y,
+            width: max(contentRect.width, attributes.frame.width),
+            height: max(contentRect.height, attributes.frame.height)
         )
     }
 
--- a/session-replay/Sources/DatadogSessionReplay/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UITextViewRecorder.swift
+++ b/session-replay/Sources/DatadogSessionReplay/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UITextViewRecorder.swift
@@ -55,7 +55,7 @@ internal struct UITextViewWireframesBuilder: NodeWireframesBuilder {
         )
     }
 
-    private var intersectedRect: CGRect {
+    private var relativeIntersectedRect: CGRect {
         CGRect(
             x: attributes.frame.origin.x - wireframeRect.origin.x,
             y: attributes.frame.origin.y - wireframeRect.origin.y,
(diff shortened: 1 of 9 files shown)


session-replay/Sources/DatadogSessionReplay/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UITextViewRecorder.swift
```diff
@@ -0,0 +1,84 @@
...
+        let builder = UITextViewWireframesBuilder(
+            wireframeID: context.ids.nodeID(for: textView),
+            attributes: attributes,
+            text: textView.text,
+            textColor: textView.textColor?.cgColor ?? UIColor.black.cgColor,
+            font: textView.font,
+            textObfuscator: context.recorder.privacy == .maskAll ? context.textObfuscator : nopTextObfuscator,
+            wireframeRect: CGRect(origin: textView.contentOffset, size: textView.contentSize)
+        )
+        return SpecificElement(wireframesBuilder: builder, recordSubtree: false)
```
ncreated, line 26: 💡 Interestingly, if the `recordSubtree` is set `true`, it is also able of recording text selection within a text view (just tested it). It behaves weird, as we don't have 1:1 matching font size in the player, but "it works" 😉. I have no opinion on leaving `false` and ignoring text selection vs setting `true` and having it done in an imperfect way for now.

session-replay/Sources/DatadogSessionReplay/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UITextViewRecorder.swift
```diff
@@ -0,0 +1,84 @@
...
+    var clip: SRContentClip {
+        let top = abs(wireframeRect.origin.y)
+        let left = abs(wireframeRect.origin.x)
+        let bottom = max(wireframeRect.height - attributes.frame.height - top, 0)
+        let right = max(wireframeRect.width - attributes.frame.width - left, 0)
+        return SRContentClip(
+            bottom: Int64(bottom),
+            left: Int64(left),
+            right: Int64(right),
+            top: Int64(top)
```
ncreated, line 54: nit/ For a sanity, I think it's better to use `Int64(withNoOverflow:)` here.

session-replay/Sources/DatadogSessionReplay/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UITextViewRecorder.swift
```diff
@@ -0,0 +1,84 @@
...
+    /// The color of the text.
+    let textColor: CGColor?
+    /// The font used by the text field.
+    let font: UIFont?
+    /// Text obfuscator for masking text.
+    let textObfuscator: TextObfuscating
+    /// The frame of the text content
+    var wireframeRect: CGRect
+
+    var clip: SRContentClip {
```
ncreated, line 45: nit/ Can be `private`. Especially that `SRContentClip` is part of the payload definition - if thinking of it as a separate layer in our architecture, it's better to not leak it further than `NodeWireframesBuilder`.
```suggestion
    private var clip: SRContentClip {
```

session-replay/Sources/DatadogSessionReplay/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UITextViewRecorder.swift
```diff
@@ -0,0 +1,84 @@
...
+        let right = max(wireframeRect.width - attributes.frame.width - left, 0)
+        return SRContentClip(
+            bottom: Int64(bottom),
+            left: Int64(left),
+            right: Int64(right),
+            top: Int64(top)
+        )
+    }
+
+    var intersectedRect: CGRect {
```
ncreated, line 58: nit/ `private`
```suggestion
    private var intersectedRect: CGRect {
```

ncreated (approved):
🎯 Nice!

session-replay/Sources/DatadogSessionReplay/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UITextViewRecorder.swift
maciejburda: Yeah - good point. I'll add it. It's very hard to fully support when take scaling into account, but usually TextView is used with scrollview enabled, so it should be fine for most cases when we update the font.

session-replay/Sources/DatadogSessionReplay/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UITextViewRecorder.swift
```diff
@@ -0,0 +1,84 @@
...
+        let right = max(wireframeRect.width - attributes.frame.width - left, 0)
+        return SRContentClip(
+            bottom: Int64(withNoOverflow: bottom),
+            left: Int64(withNoOverflow: left),
+            right: Int64(withNoOverflow: right),
+            top: Int64(withNoOverflow: top)
+        )
+    }
+
+    private var intersectedRect: CGRect {
```
maciejburda, line 58: check if `intersect()` can be used.

session-replay/Tests/DatadogSessionReplayTests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UITextViewRecorderTests.swift
```diff
@@ -0,0 +1,73 @@
...
+        let semantics = try XCTUnwrap(recorder.semantics(of: textView, with: viewAttributes, in: .mockAny()))
+        XCTAssertTrue(semantics is InvisibleElement)
+        XCTAssertNil(semantics.wireframesBuilder)
+    }
+
+    func testWhenTextViewHasText() throws {
+        // Given
+        let randomText: String = .mockRandom()
+        textView.textColor = .mockRandom()
+        textView.layoutManager.allowsNonContiguousLayout = true
```
maciejburda, line 32: Without this it fails on enigmatic:
```
testWhenTextViewHasText(): CALayer position contains NaN: [0 nan]. Layer: <CALayer:0x6000030319a0; position = CGPoint (0 15); bounds = CGRect (0 0; 0 30); delegate = <<_UITextContainerView: 0x104f0d9e0; frame = (0 0; 0 30); backgroundColor = UIExtendedGrayColorSpace 0 0; layer = <CALayer: 0x6000030319a0>> minSize = {0, 0}, maxSize = {1.7976931348623157e+308, 1.7976931348623157e+308}, textContainer = <NSTextContainer: 0x600000110640 size = (0.000000,inf); widthTracksTextView = YES; heightTracksTextView = NO>; exclusionPaths = 0x1babfefa8; lineBreakMode = 0>; sublayers = (<CALayer: 0x600003030dc0>); allowsGroupOpacity = YES; backgroundColor = <CGColor 0x600001302580> [<CGColorSpace 0x600001406880> (kCGColorSpaceICCBased; kCGColorSpaceModelMonochrome; Generic Gray Gamma 2.2 Profile; extended range)] ( 0 0 )> (CALayerInvalidGeometry)
```

session-replay/Sources/DatadogSessionReplay/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UITextViewRecorder.swift
```diff
@@ -0,0 +1,84 @@
...
+        let right = max(wireframeRect.width - attributes.frame.width - left, 0)
+        return SRContentClip(
+            bottom: Int64(withNoOverflow: bottom),
+            left: Int64(withNoOverflow: left),
+            right: Int64(withNoOverflow: right),
+            top: Int64(withNoOverflow: top)
+        )
+    }
+
+    private var relativeIntersectedRect: CGRect {
```
maciejburda, line 58: Since here `wireframeRect` is actually: `CGRect(origin: textView.contentOffset, size: textView.contentSize)`, the `CGRect.intersected()` won't work. We need a relative intersection of on the content to parent's parent.

When I look at this it feels simpler to do this way. The other solution would be to pass already relative contentRect and then use `intersected()` here.

session-replay/Tests/DatadogSessionReplayTests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UITextViewRecorderTests.swift
ncreated: How about explaining this in code comment next to it? Otherwise it will be very unclear and will require `git blame` to find the original purpose. Typically we do it somehow this:

```suggestion
        // RUMM-2681 Following is required to avoid "CALayer position contains NaN: [0 nan]. (...) (CALayerInvalidGeometry)" error
        textView.layoutManager.allowsNonContiguousLayout = true
```

PR #1100, opened 2022-12-21
RUMM-2786 Send Crash Report Through Message Bus

maxep:
### What and why?

In the effort of removing type dependencies between Crash Reporting and other Features, the communication of crashes on on the message bus needs a refactoring. The CR integration will no long create Log or RUM events to be sent on the bus, it will directly send Crash Reports for other Features to consume.

### How?

RUM and Logging are now capable of receiving `DDCrashReport` and `CashContext` from the bus to combine them into writable events.
Both `LoggingMessageReceiver` and `RUMMessageReceiver` have been fragmented into multiple receivers to better comply to the single-responsibility principle and facilitate testing. The added mutliplexer `MUXFeatureMessageReceiver` helps combining message receivers.

All related tests are re-structure to follow the new responsibilities of receivers.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


Tests/DatadogTests/Datadog/Logging/LogMessageReceiverTests.swift
```diff
@@ -0,0 +1,153 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import XCTest
+
+@testable import Datadog
+
+class LogMessageReceiverTests: XCTestCase {
```
maxep, line 11: All of this case are coming from the removed `LoggingMessageReceiverTests`.

Sources/Datadog/CrashReporting/Integrations/CrashReportingIntegration.swift
```diff
@@ -4,7 +4,70 @@
...
+    private func printError() {
+        // This case is not reachable in higher abstraction but we add sanity warning.
+        DD.logger.error(
+            """
+            In order to use Crash Reporting, RUM or Logging feature must be enabled.
+            Make sure `.enableRUM(true)` or `.enableLogging(true)` are configured
+            when initializing Datadog SDK.
+            """
+        )
+    }
```
maxep, line 72: This log won't print at initialisation like before, but after processing a crash. Customers will have less chance of seeing this. But with decoupled modules, I don't see other options..

Tests/DatadogTests/Datadog/CrashReporting/CrashReporterTests.swift
```diff
@@ -25,7 +25,7 @@ class CrashReporterTests: XCTestCase {
         let crashReporter = CrashReporter(
```
ncreated, line 25: This is part of:
```swift
testGivenPendingCrashReport_whenLoggingOrRUMIntegrationIsEnabled_itSendsAndPurgesTheCrashReport
```
test and I think it doesn't make sense anymore, because the `integration` is always there.

What still makes sense however, is IMO asserting that crash report is purged after being sent to `integration`, so renaming this test to something like:
```swift
testWhenPendingCrashReportIsFound_itIsSentToIntegrationAndPurged
```

Tests/DatadogTests/Datadog/CrashReporting/CrashReporterTests.swift
```diff
@@ -58,7 +58,7 @@ class CrashReporterTests: XCTestCase {
...
             crashReportingPlugin: plugin,
             crashContextProvider: CrashContextProviderMock(),
-            loggingOrRUMIntegration: integration,
+            integration: integration,
             messageReceiver: NOPFeatureMessageReceiver()
         )
```
ncreated, line 63: ☝️ Similar to above. IMO it would make sense for us to rename the test:
```diff
- testGivenNoPendingCrashReport_whenLoggingOrRUMIntegrationIsEnabled_itDoesNotSendTheCrashReport
+ testWhenPendingCrashReportIsNotFound_itDoesNothing
```

Tests/DatadogTests/Datadog/CrashReporting/CrashReporterTests.swift
```diff
@@ -86,7 +86,7 @@ class CrashReporterTests: XCTestCase {
...
```
ncreated, line 91: ☝️ Similar - the test name is inaccurate:
```diff
- testGivenPendingCrashReportWithUnavailableCrashContext_whenLoggingOrRUMIntegrationIsEnabled_itPurgesTheCrashReportWithNoSending
+ testWhenPendingCrashReportIsFoundButItHasUnavailableCrashContext_itPurgesTheCrashReportWithNoSending
```

Tests/DatadogTests/Datadog/CrashReporting/CrashReporterTests.swift
```diff
@@ -199,23 +231,32 @@ class CrashReporterTests: XCTestCase {
     // MARK: - Usage
 
     func testGivenPendingCrashReport_whenLoggingOrRUMIntegrationCannotBeObtained_itCannotBeInstantiated() {
```
ncreated, line 233:
```suggestion
    func testGivenNoRegisteredCrashReportReceiver_whenPendingCrashReportIsFound_itPrintsError() {
```

Tests/DatadogTests/Datadog/RUM/Integrations/CrashReportReceiverTests.swift
```diff
@@ -26,15 +67,19 @@ class CrashReportingWithRUMIntegrationTests: XCTestCase {
...
             backgroundEventTrackingEnabled: .mockRandom() // no matter BET
         )
 
         // When
-        integration.send(report: crashReport, with: crashContext)
+        XCTAssertTrue(
+            receiver.write(crash: [
+                "report": crashReport,
+                "context": crashContext
+            ], to: core)
```
ncreated, line 81: Like I mentioned earlier, there is no reason to have more than one internal API available on `receiver` (the `.receive(message:from:)`) and keep both `write()` and `send()` private. WDYT?

ncreated (changes requested):
🏅 I can see tremendous efforts here 🏋️‍♂️ and excellent work done! Everything is clear to me on the surface - well designed and it proves the arch 👍.

Going through details I left few remarks here and there and I'd like to discuss it first, so let's loop through "request changes" for better visibility.

Evidently (and this is my main feedback), some of unit tests stop making sense while transitioning to modular monolith. This will be escalating when switching to actual modules, hence I believe we need to keep them accurate with each next step in refactoring. Otherwise we risk losing coverage and it might be hard to recover all product features and edge cases that need to be tested in V2.

Sources/Datadog/Logging/LoggingV2Configuration.swift
```diff
@@ -175,3 +122,122 @@ internal struct LoggingMessageReceiver: FeatureMessageReceiver {
...
+    /// Process messages receives from the bus.
+    ///
+    /// - Parameters:
+    ///   - message: The Feature message
+    ///   - core: The core from which the message is transmitted.
+    func receive(message: FeatureMessage, from core: DatadogCoreProtocol) -> Bool {
+        guard
+            case let .custom(key, attributes) = message, key == LoggingMessageKeys.crash,
+            let report = attributes["report", type: DDCrashReport.self],
+            let context = attributes["context", type: CrashContext.self]
```
ncreated, line 140: This still couples Logs (and RUM in other place) with CR. Something we'll need to solve before modules extraction is possible. Nothing to act on right now, just raising the point here.

Tests/DatadogTests/Datadog/RUM/Integrations/CrashReportReceiverTests.swift
maciejburda: +1

maciejburda (commented):
Awesome work! 🎖️

For me personally, since I'm still onboarding to the codebase - would be great to have a little bit of live walkthrough (for the sake of faster learning experience).

Sources/Datadog/CrashReporting/Integrations/CrashReportingIntegration.swift
maxep: Before, the user was seeing this just after initialisation, so they could see any misconfiguration right away. Now, they will see this after processing a crash, which mean after the app crash and was restarted with Xcode.

Sources/Datadog/CrashReporting/Integrations/CrashReportingIntegration.swift
```diff
@@ -4,7 +4,70 @@
...
+
+/// An integration for sending crash reports to Datadog Core.
+internal struct CrashReportingCoreIntegration: CrashReportingIntegration {
+    /// Defines keys referencing Crash Report message on the bus.
+    internal enum MessageKeys {
+        /// The key for a crash message.
+        static let crash = "crash"
+
+        /// The key for a crash log message.
+        static let crashLog = "crash-log"
```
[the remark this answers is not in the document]
maxep: Yes, I will update the doc. I'm anticipating alignment with Android, where they send the crash to both RUM and Logs.

Sources/Datadog/CrashReporting/Integrations/CrashReportingIntegration.swift
```diff
@@ -4,7 +4,70 @@
...
+    /// Send the crash report et context to integrations.
+    ///
+    /// - Parameters:
+    ///   - report: The crash report.
+    ///   - context: The crash context
     func send(report: DDCrashReport, with context: CrashContext)
 }
+
+/// An integration for sending crash reports to Datadog Core.
+internal struct CrashReportingCoreIntegration: CrashReportingIntegration {
```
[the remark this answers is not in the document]
maxep: hehe yes, the pandora box. I agree, the naming here is not good.
I think it should describe the capability, but given the single function of the protocol it's hard to find something that ends with able, ible, or ing. `*Sending` doesn't really explain what it does, maybe `CrashReportSender` or `CrashReportConsumer`?
`*Type` suffix should be avoided IMO, it's not really idiomatic and doesn't read well when combining multiple conformance..

Sources/Datadog/RUM/Integrations/CrashReportReceiver.swift
```diff
@@ -73,36 +102,37 @@ internal struct CrashReportingWithRUMIntegration: CrashReportingIntegration {
...
             let rumError = createRUMError(from: crashReport, and: lastRUMViewEvent, crashDate: crashTimings.realCrashDate)
-            core.send(
-                message: .custom(
-                    key: "crash",
-                    baggage: ["rum-error": rumError]
-                )
-            )
+            core.v1.scope(for: RUMFeature.self)?.eventWriteContext(bypassConsent: true) { _, writer in
+                writer.write(value: rumError)
+            }
```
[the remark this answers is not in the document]
maxep: Yes exactly! Also this is still v1 scope, it will be replace by v2 scope when we migrate the Feature itself.

Tests/DatadogTests/Datadog/RUM/Integrations/CrashReportReceiverTests.swift
maxep: Indeed, I will change that 👍

Sources/Datadog/DatadogInternal/MessageBus/FeatureMessageReceiver.swift
```diff
@@ -35,3 +35,23 @@ public struct NOPFeatureMessageReceiver: FeatureMessageReceiver {
...
     }
 }
+
+public struct MUXFeatureMessageReceiver: FeatureMessageReceiver {
+    let receivers: [FeatureMessageReceiver]
+
+    /// Creates an instance initialized with the given receivers.
+    public init(_ elements: FeatureMessageReceiver...) {
+        receivers = Array(elements)
+    }
```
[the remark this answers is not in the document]
maxep: It's just for conveniency! It will be initialised with literal array in most cases, variadic parameters make it more straightforward.

Sources/Datadog/CrashReporting/Integrations/CrashReportingIntegration.swift
ncreated: Ah, got it 👍. I think it still makes sense to print this error here (e.g. for someone who is troubleshooting their "crashes not appearing in DD"). Somewhen later, we may think of some other mean of "early spotting misconfigurations" in V2 (depends on where do we land on with public API design).

ncreated (approved):
🎯👌

PR #1108, opened 2022-12-30
WIP: better handle cross-platform crashes

louiszawadzki:
### What and why?

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/CrashReporting/Integrations/CrashReportingWithRUMIntegration.swift
+++ b/Sources/Datadog/CrashReporting/Integrations/CrashReportingWithRUMIntegration.swift
@@ -72,8 +72,10 @@ internal struct CrashReportingWithRUMIntegration: CrashReportingIntegration {
             realDateNow: dateProvider.now.addingTimeInterval(currentTimeCorrection)
         )
 
-        if let lastRUMViewEvent = context.lastRUMViewEvent {
-            sendCrashReportLinkedToLastViewInPreviousSession(report, lastRUMViewEventInPreviousSession: lastRUMViewEvent, using: adjustedCrashTimings)
+        if let lastRUMViewEvent = crashContext.lastRUMViewEvent {
+            if (lastRUMViewEvent.view.crash?.count ?? 0 < 1) { // if a cross-platform crash was reported, do not send its native version
+                sendCrashReportLinkedToLastViewInPreviousSession(crashReport, lastRUMViewEventInPreviousSession: lastRUMViewEvent, using: adjustedCrashTimings)
+            }
         } else if let lastRUMSessionState = context.lastRUMSessionState {
             sendCrashReportToPreviousSession(report, crashContext: context, lastRUMSessionStateInPreviousSession: lastRUMSessionState, using: adjustedCrashTimings)
         } else if sessionSampler.sample() { // before producing a new RUM session, we must consider sampling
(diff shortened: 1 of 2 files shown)


Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -437,7 +439,7 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
                 action: .init(count: actionsCount.toInt64),
                 cpuTicksCount: cpuInfo?.greatestDiff,
                 cpuTicksPerSecond: cpuInfo?.greatestDiff?.divideIfNotZero(by: Double(timeSpent)),
-                crash: .init(count: 0),
+                crash: isCrash ? .init(count: 1) : .init(count: 0),
```
maciejburda, line 442: nit.

It's annoying that Swift doesn't allow conversion from Bool to Int out of the box.

We could do something like this:
```swift
extension ExpressibleByIntegerLiteral {
    init(_ booleanLiteral: BooleanLiteralType) {
        self = booleanLiteral ? 1 : 0
    }
}

Int(isCrash)
```

Sources/Datadog/CrashReporting/Integrations/CrashReportingWithRUMIntegration.swift
```diff
@@ -72,8 +72,10 @@ internal struct CrashReportingWithRUMIntegration: CrashReportingIntegration {
             realDateNow: dateProvider.now.addingTimeInterval(currentTimeCorrection)
         )
 
-        if let lastRUMViewEvent = context.lastRUMViewEvent {
-            sendCrashReportLinkedToLastViewInPreviousSession(report, lastRUMViewEventInPreviousSession: lastRUMViewEvent, using: adjustedCrashTimings)
+        if let lastRUMViewEvent = crashContext.lastRUMViewEvent {
+            if (lastRUMViewEvent.view.crash?.count ?? 0 < 1) { // if a cross-platform crash was reported, do not send its native version
+                sendCrashReportLinkedToLastViewInPreviousSession(crashReport, lastRUMViewEventInPreviousSession: lastRUMViewEvent, using: adjustedCrashTimings)
+            }
```
maciejburda, line 78: to avoid nested if, you can use`,` comma and add a statement directly in the first if let.

```suggestion
        // if a cross-platform crash was reported, do not send its native version
        if let lastRUMViewEvent = crashContext.lastRUMViewEvent, lastRUMViewEvent.view.crash?.count ?? 0 < 1
                sendCrashReportLinkedToLastViewInPreviousSession(crashReport, lastRUMViewEventInPreviousSession: lastRUMViewEvent, using: adjustedCrashTimings)
            }
```

maciejburda (commented):
Thanks for throughout explanation.
I general it feels like a valid improvement for clearing the information in the dashboard.

What puzzles me is what is the real life scenario for using `_dd.error.is_crash` manually 🤔 

Also, what is the drawback of calling stopView from the RN instead?

louiszawadzki:
@maciejburda no functional drawbacks in calling `DdRum.stopView` from the cross-platform SDK.
I believe it makes more sense to let the native SDK handle this however, as this is a "patch" over the view update event sampler that will otherwise drop the update, so I prefer to have the patch as close as possible to the issue.

We can also decide to patch directly the view update sampler to send the view event if it is a crash, if that does not have side effects.

maxep (changes requested):
Hey @louiszawadzki, thank you very much for this proposal.

It's true that a view event cannot have more than one crash (not sure why we have a count here, maybe due to browser requirements). But I wonder if this is the right place to make such assumption. It might be good to try the `viewEventThrottler` suggestion that you made, it might be a bit more complex though.
But before that, it would be best to hold on as I'm currently migrating/removing the `CrashReportingWithRUMIntegration` for v2. I will keep you posted so we can revisit this.

PR #1111, opened 2023-01-05
RUMM-2821 [V2] Simplify "consent" propagation + use `DatadogCoreProxy` in tests

ncreated:
### What and why?

📦 This PR vastly simplifies the consideration of "tracking consent" in V2 data storage. This work will facilitate introduction of `forceNewBatch` API for SR (will be done later in `RUMM-2821`).

Because this PR is big, I made rebase effort to isolate different chores, so it can be reviewed by commits.

### How?

Lack of centralised and immutable SDK context in V1 required a complex abstraction for managing "tracking consent" in thread-safe way in `FeatureStorage`. All this abstraction is now removed in favour of `context.trackingConsent` available in-place thanks to V2 migration we did so far:
```diff
- ConsentProvider.swift
- ConsentAwareDataWriter.swift
- DataOrchestrator.swift
- DataMigrator.swift
- DeleteAllDataMigrator.swift
- MoveDataMigrator.swift
- ArbitraryDataWriter.swift
+ let writer = storage.writer(for: context.trackingConsent)
+ storage.migrateData(fromConsent:toConsent:)
+ storage.clearUnauthorizedData()
```

The threading model behind "event write" sequence is still the same. Meaning:
- `eventWriteContext(_ block:)` executes the `block` closure on "context thread"
- `writer.write(value:)` implementation executes on "read/write thread"

All these removals revealed a problem in our way of mocking SDK core in tests. The previous `DatadogCoreMock` was only implementing the `DatadogCoreProtocol` and trying to "mimic" threading behaviours of the original core. That was leading to random issues when trying to replace [`InMemoryWriter` mock](https://github.com/DataDog/dd-sdk-ios/blob/1a86cdd7d2cd9de4011324e11744100575ad9281/Tests/DatadogTests/Datadog/Mocks/CoreMocks.swift#L557) with other mocks (which I had to do, because we no longer have a single "consent-aware writer" - instead we issue a writer for the current value of tracking consent). Ultimately, I decided to replace `DatadogCoreMock` with `DatadogCoreProxy`, which uses the real SDK core under the hood.

From now on, all unit tests will execute using real `DatadogCore` (wrapped into `DatadogCoreProxy` for events interception). We will benefit from depending on the actual core's threading logic in tests (similar to before, no real I/O is done when using `DatadogCoreProxy`). In fact, changing to this approach revealed two 🐞 bugs that I annotated in their disabled tests: `RUMM-2843` + `RUMM-2844`.

### A note on `DatadogCoreProxy`

The basic usage of new "core proxy" in tests is this:
```swift
let core = DatadogCoreProxy(context: .mockWith(service: "foo-bar"))
defer { core.flushAndTearDown() }
core.register(feature: LoggingFeature.mockAny())

let logger = Logger.builder.build(in: core)
logger.debug("message")

let events = core.waitAndReturnEvents(of: LoggingFeature.self, ofType: LogEvent.self)
XCTAssertEqual(events[0].serviceName, "foo-bar")
```
Unlike the previous approach of `logging.waitAndReturnEvents(count:)`, the new API doesn't require specifying number of events we want to await 🪄🎁. It returns all events that were intercepted in _event write context_ on current thread.

After this change:
- code coverage increased slightly (from `90.6%` to `90.7%`);
- tests execution time decreased from `~45s` to `~35s` on M1 (as we drastically reduce number of tests that do real I/O)

### Review checklist

- [x] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


Sources/Datadog/DatadogCore/Context/DatadogContextProvider.swift
```diff
@@ -88,7 +85,7 @@ internal final class DatadogContextProvider {
     ///
     /// - Parameter receiver: The receiver closure.
     func publish(to receiver: @escaping ContextValueReceiver<DatadogContext>) {
-        queue.async(flags: .barrier) { self.receivers.append(receiver) }
+        queue.async { self.receivers.append(receiver) }
```
ncreated, line 88: I made sure this was a leftover - this queue is not `.concurrent`, so using `.barrier` makes no effect.

Sources/Datadog/DatadogCore/Context/DatadogContextProvider.swift
```diff
@@ -149,8 +146,16 @@ internal final class DatadogContextProvider {
...
         }
     }
+
+#if DD_SDK_COMPILED_FOR_TESTING
+    func replace(context newContext: DatadogContext) {
+        queue.async {
+            self.context = newContext
+        }
+    }
+#endif
```
ncreated, line 160: I'm very much in favour of avoiding this 👍, but it will require even bigger diff in this PR.

Rationale of this hack is **solely** to facilitate the switch from previous `DatadogCoreMock` to new `DatadogCoreProxy` in tests. The way we used `DatadogCoreMock` was this:
```swift
private var core: DatadogCoreMock!

func testSomething() {
   core.context = .mockWith(
      // ... configure the mock context
   )

   // use the core in test
}
```
so I had to add analogous behaviour of changing `.context` lazily on `DatadogCoreProxy`:
```swift
// DatadogCoreProxy

var context: DatadogContext {
   didSet {
      let core = self.proxy as! DatadogCore
      core.contextProvider.replace(context: context)
   }
}
```

IMO this is a flaw in our testing convention: the `.context` should be managed only by the core, without allowing anyone else (including `DatadogCoreProxy`) to modify it arbitrarily. Migrating our tests to this following model will fulfil this idea:
```swift
func testSomething() {
   let core = DatadogCoreProxy(
      context = .mockWith(
         // ... configure the mock context
      )
   )
   defer { core.flushAndTearDown() }

   // use the core in test
}
```
The price will be to refactor all places (roughly ~100 tests). If we agree on this model, I'm OK with doing that in separate PR to isolate changes. WDYT @maxep @maciejburda ?

Sources/Datadog/DatadogCore/Storage/FilesOrchestrator.swift
```diff
@@ -123,14 +131,6 @@ internal class FilesOrchestrator {
...
     }
 
-    func deleteAllReadableFiles() {
-        do {
-            try directory.deleteAllFiles()
-        } catch {
-            DD.telemetry.error("Failed to delete all readable file", error: error)
-        }
-    }
-
```
ncreated, line 133: Rationale: this is no longer responsibility of `FilesOrchestrator` - it was moved to `FeatureStorage` instead.

Sources/Datadog/RUMMonitor.swift
```diff
@@ -132,8 +132,6 @@ public class RUMMonitor: DDRUMMonitor, RUMCommandSubscriber {
     internal let core: DatadogCoreProtocol
     /// The root scope of RUM monitoring.
     internal let applicationScope: RUMApplicationScope
-    /// Current RUM context provider for integrations with Logging and Tracing.
-    internal let contextProvider: RUMCurrentContext
```
ncreated, line 136: This wasn't needed anymore (not related to this PR, just deleting).

Tests/DatadogTests/Datadog/TracerTests.swift
```diff
@@ -655,33 +665,34 @@ class TracerTests: XCTestCase {
...
-        let spanMatcher = try tracing.waitAndReturnSpanMatchers(count: 1)[0]
-        XCTAssertEqual(
-            try spanMatcher.meta.custom(keyPath: "meta.\(RUMContextAttributes.applicationID)"),
-            rum.configuration.applicationID
-        )
-        XCTAssertValidRumUUID(try spanMatcher.meta.custom(keyPath: "meta.\(RUMContextAttributes.sessionID)"))
-        XCTAssertValidRumUUID(try spanMatcher.meta.custom(keyPath: "meta.\(RUMContextAttributes.viewID)"))
-    }
+    // TODO: RUMM-2843 [V2 regression] RUM context is not associated with span started on caller thread
+//    func testGivenBundlingWithRUMEnabledAndRUMMonitorRegistered_whenSendingSpan_itContainsCurrentRUMContext() throws {
```
ncreated, line 669: 🐞 This test revealed a hidden bug - I did JIRA for it.

Tests/DatadogTests/Datadog/RUMMonitorTests.swift
```diff
@@ -781,38 +783,39 @@ class RUMMonitorTests: XCTestCase {
...
-
-        XCTAssertNil(try? lastViewUpdate.attribute(forKeyPath: "a1") as String)
-        XCTAssertNil(try? lastViewUpdate.attribute(forKeyPath: "a2") as String)
-        XCTAssertNil(try? lastViewUpdate.attribute(forKeyPath: "a3") as String)
-        try XCTAssertEqual(lastViewUpdate.attribute(forKeyPath: "context.a1"), "bar1", "The value should be updated")
-        try XCTAssertEqual(lastViewUpdate.attribute(forKeyPath: "context.a2"), "foo2", "The attribute should not be removed")
-        try XCTAssertEqual(lastViewUpdate.attribute(forKeyPath: "context.a3"), "foo3", "The attribute should be added")
-    }
+    // TODO: RUMM-2844 [V2 regression?] RUM monitor `removeAttribute(forKey:)` behaves differently than in V1
+//    func testWhenViewIsStarted_attributesCanBeAddedOrUpdatedButNotRemoved() throws {
```
ncreated, line 787: 🐞❓ This revealed a bug that needs to be investigated in `RUMM-2844` (I'm not fully sure if this is a bug or feature).

Sources/Datadog/DatadogCore/Context/DatadogContextProvider.swift
maciejburda: I like the idea. It's better to do this now than never.

Sources/Datadog/DatadogCore/DatadogCore.swift
```diff
@@ -182,6 +186,24 @@ internal final class DatadogCore {
             self.messageBusQueue.async { messageReceiver.receive(message: .context(context), from: self) }
         }
     }
+
+    /// A list of storage units of currently registered Features.
+    private var allStorages: [FeatureStorage] {
+        let v1Storages = v1Features.values.compactMap { $0 as? V1Feature }.map { $0.storage }
+        let v2Storages = v2Features.values.map { $0.storage }
+        return v1Storages + v2Storages
```
maciejburda, line 194: I'm curious if we could end up in a state where we could have same feature in both v1 and v2 (middle of migration maybe?).

If yes - maybe it makes sense to perform union instead of adding the two? 🤔

Sources/Datadog/DatadogCore/DatadogCore.swift
```diff
@@ -283,17 +308,54 @@ extension DatadogCore: DatadogCoreProtocol {
...
+                $0.receive(message: message, from: sender)
             }
 
             if receivers.isEmpty {
                 fallback()
             }
         }
     }
+
+    /* public */ func flush() {
```
maciejburda, line 323: leftover comment

Sources/Datadog/DatadogCore/DatadogCore.swift
```diff
@@ -283,17 +308,54 @@ extension DatadogCore: DatadogCoreProtocol {
...
+        // The order of flushing below must be considered cautiously and
+        // follow our design choices around SDK core's threading.
+
+        // First, flush bus queue - because messages can lead to obtaining "event write context" (reading
+        // context & performing write) in other Features:
+        messageBusQueue.flush(numberOfTimes: 5) // 5 is arbitrary limit (we don't expect longer chains)
```
maciejburda, line 329: maybe this 5 could be a variable with a comment and then used in all three?

Sources/Datadog/DatadogCore/Utils/FoundationExtensions.swift
```diff
@@ -0,0 +1,28 @@
...
+            if countDown > 0 {
+                self.async { recursiveFlush(countDown: countDown - 1) }
+            } else {
+                semaphore.signal()
+            }
+        }
+
+        recursiveFlush(countDown: numberOfTimes)
+        _ = semaphore.wait(timeout: .distantFuture)
+    }
```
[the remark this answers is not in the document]
maciejburda: Maybe it could be moved to tests in that case?

maciejburda (approved):
LGTM 🙌

Few minor comments

Sources/Datadog/DatadogInternal/DatadogCoreProtocol.swift
```diff
@@ -106,10 +106,17 @@ public protocol DatadogCoreProtocol: AnyObject {
...
     ///               processed by any Features on the bus.
-    func send(message: FeatureMessage, else fallback: @escaping () -> Void)
+    func send(message: FeatureMessage, sender: DatadogCoreProtocol, else fallback: @escaping () -> Void)
+
+    /// Flushes all asynchronous operations in the SDK by **blocking the caller thread** till their completion.
+    func flush()
+
+    /// Awaits completion of all asynchronous operations and deinitializes this instance of the SDK.
+    /// It **blocks the caller thread**.
+    func flushAndTearDown()
```
maxep, line 116: I don't think this should be part of the protocol, it's a specific behaviour that is not a requirement of a core. It can be exposed by specific implementation depending on its inner logic, that includes blocking the caller thread.

Sources/Datadog/DatadogCore/Storage/FeatureStorage.swift
```diff
@@ -7,26 +7,101 @@
...
+        queue.async {
+            do {
+                switch (fromConsent, toConsent) {
+                case (.pending, .notGranted):
+                    try directories.unauthorized.deleteAllFiles()
+                case (.pending, .granted):
+                    try directories.unauthorized.moveAllFiles(to: directories.authorized)
+                default:
+                    break
+                }
```
maxep, line 60: With the implementation of `func writer(for trackingConsent: TrackingConsent) -> Writer`, we don't need to know the previous consent. We can only rely on the new one to perform file operations.

Sources/Datadog/DatadogCore/Utils/FoundationExtensions.swift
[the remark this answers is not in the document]
maxep: This looks odd to me TBH, especially if it's not yet necessary. Maybe it worth adding this when we need it, no?
It's might be that our threading model is at fault, we could have more straightforward data flow, by combining queue for instance, rather than having arbitrary recursive flush. WDYT?

maxep (changes requested):
The core proxy in test is very smart, and already pays off and highlighting issues, well done!
I've some remarks/question, let's discuss those first.

Sources/Datadog/DatadogCore/Utils/FoundationExtensions.swift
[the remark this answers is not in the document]
ncreated: 👍 I removed it for now - let's see when and if we will need it.

Sources/Datadog/DatadogCore/DatadogCore.swift
ncreated: > I'm curious if we could end up in a state where we could have same feature in both v1 and v2 (middle of migration maybe?).

Very unlikely - we have only 3 features registered in V1 and quite clear migration path for each. I made this code verbose to facilitate the ultimate change we will do after migrating the last one:
```diff
-        let v1Storages = v1Features.values.compactMap { $0 as? V1Feature }.map { $0.storage }
-        let v2Storages = v2Features.values.map { $0.storage }
-        return v1Storages + v2Storages
+ return features.values.map { $0.storage }
```

Sources/Datadog/DatadogCore/DatadogCore.swift
ncreated: This was removed so it's not relevant anymore. But the `/* public */ func` was not a leftover - it is a convention we have behind annotating APIs that are `internal` today, but we will make them `public` in `2.0.0` release.

Sources/Datadog/DatadogInternal/DatadogCoreProtocol.swift
ncreated: Removing it from the protocol for now 👍, but we will need it in similar form or another because of our [integration with CI Visibility](https://docs.datadoghq.com/continuous_integration/tests/swift/?tab=swiftpackagemanager#rum-integration). Basically, if app tests are instrumented with both `dd-sdk-ios` and `dd-swift-testing`, the CI Visibility SDK requests "force upload all data" in our SDK:
https://github.com/DataDog/dd-sdk-ios/blob/fcf5daa3155d3e29617a193d0fba3987b6569e01/Sources/Datadog/FeaturesIntegration/CITestIntegration.swift#L61

In V2, this logic will spawn from RUM Feature (CI Visibility sets values on RUM events), which only knows `DatadogCoreProtocol`, so we will need something similar. Indeed, it won't need to be "sync" (we can make it "async" and for the purpose of CI visibility we just stall it).

OK with no action right now, but leaving this comment for awareness.

Sources/Datadog/DatadogCore/Storage/FeatureStorage.swift
ncreated: 👍 done ✅

Sources/Datadog/DatadogInternal/DatadogCoreProtocol.swift
maxep: Yes good to know 👍  Indeed, we would need to find a way to expose flush logic in test, but I think we can make it test only.

Sources/Datadog/DatadogCore/Context/DatadogContextProvider.swift
ncreated: I made an attempt to resolve this and use new model in tests, but gave up as it requires slightly more refactoring in tests, that I don't want to conduct right now.

I got blocked on this tests (there are 2 like this in each feature: Logging, Tracing and RUM):
https://github.com/DataDog/dd-sdk-ios/blob/fcf5daa3155d3e29617a193d0fba3987b6569e01/Tests/DatadogTests/Datadog/LoggerTests.swift#L325-L331

What happens here is:
* simulate _carrier info_ / _network info_ change
* send Log | Span | RUM event
* assert the event has proper _carrier info_ / _network info_ set

Speaking V1: it made sense to test this behaviour in each Feature separately.
Speaking V2: this certain behaviour should be tested in `DatadogCoreTests`, only once.

Because we don't yet have `DatadogCoreTests` defined, I'm leaving this for now. We will revisit this later in migration.

maxep (approved):
We will greatly benefit from the core proxy, awesome work!

PR #1112, opened 2023-01-05
RUMM-2786 Create AnyEncoder and AnyDecoder

maxep:
### What and why?

Introduce implementation of `AnyEncoder` and `AnyDecoder` which will allow defragmenting data-types for transfer on the message-bus.

In the following example, a CrashContext is transmitted from the CrashReporting Feature to RUM through the message-bus: 
image

This implementation will also be useful for the Flutter SDK to replace the dependency to [DictionaryCoder](https://github.com/almazrafi/DictionaryCoder). cc @fuzzybinary 

Usage on the `FeatureBaggage` is done in #1113.

#### Encoding
The `AnyEncoder` implements the [`Encoder` protocol](https://developer.apple.com/documentation/swift/encoder) and is capable of encoding an `Encodable` object to a `Any`, `[Any?]`, or `[String: Any?]` representation.
It will result in the same `object` as the following snippet, but without having to serialize to `Data`.

```swift
let encoder = JSONEncoder()
let data = try encoder.encode(value)
let object = try JSONSerialization.jsonObject(with: data, options: .allowFragments)
```

#### Decoding
The `AnyDecoder` implements the [`Decoder` protocol](https://developer.apple.com/documentation/swift/decoder) and is capable of decoding a`Decodable` object from an `Any?` representation.
It will result in the same `value` as the following snippet, but without having to serialize to `Data`.
```swift
let decoder = JSONDecoder()
let data = try JSONSerialization.data(withJSONObject: object, options: [])
let value =  try decoder.decode(type, from: data)
```

### Review checklist

- [x] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/DatadogInternal/Codable/AnyEncoder.swift
+++ b/Sources/Datadog/DatadogInternal/Codable/AnyEncoder.swift
@@ -1,0 +1,673 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+/// An object that encodes instances of an `Encodable` type as `Any?`.
+///
+/// The example below shows how to encode an instance of a simple `GroceryProduct`
+/// type to `Any` object. The type adopts `Codable` so that it's encodable as `Any`
+/// using a `AnyEncoder` instance.
+///
+///     struct GroceryProduct: Codable {
+///         var name: String
+///         var points: Int
+///         var description: String?
+///     }
+///
+///     let pear = GroceryProduct(name: "Pear", points: 250, description: "A ripe pear.")
+///
+///     let encoder = AnyEncoder()
+///
+///     let object = try encoder.encode(pear)
+///     print(object as! NSDictionary)
+///
+///     /* Prints:
+///     {
+///         description = "A ripe pear.";
+///         name = Pear;
+///         points = 250;
+///     }
+///     */
+open class AnyEncoder {
+    /// Initializes `self`.
+    public init() { }
+
+    /// Encodes the given top-level value and returns its Any representation.
+    ///
+    /// Depending on the value and its `Encodable` implementation the returned
+    /// encoded value can be `Any`, `[Any?]`, `[String: Any?]`, or `nil`.
+    ///
+    /// - parameter value: The value to encode.
+    /// - returns: An `Any` object containing the value.
+    /// - throws: An error if any value throws an error during encoding.
+    open func encode<T>(_ value: T) throws -> Any? where T: Encodable {
+        let encoder = _AnyEncoder()
+        try value.encode(to: encoder)
+        return encoder.any
+    }
+}
+
+/// A type that can encode values into a native format for external
+/// representation.
+private class _AnyEncoder: Encoder {
+    typealias AnyEncodingStorage = (Any?) -> Void
+
+    /// The path of coding keys taken to get to this point in encoding.
+    let codingPath: [CodingKey]
+
+    /// Any contextual information set by the user for encoding.
+    let userInfo: [CodingUserInfoKey: Any] = [:]
+
+    /// The encoded value.
+    var any: Any?
+
+    init(path: [CodingKey] = []) {
+        codingPath = path
+    }
+
+    /// Returns an encoding container appropriate for holding multiple values
+    /// keyed by the given key type.
+    ///
+    /// You must use only one kind of top-level encoding container. This method
+    /// must not be called after a call to `unkeyedContainer()` or after
+    /// encoding a value through a call to `singleValueContainer()`
+    ///
+    /// - parameter type: The key type to use for the container.
+    /// - returns: A new keyed encoding container.
+    func container<Key>(keyedBy type: Key.Type) -> KeyedEncodingContainer<Key> where Key: CodingKey {
+        let container = KeyedContainer<Key>(
+            store: { self.any = $0 },
+            dictionary: any as? [String: Any?],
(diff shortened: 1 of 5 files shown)


Sources/Datadog/DatadogInternal/Codable/AnyEncoder.swift
```diff
@@ -0,0 +1,651 @@
...
+            }
+        }
+    }
+}
+
+/// A shared encodable object will skip encoding when using the `AnyEncoder`.
+///
+/// Making a Encodable as shared allow to bypass encoding when the type is
+/// known by multiple parties.
+internal protocol SharedEncodable: Encodable { }
```
maciejburda, line 646: I think we could consider `(Shared)DictionaryEncodable`. I think it would be puzzling to developers to custom encodable, until they realised it's shortening the path and allows encoding directly to a Dictionary.

maciejburda (commented):
I was curious about the performance of this solution and if measured correctly it appears that using JSONEncoder is still faster. Take a look at this example:

```swift
func measurePerformance() throws {
      let encoder = AnyEncoder()
      let jsonEncoder = JSONEncoder()

      struct Obj: Codable {
          let type: String = .mockAny()
          let id: String = .mockAny()
          let isFavorited: Bool = .mockAny()
          let attributes: Attributes = .init(title: .mockAny(), body: .mockAny())

          struct Attributes: Codable {
              let title: String
              let body: String
          }
      }
      let object = Obj()
      let iterations = 100_000
      let start = CFAbsoluteTimeGetCurrent()
      for _ in 1...iterations { // 0.1216 sec
          let dict = try encoder.encode(object)
      }
      let diff = CFAbsoluteTimeGetCurrent() - start
      print("#1 Took \(diff) seconds")
      // #1 Took 2.4502270221710205 seconds


      let start2 = CFAbsoluteTimeGetCurrent()
      for _ in 1...iterations {
          let dict2 = try JSONSerialization.jsonObject(with: jsonEncoder.encode(object), options: .allowFragments)
      }
      let diff2 = CFAbsoluteTimeGetCurrent() - start2
      print("#2 Took \(diff2) seconds")
      // #2 Took 1.6677980422973633 seconds
  }
```

maxep:
Thanks for doing a quick benchmark, @maciejburda, it helped me find some optimizations. I'm not able to reach the combination of `JSONEncoder` and `JSONSerialization` (surprisingly), but I noticed that the decoding part is actually way more performant, see the snippet below. Since we need encoding+decoding on the bus, I consider this solution more performant overall. Additionally, the `DictionaryEncodable` protocol will help skip any serialization which is not possible with `JSONEncoder`.

```swift
let encoder = AnyEncoder()
let decoder = AnyDecoder()
let jsonEncoder = JSONEncoder()
let jsonDecoder = JSONDecoder()

struct Obj: Codable {
    let type: String = .mockAny()
    let id: String = .mockAny()
    let isFavorited: Bool = .mockAny()
    let attributes: Attributes = .init(title: .mockAny(), body: .mockAny())

    struct Attributes: Codable {
        let title: String
        let body: String
    }
}

let object = Obj()
let iterations = 100_000
let start = CFAbsoluteTimeGetCurrent()
for _ in 1...iterations {
    let dict = try encoder.encode(object)
    _ = try decoder.decode(Obj.self, from: dict)
}

let diff = CFAbsoluteTimeGetCurrent() - start
print("#1 Took \(diff) seconds")
// #1 Took 2.508360981941223 seconds

let start2 = CFAbsoluteTimeGetCurrent()
for _ in 1...iterations {
    let dict2 = try JSONSerialization.jsonObject(with: jsonEncoder.encode(object), options: .allowFragments)
    _ = try jsonDecoder.decode(Obj.self, from: JSONSerialization.data(withJSONObject: dict2, options: .fragmentsAllowed))
}
let diff2 = CFAbsoluteTimeGetCurrent() - start2
print("#2 Took \(diff2) seconds")
// #2 Took 3.418305993080139 seconds
```

urseberry (approved):

maxep:
I did further benchmarks:

### Device: iPhone 11 (A13 Bionic chip), iOS 16.2

**Model to serialize:**
```swift
struct Model: Codable {
    var type: String = "type"
    var id: String = "id"
    var isFavorited: Bool = .random()
    var attributes: Attributes = .init()

    struct Attributes: Codable {
        var title: String = "title"
        var body: String = "body"
    }
}
```

**Iterations:** 100 000

#### Execution Time (in seconds)

||AnyCoder|JSONSerialization|
|---|---|---
|encoding|0.9294730424880981|0.9025610685348511|
|decoding|1.0498170852661133|1.7138639688491821|

#### CPU Profile (Weight)

||AnyCoder|JSONSerialization|
|---|---|---
|encoding|19.6%|19.4%|
|decoding|22.8%|36.5%|

#### Memory Allocation (Bytes Used)

||AnyCoder|JSONSerialization|
|---|---|---
|encoding|1.62 KB 0.0%|3.47 KB 0.1%|
|decoding|1.34 KB 0.0%|32.25 KB 1.0%|

---

Using such serialisation is obviously not free, but it performs better than the `JSONCoder`+`JSONSerialization`. I think this solution will drastically simplify our codebase by keeping a good separation of concerns between modules and keeping our flat dependency tree. Additionally, the `DictionaryEncodable` protocol is there to skip any serialisation of shared data types which will have a positive impact on perfs.

WDYT @ncreated @maciejburda ?

maciejburda (approved):
Looks great! Thanks for the extra effort on optimising 💪

ncreated (approved):
I really like the idea 🚀⭐, and I believe it will be brilliant for passing events through message bus. The performance should not be a problem - thanks for taking care of that @maciejburda @maxep  👌🏎️ .

I left minor feedback on missing tests coverage and one proposal to IMO more convenient approach in testing.

PR #1116, opened 2023-01-06
RUMM-2582 Async RUM Event Mappers

fuzzybinary:
### What and why?

This introduces async event mappers into RUM, similar to those made for the Logging API. 

This allows platforms that require async calls (such as Flutter) to support writing mappers in their native languages.

### How?

Each RUM event has a separate protocol for mapping (RUMXEventMapper) which is public, only so it can be accessed / overridden by our Cross Platform clients.  The Sync versions (SyncRUMXEventMapper) are internal, and are created by the existing public methods for creating RUM event mappers.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/Datadog+Internal.swift
+++ b/Sources/Datadog/Datadog+Internal.swift
@@ -56,4 +56,44 @@ extension DatadogExtension where ExtendedType: Datadog.Configuration.Builder {
         type.configuration.logEventMapper = mapper
         return type
     }
+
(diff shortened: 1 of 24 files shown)


maciejburda (changes requested):
Great effort 💪 

Left some suggestions/ideas and comments around potential retain cycles.

PR #1118, opened 2023-01-10
RUMM-2690 Test utilities package

maciejburda:
### What and why?

### How?

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


test-utilities/Package.swift
```diff
@@ -0,0 +1,24 @@
+// swift-tools-version: 5.7
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "TestUtilities",
```
fuzzybinary, line 7: I'm concerned about this name being a bit to general... does it make sense to maybe call it DatadogTestUtilities?

fuzzybinary:
It looks like as part of this, files were moved from `Datadog/Core` to `DatadogCore/Core`. Is there a reason we can't have them at the root in `DatadogCore?` to avoid the extra directory?

test-utilities/Package.swift
maciejburda: We have this unwritten agreement to prepend targets with `datadog` only if they are user facing. This one is purely internal. Interestingly enough - in this PR it's still quite generic, but next one will be exposing some Datadog specific mocks (so they're usable in other vertical features).

maciejburda:
@fuzzybinary These files were just moved to a proper root folder (as a part of preparation for v2). It's kind of unrelated, but at the same time it doesn't hurt. I'll try organising them better to ditch the `/Core` subfolder.

test-utilities/README.md
```diff
@@ -0,0 +1,3 @@
+# TestUtilities
+
+Package that provides foundational tools, extensions and mocks needed for testing other targets.
```
alai97, line 3:
```suggestion
Package that provides foundational tools, extensions, and mocks needed for testing other targets.
```

alai97 (approved):
Looks good!

maciejburda:
Closing if favor of https://github.com/DataDog/dd-sdk-ios/pull/1126

PR #1119, opened 2023-01-10
RUMM-2821 [V2 / SR] Add `forceNewBatch` API

ncreated:
### What and why?

📦 This PR adds `forceNewBatch: Bool` flag to _event write context_ API.

This option will be used for Session Replay (in #1121) to transport all SR requests through V2's SDK core. This will bring reliability and performance of SR uploads - the core makes sure that request are always delivered (unlike SR's own uploader that fails if device is offline).

```swift
///   - forceNewBatch: `true` to enforce that event will be written to a separate batch than previous events.
///                     Default is `false`, which means the core uses its own heuristic to split events between
///                     batches. This parameter can be leveraged in Features which require a clear separation
///                     of group of events for preparing their upload (a single upload is always constructed from a single batch).
///   - block: The block to execute.
func eventWriteContext(bypassConsent: Bool, forceNewBatch: Bool, _ block: @escaping (DatadogContext, Writer) throws -> Void)
```

### How?

The option is added to EVC and used by `FileWriter` when obtaining a file from `FilesOrchestrator` - it gets a new one if `forceNewBatch` is `true`; otherwise tries to reuse the previous one.

🎁 Because this is another piece of logic we're adding to `DatadogCore` implementation and because this component was still suffering from lack of unit tests, I created `DatadogCoreTests` file with adding few very first tests cases to assert its key behaviours (tracking consent & request building in V2 features).

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


Datadog/Datadog.xcodeproj/project.pbxproj
```diff
@@ -4457,6 +4463,8 @@
 		D2EFA873286E010100F1FAA6 /* DatadogCore */ = {
 			isa = PBXGroup;
 			children = (
+				614B78EA296D7B63009C6B92 /* DatadogCoreTests.swift */,
+				614B78EC296D7B63009C6B92 /* LowPowerModePublisherTests.swift */,
```
ncreated, line 4467: We had this file in tests, but not included 😮. I'm adding it now 👍.

Tests/DatadogTests/Datadog/DatadogCore/BatteryStatusReaderTests.swift
```diff
@@ -1,71 +0,0 @@
-/*
- * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
- * This product includes software developed at Datadog (https://www.datadoghq.com/).
- * Copyright 2019-Present Datadog, Inc.
- */
-
-import XCTest
-@testable import Datadog
-
-class BatteryStatusReaderTests: XCTestCase {
```
ncreated, line 10: Deleting this file - it wasn't used anywhere (even not referenced in Xcode project).

Tests/DatadogTests/Datadog/DatadogCore/DatadogCoreTests.swift
```diff
@@ -0,0 +1,190 @@
...
+        temporaryCoreDirectory.create()
+    }
+
+    override func tearDown() {
+        temporaryCoreDirectory.delete()
+        super.tearDown()
+    }
+
+    func testWhenWritingEventsWithDifferentTrackingConsent_itOnlyUploadsAuthorizedEvents() throws {
+        let server = ServerMock(delivery: .success(response: .mockResponseWith(statusCode: 200)))
```
ncreated, line 32: I took the approach of testing `DatadogCore` "end-to-end". We're using the [existing `ServerMock`](https://github.com/DataDog/dd-sdk-ios/blob/919ebef125b4103033400dc84bf53de5db0dd349/Tests/DatadogTests/Datadog/Mocks/ServerMock.swift#L19) here - it intercepts `URLSession` (with `URLProtocol`). We collect events in mocked V2 Feature and make an assertion on events passed by core for building upload request.

Because it is touching no internals of the `DatadogCore`, such tests should be well sustainable and require no maintenance during regular code changes.

maciejburda (approved):
LGTM

Sources/Datadog/DatadogInternal/DatadogCoreProtocol.swift
```diff
@@ -170,8 +170,12 @@ public protocol FeatureScope {
     ///   - bypassConsent: `true` to bypass the current core consent and write events as authorized.
     ///                    Default is `false`, setting `true` must still respect user's consent for
     ///                    collecting information.
+    ///   - forceNewBatch: `true` to enforce that event will be written to a separate batch than previous events.
+    ///                     Default is `false`, which means the core uses its own heuristic to split events between
+    ///                     batches. This parameter can be leveraged in Features which require a clear separation
+    ///                     of group of events for preparing their upload (a single upload is always constructed from a single batch).
     ///   - block: The block to execute.
-    func eventWriteContext(bypassConsent: Bool, _ block: @escaping (DatadogContext, Writer) throws -> Void)
+    func eventWriteContext(bypassConsent: Bool, forceNewBatch: Bool, _ block: @escaping (DatadogContext, Writer) throws -> Void)
```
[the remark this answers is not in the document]
ncreated: Good idea 👍. Turns out that methods in protocol extensions can define default parameter values, so I was able to avoid introducing new methods. Done ✅.

maxep (approved):
🚀

PR #1121, opened 2023-01-10
RUMM-2821 [SR] Send all SR requests through V2 core

ncreated:
### What and why?

📦 This PR mades use of `forceNewBatch` API added previously (#1119) when writing records to SR storage.

This makes Session Replay benefit from all transport facilities available in SDK core, i.a.:
* retrying requests failed due to bad connection,
* pausing uploads when device is offline,
* limiting number of parallel requests,
* respecting number of active threads in the app.

### How?

The "arbitrary" uploader previously added in SR is now deleted.

The logic behind building SR requests is vastly simplified, because the `SR.RequestBuilder` can now assume that all events correspond to a single segment.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


maciejburda (approved):
👌

PR #1123, opened 2023-01-11
RUMM-2821 [V2] Fix SDK core instance leaks, retain cycles and TSAN issues

ncreated:
### What and why?

📦 While trying to stabilise CI tests in #1111 I found more impact of switching them to run on the actual `DatadogCore` instance instead of basic mock. That revealed real problems in current state of V2 migration.

This PR fixes several problems of leaking `DatadogCore` memory. If not addressed, these problems will pop-up in V2 once multiple stand-alone instances of `core` are allowed for customer use.

These issues do not impact V1 nor recent release, as we only provide single, shared instance of `DatadogCore` through singleton model. Even though current V1 uses vast amount of V2 code, there is no chance through V1 API that any customer will step into these problems.

#### Added `core` reference counting

I added basic `referenceCount` property to available `DatadogCoreProtocol` mocks:
* `PassthroughCoreMock`
* `DatadogCoreProxy`

The `referenceCount` gets incremented in `init()` and decremented from `deinit {}`.

Then, I updated [DatadogTestsObserver](https://github.com/DataDog/dd-sdk-ios/blob/develop/Tests/DatadogTests/TestsObserver/DatadogTestsObserver.swift) to ensure that `referenceCount` is `0` upon completing each test. That revealed several memory leaks and retain cycles, fixed one by one.

#### Did more intensive `flush()`

Another thing was insufficient `flush()` implementation proposed previously in #1111 - while it works in most cases, when repeating some tests 100-1000 times, it still suffers from not awaiting enough operations. I made `flush()` more greedy for now, until we find a better model (that will be necessary for having public "deinitialize" API).

#### Solved 2 TSAN data races

Last, I had to solve following 2 data races in Tracing <> RUM integration, as it was responsible for occasional TSAN crashes:

Screenshot 2023-01-11 at 18 35 55

---

I've ran this branch 8 times on CI (also with ITAR disabled). It's green even if Bitrise Checks report it wrong (😞).

Screenshot 2023-01-12 at 12 18 02

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/Tracing/TracingV2Configuration.swift
+++ b/Sources/Datadog/Tracing/TracingV2Configuration.swift
@@ -51,6 +51,9 @@ internal struct TracingRequestBuilder: FeatureRequestBuilder {
 }
 
 internal struct TracingMessageReceiver: FeatureMessageReceiver {
+    /// Tracks RUM context to be associated with spans.
+    let rum = TracingWithRUMIntegration()
+
     /// Process messages receives from the bus.
     ///
     /// - Parameters:
@@ -69,14 +72,7 @@ internal struct TracingMessageReceiver: FeatureMessageReceiver {
     ///
     /// - Parameter context: The updated core context.
     private func update(context: DatadogContext) -> Bool {
-        guard
-            let tracer = Global.sharedTracer as? Tracer,
-            let integration = tracer.rumIntegration
-        else {
-            return false
-        }
-
-        integration.attribues = context.featuresAttributes["rum"]?.all()
+        rum.attribues = context.featuresAttributes["rum"]?.all()
         return true
     }
 }
--- a/Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
+++ b/Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
@@ -29,16 +29,18 @@ open class DDURLSessionDelegate: NSObject, URLSessionTaskDelegate, URLSessionDat
     }
 
     var instrumentation: URLSessionAutoInstrumentation? {
-        core().v1.feature(URLSessionAutoInstrumentation.self)
+        core?.v1.feature(URLSessionAutoInstrumentation.self)
     }
 
     let firstPartyHosts: FirstPartyHosts
 
-    private let core: () -> DatadogCoreProtocol
+    /// The instance of the SDK core notified by this delegate.
+    /// It must be a weak reference, because `URLSessionDelegate` can last longer than this instance.
+    private weak var core: DatadogCoreProtocol?
 
     @objc
     override public init() {
-        core = { defaultDatadogCore }
+        core = defaultDatadogCore
         firstPartyHosts = .init()
         super.init()
     }
@@ -77,7 +79,7 @@ open class DDURLSessionDelegate: NSObject, URLSessionTaskDelegate, URLSessionDat
     ///   - additionalFirstPartyHosts: additionalFirstPartyHosts: these hosts are tracked **in addition to** what was
     ///                                passed to `DatadogConfiguration.Builder` via `trackURLSession(firstPartyHosts:)`
     public init(
-        in core: @autoclosure @escaping () -> DatadogCoreProtocol,
+        in core: DatadogCoreProtocol,
         additionalFirstPartyHostsWithHeaderTypes: [String: Set<TracingHeaderType>]
     ) {
         self.core = core
--- a/Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
+++ b/Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
@@ -29,24 +29,22 @@ open class DDURLSessionDelegate: NSObject, URLSessionTaskDelegate, URLSessionDat
     }
 
     var instrumentation: URLSessionAutoInstrumentation? {
-        core?.v1.feature(URLSessionAutoInstrumentation.self)
+        let core = self.core ?? defaultDatadogCore
+        return core.v1.feature(URLSessionAutoInstrumentation.self)
     }
 
     let firstPartyHosts: FirstPartyHosts
 
     /// The instance of the SDK core notified by this delegate.
-    /// It must be a weak reference, because `URLSessionDelegate` can last longer than this instance.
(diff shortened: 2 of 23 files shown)


Sources/Datadog/Tracing/TracingV2Configuration.swift
```diff
@@ -69,14 +72,7 @@ internal struct TracingMessageReceiver: FeatureMessageReceiver {
...
     private func update(context: DatadogContext) -> Bool {
-        guard
-            let tracer = Global.sharedTracer as? Tracer,
-            let integration = tracer.rumIntegration
-        else {
-            return false
-        }
-
-        integration.attribues = context.featuresAttributes["rum"]?.all()
+        rum.attribues = context.featuresAttributes["rum"]?.all()
```
ncreated, line 75: This was causing TSAN data race crash, because following was possible:
* tracing feature registered in core
* → core dispatching new context
* at the moment of context arrival to this `TracingMessageReceiver`, the `tracer.rumIntegration` was in the middle of it's `alloc()`, causing data race on `tracer.rumIntegration` access.

The fix is to reverse the communication. Because `TracingMessageReceiver` is ready upon Tracing feature registration, the `Tracer` can access it safely.

Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
```diff
@@ -29,16 +29,18 @@ open class DDURLSessionDelegate: NSObject, URLSessionTaskDelegate, URLSessionDat
...
-        core().v1.feature(URLSessionAutoInstrumentation.self)
+        core?.v1.feature(URLSessionAutoInstrumentation.self)
     }
 
     let firstPartyHosts: FirstPartyHosts
 
-    private let core: () -> DatadogCoreProtocol
+    /// The instance of the SDK core notified by this delegate.
+    /// It must be a weak reference, because `URLSessionDelegate` can last longer than this instance.
+    private weak var core: DatadogCoreProtocol?
```
maxep, line 39: Creating could happen before initialising the SDK, that's why we were using a closure here, to lazily get the core.

maxep (commented):
Thank you so much for digging into this 🙏 It looks all good, I just left one comment on the session delegate which might stop working if it's initialized before core.

Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
ncreated: Good catch @maxep - I made a change (pls 🙏 review as it introduces some new concepts). Also added unit tests for all initializers, so we don't miss it again.

Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
```diff
@@ -67,22 +91,27 @@ open class DDURLSessionDelegate: NSObject, URLSessionTaskDelegate, URLSessionDat
...
-    ///   - core: Datadog SDK core.
-    ///   - additionalFirstPartyHosts: additionalFirstPartyHosts: these hosts are tracked **in addition to** what was
+    ///   - core: Datadog SDK instance (or `nil` to use default SDK instance).
+    ///   - additionalFirstPartyHosts: these hosts are tracked **in addition to** what was
     ///                                passed to `DatadogConfiguration.Builder` via `trackURLSession(firstPartyHosts:)`
     public init(
-        in core: DatadogCoreProtocol,
-        additionalFirstPartyHostsWithHeaderTypes: [String: Set<TracingHeaderType>]
+        in core: DatadogCoreProtocol? = nil,
+        additionalFirstPartyHostsWithHeaderTypes: [String: Set<TracingHeaderType>] = [:]
```
ncreated, line 112: This change in public API is backward compatible, even if someone is using `defaultDatadogCore` (for whatever reason - as we consider it "publicly internal"):
```swift
// still works after this change:
DDURLSessionDelegate(in: defaultDatadogCore, additionalFirstPartyHostsWithHeaderTypes: [:])
```

Sources/Datadog/URLSessionAutoInstrumentation/DDURLSessionDelegate.swift
```diff
@@ -29,23 +29,44 @@ open class DDURLSessionDelegate: NSObject, URLSessionTaskDelegate, URLSessionDat
...
+
+        var core: DatadogCoreProtocol? {
+            switch self {
+            case .useDefaultCore:
+                return defaultDatadogCore
+            case .useExplicitCore(let reference):
+                return reference.core
+            }
+        }
+    }
```
maxep, line 60: It make sense to either keep a weak ref to the provided core, if any, or lazy load the default one 👍 

I think we could simplify this with the following:
```swift
    var instrumentation: URLSessionAutoInstrumentation? {
        let core = core ?? defaultDatadogCore // lazy load default if no core.
        return core.v1.feature(URLSessionAutoInstrumentation.self)
    }

    let firstPartyHosts: FirstPartyHosts

    private weak var core: DatadogCoreProtocol?

    @objc
    override public init() {
        core = nil
        firstPartyHosts = .init()
        super.init()
    }

    public convenience init(in core: DatadogCoreProtocol? = nil) {
        self.init(in: core, additionalFirstPartyHostsWithHeaderTypes: [:])
    }

    @objc
    public convenience init(additionalFirstPartyHosts: Set<String>) {
        self.init(additionalFirstPartyHostsWithHeaderTypes: additionalFirstPartyHosts.reduce(into: [:], { partialResult, host in
            partialResult[host] = [.datadog]
        }))
    }

    public init(
        in core: DatadogCoreProtocol? = nil,
        additionalFirstPartyHostsWithHeaderTypes: [String: Set<TracingHeaderType>]
    ) {
        self.core = core
        self.firstPartyHosts = FirstPartyHosts(additionalFirstPartyHostsWithHeaderTypes)
        super.init()
    }
```

ncreated: 🤦 Definitely 💯! 100x Much easier - thanks 😅

maxep (approved):
Very nice that you catch that now, we would make sure that we don't keep strong refs to core anymore. Thanks again!

PR #1128, opened 2023-01-13
RUMM-2516 Remove React Native duplicate crashes

louiszawadzki:
### What and why?

Better handle cross-platform crashes to make sure:
- the view events are sent after the error
- the native crash, that is a "duplicate" from the js crash, is not sent as we limit the amount of crashes per view to 1.

Result 🎉:
image

### How?

The event is sent by modifying the `viewEventThrottler` to always accept events when the crash count is 1 - this is my favorite implementation as we keep close to the source of the issue, but I don't know if there will also be side effects if potential events come after that - there should not be so many in any case.

We don't send the duplicate by checking the last view does not contain a crash already.
A nice benefit to this implementation is that if js crash reporting is not enabled but native crash reporting is, we'll get the native crash 👍

### Potential Regression

If a user adds a crash error (can only be done by adding a `_dd.error.is_crash` attribute to the error), but then the app lives on and a distinct native crash happens, the native error won't be reported. Also all events happening in the view after the error won't be recorded.

This could happen if a cross-platform user mistakenly labels an error as a crash, but the application does not actually crash.
On react-native, we could issue a warning to let the user know that now that the app has crashed, it won't be recording any event until a new view is started to prevent this.

It could also be the case if the app uses https://github.com/a7ul/react-native-exception-handler, which will display a screen before restarting the app (useful to let users add more details for instance). 

### Review checklist
- [x] Add CHANGELOG entry for user facing changes -> I think no changelog is needed as it only is a concern for React Native

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/Utils/RUMViewUpdatesThrottler.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/Utils/RUMViewUpdatesThrottler.swift
@@ -43,7 +43,7 @@ internal final class RUMViewUpdatesThrottler: RUMViewUpdatesThrottlerType {
 
         sample = sample || event.view.isActive == false // always accept the last event in a view
 
-        sample = sample || event.view.crash?.count ?? 0 > 0 // always accept events if the view has crashed
+        sample = sample || event.view.crash.map { $0.count > 0 } ?? false
 
         if sample {
             lastSampledTimeSpentInNs = event.view.timeSpent
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/Utils/RUMViewUpdatesThrottler.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/Utils/RUMViewUpdatesThrottler.swift
@@ -43,6 +43,8 @@ internal final class RUMViewUpdatesThrottler: RUMViewUpdatesThrottlerType {
 
         sample = sample || event.view.isActive == false // always accept the last event in a view
 
+        sample = sample || event.view.crash?.count ?? 0 > 0 // always accept events if the view has crashed
+
         if sample {
             lastSampledTimeSpentInNs = event.view.timeSpent
         }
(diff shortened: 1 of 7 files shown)


Tests/DatadogTests/Datadog/RUM/Integrations/CrashReportReceiverTests.swift
```diff
@@ -90,13 +90,48 @@ class CrashReportReceiverTests: XCTestCase {
         XCTAssertEqual(core.events(ofType: RUMViewEvent.self).count, 1)
     }
 
+    func testGivenCrashDuringRUMSessionWithActiveViewCollectedLessThan4HoursAgoWithPreviousCrasg_whenSending_itSendsNothing() throws {
```
fuzzybinary, line 93: nit: 
```suggestion
    func testGivenCrashDuringRUMSessionWithActiveViewCollectedLessThan4HoursAgoWithPreviousCrash_whenSending_itSendsNothing() throws {
```

fuzzybinary (approved):

Sources/Datadog/RUM/RUMMonitor/Scopes/Utils/RUMViewUpdatesThrottler.swift
```diff
@@ -43,6 +43,8 @@ internal final class RUMViewUpdatesThrottler: RUMViewUpdatesThrottlerType {
 
         sample = sample || event.view.isActive == false // always accept the last event in a view
 
+        sample = sample || event.view.crash?.count ?? 0 > 0 // always accept events if the view has crashed
```
maxep, line 46: /nit for readability
```suggestion
        sample = sample || event.view.crash.map { $0.count > 0 } ?? false
```

maxep (approved):
Looks great, thanks, @louiszawadzki 🙏 
I left 2 minor suggestions.

PR #1113, opened 2023-01-05
RUMM-2786 Use AnyEncoder and AnyDecoder on FeatureBaggage

maxep:
### What and why?

In the effort of decoupling Feature modules, this PR removes shared types and let the core defragment data-types on the message-bus. This flow adds a new constraint on the `FeatureBaggage`: any baggage must be `Codable`.

### How?

After the introduction of [AnyEncoder and AnyDecoder](https://github.com/DataDog/dd-sdk-ios/pull/1112), we can now remove shared data-types and have model definitions in modules. The data-flow on the message-bus will look like the following:
image

All data-types that are defined in `DatadogInternal` are shared and marked as `DictionaryEncodable` to skip encoding. The `AnyCodable` is also used to transmit unknown types as-is, this is the case when transmitting the last RUM View from RUM to CrashReporting, or from sending a Crash stack from CR to RUM or Logs.

### Review checklist
- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes

### Custom CI job configuration (optional)
- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

PR #1131, opened 2023-01-17
RUMM-2688 Add NTP sync to session replay records

maciejburda:
### What and why?

Shares the server time offset of RUM view with session replay to apply NTP offset on timestamps of records.

### How?

It's passed the same way rum view id was passed and in the `EnrichedRecord` value is subtracted from the timestamp.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/DatadogSessionReplay/Sources/DatadogSessionReplay/DatadogCoreIntegration/RUMContextReceiver.swift
+++ b/DatadogSessionReplay/Sources/DatadogSessionReplay/DatadogCoreIntegration/RUMContextReceiver.swift
@@ -85,10 +85,10 @@ private extension DatadogContext {
 
 private extension FeatureBaggage {
     var rumContext: RUMContext? {
-        guard let applicationID: String = self[RUMDependency.applicationIDKey],
-              let sessionID: String = self[RUMDependency.sessionIDKey],
-              let viewID: String = self[RUMDependency.viewIDKey],
-              let serverTimeOffset: TimeInterval = self[RUMDependency.serverTimeOffset]
+        guard let applicationID: String = self[RUMDependency.ids, type: [String: String].self]?[RUMDependency.IDs.applicationIDKey],
+              let sessionID: String = self[RUMDependency.ids, type: [String: String].self]?[RUMDependency.IDs.sessionIDKey],
+              let viewID: String = self[RUMDependency.ids, type: [String: String].self]?[RUMDependency.IDs.viewIDKey],
+              let serverTimeOffset: TimeInterval = self[RUMDependency.serverTimeOffsetKey]
         else {
             // Current RUM session is not sampled
             return nil
--- a/DatadogSessionReplay/Sources/DatadogSessionReplay/DatadogCoreIntegration/RUMContextReceiver.swift
+++ b/DatadogSessionReplay/Sources/DatadogSessionReplay/DatadogCoreIntegration/RUMContextReceiver.swift
@@ -8,15 +8,31 @@ import Foundation
 import Datadog
 
 /// The RUM context received from `DatadogCore`.
-internal struct RUMContext: Equatable {
-    /// Current RUM application ID - standard UUID string, lowecased.
-    let applicationID: String
-    /// Current RUM session ID - standard UUID string, lowecased.
-    let sessionID: String
-    /// Current RUM view ID - standard UUID string, lowecased.
-    let viewID: String
-    /// Current RUM view server time offset (in seconds).
-    let serverTimeOffset: TimeInterval
+internal struct RUMContext: Decodable, Equatable {
+    internal struct IDs: Decodable, Equatable {
+        enum CodingKeys: String, CodingKey {
+            case applicationID = "application_id"
+            case sessionID = "session_id"
+            case viewID = "view.id"
+        }
+        /// Current RUM application ID - standard UUID string, lowecased.
+        let applicationID: String
+        /// Current RUM session ID - standard UUID string, lowecased.
+        let sessionID: String
+        /// Current RUM view ID - standard UUID string, lowecased.
+        let viewID: String
+    }
+
+    enum CodingKeys: String, CodingKey {
+        case ids = "ids"
+        case viewServerTimeOffset = "server_time_offset"
+    }
+
+    /// Wrapper for all RUM related IDs
+    let ids: IDs
+
+    /// Current view related server time offset
+    let viewServerTimeOffset: TimeInterval?
 }
 
 /// An observer notifying on`RUMContext` changes.
@@ -84,21 +100,12 @@ private extension DatadogContext {
 }
 
 private extension FeatureBaggage {
-    var rumContext: RUMContext? {
-        guard let applicationID: String = self[RUMDependency.ids, type: [String: String].self]?[RUMDependency.IDs.applicationIDKey],
-              let sessionID: String = self[RUMDependency.ids, type: [String: String].self]?[RUMDependency.IDs.sessionIDKey],
-              let viewID: String = self[RUMDependency.ids, type: [String: String].self]?[RUMDependency.IDs.viewIDKey],
-              let serverTimeOffset: TimeInterval = self[RUMDependency.serverTimeOffsetKey]
-        else {
-            // Current RUM session is not sampled
-            return nil
-        }
+    var rumContext: RUMContext? { try? unwrap() }
+}
 
-        return RUMContext(
-            applicationID: applicationID,
-            sessionID: sessionID,
-            viewID: viewID,
-            serverTimeOffset: serverTimeOffset
-        )
+extension FeatureBaggage {
+    func unwrap<T>() throws -> T where T: Decodable {
+        let decoder = AnyDecoder()
+        return try decoder.decode(from: attributes)
     }
 }
--- a/DatadogSessionReplay/Sources/DatadogSessionReplay/DatadogCoreIntegration/RUMContextReceiver.swift
+++ b/DatadogSessionReplay/Sources/DatadogSessionReplay/DatadogCoreIntegration/RUMContextReceiver.swift
@@ -15,6 +15,8 @@ internal struct RUMContext: Equatable {
     let sessionID: String
     /// Current RUM view ID - standard UUID string, lowecased.
     let viewID: String
+    /// Current RUM view server time offset (in seconds).
+    let serverTimeOffset: TimeInterval
 }
 
 /// An observer notifying on`RUMContext` changes.
@@ -85,12 +87,18 @@ private extension FeatureBaggage {
     var rumContext: RUMContext? {
         guard let applicationID: String = self[RUMDependency.applicationIDKey],
               let sessionID: String = self[RUMDependency.sessionIDKey],
-              let viewID: String = self[RUMDependency.viewIDKey]
+              let viewID: String = self[RUMDependency.viewIDKey],
+              let serverTimeOffset: TimeInterval = self[RUMDependency.serverTimeOffset]
         else {
             // Current RUM session is not sampled
             return nil
         }
 
-        return RUMContext(applicationID: applicationID, sessionID: sessionID, viewID: viewID)
+        return RUMContext(
+            applicationID: applicationID,
+            sessionID: sessionID,
+            viewID: viewID,
+            serverTimeOffset: serverTimeOffset
+        )
     }
 }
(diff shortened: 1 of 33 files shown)


DatadogSessionReplay/Sources/DatadogSessionReplay/Writer/Models/EnrichedRecord.swift
```diff
@@ -58,8 +58,8 @@ internal struct EnrichedRecord: Encodable {
         }
 
         self.hasFullSnapshot = hasFullSnapshot
-        self.earliestTimestamp = earliestTimestamp
-        self.latestTimestamp = latestTimestamp
+        self.earliestTimestamp = earliestTimestamp - rumContext.serverTimeOffset
+        self.latestTimestamp = latestTimestamp - rumContext.serverTimeOffset
```
maciejburda, line 62: I guess it should never go below zero? 🤔

Sources/Datadog/RUMMonitor.swift
```diff
@@ -616,7 +616,8 @@ public class RUMMonitor: DDRUMMonitor, RUMCommandSubscriber {
                     RUMContextAttributes.applicationID: context.rumApplicationID,
                     RUMContextAttributes.sessionID: context.sessionID.rawValue.uuidString.lowercased(),
                     RUMContextAttributes.viewID: context.activeViewID?.rawValue.uuidString.lowercased(),
-                    RUMContextAttributes.userActionID: context.activeUserActionID?.rawValue.uuidString.lowercased()
+                    RUMContextAttributes.userActionID: context.activeUserActionID?.rawValue.uuidString.lowercased(),
+                    RUMContextAttributes.serverTimeOffset: self.applicationScope.sessionScope?.viewScopes.last?.serverTimeOffset
```
ncreated, line 620: While usage of `.applicationID`, `.sessionID`, `.viewID` and `.userActionID` is tested [here](https://github.com/DataDog/dd-sdk-ios/blob/4f36c8a23f4416237d8b71846caab747c75b1e68/Tests/DatadogTests/Datadog/LoggerTests.swift#L542-L579), there is no coverage for `.serverTimeOffset`. The purpose of `RUMContextAttributes.serverTimeOffset` remains unclear in `Datadog` module without any test defined in `DatadogTests`.

I think we need a separate test case that will check if these attributes are set correctly in `DatadogContext`. It could explain that these 5 RUM attributes are used for integration with other Features in V2 (Logging + SR).

@maxep WDYT of this? (this is first time we're adding an attribute that has no use in V1 Features, hence we need to test it separately)

DatadogSessionReplay/Sources/DatadogSessionReplay/DatadogCoreIntegration/RUMContextReceiver.swift
```diff
@@ -85,12 +87,18 @@ private extension FeatureBaggage {
...
             // Current RUM session is not sampled
             return nil
         }
 
-        return RUMContext(applicationID: applicationID, sessionID: sessionID, viewID: viewID)
+        return RUMContext(
+            applicationID: applicationID,
+            sessionID: sessionID,
+            viewID: viewID,
+            serverTimeOffset: Int64(serverTimeOffset * 1_000)
```
ncreated, line 101: This may crash with overflow. We shouldn't make any assumption on `serverTimeOffset` being always "realistic" - I don't recall any upstream logic that will guard that.

Let's use [`Int64(withNoOverflow:)`](https://github.com/DataDog/dd-sdk-ios/blob/4f36c8a23f4416237d8b71846caab747c75b1e68/session-replay/Sources/DatadogSessionReplay/Utilities/SwiftExtensions.swift#L13) instead.

DatadogSessionReplay/Sources/DatadogSessionReplay/DatadogCoreIntegration/RUMContextReceiver.swift
```diff
@@ -85,12 +87,18 @@ private extension FeatureBaggage {
...
+        )
     }
 }
```
ncreated, line 104: Given that now we have `TestUtilities` module in place, we can define missing + add new tests in https://github.com/DataDog/dd-sdk-ios/blob/4f36c8a23f4416237d8b71846caab747c75b1e68/session-replay/Tests/DatadogSessionReplayTests/DatadogCoreIntegration/RUMContextReceiverTests.swift#L11-L19

DatadogSessionReplay/Sources/DatadogSessionReplay/DatadogCoreIntegration/RUMContextReceiver.swift
```diff
@@ -15,6 +15,8 @@ internal struct RUMContext: Equatable {
     let sessionID: String
     /// Current RUM view ID - standard UUID string, lowecased.
     let viewID: String
+    /// Current RUM view server time offeset.
```
ncreated, line 18: Typo + would be easier to read if we know the unit
```suggestion
    /// Current RUM view server time offset (in milliseconds).
```

DatadogSessionReplay/Sources/DatadogSessionReplay/Writer/Models/EnrichedRecord.swift
```diff
@@ -58,8 +58,8 @@ internal struct EnrichedRecord: Encodable {
         }
 
         self.hasFullSnapshot = hasFullSnapshot
-        self.earliestTimestamp = earliestTimestamp
-        self.latestTimestamp = latestTimestamp
+        self.earliestTimestamp = earliestTimestamp &- rumContext.serverTimeOffset
+        self.latestTimestamp = latestTimestamp &- rumContext.serverTimeOffset
```
ncreated, line 62: Applying the offset in this place doesn't make much sense - we only alter the `EnrichedRecord` which is a utility type used later to build `Segment`. In result, the segment will have correct "start" and "end" time, but all records in it will be still skewed (notice that this change ☝️ doesn't alter `self.records`).

If we apply the offset early in the `Recorder` layer - it will be automatically leveraged everywhere else (in `Processor`, `Writer` and finally `RequestBuilder`). There are two TODOs that indicate good strategic right places for date correction:

https://github.com/DataDog/dd-sdk-ios/blob/8dbd3472e52ac7156c45987f5796e47b079762c8/DatadogSessionReplay/Sources/DatadogSessionReplay/Recorder/TouchSnapshotProducer/WindowTouchSnapshotProducer.swift#L53

https://github.com/DataDog/dd-sdk-ios/blob/8dbd3472e52ac7156c45987f5796e47b079762c8/DatadogSessionReplay/Sources/DatadogSessionReplay/Recorder/Recorder.swift#L136

ncreated (changes requested):
Nice use of the `DatadogContext` 👍, but we're missing tests that will document this SR <> RUM dependency for V2.

Left few feedbacks - some are blockers. On top of that, I think the handling of offset is wrong in this PR. What we can read in RUM is that "server offset" should be **added to a date** (not subtracted):
https://github.com/DataDog/dd-sdk-ios/blob/4f36c8a23f4416237d8b71846caab747c75b1e68/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift#L49-L56

maxep:
Sharing the current view's NTP offset on the bus makes sense, but this value will be shared with [Logs attributes](https://github.com/DataDog/dd-sdk-ios/blob/develop/Sources/Datadog/Logging/RemoteLogger.swift#L128L130) and with the [tracer](https://github.com/DataDog/dd-sdk-ios/blob/develop/Sources/Datadog/Tracing/TracingV2Configuration.swift#L75).

We will need further changes to either separate the offset from the RUMContext, or update the `RemoteLogger` and `TracingMessageReceiver` to pick properties they need. I think we could benefit from #1112 and have a RUMContext types in both Logs and Tracing, I could add that in #1113! WDYT @maciejburda @ncreated ?

ncreated:
> We will need further changes to either separate the offset from the RUMContext, or update the `RemoteLogger` and `TracingMessageReceiver` to pick properties they need. I think we could benefit from #1112 and have a RUMContext types in both Logs and Tracing, I could add that in #1113! WDYT @maciejburda @ncreated ?

Indeed, because Logs take all "rum" baggage and associate it as log attributes, we end up with NTP offset being there. I agree, a way to move forward seems to be:
* having rich "rum" baggage in `DatadogContext` (as it is now)
* filtering it out to only necessary fields in each Feature (so dropping NTP offset in Logs).

This can be facilitated with #1112 yes! 👍

ncreated (changes requested):
Changes requested for better visibility on this: https://github.com/DataDog/dd-sdk-ios/pull/1131#issuecomment-1400180496

maciejburda:
@ncreated @maxep 

I made some changes. I went with nested dictionary approach as it felt simpler in the end.
I will create a ticket to refactor this to use internal feature model (preparation for v2), but I'd like to tackle it separately if it's not a problem.

ncreated (approved):
Looks very good 💪👍 with one comment on broken convention in `RUMDependency.swift`.

> I made some changes. I went with nested dictionary approach as it felt simpler in the end.
I will create a ticket to refactor this to use internal feature model (preparation for v2), but I'd like to tackle it separately if it's not a problem.

I'm okay with that, as we're clearly missing unified approach here. I think we should rather introduce it as regular V2 work instead // cc @maxep

DatadogSessionReplay/Sources/DatadogSessionReplay/DatadogCoreIntegration/RUMContextReceiver.swift
```diff
@@ -83,14 +85,20 @@ private extension DatadogContext {
...
             return nil
         }
 
-        return RUMContext(applicationID: applicationID, sessionID: sessionID, viewID: viewID)
+        return RUMContext(
+            applicationID: applicationID,
+            sessionID: sessionID,
+            viewID: viewID,
+            serverTimeOffset: serverTimeOffset
+        )
```
maxep, line 102: This is no longer necessary with the `AnyCoder`, you can make the `RUMContext` Decodable and let the `FeatureBaggage` do the decoding:

```swift
/// The RUM context received from `DatadogCore`.
internal struct RUMContext: Decodable, Equatable {
        case applicationID = "application_id"
        case sessionID = "session_id"
        case viewID = "view.id"
    }

    /// Current RUM application ID - standard UUID string, lowecased.
    let applicationID: String
    /// Current RUM session ID - standard UUID string, lowecased.
    let sessionID: String
    /// Current RUM view ID - standard UUID string, lowecased.
    let viewID: String
}

private extension FeatureBaggage {
    var rumContext: RUMContext? { self[RUMDependency.ids] }

    /// Current RUM view server time offset (in seconds).
    var viewTimeOffset: TimeInterval? { self[RUMDependency.serverTimeOffsetKey] }
}
```

With this, the enum `RUMDependency.IDs` is no longer necessary.
I would also suggest to rename the `serverTimeOffset` to `viewTimeOffset` to make it clear that it's from the current session.

maxep (changes requested):
This looks great 👏👏 I've left a minor change request to get full advantage of the new `AnyCoder`. It will simplify the way data is read from the `FeatureBaggage`

DatadogSessionReplay/Sources/DatadogSessionReplay/DatadogCoreIntegration/RUMContextReceiver.swift
maxep: The `FeatureBaggage` extension is actually not necessary, it already comes with `@dynamicMemberLookup` :)

DatadogSessionReplay/Sources/DatadogSessionReplay/DatadogCoreIntegration/RUMDependency.swift
```diff
@@ -8,30 +8,44 @@ import Foundation
...
+        /// The key for referencing RUM session ID inside RUM baggage.
+        ///
+        /// SR expects non-optional value holding lowercased, standard UUID `String`.
+        static let sessionIDKey = "session_id"
+
+        /// The key for referencing RUM view ID inside RUM baggage.
+        ///
+        /// SR expects non-optional value holding lowercased, standard UUID `String`.
+        static let viewIDKey = "view.id"
+    }
```
maxep, line 48: `serverTimeOffsetKey`, `ids`, and `enum IDs` are no longer necessary and can be removed 👍

maxep (approved):
Looks great! I see some left over in `RUMDependency`, they can be removed before merging 👍

PR #1132, opened 2023-01-17
Fix develop build issues

maciejburda:
### What and why?

Removes redundant protocol conformance from `DatadogBenchmarkTests`. They are now coming from `TestUtilities`.


ncreated (approved):
👌⭐

PR #1143, opened 2023-01-26
RUMM-2950 Fix log attribute forwarding to RUM

fuzzybinary:
### What and why?

When logs started going through the message bus, we accidentally started dropping the additional log attributes. This fixes it so we send the attributes properly. 

Additionally, we need to decode certain values specifically from the cross platform SDKs to ensure they're being set on the RUMErrorModel properly (sourceType and isCrash)

### How?

For now, we're doing "decoding" through a `decode` extension method on AttributeValue. This checks to see if the value is actually `DDAnyCodable` and casts its wrapped value the proper type if it is. Otherwise, it just returns the proper value or nil.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Tests/DatadogTests/Datadog/LoggerTests.swift
+++ b/Tests/DatadogTests/Datadog/LoggerTests.swift
@@ -622,6 +622,150 @@ class LoggerTests: XCTestCase {
         }
     }
 
+    func testWhenSendingErrorOrCriticalLogsWithAttributes_itCreatesRUMErrorForCurrentViewWithAttributes() throws {
+        let logging: LoggingFeature = .mockAny()
+        core.register(feature: logging)
+
+        let rum: RUMFeature = .mockWith(messageReceiver: ErrorMessageReceiver())
+        core.register(feature: rum)
+
+        // given
+        let logger = Logger.builder.build(in: core)
+        Global.rum = RUMMonitor(
+            core: core,
+            dependencies: RUMScopeDependencies(
+                core: core,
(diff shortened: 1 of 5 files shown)


Tests/DatadogTests/Datadog/LoggerTests.swift
```diff
@@ -622,6 +622,150 @@ class LoggerTests: XCTestCase {
...
+        logger.error("error message", attributes: [
+            "any_attribute_a": attributeValueA
+        ])
+        let attributeValueB: String = .mockRandom()
+        logger.critical("critical message", attributes: [
+            "any_attribute_b": attributeValueB
+        ])
+
+        // then
+        let rumEventMatchers = try core.waitAndReturnRUMEventMatchers()
```
ncreated, line 656: nit/ We should rather use the new `core.waitAndReturnEvents(of:ofType:)` API [added with `DatadogCoreProxy`](https://github.com/DataDog/dd-sdk-ios/wiki/How-to-use-%60DatadogCoreProxy%60-in-tests%3F#how-to-use-datadogcoreproxy). It leads to much simpler assertions and does not throw any errors.

fuzzybinary:
Agreed on finding a standardized approach. Should we create a ticket to track?

ncreated (approved):
> Agreed on finding a standardized approach. Should we create a ticket to track?

Let's add it, so we don't miss this 🙏. Thanks!

PR #1145, opened 2023-01-30
Increase error count by 1 on the view update event following a crash

plousada:
### What and why?
Following a crash, when updating a view with the crash count increase we should also increase the error count. 

### How?

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/RUM/Integrations/CrashReportReceiver.swift
+++ b/Sources/Datadog/RUM/Integrations/CrashReportReceiver.swift
@@ -399,7 +399,7 @@ internal struct CrashReportReceiver: FeatureMessageReceiver {
                 domComplete: original.view.domComplete,
                 domContentLoaded: original.view.domContentLoaded,
                 domInteractive: original.view.domInteractive,
-                error: original.view.error,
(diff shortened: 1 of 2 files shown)


Sources/Datadog/RUM/Integrations/CrashReportReceiver.swift
```diff
@@ -399,7 +399,7 @@ internal struct CrashReportReceiver: FeatureMessageReceiver {
                 domComplete: original.view.domComplete,
                 domContentLoaded: original.view.domContentLoaded,
                 domInteractive: original.view.domInteractive,
-                error: original.view.error,
+                error: .init(count: original.view.error.count + 1),
```
maciejburda, line 402: Error model for now only has count, but I think it would be safer to copy object `original.view.error` and then add 1 to the count (just so it doesn't break when we alter the model).

maciejburda (approved):
One nonblocking comment :)

maxep (approved):
LGTM 👍 
And CI is actually green:
image

xgouchet (approved):

PR #1147, opened 2023-01-31
RUMM-2507 Change ApplicationLaunch view behavior

fuzzybinary:
### What and why?

In order to more accurately track application launch and loading times, we are changing the way the `ApplicationLaunch` view and `application_start` actions work.  

Now, initializing the Datadog SDK with RUM will create the `ApplicationLaunch` view by default, staring at the process launch time. This view can then track any resources, timing, or other events (including user actions) until the first call to `startView`. 

The `application_start` event spans the entire time of the `ApplicationLaunch` view. 

In the event of an active prewarm, the staring time of the `ApplicationLaunch` view is changed to the time the SDK was initialized, and the loading time of the `application_start` event is ignored.

### How?

One big change to the UnitTests here is that now requesting RUM events from the matchers will discard the ApplicationLaunch view, as most tests assume such a view doesn't exist by default. Similarly, the matcher for the Session pulls the ApplicationLaunch view (and all associated events) into its own member, again so that tests that don't expect to see that view still work.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [x] Run smoke tests

### DIFF
--- a/Tests/DatadogTests/Datadog/Mocks/RUMFeatureMocks.swift
+++ b/Tests/DatadogTests/Datadog/Mocks/RUMFeatureMocks.swift
@@ -154,7 +154,6 @@ struct RUMCommandMock: RUMCommand {
     var time = Date()
     var attributes: [AttributeKey: AttributeValue] = [:]
     var canStartBackgroundView = false
-    var canStartApplicationLaunchView = false
     var isUserInteraction = false
 }
 
--- a/Tests/DatadogTests/Datadog/Mocks/RUMFeatureMocks.swift
+++ b/Tests/DatadogTests/Datadog/Mocks/RUMFeatureMocks.swift
@@ -787,7 +787,7 @@ extension RUMViewScope {
     }
 
     static func mockWith(
-        isInitialView: Bool = false,
+        shouldSendApplicationLaunch: Bool = false,
         parent: RUMContextProvider = RUMContextProviderMock(),
         dependencies: RUMScopeDependencies = .mockAny(),
         identity: RUMViewIdentifiable = mockView,
@@ -799,7 +799,7 @@ extension RUMViewScope {
         serverTimeOffset: TimeInterval = .zero
     ) -> RUMViewScope {
         return RUMViewScope(
-            isInitialView: isInitialView,
+            shouldSendApplicationLaunch: shouldSendApplicationLaunch,
             parent: parent,
             dependencies: dependencies,
             identity: identity,
--- a/Tests/DatadogTests/Datadog/Mocks/RUMFeatureMocks.swift
+++ b/Tests/DatadogTests/Datadog/Mocks/RUMFeatureMocks.swift
@@ -26,9 +26,17 @@ extension RUMFeature {
 }
 
 extension DatadogCoreProxy {
-    func waitAndReturnRUMEventMatchers(file: StaticString = #file, line: UInt = #line) throws -> [RUMEventMatcher] {
-        return try waitAndReturnEventsData(of: RUMFeature.self)
+    func waitAndReturnRUMEventMatchers(removeApplicationLaunch: Bool = true,
+                                       file: StaticString = #file,
+                                       line: UInt = #line) throws -> [RUMEventMatcher] {
+        var matchers = try waitAndReturnEventsData(of: RUMFeature.self)
             .map { data in try RUMEventMatcher.fromJSONObjectData(data) }
+        if removeApplicationLaunch {
+            matchers = matchers.filter({ matcher in
+                (try? matcher.attribute(forKeyPath: "view.name")) != RUMOffViewEventsHandlingRule.Constants.applicationLaunchViewName
+            })
+        }
+        return matchers
     }
 }
 
--- a/Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScopeTests.swift
+++ b/Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScopeTests.swift
@@ -246,12 +246,17 @@ class RUMSessionScopeTests: XCTestCase {
 
     // MARK: - Application Launch Events Tracking
 
-    func testGivenAppInForegroundAndInitialSessionWithNoViewTrackedBefore_whenCommandCanStartApplicationLaunchView_itCreatesAppLaunchScope() {
+    func testGivenAppInForegroundAndInitialSessionWithNoViewTrackedBefore_itCreatesAppLaunchScope() {
         // Given
         let sessionStartTime = Date()
 
         var context = self.context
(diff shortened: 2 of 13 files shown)


fuzzybinary:
I want to get documentation in as part of this PR. I'll update it when I have that done.

docs/rum_collection/data_collected.md
```diff
@@ -32,6 +32,12 @@ The following diagram illustrates the RUM event hierarchy:
 
 {{< img src="real_user_monitoring/data_collected/event-hierarchy.png" alt="RUM Event hierarchy" style="width:50%;border:none" >}}
 
+## Application Launch
```
urseberry, line 35:
```suggestion
## Application launch
```

docs/rum_collection/data_collected.md
```diff
@@ -32,6 +32,12 @@ The following diagram illustrates the RUM event hierarchy:
...
+
+During initialization the RUM iOS SDK creates a view called "ApplicationLaunch". This view's start time is set to the start of the iOS process, and can be used to track your applicaiton launch time, and will include any logs, actions, and resources made before your first call to `startView`. This view will also have an action, `applicaiton_start`, that spans the entire length of the view, and its duration can be used to time to first view.
```
urseberry, line 37:
```suggestion
During initialization, the RUM iOS SDK creates a view called "ApplicationLaunch". This view's start time matches the start of the iOS process, and can be used to track your application launch time.

The ApplicationLaunch view includes any logs, actions, and resources created before your first call to `startView`. This view has an action, `application_start`, that spans the entire length of the view. Use the duration of `application_start` to determine time to first view.
```

docs/rum_collection/data_collected.md
```diff
@@ -32,6 +32,12 @@ The following diagram illustrates the RUM event hierarchy:
...
+
+In cases where iOS decides to [prewarm your application][4], the ApplicationLaunch view will instead start when the RUM iOS SDK is initialized, and the `application_start` event will not have a duration.
```
urseberry, line 39:
```suggestion
In cases where iOS decides to [prewarm your application][4], the ApplicationLaunch view instead starts when the RUM iOS SDK is initialized, and the `application_start` event does not have a duration.
```

xgouchet (approved):

Tests/DatadogTests/Datadog/Mocks/RUMFeatureMocks.swift
```diff
@@ -26,9 +26,17 @@ extension RUMFeature {
...
+    func waitAndReturnRUMEventMatchers(removeApplicationLaunch: Bool = true,
+                                       file: StaticString = #file,
+                                       line: UInt = #line) throws -> [RUMEventMatcher] {
+        var matchers = try waitAndReturnEventsData(of: RUMFeature.self)
             .map { data in try RUMEventMatcher.fromJSONObjectData(data) }
+        if removeApplicationLaunch {
+            matchers = matchers.filter({ matcher in
+                (try? matcher.attribute(forKeyPath: "view.name")) != RUMOffViewEventsHandlingRule.Constants.applicationLaunchViewName
+            })
+        }
```
maxep, line 38: The other option is to update all tests? what impact would that be?
I think such filtering should be per test instead of having it hidden in the `waitAndReturnRUMEventMatchers`, but It might be impactful..

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScopeTests.swift
```diff
@@ -237,186 +237,13 @@ class RUMSessionScopeTests: XCTestCase {
...
         let commandTime = sessionStartTime.addingTimeInterval(1)
-        let command = RUMCommandMock(time: commandTime, canStartBackgroundView: .mockRandom(), canStartApplicationLaunchView: false)
+        let command = RUMCommandMock(time: commandTime, canStartBackgroundView: .mockRandom())
         XCTAssertTrue(scope.process(command: command, context: context, writer: writer))
 
         // Then
         XCTAssertTrue(scope.viewScopes.isEmpty, "It should not start any view scope")
     }
 
-    // MARK: - Application Launch Events Tracking
```
maxep, line 247: Most of those tests are still relevant, no?

fuzzybinary: I don't think so? Most of these deal with automatically creating the ApplicationLaunch view on commands that can do so, (the removed `canStartApplicationLaunchView` property) and should be covered by the new tests instead.

Looking through, I think I am missing a test for not creating the ApplicaitonLaunch view if it's not the initial session or booted after the background view.

maxep: I think the `testGivenAppInBackgroundAndBETDisabled_whenReceivingAnyCommand_itDoesNotCreateAnyScope` can be kept.

Tests/DatadogTests/Datadog/Mocks/RUMFeatureMocks.swift
fuzzybinary: Yes, that's the other option, it's around 30 tests.  Very few tests happen to actually care about the launch view, which is why I put this here. I could make that parameter explicit, so it's not quite as hidden?

Tests/DatadogTests/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScopeTests.swift
fuzzybinary: Probably the BET enabled one as well. I might move them into where the other ApplicaiotLaunch events are tracked as well? Yes?

maxep: Sounds good, yes 👍

fuzzybinary: Yes, `testWhenConfigurationSourceIsSet_applicationStartUsesTheConfigurationSource` definitely shouldn't have been removed. Good catch!

The other tests I think are no longer correct, but definitely please let me know if you disagree. New logic here is still a bit confusing.

`testWhenNoLoadingTime_itSendsApplicationStartAction_basedOnLoadingDate` Is no longer correct as "LoadingTime" (actually `launchTime`) is the time until `applicationDidBecomeActive`, whereas we now time until first view.

`testWhenActivePrewarm_itSendsApplicationStartAction_withoutLoadingTime` is now covered by `testWhenAppLaunchesWithActivePrewarm_itDoesNotSendApplicaitonStartDuration` (in RUMMonitorTests)

urseberry (approved):

fuzzybinary:
Closing because we're rethinking requirements.

PR #1148, opened 2023-02-01
RUMM-2939 Use TestUtilities as Module

maxep:
### What and why?

Use `TestUtilities` as independent package.

### How?

`TestUtilities` is now used as a separate module in existing test suites.
Targets for iOS and tvOS are created in the xcodeproj to facilitate the migration.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [x] Run smoke tests


maciejburda (approved):

TestUtilities/Package.swift
```diff
@@ -11,14 +11,18 @@ let package = Package(
...
     ],
     dependencies: [
         .package(name: "Datadog", path: ".."),
     ],
     targets: [
         .target(
             name: "TestUtilities",
-            dependencies: ["Datadog"])
+            dependencies: ["Datadog"],
+            path: ".",
```
maciejburda, line 24: I think it should be default path. Is there a reason to have it or it's just a leftover? 🤔

Tests/DatadogTests/Datadog/Mocks/RUMDataModelMocks.swift
```diff
@@ -4,6 +4,7 @@
  * Copyright 2019-Present Datadog, Inc.
  */
 
+import TestUtilities
```
maciejburda, line 7: One thing I just remembered. The idea was to use `@testable import TestUtilities` mainly to avoid changing all the access modifiers inside the package. In the end of the day it actually turns all internal to public automatically.

I'm wondering why you didn't have to change anything inside TestUtilities and it still worked 🤔

maciejburda (approved):
LGTM

Question so just so I can understand more deeply:

Is this PR removing compiled sources links from the "monotarget" and uses the `TestUtilities` package directly?

maxep:
image

Tests/DatadogTests/Datadog/Mocks/RUMDataModelMocks.swift
maxep: I did change all accessor to `public` :) It was only few files in the package.

maxep:
> Is this PR removing compiled sources links from the "monotarget" and uses the TestUtilities package directly?

PR #1150, opened 2023-02-01
RUMM-2889 Support untracked modals

fuzzybinary:
### What and why?

This allows modal views to be "untracked" (not call startView / stopView) when using automatic view tracking. The view disappearing under the modal will still call `stopView`, and will restart when the modal is dismissed.  This is useful for  when a user wants to manually track a view controller that is presented modally, or when a framework (such as Flutter or React Native) wants to take over view reporting.

### How?

Pre iOS 13, users will need to set `isUntrackedModal` on the view they return from the predicate. This will push the view onto the `RUMViewsHandler` stack, but this view will not be started or stopped by the `RUMViewsHandler`.

On iOS 13+, views that are marked with `isModalInPresentation`, but return null from the `rumViewPredicate` are automatically considered to be untracked modals.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Datadog/Example/Scenarios/RUM/RUMScenarios.swift
+++ b/Datadog/Example/Scenarios/RUM/RUMScenarios.swift
@@ -74,7 +74,6 @@ final class RUMModalViewsAutoInstrumentationScenario: TestScenario {
     private class Predicate: UIKitRUMViewsPredicate {
         func rumView(for viewController: UIViewController) -> RUMView? {
             if let viewName = viewController.accessibilityLabel {
-                print(viewName)
                 return .init(name: viewName)
             } else {
                 return nil
@@ -99,8 +98,8 @@ final class RUMUntrackedModalViewsAutoInstrumentationScenario: TestScenario {
         func rumView(for viewController: UIViewController) -> RUMView? {
             if let viewName = viewController.accessibilityLabel {
                 if viewController.modalPresentationStyle == .fullScreen {
-                    if #available(iOS 13, *) {
-                        // Untracked on iOS 13 via isModalInPresentation
+                    if #available(iOS 13, tvOS 13, *) {
+                        // Untracked on iOS/tvOS 13+ via isModalInPresentation
                         return nil
                     } else {
                         return .init(name: viewName, isUntrackedModal: true)
--- a/Datadog/Example/Scenarios/RUM/RUMScenarios.swift
+++ b/Datadog/Example/Scenarios/RUM/RUMScenarios.swift
@@ -74,6 +74,38 @@ final class RUMModalViewsAutoInstrumentationScenario: TestScenario {
     private class Predicate: UIKitRUMViewsPredicate {
         func rumView(for viewController: UIViewController) -> RUMView? {
             if let viewName = viewController.accessibilityLabel {
+                print(viewName)
+                return .init(name: viewName)
+            } else {
+                return nil
+            }
+        }
+    }
+
+    func configureSDK(builder: Datadog.Configuration.Builder) {
+        _ = builder
+            .trackUIKitRUMViews(using: Predicate())
+            .enableLogging(false)
+            .enableTracing(false)
+    }
+}
+
+/// Scenario based on `UINavigationController` hierarchy, which presents different VCs modally.
+/// Tracks view controllers as RUM Views.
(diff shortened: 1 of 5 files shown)


maxep (commented):
Looks great, thanks! I left 2 minor comments.
Do you think it would be valuable to have an Integration test scenario where we randomly pick a modal presentation style and we assert that the ignored modal stops the previous view?

fuzzybinary:
Yes, agree that's probably a good integration test. I'll add it.

Sources/Datadog/RUM/Instrumentation/Views/RUMViewsHandler.swift
```diff
@@ -103,11 +106,16 @@ internal final class RUMViewsHandler {
 
         // Stop the last appearing view of the stack
         if let current = stack.last {
-            stop(view: current)
+            if !current.isUntrackedModal {
```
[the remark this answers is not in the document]
fuzzybinary: Yeah, I can move them it into start / stop if you think that's better.

Datadog/Example/Scenarios/RUM/RUMScenarios.swift
```diff
@@ -74,6 +74,38 @@ final class RUMModalViewsAutoInstrumentationScenario: TestScenario {
     private class Predicate: UIKitRUMViewsPredicate {
         func rumView(for viewController: UIViewController) -> RUMView? {
             if let viewName = viewController.accessibilityLabel {
+                print(viewName)
```
maciejburda, line 77: Probably a leftover `print()` 🧹

Datadog/Example/Scenarios/RUM/RUMScenarios.swift
```diff
@@ -74,6 +74,38 @@ final class RUMModalViewsAutoInstrumentationScenario: TestScenario {
...
+                return nil
+            }
+        }
+    }
+
+    func configureSDK(builder: Datadog.Configuration.Builder) {
+        _ = builder
+            .trackUIKitRUMViews(using: Predicate())
+            .enableLogging(false)
+            .enableTracing(false)
```
maciejburda, line 89: Side note:

Would be great to improve the API in V2. IMO it should be `.loggingEnabled(true)` an so on. cc @maxep

Datadog/Example/Scenarios/RUM/RUMScenarios.swift
```diff
@@ -74,6 +74,38 @@ final class RUMModalViewsAutoInstrumentationScenario: TestScenario {
...
             if let viewName = viewController.accessibilityLabel {
+                print(viewName)
+                return .init(name: viewName)
+            } else {
+                return nil
+            }
+        }
+    }
+
+    func configureSDK(builder: Datadog.Configuration.Builder) {
```
maciejburda, line 85: I don't think it's used 🤔

Datadog/Example/Scenarios/RUM/RUMScenarios.swift
```diff
@@ -74,6 +74,38 @@ final class RUMModalViewsAutoInstrumentationScenario: TestScenario {
...
+/// Tracks view controllers as RUM Views.
+final class RUMUntrackedModalViewsAutoInstrumentationScenario: TestScenario {
+    static var storyboardName: String = "RUMModalViewsAutoInstrumentationScenario"
+
+    private class Predicate: UIKitRUMViewsPredicate {
+        func rumView(for viewController: UIViewController) -> RUMView? {
+            if let viewName = viewController.accessibilityLabel {
+                if viewController.modalPresentationStyle == .fullScreen {
+                    if #available(iOS 13, *) {
+                        // Untracked on iOS 13 via isModalInPresentation
```
maciejburda, line 103:
```suggestion
                        // Untracked on iOS 13+ via isModalInPresentation
```

maciejburda (approved):

Datadog/Example/Scenarios/RUM/RUMScenarios.swift
fuzzybinary: It is. It's called after the scenario boots. For some reason you can't set a breakpoint in it though.

maxep (approved):
Nice! thank you!

PR #1152, opened 2023-02-03
Fix a behavior change in documentation links

iGranDav:
### What and why?

It seems that `tab=us` doesn't exists anymore on those links causing empty code blocks on the page.

Removing the `tag` defaults to the first tab being selected.

Thanks @bill350 for the report

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,19 +6,19 @@
 
 ### Log Collection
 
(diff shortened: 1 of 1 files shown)


maciejburda (approved):

maxep (approved):
🙏

xgouchet (approved):

PR #1154, opened 2023-02-06
RUMM-2938 Create DatadogInternal Module

maxep:
### What and why?

Create the `DatadogInternal` module that will be shared will all Datadog modules. It includes Interface definitions and utilities to build and Core and Features.

> Sorry for the large PR, Xcode targets contains most of the added lines.

### How?

The `DatadogInternal` module is created as both a SPM target and as a Xcode target while doing the migration. Access levels have been updated accordingly 

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [x] Run smoke tests


DatadogInternal/Sources/Context/DatadogSite.swift
```diff
@@ -0,0 +1,37 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+public enum DatadogSite: String {
```
maxep, line 9: Needs to be updated with #1157

DatadogInternal/Sources/DD.swift
```diff
@@ -11,17 +11,17 @@ import Foundation
...
     /// It is meant for debugging purposes when using the SDK, hence **it should log information useful and actionable
     /// to the SDK user**. Think of possible logs that we may want to receive from our users when asking them to enable
     /// SDK verbosity and send us their console log.
-    static var logger: CoreLogger = InternalLogger(
+    public static var logger: CoreLogger = InternalLogger(
         dateProvider: SystemDateProvider(),
         timeZone: .current,
         printFunction: consolePrint,
-        verbosityLevel: { Datadog.verbosityLevel }
+        verbosityLevel: { .debug }
```
maxep, line 24: It is now set to `debug` by default, it is then changed after SDK init.

DatadogInternal/Sources/Telemetry/InternalLogger.swift
```diff
@@ -18,13 +18,13 @@ internal struct InternalLogger: CoreLogger {
     /// The print function.
     private let printFunction: (String) -> Void
     /// V1's verbosity level. Only logs above or equal to this level wil be printed.
-    private let currentVerbosityLevel: () -> LogLevel?
+    private let currentVerbosityLevel: () -> CoreLoggerLevel?
```
maxep, line 21: It no longer use the Logs Feature levels.

DatadogInternal/Sources/Extensions/DatadogExtended.swift
```diff
@@ -48,7 +48,7 @@ extension DatadogExtended {
 }
 
 /// Protocol describing the `_internal` extension points for Datadog extended types.
-public protocol DatadogInternal {
+public protocol DatadogInternalInterface {
```
maxep, line 51: This is renamed to no conflict with the package name.

Sources/Datadog/DatadogConfiguration.swift
```diff
@@ -48,66 +49,6 @@ extension Datadog {
             case never
         }
 
-        public enum DatadogEndpoint: String {
```
maxep, line 51: Moved to `DatadogInternal` as `DatadogSite`.

Tests/DatadogTests/Datadog/DatadogTests.swift
```diff
@@ -92,21 +93,6 @@ class DatadogTests: XCTestCase {
...
-            appContext: .mockAny(),
-            trackingConsent: .mockRandom(),
-            configuration: defaultBuilder.build()
-        )
-
-        XCTAssertEqual(dd.telemetry.configurations.count, 1)
-
-        Datadog.flushAndDeinitialize()
-    }
-
```
maxep, line 109: I haven't found a replacement yet :/

tools/rum-models-generator/Sources/rum-models-generator/GenerateRUMModels.swift
```diff
@@ -19,6 +19,8 @@ internal func generateRUMSwiftModels(from schema: URL) throws -> String {
              * Copyright 2019-Present Datadog, Inc.
              */
 
+            import DatadogInternal
```
maxep, line 22: The model needs the `DynamicCodingKey` from the internal module.

DatadogInternal/Sources/Context/DatadogSite.swift
maxep: Done.

Datadog/Example/AppConfiguration.swift
```diff
@@ -6,6 +6,7 @@
 
 import UIKit
 import Datadog
+import DatadogInternal
```
ncreated, line 9: Why do we need to import `DatadogInternal` in application code? In theory, `DatadogInternal` shouldn't be used by end-users. Is this required by migration some step or Example project specifics?

DatadogInternal/Sources/Upload/DataCompression.swift
```diff
@@ -16,7 +16,7 @@ import zlib
 /// - https://zlib.net/
 /// - https://datatracker.ietf.org/doc/html/rfc1950
 /// - https://datatracker.ietf.org/doc/html/rfc1951
-/* public */ internal struct Deflate {
+public struct Deflate {
```
ncreated, line 19: question/ Do we need compression things in public interface? The `URLRequestBuilder` handles compression:
```swift
// URLRequestBuilder.swift

func uploadRequest(with body: Data, compress: Bool = true)
```
so details of the actual `encode()`, `compress()` and `addler32()` could be internal, no?

Sources/Datadog/DatadogCore/DatadogCore.swift
```diff
@@ -504,7 +505,7 @@ extension DatadogContextProvider {
         serverDateProvider: ServerDateProvider
     ) {
         let context = DatadogContext(
-            site: configuration.site,
+            site: configuration.site ?? .us1,
```
ncreated, line 508: suggestion/ Defaults resolution is error-prone and dangerous if done in different places. For that reason, in V1 convention was to resolve defaults early in configuration declaration - so we don't need to struggle with optionals in downstream logic.

I think the root cause of optional here is:
```swift
/// [Datadog Site](https://docs.datadoghq.com/getting_started/site/) for data uploads. It can be `nil` in V1
/// if the SDK is configured using deprecated APIs: `set(logsEndpoint:)`, `set(tracesEndpoint:)` and `set(rumEndpoint:)`.
let site: DatadogSite?
```
I'd recommend tackling this legacy early and individually (isolated PR would be best). If we remove these deprecated APIs on `v2` branch, it will streamline further changes around `site` logic.

DatadogInternal/Sources/Telemetry/Telemetry.swift
```diff
@@ -24,12 +24,6 @@ internal protocol Telemetry {
     ///   - kind: The kind of error.
     ///   - stack: The stack trace.
     func error(id: String, message: String, kind: String?, stack: String?)
-
-    /// Sends a `TelemetryConfigurationEvent` event.
-    ///
-    /// - Parameters:
-    ///   - configuration: The current configuration
-    func configuration(configuration: FeaturesConfiguration)
```
ncreated, line 32: question/ How does configuration telemetry work now? I struggled with finding corresponding changes, only noticed disabled test 🤔💭.

ncreated (commented):
Extreme efforts, thanks @maxep 🏋️‍♂️🏅.

A bit hard to navigate in this PR. I can see diffs from few other PRs as well (SR bitmaps, `ap1()` DC, etc.) - I assume it comes from `→ develop` update. If possible, let's try smaller changes in next migration work 🙏 as it visibly impacts the quality of CR I can give here.

I left few questions to clarify some elements before proceeding with merge. But it looks great in overall 🚀!

maxep:
Thanks for the review @ncreated 🙏 

Yes, `feature/v2` needs an update with `develop`, I'm currently rebasing it so this PR will get a bit smaller.

Datadog/Example/AppConfiguration.swift
maxep: Ah good catch! we need to expose some types from `DatadogInternal`, but this should be transparent for the user.
We can either use wrapper types or use the `@_exported import` for some types. That second option is easier!

Sources/Datadog/DatadogCore/DatadogCore.swift
maxep: Good point, I agree 👍

Datadog/Example/AppConfiguration.swift
ncreated: > we need to expose some types from DatadogInternal, but this should be transparent for the user.

Which ones for example? Yes, I think `@_exported import` in our `Example` project should do the job 👍.

DatadogInternal/Sources/Telemetry/Telemetry.swift
maxep: The `Telemetry` protocol does not require this method, but the `RUMTelemetry` does expose it. The reason being that the `FeaturesConfiguration` type is not part of the `DatadogInternal`, it's define in the `Datadog` module as well as `RUMTelemetry`. [The configuration is still reported at initialisation](https://github.com/DataDog/dd-sdk-ios/pull/1154/files#diff-7fae8d57fd7c8b0dc28f762018d1019d1f75530a064e083e7828d309e182be3dR299).

It's still unclear how we will tackle reporting configuration telemetry, but it will be tackled separately.

Datadog/Example/AppConfiguration.swift
maxep: Currently, I see the following (it needs to be defined in `Datadog` module)

```swift
@_exported import enum DatadogInternal.DatadogSite
@_exported import struct DatadogInternal.AttributeKey
@_exported import protocol DatadogInternal.AttributeValue
@_exported import enum DatadogInternal.TrackingConsent
```

**Note:** I'm thinking to remove `AttributeKey` and `AttributeValue` aliases at some point

ncreated: Ok, now it makes sense 👌 - we need these symbols in `Example` app because we add custom logics around it (e.g. use it in variables declaration). Such use case won't happen in customer app, because the symbol will be visible through `Datadog` or other Feature modules.

> Note: I'm thinking to remove AttributeKey and AttributeValue aliases at some point

Their role today is to [describe with code comments](https://github.com/DataDog/dd-sdk-ios/blob/5491d29171e08c939ee09cc0d6e51ad66309daa6/Sources/Datadog/Core/Attributes/Attributes.swift#L9-L63) how keys and values work across all products. If we remove these aliases, where would we describe it? The only place would be next to each method that depends on attributes, which would be tedious. Not sure why we'd want to remove them?

Sources/Datadog/DatadogConfiguration.swift
```diff
@@ -5,7 +5,11 @@
  */
 
 import Foundation
-import DatadogInternal
+
+@_exported import enum DatadogInternal.DatadogSite
+@_exported import struct DatadogInternal.AttributeKey
+@_exported import protocol DatadogInternal.AttributeValue
+@_exported import enum DatadogInternal.TrackingConsent
```
ncreated, line 12: I was thinking we will do it in `Example` app project - do it once (in `AppDelegate`?) and have these symbols re-exported to the entire app module 🤔💭. This could be minimal and isolated solution to our inconvenience.

Unlike what we seek for in `Example` app, our users don't need to have these types available for variables declaration, e.g.:
```swift
let trackingConent: TrackingConsent
```
They will only use it for variables assignment:
```swift
.set(trackingConsent:)
```
so we don't need to re-export their visibility.

Also, `@_exported` is discouraged by Swift maintainers, same as other [underscored attributes](https://github.com/apple/swift/blob/main/docs/ReferenceGuides/UnderscoredAttributes.md). Better to avoid it if possible 😉. Here, it's perfectly possible as the problem is only in our `Example` project. WDYT @maxep ?

Datadog/Example/AppConfiguration.swift
maxep: Yes, the doc well explains it, but it is necessary because `[AttributeKey: AttributeValue]` is kindof confusing. While `[String: Encodable]` is clearer IMO. We could use a typealias like 
```swift
typealias DatadogAdditionalAttributes = [String: Encodable]
```

ncreated (approved):
Looks great! I only left one [feedback](https://github.com/DataDog/dd-sdk-ios/pull/1154/commits/b8b78e2c6ccb2ecafbc62b5bce0fa749492ff5e5#r1105517772) on `@_exported import` (please have a look). Other than this, good to go, excellent 🚀!

Datadog/Example/AppConfiguration.swift
ncreated: `DatadogAdditionalAttributes = [String: Encodable]` could work as well 👍. I like type aliases because they have some benefits of extra types without requiring us to introduce new `structs` in fact. Let's keep this for later, agree 👍.

Sources/Datadog/DatadogConfiguration.swift
maxep: I think we only need the `TrackingConsent` in the Example, I will double check.

But it's very likely that users will need it as well, to implement any consent logic, e.g. storing in UserDefault or map the type with a Consent Provider. So we would probably want to expose it somehow, no?

I saw that the underscored attributes are discouraged but I don't see the risk with this one, other than semantic change. But to avoid any, the other option is having a wrapper exposed by the `Datadog` module:
```swift
public enum DatadogTrackingConsent {
    ...
    internal var _internal: DatatogInternal.TrackingConsent { get }
}
```
I will go with that option, if that' ok with you?

maxep: Well, the wrapper has a larger impact than I thought :/ I will go with your suggestion to move forward 👍

ncreated: > But it's very likely that users will need it as well, to implement any consent logic, e.g. storing in UserDefault or map the type with a Consent Provider. So we would probably want to expose it somehow, no?

Could be 👌, but this is quite an assumption that we have no proof for. I'd say let's keep it simple for now - if one asks for it, we will think of solution. We could always add thin wrapper just for SDK configuration part (in `Datadog`) that maps into `DatadogInternal.TrackingConsent` right after configuration (with no other use than configuration it will have minimal to no impact).

PR #1158, opened 2023-02-07
RUMM-2939 Remove DatadogSessionReplay Package

maxep:
### What and why?

Make `DatadogSessionReplay` as part of the main package. This is an indirect change necessary for building the Integration Tests project.

This is necessary to rename the nested `TestUtilities` package so it no longer conflicts with the one define in the main `Package.swift`. Here how the dependency tree looks like:

```
IntegrationTests
│
├─ Datadog Package
│   └─ TestUtilities // it includes but does not expose a TestUtilities target
│
└─ TestUtilities Package
    └─ TestUtilities // should be renamed to not conflict with TestUtilities from Datadog package
```

We could avoid this by using SPM `moduleAliases` but I'm currently facing this bug: https://github.com/apple/swift-package-manager/issues/5762

rdar://104768426

### How?

Create SPM target and remove nested package.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [x] Run smoke tests


maxep:
I'm looking for another workaround, I will probably close this!

PR #1159, opened 2023-02-08
Fix a bug caught by integration test

maciejburda:
### What and why?

There was an issue caught by integration test. It was caused by existence of optional field if the dictionary.

### How?

This PR fixes the type casting to include optional strings. Way to go is to refactor the communication to always use Codable model that describes the contract. Created a ticket to tackle that next week `RUMM-3023`.

- [ ] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


maxep (approved):
Good catch!

ncreated (approved):

PR #1160, opened 2023-02-08
RUMM-2507 Create an ApplicationLaunch view during session initialization

fuzzybinary:
### What and why?

In order to more accurately track application launch and loading times, we are changing the way the ApplicationLaunch view work.

Now, initializing the Datadog SDK with RUM will create the ApplicationLaunch view by default, staring at the process launch time. This view can then track any resources, timing, or other events (including user actions) until the first call to startView.

The application_start event will now always be on the ApplicationLaunch view, but it's timing remains unchanged.

In the event of an active prewarm, the staring time of the ApplicationLaunch view is changed to the time the SDK was initialized, and the loading time of the application_start event is ignored.

### How?

One big change to the UnitTests here is that now requesting RUM events from the matchers will discard the ApplicationLaunch view, as most tests assume such a view doesn't exist by default. 

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [x] Run smoke tests


docs/rum_collection/data_collected.md
```diff
@@ -32,6 +32,14 @@ The following diagram illustrates the RUM event hierarchy:
 
 {{< img src="real_user_monitoring/data_collected/event-hierarchy.png" alt="RUM Event hierarchy" style="width:50%;border:none" >}}
 
+## Application launch
+
+During initialization, the RUM iOS SDK creates a view called "ApplicationLaunch". This view's start time matches the start of the iOS process, and can be used to track your application launch time.
+
+The ApplicationLaunch view includes any logs, actions, and resources created before your first call to `startView`. Use the duration of this view to determine time to first view. This view has an action, `application_start`, with a duration equal to the amount of time from process start until the call to `applicationDidBecomeActive`.
+
+In cases where iOS decides to [prewarm your application][4], the ApplicationLaunch view instead starts when the RUM iOS SDK is initialized, and the `application_start` event does not have a duration.
```
buraizu, line 41:
```suggestion
In cases where iOS [prewarms your application][4], the ApplicationLaunch view instead starts when the RUM iOS SDK is initialized, and the `application_start` event does not have a duration.
```

buraizu: Looks like the link for `[4]` has not beed added yet

buraizu (changes requested):
Thanks, just made one small suggestion and also noticed a missing link

buraizu (approved):

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
```diff
@@ -207,21 +205,30 @@ internal class RUMSessionScope: RUMScope, RUMContextProvider {
         )
     }
 
-    private func startApplicationLaunchView(on command: RUMCommand, context: DatadogContext) {
+    internal func startApplicationLaunchView(context: DatadogContext, writer: Writer) {
```
ncreated, line 208: suggestion/ So far the RUMScopes architecture was only allowing side-effects that happen in response to `RUMCommand` processing. Everything was implicit and the only internal API on scopes was:
```swift
func process(command: RUMCommand, context: DatadogContext, writer: Writer) -> Bool
```
Here we seem to break this convention and add a precedence: the "start app launch view" routine becomes explicit and can be called from outside, causing session scope to call arbitrary API on view scope. This adds extra complexity to an already complicated RUM architecture - to fully test it now, we'd need to cover all extra usages, like calling `startApplicationLaunchView()` in already running RUM session.

Could we consider adopting this change without introducing new internal APIs in scopes? At the surface, it looks doable, because `RUMSessionScope` has `isInitialSession` flag and `RUMViewScope` has its `isInitialView` attribute. It looks that passing necessary timings and `!= .background` state to scopes `.init()` could be enough.

Alternatively, maybe introducing `RUMStartApplicationLaunchView` command could make RUM code simpler? Such command could be sent once from `RUMMonitor` (right after initialising scopes). It shouldn't require introducing much new pieces to RUM scopes as well. Also, because it will be `RUMMonitor` who emits that command, this won't require such strong surgery in RUM scopes unit tests like we had to do in this PR - the new logic would be additive.

I wonder what you think @maxep @fuzzybinary ?

Tests/DatadogTests/Matchers/RUMEventMatcher.swift
```diff
@@ -6,6 +6,7 @@
 
 import Foundation
 import XCTest
+import Datadog
```
ncreated, line 9: I think we can't let this import here. The comment below says:

> Note: this file is individually referenced by integration tests target, so no dependency on other source files

and IIRC this won't build in UI tests (the UI tests runner doesn't link `Datadog`).

It seems to be a leftover after some other changes, no? I don't see any `Datadog` symbols used below.

ncreated (commented):
It looks okay and I really appreciate the extra care for unit tests and changes in `RUMSessionMatcher` we do here 👌. 

We need to fix at least CI build (in integration tests) to move forward.

On top of that, I left [an architectural suggestion](https://github.com/DataDog/dd-sdk-ios/pull/1160/files#r1105563896). I wonder what you think @fuzzybinary and if it sounds like something that could simplify our implementation.

fuzzybinary:
It's a good point, let's talk this morning and see which approach we like better. I'm leaning toward RUMStartApplicationLaunchView command, as we can have the initial view be not the ApplicationLaunch view.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -149,14 +149,28 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
...
 
         // Apply side effects
         switch command {
+        // Application Launch
+        case is RUMApplicationStartCommand:
+            sendApplicationStartAction(context: context, writer: writer)
+            if !isInitialView || viewPath != RUMOffViewEventsHandlingRule.Constants.applicationLaunchViewURL {
+                DD.telemetry.error(
+                    id: "application-start-error",
+                    message: "An RUMApplciationStartCommand got sent to something other than the ApplicaitonLaunch view.",
```
louiszawadzki, line 165:
```suggestion
                    message: "An RUMApplicationStartCommand got sent to something other than the ApplicationLaunch view.",
```

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -149,14 +149,28 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
...
+        // Application Launch
+        case is RUMApplicationStartCommand:
+            sendApplicationStartAction(context: context, writer: writer)
+            if !isInitialView || viewPath != RUMOffViewEventsHandlingRule.Constants.applicationLaunchViewURL {
+                DD.telemetry.error(
+                    id: "application-start-error",
+                    message: "An RUMApplicationStartCommand got sent to something other than the ApplicationLaunch view.",
+                    kind: nil,
+                    stack: nil
+                )
```
maxep, line 168: /nit You can use other signature available
```suggestion
                DD.telemetry.error("An RUMApplicationStartCommand got sent to something other than the ApplicationLaunch view.")
```
The `id` is inferred using `file:line:message`.

maxep: Maybe we can add the command type as part of the message, I'm afraid `something` is not super actionable :)

maxep (approved):
Looks great! I've left a small suggestion.

ncreated (approved):
Looks great! Thanks for looking into this alternative approach @fuzzybinary 💪. I left last question to make sure we got the requirement right - other than this, good to go 🚀.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
```diff
@@ -37,7 +37,7 @@ internal class RUMApplicationScope: RUMScope, RUMContextProvider {
 
     func process(command: RUMCommand, context: DatadogContext, writer: Writer) -> Bool {
         if sessionScope == nil {
-            startInitialSession(context: context)
+            startInitialSession(on: command, context: context, writer: writer)
```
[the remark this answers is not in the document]
ncreated: Ah, I think I get it - we always do it (start the AL view), it is just that we await for any other command to do it. Makes sense - if the app starts idle and nothing happens, no session will be started 👍.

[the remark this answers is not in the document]
fuzzybinary: That was my thought yes. It also made things a little bit cleaner as I could just use the context / writer as part of the command I was already recieving.

PR #1162, opened 2023-02-13
RUMM-2822 [SR] Set up snapshot tests for session replay

ncreated:
### What and why?

📦 🧪 This PR introduces image-based, snapshot testing workflow for iOS Session Replay.

This will fill the gap in [`NodeRecorders`](https://github.com/DataDog/dd-sdk-ios/tree/1.15.0/DatadogSessionReplay/Sources/DatadogSessionReplay/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders) test coverage that we introduced consciously (due to lack of good testing option).

Single snapshot test consists of 3 elements:
* UI code defined in `SRExample` project (most commonly a `UIViewController` configured in storyboard);
* test definition in `SRSnapshotTests` target;
* snapshot image (recorded in `recordingMode = true` or compared with previous recording).

Adding new test is very straightforward and each test executes fast, which brings the quickest loop possible for working with session replay rendering. A single renderer change can now be previewed under `0.5s` without requiring data upload and viewing replay in the player a minute or two later.

First two tests are added as part of this PR. They assert `UIView` and `UILabel` recording. Each snapshot image puts side-by-side the **actual screenshot of the app** UI and **visual description of wireframes** captured by session replay, e.g.:

testBasicTexts()-maskAll-privacy

### How?

I set up new Xcode project: `SRExample`, located in `DatadogSessionReplay` folder. The project comes with 2 targets: `SRExample` (app) and `SRSnapshotTests` (unit tests bundle). Each test displays certain view controller (`Fixture`) in full screen, then captures the snapshot (side-by-side image) and compares it with previous recording.

The image-comparison logic is made by importing minimal bits from OSS, MIT licensed [pointfreeco/swift-snapshot-testing](https://github.com/pointfreeco/swift-snapshot-testing) project. We use "perceptual precision" algorithm, meaning that two images are considered "the same" even if they slightly differ (in pixels), but the difference cannot be noticed by the human eye.

The app image is rendered by taking the snapshot of the app window (with `UIGraphicsImageRenderer()`). For turning wireframes into image, we depend on a side project I made (OSS, MIT) - [ncreated/Framing](https://github.com/ncreated/Framing/tree/framer). `Framer` is a basic utility that allows drawing simple elements (`Blueprints`) into `UIImage` in a declarative way. It works well with all attributes we use in SR data format (fills, strokes, rounded corners, texts etc.). What gets rendered is an "approximation" of how things will look in the player - enough to assert most critical logic of iOS SR (positioning, colors, view-tree flattening etc.). Snapshot tests **do not replace manual E2E testing with the real player** - with snapshots in place we can just do it less often.

Test example:
```swift
func testBasicShapes() throws {
    show(fixture: .basicShapes)
    let image = try takeSnapshot()
    try compare(image: image, referenceImage: .inFolder(named: snapshotsFolderName), record: recordingMode)
}
```

#### Note: further work required

Because each snapshot will weight `~150-300Kb`, we need to store them as large files to avoid bloating repository size. Initially, I tried using `git lfs` for that purpose, but we found it too intrusive for SPM/CP/CT users. Additional improvements need to be proposed, so this work was detached to `RUMM-3061`. We decided to merge this PR anyway - the `SRSnapshotTests` project remains a valuable add-on to SR development workflow even without asserting images on CI. It still facilitates development of SR elements, a lot.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


DatadogSessionReplay/Sources/Processor/Processor.swift
```diff
@@ -63,6 +69,10 @@ internal class Processor: Processing {
             .compactMap { node in node.semantics.wireframesBuilder }
             .flatMap { nodeBuilder in nodeBuilder.buildWireframes(with: wireframesBuilder) }
 
+        #if DEBUG
+        interceptWireframes?(wireframes)
+        #endif
```
ncreated, line 74: Here we hook into SR implementation to intercept `[SRWireframe]` for creating snapshot images. It could be done in more excellent way by refactoring `Processor` implementation and adding more abstractions - e.g. pipeline-like architecture: `flatten(snapshot) -> nodes` → `map(nodes) -> wireframes` → `export(wireframes)`, and injecting mock implementation of the last phase. Given that current hook is very minimal and non-intrusive for `Release` builds, I'd rather keep it for now and only think of sth more elegant if we find more reason for such refactoring.

DatadogSessionReplay/SRExample/SRSnapshotTests/Utils/ImageComparison.swift
```diff
@@ -0,0 +1,249 @@
...
+///   - record: if `true`, then reference file will be created / overwritten with `image` data
+///   - file: `#filePath` for eventual `XCTest` assertion failure
+///   - line: `#line` for eventual `XCTest` assertion failure
+internal func compare(
+    image: UIImage,
+    referenceImage: ReferenceImage,
+    record: Bool,
+    file: StaticString = #filePath,
+    line: UInt = #line
+) throws {
```
maxep, line 48: I think we should split this in 3 functions:
- Assert Device
- Record Image
- Compare Image

Also, I think we should leverage the `DDAssert` pattern for all assertion. For that, we would need to make `_DDEvaluateAssertion` and `DDAssertError` public, this way all module could use it to define their assertion. WDYT?

DatadogSessionReplay/SRExample/SRSnapshotTests/Utils/ImageComparison.swift
```diff
@@ -0,0 +1,249 @@
...
+import MetalPerformanceShaders
+
+// MARK: - From `swift-snapshot-testing` /Sources/SnapshotTesting/Snapshotting/UIImage.swift
+
+// remap snapshot & reference to same colorspace
+private let imageContextColorSpace = CGColorSpace(name: CGColorSpace.sRGB)
+private let imageContextBitsPerComponent = 8
+private let imageContextBytesPerPixel = 4
+
+func compare(_ old: UIImage, _ new: UIImage, precision: Float, perceptualPrecision: Float) -> String? {
```
maxep, line 99: With `_DDEvaluateAssertion ` public, this could be turned into a `DDAssertImageComparison`.

DatadogSessionReplay/SRExample/SRSnapshotTests/Utils/ImageComparison.swift
```diff
@@ -0,0 +1,249 @@
...
+            space: colorSpace,
+            bitmapInfo: CGImageAlphaInfo.premultipliedLast.rawValue
+        )
+    else { return nil }
+
+    context.draw(cgImage, in: CGRect(x: 0, y: 0, width: cgImage.width, height: cgImage.height))
+    return context
+}
+
+func perceptuallyCompare(_ old: CIImage, _ new: CIImage, pixelPrecision: Float, perceptualPrecision: Float) -> String? {
```
maxep, line 176: Same here: `DDAssertPerceptualInageComparison`, or something like that?

DatadogSessionReplay/Sources/Processor/Processor.swift
maxep: Is `DEBUG` compiler directive is the right one? can't it be `TEST` or a custom one?
not a big fan of test-specific code bit your comment makes total sense 👍

maxep (commented):
This is awesome 🚀  It is very well structured and it's crystal clear, well done!

maxep (commented):
This is awesome 🚀  It is very well structured and crystal clear, well done!

I've left comments on leveraging the `DDAssert*` pattern, so we align assertions everywhere.
Also, I don't think the project (and target) should be called `DatadogSessionReplay/SRExample`, but rather something like `DatadogSessionReplay/SnapshotTests`, as it's not an example project for users to look at.
The application target could be called like `Runner` or `Host` to better align with other test-suites (Integration, E2E) in v2. It's only naming but I think it will facilitate the navigation in the repo :)

DatadogSessionReplay/Sources/Processor/Processor.swift
ncreated: Anything other than default (`Debug` | `Release`) would be hard to configure. The SR is linked to `SRExample` as SPM package, so that flag would have to be defined in `Package.swift` AFAIK 😞. I'd rather search for architecture-level solution (more abstraction like explained ☝️), but IMO this is too early for it. WDYT?

DatadogSessionReplay/SRExample/SRSnapshotTests/Utils/ImageComparison.swift
ncreated: I like these ideas 👍 , let me look into this it this PR 👀

ncreated:
> Also, I don't think the project (and target) should be called `DatadogSessionReplay/SRExample`, but rather something like `DatadogSessionReplay/SnapshotTests`, as it's not an example project for users to look at. The application target could be called like `Runner` or `Host` to better align with other test-suites (Integration, E2E) in v2. It's only naming but I think it will facilitate the navigation in the repo :)

This makes great sense 👍. Indeed I followed old pattern of "example" project, but here we could make it better. Project renaming is not easy nor fun, but worth doing it now. I will do it in this PR 🙌.

 .gitattributes
```diff
@@ -1 +0,0 @@
-CHANGELOG.md merge=union
```
ncreated, line 1: I'm actually moving this file - there was a space in the name 😮 (`mv ' .gitattributes' .gitattributes`) which was confusing the `git lfs` setup.

DatadogSessionReplay/SRExample/SRSnapshotTests/Utils/ImageComparison.swift
ncreated: For that one, I didn't wrap it into separate `DDAssert*`. We don't use this method independently - it's part of `DDAssertSnapshotEquals(...)` logic 👍.

TestUtilities/Helpers/DDAssert.swift
```diff
@@ -10,8 +10,11 @@
 import Foundation
 import XCTest
 
-private enum DDAssertError: Error {
-    case expectedFailure(String, keyPath: [String] = [])
+public enum DDAssertError: Error {
+    /// An error indicating known coding failure that happened in evaluated `expression()` block.
+    case expectedCodingFailure(String, keyPath: [String])
```
maxep, line 15: It's not necessary a coding failure, is it? we also provide a keyPath for reflection and dictionary comparison. 
The previous `expectedFailure` with optional keyPath already covered both cases IMO. But maybe it wasn't good for auto-completion?

maxep (approved):
Very nice work here 👏 👏 
I've left a minor comment but it's good to merge 🚀

TestUtilities/Helpers/DDAssert.swift
ncreated: I reverted this change 👍. My concern was that this concept doesn't scale well if we want to provide other detail than "no" or "`keyPath: [String] = []`" (we'll have to introduce another associated value with default). However, we don't need it for now, so reverting makes sense - we can revisit it later.

bitrise.yml
```diff
@@ -249,7 +249,19 @@ workflows:
...
+    #     title: Run snapshot tests for Session Replay - iOS Simulator
+    #     run_if: '{{enveq "DD_RUN_SR_UNIT_TESTS" "1"}}'
+    #     inputs:
+    #     - scheme: SRSnapshotTests
+    #     - destination: platform=iOS Simulator,name=iPhone 14,OS=16.2
+    #     - should_build_before_test: 'no'
+    #     - is_clean_build: 'no'
+    #     - generate_code_coverage_files: 'yes'
+    #     - project_path: DatadogSessionReplay/SRSnapshotTests/SRSnapshotTests.xcworkspace
+    #     - xcpretty_test_options: --color --report html --output "${BITRISE_DEPLOY_DIR}/DatadogSessionReplay-snapshot-tests.html"
```
ncreated, line 264: As discussed on Slack, we will keep it disabled until we solve the problem of storing large files in a sustainable way. A separate task was created for this `RUMM-3061`.

PR #1165, opened 2023-02-14
RUMM-2939 Create Integration Tests Project

maxep:
### What and why?

Move `DatadogIntegrationTests` target to its own project under `IntegrationScenarios`.
This migration is part of the v2 effort as described in [RFC - iOS Project Structure](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/2637693679/RFC+-+iOS+Project+Structure)

This project currently use CocoaPods as a dependency manager, this is necessary so we can have the following dependency tree:
```
IntegrationTests
│
├─ Datadog Package
│   └─ TestUtilities // it includes but does not expose a TestUtilities target
│
└─ TestUtilities Package
    └─ TestUtilities // xcode complains about duplicated targets
```

This doesn't seems to be supported when using xcodeproj with dependency to local packages. Therefore, the `IntegrationTests` folder looks like the following:

image

#### TestUtilities

The `TestUtilities.podspec` is used for local dependency with the IntegrationTests. It could be used in other test-suites such as Benchmark and E2E. It is not meant to be published.

#### Shared Code

Sources are still shared between the main project and the `IntegrationTests`:
- `Tests/DatadogTests/Matchers`. It cannot be shared under `TestUtilities` as it depends on models. 
- `Sources/Datadog/RUM/DataModels/RUMDataModels.swift`

This is tech-debt that needs to be addressed in the future.

#### Runner Configuration

The `AppConfiguration` has been removed from the `Example` and move the `Runner`. The Example app is now configured in the app delegate, while the `Runner` is configured per test scenario.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [x] Run smoke tests


tools/ci/choose_workflows.py
```diff
@@ -57,7 +57,7 @@ def should_run_integration_tests(ctx: CIContext) -> bool:
         pr_keyword='[x] Run integration tests',
         pr_path_prefixes=[
             'Datadog/Example/',
-            'Tests/DatadogIntegrationTests/',
+            'IntegrationTests/IntegrationTests/',
             'Tests/DatadogBenchmarkTests/',
```
ncreated, line 61: 🙌  It no longer makes sense to run integration tests on code changes in `Datadog/Example` or `Tests/DatadogBenchmarkTests/`.

IntegrationTests/Runner/Utils/ConsoleOutputInterceptor.swift
```diff
@@ -0,0 +1,65 @@
...
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+#if DEBUG
+
+import UIKit
+import DatadogInternal
+import Datadog
+
+class ConsoleOutputInterceptor {
```
ncreated, line 13: Similar, AFAIK we don't use it in integration scenarios. Did a quick lookup - no reference. We don't need to copy this file 👍.

IntegrationTests/Runner/Environment.swift
```diff
@@ -0,0 +1,144 @@
...
+
+internal struct Environment {
+    /// ENV variables shared between UITests and Example targets.
+    struct Variable {
+        static let testScenarioClassName = "DD_TEST_SCENARIO_CLASS_NAME"
+        static let serverMockConfiguration = "DD_TEST_SERVER_MOCK_CONFIGURATION"
+    }
+    /// Launch arguments shared between UITests and Example targets.
+    struct Argument {
+        static let isRunningUnitTests       = "IS_RUNNING_UNIT_TESTS"
```
ncreated, line 45: Some more cleanup can be done here around `isRunningUnitTests` - unlike this new Integration Tests project, the previous `Example` project was also launched in unit tests. With this constant we were preventing some app logic to run in that scenario. We no longer need it.

ncreated (approved):
It looks great 💪. Massive work done here and everything works 💯 (including CI Visibility integration 🏅).

I left few minor feedbacks - mostly, some pieces do not need to be copied into Integration Tests project. Nothing blocking, tho it would be nice to clean it up at some point 👍.

tools/ci/choose_workflows.py
maxep: Indeed, good catch!

IntegrationTests/Runner/Utils/ConsoleOutputInterceptor.swift
maxep: 🧹

IntegrationTests/Runner/Environment.swift
maxep: I would like to keep it for now. The app could be run with Unit tests with the addition of IntegrationUnits, so this check could become handy :)

PR #1166, opened 2023-02-14
RUMM-2965 [SR] Support `UISlider` elements in session replay

ncreated:
### What and why?

📦 With this PR we're adding support to `UISlider` elements in session replay.

testSliders()-maskAll-privacy

### How?

Regular use of SR framework:
* added `UISliderRecorder` with `UISliderWireframesBuilder`
* added "Sliders" fixture in snapshot tests project.

The slider is constructed with 3 or 4 wireframes:
* 2 wireframes for "min track" and "max track"
* 1 wireframe for the "thumb" element
* 1 optional wireframe if the slider's `UIView` defines visible background

This is the first element that produces more than 2 wireframes, so I had to update the logic of generating `NodeID` (wireframe IDs) for views. Now, the `NodeIDGenerator` can create variadic number of IDs (instead of one or two before).

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/DatadogSessionReplay/Sources/Utilities/CGRectExtensions.swift
+++ b/DatadogSessionReplay/Sources/Utilities/CGRectExtensions.swift
@@ -1,0 +1,40 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+internal extension CGRect {
+    enum HorizontalAlignment {
+        case left, right, center
+    }
+
+    enum VerticalAlignment {
+        case top, bottom, middle
+    }
+
+    /// Puts this rect inside other rect with given alignment.
+    func putInside(
+        _ other: CGRect,
+        horizontalAlignment: HorizontalAlignment,
+        verticalAlignment: VerticalAlignment
+    ) -> CGRect {
+        var new = self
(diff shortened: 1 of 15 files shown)


DatadogSessionReplay/Sources/Utilities/CGRectExtensions.swift
```diff
@@ -0,0 +1,40 @@
...
+    enum HorizontalAlignment {
+        case left, right, center
+    }
+
+    enum VerticalAlignment {
+        case top, bottom, middle
+    }
+
+    /// Puts this rect inside other rect with given alignment.
+    func putInside(
```
fuzzybinary, line 19: This might be better called `alignWith`, since when aligned left or right this rectangle actually ends up outside the bounds of the rect it will be "inside".

fuzzybinary (approved):

DatadogSessionReplay/Sources/Utilities/CGRectExtensions.swift
maxep: Nice ascii art @ncreated 🎨

maxep (approved):
Looks great!

DatadogSessionReplay/Sources/Utilities/CGRectExtensions.swift
fuzzybinary: Ah, you're right! For some reason I thought `.origin` was the center, not the bottom left corner.

PR #1171, opened 2023-02-21
RUMM-2939 Create Integration Units Target

maxep:
### What and why?

// TBD

### How?

// TBD

### Review checklist
- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes

### Custom CI job configuration (optional)
- [ ] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

PR #1172, opened 2023-02-21
Remove `*.podspec.src`

maxep:
### What and why?

Remove `*.podpsec.src` for each module to simplify CocoaPods definitions.

### How?

Use `find` and `sed` to replace the podspecs version

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


fuzzybinary (approved):

DatadogSDKCrashReporting.podspec.src
```diff
@@ -1,27 +0,0 @@
...
-
-  s.swift_version      = '5.1'
-  s.ios.deployment_target = '11.0'
-  s.tvos.deployment_target = '11.0'
-
-  s.source = { :git => 'https://github.com/DataDog/dd-sdk-ios.git', :tag => s.version.to_s }
-  s.static_framework = true
-
-  s.source_files = "Sources/DatadogCrashReporting/**/*.swift"
-  s.dependency 'DatadogSDK', '__DATADOG_VERSION__'
```
ncreated, line 25: We need to change this line in the original `.podspec` in order to make it work. The solution we're replacing was depending on `__DATADOG_VERSION__` token replacement, while the new one modifies `s.version      = "__DATADOG_VERSION__"`.

If we change it to:
```
  s.dependency 'DatadogSDK', s.version.to_s
```
in the original `podspec`, it should be fine.

DatadogSDKObjc.podspec.src
```diff
@@ -1,25 +0,0 @@
...
-  }
-
-  s.swift_version      = '5.1'
-  s.ios.deployment_target = '11.0'
-  s.tvos.deployment_target = '11.0'
-
-  s.source = { :git => 'https://github.com/DataDog/dd-sdk-ios.git', :tag => s.version.to_s }
-
-  s.source_files = "Sources/DatadogObjc/**/*.swift"
-  s.dependency 'DatadogSDK', '__DATADOG_VERSION__'
```
ncreated, line 24: Similar to above, now we need:
```
  s.dependency 'DatadogSDK', s.version.to_s
```
in `DatadogSDKObjc.podspec`.

scripts/podspec_bump_version.rb
```diff
@@ -0,0 +1,21 @@
...
+if !version
+    raise "Usage: podspec_bump_version.rb <version number>"
+end
+
+podspecs = Dir.glob('*.podspec')
+
+for podspec in podspecs do
+    spec = Pod::Specification.from_file(podspec)
+    text = File.read(podspec)
+    text.gsub!(/(s.version( )*= ")#{spec.version}(")/, "\\1#{version}\\3")
```
ncreated, line 19: After all, we just do string replacement (similar to previous `sed///g`), so I wonder if it's worth introducing Ruby complexity for this PR. What do we benefit from `spec = Pod::Specification.from_file(podspec)` 🤔?

Wouldn't simple regex for matching:
```
  s.version      = "([a-z][A-Z][0-9]\.\-)*"
```
be enough?

Separate thing is that we need `Bundler` anyway as we introduced `Cocoapods` dedpendency in #1171 👍

ncreated (changes requested):

DatadogSDKCrashReporting.podspec.src
maxep: oh damn, that right!

scripts/podspec_bump_version.rb
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env ruby
```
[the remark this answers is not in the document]
maxep: I've moved it to `tools/`, it's easier to flatten this directory a bit so it's easier to find what's available.

scripts/podspec_bump_version.rb
maxep: That was a good suggestion!! I was able to do the same with a shell script using `find` and `sed`!
I've left the Gemfile and Bundler setup so we have it available, it's no longer related to this PR but as you mention we need cocoapods as a dependency in v2

ncreated (approved):
Looks good 👌

PR #1175, opened 2023-02-23
RUMM-2964 [SR] Make view-tree recording more customizable

ncreated:
### What and why?

📦 This PR adds more flexibility to the way SR traverses and records subtree hierarchies.

This change will allow each `NodeRecorder` (e.g. `UIViewRecorder` or `UILabelRecorder`) defining their own strategies and will be soon leveraged in `UIPickerViewRecorder` (and later in `UIDatePickerViewRecorder`).

### How?

Before this change, only two strategies existed, denoted by `recordSubtree: Bool` flag. For a given hierarchy, either all subviews were recorded or all could be ignored:

Screenshot 2023-02-23 at 19 18 40

In result, the recorder was visiting following nodes in DFS order:
```
ROOT, A, AA, AB, ABA, ABB, B, C
```

With this PR, we're adding one more strategy, so changing `recordSubtree` boolean into an `enum`:
```swift
internal enum NodeSubtreeStrategy {
    case record // equivalent of `recordSubtree: true`
    case ignore // equivalent of `recordSubtree: false`
    case replace(subtreeNodes: [Node]) // new one ⭐️
}
```
By using the new `.replace(subtreeNodes:)` strategy, certain node recorders (e.g. `UIPickerViewRecorder`) will be able to opt-out from their subtree traversal and instead replace the subtree with its curated `[Node]` information:

Screenshot 2023-02-23 at 19 18 45

```
ROOT, A, AA, AB, ABA, ABB, B, C, CV1, CV2, ..., CVN
```

This will enable code reuse for certain view-tree recorders, meaning:
* the root view-tree recorder could be configured with all node recorders (`UIViewRecorder`, `UILabelRecorder`, `UISwitchRecorder`, `UISliderRecorder`, ...) - as we can expect any node in the root view's hierarchy;
* the `UIViewPickerRecorder` could opt-in only `UILabelRecorder` and `UIViewRecorder` - as we expect only labels and shapes in picker's subtree;
* similar, date-picker recorder would expect texts + shapes, whereas text-field recorder could additionally search for images (to record icons).

Last not least, this required switching from tree-like nodes structure:
```swift
struct ViewTreeSnapshot {
   // ...

   let root: Node
}

struct Node {
   // ...

   let childNodes: [Node]
}
```
to flat array given by DFS-visit order:
```swift
struct ViewTreeSnapshot {
   // ...

   /// Nodes in DFS order.
   let nodes: [Node]
}

struct Node {
   // ...
}
```
The change of data structure should have no impact on the SR performance - nodes were already visited in DFS order, here we just add them to an array while visiting (instead of allocating a graph).

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/DatadogSessionReplay/Sources/Processor/Flattening/NodesFlattener.swift
+++ b/DatadogSessionReplay/Sources/Processor/Flattening/NodesFlattener.swift
@@ -7,17 +7,18 @@
 import Foundation
 import CoreGraphics
 
-/// Flattens VTS received from `Recorder` by transforming its tree-structure of nodes into array of nodes.
+/// Flattens VTS received from `Recorder` by removing invisible nodes.
 ///
-/// Flattening includes removal of nodes that are invisible because of being covered by other nodes displayed
-/// closer to the screen surface.
+/// Nodes are invisible if:
+/// - they have no appearance (e.g. nodes denoting container views)
+/// - they are covered by another opaque nodes (displayed closer to the screen surface).
 internal struct NodesFlattener {
     /// This current implementation is greedy and works in `O(n*log(n))`, wheares `O(n)` is possible.
     /// TODO: RUMM-2461 Improve flattening performance.
     func flattenNodes(in snapshot: ViewTreeSnapshot) -> [Node] {
         var flattened: [Node] = []
 
-        dfsVisit(startingFrom: snapshot.root) { nextNode in
+        for nextNode in snapshot.nodes {
             // Skip invisible nodes:
             if !(nextNode.semantics is InvisibleElement) {
                 // When accepting nodes, remove ones that are covered by another opaque node:
@@ -38,12 +39,5 @@ internal struct NodesFlattener {
         }
 
         return flattened
-    }
-
-    private func dfsVisit(startingFrom node: Node, visit: (Node) -> Void) {
-        visit(node)
-        node.children.forEach { child in
-            dfsVisit(startingFrom: child, visit: visit)
-        }
     }
 }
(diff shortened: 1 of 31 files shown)


maxep (commented):
Great idea, and testing is on point!

What I don't understand tho is how you will create the `.replace` strategy when recording 🤔 The `UIPickerViewRecorder` will need to depend on a `UILabelRecorder` for creating the nodes?

maxep:
Yes, I see! Then I'm concerned about having 2 logics for recording, one 'kind' of recorder won't depend on other recorders and will just delegate the tree traversing to its caller, they will return the root node semantics only. The other kind will depend on its own set of recorders and will provide the root node semantics and its children nodes.

If we generalise, we could say that a recorder is responsible for traversing the view tree, it could use all default recorders (equivalent to strategy `.record`), a subset of recorders (~ strategy `.replace`), or none (~ strategy `.ignore`). It would all come down to the recorder, and this would be enforced by a simple design:

```swift
internal protocol NodeRecorder {

    /// - Returns: The full list of nodes representing the view and subviews.
    func nodes(of view: UIView, with attributes: ViewAttributes, in context: ViewTreeRecordingContext) -> [Node]
}
```

Recorders would use helper functions to traverse the subviews recursively, and they will decide what nodes to return, they will use the provided attributes and semantics they found.

It's a bigger refactor I admit, so I left it as a suggestion but I believe it will scale better as we add more and more recorders. LMWYT!

maxep (approved):

ncreated:
> ```swift
> internal protocol NodeRecorder {
> 
>     /// - Returns: The full list of nodes representing the view and subviews.
>     func nodes(of view: UIView, with attributes: ViewAttributes, in context: ViewTreeRecordingContext) -> [Node]
> }
> ```

As we discussed on the meeting - it is an interesting proposal and a bit aligned with current design 👍. We may want to look into it later - for now we think it is best to cover more use cases with current architecture before considering a shift. We're still learning the complexity of views hierarchy and discovering more arch drivers.

PR #1180, opened 2023-02-27
REPLAY-1340 Support `UISwitch` elements in session replay

ncreated:
### What and why?

📦 This PR enhances the recording of `UISwitch` elements in session replay. Its preliminary support was introduced earlier, but here we solve remaining glitches and adjust the values properly for light and dark mode.

<table>
<tr><td>With no masking:</td><td>With masking:</td></tr>
<tr>
   <td></td>
   <td></td>
</tr>
<tr>
   <td></td>
   <td></td>
</tr>
</table>

### How?

A major bug 🐞 fixed in this PR corresponds to recording `UISwitch` views that are positioned inside [`UIStackView`](https://developer.apple.com/documentation/uikit/uistackview). Turned out that if rendered in stack view, the `view.frame` includes the size of the entire stack row - which was stretching the switch wireframe unreastically:

Screenshot 2023-02-27 at 19 20 50

This problem is solved by using [`intrinsicContentSize `](https://developer.apple.com/documentation/uikit/uiview/1622600-intrinsiccontentsize) for reading the size of element.

Other than this:
- adjusted color values to leverage recently introduced `SystemColors` construct;
- added optional background wireframe to render `switch.backgroundColor` if any is applied.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/DatadogSessionReplay/SRSnapshotTests/SRHost/Fixtures/Fixtures.swift
+++ b/DatadogSessionReplay/SRSnapshotTests/SRHost/Fixtures/Fixtures.swift
@@ -12,6 +12,7 @@ internal enum Fixture: CaseIterable {
     case sliders
     case segments
     case pickers
+    case switches
 
     var menuItemTitle: String {
         switch self {
@@ -25,6 +26,8 @@ internal enum Fixture: CaseIterable {
             return "Segments"
         case .pickers:
             return "Pickers"
+        case .switches:
+            return "Switches"
         }
     }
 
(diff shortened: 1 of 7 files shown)


ncreated:
> I only miss one bit. Shouldn't we commit the expected snapshot?

Thanks for review! We will host snapshots in `REPLAY-1371: Solve problem of storing large (png) files in SR snapshot tests`. Currently all snapshots have 2.7MB in size, which is more than 30% of the repository size (counting all changes, it would be over 50% now). We don't want to store them in git (unless we find LFS doable).

PR #1182, opened 2023-03-01
RUMM-3081 fix invalid cpu ticks per seconds

xgouchet:
### What and why?

When the view lifespan is shorter than a second, the estimation of ticks per second would go through the roof. 
We limit reporting it for views with a time spent greater than one second.


maciejburda (approved):
LGTM

PR #1184, opened 2023-03-01
Rename Feature Interfaces

maxep:
### What and why?

Update public interface for Feature registration as proposed in **RFC - SDK v2 Public API**.

### How?

```swift
/// A Datadog Feature that can interact with the core through the message-bus.
public protocol DatadogFeature {
    /// The feature name.
    static var name: String { get }

    /// The message bus receiver.
    ///
    /// The `FeatureMessageReceiver` defines an interface for Feature to receive any message
    /// from a bus that is shared between Features registered in a core.
    var messageReceiver: FeatureMessageReceiver { get }
}


/// A Datadog Feature with remote data store.
public protocol DatadogRemoteFeature: DatadogFeature {
    /// The URL request builder for uploading data.
    ///
    /// The `FeatureRequestBuilder` defines an interface for building a single `URLRequest`
    /// for a list of data events and the current core context.
    ///
    /// A Feature should use this interface for creating requests that needs be sent to its Datadog Intake.
    /// The request will be transported by `DatadogCore`.
    var requestBuilder: FeatureRequestBuilder { get }
}

```

The Feature name is now a static var allowing a simplified definition and usage of the `DatadogCoreProtocol`.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/DatadogInternal/Sources/DatadogCoreProtocol.swift
+++ b/DatadogInternal/Sources/DatadogCoreProtocol.swift
@@ -16,11 +16,11 @@ public var defaultDatadogCore: DatadogCoreProtocol = NOPDatadogCore()
 public protocol DatadogCoreProtocol: AnyObject {
     /// Registers a Feature instance.
     ///
-    /// Feature collects and transfers data to a Datadog Product (e.g. Logs, RUM, ...). Upon registration, the Feature can
-    /// retrieve a `FeatureScope` interface for writing events to the core. The core will store and upload events efficiently
-    /// according to the performance presets defined on initialization.
-    ///
-    /// A Feature can also communicate to other Features by sending messages on the message bus managed by core.
+    /// Feature can interact with the core and other Feature through the message bus. Some specific Features
+    /// complying to `DatadogRemoteFeature` can collects and transfers data to a Datadog Product
+    /// (e.g. Logs, RUM, ...). Upon registration, a Remote Feature can retrieve a `FeatureScope` interface
+    /// for writing events to the core. The core will store and upload events efficiently according to the performance
+    /// presets defined on initialization.
     ///
     /// - Parameter feature: The Feature instance - it will be retained and held by core.
     func register<T>(feature: T) throws where T: DatadogFeature
--- a/DatadogInternal/Sources/DatadogCoreProtocol.swift
+++ b/DatadogInternal/Sources/DatadogCoreProtocol.swift
@@ -17,7 +17,7 @@ public protocol DatadogCoreProtocol: AnyObject {
     /// Registers a Feature instance.
     ///
     /// Feature can interact with the core and other Feature through the message bus. Some specific Features
-    /// complying to `DatadogRemoteFeature` can collects and transfers data to a Datadog Product
+    /// complying to `DatadogRemoteFeature` can collect and transfer data to a Datadog Product
     /// (e.g. Logs, RUM, ...). Upon registration, a Remote Feature can retrieve a `FeatureScope` interface
     /// for writing events to the core. The core will store and upload events efficiently according to the performance
     /// presets defined on initialization.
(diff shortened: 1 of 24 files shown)


Sources/Datadog/DatadogCore/DatadogCore.swift
```diff
@@ -193,7 +193,7 @@ internal final class DatadogCore {
     /// A list of storage units of currently registered Features.
     private var allStorages: [FeatureStorage] {
         let v1Storages = v1Features.values.compactMap { $0 as? V1Feature }.map { $0.storage }
```
ncreated, line 195: question/ Do we still have V1 Features 🤔💭?

Tests/DatadogTests/Datadog/DatadogCore/DatadogCoreTests.swift
```diff
@@ -9,12 +9,13 @@ import TestUtilities
 import DatadogInternal
 @testable import Datadog
 
-private struct FeatureMock: DatadogFeature {
+private struct FeatureMock: DatadogProduct {
```
ncreated, line 12: nit/ Quite hard to figure out this lingo - `FeatureMock` that is a `DatadogProduct` in fact (whereas `DatadogFeature` is sth else).

ncreated (approved):
Looks good 👍, although what concerns me a bit is lack of conseqnence in using `*Feature` vs `*Product` in naming. We seem to use them interchangeably in SDK:
```swift
struct DatadogLogsFeature: DatadogProduct // a Product or a Feature?
```
and in tests:
```swift
struct FeatureMock: DatadogProduct // a Product or a Feature?
```
Also, because this is an effect of refactoring, some API comments in `DatadogCoreProtocol` are quite inaccurate now. None of above is blocking, but it's IMO something to be aware of in `DatadogInternal` module.

Sources/Datadog/DatadogCore/DatadogCore.swift
maxep: Yes, currently Tracing and RUM are still v1, but not for long ;)

maxep:
> what concerns me a bit is lack of conseqnence in using *Feature vs *Product in naming. 

Totally, I should fix that 👍

Tests/DatadogTests/Datadog/DatadogCore/DatadogCoreTests.swift
maxep: I will rename `DatadogProduct` to `DatadogRemoteFeature`, which is more representative of its capability. 'Product' is even more confusing,

DatadogInternal/Sources/DatadogCoreProtocol.swift
```diff
@@ -16,11 +16,11 @@ public var defaultDatadogCore: DatadogCoreProtocol = NOPDatadogCore()
 public protocol DatadogCoreProtocol: AnyObject {
     /// Registers a Feature instance.
     ///
-    /// Feature collects and transfers data to a Datadog Product (e.g. Logs, RUM, ...). Upon registration, the Feature can
-    /// retrieve a `FeatureScope` interface for writing events to the core. The core will store and upload events efficiently
-    /// according to the performance presets defined on initialization.
-    ///
-    /// A Feature can also communicate to other Features by sending messages on the message bus managed by core.
+    /// Feature can interact with the core and other Feature through the message bus. Some specific Features
+    /// complying to `DatadogRemoteFeature` can collects and transfers data to a Datadog Product
```
ncreated, line 20:
```suggestion
    /// complying to `DatadogRemoteFeature` can collect and transfer data to a Datadog Product
```

ncreated (approved):
`DatadogRemoteFeature: DatadogFeature` makes sense 👍, good idea. Good to go 🙂.

PR #1189, opened 2023-03-02
Fix memory usage issues in image capture

maciejburda:
### What and why?

Image capture algorithm started to behave funky. It's memory consumption went through the roof.
Screenshot 2023-03-02 at 10 07 50

### How?

Reasons:
- md5 for jpegs generate different values (probably apple's [pngData()](https://developer.apple.com/documentation/uikit/uiimage/1624096-pngdata) is optimised in a way where it doesn't always produce the same bitmap).
- for such image it was running into recursion loading the data each time
- since there was no autorelease pool - memory was never freed in such a loop

Fixes:
- Added cache's max cost (in bytes)
- Removed md5 hash for the time being (we can revisit when needed)
- Added autorelease pool on the `contentBase64String()` to make sure it always frees the memory (even when run in a loop or recursively).

After these changes we are back to stable:
Screenshot 2023-03-02 at 10 06 21

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/DatadogSessionReplay/Tests/Utilities/CacheTests.swift
+++ b/DatadogSessionReplay/Tests/Utilities/CacheTests.swift
@@ -40,6 +40,17 @@ class CacheTests: XCTestCase {
         XCTAssertNil(cache.value(forKey: "one"))
     }
 
+    func test_bytesLimit() {
+        let cache = Cache<String, Int>(dateProvider: {
+            return Date(timeIntervalSinceReferenceDate: 0)
+        }, totalBytesLimit: 1)
+
+        cache.insert(1, forKey: "one", size: 1)
+        cache.insert(1, forKey: "two", size: 1)
+        XCTAssertNil(cache.value(forKey: "one"))
+        XCTAssertNotNil(cache.value(forKey: "two"))
+    }
+
     func test_countLimit() {
         let cache = Cache<String, Int>(maximumEntryCount: 1)
 
--- a/DatadogSessionReplay/Sources/Recorder/Utilities/ImageDataProvider.swift
+++ b/DatadogSessionReplay/Sources/Recorder/Utilities/ImageDataProvider.swift
@@ -39,7 +39,10 @@ internal class ImageDataProvider {
                 image = image.withTintColor(tintColor)
             }
 
-            let identifier = "\(image.hash)\(String(describing: tintColor?.hash))"
+            var identifier = image.srIdentifier
+            if let tintColorIdentifier = tintColor?.srIdentifier {
+                identifier += tintColorIdentifier
+            }
             let dataLoadingStaus = cache[identifier]
             switch dataLoadingStaus {
             case .none:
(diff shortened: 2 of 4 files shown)


DatadogSessionReplay/Tests/Utilities/CacheTests.swift
```diff
@@ -40,6 +40,17 @@ class CacheTests: XCTestCase {
...
 
+    func test_bytesLimit() {
+        let cache = Cache<String, Int>(dateProvider: {
+            return Date(timeIntervalSinceReferenceDate: 0)
+        }, totalBytesLimit: 1)
+
+        cache.insert(1, forKey: "one", size: 1)
+        cache.insert(1, forKey: "two", size: 1)
+        XCTAssertNil(cache.value(forKey: "one"))
+        XCTAssertNotNil(cache.value(forKey: "two"))
```
maciejburda, line 51: By design it removes oldest entry after reaching total bytes limit

fuzzybinary (approved):

ncreated (changes requested):
Nice investigation 👌. However, it seems that this problem have been unnoticed because we lacked appropriate tests coverage. Now we fixed the problem, but the coverage is still not there. What would prevent us from falling back into this trap again? If the problem was diagnosed as "for some identical images their cache identifiers are not equal", this should be expressed in tests.

Also, it is great that we add more safety measures 💪 (cap on the cache size) and their corresponding test, but I have strong concerns on addressing the root cause by removing MD5 logic 🤔. Why exactly was it there in the first place and what are we losing now if it gets removed?

DatadogSessionReplay/Tests/Utilities/CacheTests.swift
ncreated: 💡😃 
```suggestion
        cache.insert(1, forKey: "one", size: 1)
        cache.insert(1, forKey: "two", size: 1)
        XCTAssertNil(cache.value(forKey: "one"), "It should remove oldest entry after reaching total bytes limit")
        XCTAssertNotNil(cache.value(forKey: "two"))
```

maciejburda:
Great comment @ncreated 🙌 

Initial implementation was relying on NSObject hash (as we discovered that UIImages are smartly allocated in the memory), but at some point we switched to md5 as it had potential to be more futureproof (works with assets pre-upload for instance).

I think we can forget about more elaborate hashing for the time being and only introduce it when it useful in the future. Now, test coverage for future hashing mechanism should cover scenario where we load the same image resource multiple time and we hit consistent hash for each iteration.

ncreated (approved):
> I think we can forget about more elaborate hashing for the time being and only introduce it when we find it useful. Now, test coverage for future hashing mechanism should cover scenario where we load the same image resource multiple times and we hit consistent hash for each iteration.

Makes sense 👍 - thanks for full context @maciejburda. Agree, let's bring more complex hashing up again when we need it.

PR #1194, opened 2023-03-07
REPLAY-1448 Add support for UIStepper

maciejburda:
### What and why?

📦 This PR enhances the recording of `UIStepper` elements in session replay. Its preliminary support was introduced earlier, but here we solve remaining glitches and send frames instead of images for button + and - button icons.

### How?

By adding new recorder for `UIStepper` element that checks it's frame and creates matching wireframes.

testSteppers()-allowAll-privacy
testSteppers()-allowAll-privacy

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/DatadogSessionReplay/SRSnapshotTests/SRHost/Fixtures/Fixtures.swift
+++ b/DatadogSessionReplay/SRSnapshotTests/SRHost/Fixtures/Fixtures.swift
@@ -14,6 +14,7 @@ internal enum Fixture: CaseIterable {
     case pickers
     case switches
     case textFields
+    case steppers
 
     var menuItemTitle: String {
         switch self {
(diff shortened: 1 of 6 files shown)


maciejburda:
One thing to note. Pixel perfection is not there - due to center calculation it's sometimes slightly offset. I tried mitigating with using text shape, but the results weren't as nice for font related reasons. The other solution is to use + and - images that are embedded within the control, but this means sending all the image data.

ncreated (approved):
Nice 🚀 - I left 2 feedbacks to consider.

> One thing to note. Pixel perfection is not there - due to center calculation it's sometimes slightly offset. I tried mitigating with using text shape, but the results weren't as nice for font related reasons. The other solution is to use + and - images that are embedded within the control, but this means sending all the image data.

We seek for an "approximated" look that will fulfil product requirements and what we did here is more than enough 👍. Only difference between what we do now vs we had considered is 5 vs 3-4 wireframes. This is IMO negligible (even when scrolling, sending updates to 5 wireframes is not much worse than updating 3 of them). Well done 💪.

PR #1197, opened 2023-03-08
Remove span baggage queue

maxep:
### What and why?

Use lock on `BaggageItems` for thread safety to simplify usage of our various `TracePropagationHeadersExtractor` implementation. This will ease usage of shared trace propagation in the v2 modular architecture.

### How?

Replace the Span `BaggageItems` synchronisation queue with a read-write lock.

### Review checklist

- [x] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/Tracer.swift
+++ b/Sources/Datadog/Tracer.swift
@@ -175,13 +175,11 @@ public class Tracer: OTTracer {
     }
 
     public func extract(reader: OTFormatReader) -> OTSpanContext? {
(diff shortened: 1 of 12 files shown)


maxep:
@ncreated yes I'm aware of that 👍  well queue can't completely prevent race conditions (e.g invoking `queue.[a]sync{}` simultaneously form different threads is one). But indeed, the order of execution is more predictable on a queue. Even tho `pthread_rwlock_t` is a fair lock, it's not a fifo of reads and writes. Given the nature of span's baggage I think it's safe to use a lock here 👍

PR #1200, opened 2023-03-08
REPLAY-1347 Improve image classification

maciejburda:
### What and why?

Improves image classification to get only contextual images into account.

### How?

It's accomplished by simplification of `ImageDataProvider` and moving some of it's responsibility to `UIImageViewRecorder` which enables reusability of this recorder.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/DatadogSessionReplay/Sources/Recorder/Utilities/ImageDataProvider.swift
+++ b/DatadogSessionReplay/Sources/Recorder/Utilities/ImageDataProvider.swift
@@ -18,7 +18,7 @@ internal protocol ImageDataProviding {
     ) -> String
 }
 
-final internal class ImageDataProvider: ImageDataProviding {
+internal final class ImageDataProvider: ImageDataProviding {
     private var cache: Cache<String, String>
 
     private let maxBytesSize: Int


DatadogSessionReplay/Sources/Recorder/Utilities/ImageDataProvider.swift
```diff
@@ -74,6 +60,36 @@ extension UIImage {
     }
 }
 
+fileprivate extension UIImage {
+    func compressToTargetSize(_ targetSize: Int) -> Data {
```
maciejburda, line 64: improve

DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UIImageViewWireframesBuilderTests.swift
```diff
@@ -0,0 +1,94 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import XCTest
+@testable import DatadogSessionReplay
+
+class UIImageViewWireframesBuilderTests: XCTestCase {
```
maciejburda, line 10: Coverage here could be increased. Maybe we could discuss the plan for testing more elaborate wireframes builders.

DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UIImageViewWireframesBuilderTests.swift
```diff
@@ -0,0 +1,94 @@
...
+        if case let .imageWireframe(imageWireframe) = wireframes[1] {
+            XCTAssertEqual(imageWireframe.id, imageWireframeID)
+            XCTAssertEqual(imageWireframe.base64, "")
+        } else {
+            XCTFail("Second wireframe needs to be imageWireframe case")
+        }
+    }
+}
+
+class MockImageDataProvider: ImageDataProvider {
```
maciejburda, line 90: we should probably move it somewhere more suitable

ncreated (changes requested):
It looks OK 👌 , but causes regression compared to `develop` so we need to improve it. Some images result with base64 values that exceed the SDK core's max _event size_. In result we do not send replay.

On preventing such regression in the future - I think we need to consider adding integration tests of the same fashion as we have in RUM SDK (sending SR payload to HTTP server and asserting its delivery). Let's discuss it on daily.

DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UIImageViewWireframesBuilderTests.swift
maciejburda: I like that. I'll keep these until we have snapshot in place. I'll describe the work in the ticket.

ncreated (approved):
Looks good 👌 💪

PR #1209, opened 2023-03-14
[RUMM-2997] Send a crash to both RUM and Logging features

plousada:
### What and why?

Sends crash information both to Logs and RUM Features, as opposed to only RUM if both features are enabled.
Adopted a logging strategy similar to Android.  

### How?

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Tests/DatadogTests/Datadog/CrashReporting/CrashReporterTests.swift
+++ b/Tests/DatadogTests/Datadog/CrashReporting/CrashReporterTests.swift
@@ -46,13 +46,13 @@ class CrashReporterTests: XCTestCase {
         XCTAssertTrue(plugin.hasPurgedCrashReport == true, "It should ask to purge the crash report")
     }
 
-    func testWhenPendingCrashReportIsFound_itIsSentBothToRumAndLogs() throws {
-        let expectation = self.expectation(description: "`CrashReportSender` sends the crash report to both features")
+    func testWhenPendingCrashReportIsFound_itIsSentToRumFeature() throws {
+        let expectation = self.expectation(description: "`CrashReportSender` sends the crash report to RUM feature")
         let crashContext: CrashContext = .mockRandom()
         let crashReport: DDCrashReport = .mockRandomWith(context: crashContext)
-        let crashMessageReceiver = CrashMessageReceiverMock()
+        let rumCrashReceiver = RUMCrashReceiverMock()
 
-        let core = PassthroughCoreMock(messageReceiver: crashMessageReceiver)
+        let core = PassthroughCoreMock(messageReceiver: rumCrashReceiver)
 
         let plugin = CrashReportingPluginMock()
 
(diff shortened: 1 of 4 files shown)


Tests/DatadogTests/Datadog/CrashReporting/CrashReporterTests.swift
```diff
@@ -12,7 +12,7 @@ class CrashReporterTests: XCTestCase {
     // MARK: - Sending Crash Report
 
     func testWhenPendingCrashReportIsFound_itIsSentAndPurged() throws {
-        let expectation = self.expectation(description: "`LoggingOrRUMsender` sends the crash report")
+        let expectation = self.expectation(description: "`LoggingAndRUMsender` sends the crash report")
```
ncreated, line 15: nit/ it seems to be an artifact of V2 refactoring - neither `LoggingOrRUMsender` nor `LoggingAndRUMsender` do exist and instead we use `CrashReportSender`.
```suggestion
        let expectation = self.expectation(description: "`CrashReportSender` sends the crash report")
```

Tests/DatadogTests/Datadog/CrashReporting/CrashReporterTests.swift
```diff
@@ -46,6 +46,47 @@ class CrashReporterTests: XCTestCase {
         XCTAssertTrue(plugin.hasPurgedCrashReport == true, "It should ask to purge the crash report")
     }
 
+    func testWhenPendingCrashReportIsFound_itIsSentBothToRumAndLogs() throws {
+        let expectation = self.expectation(description: "`LoggingAndRUMsender` sends the crash report to both features")
```
ncreated, line 50:
```suggestion
        let expectation = self.expectation(description: "`CrashReportSender` sends the crash report to both features")
```

Tests/DatadogTests/Datadog/CrashReporting/CrashReporterTests.swift
```diff
@@ -74,7 +115,7 @@ class CrashReporterTests: XCTestCase {
     }
 
     func testWhenPendingCrashReportIsFoundButItHasUnavailableCrashContext_itPurgesTheCrashReportWithNoSending() {
-        let expectation = self.expectation(description: "`LoggingOrRUMsender` does not send the crash report")
+        let expectation = self.expectation(description: "`LoggingAndRUMsender` does not send the crash report")
```
ncreated, line 118:
```suggestion
        let expectation = self.expectation(description: "`CrashReportSender` does not send the crash report")
```

ncreated (approved):
Looks great 🎯👌 - I left minor feedback, but for me it's good to go, anyway.

maxep (commented):
Great work 👍  another approach would be to send a single "crash" message on the bus, and make the Logs Feature consume that one instead of "crash-log". It would simplify a bit the implementation, but it would have to print a more generic warning if the crash is not consumed at all.

plousada:
> Great work 👍 another approach would be to send a single "crash" message on the bus, and make the Logs Feature consume that one instead of "crash-log". It would simplify a bit the implementation, but it would have to print a more generic warning if the crash is not consumed at all.

I can also do that if you prefer. It depends if we want to give importance to the fact that the crash is consumed by a specific feature, or by any feature. 
It's a bit easier to debug a crash not appearing in Logs/RUM with the current implementation as the message is feature specific. But it will generate a noisy warning each time a crash is treated if the developer doesn't want to activate one of the features.

maxep:
Yes, I am for simplicity :) I think we would benefit from sending the crash only once on the bus, mostly because it will save an encode/decode operation on the message-bus and it could be consume by any Feature. Also, this warning won't have much visibility as it will be printed when the app restart only.

maxep (approved):
Nice, thank you!

PR #1210, opened 2023-03-14
URLSession Auto Instrumentation

maxep:
### What and why?

Enable URLSession auto-instrumentation in v2.

Usage of the instrumentation in the Trace module: #1214 

### How?

All necessary code for URLSession instrumentation is now part of `DatadogInternal` module. This instrumentation is not yet used by RUM or Tracing, but it includes all necessary logic.

#### Network Instrumentation Feature

This `internal` Feature is capable of intercepting URL requests and tasks using Swizzling and the custom `DatadogURLSessionDelegate`, it will forward any interception to registered interceptors that will be able to inject propagation headers and record any event when a URL task starts or completes. The behaviour is similar to the v1 implementation with the difference that we can have multiple interceptor registered.

#### URLSession swizzling

URLSession swizzling is now static and enabled by calling `bind()`. Binding will increment a counter of instrumentation, when calling `unbing()`, the counter is decremented, when reaching 0 swizzling is disabled.

#### URLSession Delegate

`DatadogURLSessionDelegate` will replace the `DDURLSessionDelegate`

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionSwizzler.swift
```diff
@@ -0,0 +1,273 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+internal class URLSessionSwizzler {
```
maxep, line 9: This swizzler differs from the one currently implemented in v1, it only expose static accessors.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskInterception.swift
```diff
@@ -0,0 +1,231 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+public class URLSessionTaskInterception {
```
maxep, line 9: Almost identical to the v1 `URLSessionTaskInterception`, only the span context property has been replaced by a `trace` ids.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/DatadogURLSessionDelegate.swift
```diff
@@ -0,0 +1,93 @@
...
+ */
+
+import Foundation
+
+/// The `URLSession` delegate object which enables network requests instrumentation. **It must be
+/// used together with** `DatadogRUM` or `DatadogTrace`.
+///
+/// All requests made with the `URLSession` instrumented with this delegate will be intercepted by the SDK.
+@objc
+open class DatadogURLSessionDelegate: NSObject, URLSessionDataDelegate {
```
maxep, line 14: Almost identical to the v1 `DDURLSessionDelegate `, it will redirect tasks to the `NetworkInstrumentationFeature` in the provided `core`.

DatadogInternal/Sources/Swizzling/MethodSwizzler.swift
```diff
@@ -5,10 +5,9 @@
  */
 
 import Foundation
-import DatadogInternal
 
-internal class MethodSwizzler<TypedIMP, TypedBlockIMP> {
-    struct FoundMethod: Hashable {
+open class MethodSwizzler<TypedIMP, TypedBlockIMP> {
```
maxep, line 9: This is now exposed in `DatadogInternal` and can be used in SR

maxep (commented):
@maciejburda @ncreated sorry again for the size of this PR. I'm doing my best to open PRs by chunks but some still get big :/
The auto-instrumentation of URLSession with trace propagation in both Tracing and RUM are very tough to decouple..

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
```diff
@@ -0,0 +1,143 @@
...
+            }
+        }
+
+        let interception = URLSessionTaskInterception(
+            request: request,
+            isFirstParty: firstPartyHosts(for: session)
+                .isFirstParty(url: request.url)
+        )
+
+        interceptions[task] = interception
```
ncreated, line 79: Here we alter shared state from user-initiated `URLSession.dataTask(...)`. This can be called on any (user) thread, so it will crash 🔥 if not synchronising `interceptions[:]`.

Such behaviours are typically asserted in V1. Migrating this test should make it covered:
https://github.com/DataDog/dd-sdk-ios/blob/bdd0cef7e50b2fc478583365781f357cad963c17/Tests/DatadogTests/Datadog/URLSessionAutoInstrumentation/Interception/URLSessionInterceptorTests.swift#L515

Datadog/Datadog.xcodeproj/xcshareddata/xcschemes/DatadogInternal iOS.xcscheme
```diff
@@ -26,7 +26,18 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-      shouldUseLaunchSchemeArgsEnv = "YES">
+      shouldUseLaunchSchemeArgsEnv = "YES"
+      codeCoverageEnabled = "YES"
+      onlyGenerateCoverageForSpecifiedTargets = "YES">
+      <CodeCoverageTargets>
```
maciejburda, line 32: Curious: what's the impact of this change?

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionSwizzler.swift
```diff
@@ -0,0 +1,273 @@
...
+    /// Counts of bindings of the URL swizzler.
+    ///
+    /// This value will increment for each call to the `bind()` method.
+    /// Calling `unbind()` will decrement the count, when reaching zero, the swizzler is disabled.
+    internal private(set) static var bindingsCount: Int = 0
```
ncreated, line 14: suggestion/ Including it in test integrity checks will ensure we don't go into flakiness because of missing or not balancing `unbind()` well in some tests:  https://github.com/DataDog/dd-sdk-ios/blob/bdd0cef7e50b2fc478583365781f357cad963c17/Tests/DatadogTests/TestsObserver/DatadogTestsObserver.swift#L23

maciejburda:
I don't mind the size of the PR. It's hard to isolate work like this!

Just to reiterate:
> This instrumentation is not yet used by RUM or Tracing, but it includes all necessary logic.

So basically this PR copies the code to a destination module, but it doesn't switch the implementation yet?

TestUtilities/Mocks/SingleFeatureCoreMock.swift
```diff
@@ -0,0 +1,78 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+import XCTest
+import DatadogInternal
+
+/// Single-Feature core mocks feature-scope for a single `Feature` type.
+public final class SingleFeatureCoreMock<Feature>: DatadogCoreProtocol, FeatureScope where Feature: DatadogFeature {
```
ncreated, line 12: nit/ Not very clear when to use this instead of `PassthroughCoreMock`. As we're adding this into shared `TestUtilities`, IMO we should improve this comment.

ncreated (changes requested):
Looks OK on the surface 👍. Left few feedbacks with one blocking (lack of synchronization in interceptor).

General question: what is the strategy towards removing old instrumentation code? The `URLSessionSwizzler` and other classes are now duplicated. Will we delete them after all products are switched to new network instrumentation stack?

Datadog/Datadog.xcodeproj/xcshareddata/xcschemes/DatadogInternal iOS.xcscheme
maxep: not intended, good catch 👍

maxep:
> General question: what is the strategy towards removing old instrumentation code? The URLSessionSwizzler and other classes are now duplicated. Will we delete them after all products are switched to new network instrumentation stack?

Yes, they will be removed when RUM module uses the new instrumentation. I'm trying to do incremental changes here, mostly to facilitate the review process.

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
maxep: Of course, my bad!

maxep:
> So basically this PR copies the code to a destination module, but it doesn't switch the implementation yet?

It doesn't switch the implementation yet 👍 It's not only copied, it's also slightly modified for further use. Usage in the Trace module is applied in next PRs.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionSwizzler.swift
maxep: It is already checked with:
```swift
.init(
    assert: { activeSwizzlingNames.isEmpty },
    problem: "No swizzling must be applied.",
    solution: """
    Make sure all applied swizzling are reset by the end of test with `unswizzle()`.

    `DatadogTestsObserver` found \(activeSwizzlingNames.count) leaked swizzlings:
    \(activeSwizzlingNames.joined(separator: ", "))
    """
),
```

TestUtilities/Mocks/SingleFeatureCoreMock.swift
maxep: Indeed, I'm also making this a subclass of `PassthroughCoreMock`.

ncreated (changes requested):
Recent changes look good 👍, but I found one flag in new swizzler implementation. I left a question on dropping `__URLSessionDelegateProviding` in there.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionSwizzler.swift
```diff
@@ -0,0 +1,273 @@
...
+            self.method = try Self.findMethod(with: selector, in: klass)
+            super.init()
+        }
+
+        func swizzle() {
+            typealias Signature = @convention(block) (URLSession, URLRequest, CompletionHandler?) -> URLSessionDataTask
+            swizzle(method) { previousImplementation -> Signature in
+                return { session, request, completionHandler -> URLSessionDataTask in
+                    guard
+                        let delegate = session.delegate as? DatadogURLSessionDelegate,
```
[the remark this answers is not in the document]
maxep: Damn, I missed that, good catch!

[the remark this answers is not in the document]
maxep: I've added it as a duplicate while doing the migration.

PR #1214, opened 2023-03-17
RUMM-2617 Trace URLSession Interceptor

maxep:
### What and why?

Replace v1 `URLSessionTracingHandler` with v2 `TracingURLSessionHandler`

### How?

The `Trace` Feature now provide a `DatadogURLSessionHandler` implementation for receiving `URLSessionTaskInterception`.

#### Integration Tests

The `TestScenario` now support a `configureFeatures` for configuring Features for specific scenarios.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


DatadogInternal/Sources/MessageBus/FeatureMessageReceiver.swift
```diff
@@ -59,6 +59,6 @@ public struct CombinedFeatureMessageReceiver: FeatureMessageReceiver {
     ///   - core: An instance of the core from which the message is transmitted.
     /// - Returns: `true` if the message was processed by one of the receiver; `false` if it was ignored.
     public func receive(message: FeatureMessage, from core: DatadogCoreProtocol) -> Bool {
-        receivers.contains(where: { $0.receive(message: message, from: core) })
+        receivers.reduce(false) { $0 || $1.receive(message: message, from: core) }
```
maxep, line 62: Previously, the `CombinedFeatureMessageReceiver` would stop forwarding the message after one of its receivers consumed a message. Now, it will send the message to all receivers and it returns `true` if at least one consumes it.

Sources/Datadog/Tracing/Feature/MessageReceivers.swift
```diff
@@ -7,9 +7,28 @@
...
+    /// The RUM attributes that should be added as Span tags.
+    ///
+    /// These attributes are synchronized using a read-write lock.
+    var rum: [String: String]?
+
+    /// Provides the history of app foreground / background states.
+    var applicationStateHistory: AppStateHistory?
+}
+
+internal final class ContextMessageReceiver: FeatureMessageReceiver {
```
maxep, line 20: The `ContextMessageReceiver` replace the `TracingMessageReceiver`. It will store RUM context and Application State History received from the bus.

The RUM context will be used for span as before.
The Application State History is used for setting span tags during URLSessionTask interception.

Sources/Datadog/Tracing/Feature/MessageReceivers.swift
```diff
@@ -7,9 +7,28 @@
...
+    let bundleWithRUM: Bool
+
+    /// The RUM attributes that should be added as Span tags.
+    ///
+    /// These attributes are synchronized using a read-write lock.
+    @ReadWriteLock
+    var context: CoreContext = .init()
```
ncreated, line 27: This comment looks odd - it's copy pasted from `context.rum` (?).

Sources/Datadog/Tracing/Integrations/TracingURLSessionHandler.swift
```diff
@@ -0,0 +1,182 @@
...
+                request.setValue(value, forHTTPHeaderField: field)
+            }
+        }
+
+        return request
+    }
+
+    func isInternal(request: URLRequest) -> Bool {
+        return false
+    }
```
ncreated, line 88: I thought we will remove it - we validated that "internal URLs filtering" is no longer required. Is it a leftover?

Sources/Datadog/Tracing/TracerConfiguration.swift
```diff
@@ -41,13 +45,89 @@ extension DatadogTracer {
...
             self.sendNetworkInfo = sendNetworkInfo
             self.bundleWithRUM = bundleWithRUM
             self.globalTags = globalTags
             self.customIntakeURL = customIntakeURL
+            self.spanEventMapper = spanEventMapper
+        }
+    }
+
+    /// Datadog Tracer configuration.
+    public struct DistributedTracingConfiguration {
```
ncreated, line 61: The comment looks odd / wrong (?).

Tests/DatadogTests/Datadog/Tracing/ContextMessageReceiverTests.swift
```diff
@@ -0,0 +1,63 @@
...
+        )
+
+        XCTAssertEqual(receiver.context.rum, ["key": "value1"])
+
+        // When
+        core.set(feature: "rum", attributes: { ["ids": ["key": "value2"]] })
+
+        // Then
+        XCTAssertEqual(receiver.context.rum, ["key": "value2"])
+    }
```
ncreated, line 45: question/ Shouldn't we define such tests as Integration Units? Here it makes very little sense - the contract between RUM and Tracing that we mock here is arbitrary and false (RUM context is more than `rum.ids[key] = value1`). We don't test much more than basic V2's `MessageReceiver` logic here.

DatadogInternal/Sources/MessageBus/FeatureMessageReceiver.swift
ncreated: blocking/ Changing this behaviour here should require change in [`FeatureMessageReceiverTests`](https://github.com/DataDog/dd-sdk-ios/blob/feature/v2/DatadogInternal/Tests/MessageBus/FeatureMessageReceiverTests.swift#L58-L70), but we don't do it in this PR 🤔💭. Looking at following test (that didn't require update on this branch), something is wrong - either the test makes bad assertion or we're not altering the logic the way it's described here.

https://github.com/DataDog/dd-sdk-ios/blob/ed77be78dcb516843927e961bdbfa001de9bebaf/DatadogInternal/Tests/MessageBus/FeatureMessageReceiverTests.swift#L58-L70

ncreated (changes requested):
Nice effort 👌. It makes sense in overall.

Left few questions/remarks. Blocking until we find out if there's anything wrong in `FeatureMessageReceiverTests.swift`.

Tests/DatadogTests/Datadog/Tracing/ContextMessageReceiverTests.swift
maxep: This test is more to check the `ContextMessageReceiver(bundleWithRUM: true)` vs. `ContextMessageReceiver(bundleWithRUM: false)`, and assert that `bundleWithRUM` enables/disables the rum context propagation. So we are testing the receiver only in unit test.

Now, I agree that we should be testing the integration RUM+Trace. It was previously covered, but we have a regression that will be addressed in [RUMM-2843](https://datadoghq.atlassian.net/browse/RUMM-2843). We will re-enable that coverage while fixing it. How does that sound?

ncreated (approved):

PR #1218, opened 2023-03-21
REPLAY-1460 Enhance `UIStepper` element masking

ncreated:
### What and why?

📦 This PR enhances masking behaviour for `UIStepper` elements when `.maskAll` policy is set. Now, the MIN and MAX values cannot be inferred as we no longer grey-out their buttons:

### How?

Easy 😎.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


maxep (approved):

maciejburda (approved):
🎖️ 

Thanks for tackling this! I really love how you decided to solve it.
I was initially thinking to show a grey square, but it makes a lot of sense to do it like this.

PR #1219, opened 2023-03-21
Allow users to manually end a RUM session

fuzzybinary:
### What and why?

This adds a method to manually end a RUM session with `stopRumSession`.  This immediately stops the active view and sets the current session to inactive. A new session starts automatically when a user action is sent (`startView` or `addUserAction`).

### How?

First, may RUM events now have a new property: `session.isActive`. This can be null, and is only currently set in RUMViewEvents. The state of this member is pulled from the parent session context.

Second, I've modified `RUMApplicationScope` to support holding multiple `RUMSessionScope`s in a similar way to how Sessions can handle multiple Views until all of their events have been processed.

When a `RUMStopSession` event is received, the Application Scope forwards it to the Session, which marks itself as inactive, then forwards the event to the all views, which mark themselves as inactive and send a ViewUpdate event which captures the session now being inactive.

Provided views have finished processing all of their user actions and resources, the session is removed from available sessions on the ApplicationScope. Otherwise, it sicks around (marked as inactive) until all events are complete.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [x] Run smoke tests

### DIFF
--- a/Datadog/Example/Scenarios/RUM/StopSessionScenario/KioskSendEventsViewController.swift
+++ b/Datadog/Example/Scenarios/RUM/StopSessionScenario/KioskSendEventsViewController.swift
@@ -26,10 +26,10 @@ internal class KioskSendEventsViewController: UIViewController {
         rumMonitor.stopView(viewController: self)
     }
 
-    @IBAction func didTapDownloadResourceButton(_ sender: Any) {
+    @IBAction func didTapDownloadResourceButton(_ sender: UIButton) {
         rumMonitor.addUserAction(
             type: .tap,
-            name: (sender as! UIButton).currentTitle!,
+            name: sender.currentTitle!,
             attributes: ["button.description": String(describing: sender)]
         )
 
--- a/Datadog/Example/Scenarios/RUM/StopSessionScenario/KioskSendEventsViewController.swift
+++ b/Datadog/Example/Scenarios/RUM/StopSessionScenario/KioskSendEventsViewController.swift
@@ -1,0 +1,60 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2023-Present Datadog, Inc.
+
+import UIKit
+
+internal class KioskSendEventsViewController: UIViewController {
+    @IBOutlet private var doneButton: UIButton!
+
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        doneButton.isHidden = true
+    }
+
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+
+        rumMonitor.startView(viewController: self, name: "KioskSendEvents")
+    }
+
+    override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+
+        rumMonitor.stopView(viewController: self)
+    }
+
+    @IBAction func didTapDownloadResourceButton(_ sender: Any) {
+        rumMonitor.addUserAction(
+            type: .tap,
+            name: (sender as! UIButton).currentTitle!,
+            attributes: ["button.description": String(describing: sender)]
+        )
+
+        let simulatedResourceKey = "/resource/1"
+        let simulatedResourceRequest = URLRequest(url: URL(string: "https://foo.com/resource/1")!)
+        let simulatedResourceLoadingTime: TimeInterval = 0.1
+
+        rumMonitor.startResourceLoading(
+            resourceKey: simulatedResourceKey,
+            request: simulatedResourceRequest
+        )
+
+        DispatchQueue.main.asyncAfter(deadline: .now() + simulatedResourceLoadingTime) {
+            rumMonitor.stopResourceLoading(
+                resourceKey: simulatedResourceKey,
+                response: HTTPURLResponse(
+                    url: simulatedResourceRequest.url!,
+                    statusCode: 200,
+                    httpVersion: nil,
+                    headerFields: ["Content-Type": "image/png"]
+                )!
+            )
+
+            // Reveal "Done" so UITest can continue
+            self.doneButton.isHidden = false
+        }
+    }
+}
--- a/Sources/Datadog/RUM/RUMMonitor/RUMCommand.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/RUMCommand.swift
@@ -29,8 +29,8 @@ internal struct RUMApplicationStartCommand: RUMCommand {
 internal struct RUMStopSessionCommand: RUMCommand {
     var time: Date
     var attributes: [AttributeKey: AttributeValue] = [:]
-    var canStartBackgroundView = false // no, stopping a session should not start a backgorund session
-    var isUserInteraction = false
+    let canStartBackgroundView = false // no, stopping a session should not start a backgorund session
+    let isUserInteraction = false
 
     init(time: Date) {
         self.time = time
--- a/Sources/Datadog/RUM/RUMMonitor/RUMCommand.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/RUMCommand.swift
@@ -26,6 +26,17 @@ internal struct RUMApplicationStartCommand: RUMCommand {
     var isUserInteraction = false
 }
 
+internal struct RUMStopSessionCommand: RUMCommand {
+    var time: Date
+    var attributes: [AttributeKey: AttributeValue] = [:]
+    var canStartBackgroundView = false // no, stopping a session should not start a backgorund session
+    var isUserInteraction = false
+
+    init(time: Date) {
+        self.time = time
+    }
+}
+
 // MARK: - RUM View related commands
 
 internal struct RUMStartViewCommand: RUMCommand {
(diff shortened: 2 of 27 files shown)


Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
```diff
@@ -36,31 +48,55 @@ internal class RUMApplicationScope: RUMScope, RUMContextProvider {
...
-        if let currentSession = sessionScope {
-            sessionScope = sessionScope?.scope(byPropagating: command, context: context, writer: writer)
+        if activeSession == nil && command.isUserInteraction {
+            // No active sessions, start a new one
+            startNewSession(on: command, context: context, writer: writer)
+        }
+
+        if command is RUMStopSessionCommand {
+            // Reach in and grab the last active view
+            lastActiveView = activeSession?.viewScopes.first(where: { $0.isActiveView } )
```
fuzzybinary, line 62: I'm interested in thoughts on this approach. This feels like it's pulling in some extra dependencies in the ViewScope, but I couldn't find a good way to make activating the previous SessionScope's view work, even with the existing refresh logic.

Datadog/Example/Scenarios/RUM/StopSessionScenario/KioskSendEventsViewController.swift
```diff
@@ -0,0 +1,60 @@
...
+        rumMonitor.startView(viewController: self, name: "KioskSendEvents")
+    }
+
+    override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+
+        rumMonitor.stopView(viewController: self)
+    }
+
+    @IBAction func didTapDownloadResourceButton(_ sender: Any) {
```
maciejburda, line 29: If I'm not mistaken you can change the sender type to `UIButton` and avoid force cast

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
```diff
@@ -36,31 +48,55 @@ internal class RUMApplicationScope: RUMScope, RUMContextProvider {
...
+        sessionScopes = sessionScopes.compactMap({ scope in
+            if scope.process(command: command, context: context, writer: writer) {
+                // Returned true, keep the scope around, it still has work to do.
+                return scope
+            }
 
-            if sessionScope == nil { // if session expired
-                refresh(expiredSession: currentSession, on: command, context: context, writer: writer)
+            if scope.isActive {
+                // False, but still active means we timed out or expired, refresh the session
```
maciejburda, line 74: From the comment it feels like `isSessionActive` should be checked here as well?

I don't see this param used elsewhere.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
```diff
@@ -36,31 +48,55 @@ internal class RUMApplicationScope: RUMScope, RUMContextProvider {
...
+                return refresh(expiredSession: scope, on: command, context: context, writer: writer)
             }
+            // Else, inactive and done processing events, remove
+            return nil
+        })
+
+        // Sanity telemety, only end up with one active session
```
maciejburda, line 81: Nice! 💪

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
```diff
@@ -135,51 +158,63 @@ internal class RUMSessionScope: RUMScope, RUMContextProvider {
         }
 
         if !isSampled {
-            return true // discard all events in this session
+            // Make sure sessions end even if they are sampled
```
maciejburda, line 161: nice catch!

Sources/Datadog/RUMMonitor.swift
```diff
@@ -629,12 +635,11 @@ public class RUMMonitor: DDRUMMonitor, RUMCommandSubscriber {
         // update the core context with rum context
         core.set(feature: "rum", attributes: {
             self.queue.sync {
-                let context = self.applicationScope.sessionScope?.viewScopes.last?.context ??
-                                self.applicationScope.sessionScope?.context ??
-                                self.applicationScope.context
```
maciejburda, line 634: I think this odd formatting is more common for `??`, but if linter is not complaining I think it's fine

Sources/Datadog/RUMMonitor.swift
```diff
@@ -629,12 +635,11 @@ public class RUMMonitor: DDRUMMonitor, RUMCommandSubscriber {
...
+                let context = self.applicationScope.activeSession?.viewScopes.last?.context ??
+                              self.applicationScope.activeSession?.context ??
+                              self.applicationScope.context
 
                 guard context.sessionID != .nullUUID else {
-                    // if Session was sampled or not yet started
```
maciejburda, line 637: Is that not true anymore?

Tests/DatadogIntegrationTests/Scenarios/RUM/RUMCommonAsserts.swift
```diff
@@ -71,7 +71,12 @@ extension RUMSessionMatcher {
         let eventMatchers = try requests
             .flatMap { request in try RUMEventMatcher.fromNewlineSeparatedJSONObjectsData(request.httpBody, eventsPatch: eventsPatch) }
             .filter { event in try event.eventType() != "telemetry" }
-        let sessionMatchers = try RUMSessionMatcher.groupMatchersBySessions(eventMatchers)
+        var sessionMatchers = try RUMSessionMatcher.groupMatchersBySessions(eventMatchers)
+        sessionMatchers.sort {
+            let a = $0.viewVisits.first?.viewEvents.first?.date ?? 0
+            let b = $1.viewVisits.first?.viewEvents.first?.date ?? 0
+            return a < b
+        }
```
maciejburda, line 79: We can maintain the `let` using `sorted(by:)`

```suggestion
        let sessionMatchers = try RUMSessionMatcher.groupMatchersBySessions(eventMatchers).sorted(by: {
            return $0.viewVisits.first?.viewEvents.first?.date ?? 0 < $1.viewVisits.first?.viewEvents.first?.date ?? 0
        })
```

maciejburda (approved):
Great Effort!

Left a couple of minor suggestions 🙌

Sources/Datadog/RUMMonitor.swift
fuzzybinary: TBH I think the linter is what changed it.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMApplicationScope.swift
maciejburda: I don't get what `False` is referring to here, since we are inside of `isActive`, which is true

fuzzybinary: Ah, It's the return value of `scope.process` If it returned `true` return the scope, False, we check isActive. I'll change the comment to make it clearer.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
```diff
@@ -135,51 +158,63 @@ internal class RUMSessionScope: RUMScope, RUMContextProvider {
...
                         \(String(describing: command)) was detected, but no view is active. To track views automatically, try calling the
                         DatadogConfiguration.Builder.trackUIKitRUMViews() method. You can also track views manually using
                         the RumMonitor.startView() and RumMonitor.stopView() methods.
                         """
-                    )
+                        )
+                    }
                 }
             }
         }
```
maxep, line 203: The conditions and state management are difficult to follow and error prone IMO. It might be helpful to split this into multiple methods, would that be possible?

Sources/Datadog/RUM/RUMMonitor/RUMCommand.swift
```diff
@@ -26,6 +26,17 @@ internal struct RUMApplicationStartCommand: RUMCommand {
     var isUserInteraction = false
 }
 
+internal struct RUMStopSessionCommand: RUMCommand {
+    var time: Date
+    var attributes: [AttributeKey: AttributeValue] = [:]
+    var canStartBackgroundView = false // no, stopping a session should not start a backgorund session
+    var isUserInteraction = false
```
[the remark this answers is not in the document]
fuzzybinary: They're `var` getters in the `RUMCommand` protocol. Is it possible to change them to let in the implementation?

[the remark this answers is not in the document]
maxep: Yes, if they are `get` only, they can be `let 👍

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMSessionScope.swift
fuzzybinary: I moved the "no active view" portion to its own function. Let me know if that helps or it you think it needs more.

maxep: Perfect, thank you!

maxep (approved):
LGTM 🚀

PR #1224, opened 2023-03-24
REPLAY-1459 Session replay custom limits

maciejburda:
### What and why?

Adds ability to use custom limit for `maxFileSize`, `maxObjectSize` for session replay.
This will be replace by a proper mechanism once we migrate fully to V2.

Together with the update we added semantics for size in bytes and time which were used in the `PerformancePresets`. 

### How?

Solution boils down to adding additional optional parameter to feature protocol called `PerformancePresetOverride` which contains all the overridable fields.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


Sources/Datadog/Core/PerformancePreset.swift
```diff
@@ -113,16 +113,42 @@ internal extension PerformancePreset {
...
+            uploadDelayChangeRate: uploadDelayChangeRate
+        )
+    }
+}
+
+public struct PerformancePresetOverride {
+    let maxFileSize: UInt64?
+    let maxObjectSize: UInt64?
+
+    public init(maxFileSize: UInt64?, maxObjectSize: UInt64?) {
```
ncreated, line 150: Unless we want to make it public API for `Datadog`, it should be moved to `Sources/Datadog/DatadogInternal/**`. Because it is part of Feature <> Core contract (not User <> SDK), I think `DatadogInternal` is the right place.

Also, because it's public API (even though will be used only by us), it must include doc string and comments.

Sources/Datadog/Utils/SwiftExtensions.swift
```diff
@@ -101,3 +101,41 @@ extension Array {
...
     }
 }
+
+// MARK: - Time Convinience
+
+public extension TimeInterval {
+    var seconds: TimeInterval { return TimeInterval(self) }
+    var minutes: TimeInterval { return self.multiplyOrClamp(by: 60) }
+    var hours: TimeInterval { return self.multiplyOrClamp(by: 60.minutes) }
+    var days: TimeInterval { return self.multiplyOrClamp(by: 24.hours) }
```
ncreated, line 111: Similar, this should be part of `Sources/Datadog/DatadogInternal/**` so we don't pollute user namespaces with our utility code (now, by importing `Datadog` they will see all these helpers in V2).

Sources/Datadog/DatadogCore/DatadogCore.swift
```diff
@@ -277,12 +277,14 @@ extension DatadogCore: DatadogCoreProtocol {
...
+
         let storage = FeatureStorage(
             featureName: feature.name,
             queue: readWriteQueue,
             directories: featureDirectories,
             dateProvider: dateProvider,
-            performance: performance,
+            performance: updatedPerformance,
             encryption: encryption
         )
```
ncreated, line 289: I think there are no tests for this logic. Let's add them 🙌.

DatadogSessionReplay/Tests/Utilities/SwiftExtensionsTests.swift
```diff
@@ -44,6 +44,13 @@ class FixedWidthIntegerTests: XCTestCase {
         let expectedConvertedValue = Int.min
         XCTAssertEqual(expectedConvertedValue, convertedValue)
     }
+
+    func testBytesSizeConvenience() {
+        let size = 1
+        XCTAssertEqual(size.KB, size * 1_024)
+        XCTAssertEqual(size.MB, size * 1_024 * 1_024)
+        XCTAssertEqual(size.GB, size * 1_024 * 1_024 * 1_024)
+    }
```
ncreated, line 53: We add these extensions as part of `Sources/Datadog/Utils/SwiftExtensions.swift`, but this test file relates to Session Replay's extensions. It needs to be moved 😉.

DatadogSessionReplay/Tests/Utilities/SwiftExtensionsTests.swift
```diff
@@ -65,4 +72,11 @@ class TimeIntervalTests: XCTestCase {
         let uInt64MaxDate = Date(timeIntervalSinceReferenceDate: -.greatestFiniteMagnitude)
         XCTAssertEqual(uInt64MaxDate.timeIntervalSince1970.toInt64Milliseconds, Int64.min)
     }
+
+    func testTimeConvienience() {
+        let time = Date.mockDecember15th2019At10AMUTC().timeIntervalSince1970
+        XCTAssertEqual(time.minutes, time * 60)
+        XCTAssertEqual(time.hours, time * 60 * 60)
+        XCTAssertEqual(time.days, time * 60 * 60 * 24)
+    }
```
ncreated, line 81: Similar, we add these extensions as part of `Sources/Datadog/Utils/SwiftExtensions.swift`, but this test file relates to Session Replay's extensions. It needs to be moved.

Also, the underlying logic of this conversion distinguishes few edge cases that must be covered in tests.

ncreated (changes requested):
The idea looks good 👍, but I left few blocking comments.

As we're modifying public and V2 interfaces here: /cc @maxep. Because it changes V2 for SR: /cc @mariusc83 @0xnm

maxep (commented):
Thanks @maciejburda, it makes total sense to have presets override for Features 👍 
I agree with @ncreated's suggestions, I've added few nits as well.

Tests/DatadogTests/Datadog/Core/PerformancePresetTests.swift
```diff
@@ -137,4 +137,11 @@ class PerformancePresetTests: XCTestCase {
             )
         }
     }
+
+    func testPresetsUpdate() {
+        let preset = PerformancePreset(batchSize: .mockRandom(), uploadFrequency: .mockRandom(), bundleType: .mockRandom())
+        let updatedPreset = preset.updated(with: PerformancePresetOverride(maxFileSize: 0, maxObjectSize: 0))
```
ncreated, line 143: nit/ If we use `.mockRandom(min:max:)` (or `.mockRandom(min:max:otherThan:)` instead of hardcoding `0`-es this test would become much better. Instead of testing single value, it would be covering all possible configurations.

Tests/DatadogTests/Datadog/DatadogCore/DatadogCoreTests.swift
```diff
@@ -187,4 +188,50 @@ class DatadogCoreTests: XCTestCase {
...
+        )
+        try core.register(
+            feature: FeatureMock(
+                name: "mock",
+                performanceOverride: nil
+            )
+        )
+        let feature = core.v2Features.values.first
+        XCTAssertEqual(feature?.storage.authorizedFilesOrchestrator.performance.maxObjectSize, UInt64(512).KB)
+        XCTAssertEqual(feature?.storage.authorizedFilesOrchestrator.performance.maxFileSize, UInt64(4).MB)
```
ncreated, line 212: nit/ These numbers (`512 KB` + `4 MB`) look magical in this context and they are defined [quite far](https://github.com/DataDog/dd-sdk-ios/blob/5a657d8ed96ff49834c08fda3b673932eb80df92/Sources/Datadog/Core/PerformancePreset.swift#L116-L126) from this test implementation. We already test default performance values [here](https://github.com/DataDog/dd-sdk-ios/blob/5a657d8ed96ff49834c08fda3b673932eb80df92/Tests/DatadogTests/Datadog/Core/PerformancePresetTests.swift#L12), so repeating these assertions doesn't bring much convenience nor clarity.

I believe that what we really want to assert is: _"When performance preset is overrided, then new values are used"_. This could be asserted in a single test that registers two Features, one with override and another one without, and asserting that the latter uses expected values in their storage (which are different for first Feature). That could be simpler (one test) and more scalable (no config repetition), WDYT @maciejburda ?

ncreated (approved):
Looks good 👍👌. I left few minor feedbacks - no blocking.

Tests/DatadogTests/Datadog/DatadogCore/DatadogCoreTests.swift
maciejburda: I initially had it in one test, but I felt like it could use test name to improve clarity.

I think you're right. Let's put it back to one test - it shows the scope of the performance override better.

PR #1231, opened 2023-03-30
RUMM-2616 RUM Resource Instrumentation v2

maxep:
### What and why?

Use v2 URLSession instrumentation in RUM.

### How?

The `URLSessionRUMResourcesHandler` now complies with `DatadogURLSessionHandler` that was introduced in #1210 

The entire v1 `URLSessionAutoInstrumentation` has been removed.

#### RUM 2 APM

The `DatadogInternal.URLSessionTaskInterception` now support a `rum2APM` variable which specify if the distributed tracing is enabled in RUM. The value is set to `true` when a RUM Resource starts which will prevent the `DatadogTrace` from logging a span.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


Sources/Datadog/DatadogCore/DatadogCore.swift
```diff
@@ -384,51 +380,6 @@ extension DatadogCore: DatadogV1CoreProtocol {
...
-    /// Feature-specific part is provided explicitly with `featureSpecificConfiguration`.
-    ///
-    /// - Parameters:
-    ///   - configuration: The generic feature configuration.
-    ///   - featureSpecificConfiguration: The feature-specific configuration.
-    /// - Returns: an instance of V1 feature
-    func create<Feature: V1FeatureInitializable>(
-        configuration: DatadogFeatureConfiguration,
-        featureSpecificConfiguration: Feature.Configuration
-    ) throws -> Feature {
```
ncreated, line 400: Nice to see this finally gone 🙂💪

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskInterception.swift
```diff
@@ -65,6 +67,10 @@ public class URLSessionTaskInterception {
         )
     }
 
+    public func enableRUM2APM() {
+        self.rum2APM = true
+    }
```
ncreated, line 72: suggestion/ This API looks very odd (_"set bool flag"_). IIRC, this is set from RUM to skip further work in Tracing. Because we use `x-datadog-origin: rum` header for the same purpose (to let APM backend know that trace is managed by RUM) - have we considered using it here, instead of this new flag-API?

According to:
```diff
- /// To make sure the generated traces from RUM don’t affect APM Index Spans counts.
- /// **Note:** it is only added to requests that we create RUM Resource for (it is not injected when RUM feature is disabled and only Tracing is used). WDYT @maxep?
- public static let ddOrigin = (field: "x-datadog-origin", value: "rum")
```
if `request` was intercepted by RUM it will have `x-datadog-origin: rum` set. I can imagine that checking this header occurrence could be equivalent to `if rum2APM` check in Tracing handler - without introducing new API and only using existing constructs. WDYT @maxep ?

ncreated (approved):
It looks good 👍. No blocking feedbacks, but I wander what you think of my proposed replacement for `rum2APM` flag.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskInterception.swift
maxep: Totally, it's even better to use the origin as it is set at the very beginning of the interception.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskInterception.swift
```diff
@@ -27,26 +27,28 @@ public class URLSessionTaskInterception {
         spanID: SpanID,
         parentSpanID: SpanID?
     )?
+    /// Task metrics collected during this interception.
+    public private(set) var rum2APM: Bool = false
```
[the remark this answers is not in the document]
maxep: I've removed this in favor of https://github.com/DataDog/dd-sdk-ios/pull/1231#discussion_r1162466355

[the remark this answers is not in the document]
maxep: I've removed this in favor of https://github.com/DataDog/dd-sdk-ios/pull/1231#discussion_r1162466355

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskInterception.swift
maxep: It works for `URLRequest` but not for `URL` instrumentation as we cannot inject header into a URL object 😭

maxep: @ncreated I've changed the `rum2APM: Bool` to a `origin: String?`, this value is set automatically by reading the HTTP header field  "x-datadog-origin" and by RUM when starting an interception. LMKWYT!

DatadogTrace/Sources/Integrations/TracingURLSessionHandler.swift
```diff
@@ -90,7 +90,7 @@ internal struct TracingURLSessionHandler: DatadogURLSessionHandler {
     func interceptionDidComplete(interception: DatadogInternal.URLSessionTaskInterception) {
         guard
             interception.isFirstPartyRequest,  // `Span` should be only send for 1st party requests
-            !interception.rum2APM, // Send a span only if RUM 2 APM is disabled
+            interception.origin != "rum", // Send a span only if origin is not 'rum'
```
ncreated, line 93: nit/ This comment repeats exactly what the code says 🙂. I believe a better context for the reader would be to explain the RUM <> Tracing game.
```suggestion
            interception.origin != "rum", // if that request was tracked as RUM resource, the RUM backend will create the span on our behalf
```

Sources/Datadog/RUM/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
```diff
@@ -146,13 +146,14 @@ extension DistributedTracing {
...
+        // To make sure the generated traces from RUM don’t affect APM Index Spans counts.
+        request.setValue("rum", forHTTPHeaderField: TracingHTTPHeaders.originField)
+
         headerTypes.forEach {
             let writer: TracePropagationHeadersWriter
             switch $0 {
             case .datadog:
                 writer = HTTPHeadersWriter(sampler: sampler)
-                // To make sure the generated traces from RUM don’t affect APM Index Spans counts.
-                request.setValue("rum", forHTTPHeaderField: "x-datadog-origin")
```
ncreated: question/ Have we consulted this change with other teams? I'm referring to recent Slack discussion on setting DD-specific header in OTel-instrumented requests.

Tests/DatadogTests/Datadog/RUM/Instrumentation/Resources/URLSessionRUMResourcesHandlerTests.swift
```diff
@@ -49,7 +49,7 @@ class URLSessionRUMResourcesHandlerTests: XCTestCase {
             ]
         )
 
-        XCTAssertEqual(request.value(forHTTPHeaderField: "x-datadog-origin"), "rum")
+        XCTAssertEqual(request.value(forHTTPHeaderField: TracingHTTPHeaders.originField), "rum")
         XCTAssertEqual(request.value(forHTTPHeaderField: TracingHTTPHeaders.traceIDField), "1")
```
ncreated: suggestion/ The fact that we don't have to revamp this test after implemented change:
* from setting `x-datadog-origin` only in `.datadog` 
* to setting it for all header types (`.datadog`, `.b3`, ... `tracecontext`) 

means that this test is missing that coverage. 

Instead of having a single test that sets all header types and asserts on `x-datadog-origin` presence, we should define separate test for `x-datadog-origin`. With the recent change, it should assert that this header is set for all header types.

ncreated (approved):
Looks good 👌. I left a question and smaller suggestions, but it is good to go.

maxep:
CI is actually ✅ 
image

PR #1237, opened 2023-04-07
REPLAY-1505 Add `.maskUserInput` privacy mode

ncreated:
### What and why?

📦 This PR adds `.maskUserInput` privacy mode to Session Replay. This option will mask users input and their selections, while keeping all static text unmasked. This is to give more granular control over privacy in replays.

I made few mobile-specific adjustments (that will be validated with Product and alternatively fixed as further enhancements in next PRs), but the overall solution tries to follow [browser's SR "Mask user input mode"](https://docs.datadoghq.com/real_user_monitoring/session_replay/privacy_options#mask-user-input-mode) as close as possible.

Few examples of new mode:
<table>
<tr><td>`.maskAll`</td><td>`.maskUserInput` ⭐️</td></tr>
<tr>
   <td></td>
   <td></td>
</tr>
<tr>
   <td></td>
   <td></td>
</tr>
</table>

### How?

In this PR I propose a slight shift in the architecture of `TextObfuscators`:
* we no longer prepare `text*`, `selection*` and `sensitiveTextObfuscator` upfront (globally);
* instead, we decide which obfuscator to use in `NodeRecorder`'s configuration.

This brings more flexibility to text obfuscation in node recorders. Each recorder can now decide on their own strategy for text masking between `.maskAll` and `.maskUserInput`. Also, the strategy can be overwritten by parent recorder for certain subtrees (e.g. labels in picker views should be masked differently than standalone labels).

I made this change in response to two following observations:

#### (1) Some elements (e.g. TextFields) need to be masked differently in `.maskAll` and `.maskUserInput`
According to:
> [**Mask user input mode**](https://docs.datadoghq.com/real_user_monitoring/session_replay/privacy_options#mask-user-input-mode)
> Masks most form fields such as inputs, text areas, and checkbox values while recording all other text as is. Inputs are replaced with three asterisks (***) and text areas are obfuscated with space-preserving x characters.

the `.maskAll` option will replace regular text field's content with space-preserving number of `x`-es, but `.maskUserInput` should use fix-length `"xxx"` mask, unconditionally. This behaviour was impossible with globally-decided obfuscators.

#### (2) Some elements need to be masked differently in different subtrees
While exploring the problem, I made this diagram to realise that "`UILabels` inside `UIPickerView`" should be obfuscated differently than "`UILabels` inside `UIPickerView` inside `UIDatePicker`". This was enough evidence to decide on relaxing our architecture with recorder-specific obfuscator configuration:

<p align="center">
Screenshot 2023-04-07 at 18 06 38
</p>

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


DatadogSessionReplay/SRSnapshotTests/SRSnapshotTests.xcodeproj/xcshareddata/xcschemes/SRSnapshotTests.xcscheme
```diff
@@ -27,7 +27,17 @@
...
+      onlyGenerateCoverageForSpecifiedTargets = "YES">
+      <CodeCoverageTargets>
+         <BuildableReference
+            BuildableIdentifier = "primary"
+            BlueprintIdentifier = "DatadogSessionReplay"
+            BuildableName = "DatadogSessionReplay"
+            BlueprintName = "DatadogSessionReplay"
+            ReferencedContainer = "container:..">
+         </BuildableReference>
+      </CodeCoverageTargets>
```
ncreated, line 40: Before, it was gathering test coverage for all targets (e.g. including `TestUtilities`) which made no sense. Now it measures `DatadogSessionReplay` only. Current score is 85% (with units + snapshot tests).

DatadogSessionReplay/SRSnapshotTests/SRSnapshotTests/SRSnapshotTests.swift
```diff
@@ -21,257 +21,186 @@ final class SRSnapshotTests: SnapshotTestCase {
...
-            record: recordingMode
-        )
+        try forEachPrivacyMode { privacyMode in
+            let image = try takeSnapshot(configuration: .init(privacy: privacyMode))
+            DDAssertSnapshotTest(
+                newImage: image,
+                snapshotLocation: .folder(named: snapshotsFolderName, fileNameSuffix: "-\(privacyMode)-privacy"),
+                record: recordingMode
+            )
+        }
```
ncreated, line 31: Added `forEachPrivacyMode { _ in }` helper to not repeat the same dance for new mode.

DatadogSessionReplay/SRSnapshotTests/SRSnapshotTests/Utils/SnapshotTestCase.swift
```diff
@@ -71,6 +71,11 @@ internal class SnapshotTestCase: XCTestCase {
         waitForExpectations(timeout: seconds * 2)
     }
 
+    func forEachPrivacyMode(do work: (SessionReplayPrivacy) throws -> Void) rethrows {
+        let modes: [SessionReplayPrivacy] = [.maskAll, .allowAll, .maskUserInput]
```
ncreated, line 75: Could be `SessionReplayPrivacy.allCases`, but with explicit list we can decide on which snapshots to re-generate / assert. I found it handy 🙂.

DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/ViewTreeSnapshotBuilderTests.swift
```diff
@@ -31,33 +32,6 @@ class ViewTreeSnapshotBuilderTests: XCTestCase {
         XCTAssertEqual(queryContext.recorder, randomRecorderContext)
     }
 
-    func testItConfiguresTextObfuscatorsAccordinglyToCurrentPrivacyMode() throws {
-        // Given
```
ncreated, line 35: This test no longer makes sense - we don't configure obfuscators in `VTSBuilder` anymore.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UISegmentRecorder.swift
```diff
@@ -7,6 +7,14 @@
 import UIKit
 
 internal struct UISegmentRecorder: NodeRecorder {
+    var textObfuscator: (ViewTreeRecordingContext) -> TextObfuscating = { context in
+        switch context.recorder.privacy {
+        case .allowAll:         return context.textObfuscators.nop
+        case .maskAll:          return context.textObfuscators.spacePreservingMask
+        case .maskUserInput:    return context.textObfuscators.spacePreservingMask
+        }
+    }
```
maciejburda, line 16: The more I look at this the more it feels it should be something like:

```suggestion
    var textObfuscator: (SessionReplayPrivacy) -> TextObfuscating = { privacy in
        switch privacy {
        case .allowAll:         return TextObfuscators.nop
        case .maskAll:          return TextObfuscators.spacePreservingMask
        case .maskUserInput:    return TextObfuscators.spacePreservingMask
        }
    }
```

IMO slightly simpler because we remove dependency between context and obfuscator and highlight the fact that it's chosen based on the privacy property (from context).

maciejburda (approved):
LGTM,

One suggestion to simplify the closure a little bit.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UISegmentRecorder.swift
ncreated: Accepting `SessionReplayPrivacy` seems simpler at glance, but it enforces **global** and **static** obfuscators that would have to be freed from `SR.deinit()` if they use any state in the future (e.g. for mask caching). We tend to avoid singletons if possible (they're painful in tests).

The `ViewTreeRecordingContext` in turn will come handy for adding fine-grained privacy control later (e.g. different masking per element instance). It could be obvious / simplest choice to pass masking configuration through `context`.

How does it sound @maciejburda ?

PR #1240, opened 2023-04-10
REPLAY-1510 Button image background support

maciejburda:
### What and why?

Adds support for image backgrounds for UIButton.

As extra we provided similar solution for any image background that's coming from `UIImageView` which is direct child 
of `_UIBarBackground` which covers scenarios for `UINavigationBar`, `UITabBar` and `UIToolbar`.

Other considered solution was involving separate recorders, but this would introduce way more complexity.

Once we find reasons for adding separate recorders for any of the classes above we can gradually move the logic from the image generic to control specific image recorder.

### How?

Solution is based on UIImageView and UIImage extensions that provide number of flags representing a state of control. These states are taken into account in recording and tinting predicate of the recorder.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/DatadogSessionReplay/SRSnapshotTests/SRHost/Fixtures/Images.storyboard
+++ b/DatadogSessionReplay/SRSnapshotTests/SRHost/Fixtures/Images.storyboard
@@ -4,6 +4,7 @@
     <dependencies>
         <deployment identifier="iOS"/>
         <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="21678"/>
+        <capability name="Named colors" minToolsVersion="9.0"/>
         <capability name="Safe area layout guides" minToolsVersion="9.0"/>
         <capability name="System colors in document resources" minToolsVersion="11.0"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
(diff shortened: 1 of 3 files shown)


DatadogSessionReplay/SRSnapshotTests/SRHost/Fixtures/Images.storyboard
ncreated, line 1: At some point we could benefit from decomposing this snapshot into multiple ones. WDYT @maciejburda ?

Right now it is a single snapshot that lists dozen of images in different configurations. It's not obvious to me:
* which configurations do we cover vs which not?
* what exactly to fix if something breaks?
* how to add new configurations to this snapshot to not override existing ones?

Same as we prefer multiple unit tests over single complex one, I believe that dividing snapshots into smaller chunks could be healthy in the long term.

<p align="center">

</p>

ncreated (approved):
Looks good 👍👌. No blocking feedback 👌 

> Once we find reasons for adding separate recorders for any of the classes above we can gradually move the logic from the image generic to control specific image recorder.

I agree - it makes sense, to find simplest solution for this moment (we don't yet know what's ahead of us).

DatadogSessionReplay/SRSnapshotTests/SRHost/Fixtures/Images.storyboard
maciejburda: Yeah - perfectly valid point. I started to wonder about this as well.

For this PR let's keep it, but I'll start breaking it down + I'll provide a label explaining the use case in a follow up PR.

PR #1243, opened 2023-04-11
REPLAY-1516 Unsupported ViewController placeholder

maciejburda:
### What and why?

Adds placeholders for certain view controllers.

### How?

It's done by figuring out the private view class for each unsupported VC.
In that case we ignore it and put a placeholder wireframe in place.
testActivity()
testSafari()

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


maciejburda:
Add snapshot test

maciejburda:
As discussed - removed snapshot tests for these

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UnsupportedViewControllerRecorder.swift
```diff
@@ -0,0 +1,68 @@
...
+            )
+        ]
+    }
+}
+
+fileprivate extension UIResponder {
+    var parentViewController: UIViewController? {
+        return next as? UIViewController ?? next?.parentViewController
+    }
+}
```
ncreated, line 68: I'm quite concerned about this. With that approach, recording complexity is no longer `O(N)` (N - the number of views in hierarchy), but worse 🐌. Here we traverse the responder chain backward during each step of forward view-tree traversal - it must be slower than `O(N)`.

I collected quick measures with landmarks example app, measuring the time spend in [`snapshotBuilder.createSnapshot(of: window, with: context)`](https://github.com/DataDog/dd-sdk-ios/blob/dd3fc8077342c90a84b825455cedb9a4a9a8e7f5/DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/WindowViewTreeSnapshotProducer.swift#L21) with `UnsupportedViewControllerRecorder` enabled and without it (using the [benchmark snippet](https://gist.github.com/ncreated/db931ed0202721b50fab56c3ccdcd4e8) made a weeks ago):
```
List screen:
* without: AVG createSnapshot(): 0.8357524871826172 ms
* with: AVG createSnapshot(): 1.0014202859666612 ms (+20%)

Modal screen:
* without: AVG createSnapshot(): 0.969586107465956 ms
* with: AVG createSnapshot(): 1.1173990037706163 ms  (+15%)

UI Controls screen:
* without: AVG createSnapshot(): 0.4487183358934191 ms
* with: AVG createSnapshot(): 0.5008896191914876 ms (+12%)

With calendar control opened:
* without: AVG createSnapshot(): 1.262982686360677 ms
* with: AVG createSnapshot(): 1.3359242015414767 ms (+5%)
```
It gets worse on more complex screens. As a baseline, I did same tests with disabling some other node recorders - none of them seems to bring +20% impact.

IMHO, to be performant, view controller check requires more presence in our architecture. What I can quickly propose is:
* computing responder chain snapshot separately (that would be `O(R)` complexity)
* adding sort of `context.responderChainSnapshot.isPartOfUnsupportedVC(view)` function that is `O(1)`

This way we will end up with `O(N + R)` complexity, so `O(N)` effectively.

ncreated (changes requested):
I get the idea 👌, but backward iteration over responder chain for each view is too slow (it adds +20% to main thread time in some measured cases). We need a different approach IMO - I left one suggestion.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UnsupportedViewControllerRecorder.swift
maciejburda: Got it. Very good point. I haven't considered the performance bottleneck here.

There other strategy I was thinking about is exctracting concrete private classes from those view controllers and ignoring them using previously added `UnsupportedViewsRecorder`

ncreated: If only we find it reliable, it could be a good short-term option 👍. Given that we will eventually support "unsupported view controllers" 😉, it might be a good choice.

On the other hand, I believe that "getting into VCs hierarchy" is unavoidable in the long term (e.g. might be required for fine-grained privacy controls). We could revisit the option I proposed once we find more needs for it.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UnsupportedViewRecorder.swift
```diff
@@ -29,7 +31,7 @@ internal struct UnsupportedViewRecorder: NodeRecorder {
         let builder = UnsupportedViewWireframesBuilder(
             wireframeRect: view.frame,
             wireframeID: context.ids.nodeID(for: view),
-            unsupportedClassName: String(reflecting: type(of: view)),
+            unsupportedClassName: context.viewControllerContext.name ?? String(reflecting: type(of: view)),
```
ncreated, line 34: suggestion/ The `String(reflecting:)` produces this ugly description for SwiftUI view and I can imagine this being even more odd for more complex SwiftUI contents:
```
SwiftUI._UlHostingView<SHost.(unknown context at $100c1523c).TestView>
```

Why not describing it just as `"SwiftUI"`? The reflected description of its content doesn't bring any value.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UnsupportedViewRecorder.swift
```diff
@@ -10,17 +10,19 @@ import SwiftUI
...
-        { $0 is UIActivityIndicatorView },
-        { String(reflecting: type(of: $0)).contains("SwiftUI") }
+    private let unsupportedViewsPredicates: [(UIView, ViewTreeRecordingContext) -> Bool] = [
+        { _, context in context.viewControllerContext.isRootView(of: .safari) },
+        { _, context in context.viewControllerContext.isRootView(of: .activity) },
+        { view, _ in view is UIProgressView },
+        { view, _ in view is UIWebView },
+        { view, _ in view is WKWebView },
+        { view, _ in view is UIActivityIndicatorView },
+        { view, _ in String(reflecting: type(of: view)).contains("SwiftUI") }
```
ncreated, line 20: suggestion/ Looks like we could get more clarity and coherence if this is determined equally to `.activity` or `.safari`. Now "SwiftUI content" is recognised as an agreement and logic in multiple code units, but we could do it all in one place:
```swift
{ _, context in context.viewControllerContext.isRootView(of: .swiftUI) },
```

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/ViewTreeRecordingContext.swift
```diff
@@ -0,0 +1,89 @@
...
+                    self = .other
+                }
+            }
+        }
+
+        /// The parent view controller's type.
+        var parentType: ViewControllerType?
+
+        /// A boolean flag indicating whether the current view is the root view or not.
+        var isRootView: Bool?
```
ncreated, line 59: question/ What would `nil` value mean for this boolean flag? Looks that it should be non-optional.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/ViewTreeSnapshotBuilder.swift
```diff
@@ -57,8 +57,8 @@ extension ViewTreeSnapshotBuilder {
 }
 
 /// An arrays of default node recorders executed for the root view-tree hierarchy.
-internal func createDefaultNodeRecorders() -> [NodeRecorder] {
-    return [
+internal func createDefaultNodeRecorders(adding: [NodeRecorder] = []) -> [NodeRecorder] {
+    return adding + [
```
ncreated: cleanup/ It doesn't seem we leverage this `adding:` parameter anywhere. Isn't it a leftover?

ncreated (changes requested):
💪 It looks good in overall 👌. I left few suggestions, with spotting two places in tests that will become flaky ❄️ once we start comparing snapshot images on CI (coming pretty soon).

maciejburda:
Thanks for great review!

I was exploring static content, but unfortunately it's on possible for web view (not safari vc).

Made some changes. Pushing soon :)

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/ViewTreeRecordingContext.swift
```diff
@@ -80,10 +84,16 @@ internal extension ViewTreeRecordingContext {
...
                  .none:
                 return nil
             }
         }
     }
 }
+
+private protocol AnyUIHostingViewController: AnyObject {}
+@available(iOS 13.0, *)
+extension UIHostingController: AnyUIHostingViewController {}
```
ncreated, line 99: Nice 🥇

ncreated (approved):
👏 🤜 🤛

PR #1249, opened 2023-04-14
REPLAY-1523 Text alignment

maciejburda:
### What and why?

Adds support for text alignment.

### How?

Adds unified initialiser for UIKit->SR mapping and makes sure it's used across the board.
One snapshot test was updated with right and center alignment.

It removes logic for calculating vertical position using padding. Now it all takes whole frame of the view and achieves correct display by using text alignment.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/DatadogSessionReplay/Sources/Writer/Models/SRDataModels+UIKit.swift
+++ b/DatadogSessionReplay/Sources/Writer/Models/SRDataModels+UIKit.swift
@@ -8,6 +8,8 @@
 import UIKit
 
 extension SRTextPosition.Alignment {
+    /// Custom initializer that allows transforming UIKit's `NSTextAlignment` into `SRTextPosition.Alignment`.
+    /// Must be called on the main thread due to `UIApplication.shared.userInterfaceLayoutDirection`.
     init(textAlignment: NSTextAlignment) {
         self.vertical = .center
         switch textAlignment {
--- a/DatadogSessionReplay/Sources/Writer/Models/SRDataModels+UIKit.swift
+++ b/DatadogSessionReplay/Sources/Writer/Models/SRDataModels+UIKit.swift
@@ -1,18 +1,19 @@
-//
-//  File.swift
-//  
-//
-//  Created by Maciej Burda on 13/04/2023.
-//
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
 
 import UIKit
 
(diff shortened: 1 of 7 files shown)


DatadogSessionReplay/Sources/Writer/Models/SRDataModels+UIKit.swift
```diff
@@ -0,0 +1,32 @@
+//
+//  File.swift
+//  
+//
+//  Created by Maciej Burda on 13/04/2023.
+//
+
```
ncreated, line 7: blocking/ This header and some linter failure on our [custom rule](https://github.com/DataDog/dd-sdk-ios/blob/dd3fc8077342c90a84b825455cedb9a4a9a8e7f5/tools/lint/sources.swiftlint.yml#L80-L88):
```
⛔️ Line 25: `UIApplication.shared` is unavailable in some environments. Check `UIApplication.managedShared`.
Done linting! Found 1 violation, 1 serious in 352 files.
```
See https://github.com/DataDog/dd-sdk-ios/pull/374 for more context on that rule ☝️

DatadogSessionReplay/Sources/Writer/Models/SRDataModels+UIKit.swift
```diff
@@ -0,0 +1,32 @@
...
+            self.horizontal = .right
+        case .justified:
+            self.horizontal = .left
+        default:
+            if UIApplication.shared.userInterfaceLayoutDirection == .leftToRight {
+                self.horizontal = .left
+            } else {
+                self.horizontal = .right
+            }
+        }
```
ncreated, line 30: blocking/ We're not yet handling RTL languages and adding their support is far more than this `default` case (it will require more changes in the SDK, support to custom fonts and set of enhancements in the player). 

I don't see any value in "mimicking" RTL if it doesn't bring anything further (won't work well). In fact, we're making our life harder with this code as now it _**"Must be called on the main thread"**_. That threading constraint could be easily removed if we invest more effort: capture the layout direction on the main thread and propagate it along with the recording context down to wireframe builders (background thread). Do we want to pay this effort now given that we're unable to fully support RTL?

ncreated (changes requested):
It's IMO too much to compromise on thread safety with adding imperfect RTL support (and we can't make it perfect yet).

DatadogSessionReplay/Sources/Writer/Models/SRDataModels+UIKit.swift
maciejburda: Sounds good, I'll add a ticket 🙌

ncreated (approved):
👌 LG

PR #1250, opened 2023-04-14
RUMM-3133 Don't override ViewScope attributes with UserActionScope

ganeshnj:
### What and why?

Today, when a `UserAction`, `Resource` or `LongTask` event is sent with attributes attached, if the attributes have conflicting keys with View event, the View attributes are overwritten by the `UserAction`/`Resource`/`LongTask` attributes.

`UserActionScope` events must not update the `ViewScope` attributes, which has been the case for Android SDK.

### How?

Updated logic that overwrites the `ViewScope` attributes with `UserActionScope` attributes with unit tests to ensure that the `ViewScope` attributes are not overwritten.

Fixed `Resource` events where parent attributes were reported when there is a conflict.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
@@ -109,7 +109,7 @@ internal class RUMResourceScope: RUMScope {
     // MARK: - Sending RUM Events
 
     private func sendResourceEvent(on command: RUMStopResourceCommand, context: DatadogContext, writer: Writer) {
-        attributes.merge(rumCommandAttributes: command.attributes)
+        attributes.mergeWithoutOverwriting(rumCommandAttributes: command.attributes)
 
         let resourceStartTime: Date
         let resourceDuration: TimeInterval
--- a/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
+++ b/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
@@ -109,7 +109,7 @@ internal class RUMResourceScope: RUMScope {
     // MARK: - Sending RUM Events
 
     private func sendResourceEvent(on command: RUMStopResourceCommand, context: DatadogContext, writer: Writer) {
-        attributes.mergeWithoutOverwriting(rumCommandAttributes: command.attributes)
+        attributes.merge(rumCommandAttributes: command.attributes)
 
         let resourceStartTime: Date
         let resourceDuration: TimeInterval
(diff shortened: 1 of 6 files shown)


Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -421,7 +421,14 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
...
         version += 1
-        attributes.merge(rumCommandAttributes: command.attributes)
+
+        // RUMM-3133 Don't override View attributes with UserAction, Resource or LongTask attributes
+        switch command {
+        case is RUMUserActionCommand, is RUMResourceCommand, is RUMAddLongTaskCommand:
+            break
+        default:
+            attributes.merge(rumCommandAttributes: command.attributes)
+        }
```
maxep, line 431: I think we should do the opposite and check for view-related command types instead.
Maybe it could use a
```swift
internal protocol RUMViewCommand: RUMCommand { }
```
for selecting which command types can propagate attributes to the view-scope. It would be less error prone when we add command  types. WDYT?

maxep (commented):
Awesome @ganeshnj !
This contribution is on point and the tests properly follow our pattern 👏 
I've left a couple of comments and I would be interested to know your opinion on how we select the command types that can propagate attributes.

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift
ganeshnj: For sure having one base protocol for all the commands that we don't want to propagate would be ideal. I like the idea but I'm not sure what other commands we can expect here. This is the same rational behind using `break` with current `default` behavior.

Any hints on commands that we don't want propagation for?

Sources/Datadog/RUM/RUMMonitor/Scopes/RUMResourceScope.swift
```diff
@@ -109,7 +109,7 @@ internal class RUMResourceScope: RUMScope {
     // MARK: - Sending RUM Events
 
     private func sendResourceEvent(on command: RUMStopResourceCommand, context: DatadogContext, writer: Writer) {
-        attributes.merge(rumCommandAttributes: command.attributes)
+        attributes.mergeWithoutOverwriting(rumCommandAttributes: command.attributes)
```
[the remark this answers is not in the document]
ganeshnj: The issue I found here was as below (from test case)
```swift
monitor.addAttribute(forKey: "abc", value: "123")

// when
monitor.startView(key: "View", name: nil, attributes: [:])
monitor.addUserAction(type: .custom, name: "action1", attributes: ["abc": "456"])

monitor.startResourceLoading(resourceKey: "/resource1", url: URL(string: "https://foo.com/1")!, attributes: ["abc": "456"])
monitor.stopResourceLoading(resourceKey: "/resource1", response: .mockAny(), size: nil, attributes: ["def": "789"])
```

here want to keep the `abc` attribute value reported in the resource event as `456` (along with `def`: `789`). Currently, we overwrite it to `123` which is incorrect from my understanding. Anything that is child scope must take a precedence over the parent scope.

Do you have an example where this logic will go wrong? I can investigate further.

maxep (approved):
Perfect! CI status didn't get through but it's actually green:
image

Thanks again, @ganeshnj !

PR #1251, opened 2023-04-14
RUMM-2616 Create DatadogCrashReporting Module

maxep:
### What and why?

As part of v2 migration, create the `DatadogCrashReporting ` module.

### How?

Move sources and tests to dedicated targets.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [x] Run smoke tests


ganeshnj (commented):
Nice and clean 🥳 

Couple of comments/questions for my understanding, rest looks good to me.

DatadogCrashReporting/Sources/DatadogCrashReporter.swift
```diff
@@ -22,18 +22,30 @@ import DatadogInternal
     /// Integration enabling sending crash reports as Logs or RUM Errors.
     let sender: CrashReportSender
 
-    convenience init?(
-        core: DatadogCoreProtocol,
-        configuration: FeaturesConfiguration.CrashReporting
+    /// Initializes the Datadog Crash Reporter.
+    public static func initialize(
+        plugin: DDCrashReportingPluginType = DDCrashReportingPlugin(),
```
ncreated, line 27: Do we still need `DDCrashReportingPluginType` and `DDCrashReportingPlugin` in public interface? This design made sense when `DDCrashReportingPluginType` (interface) was defined in `Datadog` module and its implementation was provided by `DatadogCrashReporting`.

Now because both are in `DatadogCrashReporting` it seems reasonable to made plugin `internal`, WDYT?

DatadogInternal/Sources/Upload/DefaultJSONEncoder.swift
```diff
@@ -1,14 +1,15 @@
 /*
  * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
  * This product includes software developed at Datadog (https://www.datadoghq.com/).
- * Copyright 2019-Present Datadog, Inc.
+ * Copyright 2019-2020 Datadog, Inc.
```
ncreated, line 4: ⚠️  Rather not intended.

Sources/Datadog/DatadogCore/DatadogCore.swift
```diff
@@ -206,9 +206,6 @@ internal final class DatadogCore {
         allUploads.forEach { $0.flushAndTearDown() }
         allStorages.forEach { $0.setIgnoreFilesAgeWhenReading(to: false) }
 
-        // Deinitialize V2 Integrations (arbitrarily for now, until we make it into `DatadogFeatureIntegration`):
-        get(feature: CrashReporter.self)?.deinitialize()
-
```
ncreated, line 211: Doesn't this imply more cleanup possibilities? Notably:
* `var v1Features`
* `extension DatadogCore: DatadogV1CoreProtocol`
* `internal protocol V1Feature`

Tests/DatadogTests/DatadogObjc/ObjcAPITests/DDConfiguration+apiTests.m
```diff
@@ -94,7 +94,6 @@ - (void)testDDConfigurationBuilderAPI {
     [builder enableLogging:YES];
     [builder enableTracing:YES];
     [builder enableRUM:YES];
-    [builder enableCrashReportingUsing:[DDCrashReportingPlugin new]];
```
ncreated, line 97: This makes me think that `DatadogCrashReporter.initialize()` is not possible from Objective-C, right? Is this sth we will tackle later instead of now? Does it relate to other modules as well?

ncreated (commented):
Looks really good 🎯👌. I left few important questions, so I hold the approve.

Datadog/Example/ExampleAppDelegate.swift
```diff
@@ -51,7 +51,6 @@ class ExampleAppDelegate: UIResponder, UIApplicationDelegate {
             .enableLogging(true)
             .enableTracing(true)
             .enableRUM(true)
-            .enableCrashReporting(using: DDCrashReportingPlugin())
```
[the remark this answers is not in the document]
maxep: Yes, it is now enabled with `DatadogCrashReporter.initialize()`.

DatadogCrashReporting/Sources/DatadogCrashReporter.swift
```diff
@@ -7,10 +7,10 @@
 import Foundation
 import DatadogInternal
 
-/* public */ internal class CrashReporter: DatadogFeature {
-    /* public */ static let name = "crash-reporter"
+public final class DatadogCrashReporter: DatadogFeature {
+    public static let name = "crash-reporter"
 
-    /* public */ let messageReceiver: FeatureMessageReceiver
+    public let messageReceiver: FeatureMessageReceiver
```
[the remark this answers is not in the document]
maxep: It has been made public because it complies with `DatadogFeature`, but that's a good call and I think we should split the interface to only expose what's necessary for the user.

DatadogCrashReporting/Sources/DatadogCrashReporter.swift
maxep: Indeed, we can make more stuff internal now 👍

Tests/DatadogTests/DatadogObjc/ObjcAPITests/DDConfiguration+apiTests.m
maxep: Ah yes, good catch! We can easily make it accessible from objc, but we have to come up with a different name because `initialize` is reserved. We could do something like this:
```swift
@objc
public final class DatadogCrashReporter: NSObject {

    public static func initialize(in core: DatadogCoreProtocol = defaultDatadogCore)

    @available (*, unavailable) @objc
    public static func enable()
}
```

The `@available (*, unavailable) @objc` will make it available for objc only.

maxep: Or just:
```swift
@available(swift, obsoleted: 1) @objc(DatadogCrashReporter)
public final class objc_DatadogCrashReporter: NSObject {

    @objc
    public static func enable() {
        DatadogCrashReporter.initialize()
    }
}
```

I think I prefer this option for isolation, WDYT?

DatadogLogs/Sources/Feature/MessageReceivers.swift
```diff
@@ -189,7 +189,7 @@ internal struct CrashLogReceiver: FeatureMessageReceiver {
 
         // crash reporting is considering the user consent from previous session, if an event reached
         // the message bus it means that consent was granted and we can safely bypass current consent.
-        core.v1.scope(for: DatadogLogsFeature.name)?.eventWriteContext(bypassConsent: true, forceNewBatch: false) { _, writer in
+        core.scope(for: DatadogLogsFeature.name)?.eventWriteContext(bypassConsent: true, forceNewBatch: false) { _, writer in
```
maxep, line 192: This was a miss! It didn't change the behaviour since `DatadogV1CoreProtocol` was extending the `DatadogCoreProtocol`.

DatadogInternal/Sources/DatadogCoreProtocol.swift
```diff
@@ -94,14 +94,6 @@ extension DatadogCoreProtocol {
...
         send(message: message, else: {})
     }
-
-    /// Sends a message on the bus shared by features registered in this core.
-    ///
-    /// - Parameters:
-    ///   - message: The message.
-    public func send(message: FeatureMessage, else fallback: @escaping () -> Void) {
-        send(message: message, else: fallback)
-    }
```
maxep, line 104: leftover from #1245

DatadogInternal/Sources/DatadogCoreProtocol.swift
```diff
@@ -160,5 +152,5 @@ public class NOPDatadogCore: DatadogCoreProtocol {
     /// no-op
     public func set(feature: String, attributes: @escaping @autoclosure () -> FeatureBaggage) { }
     /// no-op
-    public func send(message: FeatureMessage, sender: DatadogCoreProtocol, else fallback: @escaping () -> Void) { }
+    public func send(message: FeatureMessage, else fallback: @escaping () -> Void) { }
```
maxep, line 155: leftover from 1245

Tests/DatadogTests/DatadogObjc/ObjcAPITests/DDConfiguration+apiTests.m
ncreated: Oh 😞, `initialize()` conflict is a pity. Given that we don't yet have [proper documentation](https://docs.datadoghq.com/real_user_monitoring/error_tracking/ios/?tab=cocoapods) for crash reporting in Objective-C, it will be harder to discover (and `initialize()` could be called by mistake). Let's go with `enable()` now and we can visit it when aligning public interfaces.

One more thing - it will be more consistent with other APIs and more idiomatic to Objective-C if we call it `DDCrashReporter`, not `DatadogCrashReporter`.

> The @available(swift, obsoleted: 1) @objc will make it available for objc only.

Interesting!

ncreated (approved):
Looks good 👌. Remaining problems of Obj-c interface can be solved when aligning other public APIs.

PR #1259, opened 2023-04-20
RUMM-3194 add sampling rate on Tracer

xgouchet:
### What and why?

Allow customer sampling their traces on the mobile side

### How?

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


Tests/DatadogTests/Matchers/SpanMatcher.swift
```diff
@@ -18,6 +20,9 @@ extension String: AllowedSpanAttributeValue {}
 extension UInt64: AllowedSpanAttributeValue {}
 extension Int: AllowedSpanAttributeValue {}
 
+// Only numeric values are allowed for `span.ç_dd.*`.
```
fuzzybinary, line 23: Is that supposed to be `ç_dd` or just `_dd`

xgouchet: Indeed thanks.

xgouchet:
> Do spans affect their children? If so, do we need to check / test the spans created by the URLSession delegate, and make sure the sampled Spans get encoded to the W3C / b3 headers?

The trace's sampling is being handled by the root span, meaning that this is going to be handled automatically by APM's logic on DD servers

maxep (approved):
Nice, thanks @xgouchet!

PR #1263, opened 2023-04-24
RUMM-3208 Extract WebView tracking to DatadogWebViewTracking module

ganeshnj:
### What and why?

As part of V2 migration, WebView tracking needs to be its own separate module. This will allows customers to decouple WebView tracking from the core SDK.

If a customer want to use WebView tracking, they will need to add dependency on `DatadogWebViewTracking` module.

### How?

Usage of WebView tracking doesn't change apart from a slight API name change.

The updated API is called `startTrackingDatadogEvents` which is available on `WKUserContentController`.

```swift
controller.startTrackingDatadogEvents(core: core, hosts: [])
```

The stop API remains unchanged.

WebView tracking is also used in the cross platform applications such as ReactNative which will be updated to use the new API.

Rather than depending on the static instance of `WebEventBridge`, the consumers have to create an instance of `WebViewTracking` implementation (essentially `WebViewTrackingCore`.

```swift
let webViewTracking = WebViewTrackingCore(core: core)
```

Convenience API for `WKUserContentController` to send raw events.
```swift
try webViewTracking.send(body: someBody)
```

API to convert the send modeled messages
```swift
let message = try WebViewTrackingMessage...
try webViewTracking.send(message: message)
```

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


DatadogWebViewTracking/Sources/WebViewTrackingMessage.swift
```diff
@@ -0,0 +1,73 @@
...
+
+public typealias JSON = [String: Any]
+
+/// Intermediate type to parse WebView messages and send them to the message bus
+public enum WebViewTrackingMessage {
+    /// A log message with a JSON payload
+    case log(JSON)
+
+    /// A RUM event with a JSON payload
+    case rumEvent(JSON)
```
maxep, line 17: /nit I like this enum. About the case names, I think it should be `.log` and `.rum` **or** `.logEvent` and `.rumEvent`

DatadogWebViewTracking/Sources/WebViewTracking.swift
```diff
@@ -0,0 +1,57 @@
...
+/// Defines methods to send WebView related information
+public protocol WebViewTracking {
+    /// Sends a bag of data
+    /// - Parameter body: The data to send, it must be parsable to `WebViewTrackingMessage`
+    func send(body: Any) throws
+
+    /// Sends a message
+    /// - Parameter message: The message to send
+    func send(message: WebViewTrackingMessage) throws
+}
```
maxep, line 18: For conformance it would only need the `send(message:)` method, while the convenient `send(body:)` could be provided in an extension:
```swift
/// Defines methods to send WebView related information
public protocol WebViewTracking {
    /// Sends a message
    /// - Parameter message: The message to send
    func send(message: WebViewTrackingMessage) throws
}

extension WebViewTracking {
    /// Sends a bag of data to the message bus
    /// - Parameter body: The data to send, it must be parsable to `WebViewTrackingMessage`
    public func send(body: Any) throws {
        let message = try WebViewTrackingMessage(body: body)
        try send(message: message)
    }
}
```

DatadogWebViewTracking/Sources/WebViewTracking.swift
```diff
@@ -0,0 +1,57 @@
...
+/// - Note:
+/// Cross platform SDKs should instantiate this type with a `DatadogCoreProtocol` implementation
+/// and pass WebView related messages using the message bus of the core.
+public struct WebViewTrackingCore: WebViewTracking {
+    enum MessageKeys {
+        static let browserLog = "browser-log"
+        static let browserRUMEvent = "browser-rum-event"
+    }
+
+    private let core: DatadogCoreProtocol
```
maxep, line 30: /blocking We should keep a `weak` ref to the core. Even if we won't have any retain cycle with webview tracking, it's best not to retain the core so we can control its deallocation from the future core registry.

DatadogWebViewTracking/Tests/WKUserContentController+DatadogTests.swift
```diff
@@ -144,71 +142,129 @@ class WKUserContentController_DatadogTests: XCTestCase {
...
+                        XCTAssertEqual(event["date"] as! Int64, 1_635_932_927_012)
+                        XCTAssertEqual(event["message"] as! String, "console error: error")
+                        XCTAssertEqual(event["status"] as! String, "error")
+                        XCTAssertEqual(event["view"] as! [String: String], ["referrer": "", "url": "https://datadoghq.dev/browser-sdk-test-playground"])
+                        XCTAssertEqual(event["error"] as! [String : String], ["origin": "console"])
+                        XCTAssertEqual(event["session_id"] as! String, "0110cab4-7471-480e-aa4e-7ce039ced355")
+                        logMessageExpectation.fulfill()
+                    case "browser-rum-event":
+                        let event = baggage.attributes as JSON
+                        XCTAssertEqual((event["view"] as! JSON)["id"] as! String, "64308fd4-83f9-48cb-b3e1-1e91f6721230")
```
maxep, line 163: We can avoid forcing the unwrap here, I think `XCTAssertEqual` accepts optional.

DatadogWebViewTracking/Tests/WebViewTrackingCoreTests.swift
```diff
@@ -0,0 +1,193 @@
...
+                case .custom(let key, let baggage):
+                    switch key {
+                    case "browser-log":
+                        let event = baggage.attributes as JSON
+                        XCTAssertEqual(event["date"] as! Int64, 1_635_932_927_012)
+                        XCTAssertEqual(event["message"] as! String, "console error: error")
+                        XCTAssertEqual(event["status"] as! String, "error")
+                        XCTAssertEqual(event["view"] as! [String: String], ["referrer": "", "url": "https://datadoghq.dev/browser-sdk-test-playground"])
+                        XCTAssertEqual(event["error"] as! [String : String], ["origin": "console"])
+                        XCTAssertEqual(event["session_id"] as! String, "0110cab4-7471-480e-aa4e-7ce039ced355")
```
maxep, line 69: Same here about unwrapping.

maxep (changes requested):
Amazing! I really like the new APIs, it's well designed and it will be easy for CP to migrate. Also, thanks for making the tests more readable. Well done!

I've left one minor blocking comment about retaining the `core`, which we can easily avoid with a `weak` ref.

DatadogWebViewTracking/Sources/WebViewTracking.swift
ganeshnj: Swifty, done.

DatadogWebViewTracking/Sources/WebViewTracking.swift
ganeshnj: that makes sense, I made it weak.

DatadogWebViewTracking/Sources/WebViewTracking.swift
```diff
@@ -0,0 +1,56 @@
...
+    private weak var core: DatadogCoreProtocol?
+
+    public init(core: DatadogCoreProtocol) {
+        self.core = core
+    }
+
+    /// Sends a message to the message bus
+    /// - Parameter message: The message to send
+    public func send(message: WebViewTrackingMessage) throws {
+        assert(core != nil, "Core must not be nil when using WebViewTracking")
```
ncreated, line 35: We avoid assertions in production code (in case someone enables them for release build). Instead, we use `DD.logger` to make misconfigurations verbose (if `Datadog.verbosityLevel` is enabled).

DatadogWebViewTracking/Sources/WebViewTrackingMessage.swift
ganeshnj: went with no Event suffix.

DatadogWebViewTracking/Sources/WebViewTracking.swift
ganeshnj: interesting, have we had such cases before?

happy to change it to a log message.

ganeshnj: I'm curious about this decision, I expect someway of breaking the code when there is developer error. Losing this ability and using logs seems like a back step.

ncreated: > interesting, have we had such cases before?

The very first one we added is perhaps [this](https://github.com/DataDog/dd-sdk-ios/blob/69ff9fe644098c3bc55d67af82e4600f8009bca7/Sources/Datadog/Datadog.swift#L83-L100) around `Datadog.initialize()` sequence. To not make it silent (and because of unavailability of `DD.logger` at this code point), we use `print()` directly.

> I'm curious about this decision, I expect someway of breaking the code when there is developer error. Losing this ability and using logs seems like a back step.

Early we decided on [zero crashes in our code](https://github.com/DataDog/dd-sdk-ios/blob/develop/ZEN.md#stability) and we stick to this idea to this day. Because we're "just observability" SDK, we prefer to fail silently rather than risk crashing the app in production. Also, as we rely on crash telemetry, any assertion failure of this kind would be considered a false-positive for our own monitoring.

DatadogWebViewTracking/Sources/WebViewTracking.swift
```diff
@@ -0,0 +1,58 @@
...
+            })
+        }
+    }
+}
+
+extension WebViewTracking {
+    /// Sends a bag of data to the message bus
+    /// - Parameter body: The data to send, it must be parsable to `WebViewTrackingMessage`
+    public func send(body: Any) throws {
+        let message = try WebViewTrackingMessage(body: body)
```
ncreated, line 55: non-blocking/ The API comment is a bit vague and refers to internals (_"must be parsable to `WebViewTrackingMessage`"_). Any reason to not require `body: String` and describe the requirement as _"UTF-8 JSON string"_?

Asking, because what happens next in `WebViewTrackingMessage(body:)` is:
```swift
internal init(body: Any) throws {
   guard let message = body as? String else {
```

DatadogWebViewTracking/Tests/WKUserContentController+DatadogTests.swift
```diff
@@ -144,71 +142,129 @@ class WKUserContentController_DatadogTests: XCTestCase {
     }
 
     func testSendingWebEvents() throws {
```
ncreated, line 144: 🏅 Really nice work on this one - it's much more readable now 💪

ncreated (approved):
Really nice! 🎯 👍

DatadogWebViewTracking/Sources/WebViewTracking.swift
ganeshnj: this is intentional, to allow directly using the `WebViewTracking` without any preprocessing as a convenience method. Ideally we don't want to expose this outside. Hence just made it internal.

```swift
internal init(body: Any) throws {
   guard let message = body as? String else {
```
^^ is like a `make` or `factory initializer`. We have to start from `Any` instance whether here or the caller (WK gives us Any instance). Given it is internal, it doesn't bother me much as we can change any point of time.

maxep (approved):
Great work @ganeshnj !

DatadogWebViewTracking/Sources/WebViewTracking.swift
ncreated: All clear 👍 - and more than fine with this being now internal 💪.

PR #1264, opened 2023-04-25
RUMM-3239 Fix RUM Context in Logs

maxep:
### What and why?

Fix RUM context not being attached to log when no user action exists.

### How?

Allow `nil` values in RUM context.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Tests/DatadogTests/Datadog/LoggerTests.swift
+++ b/Tests/DatadogTests/Datadog/LoggerTests.swift
@@ -568,15 +568,12 @@ class LoggerTests: XCTestCase {
                 forKeyPath: RUMContextAttributes.IDs.applicationID,
                 equals: rum.configuration.applicationID
             )
-        }
-        logMatchers.forEach {
+
             $0.assertValue(
                 forKeyPath: RUMContextAttributes.IDs.sessionID,
                 isTypeOf: String.self
             )
-        }
 
-        logMatchers.forEach {
             $0.assertValue(
                 forKeyPath: RUMContextAttributes.IDs.viewID,
                 isTypeOf: String.self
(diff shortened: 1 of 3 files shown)


Tests/DatadogTests/Datadog/LoggerTests.swift
```diff
@@ -551,28 +551,41 @@ class LoggerTests: XCTestCase {
...
+                isTypeOf: String.self
+            )
+        }
+
+        logMatchers.first?.assertNoValue(forKeyPath: RUMContextAttributes.IDs.userActionID)
+
+        logMatchers.last?.assertValue(
             forKeyPath: RUMContextAttributes.IDs.userActionID,
             isTypeOf: String.self
         )
```
ganeshnj, line 591: trivial: `!` would be better to use here than `?` as we don't want to skip assertion when `first` and `last` are nil.
```suggestion
        logMatchers.first!.assertNoValue(forKeyPath: RUMContextAttributes.IDs.userActionID)

        logMatchers.last!.assertValue(
            forKeyPath: RUMContextAttributes.IDs.userActionID,
            isTypeOf: String.self
        )
```

maxep: I have added the `XCTAssertEqual(logMatchers.count, 2)` as a first assertion for that 👍  I'm afraid that if we force unwrapping, we would crash the execution and prevent other tests to run :/

Tests/DatadogTests/Datadog/LoggerTests.swift
```diff
@@ -551,28 +551,41 @@ class LoggerTests: XCTestCase {
...
+                forKeyPath: RUMContextAttributes.IDs.sessionID,
+                isTypeOf: String.self
+            )
+        }
+
+        logMatchers.forEach {
+            $0.assertValue(
+                forKeyPath: RUMContextAttributes.IDs.viewID,
+                isTypeOf: String.self
+            )
```
fuzzybinary, line 583: Why not merge all of these into one? `forEach`

fuzzybinary (approved):

ganeshnj (approved):

maxep:
image

PR #1265, opened 2023-04-25
Fix URLSession swizzler binding operations

maxep:
### What and why?

We are experiencing flakiness in unit tests on v2 due to 2 issues introduced in #1210
1. `URLSessionSwizzler.bind()` and `URLSessionSwizzler.unbind()` were not thread-safe and could lead to faulty binding state.
2. The `.unbind()` was wrongly disabling swizzling on first call: faulty management of the `bindingsCount`.

### How?

1. `.bind()` and `.unbind()` are now acquiring a `NSLock`.
2. The `.unbind()` now properly decrements the `bindingsCount` and disables swizzling when reaching 0.

### Review checklist

- [x] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionSwizzler.swift
+++ b/DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionSwizzler.swift
@@ -11,7 +11,7 @@ internal class URLSessionSwizzler {
     ///
     /// This value will increment for each call to the `bind()` method.
     /// Calling `unbind()` will decrement the count, when reaching zero, the swizzler is disabled.
-    internal private(set) static var bindingsCount: Int = 0
+    internal private(set) static var bindingsCount: UInt = 0
     /// The binding lock.
     private static var lock = NSLock()
     /// `URLSession.dataTask(with:completionHandler:)` (for `URLRequest`) swizzling.


DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionSwizzler.swift
```diff
@@ -46,11 +51,16 @@ internal class URLSessionSwizzler {
...
+        defer { lock.unlock() }
+
+        if bindingsCount == 0 {
             return
         }
 
-        bindingsCount -= 1
+        if bindingsCount > 1 {
+            return bindingsCount -= 1
+        }
```
maxep, line 63: We had a wrong `guard` statement leading to unwanted unswizzling.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionSwizzler.swift
```diff
@@ -12,6 +12,8 @@ internal class URLSessionSwizzler {
     /// This value will increment for each call to the `bind()` method.
     /// Calling `unbind()` will decrement the count, when reaching zero, the swizzler is disabled.
     internal private(set) static var bindingsCount: Int = 0
+    /// The binding lock.
+    private static var lock = NSLock()
```
maxep, line 16: `.bind()` and `.unbind()` will acquire a lock to support concurrent calls.

ncreated (approved):
Looks good 👍👌 

The problem could be avoided in the future if we add tests integrity check for `bindingCount == 0`. We should leverage these checks for ensuring clean state of any globals (I found it extremely helpful in several battles against flakiness): https://github.com/DataDog/dd-sdk-ios/blob/58077d6e85681ac0e6f60f67e3ba07e4f8385a19/Tests/DatadogTests/TestsObserver/DatadogTestsObserver.swift#L24-L25

ncreated:
> Good call @ncreated ! I thought `activeSwizzlingNames.isEmpty` would cover that but it did not in that case, let me add that check 👍

Interestingly, it didn't cover it because (I guess) the problem was in "over unbinding" 💡.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionSwizzler.swift
```diff
@@ -22,6 +24,9 @@ internal class URLSessionSwizzler {
     internal private(set) static var dataTaskWithURL: DataTaskWithURL?
 
     static func bind() throws {
+        lock.lock()
+        defer { lock.unlock() }
+
         guard bindingsCount == 0 else {
             return bindingsCount += 1
```
ganeshnj, line 31: this statement looks so odd, the method doesn't return but still has a returning value.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionSwizzler.swift
```diff
@@ -12,6 +12,8 @@ internal class URLSessionSwizzler {
     /// This value will increment for each call to the `bind()` method.
     /// Calling `unbind()` will decrement the count, when reaching zero, the swizzler is disabled.
     internal private(set) static var bindingsCount: Int = 0
+    /// The binding lock.
```
ganeshnj, line 15: non blocking

we should either make it `UInt` as we never want it to be negative by a developer error.

Or log something when`unbind` > `bind` calls. There has to something wrong in this case IMO.

ganeshnj (approved):

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionSwizzler.swift
maxep: The operator and the `bind` both return `Void` :)

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionSwizzler.swift
maxep: Good point, I will make it `UInt` 👍

ganeshnj:
Nice finding @maxep, all checks are 🟢 .

PR #1270, opened 2023-04-27
[v2] Improve CI Stability

maxep:
### What and why?

Fix further flakiness:
1. `NetworkInstrumentationFeatureTests` needs flush of the Feature before assertion.
2. `RUMResourcesScenarioTests` was using an external 3rd party service that we don't own.
3. Bad dependency linking in `DatadogLogs iOS` target 

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/DatadogInternal/Tests/NetworkInstrumentation/NetworkInstrumentationFeatureTests.swift
+++ b/DatadogInternal/Tests/NetworkInstrumentation/NetworkInstrumentationFeatureTests.swift
@@ -240,23 +240,20 @@ class NetworkInstrumentationFeatureTests: XCTestCase {
     // MARK: - URLRequest Interception
 
     func testGivenOpenTracing_whenInterceptingRequests_itInjectsTrace() throws {
-        let notifyInterceptionStart = expectation(description: "Notify interception did start")
-
         // Given
         var request: URLRequest = .mockWith(url: "https://test.com")
         let writer = HTTPHeadersWriter(sampler: .mockKeepAll())
         handler.firstPartyHosts = .init(["test.com": [.datadog]])
-        handler.onInterceptionStart = { _ in notifyInterceptionStart.fulfill() }
 
         // When
         writer.write(traceID: .mock(1), spanID: .mock(2))
         request.allHTTPHeaderFields = writer.traceHeaderFields
 
         let task: URLSessionTask = .mockWith(request: request, response: .mockAny())
-        let feature = core.get(feature: NetworkInstrumentationFeature.self)
-        feature?.intercept(task: task, additionalFirstPartyHosts: nil)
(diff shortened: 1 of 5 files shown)


IntegrationTests/IntegrationScenarios/Scenarios/RUM/RUMResourcesScenarioTests.swift
```diff
@@ -110,7 +110,12 @@ class RUMResourcesScenarioTests: IntegrationTests, RUMCommonAsserts {
...
         // Get RUM Sessions with expected number of View visits and Resources
-        let rumRequests = try rumServerSession.pullRecordedRequests(timeout: dataDeliveryTimeout) { requests in
+        //
+        // RUMM-3151: Timeout of 30s makes the test flaky, increasing to over 30s
+        // allow the session to end. I suspect the view event debounce mechanism
+        // to postpone the end of the session when a resource completes after the
+        // stop of its parent view.
+        let rumRequests = try rumServerSession.pullRecordedRequests(timeout: dataDeliveryTimeout * 2) { requests in
             try RUMSessionMatcher.singleSession(from: requests)?.hasEnded() ?? false
         }
```
maxep, line 120: @ncreated I think that the reason for our flakiness on `develop` as well.

The `try RUMSessionMatcher.singleSession(from: requests)?.hasEnded()` stays false for more than 30s because there is an active view. After 30s, we get a request with a view update making the view inactive.

I think it's due to our debounce mechanism, it seems that the view update is sent after 30s because the view was previously stopped but with a pending resource 🤔

maxep: That wasn't it
image

IntegrationTests/Runner/Scenarios/URLSession/URLSessionScenarios.swift
```diff
@@ -214,7 +214,7 @@ class URLSessionBaseScenario: NSObject {
         }
 
         return URLSession(
-            configuration: .default,
+            configuration: .ephemeral,
```
maxep, line 217: Use `.ephemeral` to prevent using cache.

DatadogInternal/Tests/NetworkInstrumentation/NetworkInstrumentationFeatureTests.swift
```diff
@@ -15,7 +15,7 @@ class NetworkInstrumentationFeatureTests: XCTestCase {
     // swiftlint:enable implicitly_unwrapped_optional
 
     override func setUpWithError() throws {
-        super.setUp()
+        try super.setUpWithError()
```
maxep, line 18: 🤦

ganeshnj (approved):

PR #1271, opened 2023-04-28
[develop] Improve CI Stability

maxep:
### What and why?

`RUMResourcesScenarioTests` was using `https://httpbin.org` for 3rd party resources. We don't own this service and it could fail to respond in a timely manner.

Using `curl -s https://httpbin.org/get` we can have long response time:
```
    time_namelookup:  0.005084
       time_connect:  0.210384
    time_appconnect:  0.624901
   time_pretransfer:  0.414942
      time_redirect:  0.000000
 time_starttransfer:  31.134719
                    ----------
         time_total:  31.135052
```
This integration test now uses `shopist.io` that we own for 3rd party requests, it can be considered more reliable.

### How?

1. Use `https://shopist.io/categories.json` for GET and `https://api.shopist.io/checkout.json` for POST as 3rd party resources.
2. Use an `.ephemeral` URLSession instance so we don't use local cache and we can get proper metrics.

### Review checklist

- [ ] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Tests/DatadogIntegrationTests/Scenarios/RUM/RUMResourcesScenarioTests.swift
+++ b/Tests/DatadogIntegrationTests/Scenarios/RUM/RUMResourcesScenarioTests.swift
@@ -61,9 +61,9 @@ class RUMResourcesScenarioTests: IntegrationTests, RUMCommonAsserts {
         let firstPartyBadResourceURL = URL(string: "https://foo.bar")!
 
         // Requesting this third party by the app should create the RUM Resource.
-        let thirdPartyGETResourceURL = URL(string: "https://httpbin.org/get")!
+        let thirdPartyGETResourceURL = URL(string: "https://shopist.io/categories.json")!
         // Requesting this third party by the app should create the RUM Resource.
-        let thirdPartyPOSTResourceURL = URL(string: "https://httpbin.org/post")!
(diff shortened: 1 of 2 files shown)


fuzzybinary (approved):
LGTM, but I'm wondering if we could make this not hit an actual endpoint, at least for the integration tests...

Tests/DatadogIntegrationTests/Scenarios/RUM/RUMResourcesScenarioTests.swift
```diff
@@ -61,9 +61,9 @@ class RUMResourcesScenarioTests: IntegrationTests, RUMCommonAsserts {
         let firstPartyBadResourceURL = URL(string: "https://foo.bar")!
 
         // Requesting this third party by the app should create the RUM Resource.
-        let thirdPartyGETResourceURL = URL(string: "https://httpbin.org/get")!
```
ganeshnj, line 64: /comment
Interesting, in past this has been quite reliable from experience.

maxep: Maybe it's too popular now!
As @fuzzybinary mentioned, it would be nice to not hit a remote endpoint, we could use the httpbin docker image to start the service locally!

PR #1272, opened 2023-05-03
RUMM-3222 Use targetTimestamp as reference to calculate FPS

ganeshnj:
### What and why?

Currently, to find out the FPS of a view, the SDK look for the frame duration and inverses it to get the FPS.

Eg. if the frame duration is 16ms (0.016s), the FPS is 1/0.016 = 60FPS (rounded).

This generally works but when comes to variable refresh rate displays, it becomes an issue. For example, let's say OS has decided to increase the frame duration to 32ms to save battery, the FPS will be 30FPS but that doesn't mean the app is having a poor UI performance. It can very well be that the app is doing nothing.

### How?

`CADisplayLink` has a property called `targetTimestamp` which is the time when the next frame is scheduled to be displayed. This is the time when the frame is actually displayed on the screen. Hence, the SDK uses this property to calculate the expected frame duration and then inverses it to get the FPS.

At the same time, SDK continues to use the old logic to calculate the FPS using the `timestamp` property.

Finally, the SDK normalizes the FPS to 60FPS as Datadog backend only supports 60FPS rate for now described [here](https://www.datadoghq.com/blog/monitor-mobile-vitals-datadog/#slow-rendering).

> The logic for constant refresh rate displays has not changed.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
+++ b/Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
@@ -75,10 +75,16 @@ internal class VitalRefreshRateReader: ContinuousVitalReader {
 
         if let lastFrameTimestamp = self.lastFrameTimestamp {
             let currentFrameDuration = provider.currentFrameTimestamp - lastFrameTimestamp
+            guard currentFrameDuration > 0 else {
+                return nil
+            }
             let currentFPS = 1.0 / currentFrameDuration
 
             // ProMotion displays (e.g. iPad Pro and newer iPhone Pro) can have refresh rate higher than 60 FPS.
             if let expectedCurrentFrameDuration = self.nextFrameDuration, provider.maximumDeviceFramesPerSecond > 60 {
+                guard expectedCurrentFrameDuration > 0 else {
+                    return nil
+                }
                 let expectedFPS = 1.0 / expectedCurrentFrameDuration
                 fps = currentFPS * (Self.backendSupportedFrameRate / expectedFPS)
             } else {
--- a/Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
+++ b/Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
@@ -14,6 +14,7 @@ internal class VitalRefreshRateReader: ContinuousVitalReader {
     private var displayLink: CADisplayLink?
     private var lastFrameTimestamp: CFTimeInterval?
     private var nextFrameDuration: CFTimeInterval?
+    private static var backendSupportedFrameRate = 60.0
 
     init(notificationCenter: NotificationCenter = .default) {
         notificationCenter.addObserver(
@@ -79,7 +80,7 @@ internal class VitalRefreshRateReader: ContinuousVitalReader {
             // ProMotion displays (e.g. iPad Pro and newer iPhone Pro) can have refresh rate higher than 60 FPS.
             if let expectedCurrentFrameDuration = self.nextFrameDuration, provider.maximumDeviceFramesPerSecond > 60 {
                 let expectedFPS = 1.0 / expectedCurrentFrameDuration
-                fps = currentFPS * (60.0 / expectedFPS)
+                fps = currentFPS * (Self.backendSupportedFrameRate / expectedFPS)
             } else {
                 fps = currentFPS
             }
--- a/Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
+++ b/Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
@@ -68,7 +68,7 @@ internal class VitalRefreshRateReader: ContinuousVitalReader {
             let currentFPS = 1.0 / currentFrameDuration
 
             // ProMotion displays (e.g. iPad Pro and newer iPhone Pro) can have refresh rate higher than 60 FPS.
-            if let expectedCurrentFrameDuration = self.nextFrameDuration, provider.maximumDeviceFramesPerSecond > 60 {
+            if let expectedCurrentFrameDuration = self.nextFrameDuration, provider.adaptiveFrameRateSupported {
                 guard expectedCurrentFrameDuration > 0 else {
                     return nil
                 }
@@ -139,6 +139,14 @@ internal protocol FrameInfoProvider {
 
     /// Maximum number of frames per second supported by the device
     var maximumDeviceFramesPerSecond: Int { get }
+}
+
+private let adaptiveFrameRateThreshold = 60
+extension FrameInfoProvider {
+    /// `true` if the device supports adaptive frame rate
+    var adaptiveFrameRateSupported: Bool {
+        maximumDeviceFramesPerSecond > adaptiveFrameRateThreshold
+    }
 }
 
 extension CADisplayLink: FrameInfoProvider {
--- a/Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
+++ b/Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
@@ -13,6 +13,7 @@ internal class VitalRefreshRateReader: ContinuousVitalReader {
 
     private var displayLink: CADisplayLink?
     private var lastFrameTimestamp: CFTimeInterval?
+    private var nextFrameDuration: CFTimeInterval?
 
     init(notificationCenter: NotificationCenter = .default) {
         notificationCenter.addObserver(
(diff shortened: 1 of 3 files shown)


Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
```diff
@@ -53,25 +54,47 @@ internal class VitalRefreshRateReader: ContinuousVitalReader {
...
 
+    internal func framesPerSecond(provider: FrameInfoProvider) -> Double? {
+        var fps: Double? = nil
+
+        if let lastFrameTimestamp = self.lastFrameTimestamp {
+            let currentFrameDuration = provider.currentFrameTimestamp - lastFrameTimestamp
+            let currentFPS = 1.0 / currentFrameDuration
+
+            // ProMotion displays (e.g. iPad Pro and newer iPhone Pro) can have refresh rate higher than 60 FPS.
+            if let expectedCurrentFrameDuration = self.nextFrameDuration, provider.maximumDeviceFramesPerSecond > 60 {
```
maciejburda, line 80: We could export `60` to some property called i.e `maximumBackendSupportedFrameRate` just to remember the reason in the code

maciejburda (approved):
LGTM!

One tiny suggestion, and one question about expanding the support for 60+ FPS on the backend.

Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
```diff
@@ -53,25 +54,47 @@ internal class VitalRefreshRateReader: ContinuousVitalReader {
...
         }
-        lastFrameTimestamp = link.timestamp
     }
 
+    internal func framesPerSecond(provider: FrameInfoProvider) -> Double? {
+        var fps: Double? = nil
+
+        if let lastFrameTimestamp = self.lastFrameTimestamp {
+            let currentFrameDuration = provider.currentFrameTimestamp - lastFrameTimestamp
+            let currentFPS = 1.0 / currentFrameDuration
```
ncreated, line 77: nit/ I wonder if this could end up with "division by 0" - so far it didn't, this logic was already there. However, because we're processing `CADisplayLink` values here, I think it could make sense to add sanity check as a principle (_arithmetics with external values_). In looks hard to make assumptions on these values in certain edge situations.

Tests/DatadogTests/Datadog/RUM/RUMVitals/VitalRefreshRateReaderTests.swift
```diff
@@ -97,4 +97,133 @@ class VitalRefreshRateReaderTests: XCTestCase {
...
+        // second frame recorded
+        frameInfoProvider.currentFrameTimestamp = 0.008
+        frameInfoProvider.nextFrameTimestamp = 0.033
+        let secondFps = reader.framesPerSecond(provider: frameInfoProvider)
+        XCTAssertEqual(secondFps, 60)
+
+        // third frame recorded
+        frameInfoProvider.currentFrameTimestamp = 0.043
+        let thirdFps = reader.framesPerSecond(provider: frameInfoProvider)
+        XCTAssertEqual(thirdFps, 42.85714285714286)
```
ncreated, line 221: flaky ❄️/ This looks flaky - we [use `accuracy:`](https://github.com/search?q=repo%3ADataDog%2Fdd-sdk-ios+accuracy&type=code) to avoid random failures on CI (happened in the past):
```suggestion
        XCTAssertEqual(thirdFps, 42.8571, accuracy: 0.0001)
```

Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
```diff
@@ -53,25 +54,47 @@ internal class VitalRefreshRateReader: ContinuousVitalReader {
         }
     }
 
-    // MARK: - Private
+    // MARK: - Internal
```
ncreated, line 57: nit/ Any reason to change following methods from `private` to `internal`? With `private` we were enforcing thread safety for the entire reader. Now it becomes conventional - all methods must be called on the same thread.

If this is for testability of `framesPerSecond(provider:)` then it seems a good compromise 👍. Even better would be to isolate FPS logic into separate unit that would be private to the reader, but internal for its own unit tests.

It's more a question on the reasoning, than change request 🙂.

ncreated (approved):
Looks good and it reads very well 💪👍. I left few suggestions and (perhaps), flakiness alert ❄️ ⚠️.

Tests/DatadogTests/Datadog/RUM/RUMVitals/VitalRefreshRateReaderTests.swift
ganeshnj: we shouldn't get flakiness on this as this is just pure math unless Swift changes their floating point representation.

I general, I agree but here approximation is not even required.

Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
ganeshnj: that's a fair a call, I will add some logic around divide by 0 case.

Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
ganeshnj: I was kind of expecting. It is just me being lazy and didn't want to move code around a lot. But I agree, Frame detector should be its own unit.

Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
```diff
@@ -53,25 +54,47 @@ internal class VitalRefreshRateReader: ContinuousVitalReader {
...
+        var fps: Double? = nil
+
+        if let lastFrameTimestamp = self.lastFrameTimestamp {
+            let currentFrameDuration = provider.currentFrameTimestamp - lastFrameTimestamp
+            let currentFPS = 1.0 / currentFrameDuration
+
+            // ProMotion displays (e.g. iPad Pro and newer iPhone Pro) can have refresh rate higher than 60 FPS.
+            if let expectedCurrentFrameDuration = self.nextFrameDuration, provider.maximumDeviceFramesPerSecond > 60 {
+                let expectedFPS = 1.0 / expectedCurrentFrameDuration
+                fps = currentFPS * (60.0 / expectedFPS)
```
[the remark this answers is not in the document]
ganeshnj: yeah, in ideal world, we would want to have a graph with two values on a time scale. 

Expected frame rate and real frame rate but we are not designed for this (as of now).

My plan to start a conversion with backed to allow sending metrics that have such type of measurement scale.

Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
ganeshnj: I just moved back to private and kept `framesPerSecond(provider:)` internal for tests. I think it is good compromise.

ncreated: Sounds good 👍 as we're touching existing code here (vs adding new code).

Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
```diff
@@ -53,25 +55,53 @@ internal class VitalRefreshRateReader: ContinuousVitalReader {
...
+
+        if let lastFrameTimestamp = self.lastFrameTimestamp {
+            let currentFrameDuration = provider.currentFrameTimestamp - lastFrameTimestamp
+            guard currentFrameDuration > 0 else {
+                return nil
+            }
+            let currentFPS = 1.0 / currentFrameDuration
+
+            // ProMotion displays (e.g. iPad Pro and newer iPhone Pro) can have refresh rate higher than 60 FPS.
+            if let expectedCurrentFrameDuration = self.nextFrameDuration, provider.maximumDeviceFramesPerSecond > 60 {
```
maciejburda, line 71: Nit - I think we can replace the 60 at the end of the if let as well 🙌

ganeshnj: good call, done.

Sources/Datadog/RUM/RUMVitals/VitalRefreshRateReader.swift
```diff
@@ -53,25 +55,53 @@ internal class VitalRefreshRateReader: ContinuousVitalReader {
...
+                fps = currentFPS * (Self.backendSupportedFrameRate / expectedFPS)
+            } else {
+                fps = currentFPS
+            }
+        }
+
+        self.lastFrameTimestamp = provider.currentFrameTimestamp
+        self.nextFrameDuration = provider.nextFrameTimestamp - provider.currentFrameTimestamp
+
+        return fps
```
maxep, line 85: /nit/suggestion

The nested statements make it difficult to catch, maybe we could flatten these by using a `defer`, e.g:
```swift
defer {
    self.lastFrameTimestamp = provider.currentFrameTimestamp
    self.nextFrameDuration = provider.nextFrameTimestamp - provider.currentFrameTimestamp
}

guard let lastFrameTimestamp = self.lastFrameTimestamp else {
    return nil
}

let currentFrameDuration = provider.currentFrameTimestamp - lastFrameTimestamp
guard currentFrameDuration > 0 else {
    return nil
}

let currentFPS = 1.0 / currentFrameDuration

// ProMotion displays (e.g. iPad Pro and newer iPhone Pro) can have refresh rate higher than 60 FPS.
guard provider.adaptiveFrameRateSupported, let expectedCurrentFrameDuration = self.nextFrameDuration else {
    return currentFPS
}

guard expectedCurrentFrameDuration > 0 else {
    return nil
}

let expectedFPS = 1.0 / expectedCurrentFrameDuration
return currentFPS * (Self.backendSupportedFrameRate / expectedFPS)
```

> If I got it right :)

maxep (approved):
Looks great!

Thanks for the CHANGELOG, tho it needs an extra line for linking the PR. I also left a minor suggestion, but it's good to merge as is 👍

PR #1273, opened 2023-05-03
RUMM-3266 fix: use arch conditionals to assert arch specific signal

ganeshnj:
### What and why?

On x86 a `fatalError()` results on a SIGILL signal but on an ARM machine it results in a `SIGTRAP`.

Hence, the test case fail on the ARM machine. In order to update the CI to ARM, this needs to be fixed.

### How?

Use preprocessor macros to check the architecture and then use the appropriate signal.

Tested on ARM mac under Rosetta and ARM simulators.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


maxep (approved):
Nice, thank you!

maciejburda (approved):
Should this change enable M1 on Bitrise?

ganeshnj:
> Should this change enable M1 on Bitrise?

Indeed, this was the failure I was aware which stopped us running on M1.

alai97 (approved):
Looks good for docs!

PR #1278, opened 2023-05-05
RUMM-3237 Introduce SDKInit command in order to force new session on SDK initialization

plousada:
Introduces the SDKInit command when RUM Monitor is created in order to guarantee all correlated logs and traces have a valid `session_id` matching the session they were create in, even if occurring before any user interaction.

### What and why?

### How?

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


Tests/DatadogTests/Datadog/RUMMonitorTests.swift
```diff
@@ -1005,6 +1005,29 @@ class RUMMonitorTests: XCTestCase {
...
+    // MARK: - Launching SDKInit Command
+    func testWhenInitializing_itSendsSDKInitCommand_thenSetsTheSessionId() throws {
+        //Given
+        let expectation = self.expectation(description: "sessionId is called")
+
+        let rum: RUMFeature = .mockWith(
+            configuration: .mockWith(
+                onSessionStart: { sessionId, isDiscarded in
+                    XCTAssertTrue(sessionId.matches(regex: .uuidRegex))
+                    expectation.fulfill()
```
ncreated, line 1017: suggestion/ This test is totally fine 👍, but it is very minimal for the behaviour we introduce.

As described in the PR:
> when RUM Monitor is created (...) all correlated logs and traces have a valid `session_id` matching the session they were create in, **even if occurring before any user interaction**.

from now on, we shouldn't require starting a RUM view in order to get session ID in logs or spans. This in turn sounds like a new behaviour worth testing, WDYT @plousada?

I could propose 2 ways of improving coverage:
- implicit/ Test that initializing `RUMMonitor` updates V2's [`DatadogContext`](https://github.com/DataDog/dd-sdk-ios/blob/b72076ebfa7b5295e1461e50dbc14705ca68132c/Sources/Datadog/DatadogInternal/Context/DatadogContext.swift#L104-L105) respectively (sets session id in [RUM baggage](https://github.com/DataDog/dd-sdk-ios/blob/b72076ebfa7b5295e1461e50dbc14705ca68132c/Sources/Datadog/DatadogInternal/Context/DatadogContext.swift#L104-L105)). This would assume that other tests assert logs and spans created with certain `DatadogContext`.
- explicit/ Test that after initialization of `RUMMonitor`, any log or span includes RUM session id.

Because we don't yet have tests for creating logs and spans from certain `DatadogContext` and this would have to be facilitated in V2 migration, the "implicit" way looks too complex for now.

The "explicit" way however is yet living V1 convention. We could modify or add similar tests so they no longer depend on starting the RUM view:

https://github.com/DataDog/dd-sdk-ios/blob/develop/Tests/DatadogTests/Datadog/LoggerTests.swift#LL542C10-L542C107

https://github.com/DataDog/dd-sdk-ios/blob/b72076ebfa7b5295e1461e50dbc14705ca68132c/Tests/DatadogTests/Datadog/TracerTests.swift#LL700C12-L700C110

ncreated (approved):
It looks okay 👍👌, but proposed test seems very minimal. I left non-blocking suggestion.

Tests/DatadogTests/Datadog/RUMMonitorTests.swift
plousada: Makes sense, totally agree with adding the new coverage.

ncreated (approved):
🚀

maxep (approved):
LGTM! just a minor change before merging 👍

PR #1283, opened 2023-05-11
RUMM-3277 Prevent sending RUM Error from Cross Platform crash logs

louiszawadzki:
### What and why?

In the case of a Cross Platform crash log, a RUM error is already sent. 
We want to prevent this log from triggering another RUM error, which would be a duplicate.

With the change we only see 1 RUM Error:
image

### How?

Look at a cross-platform attribute before creating the log to prevent creating the error.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Tests/DatadogTests/Datadog/Logging/RemoteLoggerTests.swift
+++ b/Tests/DatadogTests/Datadog/Logging/RemoteLoggerTests.swift
@@ -1,0 +1,105 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import XCTest
+import TestUtilities
+@testable import Datadog
+
+class RemoteLoggerErrorMessageReceiver: FeatureMessageReceiver {
+    private var errors: [String] = []
+
+    /// Adds RUM Error with given message and stack to current RUM View.
+    func receive(message: FeatureMessage, from core: DatadogCoreProtocol) -> Bool {
+        guard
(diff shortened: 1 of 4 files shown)


Tests/DatadogTests/Datadog/Logging/RemoteLoggerTests.swift
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        self.errors.append(message)
+
+        return true
+    }
+
+    func getErrorMessages() -> [String] {
+        return self.errors
+    }
```
maxep, line 29: /nit you don't need this, `self.errors` is already exposed 👍

Tests/DatadogTests/Datadog/Logging/RemoteLoggerTests.swift
```diff
@@ -0,0 +1,105 @@
...
+import TestUtilities
+@testable import Datadog
+
+class RemoteLoggerErrorMessageReceiver: FeatureMessageReceiver {
+    var errors: [String] = []
+
+    /// Adds RUM Error with given message and stack to current RUM View.
+    func receive(message: FeatureMessage, from core: DatadogCoreProtocol) -> Bool {
+        guard
+            case let .error(message, attributes) = message
```
maxep:
```suggestion
            case let .error(message, _) = message
```

maxep (approved):
Thanks, Louis!

This patch is simple enough for a quick fix 👍  Now I think we should revisit Crash Reporting in RN by relying on the native crash reporter: Inject RN crash data before the native crash, instead of reporting the RN crash and preventing the native one to go through.
Maybe we could have a JIRA ticket referenced in the `errorLogIsCrash` comment?

PR #1293, opened 2023-05-30
RUMM-3222 Don't scale down refresh rate multiple times

ganeshnj:
### What and why?

During fixing the variable refresh rate calculation, the normalization was applied during the refresh rate determination phase, but the logic to normalize it again during the reporting was kept which resulted in double normalization.

Eg. First we normalize based on the logic of targetTimestamp and then divide by 2 for 120Hz display.

### How?

Remove second normalization phase during the reporting.

### What and why?

### How?

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


maxep (approved):
LGTM 👍

PR #1299, opened 2023-06-02
REPLAY-1616 Add SPM tool for managing SR snapshot files

ncreated:
### What and why?

🧰 This PR introduces `sr-snapshots` CLI tool for managing `*.png` files in Session Replay snapshot testing.

It fills the gap left intentionally in https://github.com/DataDog/dd-sdk-ios/pull/1162 - where `*.png` files weren't committed to this repo due to anticipating its growing size. Because of this, we were not able to compare snapshots in regular CI jobs, adding more effort to local development. Now we solve this problem with introducing `sr-snapshots` CLI.

Full context on why developing custom solution instead of using Git LFS can be read in this internal [RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/2923201974/RFC+-+Managing+SR+Snapshot+Files).

### How?

In agreed model, `*.png` files will be stored in separate snapshots repo as regular git files. The `dd-sdk-ios` will store only lightweight pointers that reference them and `sr-snapshots` CLI will be used to synchronise both repos.

Example of "synced" state:
```
# dd-sdk-ios:

- png/
   - .gitignore      - ignores all *.png files
   - foo.png
   - bizz.png
- pointers/
   - foo.png.json    - stores SHA256 hash of foo.png content
   - bizz.png.json   - stores SHA256 hash of bizz.png content


# dd-mobile-session-replay-snapshots:

- ios/
   - <hash of foo.png>.png      - the content of foo.png file
   - <hash of bizz.png>.png.    - the content of bizz.png file
```
"Local files" (from `dd-sdk-ios` repo) are linked to "remote files" (in `dd-mobile-session-replay-snapshots`) through "pointer files" with following rules:
* in local repo:
  * all PNGs are git-ignored,
  * each PNG has a corresponding pointer file (tracked in git),
  * each pointer file has the same name as the original PNG,
  * each pointer file stores the hash (SHA256) value of the original PNG file,
* in remote repo:
  * all PNG files are tracked in git,
  * each PNG file is named by the hash value of original local file,
  * each PNG file has the content of original local file.

Two commands are available in `sr-snapshots` CLI:
* `push` - to re-generate pointer files (from pngs) and push new files to snapshots repo,
* `pull` - to re-generate png files (from pointers) and pull files from snapshots repo.

#### Implementation details

The `sr-snapshots` CLI is a Swift Package. To enable code reuse in other similar tools, it is made of 4 separate targets:
* `SRSnapshotsCore` - implements `push` and `pull` commands
* `Files` - basic abstraction over file system: directory and files
* `Shell` - basic wrapper over `Process` APIs for calling shell commands from Swift
* `Git` - very naive implementation of git client, tailored to our needs in dd-sdk-ios

Both `Files`, `Shell` and `Git` duplicate portions of logic implemented for other tools in `dd-sdk-ios/tools/*` (in Python). By extracting these targets into separate packages, we can migrate all tooling to Swift over time.

#### Snapshots repo

The [dd-mobile-session-replay-snapshots](https://github.com/DataDog/dd-mobile-session-replay-snapshots) repo is currently internal, but we can think of making it open source in the future. An example structure after pushing all existing snapshots can be viewed on [this branch](https://github.com/DataDog/dd-mobile-session-replay-snapshots/tree/test2).

#### Next

To keep this PR relatively small, `SRSnapshotTests` project is yet not configured to leverage this new design. Remaining work of plugging it all together and enabling snapshots comparison on CI will be delivered in next PR.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


tools/sr-snapshots/Sources/Files/Directory.swift
```diff
@@ -0,0 +1,96 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+/// Enables operations on real directory in file system.
+public struct Directory: DirectoryProtocol {
```
ncreated, line 10: ⚠️ There is a visible room for improvement in implementation of some methods of this construct (e.g. transactions or I/O failure resilience). I kept it bare simple and quite naive. If seeking for more excellence, I'd rather switch to using 3rd party (e.g. [JohnSundell/Files](https://github.com/JohnSundell/Files)) for most operations in this file.

tools/sr-snapshots/Sources/Git/GitClient.swift
```diff
@@ -0,0 +1,113 @@
...
+    public func cloneIfNeeded(to directory: URL) throws {}
+    public func pull() throws {}
+    public func commit(message: String) throws {}
+    public func push() throws {}
+}
+
+/// Naive implementation of basic git client.
+///
+/// It executes git commands directly in shell and requires `gh`CLI to be preinstalled on host (https://cli.github.com/)
+/// and authorised for cloning private repositories.
```
ncreated, line 33: The requirement of [`gh` CLI](https://cli.github.com/) might look quite hard, but that tool is already [pre-installed](https://github.com/DataDog/dd-sdk-ios/blob/cd8724314d49db8ed3f755560811bee2ea69ae83/Makefile#LL61C1-L61C47) and authorised for internal repo clones in our CI. It is also leveraged for few other automations, e.g. clone [datadog-ios](https://github.com/DataDog/dd-sdk-ios/blob/cd8724314d49db8ed3f755560811bee2ea69ae83/tools/distribution/src/dogfood/repository.py#L25-L38) for dogfooding.

tools/sr-snapshots/Sources/SRSnapshotsCore/Commands.swift
```diff
@@ -0,0 +1,110 @@
...
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Git
+import Files
+import Foundation
+import ArgumentParser
+
+/// Git clone ssh for snapshots repo.
+private let ssh = "git@github.com:DataDog/dd-mobile-session-replay-snapshots.git"
```
ganeshnj, line 13: question

how would cloning via ssh work in CI? Are credentials baked in the CI environment?

tools/sr-snapshots/Sources/Shell/ShellCommand.swift
```diff
@@ -0,0 +1,80 @@
...
+    } else if let error = result.error, !error.isEmpty {
+        return error
+    } else {
+        return ""
+    }
+}
+
+/// Executes given shell command.
+@discardableResult
+public func shellResult(_ command: String) throws -> ShellResult {
```
ganeshnj, line 49: optional

you can reduce the string processing by observing the notification. For reference https://github.com/DataDog/apple-symbols-harvester/blob/ganeshnj/feat/RUM-3299-firmware-extraction/Sources/HarvesterFirmware/CommandLine.swift

ganeshnj: One benefit you get is async printing of the output. In current version, the output is printed at the end of execution which is a great experience when longer operations are in play.

tools/sr-snapshots/Sources/SRSnapshotsCore/Utils/Hashing.swift
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+import Foundation
+import CommonCrypto
+
+internal protocol Hashing {
+    func hash(from data: Data) -> String
+}
+
+internal struct SHA256Hashing: Hashing {
```
ganeshnj, line 14: SHA256 for this is overkill along with taking a dependency on the external library, time taking.

SHA1 should work for our usecase as we want hash of given data.

ganeshnj: https://developer.apple.com/documentation/cryptokit/insecure/sha1

ganeshnj:
Very nice work 🎯 

Couple of questions and suggestions but overall looks very neat and nothing blocking.

tools/sr-snapshots/Sources/SRSnapshotsCore/Commands.swift
ncreated: This is not the nicest solution, but the actual detail is implemented in `BasicGitClient` that relies on invoking shell commands directly. One of the commands uses [gh](https://cli.github.com/) which is an official GitHub CLI client. The [`gh repo clone`](https://cli.github.com/manual/gh_repo_clone) is capable of cloning any internal repo [using ENV authentication](https://cli.github.com/manual/gh_auth_login) (`GH_TOKEN`).

This setup already exists in our repo as I explained in https://github.com/DataDog/dd-sdk-ios/pull/1299#discussion_r1214541130.

Historically, we found `gh repo clone` easier than provisioning CI with SSH keys.

tools/sr-snapshots/Sources/Shell/ShellCommand.swift
ncreated: I like it, will do 👍

tools/sr-snapshots/Sources/SRSnapshotsCore/Utils/Hashing.swift
ncreated: Agree, we don't need secure hashing 👍 and SHA-1 could be better choice. I'll use it 👌

maxep (approved):
Awesome work! Very clear, it reads very well 🙏 
Agreed with @ganeshnj comments, once those are addressed it good to merge 🚀

tools/sr-snapshots/Sources/Shell/ShellCommand.swift
ncreated: Done ✅ - I used your implementation @ganeshnj as it was more versatile than mine attempt.

tools/sr-snapshots/Sources/SRSnapshotsCore/Utils/Hashing.swift
ncreated: Done ✅

tools/sr-snapshots/Sources/Git/GitClient.swift
```diff
@@ -0,0 +1,123 @@
...
+    public func pull() throws {}
+    public func commit(message: String) throws {}
+    public func push() throws {}
+}
+
+/// Naive implementation of basic git client.
+///
+/// It executes git commands directly in shell and requires `gh` CLI to be preinstalled on host (https://cli.github.com/)
+/// and authorised for cloning private repositories.
+public class BasicGitClient: GitClient {
```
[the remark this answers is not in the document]
ncreated: Good idea 👌, done ✅!

ganeshnj (approved):

maciejburda (approved):

PR #1305, opened 2023-06-07
RUMM-2985 Create Core Registry

maxep:
### What and why?

Create a registry of SDK instances to support multiple core.

### How?

Thread-safety of the registery is ensured by a read-write lock.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/DatadogInternal/Sources/CoreRegistry.swift
+++ b/DatadogInternal/Sources/CoreRegistry.swift
@@ -1,0 +1,71 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+/// A Registry for all core instances, allowing Features to retrieve the one
+/// they want from anywhere.
+public final class CoreRegistry {
+    /// Returns the default core instance if registered, `NOPDatadogCore` instance otherwise.
+    public static var `default`: DatadogCoreProtocol {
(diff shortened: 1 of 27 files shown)


DatadogInternal/Sources/CoreRegistry.swift
```diff
@@ -0,0 +1,71 @@
...
+        instances[defaultInstanceName] ?? NOPDatadogCore()
+    }
+
+    /// The name for the default core instance.
+    ///
+    /// Features should use this name as default parameter.
+    public static let defaultInstanceName = "main"
+
+    @ReadWriteLock
+    private static var instances: [String: DatadogCoreProtocol] = [:]
```
ganeshnj, line 23: question

why is this static?

Makes me think that CoreRegistry needs to instance based with a `sharedInstance`. Hence, whenever needed a shared instance can be dependency injected.

 By making `CoreRegistry` we are making an assumption that this will be shared across the target app.

maxep: That's a good point. The core instance is a best candidate for dependency injection but I don't see any reason for not allowing this 👍

maxep: In term of usage by Features, it is best to leave it as is! The injection argument is true for core instance but not for registry, so we can assume a single registry.

ganeshnj: The issue with such assumption is that, changing the interface would cause a breaking change.

Additionally, `static` methods makes testing harder if they are not pure methods.

If this is we agreed on already, I'm happy to ✅

PR #1306, opened 2023-06-09
Remove Debug RUM from Core

maxep:
### What and why?

Remove Debug RUM from core module

### How?

Move logic to RUM Monitor

### Review checklist
- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [ ] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes

### Custom CI job configuration (optional)
- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### Custom CI job configuration (optional)
- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

PR #1307, opened 2023-06-09
Use core registry in Datadog interface

maxep:
### What and why?

Follow up on #1306 

Use core registry in `Datadog` API.

### How?

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/Datadog.swift
+++ b/Sources/Datadog/Datadog.swift
@@ -110,11 +110,11 @@ public class Datadog {
         return CoreRegistry.default is DatadogCore
     }
 
-    /// Returns the core instance for the given name.
+    /// Returns the Datadog SDK instance for the given name.
     ///
     /// - Parameter name: The name of the instance to get.
     /// - Returns: The core instance if it exists, `NOPDatadogCore` instance otherwise.
-    public static func coreInstance(named name: String) -> DatadogCoreProtocol {
+    public static func sdkInstance(named name: String) -> DatadogCoreProtocol {
         CoreRegistry.instance(named: name)
     }
 
--- a/Sources/Datadog/Datadog.swift
+++ b/Sources/Datadog/Datadog.swift
@@ -110,6 +110,14 @@ public class Datadog {
         return CoreRegistry.default is DatadogCore
     }
 
+    /// Returns the core instance for the given name.
+    ///
+    /// - Parameter name: The name of the instance to get.
+    /// - Returns: The core instance if it exists, `NOPDatadogCore` instance otherwise.
+    public static func coreInstance(named name: String) -> DatadogCoreProtocol {
+        CoreRegistry.instance(named: name)
+    }
+
     /// Sets current user information.
     /// Those will be added to logs, traces and RUM events automatically.
     /// - Parameters:
@@ -121,9 +129,10 @@ public class Datadog {
         id: String? = nil,
         name: String? = nil,
         email: String? = nil,
-        extraInfo: [AttributeKey: AttributeValue] = [:]
+        extraInfo: [AttributeKey: AttributeValue] = [:],
+        in core: DatadogCoreProtocol = CoreRegistry.default
     ) {
-        let core = CoreRegistry.default as? DatadogCore
+        let core = core as? DatadogCore
         core?.setUserInfo(
             id: id,
             name: name,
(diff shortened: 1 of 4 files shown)


maxep:
wip: we are still discussing v2 public API

Sources/Datadog/Datadog.swift
```diff
@@ -125,6 +125,14 @@ public class Datadog {
         return CoreRegistry.default is DatadogCore
     }
 
+    /// Returns the core instance for the given name.
+    ///
+    /// - Parameter name: The name of the instance to get.
+    /// - Returns: The core instance if it exists, `NOPDatadogCore` instance otherwise.
+    public static func coreInstance(named name: String) -> DatadogCoreProtocol {
```
ncreated, line 132: question/ What is the counterpart of this API in `dd-sdk-android`?

Sources/Datadog/Datadog.swift
```diff
@@ -154,22 +163,23 @@ public class Datadog {
...
-        let core = CoreRegistry.default as? DatadogCore
+        let core = core as? DatadogCore
         core?.addUserExtraInfo(extraInfo)
     }
 
     /// Sets the tracking consent regarding the data collection for the Datadog SDK.
     /// - Parameter trackingConsent: new consent value, which will be applied for all data collected from now on
-    public static func set(trackingConsent: TrackingConsent) {
-        let core = CoreRegistry.default as? DatadogCore
+    public static func set(trackingConsent: TrackingConsent, in core: DatadogCoreProtocol = CoreRegistry.default) {
```
ncreated, line 175: question/ Do we do it the same way in `dd-sdk-android`? Static API that queries instance under the hood, not instance API that requires user to query the instance? Both are fine, I'm just making sure we don't diverge 👍.

Sources/Datadog/Datadog.swift
maxep: It's [`getInstance`](https://github.com/DataDog/dd-sdk-android/blob/feature/sdkv2/dd-sdk-android-core/src/main/kotlin/com/datadog/android/Datadog.kt#L138) on Android. I think this version is more explicit and swift-friendly, WDYT?

Sources/Datadog/Datadog.swift
maxep: I can't find it in Android, but this one complies with the convention. @0xnm How do you set the consent per core?

0xnm: yes, we set consent per core. In general, all the methods like `setTrackingConsent`, `setUserInfo`, etc. should be called on the instance directly, meaning the syntax will be `Datadog.getInstance().setTrackingConsent`.

The reason for that is that we are going to expose `Core` interface to the customer side (it is the instance returned by the `Datadog.getInstance`), and if we add a static method to set such things, than we would have both `Datadog.setTrackingConsent` and `Datadog.getInstance().setTrackingConsent` to do the same, which brings some ambiguity.

maxep: Thanks, @0xnm! The core interface does not expose consent and user info apis, it should be tackled in dedicated PR.

maxep: Everything declared [here](https://github.com/DataDog/dd-sdk-ios/blob/develop/Sources/Datadog/DatadogInternal/DatadogCoreProtocol.swift#L16). 

This `sdkInstance` is meant to pass the instance to static methods, like:
```swift
Logs.enable(in: Datadog.sdkInstance(named: "custom-instance"))
```
This variant allow to do this:
```swift
Datadog.set(trackingConsent: .granted)
// or
Datadog.set(trackingConsent: .granted, in: Datadog.sdkInstance(named: "custom-instance"))
```

Which complies with the approach we've discussed and aligns with other static methods across modules.

ncreated: @maxep @0xnm what is the consensus here? I consider this a blocker unless we align.

ncreated (changes requested):
Waiting until we decide on which way to align `.set(trackingConsent)` and `.addUserExtraInfo()` and `clearAllData()` with `dd-sdk-android`.

Sources/Datadog/Datadog.swift
maxep: We currently can't align without changing the `DatadogCoreProtocol`, I consider this outside of the scope of this PR.
We have [RUMM-3295](https://datadoghq.atlassian.net/browse/RUMM-3295) to align 👍

ncreated: 👌 sounds good, thanks 🙂 !

ncreated (approved):
🏅 🚀

PR #1309, opened 2023-06-15
REPLAY-1682 Add Session Replay Sampling

maciejburda:
### What and why?

The change adds a new feature to Session Replay that allows for sampling of replays. Now there is additional session replay sampling rate configuration available, with default 0 (no sampling). It's important because it allows for more flexibility in managing resources, as not all replays may need to be captured for every user.

It should be noted that Session Replay requires RUM to be enabled in order to work. If RUM is not enabled, Session Replay will also not work. Additionally, since RUM can have its own sampling, the overall Session Replay sampling rate is a multiplication of both samplings. For example, if RUM sampling is set to 50% and Session Replay sampling is set to 50%, the effective sampling rate for Session Replay would be 25%.

### How?

To introduce this feature, a new object called `RecordingCoordinator` was added. This object is responsible for getting the RUM context, randomising the sampling rate, and starting/stopping the recording scheduler as needed. By doing so, it provides a convenient and reliable way to manage the sampling rate for Session Replay.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


DatadogSessionReplay/Sources/Drafts/SessionReplay.swift
```diff
@@ -44,12 +44,6 @@ public struct SessionReplay {
 /// A draft interface of SR controller.
 /// TODO: RUMM-2268 Design convenient public API
 public protocol SessionReplayController {
-    /// Start recording.
-    func start()
-
-    /// Stop recording.
-    func stop()
-
```
maciejburda, line 52: Not needed anymore as it's internally managed by the coordinator.

DatadogSessionReplay/Sources/Recorder/Recorder.swift
```diff
@@ -128,13 +129,10 @@ internal class Recorder: Recording {
     /// **Note**: This is called on the main thread.
     private func captureNextRecord() {
         do {
-            guard let rumContext = currentRUMContext else {
-                // The RUM context was not yet received or current RUM session is not sampled.
```
maciejburda, line 132: Comment irrelevant as the guard only acts as unwrap.

DatadogSessionReplay/Tests/Recorder/RecorderTests.swift
```diff
@@ -8,42 +8,32 @@ import XCTest
...
-        self.onNew = notify
-    }
-
-    func notify(rumContext: RUMContext?) {
-        queue?.run { self.onNew?(rumContext) }
-    }
-}
-
 class RecorderTests: XCTestCase {
+    private var recorder: Recorder?
```
ncreated, line 12: question/ Why this convention 🤔💭? Storing SUT on `XCTestCase` property is wrong for components that manage global state between their `init()` and `deinit()` (such would have to be orchestrated with `setUp()` and `tearDown()`). It is not the case for `Recorder` today, but we're introducing something that is not scalable.

Sources/Datadog/Core/Utils/Sampler.swift
```diff
@@ -7,17 +7,17 @@
 import Foundation
 
 /// Sampler, deciding if events should be sent do Datadog or dropped.
-internal struct Sampler {
+public struct Sampler {
```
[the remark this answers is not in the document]
ncreated: It is OK, `Sampler` is [already public in v2 branch](https://github.com/DataDog/dd-sdk-ios/blob/feature/v2/DatadogInternal/Sources/Utils/Sampler.swift).

DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
```diff
@@ -0,0 +1,42 @@
...
+/// starting/stopping the recording scheduler as needed and propagating `has_replay` to other features.
+internal class RecordingCoordinator: RecordingCoordination {
+    /// Last received RUM context (or `nil` if RUM session is not sampled).
+    /// It's synchronized through `scheduler.queue` (main thread).
+    private(set) var currentRUMContext: RUMContext? = nil
+
+    init(
+        scheduler: Scheduler,
+        rumContextObserver: RUMContextObserver,
+        contextPublisher: SRContextPublisher,
```
ncreated, line 24: nit/ How about aligning the convention here:
```diff
rumContextObserver
- contextPublisher
+ srContextPublisher
```
It would make it cleaner: "observing RUM context (in SR)" + "publishing SR context (to RUM)".

DatadogSessionReplay/Sources/Processor/Processor.swift
```diff
@@ -75,8 +75,8 @@ internal class Processor: Processing {
 
         var records: [SRRecord] = []
         // Create records for describing UI:
-        if viewTreeSnapshot.rumContext != lastSnapshot?.rumContext {
-            // If RUM context has changed, new segment should be started.
+        if viewTreeSnapshot.rumContext.ids != lastSnapshot?.rumContext.ids {
```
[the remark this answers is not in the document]
ncreated: blocker/ Great catch ⚾! However, we must have test coverage for this in the first place. Lack of it was the root cause of the problem here.

DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
```diff
@@ -0,0 +1,42 @@
...
+                let isSampled = sampler.sample() && rumContext != nil
+                contextPublisher.setRecordingIsPending(isSampled)
+                if isSampled {
+                    scheduler.start()
+                } else {
+                    scheduler.stop()
+                }
+            }
+            self?.currentRUMContext = rumContext
+        }
```
ncreated, line 40: question/ The `scheduler` is not started initially and it will be started with the first change to RUM context. Will it work no matter of the order of enabling RUM and SR?

Imagine this situation:
```swift
RUM.enable()
RUMMonitor.shared().startView() // changes RUM context

SR.enable()
// ... recording is not started
// ... it will be only started with next change to RUM context
```
We seem to introduce logical race condition here, no? Ideally, we should be able to check the initial RUM context and perform initial `scheduler.start()`. This however is not yet doable in V2 - we lack an API to read current SDK context, we can only observe it and (I believe) observation doesn't emit initial value.

DatadogSessionReplay/Tests/Recorder/RecorderTests.swift
```diff
@@ -8,42 +8,32 @@ import XCTest
...
             rumContextObserver: rumContextObserver,
             uiApplicationSwizzler: .mockAny(),
-            scheduler: TestScheduler(numberOfRepeats: numberOfSnapshots),
+            scheduler: scheduler,
+            recordingCoordinator: RecordingCoordinationMock(currentRUMContext: .mockRandom()),
             viewTreeSnapshotProducer: ViewTreeSnapshotProducerMock(succeedingSnapshots: mockViewTreeSnapshots),
             touchSnapshotProducer: TouchSnapshotProducerMock(succeedingSnapshots: mockTouchSnapshots),
             snapshotProcessor: processor
         )
         rumContextObserver.notify(rumContext: .mockAny())
```
ncreated, line 33: change request/ This is not needed anymore, `Recorder` doesn't depend on `rumContextObserver`. This could be an argument towards the small refactor I suggested earlier - we loose clarity on what is required `Recorder` dependency.

ncreated (changes requested):
Well done 👌 - it reads very well, especially coordinator's tests 👍. There are few important questions I asked, and some blockers.

On top of inline review, we need to update `SRSnapshotTests` accordingly, because its build is now broken:

Screenshot 2023-06-16 at 12 42 31

DatadogSessionReplay/Tests/Recorder/RecorderTests.swift
maciejburda: While taking the responsibilities away from the recorder it made this test of tests very implicit. This var was introduce to just hold the reference to the recorder otherwise tests were failing due to deallocation issues. I added tearDown to make sure we nullify SUT.

DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
maciejburda: Technically order matters. If RUM starts publishing on message bus before SR gets initialized it could miss some messages. I would say - it's by design currently. We could think about potential direct dependency between RUM and SR. Personally I'm in favour of decoupling with risk above in mind.

In reality it gets frequently updated, so we should "catch" RUM context fairly quickly regardless the order.

maciejburda:
@ncreated Please give it another shot when you have a chance.

Main changes are around different context management. Now we allow viewID-less context and we act accordingaly. We only stop schedulers when we receive `nil` context from RUM, which means it's sampled out.

DatadogSessionReplay/Sources/Processor/Processor.swift
```diff
@@ -75,8 +75,10 @@ internal class Processor: Processing {
 
         var records: [SRRecord] = []
         // Create records for describing UI:
-        if viewTreeSnapshot.rumContext != lastSnapshot?.rumContext {
-            // If RUM context has changed, new segment should be started.
+        if viewTreeSnapshot.context.applicationID != lastSnapshot?.context.applicationID ||
+            viewTreeSnapshot.context.sessionID != lastSnapshot?.context.sessionID ||
+            viewTreeSnapshot.context.viewID != lastSnapshot?.context.viewID {
```
maciejburda, line 80: Now it's very much explicit + added the test.

DatadogSessionReplay/Tests/Mocks/TestScheduler.swift
```diff
@@ -19,6 +19,8 @@ internal class TestScheduler: Scheduler {
     /// Queue to execute operations on.
     let queue: Queue
 
+    var isRunning = false
```
ncreated, line 22: Because `start()` / `stop()` and `isRunning` (getter) are called from different threads in tests, we're running into data race that will be causing flaky tests in long run:

Screenshot 2023-06-20 at 09 59 52

ncreated (changes requested):

DatadogSessionReplay/Sources/DatadogCoreIntegration/RUMContextReceiver.swift
```diff
@@ -20,7 +20,7 @@ internal struct RUMContext: Decodable, Equatable {
         /// Current RUM session ID - standard UUID string, lowecased.
         let sessionID: String
         /// Current RUM view ID - standard UUID string, lowecased.
-        let viewID: String
+        let viewID: String?
```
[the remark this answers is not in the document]
maciejburda: When session is not sampled - we receive `nil` RUM context. Empty viewID happens when we load views (i.e. tapping at the tabbar item for the first time, or performing transition).

maciejburda:
@ncreated I added the refactor we discussed today. Please let me know your thoughts 🙌

DatadogSessionReplay/SRSnapshotTests/SRSnapshotTests/Utils/SnapshotTestCase.swift
```diff
@@ -37,7 +37,7 @@ internal class SnapshotTestCase: XCTestCase {
         let processor = Processor(queue: NoQueue(), writer: Writer())
         let recorder = try Recorder(
             configuration: configuration,
-            rumContextObserver: RUMContextObserverMock(),
+            recordingCoordinator: RecordingCoordinatorMock(),
             processor: processor,
             scheduler: scheduler
         )
```
ncreated, line 43: blocker/ It doesn't compile - we need to update snapshot tests. Also, we don't need `TestScheduler` anymore. BTW, I'm super happy to see how this will simplify ⭐ 🧽 👏.

ncreated (changes requested):
This looks great - so much simplification in `Recorder` 👏!

Before we go 🎬, we need to fix snapshot tests build.

DatadogSessionReplay/SRSnapshotTests/SRSnapshotTests/Utils/SnapshotTestCase.swift
```diff
@@ -36,23 +38,24 @@ internal class SnapshotTestCase: XCTestCase {
...
         var wireframes: [SRWireframe]?
 
         processor.interceptWireframes = {
             wireframes = $0
             expectation.fulfill()
         }
-
-        recorder.start()
         scheduler.triggerOnce()
-        recorder.stop()
```
ncreated, line 55: suggestion/ With `recorder.captureNextRecord()` being now internal, we can do much simpler:
```swift
// Set up wireframes interception and trigger recorder once:
var wireframes: [SRWireframe]?

processor.interceptWireframes = {
    wireframes = $0
    expectation.fulfill()
}
recorder.captureNextRecord(
    .init(privacy: configuration.privacy, applicationID: "", sessionID: "", viewID: "", viewServerTimeOffset: nil)
)
```
this way getting rid of `TestScheduler` and `RUMContextObserverMock`. Also this will solve the compiler warning:
```
Initialization of immutable value 'recordingCoordinator' was never used; consider replacing with assignment to '_' or removing it
```
Less dependencies here == less maintenance and broken builds 🙂.

ncreated (approved):
I left last suggestion - other than this, good to go 👍. Great effort 🚀!

PR #475, opened 2021-04-29
RUMM-1314: Support external resource timings

0xnm:
### What and why?

This change add support for the resource timings coming from external sources (like React Native). 

### How?

Nothing fancy, just an overload of `RUMMonitor#addResourceMetrics` with the necessary timings.

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference

PR #1310, opened 2023-06-15
RUMM-2922 Introduce `RUMMonitor` API for V2

ncreated:
### What and why?

📦 This is first PR extracting a new set of public RUM APIs for `2.0.0`.

The new API of RUM will consist of 3 main interfaces:
```swift
// Configuration:
let config: RUMConfiguration

// Feature:
RUM.enable(with: config)

// Monitor API:
RUMMonitor.shared() -> RUMMonitorProtocol
```
This PR only tackles `RUMMonitor`. Changes to `RUMConfiguration` and `RUM.enable(with:)` will come in next PRs.

### How?

First, the **size of this PR** is not because of inventing anything new, but rather moving old interface to another place and implementing agreed [public API convention](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3027638506/Public+API+Convention+in+iOS+SDK+2.x) to `RUMMonitor`. `RUMMonitor` defines 22 APIs and all are moved to dedicated protocol and duplicated in convenient protocol extension (both public). This solely gives around 1K of changes.

To simplify review, following is gist of most important changes.

#### Introduction of `RUMMonitorProtocol`

The `RUMMonitor.shared()` now returns a protocol, not base class as it was done in V1:
```swift
public class RUMMonitor {
   public static func shared(in:) -> RUMMonitorProtocol
}
```
Internally, this protocol has 2 implementations:
* `Monitor: RUMMonitorProtocol` - if SDK and RUM were enabled
* `NOPRUMMonitor: RUMMonitorProtocol` - if SDK or RUM were not enabled

#### Introduction of `extension RUMMonitorProtocol+Convenience`

Because some methods of `RUMMonitorProtocol` accept multiple optional parameters, convenient public extension is defined in `RUMMonitorProtocol+Convenience.swift`. It duplicates the `RUMMonitorProtocol` interface by providing default values.

#### Unification of RUM monitor API comments

In V1, RUM monitor APIs were added over long time, by many different contributors. This led to misalignments in the wording of public API comments. I used a chance to refresh all of them, finding a lot of opportunities for simplification.

#### Breaking changes compared to V1

List of breaking changes in RUM monitor API:
* action tracking APIs are now renamed - we now use _"action"_ instead of _"user action"_:
  * `addUserAction()` → `addAction()`
  * `startUserAction()` → `startAction()`
  * `stopUserAction()` → `stopAction()`
  * this is to align with `browser-sdk`: https://github.com/DataDog/browser-sdk/pull/939
* resource tracking APIs are renamed - we no longer use word _"Loading"_:
  * `startResourceLoading()` → `startResource()`
  * `stopResourceLoading()` → `stopResource()`
  * `stopResourceLoadingWithError()` → `stopResourceWithError()`
  * this is to align with `dd-sdk-android` where we never used _"Loading"_ phrase
* `errorMessage` parameter in `stopResourceWithError()` was renamed to `message` - for simplicity
* `source:stack:` parameters in `addError()` were reordered to `stack:source:` - because `stack` is more significant information than `source`
* the bloated API `addResourceMetrics(at:resourceKey:fetch:redirection:dns:connect:ssl:firstByte:download:responseSize:attributes:)` is now **removed** from public interface
   * it was added in https://github.com/DataDog/dd-sdk-ios/pull/475 and was made only for the purpose of DD React Native SDK
   * because now we better distinguish "extra cross-platform interface" (we separate it from "official public interface"), this variant was moved to `RUMMonitorProtocol+Internal`

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


tools/ci/choose_workflows.py
```diff
@@ -56,7 +56,7 @@ def should_run_integration_tests(ctx: CIContext) -> bool:
         build_env=context.build_env.DD_OVERRIDE_RUN_INTEGRATION_TESTS,
         pr_keyword='[x] Run integration tests',
         pr_path_prefixes=[
-            'IntegrationTests/IntegrationTests/',
+            'IntegrationTests/',
```
ncreated, line 59: Previous path was wrong and integration tests weren't triggered automatically (required selecting checkbox in PR description). Now this is fixed.

DatadogRUM/Sources/DDRUMMonitor+Internal.swift
```diff
@@ -1,49 +0,0 @@
-/*
- * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
- * This product includes software developed at Datadog (https://www.datadoghq.com/).
- * Copyright 2019-Present Datadog, Inc.
- */
-
-import Foundation
-import DatadogInternal
-
-extension DDRUMMonitor: DatadogInternalInterface {}
```
ncreated, line 10: @maxep FYI, I couldn't use `DatadogInternalInterface` for `RUMMonitorProtocol`, because it only works with classes. Decorating protocols this way seems much harder if not impossible.

To cut corners, I reused `DatadogInternalInterface` name in `RUMMonitorProtocol+Internal.swift`. IMO this shouldn't be a blocker for now - we can revisit it later.

Tests/DatadogTests/Datadog/LoggerTests.swift
```diff
@@ -28,10 +28,10 @@ class LoggerTests: XCTestCase {
         super.tearDown()
     }
 
-    /// Creates `RUMMonitor` instance for tests.
+    /// Creates `Monitor` instance for tests.
     /// The only difference vs. `RUMMonitor.initialize()` is that we disable RUM view updates sampling to get deterministic behaviour.
-    private func createTestableRUMMonitor(configuration: RUMConfiguration = .mockAny()) throws -> DDRUMMonitor {
-        let monitor = RUMMonitor(
+    private func createTestableRUMMonitor(configuration: RUMConfiguration = .mockAny()) throws -> Monitor {
+        let monitor = Monitor(
```
maciejburda, line 34: Why not RUMMonitor? I don't mind dropping RUM from names. In the end they are `DatadogRUM.Type`, but IMO we should be consistent.

tools/ci/choose_workflows.py
maciejburda: Great catch!

maciejburda (approved):
LGTM

Just one tiny question/suggestion on dropping the RUM prefix on monitor.

Tests/DatadogTests/Datadog/LoggerTests.swift
ncreated: Here we refer to `DatadogRUM.Monitor` which is internal type. The `DatadogRUM.RUMMonitor` is already reserved for appearing in the public API.
```swift
// in DatadogRUM:

public protocol RUMMonitorProtocol {}

public class RUMMonitor {
   func shared() -> RUMMonitorProtocol {
      return allGood ? Monitor() : NOPMonitor()
   }
}

internal class Monitor: RUMMonitorProtocol {}
internal class NOPMonitor: RUMMonitorProtocol {}
```
Because we can't have both, we need to chose. What I propose here is to promote `RUMMonitor` in public name as this is what our users are familiar with from V1. Does it make sense given this context 🤔?

maciejburda: ah ok! Makes sense now. I thought there's some deeper reason, just couldn't get it without this additional context. Thanks!

DatadogRUM/Sources/DDRUMMonitor+Internal.swift
maxep: Indeed, the `DatadogExtended` does not work on protocols sadly :( 

We can avoid adding a public interface to `DatadogRUM` module by doing this:
```swift
/// Extends `RUMMonitorProtocol` with additional methods designed for Datadog cross-platform SDKs.
public extension RUMMonitorProtocol {
    /// Grants access to an internal interface utilized only by Datadog cross-platform SDKs.
    /// **It is not meant for public use** and it might change without prior notice.
    var _internal: DatadogExtension<RUMMonitorProtocol> {
        DatadogExtension(self)
    }
}

extension DatadogExtension where ExtendedType: RUMMonitorProtocol {
  ...
}
```

Then each method of the extension has to cast the monitor:
```swift
extension DatadogExtension where ExtendedType: RUMMonitorProtocol {
    /// Adds RUM long task to current view.
    /// - Parameters:
    ///   - time: the time of this command in cross-platform SDK
    ///   - duration: duration of the long task
    ///   - attributes: attributes to process along with this call
    func addLongTask(
        at time: Date,
        duration: TimeInterval,
    ) {
        guard let monitor = type as? Monitor else {
            return
        }

        let longTaskCommand = RUMAddLongTaskCommand(time: time, attributes: attributes, duration: duration)
        monitor.process(command: longTaskCommand)
    }
}
```

I think it's preferable this way than adding a new public interface 👍

maxep (approved):
Looks great, thank you!

I've left a suggestion for the `_internal` extension 👍

DatadogRUM/Sources/DDRUMMonitor+Internal.swift
ncreated: I'm afraid it's not that easy. This ☝️ compiles in `DatadogRUM`, but fails to compile when used:

```swift
// ExampleAppDelegate.swift

import DatadogRUM

RUMMonitor.shared()._internal.addLongTask(at: Date(), duration: 0, attributes: [:])

// error: type 'any RUMMonitorProtocol' cannot conform to 'RUMMonitorProtocol'
//        RUMMonitor.shared()._internal.addLongTask(at: Date(), duration: 0, attributes: [:])

// note: only concrete types such as structs, enums and classes can conform to protocols
//        RUMMonitor.shared()._internal.addLongTask(at: Date(), duration: 0, attributes: [:])
//                            ^

// note: required by referencing instance method 'addLongTask(at:duration:attributes:)' on // 'DatadogExtension' where 'ExtendedType' = 'any RUMMonitorProtocol'
// public extension DatadogExtension where ExtendedType: RUMMonitorProtocol {
//       ^
```

DatadogRUM/Sources/RUMMonitorProtocol.swift
```diff
@@ -0,0 +1,354 @@
...
+    /// If enabled, a debugging outline will be displayed on top of the application, indicating the name of the active RUM view.
+    /// This can be helpful for debugging RUM instrumentation issues in your app.
+    ///
+    /// The default value is false.
+    var debug: Bool { set get }
+}
+
+// MARK: - NOP moniotor
+
+internal class NOPRUMMonitor: RUMMonitorProtocol {
```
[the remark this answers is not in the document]
ncreated: 👍 totally, done ✅

DatadogRUM/Sources/DDRUMMonitor+Internal.swift
maxep: damn... :(

ncreated: No big deal IMO - we can improve it later with no breaking change for cross-platform 👌.

maxep: Yes, I start to think that the `DatadogExtension` pattern is not well suited for mutations, especially in configuration where we aim to use `struct`...

PR #1311, opened 2023-06-17
Fix TracingUUID toString string format

changm4n:
### What and why?
There is a bug in the creation of the `traceparent` of the trace http header.
The type of `TracingUUID.rawValue` is `UInt64`, but it is converted to 32 bits when converted like [format: "%16x"].  
When `trace id : 7242571286695202638`(decimal), the `traceparent` header value is being generated as `[version]-00000000000000000000000063e94f4e-[parent id]-[trace flags]` now.

### How?

Use the `%llx` format when converting strings of type UInt64.

### Review checklist

- [x] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/Tracing/UUIDs/TracingUUID.swift
+++ b/Sources/Datadog/Tracing/UUIDs/TracingUUID.swift
@@ -40,9 +40,9 @@ internal struct TracingUUID: Equatable, Hashable {
         case .hexadecimal:
             return String(rawValue, radix: 16)
         case .hexadecimal16Chars:
-            return String(format: "%016x", rawValue)
+            return String(format: "%016llx", rawValue)
         case .hexadecimal32Chars:
-            return String(format: "%032x", rawValue)
+            return String(format: "%032llx", rawValue)
         }
     }
 }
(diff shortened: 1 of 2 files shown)


maciejburda (approved):

maciejburda:
Screenshot 2023-06-22 at 13 58 04

ncreated (approved):
Thanks a million @changm4n 🏅. We haven't heard of this causing a problem before, could you please describe what was the impact on your side?

changm4n:
@ncreated  The datadog agent on the backend was using the `trace id` in the request header received from the client, but the `trace id` sent to datadog from the client was different from the `trace id` sent to datadog from the backend, so it was being recognised as a different trace.

ncreated:
> @ncreated The datadog agent on the backend was using the `trace id` in the request header received from the client, but the `trace id` sent to datadog from the client was different from the `trace id` sent to datadog from the backend, so it was being recognised as a different trace.

Thank you @changm4n, that helps us prioritising release of this fix 👍

PR #1314, opened 2023-06-19
Prepare SPM for SessionReplay

maxep:
### What and why?

Prepare repo for including SessionReplay in SPM.

### How?

1. Remove SessionReplay dependency to `Datadog`
2. Add SessionReplay target to main `Package.swift` **without exposing the library**.
3. Fix main SPM tests

In order to keep `SessionReplay/Package.swift` available for development while doing 1. and 2., I've created symbolic links to `DatadogInternal` and `TestUtilities`.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [x] Run smoke tests

### DIFF
--- a/DatadogTrace/Sources/Utils/ActiveSpansPool.swift
+++ b/DatadogTrace/Sources/Utils/ActiveSpansPool.swift
@@ -57,7 +57,6 @@ internal class ActiveSpansPool {
         rlock.unlock()
     }
 
-#if DD_SDK_COMPILED_FOR_TESTING
     /// This explicit way of destroying `ActiveSpansPool` was added after noticing RUMM-2904. It is there to keep test coverage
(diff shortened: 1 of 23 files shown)


Package.swift
```diff
@@ -57,6 +37,10 @@ let package = Package(
             name: "DatadogWebViewTracking",
             targets: ["DatadogWebViewTracking"]
         ),
+//        .library(
```
maciejburda, line 40: Maybe we could tag it with a ticket?

maciejburda (approved):

DatadogTrace/Sources/Utils/ActiveSpansPool.swift
```diff
@@ -57,7 +57,6 @@ internal class ActiveSpansPool {
         rlock.unlock()
     }
 
-#if DD_SDK_COMPILED_FOR_TESTING
```
maxep, line 60: I haven't find a way to define compilation flags for testing only in SPM. This method is internal anyway.

PR #1315, opened 2023-06-20
RUMM-3349 Add context to crash when there's an active view

plousada:
### What and why?
We're currently not reporting user set global attributes when collecting a crash. 
This PR addresses this.

### How?

Context reported on the RUM Error created when collecting a crash is extracted from the last active RUM View.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/RUM/Integrations/CrashReportReceiver.swift
+++ b/Sources/Datadog/RUM/Integrations/CrashReportReceiver.swift
@@ -326,7 +326,7 @@ internal struct CrashReportReceiver: FeatureMessageReceiver {
             application: .init(id: lastRUMView.application.id),
             ciTest: lastRUMView.ciTest,
             connectivity: lastRUMView.connectivity,
-            context: nil,
+            context: lastRUMView.context,
             date: crashDate.timeIntervalSince1970.toInt64Milliseconds,
             device: lastRUMView.device,
             display: nil,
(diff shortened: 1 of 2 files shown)


maciejburda (approved):

Tests/DatadogTests/Datadog/RUM/Integrations/CrashReportReceiverTests.swift
```diff
@@ -498,6 +498,13 @@ class CrashReportReceiverTests: XCTestCase {
             && sendRUMErrorEvent.model.view.id == lastRUMViewEvent.view.id,
             "The `RUMErrorEvent` sent must be linked to the same RUM Session as the last `RUMViewEvent`."
         )
+
+        DDAssertJSONEqual(
+            AnyEncodable(sendRUMErrorEvent.context?.contextInfo),
+            AnyEncodable(lastRUMViewEvent.context?.contextInfo),
+            "The `RUMErrorEvent` sent must be include the context of the last `RUMViewEvent`."
+        )
```
ncreated, line 506: suggestion/ Are we sure both aren't `nil`, making this assertion not sufficient? We could consider adding `XCAssertNotNil` before or using `XCTUnwrap` to get non-optional.

ncreated (approved):
With one suggestion, looks good 👍

Sources/Datadog/RUM/Integrations/CrashReportReceiver.swift
```diff
@@ -326,7 +331,7 @@ internal struct CrashReportReceiver: FeatureMessageReceiver {
             application: .init(id: lastRUMView.application.id),
             ciTest: lastRUMView.ciTest,
             connectivity: lastRUMView.connectivity,
-            context: nil,
+            context: errorContext,
```
maxep, line 334: /suggestion I think we can pass the `lastRUMView.context` directly here, no?
```suggestion
            context: lastRUMView.context,
```

maxep (approved):
LGTM, I've left a small suggestion.

Sources/Datadog/RUM/Integrations/CrashReportReceiver.swift
plousada: Indeed, no need for the extra copy.

PR #1316, opened 2023-06-20
RUMM-2923 Logs Configuration

maxep:
### What and why?

Introduce public interface to enable the Logs Feature.

### How?

**Default:**
```swift
import DatadogLogs

Logs.enable()
```

**Advanced:**
```swift

import DatadogLogs

Logs.enable(
  with: Logs.Configuration(
    samplingRate: 20,
    eventMapper: { $0 },
    customIntakeURL: customIntakeURL,
  ),
  in: core
)
```

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [x] Run smoke tests


DatadogSDKObjc.podspec
```diff
@@ -22,4 +22,5 @@ Pod::Spec.new do |s|
 
   s.source_files = "Sources/DatadogObjc/**/*.swift"
   s.dependency 'DatadogSDK', s.version.to_s
+  s.dependency 'DatadogLogs', s.version.to_s
```
maciejburda, line 25: ObjC target won't be modularised or it's temporary?

maciejburda (approved):
LGTM

DatadogSDKObjc.podspec
maxep: Indeed, the objc interface will be exposed in a single module for now. Ultimately, we should find a way to include the objc interface in each module, without polluting swift completion.

DatadogLogs/Sources/Logs.swift
```diff
@@ -0,0 +1,113 @@
...
+        /// Sets the sampling rate for logging.
+        ///
+        /// The sampling rate must be a value between `0` and `100`. A value of `0` means no logs will be processed, `100`
+        /// means all logs will be processed.
+        ///
+        /// By default sampling is disabled, meaning that all logs are being processed).
+        public var samplingRate: Float
+
+        /// Overrides the custom server endpoint where Logs are sent.
+        public var customIntakeURL: URL?
```
ncreated, line 32: blocker/ In Android SDK [it is named `customEndpoint`](https://github.com/DataDog/dd-sdk-android/blob/129a616b921535f5308172651273676e8bcf46df/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/RumConfiguration.kt#L239) and this is also the name I'm adopting in RUM.
* it must be the same in both SDKs and all products (I will update RUM if necessary),
* I think we should avoid using the "intake" word as it can be misleading to most people (there are no _intake_ requirements behind this URL, it can be any API / server).

DatadogLogs/Sources/Logs.swift
```diff
@@ -0,0 +1,113 @@
...
+        public var processInfo: ProcessInfo = .processInfo
+
+        /// Sets the custom mapper for `LogEvent`. This can be used to modify logs before they are send to Datadog.
+        ///
+        /// - Parameter mapper: the closure taking `LogEvent` as input and expecting `LogEvent` as output.
+        /// The implementation should obtain a mutable version of the `LogEvent`, modify it and return it. Returning `nil` will result
+        /// with dropping the Log event entirely, so it won't be send to Datadog.
+        public mutating func setEventMapper(_ mapper: @escaping EventMapperClosure) {
+            eventMapper = SyncLogEventMapper(mapper)
+        }
```
ncreated, line 47: convention/ Why not `public var eventMapper: EventMapperClosure? = nil`? If we need to wrap it into `SyncLogEventMapper`, this can happen when resolving the config in `Logs.enable(with:in:)` or even later. Exposing a setter looks very inconsistent and doesn't promote simplicity.

DatadogLogs/Sources/Logs.swift
```diff
@@ -0,0 +1,113 @@
...
+/// - Log to Datadog in JSON format natively.
+/// - Use default and add custom attributes to each log sent.
+/// - Record real client IP addresses and User-Agents.
+/// - Leverage optimized network usage with automatic bulk posts.
+public struct Logs {
+    /// The Logs general configuration.
+    ///
+    /// This configuration will be applied to all Logger instances.
+    public struct Configuration {
+        public typealias EventMapperClosure = (LogEvent) -> LogEvent?
```
ncreated, line 22: For consistency with `RUM*EventMapper`:
```suggestion
        public typealias LogEventMapper = (LogEvent) -> LogEvent?
```

DatadogLogs/Sources/Logs.swift
```diff
@@ -0,0 +1,113 @@
...
+        public var samplingRate: Float
+
+        /// Overrides the custom server endpoint where Logs are sent.
+        public var customIntakeURL: URL?
+
+        /// Overrides the main bundle instance.
+        public var bundle: Bundle = .main
+
+        /// Overrides the current process info.
+        public var processInfo: ProcessInfo = .processInfo
```
ncreated, line 38: Do we need this in `public` interface 🤔💭? What would be the reason for configuring custom `ProcessInfo` and how the user would know which ENV / arguments are supported (and how) by `DatadogLogs`?
```suggestion
        /// Overrides the current process info.
        internal var processInfo: ProcessInfo = .processInfo
```

DatadogLogs/Sources/Logs.swift
```diff
@@ -0,0 +1,113 @@
...
+        /// with dropping the Log event entirely, so it won't be send to Datadog.
+        public mutating func setEventMapper(_ mapper: @escaping EventMapperClosure) {
+            eventMapper = SyncLogEventMapper(mapper)
+        }
+
+        /// Sets the custom mapper for `LogEvent`. This can be used to modify logs before they are send to Datadog.
+        ///
+        /// The implementation should obtain a mutable version of the `LogEvent`, modify it and return it. Returning `nil` will result
+        /// with dropping the Log event entirely, so it won't be send to Datadog.
+        internal var eventMapper: LogEventMapper?
```
ncreated, line 53: One public eventMapper can be enough 🙂 (see previous comment).

ncreated (changes requested):
Great to see it 💪🏋️‍♂️!

I only looked at the public API and found few convention problems. Making change request. I'll review the rest once we discuss / change my suggestions.

DatadogLogs/Sources/Logs.swift
maxep: Agreed 👍

DatadogLogs/Sources/Logs.swift
maxep: I think it conflicts with a public protocol available for cross platform, let me double check.

DatadogLogs/Sources/Logs.swift
maxep: Indeed, only used for test so it's better to not expose it 👍

DatadogLogs/Sources/Logs.swift
maxep: Currently, we only allow mapper in sync so we expose the closure. The async mapper is used by Flutter, therefor the 'internal mutation' which need to be set on the config object.

But I agree that a property would look better. We can do:

```swift

public var eventMapper: EventMapperClosure? {
  get { (_eventMapper as? SyncLogEventMapper)?.mapper }
  set { _eventMapper = SyncLogEventMapper(newValue) }
}

internal var _eventMapper: LogEventMapper? = nil
```

Not sure which one I prefer 🤔

DatadogLogs/Sources/Logs.swift
```diff
@@ -0,0 +1,110 @@
...
+        /// means all logs will be processed.
+        ///
+        /// By default sampling is disabled, meaning that all logs are being processed).
+        public var sampleRate: Float
+
+        /// Overrides the custom server endpoint where Logs are sent.
+        public var customEndpoint: URL?
+
+        /// Overrides the main bundle instance.
+        public var bundle: Bundle = .main
```
ncreated, line 35: suggestion/ Seeing this in Logs public API seems odd, so I checked if we can remove it. Looks we can:
```diff
// LogEventBuilder.swift

- loggerName: loggerName,
+ loggerName: loggerName ?? context.applicationBundleIdentifier,
```
Otherwise, the user will have to configure `bundle: Bundle` both in Core and Logs, which can be quite error-prone.

WDYT @maxep ? If you agree, we can move this update to separate JIRA for later (before GA).

DatadogLogs/Sources/Logs.swift
```diff
@@ -0,0 +1,110 @@
...
+        internal var processInfo: ProcessInfo = .processInfo
+
+        /// Sets the custom mapper for `LogEvent`. This can be used to modify logs before they are send to Datadog.
+        ///
+        /// - Parameter mapper: the closure taking `LogEvent` as input and expecting `LogEvent` as output.
+        /// The implementation should obtain a mutable version of the `LogEvent`, modify it and return it. Returning `nil` will result
+        /// with dropping the Log event entirely, so it won't be send to Datadog.
+        public mutating func eventMapper(_ mapper: @escaping EventMapperClosure) {
+            self.mapper = SyncLogEventMapper(mapper)
+        }
```
ncreated, line 47: blocker (simplicity)/ Do we need to make it that complex 🙂? Simply:
```swift
public var eventMapper: LogEventMapper?
```
can be obvious enough. It will offer both **setter** and **getter** and will be consistent with closure-based APIs in other products (_keep it simple_ is a good baseline here).

If Flutter SDK needs different mapper, we can resolve it when used:
```swift
let feature = LogsFeature(
   logEventMapper: config._internalEventMapper ?? config.eventMapper
}
```

DatadogLogs/Sources/Logs.swift
```diff
@@ -0,0 +1,110 @@
...
+    ///   - core: The instance of Datadog SDK to enable Logs in (global instance by default).
+    public static func enable(
+        with configuration: Configuration = .init(),
+        in core: DatadogCoreProtocol = CoreRegistry.default
+    ) {
+        let applicationBundleIdentifier = configuration.bundle.bundleIdentifier ?? "unknown"
+        let debug = configuration.processInfo.arguments.contains(LaunchArguments.Debug)
+        let feature = LogsFeature(
+            logEventMapper: configuration.mapper,
+            dateProvider: SystemDateProvider(),
```
ncreated, line 88: suggestion/ If we move `dateProvider` to `Logs.Configuration`:
```swift
struct Logs.Configuration {
   internal var dateProvider: DateProvider = SystemDateProvider()
}
```
then we will no longer need to mock `LogsFeature` in tests and register it to `core` every time. Both will be equal:
```swift
Logs.enable(with: .init(dateProvider: DateProviderMock()), in: core)
```
```swift
let logs: LogsFeature = .mockWith(dateProvider: DateProviderMock())
core.register(logs)
```
This way we can reduce the number of feature mocks.

DatadogLogs/Sources/Logs.swift
```diff
@@ -0,0 +1,110 @@
...
+        internal var mapper: LogEventMapper?
+
+        /// Creates a Logs configuration object.
+        ///
+        /// - Parameters:
+        ///   - eventMapper: The custom mapper for `LogEvent`. This can be used to modify logs before they are send to Datadog.
+        ///   - sampleRate: The sample rate for logging.
+        ///   - sampleRate: Overrides the custom server endpoint where Logs are sent.
+        ///   - bundle: Overrides the main bundle instance.
+        public init(
```
ncreated, line 62: suggestion/ Moving the `init` into separate `extension` will enable full initializer in `internal` scope. It will be possible to set each value in tests.

ncreated (changes requested):
Excellent work 💪. I'm resulting with very similar changes in RUM 👌.

Everything looks good, except the `eventMapper` thing. I'm up for a quick pair programming to make it simpler. Other than this, I left few minor suggestions.

DatadogLogs/Sources/Logs.swift
```diff
@@ -0,0 +1,110 @@
...
+        /// The implementation should obtain a mutable version of the `LogEvent`, modify it and return it. Returning `nil` will result
+        /// with dropping the Log event entirely, so it won't be send to Datadog.
+        internal var mapper: LogEventMapper?
+
+        /// Creates a Logs configuration object.
+        ///
+        /// - Parameters:
+        ///   - eventMapper: The custom mapper for `LogEvent`. This can be used to modify logs before they are send to Datadog.
+        ///   - sampleRate: The sample rate for logging.
+        ///   - sampleRate: Overrides the custom server endpoint where Logs are sent.
```
ncreated, line 60: blocker/ Duplicated parameter comment

DatadogLogs/Sources/Logs.swift
maxep: Indeed, that's a good point 👍

DatadogLogs/Sources/Logs.swift
maxep: Oh nice, lets do it here, it's a small impact 👍

DatadogLogs/Sources/Logs.swift
maxep: I would prefer to keep a single `init` and rely on internal property setters in tests. e.g:
```swift
func testSomething() {
  var config = Feature.Configuration(...)
  config._internal_property = ...
  Feature.enable(with: config)
}
```
Or
```swift
extension Feature.Configuration {
  func mockWith(
    ...,
    _internal_property: Type
  ) -> Self {
    ...
  }
}
```

I think it would be easier for users to have the init as part of the struct, WDYT?

DatadogLogs/Sources/Logs.swift
maxep: Resolved by your suggestion above 👍

DatadogLogs/Sources/Logs.swift
ncreated: 👌 Indeed, it's still possible. Not a big difference after all - sounds good 👍 🙂.

ncreated (approved):
👏 🤜 🤛 Awesome 🚀

maxep:
image

PR #1317, opened 2023-06-20
RUMM-2923 Logger Creation

maxep:
### What and why?

Introduce public interface to create a `LoggerProtocol` instance.

### How?

**Default:**
```swift
import DatadogLogs

let logger = Logger.create()
```

**Advanced:**
```swift

import DatadogLogs

let logger = Logger.create(
  with: Logger.Configuration(
    service = nil,
    logger = nil,
    sendNetworkInfo = false,
    bundleWithRUM = true,
    bundleWithTrace = true,
    remoteSampleRate = 100,
    remoteLogThreshold = .debug,
    consoleLogFormat = nil
  ),
  in: core
)
```

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [x] Run smoke tests


DatadogLogs/Sources/LoggerProtocol.swift
```diff
@@ -0,0 +1,253 @@
...
+    ///   - attributes: a dictionary of attributes (optional) to add for this message. If an attribute with
+    /// the same key already exist in this logger, it will be overridden (only for this message).
+    func log(
+        level: LogLevel,
+        message: String,
+        errorKind: String?,
+        errorMessage: String?,
+        stackTrace: String?,
+        attributes: [String: Encodable]?
+    )
```
ncreated, line 69: blocker/ Like the comment says: it is **only for cross-platform SDKs** in which case we should provide this method through separate `_internal` interface:
```swift
let logger: LoggerProtocol
logger._internal.log(...)
```
Part of V2, we want to strictly segregate _internal-public_ interface from _user-facing_ interface to not pollute and having to maintain backward-compatibility in the latter, endlessly.

For consistency, let's have `LoggerProtocol+Internal` like we did with [`RUMProtocol+Internal.swift`](https://github.com/DataDog/dd-sdk-ios/blob/feature/v2/DatadogRUM/Sources/RUMMonitorProtocol%2BInternal.swift).

DatadogLogs/Sources/Logger.swift
```diff
@@ -7,247 +7,183 @@
...
-    /// Sends a WARN log message.
-    /// - Parameters:
-    ///   - message: the message to be logged
-    ///   - error: the error information (optional)
-    ///   - attributes: a dictionary of attributes (optional) to add for this message. If an attribute with
-    /// the same key already exist in this logger, it will be overridden (only for this message).
-    func warn(_ message: String, error: Error? = nil, attributes: [AttributeKey: AttributeValue]? = nil) {
-        log(level: .warn, message: message, error: error, attributes: attributes)
+        /// The service name  (default value is set to application bundle identifier)
+        public var serviceName: String?
```
ncreated, line 22: blocker/ _"Rule 7: Use of standard Datadog and RUM terminology"_
```suggestion
        /// The service name  (default value is set to application bundle identifier)
        public var service: String?
```
Also in `@objc` interface ☝️

DatadogLogs/Sources/Logger.swift
```diff
@@ -7,247 +7,183 @@
...
+
+        /// The logger custom name (default value is set to main bundle identifier)
+        public var loggerName: String?
```
ncreated, line 25: suggestion/ Following [rule 7](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3027638506/Public+API+Convention+in+iOS+SDK+2.x#Rule-7%3A-Use-of-standard-Datadog-and-RUM-terminology) and _“avoid platform self referencing”_ I rather think it should be `name`. It is used as part of `Logger.create()` so it's rather obvious what the `name` would be. WDYT?
```suggestion
        /// The logger custom name (default value is set to main bundle identifier)
        public var name: String?
```

DatadogLogs/Sources/Logger.swift
```diff
@@ -7,247 +7,183 @@
...
+        ///
+        /// `true` by default.
+        public var bundleWithTrace: Bool
+
+        /// Enables logs to be sent to Datadog servers.
+        /// Can be used to disable sending logs in development.
+        /// See also: `printLogsToConsole(_:)`.
+        ///
+        /// `true` by default.
+        public var sendLogsToDatadog: Bool
```
ncreated, line 53: blocker/ [Rule 7](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3027638506/Public+API+Convention+in+iOS+SDK+2.x#Rule-7%3A-Use-of-standard-Datadog-and-RUM-terminology): _“avoid platform self referencing”_ "Datadog" is obvious implication in this context.

It makes me think if we need this option at all. Android [`Logger` defines `sampleRate`](https://github.com/DataDog/dd-sdk-android/blob/a982b35ceaec738f19f137b6f1333af325b285a0/dd-sdk-android/src/main/kotlin/com/datadog/android/log/Logger.kt#L320-L326) and this is something we should add now or later for the sake of consistency. With `sampleRate`, distinction between "remote", "console" and "remote & console" will be clear (IMO sampling should only apply to remote logs, not console logs):
```swift
// remote:
loggerConfig.sampleRate = 100
loggerConfig.consoleLogFormat = nil

// console:
loggerConfig.sampleRate = 0
loggerConfig.consoleLogFormat = .short

// remote & console:
loggerConfig.sampleRate = 100
loggerConfig.consoleLogFormat = .short
```

If we keep `sendLogsToDatadog` option, we're allowing inconsistency:
```swift
loggerConfig.sendLogsToDatadog = false
loggerConfig.sampleRate = 100
```

If we decide to keep this option anyway, I'd suggest renaming it to:
* `remoteLogging: Bool`

WDYT @maxep @0xnm? Thoughts @xgouchet ?

DatadogLogs/Sources/Logger.swift
```diff
@@ -7,247 +7,183 @@
...
+        public var consoleLogFormat: ConsoleLogFormat?
+
+        /// Set the minimum log level reported to Datadog servers.
+        /// Any log with a level equal or above the threshold will be sent.
+        ///
+        /// Note: this setting doesn't impact logs printed to the console if `printLogsToConsole(_:)`
+        /// is used - all logs will be printed, no matter of their level.
+        ///
+        /// `LogLevel.debug` by default
+        public var datadogReportingThreshold: LogLevel
```
ncreated, line 67: blocker/ [Rule 7](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3027638506/Public+API+Convention+in+iOS+SDK+2.x#Rule-7%3A-Use-of-standard-Datadog-and-RUM-terminology): _“avoid platform self referencing”_ - "Datadog" is obvious in this context.

I'd say better candidates are:
* `threshold: LogLevel` - but it lacks easy implication that this option only applies to remote not console logs
* `remoteLoggingThreshold: LogLevel`

WDYT @maxep @ganeshnj @maciejburda ?

ncreated (changes requested):
💪 Good for all changes, but I marked few important points on the public interface. Considering them blockers.

DatadogLogs/Sources/Logger.swift
maxep: Agreed!

DatadogLogs/Sources/Logger.swift
maxep: That's a good point 👍
On iOS, [sampling rate is set by the feature config](https://github.com/DataDog/dd-sdk-ios/blob/feature/v2/DatadogLogs/Sources/Logs.swift#L29) but it would be trivial to make it per logger instance.

DatadogLogs/Sources/Logger.swift
maxep: What about `reportingThreshold` or `reportingLevelThreshold`? 
Android uses `setDatadogLogsMinPriority` which doesn't work well on iOS either. 
So I guess we should sync with @0xnm on that one!

DatadogLogs/Sources/LoggerProtocol.swift
maxep: I'm adopting a slightly different approach because I have multiple instances that need to comply, `ConsoleLogger`, `RemoteLogger`, and `CombinedLogger`:
```swift
/// Internal Logger for Cross-Platform access.
public protocol InternalLoggerProtocol {
    /// General purpose logging method.
    /// Sends a log with certain `level`, `message`, `errorKind`,  `errorMessage`,  `stackTrace` and `attributes`.
    ///
    /// This method is meant for non-native or cross platform frameworks (such as React Native or Flutter) to send error information
    /// to Datadog. Although it can be used directly, it is recommended to use other methods declared on `Logger`.
    ///
    /// - Parameters:
    ///   - level: the log level
    ///   - errorKind: the kind of error reported
    ///   - errorMessage: the message attached to the error
    ///   - stackTrace: a string representation of the error's stack trace
    ///   - attributes: a dictionary of attributes (optional) to add for this message. If an attribute with
    /// the same key already exist in this logger, it will be overridden (only for this message).
    func log(
        level: LogLevel,
        message: String,
        errorKind: String?,
        errorMessage: String?,
        stackTrace: String?,
        attributes: [String: Encodable]?
    )
}

private struct NOPInternalLogger: InternalLoggerProtocol {
    func log(
        level: LogLevel,
        message: String,
        errorKind: String?,
        errorMessage: String?,
        stackTrace: String?,
        attributes: [String : Encodable]?
    ) { }
}

/// Extends `LoggerProtocol` with additional methods designed for Datadog cross-platform SDKs.
extension LoggerProtocol {
    /// Grants access to an internal interface utilized only by Datadog cross-platform SDKs.
    /// **It is not meant for public use** and it might change without prior notice.
    public var _internal: InternalLoggerProtocol {
        self as? InternalLoggerProtocol ?? NOPInternalLogger()
    }
}
```

The idea is the same but more generic 👍

DatadogLogs/Sources/Logger.swift
0xnm: Indeed `sampleRate` and `sendLogsToDatadog` are linked together, but having only `sampleRate` may be not that obvious to customers (in the end, while it may sound strange, sample rate can be applied to console logs as well). Also in Android we have a `setLogcatLogsEnabled` for console logs, and `setDatadogLogsMinPriority` to control min priority of logs sent to Datadog.

So to have a symmetry with `setLogcatLogsEnabled` on Android I would keep `sendLogsToDatadog` option, but not against maybe to rename it to `remoteLogging` (then also similar term should be used for `setDatadogLogsMinPriority` method).

DatadogLogs/Sources/LoggerProtocol.swift
ncreated: 👌 Works for me as long as it's separate and non user-facing interface 💪

maciejburda (approved):
LGTM

DatadogLogs/Sources/Logger.swift
maciejburda: I like `remoteLoggingThreshold`. To me intention is more obvious. When dealing with logging, I would need to think twice to understand that `reporting` means logging to remote (aka. sending) in that case.

DatadogLogs/Sources/Logger.swift
maxep: Following our discussion,  I'm going for:
```swift
/// Sets the sample rate for remote logging.
///
/// **When set to `0`, no log entries will be sent to Datadog servers.**
/// A value of`100` means all logs will be processed.
///
/// When setting the `remoteSampleRate` to `0`
///
/// Default is `100`, meaning that all logs will be sent.
public var remoteSampleRate: Float

/// Set the minimum log level reported to Datadog servers.
/// Any log with a level equal or above the threshold will be sent.
///
/// Note: this setting doesn't impact logs printed to the console if `printLogsToConsole(_:)`
/// is used - all logs will be printed, no matter of their level.
///
/// `LogLevel.debug` by default
public var remoteLogThreshold: LogLevel
```

ncreated (approved):
👌 Awesome!

PR #1319, opened 2023-06-21
RUMM-3353 Crash Reporter Configuration

maxep:
### What and why?

Make Crash Reporting compliant with new configuration convention.
### How?

```swift
import DatadogCrashReporting

CrashReporting.enable()
```

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [x] Run smoke tests


DatadogCrashReporting/Sources/CrashReporter.swift
```diff
@@ -0,0 +1,65 @@
...
+///
+/// - Aggregated iOS crash dashboards and attributes
+/// - Symbolicated iOS crash reports
+/// - Trend analysis with iOS error tracking
+///
+/// In order to symbolicate your stack traces, find and upload your .dSYM files to Datadog.
+/// Then, verify your configuration by running a test crash and restarting your application.
+///
+/// Your crash reports appear in [Error Tracking](https://app.datadoghq.com/rum/error-tracking).
+public final class CrashReporter {
```
ncreated, line 21: blocking until discussed/ Why `CrashReporter` not `CrashReporting`? I believe the latter will be more unified with other products:
* `Logs.enable()`
* `Trace.enable()`
* `SessionReplay.enable()`
* `RUM.enable()`

The name of `CrashReporter` is closer to how we call data collectors in each product:
* `Logger`
* `Tracer`
* `RUMMonitor`

I think we should rather reserve it for the future if (for any reason) we need to make it a shared component within `DatadogCrashReporting` product.

WDYT?

ncreated (changes requested):

DatadogCrashReporting/Sources/CrashReporter.swift
maxep: Good point, let me change that!

ncreated (approved):
LG 🚀

PR #1320, opened 2023-06-21
Stop pinning to .git URL in Package.swift.

ZevEisenberg:
### What and why?

When you pin the .git version of a URL in Package.swift, it seems to cause perturbations in the Xcode project of any client. You'll occasionally get diffs where Xcode adds or removes the .git suffix, seemingly at random. I don't know for sure if this fixes it, but it seems to maybe fix it.

### How?

Delete .git pinning from Package.swift. I didn't remove it from a couple of other Package.swift files I found in the project because they seemed to be internal, but I can update them too if it seems important.

I haven't updated the CHANGELOG or filed a ticket. Should I?

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -43,7 +43,7 @@ let package = Package(
         ),
     ],
     dependencies: [
-        .package(name: "PLCrashReporter", url: "https://github.com/microsoft/plcrashreporter.git", from: "1.11.0"),
+        .package(name: "PLCrashReporter", url: "https://github.com/microsoft/plcrashreporter", from: "1.11.0"),
     ],
     targets: [
         .target(


noremac (approved):

maxep:
Hi @ZevEisenberg ! Thanks for PR, but I'm not sure to understand the issue here.
> When you pin the .git version of a URL in Package.swift, it seems to cause perturbations in the Xcode project of any client. You'll occasionally get diffs where Xcode adds or removes the .git suffix, seemingly at random.

I'm not sure to follow, do you mean diffs in the `.pbxproj` when resolving the package?

ZevEisenberg:
Yep. But after further experimentation, I'm not sure if this is required. It seems that the issue is not the presence or absence of .git. Rather, it is consistency between packages.

That is, it's fine if you import PLCrashReporter with .git. But if another package declares it as a dependency without .git, and I import both packages, Xcode will randomly choose to use or omit the .git extension in its pinning in Package.resolved.

That's my understanding of our latest understanding, anyway. @noremac did I miss anything? I guess we don't need this merged after all, right?

maxep:
Thanks for the update @ZevEisenberg, I will close this PR then 👍

PR #1323, opened 2023-06-22
Fix data race condition in test

maxep:
### What and why?

Fix data race condition issue in `CrashReportingPluginMock`.

### How?

CI jobs seems to hang when running `testWhenPendingCrashReportIsFound_itIsSentAndPurged`, and by running the test repeatedly we can run into a data-race condition.

The `CrashReportingPluginMock` was missing a lock mechanism.

### Review checklist

- [x] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


ncreated (approved):
⚾ good spot 👌

ganeshnj (approved):

PR #1325, opened 2023-06-23
RUMM-2924 Create `DatadogTrace` API for 2.x 

ncreated:
### What and why?

📦 This PR extracts `DatadogTrace` public API for `2.0.0`.

The new API for tracing will consist of 3 main interfaces:
```swift
// Configuration:
let config: Trace.Configuration

// Feature:
Trace.enable(with: config)

// Tracer API:
Tracer.shared() -> OTTracer
```
@maxep please note that I'm not introducing `TracerProtocol` like we do for RUM monitor and Logger. Unlike these two, our tracer conforms to external Open Tracing standard and hence IMO we should using existing `OTTracer` interface. Even though OT is deprecated and it is just a conventional conformance, long term we will rather switch to OTel and not our custom interface.

### How?

Only mandatory updates and following convention:
- removal of V1 Trace configuration from `Datadog`;
- addition of V2 `Trace.Configuration` in `DatadogTrace`;
- re-creation of V2 `DDTraceConfiguration` in `DatadogObjc`;
- updates in `DatadogTests` where we still carry on `Tracer` tests;
- updates to Example app to use V2 Trace API;
- updates to IntegrationTests project to use V2 Trace API
- updates to smoke tests
- updates to Carthage setup in `pbxproj`

#### Distributed Tracing API
Because `URLSession` instrumentation can be enabled in both RUM and Trace configuration, I kept the same convention in its setup (aligned with https://github.com/DataDog/dd-sdk-ios/pull/1321).  

By default, no `URLSession` instrumentation is installed (distributed tracing is disabled):
```swift
let config = Trace.Configuration()
XCTAssertNil(config.urlSessionTracking)
```
To enable distributed tracing with `x-datadog-*` headers:
```swift
config.urlSessionTracking = .init(
   firstPartyHostsTracing: .trace(hosts: ["example.com"]) // 20% sample rate by default
)
config.urlSessionTracking = .init(
   firstPartyHostsTracing: .trace(hosts: ["example.com"], sampleRate: 100) // custom sample rate
)
```
To enable distributed tracing with custom headers:
```swift
config.urlSessionTracking = .init(
   firstPartyHostsTracing: .traceWithHeaders(
      hostsWithHeaders: ["example.com": [.b3, .datadog]] // 20% sample rate by default
   )
)
config.urlSessionTracking = .init(
   firstPartyHostsTracing: .traceWithHeaders(
      hostsWithHeaders: ["example.com": [.b3, .datadog]],
      sampleRate: 100 // custom sample rate
   )
)
```

❗ **Note**, that network instrumentation in tracing only works for first party requests. Requests sent to 3rd party URLs won't be traced: no span and no headers injection. Nothing new, just leaving it for clarity.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


DatadogTrace/Sources/DatadogTracer.swift
```diff
@@ -7,62 +7,29 @@
...
-///     import Datadog
-///
-///     // tracer reference
-///     var tracer: DDTracer!
-///
-///     // instantiate Datadog tracer
-///     tracer = DDTracer.initialize(...)
-///
-public class DatadogTracer: OTTracer {
+internal class DatadogTracer: OTTracer {
```
ncreated, line 10: Leaving it not renamed, because I feel we lack solid convention for such things. I can rename it to anything (`<whatever>Tracer`) but I wish we can make up some no-brainer pattern to follow - likely on next iOS Dev Chat.

DatadogTrace/Sources/TraceConfiguration.swift
```diff
@@ -0,0 +1,151 @@
...
+
+        /// Custom server url for sending traces.
+        ///
+        /// Default: `nil`.
+        public var customEndpoint: URL?
+
+        // MARK: - Nested Types
+
+        /// Configuration of automatic network requests tracing.
+        public struct URLSessionTracking {
```
maxep, line 76: I wonder if shouldn't combine this one and `RUM.Configuration.URLSessionTracking` into a single config struct defined in `DatadogInternal` 🤔 

Not a change request but I would like your opinion on this?

maxep (approved):
Perfect 👌 

I've left one non-blocking question but it's good to go IMO!

DatadogTrace/Sources/TraceConfiguration.swift
ncreated: I was thinking about this 👌, but IMO this wouldn't be correct nor user-friendly.

First, there is a slight but key difference in both:
```swift
// DatadogRUM:
public struct URLSessionTracking {
    public var firstPartyHostsTracing: FirstPartyHostsTracing? // <-- ❗️ optional
    public var resourceAttributesProvider: RUM.ResourceAttributesProvider?
}

// DatadogTrace:
public struct URLSessionTracking {
    public var firstPartyHostsTracing: FirstPartyHostsTracing // <-- ❗️ non-optional
}
```
Not only RUM takes `resourceAttributesProvider`, but also optionality of `firstPartyHostsTracing` is different. This comes from the fact that:
* If `URLSession` is instrumented by RUM, **it tracks all requests** as RUM resource. Additionally it will inject tracing headers to first-party requests, hence we ask for their definition optionaly. 
* If `URLSession` is instrumented by Trace, **it doesn't track all requests** (ref.: #314). It only intercepts first-party requests and injects tracing headers. Hence, it doesn't make sense to have `URLSessionTracking` without definition of hosts.

Next, because of that difference ☝️ , it could be hard to come up with a single definition of `URLSessionTracking` that works for both products and includes product-agnostic explanation in API comment.

Last, this ☝️ is the difference we have today. It is easy to imagine this drifting even more away (e.g. distinguishing the `sampleRate` value in `FirstPartyHostsTracing` enum of both products). Consider the fact that RUM and APM are separate DD products with distinct infrastructure which might translate into separate constraints.

For all that reasons, I think it is safer duplicate this definition than assume wrong abstraction. To help users with adopting this, I made sure that both `URLSessionTracking` and `FirstPartyHostsTracing` APIs stay close to each other in both products.

How does it all sound @maxep ?

ncreated: Also, keep in mind that there is no practical use case of "sharing config definition" by the user. RUM offers a superset of distributed tracing - if you enable automatic RUM resources with `urlSessionTracking`, no real point of enabling APM network tracing as RUM will manage do distributed tracing for you.

maxep: It all makes sense, thanks for the clarification 🙏

PR #1328, opened 2023-06-23
RUMM-3151 feat: reduce number of view updates by looking ahead duplicate events…

ganeshnj:
### What and why?

There are several view updates happens which are not needed. These view updates are reduced by the backend anyway, but we could reduce the number of view updates by filtering them out from the payload.

### How?

While saving RUMViewEvent to disk, we save the metadata along with it. This allows us to access key information without decoding the whole event. While preparing the request body, we filter out the events which are duplicate and will be sent in the same batch.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [x] Run smoke tests


Sources/Datadog/DatadogCore/Storage/Writing/FileWriter.swift
```diff
@@ -21,12 +21,25 @@ internal struct FileWriter: Writer {
     // MARK: - Writing data
 
     /// Encodes given value to JSON data and writes it to the file.
```
maciejburda, line 23: We could update the comment header to reflect the change

Sources/Datadog/RUM/RUMV2Configuration.swift
```diff
@@ -30,6 +33,45 @@ internal func createRUMConfiguration(configuration: FeaturesConfiguration.RUM) -
...
+    init(decoder: JSONDecoder = JSONDecoder()) {
+        self.decoder = decoder
+    }
+
+    func filter(events: [Event]) -> [Event] {
+        var seen = Set<String>()
+        var filtered = [Event]()
+
+        // reversed is O(1) and no copy because it is view on the original array
+        for event in events.reversed() {
```
maciejburda, line 48: I would slightly favour functional implementation for the whole thing (slightly more difficult to break in the future)

maciejburda (commented):
Did a first pass. Looking good. Some minor suggestions 🙌

Sources/Datadog/DatadogCore/Storage/Writing/FileWriter.swift
ganeshnj: yes, fixed it.

Sources/Datadog/RUM/RUMV2Configuration.swift
ncreated: Also, we found the imperative `for ... in` being much faster for long sequences in [SR's text obfuscation](https://github.com/DataDog/dd-sdk-ios/pull/1024#discussion_r999054231). It doesn't apply here, as the number of events is rather dozens, not thousands - I'm just sharing findings :).

Sources/Datadog/DatadogCore/Storage/Writing/FileWriter.swift
```diff
@@ -20,13 +20,30 @@ internal struct FileWriter: Writer {
...
+            // Make sure both event and event metadata are written to the same file.
+            // This is to avoid a situation where event is written to one file and event metadata to another.
+            // If this happens, the reader will not be able to match event with its metadata.
+            let writeSize = UInt64(encodedValue.count + (encodedMetadata?.count ?? 0))
             let file = try forceNewFile ? orchestrator.getNewWritableFile(writeSize: writeSize) : orchestrator.getWritableFile(writeSize: writeSize)
-            try file.append(data: data)
+            try file.append(data: encodedValue)
+            if let encodedMetadata = encodedMetadata {
+                try file.append(data: encodedMetadata)
+            }
```
ncreated, line 46: blocker/ Writes must be atomic (one transaction; both `event + meta` or nothing). Lack of atomicity was a severe issue in the past (incident-4797), so we'll run into regression with this code. 
* See: https://github.com/DataDog/dd-sdk-ios/pull/133
* Also see the number of I/O errors in iOS telemetry to get the idea on reliability of file writing 😉 

What we also need is making sure that `event + metadata` write is stress tested against flaky I/O. This test corresponds to incident-4797 and we can duplicate it / adapt to the new write variant:

https://github.com/DataDog/dd-sdk-ios/blob/73e99dd1d5813624bff9af5cb9196c3b7c6b9cf4/Tests/DatadogTests/Datadog/Core/Persistence/Writing/FileWriterTests.swift#L168-L169

🤔💭 Looking at this test, I think it became outdated after introduction of TLV during V2 migration (in V1 we were writing `data + separator byte`). One more argument to touch it. WDYT @ganeshnj ?

Sources/Datadog/DatadogCore/Storage/Reading/FileReader.swift
```diff
@@ -62,21 +62,25 @@ internal final class FileReader: Reader {
             // get event blocks only
             .compactMap {
                 switch $0.type {
-                case .event:
-                    return $0.data
+                case .event, .eventMetadata:
+                    return $0
                 }
             }
```
maxep, line 68: I guess we don't need this filter anymore, we can return all blocks 👍

Sources/Datadog/DatadogInternal/Upload/Event.swift
```diff
@@ -0,0 +1,96 @@
...
+        guard metadata.type == .eventMetadata else {
+            return Event(data: event.data, metadata: nil)
+        }
+        index += 1
+
+        return Event(data: event.data, metadata: metadata.data)
+    }
+}
+
+extension RUMViewEvent {
```
ncreated, line 78: suggestion/ This can't live in `Datadog/DatadogInternal/*` as it extends RUM type. To simplify `feature/v2` → `develop` merge, we can move it to `RUM/*`. Not sure tho if this will help much with conflicts.

Sources/Datadog/DatadogInternal/Upload/Event.swift
```diff
@@ -0,0 +1,96 @@
...
+    /// It may be used to filter out the `RUMViewEvent` from the batch.
+    struct Metadata: Codable {
+        let id: String
+        let documentVersion: Int64
+
+        private enum CodingKeys: String, CodingKey {
+            case id = "id"
+            case documentVersion = "document_version"
+        }
+    }
```
ncreated, line 89: question/ What we described in spec is using raw memory binding instead of JSON coding:
> The `view.document_version` is an unsigned integer and `view.id` is a fixed-length string, limited by the exact pattern and ASCII character-set:
>
> `^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$`
>
> It is reasonable to use direct memory binding and encode `view.document_version` as 4 bytes (way more than enough) followed by `view.id` as 36 bytes (UTF-8 string).

so we are not fulfilling the spec we agreed on.

Ultimately we want to store RUM view metadata as another TLV block:
```
+ (2 bytes) -+ (4 bytes) ----+ (4 bytes) -------+- (36 bytes) ---|
| BLOCK TYPE | DATA SIZE     |                DATA               |
+------------+---------------+-----------------------------------+
| VERSION (0)| LEGTH (40)    | DOCUMENT VERSION | VIEW ID        |
+------------+---------------+-----------------------------------+
```
To achieve this we need to expose some TLV helpers to feature modules through `DatadogInternal`.

maxep (changes requested):
Looks very good, nice work! 

I've left a blocking change about separating RUM specifics from `DatadogInternal` to facilitate merging with v2 👍

Tests/DatadogTests/Datadog/RUM/RUMFeatureTests.swift
```diff
@@ -210,22 +210,21 @@ class RUMFeatureTests: XCTestCase {
...
+        writer.write(value: RUMDataModelMock(attribute: "2nd event"), metadata: RUMViewEvent.Metadata(id: "2", documentVersion: 1))
+        writer.write(value: RUMDataModelMock(attribute: "3rd event"), metadata: RUMViewEvent.Metadata(id: "1", documentVersion: 2))
 
         let payload = try XCTUnwrap(server.waitAndReturnRequests(count: 1)[0].httpBody)
 
         // Expected payload format:
+        // event1JSON is skipped in favor of event3JSON which is same event with higher document revision
         // ```
-        // event1JSON
         // event2JSON
```
ncreated, line 222: nice integration 👌

Tests/DatadogTests/Datadog/Core/Persistence/Writing/FileWriterTests.swift
```diff
@@ -19,6 +19,42 @@ class FileWriterTests: XCTestCase {
...
     }
 
+    func testItWritesDataWithMetadataToSingleFileInTLVFormat() throws {
+        let writer = FileWriter(
+            orchestrator: FilesOrchestrator(
+                directory: temporaryDirectory,
+                performance: PerformancePreset.mockAny(),
+                dateProvider: SystemDateProvider()
+            ),
+            encryption: nil,
```
ncreated, line 29: suggestion/ How about adding one more test that makes sure metadata blocks are encrypted as well?

ncreated (changes requested):
Looks definitely good 👌👍, but I left few points with one blocker.

Sources/Datadog/DatadogInternal/Upload/Event.swift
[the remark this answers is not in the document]
maxep: Lets add the protocol when we need to have this abstraction shared across Features. For now, `FeatureRequestBuilder` is sufficient to transform `[Event]` to `URLRequest`, the transformation including filtering can stay internal to the Feature implementation and doesn't need any constrain.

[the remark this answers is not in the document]
ganeshnj: I agree, we don't have a need and it is better not to expose any public (even used internal) interface.

Sources/Datadog/DatadogCore/Storage/Reading/FileReader.swift
ganeshnj: removed.

Sources/Datadog/DatadogInternal/Upload/Event.swift
ganeshnj: Yes, `RUM/` is the place it belongs to

Tests/DatadogTests/Datadog/Core/Persistence/Writing/FileWriterTests.swift
ganeshnj: Good call, added.

Sources/Datadog/DatadogCore/Storage/Writing/FileWriter.swift
ganeshnj: This was great point, I couldn't able to fail the test with randomization we have placed and partially I think because we interrupt the writing after in between write calls. 

But like we talked, this test still has value so I modified with taking metadata into consideration.

Sources/Datadog/RUM/RUMV2Configuration.swift
```diff
@@ -30,6 +33,42 @@ internal func createRUMConfiguration(configuration: FeaturesConfiguration.RUM) -
...
+            }
+
+            guard let viewMetadata = try? decoder.decode(RUMViewEvent.Metadata.self, from: metadata) else {
+                // If we can't decode the metadata, we can't filter it.
+                return event
+            }
+
+            guard seen.contains(viewMetadata.id) == false else {
+                // If we've already seen this view, we can skip this
+                DD.logger.debug("Skipping RUMViewEvent with id: \(viewMetadata.id)")
```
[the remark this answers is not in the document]
ganeshnj: Updated the code to be less frequent and more detailed also.

maxep (approved):
Looks great !!

Sources/Datadog/DatadogCore/Storage/Writing/FileWriter.swift
```diff
@@ -27,23 +27,18 @@ internal struct FileWriter: Writer {
...
-            let encodedValue = try encode(encodable: value, blockType: .event)
-            let encodedMetadata: Data?
+            var encoded = try encode(encodable: value, blockType: .event)
             if let metadata = metadata {
-                encodedMetadata = try encode(encodable: metadata, blockType: .eventMetadata)
-            } else {
-                encodedMetadata = nil
+                let encodedMetadata = try encode(encodable: metadata, blockType: .eventMetadata)
+                encoded.append(encodedMetadata)
             }
```
ncreated: nit/ According to [the spec](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/2982249019/RFC+-+Reducing+the+number+of+view+updates+in+V2#(1)-Add-%E2%80%9Cevent-metadata%E2%80%9D-block-type-in-Core), it should be `metadata` followed by `event` and here we do reverse. We need to either update the code or the spec to be consistent.
 
> It will enable each Feature to optionally store information about the "event" block that it precedes:
>
> `[EVENT 1 METADATA] [EVENT 1] [EVENT 2 METADATA] [EVENT 2] [EVENT 3]`

Tests/DatadogTests/Datadog/Core/Persistence/Writing/FileWriterTests.swift
```diff
@@ -247,11 +294,20 @@ class FileWriterTests: XCTestCase {
...
         XCTAssertGreaterThan(events.count, 0)
         XCTAssertLessThanOrEqual(events.count, 300)
+        for event in events {
+            let actualFoo = try jsonDecoder.decode(Foo.self, from: event.data)
+            XCTAssertEqual(actualFoo, foo)
+
+            XCTAssertNotNil(event.metadata)
+            let actualMetadata = try jsonDecoder.decode(Metadata.self, from: event.metadata!)
+            XCTAssertEqual(actualMetadata, metadata)
+        }
```
ncreated, line 310: Great! 💪

Sources/Datadog/DatadogInternal/Upload/Event.swift
```diff
@@ -0,0 +1,67 @@
...
+/// Struct representing a single event.
+public struct Event: Equatable {
+    /// Data representing the event.
+    public let data: Data
+
+    /// Metadata associated with the event.
+    /// Metadata is optional and may be `nil` but of very small size.
+    /// This allows us to skip resource intensive operations in case such
+    /// as filtering of the events.
+    public let metadata: Data?
```
ncreated, line 18: nit/ Because we pass metadata in `write<E: Encodable, M: Encodable>(event: E, metadata: M)` the feature doesn't know the actual detail of `metadata: Data` encoding. We should clearly state in this comment that `metadata: Data?` is JSON encoded. Actually, the same comment should be applied to `data: Data` field.

ncreated (changes requested):
Nice work @ganeshnj 👌. Before we can go with merge:
* Please update the PR and branch name to [our convention](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/2923071440/Team+Operations#Pull-Requests) (notably, we must reference to [RUMM-3151](https://datadoghq.atlassian.net/browse/RUMM-3151) for future visibility of "why?" we decided on that change)
* What about the second part from Definition Of Done in [RUMM-3151](https://datadoghq.atlassian.net/browse/RUMM-3151)? Without dropping the old ["view update throttling" logic](https://github.com/DataDog/dd-sdk-ios/blob/dd6a509c5cba99024b0bae8423e88d3eb9f3160c/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift#L511) we are not solving the "`1ns` views problem" stated in the [spec](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/2982249019/RFC+-+Reducing+the+number+of+view+updates+in+V2). With this PR we drop unnecessary updates, but we still don't let the valid ones in. It is still an increment what we have here, so we can tackle this part separately (it will imply quite impactful change in tests). Happy to chat about this to give you more context 🙌.

ganeshnj:
> Nice work @ganeshnj 👌. Before we can go with merge:
> 
> * Please update the PR and branch name to [our convention](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/2923071440/Team+Operations#Pull-Requests) (notably, we must reference to [RUMM-3151](https://datadoghq.atlassian.net/browse/RUMM-3151) for future visibility of "why?" we decided on that change)
> * What about the second part from Definition Of Done in [RUMM-3151](https://datadoghq.atlassian.net/browse/RUMM-3151)? Without dropping the old ["view update throttling" logic](https://github.com/DataDog/dd-sdk-ios/blob/dd6a509c5cba99024b0bae8423e88d3eb9f3160c/Sources/Datadog/RUM/RUMMonitor/Scopes/RUMViewScope.swift#L511) we are not solving the "`1ns` views problem" stated in the [spec](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/2982249019/RFC+-+Reducing+the+number+of+view+updates+in+V2). With this PR we drop unnecessary updates, but we still don't let the valid ones in. It is still an increment what we have here, so we can tackle this part separately (it will imply quite impactful change in tests). Happy to chat about this to give you more context 🙌.

I realized the branch name as soon as I put the PR for review, the problem with branch name, I have to re-do the PR. 

For the 1ns issue - I want to tackle in a separate PR and not convolute this PR (keep it filtering specific). Let's chat about 1ns thing, will ping you on side.

Sources/Datadog/DatadogCore/Storage/Writing/FileWriter.swift
ganeshnj: errr, good catch.

ganeshnj: fixed.

ncreated (approved):
🏆

Sources/Datadog/DatadogInternal/Upload/Event.swift
maxep: Currently, responsibilities are not properly assigned: Core owns the encoding while Features are responsible of decoding. JSON encoding/decoding is only a convention between the core and Features, so I think it should be mentioned in comments.

Ultimately, I think we should transfer responsibility of encoding to the Features, so they are free to use the serialisation of their choices (JSON, TLV, protobuf, ...). Not a change request of this PR but something to consider in the future.

Sources/Datadog/DatadogInternal/Upload/Event.swift
ganeshnj: Fixed that.

PR #1329, opened 2023-06-23
[WIP] RUMM-2925 Create `DatadogSessionReplay` API for 2.x

ncreated:
### What and why?

// TBD

### How?

// TBD

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


DatadogSessionReplay/Sources/SessionReplayConfiguration.swift
```diff
@@ -0,0 +1,64 @@
...
+
+        internal var debugSDK: Bool = ProcessInfo.processInfo.arguments.contains(LaunchArguments.Debug)
+
+        /// Creates Session Replay configuration
+        /// - Parameters:
+        ///   - replaySampleRate: The sampling rate for Session Replay. It is applied in addition to the RUM session sample rate.
+        ///   - defaultPrivacyLevel: The way sensitive content (e.g. text) should be masked. Default: `.mask`.
+        ///   - customEndpoint: Custom server url for sending replay data. Default: `nil`.
+        public init(
+            replaySampleRate: Float,
```
ncreated, line 55: Here I make `replaySampleRate` a mandatory field for enabling SR feature although product requirement is _"`0` sample rate by default"_. 

Motivation: required parameter in Swift will work better than assuming `0` default. Otherwise, `SessionReplay.enable()` will give false promise of enabling anything, while no replay will be recorded in practise. It is more user-friendly to require explicit `sampleRate`:
```swift
SessionReplay.enable(with: .init(sampleRate: ???))
```
cc @mariusc83 @maciejburda WDYT?

DatadogSessionReplay/Tests/Recorder/PrivacyLevelTests.swift
ncreated, line 1: No real changes here - just renaming `Privacy` → `PrivacyLevel` and moving the file around.

maxep:
Indeed, we don't have the toolchain to build the xcfw from SPM yet. So yea, in the meantime the only option is to add the SR target to pbxproj :/

maxep (approved):
LGTM 👍

DatadogSessionReplay/Sources/SessionReplayConfiguration.swift
maciejburda: I like that. I actually god caught off guard by the default `0` setting once.

mariusc83: I see but I would rather discuss this with Jamie also before doing that change. This means our Public APIs are not aligned anymore in this moment.

PR #1332, opened 2023-06-26
RUMM-3355 Core v2 Configuration

maxep:
### What and why?

Introduce public interface to create a `DatadogCore` instance.

> The `Datadog` interface will be exposed as `DatadogCore` in #1335

### How?

**Default:**
```swift
import Datadog

Datadog.initialize(
  with: Datadog.Configuration(
    clientToken: "<token>",
    env: "<environment>"
  ),
  trackingConsent: .pending
)
```

**Advanced:**
```swift
import Datadog

Datadog.initialize(
  with: Datadog.Configuration(
    clientToken: "<token>",
    env: "<environment>",
    site: .us1,
    service: nil,
    bundle: .main,
    batchSize: .medium,
    uploadFrequency: .average,
    additionalConfiguration: [:],
    proxyConfiguration: nil,
    encryption: nil,
    serverDateProvider: nil
  ),
  trackingConsent: .pending
)
```

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [x] Run smoke tests


Sources/Datadog/Datadog+Internal.swift
```diff
@@ -8,7 +8,7 @@ import Foundation
 import DatadogInternal
 
 extension Datadog: InternalExtended {}
-extension Datadog.Configuration.Builder: InternalExtended {}
+//extension Datadog.Configuration.Builder: InternalExtended {}
```
ncreated, line 11: cleanup
```suggestion
```

Sources/Datadog/Datadog.swift
```diff
@@ -7,94 +7,152 @@
...
+        ///
+        /// By default, the Datadog SDK synchronizes with dedicated NTP pools provided by the
+        /// https://www.ntppool.org/ . Using different pools or setting a no-op `ServerDateProvider`
+        /// implementation will result in desynchronization of the SDK instance and the Datadog servers.
+        /// This can lead to significant time shift in RUM sessions or distributed traces.
+        public var serverDateProvider: ServerDateProvider
+
+        /// Sets additional configuration attributes.
+        /// This can be used to tweak internal features of the SDK.
+        public var additionalConfiguration: [String: Any]
```
ncreated, line 83: question, blocker-ish/ Should this be user-facing or is it only consumed by cross-platform SDKs? In the latter case, we should move it to `_internal.` scope.

I did some investigation, we introduced it in https://github.com/DataDog/dd-sdk-ios/pull/456 (the _"why?"_ is not well described) as part of [RUMM-1213](https://datadoghq.atlassian.net/browse/RUMM-1213), which belongs to old React-Native SDK epic. I doubt if we want to carry this on in public API - expecting `[String: Any]` map without giving a clue on keys and values doesn't stand for good API design.

Sources/Datadog/Datadog.swift
```diff
@@ -166,49 +224,113 @@ public class Datadog {
...
-        if Datadog.isInitialized {
+        if CoreRegistry.default is DatadogCore {
             throw ProgrammerError(description: "SDK is already initialized.")
         }
 
-        let serverDateProvider = configuration.common.serverDateProvider ?? DatadogNTPDateProvider()
+        let debug = configuration.processInfo.arguments.contains(LaunchArguments.Debug)
+        if debug {
+            consolePrint("⚠️ Overriding verbosity, and upload frequency due to \(LaunchArguments.Debug) launch argument")
+            Datadog.verbosityLevel = .debug
```
ncreated, line 266: question/ @maxep What is the plan for getting rid of global `.verbosityLevel`? Do we have necessary items in the backlog?

Sources/Datadog/Datadog.swift
```diff
@@ -166,49 +224,113 @@ public class Datadog {
...
-            useLocalEncryption: configuration.common.encryption != nil,
-            useProxy: configuration.common.proxyConfiguration != nil
+            batchSize: Int64(exactly: performance.maxFileSize),
+            batchUploadFrequency: performance.minUploadDelay.toInt64Milliseconds,
+            useLocalEncryption: configuration.encryption != nil,
+            useProxy: configuration.proxyConfiguration != nil
         )
 
-        CITestIntegration.active?.startIntegration()
-
```
ncreated, line 206: question/ Where are we moving this to? Unfortunately, `CITestIntegration` is covered with integration tests and it is easy to mess it.

Tests/DatadogTests/Datadog/DatadogConfiguration.swift
ncreated, line 1: We must rename this file: `DatadogConfiguration` → `DatadogConfigurationTests`.

ncreated (changes requested):
Really good 👍 - nice to see last V1 things disappear.

Sources/Datadog/Datadog.swift
maxep: Agreed, this is used by cross-platforms only therefore it shouldn't exposed here. Thanks for checking!

Sources/Datadog/Datadog.swift
maxep: Is there a plan to get rid of it? 🤔 
Is is used by `DD.logger` which is also global and since it's for development it makes sense keep it this way, no?

Tests/DatadogTests/Datadog/DatadogConfiguration.swift
maxep: oops!

Sources/Datadog/Datadog.swift
maxep: Good catch, it was a miss 🙏

Sources/Datadog/Datadog.swift
ncreated: TBC: sorry, I don't mean "removing it" 🙂. Do we consider attaching `.verbosityLevel` to SDK core instance, or we want to keep it global? Global would mean seeing logs from all instances of the SDK (even not mine). Whatever we decide it is rather low priority.

ncreated: Good idea 👍

ncreated (approved):
🏅 🏆

PR #1339, opened 2023-06-27
RUMM-3295 Generate 2.x `api-surface`

ncreated:
### What and why?

🧰 This PR updates `api-surface-swift` and `api-surface-obj` files with current state of V2, so we can make comparison of public APIs with Android SDK. As part of this work I also fixed obvious leftovers from V2 migration, notably aligned the visibility of `internal` and `public` declaration.

### How?

There are 3 things done in this PR (can be reviewed commit by commit):
- **Update to `api-surface` tool** - it now works with `Package.swift` instead of `*.xcworkspace`. Because we consider dropping `*.xcworkspace` in long term, this change makes us closer to it.
- **Fixed linter configuration** - it now lints `Sources/` and `Tests/` for all products. I spotted this issue by seeing unnecessary things in generated `api-surface-swift`. Pulling the thread of `explicit_top_level_acl` and `lower_acl_than_parent` SwiftLint rules not triggering, turned out the setup wasn't updated in migration. Now fixed and all linter issues are solved.
- **Generated `api-surface-(swift|objc)`** - now we can compare it with Android SDK.

Generation command remains unchanged:
```bash
make api-surface
```

Due to modular setup of V2, I enhanced the `api-surface` with capability of printing indirect ancestors (with `[?]` prefix) if they don't include explicit ACL definition. It is possible to do better (infer the ACL), but for now it can be good enough, e.g.:
```
# ----------------------------------
# API surface for Datadog:
# ----------------------------------

[?] extension InternalExtension where ExtendedType: Datadog
    public static var telemetry: _TelemetryProxy
    public static func set(customVersion: String)
```

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


api-surface-swift
```diff
@@ -1,1398 +1,1281 @@
...
- public func removeAttribute(forKey key: AttributeKey)
- public static var telemetry: _TelemetryProxy
- public static func set(customVersion: String)
+# ----------------------------------
+# API surface for Datadog:
+# ----------------------------------
+
+[?] extension InternalExtension where ExtendedType: Datadog
+    public static var telemetry: _TelemetryProxy
+    public static func set(customVersion: String)
```
ncreated, line 7: The `[?]` prefix means _"we don't know if this is public symbol, but we know it contains public descendants"_.

Typically it should be fine (both compiler and linter's `lower_acl_than_parent ` rule watch it), but not always (e.g. we have some re-exports from `DatadogInternal` that are not listed here or we sometimes publicly extend `DatadogInternal` types in product modules).

ganeshnj (approved):

rtrieu (approved):
Looks good!

maxep (approved):
Thank you 🙏

PR #1343, opened 2023-06-28
Create DatadogCore module

maxep:
### What and why?

ci checks

### How?

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [x] Run smoke tests


ncreated (commented):
As explained in PR description all 3 following `Podfiles` will enable Datadog RUM and Logs. Given that there are 3 setups, **which of these shall we promote in our online docs?**
```ruby
pod 'DatadogSDK'
pod 'DatadogSDK/RUM'
pod 'DatadogSDK/Logs'
```
```ruby
pod 'DatadogSDK/Core'
pod 'DatadogSDK/RUM'
pod 'DatadogSDK/Logs'
```
```ruby
pod 'DatadogCore'
pod 'DatadogRUM'
pod 'DatadogLogs'
```

With all above `Podfiles`, the user code will look the same:
```swift
import DatadogCore
import DatadogRUM
import DatadogLogs

Datadog.initialize(with: config, trackingConsent: .granted)
RUM.enable(with: rumConfig)
Logs.enable(with: logsConfig)
```

I kinda dislike the fact of offering multiple ways of achieving the same result. This will be cumbersome for support engineers and even for us to later validate `Podfiles` like:
```ruby
pod DatadogSDK
pod DatadogRUM
pod DatadogSDK/Logs
```
which look very inconsistent at the first glance, but turn out to be valid if knowing the trick. What are your thoughts on this @maciejburda @ganeshnj ?

@maxep How much is this a result of trying to not break compatibility of original `pod DatadogSDK`? Can we relax some of constraints to make this setup seamless?

ganeshnj:
> ```ruby
> pod 'DatadogCore'
> pod 'DatadogRUM'
> pod 'DatadogLogs'
> ```

I dislike the SDK prefix because it is implicit and doesn't provide any value. I looked at what other do, for example firebase docs https://github.com/firebase/firebase-ios-sdk/blob/master/README.md?plain=1#L71-L74 which is similar to 3rd option.

ie 
```
pod 'DatadogCore'
pod 'DatadogRUM'
pod 'DatadogLogs'
```

however, this shouldn't break the compatibility for easier migration.

maxep:
> This will be cumbersome for Datadog support engineers, can be puzzling for customers and even for us

That's true, and having multiple ways doesn't have much value actually, even for migration..

If we have to promote one, it should be the 3rd where the pods reflect the module names, as @ganeshnj mentioned 👍 
And it's a good opportunity to get rid of the `DatadogSDK*`!

If everyone agree, I will tackle the final touch to podspecs in a next PR. So we don't block dogfooding 🚀  
How does it sound?

maxep:
image

ncreated:
> Also pod name is same as actual package name in import statement for consistency.
> 
> ie
> 
> ```ruby
> pod 'DatadogCore'
> pod 'DatadogRUM'
> pod 'DatadogLogs'
> ```
> 
> however, this shouldn't break the compatibility for easier migration.

I'd be OK with above. Problem is `pod DatadogSDK` was the V1 way. However, we could forward it to `pod DatadogCore` the way proposed in this PR, [deprecate](https://guides.cocoapods.org/syntax/podspec.html#deprecated_in_favor_of) `pod DatadogSDK` in `2.0.0` and stop maintaining it from `2.1`. How does it sound @maxep @ganeshnj ?

Ultimately, starting from `2.1` the only supported way will be:
```ruby
pod 'DatadogCore'
pod 'DatadogRUM'
pod 'DatadogLogs'
```
and we could start `2.0.0` docs with that recommendation.

maxep:
@ncreated yes, to not break `pod install` of v1 users, we can do:

```ruby
Pod::Spec.new do |s|
  s.name         = "DatadogSDK"
  s.module_name  = "DatadogCore"
  ...
  
  s.deprecated_in_favor_of =  'DatadogCore'

  s.source_files = ["DatadogCore/Sources/**/*.swift",
                    "DatadogCore/Private/**/*.{h,m}"]

  s.dependency 'DatadogInternal', s.version.to_s

end
```

OR, combine modules like the following (not sure it works tho):
```ruby
Pod::Spec.new do |s|
  s.name         = "DatadogSDK"
  ...
  
  s.deprecated_in_favor_of = [
    'DatadogCore',
    'DatadogLogs',
    'DatadogRUM',
    'DatadogTrace',
    'DatadogWebViewTracking'
  ]

  s.dependency 'DatadogCore', s.version.to_s
  s.dependency 'DatadogLogs', s.version.to_s
  s.dependency 'DatadogRUM', s.version.to_s
  s.dependency 'DatadogTrace', s.version.to_s
  s.dependency 'DatadogWebViewTracking', s.version.to_s

end
```

ncreated:
@maxep If combined modules can work, and V1-like `Podfile`:
```ruby
pod DatadogSDK
```
will make following possible for `2.0.0`:
```swift
import DatadogCore
import DatadogLogs
import DatadogRUM
// ...
```
with printing the deprecation notice, then I believe this is the most prominent way.

---

Otherwise, we should optimise for:
```ruby
pod DatadogSDK
```
enabling at least `import DatadogCore` in the app with printing respective deprecation warning. Enabling `DatadogRUM` and other modules would be nice, although not mandatory IMO - we really want everyone to replace `pod DatadogSDK` with final variants when on version `2.0.0` if we plan to drop this spec in `2.1`. 

In other words: being able to use **all modules** with `pod DatadogSDK, '2.0.0'` is not a big win IMO.

---

In both ways, we could also think of adding call-to-action into the spec (before GA):
```ruby
# DatadogSDK.podspec
spec.description = <<-DESC
                     <what it used to be>
                     [!] This pod is now deprecated, check our migration guide: ...
                   DESC
```

maxep:
Sounds good 👍  Lets ignore the podspec change in this PR and I will follow up with a dedicated one.

ncreated:
> Sounds good 👍 Lets ignore the podspec change in this PR and I will follow up with a dedicated one.

🤜 🤛 👍

ncreated (approved):
🏋️‍♂️

PR #1344, opened 2023-06-28
Regenerate rum model

maciejburda:
### What and why?

This PR updates RUM model.

### How?

It required a fix for enum without explicit type in generator, after adding fix I generated the code and added blank mocks. They'll be filled in the follow up PR that adds these metrics to the event.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Sources/Datadog/RUM/DataModels/RUMDataModels.swift
+++ b/Sources/Datadog/RUM/DataModels/RUMDataModels.swift
@@ -154,14 +154,14 @@ public struct RUMActionEvent: RUMDataModel {
 
         /// Session-related internal properties
         public struct Session: Codable {
-            /// Session plan: 1 is the plan without replay, 2 is the plan with replay
-            public let plan: Plan
+            /// Session plan: 1 is the plan without replay, 2 is the plan with replay (deprecated)
+            public let plan: Plan?
 
             enum CodingKeys: String, CodingKey {
                 case plan = "plan"
             }
 
-            /// Session plan: 1 is the plan without replay, 2 is the plan with replay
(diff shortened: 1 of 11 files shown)


Sources/Datadog/RUM/DataModels/RUMDataModels.swift
```diff
@@ -300,6 +300,30 @@ public struct RUMActionEvent: RUMDataModel {
         }
     }
 
+    /// Display properties
+    public struct RUMDisplay: Codable {
```
ncreated, line 304: public API convention/ ([“avoid platform self referencing”](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3027638506/Public+API+Convention+in+iOS+SDK+2.x#Rule-7%3A-Use-of-standard-Datadog-and-RUM-terminology)) Because now we nest this model, there is no more need to prefix it with `RUM*`. We add this prefix as part of RUM code decoration:
```swift
if fixedName == "Display" {
   fixedName = "RUMDisplay"
}
```
Otherwise we're self-referencing the product in public API:
```diff
- RUMViewEvent.RUMDisplay
- RUMActionEvent.RUMDisplay
...
+ RUMViewEvent.Display
+ RUMActionEvent.Display
```

tools/rum-models-generator/Tests/CodeGenerationTests/Fixtures/fixture-schema-with-typeless-enum.json
ncreated, line 1: 💪

ncreated (approved):
One small remark on the public API name, other than this looks great 👌 🎯

PR #1353, opened 2023-06-30
RUMM-2987 Use instance name for data persistency

maxep:
### What and why?

Use core instance name instead of client-token for persistent dat folder.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


maciejburda (approved):

ganeshnj:
This change looks 👌  but should we consider client token + instance name both for it?

Imagine a scenario where customers having their own remote config for client token.

ganeshnj (approved):

maxep:
> we consider client token + instance name both for it?
> Imagine a scenario where customers having their own remote config for client token.

@ganeshnj That's the thing, we don't want to set a new folder if the customer ever rotates their client token. That's why we now use the instance name.

Issue #1292, opened 2023-05-25
APM Traces not connecting in RUM Views


lfurukawa-ml:
Hi guys,

We are trying to connect APM traces with RUM views, but when viewing the session in RUM, the trace information does not appear in the "Trace" tab:

<img width="1036" alt="img_01" src="https://github.com/DataDog/dd-sdk-ios/assets/102163451/bbccc28b-f5b6-424e-ab9a-3c9ab67ce5e7">

Upon examining the trace information in APM, we can see that it has the RUM view's attribute "view.id":

APM Trace:
<img width="733" alt="img_03" src="https://github.com/DataDog/dd-sdk-ios/assets/102163451/bf69e549-e076-41c3-85e3-4eeeb09229a9">

RUM view attributes:

<img width="1201" alt="img_02" src="https://github.com/DataDog/dd-sdk-ios/assets/102163451/4f35179c-b9ec-420b-bd4f-371483003d91">

We followed the recommended SDK configuration mentioned in this [issue](https://github.com/DataDog/dd-sdk-ios/issues/424) 

I noticed in the issue that you mentioned it might take some time for APM traces to appear in RUM, but even after waiting for 12 hours, the trace did not appear in the associated RUM view.

> **3. APM Trace <> RUM View connection.**
> Similar to Log <> RUM View connection, all APM traces created with Global.sharedTracer should be linked to the active RUM View and appear on the "Traces" tab. It may just take some more time until the trace appears there (logs appear almost instantly).

Currently, we are using the Datadog SDK 1.16.0, and we have tried updating it to version 1.19.0, but it still didn't work.

Could you please help us in identifying the problem?

Thank you!

ncreated:
Hey @lfurukawa-ml. Thanks for reporting the problem - I confirm that I'm seeing this too 👍. We need to discuss what exactly got broken and we will come back to you with updates - please stay tuned.

lfurukawa-ml:
hi @ncreated thanks for the prompt reply! we appreciate it

I wanted to use this thread to report another issue we've come across during our APM Traces testing. We have noticed that when creating traces at moments where a view has not yet been rendered, the span is sent without any RUM session information. 
For example, if we create a span during the application launch to measure the execution time of a method inside `applicationDidFinishLaunchingWithOptions` (which is automatically associated with the "view" ApplicationLaunch), this span is sent without any RUM information.

lfurukawa-ml:
hi @ncreated
[...]

PR #1355, opened 2023-06-30
RUMM-3320 [V1] Fix APM local spans not correlating with RUM views

plousada:
### What and why?

APM <-> RUM correlation was not working for local traces due to mismatch in span tag format between the SDK and APM Backend.

fix: https://github.com/DataDog/dd-sdk-ios/issues/1292

### How?

This PR removes the direct use of `RUMContextAttributes` in both Log and Span encoding. 
Instead, a mapping layer is added that transforms keys into the required format for each product. 

On the Example app, changes are made to allow enabling a simple debugging option. It's now possible to single a single log, a single span, and a single traced request on a particular RUM session.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Datadog/Example/Debugging/DebugRUMSessionViewController.swift
+++ b/Datadog/Example/Debugging/DebugRUMSessionViewController.swift
@@ -39,6 +39,12 @@ private class DebugRUMSessionViewModel: ObservableObject {
     @Published var errorMessage: String = ""
     @Published var resourceKey: String = ""
 
+    @Published var logMessage: String = ""
+    @Published var spanOperationName: String = ""
+    @Published var instrumentedRequestURL: String = "https://api.shopist.io/checkout.json"
+
+    var urlSessions: [URLSession] = []
+
     func startView() {
         guard !viewKey.isEmpty else {
             return
@@ -74,6 +80,10 @@ private class DebugRUMSessionViewModel: ObservableObject {
         )
 
         Global.rum.addUserAction(type: .custom, name: actionName)
+
+        Thread.sleep(forTimeInterval: 0.05)
+        sendSpan()
+
         self.actionName = ""
     }
 
@@ -115,6 +125,45 @@ private class DebugRUMSessionViewModel: ObservableObject {
         self.resourceKey = ""
     }
 
+    func sendLog() {
+        logger.debug(logMessage)
+        logMessage = ""
+    }
+
+    func sendSpan() {
+        let span = Global.sharedTracer.startRootSpan(operationName: spanOperationName, tags: [:])
+        Thread.sleep(forTimeInterval: 0.1)
+        span.finish()
+        spanOperationName = ""
+    }
+
+    func sendPOSTRequest() {
+        guard let url = URL(string: instrumentedRequestURL) else {
+            print("🔥 POST Request not sent - invalid url: \(instrumentedRequestURL)")
+            return
+        }
+        guard let host = url.host else {
+            print("🔥 POST Request not sent - invalid url host: \(instrumentedRequestURL)")
+            return
+        }
+
+        let delegate = DDURLSessionDelegate(additionalFirstPartyHosts: [host])
+        let session = URLSession(configuration: .ephemeral, delegate: delegate, delegateQueue: nil)
+
+        var request = URLRequest(url: url)
+        request.httpMethod = "POST"
+        let task = session.dataTask(with: request) { _, _, error in
+            if let error = error {
+                print("🌍🔥 POST \(url) completed with network error: \(error)")
+            } else {
+                print("🌍 POST \(url) sent successfully")
+            }
+        }
+        task.resume()
+
+        urlSessions.append(session) // keep session
+    }
+
     // MARK: - Private
 
     private func modifySessionItem(type: SessionItemType, label: String, change: (inout SessionItem) -> Void) {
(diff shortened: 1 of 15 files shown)


Tests/DatadogTests/Datadog/LoggerTests.swift
```diff
@@ -562,7 +562,7 @@ class LoggerTests: XCTestCase {
 
         logMatchers.forEach {
             $0.assertValue(
-                forKeyPath: RUMContextAttributes.IDs.sessionID,
+                forKeyPath: "application_id",
```
plousada, line 565: This is currently enforcing the schema used by logs backend. 
I'm wondering if it would be preferable to use the output of the mapping function `mapInternalAttributeKey` here instead of the actual expected key and leave that enforce to E2E tests.

Tests/DatadogTests/Datadog/WebView/WKUserContentController+DatadogTests.swift
```diff
@@ -180,7 +180,7 @@ class WKUserContentController_DatadogTests: XCTestCase {
         )
 
         let messageHandler = try XCTUnwrap(controller.messageHandlers.first?.handler) as? DatadogMessageHandler
-        let webLogMessage = MockScriptMessage(body: #"{"eventType":"log","event":{"date":1635932927012,"error":{"origin":"console"},"message":"console error: error","session_id":"0110cab4-7471-480e-aa4e-7ce039ced355","status":"error","view":{"referrer":"","url":"https://datadoghq.dev/browser-sdk-test-playground"}},"tags":["browser_sdk_version:3.6.13"]}"#)
+        let webLogMessage = MockScriptMessage(body: #"{"eventType":"log","event":{"date":1635932927012,"error":{"origin":"console"},"message":"console error: error","session.id":"0110cab4-7471-480e-aa4e-7ce039ced355","status":"error","view":{"referrer":"","url":"https://datadoghq.dev/browser-sdk-test-playground"}},"tags":["browser_sdk_version:3.6.13"]}"#)
```
fuzzybinary, line 183: Was this mock message made to be accurate to what the Browser SDK sends us? If so we probably shouldn't change it and verify we still send the proper attributes.

Sources/Datadog/Tracing/TracingV2Configuration.swift
```diff
@@ -72,7 +72,7 @@ internal struct TracingMessageReceiver: FeatureMessageReceiver {
     ///
     /// - Parameter context: The updated core context.
     private func update(context: DatadogContext) -> Bool {
-        if let attributes: [String: String] = context.featuresAttributes["rum"]?.ids {
+        if let attributes: [String: String?] = context.featuresAttributes["rum"]?.ids {
```
fuzzybinary, line 75: I'm curious why this change was needed?

fuzzybinary (approved):

Tests/DatadogTests/Datadog/WebView/WKUserContentController+DatadogTests.swift
plousada: You're right, my initial approach doesn't address logs coming from Browser SDK correctly.
Thanks for raising it.

Sources/Datadog/RUM/Integrations/RUMContextAttributes.swift
```diff
@@ -10,11 +10,11 @@ import Foundation
 internal enum RUMContextAttributes {
     internal enum IDs {
         /// The ID of RUM application (`String`).
-        internal static let applicationID = "application_id"
+        internal static let applicationID = "application.id"
 
         /// The ID of current RUM session (standard UUID `String`, lowercased).
         /// In case the session is rejected (not sampled), RUM context is set to empty (`[:]`) in core.
-        internal static let sessionID = "session_id"
+        internal static let sessionID = "session.id"
```
ncreated, line 17: blocker 🚫/ This change break Session Replay, which expects `"application_id"` and `"session_id"`. With being unable to recognise these new values SR thinks there is no RUM, so it doesn't record anything.

1. We need to update these values [here](https://github.com/DataDog/dd-sdk-ios/blob/10135195869d7f785f7a63201e4ecd84795b0490/DatadogSessionReplay/Sources/DatadogCoreIntegration/RUMContextReceiver.swift#L14-L16).
2. Also please update these values in comments of RUM dependency definition within SR module [here](https://github.com/DataDog/dd-sdk-ios/blob/ncreated/RUMM-3320/GH-1292-fix-apm-with-rum/DatadogSessionReplay/Sources/DatadogCoreIntegration/RUMDependency.swift) - this definition is a central place where SR defines everything it expects or provides to other modules in SDK context.

I bump the need for 2 things:
- Integration Tests for SR: it is now possible on `feature/v2` - I created [REPLAY-1863](https://datadoghq.atlassian.net/browse/REPLAY-1863)
- Integration Unit Tests: to validate feature's contract in unit tests - @maxep do we have JIRA for this?

Sources/Datadog/Logging/LoggingV2Configuration.swift
```diff
@@ -27,6 +27,17 @@ internal func createLoggingConfiguration(
...
+internal func mapInternalAttributeKey(_ originalAttribute: String) -> String {
+    switch originalAttribute {
+    case "application.id":
+        return "application_id"
+    case "session.id":
+        return "session_id"
+    default:
+        return originalAttribute
+    }
+}
```
ncreated, line 39: suggestion/ Could we declare constants for `"application.id"` and `"session.id"` as part of `RUMDependency` namespace defined in `Logging` module? Similar to how we define [`RUMDependency`](https://github.com/DataDog/dd-sdk-ios/blob/ncreated/RUMM-3320/GH-1292-fix-apm-with-rum/DatadogSessionReplay/Sources/DatadogCoreIntegration/RUMDependency.swift) in Session Replay - single place with listing all attributes that features exchange between each other.

cc @maxep

Sources/Datadog/Tracing/TracingV2Configuration.swift
```diff
@@ -18,6 +18,21 @@ internal func createTracingConfiguration(intake: URL) -> DatadogFeatureConfigura
...
+    case "session.id":
+        return "_dd.session.id"
+    case "view.id":
+        return "_dd.view.id"
+    case "user_action.id":
+        return "_dd.action.id"
+    default:
+        return originalTag
+    }
+}
```
ncreated, line 34: suggestion/ Similar ask to the one in Logging:
> suggestion/ Could we declare constants for `"application.id"` and `"session.id"` as part of `RUMDependency` namespace defined in `Tracing` module? Similar to how we define [`RUMDependency`](https://github.com/DataDog/dd-sdk-ios/blob/ncreated/RUMM-3320/GH-1292-fix-apm-with-rum/DatadogSessionReplay/Sources/DatadogCoreIntegration/RUMDependency.swift) in Session Replay - single place with listing all attributes that features exchange between each other.

cc @maxep

ncreated (changes requested):

Sources/Datadog/Logging/LoggingV2Configuration.swift
ncreated: Because we work with V1 code (no modules yet), it would have to be `enum LogsRUMDependency` (or similar) to scope it properly.

Sources/Datadog/RUM/Integrations/RUMContextAttributes.swift
maxep: > Integration Unit Tests: to validate feature's contract in unit tests - @maxep do we have JIRA for this?

Sources/Datadog/Logging/LoggingV2Configuration.swift
maxep: Having constants makes sense, should be defined alongside the message receiver IMO.

Sources/Datadog/Tracing/TracingV2Configuration.swift
ganeshnj: If the constants are shared eg "application_id", declaring them each module (Tracing, Session Replay etc) seems error prone. IMO, it should go atleast in Internal module which is shared by all other modules.

Not related to this, but having an API package for each feature makes these things easier to maintain, where there could be `RUMApi` module with plain structs, APIs and constants (ideal world) which can be taken as dependency by others without all the baggage of RUM.

Sources/Datadog/Logging/LoggingV2Configuration.swift
ncreated: Message receiver can be a good place to define "received" attributes. Where to define "provided" attributes?

Looking at mentioned [RUMDependency](https://github.com/DataDog/dd-sdk-ios/blob/ncreated/RUMM-3320/GH-1292-fix-apm-with-rum/DatadogSessionReplay/Sources/DatadogCoreIntegration/RUMDependency.swift) in SR there are two parts:
```swift
internal enum RUMDependency {
   // MARK: Contract from RUM to SR:
   static let rumBaggageKey = "rum"
   static let serverTimeOffsetKey = "server_time_offset"
   static let ids = "ids"

   // MARK: Contract from SR to RUM (mirror of `SessionReplayDependency` in RUM):
   static let srBaggageKey = "session-replay"
   static let hasReplay = "has_replay"
}
```
and I believe it is good and clear enough - single place describing the whole contract.

If we want to separate it, it's OK for me as long as it is clear and consistent across all modules.

maxep: > and I believe it is good and clear enough - single place describing the whole contract.

Yep, totally. Just to add my 2 cents: I don't think we should enforce any convention across modules on that. Some Features will use/share a lot of context values, some won't. We can have some flexibility on how Features are implemented internally IMHO

ncreated (approved):
@maxep || @ganeshnj I pushed this topic forward and now it is ready for review. 
* Only fixing Session Replay compared to previous work (the change here has low impact as we don't provide SR in V1 except private beta branch, but I did it anyway). 
* Like I mentioned on Slack, this is only minimal change to make it work. 
* We're fixing V1 here - V2 fix is in #1375 
* This PR is targeting `release/1.22.0` branch - I will do release after merge.

CI is green, Bitrise checks are false:

Screenshot 2023-07-18 at 14 11 59

Datadog/Example/Debugging/DebugRUMSessionViewController.swift
```diff
@@ -39,6 +39,12 @@ private class DebugRUMSessionViewModel: ObservableObject {
     @Published var errorMessage: String = ""
     @Published var resourceKey: String = ""
 
+    @Published var logMessage: String = ""
+    @Published var spanOperationName: String = ""
+    @Published var instrumentedRequestURL: String = "https://api.shopist.io/checkout.json"
```
ganeshnj, line 44: currently it returns 
```json
{
    "status": 404,
    "error": "Not Found"
}
```

expected ?

ganeshnj (approved):

maxep (approved):
Looks great! I left a question.

Sources/Datadog/Tracing/Span/SpanEventEncoder.swift
```diff
@@ -243,8 +243,10 @@ internal struct SpanEventEncoder {
         }
 
         try span.tags.forEach {
-            let metaKey = "meta.\($0.key)"
-            try container.encode($0.value, forKey: DynamicCodingKey(metaKey))
+            if $0.value != "null" {
```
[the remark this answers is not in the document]
ncreated: No, if context includes `nil` value (e.g. `["user_action.id": nil]`) that was being encoded as `"null"` here. But it is a good call-out. I fixed it now in a better way, by doing this after context retrieval in `TracingV2Configuration.swift`:
```swift
let attributes = attributes.compactMapValues { $0 }
```
(it's analogous to the fix in V2)

Datadog/Example/Debugging/DebugRUMSessionViewController.swift
```diff
@@ -74,6 +80,10 @@ private class DebugRUMSessionViewModel: ObservableObject {
         )
 
         Global.rum.addUserAction(type: .custom, name: actionName)
+
+        Thread.sleep(forTimeInterval: 0.05)
+        sendSpan()
+
         self.actionName = ""
```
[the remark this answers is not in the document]
ncreated: ✅ Done

Datadog/Example/Debugging/DebugRUMSessionViewController.swift
ncreated: It is `POST` endpoint (`GET` always returns `404`), but getting some % of errors on it is expected as this is part of demo environment 👍.

PR #1357, opened 2023-07-03
RUMM-3387 Align webview tracking v2 public api

maxep:
### What and why?

Align `WebViewTracking` public api with Android.

### How?

```swift
public final class WebViewTracking {
    /// Enables SDK to correlate Datadog RUM events and Logs from the WebView with native RUM session.
    ///
    /// If the content loaded in WebView uses Datadog Browser SDK (`v4.2.0+`) and matches specified
    /// `hosts`, web events will be correlated with the RUM session from native SDK.
    ///
    /// - Parameters:
    ///   - webView: The web-view to track.
    ///   - hosts: A set of hosts instrumented with Browser SDK to capture Datadog events from.
    ///   - core: Datadog SDK core to use for tracking.
    public static func enable(
        webView: WKWebView,
        hosts: Set<String> = [],
        in core: DatadogCoreProtocol = CoreRegistry.default
    )

    /// Disables Datadog iOS SDK and Datadog Browser SDK integration.
    ///
    /// Removes Datadog's ScriptMessageHandler and UserScript from the caller.
    /// - Note: This method **must** be called when the webview can be deinitialized.
    /// 
    /// - Parameter webView: The web-view to stop tracking.
    public static func disable(webView: WKWebView)
}
```

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [x] Run smoke tests


DatadogWebViewTracking/Sources/WebViewTracking.swift
```diff
@@ -4,55 +4,126 @@
...
+    ///
+    /// If the content loaded in WebView uses Datadog Browser SDK (`v4.2.0+`) and matches specified
+    /// `hosts`, web events will be correlated with the RUM session from native SDK.
+    ///
+    /// - Parameters:
+    ///   - webview: The web-view to track.
+    ///   - hosts: A set of hosts instrumented with Browser SDK to capture Datadog events from.
+    ///   - core: Datadog SDK core to use for tracking.
+    public static func enable(
+        webview: WKWebView,
```
ncreated, line 36: I think `webview` is not the right form, should be `webView` (it is `WKWebView` not `WKWebview`).

DatadogWebViewTracking/Sources/WebViewTracking.swift
```diff
@@ -4,55 +4,126 @@
...
+extension WebViewTracking: InternalExtended { }
+extension InternalExtension where ExtendedType: WebViewTracking {
+    /// Creates a web-view message emitter for cross-platform.
+    ///
+    /// Cross platform SDKs should instantiate a `MessageEmitter` implementation from
+    /// this method and pass WebView related messages using the message bus of the core.
+    ///
+    /// - Parameter core: The Datadog SDK core instance
+    /// - Returns: A `MessageEmitter` instance
+    public static func messageEmitter(in core: DatadogCoreProtocol) -> MessageEmitter {
```
maxep, line 126: I'm introducing this interface for CP SDKs so they can transmit WebView messages. It requires `MessageEmitter` and `WebViewTrackingMessage` to be public.

DatadogWebViewTracking/Sources/WebViewTracking.swift
```diff
@@ -4,55 +4,126 @@
...
+import Foundation
 import DatadogInternal
 
-/// Defines methods to send WebView related information
-internal protocol WebViewTracking {
-    /// Sends a message
-    /// - Parameter message: The message to send
-    func send(message: WebViewTrackingMessage) throws
-}
+#if !os(tvOS)
```
ganeshnj, line 10: question

seems like `WebViewTracking` should not have tvOS as supported platform. If true, please make sure we are not enabling it accidentally.

DatadogWebViewTracking/Sources/WebViewTracking.swift
```diff
@@ -4,55 +4,126 @@
...
+///
+/// # Prerequisites:
+/// Set up the web page you want rendered on your mobile iOS and tvOS application with the RUM Browser SDK
+/// first. For more information, see [RUM Browser Monitoring](https://docs.datadoghq.com/real_user_monitoring/browser/#npm).
+///
+/// You can perform the following:
+/// - Track user journeys across web and native components in mobile applications
+/// - Scope the root cause of latency to web pages or native components in mobile applications
+/// - Support users that have difficulty loading web pages on mobile devices
+public final class WebViewTracking {
```
ganeshnj, line 24: nit

```suggestion
public enum WebViewTracking {
```

personal preference when a type name is only used for namespace purpose only. I think now it has no instance methods.

DatadogWebViewTracking/Sources/WebViewTracking.swift
maxep: Yes, I didn't want to change current binary support but I think the entire `DatadogWebViewTracking` shouldnt be built for tvOS.

DatadogWebViewTracking/Sources/WebViewTracking.swift
ncreated: question/ Could `MessageEmitter` and `WebViewTrackingMessage` be nested under `InternalExtension` namespace? If so, let's do. Otherwise we pollute app's namespace with indirect and common types like `JSON` typealias.

DatadogWebViewTracking/Sources/WebViewMessage.swift
```diff
@@ -6,10 +6,10 @@
...
 import Foundation
 
-internal typealias JSON = [String: Any]
+public typealias JSON = [String: Any]
 
 /// Intermediate type to parse WebView messages and send them to the message bus
-internal enum WebViewTrackingMessage {
+public enum WebViewMessage {
     /// A log message with a JSON payload
     case log(JSON)
```
ncreated, line 14: suggestion/ Move `JSON` to `WebViewMessage` typealias. Otherwise we expose so popular type in public that it can commonly clash with user code.

DatadogWebViewTracking/Sources/WebViewTracking.swift
maxep: MessageEmitter is a protocol and cannot be nested, but it doesn't need 
to be a protocol.. Let me see how I can hide all of that 👍

maxep:
@ncreated I was able to hide everything using an abstract class definition.

ncreated (approved):
> @ncreated I was able to hide everything using an abstract class definition.

🚀 Awesome! "importance of user namespace" > "our convenience" 👍

PR #1363, opened 2023-07-06
REPLAY-1868 Configure Session Replay uploader to be more eager

ncreated:
### What and why?

📦 This PR tunes the performance of SR uploader, making it more eager. This is to increase reliability of replay uploads.

### How?

The `PerformancePreset` for Session Replay is now overwritten with values that correspond to `uploadFrequency: .fast` and `batchSize: .small` setup.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/DatadogSessionReplay/Sources/Drafts/SessionReplayFeature.swift
+++ b/DatadogSessionReplay/Sources/Drafts/SessionReplayFeature.swift
@@ -62,7 +62,9 @@ internal class SessionReplayFeature: DatadogFeature, SessionReplayController {
         self.requestBuilder = RequestBuilder(customUploadURL: configuration.customUploadURL)
         self.performanceOverride = PerformancePresetOverride(
             maxFileSize: UInt64(10).MB,
-            maxObjectSize: UInt64(10).MB
+            maxObjectSize: UInt64(10).MB,
+            meanFileAge: 5, // equivalent of `batchSize: .small` - see `DatadogCore.PerformancePreset`
+            minUploadDelay: 1 // equivalent of `uploadFrequency: .frequent` - see `DatadogCore.PerformancePreset`
         )
     }
 
--- a/DatadogSessionReplay/Sources/Recorder/WindowObserver/KeyWindowObserver.swift
+++ b/DatadogSessionReplay/Sources/Recorder/WindowObserver/KeyWindowObserver.swift
@@ -17,7 +17,6 @@ internal class KeyWindowObserver: AppWindowObserver {
         if #available(iOS 13.0, tvOS 13.0, *) {
             return findONiOS13AndLater()
         } else {
-            assertionFailure("TODO: RUMM-2409 `AppWindowObserver` isn't yet ready for this version of OS")
             return nil
         }
     }
--- a/Sources/Datadog/Core/PerformancePreset.swift
+++ b/Sources/Datadog/Core/PerformancePreset.swift
@@ -80,12 +80,11 @@ internal extension PerformancePreset {
             }
         }()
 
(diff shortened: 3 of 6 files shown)


DatadogSessionReplay/Sources/Recorder/WindowObserver/KeyWindowObserver.swift
```diff
@@ -17,7 +17,6 @@ internal class KeyWindowObserver: AppWindowObserver {
         if #available(iOS 13.0, tvOS 13.0, *) {
             return findONiOS13AndLater()
         } else {
-            assertionFailure("TODO: RUMM-2409 `AppWindowObserver` isn't yet ready for this version of OS")
```
ncreated, line 20: 🧽 This is just cleanup. We don't yet support it below iOS 13.0 - we will add it later.

Sources/Datadog/Core/PerformancePreset.swift
```diff
@@ -80,20 +80,18 @@ internal extension PerformancePreset {
             }
         }()
 
-        let uploadDelayFactors: (initial: Double, default: Double, min: Double, max: Double, changeRate: Double) = {
+        let uploadDelayFactors: (initial: Double, min: Double, max: Double, changeRate: Double) = {
```
ncreated, line 83: `default` was unused ✨

Sources/Datadog/Core/PerformancePreset.swift
```diff
@@ -126,19 +124,19 @@ internal extension PerformancePreset {
         self.uploadDelayChangeRate = uploadDelayFactors.changeRate
     }
 
-    func updated(with: PerformancePresetOverride) -> PerformancePreset {
+    func updated(with override: PerformancePresetOverride) -> PerformancePreset {
```
maciejburda, line 127: nice 🙌

DatadogSessionReplay/Sources/Drafts/SessionReplayFeature.swift
```diff
@@ -62,7 +62,9 @@ internal class SessionReplayFeature: DatadogFeature, SessionReplayController {
         self.requestBuilder = RequestBuilder(customUploadURL: configuration.customUploadURL)
         self.performanceOverride = PerformancePresetOverride(
             maxFileSize: UInt64(10).MB,
-            maxObjectSize: UInt64(10).MB
+            maxObjectSize: UInt64(10).MB,
+            meanFileAge: 5, // equivalent of `batchSize: .small` - see `DatadogCore.PerformancePreset`
+            minUploadDelay: 1 // equivalent of `uploadFrequency: .frequent` - see `DatadogCore.PerformancePreset`
```
maciejburda, line 67: a little bit ugly. Maybe we could share those enums 🤔 

Is it any easier in v2?

maciejburda (approved):
LGTM

DatadogSessionReplay/Sources/Drafts/SessionReplayFeature.swift
ncreated: In a simple way we can't because it is defined in `DatadogCore` and SR doesn't depend on that module. It depends on `DatadogInternal` and moving this enum in there would further break the isolation (it's already partially broken due to `PerformancePresetOverride`). I think long term we rather won't need overrides and invent something better for this problem.

@maxep what's your take on this?

maxep: I think settings by value enable greater flexibility. If that was exposed to user, then yes enum would have been best but since this is an internal contract between the core and Features, we can use values 👍

maxep (approved):
👍

PR #1368, opened 2023-07-11
REPLAY-1864 Enable SR Snapshot Tests on CI

ncreated:
### What and why?

📦 This PR enables SR Snapshot Tests on CI. It finalizes the work previously done in:
* https://github.com/DataDog/dd-sdk-ios/pull/1162
* https://github.com/DataDog/dd-sdk-ios/pull/1299

By running snapshot tests for each SR's PR we will ensure there is no regression in Session Replay recorder logic.

### How?

Two `make` commands are introduced:
* `make sr-push-snapshots` - To push new PNG files to [dd-mobile-session-replay-snapshots](https://github.com/DataDog/dd-mobile-session-replay-snapshots) repo (internal) and create their pointers in `dd-sdk-ios/.../_snapshots_/pointers/`.
* `make sr-pull-snapshots` - To pull all PNG files referenced by current pointers in `dd-sdk-ios` from [dd-mobile-session-replay-snapshots](https://github.com/DataDog/dd-mobile-session-replay-snapshots).

**Note:** Both commands require that [gh](https://cli.github.com/) CLI is installed and authorized on the local machine (which is already true for CI). This requirement comes from the shortcut we took when implementing #1299.

To stabilize existing snapshots on CI and make sure there is no ❄️ flakiness, I made few smaller updates to fixtures, notably:
* always using `timeZone` when configuring dates in UI;
* performing transitions without any animations;
* bumping up few timeouts, only for CI.

As a result of diagnosing and troubleshooting builds, few handy `XCTAttachements` are added for each test in case of failure. Attachments include side-by-side comparison of expected vs actual snapshot, diff image to surface difference, wireframes JSON and visual `Blueprint` model for rendered image.

#### Adding / modifying snapshot test

Typical workflow for adding or modifying snapshot tests is:
* Add / change test in `SRSnapshotTests` project.
* Generate new PNG files as usual (with `recordingMode = true`).
* Run `make sr-push-snapshots` to push PNGs and create pointers.
* Commit pointers in `dd-sdk-ios`.
* Open a PR.

#### Running snapshot tests on CI

The CI will then:
* Clone `dd-sdk-ios` commit.
* Run `make sr-pull-snapshots` to fetch PNGs referenced by pointers from cloned commit.
* Run `SRSnapshotTests`.

#### Syncing snapshot tests modified by someone else

Similarly, to sync local PNGs with certain `dd-sdk-ios` commit (e.g. after changing the branch):
* Change git branch in `dd-sdk-ios`.
* Run `make sr-pull-snapshots` to fetch PNGs referenced by pointers in current commit.
* Run `SRSnapshotTests`.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


DatadogSessionReplay/SRSnapshotTests/SRHost/Info.plist
```diff
@@ -2,6 +2,8 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>IsRunningOnCI</key>
+	<string>$(IS_CI)</string>
```
maxep, line 6: /question do we have to configure this or is it set by bitrise?

DatadogSessionReplay/SRSnapshotTests/SRSnapshotTests/Utils/ImageComparison.swift
```diff
@@ -30,66 +30,116 @@ internal struct ImageLocation {
...
-    DDAssertSimulatorDevice("iPhone14,7", "16.2", file: file, line: line)
+internal extension XCTestCase {
+    /// Compares `newImage` against the snapshot saved in `snapshotLocation` OR updates stored snapshot image data (if `record == true`).
+    func DDAssertSnapshotTest(
+        newImage: UIImage,
+        snapshotLocation: ImageLocation,
+        record: Bool,
+        file: StaticString = #filePath,
+        line: UInt = #line
+    ) {
```
maxep, line 41: I don't see `self` being used so it would best to keep it as function rather than instance method, like any other `DDAssert*`. Using function allows to use the assertion outside of a `XCTestCase` subclass, the can be useful when combining assertions for example.

maxep (commented):
Looks really great!

I minor comment: I think it should be best to keep `DDAssert*` as function.

DatadogSessionReplay/SRSnapshotTests/SRHost/Info.plist
ncreated: This is already set on CI:
https://github.com/DataDog/dd-sdk-ios/blob/670006d6c3f41ad2652c6f154fd4fda7d04a04a8/Makefile#L38

We use it in other projects through the same approach: by linking `xcconfig` and exporting it to `Info.plist`.

maxep (approved):
LGTM!

PR #1380, opened 2023-07-19
Fix xcframework GH assets

### LINKED ISSUE #1318: XCFramework - zip contained on release is not found by Carthage
### Change name of artifact

Carthage can download the xcframework from github releases. But the name should be DataDog.xcframework.zip instead of DataDog.zip.

See https://github.com/Carthage/Carthage#archive-prebuilt-frameworks-into-zip-files

maxep:
### What and why?

Update GH asset name for pre-build binaries following [Carthage requirements](https://github.com/Carthage/Carthage#archive-prebuilt-frameworks-into-zip-files).

fix #1318

### How?

With v2 and module stability enabled, GH assets will be named `Datadog.xcframework.zip`.
Prior to v2 and since we were not compliant with Carthage, we can always include the Xcode version in the asset name. The `add_xcode_version_to_github_asset` parameter is now unnecessary and has been removed.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/tools/distribution/release.py
+++ b/tools/distribution/release.py
@@ -45,12 +45,6 @@ if __name__ == "__main__":
         default=os.environ.get('DD_RELEASE_OVERWRITE_GITHUB') == '1'
     )
     parser.add_argument(
-        "--add-xcode-version-to-github-asset",
-        action='store_true',
-        help="Add Xcode version to the GH Release asset.",
-        default=os.environ.get('DD_ADD_XCODE_VERSION_TO_GITHUB_ASSET') == '1'
-    )
-    parser.add_argument(
         "--dry-run",
         action='store_true',
         help="Run validation but skip publishing.",
@@ -63,7 +57,6 @@ if __name__ == "__main__":
         only_github = True if args.only_github else False
         only_cocoapods = True if args.only_cocoapods else False
         overwrite_github = True if args.overwrite_github else False
-        add_xcode_version_to_github_asset = True if args.add_xcode_version_to_github_asset else False
         dry_run = True if args.dry_run else False
 
         # Validate arguments:
@@ -72,9 +65,6 @@ if __name__ == "__main__":
 
         if only_cocoapods and overwrite_github:
             raise Exception('`--overwrite-github` and `--only-cocoapods` cannot be used together.')
-
-        if only_cocoapods and add_xcode_version_to_github_asset:
-            raise Exception('--add-xcode-version-to-github-asset` and `--only-cocoapods` cannot be used together.')
 
         version = Version(git_tag)
         if not version:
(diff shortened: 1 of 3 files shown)


Package.swift
```diff
@@ -29,10 +29,6 @@ let package = Package(
             name: "DatadogRUM",
             targets: ["DatadogRUM"]
         ),
-        .library(
-            name: "DatadogSessionReplay",
-            targets: ["DatadogSessionReplay"]
-        ),
```
maxep, line 35: This was exposed for `dogfooding`.

tools/distribution/src/utils.py
```diff
@@ -71,9 +71,11 @@ def read_sdk_version() -> str:
     Reads SDK version from 'Sources/Datadog/Versioning.swift'.
     """
     file = 'DatadogCore/Sources/Versioning.swift'
-    version_regex = r'^.+\"([0-9]+\.[0-9]+\.[0-9]+[\-a-z0-9]*)\"'  # e.g. 'internal let __sdkVersion = "1.8.0-beta1"'
```
maxep, line 74: This regex is no longer valid following 856ed98b034713fc340452ff8d26a2ebc3834dec
It has been replace with [VERSION_PATTERN](https://packaging.pypa.io/en/latest/version.html#packaging.version.VERSION_PATTERN)

tools/distribution/src/utils.py
```diff
@@ -71,9 +71,11 @@ def read_sdk_version() -> str:
...
 
-    versions: [str] = []
+    # https://packaging.pypa.io/en/latest/version.html#packaging.version.VERSION_PATTERN
+    version_regex = re.compile(r'^.+\"' + VERSION_PATTERN + r'\"$', re.VERBOSE | re.IGNORECASE) # e.g. 'internal let __sdkVersion = "1.8.0-beta.1"
```
ganeshnj, line 76: for such trivial case, we should use simply parse the string than taking a dependency 

1.8.0-beta1
```
split by -
- 1.8.0
  - split by .
    - major: 1
    - minor: 8
    - patch: 0 
- beta1
    - read from back until there is digit
      - beta
      - 1
```

tools/distribution/release.py
```diff
@@ -73,9 +72,6 @@
         if only_cocoapods and overwrite_github:
             raise Exception('`--overwrite-github` and `--only-cocoapods` cannot be used together.')
 
-        if only_cocoapods and add_xcode_version_to_github_asset:
```
ganeshnj, line 76: image
can we remove ^^ also?

ncreated (approved):
Looks good 👍 - let's also update [CI setup for iOS SDK](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/2436007444/CI+setup+for+iOS+SDK#tagged_commit) (internal) accordingly.

tools/distribution/src/utils.py
maxep: The dependency was already introduced in 856ed98b034713fc340452ff8d26a2ebc3834dec and I think it is best to rely on the same pattern everywhere 👍

tools/distribution/release.py
maxep: good catch!

PR #1382, opened 2023-07-20
ci: fix scheme name for nightly unit tests

ganeshnj:
### What and why?

```
Failed to execute Step:
  command failed with exit status 65 (xcodebuild "-workspace" "/Users/vagrant/git/Datadog.xcworkspace" "-scheme" "Datadog iOS" "test" "-destination" "id=E467E978-443F-4E53-9A7E-C146036E2E69" "-resultBundlePath" "/var/folders/sj/_4jj239s3j33w5w1bdh98yj80000gn/T/XCUITestOutput2480199168/Test-Datadog iOS.xcresult" "-retry-tests-on-failure" "-test-iterations" "2" "-xcconfig" "/var/folders/sj/_4jj239s3j33w5w1bdh98yj80000gn/T/3713649878/temp.xcconfig"):
    xcodebuild: error: The workspace named "Datadog" does not contain a scheme named "Datadog iOS". The "-list" option can be used to find the names of the schemes in the workspace.
```

### How?

`Datadog iOS` no longer exists rather we have `DatadogCore iOS` which currently runs all the targets. Once V2 migration is fully done, we would have to add list of all the targets (given no workspace/project file be there)

Added printing `bitrise.yml` to make CI job easily debuggable.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/tools/nightly-unit-tests/generate_bitrise_yml.py
+++ b/tools/nightly-unit-tests/generate_bitrise_yml.py
@@ -118,11 +118,6 @@ def generate_bitrise_yml(test_plan: TestPlan, dry_run: bool):
 
     print('\n⚙️ Saving `bitrise.yml`...')
     bitrise_yml.write(path='bitrise.yml')
-
-    print('\n⚙️ `bitrise.yml` content:')
-    saved_bitrise_yml = open('bitrise.yml', 'r')
-    print(saved_bitrise_yml.read())
-
     print('\n⚙️ All good 👍')
 
 
--- a/tools/nightly-unit-tests/generate_bitrise_yml.py
+++ b/tools/nightly-unit-tests/generate_bitrise_yml.py
@@ -118,6 +118,11 @@ def generate_bitrise_yml(test_plan: TestPlan, dry_run: bool):
 
     print('\n⚙️ Saving `bitrise.yml`...')
     bitrise_yml.write(path='bitrise.yml')
+
+    print('\n⚙️ `bitrise.yml` content:')
+    saved_bitrise_yml = open('bitrise.yml', 'r')
+    print(saved_bitrise_yml.read())
+
     print('\n⚙️ All good 👍')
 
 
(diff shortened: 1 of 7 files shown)


ganeshnj:
bitrise night test run: pending

tools/nightly-unit-tests/generate_bitrise_yml.py
```diff
@@ -118,6 +118,11 @@ def generate_bitrise_yml(test_plan: TestPlan, dry_run: bool):
 
     print('\n⚙️ Saving `bitrise.yml`...')
     bitrise_yml.write(path='bitrise.yml')
+
+    print('\n⚙️ `bitrise.yml` content:')
+    saved_bitrise_yml = open('bitrise.yml', 'r')
+    print(saved_bitrise_yml.read())
+
```
ncreated, line 125: q/ Do we need it? AFAIR generated `yml` is deployed as build artifact and available after job completion.

ncreated (approved):
Looks good 👌. Where do the code changes come from? Misalignment in Swift versions 💭?

tools/nightly-unit-tests/generate_bitrise_yml.py
ganeshnj: ah, you are right, it is there. I didn't know that, Just removed this part.

maxep (approved):
🙏

ganeshnj:
> Looks good 👌. Where do the code changes come from? Misalignment in Swift versions 💭?

I think so, devs in the team use different version of the Xcodes and sometimes beta also. We implicitly use the new langauge features which are not supported by older compiler versions sometimes.

I don't think forcing everyone to use same Xcode version is good idea either because
- we get more people testing in different version including beta
- our customers could use any version as well

A better solution to catch such issues to build the SDK for supported Swift versions (or atleast one lowest and one highest) `in parallel`.

PR #1383, opened 2023-07-20
RUMM-3459 fix: Fix linking telemetry events to RUM session

ncreated:
### What and why?

🐞 This PR fixes broken link between RUM Telemetry events and RUM session.

Without this link we can't make data-driven decisions that include `session_id` querying or distributions.

### How?

Added integration tests for Telemetry feature, which revealed several issues. Fixed one by one.

All problems follow the same pattern of invalid assumption on RUM context retrieval from `FeatureBaggage`. The only non-optional attribute in RUM context is `application.id`, all other (`session.id`, `view.id` and `action.id`) are optional, so:
```swift
// Wrong:
let attributes: [String: String]? = context.featuresAttributes["rum"]?.ids

// Correct:
let attributes: [String: String?]? = context.featuresAttributes["rum"]?.ids
```

In result of wrong assumption, RUM context information was only attached to these telemetry logs that were recorded during full RUM context - meaning all all IDs set, so only when user action on some view was pending.

This problem wasn't surfaced by [existing unit tests](https://github.com/DataDog/dd-sdk-ios/blob/4753d05a6432237d8aa6962307d7de2699794499/DatadogRUM/Tests/Integrations/TelemetryReceiverTests.swift#L87-L94), because they mock the entire context, covering only the "full context" scenario.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/Datadog/IntegrationUnitTests/Public/TelemetryCoreIntegrationTests.swift
+++ b/Datadog/IntegrationUnitTests/Public/TelemetryCoreIntegrationTests.swift
@@ -1,0 +1,137 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import XCTest
+import DatadogCore
+import DatadogInternal
+@testable import DatadogRUM
+import TestUtilities
+
+class TelemetryCoreIntegrationTests: XCTestCase {
+    private var core: DatadogCoreProxy! // swiftlint:disable:this implicitly_unwrapped_optional
+    private var telemetry: TelemetryCore! // swiftlint:disable:this implicitly_unwrapped_optional
+
+    override func setUp() {
+        core = DatadogCoreProxy()
+        telemetry = TelemetryCore(core: core)
+    }
+
+    override func tearDown() {
+        core.flushAndTearDown()
+        core = nil
+        telemetry = nil
+    }
+
+    func testGivenRUMEnabled_telemetryEventsAreSent() {
+        // Given
+        var config = RUM.Configuration(applicationID: .mockAny())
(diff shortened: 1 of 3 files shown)


Datadog/IntegrationUnitTests/Public/TelemetryCoreIntegrationTests.swift
ncreated, line 1: Note: I'm adding `IntegrationUintTests/Public` folder here, so we can start differentiating what is an integration unit vs regular unit test. Should simplify further movement of these files to separate project / target.

DatadogRUM/Sources/Integrations/TelemetryReceiver.swift
```diff
@@ -203,7 +204,8 @@ internal final class TelemetryReceiver: FeatureMessageReceiver {
 
         rum.eventWriteContext { context, writer in
             // reset recorded events on session renewal
-            let attributes: [String: String]? = context.featuresAttributes["rum"]?.ids
+            let rumAttributes: [String: String?]? = context.featuresAttributes[RUMFeature.name]?.ids
+            let attributes = rumAttributes?.compactMapValues { $0 }
             let sessionId = attributes?[RUMContextAttributes.IDs.sessionID]
 
             if sessionId != self.currentSessionID {
```
ncreated, line 211: I'm not fully sure on what was the impact here. If no `action.id` was present in RUM context, then `sessionId` was always `nil`. Looking at further logic, I think we might have been losing significant volume of telemetries because `false` was dominant in `sessionId != self.currentSessionID` comparison (`nil != nil`). That led to never reseting the `eventIDs` buffer, hence always being capped by `TelemetryReceiver.maxEventsPerSessions`.

xgouchet (approved):

maxep (approved):
Looks great, thanks for the integration tests! Maybe we should include this in beta 2, WDYT?

PR #1384, opened 2023-07-20
RUMM-3459 feat: Add "Batch Deleted" telemetry

ncreated:
### What and why?

📦⏱️ Adding "Batch Deleted" metric.

See (internal) [RFC Batching & Upload Observability](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3076063377/RFC+-+Batching+Upload+Observability).

### How?

I added new method on `Telemetry` interface:
```swift
/// Collect metric value.
///
/// Metrics are reported as debug telemetry. Unlike regular events, they are not subject to duplicates filtering and
/// are get sampled with a different rate. Metric attributes are used to create facets for later querying and graphing.
///
/// - Parameters:
///   - name: The name of this metric.
///   - attributes: Parameters associated with this metric.
public func metric(name: String, attributes: [String: Encodable])
```

Following the RFC:
* Metrics are sent as DEBUG telemetry events with attributes.
* We sample metric events with additional 15% (on top of default 20%).
* There is no ID-based filtering for metric events.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


maxep (approved):
LGTM 👍

DatadogCore/Sources/PerformancePreset.swift
```diff
@@ -34,6 +34,12 @@ internal protocol StoragePerformancePreset {
     var maxObjectSize: UInt64 { get }
 }
 
+internal extension StoragePerformancePreset {
+    /// The default duration since last write after which the uploader will consider the file to be  “ready for upload”.
+    /// Added for computing batching & upload metrics in RUMM-3459.
+    var uploaderWindow: TimeInterval { (minFileAgeForRead + maxFileAgeForWrite) * 0.5 }
```
ganeshnj, line 40: question

what is rational behind half the read + write? a more elaborating comment would help others understanding better.

DatadogRUM/Sources/Integrations/TelemetryReceiver.swift
```diff
@@ -211,12 +258,19 @@ internal final class TelemetryReceiver: FeatureMessageReceiver {
...
+            // record up to `maxEventsPerSessions`, discard duplicates for events with `id`
+            if self.eventsCount < TelemetryReceiver.maxEventsPerSessions {
+                if id == nil {
+                    self.eventsCount += 1
+                    operation(context, writer)
+                } else if let eventID = id, !self.eventIDs.contains(eventID) {
+                    self.eventIDs.insert(eventID)
+                    self.eventsCount += 1
+                    operation(context, writer)
+                }
```
ganeshnj, line 273:
```suggestion
                if let eventID = id, !self.eventIDs.contains(eventID) {
                    self.eventIDs.insert(eventID)
                }
                self.eventsCount += 1
                operation(context, writer)
```

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
```diff
@@ -125,7 +125,7 @@ internal class DataUploadWorker: DataUploadWorkerType {
         queue.sync {
             while let nextBatch = self.fileReader.readNextBatch() {
                 _ = try? self.dataUploader.upload(events: nextBatch.events, context: contextProvider.read())
-                self.fileReader.markBatchAsRead(nextBatch)
+                self.fileReader.markBatchAsRead(nextBatch, reason: .flushed)
```
ganeshnj, line 128: blocking

we must not ignore the error with optional retry as this will hide the underlying error and we will always good flushed reason.

wrapping in a try-catch would be better option just like we are doing in `init`

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
```diff
@@ -93,7 +93,7 @@ internal class DataUploadWorker: DataUploadWorkerType {
                     }
                 } catch let error {
                     // If upload can't be initiated do not retry, so drop the batch:
-                    self.fileReader.markBatchAsRead(batch)
+                    self.fileReader.markBatchAsRead(batch, reason: .invalid)
```
ganeshnj, line 96: question

is it worth knowing the error in case of an invalid removal reason?

ganeshnj (changes requested):
Looks good overall, nicely put together. One blocking feedback and some questions.

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
ncreated: Totally, and we know it from the next line that sends error telemetry:
```swift
DD.telemetry.error("Failed to initiate '\(self.featureName)' data upload", error: error)
```
I don't think makes sense to send this error as part of the metric, because the cardinality could be very high. More practical scenario that I see is: notice huge volume of "deleted because of `.invalid`" metric and then find the actual issue by analysing "Failed to initiate with `error`" logs. WDYT?

DatadogRUM/Sources/Integrations/TelemetryReceiver.swift
ncreated: It isn't the same. Note that the original code allows one empty branch: when `id != nil`, but it is already stored in `self.eventIDs` it should do nothing (no increment and no `operation()`).

The suggested code doesn't pass `testSendTelemetry_discardDuplicates()` test, which was a good exercise 🙂👍.

DatadogCore/Sources/Utils/CoreMetrics.swift
```diff
@@ -0,0 +1,94 @@
...
+        /// The batch was deleted due to exceeding allowed max size for batches directory.
+        case purged
+        /// The feature failed to create request for that batch (e.g. data was malformed).
+        case invalid
+        /// The batch was deleted arbitrarily without considering its delivery status. This option is only used in test logic
+        /// and we don't send "Batch Deleted" metric for this case.
+        case flushed
+
+        /// Converts the removal reason to the string value expected for `batchRemovalReasonKey`.
+        var asString: String {
```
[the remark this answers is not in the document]
ncreated: Sounds better 👌, done ✅.

DatadogCore/Sources/PerformancePreset.swift
ncreated: Indeed, I added more context. Now the comment should explain it well:
```swift
/// The uploader window duration determines when a file is considered "ready for upload" by the uploader after the last write.
///
/// This value is crucial for computing batching and upload metrics within the SDK (see RUMM-3459). The uploader window is derived from the
/// original `batchSize` value, which is either configured by the user or set as an internal override. The `batchSize` represents the age of
/// "batch maturity" for uploads and is specified in seconds.
///
/// The uploader window is calculated as the average of two key parameters: `minFileAgeForRead` and `maxFileAgeForWrite`. Batches younger
/// than `maxFileAgeForWrite` are considered "writable" (available for the writer), while batches older than `minFileAgeForRead` are
/// meant to be "readable" (available for the uploader). To ensure that the writer and uploader don't access the same batch simultaneously,
/// a safe-guard window (10% of `batchSize`) is implemented within which the batch is neither writable nor readable.
var uploaderWindow: TimeInterval { (minFileAgeForRead + maxFileAgeForWrite) * 0.5 }
```

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
ncreated: I added `try-catch` 👍 because it makes sense indeed. Now it tries once, and does one additional retry on failure.

However, this doesn't change the fact that we don't want to see `.flushed` batches in metrics - no matter if flushed successfully or not (this is why we don't attach the error information if "flushed" upload failed). This code is ran only in integration tests, cannot be reached from user-facing API and we don't want to skew our metric by this utility.

An alternative that I had consider is to make `reason` optional and pass `nil` in this flow. That could make it better reflect the requirement. However, dealing with optionality in downstream logic didn't seem safe-enough to me. It would invite future maintainers to basically skip the `reason` with passing `nil`, without considering what this option is used for and possibly skewing the metric. Defining certain set of "skipped" values looks IMO safer.

WDYT @ganeshnj ?

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
ganeshnj: right, we should be able to query 

```
.invalid && message.startsWith("Failed to instantiate" 
```

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
ganeshnj: Interesting, I didn't know this is test only code. I think I can be persuaded however we have another problem that we are writing code for testing which is shipped with the SDK instead we should design our APIs which are testable without additional code. 

Not related to this PR anyway, I'm good 👌

DatadogRUM/Sources/Integrations/TelemetryReceiver.swift
ganeshnj: 💯

ganeshnj (approved):

PR #1386, opened 2023-07-21
RUMM-3460 feat: Add "Batch Closed" telemetry

ncreated:
### What and why?

📦⏱️ Adding "Batch Closed" metric.

See (internal) [RFC Batching & Upload Observability](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3076063377/RFC+-+Batching+Upload+Observability).

### How?

Leveraged metrics API added earlier in #1384.

With this PR, I'm adding full unit tests coverage for both metrics creation in `FilesOrchestrator`.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/DatadogCore/Sources/Utils/CoreMetrics.swift
+++ b/DatadogCore/Sources/Utils/CoreMetrics.swift
@@ -6,17 +6,10 @@
 
 import Foundation
 
-/// Definition of "Batch Deleted" telemetry.
-internal enum BatchDeletedMetric {
-    /// The name of this metric, included in telemetry log.
-    /// Note: the "[Mobile Metric]" prefix is added when sending this telemetry in RUM.
-    static let name = "Batch Deleted"
-
+/// Common definitions for batch telemetries.
+internal enum BatchMetric {
     /// Metric type key.
     static let typeKey = "metric_type"
-    /// Metric type value.
-    static let typeValue = "batch deleted"
-
     /// Track name key
     static let trackKey = "track"
     /// Track name value.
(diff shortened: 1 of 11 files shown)


DatadogCore/Sources/Utils/CoreMetrics.swift
```diff
@@ -92,3 +93,23 @@ internal enum BatchDeletedMetric {
         }
     }
 }
+
+/// Definition of "Batch Closed" telemetry.
+internal enum BatchClosedMetric {
```
ganeshnj, line 98: non blocking, but would have preferred this as struct with method `toAttributes() -> expected type` which makes it easier handle any conversion.

ganeshnj (approved):
💯 test cases, nice work.

maxep (approved):
👌

DatadogCore/Sources/Utils/CoreMetrics.swift
ncreated: I have been thinking of it, but I decided to keep it bare simple, as we introduce a new concept here and not yet sure which direction it will go. Moving from "static constants" model to a "type" won't cost much in the future.

For instance, one idea I had was to define `Metric` protocol that enforces common fields: `name`, `metric_type` and `attributes`. That could be passed directly to `telemetry.metric(_:)` in a type-safer way, instead of broad `telemetry.metric(name:attributes:)`.

With few options in hand, I bet on simplicity for now, so will keep it as it is.

PR #1393, opened 2023-07-26
REPLAY-1869 feat: Configure Session Replay dependency

ncreated:
### What and why?

📦 This PR brings complete definition for `DatadogSessionReplay` dependency.

It will be now available through Carthage, Cocoapods, SPM and in XCFramework.

### How?

Except updating dependency definition, I made necessary changes to ensure that smoke tests now include SR ✅ and verified that podspec and XCFramework validate properly in distribution scripts ✅.

**Changes to package distribution:**
- Added `DatadogSessionReplay iOS.xcscheme` together with defining `DatadogSessionReplay` target in the main `.pbxproj`. This brings Carthage compatibility and enables SR `.xcframework` build.
- Updated `tools/distribution/` to include `DatadogSessionReplay.podspec` and `DatadogSessionReplay.xcframework`.

**Other changes:**
- Updated smoke tests to cover Session Replay APIs for Carthage and XCFramework scenarios.
- Added first minimal coverage for Session Replay in `DatadogCoreTests`, to make sure it links properly there.
- Deleted `DatadogSessionReplay/Package.swift`, so we have only one place for editing SR code.
- Updated `SRSnapshotTests` project to use the root `Package.swift` - this required introducing `DD_TEST_UTILITIES_ENABLED` flag for exporting utility libraries in conditional builds (the entire concept is documented in code and `SRSnapshotTests/README.md`).

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


DatadogSessionReplay/Tests/Mocks/UIKitMocks.swift
```diff
@@ -105,3 +105,34 @@ extension UITextContentType: RandomMockable {
...
 }
+
+extension UIImage: RandomMockable {
+    /// Creates bitmap by randomising the value of each pixel.
+    public static func mockRandom() -> Self {
+        return mockRandom(width: .mockRandom(min: 10, max: 100), height: .mockRandom(min: 10, max: 100))
+    }
+
+    /// Creates bitmap of certain size by randomising the value of each pixel.
+    public static func mockRandom(width: Int, height: Int) -> Self {
```
ncreated, line 116: 💡 Previously, some unit tests in SR used an actual [`.png` image](https://github.com/DataDog/dd-sdk-ios/tree/2.0.0-beta1/DatadogSessionReplay/Tests/Recorder/Utilties/Media.xcassets/dd_logo_v_rgb.imageset). To avoid re-adding `Media.xcassets` to the main `Datadog.xcworkspace`, I switched to generating random bitmap at runtime.

buraizu:
Thanks, created DOCS-5955 for docs team review

Makefile
```diff
@@ -78,6 +78,12 @@ prepare-integration-tests:
 		@cd IntegrationTests/ && pod install
 		@echo "OK 👌"
 
+open-sr-snapshot-tests:
+		@echo "⚙️  Opening SRSnapshotTests with DD_TEST_UTILITIES_ENABLED ..."
+		@pgrep -q Xcode && killall Xcode && echo "- Xcode killed" || echo "- Xcode not running"
+		@sleep 0.5 && echo "- launching" # Sleep, otherwise, if Xcode was running it often fails with "procNotFound: no eligible process with specified descriptor"
+		@open --env DD_TEST_UTILITIES_ENABLED ./DatadogSessionReplay/SRSnapshotTests/SRSnapshotTests.xcworkspace
+
```
maxep, line 86: Well, I'm still not convinced by how practical this is, or will be for other UI Tests.

Killing all my instances of Xcode for running an inner project is not a good experience IMO. I found CP setup more straightforward, we could run all `pod install` in the `make dependencies` and we would be set!

maxep (approved):
Looks great!

I left a comment but it's not blocking the PR.

Makefile
ncreated: I see the point 👍, although I'm more on the side of this simple setup rather than the impact of adding CP to a project. The effect of "killing Xcodes" is mitigated by the fact that we can't have multiple projects open anyway if they link the same local package as the workspace dependency.

This is rather a topic for a wider debate on how we want to manage dependencies in the repo. I can take a lead on that and gather all options in an RFC. With V2 in place, we see better what are the needs vs limitations 🙂.

alai97 (approved):
Looks good!

Issue #1288, opened 2023-05-24
Network traffic (= traces) not captured for `URLSession.data(...)` methods


gdeff:
Hello,

In our iOS project we have set the Datadog SDK and activated trace collection.

The bug is that our requests/responses are not captured by Datadog. No trace is sent to Datadog server so we cannot find them into the Datadog UI.

After a little of investigation in the Datadog iOS SDK code, we noticed that there is a swizzling mechanism on `URLSession.dataTask()` methods. This mechanism is responsible for capturing traces.
In our project, we do not use `URLSession.dataTask()` methods but `URLSession.data()` methods which are their "async/await" version and available since iOS 15. More precisely, we observe that the following methods are not supported by this swizzling mechanism:

- `URLSession.data(for: URLRequest) -> (Data, URLResponse)`
- `URLSession.data(from: URL) -> (Data, URLResponse)`
- `URLSession.data(for: URLRequest, delegate: (URLSessionTaskDelegate)?) -> (Data, URLResponse)`
- `URLSession.data(from: URL, delegate: (URLSessionTaskDelegate)?) -> (Data, URLResponse)`

Could you, please, update the code in order that the Datadog iOS SDK support these methods ?

ganeshnj:
Since iOS 16, there have been some changes for async APIs in `URLSession` where not all delegates methods are called. This causes in failure in detection of request start and stop.

Only `NSURLSessionTask.resume` and `URLSessionTaskDelegate.urlSession(_:task:didFinishCollecting:)` are called, hence we need to modify the way we collect the metrics for async APIs (atleast) for iOS 16 and above.

I quickly checked the hypothesis on https://github.com/DataDog/dd-sdk-ios/compare/develop...ganeshnj/fix/1288-url-session-instrumentation branch and I was able to collect metrics for async APIs.

There was a similar issue in https://github.com/open-telemetry/opentelemetry-swift/pull/398 as well which was recently fixed.

ncreated:
Instrumentation for async/await APIs was introduced in [**2.5.0**](https://github.com/DataDog/dd-sdk-ios/releases/tag/2.5.0) - I'm closing this issue.

PR #1394, opened 2023-07-26
RUM-717 feat!: Add network instrumentation for async/await URLSession APIs

ganeshnj:
### What and why?

Fixes https://github.com/DataDog/dd-sdk-ios/issues/1288

### How?

https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3130426092

TLDR

Dogfood: https://github.com/DataDog/shopist-ios/pull/55

Customers no longer have to implement Datadog delegate to instrument rather they can simply use high level APIs.

1. Enable URLSessionTracking either via RUM or Trace

```swift
RUM.enable()
```
or 
```swift
Trace.enable()
```

2. Enable session tracking per delegate

```swift
let delegate = MyDelegate() // implements URLSessionTaskDelegate
URLSessionInstrumentation.enable(with: .init(delegate: delegate))
```

- instrumentation now works for async-await APIs too.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


ncreated (commented):
Only preliminary feedbacks for now, I need more time 🙏.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionSwizzler.swift
```diff
@@ -6,77 +6,70 @@
 
 import Foundation
 
+/// Swizzles `URLSession` methods.
 internal class URLSessionSwizzler {
```
maxep, line 10: /question Now that we have `URLSessionTaskSwizzler`, why do we need `URLSessionSwizzler`?
It looks like it's only there to call `interceptionWillStart` which not used by any feature.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionDataDelegateSwizzler.swift
```diff
@@ -0,0 +1,100 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+/// Swizzles `URLSessionDataDelegate` callbacks.
+internal class URLSessionDataDelegateSwizzler {
+    static var didReceiveMap: [String: DidReceive?] = [:]
+    static var lock = NSRecursiveLock()
```
maxep, line 12: /suggestion We have repetitive use of recursive lock now, so we could create a `@propertyWrapper` in a similar fashion as `ReadWriteLock`. It could be a `Mutex` using `pthread_mutex_t` with optional parameters such as `PTHREAD_MUTEX_RECURSIVE`.
Motivation being that [`NSLock` are slow compared to mutex](https://www.vadimbulavin.com/benchmarking-locking-apis/)

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
```diff
@@ -78,17 +164,23 @@ extension NetworkInstrumentationFeature {
...
     ///   - additionalFirstPartyHosts: Extra hosts to consider in the interception.
     func intercept(task: URLSessionTask, additionalFirstPartyHosts: FirstPartyHosts?) {
-        guard let request = task.originalRequest else {
+        guard let originalRequest = task.originalRequest else {
             return
         }
 
-        queue.async {
+        // async update to task leads to race condition where the currentRequest could already be sent to the transport
+        queue.sync {
```
maxep, line 172: /question We could get rid of the `queue` and use locks only, no?

maxep (changes requested):
Nice work here!

In additions to @ncreated comments, I've left some suggestions. One important topic is delegate instance vs. type when enabling the instrumentation, I think we should be clear that we are instrumenting the class and not the instance.

DatadogInternal/Sources/NetworkInstrumentation/DatadogURLSessionHandler.swift
```diff
@@ -19,6 +19,14 @@ public protocol DatadogURLSessionHandler {
     /// - Returns: The modified request.
     func modify(request: URLRequest, headerTypes: Set<TracingHeaderType>) -> URLRequest
 
+    /// Tells the interceptor that the session will start.
+    ///
+    /// Note: This method is only called for URLSessionTask based APIs.
+    /// e.g. `URLSession.dataTask(with:completionHandler:)`
```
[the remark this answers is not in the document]
ganeshnj: agree this is not perfect interface but we are constrained with URLSession and Swift.

I would have preferred it to be an internal API and not exposed outside. However, Swift limitation doesn't allow.

One idea I have is to break the interface in two
```swift
public protocol DatadogURLSessionHandler {
    var firstPartyHosts: FirstPartyHosts { get }
    func modify(request: URLRequest, headerTypes: Set<TracingHeaderType>) -> URLRequest
    func interceptionDidStart(interception: URLSessionTaskInterception)
    func interceptionDidComplete(interception: URLSessionTaskInterception)
}
```
and 
```swift
public protocol __DatadogURLSessionHandler: DatadogURLSessionHandler {
    func interceptionWillStart(interception: URLSessionTaskInterception)
}
```

This way a public vs internal APIs can be kept separately although not much of gain from usage point of view.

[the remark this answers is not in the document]
ganeshnj: WDYT?

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
```diff
@@ -41,18 +41,104 @@ internal final class NetworkInstrumentationFeature: DatadogFeature {
     /// Maps `URLSessionTask` to its `TaskInterception` object.
     ///
     /// The interceptions **must** be accessed using the `queue`.
+    @ReadWriteLock
     private var interceptions: [URLSessionTask: URLSessionTaskInterception] = [:]
```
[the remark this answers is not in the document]
ganeshnj: you are right, it is indeed redundant. Removed.

IntegrationTests/IntegrationScenarios/Scenarios/RUM/RUMResourcesScenarioTests.swift
```diff
@@ -28,7 +28,7 @@ class RUMResourcesScenarioTests: IntegrationTests, RUMCommonAsserts {
             ),
             urlSessionSetup: .init(
                 instrumentationMethod: .allCases.randomElement()!,
-                initializationMethod: .allCases.randomElement()!
+                initializationMethod: .afterSDK
```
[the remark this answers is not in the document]
ganeshnj: glad you asked, I will clean this logic as a separate PR. Don't want to mix it here.

But the rationale here is, we can't enable the URLSession instrumentation without enabling either RUM or Trace.

This will be included in the public docs too. First either enable RUM or Trace and then enable `URLSesssion` instrumentation.

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
```diff
@@ -130,7 +222,12 @@ extension NetworkInstrumentationFeature {
     ///   - task: The task that provided data.
     ///   - data: A data object containing the transferred data.
     func task(_ task: URLSessionTask, didReceive data: Data) {
-        queue.async { self.interceptions[task]?.register(nextData: data) }
+        queue.async {
+            guard let interception = self.interceptions[task] else {
+                return
+            }
+            interception.register(nextData: data)
+        }
```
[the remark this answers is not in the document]
ganeshnj: just wanted it for a breakpoint. Sometimes shorthands makes debugging harder and we know compiler will optimize it anyway.

Code for readability.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionDataDelegateSwizzler.swift
ganeshnj: Let me check on this.

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
ganeshnj: lock has been removed.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionDataDelegateSwizzler.swift
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        try bind(delegate: delegate, interceptDidReceive: interceptDidReceive)
+    }
+
+    static func bind(delegate: URLSessionDataDelegate, interceptDidReceive: @escaping (URLSession, URLSessionDataTask, Data) -> Void) throws {
+        lock.lock()
+        defer { lock.unlock() }
+
+        let didReceive = try DidReceive.build(klass: type(of: delegate))
```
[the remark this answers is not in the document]
ganeshnj: Let's discuss on this, during the RFC review we decided to have the instance because of strongly typeness we get but I'm happy to hear what do you think or you know some alternative where customers can have benefit of both worlds.

IntegrationTests/IntegrationScenarios/Scenarios/RUM/RUMResourcesScenarioTests.swift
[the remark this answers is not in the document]
ncreated: Makes sense 👍. It stands for a clear difference compared to previous API, where `DDURLSessionDelegate()` could have been (legally) constructed before SDK init. Now we don't need `.beforeSDK` nor the entire `InitializationMethod` 🧽👌.

DatadogInternal/Sources/NetworkInstrumentation/FirstPartyHosts.swift
```diff
@@ -37,6 +37,18 @@ public struct FirstPartyHosts: Equatable {
         self.init(hostsWithTracingHeaderTypes: [:])
     }
 
+    internal init?(firstPartyHosts: URLSessionInstrumentation.FirstPartyHostsTracing?) {
+        guard let firstPartyHosts = firstPartyHosts else {
+            return nil
+        }
```
maciejburda, line 43:
```
default:
    return nil
  ```
  
  is slightly less code :)

File: DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift [hunk 10]
```diff
@@ -78,17 +163,23 @@ extension NetworkInstrumentationFeature {
     ///   - task: The created task.
     ///   - additionalFirstPartyHosts: Extra hosts to consider in the interception.
     func intercept(task: URLSessionTask, additionalFirstPartyHosts: FirstPartyHosts?) {
-        guard let request = task.originalRequest else {
+        guard let originalRequest = task.originalRequest else {
             return
         }
 
-        queue.async {
+        // async update to task leads to race condition where the currentRequest could already be sent to the transport
```
maciejburda, line 170: I would rephrase to describe what we actually have in the code. Right now it looks like a typo on a first glance.

Something like: `// sync update to task prevents a race condition where the currentRequest could already be sent to the transport`

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTask+Tracking.swift
```diff
@@ -0,0 +1,38 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+private var sessionFirstPartyHostsKey: UInt8 = 1
```
maciejburda, line 9: I'm wondering if we should seek for some unique numbers for all those associated objects keys.

I imagine that some apps might use exactly the same key which might lead to unexpected behaviour

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskInterception.swift
```diff
@@ -50,6 +50,10 @@ public class URLSessionTaskInterception {
         }
     }
 
+    func register(request: URLRequest) {
+        self.request = request
+    }
```
maciejburda, line 55: Maybe it's ok to just have var request instead.

maciejburda (commented):
Overall looks good! +1 to other's comments + a few minor ones from my side

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskInterception.swift
ganeshnj: going with the established pattern like we have `metrics`, `data` etc properties here.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionDataDelegateSwizzler.swift
ganeshnj: It sounded like premature optimization to me but I still updated the code with `pthread_mutex_t`

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
ganeshnj: that's fair - updated.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTask+Tracking.swift
ganeshnj: there is a remote chance of colliding, so I changed the numbers to other than 1,2,3 etc.

DatadogCore/Tests/TestsObserver/DatadogTestsObserver.swift
```diff
@@ -42,15 +42,6 @@ internal class DatadogTestsObserver: NSObject, XCTestObservation {
...
         ),
-        .init(
-            assert: { URLSessionSwizzler.bindingsCount == 0 },
-            problem: "No `URLSessionSwizzler` must be bonded.",
-            solution: """
-            Make sure all applied `URLSessionSwizzler.bind()` are reset by the end of test with `URLSessionSwizzler.unbind()`.
-
-            `DatadogTestsObserver` found \(URLSessionSwizzler.bindingsCount) bindings left.
-            """
-        ),
```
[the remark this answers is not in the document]
ganeshnj: makes sense. updated.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionDataDelegateSwizzler.swift
ganeshnj: Actually reverted back to original - xcode complains in ThreadSanitizer and also CI run failed

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionDataDelegateSwizzler.swift
```diff
@@ -0,0 +1,100 @@
...
+
+        private init(selector: Selector, klass: AnyClass) throws {
+            do {
+                method = try Self.findMethod(with: selector, in: klass)
+            } catch {
+                // URLSessionDataDelegate doesn't implement the selector, so we inject it and swizzle it
+                let block: @convention(block) (URLSessionDataDelegate, URLSession, URLSessionDataTask, Data) -> Void = { delegate, session, task, data in
+                }
+                let imp = imp_implementationWithBlock(block)
+                class_addMethod(klass, selector, imp, "@@@")
```
ncreated, line 83: question/ What is `"@@@"`? 🤔💭 If it's the method name, I feel it's better to include "datadog" in the name - that might help to debug and understand their stack traces.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTask+Tracking.swift
```diff
@@ -0,0 +1,28 @@
...
+        get {
+            return sessionFirstPartyHosts ?? .init()
+        }
+    }
+
+    /// Extension property for storing first party hosts passed from `URLSession` to `URLSessionTask`.
+    /// This is used for `URLSessionTask` based APIs.
+    var sessionFirstPartyHosts: FirstPartyHosts? {
+        set {
+            objc_setAssociatedObject(self, &sessionFirstPartyHostsKey, newValue, objc_AssociationPolicy.OBJC_ASSOCIATION_RETAIN_NONATOMIC)
```
ncreated, line 22: question/ Is non-atomic on purpose (vs atomic [`OBJC_ASSOCIATION_RETAIN`](https://developer.apple.com/documentation/objectivec/objc_associationpolicy/objc_association_retain))?

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskDelegate+Tracking.swift
```diff
@@ -0,0 +1,28 @@
...
+import Foundation
+
+private var firstPartyHostsKey: UInt8 = 41
+private var coreKey: UInt8 = 43
+
+public extension URLSessionDelegate {
+     /// Returns the `DatadogCore` for this delegate.
+    var core: DatadogCoreProtocol? {
+        set {
+            objc_setAssociatedObject(self, &coreKey, newValue, objc_AssociationPolicy.OBJC_ASSOCIATION_RETAIN_NONATOMIC)
```
ncreated, line 16: question/ Does it create strong reference to `core` in `URLSessionDelegate`? That would mean `core` is no longer managed (solely) by `CoreRegistry`, hence we won't be able to deallocate it with `stop()` / `deinitialize()` APIs.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskDelegateSwizzler.swift
```diff
@@ -0,0 +1,115 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+internal func key(from anyClass: AnyClass) -> String {
+    String(reflecting: type(of: anyClass))
+}
```
ncreated, line 11: suggestion/ Too generic for `internal` namespace. Think of nesting or specialising.

DatadogInternal/Tests/NetworkInstrumentation/URLSessionDataDelegateSwizzlerTests.swift
ncreated, line 1: Nice 👌

IntegrationTests/Runner/Scenarios/URLSession/URLSessionScenarios.swift
```diff
@@ -142,11 +142,12 @@ class URLSessionBaseScenario: NSObject {
     }
 
     private func createInstrumentedURLSession() -> URLSession {
-        let delegate: URLSessionDelegate
+        let delegate: URLSessionDataDelegate
 
         switch setup.instrumentationMethod {
         case .directWithGlobalFirstPartyHosts:
             delegate = DDURLSessionDelegate()
+            URLSessionInstrumentation.enable(with: .init(delegateClass: DDURLSessionDelegate.self))
```
ncreated, line 150: question/ There should be no difference if using (our) `DDURLSessionDelegate` vs (their) `MyCustomDelegate`, right? If the plan is to remove `DDURLSessionDelegate`, we could start using custom delegate in integration tests right away.

ncreated (commented):
Looks neat 👌. I left few questions. On top of it:
* Shouldn't we mark `DDURLSessionDelegate` as deprecated?
* What about exposing this new API in `Objective-C`?

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
```diff
@@ -43,16 +43,103 @@ internal final class NetworkInstrumentationFeature: DatadogFeature {
...
     private var interceptions: [URLSessionTask: URLSessionTaskInterception] = [:]
 
-    init() throws {
-        try URLSessionSwizzler.bind()
+    /// Swizzles `URLSessionTaskDelegate`, `URLSessionDataDelegate`, `URLSessionTask` and `URLSession` methods
+    /// to intercept `URLSessionTask` lifecycles.
+    ///
+    /// - Parameter configuration: The configuration to use for swizzling.
+    /// Note: We are only concerned with type of the delegate here but to provide compile time safety, we
+    ///      use the instance of the delegate to get the type.
```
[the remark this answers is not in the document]
ganeshnj: yup fixed.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionDataDelegateSwizzler.swift
ganeshnj: just updated it, i hope it will make it easier to understand

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTask+Tracking.swift
ganeshnj: Initially, I didn't see a reason for it to be atomic but now for both `URLSessionDelegate.core` and `URLSessionTask.sessionFirstPartyHosts`, it makes sense to be atomic because we are not sure how these properties are called and on which thread depending on the delegate method execution and URLSession task method execution.

Good call.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskDelegateSwizzler.swift
ganeshnj: fair call, moved.

IntegrationTests/Runner/Scenarios/URLSession/URLSessionScenarios.swift
ganeshnj: these test for compatibility reasons, we want existing customers to not break.

Which means if customers are using `DDURLSessionDelegate`, we will auto enable urlsession instrumentation.

ganeshnj: however, we need to find deprecation process.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionInstrumentation.swift
```diff
@@ -0,0 +1,95 @@
...
+
+        /// Configuration of URLSession instrumentation.
+        /// - Parameters:
+        ///   - delegate: The delegate class to be used to swizzle URLSessionTaskDelegate & URLSessionDataDelegate methods.
+        ///   - firstPartyHostsTracing: Additional first party hosts to consider in the interception.
+        public init(delegateClass: AnyClass, firstPartyHostsTracing: FirstPartyHostsTracing? = nil) {
+            self.delegateClass = delegateClass
+            self.firstPartyHostsTracing = firstPartyHostsTracing
+        }
+    }
```
maxep, line 80: I think we can constrain the type of the delegate class with:
```swift
/// Configuration of URLSession instrumentation.
public struct Configuration {
    /// The delegate class to be used to swizzle URLSessionTaskDelegate & URLSessionDataDelegate methods.
    public var delegateClass: URLSessionDataDelegate.Type

    /// Additional first party hosts to consider in the interception.
    public var firstPartyHostsTracing: FirstPartyHostsTracing?

    /// Configuration of URLSession instrumentation.
    /// - Parameters:
    ///   - delegate: The delegate class to be used to swizzle URLSessionTaskDelegate & URLSessionDataDelegate methods.
    ///   - firstPartyHostsTracing: Additional first party hosts to consider in the interception.
    public init(delegateClass: URLSessionDataDelegate.Type, firstPartyHostsTracing: FirstPartyHostsTracing? = nil) {
        self.delegateClass = delegateClass
        self.firstPartyHostsTracing = firstPartyHostsTracing
    }
}
```

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
```diff
@@ -109,7 +202,7 @@ extension NetworkInstrumentationFeature {
     ///   - task: The task whose metrics have been collected.
     ///   - metrics: The collected metrics.
     func task(_ task: URLSessionTask, didFinishCollecting metrics: URLSessionTaskMetrics) {
-        queue.async {
+        queue.sync {
```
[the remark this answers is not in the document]
ganeshnj: testing around fix of CI failures - I will update it.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionInstrumentation.swift
ganeshnj: pretty neat - great find.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskDelegate+Tracking.swift
ganeshnj: `OBJC_ASSOCIATION_ASSIGN` should be weak reference and must not increment the retain 
I updated it.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/DatadogURLSessionDelegate.swift
```diff
@@ -12,12 +12,6 @@ public typealias DDURLSessionDelegate = DatadogURLSessionDelegate
 /// The implementation must ensure that required methods are called on the `ddURLSessionDelegate`.
 @objc
 public protocol __URLSessionDelegateProviding: URLSessionDelegate {
```
ganeshnj, line 14: keeping this type for not breaking customers but essentially it is of no use. The customer delegate is swizzled and we do the right thing. There is no need to forward delegate calls.

IntegrationTests/Runner/Scenarios/URLSession/URLSessionScenarios.swift
```diff
@@ -20,22 +20,18 @@ private class InheritedURLSessionDelegate: DDURLSessionDelegate {
 /// An example of instrumenting existing `URLSessionDelegate` with `DDURLSessionDelegate` through composition.
 private class CompositedURLSessionDelegate: NSObject, URLSessionTaskDelegate, URLSessionDataDelegate, __URLSessionDelegateProviding {
```
ganeshnj, line 21: this type is just like another delegate impl.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskSwizzler.swift
```diff
@@ -0,0 +1,73 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+/// Swizzles `URLSessionTask` methods.
+internal class URLSessionTaskSwizzler {
+    static var resume: Resume?
```
maxep, line 11: Exposing this property is unsafe as it must acquire the lock, using a `@propertyWrapper` would make it safe, or it should be private.

DatadogInternal/Sources/Utils/MetaTypeExtensions.swift
```diff
@@ -0,0 +1,37 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+internal enum MetaTypeExtensions {
+    static func key(from anyClass: Any) -> String {
```
maxep, line 10: /question Why do we need to sanitize the class name?

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionSwizzler.swift
maxep: So we ended up removing the `URLSessionSwizzler` right?

maxep (commented):
It looks great!! It's a complex logic but you were able to keep it readable, well done 👍 

I've one remaining concern in various swizzler where properties are exposed while a lock is required.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- [BUGFIX] fix!: swizzle URLSessionTask for network instrumentation. See [#1394][]
```
maciejburda, line 3: Probably better to describe what do we fix instead of how

Datadog/Datadog.xcodeproj/xcshareddata/xcschemes/DatadogCore iOS.xcscheme
```diff
@@ -106,7 +106,7 @@
          </EnvironmentVariable>
          <EnvironmentVariable
             key = "DD_DISABLE_NETWORK_INSTRUMENTATION"
-            value = "$(DD_DISABLE_NETWORK_INSTRUMENTATION)"
+            value = "1"
```
maciejburda, line 109: probably we should revert

Datadog/Datadog.xcodeproj/xcshareddata/xcschemes/DatadogCore tvOS.xcscheme
```diff
@@ -92,7 +92,7 @@
          </EnvironmentVariable>
          <EnvironmentVariable
             key = "DD_DISABLE_NETWORK_INSTRUMENTATION"
-            value = "$(DD_DISABLE_NETWORK_INSTRUMENTATION)"
```
maciejburda, line 95: probably we should revert

DatadogCore/Tests/DatadogObjc/DDNSURLSessionDelegateTests.swift
```diff
@@ -7,6 +7,7 @@
 import XCTest
 import TestUtilities
 import DatadogInternal
+import DatadogRUM
```
maciejburda, line 10: Do we need this? RUM and URLSessionDelegate seem unrelated

DatadogInternal/Sources/NetworkInstrumentation/URLSession/DatadogURLSessionDelegate.swift
maciejburda: Maybe we should mark it deprecated and remove on next major release?

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskDelegate+Tracking.swift
```diff
@@ -0,0 +1,28 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+private var firstPartyHostsKey: UInt8 = 41
```
maciejburda, line 9: not used

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionSwizzler.swift
ganeshnj: yes

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskDelegate+Tracking.swift
```diff
@@ -0,0 +1,28 @@
...
+private var coreKey: UInt8 = 43
```
maciejburda, line 10: And also on keys. I'm not entirely sure if we're doing this correctly.

From what I quickly read we need to provide keys that are "guaranteed to be constant, unique, and scoped for use within getters and setters"
https://stackoverflow.com/questions/29487330/proper-key-format-for-objc-setassociatedobject

boils down to either using static or #selector

We probably want to change this in all the other places in the project.
cc @ncreated

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskDelegateSwizzler.swift
```diff
@@ -0,0 +1,115 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+/// Swizzles `URLSessionTaskDelegate` callbacks.
+internal class URLSessionTaskDelegateSwizzler {
```
maciejburda, line 10: These implementations share a lot of similar code. Would it make sense to create some decoration or even inheritance?

IntegrationTests/Runner/Scenarios/URLSession/URLSessionScenarios.swift
maciejburda: Does it make sense to remove it now?

maciejburda (changes requested):
Great effort!

One potential blocker on associated object implementation

Datadog/Datadog.xcodeproj/xcshareddata/xcschemes/DatadogCore iOS.xcscheme
ganeshnj: no - we want to keep it like that because URLSession instrumentation in our SDK doesn't play well with testing SDK URLSession instrumentation

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskDelegateSwizzler.swift
ganeshnj: I prefer this way to keep them isolated and testable, have had hard time to make them this simpler to read. 

Let's keep this way.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskDelegate+Tracking.swift
ganeshnj: good call - removed

DatadogCore/Tests/DatadogObjc/DDNSURLSessionDelegateTests.swift
ganeshnj: true - not required. I added to test things before making objc interfaces. Good call. Removed.

CHANGELOG.md
ganeshnj: Yes, looks like I have to write a whole article now on this.

DatadogInternal/Sources/Utils/MetaTypeExtensions.swift
maxep: Yes, but it's only used as key, we don't need it to be human readable, no?
and why not using `String(describing: delegateClass)`?

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskDelegate+Tracking.swift
maciejburda: although now when I look at it again - it's probably fine since it's using private global var 🤔

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskSwizzler.swift
ganeshnj: actually our ReadWriteLock has something fishy and results in thread sanitizer warnings.

I moved behind a computed property with lock.

maxep: > actually our ReadWriteLock has something fishy
sure? It's quite a simple implementation. Making the property private like you did also works 👍

IntegrationTests/Runner/Scenarios/URLSession/URLSessionScenarios.swift
ganeshnj: we will keep this until the deletion of old mechnism because new solutions gives smoother migration path.

`__URLSessionDelegateProviding` will be marked deprecated though as a separate PR.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/DatadogURLSessionDelegate.swift
ganeshnj: separate PR as it is tricky, because we treat warning as errors in tests.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskDelegate+Tracking.swift
ganeshnj: we are not using this outside delegate, I don't see any issue unless someone do the same thing for delegate which is very unlikely and we use a constant other than easy to start.

ncreated (approved):
Looks good 👌. Solid work 👍. Have we ran unit tests with iOS 12 - 15 Simulators? (to make sure it works and we won't have nightly job problems after merge)

DatadogInternal/Sources/Utils/MetaTypeExtensions.swift
ganeshnj: because this gibberish changes from time to time specially the `1130f7094` part. We need a consistent fully qualified name of the type.

DatadogCore/Tests/DatadogObjc/ObjcAPITests/DDNSURLSessionDelegate+apiTests.m
```diff
@@ -18,6 +19,27 @@ @implementation DDNSURLSessionDelegate_apiTests
...
+    [super setUp];
+
+    DDConfiguration *configuration = [[DDConfiguration alloc] initWithClientToken:@"abc" env:@"def"];
+    [DDDatadog initializeWithConfiguration:configuration trackingConsent:[DDTrackingConsent notGranted]];
+
+    DDTraceConfiguration *config = [[DDTraceConfiguration alloc] init];
+    DDTraceFirstPartyHostsTracing *tracing = [[DDTraceFirstPartyHostsTracing alloc] initWithHosts:[NSSet new] sampleRate:20];
+    DDTraceURLSessionTracking *urlSessionTracking = [[DDTraceURLSessionTracking alloc] initWithFirstPartyHostsTracing:tracing];
+    [config setURLSessionTracking:urlSessionTracking];
+    [DDTrace enableWith:config];
```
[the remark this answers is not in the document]
ganeshnj: In this particular we are testing the backward compatibility in objc.

L43, creating an instance of `DDNSURLSessionDelegate` would automatically enable the URLSession instrumentation but in order to enable this atleast RUM or Trace must be enabled which this test case covers.

but we should add one for `enable` like you mentioned.

maxep (approved):
Thank you @ganeshnj for addressing all our comments and questions, it looks great!

maxep (dismissed):

ncreated (dismissed):
🏋️‍♂️ 💯

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 
 - [IMPROVEMENT] Enable cross-platform SDKs to change app `version`. See [#1447][]
 - [IMPROVEMENT] Enable cross-platform SDKs to edit more of telemetry configuration. See [#1456][]
+- [BUGFIX] fix!: swizzle URLSessionTask for network instrumentation. See [#1394][]
```
ncreated, line 18: suggestion/ I'd say we add a feature:
```
- Add network instrumentation for async/await URLSession APIs
```
not fix a bug.

ncreated (approved):

maxep (approved):
🚀

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+- [BUGFIX] feat!: Add network instrumentation for async/await URLSession APIs. See [#1394][]
```
maciejburda, line 3: I'd argue it's `IMPROVEMENT` not a `BUGFIX`.

Also, we can probably drop `feat!:` it wasn't used in the convention so far.

ganeshnj: fixed

maciejburda (approved):

ncreated (approved):

maxep (approved):
🚀

PR #1397, opened 2023-07-27
Do not propagate Error and Long Task attributes to the View

fuzzybinary:
### What and why?

A client raised an issue that Errors would get attributes from previous errors, which was making it hard for them to query on errors. It turns out this is because we were propagating attributes of errors up to the view.

The seems like this accidentally slipped in as a side affect as there were no tests checking for it, but I'm labeling it as a breaking change in the changelog.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


DatadogRUM/Sources/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -532,7 +532,6 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
 
     private func sendErrorEvent(on command: RUMAddCurrentViewErrorCommand, context: DatadogContext, writer: Writer) {
         errorsCount += 1
-        attributes.merge(rumCommandAttributes: command.attributes)
```
louiszawadzki, line 535: Question: 
Can we consider still merging the Error and View attributes together, while not modifying the View attributes?

This will reduce the scope of the change in this PR. 
Also the Browser SDK does propagate all RUM View attributes to children events, so we might move in this direction as well for mobile SDKs. 

But I'm not against aligning all mobile SDKs and events.

louiszawadzki: EDIT: Browser SDK does propagate all global attributes to children events, there is no RUM View attributes. 
So this change will actually bring all SDKs closer 👍

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 - [IMPROVEMENT] Reduce number of view updates by filtering events from payload. See [#1328][]
+- [BUGFIX / BREAKING] Do not propagate attributes from Errors and LongTasks to Views.
```
ncreated, line 4: Thanks for updating the CHANGELOG 👌, however this diff doesn't look healthy. Given that [CHANGELOG on develop](https://github.com/DataDog/dd-sdk-ios/blob/develop/CHANGELOG.md) now includes `2.0.0`, this should go alone into "Unreleased" category on top of 2.0.0. This diff shows it differently. Please rebase 🙏.

ncreated (approved):
Thanks @fuzzybinary 💪. Two suggestions, other than this good to go 👍

PR #1412, opened 2023-08-11
REPLAY-1963 Add background task coordinator

maciejburda:
### What and why?

Adds `BackgroundTaskCoordinator` which is abstraction that allows creating `UIBackgroundTask` if UIKit is available.
This is utilised in `DataUploadWorker` to register background tasks for upload works.

With this change we receive additional 30 seconds of upload time after app goes to background.

It's a starting point for other uploader improvements that will be explained in the RFC.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
+++ b/DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
@@ -153,7 +153,7 @@ internal class FilesOrchestrator: FilesOrchestratorType {
     // MARK: - `ReadableFile` orchestration
 
     func getReadableFile(
-        excludingFilesNamed excludedFileNames: Set<String> = [],
+        excludingFilesNamed excludedFileNames: Set<String>,
         context: DatadogContext
     ) -> ReadableFile? {
         do {
--- a/DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
+++ b/DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
@@ -7,10 +7,6 @@
 import Foundation
 import DatadogInternal
 
-#if canImport(UIKit)
-import UIKit
-#endif
-
 internal protocol FilesOrchestratorType: AnyObject {
     var performance: StoragePerformancePreset { get }
 
@@ -41,14 +37,11 @@ internal class FilesOrchestrator: FilesOrchestratorType {
     /// Telemetry interface.
     let telemetry: Telemetry
 
-    #if canImport(UIKit)
-    /// Application state publisher for enriching telemetry.
-    let applicationStatePublisher: ApplicationStatePublisher = .init()
-    var subscription: ContextValueSubscription?
-    #endif
     /// Flag indicating if the application is in background state.
-    @ReadWriteLock
-    var inBackground = false
+    var inBackground: Bool {
+        return contextProvider.read().applicationStateHistory.currentSnapshot.state == .background
+    }
+    let contextProvider: DatadogContextProvider
 
     /// Extra information for metrics set from this orchestrator.
     struct MetricsData {
@@ -65,24 +58,16 @@ internal class FilesOrchestrator: FilesOrchestratorType {
         directory: Directory,
         performance: StoragePerformancePreset,
         dateProvider: DateProvider,
+        contextProvider: DatadogContextProvider,
         telemetry: Telemetry,
         metricsData: MetricsData? = nil
     ) {
         self.directory = directory
         self.performance = performance
         self.dateProvider = dateProvider
+        self.contextProvider = contextProvider
         self.telemetry = telemetry
         self.metricsData = metricsData
-
-        #if canImport(UIKit)
-        self.subscription = applicationStatePublisher.subscribe { [weak self] appStateHistory in
-            self?.inBackground = appStateHistory.currentSnapshot.state == .background
-        }
-        #endif
-    }
-
-    deinit {
-        self.subscription?.cancel()
     }
 
     // MARK: - `WritableFile` orchestration
--- a/DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
+++ b/DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
@@ -268,7 +268,6 @@ internal class FilesOrchestrator: FilesOrchestratorType {
 
         let batchAge = dateProvider.now.timeIntervalSince(fileCreationDateFrom(fileName: batchFile.name))
         let inBackground = context.applicationStateHistory.currentSnapshot.state == .background
-        print("👾 In background: \(inBackground)")
         telemetry.metric(
             name: BatchDeletedMetric.name,
             attributes: [
--- a/DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
+++ b/DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
@@ -7,6 +7,10 @@
 import Foundation
 import DatadogInternal
 
+#if canImport(UIKit)
+import UIKit
+#endif
+
 internal protocol FilesOrchestratorType: AnyObject {
     var performance: StoragePerformancePreset { get }
 
@@ -244,6 +248,12 @@ internal class FilesOrchestrator: FilesOrchestratorType {
 
         let batchAge = dateProvider.now.timeIntervalSince(fileCreationDateFrom(fileName: batchFile.name))
 
+        #if canImport(UIKit)
+        let inBackground = UIApplication.dd.managedShared?.applicationState == .background
+        #else
+        let inBackground = false
+        #endif
+
         telemetry.metric(
             name: BatchDeletedMetric.name,
             attributes: [
@@ -256,7 +266,7 @@ internal class FilesOrchestrator: FilesOrchestratorType {
                 BatchDeletedMetric.uploaderWindowKey: performance.uploaderWindow.toMilliseconds,
                 BatchDeletedMetric.batchAgeKey: batchAge.toMilliseconds,
                 BatchDeletedMetric.batchRemovalReasonKey: deletionReason.toString(),
-                BatchDeletedMetric.inBackgroundKey: false,
+                BatchDeletedMetric.inBackgroundKey: inBackground,
             ]
         )
     }
--- a/DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
+++ b/DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
@@ -10,10 +10,14 @@ import DatadogInternal
 internal protocol FilesOrchestratorType: AnyObject {
     var performance: StoragePerformancePreset { get }
 
-    func getNewWritableFile(writeSize: UInt64) throws -> WritableFile
-    func getWritableFile(writeSize: UInt64) throws -> WritableFile
-    func getReadableFile(excludingFilesNamed excludedFileNames: Set<String>) -> ReadableFile?
-    func delete(readableFile: ReadableFile, deletionReason: BatchDeletedMetric.RemovalReason)
+    func getNewWritableFile(writeSize: UInt64, context: DatadogContext) throws -> WritableFile
+    func getWritableFile(writeSize: UInt64, context: DatadogContext) throws -> WritableFile
+    func getReadableFile(excludingFilesNamed excludedFileNames: Set<String>, context: DatadogContext) -> ReadableFile?
+    func delete(
+        readableFile: ReadableFile,
+        deletionReason: BatchDeletedMetric.RemovalReason,
+        context: DatadogContext
+    )
 
     var ignoreFilesAgeWhenReading: Bool { get set }
 }
@@ -37,12 +41,6 @@ internal class FilesOrchestrator: FilesOrchestratorType {
     /// Telemetry interface.
     let telemetry: Telemetry
 
-    /// Flag indicating if the application is in background state.
-    var inBackground: Bool {
-        return contextProvider.read().applicationStateHistory.currentSnapshot.state == .background
-    }
-    let contextProvider: DatadogContextProvider
-
     /// Extra information for metrics set from this orchestrator.
     struct MetricsData {
         /// The name of the track reported for this orchestrator.
@@ -58,14 +56,12 @@ internal class FilesOrchestrator: FilesOrchestratorType {
         directory: Directory,
         performance: StoragePerformancePreset,
         dateProvider: DateProvider,
-        contextProvider: DatadogContextProvider,
         telemetry: Telemetry,
         metricsData: MetricsData? = nil
     ) {
         self.directory = directory
         self.performance = performance
         self.dateProvider = dateProvider
-        self.contextProvider = contextProvider
         self.telemetry = telemetry
         self.metricsData = metricsData
     }
@@ -79,19 +75,19 @@ internal class FilesOrchestrator: FilesOrchestratorType {
     ///
     /// - Parameter writeSize: the size of data to be written
     /// - Returns: `WritableFile` capable of writing data of given size
-    func getNewWritableFile(writeSize: UInt64) throws -> WritableFile {
+    func getNewWritableFile(writeSize: UInt64, context: DatadogContext) throws -> WritableFile {
         try validate(writeSize: writeSize)
         if let closedBatchName = lastWritableFileName {
             sendBatchClosedMetric(fileName: closedBatchName, forcedNew: true)
         }
-        return try createNewWritableFile(writeSize: writeSize)
+        return try createNewWritableFile(writeSize: writeSize, context: context)
     }
 
     /// Returns writable file accordingly to default heuristic of creating and reusing files.
     ///
     /// - Parameter writeSize: the size of data to be written
     /// - Returns: `WritableFile` capable of writing data of given size
-    func getWritableFile(writeSize: UInt64) throws -> WritableFile {
+    func getWritableFile(writeSize: UInt64, context: DatadogContext) throws -> WritableFile {
         try validate(writeSize: writeSize)
 
         if let lastWritableFile = reuseLastWritableFileIfPossible(writeSize: writeSize) { // if last writable file can be reused
@@ -102,7 +98,7 @@ internal class FilesOrchestrator: FilesOrchestratorType {
             if let closedBatchName = lastWritableFileName {
                 sendBatchClosedMetric(fileName: closedBatchName, forcedNew: false)
             }
-            return try createNewWritableFile(writeSize: writeSize)
+            return try createNewWritableFile(writeSize: writeSize, context: context)
         }
     }
 
@@ -112,13 +108,13 @@ internal class FilesOrchestrator: FilesOrchestratorType {
         }
     }
 
-    private func createNewWritableFile(writeSize: UInt64) throws -> WritableFile {
+    private func createNewWritableFile(writeSize: UInt64, context: DatadogContext) throws -> WritableFile {
         // NOTE: RUMM-610 Because purging files directory is a memory-expensive operation, do it only when a new file
         // is created (we assume here that this won't happen too often). In details, this is to avoid over-allocating
         // internal `_FileCache` and `_NSFastEnumerationEnumerator` objects in downstream `FileManager` routines.
         // This optimisation results with flat allocation graph in a long term (vs endlessly growing if purging
         // happens too often).
-        try purgeFilesDirectoryIfNeeded()
+        try purgeFilesDirectoryIfNeeded(context: context)
 
         let newFileName = fileNameFrom(fileCreationDate: dateProvider.now)
         let newFile = try directory.createFile(named: newFileName)
@@ -156,11 +152,16 @@ internal class FilesOrchestrator: FilesOrchestratorType {
 
     // MARK: - `ReadableFile` orchestration
 
-    func getReadableFile(excludingFilesNamed excludedFileNames: Set<String> = []) -> ReadableFile? {
+    func getReadableFile(
+        excludingFilesNamed excludedFileNames: Set<String> = [],
+        context: DatadogContext
+    ) -> ReadableFile? {
         do {
             let filesWithCreationDate = try directory.files()
                 .map { (file: $0, creationDate: fileCreationDateFrom(fileName: $0.name)) }
-                .compactMap { try deleteFileIfItsObsolete(file: $0.file, fileCreationDate: $0.creationDate) }
+                .compactMap {
+                    try deleteFileIfItsObsolete(file: $0.file, fileCreationDate: $0.creationDate, context: context)
+                }
 
             guard let (oldestFile, creationDate) = filesWithCreationDate
                 .filter({ excludedFileNames.contains($0.file.name) == false })
@@ -186,10 +187,18 @@ internal class FilesOrchestrator: FilesOrchestratorType {
         }
     }
 
-    func delete(readableFile: ReadableFile, deletionReason: BatchDeletedMetric.RemovalReason) {
+    func delete(
+        readableFile: ReadableFile,
+        deletionReason: BatchDeletedMetric.RemovalReason,
+        context: DatadogContext
+    ) {
         do {
             try readableFile.delete()
-            sendBatchDeletedMetric(batchFile: readableFile, deletionReason: deletionReason)
+            sendBatchDeletedMetric(
+                batchFile: readableFile,
+                deletionReason: deletionReason,
+                context: context
+            )
         } catch {
             telemetry.error("Failed to delete file", error: error)
         }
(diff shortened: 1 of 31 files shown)


DatadogCore/Sources/Core/Upload/BackgroundTaskCoordinator.swift
```diff
@@ -0,0 +1,41 @@
...
+/// The `BackgroundTaskCoordinator` class provides an abstraction for managing background tasks and includes methods for registering and ending background tasks.
+/// It also serves as a useful abstraction for testing purposes.
+internal class UIKitBackgroundTaskCoordinator: BackgroundTaskCoordinator {
+    internal var tasks: [UUID: UIBackgroundTaskIdentifier] = [:]
+
+    internal func registerBackgroundTask() -> UUID {
+        let uuid = UUID()
+        tasks[uuid] = UIApplication.dd.managedShared?.beginBackgroundTask { [weak self] in
+            self?.endBackgroundTaskIfActive(uuid)
+        }
```
ncreated, line 27: In [`handler` doc](https://developer.apple.com/documentation/uikit/uiapplication/1623031-beginbackgroundtask) Apple states:
> The system calls the handler synchronously **on the main thread**, blocking the app’s suspension momentarily.

which makes the `internal var tasks:` not thread-safe. When `handler` is invoked on the main thread while `DataUploadWorker` executes `registerBackgroundTask()` (on background thread), this will crash.

This can be solved with a design that doesn't imply shared state, e.g.:
```swift
let backgroundTask = coordinator.register()
backgroundTask.end()
```

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
```diff
@@ -119,23 +73,67 @@ internal class DataUploadWorker: DataUploadWorkerType {
...
-                }
-                do {
-                    // Try uploading the batch and do one more retry on failure.
-                    _ = try self.dataUploader.upload(events: nextBatch.events, context: contextProvider.read())
-                } catch {
-                    _ = try? self.dataUploader.upload(events: nextBatch.events, context: contextProvider.read())
-                }
+        queue.sync { [weak self] in
+            while let nextBatch = self?.fileReader.readNextBatch() {
+                self?.tryToSendBatch(nextBatch)
```
ncreated, line 136: The new `tryToSendBatch()` call involves the full retry logic. This means: if calling `flushSynchronously()` under insufficient upload conditions (e.g. offline), the caller thread will run into infinite while loop.

See `flushSynchronously()` comment:
> /// - It performs arbitrary upload (without checking upload condition and without re-transmitting failed uploads).

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
```diff
@@ -34,76 +34,30 @@ internal class DataUploadWorker: DataUploadWorkerType {
...
-                let batchLabel = nextBatch != nil ? "YES" : (isSystemReady ? "NO" : "NOT CHECKED")
-                DD.logger.debug("💡 (\(self.featureName)) No upload. Batch to upload: \(batchLabel), System conditions: \(blockersForUpload.description)")
-
-                self.delay.increase()
-            }
-
-            self.scheduleNextUpload(after: self.delay.current)
+        let uploadWork = DispatchWorkItem { [weak self, delay] in
+            self?.tryToSendBatch()
+            self?.scheduleNextUpload(after: delay.current)
```
ncreated, line 60: regression/ We can't capture `delay` this way - it is a value type, so we're making a copy here. In result, the `delay` is never altered and SDK keeps sending data in fixed, initial interval, e.g.:

Screenshot 2023-08-16 at 17 44 05

ncreated (changes requested):
The idea of using OS' background API is IMO good 👍. I noted few problems with the implementation.

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
ncreated: 🏕️ boy scouting/ please check why this wasn't captured by tests - clearly we're missing something in tests and this is the core part of the SDK.

DatadogCore/Sources/Core/Upload/BackgroundTaskCoordinator.swift
```diff
@@ -0,0 +1,30 @@
...
+}
+
+internal typealias BackgroundTaskIdentifier = Int
+
+#if canImport(UIKit)
+import UIKit
+
+internal class UIKitBackgroundTaskCoordinator: BackgroundTaskCoordinator {
+    internal func registerBackgroundTask() -> BackgroundTaskIdentifier? {
+        return UIApplication.dd.managedShared?.beginBackgroundTask().rawValue
```
ganeshnj, line 23: https://developer.apple.com/documentation/uikit/uiapplication/1623031-beginbackgroundtask

> [Parameters](https://developer.apple.com/documentation/uikit/uiapplication/1623031-beginbackgroundtask#parameters)
handler
A handler to be called shortly before the app’s remaining background time reaches 0. Use this handler to clean up and mark the end of the background task. Failure to end the task explicitly will result in the termination of the app. The system calls the handler synchronously on the main thread, blocking the app’s suspension momentarily.

should there be a handler for cleanup?

ganeshnj (commented):
Looks good, just a question on cleanup.

maxep (commented):
Nice work, it's simple and well abstracted 👏 
I left some minor comments/suggestions

DatadogCore/Sources/Core/Upload/DataUploadDelay.swift
```diff
@@ -9,12 +9,12 @@ import DatadogInternal
...
     var current: TimeInterval { get }
-    mutating func decrease()
-    mutating func increase()
+    func decrease()
+    func increase()
 }
 
 /// Mutable interval used for periodic data uploads.
-internal struct DataUploadDelay: Delay {
+internal class DataUploadDelay: Delay {
```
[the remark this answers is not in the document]
maciejburda: It's solving the problem of passing the copy. I think it was mainly causing issues in tests (inability to inspect the injected Delay). But technically it prevents from capturing the immutable copy of delay within closures (which I did in the very first version of the PR).

DatadogCore/Sources/Core/Upload/BackgroundTaskCoordinator.swift
ncreated: blocking/ 👍  At minimum we must call the `endBackgroundTask()` in cleanup handler to avoid crashing the app:

> Each call to this method must be balanced by a matching call to the [endBackgroundTask(_:)](https://developer.apple.com/documentation/uikit/uiapplication/1622970-endbackgroundtask) method. (...) If you don't call `endBackgroundTask(_:)` for each task before time expires, the system kills the app.

See Apple's example code [here](https://developer.apple.com/documentation/uikit/app_and_environment/scenes/preparing_your_ui_to_run_in_the_background/extending_your_app_s_background_execution_time#overview). Keep in mind that cleanup handler is told to be called on the main thread - this Apple's example code doesn't cover `self.backgroundTaskID` thread-safety.

DatadogCore/Tests/Datadog/Core/Upload/DataUploadWorkerTests.swift
```diff
@@ -166,109 +166,105 @@ class DataUploadWorkerTests: XCTestCase {
...
-
-        let dataUploader = DataUploader(
-            httpClient: httpClient,
-            requestBuilder: FeatureRequestBuilderMock()
-        )
         let worker = DataUploadWorker(
             queue: uploaderQueue,
             fileReader: reader,
-            dataUploader: dataUploader,
+            dataUploader: DataUploaderMock(uploadStatus: .mockWith()),
```
ncreated, line 254: How do we configure "When Batch Succeeds" in this setup now? It was `.mockResponseWith(statusCode: 200)` before, but it is not clear right now.

ncreated (changes requested):

DatadogCore/Tests/Datadog/Core/Upload/DataUploadWorkerTests.swift
maciejburda: it's based `needsRetry` property, which is by default set to false in the mock. I made it explicit now.

DatadogCore/Sources/Core/Upload/BackgroundTaskCoordinator.swift
```diff
@@ -0,0 +1,38 @@
...
+/// the fact that UIBackgroundTaskIdentifier raw value is based on Int.
+internal protocol BackgroundTaskCoordinator {
+    func registerBackgroundTask() -> Int
+    func endBackgroundTaskIfActive(_ backgroundTaskIdentifier: Int)
+}
+
+#if canImport(UIKit)
+import UIKit
+
+internal class UIKitBackgroundTaskCoordinator: BackgroundTaskCoordinator {
```
ncreated, line 20: suggestion/ As we introduce reusable component here, that isn't thread safe by design, the class comment should explain what are the constraints of using it in different parts of the codebase.

ncreated (approved):
Looks ok 👍.

What about observability of batches that are sent (deleted) while the app is in background? The metrics we added recently include `is_background` attribute that was made for this. Do we have task to cover it?

DatadogCore/Sources/Core/Upload/BackgroundTaskCoordinator.swift
```diff
@@ -0,0 +1,38 @@
...
+        guard let app = UIApplication.dd.managedShared else {
+            return UIBackgroundTaskIdentifier.invalid.rawValue
+        }
+        currentBackgroundTask = app.beginBackgroundTask(expirationHandler: { [currentBackgroundTask] in
+            UIApplication.dd.managedShared?.endBackgroundTask(currentBackgroundTask)
+        })
+        return currentBackgroundTask.rawValue
+    }
+
+    func endBackgroundTaskIfActive(_ backgroundTaskIdentifier: Int) {
```
ganeshnj, line 33: question

how it is being checked whether the background task is active or not?

also, in what case `managedShared` will be nil? Just want to understand the lifecycle here.

maciejburda: managedShared was introduced to enable compilation for certain targets (i.e. extension).

Currently if it's nil, it would create a invalid task

I'll add some extra checks to make it explicit 🙌

DatadogCore/Sources/Core/Upload/DataUploadDelay.swift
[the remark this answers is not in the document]
maciejburda: I'm on a fence with this tbh.

having them mutating enforces the usage of var instead of let, but in reality mutability is achieved internally in Delay implementation.

DatadogCore/Sources/Core/Upload/BackgroundTaskCoordinator.swift
```diff
@@ -0,0 +1,61 @@
...
+
+    internal func beginBackgroundTask(expirationHandler handler: @escaping (() -> Void)) -> Int {
+        guard let app = app else {
+            return UIBackgroundTaskIdentifier.invalid.rawValue
+        }
+        return app.beginBackgroundTask(expirationHandler: { [weak self] in
+            self?.queue.async {
+                handler()
+            }
+        }).rawValue
```
ncreated, line 50: From the [`expirationHandler`](https://developer.apple.com/documentation/uikit/uiapplication/1623031-beginbackgroundtask) docs:
> The system calls the handler synchronously on the main thread, blocking the app’s suspension momentarily.

Isn't the (OS) expectation to run handler's logic also synchronously? Each "start task" must be balanced with "end task" and we call the `endBackgroundTaskIfActive(taskID)` from this closure in upstream unit, so eventually we do it async. It seems that this code is not safe:

> If you don't call `endBackgroundTask(_:)` for each task before time expires, the system kills the app.

DatadogCore/Sources/Core/Upload/DataUploadDelay.swift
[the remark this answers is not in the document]
maxep: > having them mutating enforces the usage of var instead of let

I don't see what's wrong about that, especially with `private` accessor. `Delay` protocol has no implementation, and specifying `mutating` is required because you don't know if the provided implementation is a class or a struct.

```swift
protocol Delay {
    var current: TimeInterval { get } // <- This is expected to be mutated
    mutating func decrease()
    mutating func increase()
}
```
The other option is to constrain the `Delay` to classes with `protocol Delay: AnyObject` but I don't see the need for that.

maciejburda:
Updated with some work in progress. Please refrain from reviewing for the time being.

DatadogCore/Sources/Core/Upload/DataUploadDelay.swift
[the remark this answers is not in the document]
maciejburda: The only problem I see it that using mutating enforces `DataUploadWorker` to carry `var` delay which gives a false impression of mutability. In reality it's a class with some internal logic, which actually in some cases doesn't even mutate the state (for example when it reaches top/bottom value of the delay).

I think that `mutating` was added because initially implementation of the Delay was based on struct. Now mutability is achieved by altering the current value internally in `DataUploadDelay`. IMO from the protocol level we don't need to point out that mutability.

DatadogCore/Sources/Core/Upload/BackgroundTaskCoordinator.swift
```diff
@@ -0,0 +1,60 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+/// The `BackgroundTaskCoordinator` protocol provides an abstraction for managing background tasks and includes methods for registering and ending background tasks.
+/// It serves as a useful abstraction for testing purposes as well as allows decoupling from UIKit in order to maintain Catalyst compliation. To abstract from UIKit, it leverages
+/// the fact that UIBackgroundTaskIdentifier raw value is based on Int.
+internal protocol BackgroundTaskCoordinator {
```
ganeshnj, line 12: although it is internal, we should provide some documentation on methods.

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
```diff
@@ -102,6 +109,7 @@ internal class DataUploadWorker: DataUploadWorkerType {
                 DD.logger.debug("💡 (\(self.featureName)) No upload. Batch to upload: \(batchLabel), System conditions: \(blockersForUpload.description)")
 
                 self.delay.increase()
+                self.backgroundTaskCoordinator?.endCurrentBackgroundTaskIfActive()
```
ganeshnj, line 112: ❓ How do we make sure `endCurrentBackgroundTaskIfActive` always called?

I think in a `defer` block would make more sense.

ganeshnj (changes requested):

DatadogCore/Sources/Core/Upload/DataUploadDelay.swift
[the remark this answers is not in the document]
maxep: I'm sorry to re-open this minor change, it's a nit-pick but I would like to keep things semantically correct..

`Delay` is a protocol, not a class nor a struct, the `mutating` is defined because the protocol could be applied on any type, on reference types or on value types. Its `current` property will be mutated by both of the methods, and the `DataUploadWorker` doesn't know what type of object the delay instance is, therefore it's a `var`. The previous implementation was correct, and safe with the `private` accessor.

Now if you really want to get rid of the `mutating` keyword, you have to constrain the `Delay` protocol to be class-only with:
```swift
protocol Delay: AnyObject {
   var current: TimeInterval { get }
   func decrease()
   func increase()
}
```
That's IMHO unnecessary, but it would be valid 👍

DatadogCore/Sources/Datadog.swift
```diff
@@ -105,6 +105,9 @@ public struct Datadog {
         /// The bundle object that contains the current executable.
         public var bundle: Bundle
 
+        /// Flag that determines if UIKit's background tasks are utilized to perform uploads in background.
+        public var backgroundTasksEnabled: Bool
+
         /// Creates a Datadog SDK Configuration object.
         ///
         /// - Parameters:
```
maxep, line 113: I think we should be very clear in the doc that this property will extend the lifecycle of the app up to 30s in background to proceed with data upload. It should also be added to the `init` list of parameters 👍

maxep (commented):
Thanks for the updates @maciejburda , it reads very well!

We might be lacking tests in real environment, in unit tests we are testing against `UIKitAppBackgroundTaskCoordinator ` mocks only and I wonder if it would be possible to have Integration Tests (using UITests) that push the app to background and assert data transfer. Would that be possible?

DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
```diff
@@ -61,6 +73,16 @@ internal class FilesOrchestrator: FilesOrchestratorType {
         self.dateProvider = dateProvider
         self.telemetry = telemetry
         self.metricsData = metricsData
+
+        #if canImport(UIKit)
+        self.subscription = applicationStatePublisher.subscribe { [weak self] appStateHistory in
+            self?.inBackground = appStateHistory.currentSnapshot.state == .background
+        }
+        #endif
```
ncreated, line 81: The [`ApplicationStatePublisher`](https://github.com/DataDog/dd-sdk-ios/blob/c645cce1fd43256c4acb67e4e7462e27e815eca5/DatadogCore/Sources/Core/Context/ApplicationStatePublisher.swift#L19-L22) comes with a dispatch queue. If used within `FilesOrchestrator`, it will be (pessimistically) adding 1 thread per feature to manage this subscription. This won't be scalable.

In fact, this is V1-like approach, which was modernised in V2. In V2 `AppStateHistory` is already available in [`DatadogContext.applicationStateHistory`](https://github.com/DataDog/dd-sdk-ios/blob/c645cce1fd43256c4acb67e4e7462e27e815eca5/DatadogInternal/Sources/Context/DatadogContext.swift#L73-L74), so there is no need to subscribe again and consume more resources.

We need `currentSnapshot.state` in batch removal flow, hence we can capture it from here and pass downstream to `sendBatchDeletedMetric(batchFile:deletionReason:)`:
https://github.com/DataDog/dd-sdk-ios/blob/c645cce1fd43256c4acb67e4e7462e27e815eca5/DatadogCore/Sources/Core/Upload/DataUploadWorker.swift#L60

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
ncreated, line 112: question/ To what I remember, we discussed app terminations if `end()` is not called within 30 seconds after `begin()`, no? How do we assure it 🤔 💭?

DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
maciejburda: Great input! I wasn't really sure about this one.

Thanks for the additional context - I'll make use of it 🙌

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
maciejburda: It was a problem of previous setup - for some reason calling UIKit's handler wasn't ending in a balanced state, which was causing watchdog to terminate the process.

Here it's simplified in a way that single feature hold maximum of on background task registration and it works perfectly fine. You can test by commenting this line and observing that app goes into suspended state after 30 seconds.

DatadogCore/Sources/Core/Upload/DataUploadDelay.swift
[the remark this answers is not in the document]
maciejburda: As discussed protocol is actually not needed in this case. I'll remove it :)

ganeshnj (dismissed):

maciejburda:
Added a separate task for UI test testing background upload.

DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
```diff
@@ -41,14 +37,11 @@ internal class FilesOrchestrator: FilesOrchestratorType {
...
-    let applicationStatePublisher: ApplicationStatePublisher = .init()
-    var subscription: ContextValueSubscription?
-    #endif
     /// Flag indicating if the application is in background state.
-    @ReadWriteLock
-    var inBackground = false
+    var inBackground: Bool {
+        return contextProvider.read().applicationStateHistory.currentSnapshot.state == .background
+    }
+    let contextProvider: DatadogContextProvider
```
ncreated, line 44: Instead of calling `contextProvider.read()`, would it be possible to pass the `context` from upstream routines (as method param)? This `.read()` doesn't come for free:

https://github.com/DataDog/dd-sdk-ios/blob/8be9c4c9c20443d53a1849ccc984c343e8901869/DatadogCore/Sources/Core/Context/DatadogContextProvider.swift#L92-L98

Effectively, we will be calling context queue with every batch closed. 

@maxep What's your take on this? Isn't there a risk of reaching deadlock lock on context queue?

maxep: Good catch! That's a very expensive task for only telemetry and could indeed lead to a deadlock since we don't know on which thread we are.

Preferable to use the [`applicationState`](https://developer.apple.com/documentation/uikit/uiapplication/1623003-applicationstate) from the `UIApplication`.

ncreated: > Preferable to use the [applicationState](https://developer.apple.com/documentation/uikit/uiapplication/1623003-applicationstate) from the UIApplication.

Could be an option as well 👌. @maciejburda pls make sure it is safe to use from BG thread.

From different angle tho, the `FilesOrchestrator` implements 3 main routines:
```swift
func getWritableFile(...) throws -> WritableFile
func getReadableFile(...) -> ReadableFile?
func delete(readableFile: ReadableFile, ...)
```
The first one is called from Event Write Context API and the other two are triggered from Data Upload Worker. In both these callers we know the context:
* [EVC context](https://github.com/DataDog/dd-sdk-ios/blob/8be9c4c9c20443d53a1849ccc984c343e8901869/DatadogCore/Sources/Core/DatadogCore.swift#L314-L319)
* [Data Upload Worker context](https://github.com/DataDog/dd-sdk-ios/blob/8be9c4c9c20443d53a1849ccc984c343e8901869/DatadogCore/Sources/Core/Upload/DataUploadWorker.swift#L60)

We could pass the `context` value from these calls downstream to `FilesOrchestratorType`, this way avoiding extra cost or dependency in `FilesOrchestrator`. Keep in mind that `DatadogContext` is immutable and thread-safe. @maxep WDYT?

maxep: Ah you are right, `UIApplication.applicationState` must be called from the main thread 😭  [see this](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogCore/Sources/Core/DatadogCore.swift#L397L401) 
So yea, injecting the context from caller is safer!

maciejburda: Refactored to top level context getting

DatadogCore/Sources/Datadog.swift
```diff
@@ -150,7 +158,8 @@ public struct Datadog {
             uploadFrequency: UploadFrequency = .average,
             proxyConfiguration: [AnyHashable: Any]? = nil,
             encryption: DataEncryption? = nil,
-            serverDateProvider: ServerDateProvider? = nil
+            serverDateProvider: ServerDateProvider? = nil,
+            backgroundTasksEnabled: Bool = false
```
ncreated, line 162: change-request/ This is new public API and we must document it as part of `init()` doc.

ncreated (changes requested):
LG in overall 👌. I left few small change requests, no major blockers.

ncreated (dismissed):

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
```diff
@@ -94,14 +105,15 @@ internal class DataUploadWorker: DataUploadWorkerType {
...
-                    self.fileReader.markBatchAsRead(batch, reason: .invalid)
+                    self.fileReader.markBatchAsRead(batch, reason: .invalid, context: context)
                     telemetry.error("Failed to initiate '\(self.featureName)' data upload", error: error)
                 }
             } else {
                 let batchLabel = nextBatch != nil ? "YES" : (isSystemReady ? "NO" : "NOT CHECKED")
                 DD.logger.debug("💡 (\(self.featureName)) No upload. Batch to upload: \(batchLabel), System conditions: \(blockersForUpload.description)")
 
                 self.delay.increase()
+                self.backgroundTaskCoordinator?.endBackgroundTask()
```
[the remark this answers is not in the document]
maciejburda: Yeah - I am on a fence with this. It really boils down to linking the background task documentation.

In apple docs they say it should be balanced - in reality if you close the current task within handler closure it's going to be fine (hence "may terminate").

I will try to rephrase.

ncreated (approved):

PR #1413, opened 2023-08-14
RUMM-3509 fix: Can't create custom instance of SDK after default one

ncreated:
### What and why?

🐞 This PR fixes the problem of creating custom instance of SDK **after** the default one was created.

### How?

The root cause was this check in instance registration flow, that must be more specialised (to use `instanceName`):
```swift
if CoreRegistry.default is DatadogCore {
   throw ProgrammerError(description: "SDK is already initialized.")
}
```

This problem was not discovered earlier due to lack of tests coverage for "multiple instances" setup. To fill this gap, this PR adds integration-unit:
```swift
func testGivenTwoInstancesOfSDK_whenCollectingLogs_thenEachSDKUploadsItsOwnData()
```
that validates basics of using Logs in two-instance configuration. This is also where the rest of changes in this PR originate:
* To test 2 SDK instances, the existing `ServerMock` was not enough as it only allows single `URLSession`.
* I added `HTTPClient` interface, which can be mocked and injected through `Datadog.Configuration`. This makes requests interception easy. It doesn't depend on static `URLProtocol`, so we can mock separate interceptor per SDK.
* As `HTTPClientMock` required re-implementing "requests decompression" capability, I moved that to a separate `URLRequest` extension and managed to simplify some of existing tests. 

Last, the `DatadogCoreProtocol` protocol was not re-exported from `DatadogCore` module, making following property declaration not compile. This is now solved:
```swift
import DatadogCore

let customInstance: DatadogCoreProtocol
```

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


DatadogCore/Sources/Core/Upload/URLSessionClient.swift
ncreated, line 1: 💡 Nothing new in this file, except renaming and adding protocol conformance:
```diff
- internal class HTTPClient
+ internal protocol HTTPClient
+ internal class URLSessionClient: HTTPClient
```

DatadogCore/Sources/Core/Upload/URLSessionClient.swift
```diff
@@ -0,0 +1,75 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+/// Client for sending requests over HTTP.
+internal class URLSessionClient: HTTPClient {
```
ncreated, line 10: 💡 Nothing new in this file, except renaming and adding protocol conformance:
```diff
- internal class HTTPClient
+ internal protocol HTTPClient
+ internal class URLSessionClient: HTTPClient
```

DatadogCore/Sources/Datadog.swift
```diff
@@ -315,8 +330,8 @@ public struct Datadog {
         trackingConsent: TrackingConsent,
         instanceName: String
     ) throws -> DatadogCoreProtocol {
-        if CoreRegistry.default is DatadogCore {
-            throw ProgrammerError(description: "SDK is already initialized.")
+        guard !CoreRegistry.isRegistered(instanceName: instanceName) else {
+            throw ProgrammerError(description: "The '\(instanceName)' instance of SDK is already initialized.")
         }
```
ncreated, line 335: 🎯 This is the actual fix for this PR (🐞 solves the problem of instantiating custom instance of the SDK after default one was created).

DatadogCore/Tests/Datadog/Core/Upload/DataUploaderTests.swift
```diff
@@ -10,20 +10,18 @@ import TestUtilities
...
+    // swiftlint:disable opening_brace
+    func testGivenValidRequest_whenUploadCompletesWithStatusCode_itReturnsUploadStatus() throws {
         // Given
         let randomResponse: HTTPURLResponse = .mockResponseWith(statusCode: (100...599).randomElement()!)
-        let randomRequestIDOrNil: String? = Bool.random() ? .mockRandom() : nil
-        let requestIDHeaderOrNil: URLRequestBuilder.HTTPHeader? = randomRequestIDOrNil.flatMap { randomRequestID in
-                .init(field: URLRequestBuilder.HTTPHeader.ddRequestIDHeaderField, value: { randomRequestID })
-        }
-
-        let server = ServerMock(delivery: .success(response: randomResponse))
```
ncreated, line 21: 💡 A common pattern in this PR - we can now replace most usages of `ServerMock` with `HTTPClientMock`. The latter doesn't depend on mocking `URLProtocol` and we can use multiple instances of `HTTPClientMock` in single test.

DatadogCore/Sources/Datadog.swift
```diff
@@ -169,6 +163,27 @@ public struct Datadog {
...
+        /// This can be used to tweak internal features of the SDK.
+        internal var additionalConfiguration: [String: Any] = [:]
+
+        /// Default date provider used by the SDK and all products.
+        internal var dateProvider: DateProvider = SystemDateProvider()
+
+        /// Creates `HTTPClient` with given proxy configuration attributes.
+        internal var httpClientFactory: ([AnyHashable: Any]?) -> HTTPClient = { proxyConfiguration in
+            URLSessionClient(proxyConfiguration: proxyConfiguration)
+        }
```
maxep, line 186: Lets continue to resolve the config in `Datadog.initialize` and not in `Datadog.Configuration` itself, so it's easier to read and maintain. I would remove this in favor of `URLSessionClient(proxyConfiguration: configuration .proxyConfiguration)` in `Datadog.initialize`.

DatadogCore/Sources/Datadog.swift
```diff
@@ -346,14 +361,14 @@ public struct Datadog {
...
-                in: Directory.cache(),
+                in: configuration.systemDirectory(),
                 instancenName: instanceName,
                 site: configuration.site
             ),
             dateProvider: configuration.dateProvider,
             initialConsent: trackingConsent,
             performance: performance,
-            httpClient: HTTPClient(proxyConfiguration: configuration.proxyConfiguration),
+            httpClient: configuration.httpClientFactory(configuration.proxyConfiguration),
```
maxep, line 371: Following my comment above:
```suggestion
            httpClient: URLSessionClient(proxyConfiguration: configuration.proxyConfiguration),
```

DatadogCore/Sources/Datadog.swift
```diff
@@ -169,6 +163,27 @@ public struct Datadog {
             self.encryption = encryption
             self.serverDateProvider = serverDateProvider ?? DatadogNTPDateProvider()
         }
+
+        // MARK: - Internal
+
+        /// Obtains OS directory where SDK creates its root folder.
+        /// All instances of the SDK use the same root folder, but each creates its own subdirectory.
+        internal var systemDirectory: () throws -> Directory = { try Directory.cache() }
```
maxep, line 171: `Directory` is more of a convenient struct than a data object, in this case an optional `URL?` would be more appropriate.

```swift
internal var directory: URL? = nil

...

try CoreDirectory(
    in: Directory.cache(),
    in: configuration.directory.map { Directory(url: $0) } ?? try Directory.cache(),
    instancenName: instanceName,
    site: configuration.site
)
```

maxep (commented):
Looks great 🎉  thanks for the various refacto!

I've left a couple of suggestion regarding the `Datadog.Configuration` change. I think we should keep it as data object without any logic, and keep the resolution in `Datadog.intialize`. LMKWYT!

Datadog/IntegrationUnitTests/Public/Datadog+MultipleInstancesIntegrationTests.swift
```diff
@@ -0,0 +1,100 @@
...
+        customInstanceConfig.httpClientFactory = { _ in customHTTPClient }
+        defaultInstanceConfig.bundle = .mockWith(bundleIdentifier: "com.bundle.default", CFBundleShortVersionString: "1.0-default")
+        customInstanceConfig.bundle = .mockWith(bundleIdentifier: "com.bundle.custom", CFBundleShortVersionString: "1.0-custom")
+
+        // Given
+        Datadog.initialize(with: defaultInstanceConfig, trackingConsent: .granted)
+        Datadog.initialize(with: customInstanceConfig, trackingConsent: .granted, instanceName: customInstanceName)
+
+        Logs.enable(with: .init())
+        Logs.enable(with: .init(), in: Datadog.sdkInstance(named: customInstanceName))
```
ganeshnj, line 49: is it worth checking other features? 

Specially crash reporting.

DatadogCore/Sources/Datadog.swift
ncreated: This change was motivated by testability. Notably, being able to initialize SDK using public API with swapping the single dependency in this integration-unit:
```swift
func testGivenTwoInstancesOfSDK_whenCollectingLogs_thenEachSDKUploadsItsOwnData() throws {
   let config: Datadog.Configuration = ...
   config.httpClientFactory = { _ in HTTPClientMock(responseCode: 200) }

   Datadog.initialize(with: config, trackingConsent: .granted)
   // ...
}
```
In that sense, the `Datadog.Configuration` acts as a basic DI container for passing internal dependencies and mocking them in tests. Same way as we do for `dateProvider`, `processInfo` and few other components.

TestUtilities/Helpers/Decompression.swift
```diff
@@ -58,3 +58,25 @@ public struct zlib {
...
+
+public struct DecompressionError: Error, CustomStringConvertible {
+    public var description: String
+}
+
+public extension URLRequest {
+    /// Returns a copy of the URLRequest with a potentially decompressed body.
+    /// - Returns: A copy of the URLRequest with a decompressed body if applicable, otherwise the original request.
+    /// - Throws: A `DecompressionError` if decompression fails.
+    func decompressed() throws -> URLRequest {
```
ganeshnj, line 70: This could go to Internal module. I use similar snippet to check the request body while debugging. However, its use must be avoided as it may cause memory spike in the SDK.

maciejburda (dismissed):
LGTM

TestUtilities/Helpers/Decompression.swift
ncreated: If it's only for debugging, I wonder if we should make it into release code. Especially that it rather sounds like "preference" (I personally use `rum-mock-server` for debugging).

ganeshnj (approved):

DatadogCore/Sources/Datadog.swift
maxep: Why not? It's the same as the `httpClientFactory` but in a more explicit form :)

The goal is testability and for that we need a default path and overrides for tests, we can use various solutions with more or less impacts. The closure/factory pattern is one of them but it differs from what we currently have with `additionalConfiguration`, `bundle`, or `processInfo` where the config is resolved in the `initialize` method only.

Now, it's not a big deal because it's all internal, but it's not consistent with what we have. Also, I find the following clearer than closure parameters:
```swift
/// overrides the data directory.
internal var directory: URL? = nil
/// overrides the uploader HTTP client.
internal var httpClient: HTTPClient? = nil
```

Now it's my humble opinion and I don't want to block this PR if you think otherwise.

DatadogCore/Sources/Datadog.swift
maxep: See my [comment](https://github.com/DataDog/dd-sdk-ios/pull/1413#discussion_r1295625397) below on the same topic 👍

DatadogCore/Sources/Datadog.swift
ncreated: I see the argument on line-readability 👍, but the impacts are more than this IMO. With lack of agreed guideline on this, let's keep it as it is in this PR and discuss this topic in dev chat (I added it to the agenda).

Datadog/IntegrationUnitTests/Public/Datadog+MultipleInstancesIntegrationTests.swift
ncreated: 👍 More coverage will be discussed and added after `RUMM-3514`. For now, we'd like to keep it minimal.

ncreated:
CI is green, Bitrise checks are false.

Screenshot 2023-08-16 at 16 11 08

PR #1421, opened 2023-08-17
RUMM-3218 Inject Telemetry Interface

maxep:
### What and why?

Inject `Telemetry` interface and stop using the global `DD.telemetry`.
By doing so, we enable reporting telemetry events per core with proper correlation to the org.

### How?

Implementations that need to report telemetry events now take the `Telemetry` interface as parameter.

> Special thanks to @ncreated and @maciejburda for already injecting telemetry in SR 🎖️ 

### Review checklist

- [x] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests


DatadogCore/Tests/Datadog/RUM/RUMMonitorTests.swift
```diff
@@ -1332,7 +1332,7 @@ class RUMMonitorTests: XCTestCase {
     }
 
     private func verifyGlobalAttributes(in matchers: [RUMEventMatcher]) {
-        for matcher in matchers {
+        for matcher in matchers.filterTelemetry() {
```
maxep, line 1335: Now that the telemetry is injected, we are getting events during the initialisation of the Features. Previously, we were loosing those because `DD.telemetry` was not yet set.

DatadogLogs/Sources/Feature/MessageReceivers.swift
```diff
@@ -7,61 +7,6 @@
 import Foundation
 import DatadogInternal
 
-/// Creates Logging Feature Configuration.
-/// - Parameters:
-///   - intake: The Logging intake URL.
-///   - logEventMapper: The log event mapper.
-/// - Returns: The Logging feature configuration.
-internal func createLoggingConfiguration(
```
maxep, line 15: Some v1 leftovers 🧹

DatadogCore/Tests/Datadog/Core/Upload/DataUploadWorkerTests.swift
```diff
@@ -410,26 +421,26 @@ class DataUploadWorkerTests: XCTestCase {
...
         worker.cancelSynchronously()
 
         // Then
-        XCTAssertEqual(dd.telemetry.messages.count, 1)
+        XCTAssertEqual(telemetry.messages.count, 1)
 
-        guard case .error(_, let message, _, _) = dd.telemetry.messages.first else {
-            return XCTFail("An error should be send to `DD.telemetry`.")
+        guard case .error(_, let message, _, _) = telemetry.messages.first else {
+            return XCTFail("An error should be send to `telemetry`.")
```
ncreated, line 435: suggestion/ This gets quite repetitive in all tests. Recently I faced it when working on metrics telemetry. Solved by [adding helpers in `TU.TelemetryMocks`](https://github.com/DataDog/dd-sdk-ios/blob/d0d2b5448483768273421327ef0109e1b5728ce5/TestUtilities/Mocks/TelemetryMocks.swift#L83-L98):
```swift
_ = telemetry.messages.firstMetric(...)
_ = telemetry.messages.first.asMetric()
```
WDYT on facilitating other messages in similar or any other way? Otherwise it will be always a pain to add new telemetry assertion - `guard case` is quite inconvenient.

DatadogCore/Tests/Datadog/Core/_InternalProxyTests.swift
```diff
@@ -12,8 +12,10 @@ import TestUtilities
 class _InternalProxyTests: XCTestCase {
     func testWhenTelemetryIsSentThroughProxy_thenItForwardsToDDTelemetry() throws {
```
ncreated, line 13: question/ What is the difference between tests in this file and the ones we change in 
* [DatadogCore/Tests/Datadog/InternalTelemetryTests.swift](https://github.com/DataDog/dd-sdk-ios/pull/1421/files#diff-e07ddd9caba826b1467dd5f6a90924c5f054378cee4795832284db63b33e8f95)?

I think they are so similar that we can merge them into single file. Both work with mocks and seem to assert the same things.

ncreated (dismissed):
It looks good, nice cleanups 👌.

DatadogCore/Tests/Datadog/Core/_InternalProxyTests.swift
maxep: Indeed 👍

ncreated (approved):
🎯

PR #1419, opened 2023-08-17
REPLAY-1936 Session Replay Objective-C interface

maxep:
### What and why?

Expose Session Replay to Objective-C.
Fix #1387

### How?

Add `SessionReplay` objective-c interface to `DatadogObjc` module.

### Review checklist
- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference
- [x] Add CHANGELOG entry for user facing changes

### Custom CI job configuration (optional)
- [x] Run unit tests
- [ ] Run integration tests
- [x] Run smoke tests

PR #1428, opened 2023-08-21
fix: `DatadogObjc` not fully available in `2.1.0`

ncreated:
### What and why?

🐞 Single patch for two problems:
* `DatadogObjc.podspec` not validating in `2.1.0` release, hence being unavailable for CP users;
* `DatadogObjc` module not compiling for `tvOS` for SPM users.

The issue originated in https://github.com/DataDog/dd-sdk-ios/pull/1419 where we added `DatadogSessionReplay` as dependency to `DatadogObjc` in order to provide Obj-c interface for SR. With `DatadogObjc` supporting both `iOS` and `tvOS` platforms and SR being not yet ready for `tvOS`, it introduced two issues:
* **Cocoapods:** `DatadogSessionReplay.podspec` was only declaring `ios` as deployment target, but not `tvos`. In result, the `DatadogObjc.podspec` failed validation and `pod trunk` submission for `2.1.0` spec.
* **SPM:** `DatadogObjc` library stopped compiling for `tvOS` platform as one of its dependencies (`DatadogSessionReplay`) doesn't compile for it.

#### The Fix
Seeking for a solution with minimal impact, we decided to conditionally compile the whole SR module, by adding `#if os(iOS) ... #endif` blocks to every file. It is less invasive than any other explored solution:
   * Differentiating `spec.<platform>.sources = ...` works for Cocoapods, but has no counterpart in `Package.swift`.
   * Using ["wrapper" modules](https://forums.swift.org/t/swift-package-manager-platform-conditional-build/56735) in SPM could be an option, but may surface other impacts and complexities later.
   * Given that we plan to work on `DatadogObjc` modularisation soon, that will solve this problem "by design", none of above solutions was incorporated.

#### Future mitigation:
The issue was not spotted earlier (during development), because [Cocoapods smoke tests](https://github.com/DataDog/dd-sdk-ios/tree/develop/dependency-manager-tests/cocoapods) and [SPM smoke tests](https://github.com/DataDog/dd-sdk-ios/tree/develop/dependency-manager-tests/spm) do not include `DatadogObjc` in their build. This gap is eliminated as part of this PR: https://github.com/DataDog/dd-sdk-ios/pull/1428/commits/1635102712abdfa1f06bb8f390c0c12fe352dca8.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Unreleased
 
+# 2.1.1 / 21-08-2023
+
+- [BUGFIX] `DatadogObjc` pod not available in `2.1.x` release. See [#1428][].
+
+# 2.1.0 / 18-08-2023
+
 - [BUGFIX] Manual trace injection APIs are not available in DatadogTrace. See [#1415][].
 - [BUGFIX] Fix session replay uploads to AP1 site. See [#1418][].
 - [BUGFIX] Allow instantiating custom instance of the SDK after default one. See [#1413][].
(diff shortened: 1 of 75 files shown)


maciejburda (approved):

maciejburda (approved):
LGTM

maxep (approved):

markst:
Thanks!

PR #1433, opened 2023-08-23
RUM-401 Introduce new FeatureBaggage

maxep:
### What and why?

Only allow single data types to be shared on the message-bus and prevent accessing values by keys. This solution enforces defining `Encodable` end-to-end.

### How?

Here a usage example:

From a sender:
```swift

struct MyBaggage: Encodable {
  let value: String
}

let baggage = MyBaggage(value: "test")
try core.send(message: .baggage(key: "test", value: baggage)) 
```

From a receiver:
```swift

struct MyReceiver: FeatureMessageReceiver {
  struct TheirBaggage: Decodable {
    let value: String
  }

    guard let baggage: TheirBaggage = try? message.baggage(forKey: "test") else {
      return false
    }

    // do something with baggage
  }
}
```

The proposed implementation makes the message-bus throwable in case of encoding/decoding failures allowing reporting telemetry errors with more context.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/DatadogCore/Tests/Datadog/Logs/CrashLogReceiverTests.swift
+++ b/DatadogCore/Tests/Datadog/Logs/CrashLogReceiverTests.swift
@@ -21,7 +21,7 @@ class CrashLogReceiverTests: XCTestCase {
         )
 
         // When
-        try core.send(
+        core.send(
             message: .baggage(
                 key: MessageBusSender.MessageKeys.crash,
                 value: MessageBusSender.Crash(
--- a/DatadogCore/Tests/Datadog/Logs/CrashLogReceiverTests.swift
+++ b/DatadogCore/Tests/Datadog/Logs/CrashLogReceiverTests.swift
@@ -21,19 +21,14 @@ class CrashLogReceiverTests: XCTestCase {
         )
 
         // When
-        let sent: LogEvent = .mockRandom()
-
-        core.send(
-            message: .custom(key: LoggingMessageKeys.crash, baggage: [
-                "report": DDCrashReport.mockAny(),
-                "context": CrashContext.mockWith(lastRUMViewEvent: nil)
-            ])
-        )
-
-        core.send(
-            message: .custom(key: LoggingMessageKeys.crash, baggage: [
-                "log": sent
-            ])
+        try core.send(
+            message: .baggage(
+                key: MessageBusSender.MessageKeys.crash,
+                value: MessageBusSender.Crash(
+                    report: DDCrashReport.mockAny(),
+                    context: CrashContext.mockWith(lastRUMViewEvent: nil)
+                )
+            )
         )
 
         // Then
--- a/DatadogCrashReporting/Sources/Integrations/CrashReportSender.swift
+++ b/DatadogCrashReporting/Sources/Integrations/CrashReportSender.swift
@@ -27,6 +27,13 @@ internal struct MessageBusSender: CrashReportSender {
         static let crash = "crash"
     }
 
+    struct Crash: Encodable {
+        /// The crash report.
+        let report: DDCrashReport
+        /// The crash context
+        let context: CrashContext
+    }
+
     /// The core for sending crash report and context.
     ///
     /// It must be a weak reference to avoid retain cycle (the `CrashReportSender` is held by crash reporting
@@ -39,30 +46,29 @@ internal struct MessageBusSender: CrashReportSender {
     ///   - report: The crash report.
     ///   - context: The crash context
     func send(report: DDCrashReport, with context: CrashContext) {
-        guard context.trackingConsent == .granted else {
+        guard let core = core, context.trackingConsent == .granted else {
             return
         }
 
-        sendCrash(
-            baggage: [
-                "report": report,
-                "context": context
-            ]
-        )
-    }
-
-    private func sendCrash(baggage: FeatureBaggage) {
-        core?.send(
-            message: .custom(key: MessageKeys.crash, baggage: baggage),
-            else: {
-                DD.logger.warn(
-            """
-            In order to use Crash Reporting, RUM or Logging feature must be enabled.
-            Make sure `.enableRUM(true)` or `.enableLogging(true)` are configured
-            when initializing Datadog SDK.
-            """
+        do {
+            try core.send(
+                message: .baggage(
+                    key: MessageKeys.crash,
+                    value: Crash(report: report, context: context)
+                ),
+                else: {
+                    DD.logger.warn(
+                """
+                In order to use Crash Reporting, RUM or Logging feature must be enabled.
+                Make sure `.enableRUM(true)` or `.enableLogging(true)` are configured
+                when initializing Datadog SDK.
+                """
+                )
+                }
             )
-            }
-        )
+        } catch {
+            core.telemetry
+                .error("Fails to encode crash", error: error)
+        }
     }
 }
--- a/DatadogCrashReporting/Sources/Integrations/CrashReportSender.swift
+++ b/DatadogCrashReporting/Sources/Integrations/CrashReportSender.swift
@@ -59,8 +59,7 @@ internal struct MessageBusSender: CrashReportSender {
                 DD.logger.warn(
                     """
                     In order to use Crash Reporting, RUM or Logging feature must be enabled.
-                    Make sure `.enableRUM(true)` or `.enableLogging(true)` are configured
-                    when initializing Datadog SDK.
+                    Make sure `RUM` or `Logs` are enabled when initializing Datadog SDK.
                     """
                 )
             }
--- a/DatadogCrashReporting/Sources/Integrations/CrashReportSender.swift
+++ b/DatadogCrashReporting/Sources/Integrations/CrashReportSender.swift
@@ -50,25 +50,20 @@ internal struct MessageBusSender: CrashReportSender {
             return
         }
 
-        do {
-            try core.send(
-                message: .baggage(
-                    key: MessageKeys.crash,
-                    value: Crash(report: report, context: context)
-                ),
-                else: {
-                    DD.logger.warn(
-                """
-                In order to use Crash Reporting, RUM or Logging feature must be enabled.
-                Make sure `.enableRUM(true)` or `.enableLogging(true)` are configured
-                when initializing Datadog SDK.
-                """
+        core.send(
+            message: .baggage(
+                key: MessageKeys.crash,
+                value: Crash(report: report, context: context)
+            ),
+            else: {
+                DD.logger.warn(
+                    """
+                    In order to use Crash Reporting, RUM or Logging feature must be enabled.
+                    Make sure `.enableRUM(true)` or `.enableLogging(true)` are configured
+                    when initializing Datadog SDK.
+                    """
                 )
-                }
-            )
-        } catch {
-            core.telemetry
-                .error("Fails to encode crash", error: error)
-        }
+            }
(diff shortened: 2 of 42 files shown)


DatadogInternal/Sources/MessageBus/NewFeatureBaggage.swift
```diff
@@ -0,0 +1,69 @@
...
+///     }
+///
+///     let item: CartItem = try baggage.decode()
+///
+///     print(item)
+///     // CartItem(name: "Pear", points: 250)
+///
+/// A Feature Baggage does not ensure thread-safety of values that holds references, make
+/// sure that any value can be accessibe from any thread.
+public struct NewFeatureBaggage {
```
ncreated, line 50: Is the `New*` prefix intentional? 🤔💭 Aren't we replacing the previous concept entirely in this PR?

maxep: Indeed, it will be replaced entirely in #1442
This PR applies the new baggage on the message-bus only, the second PR is about the core context.

DatadogCrashReporting/Sources/Integrations/CrashReportSender.swift
```diff
@@ -39,30 +46,29 @@ internal struct MessageBusSender: CrashReportSender {
...
+                """
+                )
+                }
             )
-            }
-        )
+        } catch {
+            core.telemetry
+                .error("Fails to encode crash", error: error)
+        }
```
ncreated, line 72: suggestion/ It feels that throwing with `try core.send(...)` is not a productive choice:
* baggage encoding can fail, so we need `try {}`,
* there is almost never a fallback we could implement in `catch {}`,
* we may always want to know the failure in telemetry (all these failures are critical).

All that together, it would be more productive to offer non-throwing `core.send(message: .baggage(...))` that will report every error to telemetry. To differentiate one failure from the other, the `send(message:)` could accept [`#fileID`](https://developer.apple.com/documentation/swift/fileid()) and `#line` as default parameters. More context about the error could be also inferred from the `baggage.key`.

In this PR, we introduce the same pattern in multiple places - all could be simplified if we go this way. Also, in `RUMSessionScope` we opt-out from capturing errors with `try?` - that would be automatically covered with this proposal. Same for next usages - we will get telemetry for free.

LMKWDYT @maxep

maxep: That's a good point, thanks! It's indeed not very practical to catch the error.
Using `#fileID` and `#line` is a good idea, but I'm not sure how it can be applied since it's [the `FeatureBaggage` method](https://github.com/DataDog/dd-sdk-ios/blob/7981a6d45cebd13a5296ef2ba65dad9190d94b03/DatadogInternal/Sources/MessageBus/FeatureMessage.swift#L37) that throw, not the core. I will do some test!

maxep: So I went another path which is more efficient and more convenient:
The `FeatureBaggage` will delay the encoding to when the message is actually received. Meaning that:
1. We won't encode the value if it's never read
2. We will throw an error only when decoding the value of the baggage

This way, we don't need to `try core.send(message: .baggage(...))`, the `FeatureMessageReceiver` will be responsible for reporting telemetry on baggage decoding failure.

DatadogInternal/Sources/MessageBus/FeatureMessage.swift
```diff
@@ -39,3 +26,30 @@ public enum FeatureMessage {
     /// The core can send telemtry data coming from all Features.
     case telemetry(TelemetryMessage)
 }
+
+extension FeatureMessage {
+    /// Creates a `.baggage` message with thegiven label and `Encodable` type.
```
ncreated, line 31:
```suggestion
    /// Creates a `.baggage` message with given key and `Encodable` value.
```

DatadogInternal/Sources/MessageBus/FeatureMessage.swift
```diff
@@ -39,3 +26,30 @@ public enum FeatureMessage {
...
+    }
+
+    /// Decodes the value of a baggage if the label matches.
+    ///
+    /// - Parameters:
+    ///   - label: The requested baggage label.
+    ///   - type: The expected type of the baggage value.
+    /// - Returns: The decoded baggage value, or nil if the label doesn't match.
+    /// - Throws: A `DecodingError` if decoding fails.
+    public func baggage<Value>(forKey key: String, type: Value.Type = Value.self) throws -> Value? where Value: Decodable {
```
ncreated, line 48: The comment isn't up-to-date with the method signature.

DatadogInternal/Sources/MessageBus/FeatureMessage.swift
```diff
@@ -39,3 +26,30 @@ public enum FeatureMessage {
...
+        guard case let .baggage(messageKey, baggage) = self, messageKey == key else {
+            return nil
```
ncreated, line 50: question/ What does it mean when the program reaches this `else { return nil }`? 
* Is it expected to end up here, hence we decide to ignore it by `return nil`?
* Or is it unexpected, but we return `nil` with lack of other options?

I wonder if we shouldn't make it strict by throwing an error, or assert it with telemetry.

DatadogInternal/Sources/MessageBus/NewFeatureBaggage.swift
```diff
@@ -0,0 +1,85 @@
...
+    ///
+    /// - Returns: The encoded baggage value.
+    public func encode() throws -> Any? {
+        // lazily encode to save from encoding
+        // if the value is never decoded.
+        if let rawValue = rawValue {
+            return rawValue
+        }
+        rawValue = try _encode()
+        return rawValue
```
ncreated, line 73: ⚠️ question/ The `rawValue` is not synchronised and `NewFeatureBaggage` is a class, so what if multiple receivers `encode()` simultaneously? I don't think the message-bus design guarantees this to not happen - baggage can be captured, escaped and processed on any thread.

DatadogInternal/Tests/MessageBus/NewFeatureBaggageTests.swift
```diff
@@ -0,0 +1,30 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import XCTest
+@testable import DatadogInternal
+
+class NewFeatureBaggageTests: XCTestCase {
```
ncreated, line 10: We're only testing the happy path scenario in this suite.

ncreated (changes requested):
The idea looks good and indeed brings more safety to the message-bus communication 👍. I left few questions and some minor feedbacks.

As general feedback, I struggle with clarity of tests that we define for both ends of the message-bus channel. They seem to be very fragmented, and could be better surfacing the important behaviours. Not a blocking thing for this PR, but rather a point to consider long term.

DatadogCore/Tests/Datadog/Logs/CrashLogReceiverTests.swift
```diff
@@ -21,19 +21,14 @@ class CrashLogReceiverTests: XCTestCase {
...
-            message: .custom(key: LoggingMessageKeys.crash, baggage: [
-                "report": DDCrashReport.mockAny(),
-                "context": CrashContext.mockWith(lastRUMViewEvent: nil)
-            ])
-        )
-
-        core.send(
-            message: .custom(key: LoggingMessageKeys.crash, baggage: [
-                "log": sent
-            ])
```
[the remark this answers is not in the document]
maxep: I did struggle with it as well TBH! The receiver is an actual instance of `CrashLogReceiver` with mocked init parameters (sample rate, date provider, ...). It testing that the `CrashLogReceiver ` is capable of reporting a `RUMCrashEvent` when receiving a crash message. But I could see the point of sending a second message, so I removed it...

DatadogInternal/Sources/MessageBus/FeatureMessage.swift
maxep: `nil` is expected here, this is a convenient method that message receivers can use to check for a specific baggage message. Maybe it doesn't read well 🤔 I could be like the following:
```swift
/// Returns the baggage if the key matches the message.
///
/// - Parameters:
///   - key: The requested baggage key.
public func baggage(forKey key: String) -> NewFeatureBaggage? {
    guard case let .baggage(messageKey, baggage) = self, messageKey == key else {
        return nil
    }

    return baggage
}
```

Then a receiver will have to do:
```swift
    do {
        guard let baggage = try message.baggage(forKey: MessageKeys.crash) else {
            return false
        }

        let crash: Crash = try bagagge.decode()
        return send(report: crash.report, with: crash.context, to: core)
    } catch {
        core.telemetry
            .error("Fails to decode crash from RUM", error: error)
    }

    return false
}
```

This might read better, WDYT?

DatadogInternal/Sources/MessageBus/NewFeatureBaggage.swift
maxep: yep that's is true!! I assumed the baggage would be read on the message-bus queue only, but that is not guaranteed by the api, so the `rawValue` should be locked.

DatadogInternal/Sources/MessageBus/FeatureMessage.swift
ncreated: Thanks, I can see it now. So the `nil` is unavoidable, but it is rather a cost of `self` being an `enum`, rather than a flaw in the logic we add. Let's keep what we have, this is enough convenience 👌.

DatadogInternal/Tests/MessageBus/NewFeatureBaggageTests.swift
maxep: Indeed, I have added tests for encoding and decoding failures as well as a test for concurrency 👍

maxep:
> As general feedback, I struggle with clarity of tests that we define for both ends of the message-bus channel. They seem to be very fragmented, and could be better surfacing the important behaviours. Not a blocking thing for this PR, but rather a point to consider long term.

@ncreated I do agree, partially due to the fact that `DatadogCoreTests` is still un umbrella for core unit tests AND Integration Test Units. It will be clearer with dedicated Integration Units but it won't avoid fragmentation, it will be split in 2 layers:
- Feature unit tests: test sending message with dummy receiver; test receiving dummy messages
- Integration unit tests: test message-bus end-to-end

DatadogInternal/Sources/MessageBus/FeatureMessage.swift
```diff
@@ -8,24 +8,11 @@ import Foundation
...
-    )
-
     /// A custom message with generic encodable
     /// attributes.
-    case custom(
+    case baggage(
         key: String,
-        baggage: FeatureBaggage
+        baggage: NewFeatureBaggage
     )
```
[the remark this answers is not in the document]
maxep: That's a good point however using `key` will align with the `baggages: [String: FeatureBaggage]` property from the core context (#1442). So either on the message-bus and in the core context, a FeatureBaggage will always be associated to a key. Does it make sense?

ncreated (approved):
Nice improvement 👌

PR #1435, opened 2023-08-24
Configuration refactor proposal

maciejburda:
It's a draft solution for configuration refactor

Benefits:
- clean mocking
- explicit configuration (no lazy computed vars)
- allows to easily use NOPDatadogCore in customer tests

Tradeoffs:
- configuration can produce `nil` if requirements are met, but we provide error logging.
- there's extra guard on initialize function that informs about creating NOP instance.
- objc wrapper needs some love

### DIFF
--- a/DatadogObjc/Sources/DatadogConfiguration+objc.swift
+++ b/DatadogObjc/Sources/DatadogConfiguration+objc.swift
@@ -237,6 +237,6 @@ public class DDConfiguration: NSObject {
     ///             To filter events on different environments (e.g. "staging" or "production").
     @objc
     public init(clientToken: String, env: String) {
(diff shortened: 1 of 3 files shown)


DatadogObjc/Sources/DatadogConfiguration+objc.swift
```diff
@@ -237,6 +237,6 @@ public class DDConfiguration: NSObject {
     ///             To filter events on different environments (e.g. "staging" or "production").
     @objc
     public init(clientToken: String, env: String) {
-        sdkConfiguration = .init(clientToken: clientToken, env: env)
+        sdkConfiguration = .init(clientToken: clientToken, env: env)!
```
maciejburda, line 240: this would need to be updated

PR #1437, opened 2023-08-24
Make pending crash report completion escaping

erichoracek:
This enables the completion closure to be called asynchronously

### What and why?

This closure is currently non-escaping, which means it must be called before the method returns, it cannot be called asynchronously. However, the documentation states that "any thread (can be used) for executing the  `completion`."

### How?

Added an escaping annotation.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 # Unreleased
+- [IMPROVEMENT] Allow readPendingCrashReport completion closure to escape.
 
 # 2.1.1 / 22-08-2023
 
--- a/DatadogCrashReporting/Sources/CrashReportingPlugin.swift
+++ b/DatadogCrashReporting/Sources/CrashReportingPlugin.swift
@@ -181,7 +181,7 @@ internal protocol CrashReportingPlugin: AnyObject {
     ///
     /// The SDK calls this method on a background thread. The implementation is free to choice any thread
     /// for executing the  `completion`.
-    func readPendingCrashReport(completion: (DDCrashReport?) -> Bool)
+    func readPendingCrashReport(completion: @escaping (DDCrashReport?) -> Bool)
 
     /// Injects custom data for describing the application state in the crash report.
     /// This data will be attached to produced crash report and will be available in `DDCrashReport`.


ncreated (commented):
@erichoracek Thanks 🙂, the change looks fine IMO. However, I wonder if this will bring anything for you. Please note that this API is `internal` in (current) `2.x`. It used to be `public` [in `1.x` of the SDK](https://github.com/DataDog/dd-sdk-ios/blob/21b62371bdbd40747ef1c3b47ea0f2fd6f154eb1/Sources/Datadog/CrashReporting/DDCrashReportingPluginType.swift#L177-L185).

If your intent was to enable some public capability with this PR, let us know what you need, so maybe we can help.

erichoracek:
Hi @ncreated, we are using this API to ingest some abbreviated crash report information (no stack traces, just type/name) from another crash reporting SDK (Sentry) in the 1.0 version of the SDK. To do this, we had our own type conforming to `CrashReportingPlugin`.

ncreated:
> Hi @ncreated, we are using this API to ingest some abbreviated crash report information (no stack traces, just type/name) from another crash reporting SDK (Sentry) in the 1.0 version of the SDK. To do this, we had our own type conforming to CrashReportingPlugin.

@erichoracek Thanks for more context. So would it still work for you if this API is now `internal`? Again, it used to be `public` in `1.x` of the SDK, but it no longer is.

erichoracek:
@ncreated Ideally this API would be public so we could continue to keep crash forwarding logic—do you have context on why it was made internal in 2.0?

ncreated:
> @ncreated Ideally this API would be public so we could continue to keep crash forwarding logic—do you have context on why it was made internal in 2.0?

The 2.0 brings no coupling between different products (Crash Reporting, RUM, Logs etc.). We made `CrashReportingPlugin` internal and the communication between CR and other modules was standardised to use _message bus_ available on _SDK core_.

In 2.0 you should rather look into [`CrashReportSender`](https://github.com/DataDog/dd-sdk-ios/blob/2.2.1/DatadogCrashReporting/Sources/Integrations/CrashReportSender.swift) interface to figure out how it interacts with `DatadogCoreProtocol` (_SDK core_) by sending crash message (through _message bus_). While it is possible to implement your own `DatadogFeature` in 2.x SDK and register it to work with features that we implemented, we're not yet providing guidelines on how to do it. Best you can do is perhaps figure out how [`DatadogCrashReporting`](https://github.com/DataDog/dd-sdk-ios/blob/2.2.1/DatadogCrashReporting/Sources/CrashReportingFeature.swift) implements [`DatadogFeature`](https://github.com/DataDog/dd-sdk-ios/blob/2.2.1/DatadogInternal/Sources/DatadogFeature.swift) to interact with [`DatadogCoreProtocol`](https://github.com/DataDog/dd-sdk-ios/blob/2.2.1/DatadogInternal/Sources/DatadogCoreProtocol.swift) and send crash information to Datadog.

Let me know if above context is anything you can utilise. We can also consider feature request to provide you with what you need - for that we must understand your case more. To send custom error information you can use Logs or RUM without utilizing Crash Reporting. With RUM you can also provide extra context to Crash Reports captured by our SDK.

erichoracek:
Gotcha, that's super helpful. Thank you! I'll investigate that approach when we upgrade to 2.0

PR #1442, opened 2023-08-28
RUM-401 Update core context with new FeatureBaggage

maxep:
### What and why?

Apply the new `FeatureBaggage` introduced in #1433 to the `DatadogContext`. The new baggages brings more type-safety in the context.

### How?

The `baggages` property replaces the `featuresAttributes` in the context.

The `DatadogCoreProtocol.update(feature:, attributes:)` has been removed because a baggage now represents a single type, and not a dictionary of types. This changes the structure of the context for Session Replay:
```swift
// before
context.featuresAttributes = [
    "session-replay" = [
        "has_replay": true,
        "records_count_by_view_id": [:]
    ]
]

// after
context.baggages = [
    "sr_has_replay-replay" = true,
    "sr_records_count_by_view_id": [:]
]
```

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


DatadogLogs/Sources/Feature/Baggages.swift
```diff
@@ -0,0 +1,75 @@
...
+
+    var internalAttributes: [String: String?] {
+        [
+            CodingKeys.traceID.rawValue: traceID,
+            CodingKeys.spanID.rawValue: spanID
+        ]
+    }
+}
+
+/// The RUM context received from `DatadogCore`.
```
ganeshnj, line 45: ❓ 

Just wondering why do we have 3 `RUMContext` types? Can they be shared so we don't have to maintain it at 3 different places. 

Interestingly, two are same type but has `server_time_offset`.

```swift
        case applicationID = "application.id"
        case sessionID = "session.id"
        case viewID = "view.id"
        case userActionID = "user_action.id"
    }
```

```swift
        case applicationID = "application.id"
        case sessionID = "session.id"
        case viewID = "view.id"
        case viewServerTimeOffset = "server_time_offset"
    }
```

```swift
        case applicationID = "application.id"
        case sessionID = "session.id"
        case viewID = "view.id"
        case userActionID = "user_action.id"
    }
```

One potential way could be the same way we forward types from `Internal` package to `RUM` and `Trace` packages. This is low effort and high value solution until we figure out something that is more elegant solution.

ganeshnj (changes requested):

DatadogLogs/Sources/Feature/Baggages.swift
maxep: That's what we are avoiding with `FeatureBaggage`, we are keeping the `Internal` module feature-agnostic. The `FeatureBaggage` allows to share encodable objects on the message-bus and in the core context, it's a loose contract between features. Types are defined end-to-end and objects are decoded to match the expected definition. 

We have some definitions that are part of the `Internal` module that comply with `PassthroughAnyCodable`, those types can be transmitted as is, without encoding/decoding, as they are known by multiple parties. But the `RUMContext` is not one of them. I believe this is the elegant solution you are referring to?

ganeshnj: IMO, a package with just APIs makes more sense but that would make our package count x2 (elegant solution)
eg, TraceAPI and TraceSDK (names are for illustrations only).

I don't see any benefit of loose coupling here. Correct me if I'm wrong here. Our packages don't work with mixed versions like `Rum version X` and `Trace version Y`. They have to be same because that's how our repo structured. With current "loose coupling" we are intentionally increasing our work of maintaining 2/3 or 4 types. With sharing types, they are still loose coupled - there is no direct dependency but provide strongly typed nature.

maxep: Please have a look to #1112 and #1113 for more context on this solution. The benefit is to avoid feature-specific logic and type definitions in the Internal module. The goal is to have a clear separation of concerns between modules and no shared responsibilities. I don't see the increase in maintenance, especially for a type as simple as the RUM context. Moreover, we are transmitting way more complex types, e.g RUM View with Crash Reporting, where this solution with type-erasure drastically simplify data transmission.

The exception to the rule is the `NetworkInstrumentation`, it was decided to share it in `DatadogInternal` for simplicity of the migration, but we should consider extracting it to its dedicated module.

DatadogCore/Tests/Datadog/FeaturesIntegration/TracingWithLoggingIntegrationTests.swift
```diff
@@ -130,8 +130,8 @@ class TracingWithLoggingIntegrationTests: XCTestCase {
         DDAssertJSONEqual(
             AnyEncodable(log.attributes.internalAttributes),
             AnyEncodable([
-                DatadogTracer.Attributes.spanID: "2",
-                DatadogTracer.Attributes.traceID: "1"
+                "dd.trace_id": "1",
+                "dd.span_id": "2"
```
maciejburda, line 134: nit - those two assertions use different keys. Maybe worth having consistent keys

DatadogInternal/Tests/MessageBus/FeatureBaggageTests.swift
```diff
@@ -5,97 +5,83 @@
...
-        XCTAssertEqual(attributes["string"], "test")
+    func testEncodeDecode() throws {
+        let pear = GroceryProduct.mockRandom()
+        let baggage = FeatureBaggage(pear)
+        let item: CartItem = try baggage.decode()
 
-        attributes["int"] = 2
-        XCTAssertEqual(attributes["int"], 2)
+        XCTAssertEqual(pear.name, item.name)
+        XCTAssertEqual(pear.points, item.points)
```
maciejburda, line 37: Worth testing the other way round and seeing if description is empty

DatadogLogs/Sources/Feature/Baggages.swift
```diff
@@ -0,0 +1,75 @@
...
+    /// Current RUM application ID - standard UUID string, lowecased.
+    let applicationID: String
+    /// Current RUM session ID - standard UUID string, lowecased.
+    let sessionID: String
+    /// Current RUM view ID - standard UUID string, lowecased. It can be empty when view is being loaded.
+    let viewID: String?
+    /// The ID of current RUM action (standard UUID `String`, lowercased).
+    let userActionID: String?
+
+    var internalAttributes: [String: String] {
```
maciejburda, line 65: could be lazy

also, I'm kinda on a fence with model preparing this dictionary. Maybe it could be externalised

Where do we use this?

maciejburda (approved):
I like flattening the structure. It actually looks cleaner.

The only problem I see is that it's now slightly easier to have key conflict (that two features use the same key to propagate data).

Gave some minor suggestions. LGTM 🙌

DatadogInternal/Tests/MessageBus/FeatureBaggageTests.swift
maxep: `item` does not have a `description` property, but I can test an optional property in `CartItem` indeed 👍

maciejburda (approved):
Good improvement!

Tested SR from the branch and eveything seems to work fine 👍

DatadogLogs/Sources/Feature/Baggages.swift
ganeshnj: For now I think it is fine but we should revisit this soon. We are going to have more and more such cases.

Another topic for dev-chat.

ganeshnj (approved):

PR #1455, opened 2023-09-06
RUM-917 Fix unnecessary type mismatch fix

maciejburda:
### What and why?

Refactors id generation for views in session replay. It prevents from reaching type mismatch for the same view but recorded by different recorders and forcing full snapshot change.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

### DIFF
--- a/DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeIDGenerator.swift
+++ b/DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeIDGenerator.swift
@@ -19,6 +19,7 @@ internal typealias NodeID = Int64
 ///
 /// **Note**: All `NodeIDGenerator` APIs must be called on the main thread.
 internal final class NodeIDGenerator {
+    /// Dictionary of generated node IDs for given `UIView` /  `NodeRecorder` pairs.
     private var ids = [String: [NodeID]]()
     /// Upper limit for generated IDs.
     /// After `currentID` reaches this limit, it will start from `0`.
@@ -36,7 +37,7 @@ internal final class NodeIDGenerator {
     /// - Parameter nodeRecorder: the `NodeRecorder` object
     /// - Returns: the `NodeID` of queried instance
     func nodeID(view: UIView, nodeRecorder: NodeRecorder) -> NodeID {
-        let key = nodeRecorder.identifier.uuidString + "-\(view.hash)"
+        let key = nodeRecorder.identifier.uuidString + String(view.hash)
         if let nodeID = ids[key]?.first {
             return nodeID
         } else {
@@ -52,7 +53,7 @@ internal final class NodeIDGenerator {
     /// - Parameter nodeRecorder: the `NodeRecorder` object
     /// - Returns: an array with given number of `NodeID` values
     func nodeIDs(_ size: Int, view: UIView, nodeRecorder: NodeRecorder) -> [NodeID] {
-        let key = nodeRecorder.identifier.uuidString + "-\(view.hash)-\(size)"
+        let key = nodeRecorder.identifier.uuidString + String(view.hash) + String(size)
         if let nodeIDs = ids[key] {
             return nodeIDs
         } else {
--- a/DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeIDGenerator.swift
+++ b/DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeIDGenerator.swift
@@ -53,19 +53,14 @@ internal final class NodeIDGenerator {
     /// - Parameter nodeRecorder: the `NodeRecorder` object
     /// - Returns: an array with given number of `NodeID` values
     func nodeIDs(_ size: Int, view: UIView, nodeRecorder: NodeRecorder) -> [NodeID] {
-        var result = [NodeID]()
-        let key = nodeRecorder.identifier.uuidString + "-\(view.hash)"
+        let key = nodeRecorder.identifier.uuidString + "-\(view.hash)-\(size)"
         if let nodeIDs = ids[key] {
-            result += nodeIDs
+            return nodeIDs
+        } else {
+            let nodeIDs = (0..<size).map { _ in getNextID() }
+            ids[key] = nodeIDs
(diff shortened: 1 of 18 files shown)


DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeIDGenerator.swift
```diff
@@ -32,28 +33,32 @@ internal final class NodeIDGenerator {
 
     /// Returns single `NodeID` for given instance of `UIView`.
     /// - Parameter view: the `UIView` object
+    /// - Parameter nodeRecorder: the `NodeRecorder` object
     /// - Returns: the `NodeID` of queried instance
-    func nodeID(for view: UIView) -> NodeID {
-        if let currentID = view.nodeID {
-            return currentID
+    func nodeID(view: UIView, nodeRecorder: NodeRecorder) -> NodeID {
+        let key = nodeRecorder.identifier.uuidString + "-\(view.hash)"
```
ncreated, line 39: How does `view.hash` work exactly? I can only find its definition on [`NSObjectProtocol.hash`](https://developer.apple.com/documentation/objectivec/nsobjectprotocol/1418859-hash) but nothing in there assures me that we can use it reliably for `UIView` tagging. In other words, are we sure the `hash` will be stable for `UIView` instances reused among view controllers or cells in recyclable lists?

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeIDGenerator.swift
```diff
@@ -63,21 +68,4 @@ internal final class NodeIDGenerator {
...
-    var nodeID: NodeID? {
-        set { objc_setAssociatedObject(self, &associatedNodeIDKey, newValue, objc_AssociationPolicy.OBJC_ASSOCIATION_RETAIN_NONATOMIC) }
-        get { objc_getAssociatedObject(self, &associatedNodeIDKey) as? NodeID }
-    }
-
-    var nodeIDs: [NodeID]? {
-        set { objc_setAssociatedObject(self, &associatedNodeIDsKey, newValue, objc_AssociationPolicy.OBJC_ASSOCIATION_RETAIN_NONATOMIC) }
-        get { objc_getAssociatedObject(self, &associatedNodeIDsKey) as? [NodeID] }
-    }
-}
```
ncreated, line 82: Why this removal?

DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeIDGeneratorTests.swift
```diff
@@ -11,6 +11,7 @@ import UIKit
 
 class NodeIDGeneratorTests: XCTestCase {
```
ncreated, line 12: There are no tests for the new behaviour we added, including the problem we solve: querying the same view by two different node recorders.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeIDGenerator.swift
```diff
@@ -32,28 +33,32 @@ internal final class NodeIDGenerator {
...
     /// Returns multiple `NodeIDs` for given instance of `UIView`.
     /// - Parameter size: the number of IDs
     /// - Parameter view: the `UIView` object
+    /// - Parameter nodeRecorder: the `NodeRecorder` object
     /// - Returns: an array with given number of `NodeID` values
-    func nodeIDs(_ size: Int, for view: UIView) -> [NodeID] {
-        if let currentIDs = view.nodeIDs, currentIDs.count == size {
-            return currentIDs
+    func nodeIDs(_ size: Int, view: UIView, nodeRecorder: NodeRecorder) -> [NodeID] {
+        let key = nodeRecorder.identifier.uuidString + "-\(view.hash)-\(size)"
```
ncreated, line 55: String concatenation and interpolation are known to be extremely slow. What is the overhead we're adding with this change? Keep in mind this will be called for each view in view-tree, every time we capture a `record()`.

ncreated (changes requested):

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeIDGenerator.swift
maciejburda: Since now key is derived from view and recorder it made more sense to me to keep dictionary instead of using associated object.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeIDGenerator.swift
maciejburda: Hash is an address in memory. We already use it for image recording.

It's been very much consistent when I was running different experiments.

maciejburda:
@ncreated Refactored accordingly to our last discussion. I like the current solution more.

Please let me know your thoughts.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeIDGenerator.swift
```diff
@@ -33,12 +33,16 @@ internal final class NodeIDGenerator {
     /// Returns single `NodeID` for given instance of `UIView`.
     /// - Parameter view: the `UIView` object
     /// - Returns: the `NodeID` of queried instance
-    func nodeID(for view: UIView) -> NodeID {
-        if let currentID = view.nodeID {
+    func nodeID(view: UIView, nodeRecorder: NodeRecorder) -> NodeID {
```
ganeshnj, line 36: looks good, you may want to update the docs with new param.

same for below method.

ganeshnj (dismissed):

ncreated (dismissed):

ganeshnj (approved):

ncreated (approved):

PR #1465, opened 2023-09-12
RUM-419 Report Application Build Number

### LINKED ISSUE #1026: Version / Build number together with App version
Good afternoon!

Currently some fields are added automatically, like this `browser` / `agent` dictionary or `version` field.

Would it be possible to also add `versionNumber` (Let's say app is `9.50.0 (1475)` - so `1475`) or should we go our own way and add that field manually the same we added `thermalState` for example?

It's possible to have logs from multiple builds of the same version so this field would be super helpful.

Screenshot 2022-10-19 at 14 40 44
Screenshot 2022-10-19 at 14 42 03

maxep:
### What and why?

Send `build_number` field as part of RUM and Logs.

fix #1026

### Review checklist

- [x] Run unit tests
- [x] Run integration tests
- [ ] Run smoke tests


xgouchet (approved):

DatadogCore/Sources/Datadog.swift
```diff
@@ -345,6 +345,9 @@ public struct Datadog {
             ?? configuration.bundle.object(forInfoDictionaryKey: "CFBundleVersion") as? String
             ?? "0.0.0"
 
+        let applicationBuildNumber = configuration.bundle.object(forInfoDictionaryKey: "CFBundleVersion") as? String
+            ?? "0"
```
ganeshnj, line 349: ❓ 
I don't recall exactly (it has been long time since I published an app) but can this be different than App Store build number?

If yes, then I expect build number (as well as application version) part of configuration (plain strings). 

So the final resolutions looks like
1. try from configuration string value
2. try from bundle
3. null (either missing from the payload or explicitly marked as `null` json value.)

❓ this is this a required field?

ganeshnj (changes requested):

DatadogCore/Sources/Datadog.swift
maxep: The `build_version` is a required field, same for the `version` field. We have not exposed an application version parameter in the core config, and I believe it should be the same for the build number.

They are both corresponding to values in the app store, but since we are getting them from the `Bundle` with optional casting to `String`, we have a fallback value that won't be reached in production.

ncreated: > I don't recall exactly (it has been long time since I published an app) but can this be different than App Store build number?

I explained our rationale [here](https://github.com/DataDog/dd-sdk-ios/issues/1026#issuecomment-1288739533). In a nutshell:
* `CFBundleShortVersionString` = [`MARKETING_VERSION`](https://developer.apple.com/documentation/xcode/build-settings-reference#Marketing-Version) = what users see in App Store
* `CFBundleVersion` = [`CURRENT_PROJECT_VERSION`](https://developer.apple.com/documentation/xcode/build-settings-reference#Current-Project-Version) = what devs see in App Store

IntegrationTests/IntegrationTests.xcodeproj/xcshareddata/xcschemes/Runner iOS.xcscheme
```diff
@@ -24,8 +24,8 @@
    </BuildAction>
    <TestAction
       buildConfiguration = "Release"
-      selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
-      selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
+      selectedDebuggerIdentifier = ""
+      selectedLauncherIdentifier = "Xcode.IDEFoundation.Launcher.PosixSpawn"
```
ncreated, line 28: q/ What is this change about? Is it intentional?

ncreated (dismissed):
Looks good to me 👍. Worth linking: https://github.com/DataDog/dd-sdk-ios/issues/1026

IntegrationTests/IntegrationTests.xcodeproj/xcshareddata/xcschemes/Runner iOS.xcscheme
maxep: not intentional, thanks 🙏

ganeshnj (approved):

PR #1494, opened 2023-09-26
RUM-1098 ci: update stack to Xcode 14.3

ganeshnj:
### What and why?

We want to build with Xcode 14.1 which is minimum version of supported Xcode for AppStore.

### How?

- Replace xcversion with xcodes.
- fix test cases

FAQs 
**Why not 14.1**
14.1 fails the CI for some unknown reason across the board, my hunch is due to CI visibility SDK but not confirmed.
failure
```
This Step failed, because it has not sent any output for 15m.
```

**Why not 14.2**
the only benefit that we get from 14.2 is to be able to install iOS 12 simulator but because bitrise runs this on Ventura, the Ventura version doesn't support iOS 12.4 simulator. Bummer.
```
The iOS 12.4 simulator runtime is not supported on hosts after macOS 12.99.0.
```
 **Why 14.3**
Our rest of the workflows run on 14.3 and it makes sense to match it if not below.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogCore/Tests/Datadog/Core/Persistence/DataBlockTests.swift
+++ b/DatadogCore/Tests/Datadog/Core/Persistence/DataBlockTests.swift
@@ -163,16 +163,16 @@ class DataBlockTests: XCTestCase {
         let stream = InputStream(url: url)!
         let reader = DataBlockReader(input: stream)
 
-        do {
-            _ = try reader.next()
-            XCTFail("Expected error to be thrown")
-        } catch DataBlockError.readOperationFailed(_, let error) {
-            XCTAssertEqual(
-                (error as? NSError)?.localizedDescription,
-                "The operation couldn’t be completed. No such file or directory"
-            )
-        } catch {
-            XCTFail("Unexpected error: \(error)")
+        XCTAssertThrowsError(try reader.next()) { error in
+            guard case DataBlockError.readOperationFailed(streamStatus: _, streamError: let streamError) = error else {
+                return XCTFail("Unexpected error: \(error)")
+            }
+
+            guard let streamError = streamError else {
+                return XCTFail("Expected stream error")
+            }
+
+            XCTAssertTrue(streamError.localizedDescription.contains("No such file or directory"))
         }
     }
 }
--- a/tools/nightly-unit-tests/generate_bitrise_yml.py
+++ b/tools/nightly-unit-tests/generate_bitrise_yml.py
@@ -25,11 +25,11 @@ from src.semver import Version
 
 def get_environment() -> (Simulators, Runtimes, Devices):
     """
-    Uses `xcversion simulators` and `xcrun simctl` CLIs to load available
+    Uses `xcodes runtimes` and `xcrun simctl` CLIs to load available
     simulators, runtimes and devices.
     """
     simulators = Simulators(
-        xcversion_simulators_output=shell_output('xcversion simulators')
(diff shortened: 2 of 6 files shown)


DatadogCore/Tests/Datadog/Core/Persistence/DataBlockTests.swift
```diff
@@ -163,16 +163,16 @@ class DataBlockTests: XCTestCase {
...
+        XCTAssertThrowsError(try reader.next()) { error in
+            guard case DataBlockError.readOperationFailed(streamStatus: _, streamError: let streamError) = error else {
+                return XCTFail("Unexpected error: \(error)")
+            }
+
+            guard let streamError = streamError else {
+                return XCTFail("Expected stream error")
+            }
+
+            XCTAssertTrue(streamError.localizedDescription.contains("No such file or directory"))
```
maxep, line 175: /suggestion alternative from the release/2.3.0 branch: 1002e348e20ab4a4b5b1a6899f0d4aa6f8ae9e9a

tools/nightly-unit-tests/src/utils.py
```diff
@@ -16,22 +16,24 @@ def shell_output(command: str):
...
             - STDOUT: {process.stdout if process.stdout != '' else '""'}
             - STDERR: {process.stderr if process.stderr != '' else '""'}
             '''
-        )
+    print(output)
+
+    if process.returncode == 0:
+        return process.stdout
+    else:
+        raise Exception(output)
```
maxep, line 36: /question Those prints were for debugging?

maxep (approved):

tools/nightly-unit-tests/src/utils.py
ganeshnj: not just debugging but for real logging because we want our CI builds as much as debuggable possible.

DatadogCore/Tests/Datadog/Core/Persistence/DataBlockTests.swift
ganeshnj: I prefer this way because it doesn't use do-try-catch rather uses prebuilt construct `XCTAssertThrowsError`, just a idiomatic preference, I wouldn't fight for it.

tools/nightly-unit-tests/generate_bitrise_yml.py
```diff
@@ -76,9 +76,8 @@ def generate_bitrise_yml(test_plan: TestPlan, dry_run: bool):
     else:
         for simulator in missing_simulators:
             start = time.time()
-            print(f' → xcversion simulators --install="{simulator.os_name} {simulator.os_version}"')
             if not dry_run:
-                print(shell_output(f'xcversion simulators --install="{simulator.os_name} {simulator.os_version}"'))
+                print(shell_output(f'sudo xcodes runtimes install "{simulator.os_name} {simulator.os_version}"'))
```
ncreated, line 80: question/ No problem with `sudo` on CI 🤔? Is `sudo` required?

ncreated (approved):
🎯 neat

tools/nightly-unit-tests/generate_bitrise_yml.py
ganeshnj: yes, sudo is required. I was surprised too how did it work in bitrise but apparently it worked.

PR #1498, opened 2023-09-27
RUM-1197 fix: Passthrough Decoding

maxep:
### What and why?

Transmitted a `PassthroughAnyCodable` property on the message-bus is failing when decoding it to `AnyDecodable`.
With the new `FeatureBaggage` introduced in #1442, the `RUMSessionState` decoding was failing on crash-reporting side.

> This is not resolving transmitting `nil` in core context.

### How?

Support decoding a `PassthroughAnyCodable` value in `AnyDecodable`.

### Review checklist

- [x] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogInternal/Sources/Codable/AnyDecodable.swift
+++ b/DatadogInternal/Sources/Codable/AnyDecodable.swift
@@ -93,7 +93,7 @@ extension _AnyDecodable {
             self.init(double)
         } else if let string = try? container.decode(String.self) {
             self.init(string)
-        } else if let passthrough = try? container.passthrough() {
+        } else if let passthrough = container.passthrough() {
             self.init(passthrough)
         } else if let array = try? container.decode([AnyDecodable].self) {
             self.init(array.map { $0.value })
--- a/DatadogInternal/Sources/Codable/AnyDecodable.swift
+++ b/DatadogInternal/Sources/Codable/AnyDecodable.swift
@@ -93,6 +93,8 @@ extension _AnyDecodable {
             self.init(double)
         } else if let string = try? container.decode(String.self) {
             self.init(string)
+        } else if let container = container as? PassthroughValueDecodingContainer, let value = try? container.decodePassthrough() {
+            self.init(value)
         } else if let array = try? container.decode([AnyDecodable].self) {
             self.init(array.map { $0.value })
         } else if let dictionary = try? container.decode([String: AnyDecodable].self) {
--- a/DatadogInternal/Sources/Codable/AnyDecodable.swift
+++ b/DatadogInternal/Sources/Codable/AnyDecodable.swift
@@ -93,8 +93,8 @@ extension _AnyDecodable {
             self.init(double)
         } else if let string = try? container.decode(String.self) {
             self.init(string)
-        } else if let container = container as? PassthroughValueDecodingContainer, let value = try? container.decodePassthrough() {
-            self.init(value)
+        } else if let passthrough = try? container.passthrough() {
+            self.init(passthrough)
         } else if let array = try? container.decode([AnyDecodable].self) {
             self.init(array.map { $0.value })
         } else if let dictionary = try? container.decode([String: AnyDecodable].self) {
--- a/DatadogInternal/Sources/Codable/AnyDecoder.swift
+++ b/DatadogInternal/Sources/Codable/AnyDecoder.swift
@@ -864,18 +864,11 @@ extension SingleValueDecodingContainer {
     ///
     /// - returns: A passthrough value if any.
     /// - throws: `DecodingError.typeMismatch` if the value was not passthrough.
-    func passthrough() throws -> PassthroughAnyCodable {
+    func passthrough() -> PassthroughAnyCodable? {
         guard let container = self as? _AnyDecoder.SingleValueContainer else {
-            throw DecodingError.typeMismatch(PassthroughAnyCodable.self, DecodingError.Context(
-                codingPath: codingPath,
-                debugDescription: "Single Value Container cannot decode passthrough value."
-            ))
+            return nil
         }
 
-        guard let value = container.value as? PassthroughAnyCodable else {
-            throw DecodingError.typeMismatch(PassthroughAnyCodable.self, in: container)
-        }
-
-        return value
+        return container.value as? PassthroughAnyCodable
     }
 }
--- a/DatadogInternal/Sources/Codable/AnyDecoder.swift
+++ b/DatadogInternal/Sources/Codable/AnyDecoder.swift
@@ -598,7 +598,7 @@ private class _AnyDecoder: Decoder {
 
     /// A container that can support the storage and direct decoding of a single
     /// nonkeyed value.
-    struct SingleValueContainer: SingleValueDecodingContainer, PassthroughValueDecodingContainer {
+    struct SingleValueContainer: SingleValueDecodingContainer {
         /// The path of coding keys taken to get to this point in encoding.
         let codingPath: [CodingKey]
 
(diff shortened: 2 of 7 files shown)


DatadogInternal/Sources/Codable/AnyDecodable.swift
```diff
@@ -93,6 +93,8 @@ extension _AnyDecodable {
             self.init(double)
         } else if let string = try? container.decode(String.self) {
             self.init(string)
+        } else if let container = container as? PassthroughValueDecodingContainer, let value = try? container.decodePassthrough() {
+            self.init(value)
```
ganeshnj, line 97: introducing non-primitive types here I think we are trying to patch a problem that needs to be fixed somewhere else. 

Can we not pass `PassthroughValueDecodingContainer` between boundaries?

maxep:
We have an issue in webview tracking integration tests, not related to this change but rather an issue with data sent by browser sdk!

DatadogInternal/Sources/Codable/AnyDecodable.swift
ganeshnj: IMO, 
```
 } else if let dictionary = try? container.decode([String: AnyDecodable].self) {
            self.init(dictionary.mapValues { $0.value })
        } 
```
should already cover this scenario. You can build any deep layer without any infinite loop.

maxep: No it's not, `let dictionary = try? container.decode([String: AnyDecodable].self)` will call `AnyDecodable.init(from decoder: Decoder)` on each value and we need to be able to get a `PassthroughAnyCodable` value.

In addition to [these](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogInternal/Sources/Codable/AnyDecodable.swift#L76L94), the `AnyDecodable` needs to support `PassthroughAnyCodable` otherwise it will fail to decode [those](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogInternal/Sources/Codable/AnyEncoder.swift#L675L678) and any additional passthrough values we may add.

maxep: It's an optimisation to avoid any useless encoding on known types, essentially types that are shared in `DatadogInternal`, see https://github.com/search?q=repo%3ADataDog%2Fdd-sdk-ios+PassthroughAnyCodable&type=code

It's indeed more difficult to handle but it's best to minimize message-bus impact on perf, so I won't be removing it.

ncreated: > I honestly think, we made it difficult to handle by introducing "Passthrough" thing.

I share the same feeling. For me, the [number of abstractions](https://github.com/DataDog/dd-sdk-ios/tree/develop/DatadogInternal/Sources/Codable) in messaging and feature context coding becomes too high to understand this part well. The `PassthroughAnyCodable` seems to promise convenience, but it makes the compromise on clarity. I'd rather see it kept bare simple.

DatadogInternal/Sources/Codable/AnyDecoder.swift
```diff
@@ -856,3 +856,26 @@ private extension DecodingError {
...
+                debugDescription: "Single Value Container cannot decode passthrough value."
+            ))
+        }
+
+        guard let value = container.value as? PassthroughAnyCodable else {
+            throw DecodingError.typeMismatch(PassthroughAnyCodable.self, in: container)
+        }
+
+        return value
+    }
```
ncreated, line 880: Any errors thrown from here are ignored in:
```swift
} else if let passthrough = try? container.passthrough() {
```
so I don't see exactly how it works 🤔 💭. It looks that this can eventually be:
```swift
func passthrough() -> PassthroughAnyCodable? {
   return (container as? _AnyDecoder.SingleValueContainer)?.value as? PassthroughAnyCodable
}
```

DatadogObjc/Sources/RUM/RUMDataModels+objc.swift
```diff
@@ -5544,4 +5544,4 @@ public class DDTelemetryConfigurationEventView: NSObject {
 
 // swiftlint:enable force_unwrapping
 
-// Generated from https://github.com/DataDog/rum-events-format/tree/1c476e469d5827aa1f4e60916f42ad35bbd950ef
+// Generated from https://github.com/DataDog/rum-events-format/tree/f69ca4664ed6e69c929855d02c4ce3d4b85d0bb4
```
ncreated, line 5547: > ~We have an issue in webview tracking integration tests, not related to this change but rather an issue with data sent by browser sdk!~ Fixed by updating RUM models

question/ What was the mentioned problem in WebViewTracking that required change to RUM data models?

ncreated (commented):
I'm not seeing directly how the change to `Passthrough` coding relates to the problem with `RUMSessionState` failing to be decoded. The `RUMSessionState` doesn't conform to `PassthroughAnyCodable` 🤔 - what is the link?

DatadogObjc/Sources/RUM/RUMDataModels+objc.swift
maxep: The format changed and IntegrationTests were failing to decode events coming from the Browser SDK: https://app.bitrise.io/build/f699b944-b3eb-48a1-9c02-67e3d427af43

DatadogInternal/Sources/Codable/AnyDecoder.swift
maxep: It could return `nil` indeed, but error escalation is a best and follow convention in `Codable`.

maxep:
> I'm not seeing directly how the change to Passthrough coding relates to the problem with `RUMSessionState` failing to be decoded. The `RUMSessionState` doesn't conform to `PassthroughAnyCodable` 🤔 - what is the link?

@ncreated, sorry for the lack of context! 

The `RUMSessionState` has a `sessionUUID: UUID` property and `UUID` conforms to `PassthroughAnyCodable`, so when transmitted on the message-bus it skips the [encoding](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogInternal/Sources/Codable/AnyEncoder.swift#L653). 
Then crash-reporting tries to decode it to AnyDecodable [here](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogCrashReporting/Sources/CrashContext/CrashContextProvider.swift#L130), which was failing because [AnyDecodable did not expect other types than primitives](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogInternal/Sources/Codable/AnyDecodable.swift#L76L94).

The fix allow decoding `PassthroughAnyCodable` in `AnyDecodable`.

DatadogInternal/Sources/Codable/AnyDecoder.swift
ncreated: Thing is, errors are ignored in caller code, so this is effectively a boilerplate. The way we use Swift's structured error handling here is not idiomatic. We throw error only to receive `nil` in `try?`, which is very confusing and makes an illusion of more control over this piece than we actually have (e.g. we customise error metadata, but other code ignores it). Tests we added only skew the real code coverage.

Not a blocker, but I'd suggest to keep it simple if we can't make use of error handling here.

DatadogObjc/Sources/RUM/RUMDataModels+objc.swift
ncreated: Oh I see, indeed https://github.com/DataDog/rum-events-format/pull/163 introduced backward-incompatible change to RUM schema 😱.

DatadogInternal/Sources/Codable/AnyDecoder.swift
maxep: The fact that it's ignored a caller doesn't mean we shouldn't escalate errors. But I get your point, it's more valuable to throw something if we have a signature like:
```swift
func passthrough<T>(_ type T.Type) throws -> T where T: PassthroughAnyCodable
```
Here, it doesn't really make sense! I can remove it

DatadogObjc/Sources/RUM/RUMDataModels+objc.swift
maxep: It's ok-ish because the SDK transmit browser events as-is without trying to decode..

ganeshnj (approved):
For unblocking release, we are good but would like to revisit this area sooner.

Thanks for answering my questions.

ncreated (approved):

PR #1507, opened 2023-10-03
fix: Flakiness in `IntegrationScenarios / RUMStopSessionScenarioTests`

maxep:
### What and why?

### How?

Increase simulate loading time to keep the view active

### Review checklist

- [x] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/IntegrationTests/Runner/Scenarios/RUM/StopSessionScenario/KioskSendEventsViewController.swift
+++ b/IntegrationTests/Runner/Scenarios/RUM/StopSessionScenario/KioskSendEventsViewController.swift
@@ -36,7 +36,7 @@ internal class KioskSendEventsViewController: UIViewController {
 
         let simulatedResourceKey = "/resource/1"
         let simulatedResourceRequest = URLRequest(url: URL(string: "https://foo.com/resource/1")!)
-        let simulatedResourceLoadingTime: TimeInterval = 0.1
+        let simulatedResourceLoadingTime: TimeInterval = 0.5
 
         rumMonitor.startResource(
             resourceKey: simulatedResourceKey,


maxep:
Other flakiness detected but the `RUMStopSessionScenarioTests` is fixed!

IntegrationTests/Runner/Scenarios/RUM/StopSessionScenario/KioskSendEventsViewController.swift
```diff
@@ -36,7 +36,7 @@ internal class KioskSendEventsViewController: UIViewController {
 
         let simulatedResourceKey = "/resource/1"
         let simulatedResourceRequest = URLRequest(url: URL(string: "https://foo.com/resource/1")!)
-        let simulatedResourceLoadingTime: TimeInterval = 0.1
+        let simulatedResourceLoadingTime: TimeInterval = 0.5
```
ganeshnj, line 39: can this be a real issue?

for example if a customer uses start and stop resource APIs right after.
```swift
        rumMonitor.startResource(
            resourceKey: simulatedResourceKey,
            request: simulatedResourceRequest
        )
        rumMonitor.stopResource(
                resourceKey: simulatedResourceKey,
                response: HTTPURLResponse(
                    url: simulatedResourceRequest.url!,
                    statusCode: 200,
                    httpVersion: nil,
                    headerFields: ["Content-Type": "image/png"]
                )!
            )
```

maxep: This won't be an issue for users, if they start and stop the resource, it will just not keep the view as active and it's completely fine.
The flakiness comes from what we expect in the integration test, where we assert that the view is first active then inactive when resources are done and the session is stoped.

ganeshnj (approved):

IntegrationTests/Runner/Scenarios/RUM/StopSessionScenario/KioskSendEventsViewController.swift
ganeshnj: perfect, thanks for quick turnaround.

PR #1510, opened 2023-10-06
 RUM-717 refactor: deprecate DatadogURLSessionDelegate and family classes

ganeshnj:
### What and why?

We want to deprecate the `DatadogURLSessionDelegate` and its related types. We already have URLSessionInstrumentation PR which now allows instrumenting web requests without the need of `DatadogURLSessionDelegate`.

### How?

We have warning as error setting is enabled for the SDK and deprecating any type and using it anywhere triggers a warning and hence, the build fails.

So, this PR extracts the URLSessionInstrumentation tests to separate targets

- iOS
- tvOS
- objc iOS
- objc tvOS

Move the relevant code new targets and include them in CI runs.

Although this doesn't solve the problem because we have to use DatadogURLSessionDelegate in objc wrappers which breaks the build again hence I disabled the warning as error setting for objc targets.

Finally, integration test targets are also updated to not use warning as error setting because we don't want to stop testing the deprecated types.

> I personally dislike this approach and would prefer to have a better solution but this is the best I could come up with. I am open to suggestions.

> Alternatively, I'm open for disabling warning as errors across the board because right now we are only using it for the SDK and in unit tests.

> IMO, deprecation must be simpler and easy to do. This is a lot of work to deprecate one type.

### Review checklist

- [x] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
```diff
@@ -103,16 +103,6 @@ internal final class NetworkInstrumentationFeature: DatadogFeature {
         }
     }
 
-    private func firstPartyHosts(configuration: URLSessionInstrumentation.Configuration, delegate: URLSessionDelegate) -> FirstPartyHosts? {
```
ganeshnj, line 106: unused method

ncreated:
🙏 Please add enough context in PR description.

ganeshnj:
> 🙏 Please add enough context in PR description.

err yes for sure. All up to date now.

IntegrationTests/IntegrationTests.xcodeproj/project.pbxproj
```diff
@@ -1115,6 +1115,7 @@
 				SDKROOT = iphoneos;
 				SWIFT_ACTIVE_COMPILATION_CONDITIONS = DEBUG;
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
+				SWIFT_TREAT_WARNINGS_AS_ERRORS = NO;
```
maciejburda, line 1118: I guess this is temporary until we remove deprecated methods (in v3?)

maciejburda (approved):
LGTM

Not a biggest fan of disabling `SWIFT_TREAT_WARNINGS_AS_ERRORS`, but I can't think of a better way if we want to leverage deprecated mechanism.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/DatadogURLSessionDelegate.swift
```diff
@@ -19,6 +21,7 @@ public protocol __URLSessionDelegateProviding: URLSessionDelegate {
 ///
 /// All requests made with the `URLSession` instrumented with this delegate will be intercepted by the SDK.
 @objc
+@available(*, deprecated, message: "Use URLSessionInstrumentation instead.")
```
ncreated, line 24: suggestion/ Most users will face this warning without much context (seeing docs or migration guide), so I wonder if it is informative enough. _What is URLSessionInstrumentation? Is this a module? A class?_

IMO, better to reference entry-point API, so users could search for it and try to figure-it out:
```swift
```

However, best IMO would be to enhance this public API comment to better explain the usage of `URLSessionInstrumentation`. Then, calling this symbol from deprecation warning could be enough 👌.
https://github.com/DataDog/dd-sdk-ios/blob/643d26e6ba2ce6582f28aee2a1789f6dccc94491/DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionInstrumentation.swift#L9-L10

DatadogInternal/Sources/NetworkInstrumentation/FirstPartyHosts.swift
```diff
@@ -8,7 +8,7 @@ import Foundation
 
 /// A struct that represents a dictionary of host names and tracing header types.
 public struct FirstPartyHosts: Equatable {
-    internal var hostsWithTracingHeaderTypes: [String: Set<TracingHeaderType>]
+    public var hostsWithTracingHeaderTypes: [String: Set<TracingHeaderType>]
```
ncreated, line 11: q/ Is this intentional? Where does it come from?

ncreated (approved):
Looks good 👌, although I left suggestion on the dev-x caused by this deprecation - we could make it more obvious IMO.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/DatadogURLSessionDelegate.swift
ganeshnj: that's fair call, I just made the changes.

DatadogInternal/Sources/NetworkInstrumentation/FirstPartyHosts.swift
ganeshnj: needed in previous interation, not anymore reverted.

PR #1513, opened 2023-10-09
RUM-1188 Add "consent" attribute to batch telemetry

ncreated:
### What and why?

📦 This PR adds observability to batches closed and deleted during `.pending` [Tracking Consent](https://swiftpackageindex.com/datadog/dd-sdk-ios/develop/documentation/datadogcore/trackingconsent). Previously, only `.granted` batches were counted in mobile metrics. To differentiate the value of consent, we add a new attribute:
```
consent: granted | pending
```

Lack of observability in `.pending` state might explain skews in volumes of "deleted" vs "closed" events (56% vs 44%). If batch was completed during `.pending` but sent in `.granted` consent, we would get "batch deleted" with no matching "batch closed".

### How?

* Enabled batching metrics for unauthorized orchestrator (the one managing `.pending` batches).
* Added new attribute to each telemetry.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogCore/Sources/Core/Storage/FeatureStorage.swift
+++ b/DatadogCore/Sources/Core/Storage/FeatureStorage.swift
@@ -120,25 +120,37 @@ extension FeatureStorage {
         encryption: DataEncryption?,
         telemetry: Telemetry
     ) {
+        let trackName = BatchMetric.trackValue(for: featureName)
+
+        if trackName == nil {
+            DD.logger.error("Can't determine track name for feature named '\(featureName)'")
+        }
+
         let authorizedFilesOrchestrator = FilesOrchestrator(
             directory: directories.authorized,
             performance: performance,
             dateProvider: dateProvider,
             telemetry: telemetry,
-            metricsData: {
-                guard let trackName = BatchMetric.trackValue(for: featureName) else {
-                    DD.logger.error("Can't determine track name for feature named '\(featureName)'")
-                    return nil
-                }
-                return FilesOrchestrator.MetricsData(trackName: trackName, uploaderPerformance: performance)
-            }()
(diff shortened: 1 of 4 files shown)


DatadogCore/Sources/Core/Storage/FeatureStorage.swift
```diff
@@ -120,25 +120,37 @@ extension FeatureStorage {
...
             dateProvider: dateProvider,
             telemetry: telemetry,
-            metricsData: {
-                guard let trackName = BatchMetric.trackValue(for: featureName) else {
-                    DD.logger.error("Can't determine track name for feature named '\(featureName)'")
-                    return nil
-                }
-                return FilesOrchestrator.MetricsData(trackName: trackName, uploaderPerformance: performance)
-            }()
+            metricsData: trackName.map { trackName in
```
xgouchet, line 134: > [!NOTE]
> Why are we using a `.map` operation here? it seems the `trackName` is a String, and not a list, so it seems not necessary

ncreated: It's [`Optional.map(_:)`](https://developer.apple.com/documentation/swift/optional/map(_:)), not [`Array.map(_:)`](https://developer.apple.com/documentation/swift/sequence/map(_:)) - a Swift counterpart of Kotlin's `?.let` I think.

xgouchet: Oh got it ;)

xgouchet (approved):

maciejburda (approved):
LGTM

PR #1517, opened 2023-10-13
Expose interfaces to add custom SR node recorders

louiszawadzki:
### What and why?

_This is a draft PR to get feedback on the implementation and possible improvements._

The goal of the feature is to enable cross-platform SDKs to add custom `NodeRecorder`s to the Session Replay configuration to map against custom elements and improve the support for these.
The feature in itself is in 2 parts: 
- adding the `additionalNodeRecorders` field and passing it along
- adding public interfaces for the CP SDKs

This could also potentially be used to add support to other 3rd party libraries or maybe SwiftUI, KMM, etc.

### DIFF
--- a/DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/ViewTreeSnapshotBuilder.swift
+++ b/DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/ViewTreeSnapshotBuilder.swift
@@ -44,9 +44,9 @@ internal struct ViewTreeSnapshotBuilder {
 }
 
 extension ViewTreeSnapshotBuilder {
-    init() {
+    init(additionalNodeRecorders: [NodeRecorder]? = nil) {
         self.init(
-            viewTreeRecorder: ViewTreeRecorder(nodeRecorders: createDefaultNodeRecorders()),
+            viewTreeRecorder: ViewTreeRecorder(nodeRecorders: createDefaultNodeRecorders() + (additionalNodeRecorders ?? [])),
             idsGenerator: NodeIDGenerator(),
             imageDataProvider: ImageDataProvider()
         )
--- a/DatadogSessionReplay/Sources/Feature/SessionReplayFeature.swift
+++ b/DatadogSessionReplay/Sources/Feature/SessionReplayFeature.swift
@@ -43,7 +43,8 @@ internal class SessionReplayFeature: DatadogRemoteFeature {
 
         let recorder = try Recorder(
             processor: processor,
-            telemetry: core.telemetry
+            telemetry: core.telemetry,
+            additionalNodeRecorders: configuration.additionalNodeRecorders
         )
         let recordingCoordinator = RecordingCoordinator(
             scheduler: scheduler,
--- a/DatadogSessionReplay/Sources/Processor/Privacy/TextObfuscator.swift
+++ b/DatadogSessionReplay/Sources/Processor/Privacy/TextObfuscator.swift
@@ -7,7 +7,7 @@
 #if os(iOS)
 import Foundation
 
-internal protocol TextObfuscating {
+public protocol TextObfuscating {
     /// Obfuscates given `text`.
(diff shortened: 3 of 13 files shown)


ncreated (commented):
Exposing all the necessary interfaces from `DatadogSessionReplay` module to accommodate cross-platform SDK features is certainly the simplest approach 👌. Because this is impactful change and comes with many consequences it might be essential to consider potential alternatives and weigh the implications.

What stands out here is the shift from `internal` to `public`. When designing `public` interfaces, we usually account for different considerations than when creating `internal` one. Simple switch from `internal` to `public` bypasses the entire process.

Being the contributor to most of interfaces changed in this PR, I feel it would be huge pain and large cost to track most of them as backward-compatible `public` API. Constructs like `NodeSemantics` or `ViewAttributes` changed significantly and multiple times since starting `DatadogSessionReplay`. Rapid refactors were not a problem for `internal` interface but it doesn't translate to `public`.

We could think of segregating required interface as _"`publicly` internal"_, like we do with some [RUM](https://github.com/DataDog/dd-sdk-ios/blob/101fe09b2dd1d444e7cca262874a5ffa3396ea40/DatadogRUM/Sources/RUMMonitorProtocol%2BInternal.swift#L22-L27), [Logs](https://github.com/DataDog/dd-sdk-ios/blob/101fe09b2dd1d444e7cca262874a5ffa3396ea40/DatadogLogs/Sources/LoggerProtocol%2BInternal.swift#L10-L11) or [telemetry](https://github.com/DataDog/dd-sdk-ios/blob/101fe09b2dd1d444e7cca262874a5ffa3396ea40/DatadogCore/Sources/Datadog%2BInternal.swift#L12-L20) APIs. Not sure how easy this might be, but at least it would make it more clear that we don't expect users to implement their own extensions for `DatadogSessionReplay`. For us it would mean less regime in tracking backward-compatibility, hence smaller cost of changes to native SR.

// cc @maciejburda @maxep @ganeshnj we need your inputs to the discussion as well. Having a meeting could make sense.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/ViewTreeSnapshotBuilder.swift
```diff
@@ -44,9 +44,9 @@ internal struct ViewTreeSnapshotBuilder {
 }
 
 extension ViewTreeSnapshotBuilder {
-    init() {
+    init(additionalNodeRecorders: [NodeRecorder]? = nil) {
```
maciejburda, line 47: can be non-nil and just default to `[]`

maciejburda:
I agree with @ncreated, and also on the side note in some particular cases you might want a recorder to be at certain index. For example `UnsupportedViewsRecorder` needs to be first to make sure we don't do any unnecessary computation before just rejecting the view.

Worth a meeting to discuss options.

ncreated:
> Also agreeing with @ncreated, this should be internal, I can look into it and add a new commit on top to make sure it's possible.
> 
> As I told @maciejburda, this functionality is already publicly available in Android: https://github.com/DataDog/dd-sdk-android/blob/develop/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/SessionReplayConfiguration.kt#L41
> 
> What I would suggest is to add it as internal first, so we can start developing and releasing cross-platform SDK extensions. Then if it makes sense to align with Android and make it public, we can make it public later, the refactoring cost should be quite easy on the RN side as we only support one iOS SDK version at a time.
> 
> What do you think @ncreated?

Fine to start internal and re-think public interface later 👌.

Android offering such API is a good argument here, however I'm not sure if the original intention on Android was to open SR for being extended by other developers (which is the promise behind `public` interface). AFAIK, it was rather to add support to [Material UI](https://github.com/DataDog/dd-sdk-android/tree/develop/features/dd-sdk-android-session-replay-material) components without making SR depend on massive `com.google.android.material.*` libraries - such, for iOS, is better to achieve with "publicly `internal`" API 🙂.

PR #1527, opened 2023-10-23
fix: add missing source directory required for JSONObjectMatcher

ganeshnj:
### What and why?

SPM only SDK fails to compile

### How?

Include the missing source files.

### Review checklist

- [ ] Run integration tests
- [x] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -186,7 +186,7 @@ let package = Package(
                 .target(name: "DatadogInternal"),
             ],
             path: "TestUtilities",
-            sources: ["Mocks", "Helpers"]
+            sources: ["Mocks", "Helpers", "Matchers"]
         )
     ]
 )


Package.swift
```diff
@@ -154,6 +154,7 @@ let package = Package(
             name: "DatadogWebViewTracking",
             dependencies: [
                 .target(name: "DatadogInternal"),
+                .target(name: "TestUtilities")
```
maxep, line 157: `DatadogWebViewTracking` should not depend on `TestUtilities`, I guess the change in L189 is enough 👍

maxep (changes requested):

Package.swift
ganeshnj: that was a mistake - already removed.

maxep (approved):
Thanks!

Issue #956, opened 2022-08-03
DataDog Performance Issues


charlie-foreflight:
We have been noticing performance issues with Datadog iOS SDK. Our application has 2 different logging methods: logging to a local file & Datadog.

After doing a stress test (10k logs every 2 seconds), the logging to a local file solution does have slight continuous memory increases, and a slight increase of CPU usage (40% or so). Whereas the Datadog iOS SDK has massive memory increases (1-2 GB in a minute, much higher than local file solution), and extremely high CPU usage (>100%).

What is really strange, is memory continues to increase for Datadog even after the stress test is turned off. After a while, it will stop using CPU usage, and memory will drop back down suddenly to normal levels. Whereas with the local file solution, memory usage will decline slowly overtime as it continues to write to disk.

From my understanding, Datadog logs all logs to disk originally, then sends to the server afterwards. Therefore I expect the performance characteristics to be somewhat similar between these two options.

However the continuous memory usage even after stopping the logging for Datadog and much higher CPU consumption is concerning.

I did notice in the code that this queue is user interactive, which is strange to me. Nothing about logging is user interactive. What was the thought process behind making this user interactive?

https://github.com/DataDog/dd-sdk-ios/blob/a7f4b50133de9243caf9d094ce9f604f8c1d60bb/Sources/Datadog/Logging/RemoteLogger.swift#L62-L65

ncreated:
Hello @charlie-foreflight 👋. Thank you for looking at our SDK 🙂. Yes, you're right - the SDK uses file storage under the hood and the expectation of measuring similar performance characteristic to regular write-to-file logging is reasonable.

We are aware of existing performance limitations and we're actively working on bringing improvements. These however include deep architectural changes and will be part of SDK `2.x` - coming soon both iOS and Android. To address your observations, let me explain what issue we've diagnosed and what actions we execute to improve it.

[...]

PR #1531, opened 2023-10-25
RUM-1837 Update logic to send N batches sequentially in each cycle

maciejburda:
### What and why?

Modifies the core logic to get a list of batches instead of single batch. It's done through a new configuration called `BatchProcessingLevel` that allows controlling the amount of batches processed sequentially without a delay within one reading/uploading cycle. Currently it exposed 3 levels: `low`, `medium` and `high` that translate to `1`, `10` and `100` of batches processed. By default it's taking up to `10` batches in a cycle.

This logic improves the data upload when batch back pressure occurs. 

### How?

Configuration is delivered to `DataUploadWorker` which implementation was modified to read up to X batches instead of single batch.

### Review checklist

- [x] Run integration tests
- [x] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -555,7 +555,6 @@ Release `2.0` introduces breaking changes. Follow the [Migration Guide](MIGRATIO
 [#1529]: https://github.com/DataDog/dd-sdk-ios/pull/1529
 [#1533]: https://github.com/DataDog/dd-sdk-ios/pull/1533
 [#1536]: https://github.com/DataDog/dd-sdk-ios/pull/1536
-[#1529]: https://github.com/DataDog/dd-sdk-ios/pull/1529
 [#1531]: https://github.com/DataDog/dd-sdk-ios/pull/1531
 [@00fa9a]: https://github.com/00FA9A
 [@britton-earnin]: https://github.com/Britton-Earnin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -555,7 +555,7 @@ Release `2.0` introduces breaking changes. Follow the [Migration Guide](MIGRATIO
 [#1529]: https://github.com/DataDog/dd-sdk-ios/pull/1529
 [#1533]: https://github.com/DataDog/dd-sdk-ios/pull/1533
 [#1536]: https://github.com/DataDog/dd-sdk-ios/pull/1536
-[#1529]: https://github.com/DataDog/dd-sdk-ios/pull/1531
+[#1529]: https://github.com/DataDog/dd-sdk-ios/pull/1529
 [#1531]: https://github.com/DataDog/dd-sdk-ios/pull/1531
 [@00fa9a]: https://github.com/00FA9A
 [@britton-earnin]: https://github.com/Britton-Earnin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -556,6 +556,7 @@ Release `2.0` introduces breaking changes. Follow the [Migration Guide](MIGRATIO
 [#1533]: https://github.com/DataDog/dd-sdk-ios/pull/1533
 [#1536]: https://github.com/DataDog/dd-sdk-ios/pull/1536
 [#1529]: https://github.com/DataDog/dd-sdk-ios/pull/1531
+[#1531]: https://github.com/DataDog/dd-sdk-ios/pull/1531
 [@00fa9a]: https://github.com/00FA9A
 [@britton-earnin]: https://github.com/Britton-Earnin
 [@hengyu]: https://github.com/Hengyu
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - [FEATURE] Change default tracing headers for first party hosts to use both Datadog headers and W3C `tracecontext` headers. See [#1529][]
 - [FEATURE] Add tracestate headers when using W3C tracecontext. See [#1536][]
 - [BUGFIX] Fix RUM ViewController leaks. See [#1533][]
+- [FEATURE] Add `BatchProcessingLevel` configuration allowing to process more batches within single read/upload cycle. See [#1531][]
 
 # 2.4.0 / 18-10-2023
 
@@ -554,6 +555,7 @@ Release `2.0` introduces breaking changes. Follow the [Migration Guide](MIGRATIO
 [#1529]: https://github.com/DataDog/dd-sdk-ios/pull/1529
 [#1533]: https://github.com/DataDog/dd-sdk-ios/pull/1533
 [#1536]: https://github.com/DataDog/dd-sdk-ios/pull/1536
+[#1529]: https://github.com/DataDog/dd-sdk-ios/pull/1531
 [@00fa9a]: https://github.com/00FA9A
 [@britton-earnin]: https://github.com/Britton-Earnin
 [@hengyu]: https://github.com/Hengyu
--- a/DatadogCore/Sources/Core/Storage/Reading/FileReader.swift
+++ b/DatadogCore/Sources/Core/Storage/Reading/FileReader.swift
@@ -30,17 +30,20 @@ internal final class FileReader: Reader {
 
     // MARK: - Reading batches
 
-    func readNextBatch() -> Batch? {
-        guard let file = orchestrator.getReadableFile(excludingFilesNamed: filesRead) else {
-            return nil
+    func readNextBatches(_ limit: Int?) -> [Batch] {
+        let files = orchestrator.getReadableFiles(excludingFilesNamed: filesRead, limit: limit)
+        guard !files.isEmpty else {
+            return []
         }
 
         do {
-            let dataBlocks = try decode(stream: file.stream())
-            return Batch(dataBlocks: dataBlocks, file: file)
+            return try files.map { file in
+                let dataBlocks = try decode(stream: file.stream())
+                return Batch(dataBlocks: dataBlocks, file: file)
+            }
         } catch {
             telemetry.error("Failed to read data from file", error: error)
-            return nil
+            return []
         }
     }
 
--- a/DatadogCore/Sources/Core/Storage/Reading/FileReader.swift
+++ b/DatadogCore/Sources/Core/Storage/Reading/FileReader.swift
@@ -30,7 +30,7 @@ internal final class FileReader: Reader {
 
     // MARK: - Reading batches
 
-    func readFiles(_ limit: Int?) -> [ReadableFile] {
+    func readFiles(limit: Int) -> [ReadableFile] {
         return orchestrator.getReadableFiles(excludingFilesNamed: filesRead, limit: limit)
     }
 
--- a/DatadogCore/Sources/Core/Storage/Reading/FileReader.swift
+++ b/DatadogCore/Sources/Core/Storage/Reading/FileReader.swift
@@ -30,19 +30,17 @@ internal final class FileReader: Reader {
 
     // MARK: - Reading batches
 
-    func readNextBatches(_ limit: Int?) -> [Batch] {
-        let files = orchestrator.getReadableFiles(excludingFilesNamed: filesRead, limit: limit)
-        guard !files.isEmpty else {
-            return []
-        }
-        return files.compactMap { file in
-            do {
-                let dataBlocks = try decode(stream: file.stream())
-                return Batch(dataBlocks: dataBlocks, file: file)
-            } catch {
-                telemetry.error("Failed to read data from file", error: error)
-                return nil
-            }
+    func readFiles(_ limit: Int?) -> [ReadableFile] {
+        return orchestrator.getReadableFiles(excludingFilesNamed: filesRead, limit: limit)
+    }
+
+    func readBatch(from file: ReadableFile) -> Batch? {
+        do {
+            let dataBlocks = try decode(stream: file.stream())
+            return Batch(dataBlocks: dataBlocks, file: file)
+        } catch {
+            telemetry.error("Failed to read data from file", error: error)
+            return nil
         }
     }
 
--- a/DatadogCore/Sources/Core/Storage/Reading/FileReader.swift
+++ b/DatadogCore/Sources/Core/Storage/Reading/FileReader.swift
@@ -35,15 +35,14 @@ internal final class FileReader: Reader {
         guard !files.isEmpty else {
             return []
         }
-
-        do {
-            return try files.map { file in
+        return files.compactMap { file in
+            do {
                 let dataBlocks = try decode(stream: file.stream())
                 return Batch(dataBlocks: dataBlocks, file: file)
+            } catch {
+                telemetry.error("Failed to read data from file", error: error)
+                return nil
             }
-        } catch {
-            telemetry.error("Failed to read data from file", error: error)
-            return []
         }
     }
 
--- a/DatadogCore/Sources/Core/Storage/Reading/Reader.swift
+++ b/DatadogCore/Sources/Core/Storage/Reading/Reader.swift
@@ -24,6 +24,6 @@ extension Batch {
 
 /// A type, reading batched data.
 internal protocol Reader {
-    func readNextBatch() -> Batch?
+    func readNextBatches(_ count: Int?) -> [Batch]
     func markBatchAsRead(_ batch: Batch, reason: BatchDeletedMetric.RemovalReason)
 }
--- a/DatadogCore/Sources/Core/Storage/Reading/Reader.swift
+++ b/DatadogCore/Sources/Core/Storage/Reading/Reader.swift
@@ -24,6 +24,23 @@ extension Batch {
 
 /// A type, reading batched data.
 internal protocol Reader {
-    func readNextBatches(_ count: Int?) -> [Batch]
+    /// Reads files from the storage.
+    /// - Parameter limit: maximum number of files to read. If `nil`, all files are read.
+    func readFiles(_ limit: Int?) -> [ReadableFile]
+    /// Reads batch from given file.
+    /// - Parameter file: file to read batch from.
+    func readBatch(from file: ReadableFile) -> Batch?
+    /// Reads next batches from the storage.
+    /// - Parameter limit: maximum number of batches to read. If `nil`, all batches are read.
+    func readNextBatches(_ limit: Int?) -> [Batch]
+    /// Marks given batch as read.
+    /// - Parameter batch: batch to mark as read.
+    /// - Parameter reason: reason for removing the batch.
     func markBatchAsRead(_ batch: Batch, reason: BatchDeletedMetric.RemovalReason)
+}
+
+extension Reader {
+    func readNextBatches(_ limit: Int? = nil) -> [Batch] {
+        return readFiles(limit).compactMap { readBatch(from: $0) }
+    }
 }
--- a/DatadogCore/Sources/Core/Storage/Reading/Reader.swift
+++ b/DatadogCore/Sources/Core/Storage/Reading/Reader.swift
@@ -25,8 +25,8 @@ extension Batch {
 /// A type, reading batched data.
 internal protocol Reader {
     /// Reads files from the storage.
-    /// - Parameter limit: maximum number of files to read. If `nil`, all files are read.
-    func readFiles(_ limit: Int?) -> [ReadableFile]
+    /// - Parameter limit: maximum number of files to read.
+    func readFiles(limit: Int) -> [ReadableFile]
     /// Reads batch from given file.
     /// - Parameter file: file to read batch from.
     func readBatch(from file: ReadableFile) -> Batch?
--- a/DatadogCore/Sources/Core/Storage/Reading/Reader.swift
+++ b/DatadogCore/Sources/Core/Storage/Reading/Reader.swift
@@ -30,17 +30,8 @@ internal protocol Reader {
     /// Reads batch from given file.
     /// - Parameter file: file to read batch from.
     func readBatch(from file: ReadableFile) -> Batch?
-    /// Reads next batches from the storage.
-    /// - Parameter limit: maximum number of batches to read. If `nil`, all batches are read.
-    func readNextBatches(_ limit: Int?) -> [Batch]
     /// Marks given batch as read.
     /// - Parameter batch: batch to mark as read.
     /// - Parameter reason: reason for removing the batch.
     func markBatchAsRead(_ batch: Batch, reason: BatchDeletedMetric.RemovalReason)
-}
-
-extension Reader {
-    func readNextBatches(_ limit: Int? = nil) -> [Batch] {
-        return readFiles(limit).compactMap { readBatch(from: $0) }
-    }
 }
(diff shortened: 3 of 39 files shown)


ncreated (commented):
How will this improvement be measured? During backlog grooming we discussed adding "batch processing level" to "Batch Deleted" telemetry, so we can correlate batch delivery with certain configuration. This part of definition of done seems lacking.

DatadogCore/Sources/Core/Storage/Reading/Reader.swift
```diff
@@ -24,6 +24,6 @@ extension Batch {
 
 /// A type, reading batched data.
 internal protocol Reader {
-    func readNextBatch() -> Batch?
+    func readNextBatches(_ count: Int?) -> [Batch]
```
ncreated, line 27: If `count` is optional, what does `count: nil` mean in the context of retrieving next batches? It deserves an API comment with explanation.

DatadogCore/Tests/Datadog/Core/Persistence/Reading/FileReaderTests.swift
ncreated, line 1: `readNextBatches(nil)` is not tested

CHANGELOG.md
```diff
@@ -550,6 +551,7 @@ Release `2.0` introduces breaking changes. Follow the [Migration Guide](MIGRATIO
 [#1394]: https://github.com/DataDog/dd-sdk-ios/pull/1394
 [#1524]: https://github.com/DataDog/dd-sdk-ios/pull/1524
 [#1529]: https://github.com/DataDog/dd-sdk-ios/pull/1529
+[#1529]: https://github.com/DataDog/dd-sdk-ios/pull/1531
```
ncreated, line 554:
```suggestion
[#1531]: https://github.com/DataDog/dd-sdk-ios/pull/1531
```

DatadogCore/Sources/Core/Storage/Reading/FileReader.swift
```diff
@@ -30,17 +30,19 @@ internal final class FileReader: Reader {
...
-            telemetry.error("Failed to read data from file", error: error)
-            return nil
+        return files.compactMap { file in
+            do {
+                let dataBlocks = try decode(stream: file.stream())
+                return Batch(dataBlocks: dataBlocks, file: file)
+            } catch {
+                telemetry.error("Failed to read data from file", error: error)
+                return nil
+            }
```
ncreated, line 45: Does it imply reading all files into memory before SDK initiates sending them? Imagine files stored in each of tracks, would that pesimisticaly mean reading up 40 files into memory with `.medium` and up to **400 files** with `.high`?

We've been struggling problems with allocations and steep memory graphs due to batch reading (like https://github.com/DataDog/dd-sdk-ios/pull/1089 or https://github.com/DataDog/dd-sdk-ios/issues/956#issuecomment-1384426485) and this change seems to be a step in opposite direction.

There is no reason to pre-read all files into memory before we start sending first of them. We should read them one-by-one for each upload, keeping the memory graph flat.

maciejburda: That's a good callout. I'll modify the implementation to only load single batch at a time

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
```diff
@@ -101,23 +102,25 @@ internal class DataUploadWorker: DataUploadWorkerType {
...
                 }
             } else {
                 let batchLabel = batches?.isEmpty == false ? "YES" : (isSystemReady ? "NO" : "NOT CHECKED")
                 DD.logger.debug("💡 (\(self.featureName)) No upload. Batch to upload: \(batchLabel), System conditions: \(blockersForUpload.description)")
 
-                self.delay.increase()
                 self.backgroundTaskCoordinator?.endBackgroundTask()
             }
 
+            allUploadsSucceeded
```
mariusc83, line 123: hmmm I think this is debatable or I do not see it well in this PR. I would only `decrease/increase` in case there was a 500 .

maciejburda: That's a good call.

It should only increase the delay for those batches that `needsRetry`.

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
```diff
@@ -62,56 +63,67 @@ internal class DataUploadWorker: DataUploadWorkerType {
...
-                        DD.logger.debug("   → (\(self.featureName)) not delivered, will be retransmitted: \(uploadStatus.userDebugDescription)")
-                    } else {
-                        self.fileReader.markBatchAsRead(batch, reason: .intakeCode(responseCode: uploadStatus.responseCode ?? -1)) // -1 is unexpected here
-                        self.delay.decrease()
-
-                        DD.logger.debug("   → (\(self.featureName)) accepted, won't be retransmitted: \(uploadStatus.userDebugDescription)")
+            let files = isSystemReady ? self.fileReader.readFiles(maxBatchesPerUpload) : nil
+            var allUploadsSucceeded = false
+            if let files = files, files.isEmpty == false {
+                for file in files {
```
maxep, line 72: /blocker I don't think we should do a for loop within the work-item, otherwise we would be blocking one thread for the entire list of batches. Instead, we should create a `DispatchWorkItem` per batch upload combined in a `DispatchGroup`, this way we will balance the load between threads.

maxep (changes requested):
Looks great! I see room for improvement in perfs: our `DataUploader` uses a semaphore which will be blocking the thread. So I think we should balance the load instead of uploading in a loop. I left a suggestion in comment, LMKWYT!

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
ncreated: > we would be blocking one thread for the entire list of batches

I agree 👍

> Instead, we should create a `DispatchWorkItem` per batch upload combined in a `DispatchGroup`, this way we will balance the load between threads

Distributing the load is one thing but the number of consumed threads is another. If I understand correctly @maxep , proposal is to use `DispatchGroup` and sync completion of multiple parallel `DispatchWorkItems`? That would explode the number of threads we use by orders of magnitude (think: 4 tracks, up to 100 batches).

Instead, IMO we should execute `DispatchWorkItems` sequentially on one queue, one after another (with no delay between them). This will link the number of threads to the number of tracks we use (so max 4).

Also, as we won't be drastically changing the routine of single `DispatchWorkItem`, this solution may not require changes to `Reader` interface (we will keep reading one batch per item like we do today). I'd also recommend to do a closer look at [this logic](https://github.com/DataDog/dd-sdk-ios/blob/05774d7c886efe98cf2cec662249fd96950cb790/DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift#L155-L165) after we implement the change. We hit problems in frequent directory enumerations in https://github.com/DataDog/dd-sdk-ios/pull/185. With the model proposed in this PR, we will be reading up to 400 batches sequentially in short delays, which requires extra measures.

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
```diff
@@ -62,56 +63,67 @@ internal class DataUploadWorker: DataUploadWorkerType {
...
-
-                    switch uploadStatus.error {
-                    case .unauthorized:
-                        DD.logger.error("⚠️ Make sure that the provided token still exists and you're targeting the relevant Datadog site.")
-                    case let .httpError(statusCode: statusCode):
-                        telemetry.error("Data upload finished with status code: \(statusCode)")
-                    case let .networkError(error: error):
-                        telemetry.error("Data upload finished with error", error: error)
-                    case .none: break
+                    DD.logger.debug("⏳ (\(self.featureName)) Uploading batch...")
```
ncreated, line 77: Do we have any thoughts about these logs? It will become extremely spammy once we enable 10 uploads per track. // cc @maciejburda @maxep

ncreated (dismissed):

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
maciejburda: This is intentional. The idea is to read and upload all the selected batches without a delay synchronously.

DatadogCore/Sources/Core/Storage/Reading/Reader.swift
```diff
@@ -24,6 +24,23 @@ extension Batch {
 
 /// A type, reading batched data.
 internal protocol Reader {
-    func readNextBatch() -> Batch?
+    /// Reads files from the storage.
+    /// - Parameter limit: maximum number of files to read. If `nil`, all files are read.
+    func readFiles(_ limit: Int?) -> [ReadableFile]
```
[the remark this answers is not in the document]
maciejburda: To me it felt cleaner to rely on `nil` rather than `.max`, but I'm happy to change 🙌

CHANGELOG.md
```diff
@@ -554,6 +555,8 @@ Release `2.0` introduces breaking changes. Follow the [Migration Guide](MIGRATIO
 [#1529]: https://github.com/DataDog/dd-sdk-ios/pull/1529
 [#1533]: https://github.com/DataDog/dd-sdk-ios/pull/1533
 [#1536]: https://github.com/DataDog/dd-sdk-ios/pull/1536
+[#1529]: https://github.com/DataDog/dd-sdk-ios/pull/1529
```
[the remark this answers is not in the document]
maciejburda: Great catch!

It's been a bad merge :/

Makes me thing - that it's probably worth updating the doc in a way where we have parametrised link (not sure if possible in markdown)

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
maciejburda: Yeah good point. I will revisit :)

maciejburda:
@maxep @ncreated 
I've made RAM tests using log pressure example from:
https://github.com/DataDog/dd-sdk-ios/pull/185

10 batches:
10-batches

develop:
develop

Since files are symlinks turned into data just before upload synchronously - we just see extended allocation, but it's not a spike. In my tests it was way faster to upload everything when using 10-batches - which is a goal of the PR.

Using multiple `DispatchWorkItem` will be cumbersome as we rely on plenty of different states. I actually explored this solution before jumping into a loop, and I would advice strongly against it. Blocking the thread is in my opinion fine. `DataUploadWorker` was designed in a way where it's in control of it's own work and since `DataUploader` is using semaphore it's playing nicely together in a synchronous fashion. Note that we only schedule next `DispatchWorkItem` after we finish current one.

I think that there is a potential in refactoring this into a DispatchQueue, but it will become quite a big change. I think it's better iterate on existing setup first.

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
maxep: The idea with `DispatchGroup` is to notify the end of the upload to plan the next cycle, it was not necessarily about uploading in parallel. But indeed, we can only use `DispatchWorkItems` 👍 

@maciejburda, re. [your comment](https://github.com/DataDog/dd-sdk-ios/pull/1531#issuecomment-1805666977): Yes, the fact that we don't load all batches is best for memory, it's even better with the `autoreleaseFrequency: .workItem` property that we use for the upload queue.

My point here is about CPU, if you are running long blocking upload tasks, GDC will be forced to create new threads for other work items (form the host app), which is suboptiomal. See [Avoiding Excessive Thread Creation](https://developer.apple.com/documentation/dispatch/dispatchqueue#3136752).
Note that all the upload queues target the `.global(qos: .utility)` which is concurrent, meaning that you might be blocking N threads for long period, N being the number of registered Features.

If you use different work-items, you will give GDC the opportunity to dispatch the load on various threads, giving also the opportunity to run higher priority tasks in between each upload work-items.

maciejburda: `DataUploadWorker` only performs single `DispatchWorkItem` at a time.

maciejburda:
@maxep @ncreated I refactored the code to use multiple `DispatchWorkItem` maintaining old the designed behaviours. I added ability to cancel, which can come in handy when we want to restart the flow under certain conditions (i.e. when app goes into background).

Please take another look 🙏

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
ncreated, line 1: Have we considered following? 
* 1/ Add [`uploadWork`](https://github.com/DataDog/dd-sdk-ios/blob/da1f1b99257e36edc114964651fa4a33f7d35d03/DatadogCore/Sources/Core/Upload/DataUploadWorker.swift#L61) `N` times to the `queue` with same `delay` to schedule `N` sequential uploads.
* 2/ Status of `N` succeeding items can be tracked in extra array and considered in every `N`th instance of the item to either increase or decrease delay and perform next schedule.

It should require much less changes to `DataUploadWorker` and (similar to https://github.com/DataDog/dd-sdk-android/pull/1686) won't require refactoring of other Core concepts like `Reader` interface or `FilesOrchestrator`.

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
```diff
@@ -55,95 +63,124 @@ internal class DataUploadWorker: DataUploadWorkerType {
...
+            if let files = files, !files.isEmpty {
+                DD.logger.debug("⏳ (\(self.featureName)) Uploading batches...")
                 self.backgroundTaskCoordinator?.beginBackgroundTask()
-                DD.logger.debug("⏳ (\(self.featureName)) Uploading batch...")
+                files.forEach { self.createAndScheduleBatchUploadWork(file: $0, context: context) }
+            } else {
+                let batchLabel = files?.isEmpty == false ? "YES" : (isSystemReady ? "NO" : "NOT CHECKED")
+                DD.logger.debug("💡 (\(self.featureName)) No upload. Batch to upload: \(batchLabel), System conditions: \(blockersForUpload.description)")
+                self.delay.increase()
+                self.backgroundTaskCoordinator?.endBackgroundTask()
```
ncreated, line 100: question/ It beings BG task each time when it finds a bach, but it ends it only when there are no more batches? How does it ensure that `begin()` is always matched with `end()`?

maciejburda: It the same behaviour as the previous implementation. In the case mentioned it's balanced by the background task coordinator internal callback.

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
maciejburda: Yes, but this makes state management more difficult. I want to maintain all the functionalities of the UploadWorker.

ncreated: With current proposal we manage multiple states:
* `private var batchReadWork: DispatchWorkItem?` to set, cancel and reset "read item"
* `private var batchUploadWorks: [DispatchWorkItem]` to set, cancel and reset multiple "upload items"
* `files.forEach { self.createAndScheduleBatchUploadWork(file: $0, context: context) }` to transfer `files` from "read item" to "upload items"

We do it from two separate routines: "read multiple batches" and "upload multiple batches".

---

What I propose is to use single, immutable `uploadWork` item and track single state:
* `private var itemResults: [ItemResult]` to reset it before scheduling `N` items and to read it upon completion of `N`th item to properly manage `delay`

That can be done from the original "read and upload single batch" routine.

maxep: Hey 👋  It is going in the right direction :) I agree with Maciek's proposal to simplify a bit. By having a single work-item, we can avoid changes in the reader + we would use a single array for tracking the state.

---
I also see a recursive way of doing it, which may simply things even more. The following is a quick draft to consider as well:
<details>
  <summary>Recursive Upload</summary>
  
```swift
@ReadWriteLock
private var batchUploadWork: DispatchWorkItem? = nil

init(...) {
    scheduleNextCycle()
}

private func scheduleNextCycle() {
    queue.asyncAfter(deadline: .now() + delay.current) { [weak self] in
        guard let count = self?.maxBatchesPerUpload else {
            return
        }

        self?.uploadBatches(count: count)
    }
}

private func uploadBatches(count: Int) {
    uploadWork = DispatchWorkItem { [weak self] in
        guard let self = self else {
            return
        }

        guard count > 0 else {
            // We have successfully uploaded all batches
            self.delay.decrease()
            self.scheduleNextCycle()
            return
        }

        let batch = readNextBatch()

        if upload(batch: batch) {
            // call next uploads recursively
            self?.scheduleBatchesUpload(count: count - 1)
        } else {
            // Stop on the first failure and schedule next cycle
            self.delay.increase()
            self?.scheduleNextCycle()
        }
    }

    batchUploadWork = uploadWork
    queue.async(execute: uploadWork)
}
```
</details>

---
In both cases, we can prevent changes in `Reader` and only focus on the upload worker 👍

maciejburda: Yeah I was considering that - but the problem is we can't predict how many batches we have available to upload. We would need to get information from the `FileOrchestrator` about total number of batches, so we know how many results means end of cycle.

Obviously we can make it work in proposed way, but it I think it will fail to deliver the simplification promise as we'll end up having quite complex state control to deal with.

It sounds like a big rework and I'm hesitant about benefits.

maciejburda: Let me try backing up my idea with an example. In the proposed solution if we have 2 batches ready and the `maxBatchesPerUpload` is 5 we'll do: RX - Read X, U - Upload

1. Successful Upload:
**New** `[R1][U][R1][U][R1][R1][R1] -> decrease interval -> Schedule Read 5`

vs

**Old** `[R2][U][U] -> decrease interval -> Schedule Read X`

2. Failure after 1st:
**New** `[R1][U Failed][R1][U Succeeded?][R1][R1][R1]  -> increase interval -> Schedule Read 5`

vs

**Old** `[R2][U Failed] -> cancel other upload -> increase interval -> Schedule Read X`

**Note**
- File orchestrator reads all files just returns limited amount after **sorting**

maxep (commented):
Great work @maciejburda 👏 
I've left one minor comment, but it looks great!

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
```diff
@@ -55,95 +63,113 @@ internal class DataUploadWorker: DataUploadWorkerType {
...
-        queue.asyncAfter(deadline: .now() + delay, execute: work)
+        queue.async(execute: uploadWork)
     }
 
     /// Sends all unsent data synchronously.
     /// - It performs arbitrary upload (without checking upload condition and without re-transmitting failed uploads).
     internal func flushSynchronously() {
-        queue.sync {
-            while let nextBatch = self.fileReader.readNextBatch() {
+        queue.sync { [fileReader, dataUploader, contextProvider] in
```
ganeshnj, line 157: is capturing of these params are required?

I wonder what happens this class is de-allocated and there is still some files are pending.

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
```diff
@@ -55,95 +63,113 @@ internal class DataUploadWorker: DataUploadWorkerType {
...
             let context = contextProvider.read()
-            let blockersForUpload = self.uploadConditions.blockersForUpload(with: context)
+            let blockersForUpload = uploadConditions.blockersForUpload(with: context)
             let isSystemReady = blockersForUpload.isEmpty
-            let nextBatch = isSystemReady ? self.fileReader.readNextBatch() : nil
-            if let batch = nextBatch {
+            let files = isSystemReady ? fileReader.readFiles(limit: maxBatchesPerUpload) : nil
+            if let files = files, !files.isEmpty {
+                DD.logger.debug("⏳ (\(self.featureName)) Uploading batches...")
                 self.backgroundTaskCoordinator?.beginBackgroundTask()
```
ganeshnj, line 84: not related to this PR but makes me think when `backgroundTaskCoordinator` be nil and what would happen in that case. IMO, we should do an early return.

maciejburda: `backgroundTaskCoordinator` only registers tasks to get some background time after app goes into background. When it's `nil` we can just ignore this code and assume that this code can happen in foreground only.

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
maciejburda: I can probably can fallback to just `[weak self]`.

This object would only get deallocated when we destroy core.

maxep (approved):
LGTM, thanks for the updates 🙏

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
```diff
@@ -55,95 +63,113 @@ internal class DataUploadWorker: DataUploadWorkerType {
...
-                DD.logger.debug("⏳ (\(self.featureName)) Uploading batch...")
+                self.uploadFile(from: files.reversed(), context: context)
```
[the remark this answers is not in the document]
maciejburda: It's for convenience. Both reversed and popLast are O(1). There is no `popFirst()` on array.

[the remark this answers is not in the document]
maxep: Ah got it 👍

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
```diff
@@ -28,9 +28,16 @@ internal class DataUploadWorker: DataUploadWorkerType {
...
-    /// Upload work scheduled by this worker.
+    /// Maximum number of batches to upload in one request.
+    private let maxBatchesPerUpload: Int
+
+    /// Batch reading work scheduled by this worker.
+    @ReadWriteLock
+    private var readWork: DispatchWorkItem?
+    /// Batch upload work scheduled by this worker.
+    @ReadWriteLock
     private var uploadWork: DispatchWorkItem?
```
maxep, line 39: /question I do prefer this than a boolean for cancellation 👍 I wonder why not using a single work-item for both? we would then have just one to cancel.

maxep (approved):

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
maciejburda: When I started the clean up it felt a little more consistent to me.

In the end of the day it's closure vs DispatchWorkItem

WDYT?

maxep: Yes, not a big deal!

PR #1538, opened 2023-11-03
RUM-1923 fix: include hex parent id in tracestate

ganeshnj:
### What and why?

Currently, if there is broken proxy in between two system, we will have the broken trace.

Hence, we need datadog trace id which can be used to recover the trace.

Eg. RUM SDK > Broken W3C Proxy > DD Tracer (recovered by reading tracestate from RUM SDK).

### How?

This PR adds the datadog trace id to the tracestate header using following format

```
p:{hex-encoded parent-id}
```

Example
```
x-datadog-trace-id: 1229782938247303441 (0x1111111111111111)
x-datadog-parent-id: 2459565876494606882 (0x2222222222222222)
x-datadog-sampling-priority: 2
x-datadog-origin: rum
```

translates to
```
traceparent: 00-00000000000000001111111111111111-2222222222222222-01
tracestate: dd=s:2;o:rum;p:2222222222222222
```

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogCore/Tests/Datadog/TracerTests.swift
+++ b/DatadogCore/Tests/Datadog/TracerTests.swift
@@ -879,7 +879,7 @@ class TracerTests: XCTestCase {
 
         // Then
         let expectedHTTPHeaders1 = [
-            "tracestate": "dd=s:1;o:rum",
+            "tracestate": "dd=s:1;o:rum;p:0000000000000002",
             "traceparent": "00-00000000000000000000000000000001-0000000000000002-01"
         ]
         XCTAssertEqual(httpHeadersWriter.traceHeaderFields, expectedHTTPHeaders1)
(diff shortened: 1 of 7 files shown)


maciejburda (dismissed):
This seems very much redundant, tbh.

ganeshnj:
> This seems very much redundant, tbh.

This is redundant for us but useful for the other tracers which are working across different systems.

Like one example I have given in the description

1. RUM SDK
2. Broken W3C Proxy
3. DD Tracer 

This Broken W3C Proxy will erase the parent information, but this extra vendor information will remain as it is. This vendor specific parent id will be read by the DD Tracer in order to get the real parent.

maciejburda (approved):
Thanks for the additional context. I haven't thought of proxy changing the header. In this case it works as safeguard.

PR #1539, opened 2023-11-06
Enable custom node recorders

louiszawadzki:
### What and why?

Add custom node recorders for cross platform SDKs to hook into.

### How?

A few notes on the implementation choices.

I've had to put more logic than I imagined into the iOS SDK since we rely quickly on functions to return types from Data models (that are defined from the schemas), that I cannot make public.
I've decided to add a adapter layer at the level of `NodeRecorders` and builders rather than at the level of Data models, which would have required more changes in the logic of Session Replay.
This means I need to write a generic mapper for each method of WireframesBuilder (`createTextWireframe`, `createImageWireframe`, etc.). I've gone with only an implementation for TextWireframe for now as it's likely the only one we'll need for RN, but it can be extended later.
This was the easiest implementation to start with, let me know if I should refactor this.

Also I could not put the `mutating` function to set the additionalNodeRecorders on the configuration object in an Extension since that extension would be readonly, so I only went with the `@_spi` solution to hide it.

I've implemented a Text mapper that works on RN, if you want to look at what implementation would look like: https://github.com/DataDog/dd-sdk-reactnative/blob/louiszawadzki/session-replay-poc/packages/core/ios/Sources/RCTTextViewRecorder.swift

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogSessionReplay/Sources/Processor/SRDataModelsBuilder/WireframesBuilder.swift
+++ b/DatadogSessionReplay/Sources/Processor/SRDataModelsBuilder/WireframesBuilder.swift
@@ -17,7 +17,7 @@ internal typealias WireframeID = NodeID
 /// It is used by the player to reconstruct individual elements of the recorded app UI.
 ///
 /// Note: `WireframesBuilder` is used by `Processor` on a single background thread.
-internal class WireframesBuilder {
+@_spi(Internal) public class SessionReplayWireframesBuilder {
     /// A set of fallback values to use if the actual value cannot be read or converted.
     ///
     /// The idea is to always provide value, which would make certain element visible in the player.
@@ -191,6 +191,8 @@ internal class WireframesBuilder {
         )
     }
 }
+
+internal typealias WireframesBuilder = SessionReplayWireframesBuilder
 
 // MARK: - Convenience
 
--- a/DatadogSessionReplay/Sources/SessionReplay+Internal.swift
+++ b/DatadogSessionReplay/Sources/SessionReplay+Internal.swift
@@ -1,0 +1,152 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2023-Present Datadog, Inc.
+ */
+
+import Foundation
+import DatadogInternal
+import UIKit
+
+@_spi(Internal) public enum SessionReplayElement {
+    case specificElement (subtreeStrategy: SessionReplayNodeSubtreeStrategy, attributes: SessionReplayViewAttributes, builder: GenericTextWireframesBuilder)
+    case invisibleElement
+}
+
+@_spi(Internal) open class CustomNodeRecorder: NodeRecorder {
+    public let identifier = UUID()
+    
+    public init () {}
+
+    internal func semantics(of view: UIView, with attributes: ViewAttributes, in context: ViewTreeRecordingContext) -> NodeSemantics? {
+        let element = getSessionReplayElement(view: view, attributes: attributes, context: context)
+        switch element {
+        case .invisibleElement:
+            return InvisibleElement.constant
+        case .none:
+            return nil
+        case .specificElement(let subtreeStrategy, let attributes, let builder):
+            let nodes = [
+                Node(viewAttributes: attributes, wireframesBuilder: builder)
+            ]
+            return SpecificElement(subtreeStrategy: subtreeStrategy, nodes: nodes)
+        }
+    }
+
+    // will be overriden with actual implementation
+    open func getSessionReplayElement(
+        view: UIView,
+        attributes: SessionReplayViewAttributes,
+        context: SessionReplayViewTreeRecordingContext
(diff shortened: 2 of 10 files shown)


DatadogSessionReplay/Sources/SessionReplay+Internal.swift
```diff
@@ -0,0 +1,152 @@
...
+        self.borderWidth = borderWidth
+        self.backgroundColor = backgroundColor
+        self.cornerRadius = cornerRadius
+        self.opacity = opacity
+        self.view = view
+        self.nodeRecorder = nodeRecorder
+        self.wireframeRect = wireframeRect
+    }
+
+    internal func buildWireframes(with builder: WireframesBuilder) -> [SRWireframe] {
```
louiszawadzki, line 84: We cannot expose `SRWireframe` since it is declared internal in the Data models.
The idea here is to create a `NodeWireframesBuilder` class that is initialized with all the needed information for the builder to be called.

DatadogSessionReplay/Sources/SessionReplay+Internal.swift
```diff
@@ -0,0 +1,152 @@
...
+        case .specificElement(let subtreeStrategy, let attributes, let builder):
+            let nodes = [
+                Node(viewAttributes: attributes, wireframesBuilder: builder)
+            ]
+            return SpecificElement(subtreeStrategy: subtreeStrategy, nodes: nodes)
+        }
+    }
+
+    // will be overriden with actual implementation
+    open func getSessionReplayElement(
```
louiszawadzki, line 37: Maybe this function can be passed in the constructor instead of being overriden, I think it'll be a bit cleaner

DatadogSessionReplay/Sources/Processor/SRDataModelsBuilder/WireframesBuilder.swift
```diff
@@ -17,7 +17,7 @@ internal typealias WireframeID = NodeID
 /// It is used by the player to reconstruct individual elements of the recorded app UI.
 ///
 /// Note: `WireframesBuilder` is used by `Processor` on a single background thread.
-internal class WireframesBuilder {
+@_spi(Internal) public class SessionReplayWireframesBuilder {
```
maciejburda, line 20: `@_spi` comes with a risk, but it looks like some [other SDKs](https://github.com/search?q=repo%3Astripe%2Fstripe-ios+spi&type=code) are using it.

[Nice read from PSPDFKit folks](https://pspdfkit.com/blog/2023/swifts-approach-to-spi/) convinced me we could pursue that path.

DatadogSessionReplay/Sources/Processor/SRDataModelsBuilder/WireframesBuilder.swift
```diff
@@ -192,6 +192,8 @@ internal class WireframesBuilder {
     }
 }
 
+internal typealias WireframesBuilder = SessionReplayWireframesBuilder
```
maciejburda, line 195: worth adding note that it's done to have more unique name that's exposed through public-internal access level.

DatadogSessionReplay/Sources/SessionReplay+Internal.swift
maciejburda: Maybe it makes sense to make it a protocol with default implementation of `semantics(of` (via extension) in that case? 🤔

DatadogSessionReplay/Sources/SessionReplay+Internal.swift
```diff
@@ -0,0 +1,152 @@
...
+    public let borderWidth: CGFloat?
+    public let backgroundColor: CGColor?
+    public let cornerRadius: CGFloat?
+    public let opacity: CGFloat?
+    public let view: UIView
+    public let nodeRecorder: CustomNodeRecorder
+    
+    public let wireframeRect: CGRect
+    
+    public init(id: NodeIDGenerator, frame: CGRect, text: String, textAlignment: TextAlignment, clip: SessionReplayContentClip, textColor: CGColor?, font: UIFont, fontScalingEnabled: Bool, borderColor: CGColor?, borderWidth: CGFloat?, backgroundColor: CGColor?, cornerRadius: CGFloat?, opacity: CGFloat?, view: UIView, nodeRecorder: CustomNodeRecorder, wireframeRect: CGRect) {
```
maciejburda, line 65: worth having vertical formatting for the init

DatadogSessionReplay/Sources/SessionReplay+Internal.swift
maciejburda: I start to see that the design here is a little bit awkward on the native side. It almost asks for making exact copies of model, which you actually ended up doing, but it's just "hidden" behind `GenericWireframeBuilder`.

IMO worth brainstorming some architecture changes to make it scale better.

Saying that - for a single wireframe builder I think it's fine.

maciejburda (commented):
Great effort!

It's becoming quite a challenge, but I think you're on a right track with the solution.

louiszawadzki:
Closing this PR as I opened a new one

PR #1522, opened 2023-10-17
RUM-1040 Dispatch Continuation

maxep:
### What and why?

A short description of what changes this PR introduces and why.

### How?

A brief description of implementation details of this PR.

### Review checklist
- [ ] Feature or bugfix MUST have appropriate tests (unit, integration)
- [ ] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes

### Custom CI job configuration (optional)
- [x] Run unit tests for Core, RUM, Trace, Logs, CR and WVT
- [x] Run unit tests for Session Replay
- [x] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

PR #1541, opened 2023-11-07
RUM-1040 Stop Core Instance

maxep:
### What and why?

Add public interface for stopping core instance of the SDK. Stopping the core will deallocate all Features and their storage & upload units.

### How?

In parity with Android, the `stopInstance` static method will stop and unregister all feature of the core. Their storage and upload process will be deallocated so no event-write-context and upload will execute.

### Note

This change was a good opportunity to streamline flushing mechanism of the core (e.g. #1522). But the number of GDC queues involved in the core and Features made the task very tedious with high risk of introducing flakiness. I believe we should continue decreasing the number of queues, by using exclusion-queue as described in [WWDC 2017 - Modernizing Grand Central Dispatch Usage](https://developer.apple.com/videos/play/wwdc2017/706/), with clear and efficient queue hierarchy. **We should do so after building continuous benchmark to make sure we don't decrease performances.** 

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - [FIX] RUM session not being linked to spans. See [#1615][]
+- [FEATURE] Allow stopping a core instance. See [#1541][]
 
 # 2.6.0 / 09-01-2024
 - [FEATURE] Add `currentSessionID(completion:)` accessor to access the current session ID.
@@ -575,6 +576,7 @@ Release `2.0` introduces breaking changes. Follow the [Migration Guide](MIGRATIO
 [#1609]: https://github.com/DataDog/dd-sdk-ios/pull/1609
 [#1615]: https://github.com/DataDog/dd-sdk-ios/pull/1615
 [#1531]: https://github.com/DataDog/dd-sdk-ios/pull/1531
+[#1541]: https://github.com/DataDog/dd-sdk-ios/pull/1541
 [#1596]: https://github.com/DataDog/dd-sdk-ios/pull/1596
 [#1597]: https://github.com/DataDog/dd-sdk-ios/pull/1597
 [@00fa9a]: https://github.com/00FA9A
--- a/DatadogCore/Sources/Datadog.swift
+++ b/DatadogCore/Sources/Datadog.swift
@@ -302,8 +302,11 @@ public enum Datadog {
     }
 
     /// Stops the initialized SDK instance attached to the given name.
+    ///
+    /// Stopping a core instance will stop all current processes by deallocating all Features registered
+    /// in the core as well as their storage & upload units.
     /// 
-    /// - Parameter instanceName: The core instance name
+    /// - Parameter instanceName: the name of the instance to stop.
     public static func stopInstance(named instanceName: String = CoreRegistry.defaultInstanceName) {
         let core = CoreRegistry.unregisterInstance(named: instanceName) as? DatadogCore
         core?.stop()
--- a/DatadogCore/Sources/Datadog.swift
+++ b/DatadogCore/Sources/Datadog.swift
@@ -301,6 +301,14 @@ public enum Datadog {
         core?.clearAllData()
     }
 
+    /// Stops the initialized SDK instance attached to the given name.
+    /// 
+    /// - Parameter instanceName: The core instance name
+    public static func stopInstance(named instanceName: String = CoreRegistry.defaultInstanceName) {
+        let core = CoreRegistry.unregisterInstance(named: instanceName) as? DatadogCore
+        core?.stop()
+    }
+
     /// Initializes the Datadog SDK.
     ///
     /// You **must** initialize the core instance of the Datadog SDK prior to enabling any Product.
@@ -489,13 +497,10 @@ public enum Datadog {
 #endif
 
     internal static func internalFlushAndDeinitialize(instanceName: String = CoreRegistry.defaultInstanceName) {
-        assert(CoreRegistry.instance(named: instanceName) is DatadogCore, "SDK must be first initialized.")
-
+        // Unregister core instance:
+        let core = CoreRegistry.unregisterInstance(named: instanceName) as? DatadogCore
         // Flush and tear down SDK core:
-        (CoreRegistry.instance(named: instanceName) as? DatadogCore)?.flushAndTearDown()
-
-        // Deinitialize `Datadog`:
-        CoreRegistry.unregisterInstance(named: instanceName)
+        core?.flushAndTearDown()
     }
 }
 
--- a/DatadogCore/Tests/Datadog/DatadogCore/DatadogCoreTests.swift
+++ b/DatadogCore/Tests/Datadog/DatadogCore/DatadogCoreTests.swift
@@ -46,7 +46,6 @@ class DatadogCoreTests: XCTestCase {
             maxBatchesPerUpload: .mockRandom(min: 1, max: 100),
             backgroundTasksEnabled: .mockAny()
         )
-        defer { core.flushAndTearDown() }
 
         let requestBuilderSpy = FeatureRequestBuilderSpy()
         try core.register(feature: FeatureMock(requestBuilder: requestBuilderSpy))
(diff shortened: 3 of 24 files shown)


DatadogCore/Tests/Datadog/DatadogCore/DatadogCoreTests.swift
```diff
@@ -45,7 +45,6 @@ class DatadogCoreTests: XCTestCase {
             applicationVersion: .mockAny(),
             backgroundTasksEnabled: .mockAny()
         )
-        defer { core.flushAndTearDown() }
```
maciejburda, line 48: I can see a new test for stopping, but I'm wondering if we need removing these?

It seems like it was here to clean things up after each test.
Do you know the exact reason of having them? Maybe a way to go is to `defer { core.stop() }` now?

maciejburda (approved):
LGTM

DatadogCore/Tests/Datadog/DatadogCore/DatadogCoreTests.swift
maxep: This was a duplicate call, the core is flushed before assertion 👍

maxep:
This PR is finally ready for review, it has increased in size due to the added integration tests.

instrumented-tests/http-server-mock/python/start_mock_server.py
```diff
@@ -47,6 +47,14 @@ def do_GET(self):
...
         ])
 
+    def do_DELETE(self):
+        """
+        Routes all incoming DELETE requests
+        """
+        self.__route([
+            (r"/requests$", self.__DELETE_requests),
+        ])
+
```
maxep, line 57: I'm adding this endpoint to remove recorded requests so we can clear the mock server before restarting the core instance.

IntegrationTests/Runner/AppDelegate.swift
maxep, line 1: I have move the initialisation part to the `AppConfiguration.swift` so we can stop and re-initialiase the default core.

IntegrationTests/Runner/Scenarios/URLSession/URLSessionScenarios.swift
```diff
@@ -39,11 +39,6 @@ private class CompositedURLSessionDelegate: NSObject, URLSessionTaskDelegate, UR
     }
 }
 
-/// An example of instrumenting existing `URLSessionDelegate` with `DDURLSessionDelegate` through inheritance.
-private class CustomURLSessionDelegate: NSObject, URLSessionDataDelegate {
```
maxep, line 43: Moving to its own file with `internal` accessor for re-use in stop-core-instance test.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 - [FEATURE] Add `BatchProcessingLevel` configuration allowing to process more batches within single read/upload cycle. See [#1531][]
 - [FIX] Use `currentRequest` instead `originalRequest` for URLSession request interception
 - [FIX] Remove weak `UIViewController` references. See [#1597][]
+- [FEATURE] Allow stopping a core instance. See [#1541][]
```
maciejburda, line 6: please rebase after the release, so we put it in the correct place

DatadogCore/Sources/Datadog.swift
```diff
@@ -489,13 +497,10 @@ public enum Datadog {
 #endif
 
     internal static func internalFlushAndDeinitialize(instanceName: String = CoreRegistry.defaultInstanceName) {
-        assert(CoreRegistry.instance(named: instanceName) is DatadogCore, "SDK must be first initialized.")
```
maciejburda, line 492: no assert anymore?

DatadogCore/Sources/Datadog.swift
```diff
@@ -301,6 +301,14 @@ public enum Datadog {
         core?.clearAllData()
     }
 
+    /// Stops the initialized SDK instance attached to the given name.
```
maciejburda, line 304: Would be great to get a short documentation stating what happens to the data processed/stored by this instance.

DatadogCore/Sources/Datadog.swift
maxep: I don't remember why I removed it TBH! probably in an attempt to flush before stopping the core. Now, I don't see why we assert initialisation here 🤔

ganeshnj (commented):
This looks good, can we do some testing that no objects are being leaked when the core instance is stopped. Ideally we want to have memory graph something like
```
     ...................
..../                   \......
 start               stop
```

DatadogCore/Tests/Datadog/DatadogCore/DatadogCoreTests.swift
```diff
@@ -243,4 +240,40 @@ class DatadogCoreTests: XCTestCase {
...
+        let requestBuilderSpy = FeatureRequestBuilderSpy()
+        try core.register(feature: FeatureMock(requestBuilder: requestBuilderSpy))
+        let scope = try XCTUnwrap(core.scope(for: FeatureMock.name))
+
+        // When
+        scope.eventWriteContext { context, writer in
+            writer.write(value: FeatureMock.Event(event: "should not be sent"))
+        }
+
+        core.stop()
```
ncreated, line 268: question/ This looks flaky ❄️ , doesn't it? If it takes more than "min file age" from  writing an event to stopping the core, data will be uploaded. What is the rationale here?

ncreated:
> This looks good, can we do some testing that no objects are being leaked when the core instance is stopped. Ideally we want to have memory graph something like
> 
> ```
>      ...................
> ..../                   \......
>  start               stop
> ```

I agree with @ganeshnj ☝️ that we must test SDK allocations. However, I don't think the `stop()` model proposed in this PR delivers this promise. It is rather "hard stop" implementation that only deallocates the state without awaiting pending asynchronous operations. Correct me if I'm wrong @maxep

IntegrationTests/IntegrationScenarios/Scenarios/Core/StopCoreScenarioTests.swift
```diff
@@ -0,0 +1,185 @@
...
+
+        assertRUM(requests: recordedRequests)
+
+        // ignore telemetry and application start view events
+        let eventMatchers = try recordedRequests
+            .flatMap { request in try RUMEventMatcher.fromNewlineSeparatedJSONObjectsData(request.httpBody) }
+            .filterTelemetry()
+            .filterApplicationLaunchView()
+
+        let session = try XCTUnwrap(RUMSessionMatcher.groupMatchersBySessions(eventMatchers).first)
```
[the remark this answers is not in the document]
maxep: It doesn't work sadly, because I run the scenario 3 times:
1. After first init
2. After stopping the core
3. After second init

The assertion aims to cover the first and last run: the first run will have the `application-launch/view` but not the 3rd. 
Also, I'm not triggering the end-session screens because I need to replay the scenario :/ 

I can still try to clean up by having 2 assertion methods for RUM.

DatadogCore/Tests/Datadog/DatadogCore/DatadogCoreTests.swift
maxep: Indeed, I can remove the first EWC. Rationale was to assert that no upload is triggered after stopping the core, event with event in the pipe.

DatadogCore/Sources/Datadog.swift
```diff
@@ -301,6 +301,17 @@ public enum Datadog {
...
+    ///
+    /// Stopping a core instance will stop all current processes by deallocating all Features registered
+    /// in the core as well as their storage & upload units.
+    /// 
+    /// - Parameter instanceName: the name of the instance to stop.
+    public static func stopInstance(named instanceName: String = CoreRegistry.defaultInstanceName) {
+        let core = CoreRegistry.unregisterInstance(named: instanceName) as? DatadogCore
+        core?.stop()
```
ncreated, line 312: question/ Why not making `stop()` a requirement in `DatadogCoreProtocol`? I have no opinion on this (both may be fine), just wondering on what was the rationale.

ncreated (approved):
Well done 👌. Great work on fixing the memory profile 💪.

ganeshnj (approved):

DatadogCore/Sources/Datadog.swift
maxep: The stop is only required from high level apis, there is no usage from Features so there no need for making it part of the protocol. The requirement might change in the future tho!

PR #1542, opened 2023-11-07
RUM-1335 Add `sessionPrecondition` to RUM events

ncreated:
### What and why?

📦 As part of **RUM Session Standarization** project, this PR introduces `sessionPrecondition` attribute to RUM events. This is to add more visibility over the reasons for creating RUM sessions.
```swift
enum RUMSessionPrecondition {
    case userAppLaunch
    case inactivityTimeout
    case maxDuration
    case backgroundLaunch
    case prewarm
    case fromNonInteractiveSession    // not implemented in this PR - TODO in RUM-1646
    case explicitStop
}
```

Mid-term goal is to simplify the logic of RUM Sessions creation and pay off portions of tech debt incurred in recent https://github.com/DataDog/dd-sdk-ios/pull/1278, https://github.com/DataDog/dd-sdk-ios/pull/1219 and https://github.com/DataDog/dd-sdk-ios/pull/1290 and previous work.

### How?

`RUMApplicationScope` is the main component modified in this PR. 

The idea behind this work is to:
* Avoid doing changes in RUM Session creation logic - we want to first add observability to the current state of affair.
* Add `sessionPrecondition` attribute to all RUM events.
* Insert sanity checks and error telemetry to validate assumptions and eventually erase ambiguity and pieces of technical debt in future increments.

Counterpart PR to RUM schema: https://github.com/DataDog/rum-events-format/pull/169

### Review checklist

- [x] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogRUM/Sources/RUMMonitor/Scopes/RUMApplicationScope.swift
+++ b/DatadogRUM/Sources/RUMMonitor/Scopes/RUMApplicationScope.swift
@@ -56,7 +56,7 @@ internal class RUMApplicationScope: RUMScope, RUMContextProvider {
         // Added in https://github.com/DataDog/dd-sdk-ios/pull/1278 to ensure that logs and traces
         // can be correlated with valid RUM session id (even if occurring before any user interaction).
         if command is RUMSDKInitCommand {
-            createInitialSession(with: context)
+            createInitialSession(with: context, on: command)
             return true // always keep application scope
         }
 
@@ -66,7 +66,7 @@ internal class RUMApplicationScope: RUMScope, RUMContextProvider {
         if sessionScopes.isEmpty && !applicationActive {
             // This flow is likely stale code as`RUMSDKInitCommand` should already start the session before reaching this point
             dependencies.telemetry.debug("Starting initial session from lazy flow")
-            createInitialSession(with: context)
+            createInitialSession(with: context, on: command)
         }
 
         // Create the application launch view on any command
@@ -125,15 +125,15 @@ internal class RUMApplicationScope: RUMScope, RUMContextProvider {
 
     // MARK: - Private
 
-    /// Sanity flag to make sure initial session is created only once.
-    private var didCreateInitialSession = false
+    /// Sanity count to make sure initial session is created only once.
+    private var didCreateInitialSessionCount = 0
 
     /// Starts initial RUM Session.
-    private func createInitialSession(with context: DatadogContext) {
-        if didCreateInitialSession { // Sanity check
-            dependencies.telemetry.error("Initial session was created more than once (previous end reason: \(lastSessionEndReason?.rawValue ?? "unknown"))")
+    private func createInitialSession(with context: DatadogContext, on command: RUMCommand) {
+        if didCreateInitialSessionCount > 0 { // Sanity check
+            dependencies.telemetry.error("Creating initial session \(didCreateInitialSessionCount) extra time(s) due to \(type(of: command)) (previous end reason: \(lastSessionEndReason?.rawValue ?? "unknown"))")
         }
-        didCreateInitialSession = true
+        didCreateInitialSessionCount += 1
 
         var startPrecondition: RUMSessionPrecondition? = nil
 
@@ -191,6 +191,11 @@ internal class RUMApplicationScope: RUMScope, RUMContextProvider {
             startPrecondition = .explicitStop
         } else {
             dependencies.telemetry.error("Failed to determine session precondition for NEW session with end reason: \(lastSessionEndReason?.rawValue ?? "unknown"))")
+        }
+
+        if didCreateInitialSessionCount > 0 { // Sanity check
+            // We assume this is not an initial session in the app (such is started with `RUMSDKInitCommand`:
+            dependencies.telemetry.error("Starting NEW session on due to \(type(of: command)), but initial sesison never existed")
         }
 
         let resumingViewScope = command is RUMStartViewCommand ? nil : lastActiveView
--- a/DatadogRUM/Sources/RUMMonitor/Scopes/RUMApplicationScope.swift
+++ b/DatadogRUM/Sources/RUMMonitor/Scopes/RUMApplicationScope.swift
@@ -131,14 +131,25 @@ internal class RUMApplicationScope: RUMScope, RUMContextProvider {
     /// Starts initial RUM Session.
     private func createInitialSession(on command: RUMCommand, context: DatadogContext, writer: Writer) {
         if didCreateInitialSession { // Sanity check
-            dependencies.telemetry.error("Initial session was created more than once")
+            dependencies.telemetry.error("Initial session was created more than once (previous end reason: \(lastSessionEndReason?.rawValue ?? "unknown"))")
         }
         didCreateInitialSession = true
+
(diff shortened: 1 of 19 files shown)


DatadogRUM/Sources/RUMMonitor/Scopes/RUMApplicationScope.swift
```diff
@@ -85,75 +93,118 @@ internal class RUMApplicationScope: RUMScope, RUMContextProvider {
...
 
+    /// Starts new RUM Session some time after previous one was ended with ``RUMMonitorProtocol.stopSession()`` API. It may re-activate the last view from previous session.
     private func startNewSession(on command: RUMCommand, context: DatadogContext, writer: Writer) {
+        var startPrecondition: RUMSessionPrecondition? = nil
+
+        if lastSessionEndReason == .stopAPI {
+            startPrecondition = .explicitStop
+        } else {
+            dependencies.telemetry.error("Failed to determine session precondition for NEW session with end reason: \(lastSessionEndReason?.rawValue ?? "unknown"))")
+        }
```
maxep, line 194: /question The condition for starting a new session is `activeSession == nil && command.isUserInteraction`, so shouldn't we set the precondition to `.userAppLaunch` by default instead?

DatadogRUM/Tests/RUMMonitor/Scopes/RUMApplicationScopeTests.swift
```diff
@@ -25,33 +37,27 @@ class RUMApplicationScopeTests: XCTestCase {
         XCTAssertNil(scope.context.activeUserActionID)
     }
 
-    func testWhenFirstEventIsReceived_itStartsNewSession() throws {
+    func testWhenInitialized_itStartsNewSession() throws {
```
maxep, line 40: /suggestion shouldn't it be 2 tests instead of a replacement? to keep testing [this statement](https://github.com/DataDog/dd-sdk-ios/pull/1542/files#diff-2a0d3031da8cdd2c56293e0221ce7a4a889ca70bf0c90f9ffad664d0e7cb1559R66).

DatadogRUM/Tests/RUMMonitor/Scopes/RUMApplicationScopeTests.swift
```diff
@@ -208,61 +214,65 @@ class RUMApplicationScopeTests: XCTestCase {
...
         // Given
         let currentTime = Date()
-        let scope = RUMApplicationScope(
+        let scope = createRUMApplicationScope(
             dependencies: .mockWith(
                 sessionSampler: .mockKeepAll()
             )
         )
-        XCTAssertTrue(scope.process(
+        _ = scope.process(
```
maxep, line 222: Why not continuing to assert the return value?

maxep (approved):
Nice work, the logic is complex but reads very well!
I've left some minor comments/questions.

DatadogRUM/Sources/RUMMonitor/Scopes/RUMApplicationScope.swift
```diff
@@ -85,75 +93,118 @@ internal class RUMApplicationScope: RUMScope, RUMContextProvider {
...
-    }
+    /// Sanity flag to make sure initial session is created only once.
+    private var didCreateInitialSession = false
+
+    /// Starts initial RUM Session.
+    private func createInitialSession(with context: DatadogContext) {
+        if didCreateInitialSession { // Sanity check
+            dependencies.telemetry.error("Initial session was created more than once (previous end reason: \(lastSessionEndReason?.rawValue ?? "unknown"))")
+        }
+        didCreateInitialSession = true
```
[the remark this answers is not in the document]
ncreated: Fair call 👍, I'll add more verbosity to this error.

DatadogRUM/Sources/RUMMonitor/Scopes/RUMApplicationScope.swift
ncreated: This is one of the tricks in the current implementation. This flow is reached due to `stopSession()` API, as proven in [this test](https://github.com/DataDog/dd-sdk-ios/blob/8aab1db2b7043435515335d45c5188825665396a/DatadogRUM/Tests/RUMMonitor/Scopes/RUMApplicationScopeTests.swift#L506-L528). The condition has two sides and while the rhs means user interaction, the lhs guards "there is no more active session". Because we start initial session with SDK start (https://github.com/DataDog/dd-sdk-ios/pull/1278) this can be effectively reached only for later sessions.

But you touched something with this question - we assume "this is not initial session". I will add telemetry to prove if this assumption is wrong.

DatadogRUM/Tests/RUMMonitor/Scopes/RUMApplicationScopeTests.swift
ncreated: Rationale here is: there is no such case (in production code) where initial session is started with first interaction - after https://github.com/DataDog/dd-sdk-ios/pull/1278 we always start it with SDK init. This test was covering non-existing behaviour, hence I aligned it with reality. How does it sound?

DatadogRUM/Tests/RUMMonitor/Scopes/RUMApplicationScopeTests.swift
ncreated: The [contract from `RUMScope` interface](https://github.com/DataDog/dd-sdk-ios/blob/8aab1db2b7043435515335d45c5188825665396a/DatadogRUM/Sources/RUMMonitor/RUMScope.swift#L10-L15) is to use return type for signalling if a scope should be kept or released by parent scope. Because `RUMApplicationScope` is the root scope retained by `Monitor`, [we ignore its return type in `Monitor`](https://github.com/DataDog/dd-sdk-ios/blob/ncreated/RUM-1335/rum-session-start-reason/DatadogRUM/Sources/RUMMonitor/Monitor.swift#L138):

```swift
_ = self.scopes.process(command: transformedCommand, context: context, writer: writer)
```

Because we don't care about its return value, asserting this in tests was giving false sense of coverage. WDYT?

DatadogRUM/Tests/RUMMonitor/Scopes/RUMApplicationScopeTests.swift
maxep: Sounds fair, but we are in `RUMApplicationScope`'s unit tests and not in the RUM/Monitor tests. IMHO, we could keep the test and not assume behaviours from upper layers. But it's really minor and I'm fine keeping this change 👍

maxep (approved):

DatadogRUM/Tests/RUMMonitor/Scopes/RUMApplicationScopeTests.swift
ncreated: I'm gonna keep it then 👍. My additional rationale is: we want `RUMApplicationScope` tests reflect existing production logic, so when changing the implementation later (as part of Session Standarization project) tests will guard behaviours that must be supported. Keeping "first event starts session" would be a false guard in that sense.

maxep (approved):

PR #1547, opened 2023-11-08
Resource endpoint support POC

maciejburda:
Draft PR presenting proposed solution

*Don't merge*


DatadogSessionReplay/Sources/Recorder/Utilities/ImageDataProvider.swift
```diff
@@ -89,14 +89,18 @@ import CryptoKit
...
+
+            let hash = computeHash()
+            objc_setAssociatedObject(self, &customHashKey, hash, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
             return hash
         }
-
-        let hash = computeHash()
-        objc_setAssociatedObject(self, &customHashKey, hash, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
-        return hash
     }
```
ncreated, line 104: Which thread will this be called on? I see no use of `customHash` in this draft PR.

maciejburda:
@ncreated Thanks a lot for taking a look!

Additional context:
POC is missing two "bigger" bits:
- ResourceProcessor (calling a customHash and data on a separate thread and writing it to a batch)
- Resource Request Builder

Technically speaking - Wireframe builders will be calling customHash simultaneously with ResourceProcessor

maciejburda:
And obviously using tuple is a streach for the sake of POC. In the actual implementation I will probably do some renaming and provide a tiny struct that would carry both arrays

DatadogSessionReplay/Sources/Feature/RequestBuilders/ResourceRequestBuilder.swift
```diff
@@ -0,0 +1,68 @@
...
+    func request(for events: [Event], with context: DatadogContext) throws -> URLRequest {
+        let decoder = JSONDecoder()
+        let resources = try events.map { event in
+            try decoder.decode(CodableResource.self, from: event.data)
+        }
+        return try createRequest(resources: resources, context: context)
+    }
+
+    private func createRequest(resources: [CodableResource], context: DatadogContext) throws -> URLRequest {
+        var multipart = multipartBuilder
```
fuzzybinary, line 28: Does the assignment to the local here perform a copy, or does this method have a side effect of modifying `multipartBuilder`?

PR #1548, opened 2023-11-08
Enable custom node recorders for cross platform

louiszawadzki:
### What and why?

Enable cross platform SDKs to provide additional node recorders to be used for session replay

### How?

I've added a layer to hide `SRWireframe` so that all the needed APIs could be made public.

I've added a few tests to check that additional node recorders are indeed passed down to the snapshot builder.

I've split some refactoring/fix commits to avoid bringing too much noise in commits where actual features happen.

`@_spi` is still hidden but seems to be used already: https://pspdfkit.com/blog/2023/swifts-approach-to-spi/
Although it is supported since XCode 12, would there be (reasonable) cases where users of the SDK would not be able to use the SDK because of this attribute?

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogSessionReplay/Sources/Processor/Processor.swift
+++ b/DatadogSessionReplay/Sources/Processor/Processor.swift
@@ -84,6 +84,7 @@ internal class Processor: Processing {
         let wireframes: [SRWireframe] = flattenedNodes
             .map { node in node.wireframesBuilder }
             .flatMap { nodeBuilder in nodeBuilder.buildWireframes(with: wireframesBuilder) }
+            .map { wireframe in wireframe.toSRWireframe() }
 
         #if DEBUG
         interceptWireframes?(wireframes)
--- a/DatadogSessionReplay/Sources/Processor/SRDataModelsBuilder/Wireframes.swift
+++ b/DatadogSessionReplay/Sources/Processor/SRDataModelsBuilder/Wireframes.swift
@@ -1,0 +1,404 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+#if os(iOS)
+import UIKit
+
+/// The border properties of this wireframe. The default value is null (no-border).
+@_spi(Internal) public struct ShapeBorder {
+    /// The border color as a String hexadecimal. Follows the #RRGGBBAA color format with the alpha value as optional.
+    public let color: String
+
+    /// The width of the border in pixels.
+    public let width: Int64
+
+    internal func toSRShapeBorder() -> SRShapeBorder {
+        return .init(color: color, width: width)
+    }
+}
+
+/// Schema of clipping information for a Wireframe.
+@_spi(Internal) public struct ContentClip {
+    /// The amount of space in pixels that needs to be clipped (masked) at the bottom of the wireframe.
+    public let bottom: Int64?
+
+    /// The amount of space in pixels that needs to be clipped (masked) at the left of the wireframe.
+    public let left: Int64?
+
+    /// The amount of space in pixels that needs to be clipped (masked) at the right of the wireframe.
+    public let right: Int64?
+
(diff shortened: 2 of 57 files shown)


DatadogSessionReplay/Sources/Processor/Processor.swift
```diff
@@ -84,6 +84,7 @@ internal class Processor: Processing {
         let wireframes: [SRWireframe] = flattenedNodes
             .map { node in node.wireframesBuilder }
             .flatMap { nodeBuilder in nodeBuilder.buildWireframes(with: wireframesBuilder) }
+            .map { wireframe in wireframe.toSRWireframe() }
```
maciejburda, line 87: `.toSRWireframe()` could be done directly in the flatmap

DatadogSessionReplay/Sources/Processor/SRDataModelsBuilder/Wireframes.swift
```diff
@@ -0,0 +1,404 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+#if os(iOS)
+import UIKit
+
+/// The border properties of this wireframe. The default value is null (no-border).
+@_spi(Internal) public struct ShapeBorder {
```
maciejburda, line 11: I'm wondering if it would make more sense to expose it as a protocol that SR objects are implementing 🤔

Slightly less boilerplate.

maciejburda (approved):
Looking good! Just some tiny suggestion on considering protocol instead of structs

DatadogSessionReplay/Sources/Processor/SRDataModelsBuilder/Wireframes.swift
louiszawadzki: Could you elaborate a bit as to what this would look like and what we'd gain from it? 
It's a bit unclear for me.

maciejburda: I guess inits "for free"

maciejburda (approved):
Great job! LGTM!

DatadogSessionReplay/Sources/Processor/SRDataModelsBuilder/Wireframes.swift
maxep, line 1: /question I don't actually see the need for this mapping, why not directly exposing the `SR*` types with `@_spi`?
You will need to update the model generator, but we would avoid having to write boilerplate code, no?

maxep (commented):
Great work @louiszawadzki! the `@_spi` will be very practical for cross-platform 👍 
I've left one question about the wireframe mapping, I think we can avoid it by updating the model generator.

DatadogSessionReplay/Sources/Processor/SRDataModelsBuilder/Wireframes.swift
maciejburda: The idea is to avoid exposing model from schema directly

maxep: But it won't be directly exposed with `@_spi`. Having it generated will be more scalable as well, if we want to use more types in cross-platform 👍

louiszawadzki: I will look into it tomorrow to see what this option would look like :) 
Adding `@_spi` might still require some other changes as well.

I am also not a big fan of exposing directly the model to the other SDKs, but if we have enough control on it we might not have too many downsides.

louiszawadzki: If that solution requires less code and is simpler to understand that might indeed be a good idea

louiszawadzki:
Closing this one as I went down a different implementation here: https://github.com/DataDog/dd-sdk-ios/pull/1559

PR #1562, opened 2023-11-21
Add GPG signing to dogfooding script commit

maciejburda:
### What and why?

Adds required GPG signing to dogfooding's automated commits.


maciejburda:
This solution doesn't work

PR #1563, opened 2023-11-21
Update Session Replay default maxID

maciejburda:
### What and why?

Updates maximum Session Replay ID to not exceed 32-bit. It's an overlooked requirement needed by the player to work correctly. Potentially we'll replace ids with String type in the future.


xgouchet (approved):

mquentin (approved):
LGTM thanks.

Question: Do you see any other id or number property that would not be maxed at 32bit ?

maciejburda:
@mquentin Great question! This is Session Replay wide id generator, it should affect everything in SR.

mquentin:
> @mquentin Great question! This is Session Replay wide id generator, it should affect everything in SR.

Ah great so all ids will be generated with this generator : nice !

PR #1564, opened 2023-11-21
RUM-1040 Fix unswizzling

maxep:
### What and why?

While we allow applying multiple swizzling to a method, we don't properly keep track the hierarchy, preventing unswizzling to fully remove implementation from the stack.

### How?

Use a linked list for each swizzled method to properly unswizzle and clean the callstack.

### Review checklist

- [x] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


DatadogInternal/Sources/Swizzling/MethodSwizzler.swift
```diff
@@ -6,127 +6,160 @@
...
-        /// NOTE: RUMM-452 as we never add/remove methods/classes at runtime,
-        /// search operation doesn't have to wrapped in sync {...} although it's visible in the interface
-        var headKlass: AnyClass? = klass
-        while let someKlass = headKlass {
-            if let foundMethod = findMethod(with: selector, in: someKlass) {
-                return FoundMethod(method: foundMethod, klass: someKlass)
-            }
-            headKlass = class_getSuperclass(headKlass)
-        }
-        throw InternalError(description: "\(NSStringFromSelector(selector)) is not found in \(NSStringFromClass(klass))")
```
maxep, line 54: I don't see any reason for doing this while the objc runtime function `class_getInstanceMethod` is doing the same thing.

DatadogRUM/Tests/Instrumentation/RUMInstrumentationTests.swift
```diff
@@ -24,8 +24,8 @@ class RUMInstrumentationTests: XCTestCase {
         // Then
         withExtendedLifetime(instrumentation) {
             DDAssertActiveSwizzlings([
-                "UIViewController.viewDidAppear:",
-                "UIViewController.viewDidDisappear:",
```
maxep, line 28: We are no longer keeping a reference to the class, only the method pointer.

DatadogSessionReplay/Sources/Recorder/TouchSnapshotProducer/UIApplicationSwizzler.swift
```diff
@@ -60,121 +56,4 @@ internal final class UIApplicationSwizzler {
         }
     }
 }
-
-internal class MethodSwizzler<TypedIMP, TypedBlockIMP> {
```
maxep, line 64: Session Replay will now use the shared `MethodSwizzler`.

DatadogInternal/Sources/Swizzling/MethodSwizzler.swift
maciejburda: Seems like it was added to have this nicely formatted `InternalError`. I think that `class_getInstanceMethod` doesn't throw, but just returns optional.

DatadogInternal/Sources/Swizzling/MethodSwizzler.swift
```diff
@@ -6,127 +6,160 @@
...
+    let closure: Any
+    init(_ closure: Any) {
+        self.closure = closure
     }
+}
 
-    public init() { }
+open class MethodSwizzler<Signature, Override> {
+    /// List of swizzling managed by this instance.
+    private var overrides: [(method: Method, `override`: OverrideBox)] = []
```
maciejburda, line 67: Would it make sense to add method to `OverrideBox`?

Otherwise maybe worth having a struct instead of tuple.

DatadogRUM/Tests/Instrumentation/RUMInstrumentationTests.swift
maciejburda: Does it change anything?

What if we have the same method signature coming from different objects? 🤔

maciejburda (changes requested):
Great effort on clean up, adding documentation and making things right! 💪 

I left some _potentially_ blocking comments. Let me know your thoughts 🙌

maciejburda (changes requested):
Great effort on clean up, adding documentation and making things right! 💪 

I left some _potentially_ blocking comments. Let me know your thoughts 🙌

DatadogInternal/Sources/Swizzling/MethodSwizzler.swift
maxep: Not just that, previously it was using `class_copyMethodList` and looked at super-class if not found.
But the `class_getInstanceMethod` doc says specifically:
```
@note This function searches superclasses for implementations, whereas \c class_copyMethodList does not.
```

DatadogRUM/Tests/Instrumentation/RUMInstrumentationTests.swift
maxep: No, the signature could be the same for different objects but the method pointer will be different 👍 
The `class` was only use for those assertions but is not necessary for swizzling.

DatadogInternal/Sources/Swizzling/MethodSwizzler.swift
maciejburda: does it make sense to use it instead and throw this exception when not found?

DatadogInternal/Sources/Swizzling/MethodSwizzler.swift
maxep: The `OverrideBox` is only a reference to a closure, and it is part of the `MethodSwizzling` linked-list so I would prefer not to have the same reference to the `Method` repeated in each element of the list.
Could you elaborate on the struct vs. tuple?  to me they are essentially the same thing, I could do a typealias for readability?

DatadogInternal/Sources/Swizzling/MethodSwizzler.swift
maxep: Yes, I have added the `dd_class_getInstanceMethod` as part of the PR. It's at the bottom of the file 👍

DatadogInternal/Tests/Swizzling/MethodSwizzlerTests.swift
```diff
@@ -41,86 +37,118 @@ class MethodSwizzlerTests: XCTestCase {
...
+        class EmptySubSubclass: EmptySubclass { }
+        XCTAssertNotNil(try dd_class_getInstanceMethod(EmptySubclass.self, selToSwizzle))
+        XCTAssertNotNil(try dd_class_getInstanceMethod(EmptySubSubclass.self, selToSwizzle))
+    }
+
     func test_swizzle_alreadySwizzledSelector() throws {
-        let foundMethod = try Swizzler.findMethod(with: selToSwizzle, in: BaseClass.self)
+        let method = try dd_class_getInstanceMethod(BaseClass.self, selToSwizzle)
+        let obj = BaseClass()
+        let before_imp = method_getImplementation(method)
```
ganeshnj, line 71: We need to increase the coverage here
- Make sure all implementation are called including the original implementation
 - Lets say first implementation returns "first", second returns first imp + " second", third returns second + " third", then you should have "first second third" as returned value in the end.
- unswizzling something that is in the middle, like second swizzling in the above example

DatadogInternal/Sources/Swizzling/MethodSwizzler.swift
```diff
@@ -6,127 +6,160 @@
...
+                origin: org_imp,
+                override: override,
+                parent: swizzlings[method]
+            )
+
+            method_setImplementation(method, ovr_imp)
         }
     }
 
     /// Removes swizzling and resets the method to its original implementation.
```
ganeshnj, line 96: docs need update here now

DatadogInternal/Sources/Swizzling/MethodSwizzler.swift
```diff
@@ -6,127 +6,160 @@
...
     public func unswizzle() {
```
ganeshnj, line 97: I had hard time understanding this, IMO it would be cleaner to read if implemented without recursion.

When you unswizzle, you pop the stack until you find the target, and then call the swizzle method again on the type.

ganeshnj (commented):
Nice effort on making the implementation clear and perhaps the most accurate I have seen around swizzling.

DatadogInternal/Sources/Swizzling/MethodSwizzler.swift
maxep: It was even harder to implement 😄 I agree on readability, I will see if I can avoid recursion!

DatadogInternal/Sources/Swizzling/MethodSwizzler.swift
maciejburda: Yeah - I guess is readablity only. `typealias` is also fine.

DatadogInternal/Sources/Swizzling/MethodSwizzler.swift
ganeshnj: Sounds good, happy with current set of tests.

ganeshnj (approved):
Thanks for updating

maciejburda (approved):

PR #1565, opened 2023-11-22
RUM-2150 Resource Request Builder

maciejburda:
### What and why?

Adds Resource request builder to Session Replay. Since it's second Request Builder it also renames the already existing one to `SegmentRequestBuilder` making clear distinction in the responsibilities.

Right now it's detached from any working code. Follow up PRs will change that, eventually adding code-level feature flag that will be enabled once backend in on production.

### How?

It's mostly following existing Session Replay setup. It could use some common parent class for both request builders, but I decided to keep them separate in case we diverge in the future.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


maxep (commented):
Looks great, all clear 👌 
I've few questions for clarification before approving

DatadogSessionReplay/Sources/Writer/Models/EnrichedResource.swift
```diff
@@ -0,0 +1,42 @@
...
+        let type: String
+        let application: Application
+
+        init(_ applicationId: String) {
+            self.type = "resource"
+            self.application = .init(id: applicationId)
+        }
+    }
+    internal var identifier: String
+    internal var data: Data
```
ganeshnj, line 25: How data is represented over the wire?

DatadogSessionReplay/Sources/Writer/Models/EnrichedResource.swift
```diff
@@ -0,0 +1,42 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+#if os(iOS)
+import Foundation
+
+/// Extends the resource information with context.
+internal struct EnrichedResource: Codable, Resource {
```
ganeshnj, line 11: Do you need to serialize and deserialize both?

If not, let's not codegen unnecessary for both.

ganeshnj (changes requested):
Looks good, a quick suggestion and question.

DatadogSessionReplay/Sources/Writer/Models/EnrichedResource.swift
maciejburda: @maxep Let me answer your question here.

It's added to a form this way:
```swift
resources.forEach {
            multipart.addFormData(
                name: "image",
                filename: $0.identifier,
                data: $0.data,
                mimeType: "image/png"
            )
        }
```

both filename and data are default form fields, so it's basically travelling over the wire in the multipart form as an file attachment, with it filename equal to the identifier.

DatadogSessionReplay/Sources/Writer/Models/EnrichedResource.swift
maciejburda: Yeah - it's serialized to a batch when writing, and deserialized before sending.

DatadogSessionReplay/Sources/Writer/Models/EnrichedResource.swift
ganeshnj: This almost seems like a browser API than a native client. I expect image uploading a POST call with body as raw bytes.

Did we consider this option?

DatadogSessionReplay/Sources/Writer/Models/EnrichedResource.swift
ganeshnj: We must not deserialize before sending, for sending the data written to disk, we should consider them raw bytes like we do any other events.

Although, I agree we deserialize for filtering but that's just the metadata, not the real payload.

maxep: I'm afraid we have no other way to store images using current api of the core, we need to serialize the image in a json and deserialise it before upload. Which mean doing base64 encoding/decoding for each one of them.
@maciejburda did you take a look at cpu overhead?

maciejburda: Yeah - for session replay it's been different since the beginning because we rely on the multipart form in a certain way.

It's on our radar to change it, but it's actually not trivial.

DatadogSessionReplay/Sources/Writer/Models/EnrichedResource.swift
maciejburda: Sorry, for missing the comment. Sent you the RFC :)

PR #1566, opened 2023-11-22
RUM-1040 Fix `URLSessionTask` interception 

maxep:
### What and why?

After first configuration, our new Network Instrumentation logic introduced in #1394 is intercepting all `URLSessionTask` regardless of their delegate.

Starting an interception can result in Start Resource in RUM. When the custom delegate is not in play, the interception will never finish and so the RUM Resource will never stop. This means that the resource event won't be transmitted, but it will keep the Resource scopes active and therefore the view scopes open.

### How?

### Review checklist

- [x] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


maxep:
Closing in favor of #1588

PR #1567, opened 2023-11-23
RUM-1820 Create full snapshot if root has changed

louiszawadzki:
### What and why?

I noticed a bug in Session Replay when the root wireframe (first wireframe in the list) is added in an incremental snapshot.

This wireframe will have no `previousId`, and the replay will add it last in the list of wireframes.

This should not happen very often, given we create a full snapshot on every RUM view.

Possible improvements for the PR:
- if the new root is already present in the previous wireframe, there is no need to create a new full snapshot

Also it's worth investigating why we're removing root wireframes in the first place (and thus adding them back afterwards).

### Review checklist

- [x] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


maciejburda:
Probably worth considering this change. WDYT @louiszawadzki @ncreated ?

maciejburda (approved):

PR #1569, opened 2023-11-29
RUM-2151 Resource feature

maciejburda:
### What and why?

Add feature separation between Segments and Resources in Session Replay.

### How?

It's done by adding a new feature registered alongside with the old one within Session Replay. New features leverages previously added `RecordsRequestBuilder`. It also adds additional writer split (Resource one, not yet used)

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


DatadogInternal/Sources/DatadogFeature.swift
```diff
@@ -37,7 +37,7 @@ public protocol DatadogFeature {
     ///
     /// The `FeatureMessageReceiver` defines an interface for Feature to receive any message
     /// from a bus that is shared between Features registered in a core.
-    var messageReceiver: FeatureMessageReceiver { get }
+    var messageReceiver: FeatureMessageReceiver? { get }
```
maciejburda, line 40: Similarly to performanceOverride, not all feature are initially interested in using messageReceiver.

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
```diff
@@ -29,7 +29,7 @@ internal final class NetworkInstrumentationFeature: DatadogFeature {
     )
 
     /// A no-op message bus receiver.
-    internal let messageReceiver: FeatureMessageReceiver = NOPFeatureMessageReceiver()
+    internal let messageReceiver: FeatureMessageReceiver? = NOPFeatureMessageReceiver()
```
ncreated, line 32: suggestion/  Now we have two ways of achieving the same: either setting `messageReceiver` to `nil` or to `NOP*` will have the same result. This would be error-prone on the long term as `!= nil` check won't reliably determine if the feature can receive messages or not. I'd suggest allowing either way, not both.

ncreated (approved):

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
maxep: I agree! I think the `NOP*` pattern can stay IMHO, but either way is fine.
Additionally, I think the `performanceOverride` is actually mis-placed and should be in the `DatadogRemoteFeature`.
Additionally 2, the `DatadogFeatureConfiguration` defined above is a leftover from the migration and no longer in use.

maxep (dismissed):
Looks great!
 I agree with @ncreated on the `nop` vs `nil`, lets address it before merging 👍

ncreated (approved):
💯

PR #1570, opened 2023-11-29
[POC] Wrap core with core proxy

louiszawadzki:
### What and why?

This proxy enables to listen to all messages that are written to the core.
We can then retrieve them and make assertions for easier E2E tests.

This PR is only for the POC, this is not mergeable as it is right now. 
Its goal is to enable discussion for an actual implementation.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


louiszawadzki:
This POC is being integrated in https://github.com/DataDog/dd-sdk-ios/pull/1629

PR #1578, opened 2023-12-06
Add OSLog support for debug logging

maciejburda:
### What and why?

Adds `OSLog` support for iOS14+ targets for internal SDK logging. It improves readability of the log and allows browsing the logs in Console application.

### How?

Mostly by updating the console print closure.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


maciejburda:
It's a proposal from small hackathon.

Tests TBD if we agree on the direction.

ncreated (changes requested):
Overall it looks interesting but OSLog cannot be seen as a simple replacement to `print()` function as proposed in this PR. Unlike `print()`, logs produced with `OSLog` will be seen by other people in system console. For that reason sensitive information should be redacted using [`OSLogPrivacy`](https://developer.apple.com/documentation/os/oslogprivacy). In context of our SDK sensitive information may include credentials, device information, event attributes or other aspects of user data.

While this PR does interesting quick attempt on the subject, we must consider privacy in order to consider it truly.

maciejburda:
@ncreated Good points! I thought of that - and apparently default value for OSLog is `.private`. Which redacts all the information. We can also make it explicit if we want.

maciejburda:
As described here:
https://stackoverflow.com/questions/62901969/os-logger-os-log-ignores-the-new-privacy-string-interpolation/62903401
https://www.avanderlee.com/debugging/oslog-unified-logging/

When debugger is on, data is always visible in the console app:
Screenshot 2024-01-08 at 15 34 26

I'll run short experiment on a device to prove that it's being properly redacted when logging is enabled on prod by mistake.

maciejburda:
Screenshot when running directly on the device in the release mode:
Screenshot 2024-01-08 at 15 59 12

maxep (dismissed):
Looks good 👍  I see great opportunity for core-specific logging!

DatadogInternal/Sources/DD.swift
```diff
@@ -25,5 +25,25 @@ public struct DD {
...
+        }
+    } else {
+        print(message)
+    }
+}
+
+#if canImport(OSLog)
+@available(iOS 14.0, tvOS 14.0, *)
+extension Logger {
+    static let datadog = Logger(subsystem: "dd-sdk-ios", category: "insights")
```
ganeshnj, line 47: can category be name of the module?

DatadogInternal/Sources/DD.swift
```diff
@@ -25,5 +25,31 @@ public struct DD {
...
+    }
+    #else
+    print(message)
+    #endif
+}
+
+#if canImport(OSLog)
+@available(iOS 14.0, tvOS 14.0, *)
+extension Logger {
+    static let datadog = Logger(subsystem: "dd-sdk-ios", category: "DatadogInternal")
```
ganeshnj, line 53: is it fair to say, we will print always `DatadogInternal` even though the log is coming from `DatadogTrace` for example?

maciejburda: ah yeah - this logger only exists in the `DatadogInternal`. We could potentially add extra parameter for the source of logs that we would put into category, but I am afraid this will add too much complexity to this setup

ganeshnj: Yes, something that we should do - I'm fine with creating a backlog if you think this will change the scope of this PR drastically.

ganeshnj (approved):

maxep (approved):

PR #1585, opened 2023-12-12
RUM-2153 Resources Recording

maciejburda:
### What and why?

Adds capability of recording image resources when traversing the view tree.

### How?

It modifies the view tree snapshot to provide references to resources along with the nodes. Those references are later used on the background thread by the resource processor and stored into batches.

It's currently disabled in the production code and marked with the `RUM-2154` ticket.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogSessionReplay/Sources/Feature/SessionReplayFeature.swift
+++ b/DatadogSessionReplay/Sources/Feature/SessionReplayFeature.swift
@@ -19,7 +19,7 @@ internal class SessionReplayFeature: DatadogRemoteFeature {
     /// Orchestrates the process of capturing next snapshots on the main thread.
     let recordingCoordinator: RecordingCoordinator
     /// Processes each new snapshot on a background thread and transforms it into records.
-    let processor: Processing
+    let processor: SnapshotProcessing
 
     // MARK: - Initialization
 
@@ -27,7 +27,7 @@ internal class SessionReplayFeature: DatadogRemoteFeature {
         core: DatadogCoreProtocol,
         configuration: SessionReplay.Configuration
     ) throws {
-        let processor = Processor(
+        let processor = SnapshotProcessor(
             queue: BackgroundAsyncQueue(named: "com.datadoghq.session-replay.processor"),
             writer: RecordWriter(core: core),
             srContextPublisher: SRContextPublisher(core: core),
--- a/DatadogSessionReplay/Sources/Feature/SessionReplayFeature.swift
+++ b/DatadogSessionReplay/Sources/Feature/SessionReplayFeature.swift
@@ -10,6 +10,7 @@ import DatadogInternal
 
 internal class SessionReplayFeature: DatadogRemoteFeature {
     static let name: String = "session-replay"
+    
     let requestBuilder: FeatureRequestBuilder
     let messageReceiver: FeatureMessageReceiver
     let performanceOverride: PerformancePresetOverride?
@@ -18,8 +19,6 @@ internal class SessionReplayFeature: DatadogRemoteFeature {
 
     /// Orchestrates the process of capturing next snapshots on the main thread.
     let recordingCoordinator: RecordingCoordinator
-    /// Processes each new snapshot on a background thread and transforms it into records.
-    let processor: SnapshotProcessing
 
     // MARK: - Initialization
 
@@ -27,10 +26,15 @@ internal class SessionReplayFeature: DatadogRemoteFeature {
         core: DatadogCoreProtocol,
         configuration: SessionReplay.Configuration
     ) throws {
-        let processor = SnapshotProcessor(
-            queue: BackgroundAsyncQueue(named: "com.datadoghq.session-replay.processor"),
-            writer: RecordWriter(core: core),
+        let snapshotProcessor = SnapshotProcessor(
+            queue: BackgroundAsyncQueue(named: "com.datadoghq.session-replay.snapshot-processor"),
+            recordsWriter: RecordWriter(core: core),
             srContextPublisher: SRContextPublisher(core: core),
+            telemetry: core.telemetry
+        )
+        let resourceProcessor = ResourceProcessor(
+            queue: BackgroundAsyncQueue(named: "com.datadoghq.session-replay.resource-processor"),
+            writer: ResourcesWriter(core: core),
             telemetry: core.telemetry
         )
 
@@ -38,7 +42,8 @@ internal class SessionReplayFeature: DatadogRemoteFeature {
         let messageReceiver = RUMContextReceiver()
 
         let recorder = try Recorder(
-            processor: processor,
+            snapshotProcessor: snapshotProcessor,
+            resourceProcessor: resourceProcessor,
             telemetry: core.telemetry,
             additionalNodeRecorders: configuration._additionalNodeRecorders
         )
@@ -53,7 +58,7 @@ internal class SessionReplayFeature: DatadogRemoteFeature {
 
         self.messageReceiver = messageReceiver
         self.recordingCoordinator = recordingCoordinator
-        self.processor = processor
+
         self.requestBuilder = SegmentRequestBuilder(
             customUploadURL: configuration.customEndpoint,
             telemetry: core.telemetry
--- a/DatadogSessionReplay/Sources/Feature/SessionReplayFeature.swift
+++ b/DatadogSessionReplay/Sources/Feature/SessionReplayFeature.swift
@@ -10,7 +10,7 @@ import DatadogInternal
 
 internal class SessionReplayFeature: DatadogRemoteFeature {
     static let name: String = "session-replay"
-    
+
     let requestBuilder: FeatureRequestBuilder
     let messageReceiver: FeatureMessageReceiver
     let performanceOverride: PerformancePresetOverride?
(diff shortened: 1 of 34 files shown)


DatadogSessionReplay/Sources/Feature/SessionReplayFeature.swift
```diff
@@ -18,42 +19,41 @@ internal class SessionReplayFeature: DatadogRemoteFeature {
 
     /// Orchestrates the process of capturing next snapshots on the main thread.
     let recordingCoordinator: RecordingCoordinator
-    /// Processes each new snapshot on a background thread and transforms it into records.
-    let processor: Processing
```
maciejburda, line 22: additional clean up - making sure that all the properties are where they belong

DatadogSessionReplay/Sources/Models/EnrichedResource.swift
```diff
@@ -8,7 +8,7 @@
 import Foundation
 
 /// Extends the resource information with context.
-internal struct EnrichedResource: Codable, Resource {
+internal struct EnrichedResource: Hashable, Codable, Resource {
```
maciejburda, line 11: Needed for `Set`, that utilised to deduplicate the resources

DatadogSessionReplay/Sources/Processor/ResourcesProcessor.swift
```diff
@@ -0,0 +1,33 @@
...
+    private let resourcesWriter: ResourcesWriting
+
+    func process(resources: [Resource], context: EnrichedResource.Context) {
+        guard !resources.isEmpty else {
+            return
+        }
+        queue.run { [resourcesWriter] in
+            resourcesWriter.write(resources: resources.map { EnrichedResource(resource: $0, context: context) })
+        }
+    }
```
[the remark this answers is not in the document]
maciejburda: `Resources` delivered to this processor in reality are references to UIImages that conform to `Resource` protocol.

Heavy operation is UIImage's `.pngData()` that's being extracted in the `EnrichedResource` initializer.

maciejburda:
@ncreated Also please take a look at the last commit. I showcased simplified threading, making both processors use the same queue.

I am a little bit on a fence with this, because it makes record upload depend on the resource processing. Having them independent sounded like a good idea at the beginning. Please let me know your thought 🙌

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UIImageViewRecorder.swift
```diff
@@ -7,6 +7,33 @@
...
+        self.tintColor = tintColor
+    }
+
+    public var identifier: String {
+        var identifier = image.srIdentifier
+        if let tintColorIdentifier = tintColor?.srIdentifier {
+            identifier += tintColorIdentifier
+        }
+        return identifier
+    }
```
ncreated, line 26: Because `image.srIdentifier` involves png compression and md5 calculation, it is very dangerous to expose this in public API. Even if it is hidden by `@_spi(Internal)`, how would we guarantee performance of this call in caller SDK (here RN SDK)?

DatadogSessionReplay/Sources/Recorder/Utilities/ImageDataProvider.swift
```diff
@@ -90,14 +100,18 @@ import CryptoKit
...
+
+            let hash = computeHash()
+            objc_setAssociatedObject(self, &customHashKey, hash, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
             return hash
         }
-
-        let hash = computeHash()
-        objc_setAssociatedObject(self, &customHashKey, hash, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
-        return hash
     }
```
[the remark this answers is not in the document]
ncreated: > And this identifier might be accessed (and lazily computed) from two different places.

TBH, it makes it very hard to read and reason about this code IMO. Depending on who calls this code first, it will cause impact on this or that thread. If tomorrow, we call it on the main thread, the SDK will cause huge impact (queue.sync {} + png compression + md5 computation) on user interface. It can be wrapped by the getter vs side-effects discussion from https://github.com/DataDog/dd-sdk-ios/pull/1585#discussion_r1439549398.

I'd prefer more streamlined design, where `srIdentifier` is set explicitly by respective thread. Other threads should only retrieve it in performant fashion.

ncreated (commented):
> I showcased simplified threading, making both processors use the same queue.

Not sure exactly how much we're winning with shared `queue`, but it is definitely a good option to use less threads if the performance is satisfactory.

My main concern in this PR is lack of clarity in threading due to causing expensive side-effect from `resource.identifier`. It might be worthy to have @maxep opinion on that, I don't want to block SR work.

The rest of code looks fine, especially I like the new return type from `record()` function which makes the code cleaner and easier to follow 👍 .

DatadogSessionReplay/Sources/Recorder/Recorder.swift
```diff
@@ -117,6 +123,11 @@ public class Recorder: Recording {
             }
             let touchSnapshot = touchSnapshotProducer.takeSnapshot(context: recorderContext)
             snapshotProcessor.process(viewTreeSnapshot: viewTreeSnapshot, touchSnapshot: touchSnapshot)
+
+            resourceProcessor?.process(
+                resources: viewTreeSnapshot.resources,
+                context: .init(recorderContext.applicationID)
+            )
```
ganeshnj, line 130: when there are `UIImage` which are not loaded yet, how are these handled?

in short - when images are loaded async.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UIImageViewRecorder.swift
```diff
@@ -7,6 +7,33 @@
...
+
+    public var data: Data {
+        var image = self.image
+        if #available(iOS 13.0, *), let tintColor = tintColor {
+            image = image.withTintColor(tintColor)
+        }
+        return image.scaledDownToApproximateSize(512.KB)
```
ganeshnj, line 33: 512 magic number? 

Should this be based on the resolution or actual size on the disk?

DatadogSessionReplay/Sources/Processor/ResourcesProcessor.swift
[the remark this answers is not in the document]
ganeshnj: regardless of whether a queue is needed here or not, the implementation of any processor/thing must not dictate where it is going to run, rather user of class decide where they want to run - delegate as much as possible.

This comes handy in troubleshooting and refactoring.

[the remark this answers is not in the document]
ganeshnj: Also `Queue` implementations are structs which holds `DispatchQueue` - a class, we are going to unnecessary create copies without any benefit.

DatadogSessionReplay/Sources/Recorder/Recorder.swift
maciejburda: This is only processing images that showed up on the screen (and met recording criteria).

Right now we don't support images downloaded in the runtime, but we might in the future.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UIImageViewRecorder.swift
maciejburda: I'll add a comment explaining the limitation. In short it's dictated by the intake limits.

DatadogSessionReplay/Sources/Recorder/Recorder.swift
ganeshnj: How do we make sure that image is loaded? Just curious on triggering logic, if you know the line of code, would be great.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/ViewTreeSnapshot.swift
```diff
@@ -54,10 +56,10 @@ internal typealias Node = SessionReplayNode
...
 @_spi(Internal)
 public protocol SessionReplayResource {
-    /// The unique identifier of the resource.
-    var identifier: String { get }
-    /// The data of the resource.
-    var data: Data { get }
+    /// Calculates the unique identifier of the resource.
+    func calculateIdentifier() -> String
+    /// Calculates the data of the resource.
+    func calculateData() -> Data
```
ncreated, line 62: flag/ These APIs are not thread safe as their implementation works on shared state. We could at least describe this in API comment, otherwise we risk a crash.

ncreated (approved):
One suggestion left. LGTM 👍

ncreated (approved):

PR #1588, opened 2023-12-13
RUM-2355 fix: Network Instrumentation

maxep:
### Task Interception

After first configuration, our new Network Instrumentation logic introduced in #1394 is intercepting all `URLSessionTask` **regardless of their delegate**.

Starting an interception can result in Start Resource in RUM. When the custom delegate is not in play, the interception will never finish and so the RUM Resource will never stop. This means that the resource event won't be transmitted, but it will keep the Resource scopes active and therefore the view scopes open.

### Task Completion

Tasks created with completion handler on pre-iOS 15 are never completed: We apply `#available(iOS 15, tvOS 15, *)` condition for completing the interception when receiving metrics, but we have no condition for pre-iOS 15.

### Swizzling Unbinding

Swizzlings on `URLSession*` are not removed when deallocating the `NetworkInstrumentationFeature`, this is preventing from re-enabling the instrumentation on a new core instance. It is blocking the 'stop core instance' feature on iOS.

### Legacy `DatadogURLSessionDelegate`

Changes in the legacy delegate are preventing from initialising multiple instances with different first-party hosts: The `DatadogURLSessionDelegate.init` is doing a static `URLSessionInstrumentation.enable` which can be done only once for a given delegate type.

### Instance-based Swizzler

In #1564, we have introduced reliable unswizzling logic which allows to create multiple instances of swizzler for the same method. I'm leveraging this to associate swizzlers to an instance of `NetworkInstrumentationFeature`, we can then unswizzle safely at deallocation.

The instance-based swizzling also allow to fix the legacy `DatadogURLSessionDelegate` by keeping a swizzler per delegate instance.

### Check for Task's delegate type

We now check for the delegate type before starting an interception,either when using the new `URLSessionInstrumentation` or the legacy `DatadogURLSessionDelegate`.

### Task Completion

The `URLSessionSwizzler` now allow to swizzle the `URLSession.dataTask(with:completionHandler:)` methods on both `URL` and `URLRequest` and swizzling is applied for all OS versions.

## Review

I suggest to review the following files first:
- [URLSessionSwizzler.swift](https://github.com/DataDog/dd-sdk-ios/blob/8174c17aa8a83adcad607ab9f554e5ddc5dcc48e/DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionSwizzler.swift)
- [NetworkInstrumentationFeature.swift](https://github.com/DataDog/dd-sdk-ios/blob/8174c17aa8a83adcad607ab9f554e5ddc5dcc48e/DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift)
- [DatadogURLSessionDelegate.swift](https://github.com/DataDog/dd-sdk-ios/blob/8174c17aa8a83adcad607ab9f554e5ddc5dcc48e/DatadogInternal/Sources/NetworkInstrumentation/URLSession/DatadogURLSessionDelegate.swift)

### Review checklist

- [x] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
```diff
@@ -38,6 +38,9 @@ internal final class NetworkInstrumentationFeature: DatadogFeature {
     @ReadWriteLock
     internal var handlers: [DatadogURLSessionHandler] = []
 
+    @ReadWriteLock
+    internal var swizzlers: [ObjectIdentifier: URLSessionSwizzler] = [:]
```
ganeshnj, line 42: Neat, never knew such type existed.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionSwizzler.swift
```diff
@@ -6,74 +6,102 @@
...
-            return _dataTaskWithURLRequestAndCompletion
-        }
-        set {
-            lock.lock()
-            defer { lock.unlock() }
-            _dataTaskWithURLRequestAndCompletion = newValue
-        }
-    }
+/// Swizzles `URLSession*` methods.
+internal final class URLSessionSwizzler {
```
ganeshnj, line 10: /Blocking

Combining all the swizzling code inside one type was never the intent but I understand the reasoning behind. However, this can be done with the existing approach by making the each swizzler type as intance type.

This makes the testing each swizzling class easy, better code navigation. You can compose those types in a `NetworkInstrumentationSwizzler` type. 

something like
```swift
class NetworkInstrumentationSwizzler {
    let urlSessionSwizzler: URLSessionSwizzler
    let urlSessionTaskDelegateSwizzler: URLSessionTaskDelegateSwizzler
    ....
}
```

ganeshnj (changes requested):

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionSwizzler.swift
maxep: As suggested, I have split the swizzler class, see the last commit. But TBH, I don't see the benefit 🤔 It just adds another layer and doesn't help much with code navigation.

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
```diff
@@ -49,80 +52,72 @@ internal final class NetworkInstrumentationFeature: DatadogFeature {
...
-                self?.queue.async { [weak self, weak session] in
-                    // prior to iOS 15, task state doesn't change to completed
-                    // hence we use didCompleteWithError to detect task completion
-                    self?._task(task, didCompleteWithError: task.error)
-                    session?.delegate?.interceptor?.task(task, didCompleteWithError: task.error)
+
+                if let currentRequest = task.currentRequest {
+                    let request = self.intercept(request: currentRequest, additionalFirstPartyHosts: configuredFirstPartyHosts)
+                    task.dd.override(currentRequest: request)
                 }
```
ncreated, line 84: question/ This condition is quite puzzling. In what circumstances we expect `task` to not have `currentRequest`? Let's explain it in the code comment if appropriate.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/DatadogURLSessionDelegate.swift
```diff
@@ -119,6 +114,37 @@ open class DatadogURLSessionDelegate: NSObject, URLSessionDataDelegate {
...
         interceptor?.task(task, didCompleteWithError: error)
     }
+
+    private func swizzle(firstPartyHosts: FirstPartyHosts) throws {
+        try swizzler.swizzle(
+            interceptResume: { [weak self] task in
+                guard
+                    let interceptor = self?.interceptor,
+                    let provider = task.dd.delegate as? __URLSessionDelegateProviding,
+                    provider.ddURLSessionDelegate == self // intercept task with self as delegate
```
ncreated, line 124: question/ Shouldn't we use `===` here?

DatadogInternal/Sources/NetworkInstrumentation/URLSession/DatadogURLSessionDelegate.swift
```diff
@@ -39,17 +46,8 @@ open class DatadogURLSessionDelegate: NSObject, URLSessionDataDelegate {
...
-        URLSessionInstrumentation.enable(
-            with: .init(
-                delegateClass: Self.self,
-                firstPartyHostsTracing: .traceWithHeaders(hostsWithHeaders: firstPartyHosts.hostsWithTracingHeaderTypes)
-            ),
-            in: core ?? CoreRegistry.default
-        )
-
         super.init()
+        try? swizzle(firstPartyHosts: .init())
```
ncreated, line 50: blocking/ At minimum we should log the problem to user console. This could save hours in troubleshooting user issues.

IntegrationTests/Runner/Environment.swift
```diff
@@ -45,16 +45,22 @@ internal struct URLSessionSetup: Codable {
     enum InstrumentationMethod: CaseIterable, Codable {
         /// Use `DDURLSessionDelegate` class directly.
         /// and define `firstPartyHosts` in feature configuration.
-        case directWithGlobalFirstPartyHosts
+        case directWithFeatureFirstPartyHosts
```
ncreated, line 48: question/ I can't see it at glance 👀 , what is the difference?

IntegrationTests/Runner/Environment.swift
```diff
@@ -45,16 +45,22 @@ internal struct URLSessionSetup: Codable {
...
         case inheritance
         /// Use `DDURLSessionDelegate` by compositing it custom delegate class (see: `CompositedURLSessionDelegate`).
         /// and define `firstPartyHosts` in feature configuration.
         case composition
+        /// Use a custom delegate.
+        /// and define `firstPartyHosts` in feature configuration.
+        case customWithFeatureFirstPartyHosts
+        /// Use a custom delegate.
+        /// and define `firstPartyHosts` with delegate's configuration.
+        case customWithAdditionalFirstyPartyHosts
```
ncreated, line 63: 🏅

ncreated (changes requested):
Massive efforts 👌 - I noticed only one blocking change, other look ok. It is really hard to see it well through (the domain is complex, code looks fine), hence I count on the quality of tests. Glad we add more integration scenarios.

Last question, about:
> **Task Interception**
> After first configuration, our new Network Instrumentation logic introduced in https://github.com/DataDog/dd-sdk-ios/pull/1394 is intercepting all URLSessionTask regardless of their delegate.

Is there a test we add to verify that not instrumented requests are ignored?

DatadogCore/Tests/TestsObserver/DatadogTestsObserver.swift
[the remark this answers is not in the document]
maxep: Previously, it was actually covered twice. Swizzling integrity is already checked by [this](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogCore/Tests/TestsObserver/DatadogTestsObserver.swift#L35L44) which will fail if any swizzling is left, printing the method signatures 👍

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
maxep: That's a good question, I don't know what could nillify the `originalRequest ` or `currentRequest` 🤔  the doc doesn't say..

DatadogInternal/Sources/NetworkInstrumentation/URLSession/DatadogURLSessionDelegate.swift
maxep: Oh good catch! the `==` operator is delegating to the `NSObjectProtocol.isEqual(_:)`, so it will be the same but the `===` should be explicit here!

DatadogInternal/Sources/NetworkInstrumentation/URLSession/DatadogURLSessionDelegate.swift
maxep: Indeed 👍

IntegrationTests/Runner/Environment.swift
maxep: Just renaming `Global` to `Feature`, there is no difference in behavior

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
```diff
@@ -38,6 +38,9 @@ internal final class NetworkInstrumentationFeature: DatadogFeature {
     @ReadWriteLock
     internal var handlers: [DatadogURLSessionHandler] = []
 
+    @ReadWriteLock
+    internal var swizzlers: [ObjectIdentifier: NetworkInstrumentationSwizzler] = [:]
```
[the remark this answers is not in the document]
maxep: I'm making this private and adding `debugDescription` to `MethodSwizzler` that should gives us enough information to troubleshoot.

maxep:
It is ready for another round of review :) 

@ncreated, to your question:
> Last question, about:
> > **Task Interception**
> > After first configuration, our new Network Instrumentation logic introduced in https://github.com/DataDog/dd-sdk-ios/pull/1394 is intercepting all URLSessionTask regardless of their delegate.
> 
> Is there a test we add to verify that not instrumented requests are ignored? 

Yes, the `NetworkInstrumentationFeatureTests` now tests not instrumented requests: [here](https://github.com/DataDog/dd-sdk-ios/pull/1588/files#diff-4985edd922141ebf762056233cbf19d4a154f9d305b00e729c82f22a5b0ae69dR235), [here](https://github.com/DataDog/dd-sdk-ios/pull/1588/files#diff-4985edd922141ebf762056233cbf19d4a154f9d305b00e729c82f22a5b0ae69dR336), and [here](https://github.com/DataDog/dd-sdk-ios/pull/1588/files#diff-4985edd922141ebf762056233cbf19d4a154f9d305b00e729c82f22a5b0ae69dR393). It covers async/await, session delegate, and task delegate.

DatadogCore/Tests/TestsObserver/DatadogTestsObserver.swift
```diff
@@ -132,34 +132,6 @@ internal class DatadogTestsObserver: NSObject, XCTestObservation {
 
             If all above conditions are met, this failure might indicate a memory leak in the implementation.
             """
-        ),
```
ganeshnj, line 135: Removing these checks seems like a back step - in case of a memory leak we will not able to catch any lingering swizzlings.

A check something like if there is an instance of swizzler in memory or not will be helpful.

something like 
```swift
func checkInstancesOfType(_ type: AnyClass) -> Bool {
    let typeString = String(describing: type)
    
    for bundle in Bundle.allBundles {
        if let className = bundle.classNamed(typeString) {
            // Check if an instance of the class exists
            if class_getInstanceSize(className) > 0 {
                return true
            }
        }
    }
    
    return false
}
```

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
```diff
@@ -49,80 +52,72 @@ internal final class NetworkInstrumentationFeature: DatadogFeature {
...
+
+        let swizzler = NetworkInstrumentationSwizzler()
+        swizzlers[identifier] = swizzler
+
+        try swizzler.swizzle(
+            interceptResume: { [weak self] task in
+                // intercept task if delegate match
+                guard let self = self, task.dd.delegate?.isKind(of: configuration.delegateClass) == true else {
+                    return
                 }
```
ganeshnj, line 79: How is the dd.delegate lifecycle is managed?

DatadogInternal/Sources/NetworkInstrumentation/URLSession/DatadogURLSessionDelegate.swift
```diff
@@ -14,6 +14,13 @@ public typealias DDURLSessionDelegate = DatadogURLSessionDelegate
 @objc
 @available(*, deprecated, message: "Use `URLSessionInstrumentation.enable(with:)` instead.")
 public protocol __URLSessionDelegateProviding: URLSessionDelegate {
+    /// Datadog delegate object.
+    /// 
+    /// The class implementing `DDURLSessionDelegateProviding` must ensure that following method calls are forwarded to `ddURLSessionDelegate`:
+    /// - `func urlSession(_:task:didFinishCollecting:)`
+    /// - `func urlSession(_:task:didCompleteWithError:)`
+    /// - `func urlSession(_:dataTask:didReceive:)`
+    var ddURLSessionDelegate: DatadogURLSessionDelegate { get }
```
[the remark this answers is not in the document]
ganeshnj: which particular behavior?

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionDataDelegateSwizzler.swift
```diff
@@ -1,127 +0,0 @@
...
- * This product includes software developed at Datadog (https://www.datadoghq.com/).
- * Copyright 2019-Present Datadog, Inc.
- */
-
-import Foundation
-
-/// Swizzles `URLSessionDataDelegate` callbacks.
-internal class URLSessionDataDelegateSwizzler {
-    private static var _didReceiveMap: [String: DidReceive?] = [:]
-    static var didReceiveMap: [String: DidReceive?] {
```
ganeshnj, line 12: what is the replacement of the didReceive delegate?

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskDelegateSwizzler.swift
```diff
@@ -8,105 +8,95 @@ import Foundation
...
-            return _didCompleteWithErrorMap
-        }
-        set {
-            lock.lock()
-            defer { lock.unlock() }
-            _didCompleteWithErrorMap = newValue
-        }
-    }
+    private let lock: NSLocking
+    private var didReceive: DidReceive?
```
ganeshnj, line 12: didReceive is only available in `URLSessionDataDelegate`

https://developer.apple.com/documentation/foundation/urlsessiondatadelegate/1410027-urlsession

It should be in its own swizzler instance.

ganeshnj (changes requested):

DatadogCore/Tests/TestsObserver/DatadogTestsObserver.swift
maxep: It is already there, see [my comment](https://github.com/DataDog/dd-sdk-ios/pull/1588#discussion_r1433930181). These checks were redundant, we are already asserting that no swizzling is left [here](https://github.com/DataDog/dd-sdk-ios/blob/738d599b46f982275a6c41c48578808403f36eb6/DatadogCore/Tests/TestsObserver/DatadogTestsObserver.swift#L35L44). Failure will print the method signature, so we will know which sizzling is at fault.

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
maxep: It's a computed `var` and not retained. see its implementation [here](https://github.com/DataDog/dd-sdk-ios/blob/738d599b46f982275a6c41c48578808403f36eb6/DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTask%2BTracking.swift#L23L35)

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskDelegateSwizzler.swift
maxep: ah yea 👍

DatadogCore/Tests/TestsObserver/DatadogTestsObserver.swift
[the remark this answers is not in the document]
ganeshnj: Nice, can you add some simple test case?

- swizzling once, Swizzling.methods.count == 1
- same method again, Swizzling.methods.count == 2
- another method, Swizzling.methods.count == 3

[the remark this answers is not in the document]
maxep: I have added this to `MethodSwizzlerTests` 👍

ncreated (approved):
Looks good 👍. Seeing a lot of changes that involve OS-version-specific swizzlings, can we verify if unit tests pass all major iOS versions from 12 to 16 🙏? Currently we can't rely on nightly tests for this.

ganeshnj (approved):

maxep:
I have ran the test-suites on the following versions:
|OS Version|Unit Tests|Integration Tests||
|---|---|---|---|
|iOS 12.4|✅|🚫|Unable to test due to SPM failure|
|iOS 13.7|❌|🚫|units fails on `master` as well|
|iOS 14.4|✅|✅||
|iOS 15.5|✅|✅||
|iOS 16.0|✅|✅||

It fails on iOS 13, from what I see it's due to the `URLSessionTask.resume` which doesnt seem to trigger the swizzled method. It fails the same on `master` so it is not due to the change in this PR nor the change in swizzling.
As this additional issue is not introduce by changes in this PR, we can tackle this seperatly.

PR #1593, opened 2023-12-15
feat: Provide a method for late initialization of URLSessionTracking

fuzzybinary:
### What and why?

This is for a limited use case of people that need to initialize native URL session tracking after RUM has been initialized.

refs: RUM-2525

### How?

The approach here is a bit weird because of the need to include some providers (the Trace ID generator and the date provider) in the configuration. We really only override these in unit tests, and they're internal, but I needed to provide a way to include the overrides, hence the new `Configuration.LateURLSessionTracking` object.

### Review checklist

- [x] Run integration tests
- [x] Run smoke tests
- [ ] Run tests for `tools/`


ganeshnj:
A bit lost with the requirement of this change. Do you have a code example where the current version doesn't work.

In current available version, either RUM or Tracing must be enabled to enable URLSession tracking.

DatadogRUM/Sources/RUM+Internal.swift
```diff
@@ -9,6 +9,42 @@ import DatadogInternal
...
+        ) {
+            self.firstPartyHostsTracing = firstPartyHostsTracing
+            self.resourceAttributesProvider = resourceAttributesProvider
+        }
+
+        public init(from: URLSessionTracking) {
+            self.firstPartyHostsTracing = from.firstPartyHostsTracing
+            self.resourceAttributesProvider = from.resourceAttributesProvider
+        }
+    }
```
maxep, line 43: /blocker The internal properties should be injected by the main config, alongside the `URLSessionTracking`. You also don't need the public accessor when using the main config. It can be something like:
```swift
// MARK: - Internal

internal let debugSDK: Bool
internal let dateProvider: DateProvider
internal let traceIDGenerator: TraceIDGenerator

public init(
    firstPartyHostsTracing: RUM.Configuration.URLSessionTracking.FirstPartyHostsTracing? = nil,
    resourceAttributesProvider: RUM.ResourceAttributesProvider? = nil,
    debugSDK: Bool = false // if you ever need to debug from CP
) {
    self.firstPartyHostsTracing = firstPartyHostsTracing
    self.resourceAttributesProvider = resourceAttributesProvider
    self.debugSDK = debugSDK
    self.dateProvider = SystemDateProvider()
    self.traceIDGenerator = DefaultTraceIDGenerator()
}

internal init(
    from configuration: URLSessionTracking,
    debugSDK: Bool,
    dateProvider: DateProvider,
    traceIDGenerator: TraceIDGenerator
) {
    self.firstPartyHostsTracing = configuration.firstPartyHostsTracing
    self.resourceAttributesProvider = configuration.resourceAttributesProvider
    self.debugSDK = debugSDK
    self.dateProvider = dateProvider
    self.traceIDGenerator = traceIDGenerator
}
```

DatadogRUM/Sources/RUM+Internal.swift
```diff
@@ -19,4 +55,56 @@ extension InternalExtension where ExtendedType == RUM {
...
+            )
+        case let .traceWithHeaders(hostsWithHeaders, sampleRate):
+            distributedTracing = DistributedTracing(
+                sampler: Sampler(samplingRate: sampleRate),
+                firstPartyHosts: FirstPartyHosts(hostsWithHeaders),
+                traceIDGenerator: configuration.traceIDGenerator
+            )
+        case .none:
+            distributedTracing = nil
+        }
```
maxep, line 98: /blocker The debug config should override the sample rate.
```suggestion
        // If first party hosts are configured, enable distributed tracing:
        switch configuration.firstPartyHostsTracing {
        case let .trace(hosts, sampleRate):
            distributedTracing = DistributedTracing(
                sampler: Sampler(samplingRate: configuration.debugSDK ? 100 : sampleRate),
                firstPartyHosts: FirstPartyHosts(hosts),
                traceIDGenerator: configuration.traceIDGenerator
            )
        case let .traceWithHeaders(hostsWithHeaders, sampleRate):
            distributedTracing = DistributedTracing(
                sampler: Sampler(samplingRate: configuration.debugSDK ? 100 : sampleRate),
                firstPartyHosts: FirstPartyHosts(hostsWithHeaders),
                traceIDGenerator: configuration.traceIDGenerator
            )
        case .none:
            distributedTracing = nil
        }
```

DatadogRUM/Sources/RUM+Internal.swift
```diff
@@ -19,4 +55,56 @@ extension InternalExtension where ExtendedType == RUM {
...
         return core.get(feature: RUMFeature.self) != nil
     }
+
+    /// Enable URL session tracking after RUM has already been enabled. This method
+    /// is only needed if the configuration of URL session tracking is not known at initialization time,
+    /// or in the case of cross platform frameworks that do not initalize native URL session tracking.
+    ///
+    /// - Parameters:
+    ///    - in: the configuration for URL session tracking
+    public static func enableUrlSessionTracking(
```
maxep, line 65: /nit
```suggestion
    public static func enableURLSessionTracking(
```

DatadogRUM/Tests/RUMTests.swift
```diff
@@ -404,4 +404,37 @@ class RUMTests: XCTestCase {
...
+        let core = FeatureRegistrationCoreMock()
+        let rumConfig = RUM.Configuration(applicationID: .mockAny())
+        RUM.enable(with: rumConfig, in: core)
+        let lateConfig = RUM.Configuration.LateURLSessionTracking()
+
+        // When
+        try RUM._internal.enableUrlSessionTracking(with: lateConfig, in: core)
+
+        // Then
+        XCTAssertNotNil(core.get(feature: NetworkInstrumentationFeature.self))
```
maxep, line 438: I think we should also assert the proper configuration of the `URLSessionRUMResourcesHandler`, making sure that properties are well propagated.

maxep (changes requested):

fuzzybinary:
@ganeshnj This is a very strange use case of a Flutter user that uses native iOS URLSessions. Flutter initializes Datadog and RUM, but they need a way to enable URLSession tracking after initialization has been performed.

ganeshnj:
> @ganeshnj This is a very strange use case of a Flutter user that uses native iOS URLSessions. Flutter initializes Datadog and RUM, but they need a way to enable URLSession tracking after initialization has been performed.

I see, I wonder if we could use the existing config type with default values and extend it with implementations that you want to inject, because this type will go to public type (in auto completions et all) and makes the API unintuitive.

fuzzybinary:
@ganeshnj I'm not sure how to do this cleanly without a new type or by adding more parameters to `enableURLSessionTracking`. Do you have an example of how you'd structure it?

Alternatively, we could put `LateURLSessionTracking` under another wrapper `Internal` or `__Internal` to avoid it showing up at the top level of autocomplete?

maxep (dismissed):
Nice, thank you for the changes and the fuzzy tests 🙏

DatadogRUM/Sources/RUM+Internal.swift
```diff
@@ -19,4 +64,56 @@ extension InternalExtension where ExtendedType == RUM {
     public static func isEnabled(in core: DatadogCoreProtocol = CoreRegistry.default) -> Bool {
         return core.get(feature: RUMFeature.self) != nil
     }
+
+    /// Enable URL session tracking after RUM has already been enabled. This method
+    /// is only needed if the configuration of URL session tracking is not known at initialization time,
+    /// or in the case of cross platform frameworks that do not initalize native URL session tracking.
+    ///
+    /// - Parameters:
+    ///    - in: the configuration for URL session tracking
```
ncreated, line 73: This list of parameters seems wrong 🖐️

DatadogRUM/Sources/RUM+Internal.swift
```diff
@@ -19,4 +64,56 @@ extension InternalExtension where ExtendedType == RUM {
...
+    public static func enableURLSessionTracking(
+        with configuration: RUM.Configuration.LateURLSessionTracking,
+        in core: DatadogCoreProtocol = CoreRegistry.default) throws {
```
ncreated, line 76: question/ Have we considered changing this to following?
```swift
    public static func enableURLSessionTracking(
        with urlSessionTrackingConfiguration: RUM.Configuration.URLSessionTracking,
        rumConfiguration: RUM.Configuration,
        in core: DatadogCoreProtocol = CoreRegistry.default) throws {
```
Effectively, this way we won't need to introduce new `LateURLSessionTracking` nor duplicate any properties. The trade-off will be on the caller-side (in Flutter), where we'll need to supply this method both configurations:
- `RUM.Configuration.URLSessionTracking`
- `RUM.Configuration`

For the sake of streamlining RUM init, the cost seems worth it. WDYT?

fuzzybinary: I'm concerned with requiring `RUM.Configuration` because of the number of parameters available that just aren't used, especially the non-optional parameters that wouldn't actually have an affect.

Remember "on the Flutter side" in this case is a client who has to make sense of the API, and requiring the `RUM.Configuration` when only a few parameters out of it are going to be used worries me. I'm open to discuss it though.

ncreated: I see 👍. One more idea that comes to my mind is attaching `RUM.Configuration` to `RUMFeature`, so we can retrieve it "after" by ourselves without requiring the client to pass it:
```swift
public static func enableURLSessionTracking(
   with urlSessionTrackingConfiguration: RUM.Configuration.URLSessionTracking,
   in core: DatadogCoreProtocol = CoreRegistry.default
) throws {
   guard let rum = core.get(feature: RUMFeature.self) else {
      throw ProgrammerError(
         description: "RUM must be initialized before calling `RUM.enableUrlSessionTracking`."
      )
   }
   let rumConfiguration = rum.configuration // <-- ⭐️ new thing
   // ...
```

This isn't a change request. Using separate `LateURLSessionTracking` type makes sense as well 👍.

ncreated (approved):

fuzzybinary:
I would actually prefer adding the configuration to the feature if we're good with that.

fuzzybinary:
Changed to saving the configuration. It's much cleaner, even if it does mean holding onto some extra info.

ncreated (approved):
LGTM 👍. Thanks for extra efforts on simplification 🙂

ncreated (approved):

PR #1596, opened 2023-12-20
fix: Fix `view.time_spent` skews

ncreated:
### What and why?

🧰 Fixes the problem of `view.time_spent` increasing infinitely if view scope tracks dangling resources (ones that were started, but never finished). 

### How?

The problem originated in the condition meant to close current view scope if next view is started. This piece of logic was executed for every next view, instead of being limited to the succeeding view only.

### Review checklist

- [x] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+# 2.5.1 / 20-12-2023
+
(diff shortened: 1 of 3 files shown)


maxep (approved):
Nice, thanks 🙏

ganeshnj (dismissed):

PR #1613, opened 2024-01-04
Add font size argument in text Wireframe builder

louiszawadzki:
### What and why?

Add font size argument in `createTextWireframe` public builder.

Today passing a `font` argument is the only way to specify a font size for the text wireframe.
With React Native text views we don't get direct access to the `UIFont`. However we get direct access to the fontSize, 
 so we can bypass this by creating a new font on every call, but that takes a lot of time and impacts performance (on an app I tested, it took up to 300ms of time over a minute of usage of the app).

I've tried passing `nil` as font (https://github.com/DataDog/dd-sdk-reactnative/pull/570), but then all elements have a small font size.

I believe enabling to specify only the font size is the best way to get a great fidelity with a small performance cost.

### How?

It seems that there is no test covering this part of the code, let me know if I should add some.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogSessionReplay/Sources/Processor/SRDataModelsBuilder/WireframesBuilder.swift
+++ b/DatadogSessionReplay/Sources/Processor/SRDataModelsBuilder/WireframesBuilder.swift
@@ -98,6 +98,7 @@ public class SessionReplayWireframesBuilder {
         clip: SRContentClip? = nil,
         textColor: CGColor? = nil,
         font: UIFont? = nil,
+        fontSize: CGFloat? = nil,
         fontScalingEnabled: Bool = false,
         borderColor: CGColor? = nil,
         borderWidth: CGFloat? = nil,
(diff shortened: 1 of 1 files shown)


DatadogSessionReplay/Sources/Processor/SRDataModelsBuilder/WireframesBuilder.swift
```diff
@@ -98,6 +98,7 @@ public class SessionReplayWireframesBuilder {
         clip: SRContentClip? = nil,
         textColor: CGColor? = nil,
         font: UIFont? = nil,
+        fontSize: CGFloat? = nil,
```
maciejburda, line 101: I don't entirely like the fact that it's now implicitly either `fontSize` or `font.size`. Maybe we could rename to `fontSizeOverride` ?

maciejburda (approved):
One non-blocking suggestion

maciejburda (approved):

PR #1615, opened 2024-01-05
RUM-699 fix: RUM context not being linked to started span

ncreated:
### What and why?

🐞 Fixes the bug of RUM information not being linked correctly on `tracer.startSpan()`.

Screenshot 2024-01-05 at 14 31 46

### How?

The problem was introduced in V2 where we started sending RUM information in `DatadogContext` by propagating its updates on message bus. Because message bus adds one more asynchronous call, the information was not available in to `tracer` at the right time:
```swift
rumMonitor.startView(key: "view", name: "FooView")
let span = tracer.startSpan(operationName: "span") // <--- here, we haven't yet 
                                                   // seen Foo's `view.id` in `DatadogContext`
// ...
span.finish()
```

Under the hood, the propagation model was this:
- (in RUM) set RUM info in `DatadogContext` (async on context queue)
- (in Core) broadcast `DatadogContext` update to all modules (async on message bus queue)
- (in Trace) receive `DatadogContext` and store RUM info in `MessageReceiver`
- (in Trace) on starting span: capture received RUM info from `MessageReceiver`
- (in Trace) on finishing span: create & write span (async on context queue)

It is now simplified and changed to:
- (in RUM) set RUM info in `DatadogContext` (async on context queue)
- (in Trace) on starting span: capture latest `DatadogContext` (async on context queue)
- (in Trace) on finishing span: create & write span with `DatadogContext` captured on starting it (async on context queue)

Ultimately, RUM context is no longer passed to Trace through message bus and `MessageReceiver`. Instead, it is queried by `DatadogTrace` with the new capability added to `FeatureScope`:

```swift
public protocol FeatureScope {
    /// Retrieve the core context.
    ///
    /// A feature can use this method to request the Datadog context valid at the moment of the call.
    ///
    /// - Parameter block: The block to execute; it is called on the context queue.
    func context(_ block: @escaping (DatadogContext) -> Void)

    // ...
}
```

🎁 Part of the requirement to enable this work was removing `tracer.queue` and replacing it with `@ReadWriteLock` to synchronise tracer and span states. According to my benchmark measures, there is no performance change coming from this PR (both in `Tracer` and in `DatadogCore`).

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


DatadogCore/Sources/Core/DatadogCore.swift
```diff
@@ -309,24 +308,21 @@ extension DatadogCore: DatadogCoreProtocol {
...
+        }
+    }
 
-            // Still on context thread: send `Writer` to EWC caller. The writer implements `AsyncWriter`, so
-            // the implementation of `writer.write(value:)` will run asynchronously without blocking the context thread.
-            do {
-                try block(context, writer)
-            } catch {
-                telemetry.error("Failed to execute feature scope", error: error)
-            }
```
ncreated, line 329: This change is intentional and makes the `block()` not throwing. Rationale is: the core being agnostic from any product-specific logic has no idea on how to handle or recover from the error. We had zero occurrence of `"Failed to execute feature scope"` in telemetry.

ganeshnj (approved):
Good cleanup. Nice work.

ganeshnj (approved):

maxep (approved):

PR #1616, opened 2024-01-05
feat: Add an accessor for the current session ID

fuzzybinary:
### What and why?

This adds a `currentSessionID` accessor onto the RUM Monitor. There are cases (usually support cases) where users want to be able to access the current session ID more easily than with the currently available "sessionStarted" notifications.

refs: RUM-1961

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogCore/Tests/Datadog/RUM/RUMMonitorTests.swift
+++ b/DatadogCore/Tests/Datadog/RUM/RUMMonitorTests.swift
@@ -42,11 +42,9 @@ class RUMMonitorTests: XCTestCase {
         RUM.enable(with: config, in: core)
         let monitor = RUMMonitor.shared(in: core)
 
-        monitor.startView(viewController: mockView)
-
         // When
-        let expectation = XCTestExpectation(description: "getCurrentSessionID callback recieved")
-        monitor.getCurrentSessionID { sessionId in
+        let expectation = XCTestExpectation(description: "currentSessionID callback recieved")
+        monitor.currentSessionID { sessionId in
             // Then
             XCTAssertNotNil(sessionId)
             XCTAssertEqual(capturedSession, sessionId)
@@ -63,12 +61,31 @@ class RUMMonitorTests: XCTestCase {
         RUM.enable(with: config, in: core)
         let monitor = RUMMonitor.shared(in: core)
 
+        // When
+        let expectation = XCTestExpectation(description: "currentSessionID callback recieved")
+        monitor.currentSessionID { sessionId in
+            // Then
+            XCTAssertNil(sessionId)
+            expectation.fulfill()
+        }
+
+        wait(for: [expectation], timeout: 0.1)
+    }
+
+    func testWhenSessionIsStopped_itReturnsNil() throws {
+        // Given
+        config.dateProvider = RelativeDateProvider(startingFrom: Date(), advancingBySeconds: 1)
+        config.sessionSampleRate = 100.0
+        RUM.enable(with: config, in: core)
+        let monitor = RUMMonitor.shared(in: core)
+
         setGlobalAttributes(of: monitor)
         monitor.startView(viewController: mockView)
 
         // When
-        let expectation = XCTestExpectation(description: "getCurrentSessionID callback recieved")
-        monitor.getCurrentSessionID { sessionId in
+        monitor.stopSession()
+        let expectation = XCTestExpectation(description: "currentSessionID callback recieved")
+        monitor.currentSessionID { sessionId in
             // Then
             XCTAssertNil(sessionId)
             expectation.fulfill()
--- a/DatadogCore/Tests/Datadog/RUM/RUMMonitorTests.swift
+++ b/DatadogCore/Tests/Datadog/RUM/RUMMonitorTests.swift
@@ -30,6 +30,53 @@ class RUMMonitorTests: XCTestCase {
         super.tearDown()
     }
 
+    // MARK: - Current Session Id
+    func testWhenSessionIsSampledIn_itReturnsCurrentSessionId() throws {
+        // Given
+        var capturedSession: String?
+        config.dateProvider = RelativeDateProvider(startingFrom: Date(), advancingBySeconds: 1)
+        config.sessionSampleRate = 100.0
+        config.onSessionStart = { session, sampled in
+            capturedSession = session
+        }
+        RUM.enable(with: config, in: core)
+        let monitor = RUMMonitor.shared(in: core)
+
+        monitor.startView(viewController: mockView)
+
+        // When
+        let expectation = XCTestExpectation(description: "getCurrentSessionID callback recieved")
+        monitor.getCurrentSessionID { sessionId in
+            // Then
+            XCTAssertNotNil(sessionId)
+            XCTAssertEqual(capturedSession, sessionId)
+            expectation.fulfill()
+        }
+
+        wait(for: [expectation], timeout: 0.1)
+    }
+
+    func testWhenSessionIsSampled_itReturnsNil() throws {
+        // Given
+        config.dateProvider = RelativeDateProvider(startingFrom: Date(), advancingBySeconds: 1)
+        config.sessionSampleRate = 0.0
+        RUM.enable(with: config, in: core)
+        let monitor = RUMMonitor.shared(in: core)
+
+        setGlobalAttributes(of: monitor)
+        monitor.startView(viewController: mockView)
+
+        // When
+        let expectation = XCTestExpectation(description: "getCurrentSessionID callback recieved")
+        monitor.getCurrentSessionID { sessionId in
+            // Then
+            XCTAssertNil(sessionId)
+            expectation.fulfill()
+        }
+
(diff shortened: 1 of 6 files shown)


DatadogRUM/Sources/RUMMonitor/Monitor.swift
```diff
@@ -195,6 +195,17 @@ internal class Monitor: RUMCommandSubscriber {
...
+    // MARK: - properties
+    var currentSessionID: String? {
+        get {
+            guard let activeSession = self.scopes.activeSession else {
+                return nil
+            }
+
+            return activeSession.sessionUUID.rawValue.uuidString
+        }
+    }
```
ncreated, line 207: blocker/ This isn't thread safe and suffers from race condition.
- thread safety: `scopes` are mutated [on monitor's `queue`](https://github.com/DataDog/dd-sdk-ios/blob/7559cdda5ef86a6e3f7294a3c8d2f85a9dfcbdf9/DatadogRUM/Sources/RUMMonitor/Monitor.swift#L134-L138) so we need to sync the getter on that queue as well.
- race condition: `scopes` mutation [is dispatched asynchronously on core's context queue](https://github.com/DataDog/dd-sdk-ios/blob/7559cdda5ef86a6e3f7294a3c8d2f85a9dfcbdf9/DatadogRUM/Sources/RUMMonitor/Monitor.swift#L134) (with `eventWriteContext { ... }`), so session ID is not yet resolved before `process(command:)` returns.

Both problems are not visible in tests this PR adds in `MonitorTests`, because it is using `PassthroughCoreMock` which is single-threaded and we end these tests by calling `monitor.flush()` to force-syncs monitor state. To surface these problems, try adding tests in [RUMMonitorTests.swift](https://github.com/DataDog/dd-sdk-ios/blob/2.5.1/DatadogCore/Tests/Datadog/RUM/) where we use a real instance of the SDK core.

Solving thread-safety is simple as it may only require using `queue.sync {}`. The price would be on performance - it will block the caller thread for the duration of current RUM event processing.

Solving race condition would be more difficult, as it may require sync-ing on core context queue as well. This might be not a good idea, as it will block the caller thread for additional duration of pending core operations. For that reason there is no `sync {}` API provided by core and we may need to use semaphores to workaround it.

Compromise might be guaranteeing thread-safety but explicitly not promising `sessionID` integrity (in API comment). It would mean that following sequence might print the same `sessionID` twice:
```swift
print(monitor.currentSessionID)
monitor.stopSession()
monitor.startView(key: "Foo")
print(monitor.currentSessionID)
```

Alternatively, we can make this API asynchronous:
```swift
monitor.currentSessionID { sessionID in
   print(sessionID) // called asynchronously
}
```
and dispatch `sessionID` in the same model as we are processing RUM commands. This would be the correct solution, but it might be less convenient for users.

DatadogRUM/Tests/RUMMonitor/MonitorTests.swift
```diff
@@ -44,6 +44,26 @@ class MonitorTests: XCTestCase {
...
+        let sampler = Sampler(samplingRate: 100)
+
+        // When
+        let monitor = Monitor(
+            core: core,
+            dependencies: .mockWith(core: core, sessionSampler: sampler),
+            dateProvider: DateProviderMock()
+        )
+        monitor.startView(key: "foo")
+        monitor.flush()
```
ncreated, line 58: The use of `monitor.flush()` and `PassthroughCoreMock` (as `core`) hides key problems with this design (it isn't thread-safe and is exposed to race condition). See my previous comment for insight. It would be better to add this test in [RUMMonitorTests.swift](https://github.com/DataDog/dd-sdk-ios/blob/2.5.1/DatadogCore/Tests/Datadog/RUM/RUMMonitorTests.swift) where we use a real instance of `DatadogCore` (not a single-threaded mock).

ncreated (changes requested):

fuzzybinary:
@ncreated Switched to the async callback. Let me know what you think.

DatadogRUM/Tests/RUMTests.swift
```diff
@@ -74,6 +74,23 @@ class RUMTests: XCTestCase {
...
+
+        // Then
+        RUMMonitor.shared().getCurrentSessionID {
+            XCTAssertNil($0)
+            expectation.fulfill()
+        }
+
+        wait(for: [expectation], timeout: 0.1)
+    }
+
```
ncreated, line 93: blocking/ This test is false positive. There is no such behaviour as "current session starts as `nil`" after https://github.com/DataDog/dd-sdk-ios/pull/1594

One problem is that `RUMMonitor.shared()` is missing the `core`:
```swift
RUMMonitor.shared(in: core)
```

However, if we pass the `core` reference this test will fail because we use `FeatureRegistrationCoreMock` in this file and that only offers feature registration capability, see: https://github.com/DataDog/dd-sdk-ios/blob/a43c63dd422ffd050ed24a1ac1dfd7b156f063c2/TestUtilities/Mocks/CoreMocks/FeatureRegistrationCoreMock.swift#L9-L23

If we change it to `PassthroughCoreMock` as suggested, it will keep failing because that [mock doesn't support `get<T>(feature:)`](https://github.com/DataDog/dd-sdk-ios/blob/a43c63dd422ffd050ed24a1ac1dfd7b156f063c2/TestUtilities/Mocks/CoreMocks/PassthroughCoreMock.swift#L94) required in [`RUMMonitor.shared(in:)](https://github.com/DataDog/dd-sdk-ios/blob/a43c63dd422ffd050ed24a1ac1dfd7b156f063c2/DatadogRUM/Sources/RUMMonitor.swift#L35).

After all, I think the problem is because we try to test `RUMMonitor` in `RUMTests.swift`, which is only configured to test `RUM.*` configuration and no behaviours. My recommendation is to remove this test from here and cover this flow in `RUMMonitorTests.swift`, which depends on real instance of the core.

DatadogCore/Tests/Datadog/RUM/RUMMonitorTests.swift
```diff
@@ -30,6 +30,53 @@ class RUMMonitorTests: XCTestCase {
...
+        var capturedSession: String?
+        config.dateProvider = RelativeDateProvider(startingFrom: Date(), advancingBySeconds: 1)
+        config.sessionSampleRate = 100.0
+        config.onSessionStart = { session, sampled in
+            capturedSession = session
+        }
+        RUM.enable(with: config, in: core)
+        let monitor = RUMMonitor.shared(in: core)
+
+        monitor.startView(viewController: mockView)
```
ncreated, line 45: suggestion/ RUM session should start immediately, so there is no need to start a view. It might be worth changing it to cover the minimal behaviour.

DatadogCore/Tests/Datadog/RUM/RUMMonitorTests.swift
```diff
@@ -30,6 +30,53 @@ class RUMMonitorTests: XCTestCase {
...
+
+    func testWhenSessionIsSampled_itReturnsNil() throws {
+        // Given
+        config.dateProvider = RelativeDateProvider(startingFrom: Date(), advancingBySeconds: 1)
+        config.sessionSampleRate = 0.0
+        RUM.enable(with: config, in: core)
+        let monitor = RUMMonitor.shared(in: core)
+
+        setGlobalAttributes(of: monitor)
+        monitor.startView(viewController: mockView)
```
ncreated, line 67: suggestion/ Both are not required to test minimal behaviour. Also using `setGlobalAttributes(of:)` doesn't make sense if we're not asserting them in the test.

ncreated (changes requested):
The implementation looks fine 👍, although we need a change in tests. Also, as we're adding user-facing API:
- please add `CHANGELOG.md` entry

DatadogRUM/Sources/RUMMonitorProtocol.swift
```diff
@@ -53,6 +53,14 @@ public protocol RUMMonitorProtocol: AnyObject {
 
     // MARK: - session
 
+    /// Get the currently active session ID. Returns `nil` if no sessions are currently active or if
+    /// the current session is sampled out.
+    /// This method uses an asynchronous callback to ensure all pending RUM events have been processed
+    /// up to the moment of the call.
+    /// - Parameters:
+    ///   - completion: the callback that will recieve the current session ID.
+    func getCurrentSessionID(completion: @escaping (String?) -> Void)
```
ncreated, line 62: suggestion/ The `get*` prefix isn't common in Swift, so `currentSessionID(completion:)` might be worth considering. Also, as we're dispatching a callback here, it is worth adding a note on what thread it will be called on (notably: not the main thread in current implementation).

DatadogRUM/Tests/RUMTests.swift
fuzzybinary: I think I'll just delete this test then, and add a test to RUMMonitorTests that checks that the session is `nil` after a call to `stopSession`.

Thoughts?

ncreated (approved):
Looks good 💪

PR #1620, opened 2024-01-09
Do not start Session Replay when it is sampled out

maciejburda:
### What and why?

Customer pointed out that when sample rate for session replay is 0, we still run it's code (including swizzling), but the code remains dormant. It makes sense to provide early return in the enable function, so in cases like using remote config, customer can use the same code and just change the parameter.

### How?

Adds a guard that early returns from the enable function when sampling rate is 0. 

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogSessionReplay/Sources/SessionReplay.swift
+++ b/DatadogSessionReplay/Sources/SessionReplay.swift
@@ -37,6 +37,9 @@ public enum SessionReplay {
                 description: "Datadog SDK must be initialized before calling `SessionReplay.enable(with:)`."
             )
         }
+        guard configuration.replaySampleRate > 0 else {
+            return
+        }
 
         let sessionReplay = try SessionReplayFeature(core: core, configuration: configuration)
         try core.register(feature: sessionReplay)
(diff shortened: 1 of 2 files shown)


DatadogSessionReplay/Sources/SessionReplay.swift
```diff
@@ -37,6 +37,9 @@ public enum SessionReplay {
                 description: "Datadog SDK must be initialized before calling `SessionReplay.enable(with:)`."
             )
         }
+        guard configuration.replaySampleRate > 0 else {
+            return
```
ncreated, line 41: question/ Doesn't this mean SR usage won't be reliably measured through RUM Telemetry once configuration telemetry is finally added to `DatadogSessionReplay`? Who will we pass the [`session_replay_sample_rate`](https://github.com/DataDog/rum-events-format/blob/83f8760b46e9a117b5975cfb592b1803d643ee3e/schemas/telemetry/configuration-schema.json#L64-L70) to core if the feature is not allocated? Note that `session_replay_sample_rate: 0` is valuable information we want to collect.

maxep: We can report the telemetry before the `guard`, not necessarily in the feature init 👍

maxep (approved):
LGTM!

PR #1625, opened 2024-01-10
RUM-1836 feat(otel-tracer): add support for status API

ganeshnj:
### What and why?

Status API

### How?

Status API saves of the status of the span indicating whether the span has recorded errors.

This is done by setting error.message tag on the span.

Along with the updating the tags, given dd-sdk-ios sends logs for errors, we replicate the same behavior and record error on the DDSpan which eventually sends a log entry.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogTrace/Tests/OpenTelemetry/OTelSpanTests.swift
+++ b/DatadogTrace/Tests/OpenTelemetry/OTelSpanTests.swift
@@ -297,7 +297,7 @@ final class OTelSpanTests: XCTestCase {
         XCTAssertEqual(recordedSpan.tags["error.msg"], nil)
     }
 
-    func testStatus_givenStatusError_whenSetStatusCalledWithOkAndUnset() {
+    func testStatus_givenStatusError_whenSetStatusCalledWithUnset() {
         let writeSpanExpectation = expectation(description: "write span event")
         let core = PassthroughCoreMock(expectation: writeSpanExpectation)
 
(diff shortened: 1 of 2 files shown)


DatadogTrace/Tests/OpenTelemetry/OTelSpanTests.swift
```diff
@@ -199,6 +199,127 @@ final class OTelSpanTests: XCTestCase {
...
+        let core = PassthroughCoreMock(expectation: writeSpanExpectation)
+
+        // Given
+        let tracer: DatadogTracer = .mockWith(core: core)
+        let span = tracer.spanBuilder(spanName: "Span").startSpan()
+        span.status = .error(description: "error description")
+
+        // When
+        span.status = .unset
+        span.end()
```
ncreated, line 311: nit/ Test name says `whenSetStatusCalledWithOkAndUnset`, so:
```suggestion
        // When
        span.status = .ok
        span.status = .unset
        span.end()
```

ncreated (approved):
Looks great 👌.

Keep in mind we can use [`oneOrMoreOf(_:)`](https://github.com/DataDog/dd-sdk-ios/blob/18056496f2c70d18acb3858e7b07551b5122bca2/TestUtilities/Helpers/FuzzyHelpers.swift#L19) for testing permutation of effects (such as overwriting group of statuses with another). Here it might be an overkill.

DatadogTrace/Tests/OpenTelemetry/OTelSpanTests.swift
ganeshnj: test name was incorrect.

PR #1626, opened 2024-01-10
RUM-1836 feat(otel-tracer): support multi level tags using key flattening

ganeshnj:
### What and why?

Currently, attributes that are array or set which is composition of primitive types are not supported.

### How?

Attributes are flattened using the recursive flattening.

This is inspired from Java version of tracer https://github.com/DataDog/dd-trace-java/blob/0f986a71765d0bad85ed4e5c3bc6ef44b8cb78fe/dd-java-agent/instrumentation/opentelemetry/opentelemetry-1.4/src/main/java/datadog/trace/instrumentation/opentelemetry14/trace/OtelSpan.java#L51-L63

The idea here is to create a key along with the path and save the value as it is in a flat dictionary (tags). In case of arrays, items are indexed with `key.{index}` pattern.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


ncreated (approved):
Looks fine and its super clear 👌. I'm thinking why such significant change to the format of tags didn't require update in any tests added earlier - perhaps we could be missing integration coverage, no?

What would you say for adding an integration-unit test for this concept (from setting some tags to seeing them in encoded event)?

ganeshnj:
> Looks fine and its super clear 👌. I'm thinking why such significant change to the format of tags didn't require update in any tests added earlier - perhaps we could be missing integration coverage, no?
> 
> What would you say for adding an integration-unit test for this concept (from setting some tags to seeing them in encoded event)?

That's because we are not working with internal of `DatadogTracer`, we know `[String: String]` works perfect and well tested. 

Currently not needed, but when we start to deprecated `OTTracer`, yes then we need to migrate those test cases.

PR #1627, opened 2024-01-10
RUM-2149 Propagate Parent Span

maxep:
### What and why?

Proposal on propagating parent span id to distributing tracing.
Additionally, stop overwriting propagation headers.

### How?
`DatadogTrace` now uses the active span as parent context in distributing tracing, if available.

### Note

Parent span propagation only works with APM, not with RUM-to-APM.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskInterception.swift
+++ b/DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskInterception.swift
@@ -22,11 +22,7 @@ public class URLSessionTaskInterception {
     public private(set) var completion: ResourceCompletion?
     /// Trace information propagated with the task. Not available when Tracing is disabled
     /// or when the task was created through `URLSession.dataTask(with:url)` on some iOS13+.
-    public private(set) var trace: (
-        traceID: TraceID,
-        spanID: SpanID,
-        parentSpanID: SpanID?
-    )?
+    public private(set) var trace: TraceContext?
     /// The Datadog origin of the Trace.
     ///
     /// Setting the value to 'rum' will indicate that the span is reported as a RUM Resource.
@@ -61,16 +57,8 @@ public class URLSessionTaskInterception {
         )
     }
 
-    public func register(
-        traceID: TraceID,
-        spanID: SpanID,
-        parentSpanID: SpanID?
-    ) {
-        self.trace = (
-            traceID: traceID,
-            spanID: spanID,
-            parentSpanID: parentSpanID
-        )
+    public func register(trace: TraceContext) {
+        self.trace = trace
     }
 
     public func register(origin: String) {
@@ -80,6 +68,22 @@ public class URLSessionTaskInterception {
     /// Tells if the interception is done (mean: both metrics and completion were collected).
     public var isDone: Bool {
         metrics != nil && completion != nil
+    }
+}
+
+public struct TraceContext {
+    public let traceID: TraceID
+    public let spanID: SpanID
+    public let parentSpanID: SpanID?
+
+    public init(
+        traceID: TraceID,
+        spanID: SpanID,
+        parentSpanID: SpanID? = nil
+    ) {
+        self.traceID = traceID
+        self.spanID = spanID
+        self.parentSpanID = parentSpanID
     }
 }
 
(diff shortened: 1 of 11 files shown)


ganeshnj:
I wonder why there is no change in the implementations of `TracePropagationHeadersWriter`?

For example
- https://github.com/DataDog/dd-sdk-ios/blob/eb50bb4d9f2a195613cd0b651eff1c8d780fa277/DatadogInternal/Sources/NetworkInstrumentation/Datadog/HTTPHeadersWriter.swift#L70C57-L70C69
- https://github.com/DataDog/dd-sdk-ios/blob/maxep/RUM-2149/propagate-parent-span/DatadogInternal/Sources/NetworkInstrumentation/W3C/W3CHTTPHeadersWriter.swift#L78C57-L78C69

DatadogCore/Tests/Datadog/Tracing/TracingURLSessionHandlerTests.swift
```diff
@@ -227,4 +227,47 @@ class TracingURLSessionHandlerTests: XCTestCase {
...
+        // When
+        span.setActive() // start root span
+        let feature = core.get(feature: NetworkInstrumentationFeature.self)
+        // simulate a URLSessionTask intereception
+        feature?.intercept(task: task, additionalFirstPartyHosts: .init(["https://www.example.com"]))
+        feature?.task(task, didFinishCollecting: .mockAny())
+        feature?.task(task, didCompleteWithError: nil)
+        feature?.flush()
+        //
+        span.finish() // finish root span
```
ncreated, line 255: blocker/ This test can work because both are called on the same thread:
- `span.setActive()`
- `feature.intercept(task:additionalFirstPartyHosts:)` 

How do we guarantee they will be called on the same thread in real use case? Note that `span` is user object and `intercept(task:additionalFirstPartyHosts:)` is an internal method in the SDK.

We must add integration test that uses high-level APIs (ideally only `public`), not a one that simulates behaviour by calling internal methods.

ncreated (changes requested):

DatadogTrace/Sources/Integrations/TracingURLSessionHandler.swift
```diff
@@ -34,7 +34,10 @@ internal struct TracingURLSessionHandler: DatadogURLSessionHandler {
             return request
         }
 
-        let spanContext = tracer.createSpanContext()
+        // Use the current active span as parent if the propagation
+        // headers support it.
+        let parentSpanContext = tracer.activeSpan?.context as? DDSpanContext
```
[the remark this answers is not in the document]
maxep: Sorry for the late reply.. that's a good call, it is not ensured and even wrong [here](https://github.com/DataDog/dd-sdk-ios/pull/1627/files#diff-735071eb9c8831222ff7fd735aa0e7b3c9a35c2a81e071c93f0eea3e03eab5bdR84). Let me revisit!

DatadogCore/Tests/Datadog/Tracing/TracingURLSessionHandlerTests.swift
maxep: We can't guarantee that the `span.setActive()` is called on the same thread. If it's not, then the parent won't be propagated and that's expected.

> Note that span is user object and intercept(task:additionalFirstPartyHosts:) is an internal method in the SDK.
Yes, I'm aware. For this test specifically, I'm following the pattern of this suite. But I can add a more high level test indeed.

maxep:
@ganeshnj 
> I wonder why there is no change in the implementations of TracePropagationHeadersWriter?

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskInterception.swift
```diff
@@ -61,16 +57,8 @@ public class URLSessionTaskInterception {
...
-        traceID: TraceID,
-        spanID: SpanID,
-        parentSpanID: SpanID?
-    ) {
-        self.trace = (
-            traceID: traceID,
-            spanID: spanID,
-            parentSpanID: parentSpanID
-        )
+    public func register(trace: TraceContext) {
```
maciejburda, line 60: We can potentially make `var trace` `public` and drop the `register()` function. I assume it's protocol conformance only?

maciejburda (approved):
One small suggestion, otherwise LGTM

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskInterception.swift
maxep: Agreed, tho I'm following the pattern in place for the class.

Issue #1618, opened 2024-01-08
How to flush SDK data for testing purpose?


matas-ramanauskas-revel:
### The issue
I'm writing tests for logging Library that uses this sdk. I want to write a few logs and flush them then test with API if they are in the correct place and accessable. The only way to do it I found so far is flushAndDeinitialize which is hidden behind "DD_SDK_COMPILED_FOR_TESTING". I set this flag in my app, but it seems not to reach inside the library still. Is there a way to make it work for testing purposes?

### Reproduction
```swift
// package.swift
    targets: [
        .executableTarget(
            name: "test-dd",
            dependencies: [
                .product(name: "DatadogCore", package: "dd-sdk-ios"),
                .product(name: "DatadogLogs", package: "dd-sdk-ios")
            ],
            path: "Sources",
            swiftSettings: [.define("DD_SDK_COMPILED_FOR_TESTING")]
        ),
```

```swift
// main.swift
...
Datadog.flushAndDeinitialize()
```

---

#### Datadog SDK version:

.upToNextMajor(from: "2.0.0") 

#### Last working Datadog SDK version:

none

#### Dependency Manager:

SPM

#### Xcode version:

15.1

#### Swift version:

5.8

#### Deployment Target:

iOS ipad (10 gen)

#### macOS version:

Ventura

ncreated:
Hello @matas-ramanauskas-revel 👋. It isn't possible to set [`swiftSettings`](https://developer.apple.com/documentation/packagedescription/target/swiftsettings) for target's dependency in SPM - like you noticed, it is available on [`Target`](https://developer.apple.com/documentation/packagedescription/package/targets), not [`Target.Dependency`](https://developer.apple.com/documentation/packagedescription/target/dependency). I think this design makes sense, as one may want to use the same package as dependency to multiple targets and if it is configured differently each time, SPM won't be able to resolve it.

We use `DD_SDK_COMPILED_FOR_TESTING` compilation condition on purpose: this API is not meant to be publicly available and we only use it internally in our tests. If you seek on basing your tests on Datadog iOS SDK's internals, you can leverage `@testable import` and call [the internal implementation of flush method](https://github.com/DataDog/dd-sdk-ios/blob/a43c63dd422ffd050ed24a1ac1dfd7b156f063c2/DatadogCore/Sources/Datadog.swift#L491):
```swift
@testable import DatadogCore

Datadog.internalFlushAndDeinitialize()
```

[...]

PR #1628, opened 2024-01-11
GH1618 Simplify (internal) flushing logic

ncreated:
### What and why?

🧰 The internal `Datadog.internalFlushAndDeinitialize()` does not work without setting `DD_SDK_COMPILED_FOR_TESTING` compiler flag. To simplify the reasoning about internal "flush", I propose to remove this macro from downstream logic. It will be still used to hide the public API behind this feature, see:

https://github.com/DataDog/dd-sdk-ios/blob/a43c63dd422ffd050ed24a1ac1dfd7b156f063c2/DatadogCore/Sources/Datadog.swift#L485-L489

Solves #1618

### How?

The "flush" logic was protected twice:
- first, the public API was hidden behind `DD_SDK_COMPILED_FOR_TESTING` flag,
- second, part of its implementation was conditionally enabled on `DD_SDK_COMPILED_FOR_TESTING`.

The second occurrence was redundant and it is removed.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
+++ b/DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
@@ -157,13 +157,11 @@ internal class FilesOrchestrator: FilesOrchestratorType {
                 .compactMap { try deleteFileIfItsObsolete(file: $0.file, fileCreationDate: $0.creationDate) }
                 .sorted(by: { $0.creationDate < $1.creationDate })
 
-            #if DD_SDK_COMPILED_FOR_TESTING
             if ignoreFilesAgeWhenReading {
                 return filesFromOldest
                     .prefix(limit)
                     .map { $0.file }
             }
-            #endif
 
             let filtered = filesFromOldest
                 .filter {


DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
```diff
@@ -157,13 +157,11 @@ internal class FilesOrchestrator: FilesOrchestratorType {
                 .compactMap { try deleteFileIfItsObsolete(file: $0.file, fileCreationDate: $0.creationDate) }
                 .sorted(by: { $0.creationDate < $1.creationDate })
 
-            #if DD_SDK_COMPILED_FOR_TESTING
```
maciejburda, line 160: Are we sure about this? When we do this "open" batch is going to be included.

ncreated: This logic is conditioned by `ignoreFilesAgeWhenReading` flag enabled from [`core.flushAndTearDown()`](https://github.com/DataDog/dd-sdk-ios/blob/36862590661791e1ac6a0d61fbaf863c94cea97d/DatadogCore/Sources/Core/DatadogCore.swift#L206-L208) which is only available when `DD_SDK_COMPILED_FOR_TESTING` macro is set:
https://github.com/DataDog/dd-sdk-ios/blob/36862590661791e1ac6a0d61fbaf863c94cea97d/DatadogCore/Sources/Datadog.swift#L485-L489

so I don't see how this may interrupt `DataUploadWorker` in production apps. The entire concept of "ignoring files age" was introduced for testing purpose (we want to force-send all data in tests). Does it sound ok to you @maciejburda?

Ultimately, I don't like using macros in this way as they shred the codebase. Since it was introduced a long time ago however, we managed to keep the impact of `DD_SDK_COMPILED_FOR_TESTING` very minimal. A better approach would be to abstract certain components and inject different implementation in tests - it implies a much bigger effort, but I hope we do it one day.

maciejburda: Not sure what am I missing, but the modified function `getReadableFiles(excludingFilesNamed excludedFileNames: Set<String> = [], limit: Int = .max)` is being called by the `DataUploadWorker``s `FileReader`.

ncreated: The logic we enable here is conditioned by `ignoreFilesAgeWhenReading` flag 🙂:
```swift
if ignoreFilesAgeWhenReading {
   return filesFromOldest
      .prefix(limit)
      .map { $0.file }
}
```
And this is only set to `true` in our tests 💡.

maciejburda: Ah, I get it now! Thanks for clarifying :)

maciejburda (approved):

maxep (approved):
👍

PR #1629, opened 2024-01-11
Add core observer for UI level testing

louiszawadzki:
### What and why?

This is the next evolution from the POC that was done in https://github.com/DataDog/dd-sdk-ios/pull/1570
I've changed the pattern from a Proxy to an observer as it is less intrusive in the core logic.

I've hidden the feature behind the `DD_SDK_COMPILED_FOR_TESTING` compilation condition so it's not available to users.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


ncreated (changes requested):
@louis, I appreciate your work on this PR and would like to discuss its goals further. 

Overall, I'm concerned about the approach taken in this PR. The use of the `DD_SDK_COMPILED_FOR_TESTING` flag pushes us further into technical debt and doesn't align with our design principles, which favor abstractions and dependency injection over such flags.

I suggest considering the proxy approach proposed in #1570 for better maintainability. It's a technique we already use in our SDK tests, reducing the introduction of redundant concepts. I'm curious about why the proxy definition is added to the SDK itself instead of implementing it directly in E2E tests.

Additionally, for E2E testing, I recommend exploring the injection of a spy implementation for the [`HTTPClient`](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogCore/Sources/Core/Upload/HTTPClient.swift) (as done in our integration tests [here](https://github.com/DataDog/dd-sdk-ios/blob/36862590661791e1ac6a0d61fbaf863c94cea97d/DatadogCore/Sources/Datadog.swift#L225-L228)). This approach provides more valuable insights, such as ensuring data upload, knowing the destination URL and authentication details, and understanding the exact data uploaded. It is much more than we can know from individual events intercepted in Event Write Context.

Let's discuss these considerations further to ensure we're on the right track.

PR #1630, opened 2024-01-15
RUM-1836 feat(otel-tracer): add support for span links

ganeshnj:
### What and why?

Span link is an otel concept which allows casual relationship among spans. This is similar to parent but it can be many-many.

### How?

In datadog, span links are supported via `_dd.span_links` tag which takes a list of links.

A link is represented with 
```json
 {
     "trace_id": "<exactly 32 character, zero-padded lower-case hexadecimal encoded trace id>",
     "span_id": "<exactly 16 character, zero-padded lower-case hexadecimal encoded span id>",
     "attributes": {"key":"value", "pairs":"of", "arbitrary":"values"},
     "dropped_attributes_count": <decimal 64 bit integer>,
     "tracestate": "a tracestate as defined in the W3C standard",
     "flags": <an integer representing the flags as defined in the W3C standard>
 }
```

Because RUM backend only supports [String: String] tags, this implementation escapes the span links JSON to make it a string value.

This
```swift
let tracer = Tracer.shared()
let span1 = tracer.spanBuilder(spanName: "span1").setSpanKind(spanKind: .client).startSpan()
span1.end()

let span2 = tracer.spanBuilder(spanName: "span2").setSpanKind(spanKind: .client).startSpan()
span2.end()

let spanBuilder = tracer.spanBuilder(spanName: "span3").setSpanKind(spanKind: .client)
spanBuilder.addLink(spanContext: span1.context)
spanBuilder.addLink(spanContext: span2.context)
let span3 = spanBuilder.startSpan()
span3.end()
```

results in 
image

We are also using lower 64 bits for trace id as the SDK doesn't support 128 bit ids yet.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


maciejburda (approved):
LGTM

PR #1640, opened 2024-01-19
RUM-2099 fix: Resolve Xcode 15 issues in local development

ncreated:
### What and why?

🧰 This PR adjusts our project so it can be developed (build → test → integrate) with Xcode 15 and iOS 17.x simulators. It makes no changes to CI setup, where we still use Xcode 14 (to build, test, integrate and release).

### How?

All changes are annotated in PR comments, with the biggest being:
- **Dropping iOS 11**: Xcode 15 drops support for iOS 11 deployment target. To solve compatibility warnings, I had to adjust the minimum deployment target in Xcode 15. Done through `Base.local.xcconfig`, which gives us flexibility to use `11.0` on CI (Xcode 14) and `12.0` in local (if developing with Xcode 15):
```
IPHONEOS_DEPLOYMENT_TARGET=12.0
```
- The order of keys in serialized JSONs is no longer stable in iOS 17. To address assertion failures, I updated failing tests to not depend on the keys order by sorting them.
- Smaller changes to fix the way of mocking `UIView` and `UIViewController` instances on iOS 17.
- Some updates in SR code to make it not depend on iOS runtime version.

### Review checklist

- [x] Run integration tests
- [x] Run smoke tests
- [ ] Run tests for `tools/`


Datadog/Datadog.xcodeproj/project.pbxproj
```diff
@@ -9098,7 +9092,6 @@
 				GENERATE_INFOPLIST_FILE = YES;
 				INFOPLIST_KEY_NSHumanReadableCopyright = "Copyright © 2023 Datadog. All rights reserved.";
 				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
-				IPHONEOS_DEPLOYMENT_TARGET = 11.0;
```
ncreated, line 9101: Removing the hardcoded value override in pbxproj. From now on it will use the one set in `base.local.xcconfig`:
```
// To fix Xcode 15 warning: 'The iOS Simulator deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is
// set to 11.0, but the range of supported deployment target versions is 12.0 to 17.0.99.'
IPHONEOS_DEPLOYMENT_TARGET=12.0
```
☝️ add this to `Base.local.xcconfig` or run `make dependencies` to fix compiler warnings in Xcode 15

DatadogCore/Tests/Datadog/Mocks/RUMFeatureMocks.swift
```diff
@@ -799,7 +799,8 @@ func createMockView(viewControllerClassName: String) -> UIViewController {
         theClass = newClass
     }
 
-    let viewController = theClass.alloc() as! UIViewController
+    let viewController = UIViewController()
+    object_setClass(viewController, theClass)
     mockWindow.rootViewController = viewController
     mockWindow.makeKeyAndVisible()
     return viewController
```
ncreated, line 806: Fixes UIKit runtime exception (tests only):
```
UIViewController is missing its initial trait collection populated during initialization. This is a serious bug,
likely caused by accessing properties or methods on the view controller before calling a UIViewController initializer.
```

DatadogInternal/Tests/Codable/AnyCodableTests.swift
```diff
@@ -184,6 +184,7 @@ class AnyCodableTests: XCTestCase {
 
     func testGivenEncodableValueWrappedIntoCodableValue_whenEncoding_itProducesExpectedJSONRepresentation() throws {
         let encoder = JSONEncoder()
+        encoder.outputFormatting = [.sortedKeys]
```
ncreated, line 187: The order of keys in serialized JSON varies between iOS 17 and prior versions. We sort them, to get stable result.

Datadog/Datadog.xcodeproj/project.pbxproj
ncreated: With `Base.local.xcconfig` being uniformly set among all targets, this also fixes `IntegrationTests` build error in Xcode 15:
```
'Compiling for iOS 11.0, but module 'SRFixtures' has a minimum deployment target of iOS 12.0'
```

DatadogSessionReplay/Tests/Mocks/SRDataModelsMocks.swift
```diff
@@ -240,8 +240,8 @@ extension SRShapeStyle: AnyMockable, RandomMockable {
     public static func mockRandom() -> SRShapeStyle {
         return SRShapeStyle(
             backgroundColor: .mockRandom(),
-            cornerRadius: .mockRandom(),
-            opacity: .mockRandom()
+            cornerRadius: .randomInteger(),
+            opacity: .randomInteger()
```
ncreated, line 244: We no longer mock random `Double` as it was giving different results when serializing high-precission numbers with `Codable` vs `JSONSerialization` on iOS 17. In some tests we compare JSON strings, so to avoid flakiness now we only randomize these values among integer numbers from a realistic range (not the full `Double` range).

DatadogSessionReplay/Tests/Recorder/Utilties/ImageDataProviderTests.swift
```diff
@@ -24,6 +24,9 @@ class ImageDataProviderTests: XCTestCase {
     }
 
     func test_returnsValidStringAndIdentifier_WhenContentIsValid() throws {
+        if #available(iOS 17, *) {
+            throw XCTSkip("TODO: RUM-2969 Fix image hashing tests for iOS 17+")
+        }
```
ncreated, line 29: The `image.pngData()` gives different results between iOS 17 and prior versions (results seem to be stable between iOS 16 and 15). I was unable to solve this problem with refactoring the code due to too much dependency on static variables and `UIImage` extension. Bigger refactoring looked risky as this code involves threading assumptions that I have no full expertise on. Behaviour Team's eye on this looks required.

Ultimately, I decided to skip these tests on iOS 17 until the problem is addressed by Behaviour Team in `RUM-2969`.

DatadogTrace/Sources/Utils/JSONEncoder.swift
ncreated, line 1: Removing - it wasn't used anywhere 🧽

Datadog/Datadog.xcodeproj/xcshareddata/xcschemes/DatadogRUM iOS.xcscheme
```diff
@@ -29,8 +29,7 @@
       shouldUseLaunchSchemeArgsEnv = "YES">
       <Testables>
          <TestableReference
-            skipped = "NO"
-            parallelizable = "YES">
+            skipped = "NO">
```
ganeshnj, line 32: errr - means longer execution time in CI?

DatadogCore/Tests/Datadog/TracerTests.swift
```diff
@@ -144,7 +144,7 @@ class TracerTests: XCTestCase {
         XCTAssertEqual(try spanMatcher.operationName(), "operation")
         XCTAssertEqual(try spanMatcher.startTime(), 1_576_404_000_000_000_000)
         XCTAssertEqual(try spanMatcher.duration(), 500_000_000)
-        XCTAssertEqual(try spanMatcher.dd.samplingRate(), 0.42)
+        XCTAssertEqual(try spanMatcher.dd.samplingRate(), 0.42, accuracy: 0.01)
```
ganeshnj, line 147: what is context behind accuracy here?

This should be well predictable value IMO.

DatadogTrace/Tests/Span/SpanEventBuilderTests.swift
```diff
@@ -410,6 +410,7 @@ class SpanEventBuilderTests: XCTestCase {
 
     func testWhenBuildingSpan_itConvertsTagValuesToString() {
         let builder: SpanEventBuilder = .mockAny()
+        builder.attributesEncoder.outputFormatting = [.sortedKeys]
```
ganeshnj, line 413: can be done in `.mockAny` to avoid repetition?

Datadog/Datadog.xcodeproj/xcshareddata/xcschemes/DatadogRUM iOS.xcscheme
ncreated: This was on purpose as running tests with this scheme in local very rarely succeeds (it mostly hangs on Simulator launch). It works fine without parallel testing. We don't have this option enabled in any other scheme.

This scheme is not used by CI. It uses [`DatadogCore iOS.xscheme`](https://github.com/DataDog/dd-sdk-ios/blob/d51149cac0560bf947e46dd9b472dba9b97c532c/Datadog/Datadog.xcodeproj/xcshareddata/xcschemes/DatadogCore%20iOS.xcscheme#L230-L239) instead, which includes all targets.

I can revert to keep this PR clean, but disabling this option is the first thing I do each time 😉.

DatadogCore/Tests/Datadog/TracerTests.swift
ncreated: It is a fair call and may reveal an issue (not related to this PR though).

We changed from decoding `Float` to decoding `Double` in this PR:
```diff
-       func samplingRate() throws -> Float { try matcher.dd(forKeyPath: "_dd.agent_psr") }
+       func samplingRate() throws -> Double { try matcher.dd(forKeyPath: "_dd.agent_psr") }
```
The original value in this test is `42` (`Float`). It gets divided by `100` (`Float`) in implementation and serialized to JSON. I went deeper and I see it gets serialized as:
```
"_dd.agent_psr" = "0.4199999868869781";
```
hence by using `accuracy` we cover the actual problem. I'd go with `accuracy` to mitigate the problem, but the serialised precision is IMO worth opening a JIRA. WDYT @ganeshnj ?

DatadogTrace/Tests/Span/SpanEventBuilderTests.swift
ncreated: done 👍

DatadogCore/Sources/Core/Context/CarrierInfoPublisher.swift
```diff
@@ -64,6 +64,12 @@ extension CarrierInfo {
 
 // MARK: - iOS 11
 
+// Xcode 15 ships with Swift 5.9 and drops support for iOS 11 deployment target.
+// The conditional check ensures that iOS 11 components are only compiled with Xcode 14, where iOS 11
+// remains a valid deployment target. Since this deployment target is no longer available in Xcode 15,
+// compiling the following code in Xcode 15 would result in compiler warnings (due to deprecation in some used APIs),
+// violating our 0 warnings policy. It can be entirely removed when Xcode 14 is dropped.
+#if swift(<5.9)
```
maxep, line 72: /question can we use `@available(iOS, deprecated: 12, obsoleted: 13, message: "something") instead or does it show a warning still?

maxep (approved):
Looks great 👍

DatadogCore/Tests/Datadog/TracerTests.swift
ganeshnj: let's open a Jira ticket.

the original root cause is 🥁 🥁 🥁 

https://github.com/apple/swift-corelibs-foundation/issues/4255

DatadogCore/Sources/Core/Context/CarrierInfoPublisher.swift
ncreated: Actually, it was even simpler. What worked is: `@available(iOS, deprecated: 12)`. No warnings with 11.0, 12.0 or 13.0 deployment target (both in Xcode 14.x and 15.x). ✅ changed. Thanks, a good call 👍.

ncreated:
> I believe here Xcode 15 + Swift 5.9 was used for testing?
> 
> Just want to confirm as this is going to be the standard combination at most local machines for quite some time.

Yes 👌, Xcode 15 and Xcode 15.2.

DatadogCore/Tests/Datadog/TracerTests.swift
ncreated: The issue in Swift stdlib is on "decoding", while we see it earlier in "encoding". I opened [RUM-3011](https://datadoghq.atlassian.net/browse/RUM-3011) with full context. Again, this is not a regression, so not blocking. Problem seems specific to iOS 16.2 rather than Xcode (Swift) version:
- 🐞 Xcode 15.0, iOS 16.2: "_dd.agent_psr":0.42500001192092896
- 🐞 Xcode 14.3.1, iOS 16.2: "_dd.agent_psr":0.42500001192092896
- ✅ Xcode 14.3.1, iOS 17.0: "_dd.agent_psr":0.425

maxep (approved):
LGTM 👍

ganeshnj (approved):

PR #1645, opened 2024-01-24
RUM-1833 feat: Link (emergency) crash Logs to crashed RUM session

ncreated:
### What and why?

📦 This PR adds link to the RUM session for crash reports sent as emergency Log event.

See:

### How?

With the "last `RUMViewEvent`" being already encoded in `CrashContext.data`, this PR adds extra logic in `DatadogLogs` to decode RUM session attributes. Decoding is limited only to these fields that are required so we don't decode the whole event (that wouldn't be even possible as `DatadogLogs` has no dependency on `DatadogRUM`).

Once decoded, these attributes are attached to `LogEvent` following the contract between SDK and Datadog app.

Except unit tests covering edge cases (like malformed RUM view information), this PR also adds integration unit tests for "Sending Crash Reports as Logs and RUM".

Last, I took extra time to document logs, spans and RUM correlations in [this spec (internal)](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3416820413/Event+Correlations+-+SDK+Datadog+App+Contracts).

### Review checklist

- [x] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogLogs/Sources/Feature/MessageReceivers.swift
+++ b/DatadogLogs/Sources/Feature/MessageReceivers.swift
@@ -85,7 +85,7 @@ internal struct LogMessageReceiver: FeatureMessageReceiver {
             return true
         } catch {
             core.telemetry
-                .error("Fails to decode crash from Logs", error: error)
+                .error("Failed to decode log message in `LogMessageReceiver`", error: error)
         }
 
         return false
@@ -152,6 +152,27 @@ internal struct CrashLogReceiver: FeatureMessageReceiver {
         let carrierInfo: CarrierInfo?
         /// Current user information.
         let userInfo: UserInfo?
+
+        /// A type representing part of last RUM view information required to link crash log with previous RUM session.
+        /// It mirrors the schema of `RUMViewEvent`, so we can decode it from the last `RUMViewEvent` coded in crash context.
+        struct PartialRUMViewEvent: Decodable {
+            struct Application: Decodable {
+                let id: String
+            }
+            struct Session: Decodable {
+                let id: String
+            }
+            struct View: Decodable {
+                let id: String
+            }
+
+            let application: Application
+            let session: Session
+            let view: View
+        }
+
+        /// The last RUM view in crashed app process.
+        let lastRUMViewEvent: PartialRUMViewEvent?
     }
 
     /// Time provider.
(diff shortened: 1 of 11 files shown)


DatadogLogs/Sources/Feature/MessageReceivers.swift
```diff
@@ -171,7 +192,7 @@ internal struct CrashLogReceiver: FeatureMessageReceiver {
             return send(report: crash.report, with: crash.context, to: core)
         } catch {
             core.telemetry
-                .error("Fails to decode crash from RUM", error: error)
+                .error("Failed to decode crash message in `LogMessageReceiver`", error: error)
```
ncreated, line 195: This is to differentiate it from [other telemetry](https://github.com/DataDog/dd-sdk-ios/blob/d51149cac0560bf947e46dd9b472dba9b97c532c/DatadogRUM/Sources/Integrations/CrashReportReceiver.swift#L146) with the same message.

DatadogLogs/Sources/Feature/MessageReceivers.swift
```diff
@@ -152,6 +152,27 @@ internal struct CrashLogReceiver: FeatureMessageReceiver {
...
+                let id: String
+            }
+            struct View: Decodable {
+                let id: String
+            }
+
+            let application: Application
+            let session: Session
+            let view: View
+        }
```
ganeshnj, line 172: Is there any public/internal doc explaining how this correlation work? In case when there an update needed we can always refer.

Also, we have been doing quite a few changes to make the correlation work and so far they have been very organic.  These things impact SDK and we should document them somewhere.

ganeshnj (approved):
Looks, good overall coverage.

A general comment on improving the documentation. Not blocking.

DatadogLogs/Sources/Feature/MessageReceivers.swift
ncreated: Fair call 👍. I took an extra time to document logs, spans and RUM correlations in [this spec (internal)](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3416820413/Event+Correlations+-+SDK+Datadog+App+Contracts).

PR #1653, opened 2024-02-05
RUM-1958 Enable SDK compilation on VisionOS

maciejburda:
### What and why?

Provides build time issue mitigations to enable compilation on VisionOS.

*This PR is not adding official VisionOS support - just resolves compilation issues in multi target environments.*

### Review checklist

- [x] Run integration tests
- [x] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogRUM/Sources/RUMVitals/VitalInfoSampler.swift
+++ b/DatadogRUM/Sources/RUMVitals/VitalInfoSampler.swift
@@ -43,7 +43,7 @@ internal final class VitalInfoSampler {
 
     private static var maximumFramesPerSecond: Double {
         get {
-            #if os(visionOS)
+            #if swift(>=5.9) && os(visionOS)
             return 120.0 // Hardcoded to enable VisionOS compilation
             #else
             return Double(UIScreen.main.maximumFramesPerSecond)
--- a/DatadogRUM/Sources/RUMVitals/VitalInfoSampler.swift
+++ b/DatadogRUM/Sources/RUMVitals/VitalInfoSampler.swift
@@ -41,12 +41,22 @@ internal final class VitalInfoSampler {
 
     private var timer: Timer?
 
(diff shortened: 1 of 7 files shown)


ncreated (changes requested):
This looks fine 👍, but we need to also update [SUPPORTED-VERSIONS.md](https://github.com/DataDog/dd-sdk-ios/blob/develop/SUPPORTED-VERSIONS.md) with a clear take on the fact that we only enable compilation for VisionOS, without really supporting it. From that point of view, it is looks similar to Catalyst. This blocker.

Suggestion/ Also, we can consider adding similar warning to VisionOS as we implement for Catalyst, WDYT?
https://github.com/DataDog/dd-sdk-ios/blob/b5417c34c484da450ae7aa506de22275a0cb7c8d/DatadogCore/Sources/Datadog.swift#L365-L367

DatadogRUM/Sources/RUMVitals/VitalInfoSampler.swift
```diff
@@ -41,12 +41,22 @@ internal final class VitalInfoSampler {
 
     private var timer: Timer?
 
+    private static var maximumFramesPerSecond: Double {
+        get {
+            #if swift(>=5.9) && os(visionOS)
+            return 120.0 // Hardcoded to enable VisionOS compilation
```
maxep, line 47: /question Would it be possible to not report the refresh rate instead of using a hardcoded and possibly wrong frame rate?

maciejburda: This is just `maximum` frames per second.

It's a non-official support only. I think that exploring the right solution here requires some more demand from customers.

WDYT?

maxep (approved):
Looks good to me!

ncreated (approved):
🤿 🚀

PR #1654, opened 2024-02-05
RUM-2700 chore: Shadow unit and integration test runs in GitLab

ncreated:
### What and why?

⚙️ This PR adds GitLab configuration to run unit and integration tests. This is to start shadowing their runs using new runner. No change to Bitrise CI configuration, which remains to be our main CI.

We use Xcode 15.0.1 on GitLab with running tests on iOS 17.

### How?

Added simple GitLab pipeline:
```
[ ENV info ]
     ↓
[ Lint ]
     ↓
[SDK unit tests (iOS)] + [SDK unit tests (tvOS)] + [SDK integration tests (iOS)]
```
In GitLab stages are sequential and jobs are parallel, meaning the test jobs will run in parallel.

🎁 I had to fix few remaining Xcode 15 x iOS 17 issues that were not spotted in #1640.

For the moment, GitLab status seems to be synced into this PR, so it may block it if the pipeline fails. My guess is we need to disable it in branch protection settings (on GitHub). This will be likely possible only after we merge this PR into `develop` (right now GitHub doesn't list these GitLab checks as possible statuses).

Screenshot 2024-02-06 at 09 59 54

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,81 @@
+stages:
+  - info
+  - lint
+  - test
+
+ENV info:
+  stage: info
+  tags:
+    - mac-ventura-preview
+  script:
+    - system_profiler SPSoftwareDataType # system info
+    - xcodebuild -version
+    - xcode-select -p # default Xcode
+    - ls /Applications/ | grep Xcode # other Xcodes
+    - xcodebuild -workspace "Datadog.xcworkspace" -scheme "DatadogCore iOS" -showdestinations -quiet # installed iOS destinations
+    - xcodebuild -workspace "Datadog.xcworkspace" -scheme "DatadogCore tvOS" -showdestinations -quiet # installed tvOS destinations
+    - xcbeautify --version
+    - swiftlint --version
+    - carthage version
+    - gh --version
+    - brew -v
+    - bundler --version
+    - python3 -V
+
+Lint:
+  stage: lint
+  tags:
+    - mac-ventura-preview
+  script:
+    - ./tools/lint/run-linter.sh
+    - ./tools/license/check-license.sh
+
+SDK unit tests (iOS):
+  stage: test
+  tags:
+    - mac-ventura-preview
+  variables:
+    TEST_WORKSPACE: "Datadog.xcworkspace"
+    TEST_DESTINATION: "platform=iOS Simulator,name=iPhone 15 Pro Max,OS=17.0.1"
+  script:
+    - make dependencies-gitlab
+    - xcodebuild -workspace "$TEST_WORKSPACE" -destination "$TEST_DESTINATION" -scheme "DatadogCore iOS" -only-testing:"DatadogCoreTests iOS" test | xcbeautify
+    - xcodebuild -workspace "$TEST_WORKSPACE" -destination "$TEST_DESTINATION" -scheme "DatadogCore iOS" -only-testing:"DatadogInternalTests iOS" test | xcbeautify
+    - xcodebuild -workspace "$TEST_WORKSPACE" -destination "$TEST_DESTINATION" -scheme "DatadogCore iOS" -only-testing:"DatadogLogsTests iOS" test | xcbeautify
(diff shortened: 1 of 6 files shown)


Datadog/Example/Utils/ConsoleOutputInterceptor.swift
```diff
@@ -21,7 +22,7 @@ class ConsoleOutputInterceptor {
         consolePrint = self.process
     }
 
-    private func process(newLog: String) {
+    private func process(newLog: String, level: CoreLoggerLevel) {
```
ncreated, line 25: This code wasn't compiling when `#if DEBUG` condition is resolved.

DatadogCore/Tests/Datadog/DatadogCore/Context/CarrierInfoPublisherTests.swift
```diff
@@ -45,6 +45,7 @@ class CarrierInfoPublisherTests: XCTestCase {
         }
     }
 
+    @available(iOS, deprecated: 12)
     func testGivenCellularServiceUnavailableOnIOS11_whenReadingCurrentCarrierInfo_itReturnsNoValue() {
         // Given
         let reader = iOS11CarrierInfoReader(networkInfo: unavailableCTTelephonyNetworkInfo)
```
ncreated, line 51: Needed to deprecate it as we're using `iOS11CarrierInfoReader` here, deprecated on iOS 12. If not deprecating the test, it was causing build warnings in Xcode 15 (`IPHONEOS_DEPLOYMENT_TARGET=12.0`) and because we treat them as errors → build failure.

maciejburda (approved):
LGTM

.gitlab-ci.yml
```diff
@@ -0,0 +1,81 @@
...
+    - xcodebuild -workspace "$TEST_WORKSPACE" -destination "$TEST_DESTINATION" -scheme "DatadogCore iOS" -only-testing:"DatadogInternalTests iOS" test | xcbeautify
+    - xcodebuild -workspace "$TEST_WORKSPACE" -destination "$TEST_DESTINATION" -scheme "DatadogCore iOS" -only-testing:"DatadogLogsTests iOS" test | xcbeautify
+    - xcodebuild -workspace "$TEST_WORKSPACE" -destination "$TEST_DESTINATION" -scheme "DatadogCore iOS" -only-testing:"DatadogTraceTests iOS" test | xcbeautify
+    - xcodebuild -workspace "$TEST_WORKSPACE" -destination "$TEST_DESTINATION" -scheme "DatadogCore iOS" -only-testing:"DatadogRUMTests iOS" test | xcbeautify
+    - xcodebuild -workspace "$TEST_WORKSPACE" -destination "$TEST_DESTINATION" -scheme "DatadogCore iOS" -only-testing:"DatadogWebViewTrackingTests iOS" test | xcbeautify
+    - xcodebuild -workspace "$TEST_WORKSPACE" -destination "$TEST_DESTINATION" -scheme "DatadogSessionReplay iOS" test | xcbeautify
+    - xcodebuild -workspace "$TEST_WORKSPACE" -destination "$TEST_DESTINATION" -scheme "DatadogCrashReporting iOS" test | xcbeautify
```
ganeshnj, line 49: food for thought, can you run test for all targets in parallel on different containers.

Something like https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

ganeshnj (approved):

.gitlab-ci.yml
ncreated: @ganeshnj that would mean running each as separate GitLab stage, right? We could see what is the overhead of this (for sure the clean build time) vs benefit of parallelization. Total run of unit tests stage is ~2-4minutes, so it might be counter-optimisation.

ganeshnj: Yes, they will be different runners. 2-4 minutes is good and looks like we don't need any parallel execution.

Ideally there shouldn't be a clean build, the past state must be forwarded to next state.

ncreated: That would be possible with passing artifacts between jobs 👌. Let's keep this in mind. For now doesn't sound necessary so we can start simple.

PR #1663, opened 2024-02-08
RUM-2818 New Replay Payload Format

maxep:
### What and why?

The replay endpoint now supports multiple segments in multipart forms.

### How?

Follow specs from https://github.com/DataDog/dd-sdk-android/pull/1805

### Review checklist

- [x] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogSessionReplay/Sources/Feature/RequestBuilders/Multipart/MultipartFormData.swift
+++ b/DatadogSessionReplay/Sources/Feature/RequestBuilders/Multipart/MultipartFormData.swift
@@ -8,27 +8,29 @@
 import Foundation
 
 internal protocol MultipartFormDataBuilder {
-    /// The boundary UUID of this multipart form.
-    var boundary: UUID { get }
+    /// The boundary  of this multipart form.
+    var boundary: String { get }
     /// Adds a field.
     mutating func addFormField(name: String, value: String)
     /// Adds a file.
     mutating func addFormData(name: String, filename: String, data: Data, mimeType: String)
     /// Returns the entire multipart body data (as it should be applied to request).
-    var data: Data { get }
+    mutating func build() -> Data
 }
 
 /// A helper facilitating creation of `multipart/form-data` body.
 internal struct MultipartFormData: MultipartFormDataBuilder {
-    let boundary: UUID
-    private var body = Data()
+    private var body: Data
 
-    init(boundary: UUID) {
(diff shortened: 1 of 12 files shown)


DatadogSessionReplay/Sources/Feature/RequestBuilders/Multipart/MultipartFormData.swift
```diff
@@ -8,46 +8,53 @@
...
-    var data: Data {
-        var data = body
-        data.append(string: "--\(boundary.uuidString)--")
-        return data
+    mutating func build() -> Data {
+        defer {
+            // reset builder
+            body = Data()
+            boundary = UUID().uuidString
+        }
```
maxep, line 54: We should reset the builder after build, otherwise we would keep appending data.

DatadogSessionReplay/Sources/Feature/RequestBuilders/Multipart/MultipartFormData.swift
```diff
@@ -8,46 +8,53 @@
 import Foundation
 
 internal protocol MultipartFormDataBuilder {
-    /// The boundary UUID of this multipart form.
-    var boundary: UUID { get }
+    /// The boundary  of this multipart form.
+    var boundary: String { get }
```
maxep, line 12: Now use a string and avoid uuid transformation each time we write the boundary.

DatadogSessionReplay/Sources/Feature/RequestBuilders/Multipart/MultipartFormData.swift
ncreated: How was it reset before this change 🤔💭?

maxep: [We were making a copy of a `struct MultipartFormData` before each request](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogSessionReplay/Sources/Feature/RequestBuilders/SegmentRequestBuilder.swift#L39), but we cannot assume that the `protocol MultipartFormDataBuilder` is applied to a value type.

maxep:
CI status should be green!

image

maciejburda (approved):
Looks good!

PR #1666, opened 2024-02-09
Fix privacy manifest collision

### LINKED ISSUE #1665: Privacy manifest causing collisions through CocoaPods
The `2.7.0` release added a privacy manifest, but its current setup with CocoaPods is causing collisions with other privacy manifest files as you can see below:

image

I believe this is because the `PrivacyInfo.xcprivacy` is referenced as a `resource` in the [podspec](https://github.com/DataDog/dd-sdk-ios/blob/9faa46ce5e93ae453cebb623771a926b44f8ac23/DatadogCore.podspec#L25) instead of using `resource_bundle`. The latter would avoid collisions by putting the privacy manifest file in subfolders for each dependency, instead of going in the app bundle's root. More context on this [here](https://github.com/google/promises/pull/226#discussion_r1447871477). I think the fix would be to simple swap `resource` with `resource_bundle` here.

maxep:
### What and why?

When using static linkage in CocoaPods with `use_frameworks! :linkage => :static`. The pod's resource gets copied at root of the app bundle, resulting in collision with the app's `PrivacyInfo.xcprivacy`.

Resolve #1665

### How?

Use `resource_bundle` to keep the `PrivacyInfo.xcprivacy` in it own bundle.

The smoke tests apps now all define privacy-manifest to catch any collision.

### Review checklist

- [ ] Run integration tests
- [x] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/dependency-manager-tests/carthage/App/ViewController.swift
+++ b/dependency-manager-tests/carthage/App/ViewController.swift
@@ -41,13 +41,6 @@ internal class ViewController: UIViewController {
         RUM.enable(with: .init(applicationID: "app-id"))
         RUMMonitor.shared().startView(viewController: self)
 
-        // DDURLSessionDelegate APIs must be visible:
-        _ = DDURLSessionDelegate()
-        _ = DatadogURLSessionDelegate()
-        class CustomDelegate: NSObject, __URLSessionDelegateProviding {
-            var ddURLSessionDelegate: DatadogURLSessionDelegate { DatadogURLSessionDelegate() }
-        }
-
         // Trace APIs must be visible:
         Trace.enable()
 
--- a/DatadogCore.podspec
+++ b/DatadogCore.podspec
@@ -22,7 +22,10 @@ Pod::Spec.new do |s|
   
   s.source_files = ["DatadogCore/Sources/**/*.swift",
                     "DatadogCore/Private/**/*.{h,m}"]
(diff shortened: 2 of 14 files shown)


dependency-manager-tests/carthage/App/ViewController.swift
```diff
@@ -41,13 +41,6 @@ internal class ViewController: UIViewController {
         RUM.enable(with: .init(applicationID: "app-id"))
         RUMMonitor.shared().startView(viewController: self)
 
-        // DDURLSessionDelegate APIs must be visible:
-        _ = DDURLSessionDelegate()
-        _ = DatadogURLSessionDelegate()
-        class CustomDelegate: NSObject, __URLSessionDelegateProviding {
-            var ddURLSessionDelegate: DatadogURLSessionDelegate { DatadogURLSessionDelegate() }
-        }
```
ncreated, line 49: q/ Why this removal? Due to deprecation warnings treated as build error?

ncreated (approved):
👌 Great we cover this in smoke tests now 👍

dependency-manager-tests/carthage/App/ViewController.swift
maxep: Yes, and because it's deprecated anyway.

PR #1685, opened 2024-02-19
RUM-2925 feat: Report non-fatal App Hangs as RUM errors

ncreated:
### What and why?

📦 First part of non-fatal App Hangs monitoring as RUM errors. App Hangs lasting more than 2s (TBD) will be reported as RUM errors.

Before this change, App Hangs were only tracked as RUM long tasks through [`RUM.LongTaskObserver`](https://github.com/DataDog/dd-sdk-ios/blob/3f2d8a043ec54b0f0aaa719797e97809551d0643/DatadogRUM/Sources/Instrumentation/LongTasks/LongTaskObserver.swift#L73), using the default threshold of 0.1s. Because long tasks are tracked on the main thread (using `CFRunLoopObserver`) it is impossible to capture their backtrace during the hang. Lack of stack traces makes it harder to troubleshoot, hence we implement additional layer of reporting hangs as RUM errors using secondary watchdog thread.

Next PRs will add logic for capturing the stack trace of the hanged main thread. In this PR we only send RUM error `"App Hang"` when the hang ends.

### How?

The solution is using secondary thread (`com.datadoghq.app-hang-watchdog`, called "watchdog thread") to schedule periodic tasks on the main queue, await their completion and determine "hang" or "no hang" situation accordingly:

Screenshot 2024-02-19 at 09 23 33

- at the beginning of watchdog thread loop, main thread **task** is scheduled;
- if **task** is executed under hang `threshold`, there is no hang;
- if **task** completion arrives after `threshold`, SDK considers it a hang.

By using secondary thread, the SDK will capture the stack trace of the main thread between `treshold` and "**task** completion" to associate it with RUM error. This work will be done in next PRs.

To not cause excessive use of CPU, the "watchdog thread" implements "IDLE" periods, which prevent it from scheduling main queue tasks too often. It is defined as `Constants.tolerance` value in watchdog thread implementation. The trade-off is adding maring of error in hangs monitoring, so the whole solution requires tuning to avoid to many false-negatives:

Screenshot 2024-02-19 at 09 32 10

"IDLE" periods make the impact of watchdog thread negligible:

Screenshot 2024-02-19 at 09 28 26

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/Datadog/IntegrationUnitTests/RUM/AppHangsMonitoringTests.swift
+++ b/Datadog/IntegrationUnitTests/RUM/AppHangsMonitoringTests.swift
@@ -14,14 +14,12 @@ class AppHangsMonitoringTests: XCTestCase {
     private var rumConfig = RUM.Configuration(applicationID: .mockAny())
 
     override func setUp() {
-        super.setUp()
         core = DatadogCoreProxy()
     }
 
     override func tearDown() {
         core.flushAndTearDown()
         core = nil
-        super.tearDown()
     }
 
     func testWhenMainThreadHangsAboveThreshold_itTracksAppHang() throws {
--- a/Datadog/IntegrationUnitTests/RUM/AppHangsMonitoringTests.swift
+++ b/Datadog/IntegrationUnitTests/RUM/AppHangsMonitoringTests.swift
@@ -1,0 +1,52 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import XCTest
+import DatadogInternal
+@testable import DatadogRUM
+
+/// Test case covering scenarios of App Hangs monitoring in RUM.
+class AppHangsMonitoringTests: XCTestCase {
+    private var core: DatadogCoreProxy! // swiftlint:disable:this implicitly_unwrapped_optional
+    private var rumConfig = RUM.Configuration(applicationID: .mockAny())
+
+    override func setUp() {
+        super.setUp()
+        core = DatadogCoreProxy()
+    }
+
+    override func tearDown() {
+        core.flushAndTearDown()
+        core = nil
+        super.tearDown()
+    }
+
+    func testWhenMainThreadHangsAboveThreshold_itTracksAppHang() throws {
+        let mainQueue = DispatchQueue(label: "main-queue", qos: .userInteractive)
+        rumConfig.mainQueue = mainQueue
+
+        // Given
+        RUM.enable(with: rumConfig, in: core)
+
+        // When
+        let beforeHang = Date()
+        mainQueue.sync {
+            Thread.sleep(forTimeInterval: self.rumConfig.defaultAppHangThreshold * 1.5)
+        }
+
+        // Then
+        Thread.sleep(forTimeInterval: 0.5) // wait to make sure watchdog thread completes hang tracking
+        RUMMonitor.shared(in: core).dd.flush() // flush RUM monitor to await hang processing
+
+        let errors = core.waitAndReturnEvents(ofFeature: RUMFeature.name, ofType: RUMErrorEvent.self)
+        let appHangError = try XCTUnwrap(errors.first)
+
+        XCTAssertEqual(appHangError.error.message, "App Hang")
+        XCTAssertEqual(appHangError.error.type, "AppHang")
+        XCTAssertEqual(appHangError.error.source, .source)
+        XCTAssertGreaterThanOrEqual(appHangError.date, beforeHang.timeIntervalSince1970.toInt64Milliseconds)
+    }
+}
--- a/DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsObserver.swift
+++ b/DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsObserver.swift
@@ -1,0 +1,59 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+import DatadogInternal
+
+internal class AppHangsObserver: RUMCommandPublisher {
+    /// Watchdog thread that monitors the main queue for App Hangs.
+    private let watchdogThread: AppHangsWatchdogThread
+    /// Weak reference to RUM monitor for sending App Hang events.
+    private(set) weak var subscriber: RUMCommandSubscriber?
+
+    init(
+        appHangThreshold: TimeInterval,
+        observedQueue: DispatchQueue,
+        dateProvider: DateProvider,
+        telemetry: Telemetry
+    ) {
+        watchdogThread = AppHangsWatchdogThread(
+            appHangThreshold: appHangThreshold,
+            queue: observedQueue,
+            dateProvider: dateProvider,
+            telemetry: telemetry
+        )
+        watchdogThread.onHangEnded = { [weak self] appHang in
+            // called on watchdog thread
+            self?.report(appHang: appHang)
+        }
+    }
+
+    func start() {
+        watchdogThread.start()
+    }
+
+    func stop() {
+        watchdogThread.cancel()
+    }
+
+    func publish(to subscriber: RUMCommandSubscriber) {
+        self.subscriber = subscriber
+    }
+
+    private func report(appHang: AppHang) {
+        let addHangCommand = RUMAddCurrentViewErrorCommand(
+            time: appHang.date,
+            message: "App Hang",
+            type: "AppHang",
+            stack: nil, // TODO: RUM-2925 Add hang stack trace
+            source: .source,
+            attributes: [
+                "hang_duration": appHang.duration
+            ]
+        )
+        subscriber?.process(command: addHangCommand)
+    }
+}
--- a/DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsWatchdogThread.swift
+++ b/DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsWatchdogThread.swift
@@ -1,0 +1,131 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+import DatadogInternal
+
+internal struct AppHang {
+    /// The date of hang end.
+    let date: Date
+    /// The duration of the hang.
+    let duration: TimeInterval
+    // TODO: RUM-2925 Add hang stack trace
+}
+
+internal final class AppHangsWatchdogThread: Thread {
+    struct Constants {
+        /// The "idle" interval for sleeping the watchdog thread before scheduling the next task on the main queue, represented as a percentage of the `appHangThreshold`.
+        ///
+        /// Introducing even a small sleep significantly reduces CPU consumption by the watchdog thread (nearly 0%). No sleep would result in close to 100% CPU usage.
+        ///
+        /// `tolerance` defines the margin of error in hang monitoring. The "measured hang duration" may differ from `0` to `tolerance` % when compared to the "actual hang duration".
+        /// This means some actual hangs very close to `appHangThreshold` might be considered false negatives and not reported.
+        ///
+        /// **Example:**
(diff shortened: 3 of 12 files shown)


Datadog/IntegrationUnitTests/RUM/AppHangsMonitoringTests.swift
ncreated, line 1: Integration-unit test.

DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsWatchdogThread.swift
```diff
@@ -0,0 +1,131 @@
...
+            } // Otherwise, `result == .timeOut`, meaning that the main thread task is delayed to more than hang threshold.
+
+            // We expect to be here roughtly `~appHangThreshold` after `wait()`. If this code is executed well after (50% margin),
+            // assume that the thread was suspended and app just woke up. In such case, ignore the hang as likely false-positive.
+            if interval(from: waitStart, to: .now()) > appHangThreshold * 1.5 {
+                DD.logger.debug("Ignoring likely false-positive App Hang")
+                continue // ignore likely false-positive
+            }
+
+            // TODO: RUM-2925 Capture stack trace of the pending App Hang
```
ncreated, line 105: It will be done in next PRs, but the actual stack trace generation will start from this line.

DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsWatchdogThread.swift
maciejburda, line 1: Super clean code and nice documentation 👏

Easy to read this crucial piece!

DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsWatchdogThread.swift
```diff
@@ -0,0 +1,131 @@
...
+        self.idleInterval = appHangThreshold * Constants.tolerance
+        self.mainQueue = queue
+        self.dateProvider = dateProvider
+        self.telemetry = telemetry
+        super.init()
+        self.name = "com.datadoghq.app-hang-watchdog"
+    }
+
+    override func main() {
+        let mainThreadTask = self.mainThreadTask
```
maciejburda, line 75: what's the benefit of doing that here?

maciejburda (approved):
Great stuff! Reads really well, and it was quite interesting to go through tests. 💪

Left some minor comments 🤓

DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsWatchdogThread.swift
ncreated: I tried to avoid capturing `[weak self]` as part of the frequently scheduled task on `mainQueue`. There's no gain, just practical choice (semaphore is a class, so this is correct).

DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsWatchdogThread.swift
```diff
@@ -0,0 +1,131 @@
...
+
+            // TODO: RUM-2925 Capture stack trace of the pending App Hang
+
+            // Previous wait timed out, so wait again for the task completion, this time infinitely until the hang ends.
+            mainThreadTask.wait()
+
+            // The hang has finished.
+            let hangDuration = interval(from: waitStart, to: .now())
+
+            if hangDuration > 30 { // sanity-check
```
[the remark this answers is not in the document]
ncreated: Because it's "roughly" 10s we may cause false-negatives and actually lost some real hangs. I really want to see some false-positive examples in telemetry, so we can adjust it better, based on real data.

DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsWatchdogThread.swift
```diff
@@ -0,0 +1,131 @@
...
+                // If the hang duration is irrealistically long (way more than assumed iOS watchdog termination limit: ~10s), send telemetry.
+                // This could be another false-positive caused by thread suspension between the two `wait()` calls.
+                telemetry.debug("Detected an App Hang with an unusually long duration", attributes: ["hang_duration": hangDuration])
+                return
```
ncreated, line 117: blocking/ This should be `continue`, not `return`. Now it tears down the entire thread after first false-positive 👎.

DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsObserver.swift
```diff
@@ -0,0 +1,59 @@
...
+        dateProvider: DateProvider,
+        telemetry: Telemetry
+    ) {
+        watchdogThread = AppHangsWatchdogThread(
+            appHangThreshold: appHangThreshold,
+            queue: observedQueue,
+            dateProvider: dateProvider,
+            telemetry: telemetry
+        )
+        watchdogThread.onHangEnded = { [weak self] appHang in
```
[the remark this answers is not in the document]
ncreated: We access `self` (`AppHangsObserver`) in this block and it doesn't yet exist until `AppHangsWatchdogThread`'s `.init()` returns. There are ways to workaround it, but all come with trade-offs.

DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsWatchdogThread.swift
```diff
@@ -0,0 +1,131 @@
...
+                date: dateProvider.now,
+                duration: hangDuration
+            )
+            onHangEnded?(appHang)
+        }
+    }
+
+    private func interval(from t1: DispatchTime, to t2: DispatchTime) -> TimeInterval {
+        TimeInterval(t2.uptimeNanoseconds - t1.uptimeNanoseconds) / 1_000_000_000
+    }
```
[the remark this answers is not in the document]
ncreated: I prefer adding convenience APIs after the time - if we see this pattern is repeated over and over, then it could be a moment for refactor and adding convenience.

Datadog/IntegrationUnitTests/RUM/AppHangsMonitoringTests.swift
```diff
@@ -0,0 +1,52 @@
...
+import DatadogInternal
+@testable import DatadogRUM
+
+/// Test case covering scenarios of App Hangs monitoring in RUM.
+class AppHangsMonitoringTests: XCTestCase {
+    private var core: DatadogCoreProxy! // swiftlint:disable:this implicitly_unwrapped_optional
+    private var rumConfig = RUM.Configuration(applicationID: .mockAny())
+
+    override func setUp() {
+        super.setUp()
```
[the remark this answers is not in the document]
ncreated: Oh, thanks a lot for bringing this up 🏅. I always wondered if this is needed or not and didn't know that Apple finally updated their docs to make it clear from their examples. Fixed ✅

DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsWatchdogThread.swift
```diff
@@ -0,0 +1,131 @@
...
+internal struct AppHang {
+    /// The date of hang end.
+    let date: Date
+    /// The duration of the hang.
+    let duration: TimeInterval
+    // TODO: RUM-2925 Add hang stack trace
+}
+
+internal final class AppHangsWatchdogThread: Thread {
+    struct Constants {
```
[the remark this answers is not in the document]
ncreated: Enum makes sense, fixed ✅. This constant is referenced from unit tests, keeping `internal`.

DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsWatchdogThread.swift
ncreated: Fixed ✅

DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsWatchdogThread.swift
```diff
@@ -0,0 +1,131 @@
...
+    override func main() {
+        let mainThreadTask = self.mainThreadTask
+
+        while !isCancelled {
+            defer {
+                // Before continuing, sleep for a while, to reduce the CPU consumption by watchdog thread.
+                Thread.sleep(forTimeInterval: idleInterval)
+            }
+
+            let waitStart: DispatchTime = .now()
```
ganeshnj, line 83: perhaps
```suggestion
            let waitStart: DispatchTime = self.dateProvider.now()
```

DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsWatchdogThread.swift
```diff
@@ -0,0 +1,131 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+import DatadogInternal
+
+internal struct AppHang {
+    /// The date of hang end.
+    let date: Date
```
ganeshnj, line 12:
```suggestion
    let endTime: Date
```

I kind of prefer having startTime and duration in general, endTime and duration seems odd from API point of view.

DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsWatchdogThread.swift
```diff
@@ -0,0 +1,131 @@
...
+    /// The "idle" interval for watchdog thread operations. Reduces the pressure on CPU.
+    private let idleInterval: TimeInterval
+    /// Semaphore used to block this thread until main thread responds.
+    private let mainThreadTask = DispatchSemaphore(value: 0)
+    /// The queue to observe by this thread (main queue).
+    private let mainQueue: DispatchQueue
+    /// SDK date provider.
+    private let dateProvider: DateProvider
+    /// Telemetry interface.
+    private let telemetry: Telemetry
```
ganeshnj, line 45: curious, having strong reference to Telemetry, can it cause reference cycle and in the end memory leak here?

maxep (approved):
Looks great! very nice work with the custom `Thread`!!

DatadogRUM/Sources/Instrumentation/RUMInstrumentation.swift
```diff
@@ -80,23 +86,32 @@ internal final class RUMInstrumentation: RUMCommandPublisher {
...
+            observedQueue: mainQueue,
+            dateProvider: dateProvider,
+            telemetry: telemetry
+        )
 
         // Enable configured instrumentations:
         self.viewControllerSwizzler?.swizzle()
         self.uiApplicationSwizzler?.swizzle()
         self.longTasks?.start()
+        self.appHangs?.start()
```
ganeshnj, line 100: enabled by default?

ganeshnj (commented):
A few minor comments, overall looks great.

DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsWatchdogThread.swift
ncreated: It is intentional. The `dateProvider.now()` returns `Date`, but we need `DispatchTime` to later await on the semaphore. We don't need true `Date` to measure the hang, CPU time is enough.

DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsWatchdogThread.swift
ncreated: This is implemented by [`CoreTelemetry`](https://github.com/DataDog/dd-sdk-ios/blob/ab930187170a1170d5131e0992f81f47ad5fd280/DatadogInternal/Sources/Telemetry/Telemetry.swift#L302-L304) which weakifies the underlying `core` reference.

DatadogRUM/Sources/Instrumentation/RUMInstrumentation.swift
ncreated: This is TBD with the product, see [this ADR: Decision #2](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3473147398/ADR+-+App+Hangs+detection+in+iOS+SDK) (internal). This PR is against a feature branch, the final API will be added before it lands on `develop`.

DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsWatchdogThread.swift
ncreated: If not blocking, I'd keep it as it is. Hangs are reported as RUM Errors, so they have no duration on view's timeline and we want to have them reported at the time of hang end.

Hang error will follow the Long Task that visualizes the duration:
Screenshot 2024-02-23 at 12 33 59

PR #1687, opened 2024-02-19
Ncreated/rum 2925/record the stack trace of app hang

ncreated:
### What and why?

// TBD

### How?

// TBD

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


DatadogCrashReporting/Sources/CrashReporting.swift
```diff
@@ -38,6 +38,10 @@ public final class CrashReporting {
 
             try core.register(feature: reporter)
 
+            if let plcr = PLCrashReporterPlugin.thirdPartyCrashReporter {
+                try core.register(backtraceReporter: BacktraceReporter(reporter: plcr))
+            }
```
ncreated, line 43: This might be more elegant, but would require revamping the startup sequence for `DatadogCrashReporting`. It still depends on [`CrashReportingPlugin`](https://github.com/DataDog/dd-sdk-ios/blob/6eab2d82441c4997ce05f9ecee7d64cc38517fc1/DatadogCrashReporting/Sources/CrashReportingPlugin.swift#L176), which was V1 dependency-inverter, no longer required in V2. This refactor is medium-size, hence I didn't want to do it for this PR.

DatadogInternal/Sources/FeatureAPIs/CrashReporting/DDCrashReport.swift
```diff
@@ -0,0 +1,102 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+/// Crash Report format supported by Datadog SDK.
+@objc
```
ganeshnj, line 10: why do we need this to be inter-operable with objc?

ganeshnj (approved):
LGTM, just a small question 👌

DatadogInternal/Sources/FeatureAPIs/CrashReporting/BacktraceReport.swift
maxep, line 1: /suggestion Can we place these under `/DatadogInternal/Sources/Models/` instead? so we make it clear that we only place models in the internal module.

maxep (approved):
Looks great!! I really appreciate the extension pattern applied to the core protocol 👌

DatadogInternal/Sources/FeatureAPIs/CrashReporting/DDCrashReport.swift
ncreated: Fair call, I traced this back to this commit https://github.com/DataDog/dd-sdk-ios/commit/2f892c267a7a4ab4892b4ced2a31aae3b3074dcc where we were adding Obj-c compatibility for Crash Reporting. This was back in V1. With V2 aligning module dependencies much better, this can be now internal to Swift domain, so it is not required. Changed here and consequently in few other places. ✅

DatadogInternal/Sources/FeatureAPIs/CrashReporting/BacktraceReport.swift
ncreated: Sure thing 👍, changed ✅

Datadog/IntegrationUnitTests/CrashReporting/GeneratingBacktraceTests.swift
```diff
@@ -0,0 +1,63 @@
...
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import XCTest
+import DatadogCrashReporting
+@testable import DatadogInternal
+
+/// Tests integration of `DatadogCore` and `DatadogCrashReporting` for backtrace generation.
+class GeneratingBacktraceTests: XCTestCase {
+    private var core: DatadogCoreProxy! // swiftlint:disable:this implicitly_unwrapped_optional
```
maciejburda, line 13: probably worth having this lint rule global to all the tests

DatadogCrashReporting/Sources/PLCrashReporterIntegration/PLCrashReporterIntegration.swift
```diff
@@ -57,4 +57,21 @@ internal final class PLCrashReporterIntegration: ThirdPartyCrashReporter {
...
     }
+
+    func generateBacktrace() throws -> BacktraceReport {
+        let liveReportData = try crashReporter.generateLiveReportAndReturnError()
+        let liveReport = try PLCrashReport(data: liveReportData)
+
+        // This is quite opportunistic - we map PLCR's live report through existing `DDCrashReport` builder to
+        // then extract essential elements for assembling `BacktraceReport`. It works for now, but be careful
+        // with how this evolves. We may need a dedicated `BacktraceReport` builder that only shares some code
+        // with `DDCrashReport` builder.
```
maciejburda, line 68: I guess it's mostly affected by the PLCR update?

Do we expect to have failed unit tests when something like this happen, or we should have some extra measures?

maciejburda (approved):
LGTM

DatadogCrashReporting/Sources/PLCrashReporterIntegration/PLCrashReporterIntegration.swift
ncreated: This is PLCR-safe 👍. Basically, both for Crash Reports and Backtrace Reports, the value returned by PLCR is `PLCrashReport`. We then map it through our `DDCrashReportBuilder`, whereas implementing distinct `BacktraceReportBuilder` would be more correct and scalable. We don't need it today, hence I'm leaving this decision context in comment.

PR #1691, opened 2024-02-20
RUM-2816 Send raw webview event on the message bus

maxep:
### What and why?

Transmit raw web-view event on the message-bus and avoid going through the `.baggage` encoding/decoding.

### How?

The `FeatureMessage` now support `.webview` case to transmit messages from the web-view integration.
The `DatadogWebViewTracking` will now send a shared `WebViewMessage` over the bus.

### Review checklist

- [x] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


DatadogInternal/Sources/MessageBus/FeatureMessage.swift
```diff
@@ -8,13 +8,24 @@ import Foundation
...
     case baggage(
         key: String,
         baggage: FeatureBaggage
     )
 
+    /// A message value.
+    ///
+    /// Use this message case when the value type is known end-to-end.
+    /// Such type can be defined as part of the `DatadogInternal` module.
+    case value(Any)
```
maxep, line 27: We could also use a case per type, such as `case webview(WebViewMessage)`. But it will be hardly scalable as we increase the number of types to transmit.

DatadogLogs/Sources/Feature/MessageReceivers.swift
```diff
@@ -277,56 +274,45 @@ internal struct WebViewLogReceiver: FeatureMessageReceiver {
...
+            } else {
+                event[tagsKey] = ddTags
+            }
 
-                if let timestampInMs = event[dateKey] as? Int64 {
-                    let serverTimeOffsetInMs = context.serverTimeOffset.toInt64Milliseconds
-                    let correctedTimestamp = Int64(timestampInMs) + serverTimeOffsetInMs
-                    event[dateKey] = correctedTimestamp
-                }
+            if let timestampInMs = event[dateKey] as? Int {
```
maxep, line 295: The `date` type is now `Int` as it's not going through the encoding/decoding of a baggage, the process resulted in `Int64`.

DatadogRUM/Sources/Integrations/WebViewEventReceiver.swift
```diff
@@ -80,7 +63,7 @@ internal final class WebViewEventReceiver: FeatureMessageReceiver {
                 let rum: RUMCoreContext = try rumBaggage.decode()
                 var event = event
 
-                if let date = event["date"] as? Int64 {
+                if let date = event["date"] as? Int {
```
maxep, line 66: Same here

> The date type is now Int as it's not going through the encoding/decoding of a baggage, the process resulted in Int64.

DatadogRUM/Tests/Integrations/WebViewEventReceiverTests.swift
```diff
@@ -44,26 +142,6 @@ class WebViewEventReceiverTests: XCTestCase {
         XCTAssertEqual(command.time, .mockDecember15th2019At10AMUTC())
     }
 
-    func testWhenReceivingWebViewTrackingMessageWithInvalidEvent_itRejectsTheMessageAndSendsErrorTelemetry() throws {
```
maxep, line 47: Replace by `WebViewMessageTests`

DatadogInternal/Sources/MessageBus/FeatureMessage.swift
maciejburda: It would show a clean list of supported types in the contract though. How many different models do we anticipate at this point?

maciejburda (approved):
LGTM

DatadogInternal/Sources/MessageBus/FeatureMessage.swift
ncreated: I'd be pretty much in favour of:
```swift
case webview(WebViewMessage)
```

The main benefit is being explicit and type-safe on both ends. An extra thing is having a clear definition point for documenting all messages that features can exchange.

> But it will be hardly scalable as we increase the number of types to transmit.

Could you elaborate? IMO with the growing number of types that message bus transmits, having clarity on what exact type both sides expect sounds critical. I don't see this against scalability.

I'd chose the simplest, typed solution to start with for maximum safety. We could see how it grows and adjust.

ncreated (commented):
I left my comment on using `Any` in `FeatureMessage`, I'd like to discuss it before we continue. The rest of PR looks fine 👍.

DatadogInternal/Sources/MessageBus/FeatureMessage.swift
maxep: Indeed, I'm probably anticipating too much here. We should go with an explicit set of messages, I will change that 👍 

Here a list of message that would need type sharing:
- error message from log
- rum view
- rum session state
- crash report
- web-view

ncreated: And that would be perfectly fine to have them all listed 👍. Long-term we can see what would be the cost of maintaining this in implementation **and tests**. Especially in tests, I can foresee the [FeatureMessageMocks.swift](https://github.com/DataDog/dd-sdk-ios/blob/2.7.1/TestUtilities/Mocks/FeatureMessageMocks.swift) constantly growing with new `first*()` and `as*` helpers, which is what concerns me.

maxep: Done, I have replaced the `.value` with `. webview` 👍

DatadogInternal/Sources/Models/WebViewMessage.swift
ncreated, line 1: suggestion/ Could we move it to `Models/WebViewTracking/WebViewMessage.swif` so we apply the same convention as earlier started in https://github.com/DataDog/dd-sdk-ios/pull/1687/ for CrashReporting models?
```
DatadogInternal/Sources/Models/CrashReporting/
DatadogInternal/Sources/Models/CrashReporting/DDCrashReport.swift
DatadogInternal/Sources/Models/CrashReporting/BacktraceReport.swift
DatadogInternal/Sources/Models/CrashReporting/BinaryImage.swift
DatadogInternal/Sources/Models/CrashReporting/DDThread.swift
```

ncreated (dismissed):
Looks good 👌

DatadogInternal/Sources/Models/WebViewMessage.swift
maxep: Sure thing 👍  it needs re-approval tho

ncreated (approved):

maciejburda (approved):

PR #1694, opened 2024-02-20
feat: Add ability to override `source_type` for crashes

fuzzybinary:
### What and why?

Add the ability to pass the attribute `_dd.native_source_type` to override the source_type for native crashes. This is needed for platforms that can supply additional information from a native crash (specifically Unity with IL2CPP).

refs: RUM-2691

### How?

This added an extra item to `DatadogContext`: `nativeSourceOverride` that is set from the cross platform additional attribute `_dd.native_source_type`. This is used to add the same value to the CrashContext when the DatadogCrashReporting feature sends a message over the bus. That value is then read and serialized by the Logs and RUM feature to provide the correct value.

### Review checklist

- [x] Run integration tests
- [x] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogCrashReporting/Sources/CrashContext/CrashContext.swift
+++ b/DatadogCrashReporting/Sources/CrashContext/CrashContext.swift
@@ -69,6 +69,9 @@ internal struct CrashContext: Codable, Equatable {
     /// The last _"Is app in foreground?"_ information from crashed app process.
     let lastIsAppInForeground: Bool
 
+    /// The override for `source_type` for crashes.
+    let nativeSourceTypeOverride: String?
+
     // MARK: - Initialization
 
     init(
@@ -86,7 +89,8 @@ internal struct CrashContext: Codable, Equatable {
         carrierInfo: CarrierInfo?,
         lastRUMViewEvent: AnyCodable?,
         lastRUMSessionState: AnyCodable?,
-        lastIsAppInForeground: Bool
+        lastIsAppInForeground: Bool,
+        nativeSourceTypeOverride: String?
     ) {
         self.serverTimeOffset = serverTimeOffset
         self.service = service
@@ -103,6 +107,7 @@ internal struct CrashContext: Codable, Equatable {
         self.lastRUMViewEvent = lastRUMViewEvent
         self.lastRUMSessionState = lastRUMSessionState
         self.lastIsAppInForeground = lastIsAppInForeground
+        self.nativeSourceTypeOverride = nativeSourceTypeOverride
     }
 
     init(
@@ -123,6 +128,7 @@ internal struct CrashContext: Codable, Equatable {
         self.networkConnectionInfo = context.networkConnectionInfo
         self.carrierInfo = context.carrierInfo
         self.lastIsAppInForeground = context.applicationStateHistory.currentSnapshot.state.isRunningInForeground
+        self.nativeSourceTypeOverride = context.nativeSourceOverride
 
         self.lastRUMViewEvent = lastRUMViewEvent
         self.lastRUMSessionState = lastRUMSessionState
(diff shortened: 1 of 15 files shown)


maciejburda (dismissed):
LGTM

DatadogInternal/Sources/Attributes/Attributes.swift
```diff
@@ -149,6 +149,10 @@ public struct CrossPlatformAttributes {
     /// It sets the GraphQL varibles as a JSON string if they were defined by the developer.
     /// Expects `String` value.
     public static let graphqlVariables = "_dd.graphql.variables"
+
+    /// Override the `source_type` of errors reported by the native crash handler. This is used on
+    /// platforms that can supply extra steps or information on a native crash (such as Unity's IL2CPP
```
ncreated, line 154: 🎯 
```suggestion
    /// platforms that can supply extra steps or information on a native crash (such as Unity's IL2CPP)
```

DatadogInternal/Sources/Context/DatadogContext.swift
```diff
@@ -39,6 +39,9 @@ public struct DatadogContext {
     ///  - See: Datadog [Reserved Attributes](https://docs.datadoghq.com/logs/log_configuration/attributes_naming_convention/#reserved-attributes).
     public let source: String
 
+    /// Denotes the source type for  crashes.. This is used for platforms that provide additional symbolication steps for native crashes.
+    public let nativeSourceOverride: String?
```
ncreated, line 43:
```suggestion
    /// Denotes the source type for  crashes. This is used for platforms that provide additional symbolication steps for native crashes.
    public let nativeSourceOverride: String?
```

DatadogCrashReporting/Sources/CrashContext/CrashContext.swift
```diff
@@ -69,6 +69,9 @@ internal struct CrashContext: Codable, Equatable {
     /// The last _"Is app in foreground?"_ information from crashed app process.
     let lastIsAppInForeground: Bool
 
+    /// The override for `source_type` for crashes.
+    let nativeSourceTypeOverride: String?
```
ncreated, line 73: I don't think we need to encode this value through `CrashContext` to have it available upon app restart (when we actually send the crash). With setting `nativeSourceOverride` in `DatadogContext` on app init, it will be set to the same value in next app session, right? That assumed, we can read it from `eventWriteContext` API before writing the actual Log event or RUM events:

- for RUM:
https://github.com/DataDog/dd-sdk-ios/blob/bdab1a1aab4b0f3e4c08d10883c508a44e435520/DatadogRUM/Sources/Integrations/CrashReportReceiver.swift#L320-L323

- for Logs:
https://github.com/DataDog/dd-sdk-ios/blob/bdab1a1aab4b0f3e4c08d10883c508a44e435520/DatadogLogs/Sources/Feature/MessageReceivers.swift#L264-L266

In both cases, we get `DatadogContext` as the first argument of the lambda. This is how we can make it simpler, given RUM example:
```swift
core.scope(for: RUMFeature.name)?.eventWriteContext(bypassConsent: true) { context, writer in
    let rumError = createRUMError(from: crashReport, and: updatedRUMView, crashDate: realCrashDate, sourceType: context.nativeSourceOverride)
    writer.write(value: rumError)
    writer.write(value: updatedRUMView)
}
```
I wouldn't mind creating error events on the context thread that this change would imply. This is negligible impact, but we can save a lot of hassle by not coding this simple value through `CrashContext`.

LTMWDYT @fuzzybinary

ncreated (commented):
I think we can do it much simpler. See the comment in `CrashContext.swift`.

DatadogCrashReporting/Sources/CrashContext/CrashContext.swift
fuzzybinary: Yeah, this was the question I think I had initially. I wasn't sure if the CrashReporting feature should send it over the message bus or if it should be pulled from context during the write.  If we're good with pulling it during the write that definitely simplifies it.

I'll adjust the PR.

ncreated: Let's do 👍

fuzzybinary:
@ncreated Fixed with the simpler approach. Let me know what you think.

ncreated (approved):
Looks great @fuzzybinary 👍, thanks for rising the concern. It is much simpler now 🎯💯

PR #1695, opened 2024-02-21
RUM-2816 [SR] Reduce segment metadata during upload only

maxep:
### What and why?

Build replay segments at the upload phase **only**.

Previously, the `start`, `end`, `records_count`, and `records_count` fields of segments metadata were computed when writing snapshots then re-computed when merging segments together at the upload phase.

With the addition of the web-view record events to the pipeline, computing the segment metadata when writing a record won't be performant - as will require parsing the event - nor meaningful because web-view will sent records one by one.

### How?

The `EnrichedRecord` now only include the RUM context and the list of records to be written to disk. The `SegmentJSON` is build at the upload phase and will compute the `start`, `end`, `records_count`, and `records_count` fields from the list of records. Segments with the same `view.id` are then merged together.

**Note:** I'm considering removing the `EnrichedRecord` type and rely on batch event metadata instead.

### Review checklist

- [x] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


maciejburda (approved):
LGTM

PR #1696, opened 2024-02-22
fix: pass through data when network request completes

### LINKED ISSUE #1680: `RUM.ResourceAttributesProvider` not providing `data` information in URLSession tracking
### Describe the bug

ResourceAttributesProvider in the urlSessionTracking initializer for RUM is not providing data information for network requests made with data(for:) methods. It provides URLRequest and URLResponse information, but data always comes back as nil and does not populate for resources. This seems to be limiting the automatic tracking experience. 

### Reproduction
Configure Datadog RUM and enable urlSessionTracking with the resourceAttributesProvider. Create a network request using async await and data(for:). Call the request and observe that neither the closure for the attributes, nor the resource in the RUM dashboard will show data. 
```swift
RUM.enable(
    with: RUM.Configuration(
        applicationID: appID,
        urlSessionTracking: RUM.Configuration.URLSessionTracking(resourceAttributesProvider: { req, resp, data, err in
            return ["dataInfo": data]
        })
    )
)
```
---

#### Datadog SDK version:
SDK V 2.7.0

#### Dependency Manager:

_Which dependency manager do you use? e.g. Cocoapods / Carthage / SPM / ..._
SPM

#### Deployment Target:

_What is the Deployment Target of your app? e.g. `iOS 12`, `iPhone` + `iPad`_
ios 17

ganeshnj:
### What and why?

Fixes #1680

### How?

This PR mimics the old behavior of the instrumentation.

When completion handler is provided in `URLSession` APIs, `urlSession(_:dataTask:didReceive:)` is not called, which means we can't buffer the data and in the called the Resource Attribute Provider.

This is not only true for the completion handler based APIs but also true for async-await APIs.

Meanwhile we use 	`urlSession(_:task:didFinishCollecting:)` to mark the completion of the network operation which is the only reliable API at this moment. We use this to mark completion for async-await APIs.

#### How did the regression happen?

For some APIs, both `urlSession(_:task:didFinishCollecting:)` and `urlSession(_:task:didCompleteWithError:)` are called. But there is not specific order guaranteed which one comes first, ie the one called first, marks the end of the network operation. In the breaking change, `urlSession(_:task:didFinishCollecting:)` and given `urlSession(_:dataTask:didReceive:)` was not called either, Resource Attribute Provider executes with nil data.

#### How does this PR fix it?

With this PR we make sure, if the API has a completion handler given, we are not going to use `urlSession(_:task:didFinishCollecting:)` as end of operation because we are sure  `urlSession(_:task:didCompleteWithError:)` will be called.

Along side, when the operation finishes, we also call handler for registering the data when the operation completes. The APIs guarantees from behavior that either `urlSession(_:dataTask:didReceive:)` is called or the compeltion handler, hence we don't register data two times.

#### Do we support async-await?

No, because they don't have completion handler, so experience remain broken with them.

#### Tested locally on

- [x] iOS 12.4
- [x] iOS 13.0
- [x] iOS 14.5
- [x] iOS 15.5

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/Datadog/IntegrationUnitTests/Public/NetworkInstrumentationIntegrationTests.swift
+++ b/Datadog/IntegrationUnitTests/Public/NetworkInstrumentationIntegrationTests.swift
@@ -8,6 +8,7 @@ import XCTest
 import TestUtilities
 import DatadogInternal
 
+@testable import DatadogRUM
 @testable import DatadogTrace
 @testable import DatadogCore
 
@@ -50,7 +51,7 @@ class NetworkInstrumentationIntegrationTests: XCTestCase {
         core.flushAndTearDown()
         core = nil
     }
-    
+
     func testParentSpanPropagation() throws {
         let expectation = expectation(description: "request completes")
         // Given
@@ -87,5 +88,118 @@ class NetworkInstrumentationIntegrationTests: XCTestCase {
     }
 
     class MockDelegate: NSObject, URLSessionDataDelegate {
+    }
+
+    func testResourceAttributesProvider_givenURLSessionDataTaskRequest() {
+        core = DatadogCoreProxy(
+            context: .mockWith(
+                env: "test",
+                version: "1.1.1",
+                serverTimeOffset: 123
+            )
+        )
+
+        let providerExpectation = expectation(description: "provider called")
+        var providerDataCount = 0
+        RUM.enable(
+            with: .init(
+                applicationID: .mockAny(),
+                urlSessionTracking: .init(
+                    resourceAttributesProvider: { req, resp, data, err in
+                        XCTAssertNotNil(data)
+                        XCTAssertTrue(data!.count > 0)
+                        providerDataCount = data!.count
+                        providerExpectation.fulfill()
+                        return [:]
+                })
+            ),
+            in: core
+        )
+
+        URLSessionInstrumentation.enable(
+            with: .init(
+                delegateClass: InstrumentedSessionDelegate.self
+            ),
+            in: core
+        )
+
+        let session = URLSession(
+            configuration: .ephemeral,
+            delegate: InstrumentedSessionDelegate(),
+            delegateQueue: nil
+        )
+        var request = URLRequest(url: URL(string: "https://www.datadoghq.com/")!)
+        request.httpMethod = "GET"
+
+        let task = session.dataTask(with: request)
+        task.resume()
+
+        wait(for: [providerExpectation], timeout: 10)
+        XCTAssertTrue(providerDataCount > 0)
+    }
+
+    func testResourceAttributesProvider_givenURLSessionDataTaskRequestWithCompletionHandler() {
+        core = DatadogCoreProxy(
+            context: .mockWith(
+                env: "test",
+                version: "1.1.1",
+                serverTimeOffset: 123
+            )
+        )
+
+        let providerExpectation = expectation(description: "provider called")
+        var providerDataCount = 0
+        var providerData: Data?
+        RUM.enable(
+            with: .init(
+                applicationID: .mockAny(),
+                urlSessionTracking: .init(
+                    resourceAttributesProvider: { req, resp, data, err in
+                        XCTAssertNotNil(data)
+                        XCTAssertTrue(data!.count > 0)
+                        providerDataCount = data!.count
+                        data.map { providerData = $0 }
+                        providerExpectation.fulfill()
+                        return [:]
+                })
+            ),
+            in: core
+        )
+
+        URLSessionInstrumentation.enable(
+            with: .init(
+                delegateClass: InstrumentedSessionDelegate.self
+            ),
+            in: core
+        )
+
+        let session = URLSession(
+            configuration: .ephemeral,
+            delegate: InstrumentedSessionDelegate(),
+            delegateQueue: nil
+        )
+        let request = URLRequest(url: URL(string: "https://www.datadoghq.com/")!)
+
+        let taskExpectation = self.expectation(description: "task completed")
+        var taskDataCount = 0
+        var taskData: Data?
+        let task = session.dataTask(with: request) { data, _, _ in
+            XCTAssertNotNil(data)
+            XCTAssertTrue(data!.count > 0)
+            taskDataCount = data!.count
+            data.map { taskData = $0 }
+            taskExpectation.fulfill()
+        }
+        task.resume()
+
+        wait(for: [providerExpectation, taskExpectation], timeout: 10)
+        XCTAssertEqual(providerDataCount, taskDataCount)
+        XCTAssertEqual(providerData, taskData)
+    }
+
+    class InstrumentedSessionDelegate: NSObject, URLSessionDataDelegate {
+        func urlSession(_ session: URLSession, dataTask: URLSessionDataTask, didReceive data: Data) {
+            print(data)
+        }
     }
 }
--- a/Datadog/IntegrationUnitTests/Public/NetworkInstrumentationIntegrationTests.swift
+++ b/Datadog/IntegrationUnitTests/Public/NetworkInstrumentationIntegrationTests.swift
@@ -199,7 +199,6 @@ class NetworkInstrumentationIntegrationTests: XCTestCase {
 
     class InstrumentedSessionDelegate: NSObject, URLSessionDataDelegate {
         func urlSession(_ session: URLSession, dataTask: URLSessionDataTask, didReceive data: Data) {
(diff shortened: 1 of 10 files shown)


Datadog/IntegrationUnitTests/Public/NetworkInstrumentationIntegrationTests.swift
```diff
@@ -88,4 +89,117 @@ class NetworkInstrumentationIntegrationTests: XCTestCase {
...
+        task.resume()
+
+        wait(for: [providerExpectation, taskExpectation], timeout: 10)
+        XCTAssertEqual(providerDataCount, taskDataCount)
+        XCTAssertEqual(providerData, taskData)
+    }
+
+    class InstrumentedSessionDelegate: NSObject, URLSessionDataDelegate {
+        func urlSession(_ session: URLSession, dataTask: URLSessionDataTask, didReceive data: Data) {
+            print(data)
```
ncreated, line 202: Doesn't seem useful except when debugging the new test.
```suggestion
```

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
```diff
@@ -92,7 +92,7 @@ internal final class NetworkInstrumentationFeature: DatadogFeature {
             interceptDidFinishCollecting: { [weak self] session, task, metrics in
                 self?.task(task, didFinishCollecting: metrics)
 
-                if #available(iOS 15, tvOS 15, *) {
+                if #available(iOS 15, tvOS 15, *), !task.dd.hasCompletion {
                     // iOS 15 and above, didCompleteWithError is not called hence we use task state to detect task completion
                     // while prior to iOS 15, task state doesn't change to completed hence we use didCompleteWithError to detect task completion
                     self?.task(task, didCompleteWithError: task.error)
```
ncreated, line 98: suggestion/ Don't we leak the abstraction here? It seems that `swizzler` implementation could perform this whole check. This way `NetworkInstrumentationFeature` won't be infected with OS-specific checks as it will rely on the `swizzler` to call `interceptDidCompleteWithError` uniformly, no matter of `task` state or OS specifics.

This complexity is much easier to unit test in downstream components. A broken test will be much shorter to troubleshoot on this or that OS version and testing different edge cases will be more isolated.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTask+Tracking.swift
```diff
@@ -33,4 +33,20 @@ extension DatadogExtension where ExtendedType: URLSessionTask {
...
     }
+
+    var hasCompletion: Bool {
+        get {
+            let value = objc_getAssociatedObject(type, &hasCompletionKey) as? Bool
+            return value == true
+        }
+        set {
+            if newValue {
+                objc_setAssociatedObject(type, &hasCompletionKey, true, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
```
ncreated, line 44: blocking/ We rather mean following, no?
```suggestion
                objc_setAssociatedObject(type, &hasCompletionKey, newValue, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
```

DatadogRUM/Sources/RUMConfiguration.swift
```diff
@@ -221,6 +221,8 @@ extension RUM {
             /// attributes for RUM resource based on the provided request, response, data, and error.
             /// Keep the implementation fast and do not make any assumptions on the thread used to run it.
             ///
+            /// Note: This is not supported for async-await APIs.
+            ///
```
ncreated, line 225: question/ Meaning that we don't fully solve https://github.com/DataDog/dd-sdk-ios/issues/1680 as it was originally reported for async/await APIs. We may need to update the PR description or add more details to the issue, otherwise it will be closed automatically with the merge.

ncreated (changes requested):
blocking/ On top of in-line comments, I think we're missing the fact that `tasks` are pooled in `URLSession` and may be reused for succeeding operations. That means the `task` state set with `task.dd.hasCompletion` must be reset before putting the task back to pool.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTask+Tracking.swift
ganeshnj: not blocking, if check the `if block`, if the newValue is false, we remove it.

ganeshnj:
> blocking/ On top of in-line comments, I think we're missing the fact that `tasks` are pooled in `URLSession` and may be reused for succeeding operations. That means the `task` state set with `task.dd.hasCompletion` must be reset before putting the task back to pool.

What do you mean here? I don't expect same instance of `URLSessionTask` being used again. Rather I expect the connection to be used.

DatadogRUM/Sources/RUMConfiguration.swift
ganeshnj: For sure, this is regression fix. There is no path forward I could see which can solve this issue for async-await APIs like I mentioned in the description of the PR.

Yes, we are not going to close the issue but update it with what we did.

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
ganeshnj: it is matter of where to handle this kind of complexity. We have simpler swizzler, easy to unit test but complex usage which is covered with test cases.

I like this because complexity stays at one place rather distributing to all over.

If you feel strong about it with some clear benefit, we have to do a refactor which I think must not be part of this PR.

ncreated: No strong opinion 👍. If tests coverage is there, at this point it would only imply refactoring with no immediate gain.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTask+Tracking.swift
ncreated: Oh, I've read it as `if let value`! Forget it 🙂👍

DatadogInternal/Sources/NetworkInstrumentation/URLSession/DatadogURLSessionDelegate.swift
```diff
@@ -139,6 +139,7 @@ open class DatadogURLSessionDelegate: NSObject, URLSessionDataDelegate {
             try swizzler.swizzle(
                 interceptCompletionHandler: { [weak self] task, _, error in
                     self?.interceptor?.task(task, didCompleteWithError: error)
+                }, didReceive: { _, _ in
```
maciejburda, line 142: nit: didReceive param could be optional instead

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTask+Tracking.swift
```diff
@@ -33,4 +33,20 @@ extension DatadogExtension where ExtendedType: URLSessionTask {
...
+            } else {
+                objc_setAssociatedObject(type, &hasCompletionKey, nil, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
+            }
```
maciejburda, line 47: Is there any benefit to clear it on false? Maybe that is enough:
```suggestion
             objc_setAssociatedObject(type, &hasCompletionKey, newValue, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
```

not a big deal - just less code to read

maciejburda (approved):
Looks good. Some minor nits to consider :)

DatadogInternal/Sources/NetworkInstrumentation/URLSession/DatadogURLSessionDelegate.swift
ganeshnj: yeah but not worth the refactor.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTask+Tracking.swift
ganeshnj: more of personal preference, I don't think we would gain or lose either way.

PR #1697, opened 2024-02-22
RUM-3222 Move Session Replay obj-c interfaces directly to the target

maciejburda:
### What and why?

Moves obj-c interfaces of Session Replay as described in the [RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3157787243/RFC+-+Modular+Objective-C+Interface)

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


ncreated (commented):
💡 Moving the SR API from `DatadogObjc` to `DatadogSessionReplay` is a breaking-change, no? What is the rationale and mitigation?

maciejburda:
@ncreated 
Truth be told, ticket originated because we missed the fact we already support obj-c in the wrapper package.

We discussed this with @maxep and it should work without a change on the customer side.

It's following the v2 idea of moving all the obj-c api directly to targets and deprecating the obj-c wrapper.

ncreated (dismissed):

maciejburda:
Getting this in snapshot tests:

`target at '~/dd-sdk-ios/DatadogSessionReplay/Tests' contains mixed language source files; feature not supported`

Will try to resolve later. But sounds like it may require slightly different approach.

ncreated (approved):
🎯 nice!

PR #1702, opened 2024-02-28
feat: Add source_type and variant support to logs

fuzzybinary:
### What and why?

Add variant support to logs to assist in symbolication. 
Also add source_type support, which was already partially supported through RUM but not completely on logs.

refs: DataDog/dd-sdk-flutter#568

### Review checklist

- [x] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


DatadogCore/Tests/Datadog/LoggerTests.swift
```diff
@@ -234,6 +235,38 @@ class LoggerTests: XCTestCase {
...
+        let errorMessage = String.mockRandom()
+        let stackTrace = String.mockRandom()
+        logger._internal.log(level: .info,
+                   message: .mockAny(),
+                   errorKind: errorKind,
+                   errorMessage: errorMessage,
+                   stackTrace: stackTrace,
+                   attributes: [
+                    "_dd.error.source_type": "flutter"
+                   ]
```
maciejburda, line 259: super nit: formatting :) (not a blocker)

line should be broken after `log(`

DatadogInternal/Sources/Utils/DDError.swift
```diff
@@ -24,11 +24,13 @@ public struct DDError: Equatable, Codable, PassthroughAnyCodable {
     public let type: String
     public let message: String
     public let stack: String
+    public let sourceType: String
 
-    public init(type: String, message: String, stack: String) {
+    public init(type: String, message: String, stack: String, sourceType: String = "ios") {
```
maciejburda, line 29: I wonder if we should care if it is iOS, iPadOS, tvOS or VisionOS here? 🤔

maciejburda (commented):
Looks good. Not sure if hardcoding `ios` is good enough. We officially support `tvOS` and `iOS`.

DatadogInternal/Sources/Utils/DDError.swift
maxep: It's fine to use `ios` as default here 👍 `source_type` is an [enum](https://github.com/DataDog/rum-events-format/blob/master/schemas/rum/error-schema.json#L112L125) with fixed values. It is used for symbolication and `ios` is an umbrella for apple platforms.

maxep (approved):
LGTM 👍

DatadogCore/Tests/Datadog/LoggerTests.swift
fuzzybinary: 🤔 Why didn't swiftlint complain?

DatadogInternal/Sources/Utils/DDError.swift
fuzzybinary: Yeah, `source_type` here is more "what's the format of the crash report" rather than what's the OS.

PR #1705, opened 2024-03-05
RUM-2925 feat: Add public API for non-fatal App Hangs monitoring

ncreated:
### What and why?

📦 The final touch on https://github.com/DataDog/dd-sdk-ios/pull/1699. Adding public API for App Hangs monitoring in `RUM.Configuration`:

```swift
/// Enables App Hangs monitoring with the given threshold (in seconds).
///
/// Only App Hangs that last more than this threshold will be reported. The minimal allowed value for this option is `0.1` seconds.
/// To disable hangs monitoring, set this parameter to `nil`.
///
/// - Note: Be cautious when setting the threshold to very small values, as it may lead to excessive reporting of hangs.
///         The SDK implements a secondary thread for monitoring App Hangs. To reduce CPU utilization, it tracks hangs with a tolerance of 2.5%, meaning that
///         some hangs lasting very close to this threshold may not be reported.
///
/// - Note: App Hangs monitoring requires Datadog Crash Reporting to be enabled. Otherwise stack trace will be not reported in App Hang errors.
///
/// - Default: `nil` (hangs monitoring disabled).
public var appHangThreshold: TimeInterval?
```

This is limited to non-fatal hangs. Once fatal hangs monitoring is available, it will leverage this option and a new one will be introduced to select which kinds of hangs should be tracked. This follows the agreement from [ADR - App Hangs detection in iOS SDK](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3473147398/ADR+-+App+Hangs+detection+in+iOS+SDK) (internal).

### How?

Adding public API to condition hangs instrumentation.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogRUM/Sources/Instrumentation/RUMInstrumentation.swift
+++ b/DatadogRUM/Sources/Instrumentation/RUMInstrumentation.swift
@@ -9,6 +9,13 @@ import DatadogInternal
 
 /// Bundles RUM instrumentation components.
 internal final class RUMInstrumentation: RUMCommandPublisher {
+    private enum Constants {
+        /// Minimum allowed value for long task threshold configuration.
+        static let minLongTaskThreshold: TimeInterval = 0
+        /// Minimum allowed value for app hang threshold configuration.
+        static let minAppHangThreshold: TimeInterval = 0.1
+    }
+
     /// Swizzles `UIViewController` for intercepting its lifecycle callbacks.
     /// It is `nil` (no swizzling) if RUM View instrumentaiton is not enabled.
     let viewControllerSwizzler: UIViewControllerSwizzler?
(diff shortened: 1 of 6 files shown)


maciejburda (approved):
LGTM 🚀

DatadogRUM/Sources/Instrumentation/RUMInstrumentation.swift
```diff
@@ -78,22 +86,34 @@ internal final class RUMInstrumentation: RUMCommandPublisher {
...
+                appHangs = AppHangsObserver(
+                    appHangThreshold: appHangThreshold,
+                    observedQueue: mainQueue,
+                    backtraceReporter: backtraceReporter,
+                    dateProvider: dateProvider,
+                    telemetry: telemetry
+                )
+            } else {
+                DD.logger.error("`RUM.Configuration.appHangThreshold` cannot be less than 0.1s. App Hangs monitoring will be disabled.")
+            }
```
maxep, line 108: /question When the threshold is lower that the min, shouldn't we still enable the feature using the min value + print an warning? instead of not enable the instrumentation?

maxep (approved):
Looks good! I left a question.

PR #1711, opened 2024-03-07
[Innovation week] Enable compilation for macOS

maciejburda:
### What and why?

These changes enable macOS support for Core, Logs, Trace, Crash Reporting and Web View Tracking as well as add some minor tweaks for VisionOS support.

### Review checklist

- [x] Run integration tests
- [x] Run smoke tests
- [x] Run tests for `tools/`


DatadogCore/Private/ObjcAppLaunchHandler.m
```diff
@@ -39,7 +42,7 @@ + (void)load {
     // This is called at the `DatadogPrivate` load time, keep the work minimal
     _shared = [[self alloc] initWithProcessInfo:NSProcessInfo.processInfo
                                        loadTime:CFAbsoluteTimeGetCurrent()];
-
+#if TARGET_OS_IPHONE
     NSNotificationCenter * __weak center = NSNotificationCenter.defaultCenter;
```
ncreated, line 46: What about `tvOS`?

ncreated (commented):
With adding so many `#if <os>` directives we need at least smoke tests that will verify build compatibility for `macOS`.

DatadogInternal/Sources/Context/DeviceInfo.swift
```diff
@@ -96,4 +96,34 @@ extension DeviceInfo {
         #endif
     }
 }
+#else
```
maxep, line 99: /suggestion I think it should be `#elseif macOS` because we are hard-coding `osName: "macOS"` in this case.

DatadogRUM/Sources/RUMVitals/VitalRefreshRateReader.swift
```diff
@@ -172,7 +172,9 @@ extension FrameInfoProvider {
 extension CADisplayLink: FrameInfoProvider {
     var maximumDeviceFramesPerSecond: Int {
         #if swift(>=5.9) && os(visionOS)
-        120
+        // Hardcoded as for now there's no good way of extracting maximum FPS on VisionOS
+        // https://developer.apple.com/documentation/visionos/analyzing-the-performance-of-your-visionos-app/
+        90
```
maxep, line 177: /question I'm good with the comment but why changing the value here?

maxep: I found it in the doc, maybe the link should https://developer.apple.com/documentation/visionos/analyzing-the-performance-of-your-visionos-app#Inspect-frame-rendering-performance instead?

Package.swift
```diff
@@ -7,7 +7,8 @@ let package = Package(
     name: "Datadog",
     platforms: [
         .iOS(.v11),
-        .tvOS(.v11)
+        .tvOS(.v11),
+        .macOS(.v10_15)
```
ncreated, line 11: question/ 10.15 is macOS Catalina, released in 2019 which means supporting 5 major versions of macOS. Was there an extra motivation to reach that far? 

I think  we can save multiple added complexities if targeting macOS Sonoma and Ventura (`.macOS(.v13)`). That includes all the availability checks we added:
```
#available(..., macOS 12.0, *)
#available(..., macOS 10.15, *)
...
```

dependency-manager-tests/spm/App macOS/ContentView.swift
```diff
@@ -0,0 +1,22 @@
...
+            Image(systemName: "globe")
+                .imageScale(.large)
+            Text("Testing...")
+        }
+        .padding()
+    }
+
+    init() {
+        DatadogSetup.initialize()
+    }
```
maxep, line 21: /suggestion This should be in the init of the App not the view, right?

ncreated (changes requested):
- [x] I think we can limit the impact of this PR if targeting a higher macOS version (more might come with official support). I left a question on this.
- [x] Also, let's change the title of this PR to be more descriptive. Otherwise future selves won't find it.

💪  Other than this is it looks fine - great work on adding a smoke test 🚬 🏅 🚭.

maxep (changes requested):
Looks great, very nice effort! I think we should update the core context provider enablement, especially the `LowPowerModePublisher` which could be enabled for macOS 12 and above.

DatadogInternal/Sources/Context/DeviceInfo.swift
maciejburda: the if is based on `#if canImport(UIKit)` because of `UIDevice` dependency, but I can add it for sanity

SUPPORTED-VERSIONS.md
```diff
@@ -5,13 +5,18 @@
 | **iOS**    |     ✅    |  `11+` |
 | **tvOS**   |     ✅    |  `11+` |
 | **iPadOS** |     ✅    |  `11+` |
-| **VisionOS** |   ⚠️    |  `1.1+` |
+| **visionOS** |   ⚠️    |  `1.0+` |
+| **macOS**  |     ⚠️    |  `10.15+` |
```
ganeshnj: What kind of support we are offering here, I see 
- Mac (I think this is most interesting)
- Mac Catalyst
- Mac Designed for iPad

image

SUPPORTED-VERSIONS.md
```diff
@@ -5,13 +5,18 @@
...
 | **Linux**  |     ❌    |  `n/a` |
 
 ## VisionOS
-VisionOS is not officially supported by Datadog SDK. Some features may not be fully functional.
+
+VisionOS is not officially supported by Datadog SDK. Some features may not be fully functional. Note that `CrashReporting` is not supported on VisionOS, due to lack of support on the [PLCrashReporter side](https://github.com/microsoft/plcrashreporter/issues/288).
+
+## MacOS
+
+MacOS is not officially supported by Datadog SDK. Some features may not be fully functional. Note that `RUM`, which heavily depends on `UIKit` does not build on macOS.
```
ganeshnj: If we know this is not going to build, we must not add RUM support in project. This will cause unnecessary friction.

bitrise.yml
```diff
@@ -360,6 +360,18 @@ workflows:
                 -project "$BITRISE_SOURCE_DIR/dependency-manager-tests/spm/SPMProject.xcodeproj" \
                 -destination "platform=macOS,variant=Mac Catalyst" \
                 | xcpretty
+    - script:
+        title: Check macOS compatibility (build SPMProject for macOS)
```
ganeshnj, line 364: ❓ 
Only SPM?

ganeshnj (changes requested):
```
Compiling for macOS 11, but module 'DatadogCore' has a minimum deployment target of macOS 12.0
```

I tried to compile the SPM dependency manager test app with Xcode 15 and couldn't build. There seems some version mismatch.

DatadogInternal/Sources/Context/DeviceInfo.swift
```diff
@@ -96,4 +96,34 @@ extension DeviceInfo {
...
+        processInfo: ProcessInfo = .processInfo
+    ) {
+        var architecture = "unknown"
+        if let archInfo = NXGetLocalArchInfo()?.pointee {
+            architecture = String(utf8String: archInfo.name) ?? "unknown"
+        }
+        Host.current().name
+
+        let build = (try? Sysctl.osVersion()) ?? ""
+        let model = (try? Sysctl.model()) ?? ""
```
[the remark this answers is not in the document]
maciejburda: They're required by schema

SUPPORTED-VERSIONS.md
maciejburda: I made some adjustment based on previous comment. Now we only support 12+

Now target destination is a pandora box.

I'd like to discuss during the SDK meeting.
I think we can simplify and have only one target with all the destinations, instead of current iOS/tvOS split

SUPPORTED-VERSIONS.md
maciejburda: We can only have one Package definition and as far as I know, and we can't do selective destination (Session Replay is iOS only for example)

bitrise.yml
maciejburda: It follows what we did for Mac Catalyst. It's minimal effort for the smoke test, as the support is not official

DatadogInternal/Sources/Context/DeviceInfo.swift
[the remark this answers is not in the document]
ganeshnj: I see, instead empty string some other string would be helpful in querying like `<unknown>`  or `<unrecognized>`

ganeshnj:
I tried the latest version 7b094e829a58ffba2d11fbce42a5ff0f10d149c2 but still couldn't get the sample SPM app logs and tracing working yet. I think this is minimum bar that the project not only compiles but have some basic things working.

dependency-manager-tests/spm/Shared/DatadogSetup.swift
```diff
@@ -0,0 +1,37 @@
...
+                remoteSampleRate: 0,
+                consoleLogFormat: .short
+            )
+        )
+
+        // Trace APIs must be visible:
+        Trace.enable()
+
+        logger?.info("It works")
+        _ = Tracer.shared().startSpan(operationName: "this too")
```
ganeshnj, line 35: probably because of not finishing I'm not seeing the span.
```suggestion
        let span = Tracer.shared().startSpan(operationName: "this too")
        span.finish()
```

ganeshnj (dismissed):

ncreated (approved):
Looks good 👍. I changed the PR title to be more descriptive. Great initiative, thanks 🏅

ncreated (approved):

maxep (approved):
Thanks!

PR #1715, opened 2024-03-08
RUM-3437 Inject Container to Browser Events

maxep:
### What and why?

As part of WebView Replay requirements, each Browser event should include a `container` field with the native source and `view.id`. As described in [_view-container-schema.json](https://github.com/DataDog/rum-events-format/blob/master/schemas/rum/_view-container-schema.json).

This field will help the player to retrieve the Browser RUM View contained in a native replay.

### How?

The Browser SDK sends views through the JS bridge asynchronously with delays, meaning that a browser event could come from a previous view. See the following user-journey:
```
View 1 (id: 1) -> View 2 (id: 2) -> Back to View 1 (id: 3)
```

Assuming that View 1 presents a web-view, we could receive Browser events contained in view.id 1 while being in view.id 3.

To assign the right container to the right Browser event, we need to keep previous view ids in cache so they can be fetch based on timestamp. This is the role of the introduced `ViewCache`.
The `ViewCache` class is a new RUM scope dependency that is used to store view ids when a new view starts. The `WebViewEventReceiver` can then use that cache to inject the right container to the Browser events.

### Review checklist

- [x] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogRUM/Sources/Integrations/WebViewEventReceiver.swift
+++ b/DatadogRUM/Sources/Integrations/WebViewEventReceiver.swift
@@ -17,6 +17,9 @@ internal final class WebViewEventReceiver: FeatureMessageReceiver {
     /// The date provider.
     let dateProvider: DateProvider
 
+    /// The view cache containing ids of current and previous views.
+    let viewCache: ViewCache
+
     /// Creates a new receiver.
     ///
     /// - Parameters:
@@ -24,10 +27,12 @@ internal final class WebViewEventReceiver: FeatureMessageReceiver {
     ///   - commandSubscriber: Subscriber that can process a `RUMKeepSessionAliveCommand`.
     init(
         dateProvider: DateProvider,
-        commandSubscriber: RUMCommandSubscriber
+        commandSubscriber: RUMCommandSubscriber,
+        viewCache: ViewCache
     ) {
         self.commandSubscriber = commandSubscriber
         self.dateProvider = dateProvider
+        self.viewCache = viewCache
     }
 
     func receive(message: FeatureMessage, from core: DatadogCoreProtocol) -> Bool {
@@ -68,6 +73,14 @@ internal final class WebViewEventReceiver: FeatureMessageReceiver {
                     let serverTimeOffsetInMs = self.getOffsetInMs(viewID: viewID, context: context)
                     let correctedDate = Int64(date) + serverTimeOffsetInMs
                     event["date"] = correctedDate
+
+                    // Inject the container source and view id
+                    if let viewID = self.viewCache.lastView(before: date, hasReplay: true) {
+                        event[RUMViewEvent.CodingKeys.container.rawValue] = RUMViewEvent.Container(
+                            source: RUMViewEvent.Container.Source(rawValue: context.source) ?? .ios,
+                            view: RUMViewEvent.Container.View(id: viewID)
+                        )
+                    }
                 }
 
                 if var application = event["application"] as? JSON {
--- a/DatadogRUM/Sources/RUMMonitor/Scopes/Utils/ViewCache.swift
+++ b/DatadogRUM/Sources/RUMMonitor/Scopes/Utils/ViewCache.swift
@@ -33,7 +33,7 @@ internal final class ViewCache {
     ///   - capacity: The maximum number of ids to store.
     init(
         dateProvider: DateProvider = SystemDateProvider(),
-        ttl: TimeInterval = 3 * 60,
+        ttl: TimeInterval = 3.minutes,
         capacity: Int = 30
     ) {
         self.dateProvider = dateProvider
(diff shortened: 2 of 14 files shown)


DatadogRUM/Sources/Integrations/WebViewEventReceiver.swift
```diff
@@ -68,6 +73,14 @@ internal final class WebViewEventReceiver: FeatureMessageReceiver {
                     let serverTimeOffsetInMs = self.getOffsetInMs(viewID: viewID, context: context)
                     let correctedDate = Int64(date) + serverTimeOffsetInMs
                     event["date"] = correctedDate
```
ncreated, line 75: question/ Could `ViewCache` be used to cleanup this class from NTP offsets caching? I mean [this code](https://github.com/DataDog/dd-sdk-ios/blob/54fae19f7b45787301ae8d0b99d701fc0f69e9ba/DatadogRUM/Sources/Integrations/WebViewEventReceiver.swift#L97-L122), that has similar concepts to what we introduced in `ViewCache`: _get by ID_, _purge_ etc.

DatadogRUM/Sources/RUMMonitor/Scopes/Utils/ViewCache.swift
```diff
@@ -0,0 +1,99 @@
...
+        purge()
+    }
+
+    /// Gets the last view id before the specified timestamp.
+    ///
+    /// - Parameters:
+    ///   - timestamp: The requested epoch timestamp in milliseconds.
+    ///   - hasReplay: Specify `true` to get the last view with replay.
+    /// - Returns: The view id if found.
+    func lastView<Integer>(before timestamp: Integer, hasReplay: Bool? = nil) -> String? where Integer: BinaryInteger {
```
ncreated, line 71: question/ Why do we need generic `BinaryInteger`?

IntegrationTests/IntegrationScenarios/Scenarios/WebView/WebViewScenarioTest.swift
```diff
@@ -16,13 +16,16 @@ class WebViewScenarioTest: IntegrationTests, RUMCommonAsserts {
         let rumServerSession = server.obtainUniqueRecordingSession()
         // Server session recording Logs send to `HTTPServerMock`.
         let loggingServerSession = server.obtainUniqueRecordingSession()
+        // Server session recording Replays send to `HTTPServerMock`.
+        let srServerSession = server.obtainUniqueRecordingSession()
```
ncreated, line 20: question/ Does the web page we display in WebView as part of this test scenario support browser SR? I wonder how this work, it looks great 👍, easy but great coverage.

ncreated (approved):
It looks good 👍. I like the cache implementation - neat and simple.

I left few questions, but no blocking.

maciejburda (approved):
Looking good!

Some minor suggestions

DatadogRUM/Sources/Integrations/WebViewEventReceiver.swift
maxep: The NTP caching is a simple `key->value` cache while the `ViewCache` will look for the closest view in the past using timestamp. So it's different key type and query that are not compatible :/

DatadogRUM/Sources/RUMMonitor/Scopes/Utils/ViewCache.swift
maxep: Because RUM uses `Int64` timestamps while the Browser RUM events coming from the bridge are serialised to `Int`. I'm using `BinaryInteger` to allow querying from both sides.

IntegrationTests/IntegrationScenarios/Scenarios/WebView/WebViewScenarioTest.swift
maxep: Not yet, but I have enabled native SR in this scenario so that the native RUM views have `has_replay` true. It is the condition to inject the `container` field in Browser RUM events.

DatadogRUM/Sources/RUMMonitor/Scopes/Utils/ViewCache.swift
```diff
@@ -0,0 +1,99 @@
...
+ */
+
+import Foundation
+import DatadogInternal
+
+/// The ``ViewCache``  keeps previous view ids in memory.
+///
+/// This class can be used to store and retrieve previous RUM views based
+/// on timestamp.
+internal final class ViewCache {
```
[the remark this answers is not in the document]
maxep: `NSCache` is a simple `key->value` storage, while what we need here is to query the closest view in the past given a timestamp. That's why I keep view ids from newest to oldest. Does it makes sense?

ncreated (approved):

PR #1721, opened 2024-03-12
RUM-1828 feat! (128 bit trace id) support 128 bit trace ids in distributed tracing, spans, rum and log events

ganeshnj:
### What and why?

RFC https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3545630931

### How?

This PR adds support for 128 bit trace ids in the SDK. There are several behind the scenes changes to support 128 bit trace ids which includes new TraceID type to keep the implementation decoupled from SpanID.

1. All distributed tracing headers are updated to support 128 bit trace ids
2. All span and rum events are updated to support 128 bit trace ids
3. All log events are updated to support 128 bit trace ids

Tests are updated accordingly and I changed trace and span ids to the values which are represented differently in hex and decimal unlike 1, 2. This forces us to make sure we are encoding the ids correctly where needed.

This is a breaking change as we are making changes in public types but in Internal module.

### Review checklist

- [x] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogInternal/Sources/NetworkInstrumentation/SpanID.swift
+++ b/DatadogInternal/Sources/NetworkInstrumentation/SpanID.swift
@@ -1,0 +1,161 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+public struct SpanID: RawRepresentable, Equatable, Hashable {
+    public static let invalidId: UInt64 = 0
+    public static let invalid = SpanID()
+
+    /// The `String` representation format of a `SpanID`.
+    public enum Representation {
+        case decimal
+        case hexadecimal
+        case hexadecimal16Chars
+        case hexadecimal32Chars
+    }
+
+    /// The unique integer (64-bit unsigned) ID of the trace containing this span.
+    /// - See also: [Datadog API Reference - Send Traces](https://docs.datadoghq.com/api/?lang=bash#send-traces)
+    public let rawValue: UInt64
+
+    /// Creates a new instance with the specified raw value.
+    ///
+    /// - Parameter rawValue: The raw value to use for the new instance.
+    public init(rawValue: UInt64) {
+        self.rawValue = rawValue
+    }
+
+    public init() {
+        self.rawValue = Self.invalidId
+    }
+
+    public func toString(representation: Representation) -> String {
+        String(self, representation: representation)
+    }
+}
+
+extension SpanID {
+    /// Creates a `SpanID` from a `String` representation.
+    ///
+    /// - Parameters:
+    ///   - string: The `String` representation.
+    ///   - representation: The representation, `.decimal` by default.
+    public init?(_ string: String, representation: Representation = .decimal) {
+        switch representation {
+        case .decimal:
+            guard let rawValue = UInt64(string) else {
+                return nil
+            }
+
+            self.init(rawValue: rawValue)
+        case .hexadecimal, .hexadecimal16Chars, .hexadecimal32Chars:
+            guard let rawValue = UInt64(string, radix: 16) else {
+                return nil
+            }
+
+            self.init(rawValue: rawValue)
+        }
+    }
+}
+
+extension SpanID: ExpressibleByIntegerLiteral {
+    /// Creates an instance initialized to the specified integer value.
+    ///
+    /// Do not call this initializer directly. Instead, initialize a variable or
+    /// constant using an integer literal. For example:
+    ///
+    ///     let id: SpanID = 23
+    ///
+    /// In this example, the assignment to the `id` constant calls this integer
+    /// literal initializer behind the scenes.
+    ///
+    /// - Parameter value: The value to create.
+    public init(integerLiteral value: UInt64) {
+        self.init(rawValue: value)
+    }
+}
+
+extension String {
+    /// Creates a `String` representation of a `SpanID`.
+    ///
+    /// - Parameters:
+    ///   - spanID: The Trace ID
+    ///   - representation: The required representation. `.decimal` by default.
+    public init(_ spanID: SpanID, representation: SpanID.Representation = .decimal) {
+        switch representation {
+        case .decimal:
+            self.init(spanID.rawValue)
+        case .hexadecimal:
+            self.init(spanID.rawValue, radix: 16)
+        case .hexadecimal16Chars:
+            self.init(format: "%016llx", spanID.rawValue)
+        case .hexadecimal32Chars:
+            self.init(format: "%032llx", spanID.rawValue)
+        }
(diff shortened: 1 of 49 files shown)


DatadogInternal/Sources/NetworkInstrumentation/SpanID.swift
```diff
@@ -0,0 +1,161 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-2020 Datadog, Inc.
+ */
+
+import Foundation
+
+public struct SpanID: RawRepresentable, Equatable, Hashable {
+    public static let invalidId: UInt64 = 0
```
maciejburda, line 10: I was wondering when reading tests how it's been done 🤔 

Great idea! 👍 - improves readability!

DatadogInternal/Sources/NetworkInstrumentation/TraceID.swift
```diff
@@ -6,8 +6,6 @@
 
 import Foundation
 
-public typealias SpanID = TraceID
```
maciejburda, line 9: The fact that SpanID is 64-bit and TraceID is 128-bit feels not obvious to me just by name.

For example in the b3 definition it's either 64 or 128 for TraceID and always 64-bit for SpanID.
For W3C and OTel they seem to be aligned.

Maybe worth just naming them more agnostically, like `Identifier64` and `Identifier128`

maciejburda: Actually scratch that. OTel lib is doing exactly that:
https://github.com/open-telemetry/opentelemetry-swift/blob/812e3fca27a0bbc6a281b0b54546694251b54243/Sources/OpenTelemetryApi/Trace/TraceId.swift
https://github.com/open-telemetry/opentelemetry-swift/blob/812e3fca27a0bbc6a281b0b54546694251b54243/Sources/OpenTelemetryApi/Trace/SpanId.swift

maciejburda (dismissed):
LGTM 🚀 

Some heavy work on adjusting all the tests!

DatadogTrace/Sources/Span/SpanEventEncoder.swift
```diff
@@ -247,5 +247,8 @@ internal struct SpanEventEncoder {
             let metaKey = "meta.\($0.key)"
             try container.encode($0.value, forKey: DynamicCodingKey(metaKey))
         }
+
+        let metaDdPTid = "meta._dd.p.tid"
+        try container.encode(span.traceID.idHiHex, forKey: DynamicCodingKey(metaDdPTid))
```
ncreated, line 252: change-request/ This method is made for encoding "custom meta" fields using dynamic coding keys:
```swift
/// Encodes `meta.*` attributes coming from user
private func encodeCustomMeta(_ span: SpanEvent, to container: inout KeyedEncodingContainer<DynamicCodingKey>) throws {
```

There is another one, declared above, which encodes default tags using type-safe keys. We should prefer it for this use case. It will add more clarity and avoid allocating string literal each time:
```swift
/// Encodes default `meta.*` attributes
private func encodeDefaultMeta(_ span: SpanEvent, to container: inout KeyedEncodingContainer<StaticCodingKeys>) throws {
```

TestUtilities/Mocks/NetworkInstrumentationMocks.swift
```diff
@@ -28,15 +47,37 @@ public class RelativeTracingUUIDGenerator: TraceIDGenerator {
...
+            defer { uuid = uuid + (0, count) }
+            return uuid
+        }
+    }
+}
+
+public class RelativeSpanIDGenerator: SpanIDGenerator {
+    private(set) var uuid: SpanID
+    internal let count: UInt64
+    private let queue = DispatchQueue(label: "queue-RelativeTracingUUIDGenerator-\(UUID().uuidString)")
```
ncreated, line 60: suggestion/ Using `@ReadWriteLock var uuid: SpanID` might be much cleaner than syncing with queue. We can update `RelativeTracingUUIDGenerator` with this change as well. Rule of thumb: with property wrappers being available, we can simplify a lot of synchronisation in tests by replacing dummy queues with this convenient lock.

DatadogInternal/Sources/NetworkInstrumentation/SpanID.swift
```diff
@@ -0,0 +1,161 @@
...
+            }
+            self = spanID
+        }
+    }
+
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.singleValueContainer()
+        try container.encode(rawValue)
+    }
+}
```
ncreated, line 161: suggestion/ We're missing unit tests coverage for `SpanID` coding (in `DatadogInternalTests`). The overall coverage for `SpanID` looks quite low given the complexity we manage here - could we improve it?

Screenshot 2024-04-09 at 10 29 02

DatadogInternal/Sources/NetworkInstrumentation/TraceID.swift
```diff
@@ -38,17 +107,41 @@ extension TraceID {
...
-        case .hexadecimal, .hexadecimal16Chars, .hexadecimal32Chars:
-            guard let rawValue = UInt64(string, radix: 16) else {
+            self.init(idLo: idLo)
+        case .hexadecimal32Chars:
+            guard let idLo = UInt64(string, radix: 16), let idHi = UInt64(string.prefix(16), radix: 16) else {
                 return nil
             }
 
-            self.init(rawValue: rawValue)
+            self.init(rawValue: (idHi, idLo))
```
ncreated, line 126: suggestion/ These two seem to have no unit tests coverage (in `DatadogInternalTests`). Overall score for `TraceID` seems low given that this is critical piece. Could we improve it?

Screenshot 2024-04-09 at 10 28 51

ncreated (changes requested):
It looks great 💪 - I really appreciate the level of detail in this implementation 🏅. Leaving one change-request and few comments - mainly about the unit tests coverage which seems a bit too low given the importance of `TraceID` and `SpanID` in our SDK.

DatadogTrace/Sources/Span/SpanEventEncoder.swift
ganeshnj: sure. fixed.

DatadogInternal/Sources/NetworkInstrumentation/SpanID.swift
ganeshnj: fair call, I added bunch of cases now.

DatadogInternal/Sources/NetworkInstrumentation/TraceID.swift
ganeshnj: fair call, I added bunch of cases now.

ncreated (approved):
💯 🏆

ncreated (approved):

PR #1726, opened 2024-03-13
RUM-2925 fix: Refine non-fatal App Hang tracking

ncreated:
### What and why?

📦 🧰 This PR refines non-fatal App Hangs tracking after gathering first feedback in dogfooding in Shopist iOS.

### How?

Two improvements are applied
- 1/ The hanged thread is not marked as "crashed" in hang error (`error.threads[0].crashed`);
- 2/ An extra telemetry is sent when backtrace generation fails.

The root cause of second issue is yet unknown. It can't be reproduced in local, but it is prominent in dogfooding with Shopist and s8s devices. To get more visibility into problem, an error telemetry is collected when backtrace generation fails. To make it clear for early adopters, an error message is adjusted for this case as agreed with @MaelNamNam:

```
Failed to generate stack trace. This is a known issue and we work on it.
```

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


maciejburda (dismissed):
Nice! Adding this telemetry actually brought some order to error handling 💪

maciejburda (approved):

PR #1725, opened 2024-03-13
RUM-3462 refactor: Make TLV implementation reusable

ncreated:
### What and why?

📦 This PR makes existing TLV implementation generic, so it can be used in upcoming implementation of `DataStore` to persist feature-specific data between app launches.

### How?

// TBD

### Review checklist
- [ ] Feature or bugfix MUST have appropriate tests (unit, integration)
- [ ] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes

### Custom CI job configuration (optional)
- [ ] Run unit tests for Core, RUM, Trace, Logs, CR and WVT
- [ ] Run unit tests for Session Replay
- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

PR #1730, opened 2024-03-19
RUM-3462 feat: Introduce `DataStore`

ncreated:
### What and why?

📦 This PR introduces a `DataStore` interface designed to store key-value pairs for a given feature. The primary goal is to provide a mechanism for persisting data between app launches.

For the full list of use cases and requirements considered in this implementation, see [RFC - Local Datastore (internal)](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3327557955/RFC+-+Local+Datastore+Problem+Statement).

### How?

**New APIs**

`DataStore` interface:
```swift
public protocol DataStore {
    func setValue(_ value: Data, forKey key: String, version: DataStoreKeyVersion = 0) {}
    func value(forKey key: String, callback: @escaping (DataStoreValueResult) -> Void) {}
    func removeValue(forKey key: String) {}
}

public enum DataStoreValueResult {
    case value(Data, DataStoreKeyVersion)
    case noValue
    case error(Error)
}

public typealias DataStoreKeyVersion = UInt16
```
`DataStoreKeyVersion` (an `UInt16`): Values are persisted in files and SDK upgrades may occur between app launches. To ensure proper data reading, the caller can specify the version of expected data on read. This way it can handle the situation when data was written in old format, so it must skip decoding it in the new format.

The `DataStore` is specific for a feature. It can be obtained through `FeatureScope` interface:
```swift
public extension FeatureScope {
   func dataStore() -> DataStore
   func dataStoreContext(_ block: @escaping (DatadogContext, DataStore) -> Void) {
}
```
This comes in-line with Event Write Context API that already exists on the `FeatureScope`:
```swift
public protocol FeatureScope {
   func context(_ block: @escaping (DatadogContext) -> Void)
   func eventWriteContext(_ block: @escaping (DatadogContext, Writer) -> Void) {
}
```

**TLV format**

Data is encoded in TLV format, leveraging prior work on batch files and recent refinements in  https://github.com/DataDog/dd-sdk-ios/pull/1725.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogCore/Sources/Core/TLV/TLVBlock.swift
+++ b/DatadogCore/Sources/Core/TLV/TLVBlock.swift
@@ -8,10 +8,10 @@ import Foundation
 import DatadogInternal
 
 /// Block size binary type
-internal typealias BlockSize = UInt32
+internal typealias TLVBlockSize = UInt32
 
 /// Reported errors while manipulating data blocks.
-internal enum DataBlockError: Error {
+internal enum TLVBlockError: Error {
     case readOperationFailed(streamStatus: Stream.Status, streamError: Error?)
     case invalidDataType(got: Any)
     case invalidByteSequence(expected: Int, got: Int)
@@ -50,8 +50,8 @@ internal struct TLVBlock<BlockType: TLVBlockType> {
         // T
         withUnsafeBytes(of: type.rawValue) { buffer.append(contentsOf: $0) }
         // L
-        guard let length = BlockSize(exactly: data.count), length <= maxLength else {
-            throw DataBlockError.bytesLengthExceedsLimit(limit: maxLength)
+        guard let length = TLVBlockSize(exactly: data.count), length <= maxLength else {
+            throw TLVBlockError.bytesLengthExceedsLimit(limit: maxLength)
         }
         withUnsafeBytes(of: length) { buffer.append(contentsOf: $0) }
         // V
--- a/DatadogCore/Sources/Core/TLV/TLVBlock.swift
+++ b/DatadogCore/Sources/Core/TLV/TLVBlock.swift
@@ -19,17 +19,10 @@ internal enum TLVBlockError: Error {
     case dataAllocationFailure
     case endOfStream
 }
-
-internal protocol TLVBlockType {
-    associatedtype RawValue = UInt16
-    init?(rawValue: RawValue)
-    var rawValue: RawValue { get }
-}
-
 /// A data block in defined by its type and a byte sequence.
 ///
 /// A block can be serialized in data stream by following TLV format.
-internal struct TLVBlock<BlockType: TLVBlockType> {
+internal struct TLVBlock<BlockType> where BlockType: RawRepresentable, BlockType.RawValue == UInt16 {
     /// Type describing the data block.
     let type: BlockType
 
--- a/DatadogInternal/Sources/DataStore/DataStore.swift
+++ b/DatadogInternal/Sources/DataStore/DataStore.swift
@@ -1,0 +1,83 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+/// Represents the version of data format stored for a key in the data store.
+///
+/// As values are persisted, the format of data may change between different SDK versions. To prevent decoding errors,
+/// callers can utilize `DataStoreKeyVersion` to explicitly specify the version of serialized `Data` before attempting deserialization.
+public typealias DataStoreKeyVersion = UInt16
+
+/// The default version of data stored for keys (equals `0`).
+public let dataStoreDefaultKeyVersion: DataStoreKeyVersion = 0
+
+/// Possible results of retrieving a value from the data store.
+public enum DataStoreValueResult {
+    /// The value was found and serialized using the format defined by the specified version.
+    case value(Data, DataStoreKeyVersion)
+    /// There was no value associated with the requested key.
+    case noValue
+    /// The value could not be read due to an underlying error.
+    /// This may represent an error with the file format or an I/O exception that occurred during reading.
+    case error(Error)
+
+    /// Retrieves the data value associated with the result, if it matches the expected version.
+    ///
+    /// - Parameter expectedVersion: The version expected for the retrieved data.
+    /// - Returns: The data value if the version matches the expected version; otherwise, nil.
+    public func data(expectedVersion: DataStoreKeyVersion = dataStoreDefaultKeyVersion) -> Data? {
+        guard case .value(let data, let storedVersion) = self, storedVersion == expectedVersion else {
+            return nil
+        }
+        return data
+    }
+}
+
+/// Defines the interface for a data store capable of storing key-value pairs for a given feature.
+public protocol DataStore {
+    /// Sets the value for the specified key in the data store.
+    ///
+    /// - Parameters:
+    ///   - value: The data to be stored.
+    ///   - key: The unique identifier for the data. Must be a valid file name, as it will be persisted in files.
+    ///   - version: The version of the data format. Defaults to `0`.
+    func setValue(_ value: Data, forKey key: String, version: DataStoreKeyVersion)
+
+    /// Retrieves the value associated with the specified key from the data store.
+    ///
+    /// - Parameters:
+    ///   - key: The unique identifier for the data. Must be a valid file name, as it will be persisted in files.
+    ///   - callback: A closure providing the result asynchronously on an internal queue.
+    func value(forKey key: String, callback: @escaping (DataStoreValueResult) -> Void)
+
+    /// Deletes the value associated with the specified key from the data store.
+    ///
+    /// - Parameter key: The unique identifier for the value to be deleted. Must be a valid file name, as it will be persisted in files.
+    func removeValue(forKey key: String)
+}
+
+public extension DataStore {
+    /// Sets the value for the specified key in the data store with the default `version` of `0`.
+    ///
+    /// - Parameters:
+    ///   - value: The data to be stored.
+    ///   - key: The unique identifier for the data. Must be a valid file name, as it will be persisted in files.
+    func setValue(_ value: Data, forKey key: String) {
+        setValue(value, forKey: key, version: dataStoreDefaultKeyVersion)
+    }
+}
+
+public struct NOPDataStore: DataStore {
+    public init() {}
+
+    /// no-op
+    public func setValue(_ value: Data, forKey key: String, version: DataStoreKeyVersion) {}
+    /// no-op
+    public func value(forKey key: String, callback: @escaping (DataStoreValueResult) -> Void) {}
+    /// no-op
+    public func removeValue(forKey key: String) {}
+}
--- a/DatadogInternal/Sources/DatadogCoreProtocol.swift
+++ b/DatadogInternal/Sources/DatadogCoreProtocol.swift
@@ -224,9 +224,8 @@ public protocol FeatureScope {
     /// - Parameter block: The block to execute; it is called on the context queue.
     func context(_ block: @escaping (DatadogContext) -> Void)
 
-    /// Retrieve data store.
-    /// - Returns: Data store configured for storing data for this feature.
-    func dataStore() -> DataStore
+    /// Data store configured for storing data for this feature.
+    var dataStore: DataStore { get }
 }
 
 /// Feature scope provides a context and a writer to build a record event.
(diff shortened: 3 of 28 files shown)


DatadogCore/Sources/Core/DataStore/CoreDataStore.swift
```diff
@@ -0,0 +1,186 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+import DatadogInternal
+
+/// An entry stored in Data Store files.
+internal struct CoreDataStoreEntry: Equatable {
```
maciejburda, line 11: `CoreData` is very misfortunate name for this one 😄 
One thing I considered when working on mine was using `LocalStore`. So this would be `CoreLocalStoreEntry`. But maybe just dropping `Core` from everything would be enough.

Not a hard blocker, but IMO `CoreData` thing may confuse future developers.

DatadogCore/Sources/Core/DataStore/CoreDataStore.swift
```diff
@@ -0,0 +1,186 @@
...
+    let policyData: PolicyData
+    /// The version of the value associated with the entry.
+    let valueVersion: DataStoreKeyVersion
+    /// The data for this entry.
+    let valueData: Data
+}
+
+/// A concrete implementation of the DataStore protocol using file storage.
+internal class CoreDataStore: DataStore {
+    enum Constants {
```
maciejburda, line 30:
```suggestion
    private enum Constants {
```

DatadogCore/Sources/Core/DataStore/CoreDataStore.swift
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    private func deleteData(forKey key: String) throws {
+        if directory.hasFile(named: key) {
+            try directory.file(named: key).delete()
+        }
+    }
+}
+
+extension CoreDataStore: Flushable {
```
maciejburda, line 181: nice!

maciejburda (dismissed):
Wow! Really nice piece 💪 

Great tests, nice docs. I like how it's exposed to features 🎖️ 

One thing I would consider around naming is dropping `Core` to avoid having `CoreData`. It might confuse some developers.

DatadogInternal/Sources/DataStore/DataStore.swift
```diff
@@ -0,0 +1,96 @@
...
+    ///   - policy: The policy governing access to the stored data.
+    func set(value: Data, forKey key: String, version: DataStoreKeyVersion, policy: DataStoreKeyPolicy)
+
+    /// Retrieves the value associated with the specified key from the data store.
+    ///
+    /// - Parameters:
+    ///   - key: The unique identifier for the data. Must be a valid file name, as it will be persisted in files.
+    ///   - version: The version expected for the retrieved data.
+    ///   - callback: A closure to be executed once the retrieval operation completes, providing the result.
+    func value(forKey key: String, version: DataStoreKeyVersion, callback: @escaping (DataStoreValueResult) -> Void)
```
ncreated, line 52: @maciejburda @maxep I'm happy to discuss alternatives to this design, notably for removing `policy` from the "set value" API. The interface proposed in this PR might be a bit confusing for values set in one place and read in another part of the codebase (like "last RUM view" information). The more "set" and "get" are distanced, the harder it is to plan the right storage policies.

Considering policies we introduce in this PR:

`deleteUponRead` could be replaced with either:
-  A return value in "get value" callback - this would let the caller decide if delete or not based on the value that was read. It will make a compromise on atomicity assuming that the caller makes a synchronous side-effect right upon read and app is terminated due to crash (data store may not be given enough time to delete). In practice tho, all writes in our codebase are async, hence the risk of race condition is very low.
- A `delete` parameter in "get value" API - this way the caller can decide upfront if the value should be deleted after read. It will give the same atomicity as current implementation.

`sameAppVersion` and `sameOSVersion` can be replaced with either:
-  The caller encoding both informations as part of the value. That would require the caller to access `DatdogContext` which is not always convenient (e.g. we don't have it in App Hangs observer).
-  By considering them standard attributes of `DataStoreValueResult` (set by data store implementation).

There are perhaps other options that we can think of. I don't mind doing a short 🧠⚡  meeting on this design. LMK.

maxep: I was writing a comment about that :) The `PolicyData` is too opinionated and add a level of complexity regarding queues sync. IMO, we should let the feature handling its data and constraints: If its data is linked to a app/os version or any other context data, then the caller will have to get that info before writing.
Same goes for the `DataStoreKeyVersion`: If the format has changed, then the feature will fail to decode and that's it, no need to care about it from the core perspective.

I think the policy should be more about rules on how to manage persistency of the data: like `deleteUponRead` that is present here. We could also think of ttl or expiration date policy, tho it should be added only if we have a use case for it.

I'm happy to have brainstorming session on this 👍

DatadogInternal/Sources/Utils/WithErrorContext.swift
```diff
@@ -0,0 +1,65 @@
...
+///
+/// - Throws: An `EnrichedError` containing the original error and the provided context.
+///
+/// - Returns: The result of the closure if no error is thrown.
+public func withErrorContext<T>(
+    fileID: StaticString = #fileID,
+    line: UInt = #line,
+    _ context: @autoclosure () -> String,
+    _ block: () throws -> T
+) throws -> T {
```
maxep, line 25: I'm not a fan of this TBH, but I see the point. I wonder if we could have an `InternalError` extension or something 🤔

maxep: Also, the function should `rethrows` instead.

DatadogInternal/Sources/DataStore/DataStore.swift
```diff
@@ -0,0 +1,83 @@
...
+    /// Retrieves the data value associated with the result, if it matches the expected version.
+    ///
+    /// - Parameter expectedVersion: The version expected for the retrieved data.
+    /// - Returns: The data value if the version matches the expected version; otherwise, nil.
+    public func data(expectedVersion: DataStoreKeyVersion = dataStoreDefaultKeyVersion) -> Data? {
+        guard case .value(let data, let storedVersion) = self, storedVersion == expectedVersion else {
+            return nil
+        }
+        return data
+    }
```
ncreated, line 37: Alternatively we can change the semantic to make it always return `Data` (if found) with performing an optional check on `expectedVersion` (if provided) as in following code. I have no strong opinion.
```swift
public func data(expectedVersion: DataStoreKeyVersion? = nil) -> Data? {
    guard case .value(let data, let storedVersion) = self else {
        return nil
    }
    if let expectedVersion == expectedVersion, storedVersion != expectedVersion {
        return nil
    }

    return data
}
```

maciejburda: not sure if version expectation should be built in.

I would assume that the consumer of the api will handle different model version migration

DatadogInternal/Sources/DatadogCoreProtocol.swift
```diff
@@ -223,6 +223,10 @@ public protocol FeatureScope {
     ///
     /// - Parameter block: The block to execute; it is called on the context queue.
     func context(_ block: @escaping (DatadogContext) -> Void)
+
+    /// Retrieve data store.
+    /// - Returns: Data store configured for storing data for this feature.
+    func dataStore() -> DataStore
```
ncreated, line 229: The presence of `DataStore` in `FeatureScope` interface [limits its usage](https://github.com/DataDog/dd-sdk-ios/blob/60eae0ade10d9273f55636c38c6523085de6b339/DatadogCore/Sources/Core/DatadogCore.swift#L294-L296) to only `DatadogRemoteFeatures` (the ones that can write and read batch files). In consequence, it won't be available in following features:
- `CrashReportingFeature`
- `NetworkInstrumentationFeature`
- `BacktraceReportingFeature`

While last two are rather fine, there might be use cases for `DataStore` in Crash Reporting. I don't foresee any at this point, hence I don't consider it blocking.

Even if `DataStore` is required for `CrashReportingFeature` at some point, I rather thing we should work on the `FeatureScope` interface rather than re-design the `DataStore` management. I'm open for discussion 🙂.

DatadogCore/Sources/Core/DataStore/CoreDataStore.swift
ncreated: Changed to `FeatureDataStore` as it better follows the agreed design 👍.

DatadogInternal/Sources/DataStore/DataStore.swift
ncreated: The PR is now updated with the new discussed design 👍. Thanks for sharing your points @maxep, I think we have something (simpler and) better now 💪. No more `policy` consideration in `DS` implementation.

DatadogInternal/Sources/DataStore/DataStore.swift
ncreated: The `version` is now made a pure convenience. It is fully hidden in the basic usage:
```swift
store.setValue(data, forKey: "key")
store.value(forKey: "key") { result in
   _ = result.data()
}
```

If only versioning problem occurs, we can leverage it:
```swift
store.setValue(data_in_new_version, forKey: "key", version: 2)
store.value(forKey: "key") { result in
   _ = result.data(expectedVersion: 2)
}
```

I think this is the best balance between having this support and keeping a clean API. Requiring the caller to handle versioning will cause unnecessary repetitive code infection across different places.

DatadogInternal/Sources/Utils/WithErrorContext.swift
ncreated: Removed 👍 as now we encode less blocks in `DS`. TLV errors are now enriched manually.

DatadogInternal/Sources/DataStore/DataStore.swift
ncreated: > Requiring the caller to handle versioning will cause unnecessary repetitive code infection across different places.

Also, if `version` is not encoded for every data by upfront, we won't be able to ever support migration from "no version" to "version: 2". The decoding failure will become ambiguous and (if not filtered somehow) will flood telemetry with false-positives.

DatadogCore/Sources/Core/TLV/TLVBlock.swift
maxep, line 1: /suggestion I have noticed the definition of `protocol TLVBlockType` in #1725 but it's actually not necessary because any `enum` with raw value is already compliant with `RawRepresentable`. So we can use `RawRepresentable` directly, like so:

```swift
internal struct TLVBlock<BlockType> where BlockType: RawRepresentable, BlockType.RawValue == UInt16 {
  ...
}

internal final class TLVBlockReader<BlockType> where BlockType: RawRepresentable, BlockType.RawValue == UInt16 {
  ...
}
```

TestUtilities/Helpers/DDAssert.swift
```diff
@@ -208,3 +208,17 @@ public func DDAssertDictionariesNotEqual(_ expression1: @autoclosure () throws -
...
+    _DDEvaluateAssertion(message: message(), file: file, line: line) {
+        do {
+            let result = try expression()
+            throw DDAssertError.expectedFailure("Did not throw an error (returned `\(result)` instead)")
+        } catch let error where !(error is DDAssertError) {
+            guard let expectedError = error as? E else {
+                throw DDAssertError.expectedFailure("Did throw an error but it is not of `\(E.self)` type (got `\(type(of: error))` instead)")
+            }
+            errorHandler(expectedError) // expected
+        }
```
maxep, line 222: /nit
```suggestion
        do {
            let result = try expression()
            throw DDAssertError.expectedFailure("Did not throw an error (returned `\(result)` instead)")
        } catch let error as E {
            errorHandler(error) // expected
        } catch let error as DDAssertError {
            throw error
        } catch {
            throw DDAssertError.expectedFailure("Did throw an error but it is not of `\(E.self)` type (got `\(type(of: error))` instead)")
        }
```

DatadogInternal/Sources/DatadogCoreProtocol.swift
maxep: >  I rather thing we should work on the FeatureScope interface rather than re-design the DataStore management.

Totally 👍 

/suggestion can we use a `var` instead?

maxep (approved):
It looks great, well done!!

I have left minor suggestions but it's good to go 👍

maciejburda (approved):
Nice! Thanks for taking the extra effort and improving the API 💪

maxep (approved):
Nice!

Issue #1739, opened 2024-03-20
Swift Package Manager warning: No rule to process file Datadog.modulemap


ejensen:
### Describe the bug

When integrating the DataDog SDK through Swift Package Manager, a warning is generated about `Datadog.modulemap` being included in the package.

### Reproduction steps

Integrate the DataDog SDK with Swift Package Manager, or just open the `Package.swift` in this repository. Notice the warning about "No rule to process file Datadog.modulemap"

### Affected SDK versions

2.8.1

### Latest working SDK version

N/A

### Did you confirm if the latest SDK version fixes the bug?

Yes

### Integration Methods

SPM

### Xcode Version

Xcode 15.3 (15E204a)

### Swift Version

Swift 5.10

maciejburda:
Thanks reporting and providing the fix! 

We'll review with the team as soon as possible and let you know in the Pull Request.

maxep:
We have removed the modulemap in #1746

PR #1740, opened 2024-03-20
Fix Swift Package Manager warning for Datadog.modulemap

ejensen:
### What and why?
Fix the Swift Package warning for unprocessable files. Issue #1739

### How?
Added the modulemap file to the package target's exclusion list.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -54,6 +54,9 @@ let package = Package(
                 .target(name: "DatadogPrivate"),
             ],
             path: "DatadogCore",
+            exclude: [
+                "Sources/Datadog.modulemap"
+            ],
             sources: ["Sources"],
             resources: [
                 .copy("Resources/PrivacyInfo.xcprivacy")


ejensen:
> Unfortunately it fails on smoke test for Mac Catalyst:
> 
> `xcodebuild build -scheme 'App iOS' -project $PATH/dependency-manager-tests/spm/SPMProject.xcodeproj -destination 'platform=macOS,variant=Mac Catalyst'`
> 
> I'll try to take a deeper look next week, but on the first glance it looks hard to resolve as there's no conditional exclude available.

Running that command completes successfully with these changes. The only warning is the following, but it is not new to these changes:
> warning: Run script build phase '⚙️ Run linter' will be run during every build because it does not specify any outputs. To address this warning, either add output dependencies to the script phase, or configure it to run in every build by unchecking "Based on dependency analysis" in the script phase. (in target 'App iOS' from project 'SPMProject')

The original Swift Package warning detailed in Issue #1739 states the module map is not being processed by the build system since it has no rules that specify how to process the file. This warning is for all platforms including Mac Catalyst. So would be counter-intuitive if the file the build system is warning about and skipping would cause a new build error by explicitly skipping.

maciejburda:
Maybe it's CI flakiness. I'll try restarting

maxep:
Hey @ejensen 👋 

I'm closing this in favor of #1746. The modulemap was added a while ago for compatibility with early SPM, it is no longer necessary so we can remove it.

Thanks again for the report and contribution!

PR #1741, opened 2024-03-20
feat: Support eventMapper on crashes for logs

fuzzybinary:
### What and why?

This adds a call to the configured log event mapper when a crash is reported to the Logs feature.  It's tested in the integration test app by adding a fingerprint to the reported crash.

refs: RUM-2209

### Review checklist

- [x] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/Datadog/Example/ExampleAppDelegate.swift
+++ b/Datadog/Example/ExampleAppDelegate.swift
@@ -21,6 +21,12 @@ var rumMonitor: RUMMonitorProtocol { RUMMonitor.shared() }
 class ExampleAppDelegate: UIResponder, UIApplicationDelegate {
     var window: UIWindow?
 
+    func logEventMapper(event: LogEvent) -> LogEvent {
+        var event = event
+        event.error?.fingerprint = "mapped"
+        return event
+    }
+
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
         if Environment.isRunningUnitTests() {
             return false
@@ -44,6 +50,7 @@ class ExampleAppDelegate: UIResponder, UIApplicationDelegate {
         // Enable Logs
         Logs.enable(
             with: Logs.Configuration(
+                eventMapper: logEventMapper,
                 customEndpoint: Environment.readCustomLogsURL()
             )
         )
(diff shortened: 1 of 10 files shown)


DatadogLogs/Sources/Feature/MessageReceivers.swift
```diff
@@ -191,6 +191,7 @@ internal struct CrashLogReceiver: FeatureMessageReceiver {
 
     /// Time provider.
     let dateProvider: DateProvider
+    let logEventMapper: LogEventMapper?
```
maxep, line 194: /siggestion For convenience, we can have an `init` that default the mapper to `nil`

Datadog/Example/ExampleAppDelegate.swift
```diff
@@ -21,6 +21,12 @@ var rumMonitor: RUMMonitorProtocol { RUMMonitor.shared() }
 class ExampleAppDelegate: UIResponder, UIApplicationDelegate {
     var window: UIWindow?
 
+    func logEventMapper(event: LogEvent) -> LogEvent {
+        var event = event
+        event.error?.fingerprint = "mapped"
+        return event
+    }
+
```
maxep, line 29: /question Do we need this in the Example app?

maxep (commented):
Looks good 👍  would be convenient to default the mapper to `nil` in the crash receiver, WDYT?

Datadog/Example/ExampleAppDelegate.swift
fuzzybinary: No, sorry, I put this in for testing and forgot to remove it.

DatadogLogs/Sources/Feature/MessageReceivers.swift
ncreated: I also prefer being explicit. This way compiler will guide us on all the places where `MessageReceiver` is used and will make us think of injecting mapper where it makes sense. If we seek for testing convenience, it is safer to add it through extension in test module.

ncreated (dismissed):
👍 Looks good, but please add `CHANGELOG` entry 🙂

ncreated (approved):

maxep (approved):

PR #1742, opened 2024-03-21
feat: Allow crash reporting to use event mappers

fuzzybinary:
### What and why?

CrashReportReceiver now supports calling both the view and event mapper before writing the crash to storage.

refs: RUM-2206

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+* [FEATURE] Call RUM's `errorEventMapper` for crashes.
+
 # 2.8.0 / 19-03-2024
 
 - [FEATURE] App Hangs are tracked as RUM errors. See [#1685][]
--- a/DatadogCore/Tests/Datadog/RUM/Integrations/CrashReportReceiverTests.swift
+++ b/DatadogCore/Tests/Datadog/RUM/Integrations/CrashReportReceiverTests.swift
@@ -402,6 +402,7 @@ class CrashReportReceiverTests: XCTestCase {
         DDAssertReflectionEqual(sendRUMViewEvent.os, lastRUMViewEvent.os)
     }
 
+    //
     func testGivenCrashDuringRUMSessionWithActiveView_whenSendingRUMErrorEvent_itIsLinkedToPreviousRUMSessionAndIncludesCrashInformation() throws {
         let lastRUMViewEvent: RUMViewEvent = .mockRandomWith(crashCount: 0)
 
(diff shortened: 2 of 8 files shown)


DatadogCore/Tests/Datadog/RUM/Integrations/CrashReportReceiverTests.swift
```diff
@@ -993,4 +1129,90 @@ class CrashReportReceiverTests: XCTestCase {
             expectViewURL: RUMOffViewEventsHandlingRule.Constants.backgroundViewURL
         )
     }
+
+    func testGivenCrashDuringAppLaunchWithEventMappers_whenSending_itSendsWithMappedEvents() throws {
```
ncreated, line 1133: suggestion/ This test feels redundant as we already cover the two control flows that call mappers (the new "with active view" and "without active view" tests). What seems lacking tho is a test that will ensure that view event is not sent when crash (error) is dropped from the mapper. // depends on what we decide for dropping crashes 🙂

ncreated (dismissed):
Looks great 👍. I left one suggestion, non-blocking. LMK what we decide on the possibility of dropping crashes from mapper.

Also, please update `CHANGELOG.md` 🙏

DatadogCore/Tests/Datadog/RUM/Integrations/CrashReportReceiverTests.swift
fuzzybinary: Replaced this test with one that checks that we don't drop crashes from the mapper.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+* [FEATURE] Call RUM's `errorEventMapper` for crashes.
 * [FEATURE] Support calling log event mapper for crashes.
```
ncreated, line 4: nit/ Let's add PR references to both to be consistent with the rest of this file. It helps when generating release notes and later when tracking changes.

DatadogRUM/Sources/Integrations/CrashReportReceiver.swift
```diff
@@ -203,7 +206,9 @@ internal struct CrashReportReceiver: FeatureMessageReceiver {
             DD.logger.debug("Sending crash as RUM error.")
             core.scope(for: RUMFeature.self).eventWriteContext(bypassConsent: true) { context, writer in
                 let rumError = createRUMError(from: crashReport, and: lastRUMViewEvent, crashDate: crashTimings.realCrashDate, sourceType: context.nativeSourceOverride)
-                writer.write(value: rumError)
+                if let mappedError = self.eventsMapper.map(event: rumError) {
+                    writer.write(value: mappedError)
+                }
```
ncreated, line 211: question/ Why do we allow dropping crash in this mapper call? Seems inconsistent with the next crash write in this file. Is it because in this case it goes without corresponding view event? If it's on purpose, I think it may make sense to add code comment on this decision.

ncreated (dismissed):
🎯 Well integrated 🔗. I left two minor comments.

ncreated (approved):
@fuzzybinary I updated this branch with git conflicts resolution after my merge of https://github.com/DataDog/dd-sdk-ios/pull/1750. It should be all fine now 👍.

The notable change is that from #1750 all RUM components (including `CrashReportReceiver`) depend on `FeatureScope` instead of `DatadogCoreProtocol`. That removes the use of `PassthroughCoreMock` in all RUM unit tests, replacing them with simpler `FeatureScopeMock`. This achieves goals outlined in #1744

ncreated (approved):

DatadogRUM/Sources/Integrations/CrashReportReceiver.swift
fuzzybinary: No it's not on purpose. I'll make sure the test covers this and fix.

DatadogCore/Tests/Datadog/RUM/Integrations/CrashReportReceiverTests.swift
```diff
@@ -551,6 +552,174 @@ class CrashReportReceiverTests: XCTestCase {
...
+                }
+            )
+        )
+
+        // When
+        XCTAssertTrue(
+            receiver.receive(message: .baggage(
+                key: MessageBusSender.MessageKeys.crash,
+                value: MessageBusSender.Crash(report: crashReport, context: crashContext)
+            ), from: core)
```
ncreated, line 666: This doesn't compile. With my latest changes on `develop`, we should now test against `FeatureScopeMock` not the `core`. See surrounding tests in this file, I updated them this morning.

fuzzybinary: Ugh, sorry, pulled and rebased and didn't see any conflicts so assumed it was good. Fixed now.

ncreated (approved):

PR #1751, opened 2024-03-27
RUM-3461 feat: track Fatal App Hangs

ncreated:
### What and why?

📦 This PR brings the main part of Fatal App Hangs tracking. It tracks hangs from _start_ to _end_ or _cancellation_, persisting the fatal error information to be considered after app restart.

🚧 Sending RUM error and RUM view update will be added in following PR, to keep the code review scoped and lean.

### How?

Fatal App Hangs tracking has a lot of similarity with Crash tracking. Notably, the context of the fatal error must be persisted in the failed process so it is available after restart (that includes: _last RUM view_, _last RUM session info_, _current NTP offset_ and [more](https://github.com/DataDog/dd-sdk-ios/blob/a3a1d7f55aeb1b96b438a64b9451d2e8f610a750/DatadogCrashReporting/Sources/CrashContext/CrashContext.swift#L15)).

To embrace this similarity and enable later code reuse this PR introduces `FatalErrorContextNotifier` that talks to both `DatadogCrashReporting` (over message bus) and RUM's `AppHangsMonitor`. It is meant to be extended even more for later watchdog terminations tracking. This way all fatal errors (crashes, fatal hangs and OOMs) will utilize [the same logic](https://github.com/DataDog/dd-sdk-ios/blob/a3a1d7f55aeb1b96b438a64b9451d2e8f610a750/DatadogRUM/Sources/Integrations/CrashReportReceiver.swift#L152-L601) of uploading them to previous RUM session (covering all edge cases like fail in background vs BET enabled/disabled, fail with no RUM session started, restart after more than 4h of the issue etc.).

The overall architecture introduced in this PR can be illustrated as follows:

Screenshot 2024-03-28 at 11 10 37

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


DatadogRUM/Sources/Feature/RUMDataStore.swift
```diff
@@ -0,0 +1,77 @@
...
+            block(context, rumDataStore)
+        }
+    }
+}
+
+/// RUM interface for data store.
+///
+/// It stores values in JSON format and implements convenience for type-safe key referencing and data serialization.
+/// Serialization errors are logged to telemetry.
+internal struct RUMDataStore {
```
ncreated, line 28: 💡 This adds JSON-coding convenience for interacting with Data Store. It leverages SRP to separate concerns by abstracting error handling and failure reporting. It will be vastly leveraged for watchdog terminations, where the number of stored values will grow.

DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsMonitor.swift
```diff
@@ -0,0 +1,99 @@
...
+        /// The standardized `error.stack` when backtrace generation failed due to an internal error.
+        static let appHangStackGenerationFailedErrorMessage = "Failed to generate stack trace. This is a known issue and we work on it."
+    }
+
+    /// Watchdog thread that monitors the main queue for App Hangs.
+    private let watchdogThread: AppHangsObservingThread
+    /// Handles non-fatal App Hangs.
+    internal let nonFatalHangsHandler: NonFatalAppHangsHandler
+    /// Handles non-fatal App Hangs.
+    internal let fatalHangsHandler: FatalAppHangsHandler
```
ncreated, line 27: 💡 The previous non-fatal App Hang logic went to `NonFatalAppHangsHandler`. Fatals are implemented in separate handler. This is to separate concerns as both will evolve significantly different ways (with fatal handler reaching significantly higher complexity).

DatadogRUM/Sources/Instrumentation/AppHangs/FatalAppHangsHandler.swift
```diff
@@ -0,0 +1,78 @@
...
+
+            self?.send(fatalHang: fatalHang)
+        }
+    }
+
+    private func send(fatalHang: FatalAppHang) {
+        // TODO: RUM-3461
+        // Similar to how we send Crash report in `CrashReportReceiver`:
+        // - construct RUM error from `fatalHang.hang` information
+        // - update `error.count` in `fatalHang.lastRUMView`
```
ncreated, line 76: 💡 This will come in following PR. This PR is merged against feature branch.

DatadogRUM/Sources/Instrumentation/AppHangs/ProcessIdentifier.swift
```diff
@@ -0,0 +1,19 @@
...
+/// process but different for RUM instances after app restart.
+///
+/// Use this identifier to distinguish data collected between different process instances and between SDK instances:
+/// - Data collected in two processes will have different `processID`.
+/// - Data collected in two SDK instances within the same process will share the same `processID`.
+///
+/// Example use case in fatal App Hangs tracking:
+/// - SDK started → RUM enabled → [hang occurs] → pending App Hang saved → SDK stopped → SDK started again → RUM enabled again → pending App Hang loaded
+/// - When restarting RUM , the `processID` check ensures dropping pending hang from the previous instance, preventing false "fatal" hang detection.
+internal let currentProcessID = UUID()
```
ncreated, line 19: 💡 This is to achieve both:
- cover the edge case of pending hang being written before stopping the SDK and being considered a "fatal" when restarting it within the same process;
- add a safety net for limiting false-positive fatal hangs (by only reporting if their tracked `processID` is different from `currentProcessID`).

DatadogRUM/Sources/Feature/RUMDataStore.swift
maciejburda: I like that.
I did manual encoding/decoding in my PR using DataStore.

IMO worth adding similar kind of codable support extension as public. It's going to be very common use case.

cc @maxep

maciejburda (approved):
Gave it general overview. Looks good. Well commented and separated!

DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsWatchdogThread.swift
```diff
@@ -35,16 +56,27 @@ internal final class AppHangsWatchdogThread: Thread {
...
+    @ReadWriteLock
+    internal var onHangCancelled: ((AppHang) -> Void)?
+    /// Closure to be notified when App Hang ends. It passes the hang and its duration.
+    /// It is executed on the watchdog thread.
     @ReadWriteLock
-    internal var onHangEnded: ((AppHang) -> Void)?
+    internal var onHangEnded: ((AppHang, TimeInterval) -> Void)?
     /// A block called after this thread finished its pass and will become idle.
     @ReadWriteLock
     internal var onBeforeSleep: (() -> Void)?
```
maxep, line 82: /suggestion That makes 5 locks :/ The property wrapper might not fit very well here, unless we use a companion object with a single lock.

Another approach would be to create another read-write lock object that is not a property wrapper and that we can use to lock multiple accesses.

DatadogRUM/Sources/RUMMonitor/Scopes/FatalErrorContextNotifier.swift
```diff
@@ -0,0 +1,43 @@
...
+    @ReadWriteLock
+    var view: RUMViewEvent? {
+        didSet {
+            if let lastRUMView = view {
+                messageBus.send(message: .baggage(key: RUMBaggageKeys.viewEvent, value: lastRUMView))
+            } else {
+                messageBus.send(message: .baggage(key: RUMBaggageKeys.viewReset, value: true))
+            }
+        }
+    }
```
maxep, line 42: /suggestion We only need the setter here, if we only provide set methods we can remove the locks:
```swift
internal final class FatalErrorContextNotifier {
  func set(state: RUMSessionState)
  func set(view: RUMViewEvent?)
}
```

maxep (commented):
Looks great! I left some comments regarding locks, LMWYT!

DatadogRUM/Sources/RUMMonitor/Scopes/FatalErrorContextNotifier.swift
ncreated: We need a getter as well. The `fatalErrorContext.view` is read from `FatalAppHangsHandler`. The same will happen to `RUMSessionState` to consider edge-cases of uploading fatal error to previous RUM session:

Screenshot 2024-04-02 at 15 25 06

DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsWatchdogThread.swift
ncreated: There is only one write to each `onHang*` lock during SDK startup on the main thread. Later, SDK only reads them from the watchdog (background) thread.

I could combine these callbacks to `Observer` struct, e.g.:
```swift
thread.hangObserver = .init(
   onStarted: { _ in },
   onCancelled: { _ in }
   onEnded: { _, _ in }
)
```
Alternatively, we can leverage `delegate` model (`thread.delegate = self`). WDYT @maxep ?

DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsWatchdogThread.swift
```diff
@@ -7,7 +7,22 @@
 import Foundation
 import DatadogInternal
 
-internal final class AppHangsWatchdogThread: Thread {
+internal protocol AppHangsObservingThread: AnyObject {
```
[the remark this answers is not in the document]
ncreated: This is to enable mutability for callbacks:
```swift
var onHangStarted: ((AppHang) -> Void)? { set get }
var onHangCancelled: ((AppHang) -> Void)? { set get }
var onHangEnded: ((AppHang, TimeInterval) -> Void)? { set get }
var onBeforeSleep: (() -> Void)? { set get }
```

Alternatively, we can manage the thread reference as `var`, not `let`:
```diff
- private let watchdogThread: AppHangsObservingThread
+ private var watchdogThread: AppHangsObservingThread
```

DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsWatchdogThread.swift
maxep: The good old delegate pattern :) It is quite appropriate in this case 👍 Do you mind trying out?

DatadogRUM/Sources/RUMMonitor/Scopes/FatalErrorContextNotifier.swift
maxep: Ah sorry, I missed that part, nevermind!

DatadogRUM/Sources/Feature/RUMDataStore.swift
ncreated: 👍 Sounds good. Let's move on with `scope.rumDataStore` in `RUM`. I can follow with refactoring PR to introduce `scope.jsonDataStore` in `DatadogInternal`.

DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsWatchdogThread.swift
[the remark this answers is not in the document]
ncreated: `AnyObject` is no longer required after switch to delegate pattern.

DatadogRUM/Sources/Instrumentation/AppHangs/AppHangsWatchdogThread.swift
ncreated: Sure, I changed to delegate 👍. It is `weak var` [which implies "atomic"](https://github.com/apple/swift/pull/1454), so no locks required.

maxep (approved):
Looks great, well done!

PR #1750, opened 2024-03-26
RUM-3461 refactor: Replace `core` dependency in RUM with `FeatureScope`

ncreated:
### What and why?

📦 Utilizing the change from #1744, this PR updates the whole RUM to depend on `FeatureScope` interface rather than `DatadogCoreProtocol` directly.

This will be leveraged in fatal App Hangs monitoring, where `AppHangObserver` will take a dependency on `FeatureScope` to access DataStore, Event Write Context and Telemetry.

### How?

With `FeatureScope` being available during RUM feature registration (https://github.com/DataDog/dd-sdk-ios/pull/1744), it is now injected into all downstream components in RUM including the `Monitor` and all message receivers.

### Review checklist
- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes

### Custom CI job configuration (optional)
- [ ] Run unit tests for Core, RUM, Trace, Logs, CR and WVT
- [ ] Run unit tests for Session Replay
- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

PR #1757, opened 2024-04-02
RUM-2813 Add support for `internal_telemetry` from browser

maxep:
### What and why?

Following #1691, the telemetry events coming from Browser were no longer forwarded to RUM track.

### How?

Support `internal_telemetry` message type and forward them to RUM.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogRUM/Sources/Integrations/WebViewEventReceiver.swift
+++ b/DatadogRUM/Sources/Integrations/WebViewEventReceiver.swift
@@ -49,9 +49,9 @@ internal final class WebViewEventReceiver: FeatureMessageReceiver {
     func receive(message: FeatureMessage, from core: DatadogCoreProtocol) -> Bool {
         switch message {
         case let .webview(.rum(event)):
-            receive(rum: event, from: core)
+            receive(rum: event)
         case let .webview(.telemetry(event)):
-            receive(telemetry: event, from: core)
+            receive(telemetry: event)
         default:
             return false
         }
@@ -59,7 +59,7 @@ internal final class WebViewEventReceiver: FeatureMessageReceiver {
         return true
     }
 
-    private func receive(rum event: JSON, from core: DatadogCoreProtocol) {
+    private func receive(rum event: JSON) {
         commandSubscriber.process(
             command: RUMKeepSessionAliveCommand(
                 time: dateProvider.now,
@@ -115,9 +115,9 @@ internal final class WebViewEventReceiver: FeatureMessageReceiver {
         }
     }
 
-    private func receive(telemetry event: JSON, from core: DatadogCoreProtocol) {
+    private func receive(telemetry event: JSON) {
         // RUM-2866: Update with dedicated telemetry track
-        featureScope.eventWriteContext { [weak core] context, writer in
+        featureScope.eventWriteContext { [featureScope] context, writer in
             guard let rumBaggage = context.baggages[RUMFeature.name] else {
                 return // Drop event if RUM is not enabled or RUM session is not sampled
             }
@@ -142,7 +142,7 @@ internal final class WebViewEventReceiver: FeatureMessageReceiver {
 
                 writer.write(value: AnyEncodable(event))
             } catch {
-                core?.telemetry.error("Failed to decode `RUMCoreContext`", error: error)
+                featureScope.telemetry.error("Failed to decode `RUMCoreContext`", error: error)
             }
         }
     }
--- a/DatadogRUM/Sources/Integrations/WebViewEventReceiver.swift
+++ b/DatadogRUM/Sources/Integrations/WebViewEventReceiver.swift
@@ -39,23 +39,27 @@ internal final class WebViewEventReceiver: FeatureMessageReceiver {
         self.viewCache = viewCache
     }
 
-    func receive(message: FeatureMessage, from core: DatadogCoreProtocol) -> Bool {
-        guard case let .webview(.rum(event)) = message else {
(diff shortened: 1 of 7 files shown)


DatadogRUM/Sources/Integrations/WebViewEventReceiver.swift
```diff
@@ -39,23 +39,27 @@ internal final class WebViewEventReceiver: FeatureMessageReceiver {
...
+    /// - Parameters:
+    ///   - message: The message containing the Browser RUM event.
+    ///   - core: The core to write the event.
     func receive(message: FeatureMessage, from core: DatadogCoreProtocol) -> Bool {
-        guard case let .webview(.rum(event)) = message else {
+        switch message {
+        case let .webview(.rum(event)):
+            receive(rum: event, from: core)
+        case let .webview(.telemetry(event)):
+            receive(telemetry: event, from: core)
```
ncreated, line 54: obsolete/ Starting from https://github.com/DataDog/dd-sdk-ios/pull/1744 we want to avoid passing the `core` reference around if not necessary and that includes message receiver APIs. We want to [delete the `core` parameter long term](https://github.com/DataDog/dd-sdk-ios/blob/723dc0d61198fad2c627fc2579d24bf991b015ef/DatadogInternal/Sources/MessageBus/FeatureMessageReceiver.swift#L29-L32). To move us closer to the goal, the https://github.com/DataDog/dd-sdk-ios/pull/1750 has cleaned up `WebViewEventReceiver`, but now we bring it back in this change. With availability of `featureScope` in this receiver, the `core` reference is obsolete in this file. Let's stick to one convention and goal 🙏.

DatadogRUM/Sources/Integrations/WebViewEventReceiver.swift
```diff
@@ -111,6 +115,38 @@ internal final class WebViewEventReceiver: FeatureMessageReceiver {
...
+                }
+
+                if var session = event["session"] as? JSON {
+                    session["id"] = rum.sessionID
+                    event["session"] = session
+                }
+
+                writer.write(value: AnyEncodable(event))
+            } catch {
+                core?.telemetry.error("Failed to decode `RUMCoreContext`", error: error)
```
ncreated, line 145: change-request/ The telemetry endpoint is now available on `featureScope` and that can be even captured strongly to the closure (it manages weak reference to `core`):
```swift
featureScope.eventWriteContext { [featureScope] context, writer in
   // ...
   featureScope.telemetry.error("...")
}
```

DatadogWebViewTracking/Tests/MessageEmitterTests.swift
```diff
@@ -91,7 +91,37 @@ class MessageEmitterTests: XCTestCase {
...
+        XCTAssertEqual(try json.value("attribute1"), 123)
+        XCTAssertEqual(try json.value("attribute2"), "foo")
+        XCTAssertEqual(try json.array("attribute3").values(), ["foo", "bar", "bizz"])
+    }
+
+    func testWhenReceivingTelemetryEvent_itForwardsToTelemetry() throws {
+        // Given
+        let receiverMock = FeatureMessageReceiverMock()
+        let core = PassthroughCoreMock(messageReceiver: receiverMock)
+        let emitter = MessageEmitter(logsSampler: .mockRandom(), core: core)
```
ncreated, line 107: cleanup/ To achieve mid-term goals (no managing `core` reference), could we refactor `MessageEmitter` to depend on feature scope instead? It looks trivial and it would clean up the implementation and tests.
```diff
-        let receiverMock = FeatureMessageReceiverMock()
-        let core = PassthroughCoreMock(messageReceiver: receiverMock)
-        let emitter = MessageEmitter(logsSampler: .mockRandom(), core: core)
+        let featureScope = FeatureScopeMock() 
+        let emitter = MessageEmitter(logsSampler: .mockRandom(), featureScope: featureScope)

-        receiverMock.messages.firstWebViewMessage
+        featureScope.messagesSent().firstWebViewMessage
```
The idea is: `MessageEmitter` doesn't need to depend on the entire `core` in order to send messages. Same applies to tests - we don't need to mock the entire `core` in order to test if emitter sends expected message.

DatadogRUM/Sources/Integrations/WebViewEventReceiver.swift
maxep: Sure, it was brought back while rebasing on develop, let me change that 👍

DatadogWebViewTracking/Tests/MessageEmitterTests.swift
maxep: WebViewTracking does not register any feature, it is enabled per webview instance and only sends messages on the bus. It doesn't need the full scope machinery, with data-store and such.. I start to think that we would need a bit a interface segregation instead of having everything in a feature scope 🤔 

Lets have a further discussion on that and revisit with another PR, how does it sound?

ncreated: I see, indeed. Well, we have [`MessageSending`](https://github.com/DataDog/dd-sdk-ios/blob/723dc0d61198fad2c627fc2579d24bf991b015ef/DatadogInternal/Sources/DatadogCoreProtocol.swift#L57) in place - if mocked, it can make these tests lean.

ncreated (approved):

Issue #1638, opened 2024-01-18
Crashes in  `NetworkInstrumentationFeature.extractTrace` and ` NetworkInstrumentationFeature.intercept`


JaviSoto:
### The crash

We recently set up distributed network tracing, and we're hitting these crashes in production, although we're not able to reproduce, so they don't happen 100% of the time. Here's a few different stack traces, it's not always the same crash, but they happen in the same codepath:

<details><summary>Stack traces</summary>
<p>

```
OS Version: iOS 17.2 (21C62)
Report Version: 104

Exception Type: EXC_BREAKPOINT (SIGTRAP)
Crashed Thread: 10

Application Specific Information:
allHTTPHeaderFields

Thread 10 Crashed:
0   CoreFoundation                  0x31fe4ac54         CF_IS_OBJC
1   CoreFoundation                  0x31fe694e4         [inlined] CFDictionaryGetCount
2   CoreFoundation                  0x31fe694e4         CFDictionaryGetCount
3   CFNetwork                       0x3221fb2b8         _CFHTTPServerResponseEnqueue
4   CFNetwork                       0x3221e2be0         CFHTTPCookieStorageUnscheduleFromRunLoop
5   Foundation                      0x31e1a2718         URLRequest.allHTTPHeaderFields.getter
6   <AppName>                         0x104ec81c4         NetworkInstrumentationFeature.extractTrace (NetworkInstrumentationFeature.swift:246)
7   <AppName>                         0x104ec6ba4         [inlined] NetworkInstrumentationFeature.extractTrace
8   <AppName>                         0x104ec6ba4         NetworkInstrumentationFeature.intercept (NetworkInstrumentationFeature.swift:169)
9   <AppName>                         0x104ec6dbc         thunk for closure
10  libdispatch.dylib               0x32fe246a4         _dispatch_call_block_and_release
11  libdispatch.dylib               0x32fe262fc         _dispatch_client_callout
12  libdispatch.dylib               0x32fe2d890         _dispatch_lane_serial_drain
13  libdispatch.dylib               0x32fe2e3c0         _dispatch_lane_invoke
14  libdispatch.dylib               0x32fe39000         _dispatch_root_queue_drain_deferred_wlh
15  libdispatch.dylib               0x32fe38874         _dispatch_workloop_worker_thread
16  libsystem_pthread.dylib         0x3f3c2d960         _pthread_wqthread
```

```
OS Version: iOS 17.2.1 (21C66)
[...]

PR #1767, opened 2024-04-08
RUM-2916 fix: Crash in `NetworkInstrumentationFeature`

ncreated:
### What and why?

📦 This PR addresses the crashes reported in #1638 by implementing extra thread safety measures.

```
Thread 10 Crashed:
0   CoreFoundation                  0x31fe4ac54         CF_IS_OBJC
1   CoreFoundation                  0x31fe694e4         [inlined] CFDictionaryGetCount
2   CoreFoundation                  0x31fe694e4         CFDictionaryGetCount
3   CFNetwork                       0x3221fb2b8         _CFHTTPServerResponseEnqueue
4   CFNetwork                       0x3221e2be0         CFHTTPCookieStorageUnscheduleFromRunLoop
5   Foundation                      0x31e1a2718         URLRequest.allHTTPHeaderFields.getter
6   <AppName>                         0x104ec81c4         NetworkInstrumentationFeature.extractTrace (NetworkInstrumentationFeature.swift:246)
7   <AppName>                         0x104ec6ba4         [inlined] NetworkInstrumentationFeature.extractTrace
8   <AppName>                         0x104ec6ba4         NetworkInstrumentationFeature.intercept (NetworkInstrumentationFeature.swift:169)
9   <AppName>                         0x104ec6dbc         thunk for closure
```

I couldn't reproduce the issue, so the fix is based on the following considerations:

- There might be a thread safety issue when accessing `task.currentRequest` from a private queue.
- There could be a data race when reading `request.allHTTPHeaderFields` from a private queue, possibly due to underlying `CFDictionary` operations.
- It's possible that a combination of the above scenarios is occurring, as suggested by multiple different stack traces reported in #1638.

### How?

Two main changes to resolve the issue:
- `let currentRequest = task.currentRequest` has been moved from the private queue to the `URLSession` thread to ensure safe access.
- A new type, `ImmutableRequest`, has been introduced. This type captures an immutable copy of the necessary fields from `URLRequest`, thereby avoiding direct referencing of underlying Foundation objects from the private queue.

### Reproducibility

While attempting to reproduce the problem in our codebase, I encountered several key findings:

- Modifying `URLRequest` from multiple concurrent threads causes thread safety issues. This arises from `URLRequest` being a Swift struct backed by an underlying Foundation object, as it conforms to [ReferenceConvertible](https://developer.apple.com/documentation/foundation/referenceconvertible):
```swift
// 🔥 Thread 12: EXC_BAD_ACCESS
// #0 ___lldb_unnamed_symbol6147 ()
// #1 ___lldb_unnamed_symbol6175 ()
// #2 CFURLRequestSetHTTPHeaderFieldValue ()
// #3 specialized URLRequest._applyMutation<τ_0_0>(_:) ()
// #4 closure #1 in CrashInNetworkFeatureExperiment.testRequestMutation() at CrashInNetworkFeatureExperiment.swift:28
// #5 partial apply for closure #1 in CrashInNetworkFeatureExperiment.testRequestMutation() ()
// #6 partial apply for thunk for @callee_guaranteed (@unowned Int) -> () ()
// #7 thunk for @escaping @callee_guaranteed (@unowned Int) -> () ()
// #8 __wrap_dispatch_apply_block_invoke ()
func testRequestMutation() {
    var request = URLRequest(url: URL(string: "https://example.com/")!)

    DispatchQueue.concurrentPerform(iterations: 10_000) { idx in
        request.setValue("header-\(idx)", forHTTPHeaderField: "field") // #28
        _ = request.allHTTPHeaderFields?["field"]
    }
}
``` 

- It appears safe [to modify](https://github.com/DataDog/dd-sdk-ios/blob/5334eab9cb5f9de2f47f62d9a951b3b05230938e/DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTask%2BTracking.swift#L19) `task.currentRequest` between threads, even if the request undergoes multiple redirections.
```swift
// ✅ no crash
func testTaskMutation() {
    let request = URLRequest(url: URL(string: "https://httpbin.org/redirect/10")!)
    let task = session.dataTask(with: request)
    task.resume()

    DispatchQueue.concurrentPerform(iterations: 10_000) { idx in
        let otherRequest = URLRequest(url: URL(string: "https://httpbin.org/redirect/\(idx)")!)
        task.setValue(otherRequest, forKey: "currentRequest")
        _ = task.currentRequest
    }
}
```

- Isolating our `NetworkInstrumentationFeature` code and subjecting it to stress testing with simultaneous thread access did not reveal any issues.
```swift
// ✅ no crash
func testNetworkInstrumentationFeature() {
    let swizzler = NetworkInstrumentationSwizzler()

    try! swizzler.swizzle(
        interceptResume: { task in
            var modifiedRequest = URLRequest(url: URL(string: "https://httpbin.org/redirect/15")!)
            modifiedRequest.setValue("header-value", forHTTPHeaderField: "header-field")
            task.dd.override(currentRequest: modifiedRequest)
        }
    )

    let request = URLRequest(url: URL(string: "https://httpbin.org/redirect/10")!)
    let task = session.dataTask(with: request)
    task.resume()

    DispatchQueue.concurrentPerform(iterations: 10_000) { idx in
        _ = task.currentRequest?.allHTTPHeaderFields
        _ = task.currentRequest?.allHTTPHeaderFields?["header-field"]
    }
}
```

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
+++ b/DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
@@ -153,11 +153,16 @@ extension NetworkInstrumentationFeature {
     ///   - task: The created task.
     ///   - additionalFirstPartyHosts: Extra hosts to consider in the interception.
     func intercept(task: URLSessionTask, additionalFirstPartyHosts: FirstPartyHosts?) {
+        guard let currentRequest = task.currentRequest else {
+            return
+        }
+
         // Get the current trace context from all handlers.
         let traceContexts = handlers.compactMap { $0.traceContext() }
+        let request = ImmutableRequest(request: currentRequest)
 
         queue.async { [weak self] in
-            guard let self = self, let request = task.currentRequest else {
+            guard let self = self else {
                 return
             }
 
@@ -187,7 +192,7 @@ extension NetworkInstrumentationFeature {
                 ))
             }
 
-            if let origin = request.value(forHTTPHeaderField: TracingHTTPHeaders.originField) {
+            if let origin = request.allHTTPHeaderFields?[TracingHTTPHeaders.originField] {
                 interception.register(origin: origin)
             }
 
@@ -259,7 +264,7 @@ extension NetworkInstrumentationFeature {
         interceptions[task] = nil
     }
 
-    private func extractTrace(firstPartyHosts: FirstPartyHosts, request: URLRequest) -> (traceID: TraceID, spanID: SpanID, parentSpanID: SpanID?)? {
+    private func extractTrace(firstPartyHosts: FirstPartyHosts, request: ImmutableRequest) -> (traceID: TraceID, spanID: SpanID, parentSpanID: SpanID?)? {
         guard let headers = request.allHTTPHeaderFields else {
             return nil
         }
--- a/DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
+++ b/DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
@@ -153,13 +153,16 @@ extension NetworkInstrumentationFeature {
     ///   - task: The created task.
     ///   - additionalFirstPartyHosts: Extra hosts to consider in the interception.
     func intercept(task: URLSessionTask, additionalFirstPartyHosts: FirstPartyHosts?) {
+        // In response to https://github.com/DataDog/dd-sdk-ios/issues/1638 capture the current request object on the
+        // caller thread and freeze its attributes through `ImmutableRequest`. This is to avoid changing the request
+        // object from multiple threads:
         guard let currentRequest = task.currentRequest else {
             return
         }
+        let request = ImmutableRequest(request: currentRequest)
 
         // Get the current trace context from all handlers.
         let traceContexts = handlers.compactMap { $0.traceContext() }
-        let request = ImmutableRequest(request: currentRequest)
 
         queue.async { [weak self] in
             guard let self = self else {
--- a/DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskInterception.swift
+++ b/DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskInterception.swift
@@ -11,7 +11,7 @@ public class URLSessionTaskInterception {
     public let identifier: UUID
     /// The initial request send during this interception. It is, the request send from `URLSession`, not the one
     /// given by the user (as the request could have been modified in `URLSessionSwizzler`).
-    public private(set) var request: URLRequest
+    public private(set) var request: ImmutableRequest
     /// Tells if the `request` is send to a 1st party host.
     public let isFirstPartyRequest: Bool
     /// Task metrics collected during this interception.
@@ -28,7 +28,7 @@ public class URLSessionTaskInterception {
     /// Setting the value to 'rum' will indicate that the span is reported as a RUM Resource.
     public private(set) var origin: String?
 
-    init(request: URLRequest, isFirstParty: Bool) {
+    init(request: ImmutableRequest, isFirstParty: Bool) {
         self.identifier = UUID()
         self.request = request
         self.isFirstPartyRequest = isFirstParty
@@ -46,7 +46,7 @@ public class URLSessionTaskInterception {
         }
     }
 
-    func register(request: URLRequest) {
+    func register(request: ImmutableRequest) {
         self.request = request
     }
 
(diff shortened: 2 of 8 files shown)


DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskInterception.swift
```diff
@@ -97,6 +97,30 @@ public struct ResourceCompletion {
     }
 }
 
+/// An immutable version of `URLRequest`.
+///
+/// Introduced in response to concerns raised in https://github.com/DataDog/dd-sdk-ios/issues/1638 
+/// it makes a copy of request attributes, safeguarding against potential thread safety issues arising from concurrent 
+/// mutations (see more context in https://github.com/DataDog/dd-sdk-ios/pull/1767 ).
+public struct ImmutableRequest {
```
ganeshnj, line 105: does this have to be public?

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskInterception.swift
```diff
@@ -97,6 +97,30 @@ public struct ResourceCompletion {
...
+    /// The URL of the request.
+    public let url: URL?
+    /// The HTTP method of the request.
+    public let httpMethod: String?
+    /// The HTTP header fields of the request.
+    public let allHTTPHeaderFields: [String: String]?
+    /// A reference to the original `URLRequest` object provided during initialization. Direct use is discouraged
+    /// due to thread safety concerns. Instead, necessary attributes should be accessed through `ImmutableRequest` fields.
+    public let unsafeOriginal: URLRequest
```
ganeshnj, line 114: we shouldn't keep a copy here else we would end up having the same issue again unknowingly.

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
```diff
@@ -153,11 +153,19 @@ extension NetworkInstrumentationFeature {
     ///   - task: The created task.
     ///   - additionalFirstPartyHosts: Extra hosts to consider in the interception.
     func intercept(task: URLSessionTask, additionalFirstPartyHosts: FirstPartyHosts?) {
+        // In response to https://github.com/DataDog/dd-sdk-ios/issues/1638 capture the current request object on the
+        // caller thread and freeze its attributes through `ImmutableRequest`. This is to avoid changing the request
+        // object from multiple threads:
+        guard let currentRequest = task.currentRequest else {
+            return
+        }
+        let request = ImmutableRequest(request: currentRequest)
```
ganeshnj, line 162: good idea of keeping the ImmutableRequest init outside the queue.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskInterception.swift
ncreated: It has `public` requirement [in `URLSessionTaskInterception`](https://github.com/DataDog/dd-sdk-ios/blob/5334eab9cb5f9de2f47f62d9a951b3b05230938e/DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskInterception.swift#L12-L14) which is later leveraged in:
- `DatadogRUM`: [example](https://github.com/DataDog/dd-sdk-ios/blob/5334eab9cb5f9de2f47f62d9a951b3b05230938e/DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift#L74)
- `DatadogTrace`: [example](https://github.com/DataDog/dd-sdk-ios/blob/5334eab9cb5f9de2f47f62d9a951b3b05230938e/DatadogTrace/Sources/Integrations/TracingURLSessionHandler.swift#L132)

Unless we want to invest into bigger refactoring, I think there is no better way around. Keep in mind this is public in `DatadogInternal`, meaning not user-facing.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskInterception.swift
ncreated: Agree 👍, however we provide `URLRequest` in RUM Attributes Provider API, which requires passing the original request [here](https://github.com/DataDog/dd-sdk-ios/blob/5334eab9cb5f9de2f47f62d9a951b3b05230938e/DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift#L102) in `DatadogRUM`. I share the same concern - we may overlook it and add a new usage somewhere else. To mitigate this risk, I named this property as `unsafeOriginal` so it should raise attention. As an extra layer, we could add linter rule to prevent more usage, but even this can be muted.

ganeshnj: I think we have made interesting choices with `rumAttributesProvider` API and now we are hand locked. Thanks for background.

maxep (approved):
Looks good, tho I'm a bit concerned we are not able to reproduce the crash :/ lets keep an eye on it!

Limwin94:
https://github.com/DataDog/dd-sdk-ios/issues/1638#issuecomment-2058111010

PR #1768, opened 2024-04-09
feat: Support adding binary images to non-crashes

fuzzybinary:
### What and why?

This adds a cross platform attribute `_dd.error.include_binary_images` that will attach binary images to error reports by creating a backtrace and pulling the binary images from it.

It contains a TODO to replace backtrace generation with a different method of binary image attachment at a future time.

### Review checklist

- [x] Run integration tests
- [x] Run smoke tests
- [ ] Run tests for `tools/`


DatadogRUM/Sources/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -559,6 +559,15 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
         var commandAttributes = command.attributes
         let errorFingerprint = commandAttributes.removeValue(forKey: RUM.Attributes.errorFingerprint) as? String
 
+        var binaryImages = command.binaryImages?.compactMap { $0.toRUMDataFormat }
+        if let addBinaryImages = commandAttributes.removeValue(forKey: CrossPlatformAttributes.includeBinaryImages) {
+            // Don't try to get binary images if we already have them.
+            if binaryImages == nil {
+                // TODO: RUM-000 Replace full backtrace reporter with simpler binary image fetcher
```
ncreated, line 566: That would make sense 👍. Also it could abstract the caching problem.

fuzzybinary:
Marking this as ready for review, but we can talk about if we want to do RUM-4072 before actually merging this in.

DatadogLogs/Sources/Log/LogEventEncoder.swift
```diff
@@ -75,6 +75,8 @@ public struct LogEvent: Encodable {
         public var sourceType: String = "ios"
         /// The custom fingerprint supplied for this error, if any
         public var fingerprint: String?
+        /// Binary images needed to decode the provided stack (if any)
+        public var binaryImages: [BinaryImage]?
```
ncreated, line 79: change-request/ There are two problems with this code:
- Separation of concerns: We encode `DatadogInternal.BinaryImage` directly into log event sent to EvP. Because `BinaryImage` is considered an internal code, it can be changed at any time. An unaware change contributor may not even notice that altering it will break the contract with logs intake.
- The property is marked mutable, hence user may want to interact with and redact the `BinaryImage` through log event mapper. However, without [re-exporting the `DatadogInternal.BinaryImage` from `DatadogLogs`](https://github.com/DataDog/dd-sdk-ios/blob/b63af60df8c6417866b5acd70689d8622962641c/DatadogRUM/Sources/RUMConfiguration.swift#L12) the compiler and / or autocompletion may be lost with dynamic lookup.

To address both, we should rather introduce a new type (`LogEvent.BinaryImage`?) that will mirror the `DatadogInternal.BinaryImage`. This way we achieve both: this model will prevent from accidentally breaking the contract due to the internal change and it will serve as first-class-citizen in `DatadogLogs` public interface.

Note, that this is exactly what we do with `BinaryImages` in RUM. They're [mapped here](https://github.com/DataDog/dd-sdk-ios/blob/b63af60df8c6417866b5acd70689d8622962641c/DatadogRUM/Sources/DataModels/RUMDataModelsMapping.swift#L98-L113).

DatadogLogs/Sources/RemoteLogger.swift
```diff
@@ -178,14 +189,21 @@ internal final class RemoteLogger: LoggerProtocol {
...
                         key: ErrorMessage.key,
                         value: ErrorMessage(
                             message: log.error?.message ?? log.message,
                             type: log.error?.kind,
                             stack: log.error?.stack,
-                            attributes: .init(combinedAttributes)
+                            attributes: .init(rumAttributes)
                         )
                     )
                 )
```
ncreated, line 209: suggestion/ I think there is an easier way. We could add `binaryImages` to the `ErrorMessage` that is sent over the bus. Note that [`BinaryImage` conforms](https://github.com/DataDog/dd-sdk-ios/blob/b63af60df8c6417866b5acd70689d8622962641c/DatadogInternal/Sources/Models/CrashReporting/BinaryImage.swift#L10) to [`PassthroughAnyCodable`](https://github.com/DataDog/dd-sdk-ios/blob/b63af60df8c6417866b5acd70689d8622962641c/DatadogInternal/Sources/Codable/AnyEncoder.swift#L664-L673) meaning that it is transmitted without actual coding (no perf hit).

Upon [receiving the new `ErrorMessage` in RUM](https://github.com/DataDog/dd-sdk-ios/blob/b63af60df8c6417866b5acd70689d8622962641c/DatadogRUM/Sources/Integrations/ErrorMessageReceiver.swift#L41) it is passed to internal [`Monitor.addError()`](https://github.com/DataDog/dd-sdk-ios/blob/b63af60df8c6417866b5acd70689d8622962641c/DatadogRUM/Sources/RUMMonitor/Monitor.swift#L533) which emits `RUMAddCurrentViewErrorCommand`. This command has a [broad initializer that accepts `binaryImages`](https://github.com/DataDog/dd-sdk-ios/blob/b63af60df8c6417866b5acd70689d8622962641c/DatadogRUM/Sources/RUMMonitor/RUMCommand.swift#L179-L191), so we should be fine 🔗 👍.

ncreated (changes requested):
👍 It's on track. I left one change request and potential simple solution to doubling BI generation for log errors.

DatadogLogs/Sources/RemoteLogger.swift
fuzzybinary: So I remembered the reason I didn't do this, though it's potentially trivial to fix. `ErrorMessageReceiver` uses the [public `Monitor.addError` method](https://github.com/DataDog/dd-sdk-ios/blob/258d9ff2e144f393587e031d2a2719e6826b82c7/DatadogRUM/Sources/Integrations/ErrorMessageReceiver.swift#L41) to add the error to the monitor, which we obviously do not want to modify to accept binary images.

The answer is fairly straightforward, but adds a new method to the internal interface on Monitor just for this.

PR #1773, opened 2024-04-11
RUM-3853 feat(otel-tracer): remove support for event API

ganeshnj:
### What and why?

`event` is not supported. 

### How?

- Make event API not supported
- Test case for logs integration

### Review checklist

- [x] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogCore/Tests/Datadog/LoggerTests.swift
+++ b/DatadogCore/Tests/Datadog/LoggerTests.swift
@@ -7,6 +7,7 @@
 import XCTest
 import TestUtilities
 import DatadogInternal
+import OpenTelemetryApi
 
 @testable import DatadogLogs
 @testable import DatadogTrace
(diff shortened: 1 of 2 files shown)


DatadogCore/Tests/Datadog/LoggerTests.swift
```diff
@@ -875,6 +876,45 @@ class LoggerTests: XCTestCase {
...
+            .tracerProvider
+            .get(instrumentationName: "", instrumentationVersion: nil)
+
+        // when
+        let span = tracer.spanBuilder(spanName: "span")
+            .setActive(true)
+            .startSpan()
+        logger.info("info message 1")
+        span.end()
+        logger.info("info message 2")
```
ncreated, line 902: food for thought/ One thing that bothers me is that in V2 the "active span" [we store in `DatadogContext`](https://github.com/DataDog/dd-sdk-ios/blob/eb5515c6bf28f3cd6faa15719a970f39c46183d0/DatadogTrace/Sources/DatadogTracer.swift#L134) is actually the "last span marked active in `DatadogTrace`" not the "active span in current thread". To illustrate this, following change will keep the test green although it shouldn't if we stick to the definition of "active span" per OT or OTel.

Not saying there's something to do in this PR, just bringing this problem up as we're stepping over it. I wonder what's your take on this @ganeshnj.

```swift
let queue = DispatchQueue(label: "foo")
var span: Span!

queue.sync {
    span = tracer.spanBuilder(spanName: "span")
        .setActive(true)
        .startSpan()
}
logger.info("info message 1")
queue.sync {
    span.end()
}
logger.info("info message 2")
```

DatadogTrace/Sources/OpenTelemetry/OTelSpan.swift
```diff
@@ -136,42 +136,20 @@ internal class OTelSpan: OpenTelemetryApi.Span {
         )
     }
 
-    /// Sends a span event which is akin to a log in Datadog
-    /// - Parameter name: name of the event
     func addEvent(name: String) {
-        addEvent(name: name, timestamp: .init())
+        // not supported
```
ncreated, line 140: suggestion/ We may consider warning the user of depending on not supported API:
```swift
DD.logger.warn("\(#function) is not yet supported in `DatadogTrace`")
```

ncreated (approved):
Looks good 👌 🎯

DatadogCore/Tests/Datadog/LoggerTests.swift
ganeshnj: agree, current version is very naive. This will be more troublesome with swift-concurrency (actors specifically).

I have this back of my mind, and hopefully will follow up once otel topic is out.

DatadogTrace/Sources/OpenTelemetry/OTelSpan.swift
ganeshnj: I was hesitant that it can be too verbose but glad you mentioned.

Updated.

PR #1777, opened 2024-04-15
RUM-3185 feat(otel-tracer): use DatadogSDKTesting as SPM dependency

ganeshnj:
### What and why?

DatadogSDKTesting is not enabled on otel-tracer branch.

### How?

There is a history why? It is mainly because `OpenTelemetryApi` package is imported by both `DatadogTrace` and `DatadogSDKTesting`. Linker is not intelligent enough to decide which one to use and it randomly select one. This random from the experience is frameworks ordered by name and then whichever comes first.

I had success with two solutions.

1. Use `OpenTelemetryApi` and `DatadogSDKTesting` as SPM. But we have big limitation, lack of iOS 12 support https://github.com/open-telemetry/opentelemetry-swift/issues/521
2. Use `OpenTelemetryApi` as xcframework and `DatadogSDKTesting` as SPM which worked as well and hence this PR.

image

One thing to notice here, now the `DatadogSDKTesting` SDK will be added to the workspace. It is no longer dynamically injected which is good and bad. Good, we can now bump the version and maintain better. Bad, it has to be downloaded in our development cycle.

### Review checklist

- [x] Run integration tests
- [x] Run smoke tests
- [x] Run tests for `tools/`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,17 +1,6 @@
 all: dependencies templates
 
-# The release version of `dd-sdk-swift-testing` to use for tests instrumentation.
-DD_SDK_SWIFT_TESTING_VERSION = 2.3.2
-DD_DISABLE_TEST_INSTRUMENTING = false
-
 define DD_SDK_TESTING_XCCONFIG_CI
-DD_SDK_TESTING_PATH=$$(DD_SDK_TESTING_OVERRIDE_PATH:default=$$(SRCROOT)/../instrumented-tests/)\n
-FRAMEWORK_SEARCH_PATHS[sdk=iphonesimulator*]=$$(inherited) $$(DD_SDK_TESTING_PATH)/DatadogSDKTesting.xcframework/ios-arm64_x86_64-simulator/\n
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -2,16 +2,8 @@ all: dependencies templates
 
 # The release version of `dd-sdk-swift-testing` to use for tests instrumentation.
 DD_SDK_SWIFT_TESTING_VERSION = 2.3.2
```
ncreated, line 4: This can be removed as well if we no longer fetch this dependency manually.

ncreated (commented):
question/ Except the need of fetching and linking the `DatadogSDKTesting` dependency, is this change fully transparent to local development? I wonder if local test runs won't be instrumented or even reported to DD by default? Or will it depend on the presence and configuration in `DatadogSDKTesting.local.xcconfig`?

ganeshnj:
> question/ Except the need of fetching and linking the `DatadogSDKTesting` dependency, is this change fully transparent to local development? I wonder if local test runs won't be instrumented or even reported to DD by default? Or will it depend on the presence and configuration in `DatadogSDKTesting.local.xcconfig`?

I verified it depends on the existence of `DatadogSDKTesting.local.xcconfig` hence local tests are not instrumented unless you wish to.

Makefile
ganeshnj: true.

ncreated (approved):

Issue #1549, opened 2023-11-09
Setting the log verbosity level produces a strict concurrency warning


jpsim:
Setting the log verbosity level produces a strict concurrency warning when compiling with `SWIFT_STRICT_CONCURRENCY = complete` or `-strict-concurrency=complete` produces the following warning:

```swift
Datadog.verbosityLevel = .debug
// 👆 Reference to static property 'verbosityLevel' is not concurrency-safe because it involves shared mutable state
```

Ideally Datadog's iOS SDKs would be audited for strict concurrency warnings so that consumers can enable these strict flags if they choose to do so without introducing warnings.

jpsim:
I can file a separate issue for this, but it would also be nice if `LoggerProtocol` was `Sendable`.

ncreated:
Hello @jpsim 👋. This request seems to overlap with https://github.com/DataDog/dd-sdk-ios/issues/1440.

> Ideally Datadog's iOS SDKs would be audited for strict concurrency warnings so that consumers can enable these strict flags if they choose to do so without introducing warnings.

This is something we already track in our backlog. I will bump up the priority with this new ask.

farshadtx:
Any update here?

ncreated:
@farshadtx Unfortunately no, this topic is still in our backlog.

jaredsinclair:
@ncreated Howdy, I would like to know, when you write `...is still in our backlog...` did that mean it's being actively worked on? Do ya'll have a target date by which you plan to be ready for Swift 6 and strict concurrency?

jaredsinclair:
I've opened a PR from my fork of the repo that addresses the defect using an existing synchronization within the SDK:

#1778

ncreated:
Thanks a lot @jaredsinclair 👍. I just merged the PR, so it will be available in the next release (likely `2.11.0`).

PR #1778, opened 2024-04-15
[#1549] Address Swift 6 concurrency warning by synchronizing access to verbosityLevel.

jaredsinclair:
### What and why?

When building a host application target that imports the Datadog iOS SDK and has `complete` concurrency checking enabled on the application target, `Datadog.verbosityLevel` emits the following warning:

> Static property 'verbosityLevel' is not concurrency-safe because it is non-isolated global shared mutable state; this is an error in Swift 6

This PR addresses the defect using an existing synchronization utility within _DatadogInternal_.

### How?

By downgrading `Datadog.verbosityLevel` to a computed property, backed by a private property using an instance of `ReadWriteStorage<Value>` to ensure efficient synchronization, the concurrency checker warning is fully resolved.

The warning that the concurrency checker emits, according to the setup described above, is a valid one. It's an example of the tooling doing exactly what it was designed to address, among other types of violations: shared mutable state not protected via explicit synchronization or actor isolation. Since this property has been and must continue to be readable and writable from any thread, and without regard to actor isolation, the most appropriate change is to wrap access to the property in a synchronization mechanism.

The ReadWriteLock class is the right tool for the job. It was missing only one key ingredient: an explicit declaration that it conforms to Sendable. It cannot directly conform to Sendable because it contains a `private var` property, mutable state. But the maintainers of the Datadog SDK know that it was designed to be Sendable before Sendable was a term, through the use of a POSIX reader/writer lock. Therefore it can be reasonably declared `@unchecked Sendable`, satisfying the concurrency checker through years of peer review and production exercise.

### Review checklist
- [x] Feature or bugfix MUST have appropriate tests (unit, integration) *Existing tests that exercise verbosityLevel should suffice.*
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference #1549 

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogCore/Sources/Datadog.swift
+++ b/DatadogCore/Sources/Datadog.swift
@@ -231,7 +231,14 @@ public enum Datadog {
     /// Verbosity level of Datadog SDK. Can be used for debugging purposes.
     /// If set, internal events occuring inside SDK will be printed to debugger console if their level is equal or greater than `verbosityLevel`.
     /// Default is `nil`.
-    public static var verbosityLevel: CoreLoggerLevel? = nil
+    public static var verbosityLevel: CoreLoggerLevel? {
+        get { _verbosityLevel.wrappedValue }
+        set { _verbosityLevel.wrappedValue = newValue }
+    }
+
+    /// The backing storage for `verbosityLevel`, ensuring efficient synchronized
+    /// read/write access to the shared value.
+    private static let _verbosityLevel = ReadWriteLock<CoreLoggerLevel?>(wrappedValue: nil)
 
     /// Returns `true` if the Datadog SDK is already initialized, `false` otherwise.
     ///
--- a/DatadogInternal/Sources/Concurrency/ReadWriteLock.swift
+++ b/DatadogInternal/Sources/Concurrency/ReadWriteLock.swift
@@ -13,7 +13,7 @@ import Foundation
 /// An additional method `mutate` allow to safely mutate the value in-place (to read it
 /// and write it while obtaining the lock only once).
 @propertyWrapper
-public final class ReadWriteLock<Value> {
+public final class ReadWriteLock<Value>: @unchecked Sendable {
     /// The wrapped value.
     private var value: Value
 


DatadogCore/Sources/Datadog.swift
```diff
@@ -231,7 +231,14 @@ public enum Datadog {
...
     /// Default is `nil`.
-    public static var verbosityLevel: CoreLoggerLevel? = nil
+    public static var verbosityLevel: CoreLoggerLevel? {
+        get { _verbosityLevel.wrappedValue }
+        set { _verbosityLevel.wrappedValue = newValue }
+    }
+
+    /// The backing storage for `verbosityLevel`, ensuring efficient synchronized
+    /// read/write access to the shared value.
+    private static let _verbosityLevel = ReadWriteLock<CoreLoggerLevel?>(wrappedValue: nil)
```
ganeshnj, line 241:
```suggestion
    @ReadWriteLock
    public static var verbosityLevel: CoreLoggerLevel? = nil
```

I wonder if you really need a backing property here.

ganeshnj (changes requested):

DatadogCore/Sources/Datadog.swift
jaredsinclair: Unfortunately, property wrappers are unusable here; the Swift 6 / `complete` concurrency checker sees this as mutable state:

Screenshot 2024-04-17 at 7 19 21 AM

These two requirements are at odds:

- Property wrappers must be defined as `var` not `let`.
- The concurrency checker disallows the use of `var` because it's shared mutable state

Something has to give. Until the day that Swift gains the ability to overlook a `var` wrapped in a sendable property wrapper, it is not possible to use a property wrapper to resolve concurrency issues with shared mutable state.

ganeshnj (approved):

ganeshnj:
@jaredsinclair  I just kicked off the build.

jaredsinclair:
@ganeshnj What does this mean? It looks like all the steps passed except...whatever this top one represents:

Screenshot 2024-04-22 at 9 59 30 AM

ganeshnj:
This is just false positive, I verified, everything looks fine. I think bitrise can't handle the PR created from forks correctly. This PR is good to go and we will take care of it. Someone from the team will merge and release in the next release.

PR #1782, opened 2024-04-16
RUM-885 Session Replay Configuration Telemetry

maxep:
### What and why?

Report configuration telemetry when Session Replay is enabled.

### How?

Set `session_replay_sample_rate` and `default_privacy_level` in configuration telemetry.
This PR also removes deprecated configuration properties.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


DatadogInternal/Sources/Telemetry/Telemetry.swift
```diff
@@ -253,10 +248,8 @@ extension Telemetry {
         forwardErrorsToLogs: Bool? = nil,
         initializationType: String? = nil,
         mobileVitalsUpdatePeriod: Int64? = nil,
-        premiumSampleRate: Int64? = nil,
         reactNativeVersion: String? = nil,
         reactVersion: String? = nil,
-        replaySampleRate: Int64? = nil,
```
ncreated, line 259: > This PR also removes deprecated configuration properties.

question/ We never used these removed attributes, right? Was it that we made this type too verbose at the beginning, and now we pay down debt? Or were they actually removed from the schema?

ncreated (approved):

DatadogInternal/Sources/Telemetry/Telemetry.swift
maxep: Only the `trackInteractions` was in use, we made it too verbose at first. They are not yet removed from the schema, only marked as deprecated.

The `trackInteractions` has been replaced by `trackUserInteractions `, [we will send the 2 for backward compatibility](https://github.com/DataDog/dd-sdk-ios/pull/1782/commits/9cdf221f9fc88363e78735754f87fc012b0aca18#diff-c096953c4d5e7e09d176398ba8cea0413bcc7f770459e420555200187bdfabfeL317-R319).

PR #1772, opened 2024-04-11
RUM-3470 chore: Add tests for head-based trace sampling

ncreated:
### What and why?

📦 🧰 This PR prepares `DatadogTrace` for the implementation of head-based sampling. It only brings refactoring and fills gaps in tests coverage. There are **no changes to existing behaviours**.

Notably, it adds `HeadBasedSamplingTests` suite that covers main scenarios we want to achieve with head-based sampling. That includes:

Head-based sampling of local trace, where all spans must be either "kept" or "dropped":
```
client-ios-app:     [---- custom.span.1 -----]
client-ios-app:       [-- custom.span.2 ---]
client-ios-app:         [- custom.span.3 -]
```

Head-based sampling of distributed trace, where all spans must be either "kept" or "dropped":
```
dd-sdk-ios:         [--- urlsession.request ---]
client backend:        [--- backend span ---]
```

Head-based sampling of distributed trace with local parent, where all spans must be either "kept" or "dropped":
```
client-ios-app:     [-------- active.span ---------]
dd-sdk-ios:           [-- urlsession.request --]
client backend:        [--- backend span ---]
```

The work has to be split between `RUM-3470` (next PR) and `RUM-3535` (_"trace context injection"_). To embrace common understanding of what we want to do, I propose to go red → green refactoring, hence I'm adding multiple tests that are yet disabled.

### How?

- Added `HeadBasedSamplingTests` to add missing coverage and set common goals for `RUM-3470` and `RUM-3535`;
- Renamed samplers in `DatadogTrace` for more clarity:
   - `sampler` → `localTraceSampler` in `DatadogTracer`;
   - `tracingSampler` → `distributedTraceSampler` in `TracingURLSessionHandler`;
- To facilitate further work and better fit tests pyramid, moved first tracer tests from `DatadogCoreTests` (integration testing) to `DatadogTraceTests` (unit testing).

### Review checklist
- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes

### Custom CI job configuration (optional)
- [ ] Run unit tests for Core, RUM, Trace, Logs, CR and WVT
- [ ] Run unit tests for Session Replay
- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

PR #1783, opened 2024-04-17
RUM-3470 feat: Head-based sampling for local and (automatic) distributed tracing

ncreated:
### What and why?

📦 This PR implements head-based sampling for the scenarios defined earlier in https://github.com/DataDog/dd-sdk-ios/pull/1772. It also adds tests for one more class of scenarios: using distributed trace created manually with `OTFormatWriter` API. 

<details><summary> See all scenarios for head-based sampling vs their support level:</summary>
<p>

- ✅ Local trace with explicit parent:
```
client-ios-app:     [---- custom.span.1 -----]
client-ios-app:       [-- custom.span.2 ---]
client-ios-app:         [- custom.span.3 -]
```
- ✅ Local trace with implicit parent:
```
client-ios-app:     [----- active.span ------]
client-ios-app:       [-- custom.span.1 ---]
client-ios-app:         [- custom.span.2 -]
```
- ✅ Distributed trace sent via `URLSessionInstrumentation`:
```
dd-sdk-ios:         [--- urlsession.request ---]
client backend:        [--- backend span ---]
```
- ✅ Distributed trace sent via `URLSessionInstrumentation` with implicit parent:
```
client-ios-app:     [-------- active.span ---------]
dd-sdk-ios:           [--- urlsession.request ---]
client backend:          [--- backend span ---]
```
- ❌ Distributed trace sent via Tracer API (`*HTTPHeadersWriter`):
```
client-ios-app:     [------ network.span -------]
client backend:         [--- backend span ---]
```
- ❌ Distributed trace sent via Tracer API (`*HTTPHeadersWriter`) with implicit parent:
```
client-ios-app:     [------ active.span -------]
client-ios-app:       [---- network.span ----]
client backend:         [-- backend span --]
```
The last 2 scenarios are defined as failing (and disabled) tests. They will be supported and enabled in next PR.

</p>
</details>

### How?

I did few changes that enable this feature.

#### The sampling decision is now a part of `DDSpanContext`

In OT, creating child span requires passing the `OTSpanContext` of the parent. This makes a good place for propagating the sampling decision for head-based sampling. For that reason, `sampleRate` and `isKept` are added to `DDSpanContext`.

```
DDSpanContext (parent): {traceID, spanID, nil, isKept, sampleRate}

                                   ↓

DDSpanContext (child):  {parent.traceID, spanID, parent.spanID, parent.isKept, parent.sampleRate}
```

Note: While `isKept` is crucial for sampling the whole trace equally, the `sampleRate` is not. We need to propagate it, so it can be later sent [in `SpanEvent`](https://github.com/DataDog/dd-sdk-ios/blob/973fb6eb1820faca459e1c14fe4391284ca8290d/DatadogTrace/Sources/Span/SpanEventEncoder.swift#L60) for computing BE-side metrics.

#### The `OTFormatReader` implementations can now extract the sampling decision

Because the `OTTracer` can encode and decode the `OTSpanContext`:
```swift
// OTTracer
func inject(spanContext: OTSpanContext, writer: OTFormatWriter)
func extract(reader: OTFormatReader) -> OTSpanContext?
```
the implementations of HTTP header readers (DD, W3C and B3) were extended to extract the `isKept` and `sampleRate`.

Note: The `sampleRate` is not encoded as header so it needs to be supplied by the `tracer` in `extract(reader:)`.

#### `NetworkInstrumentationFeature` no longer decodes `TraceContext` from request headers

I enhanced the implementation of passing `TraceContext` between task interception methods. It now avoids coding it through request headers, which enables passing extra information not available in standard headers (the `sampleRate`).

Before:
```swift
// NetworkInstrumentationFeature
let request = self.intercept(request: currentRequest, ...) // <-- encode TraceContext in request headers
self.intercept(task: task, ...) // <-- decode TraceContext from task.currentRequest headers
```
Now:
```swift
let (request, traceContexts) = self.intercept(request: currentRequest, ...)
self.intercept(task: task, with: traceContexts, ...)
```

Note: This simplifies and enhances task interception in the main flow. However, it required additional changes in `URLSessionInterceptor` which is required for Alamofire Extension. 

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/Datadog/Datadog.xcodeproj/xcshareddata/xcschemes/DatadogCore iOS.xcscheme
+++ b/Datadog/Datadog.xcodeproj/xcshareddata/xcschemes/DatadogCore iOS.xcscheme
@@ -198,22 +198,22 @@
             </BuildableReference>
             <SkippedTests>
                <Test
-                  Identifier = "HeadBasedSamplingTests/testSamplingLocalTrace()">
+                  Identifier = "HeadBasedSamplingTests/testSendingDroppedDistributedTraceWithNoParent_throughTracerAPI()">
                </Test>
                <Test
-                  Identifier = "HeadBasedSamplingTests/testSamplingLocalTraceWithImplicitParent()">
+                  Identifier = "HeadBasedSamplingTests/testSendingDroppedDistributedTraceWithNoParent_throughURLSessionInstrumentationAPI()">
                </Test>
                <Test
-                  Identifier = "HeadBasedSamplingTests/testSendingDroppedDistributedTraceWithNoParent()">
+                  Identifier = "HeadBasedSamplingTests/testSendingDroppedDistributedTraceWithParent_throughTracerAPI()">
                </Test>
                <Test
-                  Identifier = "HeadBasedSamplingTests/testSendingDroppedDistributedTraceWithParent()">
+                  Identifier = "HeadBasedSamplingTests/testSendingDroppedDistributedTraceWithParent_throughURLSessionInstrumentationAPI()">
                </Test>
                <Test
-                  Identifier = "HeadBasedSamplingTests/testSendingSampledDistributedTraceWithNoParent()">
+                  Identifier = "HeadBasedSamplingTests/testSendingSampledDistributedTraceWithNoParent_throughTracerAPI()">
                </Test>
                <Test
-                  Identifier = "HeadBasedSamplingTests/testSendingSampledDistributedTraceWithParent()">
+                  Identifier = "HeadBasedSamplingTests/testSendingSampledDistributedTraceWithParent_throughTracerAPI()">
                </Test>
             </SkippedTests>
          </TestableReference>
--- a/DatadogInternal/Sources/NetworkInstrumentation/Datadog/HTTPHeadersWriter.swift
+++ b/DatadogInternal/Sources/NetworkInstrumentation/Datadog/HTTPHeadersWriter.swift
@@ -38,7 +38,7 @@ public class HTTPHeadersWriter: TracePropagationHeadersWriter {
     ///
     /// This value will decide of the `x-datadog-sampling-priority` header field value
     /// and if `x-datadog-trace-id` and `x-datadog-parent-id` are propagated.
-    private let sampler: Sampler
+    private let sampler: Sampling
 
     /// Initializes the headers writer.
     ///
@@ -58,7 +58,7 @@ public class HTTPHeadersWriter: TracePropagationHeadersWriter {
     /// Initializes the headers writer.
     ///
     /// - Parameter sampler: The sampler used for headers injection.
-    public init(sampler: Sampler) {
+    public init(sampler: Sampling) {
         self.sampler = sampler
     }
 
--- a/DatadogInternal/Sources/NetworkInstrumentation/DatadogURLSessionHandler.swift
+++ b/DatadogInternal/Sources/NetworkInstrumentation/DatadogURLSessionHandler.swift
@@ -8,35 +8,33 @@ import Foundation
 
 /// An interface for processing `URLSession` task interceptions.
 public protocol DatadogURLSessionHandler {
-    /// The interceptor's first party hosts
+    /// The first party hosts configured for this handler.
     var firstPartyHosts: FirstPartyHosts { get }
 
-    /// Tells the interceptor to modify a URL request.
+    /// Modifies the provided request by injecting trace headers.
     ///
     /// - Parameters:
-    ///   - request: The request to intercept.
-    ///   - additionalFirstPartyHosts: Additional 1st-party hosts.
-    /// - Returns: The modified request.
-    func modify(request: URLRequest, headerTypes: Set<TracingHeaderType>) -> URLRequest
+    ///   - request: The request to be modified.
+    ///   - headerTypes: The types of tracing headers to inject into the request.
+    /// - Returns: A tuple containing the modified request and the injected TraceContext. If no trace is injected (e.g., due to sampling),
+    ///            the returned request remains unmodified, and the trace context will be nil.
+    func modify(request: URLRequest, headerTypes: Set<TracingHeaderType>) -> (URLRequest, TraceContext?)
 
-    /// Returns the trace of the current execution context.
-    func traceContext() -> TraceContext?
-
-    /// Tells the interceptor that the session did start.
+    /// Notifies the handler that the interception has started.
     ///
-    /// - Parameter interception: The URLSession interception.
+    /// - Parameter interception: The URLSession task interception.
     func interceptionDidStart(interception: URLSessionTaskInterception)
 
-    /// Tells the interceptor that the session did complete.
+    /// Notifies the handler that the interception has completed.
     ///
-    /// - Parameter interception: The URLSession interception.
+    /// - Parameter interception: The URLSession task interception.
     func interceptionDidComplete(interception: URLSessionTaskInterception)
 }
 
 extension DatadogCoreProtocol {
     /// Core extension for registering `URLSession` handlers.
     ///
-    /// - Parameter urlSessionHandler: The `URLSession` handlers to register.
+    /// - Parameter urlSessionHandler: The `URLSession` handler to register.
     public func register(urlSessionHandler: DatadogURLSessionHandler) throws {
         let feature = get(feature: NetworkInstrumentationFeature.self) ?? .init()
         feature.handlers.append(urlSessionHandler)
--- a/DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionInterceptor.swift
+++ b/DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionInterceptor.swift
@@ -6,6 +6,8 @@
 
 import Foundation
 
+// TODO: RUM-3470 Add tests to `URLSessionInterceptor`
+
 /// The `URLSession` Interceptor provides methods for injecting distributed-traces
 /// headers into a `URLRequest`and to instrument a `URLURLSessionTask` lifcycle,
 /// from its creation to completion.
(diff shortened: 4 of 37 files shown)


DatadogTrace/Sources/DatadogTracer.swift
```diff
@@ -110,12 +112,14 @@ internal class DatadogTracer: OTTracer {
...
-    internal func createSpanContext(parentSpanContext: DDSpanContext? = nil) -> DDSpanContext {
+    internal func createSpanContext(parentSpanContext: DDSpanContext?, using sampler: Sampler) -> DDSpanContext {
         return DDSpanContext(
             traceID: parentSpanContext?.traceID ?? traceIDGenerator.generate(),
             spanID: spanIDGenerator.generate(),
             parentSpanID: parentSpanContext?.spanID,
-            baggageItems: BaggageItems(parent: parentSpanContext?.baggageItems)
+            baggageItems: BaggageItems(parent: parentSpanContext?.baggageItems),
+            sampleRate: parentSpanContext?.sampleRate ?? sampler.samplingRate,
+            isKept: parentSpanContext?.isKept ?? sampler.sample()
```
ncreated, line 122: 💡 This is the main point for head-based sampling implementation in this PR:
- use the sampling decision from the parent span (if available)
- or determine it based on provided sampler (which can be local or distributed trace sampler).

Datadog/Datadog.xcodeproj/xcshareddata/xcschemes/DatadogCore iOS.xcscheme
```diff
@@ -198,22 +198,22 @@
             </BuildableReference>
             <SkippedTests>
                <Test
-                  Identifier = "HeadBasedSamplingTests/testSamplingLocalTrace()">
+                  Identifier = "HeadBasedSamplingTests/testSendingDroppedDistributedTraceWithNoParent_throughTracerAPI()">
```
ncreated, line 201: 💡 Some tests were renamed, some "red" ones were enabled (now "green") and some new "red" tests were added (disabled) to be fixed in next PR.

DatadogInternal/Sources/NetworkInstrumentation/DatadogURLSessionHandler.swift
```diff
@@ -8,35 +8,33 @@ import Foundation
...
-    /// - Returns: The modified request.
-    func modify(request: URLRequest, headerTypes: Set<TracingHeaderType>) -> URLRequest
+    ///   - request: The request to be modified.
+    ///   - headerTypes: The types of tracing headers to inject into the request.
+    /// - Returns: A tuple containing the modified request and the injected TraceContext. If no trace is injected (e.g., due to sampling),
+    ///            the returned request remains unmodified, and the trace context will be nil.
+    func modify(request: URLRequest, headerTypes: Set<TracingHeaderType>) -> (URLRequest, TraceContext?)
 
-    /// Returns the trace of the current execution context.
-    func traceContext() -> TraceContext?
```
ncreated, line 23: 💡 No longer needed. It was actually doubling the effort:
- active span was obtained once in `modify(request:)` of Trace handler,
- obtained second time in `intercept(task:)` in `NetworkInstrumentationFeature`,
- it was no-op in RUM handler.

Now we read it only once, in Trace handler and pass the `parentSpanID` in `TraceContext`.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionInterceptor.swift
```diff
@@ -6,6 +6,8 @@
 
 import Foundation
 
+// TODO: RUM-3470 Add tests to `URLSessionInterceptor`
```
ncreated, line 9: 💡 `URLSessionInterception` has no tests coverage. I will add tests in next PR. Here, we're merging against feature branch.

DatadogInternal/Tests/NetworkInstrumentation/NetworkInstrumentationFeatureTests.swift
```diff
@@ -466,101 +466,23 @@ class NetworkInstrumentationFeatureTests: XCTestCase {
         )
     }
 
-    // MARK: - URLRequest Interception
+    // MARK: - URLSessionTask Interception
 
-    func testGivenOpenTracing_whenInterceptingRequests_itInjectsTrace() throws {
-        // Given
```
ncreated, line 472: 💡 All these tests are not necessary because we removed `traceContext()` in `DatadogURLSessionHandler`.

DatadogTrace/Sources/DatadogTracer.swift
```diff
@@ -101,7 +101,9 @@ internal class DatadogTracer: OTTracer {
 
     func extract(reader: OTFormatReader) -> OTSpanContext? {
         // TODO: RUMM-385 - make `HTTPHeadersReader` available in public API
-        reader.extract()
+        var reader = reader as? TracePropagationHeadersReader
+        reader?.tracerSampleRate = localTraceSampler.samplingRate
+        return (reader as? OTFormatReader)?.extract()
```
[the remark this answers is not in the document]
ncreated: Fair call 👍. Like you said @maxep, head-based sampling requires touching multiple moving parts and I found it impossible fit it in seamlessly. This piece is a clear stain, let's work on it 🙌.

I like the 2nd proposal more as it is better scoped. The `DDSpanContext` is already a `struct`, so it is cheap to create it again. Also, in next PR we will likely deprecate HTTP header writers and readers as being not backward-compatible for the 2 remaining scenarios outlined in the PR description. `TracePropagationHeadersReader` may become obsolete in consequence.

Keep in mind tho, that 2nd approach will include a trap. There will be no good default we could extract out of headers for non-optional `DDSpanContext.sampleRate`, so we'd need to pass it a "whatever" value that will be overwritten later in `Tracer`:
```swift
extension TracePropagationHeadersReader where Self: OTFormatReader {
    public func extract() -> OTSpanContext? {
        guard let ids = read() else {
            return nil
        }
        return DDSpanContext(
            traceID: ids.traceID,
            spanID: ids.spanID,
            parentSpanID: ids.parentSpanID,
            baggageItems: BaggageItems(),
            sampleRate: 0, // [!!!] placeholder, it will be overwritten in Tracer.extract()
            isKept: sampled ?? false // unreachable default: we got trace ID, so this request must have been instrumented
        )
    }
}
```

By principle, **this situation should be solved by making `sampleRate` an optional member of `DDSpanContext`**. This however moves another part → the `DatadogInternal.TraceContext` that will also require optionality in `sampleRate`. This begins the chain reaction and `guard let { /* not reachable */ }` surgery in network instrumentation code. All that for the sake of principle → not worth it 🙅. I think we should accept `0` default as it has no practical consequence (it will be always overwritten by `Tracer`).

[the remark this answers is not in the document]
maxep: That would make sense 👍 We could still require an optional sample rate in the reader, such as:
```swift
public protocol TracePropagationHeadersReader {
    /// The sample rate used to sample this trace.
    var tracerSampleRate: Float? { get } // getter only that can be set at init

    func read() -> (
        traceID: TraceID,
        spanID: SpanID,
        parentSpanID: SpanID?,
        sampled: Bool? // can this be part of the read method?
    )?
}
```
But as you said, if `0` as default as no consequences lets remove the sampling rate from the reader 👍 Especially if the current writer/reader become deprecated!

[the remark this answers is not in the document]
ncreated: ✅ done

DatadogInternal/Sources/NetworkInstrumentation/Datadog/HTTPHeadersWriter.swift
```diff
@@ -58,7 +58,7 @@ public class HTTPHeadersWriter: TracePropagationHeadersWriter {
     /// Initializes the headers writer.
     ///
     /// - Parameter sampler: The sampler used for headers injection.
-    public init(sampler: Sampler) {
+    public init(sampler: Sampling) {
```
ganeshnj, line 61: ❗ 
This is a breaking change

DatadogInternal/Sources/Utils/Sampler.swift
```diff
@@ -21,3 +30,25 @@ public struct Sampler {
         return Float.random(in: 0.0..<100.0) < samplingRate
     }
 }
+
+/// A sampler that determines sampling decisions deterministically (the same each time).
+public struct DeterministicSampler: Sampling {
```
ganeshnj, line 35: for inspiration, I think we can multiple samplers like here https://github.com/open-telemetry/opentelemetry-swift/blob/4eb75bc8f0d6449bc197637b5c6044b51ab8c4ed/Sources/OpenTelemetrySdk/Trace/Samplers/Samplers.swift#L10

I'm particularly interested in `ParentBasedSampler`

ganeshnj (commented):
Looks good nothing block.

ganeshnj:
✅ Local trace with implicit parent:
```
client-ios-app:     [----- active.span ------]
client-ios-app:       [-- custom.span.1 ---]
client-ios-app:         [- custom.span.2 -]
```

This must be rather from parent-child point of view.
```
client-ios-app:     [----- active.span ------]
client-ios-app:       [-- custom.span.1 ---]
client-ios-app:       [-- custom.span.2 ---]
```

DatadogInternal/Sources/NetworkInstrumentation/Datadog/HTTPHeadersWriter.swift
ncreated: Technically yes, but the `Sampler` type is only defined in `DatadogInternal` without being re-exported from user-facing modules. In practise, the only supported public initializer for `HTTPHeadersWriter` is `init(sampleRate:)`, which is not altered by this change.

DatadogInternal/Sources/Utils/Sampler.swift
ncreated: @ganeshnj I'm sure what is the actual ask here 🙂, could you clarify? The SDK doesn't expose any sampler implementation to the end-user (interface + impls exist only in `DatadogInternal`). Because end-users only work with `sampleRate: Float` API today, I don't see reason for introducing more samplers than we need for SDK internals.

ganeshnj: apologies, it wasn't clear. I mean to say - for our internal working we can use some existing pattern. I wasn't expecting any action on the comment. Just for information if we have tricky cases around sampling decision.

ganeshnj (approved):

maxep (approved):
Well done!

PR #1784, opened 2024-04-17
RUM-4086 Synchronize SR info with WebViews

maxep:
### What and why?

This fixes some discrepancies between the Webview and host app content regarding Session Replay.

Same as  DataDog/dd-sdk-android#1990

### How?

- The `_dd.session.plan` field has been deprecated and should be removed
- The `session.has_replay` and `_dd.replay_stats` fields should be discarded from the webview RUM event if SR is disable in the host application, to avoid orphan SR sessions

### Review checklist

- [x] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/Datadog/IntegrationUnitTests/RUM/WebEventIntegrationTests.swift
+++ b/Datadog/IntegrationUnitTests/RUM/WebEventIntegrationTests.swift
@@ -61,7 +61,7 @@ class WebEventIntegrationTests: XCTestCase {
             "session": {
               "id": "0110cab4-7471-480e-aa4e-7ce039ced355",
               "type": "user",
-                            "has_replay": true
+              "has_replay": true
             },
             "type": "view",
             "view": {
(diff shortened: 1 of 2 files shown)


Datadog/IntegrationUnitTests/RUM/WebEventIntegrationTests.swift
```diff
@@ -60,7 +60,8 @@ class WebEventIntegrationTests: XCTestCase {
             "service": "super",
             "session": {
               "id": "0110cab4-7471-480e-aa4e-7ce039ced355",
-              "type": "user"
+              "type": "user",
+                            "has_replay": true
```
maciejburda, line 64:
```suggestion
              "type": "user",
              "has_replay": true
```

DatadogRUM/Sources/Integrations/WebViewEventReceiver.swift
```diff
@@ -100,13 +100,17 @@ internal final class WebViewEventReceiver: FeatureMessageReceiver {
 
                 if var session = event["session"] as? JSON {
                     session["id"] = rum.sessionID
+                    // RUM-4086: unset `has_replay` if native replay is disabled
```
maciejburda, line 103: Not sure if we need to quote the ticket number in both cases. Probably comment explaining is enough. We can track ticket through the PR

maciejburda (approved):
LGTM

PR #1787, opened 2024-04-22
Fix color SR identifier

maciejburda:
### What and why?

Fixes the SR color identifier, which is causing unnecessary upload of resources.

### How?

Previous implementation was relying on UIColor (NSObject) hash. It appears to be stable only in the developer environment, but it feels random on production.

Testing is tricky, because different versions of iOS produce different images, and hashes.
Improved tests to check if length of the identifier, as well as if it contains color for tinted images.

Also, it resets the store version.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogSessionReplay/Sources/Recorder/Utilities/UIImage+SessionReplay.swift
+++ b/DatadogSessionReplay/Sources/Recorder/Utilities/UIImage+SessionReplay.swift
@@ -42,10 +42,7 @@ extension UIColor {
         var b: CGFloat = 0
         var a: CGFloat = 0
         getRed(&r, green: &g, blue: &b, alpha: &a)
-        return String(
-            format: "%02X%02X%02X%02X",
-            Int(round(r * 255)), Int(round(g * 255)), Int(round(b * 255)), Int(round(a * 255))
-        )
+        return String(format: "%02X%02X%02X%02X", Int(round(r * 255)), Int(round(g * 255)), Int(round(b * 255)), Int(round(a * 255)))
     }
 }
 #endif
--- a/DatadogSessionReplay/Sources/Recorder/Utilities/UIImage+SessionReplay.swift
+++ b/DatadogSessionReplay/Sources/Recorder/Utilities/UIImage+SessionReplay.swift
@@ -37,7 +37,15 @@ extension DatadogExtension where ExtendedType: UIImage {
 
 extension UIColor {
     var srIdentifier: String {
-        return "\(hash)"
+        var r: CGFloat = 0
+        var g: CGFloat = 0
+        var b: CGFloat = 0
+        var a: CGFloat = 0
+        getRed(&r, green: &g, blue: &b, alpha: &a)
+        return String(
+            format: "%02X%02X%02X%02X",
+            Int(round(r * 255)), Int(round(g * 255)), Int(round(b * 255)), Int(round(a * 255))
+        )
     }
 }
 #endif
--- a/DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UIImageResourceTests.swift
+++ b/DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UIImageResourceTests.swift
@@ -20,7 +20,7 @@ final class UIImageResourceTests: XCTestCase {
         let image = createSinglePixelImage()
         let imageResource = UIImageResource(image: image, tintColor: nil)
 
-        XCTAssertNotEqual(imageResource.calculateIdentifier(), "")
+        XCTAssertEqual(imageResource.calculateIdentifier().count, 32)
         XCTAssertGreaterThan(imageResource.calculateData().count, 0)
     }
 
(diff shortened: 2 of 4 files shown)


DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UIImageResourceTests.swift
```diff
@@ -20,7 +20,7 @@ final class UIImageResourceTests: XCTestCase {
         let image = createSinglePixelImage()
         let imageResource = UIImageResource(image: image, tintColor: nil)
 
-        XCTAssertNotEqual(imageResource.calculateIdentifier(), "")
+        XCTAssertEqual(imageResource.calculateIdentifier().count, 32)
```
maxep, line 23: /question Could you explain why we cannot assert the value? The tint color will be different based on the OS version? If so, I wonder if the tint color should actually be part of the identifier 🤔

maciejburda: the image hash will be different. Different iOS versions produce slightly different data, hence different hash.

We could potentially lock it in for currently selected iOS/Simu

DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UIImageResourceTests.swift
```diff
@@ -38,7 +39,7 @@ final class UIImageResourceTests: XCTestCase {
         let image = UIImage(systemName: "circle.fill")!
         let imageResource = UIImageResource(image: image, tintColor: nil)
 
-        XCTAssertNotEqual(imageResource.calculateIdentifier(), "")
+        XCTAssertEqual(imageResource.calculateIdentifier().count, 32)
         XCTAssertGreaterThan(imageResource.calculateData().count, 0)
```
ganeshnj, line 43: how come these asserts are different.

Like 32 vs 40 characters and for 40 char there is contains check but not for 32 version?

maciejburda: We append identifier of the color at the end if we recognise that the image should be tinted.

Thing is, icons are in a solid color, and the tint is only applied within the UIImageView.

DatadogSessionReplay/Sources/Recorder/Utilities/UIImage+SessionReplay.swift
```diff
@@ -37,7 +37,12 @@ extension DatadogExtension where ExtendedType: UIImage {
 
 extension UIColor {
     var srIdentifier: String {
-        return "\(hash)"
+        var r: CGFloat = 0
+        var g: CGFloat = 0
+        var b: CGFloat = 0
+        var a: CGFloat = 0
+        getRed(&r, green: &g, blue: &b, alpha: &a)
+        return String(format: "%02X%02X%02X%02X", Int(round(r * 255)), Int(round(g * 255)), Int(round(b * 255)), Int(round(a * 255)))
```
ncreated, line 45: question/ What was the actual issue with using `hash`? Can we elaborate on this in PR description?

What is unclear to me:

> Fixes the SR color identifier, which is causing unnecessary upload of resources
> ...
> different versions of iOS produce different images, and hashes

So, if the `hash` is stable on iOS version, why  was it causing non-deterministic resource IDs?

maciejburda: It appears to be stable only in the developer environment. It's totally random on production.

DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UIImageResourceTests.swift
maciejburda, line 1: I made them explicit for the CI's simulator

DatadogSessionReplay/Sources/Recorder/Utilities/UIImage+SessionReplay.swift
ncreated: Ok, pretty clear 👌

DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UIImageResourceTests.swift
maciejburda: Reverted, as CI was producing different images

ncreated (approved):

ganeshnj (approved):

Issue #1690, opened 2024-02-19
Crash on widget extension


ivopintodasilva:
### The crash

This potential crash causes widgets to be inactive (and some times totally blank).
The only common clue in the stacktrace is Datadog's `DataUploadWorker` reading files from the disk:

<img width="836" alt="image" src="https://github.com/DataDog/dd-sdk-ios/assets/4477378/d9b45966-7116-47d9-8733-358ee1729ecc">

---

<img width="813" alt="image" src="https://github.com/DataDog/dd-sdk-ios/assets/4477378/9efa6816-439e-4706-8049-ccdc34614696">

---

<img width="846" alt="image" src="https://github.com/DataDog/dd-sdk-ios/assets/4477378/34ffe065-4270-490e-ac81-b5770d1a6da4">
<img width="846" alt="image" src="https://github.com/DataDog/dd-sdk-ios/assets/4477378/3a4d53cd-38fe-4354-8f56-02943f8143c5">

---

In order to avoid uploading logs from the widget extension, I tried to disable the upload altogether (and only write them to disk so that they are uploaded from the main app) but afaik it's not possible to achieve this in the current SDK.

Any suggestions on what I could try next?

### Reproduction

This crash is very difficult to replicate and I can't provide a project that does it consistently.
One pattern that I've detected is that it happens more frequently once the app is updated to a new version (eg. whenever we release an update to the app store).

---

#### Datadog SDK versions:

1.22.0

#### OS version:

<img width="246" alt="image" src="https://github.com/DataDog/dd-sdk-ios/assets/4477378/092a4ac7-ab9f-4206-a561-8bf838639906">

#### Deployment Target:

iOS 15

#### Environment:

Seems to happen more frequently once the app is updated to a new version (eg. whenever we release an update to the app store).

maciejburda:
Thanks for reporting!

We'll look into it and let you know.

maciejburda:
@ivopintodasilva I am trying to recreate the issue locally. Can you give me a hint about the exact type of widget you are using? Also, where in it's lifecycle are you using logger?

ivopintodasilva:
thank you for looking into it! 🙏 

> Can you give me a hint about the exact type of widget you are using?

our widget extension has 4 different widgets:

- 3 "regular" widgets (using `StaticConfiguration` as the `WidgetConfiguration`)
[...]

PR #1803, opened 2024-04-26
RUM-3284 Extension background upload

maciejburda:
### What and why?

This PR makes extension targets use `beginActivity(options:reason:)` and `endActivity(_:)` APIs when `backgroundTasksEnabled` configuration is enabled. This allows slightly extending the lifetime of the extension in order to upload all the necessary data.

https://developer.apple.com/documentation/foundation/processinfo/1415995-beginactivity
https://developer.apple.com/documentation/foundation/processinfo/1411321-endactivity

It is recommended to use this configuration of the SDK when initialising from extensions:
```swift
Datadog.initialize(
    with: Datadog.Configuration(
        clientToken: clientToken,
        env: environment,
        site: .us1,
        backgroundTasksEnabled: true // This will give more time for extensions to upload SDK data
    ),
    trackingConsent: .granted,
    instanceName: "UniqueExtensionName" // This line is important when using the SDK in multiple extensions
)
```

### How?

This PR adds extension compatible mechanism for background upload as well as mentions the recommended configuration setup when using the SDK from within extensions. Currently adopted mechanism only works from app targets.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogCore/Sources/Core/Storage/Directories.swift
+++ b/DatadogCore/Sources/Core/Storage/Directories.swift
@@ -44,8 +44,8 @@ internal extension CoreDirectory {
     ///   - osDirectory: the root OS directory (`/Library/Caches`) to create core directory inside.
     ///   - instancenName: The core instance name.
     ///   - site: The cor instance site.
-    init(in osDirectory: Directory, instancenName: String, site: DatadogSite) throws {
-        let sdkInstanceUUID = sha256("\(instancenName)\(site)")
+    init(in osDirectory: Directory, instanceName: String, site: DatadogSite) throws {
+        let sdkInstanceUUID = sha256("\(instanceName)\(site)")
         let path = "com.datadoghq/v2/\(sdkInstanceUUID)"
 
         self.init(
--- a/DatadogCore/Sources/Core/Storage/Directories.swift
+++ b/DatadogCore/Sources/Core/Storage/Directories.swift
@@ -42,7 +42,7 @@ internal extension CoreDirectory {
     /// 
     /// - Parameters:
     ///   - osDirectory: the root OS directory (`/Library/Caches`) to create core directory inside.
-    ///   - instancenName: The core instance name.
+    ///   - instanceName: The core instance name.
     ///   - site: The cor instance site.
     init(in osDirectory: Directory, instanceName: String, site: DatadogSite) throws {
         let sdkInstanceUUID = sha256("\(instanceName)\(site)")
--- a/DatadogCore/Sources/Core/Upload/BackgroundTaskCoordinator.swift
+++ b/DatadogCore/Sources/Core/Upload/BackgroundTaskCoordinator.swift
@@ -61,28 +61,36 @@ internal class AppBackgroundTaskCoordinator: BackgroundTaskCoordinator {
     }
 }
 
+/// Bridge protocol that matches `UIApplication` interface for background tasks. Allows easier testablity.
+internal protocol ProcessInfoActivityCoordinator {
+    func beginActivity(options: ProcessInfo.ActivityOptions, reason: String) -> any NSObjectProtocol
+    func endActivity(_ activity: any NSObjectProtocol)
+}
+
+extension ProcessInfo: ProcessInfoActivityCoordinator {}
+
 internal class ExtensionBackgroundTaskCoordinator: BackgroundTaskCoordinator {
-    private let processInfo: ProcessInfo
+    private let processInfo: ProcessInfoActivityCoordinator
 
     @ReadWriteLock
     private var currentActivity: NSObjectProtocol?
 
     internal init(
-        processInfo: ProcessInfo = .init()
+        processInfo: ProcessInfoActivityCoordinator = ProcessInfo()
     ) {
         self.processInfo = processInfo
     }
 
     internal func beginBackgroundTask() {
         endBackgroundTask()
-        currentActivity = processInfo.beginActivity(options: [.background], reason: "Datadog background upload")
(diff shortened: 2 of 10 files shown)


fuzzybinary (approved):

ganeshnj (commented):
I have couple of question on the approach here.

- When a batch is written - there can only be one and only one writer, because the way we name it. Once the file is written to the disk it is immutable.
	- in this process, the lock must be on the file, not on the directory, if we need.
- Now reading the file from multiple sources must not make a difference.

While working on a personal project, I realized the programming model of widgets is quite different from lifecycle point of view. Hence I wonder if it even makes sense to upload batches from extension itself. 

https://developer.apple.com/documentation/widgetkit/making-network-requests-in-a-widget-extension

DatadogCore/Sources/Core/Storage/Directories.swift
```diff
@@ -44,8 +44,8 @@ internal extension CoreDirectory {
     ///   - osDirectory: the root OS directory (`/Library/Caches`) to create core directory inside.
     ///   - instancenName: The core instance name.
```
ncreated, line 45:
```suggestion
    ///   - instanceName: The core instance name.
```

maciejburda:
> I have couple of question on the approach here.
> 
> * When a batch is written - there can only be one and only one writer, because the way we name it. Once the file is written to the disk it is immutable.
>   
>   * in this process, the lock must be on the file, not on the directory, if we need.
> * Now reading the file from multiple sources must not make a difference.
> 
> While working on a personal project, I realized the programming model of widgets is quite different from lifecycle point of view. Hence I wonder if it even makes sense to upload batches from extension itself.
> 
> https://developer.apple.com/documentation/widgetkit/making-network-requests-in-a-widget-extension

Problem is we don't have access to app's directory from the extension. For example I get path like this:
App:
```"com.datadoghq/v2/f5da9c34155155fa15554148a8a2d145170e0a082ddf9cba39ea3a198b2bb947"
file:///Users/maciej.burda/Library/Developer/CoreSimulator/Devices/7A836CF3-9C5A-4B90-BA73-76D0F1B5EB87/
data/Containers/Data/Application/93B8D5CA-40AD-4AAD-9B2F-C5F8DEBAB04E/Library/Caches/com.datadoghq/v2/f5da9c34155155fa15554148a8a2d145170e0a082ddf9cba39ea3a198b2bb947/rum/v2/```

Extension:
```"com.datadoghq/v2/f5da9c34155155fa15554148a8a2d145170e0a082ddf9cba39ea3a198b2bb947"\
file:///Users/maciej.burda/Library/Developer/CoreSimulator/Devices/5A69FAE7-6D74-49CC-803B-B7E64C895AB2/
data/Library/Caches/com.datadoghq/v2/f5da9c34155155fa15554148a8a2d145170e0a082ddf9cba39ea3a198b2bb947/logging/v2/```

For sharing the storage we could think of using app groups for instance, but this requires bigger planning and probably an RFC.

This PR helps but doesn't fully resolves the `0xdead10cc` which happens when file handle is open, while extension is being suspended. BG tasks will increase the amount of time we have to perform all the SDK work.

File: DatadogCore/Sources/Core/Upload/BackgroundTaskCoordinator.swift [hunk 2]
```diff
@@ -60,4 +60,38 @@ internal class UIKitBackgroundTaskCoordinator: BackgroundTaskCoordinator {
...
+    internal init(
+        processInfo: ProcessInfoActivityCoordinator = ProcessInfo()
+    ) {
+        self.processInfo = processInfo
+    }
+
+    internal func beginBackgroundTask() {
+        endBackgroundTask()
+        currentActivity = processInfo.beginActivity(options: [.background], reason: "Datadog SDK background upload")
+    }
```
ncreated, line 87: question/ Because `beginBackgroundTask()` [is called](https://github.com/DataDog/dd-sdk-ios/blob/ad20f2d4ba50de8f01af7ffcbd18348ebe9fbf5b/DatadogCore/Sources/Core/Upload/DataUploadWorker.swift#L80C49-L80C68) right before the SDK starts uploading data, how does it correspond to the [explanation of `0xdead10cc`](https://developer.apple.com/documentation/xcode/sigkill)?

> (...) Make this request **well before starting to write to the file** in order to complete those operations and relinquish the lock before the app suspends.

Apple suggestion is to wrap "file **write**" operations with `.beginActivity()` and `.endActivity()`. What we seem to do instead is wrapping data uploads. Considering that [upload starts](https://github.com/DataDog/dd-sdk-ios/blob/ad20f2d4ba50de8f01af7ffcbd18348ebe9fbf5b/DatadogCore/Sources/Core/Upload/DataUploadWorker.swift#L80-L81) after the [file was **read**](https://github.com/DataDog/dd-sdk-ios/blob/ad20f2d4ba50de8f01af7ffcbd18348ebe9fbf5b/DatadogCore/Sources/Core/Upload/DataUploadWorker.swift#L77) and it has nothing to do with file writes (which are decoupled SDK routine), I'm not seeing clearly how this solves the `0xdead10cc` problem.

Shouldn't we instead use `.beginActivity()` and `.endActivity()` when writing to files, instead of when performing data upload 🤔💭?

maciejburda:
As described in https://github.com/DataDog/dd-sdk-ios/issues/1690 we ruled out some potential root causes in the SDK code.

Now, this PR is still relevant, but it acts more as a fix of background tasks for extension targets. I updated the description accordingly. Please review again :)

ncreated (dismissed):
Looks good 👌, thanks for doing an extra iteration on this 💪

PR #1804, opened 2024-04-29
Add `DatadogSessionReplay` to the Swift API surface generation

0xnm:
### What and why?

This PR adds `DatadogSessionReplay` to the Swift API surface generation.

One thing which comes from this PR is that since `DatadogSessionReplay` also contains some Objective-C bindings, they leak into Swift API surface file. But to address this changes are needed in generation logic, it seems.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/api-surface-swift
+++ b/api-surface-swift
@@ -1573,3 +1573,390 @@ public enum WebViewTracking
     public class AbstractMessageEmitter
         public func send(body: Any, slotId: String? = nil)
     public static func messageEmitter(in core: DatadogCoreProtocol,logsSampleRate: Float = 100) -> AbstractMessageEmitter
+
+
+# ----------------------------------
+# API surface for DatadogSessionReplay:
+# ----------------------------------
+
+[?] extension SRTextPosition.Alignment
+    public init(systemTextAlignment: NSTextAlignment,vertical: SRTextPosition.Alignment.Vertical = .center)
+public struct SRSegment: SRDataModel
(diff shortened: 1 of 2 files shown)


api-surface-swift
```diff
@@ -1573,3 +1573,390 @@ public enum WebViewTracking
...
+    public let nodes: [SessionReplayNode]
+    public let resources: [SessionReplayResource]
+    public init(subtreeStrategy: SessionReplayNodeSubtreeStrategy,nodes: [SessionReplayNode],resources: [SessionReplayResource] = [])
+public final class DDSessionReplay: NSObject
+    public static func enable(with configuration: DDSessionReplayConfiguration)
+public final class DDSessionReplayConfiguration: NSObject
+    @objc public var replaySampleRate: Float
+    @objc public var defaultPrivacyLevel: DDSessionReplayConfigurationPrivacyLevel
+    @objc public var customEndpoint: URL?
+    public required init(replaySampleRate: Float)
```
0xnm, line 1945: @DataDog/rum-mobile-ios Since `DatadogSessionReplay` module contains ObjC bindings, they leak into `api-surface-swift` file. This needs to be addressed in generator.

maxep: Yes indeed!

maxep (approved):

PR #1825, opened 2024-05-09
[RUM-2911] Report time since applications start in RUM errors

plousada:
### What and why?

This PR adds the time measurement between application start and RUM Errors (in milliseconds). 
It also addresses a loss in millisecond precision that was found in the collection of process start timings. 

### How?

`error.time_since_app_start` property is added to RUM Errors in order to propagate this information. 
Application start is defined as the process start, the same mechanism as for `ApplicationLaunch` view. 

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


ncreated (changes requested):
Well scoped change, touching the right things 🎯 

I left few suggestions and some request changes. Notable problems are:
- no consideration of app prewarm state, which will lead to oddly large "time since app start" values 
- smaller problems with overflow consideration in `Double` → `Int` casting

I'll also help with a change simplifying and better testing `Date` coding in `CrashContextTests`.

DatadogRUM/Sources/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -552,6 +552,10 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
 
         var commandAttributes = command.attributes
         let errorFingerprint = commandAttributes.removeValue(forKey: RUM.Attributes.errorFingerprint) as? String
+        var timeSinceAppStart: Int64? = nil
+        if let startTime = context.launchTime?.launchDate {
```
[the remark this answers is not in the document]
plousada: I feel it's still valid to report it even if the app has been pre-warmed. 
It could serve as a useful indication to the user that this error occurred after the app has been pre-warmed, which might otherwise not be the case no cold launches. 

It also aligns with Android's behavior.

DatadogRUM/Sources/Instrumentation/AppHangs/FatalAppHangsHandler.swift
```diff
@@ -92,6 +92,10 @@ internal final class FatalAppHangsHandler {
 
             let realErrorDate = fatalHang.hang.startDate.addingTimeInterval(fatalHang.serverTimeOffset)
             let realDateNow = dateProvider.now.addingTimeInterval(context.serverTimeOffset)
+            var timeSinceAppStart: TimeInterval? = nil
+            if let startTime = context.launchTime?.launchDate {
+                timeSinceAppStart = realErrorDate.timeIntervalSince(startTime) * 1_000
+            }
```
ncreated, line 98: change request/ The `timeSinceAppStart` computation here is wrong:
- `fatalHang.hang.startDate` is the start date of the hang in **previous app process** (terminated one).
- `context.launchTime.launchDate` is the launch date of the **current app process** (the one sending error).

so the "time since" value will become negative.


We must persist the `appLaunchDate` information [in `FatalAppHang`](https://github.com/DataDog/dd-sdk-ios/blob/caa508b7097f0a0bdc9dab1ae668bb162c2c7fee/DatadogRUM/Sources/Instrumentation/AppHangs/AppHang.swift#L31-L34) similar way to how we do it for crash reports (in `CrashContext`).

ncreated (changes requested):
The way we compute `timeSinceAppStart` for FatalAppHangs seems wrong (more info in in-line comment). Once fixed, I can help with adding tests coverage for this value in fatal hangs 👌.

DatadogRUM/Sources/Instrumentation/AppHangs/FatalAppHangsHandler.swift
```diff
@@ -92,6 +93,7 @@ internal final class FatalAppHangsHandler {
 
             let realErrorDate = fatalHang.hang.startDate.addingTimeInterval(fatalHang.serverTimeOffset)
             let realDateNow = dateProvider.now.addingTimeInterval(context.serverTimeOffset)
+            let timeSinceAppStart = fatalHang.appLaunchDate.map { realErrorDate.timeIntervalSince($0) }
```
ncreated, line 96: FYI, there was a bug in unit conversion (multiplying by `1_000` while the value is `TimeInterval`, so must be in seconds). I pushed a fix with adding tests to `timeSinceAppStart` for Fatal App Hangs. It is all good now.

ncreated (approved):
Looks great 👌. Thanks for extra efforts 💪!

ncreated (approved):

PR #1830, opened 2024-05-13
RUM-3588 chore: Add tests for global RUM attributes

ncreated:
### What and why?

📦 This PR adds tests to existing RUM global attributes logic. This is to increase unit tests coverage in different scenarios:
- changing attributes right after SDK init;
- changing attributes after view is started;
- changing attributes while inactive view (view with pending resource).

### How?

Defined new **unit test** suite in `DatadogRUMTests` - asserting all existing behaviours for global RUM attributes handling.

I'm not yet removing the [existing RUM attributes **integration tests** in `DatadogCoreTests`](https://github.com/DataDog/dd-sdk-ios/blob/2e74021bfd7a1fd52a7e1c89a9da26eca776b580/DatadogCore/Tests/Datadog/RUM/RUMMonitorTests.swift#L1426-L1446) as they cover also the situation logic of merging command-specific attributes with global ones.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogRUM/Tests/RUMMonitor/Monitor+GlobalAttributesTests.swift
+++ b/DatadogRUM/Tests/RUMMonitor/Monitor+GlobalAttributesTests.swift
@@ -116,6 +116,24 @@ class Monitor_GlobalAttributesTests: XCTestCase {
         XCTAssertEqual(lastView.attribute(forKey: "attribute2"), "value2")
     }
 
+    func testUpdatingGlobalAttributesAfterSDKInit_thenStartingAndStoppingView() throws {
+        // Given
+        monitor.notifySDKInit()
+
+        // When
+        monitor.addAttribute(forKey: "attribute", value: "value")
+        monitor.addAttribute(forKey: "attribute", value: "new-value")
+        monitor.removeAttribute(forKey: "unknown-attribute")
+        monitor.startView(key: "View")
+        monitor.stopView(key: "View")
+
+        // Then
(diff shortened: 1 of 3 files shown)


maciejburda (approved):

DatadogRUM/Tests/RUMMonitor/Monitor+GlobalAttributesTests.swift
```diff
@@ -0,0 +1,419 @@
...
+    }
+
+    func testAddingGlobalAttributesAfterSDKInit_thenRemovingAttribute() throws {
+        // Given
+        monitor.notifySDKInit()
+        monitor.addAttribute(forKey: "attribute1", value: "value1")
+        monitor.addAttribute(forKey: "attribute2", value: "value2")
+
+        // When
+        monitor.removeAttribute(forKey: "attribute1")
```
ganeshnj, line 75: /suggestion

- can we add a test something like removing an attribute which doesn't exist? Ideally should be a noop.

- also adding the same key attribute again

ganeshnj (approved):

DatadogRUM/Tests/RUMMonitor/Monitor+GlobalAttributesTests.swift
ncreated: Fair point 👍. Added extra coverage for updating attributes and removing non-existing ones.

maciejburda (approved):

PR #1834, opened 2024-05-14
RUM-3588 fix: Always update crash context with RUM attributes

ncreated:
### What and why?

📦 With this PR, RUM error and RUM view sent after crash will always include up-to-date attributes set on `RUMMonitor`.

Because we don't update crash context with the "last RUM view" info on every change to global attributes, the global RUM attributes held in "last view" were not always correct. Global attributes change that happened during the last active view was never included in the crash error.

### How?

The `RUM.FatalErrorContextNotifier` is updated with `globalAttributes` property. Upon change, it sends a baggage over message bus holding `DatadogInternal.GlobalRUMAttributes` (new type). It is received in `DatadogCrashReporting` and gets encoded into `CrashContext`. 

Upon restarting the app from crash, the value is decoded to `RUM.CrashContext` on regular basis.

🎁 On testing front, I separated tests for updating the `FatalErrorContextNotifier` (1 and 2) from sending encoded baggage over message bus (3 and 4):

<p align="center">
Screenshot 2024-05-16 at 18 19 42
</p>

#### ♻️ Considered Alternative Approach

In my initial attempt, I tried a simpler approach of updating `crashContext.lastRUMView` with latest RUM attributes. That only required producing view updates on every global attribute change and [updating the view event in `FatalErrorContext`](https://github.com/DataDog/dd-sdk-ios/blob/c044ef9c69369a44c772d395c838d8c6951c2ab3/DatadogRUM/Sources/RUMMonitor/Scopes/RUMViewScope.swift#L548-L549) inside `RUMViewScope`. This however added a linear backpressure to the RUM queue (red line), growing with the number of calls to `monitor.addAttribute(forKey:)`. With lack of "add multiple attribute**s**" API, this perf hit is not acceptable.

The solution implemented in this PR adds much less to the RUM backpressure (yellow line). The impact is related to message bus communication which is done on shared thread. 

<p align="center">
Screenshot 2024-05-16 at 18 33 20
</p>

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogCore/Tests/Datadog/CrashReporting/CrashContext/CrashContextProviderTests.swift
+++ b/DatadogCore/Tests/Datadog/CrashReporting/CrashContext/CrashContextProviderTests.swift
@@ -302,6 +302,7 @@ class CrashContextProviderTests: XCTestCase {
     // MARK: - Helpers
 
     private func DDAssert(crashContext: CrashContext, includes sdkContext: DatadogContext, file: StaticString = #filePath, line: UInt = #line) {
+        XCTAssertEqual(crashContext.appLaunchDate, sdkContext.launchTime?.launchDate, file: file, line: line)
         XCTAssertEqual(crashContext.serverTimeOffset, sdkContext.serverTimeOffset, file: file, line: line)
         XCTAssertEqual(crashContext.service, sdkContext.service, file: file, line: line)
         XCTAssertEqual(crashContext.env, sdkContext.env, file: file, line: line)
--- a/DatadogCore/Tests/Datadog/CrashReporting/CrashContext/CrashContextProviderTests.swift
+++ b/DatadogCore/Tests/Datadog/CrashReporting/CrashContext/CrashContextProviderTests.swift
@@ -12,6 +12,7 @@ import CoreTelephony
 import DatadogInternal
 import TestUtilities
 
+@testable import DatadogLogs
 @testable import DatadogRUM
 @testable import DatadogCrashReporting
 @testable import DatadogCore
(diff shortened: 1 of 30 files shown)


DatadogCore/Tests/Datadog/CrashReporting/CrashContext/CrashContextProviderTests.swift
ncreated, line 1: 💡 I did revamp of the way we test `CrashContextCoreProvider` to pay down tech debt taken during V2-migration. The debt was on testing the `provider` through sending messages to mocked core.

#### Full Rationale (from my Slack thread):
The way we manage CrashContext updates is super fragile 🤹. [Here](https://github.com/DataDog/dd-sdk-ios/blob/3014ba81253c6e24a7465951532bfd9945ec355d/DatadogCrashReporting/Sources/CrashContext/CrashContextProvider.swift#L100-L109) we replace the whole context on every `DatadogContext` update. SDK context updates are very frequent, meaning we do it often. To make it correct, there are few elements that require contributor to have deep understanding of following bits:
- The [custom Equatable implementation for CrashContext](https://github.com/DataDog/dd-sdk-ios/blob/3014ba81253c6e24a7465951532bfd9945ec355d/DatadogCrashReporting/Sources/CrashContext/CrashContext.swift#L138-L152) must be updated with the new value added.
- The [convenience `CrashContext.init()`](https://github.com/DataDog/dd-sdk-ios/blob/3014ba81253c6e24a7465951532bfd9945ec355d/DatadogCrashReporting/Sources/CrashContext/CrashContext.swift#L113-L118) must be updated with transporting the previous value, otherwise it will erase previous state through `DatadogContext` overwrite
- The [initial `CrashContext`](https://github.com/DataDog/dd-sdk-ios/blob/3014ba81253c6e24a7465951532bfd9945ec355d/DatadogCrashReporting/Sources/CrashContext/CrashContextProvider.swift#L48) is `nil` until the first `DatadogContext` arrives — which was not visible in previous tests as they [followed the pattern of installing it inside core mock](https://github.com/DataDog/dd-sdk-ios/blob/3014ba81253c6e24a7465951532bfd9945ec355d/DatadogCore/Tests/Datadog/CrashReporting/CrashContext/CrashContextProviderTests.swift#L63-L64) that accidentally guarantees the availability of first `DatadogContext`.

---

With the testing convention I introduced to this file, now all above criteria are covered.

DatadogCore/Tests/Datadog/CrashReporting/CrashContext/CrashContextTests.swift
```diff
@@ -64,6 +64,40 @@ class CrashContextTests: XCTestCase {
...
+
+        // Then
+        let deserializedContext = try decoder.decode(CrashContext.self, from: serializedContext)
+        DDAssertJSONEqual(
+            deserializedContext.lastRUMAttributes,
+            randomRUMAttributes
+        )
+    }
+
+    func testGivenContextWithLastLogttributesSet_whenItGetsEncoded_thenTheValueIsPreservedAfterDecoding() throws {
```
ncreated, line 84: 💡 Adding coverage to coding global Log attributes as it was missing.

DatadogRUM/Tests/RUMMonitor/Scopes/RUMSessionScopeTests.swift
```diff
@@ -361,13 +363,13 @@ class RUMSessionScopeTests: XCTestCase {
             hasTrackedAnyView: false,
             didStartWithReplay: randomIsReplayBeingRecorded
         )
-        let baggageSent = try XCTUnwrap(featureScope.messagesSent().lastBaggage(withKey: RUMBaggageKeys.sessionState))
-        let sessionStateSent: RUMSessionState = try baggageSent.decode()
-        XCTAssertEqual(sessionStateSent, expectedSessionState, "It must send 'session state' message")
+        let actualSessionState = try XCTUnwrap(fatalErrorContext.sessionState)
+        XCTAssertEqual(actualSessionState, expectedSessionState)
```
ncreated, line 367: 💡 This change in `RUMSessionScopeTests` is around asserting the update on `FatalErrorContext` instead of inspecting the baggage sent over message bus. The baggage is now tested in `FatalErrorContextTests` which streamlines things and better separates concerns.

DatadogRUM/Tests/RUMMonitor/Scopes/RUMSessionScopeTests.swift
```diff
@@ -557,60 +588,6 @@ class RUMSessionScopeTests: XCTestCase {
...
-
-        // When
-        _ = scope.process(command: RUMStopSessionCommand.mockWith(time: Date()), context: context, writer: writer)
-        let stopResourceCommand = RUMStopResourceCommand.mockWith(resourceKey: startResourceCommand.resourceKey, time: Date())
-        _ = scope.process(command: stopResourceCommand, context: context, writer: writer)
-
-        // Then
-        let viewResetMessages = featureScope.messagesSent().filter { $0.asBaggage?.key == RUMBaggageKeys.viewReset }
-        XCTAssertEqual(viewResetMessages.count, 1, "It must send only one 'view reset' message")
-    }
```
ncreated, line 612: 💡 I remove this test as it covers non-existing behaviour. Once `RUMStopSessionCommand` is received, the RUM monitor deallocates its scope, so it can no longer receive future commands (`RUMStopResourceCommand` in this case).

DatadogRUM/Tests/RUMMonitor/Scopes/RUMViewScopeTests.swift
```diff
@@ -2537,8 +2538,7 @@ class RUMViewScopeTests: XCTestCase {
 
         // Then
         let rumViewWritten = try XCTUnwrap(featureScope.eventsWritten(ofType: RUMViewEvent.self).last, "It should send view event")
-        let baggageSent = try XCTUnwrap(featureScope.messagesSent().firstBaggage(withKey: RUMBaggageKeys.viewEvent))
-        let rumViewSent: RUMViewEvent = try baggageSent.decode()
-        DDAssertReflectionEqual(rumViewSent, rumViewWritten, "It must sent written event over message bus")
+        let rumViewInFatalErrorContext = try XCTUnwrap(fatalErrorContext.view)
+        DDAssertReflectionEqual(rumViewWritten, rumViewInFatalErrorContext, "It must update fatal error context with the view event written")
```
ncreated, line 2542: 💡 This change in `RUMViewScopeTests` is around asserting the update on `FatalErrorContext` instead of inspecting the baggage sent over message bus. The baggage is now tested in `FatalErrorContextTests` which streamlines things and better separates concerns.

DatadogCrashReporting/Sources/CrashContext/CrashContextProvider.swift
```diff
@@ -155,6 +165,17 @@ extension CrashContextCoreProvider: FeatureMessageReceiver {
             }
         }
     }
+
+    private func updateRUMAttributes(with baggage: FeatureBaggage, to core: DatadogCoreProtocol) {
+        queue.async { [weak core] in
+            do {
+                self.rumAttributes = try baggage.decode(type: GlobalRUMAttributes.self)
```
maciejburda, line 172: shouldn't we weakify self as well? 🤔

DatadogInternal/Sources/Models/RUM/GlobalRUMAttributes.swift
```diff
@@ -0,0 +1,28 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+public struct GlobalRUMAttributes: Codable, PassthroughAnyCodable {
```
maciejburda, line 9: would it make sense to write extension for `[AttributeKey: AttributeValue]` instead?

maciejburda (approved):
Looks good! Great coverage!

Some minor comments

DatadogCrashReporting/Sources/CrashContext/CrashContextProvider.swift
ncreated: 👍 Good point, ✅ updated in the whole file

DatadogInternal/Sources/Models/RUM/GlobalRUMAttributes.swift
ncreated: Extending standard type would bring much less clarity to the whole concept. `GlobalRUMAttributes` is a shared type (it is defined in `DatadogInternal`), standing for contract between RUM and Crash Reporting. Whenever we step on this type in code, we know it is meant for message bus communication. Same could not be achieved with using standard ubiquitous type.

Also, the type defines two aspects that could not be supported in dictionary:
- `Codable` - so we can code it back and forth (`[String: Endocable]` can't be `Codable`);
- `PassthroughAnyCodable` - to avoid marshalling the type when passing it over message bus ([`String: Encodable]` can't be managed this way as we use it transitively in multiple other types exchanged over MB, e.g. for encoding user attributes in RUM view event);

DatadogRUM/Sources/RUMMonitor/Scopes/FatalErrorContextNotifier.swift
```diff
@@ -7,9 +7,22 @@
...
+    var view: RUMViewEvent? { set get }
+    /// The current set of RUM attributes.
+    /// It gets updated with global attributes being added or removed in `RUMMonitor`.
+    var globalAttributes: [String: Encodable] { set get }
+}
+
+/// Manages RUM information necessary for building context of fatal errors such as Crashes or Fatal App Hangs.
 /// It tracks value changes and notifies updates on message bus.
-internal final class FatalErrorContextNotifier {
+internal final class FatalErrorContextNotifier: FatalErrorContextNotifying {
```
maxep, line 25: /suggestion I think the names should reflect responsibility (interface) vs. behavior (imp). Here it would sending the context vs. sending on the bus.
```swift
internal final class FatalErrorContextBus: FatalErrorContextSender
```

maxep (approved):
It looks great! Thanks for tackling the debt, it is way bette now.
I have left a small suggestion on naming but it can be merged as is.

DatadogRUM/Sources/RUMMonitor/Scopes/FatalErrorContextNotifier.swift
ncreated: It's true for strategy pattern, but is not the case here. The protocol is made purely for dependency injection and testability. It is not for introducing multiple strategies of sending `FatalErrorContext`. 

If at any point we want to notify FEC through different implementations, then I'll be 100% on the side of this suggestion. For now I'll keep it as it is 🙂.

PR #1835, opened 2024-05-14
RUM-3535 feat: add support for trace context injection configuration to allow selective injection

ganeshnj:
### What and why?

RFC https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3529509193

### How?

There is one big difference has been in this PR along with the implementation of the RFC.  If we have a distributed span which has a parent, the parent's sampling strategy will be applied to the distributed span.

Now if the parent span is not sampled, the distributed span will not be sampled regardless of the distributed sampling configuration and vice versa.

This is similar to the Trace Agent behavior where the decision stays consistent across the board on the spans. This way customers will not have a broken trace ever.

### Review checklist

- [x] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


maxep (dismissed):
Looks great, very clear!

DatadogRUM/Sources/RUMConfiguration.swift
```diff
@@ -294,13 +294,23 @@ extension RUM.Configuration.URLSessionTracking {
         /// - Parameters:
         ///   - hosts: The set of hosts to inject tracing headers. Note: Hosts must not include the "http(s)://" prefix.
         ///   - sampleRate: The sampling rate for tracing. Must be a value between `0.0` and `100.0`. Default: `20`.
-        case trace(hosts: Set<String>, sampleRate: Float = 20)
+        ///   - traceControlInjection: The strategy for injecting trace context into requests. Default: `.all`.
+        case trace(
+            hosts: Set<String>,
+            sampleRate: Float = 20,
+            traceControlInjection: TraceContextInjection = .all
```
ncreated, line 301: question/ I wonder if the `TraceContextInjection` type shouldn't be re-exported through `DatadogRUM` [like we do for other internal types](https://github.com/DataDog/dd-sdk-ios/blob/c044ef9c69369a44c772d395c838d8c6951c2ab3/DatadogRUM/Sources/RUMConfiguration.swift#L15). My feeling is we don't need it unless someone wants to define a property of this type in their code, which is not the case here. For extra check, we can call this API in [smoke tests](https://github.com/DataDog/dd-sdk-ios/blob/c044ef9c69369a44c772d395c838d8c6951c2ab3/dependency-manager-tests/spm/App/ViewController.swift#L20) to see if it works without imoprting `DatadogInternal`.

The same applies to [re-exporting types in `DatadogTrace`](https://github.com/DataDog/dd-sdk-ios/blob/c044ef9c69369a44c772d395c838d8c6951c2ab3/DatadogTrace/Sources/TraceConfiguration.swift#L11-L21).

DatadogInternal/Tests/NetworkInstrumentation/HTTPHeadersWriterTests.swift
```diff
@@ -28,7 +28,7 @@ class HTTPHeadersWriterTests: XCTestCase {
     }
 
     func testWritingDroppedTraceContext_withAutoSamplingStrategy() {
```
ncreated, line 30: suggestion/ I think the `*AutoSamplingStrategy*` is a leftover after renaming `samplingStrategy` from `.auto` to `.headBased`. Could we update this name in all tests, so it is correct? 🏕️ 🧹

DatadogInternal/Tests/NetworkInstrumentation/W3CHTTPHeadersReaderTests.swift
ncreated, line 1: change-request/ I only see updates to existing tests with `.all` injection, but no new tests added for `.sampled`. Aren't we missing enough coverage for new variants?

DatadogInternal/Tests/NetworkInstrumentation/W3CHTTPHeadersWriterTests.swift
ncreated, line 1: change-request/ Similar to `W3C*ReaderTests`, shall we add coverage to new variants with using `.sampled` injection?

ncreated (changes requested):
Looks good and reads really well 🎯 👍

My only concern is on added tests coverage - it seems we updated existing unit tests for Header Writers and Readers 👍, but we haven't added coverage for new permutations of their `init()`, no?

DatadogRUM/Sources/RUMConfiguration.swift
ganeshnj: actually we should, else customer have to import internal. Good call.

DatadogInternal/Tests/NetworkInstrumentation/HTTPHeadersWriterTests.swift
ganeshnj: sure

DatadogInternal/Tests/NetworkInstrumentation/W3CHTTPHeadersWriterTests.swift
ganeshnj: Good call.

DatadogInternal/Tests/NetworkInstrumentation/W3CHTTPHeadersReaderTests.swift
ganeshnj: Good call.

ncreated (dismissed):
🎯

maciejburda (dismissed):
LGTM

maxep (approved):

maxep (approved):

PR #1798, opened 2024-04-25
RUM-3574 perf: Start sending batches immediately after feature is initialized

ncreated:
### What and why?

📦⏱️ This PR removes the initial delay for sending batches. Now, first upload will be triggered soon after feature is initialized.

The motivation is to maximise a chance of sending data if application is impacted by early crash.
Related to https://github.com/DataDog/dd-sdk-flutter/issues/573

### How?

Removing initial delay from `DataUploadWorker`.

Before:
```
----[I]------------[U1]----------------[U2]-->
     ^ SDK init
                    ^ initial upload
                                         ^ next upload
```

Now:
```
----[I][U1]----------------[U2]-->
     ^ SDK init
        ^ initial upload
                             ^ next upload
```

### Review checklist
- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference
- [x] Add CHANGELOG entry for user facing changes

### Custom CI job configuration (optional)
- [ ] Run unit tests for Core, RUM, Trace, Logs, CR and WVT
- [ ] Run unit tests for Session Replay
- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

PR #1840, opened 2024-05-15
fix: Remove flaky assertion due to upload at init

maxep:
### What and why?

The upload at init could pick files in the authorized folder, so we cannot guarantee that the files will be there while running the test.

The assertion can be removed as we only need to assert that the `clearAllData` removes all files.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


ncreated (approved):
Looks good 👍 

Mentioning https://github.com/DataDog/dd-sdk-ios/pull/1798 for GH link reference.

PR #1843, opened 2024-05-17
RUM-3183 fix: Fix crash on accessing `request.allHTTPHeaderFields`

ncreated:
### What and why?

📦 🧯 Fixes https://github.com/DataDog/dd-sdk-ios/issues/1638

As we observed in https://github.com/DataDog/dd-sdk-ios/issues/1638, accessing `request.allHTTPHeaderFields` is not safe and leads to crashes with undefined root cause. To avoid this, instead we use `request.value(forHTTPHeaderField:)` to only read headers known by the SDK.

### How?

Removed usage of `.allHTTPHeaderFields` by replacing it with higher-level `.value(forHTTPHeaderField:)` API.

I also added linter rule to avoid using `allHTTPHeaderFields` if possible.

#### Suspected problem with `.allHTTPHeaderFields` safety

I failed to reproduce the crash, although by exploring `swift-corelibs-foundation` code, I suspect it to relate to the Obj-c → Swift bridging, somewhere in [this code](https://github.com/apple/swift-corelibs-foundation/blob/06072877b110c4ba5501ddc012830c1072f7198c/Sources/FoundationNetworking/URLRequest.swift#L285-L309). Symbols ran during the bridging correspond to some reported in https://github.com/DataDog/dd-sdk-ios/issues/1638 stack traces. This may invalidate our earlier assumption on the problem being caused by lack of thread safety in `.allHTTPHeaderFields` implementation.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


DatadogInternal/Sources/NetworkInstrumentation/URLSession/ImmutableRequest.swift
```diff
@@ -0,0 +1,44 @@
...
+
+    public init(request: URLRequest) {
+        self.url = request.url
+        self.httpMethod = request.httpMethod
+
+        // As observed in https://github.com/DataDog/dd-sdk-ios/issues/1638, accessing `request.allHTTPHeaderFields` is not
+        // safe and leads to crashes with undefined root cause. To avoid this, instead we use `request.value(forHTTPHeaderField:)`
+        // to only read headers known by the SDK.
+        var knownHTTPHeaderFields: [String: String] = [:]
+        addHeaderIfExists(request: request, field: TracingHTTPHeaders.originField, to: &knownHTTPHeaderFields)
```
ncreated, line 33: 💡 We could consider adding other known headers here, notably the ones related to distributed tracing (`x-datadog-*`, W3C and B3). Not doing this as these headers are not required for upstream code. The only reason for `allHTTPHeaderFields` existence, was accessing the `x-datadog-origin` header which is added here.

Alternatively, we could replace the entire `knownHTTPHeaderFields` prop with `datadogOriginField`. This way we will move even further away from the unsafe consideration of "all headers".

Open for discussion 🧠.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/ImmutableRequest.swift
```diff
@@ -0,0 +1,44 @@
...
+        self.knownHTTPHeaderFields = knownHTTPHeaderFields
+        self.unsafeOriginal = request
+    }
+}
+
+private func addHeaderIfExists(request: URLRequest, field: String, to knownHeaders: inout [String: String]) {
+    if let value = request.value(forHTTPHeaderField: field) {
+        knownHeaders[field] = value
+    }
+}
```
ganeshnj, line 44: /question

would be cleaner to not use `inout` ? Or does it have any benefit? Personal opinion but pure functions are usually less ambiguous.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/ImmutableRequest.swift
ncreated: > Can we copy all headers for instance?

I don't see any option without using the problematic `allHTTPHeaderFields` API.

> This seems risky, by only looking for some headers. We will be blind where knownHTTPHeaderFields is actually used.

Makes sense 👍. I changed it to the alternative approach by replacing `knownHTTPHeaderFields` with `ddOriginHeaderValue` prop.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/ImmutableRequest.swift
ncreated: n/a after change to `ddOriginHeaderValue`

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionInterceptor.swift
```diff
@@ -98,7 +98,7 @@ public struct URLSessionInterceptor {
     // MARK: - Private
 
     private func extractTraceID(from request: URLRequest) -> TraceID? {
-        guard let headers = request.allHTTPHeaderFields else {
+        guard let headers = request.allHTTPHeaderFields else { // swiftlint:disable:this unsafe_all_http_header_fields
```
ncreated, line 101: 💡 It is IMO fine to ignore this warning here as `URLSessionInterceptor` is only used by Alamofire extension, which should be deprecated and removed. We have no crash reports pointing this usage, hence this should be safe.

Alternatively, we could use `value(forHTTPHeaderField:)` to read (one by one) all headers required by DD, B3 and W3C header readers in following lines.

maxep: I prefer the alternative mentioned, it's clearer 👍 thanks for the change :)

tools/lint/sources.swiftlint.yml
```diff
@@ -89,6 +89,15 @@ custom_rules:
...
     message: "`UIApplication.shared` is unavailable in some environments. Check `UIApplication.managedShared`."
     severity: error
+  unsafe_all_http_header_fields: # prevents from using `URLRequest.allHTTPHeaderFields` API
+    included: Sources
+    name: "Unsafe API: `URLRequest.allHTTPHeaderFields`"
+    regex: '\.allHTTPHeaderFields'
+    excluded_match_kinds: 
+      - comment
+      - doccomment
+    message: "`URLRequest.allHTTPHeaderFields` is considered unsafe. Prefer using `URLRequest.value(forHTTPHeaderField:)`. See full context in https://github.com/DataDog/dd-sdk-ios/pull/1843"
```
maxep, line 99: 🥇

DatadogInternal/Sources/NetworkInstrumentation/URLSession/ImmutableRequest.swift
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+import Foundation
+
+/// An immutable version of `URLRequest`.
+///
+/// Introduced in response to concerns raised in https://github.com/DataDog/dd-sdk-ios/issues/1638
+/// it makes a copy of request attributes, safeguarding against potential thread safety issues arising from concurrent
+/// mutations (see more context in https://github.com/DataDog/dd-sdk-ios/pull/1767 ).
+public struct ImmutableRequest {
```
maxep, line 14: /suggestion I guess this could be merged with `URLSessionTaskInterception`, or an inner type.

maxep (approved):
🤞

DatadogInternal/Sources/NetworkInstrumentation/URLSession/ImmutableRequest.swift
ncreated: In this PR we just move this type to separate file (it was added earlier). While merging types is simple in code, updating tests is harder as it adds more permutations to cover. For the sake of isolated testing, let's keep it this way.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/ImmutableRequest.swift
ganeshnj: This is those places where I prefer using assertions (debug only) but not sure how customer can selectively disable them for libraries only else it is too much of noise. 

I'm okay.

ganeshnj (approved):

maxep (approved):

PR #1851, opened 2024-05-20
RUM-347 fix: Fix `batch_duration` in "Batch Closed" metric

ncreated:
### What and why?

📦 🔭 This PR fixes the wrong value reported in `batch_duration` for "Batch Closed" metrics.

For 1w data, Android SDK reports it as 1.41s in p50 and 9.42s in p90, whereas for iOS SDK it is respectively ~30s and ~3min. 

### How?

Now we measure `batch_duration` in the same way [as Android SDK does](https://github.com/DataDog/dd-sdk-android/blob/a0e05c37a8ec3ee78415e082fb18f8e3a10668e2/dd-sdk-android-core/src/main/kotlin/com/datadog/android/core/internal/persistence/file/batch/BatchFileOrchestrator.kt#L49C17-L49C40) - by marking the last writable access to the batch and computing `batch_duration` as the duration from batch creation to last access.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogCore/Tests/Datadog/Core/Persistence/FilesOrchestratorTests.swift
+++ b/DatadogCore/Tests/Datadog/Core/Persistence/FilesOrchestratorTests.swift
@@ -55,17 +55,20 @@ class FilesOrchestratorTests: XCTestCase {
     }
 
     func testWhenSameWritableFileWasUsedMaxNumberOfTimes_itCreatesNewFile() throws {
-        let orchestrator = configureOrchestrator(using: RelativeDateProvider(advancingBySeconds: 0.001))
+        let dateProvider = DateProviderMock()
+        let orchestrator = configureOrchestrator(using: dateProvider)
         var previousFile: WritableFile = try orchestrator.getWritableFile(writeSize: 1) // first use of a new file
         var nextFile: WritableFile
 
         for _ in (0..<5) {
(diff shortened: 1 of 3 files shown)


DatadogCore/Tests/Datadog/Core/Persistence/FilesOrchestratorTests.swift
```diff
@@ -55,17 +55,20 @@ class FilesOrchestratorTests: XCTestCase {
     }
 
     func testWhenSameWritableFileWasUsedMaxNumberOfTimes_itCreatesNewFile() throws {
-        let orchestrator = configureOrchestrator(using: RelativeDateProvider(advancingBySeconds: 0.001))
+        let dateProvider = DateProviderMock()
+        let orchestrator = configureOrchestrator(using: dateProvider)
```
ncreated, line 59: 💡 This test started being flaky after adding more calls to `dateProvider.now` in files orchestrator. It is because we were using relative date provider mock that advances the date after each call to `.now`. In general, we should avoid using `RelativeDateProvider` and be more specific about the "now" value required in each test.

maciejburda (approved):

maxep (approved):
Thanks 🙏

PR #1857, opened 2024-05-22
refactor: Standardize locating SDK metrics code

ncreated:
### What and why?

📦 ✏️ This small refactoring PR aims at standardizing the way of placing SDK metrics code in our codebase.

With existing "Batch Closed", "Batch Deleted" and "Method Called" metrics and before adding "RUM Session Ended" it is a good moment to re-think how we locate SDK metrics. The goal is to introduce minimal convention that can be followed uniformly, so it is easier to navigate through existing metrics (adding / removing / updating).

### How?

This PR moves metric-related files around and applies minimal renaming.

Proposed convention is to store metric-related code in `SDKMetrics` folder that can be defined in each module:
```
DatadogCore/
   → SDKMetrics/
      → BatchMetrics.swift // ("Batch Closed" and "Batch Deleted" metric definitions)
DatadogInternal/
   → SDKMetrics/
      → SDKMetricFields.swift // (common metric fields)
      → MethodCalledMetric.swift // ("Method Called" metric definition)
```

For the purpose of "RUM Session Ended" it will be extended to:
```
DatadogRUM/
   → SDKMetrics/
      → RUMSessionEndedMetric.swift // (metric definition)
```

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


maciejburda (approved):
Nice!

At some point might be worth it to create structs for those metrics with manually implemented `.asAttributes()` function (to avoid encoding/decoding). This way it will be simpler to implement new models in the future.

ncreated:
> At some point might be worth it to create structs for those metrics with manually implemented `.asAttributes()` function (to avoid encoding/decoding). This way it will be simpler to implement new models in the future.

👍 Definitely 💯. A common "metric" type will help to standardize SDK metrics even more (and keep the whole concept type safe). It might be something to do in the future - likely through RFC to avoid 🏓. In this PR I only wanted to achieve minimal unification of the file placement convention.

PR #1863, opened 2024-05-27
RUM-917 Improve telemetry of type mismatch

maciejburda:
### What and why?

Adds extra telemetry info to to help resolving type mismatch error in Session Replay.

### How?

Expands error type to contain from and to types of the wireframe that conflicted.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogSessionReplay/Sources/Processor/Diffing/Diff+SRWireframes.swift
+++ b/DatadogSessionReplay/Sources/Processor/Diffing/Diff+SRWireframes.swift
@@ -41,6 +41,21 @@ extension SRWireframe: Diffable {
             return true
         }
     }
+
+    var type: String {
+        switch self {
+        case let (.shapeWireframe(value)):
+            return value.type
+        case let (.textWireframe(value)):
+            return value.type
+        case let (.imageWireframe(value)):
+            return value.type
+        case let (.placeholderWireframe(value)):
+            return value.type
+        case let (.webviewWireframe(value)):
+            return value.type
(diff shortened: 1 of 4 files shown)


DatadogSessionReplay/Sources/Processor/Diffing/Diff+SRWireframes.swift
```diff
@@ -41,17 +41,32 @@ extension SRWireframe: Diffable {
...
+        }
+    }
 }
 
 // MARK: - Resolving Mutations
 
 internal typealias WireframeMutation = SRIncrementalSnapshotRecord.Data.MutationData.Updates
 
-internal enum WireframeMutationError: Error {
+internal enum WireframeMutationError: Error, Equatable {
```
mariedm, line 65: I believe this enum is automatically conforming to `Equatable` since the associated values you added are `String`. I don't think it's an issue to add it explicitly, but I was wondering if I missed anything here?

mariedm (approved):

maxep (approved):

DatadogSessionReplay/Sources/Processor/Diffing/Diff+SRWireframes.swift
maciejburda: Assertions in test were complaining about it. I think that once enum has associated values it needs to be explicitly specified

PR #1866, opened 2024-05-29
RUM-1660 chore: Send "RUM Session Ended" telemetry

ncreated:
### What and why?

📦 🔭 Following [the spec](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3663167927/RUM+Session+Ended+Telemetry+-+Spec) (internal), this PR adds "RUM Session Ended" telemetry.

The "RUM Session Ended" telemetry marks the conclusion of a RUM session and provides meta parameters that describe the session's quality and diagnostic details.

### How?

The goal of this implementation is to better understand the possible states of a RUM session by making minimal assumptions about how sessions are tracked in RUM scopes. For instance, it can track multiple sessions simultaneously, which, though unlikely, is observed in some telemetry data. The design aims to be loosely coupled with the RUM architecture, allowing flexible and convenient access to the metric object rather than restricting its lifecycle to RUM scopes.

The high-level design of this telemetry consists of two main components:
- `SessionEndedMetricController`: Manages the lifecycle of metric objects.
- `SessionEndedMetric`: Tracks the state of a RUM session and exports metric attributes.

Both are thoroughly covered in unit tests. Integration tests ensure the proper integration of the metric with RUM.

The interface of `SessionEndedMetricController`:
```swift
internal final class SessionEndedMetricController {
    // Starts a new metric for a given session.
    func startMetric(sessionID: RUMUUID, precondition: RUMSessionPrecondition?, context: DatadogContext)

    // 3 APIs tracking session state for given session ID (or `nil` for latest session):
    func track(view: RUMViewEvent, in sessionID: RUMUUID?)
    func track(sdkErrorKind: String, in sessionID: RUMUUID?)
    func trackWasStopped(sessionID: RUMUUID?)

    // Ends the metric for a given session, sending it to telemetry and removing it from pending metrics.
    func endMetric(sessionID: String)
}
```
The interface of `SessionEndedMetric`:
```swift
internal final struct SessionEndedMetric {
    mutating func track(view: RUMViewEvent)
    mutating func track(sdkErrorKind: String)
    mutating func trackWasStopped()
    func asMetricAttributes() -> [String: Encodable]
}
```

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/Datadog/Datadog.xcodeproj/project.pbxproj
+++ b/Datadog/Datadog.xcodeproj/project.pbxproj
@@ -533,6 +533,8 @@
 		61C363802436164B00C4D4E6 /* ObjcExceptionHandlerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61C3637F2436164B00C4D4E6 /* ObjcExceptionHandlerTests.swift */; };
 		61C3638324361BE200C4D4E6 /* DatadogPrivateMocks.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61C3638224361BE200C4D4E6 /* DatadogPrivateMocks.swift */; };
 		61C3638524361E9200C4D4E6 /* Globals.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61C3638424361E9200C4D4E6 /* Globals.swift */; };
+		61C4534A2C0A0BBF00CC4C17 /* TelemetryInterceptorTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61C453492C0A0BBF00CC4C17 /* TelemetryInterceptorTests.swift */; };
+		61C4534B2C0A0BBF00CC4C17 /* TelemetryInterceptorTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61C453492C0A0BBF00CC4C17 /* TelemetryInterceptorTests.swift */; };
 		61C5A89624509BF600DA608C /* TracerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61C5A89524509BF600DA608C /* TracerTests.swift */; };
 		61C713A32A3B78F900FA735A /* RUMMonitorProtocol.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61C713A02A3B78F900FA735A /* RUMMonitorProtocol.swift */; };
 		61C713A42A3B78F900FA735A /* RUMMonitorProtocol.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61C713A02A3B78F900FA735A /* RUMMonitorProtocol.swift */; };
@@ -2542,6 +2544,7 @@
 		61C3E63824BF19B4008053F2 /* RUMContext.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = RUMContext.swift; sourceTree = "<group>"; };
 		61C3E63A24BF1A4B008053F2 /* RUMCommand.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = RUMCommand.swift; sourceTree = "<group>"; };
 		61C3E63D24BF1B91008053F2 /* RUMApplicationScope.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = RUMApplicationScope.swift; sourceTree = "<group>"; };
+		61C453492C0A0BBF00CC4C17 /* TelemetryInterceptorTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = TelemetryInterceptorTests.swift; sourceTree = "<group>"; };
 		61C5A87824509A0C00DA608C /* DDSpan.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = DDSpan.swift; sourceTree = "<group>"; };
 		61C5A87924509A0C00DA608C /* DDNoOps.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = DDNoOps.swift; sourceTree = "<group>"; };
 		61C5A87C24509A0C00DA608C /* Casting.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = Casting.swift; sourceTree = "<group>"; };
@@ -5846,6 +5849,7 @@
 				D21C26ED28AFB65B005DD405 /* ErrorMessageReceiverTests.swift */,
 				9E53889B2773C4B300A7DC42 /* WebViewEventReceiverTests.swift */,
 				D248ED4728081B9B00B315B4 /* TelemetryReceiverTests.swift */,
+				61C453492C0A0BBF00CC4C17 /* TelemetryInterceptorTests.swift */,
 			);
 			path = Integrations;
 			sourceTree = "<group>";
@@ -8693,6 +8697,7 @@
 				6188697D2A4376F700E8996B /* RUMConfigurationTests.swift in Sources */,
 				61DCC8482C05CD0000CB59E5 /* SessionEndedMetricControllerTests.swift in Sources */,
 				D23F8EA029DDCD38001CFAE8 /* RUMOffViewEventsHandlingRuleTests.swift in Sources */,
+				61C4534B2C0A0BBF00CC4C17 /* TelemetryInterceptorTests.swift in Sources */,
 				D23F8EA229DDCD38001CFAE8 /* RUMSessionScopeTests.swift in Sources */,
 				D23F8EA329DDCD38001CFAE8 /* RUMUserActionScopeTests.swift in Sources */,
 				615B0F8C2BB33C2800E9ED6C /* AppHangsMonitorTests.swift in Sources */,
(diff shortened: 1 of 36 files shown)


DatadogRUM/Sources/SDKMetrics/SessionEndedMetric.swift
```diff
@@ -0,0 +1,251 @@
...
+    ) {
+        self.sessionID = sessionID
+        self.bundleType = context.applicationBundleType
+        self.precondition = precondition
+    }
+
+    /// Tracks the view event that occurred during the session.
+    func track(view: RUMViewEvent) {
+        guard view.session.id == sessionID else {
+            return // sanity check, unexpected
```
maciejburda, line 87: maybe worth telemetry?

DatadogRUM/Sources/SDKMetrics/SessionEndedMetricController.swift
```diff
@@ -0,0 +1,65 @@
...
+import DatadogInternal
+
+/// A controller responsible for managing "RUM Session Ended" metrics.
+internal final class SessionEndedMetricController {
+    /// Dictionary to keep track of pending metrics, keyed by session ID.
+    @ReadWriteLock
+    private var metricsBySessionID: [String: SessionEndedMetric] = [:]
+    /// Array to keep track of pending metrics in their start order.
+    @ReadWriteLock
+    private var metrics: [SessionEndedMetric] = []
```
maxep, line 17: /discussion Mutating a metric property will acquire at least 2 locks, 1 in the controller and one in the metric. The perf penalty will be minimal because we don't expect many concurrent calls. But I want to highlight that the `@ReadWriteLock` wrapper uses a fair pthread lock which is not cheap. 

Moreover, the `metric.track(*_:)` methods are acquiring the property lock multiple times, which could be prevented by calling the [`mutate`](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogInternal/Sources/Concurrency/ReadWriteLock.swift#L49) closure.

I think it could be redesigned to use a single lock in the controller and perform all mutations at the controller level, something like the following where `SessionEndedMetric` could be a `struct`:

```swift
internal final class SessionEndedMetricController {
    // Starts a new metric for a given session.
    func startMetric(sessionID: String, precondition: RUMSessionPrecondition?, context: DatadogContext)

    // Tracks the view event that occurred during the session.
    func track(view: RUMViewEvent, for sessionID: String)

    // Tracks the kind of SDK error that occurred during the session.
    func track(sdkErrorKind: String, for sessionID: String)

    // Signals that the session was stopped with `stopSession()` API.
    func trackWasStopped(for sessionID: String)

    // Ends the metric for a given session, sending it to telemetry and removing it from pending metrics.
    func endMetric(sessionID: String)

    // Retrieves the last started metric.
    var latestMetric: SessionEndedMetric? { get }
}
```
What do you think?

Note: the `@ReadWriteLock` is convenient but shouldn't be our only lock. We could provide a lock object as well (similar to NSLock) so we can acquire a lock for operations and not only for property access. We could also provide unfair lock such as `os_unfair_lock` which are cheap.

maxep (commented):
Looks great, thanks for the test coverage!
I have left a comment re. usage of locks, LMKWYT.

DatadogRUM/Sources/SDKMetrics/SessionEndedMetricController.swift
ncreated: 👍 I had this concern on double locking, but ignored it with assuming minimal impact and trying to keep the solution simple. With this feedback I'm happy to improve this 👌.

DatadogRUM/Sources/Integrations/TelemetryInterecptor.swift
```diff
@@ -0,0 +1,33 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+import DatadogInternal
+
+/// Intercepts telemetry events sent through message bus.
+internal struct TelemetryInterecptor: FeatureMessageReceiver {
```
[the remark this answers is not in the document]
ncreated: I'd say it always starts simple at the beginning. The `TelemetryReceiver` already covers multiple concerns, I didn't want to introduce another one. For that reason I've chosen composition and I'd prefer to stay with this unless this is a change request.

While it can be easily moved to `TelemetryReceiver`, adding unit tests will be way harder as it operates on much higher abstractions than the `TelemetryInterceptor` we introduce here. LTMKWDYT @maxep

DatadogRUM/Sources/SDKMetrics/SessionEndedMetric.swift
ncreated: Indeed 👍. I added an error so we will see if views are tracked in wrong sessions (although unexpected, it is possible from upstream code).

DatadogRUM/Sources/SDKMetrics/SessionEndedMetricController.swift
ncreated: ✅ Addressed 👍. I made the `SessionEndedMetric` a `struct` with mutable methods. The controller is now a class which mutates the right `SEM` using single RW lock 👍.

DatadogRUM/Sources/SDKMetrics/SessionEndedMetricController.swift
```diff
@@ -0,0 +1,88 @@
...
+        guard let metric = metricsBySessionID[sessionID] else {
+            return
+        }
+        telemetry.metric(name: SessionEndedMetric.Constants.name, attributes: metric.asMetricAttributes())
+        metricsBySessionID[sessionID] = nil
+        pendingSessionIDs.removeAll(where: { $0 == sessionID }) // O(n), but "ending the metric" is very rare event
+    }
+
+    private func updateMetric(for sessionID: RUMUUID?, _ mutation: (inout SessionEndedMetric?) throws -> Void) {
+        guard let sessionID = (sessionID ?? pendingSessionIDs.last) else {
```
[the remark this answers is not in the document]
ncreated: @ambushwork The `pendingSessionIDs` is more a history of pending sessions. It is true that we query only for the last element, but it might change from one query to another if `endMetric(sessionID:)` is called in between (we remove the ID from array as part of that call). If using single property to store the "last ID", we won't know how to change it after `endMetric(sessionID:)` as we wouldn't know the previous ID.

Also, even though sessions are expected to end in the same order as they were started, some iOS telemetry shows rare but existing consistency problems in session management. For that reason, we prefer to not introduce too many assumptions in "Session Ended" metric, which is another reason why we track the whole array of IDs rather than only the "latest" one.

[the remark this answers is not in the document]
ambushwork: Thanks, quite clear for me now !

DatadogRUM/Sources/RUMMonitor/Scopes/RUMApplicationScope.swift
```diff
@@ -104,6 +105,10 @@ internal class RUMApplicationScope: RUMScope, RUMContextProvider {
                 return scope
             }
 
+            // proccss(command:context:writer) returned false, so the scope will be deallocated at the end of
+            // this execution context. End the "RUM Session Ended" metric:
+            defer { dependencies.sessionEndedMetric.endMetric(sessionID: scope.sessionUUID) }
+
```
ambushwork, line 111: If I understand correctly, the `endMetric` can be triggered only when there is a command to process and it returns false, otherwise it can not be triggered it self even if the duration and timeout has passed the limit?

ncreated: That is correct 👍. Every state change in RUM is caused by command. Even in case of "15min timeout" or "4h max duration", the `RUMApplicationScope` only recognizes these upon receiving next command.

DatadogRUM/Sources/Integrations/TelemetryInterecptor.swift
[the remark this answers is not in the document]
maxep: Yes, I don't have strong opinion on this.
But the typos still persist ;)

DatadogRUM/Sources/SDKMetrics/SessionEndedMetricController.swift
```diff
@@ -0,0 +1,88 @@
...
+import DatadogInternal
+
+/// A controller responsible for managing "RUM Session Ended" metrics.
+internal final class SessionEndedMetricController {
+    /// Dictionary to keep track of pending metrics, keyed by session ID.
+    @ReadWriteLock
+    private var metricsBySessionID: [RUMUUID: SessionEndedMetric] = [:]
+    /// Array to keep track of pending session IDs in their start order.
+    @ReadWriteLock
+    private var pendingSessionIDs: [RUMUUID] = []
```
maxep, line 17: /suggestion As this is private, and tightly linked to `metricsBySessionID`, it can be mutated while acquiring the lock of `metricsBySessionID`. Starting a metric would look like:
```swift
_metricsBySessionID.mutate { metrics in
    metrics[sessionID] = SessionEndedMetric(sessionID: sessionID, precondition: precondition, context: context)
    pendingSessionIDs.append(sessionID)
}
```

maxep (approved):
Looks great! I left a minor suggestion on removing the extra lock on the `pendingSessionIDs` property.

DatadogRUM/Sources/Integrations/TelemetryInterecptor.swift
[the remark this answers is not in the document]
ncreated: > But the typos still persist ;)

Ops! Fixed ✅

DatadogRUM/Sources/SDKMetrics/SessionEndedMetricController.swift
ncreated: > As this is private and tightly linked (...)

Fair call, done ✅

maxep (approved):
Well done, it looks great!!

Datadog/Datadog.xcodeproj/project.pbxproj
```diff
@@ -8987,7 +8987,7 @@
 				D29A9F8929DD85BB005C54A4 /* VitalRefreshRateReader.swift in Sources */,
 				D29A9F6929DD85BB005C54A4 /* UIKitRUMUserActionsHandler.swift in Sources */,
 				D29A9F5229DD85BB005C54A4 /* RUMUUIDGenerator.swift in Sources */,
-				61DCC84E2C071DCD00CB59E5 /* TelemetryInterecptor.swift in Sources */,
+				61DCC84E2C071DCD00CB59E5 /* TelemetryInterceptor.swift in Sources */,
```
maciejburda, line 8990: 🎖️

maciejburda (approved):

PR #1876, opened 2024-06-03
Fix GH asset otel validation

maxep:
### What and why?

Validate `OpenTelemetryApi.xcframework` during distribution.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/tools/distribution/src/release/assets/gh_asset.py
+++ b/tools/distribution/src/release/assets/gh_asset.py
@@ -167,7 +167,7 @@ class CrashReporterXCFrameworkValidator(XCFrameworkValidator):
         if in_version < min_cr_version:
             return False # Datadog Crash Reporting.xcframework was introduced in `1.7.0`
 
-        dir = zip_directory.get('CrashReporter.xcframework')
+        dir = zip_directory.get(self.name)
 
         min_xc14_version = Version('1.12.1')
         if in_version >= min_xc14_version:
@@ -204,7 +204,7 @@ class KronosXCFrameworkValidator(XCFrameworkValidator):
         if in_version < min_version or in_version >= max_version:
             return False
             
-        zip_directory.get('Kronos.xcframework').assert_it_has_files([
+        zip_directory.get(self.name).assert_it_has_files([
             'ios-arm64_arm64e',
             'ios-arm64_arm64e/dSYMs/*.dSYM',
             'ios-arm64_arm64e/**/*.swiftinterface',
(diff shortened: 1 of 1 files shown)


tools/distribution/src/release/assets/gh_asset.py
mariedm, line 1: Is this script executed automatically as part of a CI/CD pipeline or other workflow?

tools/distribution/src/release/assets/gh_asset.py
```diff
@@ -167,7 +167,7 @@ def validate(self, zip_directory: DirectoryMatcher, in_version: Version) -> bool
         if in_version < min_cr_version:
             return False # Datadog Crash Reporting.xcframework was introduced in `1.7.0`
 
-        dir = zip_directory.get('CrashReporter.xcframework')
+        dir = zip_directory.get(self.name)
```
mariedm, line 170: Is there a reason for not doing this change in other `validate` functions?

tools/distribution/src/release/assets/gh_asset.py
```diff
@@ -251,6 +251,33 @@ def validate(self, zip_directory: DirectoryMatcher, in_version: Version) -> bool
             ])
 
         return True
+    
+class OpenTelemetryXCFrameworkValidator(XCFrameworkValidator):
+    name = 'OpenTelemetryApi.xcframework'
+
+    def validate(self, zip_directory: DirectoryMatcher, in_version: Version) -> bool:
+        min_otel_version = Version('2.12.0')
```
mariedm, line 259: nit: this could be added at the top of the file with the other version requirements (though it's only used by this function for now)

mariedm (approved):

tools/distribution/src/release/assets/gh_asset.py
maxep: It's part of CD pipeline and executed when distributing a new tag 👍

tools/distribution/src/release/assets/gh_asset.py
maxep: That's a miss, let me update all of them 👍

tools/distribution/src/release/assets/gh_asset.py
maxep: This min version only applies to otel, so it can be contained in `OpenTelemetryXCFrameworkValidator`. We can make it global when we need to share it 👍

mariedm (approved):

ncreated (approved):

PR #1885, opened 2024-06-06
RUM-817 [Session Replay] Handle Objective-C Runtime Exception

maxep:
### What and why?

Recover from Objective-C runtime exception while recoding snapshot.

### How?

1. Define abstraction in `DatadogInternal` to register an Objective-C exception handler.
2. Register the Objective-C exception handler from `DatadogCore`.
3. Catch any Objective-C exception while recording a snapshot.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


DatadogCore/Sources/Core/DatadogCore.swift
```diff
@@ -491,3 +491,12 @@ extension DatadogCore: Flushable {
         }
     }
 }
+
+#if SPM_BUILD
+import DatadogPrivate
+#endif
+
+internal let registerObjcExceptionHandlerOnce: () -> Void = {
+    ObjcException.rethrow = __dd_private_ObjcExceptionHandler.rethrow
```
mariedm, line 500: /question What I understand is that `ObjcException.rethrow` is set internally and not intended for customers to use, but is it safe? Is there a way to validate what is assigned to `ObjcException.rethrow` to ensure it is an Objective-C compatible function? It might over-complicate the current code for a minimal gain, but I'm curious.

DatadogCore/Sources/Core/DatadogCore.swift
```diff
@@ -491,3 +491,12 @@ extension DatadogCore: Flushable {
         }
     }
 }
+
+#if SPM_BUILD
+import DatadogPrivate
+#endif
+
+internal let registerObjcExceptionHandlerOnce: () -> Void = {
```
mariedm, line 499: /question Do we need to make sure that `registerObjcExceptionHandlerOnce` is called only once?

mariedm (approved):

DatadogCore/Sources/Core/DatadogCore.swift
```diff
@@ -491,3 +491,12 @@ extension DatadogCore: Flushable {
         }
     }
 }
+
+#if SPM_BUILD
+import DatadogPrivate
```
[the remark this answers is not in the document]
maxep: SPM does not support mixed language target, so we created the `DatadogPrivate` to expose objc code to the core. In Xcode targets, the objc classes are part of the core target.

DatadogInternal/Sources/Utils/DDError.swift
```diff
@@ -90,3 +90,48 @@ public struct InternalError: Error, CustomStringConvertible {
...
+                value = try block()
+            } catch {
+                swiftError = error
+            }
+        }
+    } catch {
+        // wrap the underlying objc runtime exception in
+        // a `ObjcException` for easier matching during
+        // escalation.
+        throw ObjcException(error: error)
```
ncreated, line 133: suggestion/ Would it make sense to pass `#fileID` and `#line` to `ObjcException` as additional properties, so we can see the source code that raised exception?

DatadogSessionReplay/Sources/Recorder/Recorder.swift
```diff
@@ -119,37 +109,19 @@ public class Recorder: Recording {
...
-
-            resourceProcessor.process(
-                resources: viewTreeSnapshot.resources,
-                context: .init(recorderContext.applicationID)
-            )
-        } catch let error {
-            isSuccessful = false
-            telemetry.error("[SR] Failed to take snapshot", error: DDError(error: error))
+    func captureNextRecord(_ recorderContext: Context) throws {
+        guard let viewTreeSnapshot = try viewTreeSnapshotProducer.takeSnapshot(with: recorderContext) else {
```
ncreated, line 113: question/ Won't the Swift error thrown from here be wrapped into `ObjcException` and falsely reported as Obj-c issue to telemetry? Maybe we should split this call into two succeeding calls and pass the `viewTreeSnapshot` result to `captureNextRecord(_:viewTreeSnapshot:)`?

DatadogInternal/Sources/Utils/DDError.swift
```diff
@@ -90,3 +90,48 @@ public struct InternalError: Error, CustomStringConvertible {
...
+
+/// Rethrow Objective-C runtime exception as `Swift.Error`.
+///
+/// - Warning: As stated in [Objective-C Automatic Reference Counting (ARC)](https://clang.llvm.org/docs/AutomaticReferenceCounting.html#exceptions),
+/// in Objective-C, ARC is not exception-safe and  does not perform releases which would occur at the end of a
+/// full-expression if that full-expression throws an exception. Therefore, ARC-generated code leaks by default
+/// on exceptions.
+/// - throws: `ObjcException` if an exception was raised by the Objective-C runtime.
+@discardableResult
+public func objc_rethrow<T>(_ block: () throws -> T) throws -> T {
```
ncreated, line 118: change-request/ It seems we're missing explicit tests coverage for the logic in this code. What we test in `ObjcExceptionHandlerTests.swift` is the direct obj-c implementation, bypassing this function. Wouldn't it make sense to change it to utilize `objc_rethrow` instead?

ncreated (changes requested):
Looks good, exactly what we need 👍. I left few questions and requested change for adding unit test coverage to the `objc_rethrow` call.

DatadogCore/Sources/Datadog.swift
```diff
@@ -492,6 +492,8 @@ public enum Datadog {
             verbosityLevel: { Datadog.verbosityLevel }
         )
 
+        registerObjcExceptionHandlerOnce()
```
ncreated, line 495: suggestion/ It might make sense to register this handler before we instantiate `DatadogCore`. As part of core init SDK interacts with file system (to create directory structure), which I believe takes advantage of obj-c error handling.

DatadogCore/Sources/Core/DatadogCore.swift
ncreated: I believe this is because we support multiple SDK instances. This handler can be considered a shared resource between them. In any case, we could make it clear with minimal API comment.

DatadogCore/Sources/Core/DatadogCore.swift
maxep: Good question! The `ObjcException.rethrow` is explicitly internal as part of the `DatadogInternal` module, if a customer provides a non objc-try-catch implementation it will be at their own risk but it won't affect the SDK behaviour. So I don't see the don't need for compilation or runtime validation, if any existed.

DatadogCore/Sources/Datadog.swift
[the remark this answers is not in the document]
maxep: The `registerObjcExceptionHandlerOnce` does not invoke the `ObjcException.rethrow`, it's setting an implementation to a closure property. It won't ever fail.

maxep: Yes, good point 👍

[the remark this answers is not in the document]
mariedm: You're right, actually it's only doing the assignment, not calling it.

DatadogSessionReplay/Sources/Recorder/Recorder.swift
maxep: No, because `objc_rethrow` implementation catches Swit errors and [rethrow them as-is](https://github.com/DataDog/dd-sdk-ios/blob/f824015317e2f3f7a01fa3578b94983ed39c1512/DatadogInternal/Sources/Utils/DDError.swift#L136).
I will add test coverage on this 👍

mariedm (approved):

ncreated (approved):
🎯

PR #1864, opened 2024-05-27
[Session replay] Support bg color + hidden nav bar

mariedm:
### What and why?

This PR adds support for hidden navigation bar and navigation bar with colors (both `backgroundColor` and `barTintColor`) for session replay.

See an example of session replay [here](https://mobile-integration.datadoghq.com/rum/replay/sessions/4218d75a-dbe8-451a-ab69-2549e9482cf6?applicationId=5e20092a-ba66-4868-bfa3-ee8a0f1a54a1&seed=003ccbcd-4599-4973-9d52-9a8b784111bf&ts=1716823072409) with different colors used.

Screenshot 2024-06-03 at 11 36 57

### How?

- If the navigation bar is hidden, it is identified as an `InvisibleElement` ([session replay example](https://mobile-integration.datadoghq.com/rum/replay/sessions/4218d75a-dbe8-451a-ab69-2549e9482cf6?applicationId=5e20092a-ba66-4868-bfa3-ee8a0f1a54a1&seed=003ccbcd-4599-4973-9d52-9a8b784111bf&ts=1716823072409))
- The recorder now accounts for the navigation bar's background color ([session replay example](https://mobile-integration.datadoghq.com/rum/replay/sessions/9956410b-35f8-41bd-8cba-8b66d234d50b?applicationId=5e20092a-ba66-4868-bfa3-ee8a0f1a54a1&seed=4fbb32e1-4da5-4aad-8f20-c1685d746e6d&ts=1716819952461))

### Context on Navigation Bar Style

The navigation bar can be styled (default or black, translucent or not) without specifying colors in the settings (background color, bar tint color). In these cases, the system assigns colors that are not reflected in the actual colors detected in code or view debugger—they remain nil. 

Since these colors are inaccessible, we must check the navigation bar settings directly, rather than using a general approach (eg. at the `UIViewRecorder` level).

### Note on Title's Font Color

Similarly, switching the nav style from Default to Black changes the title font color, but this change is not detected in the actual colors. Unless specified, the title’s font color is black with the Default style and white with the Black style. Since we have to check the navigation bar's style, we might have to add a property to `UINavigationBarWireframesBuilder`. The title's font color is not addresssed in this PR. 

### Note on Border Color

[...]

PR #1890, opened 2024-06-10
RUM-4151 [SR] Support background color for Tab Bars

mariedm:
### What and why?

Session Replay currently does not support custom background colors for tab bars, defaulting to white or black based on the user's trait collection.

### How?

This PR:
- Modifies the SR Tab Bar Recorder to support custom background colors.
- Adds new snapshot tests for both static and embedded tab bars.

This work is similar to a previous PR that added support for custom background colors on navigation bars. For more context, see [PR #1864](https://github.com/DataDog/dd-sdk-ios/pull/1864).

## Notes on deselected items’ Tint Color

TLDR: if no `unselectedTintColor` is set, icons will be rendered in white.

- Both selected and unselected tint colors for tab bar items can be set using `tintColor` and `unselectedItemTintColor`.
- If `unselectedItemTintColor` is not set, the default is gray, though it appears white in the view debugger and is not accessible via code.
- An item's icon is recorded using the `UIImageViewRecorder`. The correct tint color for a selected item is obtained through the `tintColorProvider`.
- It is not possible to detect selection status at the `image` or `imageView` level within the `UIImageViewRecorder`.
    - `image.isTinted` always returns `true` if a `tintColor` is set, regardless of selection status.
    - Without access to the unselected tint color, we cannot determine the rendered color for unselected items.
    - Comparing the recorded image with the `withTintColor(_:)` effect applied to it does not work.
    - Accessing the Tab Bar and Tab Bar button from the image's superviews is possible, but the button is of type `UITabBarButton`, a private API, preventing state checks to determine selection status and appropriate tint color.
## Snapshot tests screenshots

**Static Tab Bar**

testTabBars()-allow-privacy

**Embedded Tab Bar**

testTabBars()-embeddedtabbar-allow-privacy

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UITabBarRecorder.swift
```diff
@@ -25,7 +25,6 @@ internal struct UITabBarRecorder: NodeRecorder {
     }
 
     private func inferOccupiedFrame(of tabBar: UITabBar, in context: ViewTreeRecordingContext) -> CGRect {
-        // TODO: RUMM-2791 Enhance appearance of `UITabBar` and `UINavigationBar` in SR
```
ncreated, line 28: 🎯👌

ncreated (approved):
Looks great 🎯. I appreciate the level of details in PR description and snapshot tests - all clear 🚀! 

Minor comment on conventions: we usually mention the JIRA ref (RUM-4151 in this case) as part of every commit (prefix) and in the PR title. This way we establish automatic link between GH and JIRA, which helps tracking decisions later. For this PR, let's mention it in the title at least 🙂.

DatadogSessionReplay/SRSnapshotTests/SRFixtures/Sources/SRFixtures/ViewControllers/TabbarControllers.swift
```diff
@@ -0,0 +1,32 @@
...
+
+public class TabbarViewControllers: UIViewController {
+
+    @IBOutlet var tabbars: [UITabBar]!
+    
+    public override func viewDidLoad() {
+        for tabbar in tabbars {
+            // Select the first tabbar's item
+            // so we can see both a selected and unselected item
+            tabbar.selectedItem = tabbar.items?[0]
```
maxep, line 17: /nit
```suggestion
            tabbar.selectedItem = tabbar.items?.first
```

maciejburda (commented):
Nice!

I would only suggest defaulting the icon image tint to gray whenever we deal with `nil`.

Also - we probably can drop the bottom border. Seems like native component has only a gray line on top.

maxep (dismissed):
Looks great! nice work on the snapshot tests 👏

mariedm:
> Minor comment on conventions: we usually mention the JIRA ref (RUM-4151 in this case) as part of every commit (prefix) and in the PR title. This way we establish automatic link between GH and JIRA, which helps tracking decisions later. For this PR, let's mention it in the title at least 🙂.

Got it! I will follow this pattern next time. In the meantime, I added it to the PR title.

mariedm:
> I would only suggest defaulting the icon image tint to gray whenever we deal with `unselectedItemTintColor == nil`.
> 
> Also - we probably can drop the bottom border. Seems like native component has only a gray line on top.

@maciejburda I can remove the bottom border color. 👍 However, how do you suggest we default the icon to gray when `unselectedItemTintColor == nil`? We don't have access to this property within the `UIImageViewRecorder`, and we can't determine if the tab bar item it belongs to is selected or not. Therefore, we don't know whether to apply the `tintColor` or the `unselectedItemTintColor`.

mariedm:
@maciejburda Actually, the current implementation in `UITabBarRecorder` will also remove the top border, which might not be what we want?

testTabBars()-mask-privacy

testTabBars()-embeddedtabbar-allow-privacy

mariedm:
@maciejburda as discussed in the Daily today, I can merge this PR once I have the required approvals to unblock @ncreated's work. I created a follow-up [ticket](https://datadoghq.atlassian.net/browse/RUM-4883) from your comment.

maciejburda (approved):
LGTM!

As discussed - we can try to tackle white tint images in a separate PR

PR #1893, opened 2024-06-11
RUM-2442 [S8S] Emit `_dd.session.id` and `_dd.application.id` on view change

maxep:
### What and why?

Same as https://github.com/DataDog/dd-sdk-android/pull/1813

Emit the Application and Session ID with the view so S8S can retrieve them in the device logs.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogRUM/Sources/RUMMonitor/Scopes/RUMApplicationScope.swift
+++ b/DatadogRUM/Sources/RUMMonitor/Scopes/RUMApplicationScope.swift
@@ -43,11 +43,6 @@ internal class RUMApplicationScope: RUMScope, RUMContextProvider {
             activeViewName: nil,
             activeUserActionID: nil
         )
-
-        // Notify Synthetics if needed
-        if dependencies.syntheticsTest != nil {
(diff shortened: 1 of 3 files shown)


vbarbaresi (approved):
Thanks!

DatadogRUM/Sources/RUMMonitor/Scopes/RUMApplicationScope.swift
```diff
@@ -43,11 +43,6 @@ internal class RUMApplicationScope: RUMScope, RUMContextProvider {
             activeViewName: nil,
             activeUserActionID: nil
         )
-
```
mariedm, line 46: /question I know that's how Android did it, but could it be useful to not delete these, and just add the new logs?

maxep: The logs in the view scope will be sufficient for S8S to correlate with the session, no need to keep other prints 👍

PR #1895, opened 2024-06-11
Fix Xcode 15 compilation

NachoSoto:
### What and why?

`UIApplication.beginBackgroundTask` added a default value of `nil`. This allows compiling on both Xcode 15 and 16.


NachoSoto:
No problem, thank you!

PR #1904, opened 2024-06-13
RUM-4591 chore: Add diagnostic attributes to "RUM Session Ended" telemetry

ncreated:
### What and why?

📦 🔭 According to [the spec](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3663167927/RUM+Session+Ended+Telemetry+-+Spec) (internal), this PR adds "diagnostic attributes" to "RUM Session Ended" telemetry.

This is a follow-up on https://github.com/DataDog/dd-sdk-ios/pull/1866 where we implemented a handset of "quality attributes".

### How?

The list of added attributes:
#### `rse.has_background_events_tracking_enabled`
Indicates whether [tracking of background events](https://github.com/DataDog/dd-sdk-ios/blob/888d39fc5407e01b72deab78451be689232b3342/DatadogRUM/Sources/RUMConfiguration.swift#L105-L112) was enabled in the RUM configuration.

#### `rse.views_count.by_instrumentation`
Holds the map of instrumentation type to the number of distinct views (view UUIDs) tracked through it (`manual` | `swiftui` | `uikit`).

#### `rse.views_count.with_has_replay`
The number of views with `has_replay == true`.

#### `rse.ntp_offset`
The NTP offset at the beginning and end of the session.

#### `rse.no_view_events_count`
The number of events (`actions` | `resources` | `errors` | `long_tasks`) dropped because there was no active view at the time they occurred.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogRUM/Sources/SDKMetrics/SessionEndedMetric.swift
+++ b/DatadogRUM/Sources/SDKMetrics/SessionEndedMetric.swift
@@ -81,9 +81,11 @@ internal struct SessionEndedMetric {
     /// Indicates if the session was stopped through `stopSession()` API.
     private var wasStopped = false
 
+    /// If `RUM.Configuration.trackBackgroundEvents` was enabled for this session.
+    private let tracksBackgroundEvents: Bool
+
     // TODO: RUM-4591 Track diagnostic attributes:
     // - no_view_events_count
-    // - has_background_events_tracking_enabled
     // - has_replay
     // - ntp_offset
 
@@ -94,14 +96,17 @@ internal struct SessionEndedMetric {
     ///   - sessionID: An ID of the session that is being tracked with this metric.
     ///   - precondition: The precondition that led to starting this session.
     ///   - context: The SDK context at the moment of starting this session.
+    ///   - tracksBackgroundEvents: If background events tracking is enabled for this session.
     init(
         sessionID: RUMUUID,
         precondition: RUMSessionPrecondition?,
-        context: DatadogContext
+        context: DatadogContext,
+        tracksBackgroundEvents: Bool
     ) {
         self.sessionID = sessionID
         self.bundleType = context.applicationBundleType
         self.precondition = precondition
+        self.tracksBackgroundEvents = tracksBackgroundEvents
     }
 
     /// Tracks the view event that occurred during the session.
@@ -165,6 +170,8 @@ internal struct SessionEndedMetric {
         let duration: Int64?
         /// Indicates if the session was stopped through `stopSession()` API.
         let wasStopped: Bool
+        /// If background events tracking is enabled for this session.
+        let hasBackgroundEventsTrackingEnabled: Bool
 
         struct ViewsCount: Encodable {
             /// The number of distinct views (view UUIDs) sent during this session.
@@ -207,6 +214,7 @@ internal struct SessionEndedMetric {
             case precondition
             case duration
             case wasStopped = "was_stopped"
+            case hasBackgroundEventsTrackingEnabled = "has_background_events_tracking_enabled"
             case viewsCount = "views_count"
             case sdkErrorsCount = "sdk_errors_count"
         }
@@ -249,6 +257,7 @@ internal struct SessionEndedMetric {
                 precondition: precondition?.rawValue,
                 duration: durationNs,
                 wasStopped: wasStopped,
+                hasBackgroundEventsTrackingEnabled: tracksBackgroundEvents,
                 viewsCount: .init(
                     total: totalViewsCount,
                     background: backgroundViewsCount,
--- a/DatadogRUM/Sources/SDKMetrics/SessionEndedMetric.swift
+++ b/DatadogRUM/Sources/SDKMetrics/SessionEndedMetric.swift
@@ -84,10 +84,12 @@ internal struct SessionEndedMetric {
     /// If `RUM.Configuration.trackBackgroundEvents` was enabled for this session.
     private let tracksBackgroundEvents: Bool
 
+    /// The current value of NTP offset at session start.
+    private let ntpOffsetAtStart: TimeInterval
+
     // TODO: RUM-4591 Track diagnostic attributes:
     // - no_view_events_count
     // - has_replay
-    // - ntp_offset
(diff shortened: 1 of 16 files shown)


DatadogRUM/Sources/SDKMetrics/SessionEndedMetric.swift
```diff
@@ -81,29 +108,40 @@ internal struct SessionEndedMetric {
...
+        self.ntpOffsetAtStart = context.serverTimeOffset
     }
 
     /// Tracks the view event that occurred during the session.
-    mutating func track(view: RUMViewEvent) throws {
+    /// - Parameters:
+    ///   - view: the view event to track
+    ///   - instrumentationType: the type of instrumentation used to start this view (only the first value for each `view.id` is tracked; succeeding values
+    ///   will be ignored so it is okay to pass value on first call and then follow with `nil` for next updates of given `view.id`)
+    mutating func track(view: RUMViewEvent, instrumentationType: ViewInstrumentationType?) throws {
```
ganeshnj, line 130: can this type be a reference type? 

struct with mutating methods are expensive and forces new copy of the object everytime we mutate.

ganeshnj (commented):
A more fundamental question on choice of type in the PR. The changes looks good.

DatadogRUM/Sources/SDKMetrics/SessionEndedMetric.swift
ncreated: We started with reference type (`class SessionEndedMetric`), but [as discussed here](https://github.com/DataDog/dd-sdk-ios/pull/1866#discussion_r1627535883) the cost was on excessive locks for the sake of state synchronization. By its nature, updates to `SessionEndedMetric` will come from various places in the code and different threads. With that in mind we preferred simplicity in thread safety (single lock from `SessionEndedMetricController`) over memory footprint due to struct mutation.

To address memory footprint concerns:
- the SEM struct is optimized to not store large values - it pre-computes metric attributes
- the most frequent mutation (`track(view:)`) is linked to app lifecycle, and should be spaced out enough to be considered a rare event

ganeshnj: Makes sense, 

a followup to this how do we make sure the usage of the struct. If the updates are being called from different threads, a mutating struct is not thread safe.

Try yourself

```swift
    func testConcurrency() {
        var counter = Counter()

        DispatchQueue.concurrentPerform(iterations: 100) { _ in
            counter.increment()
        }

        DispatchQueue.concurrentPerform(iterations: 100) { _ in
            counter.decrement()
        }

        XCTAssertEqual(counter.value, 0)
    }

    struct Counter {
        var value = 0

        mutating func increment() {
            value += 1
        }

        mutating func decrement() {
            value -= 1
        }
    }
```

You have to call mutating methods from same queue (thread) with `sync` API if we want this to be thread safe mutating struct.

DatadogRUM/Sources/SDKMetrics/SessionEndedMetric.swift
```diff
@@ -121,6 +159,15 @@ internal struct SessionEndedMetric {
         }
     }
 
+    /// Tracks an event missed due to absence of an active view.
+    mutating func track(missedEventType: MissedEventType) {
```
ganeshnj, line 163: missing the param docs.

DatadogRUM/Sources/SDKMetrics/SessionEndedMetric.swift
```diff
@@ -179,18 +234,61 @@ internal struct SessionEndedMetric {
 
         let sdkErrorsCount: SDKErrorsCount
 
+        struct NTPOffset: Encodable {
```
ganeshnj, line 237: docs

DatadogRUM/Sources/SDKMetrics/SessionEndedMetric.swift
ncreated: Thread safety is enforced by design. The [`SessionEndedMetricController`](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogRUM/Sources/SDKMetrics/SessionEndedMetricController.swift) is the only type that mutates the `SessionEndedMetric` struct. The controller is injected into codebase, making the struct a private detail. All other systems interact with the controller, not the `struct`.

By principle, the `struct SessionEndedMetric` should be made `private` to `SessionEndedMetricController` so it is not accessible in `internal` namespace. However, we keep it `internal` so it can be unit tested. Covering all edge cases for `SessionEndedMetric` mutation is trivial in `SessionEndedMetricTests`. Achieving the same through `SessionEndedMetricControllerTests` will visibly more hassle.

How does it sound @ganeshnj ?

ncreated: Also, `SEM` mutations through `SEMController` are tested for thread safety, see:

- [`func testTrackingSessionEndedMetricIsThreadSafe()`](https://github.com/DataDog/dd-sdk-ios/blob/07a23ef31523c19518b97f1cea6ba96c94f2ef9c/DatadogRUM/Tests/SDKMetrics/SessionEndedMetricControllerTests.swift#L93-L115)

ganeshnj: I see the lock on the [dictionary](https://github.com/DataDog/dd-sdk-ios/blob/07a23ef31523c19518b97f1cea6ba96c94f2ef9c/DatadogRUM/Sources/SDKMetrics/SessionEndedMetricController.swift#L14) which makes it thread safe.

ncreated: Thanks for clarifying this point on call 👍. The `SEM` is now a reference type to reduce memory footprint on mutations. Thread safety is still ensured by `SEMController`. After all it was a small but nice change, thanks!

ganeshnj (approved):

ganeshnj (approved):

PR #1906, opened 2024-06-14
RUM-4883 Tabbar Icon Default Tint Color

mariedm:
### What and why?

This PR introduces changes to the Session Replay feature to handle the default color of tab bar icons when no unselectedItemTintColor is set. This enhances the visual consistency and clarity in the session replays.

### How?

This PR updates the `UITabBarRecorder` to handle the default color for unselected tab bar items. It introduces a logic that sets the tab bar's icon color to gray when the `unselectedItemTintColor` is not provided.  The `UITabBarRecorder` now includes its own `UIImageViewRecorder` to handle icons properly.

Here's a SR example with `unselectedTintColor` not set on tab bar: [Session Replay Example 1](https://mobile-integration.datadoghq.com/rum/replay/sessions/9eaaf371-f8dd-46c2-9106-0bc986c074d9?seed=99a49f69-ce1d-4d26-adb1-f83fb69fdc8c&ts=1718378059952).

### Notes on Image Comparison

This PR uses image comparison to identify the selected item in the tab bar. I considered different options for this:
- Image Data Comparison: accurate but too costly in terms of performance, especially if executed on the main thread
- Using `srIdentifier`: provides a unique string, and while cached, still not ideal for performance
- Using Image Description, idea suggested by @maciejburda: provides a sort of unique identifier with minimal overhead (though descriptions _could_ be similar for images with similar properties (?))
- Using `cgImage` properties: generates a unique identifier, still theoretically possible for collisions, but highly unlikely

To make a decision between the last 2 options, I did a small benchmark on real devices. Results show that `cgImage` is much more efficient (24x faster on iPhone 13 mini, 12x faster on iPhone SE 2nd gen.).

### Screenshots

Updated screenshots from snapshot tests.

<table>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`


DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UITabBarRecorder.swift
```diff
@@ -7,21 +7,73 @@
...
+                            // our hypothesis is that each item uses a different image.
+                            return itemSelectedImage?.uniqueDescription == imageViewImage.uniqueDescription
+                        }
+
+                        // If the item is not selected,
+                        // return the unselectedItemTintColor,
+                        // or the default gray color if not set.
+                        if currentItemInSelectedState == nil || tabBar.selectedItem != currentItemInSelectedState {
+                            let unselectedColor = tabBar.unselectedItemTintColor ?? .lightGray.withAlphaComponent(0.5)
+                            return tabBar.unselectedItemTintColor ?? .systemGray.withAlphaComponent(0.5)
```
ncreated, line 40: suggestion/ In some other places we avoid referencing system colors directly. Instead we bridge them through [`SystemColors`](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogSessionReplay/Sources/Recorder/Utilities/SystemColors.swift) utility that provides fallbacks for older OS-es. Might be used here for consistency and better look-up (it already defines `.systemGray`).

ncreated (approved):
> I considered different options for this:
> ...
> To make a decision between the last 2 options, I did a small benchmark on real devices. Results show that cgImage is much more efficient (24x faster on iPhone 13 mini, 12x faster on iPhone SE 2nd gen.).

Excellent approach 🏅! We need more benchmark driven decisions in SR, well done 💪

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UITabBarRecorder.swift
```diff
@@ -76,3 +128,20 @@ internal struct UITabBarWireframesBuilder: NodeWireframesBuilder {
     }
 }
 #endif
+
+fileprivate extension UIImage {
+    var uniqueDescription: String? {
```
ncreated, line 133: suggestion/ We may want to add API comment with some of the contexts added in implementation. Notably, from the caller stand-point it might be interesting:
- that it is calculated from `CGImage` props,
- it prefers performance over acurracy (collisions are possible),
- when it may return `nil`.

mariedm: Good idea, I've updated ✅

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UITabBarRecorder.swift
mariedm: I followed your suggestion ✅

ncreated (approved):

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UITabBarRecorder.swift
```diff
@@ -7,21 +7,71 @@
 #if os(iOS)
 import UIKit
 
-internal struct UITabBarRecorder: NodeRecorder {
+internal final class UITabBarRecorder: NodeRecorder {
     let identifier = UUID()
+
+    private var currentlyProcessedTabbar: UITabBar? = nil
+    private lazy var subtreeViewRecorder: ViewTreeRecorder = {
```
maxep, line 14: /change-request The `UITabBarRecorder` is static and the `currentlyProcessedTabbar ` var will retain the last recorded `UITabBar`. We should release the tab bar when we are done recording, for that we could use a method that instantiate a `ViewTreeRecorder` on the fly:
```swift
func recordSubtree(of tabbar: UITabBar, in context: ViewTreeRecordingContext) -> NodeSemantics? {
  ViewTreeRecorder([...]).record(tabBar, in: context)
}
```
I prefer this approach instead of a weak var, the flow is clearer this way IMO.

maxep (changes requested):
Very nice work, I really appreciate the performance driven solution and the level of comments 👏 
I have left a minor change request on releasing the `UITabBar` in the recorder.

maxep (changes requested):
Very nice work, I really appreciate the performance driven solution and the level of comments 👏 
I have left a minor change request on releasing the `UITabBar` in the recorder.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UITabBarRecorder.swift
mariedm: That's a valid point, I made the requested change. Thanks for the suggestion!

maxep (approved):
Thanks, well done!

PR #1908, opened 2024-06-18
Allow disabling app hang monitoring in ObjC API

0xnm:
### What and why?

Unlike long tasks tracking, app hang monitoring can be disabled. This PR treats 0 passed as an argument in ObjC API as a special value to disable app hang monitoring (to avoid changing property type to the optional).

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogCore/Tests/DatadogObjc/DDRUMConfigurationTests.swift
+++ b/DatadogCore/Tests/DatadogObjc/DDRUMConfigurationTests.swift
@@ -107,10 +107,16 @@ class DDRUMConfigurationTests: XCTestCase {
     }
 
     func testAppHangThreshold() {
-        let random: TimeInterval = .mockRandom()
+        let random: TimeInterval = .mockRandom(min: 0.01, max: .greatestFiniteMagnitude)
         objc.appHangThreshold = random
         XCTAssertEqual(objc.appHangThreshold, random)
         XCTAssertEqual(swift.appHangThreshold, random)
+    }
+
+    func testAppHangThresholdDisable() {
+        objc.appHangThreshold = 0
+        XCTAssertEqual(objc.appHangThreshold, 0)
+        XCTAssertEqual(swift.appHangThreshold, nil)
     }
 
     func testVitalsUpdateFrequency() {
--- a/DatadogObjc/Sources/RUM/RUM+objc.swift
+++ b/DatadogObjc/Sources/RUM/RUM+objc.swift
@@ -378,7 +378,7 @@ public class DDRUMConfiguration: NSObject {
     }
 
     @objc public var appHangThreshold: TimeInterval {
-        set { swiftConfig.appHangThreshold = newValue }
+        set { swiftConfig.appHangThreshold = newValue == 0 ? nil : newValue }
         get { swiftConfig.appHangThreshold ?? 0 }
     }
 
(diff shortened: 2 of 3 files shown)


DatadogObjc/Sources/RUM/RUM+objc.swift
```diff
@@ -378,7 +378,7 @@ public class DDRUMConfiguration: NSObject {
     }
 
     @objc public var appHangThreshold: TimeInterval {
-        set { swiftConfig.appHangThreshold = newValue }
+        set { swiftConfig.appHangThreshold = newValue == 0 ? nil : newValue }
```
mariedm, line 381: I'm not sure this is related, but I see a `minAppHangThreshold` set to `0.1` in `RUMInstrumentation`. Can it be bypassed and actually set to `0`?

mariedm (dismissed):

DatadogObjc/Sources/RUM/RUM+objc.swift
0xnm: this is per documentation https://github.com/DataDog/dd-sdk-ios/blob/085da2c4a390babc3bcc6097fc9f6f3a34d94a8e/DatadogRUM/Sources/RUMConfiguration.swift#L124-L125 which says to use `nil` value to disable it.

mariedm: Got it, thanks!

DatadogCore/Tests/DatadogObjc/DDRUMConfigurationTests.swift
```diff
@@ -110,7 +110,7 @@ class DDRUMConfigurationTests: XCTestCase {
         let random: TimeInterval = .mockRandom()
         objc.appHangThreshold = random
         XCTAssertEqual(objc.appHangThreshold, random)
-        XCTAssertEqual(swift.appHangThreshold, random)
+        XCTAssertEqual(swift.appHangThreshold, random == 0 ? nil : random)
```
mariedm, line 113: I feel like this test modification is trying to accommodate the changes in logic, rather than ensuring it accurately validates the new logic.

We could add a new test specifically for the `0` value. And to avoid avoid `mockRandom` returning `0` in the original test, making it fail, we could use another extension, such as the `mockRandom` `Double extension` in `FoundationMocks`, which takes parameters.

0xnm: yes, it can be see as an attempt to cut the corner.

I've pushed an update: one test will test strictly positive value (I didn't introduce additional extension method, but rather re-used existing one providing explicit args), another will test 0 value to disable app hangs monitoring.

Let me know what you think.

mariedm (approved):

DatadogCore/Tests/DatadogObjc/DDRUMConfigurationTests.swift
mariedm: It looks great! 👌✅

DatadogObjc/Sources/RUM/RUM+objc.swift
ganeshnj, line 381: This file is part of codegen and must be not be modified manually.

0xnm: can you point me to the command which generates it? I was adding methods to this class before manually.

PR #1912, opened 2024-06-19
RUM-4911 feat(watchdog-termination): ability to find most recently updated file in a dir

ganeshnj:
### What and why?

We need to know the timestamp of file update to get the approximate timing for the watchdog termination. This will be used to build the rum error event.

### How?

The implementation is quite straight forward, with the new API `Directory.mostRecentModifiedFile(before:)`, we can find the most recently updated file. It recursively checks all the subdirectories, files and itself and returns the file which is most recently updated.

### Review checklist

- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogCore/Tests/Datadog/Core/Persistence/Files/DirectoryTests.swift
+++ b/DatadogCore/Tests/Datadog/Core/Persistence/Files/DirectoryTests.swift
@@ -171,6 +171,156 @@ class DirectoryTests: XCTestCase {
         XCTAssertNoThrow(try destinationDirectory.file(named: "f3"))
     }
 
+   func testModifiedAt() throws {
+        // when directory is created
+        let before = Date.timeIntervalSinceReferenceDate - 1
+        let directory = try Directory(withSubdirectoryPath: uniqueSubdirectoryName())
+        defer { directory.delete() }
+        let creationDate = try directory.modifiedAt()
+        let after = Date.timeIntervalSinceReferenceDate + 1
+
+        XCTAssertNotNil(creationDate)
+        XCTAssertGreaterThanOrEqual(creationDate!.timeIntervalSinceReferenceDate, before)
+        XCTAssertLessThanOrEqual(creationDate!.timeIntervalSinceReferenceDate, after)
+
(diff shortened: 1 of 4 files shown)


DatadogCore/Tests/Datadog/Core/Persistence/Files/DirectoryTests.swift
```diff
@@ -171,6 +171,156 @@ class DirectoryTests: XCTestCase {
...
+
+        try file1.append(data: .mock(ofSize: 1))
+        try file2.append(data: .mock(ofSize: 2))
+        try file3.append(data: .mock(ofSize: 3))
+
+        try file2.delete()
+
+        let modifiedFile = try directory.mostRecentModifiedFile(before: .init())
+        XCTAssertNotNil(modifiedFile)
+        XCTAssertEqual(directory.url, modifiedFile?.url)
```
mariedm, line 281: nit: should it also compare `modifiedFile` and `file3` urls?

mariedm (approved):
LGMT! I added a small nit, but feel free to disregard. 

Another nit: I feel like we could use some comments in the tests, but previous tests don't have any.

DatadogCore/Tests/Datadog/Core/Persistence/Files/DirectoryTests.swift
mariedm: Oh right, so it returns the directory itself? So for watchdog termination, it doesn't matter if we get the timestamp of the last modified file or the directory itself?

ganeshnj: correct, what matters is it has to be lower than the given date, because we don't want to take any updated dir/file which was affect of current session.

mariedm: Got it, thanks for the explanation.

ncreated (approved):
Neat 👌

PR #1916, opened 2024-06-21
RUM-4964 Improve Nav Bar Support in SR

mariedm:
### What and why?

Improving Navigation Bar support in Session Replay.

### How?

As described in this [archived PR](https://github.com/DataDog/dd-sdk-ios/pull/1915), the navigation bar appearance changes when scrolling the view. Currently, we can't access the scrolling state within the recorder, so we can't adapt the navigation bar's appearance dynamically.

Given that the current navigation bar recorder already supports most common use cases, this PR only enhances support by changing the navigation bar background color when it is set to non-translucent.

Additionally, the snapshot tests are updated to better assess the rendering:
- Removing the navigation bar present in the image tests, as they now have dedicated tests.
- Adding a test to apply a tint color to the navigation bar's title and items.
- Adding a method to push a view controller onto the navigation stack to test scenarios closer to real-world use cases.

#### Static Navigation Bars
<table>
  <tr>
    <td>No Tint Color set</td>
    <td>Tint Color set</td>
  </tr>
</table>

#### Embedded Navigation Bars

<table>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

### Review checklist

- [ ] Run smoke tests
- [ ] Run tests for `tools/`


DatadogSessionReplay/SRSnapshotTests/SRSnapshotTests/SRSnapshotTests.swift
```diff
@@ -292,9 +332,12 @@ final class SRSnapshotTests: SnapshotTestCase {
         ]
 
         for fixture in navBarFixtures {
-            show(fixture: fixture)
+            let navController = show(fixture: fixture) as! TestNavigationController
+            navController.pushNextView()
+            // Wait for view controller to be pushed onto the navigation stack.
+            wait(seconds: 0.5)
```
ncreated, line 338: question/ This might become flaky. Is there a way to perform the segue without animation, so the `pushNextView()` could be blocking API that guarantees expected state after return?

DatadogSessionReplay/SRSnapshotTests/SRSnapshotTests/SRSnapshotTests.swift
```diff
@@ -279,8 +280,47 @@ final class SRSnapshotTests: SnapshotTestCase {
...
+                snapshotLocation: .folder(named: snapshotsFolderPath, fileNameSuffix: "-\(fileNamePrefix)-\(privacyMode)-privacy"),
+                record: recordingMode
+            )
+        }
+
+        // Test Static Navigation Bars with tinted color
+        let vc2 = show(fixture: .navigationBars) as! NavigationBarControllers
+        vc2.setTintColor()
+        // Wait for colors to appear on screen, so the recorder can capture them.
+        wait(seconds: 0.5)
```
ncreated, line 306: question/ This might become flaky. Is there an option to assure that colors are set upon `setTintColor()` return? Might be tricky 🤔 💭

DatadogSessionReplay/SRSnapshotTests/SRSnapshotTests/SRSnapshotTests.swift
```diff
@@ -7,6 +7,7 @@
 import XCTest
 import SRFixtures
 import TestUtilities
+import DatadogInternal
```
ncreated, line 10: question/ Why we need `import DatadogInternal` in this scope?

DatadogSessionReplay/SRSnapshotTests/SRSnapshotTests/SRSnapshotTests.swift
ncreated, line 1: food-for-thought/ Not a request nor suggestion, rather raising a point on splitting this file into multiple ones for better segregation of snapshot tests. It's getting large 🙂. As we're adding more and more (🏅) snapshots, maybe it can benefit from a split and helpers reuse (dunno how repetitive the work here is tho).

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UINavigationBarRecorder.swift
```diff
@@ -40,11 +39,11 @@ internal struct UINavigationBarRecorder: NodeRecorder {
         return occupiedFrame
     }
 
-    private func inferColor(of navigationBar: UINavigationBar) -> CGColor {
-        // TODO: RUMM-2791 Enhance appearance of `UITabBar` and `UINavigationBar` in SR
```
ncreated, line 44: SuccessKid

ncreated (dismissed):
Looks great 👌 🎯. I left few questions and hints, nothing blocking.

DatadogSessionReplay/SRSnapshotTests/SRSnapshotTests/SRSnapshotTests.swift
mariedm: I can create another view controller in the storyboard with all the tint colors set, then no need to call a specific method or wait for half a sec.

DatadogSessionReplay/SRSnapshotTests/SRSnapshotTests/SRSnapshotTests.swift
mariedm: I tried to push the view without animation, but it's still necessary to wait for at least 0.1 seconds.

Here's an alternative solution (I haven't tried it): We can create a `UIViewController` subclass with a handler that notifies the navigation controller once `viewDidAppear` has been called. Using `XCTWaiter`, we can then wait for the view to appear. This approach ensures we wait for the view to actually appear rather than relying on an arbitrary delay. It also avoids unnecessarily long waits, making the tests run a bit faster.

While we could consider removing this navigation test altogether, doing so would reduce the test coverage of the recorder.

A small note: although this issue is worth addressing, there are many instances of `wait(seconds:)` in these snapshot tests.

ncreated: I wonder if [`UINavigationController.setViewControllers(_:animated:)`](https://developer.apple.com/documentation/uikit/uinavigationcontroller/1621861-setviewcontrollers) could help (AFAIK it is used by push / performSegue under the hood).

> there are many instances of wait(seconds:) in these snapshot tests

Fair point 👌 and the main reason I don't consider this a blocking feedback. I'm totally ok with what is implemented here 👍.

DatadogSessionReplay/SRSnapshotTests/SRSnapshotTests/SRSnapshotTests.swift
ncreated: It feels not worth it and I don't see any other option. Let's maybe check with x100 reps (in local) that `0.5s` delay isn't flaky and we're good to go 👍.

maxep (dismissed):
Looks great! nothing to add to @ncreated's comment 👍

DatadogSessionReplay/SRSnapshotTests/SRSnapshotTests/SRSnapshotTests.swift
mariedm: Good point. I needed it because of this line: `let privacyModes: [SessionReplayPrivacyLevel] = [.allow, .mask]`. It was convenient during tests, but I can remove it now.

DatadogSessionReplay/SRSnapshotTests/SRSnapshotTests/SRSnapshotTests.swift
mariedm: Agreed. I'd be happy to look into a small refactoring in a separate PR.

DatadogSessionReplay/SRSnapshotTests/SRSnapshotTests/SRSnapshotTests.swift
ncreated: Oh, this is actually a bug 🐞 that will require our customers to also import `DatadogInternal`, which we rather avoid. To fix it, the `SessionReplayPrivacyLevel` must be re-exported from `DatadogInternal` in [`SessionReplayConfiguration.swift`](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogSessionReplay/Sources/SessionReplayConfiguration.swift). The same way as we do for some types in RUM: [example](https://github.com/DataDog/dd-sdk-ios/blob/380bdf4938519b676d5a1d06bd2ce6928637a0a6/DatadogRUM/Sources/RUMConfiguration.swift#L10-L17). // cc @maxep 

For simplicity, can we fix it within this PR so we get rid of this import in snapshot tests?

DatadogSessionReplay/SRSnapshotTests/SRSnapshotTests/SRSnapshotTests.swift
mariedm: Unfortunately it's not working; a delay is still necessary. I was able to reduce the delay for the tint color test to 0.1 seconds. I ran both tests repeatedly, and they passed ✅

DatadogSessionReplay/SRSnapshotTests/SRSnapshotTests/SRSnapshotTests.swift
mariedm: I already removed the import statement from this file, but I could still do this requested change in the follow-up, refactoring PR?

maxep: Ah yes, that's a miss from me 🤦  I can open another PR to fix it 👍

maxep (approved):

PR #1917, opened 2024-06-21
RUM-4911 feat(watchdog-termination): send Watchdog Termination

ganeshnj:
### What and why?

We already have ability to detect the WT https://github.com/DataDog/dd-sdk-ios/pull/1889

Now we want to report them.

### How?

Related PR: https://github.com/DataDog/rum-events-format/pull/213

We already added capability to detect a watchdog termination https://github.com/DataDog/dd-sdk-ios/pull/1889, this PR implements the event sending mechanism. It uses basic building blocks added during the App hang reporting.

There are two fundamental changes done in this
1. The crash reporter now sets the core context baggage instead using point-to-point messaging. Point-to-point messaging introduced a strict ordering of feature enablement. ie first RUM must be enabled and then CR. This is a lot of cognitive load for our customers and we would like to make it simpler to use but this eventually makes the implementation complex for the monitor.
2. How we keep track of the RUM view event, which is used to build the error event. In current implementation, the monitor keeps track of most recent view event which is simpler than the initial concept.

   > In the initial concept, view events were read from the written batch which is more complex from implementation point of view but better for customer.

The `WatchdogTerminationAppStateManager` is now simpler as `WatchdogTerminationMonitor` takes care of message processing. The monitor starts when the launch report is available and checks for the occurrence of the watchdog termination, if found, it writes the event else start observing the application lifecycle and RUM view events. `WatchdogTerminationMonitor.State` enum is introduced for the state management.

### Review checklist

- [ ] Run smoke tests
- [ ] Run tests for `tools/`


DatadogRUM/Sources/Instrumentation/WatchdogTerminations/WatchdogTerminationMonitor.swift
```diff
@@ -9,73 +9,137 @@ import DatadogInternal
...
+        static let rumViewEventUpdated = "RUM View event updated"
+        static let failedToSendWatchdogTermination = "Failed to send Watchdog Termination event"
+        static let failedToDecodeLaunchReport = "Fails to decode LaunchReport in RUM"
     }
 
     let checker: WatchdogTerminationChecker
     let appStateManager: WatchdogTerminationAppStateManager
     let feature: FeatureScope
     let reporter: WatchdogTerminationReporting
+    weak var core: DatadogCoreProtocol?
```
ncreated, line 36: suggestion/ This looks against our goals from #1744 where we started removing direct core references from Features code. Direct refs are error prone (can leak if not weakified) and less scalable (they do not isolate core capabilities, instead bloat `DatadogCoreProtocol` with new things).

While I totally get the idea of having this reference here and `.mostRecentModifiedFileAt(before:)` doesn't fit into [`FeatureScope`](https://github.com/DataDog/dd-sdk-ios/blob/380bdf4938519b676d5a1d06bd2ce6928637a0a6/DatadogCore/Sources/Core/DatadogCore.swift#L311-L374), I wonder if we could avoid arbitrary `weak var core` by introducing `CoreUtils` type that manages the `core` reference similar to how it is done for [`CoreTelemetry`](https://github.com/DataDog/dd-sdk-ios/blob/380bdf4938519b676d5a1d06bd2ce6928637a0a6/DatadogInternal/Sources/Telemetry/Telemetry.swift#L366-L389). Just like with `Telemetry`, it can be safe to pass `CoreUtils` reference around and retain it on non-optional value.

cc @maxep what would be your take on this?

DatadogRUM/Sources/Instrumentation/WatchdogTerminations/WatchdogTerminationReporter.swift
```diff
@@ -10,13 +10,45 @@ import DatadogInternal
...
+                errorStack: nil,
+                errorThreads: nil,
+                errorBinaryImages: nil,
+                errorWasTruncated: nil,
+                errorMeta: nil
+            )
+            let error = builder.createRUMError(with: viewEvent)
+            let view = builder.updateRUMViewWithError(viewEvent)
+            writer.write(value: error)
+            writer.write(value: view)
```
ncreated, line 51: change-request/ We need to consider event timings - RUM intake doesn't accept events in certain situation. Notably, the [`viewEventAvailabilityThreshold`](https://github.com/DataDog/dd-sdk-ios/blob/380bdf4938519b676d5a1d06bd2ce6928637a0a6/DatadogRUM/Sources/FatalErrorBuilder.swift#L13-L19) defines the last moment when we can issue a view update to previous session.

It should be okay for the product to only support the main flow initially, [like it was done in `FatalAppHangsHandler`](https://github.com/DataDog/dd-sdk-ios/blob/380bdf4938519b676d5a1d06bd2ce6928637a0a6/DatadogRUM/Sources/Instrumentation/AppHangs/FatalAppHangsHandler.swift#L114-L124). We can cover edge-case situations later, ideally by moving this logic to `FatalErrorBuilder` instead of duplicating it between all kind of fatal errors:
- [`RUM.CrashReportReceiver`](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogRUM/Sources/Integrations/CrashReportReceiver.swift#L156-L173) - supports all edge cases for sending crashes
- [`FatalAppHangsHandler`](https://github.com/DataDog/dd-sdk-ios/blob/380bdf4938519b676d5a1d06bd2ce6928637a0a6/DatadogRUM/Sources/Instrumentation/AppHangs/FatalAppHangsHandler.swift#L114-L124) - supports only the happy-path for sending fatal hangs
- now `WatchdogTerminationReporter`

ncreated (changes requested):
The code looks great 👌. I left few comments on integrating watchdog error events into RUM - notably we must consider session availability constraints in RUM intake. All described through in-line comments.

On top of that, as this is final PR on this matter, we must add [integration-unit tests](https://github.com/DataDog/dd-sdk-ios/tree/develop/Datadog/IntegrationUnitTests/RUM) for watchdog terminations monitoring.

DatadogCore/Tests/Datadog/Mocks/DatadogInternal/DatadogCoreProxy.swift
```diff
@@ -91,6 +91,10 @@ internal class DatadogCoreProxy: DatadogCoreProtocol {
     func send(message: FeatureMessage, else fallback: @escaping () -> Void) {
         core.send(message: message, else: fallback)
     }
+
+    func mostRecentModifiedFileAt(before: Date) throws -> Date? {
+        return nil
```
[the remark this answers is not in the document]
ganeshnj: it wasn't needed for testing. I fixed it for completeness.

DatadogCore/Sources/Core/DatadogCore.swift
```diff
@@ -486,6 +486,15 @@ extension DatadogCore: Flushable {
     }
 }
 
+extension DatadogCore: CoreStorage {
+    /// Returns the most recent modification date of a file in the core directory.
+    /// - Parameter before: The date to compare the last modification date of files.
+    /// - Returns: The latest modified file or `nil` if no files were modified before given date.
+    func mostRecentModifiedFileAt(before: Date) throws -> Date? {
+        let file = try directory.coreDirectory.mostRecentModifiedFile(before: before)
+        return try file?.modifiedAt()
```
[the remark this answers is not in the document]
ganeshnj: There must not be an issue with file IO happening at the same time

```swift
   func testExample() throws {
        let tempPath = NSTemporaryDirectory()
        let tempFile = tempPath + "test_file_attributes.txt"
        let tempURL = URL(fileURLWithPath: tempFile)

        let exp = expectation(description: "test")

        // update file and query file attributes in parallel
        DispatchQueue.global().async {
            while true {
                self.updateFile(tempURL)
            }
        }

        DispatchQueue.global().async {
            while true {
                self.modifiedAt(tempURL)
            }
        }

        waitForExpectations(timeout: 10)
    }

    func modifiedAt(_ url: URL) -> Date? {
        let attributes = try? FileManager.default.attributesOfItem(atPath: url.path)
        let date = attributes?[.modificationDate] as? Date
        print("updatedAt: \(date ?? Date())")
        return date
    }

    func updateFile(_ url: URL) {
        let text = String(repeating: "a", count: 1000000)
        print(text.first)
        let data = text.data(using: .utf8)
        try? data?.write(to: url)
    }
```

Attribute querying is not part of critical section. However, a file deletion could be interesting case where we might end up returning a nil date. I still switch to queue.

DatadogObjc/Sources/RUM/RUMDataModels+objc.swift
```diff
@@ -7629,4 +7662,4 @@ public class DDTelemetryConfigurationEventView: NSObject {
 
 // swiftlint:enable force_unwrapping
 
-// Generated from https://github.com/DataDog/rum-events-format/tree/30d4b773abb4e33edc9d6053d3c12cd302e948a5
+// Generated from https://github.com/DataDog/rum-events-format/tree/ab365359e88ec5e24fc26397695af7e539b8d670
```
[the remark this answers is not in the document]
ganeshnj: sure thing, you can see can check the commit in tree is it was ever pushed to remote

https://github.com/DataDog/rum-events-format/tree/ab365359e88ec5e24fc26397695af7e539b8d670

DatadogRUM/Sources/Instrumentation/WatchdogTerminations/WatchdogTerminationReporter.swift
ganeshnj: Can you elaborate how does https://github.com/DataDog/dd-sdk-ios/blob/380bdf4938519b676d5a1d06bd2ce6928637a0a6/DatadogRUM/Sources/Instrumentation/AppHangs/FatalAppHangsHandler.swift#L114-L124 logic helps here? 

What makes it confusing is that `else` part only writes the error but this should be covered with view + error already.

Currently we are sending both view update and error. Which one is taken or rejected, is up to the backend logic.

DatadogRUM/Sources/Instrumentation/WatchdogTerminations/WatchdogTerminationMonitor.swift
ganeshnj: This is recurrent problem and I think we can do better here, this perhaps beyond the scope of this PR but one idea that I have is to introduce something like
```swift
public struct Weak<T> {
    public var value: T?
    
    public init(value: T? = nil) {
        self.value = value
    }
}
```

and update the protocols to not conform AnyObject
```
public protocol DatadogCoreProtocol: MessageSending, BaggageSharing, CoreStorage {
```

this will be enforced at compiler level and we can use core using `Weak<DatadogCoreProtocol>`. This is must better approach than keep wrapping the core into different structs like we did for telemetry.

Let's sync on this.

DatadogRUM/Sources/Instrumentation/WatchdogTerminations/WatchdogTerminationReporter.swift
ncreated: > Which one is taken or rejected, is up to the backend logic.

If we want to challenge existing logic it must be first verified with RUM Ingestion. The decision to send error with no view if 4h lasted since crash was based on inputs from BE team. I found first minimal record of this decision [here (see: PROBLEM 1)](https://docs.google.com/document/d/1u4Yr96NtvESbfltDMvqpCJrzl6Sa5KttmKqLSVtbtqQ/edit?usp=sharing). It is dated 2020, so it might be worth revisiting this topic. If this constraint is no longer there, we can vastly simplify App Hangs and Crash Reporting code.

I consider this a blocker as it introduces a divergence in fatal errors handling (WDT / CR / AH).

ncreated (commented):
👌 Looks good with the remaining blocker on sending view event more than 4h after session was closed (see my comment). I'd rather not end up in situation where one feature implements different view upload logic than other. If 4h constraint is no longer a thing (please verify), we must remove it from Crash Reporting and App Hangs monitoring too (this can happen through separate JIRA).

DatadogRUM/Sources/Instrumentation/WatchdogTerminations/WatchdogTerminationReporter.swift
ganeshnj: Next: we should do if-else with 4 hours window (constant) and unblock this PR and continue the conversation with RUM BE if we can simplify this.

DatadogRUM/Sources/Instrumentation/WatchdogTerminations/WatchdogTerminationMonitor.swift
ganeshnj: Action: for now, we do with wrapper mechanism as we do today and let's try to simplify after.

DatadogRUM/Sources/Instrumentation/AppHangs/ProcessIdentifier.swift
```diff
@@ -17,3 +17,10 @@ import Foundation
 /// - SDK started → RUM enabled → [hang occurs] → pending App Hang saved → SDK stopped → SDK started again → RUM enabled again → pending App Hang loaded
 /// - When restarting RUM , the `processID` check ensures dropping pending hang from the previous instance, preventing false "fatal" hang detection.
 internal let currentProcessID = UUID()
+
+/// Time since the application process started.
+///
+/// Example use case in watchdog termination tracking:
+/// - SDK started -> RUM enabled -> [watchdog termination] -> SDK stopped -> SDK started again -> RUM enabled again -> check if the app was terminated by watchdog
+/// - If true, check any file updates that were done before current process started, that is most close to the watchdog termination.
+internal let runningSince = Date()
```
ganeshnj, line 26: Action: we can use `LaunchTime`

ncreated (approved):
🎯 Great efforts!

PR #1922, opened 2024-06-24
RUM-1000 chore: improve telemetry messages for file I/O errors

ganeshnj:
### What and why?

Current telemetry message is not very helpful to diagnose file IO related issues

For example

```
Failed to write data to file - Cannot create file at path: /var/mobile/Containers/Data/Application/*/Library/Caches
```

### How?

Split the big `do-try-catch` block into individual ones to get more insights where the issue.

### Review checklist

- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### DIFF
--- a/DatadogCore/Sources/Core/Storage/Writing/FileWriter.swift
+++ b/DatadogCore/Sources/Core/Storage/Writing/FileWriter.swift
@@ -37,25 +37,44 @@ internal struct FileWriter: Writer {
     ///  - value: Encodable value to write.
     ///  - metadata: Encodable metadata to write.
     func write<T: Encodable, M: Encodable>(value: T, metadata: M?) {
-        do {
-            var encoded: Data = .init()
-            if let metadata = metadata {
+        var encoded: Data = .init()
+        if let metadata = metadata {
+            do {
                 let encodedMetadata = try encode(value: metadata, blockType: .eventMetadata)
                 encoded.append(encodedMetadata)
+            } catch {
+                DD.logger.error("(\(orchestrator.trackName)) Failed to encode metadata", error: error)
+                telemetry.error("(\(orchestrator.trackName)) Failed to encode metadata", error: error)
             }
+        }
 
+        do {
             let encodedValue = try encode(value: value, blockType: .event)
             encoded.append(encodedValue)
+        } catch {
+            DD.logger.error("(\(orchestrator.trackName)) Failed to encode value", error: error)
+            telemetry.error("(\(orchestrator.trackName)) Failed to encode value", error: error)
+            return
+        }
 
-            // Make sure both event and event metadata are written to the same file.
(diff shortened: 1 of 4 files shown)


mariedm (approved):

DatadogCore/Sources/Core/Storage/Writing/FileWriter.swift
```diff
@@ -37,21 +37,40 @@ internal struct FileWriter: Writer {
...
+        // Make sure both event and event metadata are written to the same file.
+        // This is to avoid a situation where event is written to one file and event metadata to another.
+        // If this happens, the reader will not be able to match event with its metadata.
+        let writeSize = UInt64(encoded.count)
+        let file: WritableFile
+        do {
+            file = try orchestrator.getWritableFile(writeSize: writeSize)
+        } catch {
+            DD.logger.error("Failed to get writable file", error: error)
+            telemetry.error("Failed to get writable file", error: error)
```
ncreated, line 69: suggestion/ For more context, would it make sense to include `writeSize`? To control the cardinality of `error.message`, this could be done by composing the `error`:
```swift
enum FileWriterError: Error { /* ... */ }

FileWriterError.writeError(error: error, writeSize: writeSize)
```

ncreated (commented):
I like the direction - definitely better than what we have today. However, to me the most critical piece that is missing is the actual track we write to (`rum`, `trace`, `logs`, `sr`, `sr-resources`). Given that [we have `trackName` available in parent `FilesOrchestrator`](https://github.com/DataDog/dd-sdk-ios/blob/538ffa6a5b9d8fbfcfaa9dc45dd39bd1e644ab19/DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift#L45), could we make it into `FileWriter` and attach to the `error`?

DatadogCore/Sources/Core/Storage/Writing/FileWriter.swift
ganeshnj: this would mask the underlying error and wrapping one error inside another is something not very encouraged. However I do like the idea of using the writeSize in the error message. Given these are logs, cardinality must not be an issue.

ganeshnj:
> I like the direction - definitely better than what we have today. However, to me the most critical piece that is missing is the actual track we write to (`rum`, `trace`, `logs`, `sr`, `sr-resources`). Given that [we have `trackName` available in parent `FilesOrchestrator`](https://github.com/DataDog/dd-sdk-ios/blob/538ffa6a5b9d8fbfcfaa9dc45dd39bd1e644ab19/DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift#L45), could we make it into `FileWriter` and attach to the `error`?

fair call, I just updated error messages with track name following the same format as we do at other places `({trackname})`

PR #1930, opened 2024-06-28
chore: use otel-swift fork that only has APIs

ganeshnj:
### What and why?

Related https://github.com/DataDog/opentelemetry-swift-packages/pull/17

We are not sure when the dependency issue will be resolved at the root level. 

### How?

Take a dependency on the forked version of otel-swift that has only APIs.

### Review checklist

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -46,7 +46,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/microsoft/plcrashreporter.git", from: "1.11.2"),
-        .package(url: "https://github.com/open-telemetry/opentelemetry-swift.git", exact: "1.6.0")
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -46,7 +46,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/microsoft/plcrashreporter.git", from: "1.11.2"),
-        .package(url: "https://github.com/open-telemetry/opentelemetry-swift.git", exact: "1.6.0")
```
ganeshnj, line 49: once https://github.com/DataDog/opentelemetry-swift-packages/pull/17 is merged, we will take it back to 1.6.0.

ncreated (approved):

PR #1940, opened 2024-07-03
RUM-5084 feat: add ability to clear data in feature data storage

ganeshnj:
### What and why?

We have `clearAllData` today which clears the events but doesn't clear the feature data storage that we use for persisting data for feature likes app hangs, watchdog terminations etc.

### How?

Introduce new API to clear data in the feature data storage.

### Review checklist

### DIFF
--- a/DatadogCore/Sources/Core/Storage/Directories.swift
+++ b/DatadogCore/Sources/Core/Storage/Directories.swift
@@ -35,11 +35,22 @@ internal struct CoreDirectory {
             authorized: try coreDirectory.createSubdirectory(path: "\(name)/v2")
         )
     }
+
+    /// Obtains the path to the data store for given Feature.
+    ///
+    /// Note: `FeatureDataStore` directory is created on-demand which may happen before `FeatureDirectories` are created.
+    /// Hence, this method only returns the path and let the caller decide if the directory should be created.
+    ///
+    /// - Parameter name: The given Feature name.
+    /// - Returns: The path to the data store for given Feature.
+    func getDataStorePath(forFeatureNamed name: String) -> String {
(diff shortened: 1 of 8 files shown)


DatadogCore/Sources/Core/Storage/Directories.swift
```diff
@@ -31,6 +31,7 @@ internal struct CoreDirectory {
     /// - Returns: The Feature's directories
     func getFeatureDirectories(forFeatureNamed name: String) throws -> FeatureDirectories {
         return FeatureDirectories(
+            root: try coreDirectory.createSubdirectory(path: "\(name)"),
```
ncreated, line 34: question/ The `root` doesn't seem to be used in production code. Is it necessary?

I only see `root` being used in tests (to access DS directory), which introduces large inconsistency:
- the real path to feature DS is [created here](https://github.com/DataDog/dd-sdk-ios/blob/72c95e5edadc6c3bfc2dbaf563f2e92854c7e5cd/DatadogCore/Sources/Core/DataStore/FeatureDataStore.swift#L38);
- and here, the `root` targets exactly the same path without keeping both in sync.

We may benefit from having `dataStorePath` defined in `FeatureDirectories` and used by `FeatureDataStore`. This will centralize the DS path management and solve the inconsistency. WDYT @ganeshnj ?

ganeshnj: On a high level it makes sense but current `getFeatureDirectories` API is throws which makes it harder to integrate with existing system.

However, I made some changes and introduced a new API which is similar that returns the path of the data store which can be used to create dir later on. This also aligns perfectly with the approach we took where data store dir is created on demand.

ncreated (approved):
🎯

ncreated (approved):

PR #1921, opened 2024-06-24
RUM-4079 chore: Migrate tools and smoke tests to GitLab

ncreated:
### What and why?

📦 🧰 Following https://github.com/DataDog/dd-sdk-ios/pull/1910, this PR migrates next portion of CI setup to GitLab:
- ✅ Smoke tests:
   - whole SDK integration (iOS + tvOS) with Cocoapods, Carthage, XCFrameworks and SPM
   - SPM build for iOS, tvOS, macOS, macOS through Mac Catalyst and visionOS
- ✅ Tooling tests
   - unit tests for `http-server-mock`, `rum-models-generator`, `sr-snapshots` and distribution package (python)

The following integrations are still running on Bitrise:

- ❌ SR snapshot tests - these will be moved to GitLab after upgrading our macOS runner to Xcode 15.4 and installing the iOS 17.5 runtime.
- ❌ Dogfooding automation
- ❌ Release automation

I will address the remaining items in subsequent PR(s).

### How?

General remarks:
- The new `smoke-test` stage uses `macos:ventura` runner (lower spec). This is to not consume parallelisation budget of `macos:sonoma` which can run `test` and `test-ui` for other branches while smoke tests are still performed on the current branch.
- Xcode 15 is not activated by default in `macos:ventura`, so I added an option to `runner-setup.sh` to activate it. Ultimately, this will be done in the AMI (runner image).
- I renamed `dependency-manager-tests` to `SmokeTests` for clarity and to better fit our conventions.
- All smoke tests are ran via distinct `Makefiles` stored at `SmokeTests/<dep manager>/Makefile`. This is not different from what we had before. To better streamline these tests, each smoke `Makefile` now distinguishes `make clean install test` commands.
- I also moved `instrumented-tests/http-server-mock` to `tools/http-server-mock` to better fit repo layout.

### Makefile

Same as in #1910, smoke and tool tests are ran from the main Makefile. List of new key commands:
- Tool Tests:
   - `make tools-test` - Runs tests for all repo tools.
- Smoke Tests:
   - `make smoke-test` - Runs smoke tests in specified TEST_DIRECTORY (using OS, PLATFORM and DEVICE)
   - `make smoke-test-ios-all` - Runs all smoke tests adequate for iOS.
   - `make smoke-test-tvos-all` - Runs all smoke tests adequate for tvOS.
   - `make spm-build` - Builds `Package.swift` for specified SCHEME and DESTINATION.
   - `make spm-build-ios` - Builds `Package.swift` for iOS.
[...]

PR #1910, opened 2024-06-18
RUM-4079 chore: Migrate lint, tests and ui-tests to GitLab

ncreated:
### What and why?

📦 🧰 This PR migrates a significant portion of our CI setup to GitLab:
- ✅ Lint
- ✅ Unit tests
- ✅ UI tests

The following integrations are still running on Bitrise:
- ❌ SR snapshot tests - these will be moved to GitLab after upgrading our macOS runner to Xcode 15.4 and installing the iOS 17.5 runtime.
- ❌ Smoke tests
- ❌ Tooling tests
- ❌ Dogfooding automation
- ❌ Release automation

I will address the remaining items in subsequent PR(s).

### How?

General Remarks:
- The performance and stability of tests on the GitLab runner are comparable to running them on a local laptop.
- We're using a dedicated `macos:sonoma` runner, which currently lacks some components. I plan to update the AMI; however, in the meantime, missing components are installed at runtime and persisted, minimizing setup times.
- For convenience of argument parsing in `.sh` scripts, I employed [argparse-sh](https://github.com/yaacov/argparse-sh) via my fork [argparse-zsh](https://github.com/ncreated/argparse-zsh).

### Makefile
GitLab integrates changes through `make` commands, structured as follows:
```
.gitlab-ci.yml → Makefile → ./tools/*
```

Main motivations for the Makefile design:
- Modularity: This allows us to easily replicate CI commands locally by copying them directly from the CI logs, e.g.:

Screenshot 2024-06-19 at 11 47 07

- Utility: Several commands facilitate local development.

List of key commands:
- Utility & Setup:
  - `make env-check` - Validates the presence of required CLI tools and displays their versions.
  - `make repo-setup` - Prepares the repo for development and testing (ENV="dev" | "ci").
  - `make clean` - Resets the repository state, clearing prebuilt artifacts and configurations.
- Lint:
  - `make lint` - Lints SDK and tests code.
  - `make license-check` - Ensures Apache 2.0 license headers are present.
- Unit Testing:
  - `make test` - Runs unit tests based on specified parameters (SCHEME, OS, PLATFORM, DEVICE).
  - `make test-ios` - Runs iOS Simulator unit tests for a specific scheme.
  - `make test-ios-all` - Runs iOS tests for all schemes.
  - `make test-tvos` - Runs tvOS Simulator unit tests for a specific scheme.
  - `make test-tvos-all` - Runs tvOS tests for all schemes.
- UI Testing:
[...]

PR #636, opened 2021-10-14
RUMM-1667 Enhance release automation

ncreated:
### What and why?

📦 This PR proposes enhancement to our release automation model. Compared to previous model:
- It no longer creates `Datadog-x.y.z.zip` archive in current directory, instead it clones the repo to temporary folder which ensures clean state for building `XCFrameworks`. It removes the need for [`prepare()` and `cleanup()`](https://github.com/DataDog/dd-sdk-ios/blob/920464537fd09d0b7166ff030fae9040138aab01/tools/distribution/make_distro_builds.sh#L26-L42) operations and guarantees that no dirty artefacts will be included in the `.zip` (in `1.7.0`, `1.7.1` and `1.8.0-beta1` we accidentally included `HTTPServerMock.xcframework`).
- It runs integrity checks to ensure that [`sdk_version`](https://github.com/DataDog/dd-sdk-ios/blob/920464537fd09d0b7166ff030fae9040138aab01/Sources/Datadog/Versioning.swift#L3) and [`spec.version`](https://github.com/DataDog/dd-sdk-ios/blob/920464537fd09d0b7166ff030fae9040138aab01/DatadogSDK.podspec#L4) match the git tag and validates the content of produced `Datadog-x.y.z.zip` archive.
- It supports `--dry-run` mode for testing and debugging locally and on CI (with `DD_RELEASE_DRY_RUN=1` ENV).
- It allows selective run for particular release tags (also on CI) which will be necessary for fixing `1.7.0`, `1.7.1` and `1.8.0-beta1` GH assets.

#### Updated Bitrise pipelnie

I reshaped our Bitrise CI pipeline to spin off two distinct jobs, one for publishing GH Release asset, one for pushing pods:
```
tagged_commit:               --- (tests) ---> ✅
publish_github_asset:                       ↘ --(make .zip + validate + upload)--> ✅
publish_cocoapods_podspecs:                 ↘ --(lint specs + push specs)-----------------------> ✅
```
I used the official [Bitrise Start Build](https://www.bitrise.io/integrations/steps/build-router-start) step instead of calling Bitrise API with `curl` (it gives us ENV propagation from `tagged_commit` to child jobs for free). It is possible to support Github Check by making the `tagged_commit` await child jobs completion, but in our case this will always end up with a CI timeout (`90min`) as Cocoapods specs push might take `~1h`. Instead, eventual failure from child jobs is reported to our Slack channel.

#### New CLI tooling

I replaced Bash scripts with more flexible Python CLI tool:
```
[...]

PR #1945, opened 2024-07-05
RUM-4079 chore: Migrate release automation to GitLab

ncreated:
### What and why?

📦 🧰 This PR migrates release automation to GitLab, following PRs [#1921](https://github.com/DataDog/dd-sdk-ios/pull/1921) and [#1910](https://github.com/DataDog/dd-sdk-ios/pull/1910).

Remaining automations in Bitrise:
- ❌ SR snapshot tests - will be migrated next
- ❌ Dogfooding automation
- ❌ E2E tests app upload to s8s

#### Overview

The previous `release.py` (Python) automation has been replaced with `make` + shell scripts, retaining the same release model. I made few improvements to the failure resistance of publishing Cocoapods podspecs, which are prone to timeouts. Failed CP jobs can now be safely retried, and already pushed specs will be skipped.

#### GitLab Release Pipeline

The release pipeline consists of 5 sequential jobs:
1. **Build Artifacts** - Creates and validates two artifacts for the next jobs:
   - `artifacts/<tag>/Datadog.xcframework.zip` - XCFrameworks bundle for GH Release;
   - `artifacts/<tag>/dd-sdk-ios` - fresh clone of the repo for linting and publishing Cocoapods podspecs.
2. **Publish GH Asset** - Publishes `Datadog.xcframework.zip` to GH Release.
3. **Publish CP podspecs (internal)** - Lints and pushes `DatadogInternal.podspec` to trunk, starting only after (2) succeeds to avoid leaving the release in an inconsistent state.
4. **Publish CP podspecs (dependent)** - Lints and pushes main SDK podspecs (Core, Logs, Trace, RUM, SR, CR, and WVT) 20 minutes after (3) succeeds, anticipating the availability of `DatadogInternal.podspec` in the CP trunk.
5. **Publish CP podspecs (legacy)** - Lints and pushes legacy podspecs (for backward compatibility with V1), including `DatadogObjc.podspec`, 20 minutes after (4) to ensure main podspecs are available.

<p align="center">

</p>

All "publish" jobs use the [`needs`](https://docs.gitlab.com/ee/ci/yaml/#needs) syntax for dependencies.

#### Starting Release Pipeline Manually (`DRY_RUN`)

For debugging and troubleshooting, the release pipeline can be manually started with `RELEASE_GIT_TAG` and `RELEASE_DRY_RUN` CI env variables. Setting `RELEASE_DRY_RUN=1` runs the entire pipeline but skips publishing artifacts. `RELEASE_GIT_TAG` can point to any existing release tag for testing.

<p align="center">

</p>

#### CI Secrets

Introducing this automation required implementing the flow for managing CI secrets. Following [internal guidelines](https://datadoghq.atlassian.net/wiki/x/54EE1), secrets are managed using [Vault](https://www.vaultproject.io/). Vault is already configured in CI runners. To authenticate it on local machines, I added convenient automation in `tools/secrets` accessible through `make set-ci-secret`:

<p align="center">

</p>

It uses OpenID to authorize operations through a web browser and existing Google account sessions. Everyone on the team should already be authorized to manage CI secrets. Follow [this guide](https://datadoghq.atlassian.net/wiki/x/cIEB4w) (internal) for more info.

#### Building XCFrameworks

The `tools/release/build-xcframeworks.sh` added in this PR is an enhanced version of the [previous `build-xcframework.sh` script](https://github.com/DataDog/dd-sdk-ios/blob/2.14.0/tools/distribution/build-xcframework.sh). The core logic remains unchanged, but it now supports `--ios` and `--tvos` options for platform-specific slices. While both flags are used in the release pipeline, they are used separately in iOS and tvOS smoke tests, improving performance:
- **Smoke Tests (iOS)** is now 3 minutes faster 🏎️ (~22min -> ~19min)
- **Smoke Tests (tvOS)** is now 6 minutes faster 🏎️ (~19min -> ~13min)

#### 🎁 Extra bits / not done in this PR

No Slack notifications are implemented in the release pipeline yet. I plan to work on this separately, implementing notifications for all key pipelines.

Future plans include automating:
- Git-flow PRs for releases (with [`gh pr create`](https://cli.github.com/manual/gh_pr_create))
- GH Release for tags, including release notes (with [`gh release create --generate-notes`](https://cli.github.com/manual/gh_release_create))

### Review checklist


tools/release/build-xcframeworks.sh
```diff
@@ -0,0 +1,151 @@
...
+    # Therefore, we cannot provide ABI stability and we have to supply '-allow-internal-distribution'.
+    xcodebuild -create-xcframework -allow-internal-distribution ${xcoptions[@]} -output "$XCFRAMEWORKS_OUTPUT/$product.xcframework" | xcbeautify
+
+    echo_succ "The '$product.xcframework' was created successfully in '$XCFRAMEWORKS_OUTPUT'"
+}
+
+echo_info "cd '$REPO_PATH'"
+cd $REPO_PATH
+
+check_repo_clean_state
```
ganeshnj, line 117: in what case this can fail in CI?

Would it makes sense to have an option to clear the repo state which is `true` in CI and `false` for our local testing?

tools/release/build.sh
```diff
@@ -0,0 +1,62 @@
...
+create_zip_archive () {
+    echo_subtitle "Create '$XCF_ZIP_NAME' in '$ARTIFACTS_PATH'"
+    cd "$ARTIFACTS_PATH" && zip -r -q $XCF_ZIP_NAME $XCF_DIR_NAME
+    cd -
+}
+
+rm -rf "$ARTIFACTS_PATH"
+mkdir -p "$ARTIFACTS_PATH"
+
+clone_repo
```
ganeshnj, line 54: Could we clean the repo state rather cloning a new copy?

Essentially `git clean -fxd`

tools/runner-setup.sh
```diff
@@ -103,3 +106,27 @@ if [ "$visionOS" = "true" ]; then
...
+        get_secret $DD_IOS_SECRET__SSH_KEY > $SSH_KEY_PATH
+        chmod 600 "$SSH_KEY_PATH"
+
+        cat <<EOF > "$HOME/.ssh/config"
+Host github.com
+    HostName github.com
+    User git
+    IdentityFile $SSH_KEY_PATH
+    StrictHostKeyChecking no
+EOF
```
ganeshnj, line 127: I imagine we need this to clone the repo but per my comment above, we shouldn't need to clone it again and just reset the state with some git commands.

tools/secrets/set-secret.sh
```diff
@@ -0,0 +1,99 @@
...
+
+    if [[ -f "$SECRET_FILE" ]]; then
+        SECRET_VALUE=$(cat "$SECRET_FILE")
+    else
+        echo_err "Error: File '$SECRET_FILE' does not exist."
+        exit 1
+    fi
+}
+
+select_input_method() {
```
ganeshnj, line 63: how is it going to work in a CI env? How can you enter manually in gitlab CI during the execution?

🤔

ganeshnj (commented):
Some minor comments but overall right on the track. Great continuation.

tools/secrets/set-secret.sh
ncreated: The `set-secret.sh` flow is made for local use only. We don't **set** secrets from CI - CI only **gets** them (through `get-secret.sh`).

tools/release/build.sh
ncreated: Yes, we can, but this would significantly change the existing release model introduced to Bitrise in [this PR](https://github.com/DataDog/dd-sdk-ios/pull/636). While migrating things to GitLab I wanted to keep it simple, not exposed to new challenges.

In [#636](https://github.com/DataDog/dd-sdk-ios/pull/636), we switched from using "cleanup" scripts to performing a fresh clone of the repo for two reasons:
1. Overlooking some dirty state during "cleanup" led to shipping unnecessary content in three release artifacts.
2. There was no CI-automated way to re-build artifacts for past releases using the new tooling.

While `git clean -fxd` is definitely better than the [custom "cleanup"](https://github.com/DataDog/dd-sdk-ios/blob/920464537fd09d0b7166ff030fae9040138aab01/tools/distribution/make_distro_builds.sh#L26-L42) we used to have before [#636](https://github.com/DataDog/dd-sdk-ios/pull/636), performing a release from the current repo doesn't offer the same level of flexibility in CI. Specifically:

- With the current model, it is possible to test the entire release pipeline on any branch by targeting an existing tag (using `RELEASE_GIT_TAG` and `RELEASE_DRY_RUN` options). I leveraged this several times to test this PR.
- In rare situations, we had to modify release tooling and re-build artifacts for existing tags using a new version of Xcode on CI (possible by using `RELEASE_GIT_TAG=<past release>`, `RELEASE_DRY_RUN=0`, and `OVERWRITE_EXISTING=1`).

Likely, we don't need such a level of CI flexibility these days ☝️, as module stability [was finally fixed in `2.x`](https://datadoghq.atlassian.net/wiki/x/KQFfiw).

However, cleaning the current repo for developing release tooling **in local** might be very cumbersome and impractical, given that we track a lot of dev configuration through `.local.xcconfig` files (which would be wiped out). Performing a fresh clone of the repo into `artifacts/<tag>/` is more convenient and less invasive.

Let me know how does it sound @ganeshnj

tools/runner-setup.sh
ncreated: Sure 👌 (answered in other comment). However, the `--ssh` setup for runner will be also utilized for [SR Snapshot Tests](https://datadoghq.atlassian.net/wiki/x/nIxdu) and dogfooding automations that I will be migrating next (both do git clone). It is also used for XCFramework Smoke Tests in this PR.

tools/release/build.sh
ncreated: > but I don't think it is a good approach to have something released other than what the current branch is triggered from

This is definitely not the case here. For `tag` triggers, the flow is this:
- GitLab clones the repo to `~/dd-sdk-ios` for given `GIT_TAG`
- `~/dd-sdk-ios/tools/release/build.sh` starts
- it clones the repo for **the same** `GIT_TAG` into `~/dd-sdk-ios/artifacts/<GIT_TAG>/dd-sdk-ios`
- finally, `~/dd-sdk-ios/tools/release/` automation:
   - builds and publishes XCFramework for `~/dd-sdk-ios/artifacts/<GIT_TAG>/dd-sdk-ios`
   - publishes Cocoapod podspecs for `~/dd-sdk-ios/artifacts/<GIT_TAG>/dd-sdk-ios`

Ultimately, the automation will release exactly the hash that it was triggered for, so we can't speak of anti-patterns.

Then, I agree that this implementation detail might look custom, but it was made intentionally to address concrete problems we hit when releasing SDK that supports several iOS dependency managers, each operating in different model.

> Let's not block the PR 

👍 Also note that all `tools/release/` scripts introduced in this PR do support path parameters making the release of `.` instead of `artifacts/<GIT_TAG>/dd-sdk-ios` very simple to achieve.

tools/release/build-xcframeworks.sh
ncreated: ✅ Replaced with `git clean -fxd` inside `artifacts/<GIT_TAG>/dd-sdk-ios`. It's still defensive (the repo is cloned so must be clean), but I'd like to keep it so we can leverage this if we decide to use current repo for building artifacts as discussed in other comments.

ganeshnj (dismissed):

tools/release/build.sh
ganeshnj: Thanks for clarification.

mariedm (approved):

ganeshnj (approved):

ncreated:
Dry run: ✅ https://gitlab.ddbuild.io/DataDog/dd-sdk-ios/-/pipelines/38803721

PR #1918, opened 2024-06-21
[Issue #125] Enable DatadogCore and DatadogLogs to compile on watchOS platform

jfiser-paylocity:
### What and why?
Related to issue https://github.com/DataDog/dd-sdk-ios/issues/125.

This PR resolves compilation errors for watchOS platform in the following targets:
- DatadogInternal
- DatadogCore
- DatadogLogs

### How?
Minimum required version of watchOS 7 has been added to SPM/Cocoapods/Carthage definitions. The biggest issue on lower version is the lack of system notifications (didBecomeActive, didEnterBackground, willResignActive, willEnterForeground).

There are couple of limitations on watchOS platform overall, however I don’t see these as blockers:
- No background task - Logs will still sync when the app is on the foreground.
- HTTP header `Proxy-Authorization` is not set - Could be mitigated by using String static values
- Kronos library always returns local time

To test this, you can temporarily add Apple Watch under Supported Destinations in respective targets. However it might be good to ensure at compilation time that watchOS platform does not break in the future.

### Review checklist
- [ ] Feature or bugfix MUST have appropriate tests (unit, integration)
- [ ] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes

### Custom CI job configuration (optional)
- [ ] Run unit tests for Session Replay
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

### Example
I tried to integrate fork changes directly in the POC project of mine to validate it works. See below is the example log opened in web Datadog dashboard.

image

Issue #125, opened 2020-06-03
Support watchOS


mozeryansky:
I had implemented a custom datadog internal sdk for logging before this official SDK was available. I have both watchOS and iOS applications. However, I can't switch to using this SDK because it relies on certain UIKit components not in watchOS.

Could we get watchOS support?

ncreated:
Hello @mozeryansky 👋 , thanks for reaching us out 🙂! This question is similar to #78 (`macOS` support). Our strategy for the iOS SDK is to first deliver all three major features: [Logging](https://docs.datadoghq.com/logs/), [Tracing](https://docs.datadoghq.com/tracing/) and [Real User Monitoring](https://docs.datadoghq.com/real_user_monitoring/installation/?tab=us). After that point, we will re-evaluate supporting more platforms (including `watchOS`).

Although we don't have any time estimations for more platforms support, the facts are: Logging feature is already done and we're close to starting Tracing beta. After this, we'll work on RUM.

Please stay updated to our releases, to be notified when the `watchOS` is eventually supported 🙂. And we're glad you use Datadog 🚀.

ncreated:
@mozeryansky Sure, let's keep it alive for more visibility 👌.

n-ravichandran:
@ncreated Is there are update on this?

antonmartinsson:
Still no updates on this @ncreated? Has any kind of evaluation like you mentioned taken place?

MaelNamNam:
Hello and thanks for bumping this up, but unfortunately we still don’t have plans to support WatchOS with our SDK in the near future. However, the need has surfaced again a few times in the past few months and we might re-evaluate our priorities. Therefore, could you escalate a feature request through your support channels so we can properly track demand please?

antonmartinsson:
Bummer! I'll gladly create a feature request. As someone who hasn't been in contact with your support before, what's the best way to do so? [This page](https://help.datadoghq.com/hc/en-us/requests/new)?

MaelNamNam:
@antonmartinsson yes this is it! Please select `General Issue` in the "Impact" dropdown :)

dfed:
Was this meant to be closed as "completed"? Or is this a "won't do"?

maciejburda:
@dfed Thanks for double checking with us!

In this case it's neither. It's on our list, but unfortunately it falls behind other priorities as of now.

[...]

PR #1946, opened 2024-07-08
[Issue #125] Enable `DatadogCore` and `DatadogLogs` to compile on watchOS platform 

ncreated:
### What and why?

Merging https://github.com/DataDog/dd-sdk-ios/pull/1918 (external) to `develop`.

### How?

The https://github.com/DataDog/dd-sdk-ios/pull/1918 was first merged to internal branch. Here w run full CI and do last QA before merging it to `develop`.

### Review checklist


mariedm (approved):
LGMT, but I haven't tried to compile it nor ran this branch in a watchOS app.

DatadogInternal/Sources/Context/DeviceProtocol.swift
```diff
@@ -0,0 +1,38 @@
...
+    /// Get the current device
+    public static var current: DeviceProtocol {
+        #if canImport(WatchKit)
+        WKInterfaceDevice.current()
+        #else
+        UIDevice.current
+        #endif
+    }
+}
+#endif
```
maxep, line 38: /blocking Looks like the compilation statement is wrong here, does it actually compile on watchOS? 🤔

maxep: Actually, I think we should remove this file and only relay on `DeviceInfo` implementation, which is already based on compile statement for `UIKit` and `macOS`. We should add `watchOS` there IMO.

maxep (changes requested):

DatadogInternal/Sources/Context/DeviceProtocol.swift
maxep: @ncreated, if you agree I can make the change 👍

maxep (approved):

PR #1960, opened 2024-07-17
RUM-1000 chore: use throwing API to create the file

ganeshnj:
### What and why?

The error message we get today is non actionable and doesn't provide information why the file create failed.

```
InternalError: Failed to write data to file - Cannot create file at path: /var/mobile/Containers/Data/Application/E08E89DC-F1F2-496E-BEB0-B5D9E9359F8E/Library/Caches/com.datadoghq/v2/5168944e862876aa5a9ad4e5df4a654f6d54deca3edc50e66544bff2a690103e/rum/v2/742819228496
```

### How?

Replaced file creation with a API which throws.

### Review checklist

### DIFF
--- a/DatadogCore/Sources/Core/Storage/Files/Directory.swift
+++ b/DatadogCore/Sources/Core/Storage/Files/Directory.swift
@@ -7,6 +7,10 @@
 import Foundation
 import DatadogInternal
 
+extension Data {
+    static let empty = Data()
+}
+
 /// Provides interfaces for accessing common properties and operations for a directory.
 internal protocol DirectoryProtocol: FileProtocol {
     /// Returns list of subdirectories in the directory.
@@ -106,9 +110,7 @@ internal struct Directory: DirectoryProtocol {
     /// Creates file with given name.
     func createFile(named fileName: String) throws -> File {
         let fileURL = url.appendingPathComponent(fileName, isDirectory: false)
-        guard FileManager.default.createFile(atPath: fileURL.path, contents: nil, attributes: nil) == true else {
-            throw InternalError(description: "Cannot create file at path: \(fileURL.path)")
-        }
+        try Data.empty.write(to: fileURL, options: .atomic)
         return File(url: fileURL)
     }
 


DatadogCore/Sources/Core/Storage/Files/Directory.swift
```diff
@@ -106,9 +110,7 @@ internal struct Directory: DirectoryProtocol {
     /// Creates file with given name.
     func createFile(named fileName: String) throws -> File {
         let fileURL = url.appendingPathComponent(fileName, isDirectory: false)
-        guard FileManager.default.createFile(atPath: fileURL.path, contents: nil, attributes: nil) == true else {
-            throw InternalError(description: "Cannot create file at path: \(fileURL.path)")
-        }
+        try Data.empty.write(to: fileURL, options: .atomic)
```
ncreated, line 113: question/ Do we have any insights into what kind of errors it can throw? What can we expect to see in telemetry?

ncreated (approved):
I like the idea, very simple 👌. I was wondering if it would make sense to include the `track` (`rum` | `logs` | ...) as an extra context, but rather it won't tell us anything interesting. By their mean, I/O errors are rather generic and do not relate to particular feature. Sharing in case we have more thoughts on this @ganeshnj

DatadogCore/Sources/Core/Storage/Files/Directory.swift
ganeshnj: Assuming this is 1:1 mapping of https://developer.apple.com/documentation/foundation/nsdata/1414800-writetofile and it uses syscall to perform file IO, I can expect https://man7.org/linux/man-pages/man2/write.2.html#top_of_page:~:text=are%20not%20specified.-,ERRORS,-top errors.

ganeshnj:
> I like the idea, very simple 👌. I was wondering if it would make sense to include the `track` (`rum` | `logs` | ...) as an extra context, but rather it won't tell us anything interesting. By their mean, I/O errors are rather generic and do not relate to particular feature. Sharing in case we have more thoughts on this @ganeshnj

I would start with this and it we are still missing some context, then try to add more information, IMO track wouldn't be that useful in this case (just a hunch) but let's see.

PR #1957, opened 2024-07-11
RUM-4079 chore: Migrate E2E Tests app workflow to GitLab

ncreated:
### What and why?

📦 🧰 This PR migrates E2E Tests workflow to GitLab (build → sign → upload to synthetics). It follows other GitLab migration efforts: https://github.com/DataDog/dd-sdk-ios/pull/1953, https://github.com/DataDog/dd-sdk-ios/pull/1945, https://github.com/DataDog/dd-sdk-ios/pull/1921 and https://github.com/DataDog/dd-sdk-ios/pull/1910.

The only remaining automation in Bitrise is ❌ Dogfooding. It will be migrated next.

### How?

This work is leveraging existing [`code-sign.sh`](https://github.com/DataDog/dd-sdk-ios/blob/2.14.1/tools/code-sign.sh) flow. Previous script was renamed and modularized with enhanced errors handling. It now creates all code-signing files (certificate, provisioning profile, and xcconfig) from CI secrets and delegates `archive`, `export` and `upload` routines to `E2ETests/Makefile` as before.

### Review checklist
- [ ] Feature or bugfix MUST have appropriate tests (unit, integration)
- [ ] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes

### Custom CI job configuration (optional)
- [ ] Run unit tests for Session Replay

PR #1961, opened 2024-07-18
RUM-4079 chore: Migrate dogfooding automation to GitLab

ncreated:
### What and why?

📦 🧰 This PR migrates Dogfooding automation to GitLab. It comes with multiple quality of life improvements to streamline our dogfooding process:
- Dogfooding from `develop` directly.
- The `sdk_version` will now ponit to dogfooded version (e.g. `2.14.1+f3024dc4`).
- Dogfooding PRs will now include CHANGELOG, listing PRs since previous dogfooding.
- Commits in dogfooding PRs are now signed 🔐 🎁.

It follows other GitLab migration efforts: https://github.com/DataDog/dd-sdk-ios/pull/1957, https://github.com/DataDog/dd-sdk-ios/pull/1953, https://github.com/DataDog/dd-sdk-ios/pull/1945, https://github.com/DataDog/dd-sdk-ios/pull/1921 and https://github.com/DataDog/dd-sdk-ios/pull/1910. With this merge, there will be no other meaningful automations remaining in Bitrise.

### How?

Dogfooding is a new GitLab stage, which can be started manually for `develop` commits:

Screenshot 2024-07-18 at 10 43 08

The previous `dogfood.py` (Python) automation has been replaced with `make` + `dogfood.sh`:
- `make dogfood-shopist`
- `make dogfood-datadog-app`

There are two Python scripts called by `dogfood.sh`:
- `update-dependency.py` - updates `Package.resolved` in dependant project.
- `read-dogfooded-hash.py` - reads last dogfooded commit from dependant project to generate changelog.

Same as before, the dogfooding process is:
- clone dependant repo
- update `Package.resolved` in dependant project
- commit & push to `dogfooding-<dogfooded commit>` branch
- open PR in dependant repo

#### 🎁 New things
We're now dogfooding from `develop` branch directly, meaning no more merges required to `dogfooding` branch. The `dogfooding` can be deleted after we make sure dependant projects no longer need it.

Each PR opened to dependant repo will now include CHANGELOG that lists all PRs merged to `dd-sdk-ios` since previous dogfooding. See these test PRs:
- https://github.com/DataDog/shopist-ios/pull/101
- https://github.com/DataDog/datadog-ios/pull/1544

The `sdk_version` in dependant projects is now updated to include dogfooded commit hash. This will tell us exactly which SDK version is being used:

349851758-7587ae3d-73cb-4e02-8a19-ad744ad3f72e

^ after round of feedback, this is now using `+` instead of `-` to conform with https://semver.org/

In order to make it all work, these two PRs to dependant repos must be merged before we start using it:
- https://github.com/DataDog/shopist-ios/pull/98
- https://github.com/DataDog/datadog-ios/pull/1543

### Review checklist


tools/dogfooding/dogfood.sh
```diff
@@ -0,0 +1,271 @@
...
+    echo_info "▸ Changelog:" >&2
+    echo_info ">>> changelog begin" >&2
+    echo_succ "$CHANGELOG" >&2
+    echo_info "<<< changelog end" >&2
+
+    echo "$CHANGELOG"
+}
+
+# Updates dd-sdk-ios version in dependant project to DOGFOODED_COMMIT.
+update_dependant_package_resolved() {
```
ganeshnj, line 196: /improvement

given we are always going to have `develop` as target branch, we should use `xcodebuild -resolvePackageDependencies`. This is future proof solution and we don't have to maintain updating Package.resolved file logic.

This may eventually update all the dependencies but that something is expected because that's how they are being configured in the project. (minor, patch upgrade rules).

tools/dogfooding/read-dogfooded-hash.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    print(f'DOGFOODED_HASH={hash}')
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Reads the hash of dd-sdk-ios dependency in "Package.resolved" of SDK-dependant project and prints it to STDOUT.')
+    parser.add_argument('--repo-package-resolved-path', type=str, required=True, help='Path to "Package.resolved" file in SDK-dependant project')
+    args = parser.parse_args()
+    
+    try:
+        dependant_package = PackageResolvedFile(path=args.repo_package_resolved_path)
```
ganeshnj, line 28: /question

Could we rely on `__dogfoodedSDKVersion` to keep it simple?

ganeshnj (commented):
Looks great, couple of suggestion to make it robust to internal working of package.resolved file.

tools/dogfooding/read-dogfooded-hash.py
ncreated: Good idea, I like the concept of using `__dogfoodedSDKVersion` as source of truth here 👍. However, we use `dogfooded-hash` for later query in `git log` (changelog) and that needs full commit hash vs `__dogfoodedSDKVersion` only encodes the short one for readability. I will check what we can do here. 

Ideas @ganeshnj ?

tools/dogfooding/dogfood.sh
ncreated: Unless I miss something I think this could only work for Shopist which only depends on `dd-sdk-ios`. It won't work for Datadog app which has dozens of other dependencies and uses [custom build system](https://github.com/DataDog/datadog-ios/blob/develop/README.md#build). We don't want to dictate the way of managing dependencies in `datadog-ios`.

tools/dogfooding/read-dogfooded-hash.py
ganeshnj: the short form is totally fine here, git handles it gracefully. You will almost never hit a conflict.

tools/dogfooding/dogfood.sh
ganeshnj: this is actually a blessing and curse, now our SDK has dependencies on other SPM packages, we want them also get updated when `dd-sdk-ios` updates them. I think, the current solution won't work.

Also, this is going to make PR and we are going to review. So should be fine to spot things that are not looking good. If the other dozens of dependencies we have are not suppose to be updated, we must use `exact` condition to version control.

ganeshnj: and probably the reality is, when you do xcodebuild, it is eventually going to update the dependencies anyway. We are not gaining anything here.

ncreated: > now our SDK has dependencies on other SPM packages, we want them also get updated when dd-sdk-ios updates them. I think, the current solution won't work

We [have this covered](https://github.com/DataDog/dd-sdk-ios/blob/65854b3fe0a2aff3dec470e3f74835d1abffe2c7/tools/distribution/dogfood.py#L62-L81) for a long time and it is [migrated as part of this PR](https://github.com/DataDog/dd-sdk-ios/blob/201b40247c518297de875597415402f019c61327/tools/dogfooding/update-dependency.py#L37-L55) with no change. In terms of updating `Package.resolved` in dependent projects, this PR doesn't bring any new logic. 

> If the other dozens of dependencies we have are not suppose to be updated, we must use exact condition to version control

Again, this is asking to change process of dependency management in different repo (`datadog-ios`) and another team, which is not something we can conclude here. I get the improvement idea, but it doesn't seem doable in this PR.

ganeshnj: If `datadog-ios` team is not using `disableAutomaticPackageResolution` flag during xcodebuild, they are already doing what I'm proposing, but I also agree on the ownership ambiguity.

tools/dogfooding/read-dogfooded-hash.py
ncreated: Done ✅

tools/dogfooding/update-dependency.py
ncreated, line 1: TBC: This script isn't bringing anything new. It is just extract of the minimal piece from the previous dogfooding model.

ganeshnj (approved):
🚀  great work with quality of life improvements.

PR #1966, opened 2024-07-22
RUM-3060 feat: Deprecate Alamofire extension pod

ncreated:
### What and why?

📦 Starting from [2.5.0](https://github.com/DataDog/dd-sdk-ios/releases/tag/2.5.0), RUM can automatically track Alamofire requests, making the DatadogAlamofireExtension pod obsolete. Maintaining this pod is cumbersome as it adds an extra code path to the network instrumentation logic, which became problematic in [issue #1638](https://github.com/DataDog/dd-sdk-ios/issues/1638).

The new network instrumentation API introduced in 2.5.0 can handle Alamofire requests without requiring a separate pod. This solution also works for SPM, Carthage, and XCFramework integration. This PR deprecates DatadogAlamofireExtension in favor of the new instructions explained in the following documentation PR:
- https://github.com/DataDog/documentation/pull/24253

### How?

- Created https://github.com/DataDog/documentation/pull/24253 to explain how to instrument Alamofire (and Apollo GraphQL) requests;
- `DatadogAlamofireExtension.podspec` is marked deprecated;

### Review checklist

### DIFF
--- a/DatadogAlamofireExtension.podspec
+++ b/DatadogAlamofireExtension.podspec
@@ -2,6 +2,11 @@ Pod::Spec.new do |s|
   s.name         = "DatadogAlamofireExtension"
   s.version      = "2.14.1"
   s.summary      = "An Official Extensions of Datadog Swift SDK for Alamofire."
+  s.description  = <<-DESC
+                   The DatadogAlamofireExtension pod is deprecated and will no longer be maintained.
+                   Please refer to the following documentation on how to instrument Alamofire with the Datadog iOS SDK:
+                   https://docs.datadoghq.com/real_user_monitoring/mobile_and_tv_monitoring/integrated_libraries/ios
+                   DESC
 
   s.homepage     = "https://www.datadoghq.com"
   s.social_media_url   = "https://twitter.com/datadoghq"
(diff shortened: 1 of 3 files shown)


mariedm (approved):

DatadogAlamofireExtension.podspec
```diff
@@ -14,6 +19,8 @@ Pod::Spec.new do |s|
     "Maciej Burda" => "maciej.burda@datadoghq.com"
   }
 
+  s.deprecated = true
```
ganeshnj, line 22: do we need to publish a version in order deprecate?

Can we simply delete this? So when customer tries to update (it is customer trigger action) they see an error and fix the issue instead keep using the deprecated version.

ganeshnj (approved):

DatadogAlamofireExtension.podspec
ncreated: We aim at deprecation not removal. The `DatadogAlamofireExtension` will continue to work for all customers in next releases until we finally delete it (in `3.0.0`).

The `deprecated = true` will produce a warning during `pod update` indicating that this pod is being sunsetted. Customers are expected to follow the notice from pod description and land on https://docs.datadoghq.com/real_user_monitoring/mobile_and_tv_monitoring/integrated_libraries/ios (available once https://github.com/DataDog/documentation/pull/24253 is merged) to instrument Alamofire using our `URLSessionInstrumentation` directly.

PR #1967, opened 2024-07-24
Fix: Session Replay TabBar Recorder id generation

maciejburda:
### What and why?

Fixes an issue where TabBar recorder subviews ID generation was inconsistent.

### How?

It was caused by the fact that subtree recorders in the TabBarRecorder where created inline with different UUID each time. I refactored the entire setup and provided additional documentation about providing identifiers.

### Review checklist


mariedm (dismissed):

ncreated (dismissed):
🎯 Good finding, it was tricky one 👍

I like the solution - small and right to the point 👍

maxep (approved):
Thank you 👍

ncreated (approved):

PR #1968, opened 2024-07-24
fix: do not send context updates on context read

ganeshnj:
### What and why?

TLDR: when reading `context` from `FeatureScope` triggers a context update which again publishes context updates again to all the subscribes and we end up in a loop.

This shoots up CPU usage and an app termination likely.

### How?

- Now, when we read the context, we no longer update it.
- a small improvement in WT message receiver

There a few learnings from this bug
- The current pub-sub mechanism is very complex and requires a lot of dive deep before you can start making sense.
- The value pub-sub delivers to its consumers is very simple which we have talked before also. For example, on every context, telemetry, webview or baggage message, each receiver is notified which is wastage of CPU cycles. The receivers must be able to subscribe to individual message or part of message.

### Review checklist

### DIFF
--- a/DatadogCore/Sources/Core/Context/DatadogContextProvider.swift
+++ b/DatadogCore/Sources/Core/Context/DatadogContextProvider.swift
@@ -40,9 +40,7 @@ internal final class DatadogContextProvider {
     /// The current `context`.
     ///
     /// The value must be accessed from the `queue` only.
-    private var context: DatadogContext {
-        didSet { receivers.forEach { $0(context) } }
-    }
+    private var context: DatadogContext
 
     /// The queue used to synchronize the access to the `DatadogContext`.
     internal let queue = DispatchQueue(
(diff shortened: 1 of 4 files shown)


DatadogCore/Sources/Core/Context/DatadogContextProvider.swift
```diff
@@ -110,7 +108,12 @@ internal final class DatadogContextProvider {
     ///
     /// - Parameter block: The block closure called with the current context.
     func write(block: @escaping (inout DatadogContext) -> Void) {
-        queue.async { block(&self.context) }
+        queue.async {
```
mariedm, line 111: Given what you mentioned in the PR description, should we check if the context changed before notifying the receivers?

mariedm (approved):

DatadogCore/Sources/Core/Context/DatadogContextProvider.swift
ganeshnj: this is not so possible with current context definition, we are going to do some followup with a larger refactor.

ncreated (approved):
🎯 Well done 👍

> The current pub-sub mechanism is very complex and requires a lot of dive deep before you can start making sense.
The value pub-sub delivers to its consumers is very simple which we have talked before also. For example, on every context, telemetry, webview or baggage message, each receiver is notified which is wastage of CPU cycles. The receivers must be able to subscribe to individual message or part of message.

💯 agree, looking forward to follow up on this

PR #1891, opened 2024-06-10
RUM-4829 Bump `IPHONEOS_DEPLOYMENT_TARGET` to `12.0`

ncreated:
### What and why?

// TBD

### How?

// TBD

### Review checklist
- [ ] Feature or bugfix MUST have appropriate tests (unit, integration)
- [ ] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes

### Custom CI job configuration (optional)
- [x] Run unit tests for Core, RUM, Trace, Logs, CR and WVT
- [x] Run unit tests for Session Replay
- [x] Run integration tests
- [x] Run smoke tests
- [x] Run tests for `tools/`

PR #1970, opened 2024-07-25
fix: Remove ContextValueReader

maxep:
### What and why?

Follow up #1968 & #1891

The context dispatch on read was keeping the context up-to-date on pre-iOS 12 devices. Because some context values were only accessible by readers (as opposed to publishers) on these legacy OSs.
Now that the minimum OS version is 12, all context values are being published and we no longer need the `ContextValueReader` abstraction.

### Review checklist


mariedm (approved):
Thanks for the clean-up! 🧹

ncreated (approved):
Nice cleanup 👌.

question/ I get the idea, but the rationale is a bit vague to me. Which exact providers from #1891 required this logic? Both:
- `CarrierInfoPublisher` 
- and `NetworkConnectionInfoPublisher`

?

DatadogCore/Sources/Core/Context/DatadogContextProvider.swift
```diff
@@ -65,21 +62,12 @@ internal final class DatadogContextProvider {
...
         subscriptions.forEach { $0.cancel() }
     }
 
-    /// Reads current context.
-    ///
-    /// **Warning:** Must be called from the `queue`.
-    private func unsafeRead() -> DatadogContext {
-        reader.read(to: &self.context)
-        return context
-    }
```
ganeshnj, line 81: It was harder to make sense before given `context` is private property and we were passing the reference outside which can do whatever (essentially modify). Good clearance 🧹

DatadogCore/Sources/Core/Context/DatadogContextProvider.swift
```diff
@@ -138,23 +126,6 @@ internal final class DatadogContextProvider {
...
-    /// The context provider has the ability to a assign a value reader that complies to
-    /// ``ContextValueReader`` to a specific context property. e.g.:
-    ///
-    ///     let reader = ServerOffsetReader<TimeInterval>()
-    ///     provider.assign(reader: reader, to: \.serverTimeOffset)
-    ///
-    /// - Parameters:
-    ///   - reader: The value reader.
-    ///   - keyPath: A context's key path that supports reading from and writing to the resulting value.
-    func assign<Reader>(reader: Reader, to keyPath: WritableKeyPath<DatadogContext, Reader.Value>) where Reader: ContextValueReader {
```
ganeshnj, line 152: 🧹  good clearance, there wasn't any usage of this and just causing confusion.

DatadogCore/Sources/Core/Context/DatadogContextProvider.swift
ganeshnj: Although we are still leaking it in[ L100](https://github.com/DataDog/dd-sdk-ios/blob/maxep/remove-context-reader/DatadogCore/Sources/Core/Context/DatadogContextProvider.swift#L100) which isn't obvious for someone new jumping in.

ganeshnj (approved):
Very good cleanup. Thanks for jumping on this.

DatadogCore/Sources/Core/Context/DatadogContextProvider.swift
maxep: Yes, it wasn't properly documented..
Good catch on [L100](https://github.com/DataDog/dd-sdk-ios/blob/maxep/remove-context-reader/DatadogCore/Sources/Core/Context/DatadogContextProvider.swift#L100) 👍

maxep: It's not actually a leak, it's the mutation block of the context. It would be great to make the context `Equitable`...

maxep:
@ncreated, It was the followings:
- [iOS11CarrierInfoReader](https://github.com/DataDog/dd-sdk-ios/pull/1891/files#diff-40ffc89e84e51099109e16562f6e2cc761c0656689375a611dfce2dffeac5acfL70)
- [SCNetworkReachabilityReader](https://github.com/DataDog/dd-sdk-ios/pull/1891/files#diff-0c266be18d70373a9154097b0cb50a1a24232f532f0920d9422566b2ee894565L107)

Both were readers because of iOS 11 limitations. Now, we no longer use readers and we shouldn't allow them because of the risk it impose to the message-bus.

DatadogCore/Sources/Core/Context/DatadogContextProvider.swift
ganeshnj: that's something we must do, it is in my list
- a simple set to context is not a good indicator to trigger receivers. We must check the diff which eventually means  we need to have `Equatable` conformance.

maxep: Agreed! The issue is with the `featureBaggages` and `userInfo` which is a free form dict.

PR #1986, opened 2024-08-02
RUM-5536 [SR] Start / Stop API

mariedm:
### What and why?

This PR adds the ability for customers to:
- Start and stop a recording 
- Disable automatic recording when enabling Session Replay

See [RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3897426405/RFC+-+Session+Replay+Start+Stop+API#Actions) for more details.

### How?

- Introduce a new `startRecordingImmediately` optional parameter when enabling Session Replay, defaulting to `true`
  - A new field has been added for telemetry as well
  - Models have already been updated in [#216](https://github.com/DataDog/rum-events-format/pull/216) and #1983
- Add new set of `start` and `stop` public APIs to the existing `SessionReplay` public enum, through static functions
- Introduce a `recordingEnabled` variable to `RecordingCoordinator` to track when the user wants to start or stop a recording
- Update the RUM flag `has_replay` accordingly when starting or stopping a recording 
- Modified a bit the logic in `RecordingCoordinator`:
  - The scheduler is only started if the recording has been enabled by the user and the RUM session is sampled
  - We decouple the update of the `has_replay` flag from the RUM view ID; whether it's `nil` or not has no impact on this flag anymore

**Note**: ignoring webviews' snapshots will be tackled separately

### Review checklist

### DIFF
--- a/DatadogSessionReplay/Sources/Feature/Baggages.swift
+++ b/DatadogSessionReplay/Sources/Feature/Baggages.swift
@@ -5,7 +5,6 @@
  */
 
 import Foundation
-import DatadogInternal
 
 /// Defines dependency between Session Replay (SR) and RUM modules.
 /// It aims at centralizing documentation of contracts between both products.
@@ -38,11 +37,4 @@ internal struct RUMContext: Codable, Equatable {
     let viewID: String?
     /// Current view related server time offset
     let viewServerTimeOffset: TimeInterval?
-}
-
-extension DatadogContext {
-    /// The value indicating if replay is being performed by Session Replay.
-    var hasReplay: Bool? {
-        try? baggages[RUMDependency.hasReplay]?.decode()
-    }
 }
--- a/DatadogSessionReplay/Sources/Feature/Baggages.swift
+++ b/DatadogSessionReplay/Sources/Feature/Baggages.swift
@@ -5,6 +5,7 @@
  */
 
 import Foundation
+import DatadogInternal
 
 /// Defines dependency between Session Replay (SR) and RUM modules.
 /// It aims at centralizing documentation of contracts between both products.
@@ -37,4 +38,11 @@ internal struct RUMContext: Codable, Equatable {
     let viewID: String?
     /// Current view related server time offset
     let viewServerTimeOffset: TimeInterval?
+}
+
+extension DatadogContext {
+    /// The value indicating if replay is being performed by Session Replay.
+    var hasReplay: Bool? {
+        try? baggages[RUMDependency.hasReplay]?.decode()
+    }
 }
--- a/DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
+++ b/DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
@@ -20,6 +20,7 @@ internal class RecordingCoordinator {
     let srContextPublisher: SRContextPublisher
 
     private var currentRUMContext: RUMContext? = nil
+    @ReadWriteLock
     private var isSampled = false
 
     /// `recordingEnabled` is used to track when the user 
--- a/DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
+++ b/DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
@@ -20,13 +20,11 @@ internal class RecordingCoordinator {
     let srContextPublisher: SRContextPublisher
 
     private var currentRUMContext: RUMContext? = nil
-    @ReadWriteLock
     private var isSampled = false
 
     /// `recordingEnabled` is used to track when the user 
     /// has enabled or disabled the recording for Session Replay.
-    @ReadWriteLock
-    private(set) var recordingEnabled = false
+    private var recordingEnabled = false
 
     /// Sends telemetry through sdk core.
     private let telemetry: Telemetry
@@ -67,14 +65,18 @@ internal class RecordingCoordinator {
 
     /// Enables recording based on user request.
     func startRecording() {
-        recordingEnabled = true
-        evaluateRecordingConditions()
+        scheduler.queue.run { [weak self] in
+            self?.recordingEnabled = true
+            self?.evaluateRecordingConditions()
+        }
     }
 
     /// Disables recording based on user request.
     func stopRecording() {
-        recordingEnabled = false
-        evaluateRecordingConditions()
+        scheduler.queue.run { [weak self] in
+            self?.recordingEnabled = false
+            self?.evaluateRecordingConditions()
+        }
     }
 
     // MARK: Private
--- a/DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
+++ b/DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
@@ -9,7 +9,9 @@ import Foundation
 import DatadogInternal
 
 /// Object is responsible for getting the RUM context, randomising the sampling rate,
-/// starting/stopping the recording scheduler as needed and propagating `has_replay` to other features.
+/// managing the recording state, starting/stopping the recording scheduler as needed,
+/// and propagating `has_replay` to other features.
+
 internal class RecordingCoordinator {
     let recorder: Recording
     let scheduler: Scheduler
@@ -20,9 +22,14 @@ internal class RecordingCoordinator {
     private var currentRUMContext: RUMContext? = nil
     private var isSampled = false
 
+    /// `recordingEnabled` is used to track when the user 
+    /// has enabled or disabled the recording for Session Replay.
+    @ReadWriteLock
+    private(set) var recordingEnabled = false
+
     /// Sends telemetry through sdk core.
     private let telemetry: Telemetry
-    /// The sampling rate for internal telemetry of method call.
+    /// The sampling rate for internal telemetry of method calls.
     private let methodCallTelemetrySamplingRate: Float
 
     init(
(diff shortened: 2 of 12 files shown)


DatadogSessionReplay/Sources/SessionReplay.swift
```diff
@@ -19,20 +19,53 @@ public enum SessionReplay {
     /// - Parameters:
     ///   - configuration: Configuration of the feature.
     ///   - core: The instance of Datadog SDK to enable Session Replay in (global instance by default).
+    ///   - startRecordingImmediately: When `true`, the recording starts automatically; when `false` it doesn't and the recording will need to be started manually (`true` by default).
     public static func enable(
-        with configuration: SessionReplay.Configuration, in core: DatadogCoreProtocol = CoreRegistry.default
+        with configuration: SessionReplay.Configuration,
+        in core: DatadogCoreProtocol = CoreRegistry.default,
+        startRecordingImmediately: Bool = true
```
ncreated, line 26: question/ Is there any extra motivation behind introducing a new parameter to `SR.enable()` vs adding it to the existing `SessionReplay.Configuration`? The latter could be more idiomatic choice, no?

In all our modules (RUM | Logs | Trace | ...) we aim at keeping the `enable()` functions lean by separating `configuration:` and `core:` parameters. Here we seem to diverge from existing convention.

mariedm: Not really, I started like this but actually thought of including it in the configuration. I will make the change!

ncreated: Great! Also, the [iOS > Public API Convention in iOS SDK 2.x > Rule 2](https://datadoghq.atlassian.net/wiki/x/6hh2t) makes a good guideline on this element. Keeping `.enable(with:in:)` should be preferred if possible.

DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
```diff
@@ -46,30 +54,61 @@ internal class RecordingCoordinator {
...
+    func stopRecording() {
+        recordingEnabled = false
+        evaluateRecordingConditions()
+    }
+
+    // MARK: Private
+
+    /// Evaluates whether recording should start or stop based on user request and sampling.
+    private func evaluateRecordingConditions() {
+       if recordingEnabled && isSampled {
```
ncreated, line 83: change-request/ This adds a new use of `isSampled` which isn't synchronized through `@ReadWriteLock`. Because `evaluateRecordingConditions()` can be called on any thread (through start / stop recording), this may crash due to data race if `isSampled` is changed from `onRUMContextChanged()` at the same time.

DatadogSessionReplay/Sources/SessionReplay.swift
```diff
@@ -45,13 +78,46 @@ public enum SessionReplay {
...
     }
+
+    internal static func startRecording(core: DatadogCoreProtocol) throws {
+        guard !(core is NOPDatadogCore) else {
+            throw ProgrammerError(
+                description: "Datadog SDK must be initialized before calling `SessionReplay.startRecording()`."
+            )
+        }
+
+        guard let sr = core.feature(named: SessionReplayFeaturneName, type: SessionReplayFeature.self) else {
```
ncreated, line 98: nit/ It might be safer to get it by type: `core.get(feature: SessionReplayFeature.self)`

DatadogSessionReplay/Sources/Feature/Baggages.swift
```diff
@@ -38,3 +39,10 @@ internal struct RUMContext: Codable, Equatable {
     /// Current view related server time offset
     let viewServerTimeOffset: TimeInterval?
 }
+
+extension DatadogContext {
+    /// The value indicating if replay is being performed by Session Replay.
+    var hasReplay: Bool? {
+        try? baggages[RUMDependency.hasReplay]?.decode()
+    }
+}
```
ncreated, line 48: question/ Where do we use it? Having this defined in SR module looks odd - is there really a use case where SR needs to decode its own baggage?

ncreated (changes requested):
Looks great 💪 and I appreciate the work on tests coverage 🏅 👌. I left one blocking commend on potential problem in synchronizing the `isSampled` state (I think now it can be mutated from different threads).

Also, let's move the new option to `SessionReplay.Configuration` to be consistent with other feature APIs 🙏.

DatadogSessionReplay/Sources/Feature/Baggages.swift
mariedm: I had added it to access `hasReplay` from `WebViewRecordReceiver`, so I can ignore webviews' snapshots accordingly. Since I'm not taking care of it in this PR, I can remove it for now, but I might need to re-add it in the future. If that's the case, did you have another place in mind, that would be more appropriate?

DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
mariedm: That's a good point! I'll make the change.

DatadogSessionReplay/Sources/SessionReplayConfiguration.swift
```diff
@@ -49,10 +54,12 @@ extension SessionReplay {
         public init(
             replaySampleRate: Float,
             defaultPrivacyLevel: SessionReplayPrivacyLevel = .mask,
+            startRecordingImmediately: Bool = true,
```
ncreated, line 57: nit/ We should also list it within `/// - Parameters:` list in this API comment.

ncreated (approved):
Looks great 🎯. I left one small feedback on API comment.

DatadogSessionReplay/Sources/Feature/Baggages.swift
ncreated: Got it 👍. No strong opinion on this, but perhaps better to introduce alongside with usage. I thought it's a leftover 🙂

DatadogSessionReplay/Sources/SessionReplayConfiguration.swift
mariedm: Totally! That's a miss from my side when making the changes. Good catch.

DatadogSessionReplay/Sources/Feature/Baggages.swift
mariedm: What do you mean by "perhaps better to introduce alongside with usage"?

maxep: Yes, lets remove it if it's not used 👍

mariedm: Already done ✅

DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
```diff
@@ -18,11 +20,17 @@ internal class RecordingCoordinator {
     let srContextPublisher: SRContextPublisher
 
     private var currentRUMContext: RUMContext? = nil
+    @ReadWriteLock
     private var isSampled = false
 
+    /// `recordingEnabled` is used to track when the user 
+    /// has enabled or disabled the recording for Session Replay.
+    @ReadWriteLock
+    private(set) var recordingEnabled = false
```
maxep, line 29: /suggestion Instead of using locks here (I know I made the suggestion 😄), we could use the scheduler's queue to synchronise updated. WDYT?

DatadogSessionReplay/Sources/SessionReplay.swift
```diff
@@ -50,8 +80,41 @@ public enum SessionReplay {
...
+        }
+
+        guard let sr = core.get(feature: SessionReplayFeature.self) else {
+            throw ProgrammerError(
+                description: "Session Replay must be initialized before calling `SessionReplay.stopRecording()`."
+            )
+        }
+
+        sr.stopRecording()
+    }
```
maxep, line 118: /question do we need this level of logs? When starting we could only check if the feature is enabled, not necessarily if core is initialised. And stopping a no-op feature is not necessarily a programmer error, it could be silent.

maxep (commented):
Looks great! I have left couple of question/suggestion.

DatadogSessionReplay/Sources/SessionReplay.swift
mariedm: Good point, I'll make the change.

DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
mariedm: So, something like this? 

```
private var _isRecordingPaused = true
private var isRecordingPaused: Bool {
    get {
            scheduler.queue.sync { _isRecordingPaused }
    }
    set {
            scheduler.queue.sync { _isRecordingPaused = newValue }
    }
}
```

mariedm:
@maxep Following our discussion, I’ve updated the logic to use the scheduler’s queue. Sharing here the reasoning for later reference.

With this approach, all state changes are synchronized on the same queue. It maintains consistency without the overhead of locks.

A trade-off of this approach is that `recordingEnabled` is now private to prevent direct access from outside the class, which means we can no longer access it directly in tests. We can still rely on the scheduler’s `isRunning` variable to verify when the recording is active, though we lose the ability to directly test whether the recording was enabled or disabled by the user.

This change prioritizes thread and synchronization safety, which can avoid hard-to-debug issues. This should save us time and prevent potential problems in the future.

mariedm:
Note: I logged a [ticket](https://datadoghq.atlassian.net/browse/RUM-5845) to deal with webviews' snapshots later, as this should be an edge case not blocking GA.

maxep (approved):
Looks great, thanks!

PR #1992, opened 2024-08-09
RUM-5762 [SR] Create touchPrivacyLevel masking option

mariedm:
### What and why?

This PR is the first step towards providing more granular control over global masking in Session Replay. It introduces a new privacy level for user touches (e.g., tap, swipe), allowing customers to choose whether to show or hide these touches in replays. Currently, touches are always visible.

For full context, refer to the [RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3884548349/RFC+-+Session+Replay+Global+Masking+Configuration).

### How?

This PR establishes the new privacy level within the Session Replay Configuration but does not yet implement the logic for touch visibility in snapshots.

Future PRs will introduce additional privacy levels for images, text, and user inputs. The current Privacy Level will be deprecated once the global masking feature is fully implemented and released.

This PR:
- Creates an enum for `SessionReplayTouchPrivacyLevel`
- Adds a new initializer for `SessionReplayConfiguration` that accepts the new `touchPrivacyLevel`
  - The new API for initializing `SessionReplayConfiguration` uses a mandatory `sampleRate` parameter and an optional `touchPrivacyLevel` parameter, whereas the former initializer uses a mandatory `replaySampleRate` parameter and an optional `defaultPrivacyLevel`. This change in parameter naming was necessary to avoid ambiguity in method calls.
- Updates various tests accordingly

### Review checklist

### DIFF
--- a/DatadogSessionReplay/Sources/SessionReplayConfiguration.swift
+++ b/DatadogSessionReplay/Sources/SessionReplayConfiguration.swift
@@ -10,6 +10,7 @@ import DatadogInternal
 
 // swiftlint:disable duplicate_imports
 @_exported import enum DatadogInternal.SessionReplayPrivacyLevel
+@_exported import enum DatadogInternal.SessionReplayTouchPrivacyLevel
 // swiftlint:enable duplicate_imports
 
 extension SessionReplay {
@@ -30,6 +31,11 @@ extension SessionReplay {
         /// Default: `.mask`.
         public var defaultPrivacyLevel: SessionReplayPrivacyLevel
 
+        /// Defines the way user touches (e.g. tap) should be masked.
+        ///
+        /// Default: `.hide`.
+        public var touchPrivacyLevel: SessionReplayTouchPrivacyLevel
+
         /// Custom server url for sending replay data.
         ///
         /// Default: `nil`.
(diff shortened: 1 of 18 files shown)


DatadogSessionReplay/Sources/SessionReplayConfiguration.swift
```diff
@@ -43,6 +49,23 @@ extension SessionReplay {
 
         /// Creates Session Replay configuration
         /// - Parameters:
+        ///   - sampleRate: The sampling rate for Session Replay. It is applied in addition to the RUM session sample rate.
+        ///   - touchPrivacyLevel: The way user touches (e.g. tap) should be masked. Default: `.hide`.
+        ///   - customEndpoint: Custom server url for sending replay data. Default: `nil`.
+        public init(
+            sampleRate: Float,
```
mariedm, line 56: Note: The parameter name was changed to `sampleRate` in this initializer to distinguish it from the other initializer. Without this change, this would result in an ambiguous call, and Swift not being able to determine which initializer to use.

mariedm:
General question: is this the sort of PR to add to the CHANGELOG, given that it provides a new public API?

DatadogSessionReplay/Sources/SessionReplayConfiguration.swift
maxep: /change-request Lets not expose the initialiser until the feature is implemented.

maxep (changes requested):
Looks good 👍  Just a change request to not expose the config yet, until the feature is implemented.

maxep:
> General question: is this the sort of PR to add to the CHANGELOG, given that it provides a new public API?

Lets not exposed the public config while in progress 👍

mariedm:
> Lets not exposed the public config while in progress
@maxep right! Actually I was planning to keep this on a feature branch until the end but I forgot when submitting this PR. I just changed the base branch. 

Do you think I should still make this new API private in the meantime?

maxep (approved):

mariedm:
Following our discussion on Slack, I've updated the new API. We’re keeping the same parameter name for `replaySampleRate` and requiring users to explicitly pass the new privacy levels. This remains a non-breaking change solution, and we can always introduce some default values for the new API in a future major version.

An action item has been added to the [Changes For Next Major Version](https://docs.google.com/document/d/1YPSS5IKQcA-sLKnVXF2oNPz6bjYbtM41KopNoHE0dcs/edit?usp=sharing) (internal) to remove the former initializer.

ganeshnj (approved):

PR #1993, opened 2024-08-13
RUM-5555 Benchmarks: Collect Memory Metric

maxep:
### What and why?

### How?

### Review checklist


BenchmarkTests/Benchmarks/Sources/Exporter.swift
```diff
@@ -0,0 +1,26 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
```
mariedm, line 6: nit: a brief comment explaining the purpose of this class could be helpful in the future

BenchmarkTests/Benchmarks/Sources/Exporter.swift
```diff
@@ -0,0 +1,26 @@
...
+    init() {
+        let configuration: URLSessionConfiguration = .ephemeral
+        configuration.urlCache = nil
+        self.session = URLSession(configuration: configuration, delegate: SessionDelegate(), delegateQueue: nil)
+    }
+}
+
+extension Exporter.SessionDelegate: URLSessionTaskDelegate {
+    func urlSession(_ session: URLSession, task: URLSessionTask, didCompleteWithError error: (any Error)?) {
```
mariedm, line 23: Is it left empty on purpose for now?

BenchmarkTests/Benchmarks/Sources/MetricExporter.swift
```diff
@@ -0,0 +1,159 @@
...
+            resources: [],
+            tags: Array(tags)
+        )
+    }
+    
+    /// Submit timeseries to the Metrics intake.
+    ///
+    /// - Parameter series: The timeseries.
+    func submit(series: [Serie]) throws {
+        var data = try series.reduce(Data()) { data, serie in
```
mariedm, line 126: Should we expect large payloads here? If so, what happens if it exceeds a certain size?

BenchmarkTests/Runner/Scenarios/SyntheticScenario.swift
```diff
@@ -0,0 +1,69 @@
...
+    /// The scenario's name.
+    let name: String
+    
+    /// The underlying scenario.
+    private let _scenario: Scenario
+    
+    /// Creates the scenario by reading the `BENCHMARK_SCENARIO` value from the
+    /// environment vairables.
+    ///
+    /// - Parameter processInfo: The `ProcessInfo` with environment vairables
```
mariedm, line 22:
```suggestion
    /// - Parameter processInfo: The `ProcessInfo` with environment variables
```

BenchmarkTests/Runner/Scenarios/SyntheticScenario.swift
```diff
@@ -0,0 +1,69 @@
...
+/// variable to instantiate a `Scenario` compliant object.
+internal struct SyntheticScenario: Scenario {
+    /// The scenario's name.
+    let name: String
+    
+    /// The underlying scenario.
+    private let _scenario: Scenario
+    
+    /// Creates the scenario by reading the `BENCHMARK_SCENARIO` value from the
+    /// environment vairables.
```
mariedm, line 20:
```suggestion
    /// environment variables.
```

BenchmarkTests/Runner/Scenarios/SyntheticScenario.swift
```diff
@@ -0,0 +1,69 @@
...
+    ///
+    /// - Parameter processInfo: The `ProcessInfo` with environment vairables
+    /// configured
+    init?(processInfo: ProcessInfo = .processInfo) {
+        guard let name = processInfo.environment["BENCHMARK_SCENARIO"] else {
+            return nil
+        }
+
+        switch name {
+        case "sessionReplay":
```
mariedm, line 30: Should we use an enum for this, instead of a raw string?

BenchmarkTests/Runner/AppDelegate.swift
```diff
@@ -5,20 +5,76 @@
...
+        case .baseline, .metrics:
+            // measure metrics during baseline and metrics runs
+            Benchmarks.metrics(
+                with: Benchmarks.Configuration(
+                    info: applicationInfo,
+                    scenario: scenario,
+                    run: run
+                )
+            )
+        case .profiling:
```
mariedm, line 35: How will the `.profiling` use case differ from `.metrics` in practice?

mariedm (commented):
Nice work, looks great! 🙌

I have a general question: will this benchmark work on iOS only, or is it compatible with all supported platforms?

michaelcretzman (approved):

BenchmarkTests/Benchmarks/Sources/Exporter.swift
maxep: I'm using it mostly for debugging, but it can be removed :+1

BenchmarkTests/Runner/AppDelegate.swift
maxep: Yes, we will perform different runs from baseline, metrics, and profiling. This way we avoid skewing metrics data with profiling overhead.

BenchmarkTests/Benchmarks/Sources/Benchmarks.swift
```diff
@@ -0,0 +1,133 @@
...
+let instrumentationName = "benchmarks"
+let instrumentationVersion = "1.0.0"
+
+/// Benchmark entrypoint to configure opentelemetry with metrics meters
+/// and tracer.
+public enum Benchmarks {
+    /// Configuration of the Benchmarks library.
+    public struct Configuration {
+        /// Context of Benchmarks measures.
+        /// The context properties will be added metrics as tags.
```
ncreated, line 24:
```suggestion
        /// The context properties will be added to metrics as tags.
```

BenchmarkTests/Benchmarks/Sources/MetricExporter.swift
maxep: There is no caching so we will upload 3 metrics per payload, with only one point per metrics. Each upload will happen every 10 seconds as configured [here](https://github.com/DataDog/dd-sdk-ios/pull/1993/files#diff-b5ae3e89d1c4c79036d3ea45588a247fff697d5ccf90f172c0be2856a4542e79R92). So each payload will be very small 👍

BenchmarkTests/Benchmarks/Sources/Benchmarks.swift
```diff
@@ -0,0 +1,133 @@
...
+            "scenario": configuration.context.scenario,
+            "application_id": configuration.context.applicationIdentifier,
+            "sdk_version": configuration.context.sdkVersion,
+            "branch": configuration.context.branch,
+        ]
+
+        _ = meter.createDoubleObservableGauge(name: "ios.benchmark.memory") { metric in
+            do {
+                let mem = try Memory.footprint()
+                metric.observe(value: mem, labels: labels)
```
ncreated, line 121: question/ How does it work? We create Gauge metric ✅ but when / how frequently will this closure be called?

BenchmarkTests/Benchmarks/Sources/MetricExporter.swift
```diff
@@ -0,0 +1,162 @@
...
+import OpenTelemetrySdk
+
+enum MetricExporterError: Error {
+    case unsupportedMetric(aggregation: AggregationType, dataType: Any.Type)
+}
+
+/// Replacement of otel `DatadogExporter` for metrics.
+///
+/// This version does not store data to disk, it uploads to the intake directly.
+/// Additionally, it does not crash.
```
ncreated, line 17: Is this crash a known issue in `opentelemetry-swift`? If so let's link it here, so we can go back to the OOB exporter once fixed. If there are no reports, let's make one.

BenchmarkTests/Runner/AppDelegate.swift
```diff
@@ -5,20 +5,76 @@
...
+            apiKey: info.apiKey,
+            context: Benchmarks.Configuration.Context(
+                applicationIdentifier: bundle.bundleIdentifier!,
+                applicationName: bundle.object(forInfoDictionaryKey: "CFBundleExecutable") as! String,
+                applicationVersion: bundle.object(forInfoDictionaryKey: "CFBundleShortVersionString") as! String,
+                sdkVersion: "",
+                deviceModel: try! sysctl.model(),
+                osName: device.systemName,
+                osVersion: device.systemVersion,
+                run: run.rawValue,
```
ncreated, line 74: blocking/ The context properties are added as metric tags and the `run` tag we agreed upon in [the RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3889660018/RFC+Session+Replay+Performance+Metrics) is:
```
run:baseline|instrumented|profiling
```
So, passing the `run.rawValue` seems wrong as it will be one of:
```swift
internal enum SyntheticRun: String {
    case baseline
    case metrics
    case profiling
}
```

BenchmarkTests/Runner/Scenarios/SyntheticScenario.swift
```diff
@@ -0,0 +1,74 @@
...
+    func instrument(with info: AppInfo) {
+        _scenario.instrument(with: info)
+    }
+}
+
+/// The Synthetics benchark run value.
+internal enum SyntheticRun: String {
+    case baseline
+    case metrics
+    case profiling
```
ncreated, line 57: suggestion/ This is quite crucial piece to understand the whole Benchmark automation. Can we add comments explaining what instrumentation is activated during each run?

ncreated (changes requested):
It looks great 👌 and the abstraction makes sense. I left few suggestions and found one blocking issue.

BenchmarkTests/Benchmarks/Sources/Exporter.swift
mariedm: Up to you!

BenchmarkTests/Benchmarks/Sources/Benchmarks.swift
maxep: It is based on [Asynchronous Gauge](https://opentelemetry.io/docs/specs/otel/metrics/api/#asynchronous-gauge) from otel specs where _Callback functions will be called only when the Meter is being observed._
The meter is observed when there is a push, [in our case it will be every 10s](https://github.com/DataDog/dd-sdk-ios/pull/1993/files#diff-b5ae3e89d1c4c79036d3ea45588a247fff697d5ccf90f172c0be2856a4542e79R92)

BenchmarkTests/Benchmarks/Sources/MetricExporter.swift
maxep: I haven't reported it just yet, will do.
Basically, [these](https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/Exporters/DatadogExporter/Metrics/MetricsExporter.swift#L69-L84) force unwrap can fail.

BenchmarkTests/Benchmarks/Sources/Benchmarks.swift
ncreated: Got it, thanks!

ncreated (approved):

PR #1997, opened 2024-08-16
fix: Ensure needed attributes are sent from logs on the message bus

fuzzybinary:
### What and why?

Certain attributes (fingerprint and error source type) are removed as part of log processing because they are moved onto official properties of errors in logs.  However, when these errors are transmitted to RUM, the properties are not replaced, which they should be, so that RUM has all of the information about the error.

### Review checklist

### DIFF
--- a/DatadogLogs/Sources/RemoteLogger.swift
+++ b/DatadogLogs/Sources/RemoteLogger.swift
@@ -189,6 +189,15 @@ internal final class RemoteLogger: LoggerProtocol {
                     return
                 }
 
+                // Add back in fingerprint and error source type
+                var busCombinedAttributes = combinedAttributes
+                if let errorSourcetype = error?.sourceType {
+                    busCombinedAttributes[CrossPlatformAttributes.errorSourceType] = errorSourcetype
+                }
+                if let errorFingerprint = errorFingerprint {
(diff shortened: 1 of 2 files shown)


DatadogLogs/Sources/RemoteLogger.swift
```diff
@@ -189,6 +189,15 @@ internal final class RemoteLogger: LoggerProtocol {
...
 
+                // Add back in fingerprint and error source type
+                var busCombinedAttributes = combinedAttributes
+                if let errorSourcetype = error?.sourceType {
+                    busCombinedAttributes[CrossPlatformAttributes.errorSourceType] = errorSourcetype
+                }
+                if let errorFingerprint = errorFingerprint {
+                    busCombinedAttributes[Logs.Attributes.errorFingerprint] = errorFingerprint
+                }
+
```
maxep, line 200: /suggestion I think it would be clearer if that these properties were added to `ErrorMessage` object, would that be possible?

fuzzybinary: That's a much larger refactor is the issue. 😓

maxep (approved):
Good to merge as we have RUM-5893 to refactor 👍

PR #2005, opened 2024-08-20
Always assume device is online on Apple Watch

jfiser-paylocity:
### What and why?

Workaround for the issue https://github.com/DataDog/dd-sdk-ios/issues/2004.

On a real Apple Watch device, `NWPathMonitor` component always returns value of `.unsatisfied` that results in no logs being ever uploaded to API. As there is no straight-forward replacement for watchOS platform, this PR assumes that Apple Watch device is always online (another solution could be to implement periodic URL check in time interval).

From _https://developer.apple.com/documentation/Technotes/tn3135-low-level-networking-on-watchOS_:
Capture-2024-08-20-142242

Datadog dashboard logs using the fix:
| Simulator | Real device |
| - | - |
| Capture-2024-08-20-141120 | Capture-2024-08-20-140535 |

### How?

This PR specifically fixes logs upload by bypassing reachability check in `DataUploadConditions`. 

### Review checklist
- [ ] Feature or bugfix MUST have appropriate tests (unit, integration)
- [ ] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes

PR #2009, opened 2024-08-21
fix: Always assume device is online on Apple Watch

maxep:
### What and why?

Skip `networkReachability` check for uploading on watchOS.

ext contribution: #2005

### Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - [FIX] Propagate global Tracer tags to OpenTelemetry span attributes. See [#2000][]
 - [FEATURE] Add Logs event mapper to ObjC API. See [#2008][]
 - [IMPROVEMENT] Send retry information with network requests (eg. retry_count, last_failure_status and idempotency key). See [#1991][]
+- [FIX] Ignore network reachability on watchOS . See [#2005][] (Thanks [@jfiser-paylocity][])
 
 # 2.16.0 / 20-08-2024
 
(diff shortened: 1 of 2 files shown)


ncreated (commented):
Question/ What will happen to the "failed" connections and how will they impact the next upload delay? In case of "offline" is the failure immediate or comes with certain delay?

maxep:
@ncreated Good question, we might fall under [this](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogCore/Sources/Core/Upload/DataUploadStatus.swift#L53) and never retry the request..
We should test this, preferably on a physical watch. @mariedm would you be able to test this?

mariedm:
> @ncreated Good question, we might fall under [this](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogCore/Sources/Core/Upload/DataUploadStatus.swift#L53) and never retry the request.. We should test this, preferably on a physical watch. @mariedm would you be able to test this?

Yes, I will!

mariedm:
I tested on an Apple Watch Series 7 running watchOS 10.6.1. I can confirm that the network is reachable on the simulator but not on my device.

When bypassing the network reachability check on watchOS, [uploadStatus.needsRetry](https://github.com/DataDog/dd-sdk-ios/blob/7a57ec74a7150e761d700a1471cb45b0a9605930/DatadogCore/Sources/Core/Upload/DataUploadWorker.swift#L123) returns `true`. After restoring the internet connection, uploads resume as expected, so this approach appears safe.

I found out about the [Watch Connectivity](https://developer.apple.com/documentation/watchconnectivity) framework which can determine if the paired iPhone is reachable. This might be useful in the future in case we want to make this solution more robust.

maxep:
Thanks for testing @mariedm 🙏  It is good to merge then, WDYT @ncreated ?

ncreated (approved):

ncreated:
> Thanks for testing @mariedm 🙏 It is good to merge then, WDYT @ncreated ?

Yes 👌. Thanks for assurance @mariedm 💪

PR #2013, opened 2024-08-22
Fix lazy image SR identifier

maciejburda:
### What and why?

While working on the configuration we have discovered that lazy identifier for UIImages in Session Replay doesn't work.

Mitigation is simple. When introducing `DatadogExtension` instead of direct extension we forget to change the object that maintains associated objects.

This is hard to test. We could potentially measure the time. Second time should be way faster than the first one, but it has a risk of being flaky.

### Review checklist


mariedm (approved):
Thanks for the fix! 🙌

maxep (approved):
Thank you!

PR #2014, opened 2024-08-22
RUM-5755 Add Benchmark Profiler Interface

maxep:
### What and why?

Expose profiler interface for benchmarking.

During benchmarking, a concrete implementation of the profiler will be injected to collect data during execution of the SDK. In production, the profiler is no-op and immutable.

Current profiler interface can create a tracer instance for a given operation.

### How?

The `BenchmarkProfiler` is added to the `DatadogInternal` module with a single static variable allowing injecting a profiler instance from the Benchmark project.

### Review checklist

### DIFF
--- a/BenchmarkTests/Benchmarks/Sources/Benchmarks.swift
+++ b/BenchmarkTests/Benchmarks/Sources/Benchmarks.swift
@@ -11,6 +11,7 @@
 import Foundation
 import OpenTelemetryApi
 import OpenTelemetrySdk
+import DatadogExporter
 
 let instrumentationName = "benchmarks"
 let instrumentationVersion = "1.0.0"
@@ -151,5 +152,30 @@ public enum Benchmarks {
         }
 
         OpenTelemetry.registerMeterProvider(meterProvider: meterProvider)
+    }
+
+    /// Configure and register a OpenTelemetry Tracer.
+    ///
+    /// - Parameter configuration: The Benchmark configuration.
+    public static func tracer(with configuration: Configuration) {
+        let exporterConfiguration = ExporterConfiguration(
+            serviceName: configuration.context.applicationIdentifier,
+            resource: "Benchmark Tracer",
+            applicationName: configuration.context.applicationName,
+            applicationVersion: configuration.context.applicationVersion,
+            environment: "benchmarks",
+            apiKey: configuration.apiKey,
+            endpoint: .us1,
+            uploadCondition: { true }
+        )
+        
+        let exporter = try! DatadogExporter(config: exporterConfiguration)
+        let processor = SimpleSpanProcessor(spanExporter: exporter)
+
+        let provider = TracerProviderBuilder()
+            .add(spanProcessor: processor)
+            .build()
+
+        OpenTelemetry.registerTracerProvider(tracerProvider: provider)
     }
 }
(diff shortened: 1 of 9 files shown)


Package.swift
```diff
@@ -7,6 +7,9 @@ let opentelemetry = ProcessInfo.processInfo.environment["OTEL_SWIFT"] != nil ?
     (name: "opentelemetry-swift", url: "https://github.com/open-telemetry/opentelemetry-swift.git") :
     (name: "opentelemetry-swift-packages", url: "https://github.com/DataDog/opentelemetry-swift-packages.git")
 
+let internalSwiftSettings: [SwiftSetting] = ProcessInfo.processInfo.environment["BENCHMARK"] != nil ?
+    [.define("BENCHMARK")] : []
```
maxep, line 11: I'm open to suggestions if there is another way to specify the `BENCHMARK` env.

BenchmarkTests/Runner/BenchmarkProfiler.swift
```diff
@@ -0,0 +1,26 @@
...
+
+internal final class DummyTracer: DatadogInternal.BenchmarkTracer {
+    func startSpan(named: @autoclosure () -> String) -> any DatadogInternal.BenchmarkSpan {
+        DummySpan()
+    }
+}
+
+internal final class DummySpan: DatadogInternal.BenchmarkSpan {
+    func stop() { }
+}
```
maxep, line 26: It's temporary, concrete implementation will come in another PR

BenchmarkTests/Benchmarks/Sources/Benchmarks.swift
```diff
@@ -152,4 +153,29 @@ public enum Benchmarks {
 
         OpenTelemetry.registerMeterProvider(meterProvider: meterProvider)
     }
+
+    /// Configure and register a OpenTelemetry Tracer.
+    ///
+    /// - Parameter configuration: The Benchmark configuration.
+    public static func tracer(with configuration: Configuration) {
```
ncreated, line 160: nit/ This API name has no verb, making it impossible to understand what `Benchmarks.tracer(with:)` will do. I'd assume it returns a tracer, but in fact it registers one. Same applies to `Benchmark.metrics(with:)`.

I had to scroll to this implementation from `AppDelegate's` call in order to understand what this code does.

ncreated (dismissed):
Looks good 👍. I left few non-blocking feedbacks.

DatadogInternal/Sources/Benchmarks/BenchmarkProfiler.swift
```diff
@@ -0,0 +1,77 @@
...
+        NOPBenchmarkSpan.shared
+    }
+}
+
+private final class NOPBenchmarkSpan: BenchmarkSpan {
+    /// The no-op span shared instance.
+    static let shared: BenchmarkSpan = NOPBenchmarkSpan()
+    /// no-op
+    func stop() {}
+}
```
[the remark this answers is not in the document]
maxep: Good point 👍

BenchmarkTests/Benchmarks/Sources/Benchmarks.swift
maxep: Yes indeed, I've renamed them to `enableMetrics` and `enableTracer` 👍

ncreated (approved):
🚀

PR #2015, opened 2024-08-22
RUM-5083 refactor: Allow different sampling rates for each metric

ncreated:
### What and Why?

📦 This PR introduces the ability to sample each SDK metric using distinct sample rates.

Previously, [all metrics](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3534554673/Telemetry+Metric+Specs) shared the same [`metricTelemetrySampleRate`](https://github.com/DataDog/dd-sdk-ios/blob/691fc07fbc32267ec313cd3c3590f1155a9573ac/DatadogRUM/Sources/RUMConfiguration.swift#L273-L276) (15%). To sample batch metrics at different rates in `RUM-5083`, this refactoring is necessary.

### How?

A new property, `sampleRate`, has been added to the telemetry message that carries metric data:

```swift
public struct MetricTelemetry {
    public let name: String
    public let attributes: [String: Encodable]

    /// The sample rate for this metric, applied in addition to the telemetry sample rate.
    /// ...
    /// This sample rate is applied in the telemetry receiver, after the metric has been processed by the SDK core (tail-based sampling).
    public let sampleRate: Float
}

public enum TelemetryMessage {
    // ...
    case metric(MetricTelemetry)
}
```
The [`metricTelemetrySampleRate`](https://github.com/DataDog/dd-sdk-ios/blob/691fc07fbc32267ec313cd3c3590f1155a9573ac/DatadogRUM/Sources/RUMConfiguration.swift#L273-L276) has been removed, and now each metric is sampled individually using the metric-specific `sampleRate` (15% by default).

As before, this sample rate is applied in addition to the user-controlled [`telemetrySampleRate`](https://github.com/DataDog/dd-sdk-ios/blob/691fc07fbc32267ec313cd3c3590f1155a9573ac/DatadogRUM/Sources/RUMConfiguration.swift#L213-L218) (20% by default) in the RUM `TelemetryReceiver`. Both rates are applied after telemetry is passed over the `MessageBus` to RUM (tail-based sampling):

Screenshot 2024-08-23 at 09 46 44

#### Alternative Considered: Head-Based Telemetry Sampling

I also explored the potential benefits of switching to head-based sampling for the entire telemetry (not just metrics). By dropping telemetry samples earlier, we could reduce the cost of processing them over the `MessageBus`.

To evaluate this, I benchmarked Shopist iOS app, running one of its scenarios for 50s on an iPhone 13 Pro Max in "Release" configuration. Each time, I measured the number and timings of messages passed over the `MessageBus`:

Screenshot 2024-08-23 at 09 54 48

**Conclusion:** Head-based telemetry sampling doesn't offer significant gains. In this benchmark, telemetry accounted for the fewest type of messages sent over the message bus (~7%) and contributed only a negligible processing duration (~17ms) to the overall 50s run.

Instead, we should focus on the number of `context` and `baggage` messages exchanged over the bus which are more numerous and way heavier. The full results of this benchmark and further conclusions will be published in a separate document.

### Review checklist

### DIFF
--- a/DatadogInternal/Sources/Telemetry/Telemetry.swift
+++ b/DatadogInternal/Sources/Telemetry/Telemetry.swift
@@ -58,11 +58,18 @@ public struct ConfigurationTelemetry: Equatable {
     public let useWorkerUrl: Bool?
 }
 
+public struct MetricTelemetry {
+    /// The name of the metric.
+    public let name: String
+    /// The attributes associated with this metric.
+    public let attributes: [String: Encodable]
+}
+
 public enum TelemetryMessage {
     case debug(id: String, message: String, attributes: [String: Encodable]?)
     case error(id: String, message: String, kind: String, stack: String)
     case configuration(ConfigurationTelemetry)
-    case metric(name: String, attributes: [String: Encodable])
+    case metric(MetricTelemetry)
 }
 
 /// The `Telemetry` protocol defines methods to collect debug information
@@ -106,7 +113,7 @@ public extension Telemetry {
     ///   - isSuccessful: A flag indicating if the method call was successful.
     func stopMethodCalled(_ metric: MethodCalledTrace?, isSuccessful: Bool = true) {
         if let metric = metric {
-            send(telemetry: metric.asTelemetryMetric(isSuccessful: isSuccessful))
+            send(telemetry: .metric(metric.asTelemetryMetric(isSuccessful: isSuccessful)))
         }
     }
 }
@@ -117,12 +124,12 @@ public struct MethodCalledTrace {
     let callerClass: String
     let startTime = Date()
 
-    var exectutionTime: Int64 {
+    private var exectutionTime: Int64 {
         return -startTime.timeIntervalSinceNow.toInt64Nanoseconds
     }
 
-    func asTelemetryMetric(isSuccessful: Bool) -> TelemetryMessage {
-        return .metric(
+    func asTelemetryMetric(isSuccessful: Bool) -> MetricTelemetry {
+        return MetricTelemetry(
             name: MethodCalledMetric.name,
             attributes: [
                 MethodCalledMetric.executionTime: exectutionTime,
@@ -354,7 +361,7 @@ extension Telemetry {
     ///   - name: The name of this metric.
     ///   - attributes: Parameters associated with this metric.
     public func metric(name: String, attributes: [String: Encodable]) {
-        send(telemetry: .metric(name: name, attributes: attributes))
+        send(telemetry: .metric(MetricTelemetry(name: name, attributes: attributes)))
     }
 }
 
--- a/DatadogInternal/Sources/Telemetry/Telemetry.swift
+++ b/DatadogInternal/Sources/Telemetry/Telemetry.swift
@@ -59,10 +59,24 @@ public struct ConfigurationTelemetry: Equatable {
 }
 
 public struct MetricTelemetry {
+    /// The default sample rate for metric events, applied in addition to the telemetry sample rate (20% by default).
+    public static let defaultSampleRate: Float = 15
+
     /// The name of the metric.
     public let name: String
+
     /// The attributes associated with this metric.
     public let attributes: [String: Encodable]
+
+    /// The sample rate for this metric, applied in addition to the telemetry sample rate.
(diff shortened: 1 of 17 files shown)


DatadogInternal/Sources/Telemetry/Telemetry.swift
```diff
@@ -58,11 +58,32 @@ public struct ConfigurationTelemetry: Equatable {
     public let useWorkerUrl: Bool?
 }
 
+public struct MetricTelemetry {
+    /// The default sample rate for metric events, applied in addition to the telemetry sample rate (20% by default).
```
mariedm, line 62: The default telemetry sample rate is 20% and the default metric sample rate is 15%, correct? The comment might be a bit confusing here.

mariedm, line 62:
```suggestion
    /// The default sample rate for metric events (15%), applied in addition to the telemetry sample rate (20% by default).
```

DatadogInternal/Sources/Telemetry/Telemetry.swift
```diff
@@ -75,20 +96,28 @@ public protocol Telemetry {
...
+    /// Unlike the telemetry sample rate and tail-based sampling in `stopMethodCalled()`, this sample rate is applied at the start of the method call timing.
+    /// This head-based sampling reduces the impact of processing high-frequency metrics in the SDK core, as most samples can be dropped
+    /// before being passed to the message bus.
+    ///
+    /// - Returns: A `MethodCalledTrace` instance for stopping the method call and measuring its execution time, or `nil` if the method call is not sampled.
     func startMethodCalled(
         operationName: String,
         callerClass: String,
-        samplingRate: Float = 100.0
+        sampleRate: Float
```
mariedm, line 118: What is the reason behind removing the default value here?

DatadogInternal/Sources/Telemetry/Telemetry.swift
```diff
@@ -98,15 +127,38 @@ public extension Telemetry {
...
     ///   - metric: The `MethodCalledTrace` instance.
-    ///   - isSuccessful: A flag indicating if the method call was successful.
-    func stopMethodCalled(_ metric: MethodCalledTrace?, isSuccessful: Bool = true) {
+    ///   - isSuccessful: A flag indicating whether the method call was successful.
+    ///   - sampleRate: The sample rate for **tail-based** sampling of the metric, applied after the metric is processed by the SDK core.
+    ///     Defaults to `MetricTelemetry.defaultSampleRate` (15%).
+    func stopMethodCalled(
+        _ metric: MethodCalledTrace?,
+        isSuccessful: Bool = true,
+        sampleRate: Float = MetricTelemetry.defaultSampleRate
```
mariedm, line 143: Would it be clearer for customers to have this parameter named `tailBasedSampleRate` here, and `headBasedSampleRate` in `startMethodCalled`?

mariedm (commented):
Awesome description! 🔥 I just have a couple of questions, but it looks great.

DatadogInternal/Sources/Telemetry/Telemetry.swift
ncreated: Yes, it's exactly like you said. And this is very fair call, I updated the comment as suggested 👍.

DatadogInternal/Sources/Telemetry/Telemetry.swift
ncreated: At first, I wanted compiler to guide me on all places where this is used 🙂, but after I was done I thought it's simply safer to not provide default for this particular call. The `startMethodCall()` is a high-frequency metric, defaulting it to 100% and overlooking it by mistake may bring large perf regression.

DatadogInternal/Sources/Telemetry/Telemetry.swift
ncreated: Oh, this was great idea 👍 - I also found `sampleRate` confusing in both `start()` and `stop()` but changing to `headSampleRate` and `tailSampleRate` makes total sense here. Done ✅.

BTW, this is `DatadogInternal` API, so not meant for customer use.

ganeshnj (approved):

PR #2019, opened 2024-08-26
RUM-4571 feat: add support for usage telemetry

ganeshnj:
### What and why?

We don't support usage telemetry message (similar to configuration telemetry message)

### How?

The codegen of telemetry event is tricky because the way they are defined which are JS friendly (loosely typed).

I considered various options

#### Option 1 (Preferred)

- Define an enum for each feature

```swift
enum TelemetryCommonFeaturesUsage: Codable, Equatable {
    case setTrackingConsent(value: SetTrackingConsent)
    case stopSession(value: StopSession)
    case startView(value: StartView)
}

struct SetTrackingConsent: Codable, Equatable {
    let feature: String = "set-tracking-consent"
    let trackingConsent: TrackingConsent
}

enum TrackingConsent: String, Codable {
    case granted = "granted"
    case notGranted = "not-granted"
    case pending = "pending"
}

struct StopSession: Codable, Equatable {
    let feature: String = "stop-session"
}

struct StartView: Codable, Equatable {
    let feature: String = "start-view"
}
```

Pros:
+ verbose
+ strongly typed
+ forward compatibility

Cons:
- more code
- decoding can be an issue for cases where only `feature` is there, but we don't need decoding here.

#### Option 2

- Club all common feature usage in a single enum.

```swift
enum TelemetryCommonFeaturesUsage: Codable, Equatable {
    case featureTrackingConsent(feature: Feature, trackingConsent: TrackingConsent)
    case feature(feature: Feature)
}

enum Feature: String, Codable {
    case setTrackingConsent = "set-tracking-consent"
    case stopSession = "stop-session"
    case startView = "start-view"
}

enum TrackingConsent: String, Codable {
    case granted = "granted"
    case notGranted = "not-granted"
    case pending = "pending"
}
```

Pros:
+ less code
+ strongly typed

Cons:
- backward compatibility if more associated values are added to the enum
- codegen doesn't support today
- can get overwhelming if there is a feature with many associated values

#### Option 3

- Club all common feature usage in a single enum but keep one of the associated value as a struct.

```swift
enum TelemetryCommonFeatureUsage: Codable, Equatable {
    case featureTrackingConsent(value: FeatureTrackingConsent)
    case feature(value: Feature)
}

struct FeatureTrackingConsent: Codable, Equatable {
    let feature: Feature
    let trackingConsent: TrackingConsent
}

enum Feature: String, Codable {
    case setTrackingConsent = "set-tracking-consent"
    case stopSession = "stop-session"
    case startView = "start-view"
}

enum TrackingConsent: String, Codable {
    case granted = "granted"
    case notGranted = "not-granted"
    case pending = "pending"
}
```

Pros:
+ less code

Cons:
- backward compatibility if more associated values are added to the enum
- codegen doesn't support today

In the end, went with Option 1 because it is more verbose and strongly typed. Biggest plus is this is something we already support to some extent in the codegen.

We are not generating the obj-c types but generating them will be easier with Option 1.

### Review checklist


mariedm (dismissed):
Thank you for the detailed description! I agree that option 1 makes sense.

If I understand correctly, these usage telemetry messages are intended for internal use only, is that right? And is that the reason why we don't need the ObjC interoperability?

ganeshnj:
> Thank you for the detailed description! I agree that option 1 makes sense.
> 
> If I understand correctly, these usage telemetry messages are intended for internal use only, is that right? And is that the reason why we don't need the ObjC interoperability?

Yes, the telemetry APIs are internal only and we don't use any objc code to build the SDK (hopefully in future too) hence there is no immediate need for objc APIs. If we ever do, have to go with base + child classes approach (depending on number of cases) to build the equivalent.

tools/rum-models-generator/Sources/CodeGeneration/Print/ObjcInteropPrinter.swift
```diff
@@ -36,15 +36,27 @@ import Foundation
...
 
     // MARK: - CodePrinter
 
     public func print(code: GeneratedCode) throws -> String {
+        let filteredTypes = code.swiftTypes.filter {
+            guard let typeName = $0.typeName else {
+                return true
+            }
+            return !typesToSkip.contains(typeName)
+        }
```
ncreated, line 57: suggestion/ Here we're violating OCP by performing domain-specific (usage telemetry) filtering in `ObjcInteropPrinter` that is meant for general purpose (_print objc interop for `GeneratedCode`_).

This can be avoided if this filtering is done at the highest level or `rum-models-generator` CLI:
https://github.com/DataDog/dd-sdk-ios/blob/e5a7b3e0e9dc165a5fd85985dd4e623ab042616e/tools/rum-models-generator/Sources/rum-models-generator/GenerateRUMModels.swift#L70-L73

We could go even further and move this work to `RUMCodeDecorator` and only for objc-interop generation. This is where this filtering should really belong after all.

tools/rum-models-generator/Sources/CodeGeneration/Generate/JSONSchema.swift
```diff
@@ -210,11 +223,28 @@ internal class JSONSchema: Decodable {
...
+                guard let oneOf = obj.oneOf else {
+                    return obj
+                }
+
+                // promote `feature` property to `title` if `title` is missing
+                for one in oneOf {
+                    if one.title == nil {
+                        one.title = one.properties?["feature"]?.const?.value.description
+                    }
+                }
```
ncreated, line 238: question/ What do we use the `title` for? I'm concerned about this change as effectively it produces different JSON schema than was loaded from the file. I wish we could avoid adding this hack and solve the problem in a different way. Are there cleaner options?

DatadogRUM/Sources/Integrations/TelemetryReceiver.swift
```diff
@@ -166,6 +168,35 @@ internal final class TelemetryReceiver: FeatureMessageReceiver {
         }
     }
 
+    private func send(usage: DatadogInternal.UsageTelemetry) {
+        let date = dateProvider.now
+
+        self.record(event: nil) { context, writer in
```
ncreated, line 174: question\blocking/ We're missing any mechanism for volume control, so effectively we will be collecting 20% of all samples (see: [`telemetrySampleRate`](https://github.com/DataDog/dd-sdk-ios/blob/e5a7b3e0e9dc165a5fd85985dd4e623ab042616e/DatadogRUM/Sources/RUMConfiguration.swift#L213-L218)). According to the [original RFC](https://docs.google.com/document/d/1NBe-Qms6mlf0S2009aXmKzUuPR0y5DUZO8K8xMKCR1o/edit?usp=sharing) this is considered too much. We may want to add extra sample rate similar to [`configurationTelemetrySampleRate`](https://github.com/DataDog/dd-sdk-ios/blob/e5a7b3e0e9dc165a5fd85985dd4e623ab042616e/DatadogRUM/Sources/RUMConfiguration.swift#L267-L268).

ganeshnj: Not touching the sampling priority part of this RP, this will be in a followup when we integrate it.

tools/rum-models-generator/Sources/CodeGeneration/Generate/JSONSchema.swift
ganeshnj: this is a problem with model itself, because title was never required. 

I have backfilled the titles here https://github.com/DataDog/rum-events-format/pull/221 but we can never be sure. During the codegen, it is common practice to transform the given specs to something that makes more sense, but under the limits and this change is not altering the schema but rather backfilling when certain property is not available such as title.

I would prefer to not have this and make title required but that will be breaking.

tools/rum-models-generator/Sources/CodeGeneration/Print/ObjcInteropPrinter.swift
ganeshnj: good call, I just made changes to make it configurable via CLI directly

tools/rum-models-generator/Sources/CodeGeneration/Generate/JSONSchema.swift
ncreated: I'm more concerned by inferring the title from a fixed field of the schema (`"feature"`) - and doing this only for certain schemas (`"TelemetryCommonFeaturesUsage" ||  "TelemetryMobileFeaturesUsage"`). I basically wonder if there could be a general-purpose fallback that can benefit any subschemas. Otherwise, we will break the abstraction by leaking particular schema details to generic schema parser.

Alternatively, we could make it into configuration of `JSONSchemaReader` and get rid of hardcoded strings here - but effectively this will only blurry the solution and likely make it harder to read.

ncreated (commented):
I did a test with [`rum-mock-server`](https://github.com/DataDog/rum-mock-server) and sent some usage telemetry to see if it will pass JSON schema validation. Unfortunately it didn't, but I'm not sure if the problem is coming from this PR or it is due to a known [RUM-2941](https://datadoghq.atlassian.net/browse/RUM-2941) bug. If [RUM-2941](https://datadoghq.atlassian.net/browse/RUM-2941) is the actual problem here, we may hold with sending usage telemetry until it is fixed.

Here's an example usage telemetry event sent to the mock server.- it is missing required [`telemetry.type: usage`](https://github.com/DataDog/rum-events-format/blob/ec07c062cbbb2f19b49d08f72bc95703b502906d/schemas/telemetry/usage-schema.json#L19-L23) attribute:
```json
{
    "session": {
        "id": "b517...c64a"
    },
    "view": {
        "id": "4306...1ef1"
    },
    "source": "ios",
    "version": "2.16.0",
    "application": {
        "id": "9d1b...f645"
    },
    "telemetry": {
        "device": {
            "brand": "Apple",
            "architecture": "arm64e",
            "model": "iPhone16,1 Simulator"
        },
        "os": {
            "build": "23G80",
            "version": "17.5",
            "name": "iOS"
        },
        "usage": {
            "overwritten": false,
            "no_active_view": false,
            "feature": "addViewLoadingTime",
            "no_view": false
        }
    },
    "_dd": {
        "format_version": 2
    },
    "type": "telemetry",
    "service": "dd-sdk-ios",
    "date": 1724916301426
}
```

tools/rum-models-generator/Sources/CodeGeneration/Generate/JSONSchema.swift
ganeshnj: I would refrain to apply this across the board, atleast we know where the problem is and this fix actually solves it. If we encounter similar problem, then we can think to generalize it like using some block for specific type names.

additionally, if we get the schema updates merged in, we don't need this workaround anyway as long as people will follow the existing pattern and never skip title. My problem is, it is not enforced and good intentions not always work.

ganeshnj:
> I did a test with [`rum-mock-server`](https://github.com/DataDog/rum-mock-server) and sent some usage telemetry to see if it will pass JSON schema validation. Unfortunately it didn't, but I'm not sure if the problem is coming from this PR or it is due to a known [RUM-2941](https://datadoghq.atlassian.net/browse/RUM-2941) bug. If [RUM-2941](https://datadoghq.atlassian.net/browse/RUM-2941) is the actual problem here, we may hold with sending usage telemetry until it is fixed.
> 
> Here's an example usage telemetry event sent to the mock server.- it is missing required [`telemetry.type: usage`](https://github.com/DataDog/rum-events-format/blob/ec07c062cbbb2f19b49d08f72bc95703b502906d/schemas/telemetry/usage-schema.json#L19-L23) attribute:
> 
> ```json
> {
>     "session": {
>         "id": "b517...c64a"
>     },
>     "view": {
>         "id": "4306...1ef1"
>     },
>     "source": "ios",
>     "version": "2.16.0",
>     "application": {
>         "id": "9d1b...f645"
>     },
>     "telemetry": {
>         "device": {
>             "brand": "Apple",
>             "architecture": "arm64e",
>             "model": "iPhone16,1 Simulator"
>         },
>         "os": {
>             "build": "23G80",
>             "version": "17.5",
>             "name": "iOS"
>         },
>         "usage": {
>             "overwritten": false,
>             "no_active_view": false,
>             "feature": "addViewLoadingTime",
>             "no_view": false
>         }
>     },
>     "_dd": {
>         "format_version": 2
>     },
>     "type": "telemetry",
>     "service": "dd-sdk-ios",
>     "date": 1724916301426
> }
> ```

I can see it being generated here https://github.com/DataDog/dd-sdk-ios/pull/2019/files#diff-14b43220986548c41cfa60ab0822b1dd9452198fd41a4bab2bfc14108f720c9bR4128

also, I tried ingesting the tracking consent and it was correctly received. I don't see any reason from codegen point of view why your test failed.

tools/rum-models-generator/Sources/CodeGeneration/Generate/JSONSchema.swift
ncreated: > additionally, if we get the schema updates merged in, we don't need this workaround anyway as long as people will follow the existing pattern and never skip title. My problem is, it is not enforced and good intentions not always work.

I see 👍. So after https://github.com/DataDog/rum-events-format/pull/221 is merged maybe we need better error handling rather than this hack? If `title` is missing in any schema, this can result with clear error describing the problem and a fix that should happen in in `rum-events-format`.

ncreated:
> Actually it is the custom encodable conformance which is not taking account of const properties.

Okay, so this also the root cause diagnosed in [RUM-2941](https://datadoghq.atlassian.net/browse/RUM-2941), meaning it is the same problem. Should not block this PR then.

ncreated (approved):

mariedm (approved):

PR #2022, opened 2024-08-26
RUM-5841 feat: setup addViewLoadingTime usage telemetry

ganeshnj:
### What and why?

We want to use usage telemetry to trace the `addViewLoadingTime` usage.

### How?

There are different cases we want to track per https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3951001790/RFC+View+loading+time+manual+API

- when API is used
- when there is no view to report
- when there is a view but not active
- when there is view, active and already has the timing

### Review checklist


maciejburda:
Seems like it's missing some changes.

I'm only seeing RUM model regenerated

ganeshnj:
Closing in favor of https://github.com/DataDog/dd-sdk-ios/pull/2022

PR #2031, opened 2024-08-30
feat: Improve metrics collected during benchmarks

maxep:
### What and why?

Memory and CPU metrics were only reporting the last values collected after 10 s interval. It cannot provide accurate data, instead we are now collecting metrics more frequently and report the aggregated values during push.

### How?

Collect CPU usage and Memory footprint using timers configured to read values every 100 ms, then report the aggregation depending on the metric type:
- The max memory footprint
- The average CPU usage
- The min frame rate.

### Review checklist

### DIFF
--- a/BenchmarkTests/Benchmarks/Sources/Metrics.swift
+++ b/BenchmarkTests/Benchmarks/Sources/Metrics.swift
@@ -12,13 +12,125 @@ import QuartzCore
 let TASK_VM_INFO_COUNT = mach_msg_type_number_t(MemoryLayout<task_vm_info_data_t>.size / MemoryLayout<integer_t>.size)
 let TASK_VM_INFO_REV1_COUNT = mach_msg_type_number_t(MemoryLayout.offset(of: \task_vm_info_data_t.min_address)! / MemoryLayout<integer_t>.size)
 
-enum MachError: Error {
+internal enum MachError: Error {
     case task_info(return: kern_return_t)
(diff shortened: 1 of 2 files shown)


BenchmarkTests/Benchmarks/Sources/Metrics.swift
```diff
@@ -12,21 +12,133 @@ import QuartzCore
...
+                sum: sum,
+                count: count,
+                avg: Double(sum) / Double(count)
+            )
+        } ?? Aggregation(min: value, max: value, sum: value, count: 1, avg: Double(value))
+        pthread_mutex_unlock(&mutex)
+    }
+}
+
+extension MetricAggregator where T: BinaryFloatingPoint {
```
mariedm, line 73: I was wondering if we could merge these two extensions since the logic inside is the same, but it doesn't seem straightforward.

mariedm (approved):
Looks great! 🙌

General question: do you know the impact of the more frequent metric collection on the app's performance?

BenchmarkTests/Benchmarks/Sources/Metrics.swift
maxep: I tried, but failed :/ Problem is that there is 2 distinct `Double.init` that take either a `BinaryInteger` or a `BinaryFloatingPoint`. There is no common method that can help computing the `avg` value.

PR #2034, opened 2024-09-03
RUM-5766 Pass correct privacy level to webviews

mariedm:
### What and why?

As part of the Fine-Grained Masking (FGM) initiative (see [RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3884548349/RFC+-+Session+Replay+Global+Masking+Configuration)), we have introduced new privacy levels for configuring Session Replay. We have an integration with the browser SDK, however, they don't support these fine-grained privacy levels. As a result, we need to ensure we pass the appropriate level to webviews.

### How?

- Although the former privacy level has been deprecated, we still rely on it in [WebViewTracking](https://github.com/DataDog/dd-sdk-ios/blob/8d26db63bb8856ab3b6b143ab00a3a1397d22fb1/DatadogWebViewTracking/Sources/WebViewTracking.swift#L114) to pass the correct privacy level.
- Using the conversion matrix below, the new `determineWebViewPrivacyLevel` method converts the new privacy levels (text and input, image, touch) to the corresponding global privacy level.

Screenshot 2024-09-03 at 18 15 19

### Review checklist


DatadogInternal/Sources/Models/SessionReplay/SessionReplayConfiguration.swift
```diff
@@ -17,7 +17,7 @@ public enum SessionReplayPrivacyLevel: String {
     case mask
 
     /// Mask input elements, but record all other content.
-    case maskUserInput = "mask_user_input"
+    case maskUserInput = "mask-user-input"
```
mariedm, line 20: It seems that we were passing the wrong privacy level to webviews. See [doc](https://docs.datadoghq.com/real_user_monitoring/session_replay/privacy_options) for ref.

DatadogSessionReplay/Tests/SessionReplayTests.swift
```diff
@@ -63,6 +63,7 @@ class SessionReplayTests: XCTestCase {
         // Then
         let sr = try XCTUnwrap(core.get(feature: SessionReplayFeature.self))
         XCTAssertEqual(sr.recordingCoordinator.sampler.samplingRate, 42)
+        XCTAssertFalse(config.isConfiguredWithNewApi)
```
mariedm, line 66: `isConfiguredWithNewApi` was added in #2030 to distinguish what API customers have used and only send the new or deprecated privacy levels in telemetry.

DatadogInternal/Sources/Models/SessionReplay/SessionReplayConfiguration.swift
maxep: Nice catch!!

DatadogInternal/Sources/Models/SessionReplay/SessionReplayConfiguration.swift
```diff
@@ -32,9 +32,10 @@ public enum SessionReplayPrivacyLevel: String {
...
 public protocol SessionReplayConfiguration {
-    // TODO: RUM-5766 - Pass correct privacy level to webviews
-    /// Deprecated
-    /// Former global privacy level to use in Session Replay.
+    /// Global privacy level to use in Session Replay.
+    /// Although this setting is deprecated on mobile,
+    /// it is still needed to configure the browser SDK,
+    /// which currently does not support fine-grained privacy options.
     var privacyLevel: SessionReplayPrivacyLevel { get }
 }
```
maxep, line 40: /suggestion We have now privacy levels defined in 2 places, the configuration flow is not super clear and it won't be easy to change if Browser SDK ever support fine-grained privacy. IMO, we should follow the pattern in place:
1. Define all privacy levels in `DatadogInternal/Sources/Models/SessionReplay/SessionReplayConfiguration.swift`
2. Move `determineWebViewPrivacyLevel` to `WebViewTracking` domain and make the conversion there

This way, if we have to inject fine-grained privacy options to the JS bridge, we would only have to change `WebViewTracking` module.

mariedm: It makes sense. But I need `determineWebViewPrivacyLevel` in `SessionReplay.Configuration` to assign a value to `defaultPrivacyLevel` when using the new API. Effectively, it's not being used except for the webview integration, so I could make `defaultPrivacyLevel` `optional` so that it doesn't require a value. What do you think?

mariedm: Or add a default value to `defaultPrivacyLevel`, which is kind of what we already do in the API.
`public var defaultPrivacyLevel: SessionReplayPrivacyLevel = .mask`

DatadogCore/Tests/DatadogObjc/DDSessionReplayTests.swift
```diff
@@ -31,7 +31,7 @@ class DDSessionReplayTests: XCTestCase {
 
     func testConfigurationWithNewApi() {
         // Given
-        let textAndInputPrivacy: DDSessionReplayConfigurationTextAndInputPrivacyLevel = [.maskAll, .maskAllInputs, .maskSensitiveInputs].randomElement()!
+        let textAndInputPrivacy: DDTextAndInputPrivacyLevel = [.maskAll, .maskAllInputs, .maskSensitiveInputs].randomElement()!
```
mariedm, line 34: Renamed var to align with other new privacy levels.

DatadogSessionReplay/Tests/SessionReplayTests.swift
maxep: /question As discussed with the team, we will report only new levels in telemetry so we can remove the `isConfiguredWithNewApi`. Do you want to do it as part of this PR?

mariedm: I'm planning to take care of this in a separate PR.

mariedm: I might keep it though, if we can indicate which API was used in telemetry.

DatadogSessionReplay/Tests/SessionReplayTests.swift
```diff
@@ -96,117 +96,31 @@ class SessionReplayTests: XCTestCase {
         XCTAssertNil((r.requestBuilder as? ResourceRequestBuilder)?.customUploadURL)
     }
 
-    func testWhenEnabled_itSendsConfigurationTelemetry() throws {
```
mariedm, line 99: Just moved telemetry tests lower, separate from configuration tests

DatadogSessionReplay/Tests/SessionReplayTests.swift
mariedm: I removed it completely, and also removed the deprecated privacy level from the `SessionReplayConfiguration` protocol, as we don't rely on it anymore.

DatadogWebViewTracking/Sources/WebViewTracking.swift
```diff
@@ -111,8 +111,7 @@ public enum WebViewTracking {
             type: SessionReplayConfiguration.self
         )
 
-        // TODO: RUM-5766 - Pass correct privacy level to webviews
-        let privacyLevel = sessionReplay?.privacyLevel ?? .mask
+        let privacyLevel = sessionReplay != nil ? Self.determineWebViewPrivacyLevel(textPrivacy: sessionReplay!.textAndInputPrivacyLevel, imagePrivacy: sessionReplay!.imagePrivacyLevel, touchPrivacy: sessionReplay!.touchPrivacyLevel) : .mask   // swiftlint:disable:this force_unwrapping
```
maxep, line 114: /nit
```suggestion
        let privacyLevel = sessionReplay.map {
            Self.determineWebViewPrivacyLevel(
                textPrivacy: $0.textAndInputPrivacyLevel,
                imagePrivacy: $0.imagePrivacyLevel,
                touchPrivacy: $0.touchPrivacyLevel
            )
        } ?? .mask
```

maxep (approved):
Looks great, thanks for the effort on this!

maxep (approved):

PR #2035, opened 2024-09-03
RUM-1053 Make `DatadogLogs` public API concurrency-safe

ncreated:
### What and why?

📦 This PR prepares DatadogLogs for compatibility with apps that [enable "complete" concurrency checking](https://www.swift.org/migration/documentation/swift-6-concurrency-migration-guide/completechecking/) in Swift 6.

### How?

I identified one scenario where the `DatadogLogs` API could raise a data-race safety issue in Swift 6. Specifically, this occurs when a `LoggerProtocol` instance is passed across isolation boundaries, such as when it’s used from a different thread:

```swift
import DatadogLogs


let logger = Logger.create(with: .init())

DispatchQueue.global(qos: .userInitiated).async {
    logger.debug("foo") 
    // ^ ❌ Error: Capture of 'logger' with non-sendable type 'any LoggerProtocol' in a `@Sendable` closure
}
```

This issue is resolved by [declaring Sendable conformance](https://www.swift.org/migration/documentation/swift-6-concurrency-migration-guide/dataracesafety#Sendable-Types) for LoggerProtocol:

> A conformance to `Sendable` means the given type is thread safe, and values of the type can be shared across arbitrary isolation domains without introducing a risk of data races.

Since our logger implementation has been thread-safe from the beginning, this conformance ensures that it meets the upcoming requirements of the Swift 6 language model.

### Review checklist

### DIFF
--- a/DatadogLogs/Sources/LoggerProtocol.swift
+++ b/DatadogLogs/Sources/LoggerProtocol.swift
@@ -39,7 +39,7 @@ extension CoreLoggerLevel {
 ///
 ///     // logger reference
 ///     var logger = Logger.create()
-public protocol LoggerProtocol {
+public protocol LoggerProtocol: Sendable {
     /// General purpose logging method.
     /// Sends a log with certain `level`, `message`, `error` and `attributes`.
     ///


DatadogInternal/Sources/Utils/DateFormatting.swift
```diff
@@ -6,31 +6,20 @@
...
-        formatter.formatOptions.insert(.withFractionalSeconds)
-        return formatter
-    } else {
-        let iso8601Formatter = DateFormatter()
-        iso8601Formatter.locale = Locale(identifier: "en_US_POSIX")
-        iso8601Formatter.timeZone = TimeZone(abbreviation: "UTC")! // swiftlint:disable:this force_unwrapping
-        iso8601Formatter.calendar = Calendar(identifier: .gregorian)
-        iso8601Formatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'" // ISO8601 format
-        return iso8601Formatter
-    }
```
ncreated, line 33: 💡 not needed anymore after [we dropped iOS 11](https://github.com/DataDog/dd-sdk-ios/pull/1891)

DatadogLogs/Sources/Log/SynchronizedAttributes.swift
```diff
@@ -0,0 +1,47 @@
...
+import Foundation
+import DatadogInternal
+
+/// A thread-safe container for managing attributes in a key-value format.
+/// This class allows concurrent access and modification of attributes, ensuring data consistency
+/// through the use of a `ReadWriteLock`. It is designed to be used in scenarios where attributes
+/// need to be safely managed across multiple threads or tasks.
+internal final class SynchronizedAttributes: Sendable {
+    /// The underlying dictionary of attributes, wrapped in a `ReadWriteLock` to ensure thread safety.
+    private let attributes: ReadWriteLock<[String: Encodable]>
```
maxep, line 16: /question We are not using the `ReadWriteLock` as a property wrapper in multiple places, is it because it would require it to be a `var`? Can't this be solved by making `ReadWriteLock` Sendable?

ncreated: The `ReadWriteLock` is [`Sendable` already](https://github.com/DataDog/dd-sdk-ios/blob/55ded08837526cabafe55d1bb3af77caacd0e7ae/DatadogInternal/Sources/Concurrency/ReadWriteLock.swift#L16) - it was done in [this PR](https://github.com/DataDog/dd-sdk-ios/pull/1778). And yes, if using property wrapper, this will become `var`, raising concurrency issue. I agree with what was [discussed here](https://github.com/DataDog/dd-sdk-ios/pull/1778#discussion_r1568741596) - property wrappers become quite unusable with strict concurrency.

maxep: Yes, makes sense 👍

maxep (approved):
LGTM 👍

PR #2037, opened 2024-09-04
gh-2032 chore: update TraceID docs to reflect the generation implementation

### LINKED ISSUE #2032: Incorrect doc comment
### Question

In `DatadogInternal/Sources/NetworkInstrumentation/TraceID.swift`, within `struct DefaultTraceIDGenerator`, the following constant declaration seems to have an incorrect doc comment:

```swift
    /// Describes the lower and upper boundary of tracing ID generation.
    ///
    /// * Lower: starts with `1` as `0` is reserved for historical reason: 0 == "unset", ref: dd-trace-java:DDId.java.
    /// * Upper: equals to `2 ^ 63 - 1` as some tracers can't handle the `2 ^ 64 -1` range, ref: dd-trace-java:DDId.java.
    public static let defaultGenerationRange = (1...UInt64.max)
```

Issues:
1. `UInt64.max`, according to both the documentation and empirical testing, evaluates to 2 ^ 64 - 1 (18446744073709551615), not 2 ^ 63 - 1 (9223372036854775807). 
1. Also, the [dd-trace-java](https://github.com/DataDog/dd-trace-java) package does not include any file named `DDId.java`.

I also noticed that in `DatadogInternal/Sources/NetworkInstrumentation/SpanID.swift`, within `struct DefaultSpanIDGenerator` there is a similar constant declaration that uses `UInt64.max >> 1` (instead of just `UInt64.max`) which _does_ evaluate to 2 ^ 63 - 1. Should that be the case in `DefaultTraceIDGenerator` as well?

---
Also related, both `DefaultTraceIDGenerator` and `DefaultSpanIDGenerator` use a `repeat {} while {}` loop to skip generated values of `0`, but those are already eliminated since the range starts at 1, not 0. Isn't that unnecessary code then?

ganeshnj:
### What and why?

Fixes #2032 

### How?

Updated docs to match the trace id generation

### Review checklist

### DIFF
--- a/DatadogInternal/Sources/NetworkInstrumentation/TraceID.swift
+++ b/DatadogInternal/Sources/NetworkInstrumentation/TraceID.swift
@@ -195,12 +195,11 @@ public protocol TraceIDGenerator {
     func generate() -> TraceID
 }
 
-/// A Default `TraceID` genarator.
+/// A Default `TraceID` generator.
+/// TraceId are 128 bit and follows a specific format:
+/// <32-bit unix seconds> <32 bits of zero> <64 random bits>
 public struct DefaultTraceIDGenerator: TraceIDGenerator {
-    /// Describes the lower and upper boundary of tracing ID generation.
-    ///
-    /// * Lower: starts with `1` as `0` is reserved for historical reason: 0 == "unset", ref: dd-trace-java:DDId.java.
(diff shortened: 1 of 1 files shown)


liambutler-lawrence (approved):
Thanks!

maxep (approved):

PR #2039, opened 2024-09-05
Add background upload configuration info to batch deleted telemetry

maciejburda:
### What and why?

Enriches `Batch Deleted` telemetry metric with `backgroundTaskEnabled` configuration info.
This is needed for measuring batch delivery success rate based on existence of the configuration.

### How?

I have added new field `background_tasks_enabled`.

`in_background` is still hardcoded to false, because knowing if the deletion happen in the background is not trivial.
Although I think we're mostly interested in measuring the difference is success rate for two different configurations, rather than knowing if batch file was deleted in background/foreground.

### Review checklist

### DIFF
--- a/DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
+++ b/DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
@@ -48,6 +48,8 @@ internal class FilesOrchestrator: FilesOrchestratorType {
         let consentLabel: String
         /// The preset for uploader performance in this feature to include in metric.
         let uploaderPerformance: UploadPerformancePreset
+        /// The present configuration of background upload in this feature to include in metric.
+        let backgroundTasksEnabled: Bool
     }
 
     /// An extra information to include in metrics or `nil` if metrics should not be reported for this orchestrator.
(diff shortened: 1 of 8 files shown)


DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
```diff
@@ -249,7 +251,7 @@ internal class FilesOrchestrator: FilesOrchestratorType {
                 BatchDeletedMetric.uploaderWindowKey: performance.uploaderWindow.toMilliseconds,
                 BatchDeletedMetric.batchAgeKey: batchAge.toMilliseconds,
                 BatchDeletedMetric.batchRemovalReasonKey: deletionReason.toString(),
-                BatchDeletedMetric.inBackgroundKey: false
+                BatchDeletedMetric.inBackgroundKey: metricsData.backgroundTasksEnabled
```
mariedm, line 254: I understand you want to use an existing, unused key, but isn't it a bit confusing to mix them? Can't we create a `BatchDeletedMetric.backgroundTasksEnabledKey`?

ncreated: Agree with @mariedm, let's create a new key and add it to the [spec](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3287975594/Batch+Deleted+Telemetry+-+Spec). The `in_background` has different purpose:
> `in_background` - if the batch was deleted in background; true or false

Having `backgroundTasksEnabled` enabled doesn't imply that all batches are deleted in background, so we set this value wrong. It was originally hardcoded to `false`, because iOS SDK was not supporting background uploads.

Supporting the actual value in `in_background` is separate task that will require complex implementation.

ncreated (changes requested):
I like the idea of collecting this attribute 👍. Although we should introduce a new field rather than re-using existing one for different purpose. Otherwise we will invalidate the [spec](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3287975594/Batch+Deleted+Telemetry+-+Spec) through this PR.

PR #1378, opened 2023-07-18
RUMM-3458 Enable adding custom attributes to RUM telemetry (debug) logs

ncreated:
### What and why?

📦 This PR enables sending custom attributes with debug messages through `Telemetry` interface:

```diff
- public func debug(id: String, message: String)
+ public func debug(id: String, message: String, attributes: [String: Encodable]? = nil) {
```

It makes a preliminary step to enable metrics in this interface. Although we will introduce separate method for metrics, the foundation of telemetry attributes is done here to be leveraged in `debug()` logs as well.

### How?

* Generated RUM data models after https://github.com/DataDog/rum-events-format/pull/154
   * → this required small enhancement in `rum-models-generator`
* Introduced `attributes: [String: Encodable]` property to `.debug` telemetry message and forwarded it from public `Telemetry` API down to telemetry receiver in RUM.

#### Fix in `rum-models-generator`

Because https://github.com/DataDog/rum-events-format/pull/154 introduced `"additionalProperties": true` to telemetry schema, it surfaced a gap in our code generation where `rum-models-generator` was printing invalid Swift code for custom `encode(to:)` and `decode(from:)` conformance when a type includes constant property (immutable with default value).

### Review checklist
- [ ] Feature or bugfix MUST have appropriate tests (unit, integration)
- [ ] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes

### Custom CI job configuration (optional)
- [ ] Run unit tests
- [ ] Run integration tests
- [ ] Run smoke tests

PR #2042, opened 2024-09-09
RUM-2491 fix: Fix missing `status` and `type` fields in RUM telemetry events

ncreated:
### What and why?

📦 This fix addresses the issue of missing status and type fields in RUM telemetry events. The absence of these fields reduced SDK observability, causing logs classified as "error" to be incorrectly marked as "debug."

### How?

The root cause was in the `rum-models-generator`, which started skipping encoding of immutable properties with default values after https://github.com/DataDog/dd-sdk-ios/pull/1378/commits/56ec883dca09d29036f2f1015789e81ec637505d. As a result, fields such as the following were excluded from telemetry event encoding:
```swift
/// Level/severity of the log
public let status: String = "debug"

/// Telemetry type
public let type: String? = "log"
```
The issue is resolved by ensuring these fields are included during encoding, but not during decoding (which was the original problem addressed in https://github.com/DataDog/dd-sdk-ios/pull/1378/commits/56ec883dca09d29036f2f1015789e81ec637505d. With this fix, all telemetry events now validate against the schema, confirmed through tests using the RUM Mock Server.

Additionally, a new UI was added to the Example project to help test various telemetry events:

<p align="center">
Screenshot 2024-09-09 at 18 33 12
</p>

### Review checklist


Datadog/Example/Debugging/DebugRUMViewController.swift
```diff
@@ -181,6 +180,51 @@ class DebugRUMViewController: UIViewController {
...
         sendErrorEventButton.disableFor(seconds: 0.5)
     }
+
+    // MARK: - Telemetry Events
+
+    @IBAction func didTapTelemetryEvent(_ sender: Any) {
+        guard let button = sender as? UIButton, let title = button.currentTitle else {
+            return
+        }
+        button.disableFor(seconds: 0.5)
```
mariedm, line 190:
```suggestion
        button.isUserInteractionEnabled = false
        defer { button.isUserInteractionEnabled = true } 
```

tools/rum-models-generator/Sources/CodeGeneration/Print/SwiftPrinter.swift
```diff
@@ -157,9 +157,11 @@ public class SwiftPrinter: BasePrinter, CodePrinter {
     /// This will only be effective for types that support dynamically coded properties. If all properties are statically coded,
     /// the `Codable` implementation is generated by compiler.
     private func printExtensionsWithCodableImplementation(_ swiftStruct: SwiftStruct, parentSwiftStructs: [SwiftStruct] = []) throws {
-        let codedProperties = swiftStruct.properties.filter { $0.defaultValue == nil || $0.mutability != .immutable }
-        let staticallyCodedProperties = codedProperties.filter { $0.codingKey.isStatic }
-        let dynamicallyCodedProperties = codedProperties.filter { !$0.codingKey.isStatic }
+        let staticallyEncodedProperties = swiftStruct.properties.filter { $0.codingKey.isStatic }
+        let staticallyDecodedProperties = staticallyEncodedProperties.filter { !($0.defaultValue != nil && $0.mutability == .immutable) }
```
mariedm, line 161: Isn't it the same as `($0.defaultValue == nil || $0.mutability != .immutable)`? Which is a bit easier to read.

mariedm (dismissed):

Datadog/Example/Debugging/DebugRUMViewController.swift
ncreated: @mariedm 🤔 why? The `.disableFor(seconds: 0.5)` is pure utility, blocks the button for 0.5s and that's it.

mariedm: I imagine this is to prevent accidental gestures, such as double-tapping? My suggestion ensures the button is disabled while processing and re-enabled once the action is completed. But this is a nit and I'm fine with what you wrote.

ncreated: Actually, [`disableFor(seconds:)`](https://github.com/DataDog/dd-sdk-ios/blob/76fdab4a48c206e6d78f3e93972672c988f554fe/Datadog/Example/Utils/UIButton%2BDisabling.swift#L10-L27) does both: it prevents over-tapping and dims-out the button visually for given number of seconds.
```swift
self.isEnabled = false
self.backgroundColor = .systemGray
```

tools/rum-models-generator/Sources/CodeGeneration/Print/SwiftPrinter.swift
ncreated: ✅ done

PR #2050, opened 2024-09-12
RUM-872 Keep Secured Text Hidden

maxep:
### What and why?

`UITextInputTraits`'s `isSecureTextEntry` and `textContentType` properties are mutable, meaning that a sensitive texts could be revealed to the end user. e.g.: Revealing a password for verification.

But in Session Replay, we should keep considering the text as sensitive even if it was revealed.

### How?

Keep sensitivity in memory using associated object on views conforming to `UITextInputTraits`.

### Review checklist


mariedm (approved):
LGMT! Thank you for taking care of it!

mariedm (approved):

mariedm (approved):

PR #2051, opened 2024-09-13
Benchmark: Allow run without measure

maxep:
### What and why?

Allow running benchmark scenario without measures so we can debug and define s8s tests without collecting data.

### Review checklist

### DIFF
--- a/BenchmarkTests/BenchmarkTests.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
+++ b/BenchmarkTests/BenchmarkTests.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
@@ -54,7 +54,7 @@
          <EnvironmentVariable
             key = "BENCHMARK_RUN"
             value = "instrumented"
-            isEnabled = "YES">
+            isEnabled = "NO">
          </EnvironmentVariable>
          <EnvironmentVariable
             key = "BENCHMARK_SCENARIO"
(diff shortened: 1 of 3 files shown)


BenchmarkTests/BenchmarkTests.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
```diff
@@ -53,8 +53,8 @@
       <EnvironmentVariables>
          <EnvironmentVariable
             key = "BENCHMARK_RUN"
-            value = "instrumented"
-            isEnabled = "YES">
+            value = "metrics"
+            isEnabled = "NO">
```
mariedm, line 57: Will there still be instrumented scenario?

mariedm (dismissed):

BenchmarkTests/BenchmarkTests.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
maxep: This change was not intented, I reverted 👍

mariedm: My question remains though: does this change from `instrumented` set to `YES` to `NO` mean there will be no instrumented benchmark anymore?

maxep: No, this is the env variable for running this app in Debug:
image

When set to `NO` (unselected), the `none` run will be selected. This is for development only, to simply select a run and a scenario, it is not applied to s8s build.

mariedm: Got it, thanks!

PR #2058, opened 2024-09-19
RUM-6224 [Session Replay FGM] Create privacy overrides

mariedm:
### What and why?

This PR introduces the ability to override Session Replay's fine-grained privacy settings (text and input, images, touches) at the view level. This allows customers to apply specific privacy levels to specific views, overriding the global privacy configuration, enabling customers to control masking at a more granular level.

A new `HiddenPrivacyLevel` is also introduced, which, when applied to a view, will replace the view with an opaque wireframe and prevent any interactions or subviews from being recorded. This privacy level is only available as an override and takes precedence over other privacy settings.

### How?

- Created a `HiddenPrivacyLevel` enum
- Added a `SessionReplayOverride` struct to encapsulate all overrideable privacy settings
- Created an extension for `UIView` to access `SessionReplayOverride` via the `.dd.sessionReplayOverride` property
- Used associated objects to store the privacy overrides for each view
- Created a separate set of enums for override properties for Objective-C compatibility, to mimic `nil` values in Swift

Note: This PR only creates the new properties; the logic to apply the overrides will follow in another PR.

### Review checklist


DatadogInternal/Sources/Models/SessionReplay/PrivacyLevel+SessionReplay.swift
```diff
@@ -0,0 +1,81 @@
...
+
+/// Extension on `DatadogExtension` to provide access to `SessionReplayOverride` for any `UIView`.
+extension DatadogExtension where ExtendedType: UIView {
+    /// Provides access to Session Replay override settings for the view.
+    /// Usage: `myView.dd.sessionReplayOverride.textAndInputPrivacy = .maskNone`.
+    public var sessionReplayOverride: SessionReplayOverrideExtension<ExtendedType> {
+        get {
+            return SessionReplayOverrideExtension(self.type)
+        }
+        set {}
```
ganeshnj, line 20: /improvement

intentional? looks like it should be get only from the getter definition.

DatadogInternal/Sources/Models/SessionReplay/PrivacyLevel+SessionReplay.swift
```diff
@@ -0,0 +1,81 @@
...
+        self.view = view
+    }
+
+    /// Text and input privacy override (e.g., mask or unmask specific text fields, labels, etc.).
+    public var textAndInputPrivacy: TextAndInputPrivacyLevel? {
+        get {
+            return objc_getAssociatedObject(view as AnyObject, &associatedTextAndInputPrivacyKey) as? TextAndInputPrivacyLevel
+        }
+        set {
+            objc_setAssociatedObject(view as AnyObject, &associatedTextAndInputPrivacyKey, newValue, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
```
ganeshnj, line 47: /question

as per public docs of `OBJC_ASSOCIATION_RETAIN_NONATOMIC` 

https://developer.apple.com/documentation/objectivec/objc_associationpolicy/objc_association_retain_nonatomic
>Specifies a strong reference to the associated object, and that the association is not made atomically.

This means any strong reference to these extension properties will not allow deinit of views which could be potential issue for customers apps. We would like to stay away from the view life cycle and let the customer decide it.

DatadogInternal/Sources/Models/SessionReplay/PrivacyLevel+SessionReplay.swift
```diff
@@ -0,0 +1,81 @@
...
+
+    /// Hidden privacy override (e.g., mark a view as hidden, rendering it as an opaque wireframe in replays).
+    public var hiddenPrivacy: HiddenPrivacyLevel? {
+        get {
+            return objc_getAssociatedObject(view as AnyObject, &associatedHiddenPrivacyKey) as? HiddenPrivacyLevel
+        }
+        set {
+            objc_setAssociatedObject(view as AnyObject, &associatedHiddenPrivacyKey, newValue, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
+        }
+    }
```
maciejburda, line 79: Maybe simple `Bool` is good enough for this? `.hidden` and `.none` options cause a little bit of confusion imo 🤔

DatadogSessionReplay/Sources/SessionReplay+objc.swift
```diff
@@ -213,7 +213,7 @@ public enum DDImagePrivacyLevel: Int {
     }
 }
 
-/// Available privacy levels for content masking.
+/// Available privacy levels for touch masking.
```
maciejburda, line 216: 👌

DatadogInternal/Sources/Models/SessionReplay/SessionReplayConfiguration.swift
```diff
@@ -57,6 +57,15 @@ public enum TouchPrivacyLevel: String {
     case hide
 }
 
+/// Privacy level for overriding global privacy settings in Session Replay.
+public enum HiddenPrivacyLevel: String {
```
ganeshnj, line 61: /question

is this something customer can set or get? If yes, we must forward pass through SR package. We don't want customers to force import internal.

ganeshnj (commented):
Nothing concerning from business point of view, some questions about the implementation details.

DatadogInternal/Sources/Models/SessionReplay/PrivacyLevel+SessionReplay.swift
maciejburda, line 20: IMO worth writing tests accessing feature through this API directly. In the end this the public API used by the customer

maciejburda (commented):
In general looks good! 👏 

Left some suggestions 🙇

DatadogInternal/Sources/Models/SessionReplay/PrivacyLevel+SessionReplay.swift
mariedm: In addition to the tests I created in SessionReplayOverrideTests? What more would you test?

DatadogInternal/Sources/Models/SessionReplay/PrivacyLevel+SessionReplay.swift
mariedm: I thought of it. I agree that `.none` is a bit confusing, especially since it's used to represent an optional in Swift, which means customers will have to explicitly specify `HiddenPrivacyLevel.none` for the compiler to resolve it correctly. One alternative is to rename it to something like `.disabled`. Another possibility is for the `enum` to have a single case, `.hide` (and the privacy can be clearer by setting it to `nil` since it's optional in the SR extension).

The main advantage of using an `enum` over a `Bool` is that we can extend it later, in case we want to add other use cases (a point @ganeshnj raised at the time of writing the [RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3901129961/RFC+-+Session+Replay+Fine+Grained+Masking+Overrides)). However, for the current use case, a `Bool` might make more sense.

And ideally we want iOS and Android to align (cc @3Ajonathanmos).

DatadogInternal/Sources/Models/SessionReplay/PrivacyLevel+SessionReplay.swift
mariedm: That's a valid point, I can probably use `OBJC_ASSOCIATION_COPY_NONATOMIC` instead.

DatadogInternal/Sources/Models/SessionReplay/SessionReplayConfiguration.swift
mariedm: Yes, customers can set it. I placed it here along the other privacy levels. If we move this enum to the SR module, then we should move the others as well imo.

DatadogInternal/Sources/Models/SessionReplay/PrivacyLevel+SessionReplay.swift
mariedm: Following our meeting, we decided to use a `Boolean` instead.

DatadogInternal/Sources/Models/SessionReplay/PrivacyLevel+SessionReplay.swift
mariedm: It's intentional. The empty setter allows to access the individual properties without requiring additional logic in the setter itself, like this:
`someLabel.dd.sessionReplayOverride.textAndInputPrivacy = .maskAll`

This approach is used with other similar extensions in the codebase (e.g., [DatadogExtended](https://github.com/DataDog/dd-sdk-ios/blob/597c06ea9bfd777b62879678d1b37461a7bb2a29/DatadogInternal/Sources/Extensions/DatadogExtended.swift#L36)).

DatadogInternal/Sources/Models/SessionReplay/PrivacyLevel+SessionReplay.swift
mariedm: I've looked into it, and since the properties being associated are value types, they are copied by default, so using `OBJC_ASSOCIATION_RETAIN_NONATOMIC` or `OBJC_ASSOCIATION_COPY_NONATOMIC` will result in the same behavior. That said, switching to `OBJC_ASSOCIATION_COPY_NONATOMIC` can make sense because the overrides are meant to be isolated per view, not shared between views. This is safer in case we later decide to use reference types.

ganeshnj (approved):

DatadogInternal/Sources/Models/SessionReplay/SessionReplayConfiguration.swift
ganeshnj: You would need to do something like

https://github.com/DataDog/dd-sdk-ios/blob/783407bc8ea9d5027e874b59212116993a8bea2a/DatadogTrace/Sources/TraceConfiguration.swift#L22

if you keep them in internal, but moving to SR module makes more sense if these types are not shared in other modules.

not blocking as we have other similar things already, but we must address this before GA.

maxep (changes requested):
Looks great overall! It's a good use of the extension pattern.

I think we should move the interface and objc tests to the SR module. I also left a suggestion to simplify the `SessionReplayOverrideExtension` definition.

maciejburda (approved):

DatadogInternal/Sources/Models/SessionReplay/PrivacyLevel+SessionReplay.swift
maciejburda: Missed it! thanks

DatadogInternal/Sources/Models/SessionReplay/PrivacyLevel+SessionReplay.swift
```diff
@@ -0,0 +1,81 @@
...
+
+private var associatedTextAndInputPrivacyKey: UInt8 = 3
+private var associatedImagePrivacyKey: UInt8 = 4
+private var associatedTouchPrivacyKey: UInt8 = 5
+private var associatedHiddenPrivacyKey: UInt8 = 6
+
+// MARK: - SessionReplayOverrideExtension
+
+/// `UIView` extension  to manage the Session Replay privacy override settings.
+public struct SessionReplayOverrideExtension<ExtendedType> {
```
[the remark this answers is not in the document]
mariedm: @ganeshnj made a similar comment. I placed the extension in `DatadogInternal` because that's where the privacy levels are defined. Do you know what was the reasoning for placing them in the internal module initially?

DatadogCore/Tests/DatadogObjc/DDSessionReplayOverridesTests.swift
```diff
@@ -0,0 +1,107 @@
...
+
+#if os(iOS)
+
+import XCTest
+import TestUtilities
+import DatadogInternal
+
+@testable import DatadogSessionReplay
+
+class DDSessionReplayOverrideTests: XCTestCase {
```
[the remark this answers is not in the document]
mariedm: @ganeshnj made a similar comment. I placed the extension in DatadogInternal because that's where the privacy levels are defined. Do you know what was the reasoning for placing them in the internal module initially?

DatadogInternal/Sources/Models/SessionReplay/PrivacyLevel+SessionReplay.swift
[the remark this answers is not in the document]
mariedm: Thanks for the suggestion! But the constraint on `UIView` is wanted, as we specifically want these properties to be available only on `UIView`. Don't you think?

DatadogCore/Tests/DatadogObjc/DDSessionReplayOverridesTests.swift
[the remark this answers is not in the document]
maxep: That's because the privacy levels are shared with the webview tracking module. But the override interface won't be shared.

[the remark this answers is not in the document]
mariedm: Got it, I'll move the extension to SR then.

DatadogInternal/Sources/Models/SessionReplay/PrivacyLevel+SessionReplay.swift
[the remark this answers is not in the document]
maxep: Ys, the constraint for `UIView` stays:
```swift
extension DatadogExtension where ExtendedType: UIView 
```
What I'm proposing is to make `SessionReplayOverrideExtension` a class and to use `where ExtendedType: AnyObject` so you don't have to do `view as AnyObject` in associated object.
To go even further, you don't really need a generic type here. You can directly use `private let view: UIView`, unless I'm missing something.

DatadogInternal/Sources/Models/SessionReplay/PrivacyLevel+SessionReplay.swift
maxep: Just to clarify: using `OBJC_ASSOCIATION_RETAIN_NONATOMIC` will retain the associated value, not the view. But in this case, using `COPY` is more explicit 👍

DatadogInternal/Sources/Models/SessionReplay/PrivacyLevel+SessionReplay.swift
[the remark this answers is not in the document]
mariedm: Ok, I get it now. It makes sense, I will use your suggestion.

DatadogSessionReplay/Sources/PrivacyLevel+SessionReplay.swift
```diff
@@ -0,0 +1,79 @@
...
+    private let view: UIView
+
+    public init(_ view: UIView) {
+        self.view = view
+    }
+
+    /// Text and input privacy override (e.g., mask or unmask specific text fields, labels, etc.).
+    public var textAndInputPrivacy: TextAndInputPrivacyLevel? {
+        get {
+            return objc_getAssociatedObject(view as AnyObject, &associatedTextAndInputPrivacyKey) as? TextAndInputPrivacyLevel
```
maxep, line 42: /nit it's now redundant
```suggestion
            return objc_getAssociatedObject(view, &associatedTextAndInputPrivacyKey) as? TextAndInputPrivacyLevel
```

maxep (approved):
Nice!!

DatadogSessionReplay/Sources/PrivacyLevel+SessionReplay.swift
mariedm: Yes, I noticed and fixed it in the next PR :p

PR #2063, opened 2024-09-25
RUM-3175 fix: Fix race condition during consent change (pending → granted)

ganeshnj:
### What and why?

📦🐞 This PR fixes a race condition that occurs during the change of tracking consent from `.pending` to `.granted`. This issue could result in the loss of events recorded on the current thread right before the consent change. 

### How?

The fix involves synchronizing data migration on the context queue to prevent conflicts with ongoing "event write" operations. This change guarantees that all latest events are accurately migrated based on the updated consent state.

### Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- [FIX] Fix race condition during consent change, preventing loss of events recorded on the current thread. See [#2063][]
 - [IMPROVEMENT] Support mutation of events' attributes. See [#2099][]
 
 # 2.19.0 / 28-10-2024
@@ -786,6 +787,7 @@ Release `2.0` introduces breaking changes. Follow the [Migration Guide](MIGRATIO
 [#2088]: https://github.com/DataDog/dd-sdk-ios/pull/2088
 [#2083]: https://github.com/DataDog/dd-sdk-ios/pull/2083
 [#2099]: https://github.com/DataDog/dd-sdk-ios/pull/2099
+[#2063]: https://github.com/DataDog/dd-sdk-ios/pull/2063
 [@00fa9a]: https://github.com/00FA9A
 [@britton-earnin]: https://github.com/Britton-Earnin
 [@hengyu]: https://github.com/Hengyu
--- a/DatadogCore/Sources/Core/DatadogCore.swift
+++ b/DatadogCore/Sources/Core/DatadogCore.swift
@@ -160,7 +160,12 @@ internal final class DatadogCore {
     /// - Parameter trackingConsent: new consent value, which will be applied for all data collected from now on
     func set(trackingConsent: TrackingConsent) {
         if trackingConsent != consentPublisher.consent {
-            allStorages.forEach { $0.migrateUnauthorizedData(toConsent: trackingConsent) }
+            contextProvider.queue.async { [allStorages] in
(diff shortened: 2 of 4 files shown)


DatadogCore/Sources/Core/Storage/Files/Directory.swift
```diff
@@ -155,7 +155,7 @@ internal struct Directory: DirectoryProtocol {
         try retry(times: 3, delay: 0.001) {
             try files().forEach { file in
                 let destinationFileURL = destinationDirectory.url.appendingPathComponent(file.name)
-                try? retry(times: 3, delay: 0.0001) {
+                try retry(times: 3, delay: 0.0001) {
```
ganeshnj, line 158: i don't see why we should ignore the error here, looks like miss in past.

DatadogCore/Sources/Core/DatadogCore.swift
```diff
@@ -159,8 +159,15 @@ internal final class DatadogCore {
     ///
     /// - Parameter trackingConsent: new consent value, which will be applied for all data collected from now on
     func set(trackingConsent: TrackingConsent) {
+        // There could be events in the queue that were collected before the consent change
+        // but not written to the disk yet, which must be migrated to the new consent.
+        // Flush ensures we are in a consistent state before the consent change.
+        flush()
```
mariedm, line 165: Do we want to flush the data whatever the consent value is?

maxep (changes requested):
The `flush` is a sync method:
```
/// Flushes asynchronous operations related to events write, context and message bus propagation in this instance of the SDK
/// with **blocking the caller thread** till their completion.
```
We should not block the user thread here, the flush method is mostly used for testing.
Instead, it might be sufficient to just publish the new consent before doing the migration. Or to call the migration from the context queue, also after publishing the new consent.

I would also be great to have a test case covering this, to prevent regression.

CHANGELOG.md
```diff
@@ -1,9 +1,12 @@
 # Unreleased
 
+- [FIX] Force events flushing by performing consent change on context queue. See [#2063][]
+
 # 2.18.0 / 25-09-2024
 - [IMPROVEMENT] Add overwrite required (breaking) param to addViewLoadingTime & usage telemetry. See [#2040][]
 - [FEATURE] Prevent "show password" features from revealing sensitive texts in Session Replay. See [#2050][]
 - [FEATURE] Add Fine-Grained Masking configuration options to Session Replay. See [#2043][]
+- [FEATURE] Prevent "show password" features from revealing sensitive texts. See [#2050][]
```
mariedm, line 9: This is already there, 2 lines above

ganeshnj: bad rebase

mariedm (dismissed):

DatadogCore/Sources/Core/DatadogCore.swift
ncreated: N/A anymore with the new fix we do it without flushing

ncreated:
> (...) to call the migration from the context queue, also after publishing the new consent.

@maxep Fair call ✅. I continued this work fixing it in proposed way ☝️. It's not necessary to publish the consent before migration task as migration depends on the new consent value explicitly.

mariedm (approved):

maxep (approved):
Nice, on target 🎯 😉

PR #2065, opened 2024-09-25
RUM-6223 Implement Hidden Override logic

mariedm:
### What and why?

This PR implements the logic for the hidden property in Session Replay, as a continuation of #2058, which introduces Fine-Grained Masking (FGM) overrides.

### How?

- Added a new `SessionReplayOverrideExtension` property to `ViewAttributes`.
- In the view recorder, it now checks for the `hiddenPrivacy` property.
- When this property is set, it returns a placeholder frame for the hidden view and ignores its child views.
- Renamed `hiddenPrivacy` to `hide`, as it is now a `Bool`, making the name more intuitive after converting from an `enum`.
- Addressed [this comment](https://github.com/DataDog/dd-sdk-ios/pull/2058#discussion_r1775231721) from the previous PR to clean up the `SessionReplayOverrideExtension` implementation.

### Review checklist

### DIFF
--- a/DatadogSessionReplay/Sources/SessionReplayOverrides+objc.swift
+++ b/DatadogSessionReplay/Sources/SessionReplayOverrides+objc.swift
@@ -9,6 +9,27 @@ import Foundation
 import DatadogInternal
 import UIKit
 
+private var associatedSROverrideKey: UInt8 = 0
+
+/// Objective-C accessible extension for UIView
+@objc
+public extension UIView {
+    @objc var ddSessionReplayOverride: DDSessionReplayOverride {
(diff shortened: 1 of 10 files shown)


DatadogSessionReplay/Sources/SessionReplayOverrides+objc.swift
maxep, line 1: /question not related to this change:
How an objc user will set the override? it doesn't look like the UIView extension is exposed.

maxep (approved):
Looks good 👍 

I just left a question which is most-likely a follow up on the objc interface.

PR #2074, opened 2024-10-03
RUM-6263 Trigger based approach

maciejburda:
### What and why?

This PR changes the default behaviour of the SR recorder from scheduler to trigger based approach. Significantly reducing the SR footprint on idle.

### How?

It's achieved by introducing `RecordingTrigger` which is responsible for subscribing to two events: `UIView.layoutSubviews` and `UIApplication.sendEvent` using obj-c swizzling. Trigger takes a weak dependency on the recording coordinator to start/stop it, as well as invoke captureNextRecord.

Those events happen frequently (every touch interaction and each view's layout change), they are throttled for 100ms (which matches scheduler capture frequency).

Below you can find screenshots from the basic Xcode profiler showcasing CPU usage for similar scenario (go through checkout and stay idle on dashboard):

*Before*
Screenshot 2024-10-09 at 13 15 16

*After*
Screenshot 2024-10-09 at 13 09 47

### Review checklist


DatadogSessionReplay/Sources/Recorder/RecordingTrigger.swift
```diff
@@ -0,0 +1,62 @@
...
+    deinit {
+        uiViewSwizzler?.unswizzle()
+        uiApplicationSwizzler?.unswizzle()
+    }
+
+    func startWatchingTriggers() {
+        if recordingCoordinator?.startRecording() == true {
+            uiViewSwizzler?.swizzle()
+            uiApplicationSwizzler?.swizzle()
+        }
```
ncreated, line 42: question/ Is is this check somehow evaluated after RUM context is changed to "new session"? The underlying return value from `startRecording()` is:
```swift
return recordingEnabled && isSampled
```
where `isSampled` is updated in `onRUMContextChanged(rumContext:)`. It seems that `RecordingTrigger` doesn't re-evaluate swizzling on context change. Ultimately, this may lead to "no recording" when first session was not sampled so swizzlings are not installed, but the succeeding session is sampled but swizzlings remain to be off.

DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
```diff
@@ -115,46 +113,68 @@ internal class RecordingCoordinator {
...
-    /// Captures the next recording if conditions are met.
-    private func captureNextRecord() {
-        /// We don't capture any snapshots if the RUM context has no view ID.
-        guard let rumContext = currentRUMContext,
-              let viewID = rumContext.viewID else {
-            return
-        }
+    private var lastFrameTimestamp: Date?
+    private var shouldSkipFrame: Bool {
+        return Date().timeIntervalSince(lastFrameTimestamp ?? .distantPast) < Constants.throttingRate
```
ncreated, line 118: suggestion/ Using `dateProvider` would let us add coverage to skipping frames without blocking tests thread. No risk of flakiness also as `dateProvider` mock is more explicit than "current real time".

DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
```diff
@@ -115,46 +113,68 @@ internal class RecordingCoordinator {
...
+            // We don't capture any snapshots if the RUM context has no view ID.
+            guard let rumContext = currentRUMContext,
+                  let viewID = rumContext.viewID else {
+                return
+            }
+            // Skip frame if there's pressure on processing
+            guard shouldSkipFrame == false else {
+                return
+            }
+            lastFrameTimestamp = Date()
```
ncreated, line 139: change-request/ Throttling from inside the async operation is quite inefficient as it leads to flooding the `main` queue with bunch of unnecessary tasks. From my test in Shopist app:
- when app is launched, the `main` queue is flooded with nearly 200 async tasks due to `layoutSubviews` trigger;
- when touching the screen and moving finger, it is again flooded with about 50 tasks per second.

It can be avoided if the "should skip frame" check is moved before the task gets dispatched. Note that we discuss the performance impact on the UI thread, so we must do whatever it takes to unload it from unnecessary consumption.

ncreated (changes requested):
Great optimization idea 👍 💯 . I left few feedbacks, but most importantly a change request to offload the main thread from unnecessary tasks.

DatadogSessionReplay/Sources/Recorder/RecordingTrigger.swift
maciejburda: Yeah - that's a good point. Trigger will keep firing even the next session is sampled out.

maciejburda:
I have addressed all the feedback, which led to small refactor. But I think it makes more sense now.

Key change is that trigger is a dependency of coordinator and it notifies it through a delegate pattern.

On checking `shouldSkipFrame` outside of queue. I made some tests and it's significantly more performant when done from within the queue.
Screenshot 2024-10-23 at 13 37 00
Screenshot 2024-10-23 at 13 38 20

DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
```diff
@@ -115,46 +119,71 @@ internal class RecordingCoordinator {
         srContextPublisher.setHasReplay(hasReplay)
     }
 
+    private var lastFrameTimestamp: Date?
+    private var shouldSkipFrame: Bool {
+        return dateProvider.now.timeIntervalSince(lastFrameTimestamp ?? .distantPast) < Constants.throttingRate
+    }
```
maxep, line 125: /suggestion We could have a `Debouncer` object that could take care of this, similar to the future `TimeBank`, WDYT?

DatadogSessionReplay/Sources/Recorder/RecordingTrigger.swift
```diff
@@ -0,0 +1,62 @@
...
+    func startWatchingTriggers()
+    func stopWatchingTriggers()
+
+    var delegate: RecordingTriggerDelegate? { get set }
+}
+
+/// Delegate to notify about recording trigger.
+internal protocol RecordingTriggerDelegate: AnyObject {
+    func didTrigger()
+}
```
maxep, line 23: /suggestion I'm not a fan of delegate property in a protocol, because you cannot enforce `weak` from the protocol property. So one could assume it's a retain while it is not in the concrete type. Instead, we could a simple callback such as
```swift
internal protocol RecordingTriggering {
    func startWatchingTriggers(_ callback: @escaping () -> Void)
    func stopWatchingTriggers()
}
```
Would that work?

maxep (changes requested):
Looks good 👍 I have left some suggestions and a change-request regard thread safety

DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
maciejburda: Totally - but for now I'd suggest merging this as is - and refactor once we have timebank in place. It will shed some more slight on our approach of unit testing this

DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
```diff
@@ -115,46 +119,71 @@ internal class RecordingCoordinator {
...
-        } catch let objc as ObjcException {
-            telemetry.error("[SR] Failed to take snapshot due to Objective-C runtime exception", error: objc.error)
-            // An Objective-C runtime exception is a severe issue that will leak if
-            // the framework is not built with `-fobjc-arc-exceptions` option.
-            // We recover from the exception and stop the scheduler as a measure of
-            // caution. The scheduler could start again at a next RUM context change.
-            stopRecording()
-        } catch {
-            telemetry.error("[SR] Failed to take snapshot", error: error)
+        queue.run { [weak self] in
```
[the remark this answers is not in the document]
maciejburda: this runs async job on main thread

maciejburda:
Addressed all the requested changes, and did [some extra tests (internal)](https://mobile-integration.datadoghq.com/rum/replay/sessions/c438d5df-8db9-4c47-b293-7635895db57b?applicationId=5e20092a-ba66-4868-bfa3-ee8a0f1a54a1&p_devtools_tab=performance&seed=70bffd1a-f83d-4c22-bded-f305db5397ba&ts=1729694271248).

Screenshot 2024-10-23 at 15 39 04
Screenshot 2024-10-23 at 15 40 40

DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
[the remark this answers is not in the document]
maxep: I mean the `captureNextRecord` is called by triggers from the main thread, so I wonder why we jump on the main thread again?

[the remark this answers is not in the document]
maciejburda: Well - that's actually a good question. Implementation-wise it's always main thread. Although it's not enforced by design right now.

Double checking.

maxep (dismissed):
Looks great! Thanks for the changes 🙏

DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
```diff
@@ -115,9 +123,26 @@ internal class RecordingCoordinator {
...
+        guard shouldSkipTrigger == false else {
+            return
+        }
+        lastTriggerDate = dateProvider.now
+        captureNextRecord()
+    }
+
     private func captureNextRecord() {
-        /// We don't capture any snapshots if the RUM context has no view ID.
+        guard isSampled == true && recordingEnabled == true else {
```
mariedm, line 141: It seems like this check isn't necessary since it’s already handled in `evaluateRecordingConditions`. If you’d like to keep it as an additional safety check, we could centralize this logic with a computed property:

```
private var shouldRecord: Bool {
    return isSampled && recordingEnabled
}
```

DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
```diff
@@ -43,59 +51,59 @@ internal class RecordingCoordinator {
...
         sampler: Sampler,
         telemetry: Telemetry,
-        startRecordingImmediately: Bool,
-        methodCallTelemetrySamplingRate: Float = 0.1
-    ) {
+        recordingTrigger: RecordingTriggering,
+        methodCallTelemetrySamplingRate: Float = 0.1,
+        dateProvider: DateProvider = SystemDateProvider(),
+        queue: Queue = MainAsyncQueue()
+    ) throws {
```
mariedm, line 58: Would it make sense to add some telemetry in cases the swizzling or recorder initialization fails? This could help in diagnosing potential customer issues.

mariedm (commented):
Awesome work! And this CPU improvement is very welcome! 👏

General question: Since we're fully transitioning to swizzling for SR recording, should we consider updating the documentation to add warnings for customers? E.g., potential conflicts with other SDKs or their own code that might also be using swizzling.

DatadogSessionReplay/Tests/Recorder/RecordingCoordinatorTests.swift
```diff
@@ -164,56 +180,13 @@ class RecordingCoordinatorTests: XCTestCase {
...
+        recordingTriggerMock.triggerCallback?()
 
         // Then
         let metric = try XCTUnwrap(telemetry.messages.last?.asMetric)
         XCTAssertEqual(metric.name, "Method Called")
     }
 
-    // MARK: StartRecordingImmediately Initialization Tests
-
-    func test_whenStartRecordingImmediatelyIsDefault_itShouldRecord() throws {
```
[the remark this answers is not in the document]
maciejburda: That's interesting point. It's mostly about lack of ability to test `SessionReplayFeature` properly, because after the refactor - that's the object responsible for calling it.

IMO we should revisit how we serve all the dependencies to it, and make it testable. Ideally in a separate PR.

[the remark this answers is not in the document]
maciejburda: Entire thing happens in `SessionReplayFeature` and that's the object responsible for starting the recording immediately when `configuration.startRecordingImmediately` is set to `true`. We can move this responsibility down to `RecordingCoordinator`, but it will only check if the new implementation is working.

In order to test `SessionReplayFeature` we would need to introduce dependency factory to be able to inject a mock. This way we could check that if `startRecordingImmediately == true`,  `SessionReplayFeature` calls `startRecording()` on `recordingCoordinator`.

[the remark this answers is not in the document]
maxep: I was able to re-enable this coverage in 9019d8d4a47416356005d7a53603ef0c87508149 by simulating start/stop with triggers 👍  I also fixed a flaky tests

maxep (dismissed):

maxep (approved):

maxep (approved):

DatadogSessionReplay/Tests/Recorder/RecordingCoordinatorTests.swift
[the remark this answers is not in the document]
mariedm: @maciejburda it looks like you forgot to re-add these tests?

DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
```diff
@@ -43,62 +51,62 @@ internal class RecordingCoordinator {
...
+            recordingTrigger.stopWatchingTriggers()
+        }
+        updateHasReplay()
+    }
+
+    // MARK: Private
 
     private func onRUMContextChanged(rumContext: RUMContext?) {
-        if currentRUMContext?.sessionID != rumContext?.sessionID || currentRUMContext == nil {
+        if currentRUMContext?.sessionID != rumContext?.sessionID || currentRUMContext == nil && rumContext != nil {
```
mariedm, line 109: I'm not sure I understand this addition. In which scenario `currentRUMContext` would be `nil` and `rumContext` **not** `nil`? When switching views? Is it really what we want to check here? 

Parenthesis could improve readability.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - [FEATURE] Add Privacy Overrides in Session Replay. See [#2088][]
 - [IMPROVEMENT] Add ObjC API for the internal logging/telemetry. See [#2073][]
+- [IMPROVEMENT] Session Replay recording is event-triggered instead of scheduled. See [#2074]
```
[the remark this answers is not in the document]
mariedm: It's still there and duplicated; it should only be present in the `Unreleased` section.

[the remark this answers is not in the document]
maciejburda: Good catch! rebase issue :/

DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
maciejburda: Yeah - it's actually unnecessary, because we have checks in other places.

Normally this scenario is impossible, maybe except for some edge cases where RUM gets deactivated during the runtime. It's here mostly to satisfy unit tests.

mariedm (dismissed):

DatadogSessionReplay/Tests/Recorder/RecordingCoordinatorTests.swift
```diff
@@ -33,38 +35,53 @@ class RecordingCoordinatorTests: XCTestCase {
...
         // When
         rumContextObserver.notify(rumContext: .mockRandom())
+        recordingTriggerMock.triggerCallback?()
 
         // Then
-        XCTAssertFalse(scheduler.isRunning)
         XCTAssertEqual(try core.context.baggages["sr_has_replay"]?.decode(), false)
         XCTAssertEqual(recordingMock.captureNextRecordCallsCount, 0)
+        XCTAssertEqual(recordingTriggerMock.startWatchingTriggersCallsCount, 0)
+        XCTAssertEqual(recordingTriggerMock.stopWatchingTriggersCallsCount, 2)
```
ncreated, line 67: question/ Why `2`? I'd expect it to be stopped only once. Asking because `stopWatchingTriggers()` implies unswizzling which manages a `lock`. If this is vastly overcalled, may cause perf penalty.

DatadogSessionReplay/Tests/Recorder/RecordingTriggerTests.swift
```diff
@@ -0,0 +1,65 @@
...
+        recordingTrigger.startWatchingTriggers {
+            didTriggerCalledCount += 1
+        }
+        UIApplication.shared.sendEvent(UIEvent())
+
+        XCTAssertEqual(didTriggerCalledCount, 0)
+    }
+
+    private func randomTrigger() {
+        if Int.random(in: 1...100) > 50 {
```
ncreated, line 56: nit/ It's just `Bool.random()` but doesn't change anything.

ncreated (dismissed):
Two minor comments, looks good overall 👍 💪

DatadogSessionReplay/Tests/Recorder/RecordingCoordinatorTests.swift
maciejburda: It's because `rumContext` change triggers recording `evaluateRecordingConditions` which starts/stops based on some criteria.

There's a simple way to improve this by knowing if it's recording or not. I'll add that change.

DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
```diff
@@ -93,14 +95,17 @@ internal class RecordingCoordinator {
...
+            isRecording = true
             recordingTrigger.startWatchingTriggers { [weak self] in
                 self?.didTrigger()
             }
-        } else {
+            updateHasReplay()
+        } else if isRecording {
+            isRecording = false
             recordingTrigger.stopWatchingTriggers()
+            updateHasReplay()
```
mariedm, line 107: So this change is to avoid updating the context through `updateHasReplay`, correct? 

What happens if we are recording and we receive a context update? It looks like since `isRecording` is `true`, it will stop the recording.

maciejburda: Yeah - it got caught by the integration test. Great catch!

It should be fixed now.

maciejburda:
We decided to slightly revamp the approach. Closing this one for the time being.

maciejburda:
Just noticed I need to add tests for the new swizzlers. On it. Will reopen when done.

DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
```diff
@@ -43,59 +53,62 @@ internal class RecordingCoordinator {
         recorder: Recording,
         sampler: Sampler,
         telemetry: Telemetry,
-        startRecordingImmediately: Bool,
-        methodCallTelemetrySamplingRate: Float = 0.1
-    ) {
+        recordingTrigger: RecordingTriggering,
+        methodCallTelemetrySamplingRate: Float = 0.1,
+        dateProvider: DateProvider = SystemDateProvider(),
+        queue: Queue = MainAsyncQueue()
```
maxep, line 59: /suggestion To prevent test flakiness in the future, it might be best to remove the default `queue` value and let the feature inject the instance. This way, we are forced to specify a queue in tests.

DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
maxep: /suggestion The statements are difficult to understand and redundant with the `updateHasReplay()`. We could make it both easier to read and more straightforward with:
```swift
let hasReplay = isSampled && recordingEnabled
if !isRecording && hasReplay && currentRUMContext != nil {
    isRecording = true
    recordingTrigger.startWatchingTriggers { [weak self] in
        self?.didTrigger()
    }
    srContextPublisher.setHasReplay(hasReplay)
} else if isRecording && !hasReplay {
    isRecording = false
    recordingTrigger.stopWatchingTriggers()
    srContextPublisher.setHasReplay(hasReplay)
}

```
Also, I don't see the reason why adding `currentRUMContext` here?

DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
```diff
@@ -115,9 +128,22 @@ internal class RecordingCoordinator {
...
 
-    /// Captures the next recording if conditions are met.
+    private var lastTriggerDate: Date?
+
+    private var shouldSkipTrigger: Bool {
+        return dateProvider.now.timeIntervalSince(lastTriggerDate ?? .distantPast) < Constants.throttlingRate
+    }
+
+    private func didTrigger() {
+        guard shouldSkipTrigger == false else {
```
maxep, line 138: /nit `guard !shouldSkipTrigger else`

DatadogSessionReplay/Sources/Recorder/RecordingCoordinator.swift
```diff
@@ -157,6 +185,10 @@ internal class RecordingCoordinator {
         telemetry.stopMethodCalled(methodCalledTrace, isSuccessful: isSuccessful)
     }
 
+    private enum Constants {
+        static let throttlingRate: TimeInterval = 0.1 // 100ms
```
maxep, line 189: /question Why changing from [config](https://github.com/DataDog/dd-sdk-ios/pull/2074/files#diff-dde69db27680d1c31b894abbb9a1ddc5009ecb3a51e555a112302faed0306834L52) to constant?

DatadogSessionReplay/Sources/Recorder/RecordingTrigger.swift
```diff
@@ -0,0 +1,71 @@
...
+        triggerCallback?()
+    }
+
+    func notify_viewDidAppear(viewController: UIViewController) {
+        triggerCallback?()
+    }
+
+    func notify_sendEvent(application: UIApplication, event: UIEvent) {
+        triggerCallback?()
+    }
```
maxep, line 69: Could elaborate on why these 5 methods? The PR description only mention 2 and I don't see why we would need `notify_viewDidAppear` or `notify_setNeedsDisplay` vs. `notify_draw` vs. `notify_layoutSubviews` 🤔

DatadogSessionReplay/Sources/Recorder/UIViewSwizzler.swift
```diff
@@ -0,0 +1,57 @@
...
+            self.method = try dd_class_getInstanceMethod(UIView.self, Self.selector)
+            self.handler = handler
+        }
+
+        func swizzle() {
+            typealias Signature = @convention(block) (UIView) -> Void
+            swizzle(method) { previousImplementation -> Signature in
+                return { [weak handler = self.handler] view  in
+                    handler?.notify_layoutSubviews(view: view)
+                    previousImplementation(view, Self.selector)
```
maxep, line 51: /question/change-request Is there a reason why it notifies **before** the previous imp? Maybe we should let the rendering happen before notifying? Same goes for other swizzlers.

DatadogSessionReplay/Tests/Recorder/RecordingCoordinatorTests.swift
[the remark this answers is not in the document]
maxep: Yes, these are missing again 🤔

IntegrationTests/IntegrationScenarios/Scenarios/SessionReplay/SRMultipleViewsRecordingScenarioTests.swift
```diff
@@ -13,6 +13,11 @@ private extension ExampleApplication {
         buttons["NEXT"].safeTap(within: 5)
     }
 
+    /// Triggers SR recording
+    func deadClick() {
+        windows.firstMatch.tap()
+    }
```
maxep, line 19: Why do we need this now?

maxep (approved):
It looks good, I like the split between `RecordingCoordinator` and `RecordingTrigger` 👍 

What's less clear to me is the various method that we swizzle, it seems empirical and I'm not sure why we need so many. For example, `CALayer.draw` seems sufficient to catch all rendering, then why do we need `CALayer.setNeedsDisplay` and `UIView.layoutSubviews`? In this context, `UIViewController.viewDidAppear` seems redundant.

Also, I think we should let the previous swizzled implementation execute and then notify to coordinator. To not delay the CoreGraphics rendering and record the new layout.

I also left some minor question/nit but these 2 points are the most important.

DatadogSessionReplay/Sources/Utilities/Schedulers/Scheduler.swift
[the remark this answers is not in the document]
maciejburda: It will be done differently as a part of a separate PR. Most likely trigger will have two modes. Time based and event based.

DatadogSessionReplay/Sources/Recorder/RecordingTrigger.swift
maciejburda: I cleaned it up. I was playing with many different permutations and this one seems the most stable.

Although I'll give it another shot with Maxime's suggestion below to trigger the notification before the implementation.

IntegrationTests/IntegrationScenarios/Scenarios/SessionReplay/SRMultipleViewsRecordingScenarioTests.swift
maciejburda: Still needed as integration tests view are quite static.

PR #2077, opened 2024-10-04
Fix Session Replay ObjC interface

mariedm:
### What and why?

The current Objective-C interface for Session Replay doesn't adhere to the internal guidelines for Objective-C interoperability (see [RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3157787243/RFC+-+Modular+Objective-C+Interface#Recommended-solution)).

### How?

- Applied `@_spi(objc)` to hide the ObjC interfaces from Swift consumers while keeping them accessible to ObjC clients.
- Prefixed types with `objc_*` to differentiate them from Swift-native types.
- Updated the ObjC interface for `SessionReplay`, `SessionReplayConfiguration`, and the different privacy levels following the modular interface guidelines.
- Moved `DDSessionReplayTests` to the Session Replay module

### Review checklist


mariedm:
> Is the following a potentially backward-incompatible change?
> 
> > Prefixed types with objc_* to differentiate them from Swift-native types.
> 
> It is changing the name of the exported symbols, right?

This won't cause any backward compatibility issues. ObjC users will still use `DDSessionReplay` and and `DDSessionReplayConfiguration` thanks to the `@objc(xxx)` annotation. The exported symbols will stay the same.

ncreated (approved):
👍 Happy to see this cleanup, well done.

PR #2080, opened 2024-10-10
RUM-6376 Only enable single core for Session Replay

mariedm:
### What and why?

While we support multiple core instances for most features, we want to restrict this for Session Replay. 

### How?

- Before enabling Session Replay, check if any instance has already enabled the feature.
- If Session Replay has been enabled, then an error is thrown, and a telemetry debug message is logged

Screenshot 2024-10-10 at 13 40 23

### Review checklist

### DIFF
--- a/DatadogInternal/Sources/CoreRegistry.swift
+++ b/DatadogInternal/Sources/CoreRegistry.swift
@@ -76,4 +76,17 @@ public final class CoreRegistry {
     public static func instance(named name: String) -> DatadogCoreProtocol {
         instances[name] ?? NOPDatadogCore()
     }
+
+    /// Checks if the specified `DatadogFeature` is enabled for any registered core instance.
+    ///
+    /// - Parameter feature: The feature type to check for.
+    /// - Returns: `true` if the feature is enabled in at least one instance, otherwise `false`.
+    public static func isFeatureEnabled<T>(feature: T.Type) -> Bool where T: DatadogFeature {
+        for instance in instances.values {
+            if instance.get(feature: T.self) != nil {
+                return true
+            }
+        }
+        return false
+    }
 }
--- a/DatadogSessionReplay/Sources/SessionReplay.swift
+++ b/DatadogSessionReplay/Sources/SessionReplay.swift
@@ -69,6 +69,14 @@ public enum SessionReplay {
                 description: "Datadog SDK must be initialized before calling `SessionReplay.enable(with:)`."
             )
         }
+
+        guard core.get(feature: SessionReplayFeature.self) == nil else {
+            core.telemetry.send(telemetry: .debug(id: "1", message: "Session Replay has already been enabled", attributes: nil))
+            throw ProgrammerError(
+                description: "Session Replay is already enabled and does not support multiple instances. The existing instance will continue to be used."
+            )
+        }
+
         guard configuration.replaySampleRate > 0 else {
             return
         }
--- a/DatadogSessionReplay/Sources/SessionReplay.swift
+++ b/DatadogSessionReplay/Sources/SessionReplay.swift
@@ -70,8 +70,8 @@ public enum SessionReplay {
             )
         }
 
-        guard core.get(feature: SessionReplayFeature.self) == nil else {
-            core.telemetry.send(telemetry: .debug(id: "1", message: "Session Replay has already been enabled", attributes: nil))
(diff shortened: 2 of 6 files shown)


DatadogSessionReplay/Sources/SessionReplay.swift
```diff
@@ -69,6 +69,14 @@ public enum SessionReplay {
                 description: "Datadog SDK must be initialized before calling `SessionReplay.enable(with:)`."
             )
         }
+
+        guard core.get(feature: SessionReplayFeature.self) == nil else {
```
maxep, line 73: /change-request It is only checking the core given as parameter, but we should check all registered cores in [CoreRegistry](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogInternal/Sources/CoreRegistry.swift).

DatadogSessionReplay/Sources/SessionReplay.swift
```diff
@@ -69,6 +69,14 @@ public enum SessionReplay {
...
+            core.telemetry.send(telemetry: .debug(id: "1", message: "Session Replay has already been enabled", attributes: nil))
```
maxep, line 74: /change-request The `id` is made for deduplication, you can use [this method](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogInternal/Sources/Telemetry/Telemetry.swift#L288) instead.
```suggestion
            core.telemetry.debug("Session Replay has already been enabled")
```

maxep (changes requested):

DatadogSessionReplay/Sources/SessionReplay.swift
mariedm: Absolutely! Thanks for the catch.

DatadogInternal/Sources/CoreRegistry.swift
```diff
@@ -76,4 +76,20 @@ public final class CoreRegistry {
...
     }
+
+    /// Checks if the specified `DatadogFeature` is enabled for any registered core instance.
+    ///
+    /// - Parameter feature: The feature type to check for.
+    /// - Returns: `true` if the feature is enabled in at least one instance, otherwise `false`.
+    public static func isFeatureEnabled<T>(feature: T.Type) -> Bool where T: DatadogFeature {
+        guard instances.count > 0 else {
+            return false
+        }
```
maxep, line 87: /nit not really necessary, the for loop will resume immediately if the array is empty :)

DatadogSessionReplay/Tests/SessionReplayTests.swift
```diff
@@ -17,12 +17,14 @@ class SessionReplayTests: XCTestCase {
 
     override func setUpWithError() throws {
         core = FeatureRegistrationCoreMock()
+        CoreRegistry.register(default: core)
```
maxep, line 20: /question is it necessary for this test?

mariedm: Yes, I need it for the new test, otherwise there are no instance registered in `CoreRegistry` and `isFeatureEnabled` will always return `false`. I can also include it only where it's needed, in `testWhenEnabledMultipleTimes_itPrintsError`. Or remove the test altogether.

maxep: ah sorry, I missed the test below. Let's keep it as is 👍

maxep (approved):
Thanks!

DatadogSessionReplay/Sources/SessionReplay.swift
```diff
@@ -69,6 +69,14 @@ public enum SessionReplay {
                 description: "Datadog SDK must be initialized before calling `SessionReplay.enable(with:)`."
             )
         }
+
+        guard !CoreRegistry.isFeatureEnabled(feature: SessionReplayFeature.self) else {
+            core.telemetry.debug("Session Replay has already been enabled")
```
ncreated, line 74: question/ What is the rationale of this telemetry? It reports user error (wrong use of the SDK), hence is not actionable on our side and doesn't say anything about the SDK functioning. In other words: what could we do if seeing these logs 🙂?

mariedm: I believe the intent behind this telemetry is to understand if customers are trying to enable Session Replay multiple times, which could indicate interest in having support for multiple cores with this feature.

@maciejburda: was there any other reasoning behind this telemetry?

PR #2082, opened 2024-10-11
RUM-6354 fix: Prioritize SwiftUI RUM actions over UIKit actions

ncreated:
### What and why?

📦 This PR resolves a conflict in RUM actions instrumentation that occurs when a user taps a SwiftUI button embedded inside a UIKit table view cell. This issue arises in apps that mix both UI frameworks, leading to inaccurate tracking of the tap action.

```
+--------UITableView----------+
|   +----UITableViewCell----+  |
|   |                       |  |
|   |  [  SwiftUI Button ]  |  |
|   |                       |  |
|   +-----------------------+  |       
+-------------------------------+
```

In this case, UIKit’s instrumentation [handled by `UIKitRUMActionsHandler`](https://github.com/DataDog/dd-sdk-ios/blob/d4e2247a6f55b95292e4b1b2953a25bdd8c67306/DatadogRUM/Sources/Instrumentation/Actions/UIKit/UIKitRUMUserActionsHandler.swift#L137) treats the `UITableViewCell` as the best target for the action, assigning its name to `action.target.name`. However, if the [`trackRUMAction()`](https://github.com/DataDog/dd-sdk-ios/blob/d4e2247a6f55b95292e4b1b2953a25bdd8c67306/DatadogRUM/Sources/Instrumentation/Actions/SwiftUI/SwiftUIActionModifier.swift#L36) modifier is used on the SwiftUI button, it already provides a more specific target name, which should take precedence.

### How?

This conflict is resolved by introducing `InstrumentationType` to the `RUMAddUserActionCommand`:
```swift
internal enum InstrumentationType: Int {
    case uikit
    case swiftui
    case manual

    var priority: Int { rawValue }
}
```
The priority field allows the `RUMViewScope` to determine whether to replace the existing `RUMUserActionScope` with one from a higher-priority instrumentation type. This replacement occurs if the new command is received within 100ms of the original, a window determined by the [existing action timeout](https://github.com/DataDog/dd-sdk-ios/blob/5b5105dd6d50af7e66ad14092d4fc101fc83be90/DatadogRUM/Sources/RUMMonitor/Scopes/RUMUserActionScope.swift#L12-L14).

Because UIKit instrumentation is triggered first, followed by SwiftUI gestures, this change ensures that the correct `action.target.name` is set. Similarly, if manual tracking (via `RUMMonitor` APIs) is mixed with UIKit or SwiftUI instrumentation, manual actions will always take precedence due to their explicit nature.

**Before:**
Screenshot 2024-10-11 at 15 38 55

**After:**
Screenshot 2024-10-11 at 15 39 00

### Review checklist

### DIFF
--- a/DatadogRUM/Sources/Instrumentation/RUMInstrumentation.swift
+++ b/DatadogRUM/Sources/Instrumentation/RUMInstrumentation.swift
@@ -27,9 +27,9 @@ internal final class RUMInstrumentation: RUMCommandPublisher {
     /// Swizzles `UIApplication` for intercepting `UIEvents` passed to the app.
     /// It is `nil` (no swizzling) if RUM Action instrumentaiton is not enabled.
     let uiApplicationSwizzler: UIApplicationSwizzler?
-    /// Receives interceptions from `UIApplicationSwizzler`.
-    /// It is `nil` if RUM Action instrumentaiton is not enabled.
-    let actionsHandler: UIEventHandler?
+    /// Receives interceptions from `UIApplicationSwizzler` and from SwiftUI instrumentation.
+    /// It is non-optional as we can't know if SwiftUI instrumentation will be used or not.
+    let actionsHandler: RUMActionsHandling
 
     /// Instruments RUM Long Tasks. It is `nil` if long tasks tracking is not enabled.
     let longTasks: LongTaskObserver?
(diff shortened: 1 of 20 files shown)


DatadogRUM/Sources/Instrumentation/RUMInstrumentation.swift
```diff
@@ -61,41 +61,41 @@ internal final class RUMInstrumentation: RUMCommandPublisher {
...
+                consolePrint(
+                    "🔥 Datadog SDK error: UIKit RUM Views tracking can't be enabled due to error: \(error)",
+                    .error
+                )
+            }
+            return nil
+        }()
+
+        // Always create actions handler (we can't know if it will be used by SwiftUI instrumentation)
+        // and only swizzle `UIApplicationSwizzler` if UIKit instrumentation is coUIKitRUMActionsHandlingUIKitRUMActionsHandlingnfigured:
```
maxep, line 79:
```suggestion
        // and only swizzle `UIApplicationSwizzler` if UIKit instrumentation is configured:
```

maxep: Nice catch btw!

maxep (approved):
Makes sense and looks great 👍

maxep (approved):

PR #2084, opened 2024-10-18
RUM-6569 [SR][FGM] Touch Override logic

mariedm:
### What and why?

This PR builds on the work of #2058 and #2067, which introduced Fine-Grained Masking (FGM) overrides (see the [RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3901129961/RFC+-+Session+Replay+Fine+Grained+Masking+Overrides) [internal] for more details). It implements the logic for applying touch privacy overrides at the view level, allowing views to have specific privacy settings that take precedence over global configurations.

### How?

- The global touch privacy level is now passed down to the `WindowTouchSnapshotProducer`.
- When a touch is detected (in `notify_sendEvent`), the applicable privacy setting is resolved, considering both view-level overrides and the global setting.
- The `overrideForTouch` dictionary is used to track privacy overrides for each ongoing touch event, ensuring the resolved privacy setting is reused throughout the lifecycle of the touch (from began to ended).
- Touch events are recorded or ignored based on the resolved privacy setting, with the view-level override taking precedence over the global configuration.

---

**🧪 Note**: A **flaky test** in `SessionReplayOverrideTests` was also identified and fixed in this PR.

**👩‍🎨 Design Consideration**: An alternative solution considered was to store the `Recorder.Context` within the `WindowTouchSnapshotProducer`, injected during initialization, and updated from the recorder when necessary. This stored context would then be used in both `takeSnapshot` and `notify_sendEvent`. However, this approach introduced complexity in managing the context state inside the snapshot producer, with potential issues around timing, synchronization, and the risk of stale contexts being used.

Another alternative was to pass the entire recording context directly into `notify_sendEvent`, but this was not ideal since touch event processing only requires the global touch privacy, not the full recording context.

The design decision was made to pass the global touch privacy during initialization because it is static and does not change during the session, meaning it only needs to be provided once. In contrast, the recording context can change and is therefore passed to `takeSnapshot` as needed, ensuring the correct context is used at the time of snapshot capture.

**📈 Performance Impact Note**: Running the Shopist application [internal] on an iPhone 15, I measured the performance impact of `notify_sendEvent` of my feature branch (mariedm/FGM-overrides) compared to the current branch. The results are as follows:
- Current branch: 52 μs on average
- Feature branch: 72.92 μs on average

While this indicates a **40% slowdown**, the absolute difference is just **20 μs**, which is unlikely to be noticeable in most real-world scenarios. In most applications, an additional 20 μs per touch event is unlikely to affect performance in a meaningful way.

### Review checklist


DatadogSessionReplay/Sources/Recorder/Recorder.swift
```diff
@@ -117,7 +119,7 @@ public class Recorder: Recording {
             return
         }
 
-        let touchSnapshot = recorderContext.touchPrivacy == .show ? touchSnapshotProducer.takeSnapshot(context: recorderContext) : nil
+        let touchSnapshot = touchSnapshotProducer.takeSnapshot(context: recorderContext)
```
mariedm, line 122: We don't check for the global touch privacy here anymore, it's now done in `WindowTouchSnapshotProducer`.

DatadogSessionReplay/Tests/SessionReplayOverrideTests.swift
```diff
@@ -160,6 +160,9 @@ class SessionReplayOverridesTests: XCTestCase {
         // Given
         let childOverrides: PrivacyOverrides = .mockAny()
         let parentOverrides: PrivacyOverrides = .mockRandom()
+        /// We explicitly set `hide` to `true` in the parent override because the child’s `hide` is `nil`.
+        /// In the merge logic, `true` takes precedence, and `false` behaves the same as `nil`, meaning no override.
+        parentOverrides.hide = true
```
mariedm, line 165: Flaky test, unrelated to this change

DatadogSessionReplay/Sources/Recorder/TouchSnapshotProducer/WindowTouchSnapshotProducer.swift
```diff
@@ -14,12 +14,20 @@ internal class WindowTouchSnapshotProducer: TouchSnapshotProducer, UIEventHandle
...
-    init(windowObserver: AppWindowObserver) {
+    init(
+        windowObserver: AppWindowObserver,
+        globalTouchPrivacy: TouchPrivacyLevel
+    ) {
         self.windowObserver = windowObserver
+        self.globalTouchPrivacy = globalTouchPrivacy
     }
 
     func takeSnapshot(context: Recorder.Context) -> TouchSnapshot? {
```
maxep, line 33: /suggestion If we want to follow the design in place, we should instead add the `globalTouchPrivacy` as a property of the `Recorder.Context`. The override value could be added to the `TouchSnapshot.Touch` when receiving touches. Then the `takeSnapshot` method will be about getting the first touch that satisfies the privacy level. Would that make sense?

maxep (commented):
I have left a suggestion, LMKWYT!

DatadogSessionReplay/Sources/Recorder/TouchSnapshotProducer/WindowTouchSnapshotProducer.swift
mariedm: Yes, it makes sense! I wish I had thought of that. And actually, the `globalTouchPrivacy` is already a part of [Recorder.Context](https://github.com/DataDog/dd-sdk-ios/blob/2d62ca880be7cca0cb03a30c13d95aad37744cf4/DatadogSessionReplay/Sources/Recorder/Recorder.swift#L30).

mariedm:
**Logic Update**

- **Touch Privacy Level Handling**: Updated `WindowTouchSnapshotProducer` to use the global touch privacy from the recorder context during `takeSnapshot`, allowing for dynamic context-based privacy decisions. Previously, it was passed as a global value in the initializer.
- **Touch Override Logic**: Introduced an optional `touchOverride` property to `TouchSnapshot.Touch` to allow privacy overrides to be resolved only when a snapshot is taken, in `takeSnapshot`.
- **Override Cache Cleanup**: Added a mechanism to remove stale entries from the `overrideForTouch` cache if touches do not properly reach an end state.

maxep (approved):
Nice!!

mariedm:
Created a [ticket](https://datadoghq.atlassian.net/browse/RUM-6862) [internal] to follow-up on the cache cleanup logic.

PR #2091, opened 2024-10-24
Fix Flaky Test - Remove force unwrapping in URL response validation

maxep:
### What and why?

Some URL session tests depend on `https://www.datadoghq.com` which could fail to load (somehow). But our test are expecting data by force unwrapping the response.

### How?

Loose the expectation by comparing the URL response with the response that was instrumented, regardless of its data.

### Review checklist


mariedm (approved):
LGMT!

PR #2092, opened 2024-10-24
RUM-917 Fix type mistmatch error

maciejburda:
### What and why?

Fixes the issues of type mismatch error which was causing full snapshot enforcement in some cases.

### How?

We've been wrongly passing identifier of the parent recorder to it's children which was causing in some cases issue where two views would be recorded with different type but the same id. It's because `NodeIDGenerator` depends on the identifier of the recorder.

### Review checklist

### DIFF
--- a/DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UIDatePickerRecorder.swift
+++ b/DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UIDatePickerRecorder.swift
@@ -16,9 +16,9 @@ internal struct UIDatePickerRecorder: NodeRecorder {
 
     init(identifier: UUID) {
         self.identifier = identifier
-        self.wheelsStyleRecorder = WheelsStyleDatePickerRecorder(identifier: identifier)
-        self.compactStyleRecorder = CompactStyleDatePickerRecorder(identifier: identifier)
-        self.inlineStyleRecorder = InlineStyleDatePickerRecorder(identifier: identifier)
+        self.wheelsStyleRecorder = WheelsStyleDatePickerRecorder(identifier: UUID())
+        self.compactStyleRecorder = CompactStyleDatePickerRecorder(identifier: UUID())
+        self.inlineStyleRecorder = InlineStyleDatePickerRecorder(identifier: UUID())
     }
 
     func semantics(of view: UIView, with attributes: ViewAttributes, in context: ViewTreeRecordingContext) -> NodeSemantics? {
@@ -104,9 +104,9 @@ private struct InlineStyleDatePickerRecorder {
     let subtreeRecorder: ViewTreeRecorder
 
     init(identifier: UUID) {
-        self.viewRecorder = UIViewRecorder(identifier: identifier)
+        self.viewRecorder = UIViewRecorder(identifier: UUID())
         self.labelRecorder = UILabelRecorder(
-            identifier: identifier,
+            identifier: UUID(),
             textObfuscator: { context, viewAttributes in
                 return viewAttributes.resolveTextAndInputPrivacyLevel(in: context).staticTextObfuscator
             }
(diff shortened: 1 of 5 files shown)


maxep:
Seems like it's reverting #1967, wouldn't this introduce a regression?

maxep:
Also, I think we should target the release branch, WDYT?

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UIDatePickerRecorder.swift
```diff
@@ -16,9 +16,9 @@ internal struct UIDatePickerRecorder: NodeRecorder {
 
     init(identifier: UUID) {
         self.identifier = identifier
-        self.wheelsStyleRecorder = WheelsStyleDatePickerRecorder(identifier: identifier)
-        self.compactStyleRecorder = CompactStyleDatePickerRecorder(identifier: identifier)
-        self.inlineStyleRecorder = InlineStyleDatePickerRecorder(identifier: identifier)
+        self.wheelsStyleRecorder = WheelsStyleDatePickerRecorder(identifier: UUID())
+        self.compactStyleRecorder = CompactStyleDatePickerRecorder(identifier: UUID())
+        self.inlineStyleRecorder = InlineStyleDatePickerRecorder(identifier: UUID())
```
maxep, line 21: /suggestion We could make an init with default value instead:
```swift
init(identifier: UUID = UUID()) { }
```
---
It is still unclear when to create a recorder with new uuid and when to inject one. If I understand correctly, the fix in #1967 introduced another issue, the actual fix is in this PR in `UITabBarRecorder.swift`, right?

I guess it might be best to revert changes in #1967 to prevent any mis-usage in the future: expecting an `identifier` parameter could be error-prone, WDYT?

maciejburda: The bottom line strategy is: recorders should have a stable and unique id. We should avoid creating temporary recorders inline inside of function.

Both changes are relevant. IMO it's better to keep explicit id (it makes you stop for a second).

I'll add a documentation within Recorder.

DatadogSessionReplay/Sources/Processor/SnapshotProcessor.swift
```diff
@@ -96,8 +96,7 @@ internal class SnapshotProcessor: SnapshotProcessing {
         var records: [SRRecord] = []
         // Create records for describing UI:
         if viewTreeSnapshot.context.applicationID != lastSnapshot?.context.applicationID ||
-            viewTreeSnapshot.context.sessionID != lastSnapshot?.context.sessionID ||
-            viewTreeSnapshot.context.viewID != lastSnapshot?.context.viewID {
```
maxep, line 100: /question Is this also part of the fix?

DatadogSessionReplay/Sources/Processor/SnapshotProcessor.swift
```diff
@@ -119,7 +118,7 @@ internal class SnapshotProcessor: SnapshotProcessing {
                 .flatMap { records.append($0) }
             }
         } else {
-            telemetry.error("[SR] Unexpected flow in `Processor`: no previous wireframes and no previous RUM context")
```
maxep, line 122: /question We have been incorrectly loggin errors here?

maxep (dismissed):

maxep:
I have added a couple of question to make sure before merging!

DatadogSessionReplay/Sources/Processor/SnapshotProcessor.swift
maciejburda: I reverted that change. Will be tackled in a separate PR.

DatadogSessionReplay/Sources/Processor/SnapshotProcessor.swift
maciejburda: Reverted. Will log a separate ticket for the double full snapshot issue.

mariedm (approved):

PR #2093, opened 2024-10-25
chore: Cleanup local simulators on CI

maxep:
### What and why?

CI builds started to fail with the following error on random tests:
```
🔥 Failed to delete `TestsDirectory`: Error Domain=NSCocoaErrorDomain Code=512 ""com.datadoghq.ios-sdk-tests-434C4A49-7195-4957-AE16-9A943737BDBF" couldn't be removed."
```
From [this thread](https://developer.apple.com/forums/thread/128927?answerId=631839022#631839022), the error code `512` could mean that we have reached the maximum number of items in cache.

### How?

Add a step to clean up simulators before running the pipeline. After running this, builds are passing again.

### Review checklist

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -70,15 +70,15 @@ GEM
     ethon (0.16.0)
       ffi (>= 1.15.0)
     ffi (1.17.0)
-    ffi (1.17.0-aarch64-linux-gnu)
+    ffi (1.17.0-aarch64-linux)
     ffi (1.17.0-aarch64-linux-musl)
-    ffi (1.17.0-arm-linux-gnu)
+    ffi (1.17.0-arm-linux)
     ffi (1.17.0-arm-linux-musl)
     ffi (1.17.0-arm64-darwin)
-    ffi (1.17.0-x86-linux-gnu)
+    ffi (1.17.0-x86-linux)
     ffi (1.17.0-x86-linux-musl)
     ffi (1.17.0-x86_64-darwin)
-    ffi (1.17.0-x86_64-linux-gnu)
+    ffi (1.17.0-x86_64-linux)
     ffi (1.17.0-x86_64-linux-musl)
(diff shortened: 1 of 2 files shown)


maciejburda (dismissed):

Gemfile.lock
mariedm, line 1: Is this change wanted?

maxep: It's unrelated but looks like gems are out of sync, so I committed it. If you run `make dependencies` locally it should give the same result, can you confirm?

mariedm (approved):

tools/clean.sh
```diff
@@ -25,3 +28,13 @@ clean_dir ./IntegrationTests/Pods
 echo_warn "Cleaning local xcconfigs"
 rm -vf ./xcconfigs/Base.ci.local.xcconfig
 rm -vf ./xcconfigs/Base.dev.local.xcconfig
+
+ENV_DEV="dev"
+ENV_CI="ci"
```
ncreated, line 33: Doesn't seem used 🧽
```suggestion
```

tools/clean.sh
```diff
@@ -25,3 +28,13 @@ clean_dir ./IntegrationTests/Pods
...
+
+if [[ "$ENV" == "ci" ]]; then
```
ncreated, line 35: nit/ This works rather by coincidence than explicit check, because this param is made for `repo-setup` command in `make clean repo-setup ENV=ci`.

More explicit is to check `$CI` env which is default for GitLab runner:
```suggestion
if [ "$CI" = "true" ]; then
```

ncreated (approved):
Thanks! 🚀

mariedm (approved):

ncreated (approved):

maxep:
Closing in favour of #2096

PR #2094, opened 2024-10-28
RUM-6698 refactor: Enable usage of real DatadogCore initializer in integration tests

ncreated:
### What and why?

📦 This PR is a refactoring only, with no new test coverage or bug fixes included. It improves our ability to perform comprehensive integration testing for SDK behavior **during the application launch sequence**.

The main goal of this and upcoming changes is to enhance test coverage for important SDK processes, like [starting RUM sessions](https://github.com/DataDog/dd-sdk-ios/blob/develop/Datadog/IntegrationUnitTests/RUM/StartingRUMSessionTests.swift), which have been found to work under incorrect assumptions, hiding bugs such as `RUM-6698`.

Previously, to simulate different app launch environments, we used context mocking and the internal core's `init()`:
```swift
core = DatadogCore(
   directory: ...,
   dateProvider: ...,
   initialConsent: ...,
   performance: ...,
   httpClient: ...,
   encryption: ...,
   contextProvider: DatadogContextProvider.mockWith(
      applicationStateHistory: .mockAppInBackground(since: dateInPast)
   ),
   applicationVersion: ...,
   maxBatchesPerUpload: ...,
   backgroundTasksEnabled: ...
)
```

As a result, our integration tests coverage is missing the actual code of setting up the `DatadogContextProvider`, leading to overlooked issues with resolving the application state, such as:
- [This incorrect assumption](https://github.com/DataDog/dd-sdk-ios/blob/b7ebb2c051b2495fd04868a9ec9a9d97deb0e493/DatadogCore/Sources/Core/DatadogCore.swift#L424-L427) that the process always starts in `.active` (which it never does).
- [This delay](https://github.com/DataDog/dd-sdk-ios/blob/b7ebb2c051b2495fd04868a9ec9a9d97deb0e493/DatadogCore/Sources/Core/DatadogCore.swift#L450-L454) that causes the first events to sometimes miss the actual application state.
- The wrong order of sending lifecycle notifications to the core and RUM instrumentation that causes some events consider wrong application state.

These problems will be addressed in `RUM-6698` in future PRs. We expect these changes may have some impact, so I will work in a feature branch and merge only after the upcoming SR releases to isolate changes.

### How?

This PR separates `NotificationCenter` and `AppStateProvider` (a new type) so that both can be injected into the core. It also moves the core `init()` method to `DatadogCore`, making it possible to use in tests:
```swift
var config: Datadog.Configuration = .mockAny()
config.appStateProvider = AppStateProviderMock()
config.notificationCenter = NotificationCenter()
config.dateProvider = DateProviderMock()

core = try DatadogCore(
   configuration: config, trackingConsent: .granted, instanceName: "test"
)
```
As before, the `core` can be wrapped in `DatadogCoreProxy` to observe events.

Both `config.notificationCenter` and `config.appStateProvider` will be used in upcoming PRs to simulate the actual process environment for testing "RUM session start" and "RUM session duration."

### Review checklist


DatadogCore/Sources/Core/DatadogCore.swift
```diff
@@ -71,6 +71,93 @@ internal final class DatadogCore {
     /// Maximum number of batches per upload.
     internal let maxBatchesPerUpload: Int
 
+    convenience init(
+        configuration: Datadog.Configuration,
+        trackingConsent: TrackingConsent,
+        instanceName: String
+    ) throws {
```
ncreated, line 78: 💬 This initializer is extracted from what was originally in `Datadog.swift` (behind the static facade). No other code changes were done.

maxep: May I ask why?

The problem now is that `DatadogCore` depends on `Datadog.*` data types (if we take the clean architecture, this breaks the dependency rule). I would rather have a `DatadogCore` extension defined at the upper level of the module than make `DatadogCore` depend on `Datadog`. If that make sense?

ncreated: ✅ done

DatadogCore/Sources/Core/Context/BatteryStatusPublisher.swift
```diff
@@ -24,11 +24,11 @@ internal final class BatteryStatusPublisher: ContextValuePublisher {
...
     /// 
     /// - Parameters:
-    ///   - device: The `UIDevice` instance. `.current` by default.
     ///   - notificationCenter: The notification center for observing the `UIDevice` battery changes,
+    ///   - device: The `UIDevice` instance. `.current` by default.
     init(
-        device: UIDevice = .current,
-        notificationCenter: NotificationCenter = .default
+        notificationCenter: NotificationCenter,
+        device: UIDevice = .current
```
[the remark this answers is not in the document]
ncreated: I was just lazy, but I now fixed it ✅ ⛺ 🧽 

However, there are other places where default values are provided on init basis rather than passed as dependency injection. This makes integration testing impossible. I wish we pay more attention to this in general.

DatadogInternal/Sources/Context/AppStateProvider.swift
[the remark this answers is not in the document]
ncreated: Good point, moved ✅

DatadogInternal/Sources/Context/AppState.swift
mariedm, line 1: Thanks, it's easier to understand now!

DatadogInternal/Sources/Context/AppState.swift
```diff
@@ -142,31 +151,66 @@ extension AppStateHistory {
...
+        @unknown default: self = .active // in case a new state is introduced, default to most expected state
+        }
+    }
+}
+
+#else // macOS (no UIKit and no WatchKit)
+
+public struct DefaultAppStateProvider: AppStateProvider {
+    public init() {}
+    public let current: AppState = .active
```
mariedm, line 213: I'm wondering if it might better to use `.background` as the fallback state of `.active`. This could potentially avoid false positives, which might contributed to issues like the 1ns sessions. What happens in the case of cold starts and pre-warmings? Is the state correctly initialized?

mariedm (approved):
Great job on the refactoring! 💪

I have a general question. This refactor got me thinking about our issue around the 1ns sessions. Is it possible that we miss events in case of rapid app state changes, e.g., minimizing and reopening the app quickly, or when receiving multiple notifications at once that aren't processed quickly enough? This might be something worth testing in future PRs.

DatadogCore/Sources/Core/Context/BatteryStatusPublisher.swift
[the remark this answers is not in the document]
maxep: No problem, I was just curious :)

But now I'm wondering why providing default values make integration testing impossible? 🤔 
I understand that removing the defaults enforces dependency injection, and it's probably a better pattern to follow. Maybe we should add this as a rule somewhere, wdyt?

maxep (approved):
🚀

DatadogCore/Sources/Core/Context/BatteryStatusPublisher.swift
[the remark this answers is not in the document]
ncreated: > But now I'm wondering why providing default values make integration testing impossible? 🤔

Sorry, wrong wording 🙂. It is not "impossible" as long as the actual value is injected from upstream no matter the default value. However, this was not the case for all dependencies refined in this PR - I had to make an extra effort to first find these dependencies and then to inject them from the root configuration. Otherwise, my integration tests would not work with assumed environment.

> Maybe we should add this as a rule somewhere, wdyt?

This is good idea 👍. I created [RUM-6977](https://datadoghq.atlassian.net/browse/RUM-6977) to look into possibility of adding linter rule for this.

DatadogInternal/Sources/Context/AppState.swift
ncreated: Please mind that this default is only for macOS where we don't offer official support. Because we don't really test SDK for macOS and knowing that we do not subscribe to `AppKit's` lifecycle notifications anywhere in the SDK, always assuming `.active` may seem safer. In macOS apps, this state will very likely become the only one that is assumed (which is fine for non-official support and no RUM offering for macOS).

PR #2103, opened 2024-11-06
chore: Enable CI Tests visibility back

ncreated:
### What and why?

📦 We temporarily disabled CI Tests Visibility in https://github.com/DataDog/dd-sdk-ios/pull/2096. This PR enables it back.

### How?

Before merge, I'll run unit tests several times to see if the problem is gone.

### Review checklist

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -70,7 +70,7 @@ default:
       when: delayed
       start_in: 40 minutes
 
-ENV check:
+.ENV check:
   stage: pre
   rules: 
     - !reference [.test-pipeline-job, rules]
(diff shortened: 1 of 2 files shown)


mariedm:
Did we change anything about it, or are we just trying again to enable it and see how it goes?

ncreated:
> Did we change anything about it, or are we just trying again to enable it and see how it goes?

@mariedm It started causing problems in our tests somewhen around `2.5.2` release of `dd-sdk-swift-testing`. I change it now to pin the exact `2.5.1` version of this dependency (vs `2.5.*` before) and it seems to work. I'm going to chat with @ypopovych, but it seems we can stay on `2.5.1` until we solve the problem.

ncreated:
Closing, there will be separate attempt.

PR #2111, opened 2024-11-15
RUM-6867 Implement id based deterministic sampling

xgouchet:
### What and why?

Implement an id based deterministic sampling. Next PR will make use of it for our tracing decisiong, allowing a more consistent APM implementation with other platform agents

### DIFF
--- a/DatadogInternal/Sources/Utils/Sampler.swift
+++ b/DatadogInternal/Sources/Utils/Sampler.swift
@@ -33,6 +33,12 @@ public struct Sampler: Sampling {
 
 /// A sampler that determines sampling decisions deterministically (the same each time).
 internal struct DeterministicSampler: Sampling {
+    enum Constants {
+        /// Good number for Knuth hashing (large, prime, fit in 64 bit long)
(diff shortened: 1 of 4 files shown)


maxep (dismissed):
Nice 👌

DatadogInternal/Sources/Utils/Sampler.swift
```diff
@@ -43,5 +47,13 @@ internal struct DeterministicSampler: Sampling {
         self.shouldSample = shouldSample
     }
 
+    init(baseId: UInt64, samplingRate: Float) {
```
ncreated, line 50: question/ Is `baseId` a `TraceID`? If so, we could make it accept [`TraceID`](https://github.com/DataDog/dd-sdk-ios/blob/52842a99d3eced599fc59185f0c7c166aa00cdf1/DatadogInternal/Sources/NetworkInstrumentation/TraceID.swift#L9) to avoid ambiguity with accepting `UInt64` type. In fact, our `TraceID` is not as simple as `UInt64`. It has `idHi` and `idLo` making it unclear what should be the value passed here.

ncreated (approved):

PR #2112, opened 2024-11-15
RUM-7285 Add api-surface step to Lint stage

mariedm:
### What and why?

To ensure public API changes are properly updated, I previously added a reminder in our GitHub PR template checklist to run `make api-surface`. However, this has proven insufficient 🙃 To address this, I am adding an automated step to the CI Lint phase to enforce this check.

### How?

- Add CI Validation:
  - A step is added to the Lint phase that runs `make api-surface`.
  - The generated API surface files are written to temporary files (`api-surface-swift-generated` and `api-surface-objc-generated`).
  - These temporary files are compared with the committed reference files (`api-surface-swift` and `api-surface-objc`).
- CI Behavior:
  - If differences are detected, the CI step exits with an error, prompting contributors to run `make api-surface` locally and commit the updated reference files.
- Updated Reference Files:
  - The reference files are updated as needed to reflect recent changes to the public API.

### Review checklist

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -128,7 +128,7 @@ Lint:
 API Surface Verify:
   stage: lint
   rules:
-    - if: '$CI_COMMIT_BRANCH == $MAIN_BRANCH'
+    - if: '$CI_COMMIT_BRANCH =~ /^(release|hotfix)\/.*/'
   script:
     - make clean repo-setup ENV=ci
     - make api-surface-verify


.gitlab-ci.yml
```diff
@@ -91,6 +91,17 @@ Lint:
...
+    - if ! diff api-surface-swift api-surface-swift-generated; then
+        echo "❌ Swift API surface mismatch";
+        echo "👉 Run \`make api-surface\` locally to update \`api-surface-swift\` and commit the changes.";
+        exit 1;
+      fi
+    - if ! diff api-surface-objc api-surface-objc-generated; then
+        echo "❌ Objective-C API surface mismatch";
+        echo "👉 Run \`make api-surface\` locally to update \`api-surface-objc\` and commit the changes.";
+        exit 1;
+      fi
```
ncreated, line 104: suggestion/ Could we move these checks to separate script in `tools/` directory? The convention we started for [.gitlab-ci.yml](https://github.com/DataDog/dd-sdk-ios/blob/develop/.gitlab-ci.yml) is to avoid bare scripts in `yml` as they are very inconvenient to reproduce in local shell.

Ideally, we could get inspired by `make rum-models-verfy`:
* `make api-surface` - to generate API surface files; ran in local
* `make api-surface-verify` - to verify API surface; ran on CI

Makefile
```diff
@@ -356,6 +356,16 @@ sr-snapshot-tests-open:
...
+
+# Use different paths when running in CI
+ifeq ($(ENV),ci)
+  SWIFT_OUTPUT_PATH := api-surface-swift-generated
+  OBJC_OUTPUT_PATH := api-surface-objc-generated
+endif
+
 # Generate api-surface files for Datadog and DatadogObjc.
 api-surface:
 		@echo "Generating api-surface-swift"
```
ncreated, line 371: suggestion/ Let's follow the convention in of [printing command's title before it is ran](https://github.com/DataDog/dd-sdk-ios/blob/004ca93c6036c0ceead2f09fedebfae9496c9da3/Makefile#L313). This brings a lot of clarity to CI logs:
```
	@$(ECHO_TITLE) "make api-surface"
```

Also, let's fix the indentation for this command. It uses the old one, all new commands (after migration to GitLab) are indented with less whitespaces. See [these examples](https://github.com/DataDog/dd-sdk-ios/blob/004ca93c6036c0ceead2f09fedebfae9496c9da3/Makefile#L311-L328).

ncreated (approved):
This is great 🤩! I left two feedbacks on conventions, but well done overall 💪

mariedm:
**Update**

- Addressed comments: 
  - Removed inline scripts from `.gitlab-ci.yml` and moved logic to `tools/`
  - Followed command title printing style
  - Corrected `Makefile` indentation
- Split Commands:
  - `Generate` to create API surface files and `Verify` to compare with reference files
  - Added new tests for the new `Verify` command 
- Fixes:
  - Adjusted lint rule in `SessionReplayConfiguration` to fix api-surface file

tools/api-surface/Tests/APISurfaceCoreTests/IntegrationTests.swift
```diff
@@ -6,71 +6,155 @@
...
+import ArgumentParser
 
 class IntegrationTests: XCTestCase {
     func testCreatingSurfaceForFixtureLibraries() throws {
-        var outputs: [String] = []
-        APISurfaceCore.printFunction = { outputs.append($0) }
-
         // Given
-        let command = try SPMLibrarySurfaceCommand.parse([
+        let temporaryFile = "/tmp/api-surface-test-output"
```
ncreated, line 14: suggestion/ Using a dynamically created temp directory could bring some benefits:
- Avoid conflicts between test runs and ensure isolation.
- Automatically clean up after tests to prevent leftover files from affecting others.
- More robust (in other environments) than using hardcoded paths.

Example:
```swift
let tempDirectory = FileManager.default.temporaryDirectory.appendingPathComponent("com.datadoghq.api-surface-" + UUID().uuidString)
try FileManager.default.createDirectory(at: tempDirectory, withIntermediateDirectories: true, attributes: nil)
        
let temporaryFile = tempDirectory.appendingPathComponent("api-surface-test-output").path
```

ncreated (approved):
Well done 🏅💪!

I did notice however, that integrating `api-surface` has increased the `lint` job duration from 30 seconds to around **3 minutes**. This is a significant change, so before merging, it might be worth discussing with the team:

- How do we feel about this increased duration?
- How might this impact development or the feedback loop?
- Are there any optimization opportunities?
- When should we consider optimizing?

simaoseica-dd (dismissed):
🏅
This will be very helpful

tools/api-surface/Tests/APISurfaceCoreTests/IntegrationTests.swift
mariedm: Good idea! I've applied your suggestion.

mariedm:
I'm reviving this PR after its priority was bumped; here are the new changes:
- rebased, which includes `v3` release, where each ObjC API has been moved to their feature module (and ObjC module has been removed)
- added a test for ObjC use case (and a fixture)
- updated some dependencies, including `SourceKitten`
- updated tests to use a temporary directory dynamically created for file generation

As discussed, because this step is adding significant time to CI (~3min), we decided to only run it on commits to the `master` branch. This will allow us, when preparing a release, to update the listing of APIs if needed. 

Additionally, I created a [Jira ticket](https://datadoghq.atlassian.net/browse/RUM-11641) listing some potential optimizations. This might allow us to run this step against the `develop` branch as well in future iterations.

maxep (approved):
LGTM 👍

.gitlab-ci.yml
```diff
@@ -125,6 +125,14 @@ Lint:
     - make lint license-check
     - make rum-models-verify sr-models-verify
 
+API Surface Verify:
+  stage: lint
+  rules:
+    - if: '$CI_COMMIT_BRANCH == $MAIN_BRANCH'
```
ncreated: This means it will run only after `(release|hotfix)/*` gets merged to the main branch, which is too late in the process, no? I think it would be better to run it for commits on `release/*` and `hotfix/*` branches - that would give us a chance to align APIs before release is done.

PR #2114, opened 2024-11-19
RUM-7251 Static trace context caching

maxep:
### What and why?

The `URLSessionInterceptor` keeps a cache of trace context but the `shared()` methods returns a new instance each time, leading to trace discontinuity.

### How?

Make the trace-context cache static.

### Review checklist

### DIFF
--- a/DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionInterceptor.swift
+++ b/DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionInterceptor.swift
@@ -34,7 +34,7 @@ public struct URLSessionInterceptor {
     /// This is to bridge the gap between what is encoded into HTTP headers and what is later needed for processing 
     /// the interception (unlike request headers, the `TraceContext` holds the original information on trace sampling).
     @ReadWriteLock
-    private var contextsByTraceID: [TraceID: [TraceContext]] = [:]
+    static var contextsByTraceID: [TraceID: [TraceContext]] = [:]
 
     /// Tells the interceptor to modify a URL request.
     ///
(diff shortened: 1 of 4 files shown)


DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionInterceptor.swift
```diff
@@ -34,7 +34,7 @@ public struct URLSessionInterceptor {
     /// This is to bridge the gap between what is encoded into HTTP headers and what is later needed for processing 
     /// the interception (unlike request headers, the `TraceContext` holds the original information on trace sampling).
     @ReadWriteLock
-    private var contextsByTraceID: [TraceID: [TraceContext]] = [:]
```
maxep, line 37: @ncreated, That's the most straightforward option. Another option would be to make this dict a property of the `NetworkInstrumentationFeature` but it's out of its responsibility...

ncreated: I think this fix is good enough given that the entire `URLSessionInterceptor` is only used by the deprecated Alamofire instrumentation code. There is no other code that depends on this type, so we can keep it simple 👍.

>  Another option would be to make this dict a property of the `NetworkInstrumentationFeature` but it's out of its responsibility...

Or another option would be to fix the `shared(in:)` method, so it really implements singleton behavior. It should return a single instance of interceptor for given `core`. That would be the only change required in this fix, but it will require introducing a new concept of attaching arbitrary objects to `core`. Not worth it because `URLSessionInterceptor` is already deprecated.

ncreated (approved):

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionInterceptor.swift
maxep: Is it deprecated? `URLSessionInterceptor` is still the entry-point for manual URLSession instrumentation, it is used by the deprecated Alamofire extension but its definition is public and not marked as deprecated.

PR #2116, opened 2024-11-20
Make Crash Reporter Plugin Public

naftaly:
This PR makes the crash reporter public and adds some hooks to send reports at other times than startup.

---

This pull request includes several changes to the `DatadogCrashReporting` module, focusing on enhancing crash reporting functionality and improving compatibility with conditional imports. The most important changes include adding a new method to merge RUM attributes in `CrashContext`, updating the `CrashReporting` class to handle conditional imports, and making the `CrashReportingFeature` class and `CrashReportingPlugin` protocol public.

Enhancements to crash reporting functionality:

* [`DatadogCrashReporting/Sources/CrashContext/CrashContext.swift`](diffhunk://#diff-b3539333f9d72967f5061090c4d4ba9a73edadb9334e088f8815fd032199e6f0R168-R195): Added a new method `merging(rumAttributes:)` to the `CrashContext` struct for merging RUM attributes.
* [`DatadogCrashReporting/Sources/CrashReporting.swift`](diffhunk://#diff-c32640359b93857c4d85ff08e2832d1759d5dcd294491d5dedd722ef828d69aaR59-R62): Introduced a new method `send(_:attributes:in:)` to the `CrashReporting` class for sending crash reports with optional attributes.

Improvements for conditional imports:

* [`DatadogCrashReporting/Sources/CrashReporting.swift`](diffhunk://#diff-c32640359b93857c4d85ff08e2832d1759d5dcd294491d5dedd722ef828d69aaR24-R31): Updated the `enable(in:)` method to conditionally compile code based on the availability of the `CrashReporter` module. [[1]](diffhunk://#diff-c32640359b93857c4d85ff08e2832d1759d5dcd294491d5dedd722ef828d69aaR24-R31) [[2]](diffhunk://#diff-c32640359b93857c4d85ff08e2832d1759d5dcd294491d5dedd722ef828d69aaR45-R49)
* Various files in `DatadogCrashReporting/Sources/PLCrashReporterIntegration/`: Added `#if canImport(CrashReporter)` checks to conditionally compile code depending on the availability of the `CrashReporter` module. [[1]](diffhunk://#diff-bd97abaa913473bd362bf4e20b8c560f763122c1361e22e00b3d1a7ab929e07aR7-R8) [[2]](diffhunk://#diff-bd97abaa913473bd362bf4e20b8c560f763122c1361e22e00b3d1a7ab929e07aR323-R324) [[3]](diffhunk://#diff-47b3f6e87bca1bae23fc704f20a51f8dbdd18171cbeefb8061c545c4fb657654R7-R8) [[4]](diffhunk://#diff-47b3f6e87bca1bae23fc704f20a51f8dbdd18171cbeefb8061c545c4fb657654R97-R98) [[5]](diffhunk://#diff-f18159b2ed56410cd88af22dfe33406cc3ab08672b9e37c15f4cddc1be5cbbe0R7-R8) [[6]](diffhunk://#diff-f18159b2ed56410cd88af22dfe33406cc3ab08672b9e37c15f4cddc1be5cbbe0R28-R29) [[7]](diffhunk://#diff-f4aef31e1317d349b97407fcca5ced4ca39332260e6cff23c57fad3e4aaa851aR7-R8) [[8]](diffhunk://#diff-f4aef31e1317d349b97407fcca5ced4ca39332260e6cff23c57fad3e4aaa851aR228-R229) [[9]](diffhunk://#diff-ae2462de437de202b1a7d4fcc19bb98534b68ecbf312cd81540684a36a8eae78R7-R8) [[10]](diffhunk://#diff-ae2462de437de202b1a7d4fcc19bb98534b68ecbf312cd81540684a36a8eae78R87-R88) [[11]](diffhunk://#diff-8cb7517e08ef120d5f94d9e067ff28361956fbda8f941676c29243b6d0b53510R7-R8) [[12]](diffhunk://#diff-8cb7517e08ef120d5f94d9e067ff28361956fbda8f941676c29243b6d0b53510R64-R65)

Public API changes:

* [`DatadogCrashReporting/Sources/CrashReportingFeature.swift`](diffhunk://#diff-552b6319a97696000b5d646e20f1a0d2f2c68aa9f422e199007e77843b4e8a6aL10-R13): Changed the `CrashReportingFeature` class and its `name` and `messageReceiver` properties to be public. [[1]](diffhunk://#diff-552b6319a97696000b5d646e20f1a0d2f2c68aa9f422e199007e77843b4e8a6aL10-R13) [[2]](diffhunk://#diff-552b6319a97696000b5d646e20f1a0d2f2c68aa9f422e199007e77843b4e8a6aR57-R66) [[3]](diffhunk://#diff-552b6319a97696000b5d646e20f1a0d2f2c68aa9f422e199007e77843b4e8a6aL147-R157)
* [`DatadogCrashReporting/Sources/CrashReportingPlugin.swift`](diffhunk://#diff-f0a07cff9ca12bd4e3b8c439ec2c30628213fb34a2de6086eccda055296e75e9L13-R13): Made the `CrashReportingPlugin` protocol public and added a new `NoopCrashReportingPlugin` class. [[1]](diffhunk://#diff-f0a07cff9ca12bd4e3b8c439ec2c30628213fb34a2de6086eccda055296e75e9L13-R13) [[2]](diffhunk://#diff-f0a07cff9ca12bd4e3b8c439ec2c30628213fb34a2de6086eccda055296e75e9R30-R36)

### DIFF
--- a/DatadogCrashReporting/Sources/CrashContext/CrashContext.swift
+++ b/DatadogCrashReporting/Sources/CrashContext/CrashContext.swift
@@ -165,4 +165,32 @@ internal struct CrashContext: Codable, Equatable {
             lhs.userInfo?.email == rhs.userInfo?.email &&
             lhs.appLaunchDate == rhs.appLaunchDate
     }
+
+    func merging(rumAttributes: GlobalRUMAttributes) -> Self {
+        let sourceAttribtues = lastRUMAttributes?.attributes ?? [:]
+        let mergedAttributes = sourceAttribtues.merging(rumAttributes.attributes) { new, _ in new }
+
+        return CrashContext(
+            serverTimeOffset: serverTimeOffset,
+            service: service,
+            env: env,
+            version: version,
+            buildNumber: buildNumber,
+            device: device,
+            sdkVersion: sdkVersion,
+            source: source,
+            trackingConsent: trackingConsent,
+            userInfo: userInfo,
+            networkConnectionInfo: networkConnectionInfo,
+            carrierInfo: carrierInfo,
+            lastIsAppInForeground: lastIsAppInForeground,
+            appLaunchDate: appLaunchDate,
+            lastRUMViewEvent: lastRUMViewEvent,
+            lastRUMSessionState: lastRUMSessionState,
+            lastRUMAttributes: GlobalRUMAttributes(
+                attributes: mergedAttributes
+            ),
+            lastLogAttributes: lastLogAttributes
+        )
+    }
 }
--- a/DatadogCrashReporting/Sources/CrashReporting.swift
+++ b/DatadogCrashReporting/Sources/CrashReporting.swift
@@ -21,10 +21,14 @@ import DatadogInternal
 public final class CrashReporting {
     /// Initializes the Datadog Crash Reporter.
     public static func enable(in core: DatadogCoreProtocol = CoreRegistry.default) {
+#if canImport(CrashReporter)
         enable(with: PLCrashReporterPlugin(), in: core)
+#else
+        assertionFailure()
+#endif
     }
 
-    internal static func enable(with plugin: CrashReportingPlugin, in core: DatadogCoreProtocol) {
+    public static func enable(with plugin: CrashReportingPlugin, in core: DatadogCoreProtocol = CoreRegistry.default) {
         do {
             let contextProvider = CrashContextCoreProvider()
 
@@ -38,9 +42,11 @@ public final class CrashReporting {
 
             try core.register(feature: reporter)
 
+#if canImport(CrashReporter)
             if let plcr = PLCrashReporterPlugin.thirdPartyCrashReporter {
                 try core.register(backtraceReporter: BacktraceReporter(reporter: plcr))
             }
+#endif
 
             reporter.sendCrashReportIfFound()
 
@@ -49,6 +55,10 @@ public final class CrashReporting {
         } catch {
             consolePrint("\(error)", .error)
         }
+    }
+
+    public static func send(_ report: DDCrashReport, attributes: [AttributeKey: AttributeValue]? = nil, in core: DatadogCoreProtocol = CoreRegistry.default) {
+        core.get(feature: CrashReportingFeature.self)?.send(report, attributes: attributes)
     }
 }
 
--- a/DatadogCrashReporting/Sources/CrashReportingFeature.swift
+++ b/DatadogCrashReporting/Sources/CrashReportingFeature.swift
@@ -7,10 +7,10 @@
 import Foundation
 import DatadogInternal
 
-internal final class CrashReportingFeature: DatadogFeature {
-    static let name = "crash-reporter"
+public final class CrashReportingFeature: DatadogFeature {
+    public static let name = "crash-reporter"
 
-    let messageReceiver: FeatureMessageReceiver
+    public let messageReceiver: FeatureMessageReceiver
 
     /// Queue for synchronizing internal operations.
     private let queue: DispatchQueue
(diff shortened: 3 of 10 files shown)


maxep:
Hello @naftaly 👋 

Thank you very much for your contribution! Enabling the injection of a different crash reporting library is a great idea.  

I have a few questions and comments regarding the proposed changes:  

1. Our `DatadogCrashReporting` module has a direct dependency on `PLCrashReporter`, so the conditional import statements will always evaluate to `true`. Could you elaborate on the motivation behind it?  

2. I noticed two approaches for reporting crashes: one active, using the `send(_ report:)` method, and another reactive, through the `CrashReportingPlugin` protocol and its callback. What specific use cases are these designed to address?  
   Typically, crash reporting libraries provide the crash report at app launch. In this case, having the `DatadogCrashReporting` module retrieve the report from its plugin when ready seems sufficient. Am I missing a scenario where both methods are necessary?  

3. Regarding the `send(_ report:)` method, what type of `attributes` are you intending to inject? Are they related to the application's current state, or do they refer to information from the previous session?  

Thank you again for your effort in improving our crash reporting!

naftaly:
> Hello @naftaly 👋
> 
> Thank you very much for your contribution! Enabling the injection of a different crash reporting library is a great idea.

Thank you, I wasa hoping you'd like it :)

> 
> I have a few questions and comments regarding the proposed changes:
> 
> 1. Our `DatadogCrashReporting` module has a direct dependency on `PLCrashReporter`, so the conditional import statements will always evaluate to `true`. Could you elaborate on the motivation behind it?

The source I'm working with ends up importing through Cocoapod which for some reason leads to "CrashReporter" not being available. I didn't get into it too much, but I thought it would be a good solution until I got around to making a feature request to split out the plugin implementation from the CrashReporting infrastructure since we'd prefer to not include it since we wouldn't be using it, PLCrashReporter that is.

> 2. I noticed two approaches for reporting crashes: one active, using the `send(_ report:)` method, and another reactive, through the `CrashReportingPlugin` protocol and its callback. What specific use cases are these designed to address?
>    Typically, crash reporting libraries provide the crash report at app launch. In this case, having the `DatadogCrashReporting` module retrieve the report from its plugin when ready seems sufficient. Am I missing a scenario where both methods are necessary?

You're totally right, that is a typical scenario. But there are also other scenarios where dependencies might not be ready yet, or simply that the infra responsible to crashes wants to send them at some other point in time, such as during downtime, or not during startup. 

In our case, we have crash data coming from some other locations, and we enhance it with multiple different sets of data that for some reason or another are not ready as early as we need to call CrashReporter.enable().

Overall though, we want to start up the reporting system very very early to ensure we have the most complete set of data, and we want to report the crashes when we feel it is appropriate later during the session, or even in some future session (not necessarily the next session).

> 3. Regarding the `send(_ report:)` method, what type of `attributes` are you intending to inject? Are they related to the application's current state, or do they refer to information from the previous session?

The attributes can really be anything, but they are the extended set of data I was talking about earlier that come from other datasets. They don't always refer to the previous session, but they refer to the session of report we are sending.

> 
> Thank you again for your effort in improving our crash reporting!

Thank you for considering our PR, we love using Datadog and hope we can continue helping out as much we can.

naftaly:
> Hello @naftaly
> 
> Thank you for the additional context and clarification, it makes sense.
> 
> > The source I'm working with ends up importing through Cocoapod which for some reason leads to "CrashReporter" not being available. I didn't get into it too much, but I thought it would be a good solution until I got around to making a feature request to split out the plugin implementation from the CrashReporting infrastructure since we'd prefer to not include it since we wouldn't be using it, PLCrashReporter that is.
> 
> Our [podspec](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogCrashReporting.podspec#L26) depends on PLCR, so I'm curious—what specific issue are you trying to address? It seems preferable to resolve the CocoaPods problem directly rather than introducing conditional imports into the source code.

We end up working with a Bazel build system which does special things with Pods, so it's probably an issue on our side that I guess I simply swept away :) I'll figure it out on our side so we can simply forget about those import checks I think.

> 
> Splitting the plugin interface from its implementation would have a broader impact, as it would require architectural changes, including adding a new module. Submitting a feature request is the way to go 👍

Great, sounds good. Can we consider this a feature request or is there another route?

> 
> Regarding attributes, I believe it would make sense to add them as a property of `DDCrashReport`. This way, both active and reactive reporting methods could inject attributes. However, implementing this would require changes to the RUM and Logs modules to ensure proper attribute propagation. Of course, we don’t expect you to make such changes. Our usual process for external PRs is to merge them into a dedicated branch, where we handle reworking and add the necessary tests ourselves.
> 

That sounds good to me, thank you.


> One other change I noticed: making the `CrashReportingFeature` public. Could you explain the reasoning behind this?
> 

It could be internal instead. Due to `CrashReporting.enabled/send` becoming public, it needed to be upgraded as well.

PR #2126, opened 2024-11-29
RUM-7421 Expose Crash Plugin

maxep:
### What and why?

Follow-up #2116

Expose `CrashReportingPlugin` publicly and allow reporting crashes asynchronously with additional attributes.
The plugin can also provide an optional backtrace reporter.

#### Crash Reporting Plugin

The plugin interface is now `public` with `@escaping` reporting closure, allowing reporting crashes at any time.
The interface also define an optional `backtraceReporter` property for reading backtraces.
This new plugin interface allows integration to fully replace the `PLCrashReporter` implementation.

#### Crash Report Additional Attributes

The `DDCrashReport` now defines `additionalAttributes` property to propagate atrributes.
These attributes are merged with logs' user attributes and RUM Error's context.

### Review checklist

### DIFF
--- a/Datadog/IntegrationUnitTests/CrashReporting/SendingCrashReportTests.swift
+++ b/Datadog/IntegrationUnitTests/CrashReporting/SendingCrashReportTests.swift
@@ -19,6 +19,8 @@ private class CrashReporterMock: CrashReportingPlugin {
     internal var pendingCrashReport: DDCrashReport?
     @ReadWriteLock
     internal var injectedContext: Data? = nil
+    /// Custom backtrace reporter injected to the plugin.
+    var injectedBacktraceReporter: BacktraceReporting?
 
     init(pendingCrashReport: DDCrashReport? = nil) {
         self.pendingCrashReport = pendingCrashReport
@@ -26,6 +28,7 @@ private class CrashReporterMock: CrashReportingPlugin {
 
     func readPendingCrashReport(completion: (DDCrashReport?) -> Bool) { _ = completion(pendingCrashReport) }
     func inject(context: Data) { injectedContext = context }
+    var backtraceReporter: BacktraceReporting? { injectedBacktraceReporter }
 }
 
 /// Covers broad scenarios of sending Crash Reports.
--- a/Datadog/IntegrationUnitTests/CrashReporting/SendingCrashReportTests.swift
+++ b/Datadog/IntegrationUnitTests/CrashReporting/SendingCrashReportTests.swift
@@ -85,7 +85,9 @@ class SendingCrashReportTests: XCTestCase {
         XCTAssertNotNil(rumEvent.error.binaryImages)
         XCTAssertNotNil(rumEvent.error.meta)
         XCTAssertNotNil(rumEvent.error.wasTruncated)
-        DDAssertJSONEqual(rumEvent.context!.contextInfo, crashContext.lastRUMAttributes!)
+        let contextAttributes = try XCTUnwrap(rumEvent.context?.contextInfo)
+        let lastRUMAttributes = try XCTUnwrap(crashContext.lastRUMAttributes?.attributes)
+        DDAssertJSONEqual(contextAttributes, lastRUMAttributes.merging(crashReportAttributes) { $1 })
     }
 
     func testWhenSendingCrashReportAsLog_itIsLinkedToTheRUMSessionThatHasCrashed() throws {
--- a/Datadog/IntegrationUnitTests/CrashReporting/SendingCrashReportTests.swift
+++ b/Datadog/IntegrationUnitTests/CrashReporting/SendingCrashReportTests.swift
@@ -55,6 +55,7 @@ class SendingCrashReportTests: XCTestCase {
             lastLogAttributes: .init(mockRandomAttributes())
         )
         let crashReport: DDCrashReport = .mockRandomWith(context: crashContext)
+        let crashReportAttributes: [String: Encodable] = try XCTUnwrap(crashReport.additionalAttributes.dd.decode())
 
         // When
         Logs.enable(with: .init(), in: core)
@@ -68,8 +69,8 @@ class SendingCrashReportTests: XCTestCase {
         XCTAssertEqual(log.error?.message, crashReport.message)
         XCTAssertEqual(log.error?.kind, crashReport.type)
         XCTAssertEqual(log.error?.stack, crashReport.stack)
-        XCTAssertFalse(log.attributes.userAttributes.isEmpty)
-        DDAssertJSONEqual(log.attributes.userAttributes, crashContext.lastLogAttributes!)
+        let lastLogAttributes: [String: Encodable] = try XCTUnwrap(crashContext.lastLogAttributes.dd.decode())
+        DDAssertJSONEqual(log.attributes.userAttributes, lastLogAttributes.merging(crashReportAttributes) { $1 })
         XCTAssertNotNil(log.attributes.internalAttributes?[DDError.threads])
         XCTAssertNotNil(log.attributes.internalAttributes?[DDError.binaryImages])
         XCTAssertNotNil(log.attributes.internalAttributes?[DDError.meta])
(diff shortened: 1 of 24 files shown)


DatadogCrashReporting/Sources/CrashReportingFeature.swift
```diff
@@ -31,59 +28,57 @@ internal final class CrashReportingFeature: DatadogFeature {
...
-                    // `CrashContext` is malformed and and cannot be read. Return `true` to let the crash reporter
-                    // purge this crash report as we are not able to process it respectively.
-                    self.sender.send(launch: .init(didCrash: true))
-                    return true
-                }
-
-                self.sender.send(report: availableCrashReport, with: crashContext)
+        // The plugin.readPendingCrashReport method is escaping, so
+        // every call in the closure must be thread-safe
+        plugin.readPendingCrashReport { [weak self] crashReport in
```
ncreated, line 53: blocking/ The removal of the `queue` causes the plugin work to be executed on the main thread ⚠️. For the default plugin, this will result in all I/O operations related to loading and purging crash data being performed on the main thread (see [here](https://github.com/DataDog/dd-sdk-ios/blob/ab8f2fbfb275e8ab7afc962c87794721854358fc/DatadogCrashReporting/Sources/PLCrashReporterIntegration/PLCrashReporterPlugin.swift#L39-L55)). That will likely cause perf hit and will impact app startup durations.

Additionally, by removing the queue we're breaking the contract expected by the plugin implementation. As specified [here](https://github.com/DataDog/dd-sdk-ios/blob/ab8f2fbfb275e8ab7afc962c87794721854358fc/DatadogCrashReporting/Sources/CrashReportingPlugin.swift#L19-L20), both plugin methods are expected to be called on a secondary thread:
```swift
    /// The SDK calls this method on a background thread. The implementation is free to choice any thread
    /// for executing the  `completion`.
    func readPendingCrashReport(completion: (DDCrashReport?) -> Bool)

    /// The SDK calls this method for each significant application state change.
    /// It is called on a background thread and succeeding calls are synchronized.
    func inject(context: Data)
```

ncreated (changes requested):
👌  Everything looks great except the removal of `queue` that will cause performance regression (see my other comment).

Datadog/IntegrationUnitTests/CrashReporting/SendingCrashReportTests.swift
```diff
@@ -81,7 +85,9 @@ class SendingCrashReportTests: XCTestCase {
         XCTAssertNotNil(rumEvent.error.binaryImages)
         XCTAssertNotNil(rumEvent.error.meta)
         XCTAssertNotNil(rumEvent.error.wasTruncated)
-        DDAssertJSONEqual(rumEvent.context!.contextInfo, crashContext.lastRUMAttributes!)
+        let contextAttributes = try XCTUnwrap(rumEvent.context?.contextInfo)
+        let lastRUMAttributes = try XCTUnwrap(crashContext.lastRUMAttributes?.attributes)
+        DDAssertJSONEqual(contextAttributes, lastRUMAttributes.merging(crashReportAttributes) { $1 })
```
ncreated, line 90: 🏅

DatadogInternal/Sources/Models/CrashReporting/DDCrashReport.swift
```diff
@@ -97,5 +100,6 @@ public struct DDCrashReport: Codable, PassthroughAnyCodable {
         self.meta = meta
         self.wasTruncated = wasTruncated
         self.context = context
+        self.additionalAttributes = AnyCodable(additionalAttributes)
```
ncreated, line 103: question/ Am I correct by saying that `Codable` requirement for `DDCrashReport` only comes from the message bus interface? We're using `PassthroughAnyCodable` so there's no actual coding and perf hit, but things like passing attribtues ☝️ are far from easy.

DatadogCrashReporting/Sources/CrashReportingFeature.swift
maxep: Ah yes, indeed! Good catch!

DatadogInternal/Sources/Models/CrashReporting/DDCrashReport.swift
maxep: Exactly, it is only for conformance.
The `DDCrashReport` and `CrashContext` are now both in `DatadogInternal`, so it's no longer required to pass them as `FeatureBaggage`. We could pass them directly as a `.crash(report: DDCrashReport, context: CrashContext)` message on the bus, but such change has wider impact and should be treated separately IMO.

mariedm (dismissed):

ncreated (dismissed):
Looks great 💪

maxep:
Hello @naftaly 👋 

Here is the follow up of your contribution. The 2 main changes are:
1. The added attributes are now part of the `DDCrashReport` object. They are merged with either the RUM Error context-info or with the error log's user attributes.
2. We have removed the public interface to `send` a report. Instead, we have made the `CrashReportingPlugin.readPendingCrashReport` closure escaping so you can report the crash at any time. In practice, you will store the closure in your plugin implementation to report the crash later.

I hope this will answer your needs!

naftaly:
> Hello @naftaly 👋
> 
> Here is the follow up of your contribution. The 2 main changes are:
> 
> 1. The added attributes are now part of the `DDCrashReport` object. They are merged with either the RUM Error context-info or with the error log's user attributes.
> 2. We have removed the public interface to `send` a report. Instead, we have made the `CrashReportingPlugin.readPendingCrashReport` closure escaping so you can report the crash at any time. In practice, you will store the closure in your plugin implementation to report the crash later.
> 
> I hope this will answer your needs!

Thank you for your response. I'm curious why you chose to have the user store a block. Approaches like this can easily lead to retain cycles and unexpected behavior. In contrast, a simple send call is more explicit and easier to understand.

maxep:
> Thank you for your response. I'm curious why you chose to have the user store a block. Approaches like this can easily lead to retain cycles and unexpected behavior. In contrast, a simple send call is more explicit and easier to understand.

Internally, we make sure that the closure does not retain anything, if the Datadog SDK instance is deallocated, the closure will have no effect. Additionally, the crash-reporting feature is immutable and the closure is only about transmitting the crash report by jumping on a dedicated queue (we've called the message-bus), so the closure is also thread-safe.

The reason we went with this approach is that we want a single entry-point for interacting with the crash reporter lib, and that entry-point is the `CrashReportingPlugin`. By solely relying on the closure, we also control the lifecycle: The SDK will provide the closure when it's ready to make a report.
Providing a custom plugin is an advanced usage that is far from standard integration, one that require data injection and report formatting. Having a `send` method is indeed simple, but could also lead to mis-use.

I hope this make sense!

mariedm (approved):

ncreated (approved):

maciejburda (approved):

naftaly:
> > Thank you for your response. I'm curious why you chose to have the user store a block. Approaches like this can easily lead to retain cycles and unexpected behavior. In contrast, a simple send call is more explicit and easier to understand.
> 
> Internally, we make sure that the closure does not retain anything, if the Datadog SDK instance is deallocated, the closure will have no effect. Additionally, the crash-reporting feature is immutable and the closure is only about transmitting the crash report by jumping on a dedicated queue (we've called the message-bus), so the closure is also thread-safe.
> 
> The reason we went with this approach is that we want a single entry-point for interacting with the crash reporter lib, and that entry-point is the `CrashReportingPlugin`. By solely relying on the closure, we also control the lifecycle: The SDK will provide the closure when it's ready to make a report. Providing a custom plugin is an advanced usage that is far from standard integration, one that require data injection and report formatting. Having a `send` method is indeed simple, but could also lead to mis-use.
> 
> I hope this make sense!

Hi @maxep ,

I just wanted to validate that you don't expect the completion block to be called on any specific if called in the future by a 3rd party. For example, if I hold onto the block and call it at a later point, it won't be on the internal `queue` as if it was called in a non-escaping manner in the crash plugin.

thanks.

maxep:
Hey @naftaly 👋 

> I just wanted to validate that you don't expect the completion block to be called on any specific if called in the future by a 3rd party. For example, if I hold onto the block and call it at a later point, it won't be on the internal queue as if it was called in a non-escaping manner in the crash plugin.

Exactly, you can hold the `completion` block and call it later from any thread 👍

PR #2127, opened 2024-12-03
RUM-6850: Add "Effective Sample Rate" to telemetry events

simaoseica-dd:
### What and why?

This PR adds a new property `effectiveSampleRate` to all types of telemetry events.
We want to know the actual rate applied to the events that is composed by all the sample rates, in the different layers, through which the event passes.

### How?

- Composes `effectiveSampleRate` with the global `telemetrySampleRate` for:
   -  Configuration telemetry events, with "configuration telemetry sample rate" of 20%.
   -  Metrics telemetry events, with "metrics telemetry sample rate" of 15%.
   -  Individual metrics events, like debug and error events.

- Logic centralized on `SampleRate` type alias:
   - Standard type for sample rate parameters
   - Logic for the composition of sample rates centralized in one place

### Review checklist


DatadogRUM/Sources/Integrations/TelemetryReceiver.swift
```diff
@@ -255,11 +259,18 @@ internal final class TelemetryReceiver: FeatureMessageReceiver {
             let sessionIDOverride: String? = attributes.removeValue(forKey: SDKMetricFields.sessionIDOverrideKey)?.dd.decode()
             let sessionID = sessionIDOverride ?? rum?.sessionID
 
+            // Calculates the composition of sample rates. The metric can have up to 3 layers of sampling.
+            var effectiveSampleRate = metric.sampleRate.composed(with: self.sampler.samplingRate)
+            if let headSampleRate = metric.attributes[MethodCalledMetric.headSampleRate] as? SampleRate {
+                effectiveSampleRate = effectiveSampleRate.composed(with: headSampleRate)
+            }
```
ncreated, line 266: suggestion/ This is generic flow for any type of metric, but here we use `MethodCalledMetric.headSampleRate` which is specific to "Method Called" metric. It could make sense to move `.headSampleRate` to more generic [`SDKMetricFields`](https://github.com/DataDog/dd-sdk-ios/blob/a4c6a1cb49a22eec9f2d5dada8be160082533d35/DatadogInternal/Sources/SDKMetrics/SDKMetricFields.swift#L10) namespace, WDYT?


Also, the `.headSampleRate` is never removed from `attributes`, which means it is uploaded with the metric log, no? I think we should remove it here, similar to how we handle `SDKMetricFields.sessionIDOverrideKey` above.

DatadogRUM/Sources/Integrations/TelemetryReceiver.swift
```diff
@@ -255,11 +259,18 @@ internal final class TelemetryReceiver: FeatureMessageReceiver {
...
+
             let event = TelemetryDebugEvent(
                 dd: .init(),
                 action: rum?.userActionID.map { .init(id: $0) },
                 application: rum.map { .init(id: $0.applicationID) },
                 date: date.addingTimeInterval(context.serverTimeOffset).timeIntervalSince1970.toInt64Milliseconds,
+                effectiveSampleRate: Int64(effectiveSampleRate),
```
ncreated, line 273: blocking/ This can overflow, especially that `telemetrySampleRate` is controlled by the user. We should use [`Int64(withNoOverflow:)`](https://github.com/DataDog/dd-sdk-ios/blob/a4c6a1cb49a22eec9f2d5dada8be160082533d35/DatadogInternal/Sources/Utils/SwiftExtensions.swift#L96-L99) instead.

Datadog/IntegrationUnitTests/Public/CoreTelemetryIntegrationTests.swift
```diff
@@ -204,4 +204,51 @@ class CoreTelemetryIntegrationTests: XCTestCase {
...
+        if let debug = debugEvents.first,
+           debug.telemetry.message == "Debug Telemetry" {
+         
+            XCTAssertEqual(debug.effectiveSampleRate, 90)
+        }
+        
+        if let error = errorEvents.first {
+            XCTAssertEqual(error.telemetry.message, "Error Telemetry")
+            XCTAssertEqual(error.effectiveSampleRate, 90)
+        }
```
ncreated, line 238: suggestion/ As I understand, we use if let here because the events may not be sent due to the sample rate values we are testing, which makes sense 👍.

However, this approach could become unreliable if an issue is introduced, but no events are sampled, meaning assertions won’t catch potential regressions. In such cases, CI could pass even though the test might fail at another time due to missed samples.

To mitigate this, we could increase the sample rate and send more events, reducing the likelihood of having zero samples in the resulting arrays. By sending multiple telemetry events, the chances of not having a single sample would become negligible, and we could safely assert that the events are sent.

Here’s an example of how this can be done:
```swift
// When
for idx in (0..<100) {
    core.telemetry.debug("Debug Telemetry \(idx)")
    core.telemetry.error("Error Telemetry \(idx)")
    core.telemetry.metric(name: "Metric Name", attributes: [:], sampleRate: 99)
    core.telemetry.send(telemetry: .usage(.init(event: .setUser, sampleRate: 100)))
    core.telemetry.stopMethodCalled(
        core.telemetry.startMethodCalled(operationName: .mockRandom(), callerClass: .mockRandom(), headSampleRate: 80),
        tailSampleRate: 99
    )
}
```

With this approach, the chances of not having a single sample in the resulting arrays are extremely low, allowing us to assert:
```swift
// Then
XCTAssertGreaterThan(debugEvents.count, 0)
XCTAssertGreaterThan(errorEvents.count, 0)
XCTAssertGreaterThan(usageEvents.count, 0)

// loop through each and assert on `effectiveSampleRate`
```

**Rationale:**
Given an effectiveSampleRate of 80%, the probability of having no sample in 100 consecutive attempts is `(0.2)^100`, which is effectively zero. Therefore, we can confidently assert the presence of events in the arrays.

Datadog/IntegrationUnitTests/Public/CoreTelemetryIntegrationTests.swift
```diff
@@ -204,4 +204,51 @@ class CoreTelemetryIntegrationTests: XCTestCase {
...
+
+        if let metric = debugEvents.first(where: { $0.telemetry.message == "[Mobile Metric] Metric Name" }) {
+
+            XCTAssertEqual(metric.effectiveSampleRate, 89)
+        }
+
+        if let metric = debugEvents.first(where: { $0.telemetry.message == "[Mobile Metric] Method Called" }) {
+
+            XCTAssertEqual(metric.effectiveSampleRate, 71)
```
ncreated, line 247: suggestion/ `71` looks quite magical here. What if we extract all sample rates to separate properties and define it as multiplication of certain rates?
```swift
XCTAssertEqual(metric.effectiveSampleRate, telemetrySampleRate * metricsSampleRate)
```

DatadogRUM/Sources/Integrations/TelemetryReceiver.swift
simaoseica-dd: You're right. ty

DatadogRUM/Sources/Integrations/TelemetryReceiver.swift
simaoseica-dd: Makes sense. It is generic and has an appropriate enum. Also removed it from being uploaded. ty

Datadog/IntegrationUnitTests/Public/CoreTelemetryIntegrationTests.swift
simaoseica-dd: Good point here 🙏  It could hide some regressions. Changed based on the suggestion!

mariedm (dismissed):

DatadogRUM/Sources/RUMConfiguration.swift
```diff
@@ -399,14 +397,3 @@ extension RUM.Configuration {
 }
 
 extension RUM.Configuration: InternalExtended {}
-extension InternalExtension where ExtendedType == RUM.Configuration {
-    /// The sampling rate for configuration telemetry events. When set, it overwrites the value
-    /// of `configurationTelemetrySampleRate` in `RUM.Configuration`.
-    ///
-    /// It is mostly used to enable or disable telemetry events when running test scenarios.
-    /// Expects value between `0.0` and `100.0`.
-    public var configurationTelemetrySampleRate: Float {
```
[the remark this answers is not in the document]
simaoseica-dd: It seems that the flutter's `DatadogRumPlugin` has its own configuration for `configurationTelemetrySampleRate` and that we don't want to expose it publicly in the `RUM.Configuration`.
I rolled back the changes.

ncreated (approved):
🏅 🎯 Well done, LG!

mariedm (approved):

PR #2129, opened 2024-12-04
RUM-6583: Fix flaky test for RUM CPU vital

simaoseica-dd:
### What and why?

The test [testWhenCPUUnderHeavyLoadItMeasuresHigherCPUTicks()](https://github.com/datadog/dd-sdk-ios/blob/965f1e09f704a79c7158233d00608c99d38ac498/DatadogCore/Tests/Datadog/RUM/RUMVitals/VitalCPUReaderTests.swift#L15-L25) compares the CPU ticks under heavy and light processing and it started being flaky in GitLab. The CPU ticks measured should be greater on heavy processing.

This PR mitigates the flakiness of the test.

### How?

- To avoid false positives, the blocks of processing are run more than once. This way we can have an average of measures that dilute the false positives. 
- The duration of both tasks are now the same, to avoid comparisons of CPU ticks under different time intervals. 
- It uses `CFAbsoluteTimeGetCurrent` instead of `Date` to get the time since it is more suitable for performance measurements providing high-precision timing.

### Review checklist
- [ ] ~~Add Objective-C interface for public APIs (see our [guidelines](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3157787243/RFC+-+Modular+Objective-C+Interface#Recommended-solution) [internal]) and run `make api-surface`)~~

### DIFF
--- a/DatadogCore/Tests/Datadog/RUM/RUMVitals/VitalCPUReaderTests.swift
+++ b/DatadogCore/Tests/Datadog/RUM/RUMVitals/VitalCPUReaderTests.swift
@@ -13,15 +13,28 @@ class VitalCPUReaderTest: XCTestCase {
     lazy var cpuReader = VitalCPUReader(notificationCenter: testNotificationCenter)
 
     func testWhenCPUUnderHeavyLoadItMeasuresHigherCPUTicks() throws {
-        let highLoadAverage = try averageCPUTicks {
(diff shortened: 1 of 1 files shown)


mariedm (approved):
Nice improvement! 🙌 Thank you for taking care of this flakiness. ❄

PR #2131, opened 2024-12-04
RUM-7475 [SR][SwiftUI] Update snapshot tests

mariedm:
### What and why?

With the introduction of SwiftUI support in Session Replay, our snapshot tests need to be updated to cover a wider range of components.

### How?

- Created a dedicated `SwiftUIView` to test SwiftUI components.
- Added tests for additional components: TextField, Button, Divider, Image, List
- Added snapshot tests for different privacy configurations

| Mask none | Mask user input + non bundled images | Mask all | 
| ------------- | ------------- | ------------- |
| testSwiftUI()-maskAll_images-maskSensitiveInputs-privacy | testSwiftUI()-maskNonBundledOnly_images-maskAllInputs-privacy | testSwiftUI()-maskAll_images-maskAll-privacy |

### Review checklist


DatadogSessionReplay/SRSnapshotTests/SRFixtures/Sources/SRFixtures/ViewControllers/SwiftUIView.swift
```diff
@@ -0,0 +1,66 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import SwiftUI
+
+@available(iOS 13.0, *)
```
simaoseica-dd, line 9: question/ do we have a target only for snapshot testing? I'm wondering why the View is set for a low version while being used only for snapshots. 
Looking at the env var, it seems pointing to 17.5 (`DEFAULT_SR_SNAPSHOT_TESTS_OS := 17.5`)

DatadogSessionReplay/SRSnapshotTests/SRFixtures/Sources/SRFixtures/ViewControllers/SwiftUIView.swift
```diff
@@ -0,0 +1,66 @@
...
+struct SwiftUIView: View {
+    var body: some View {
+        VStack(spacing: 10) {
+            Text("Hello, SwiftUI!")
+                .font(.headline)
+
+            if #available(iOS 14.0, *) {
+                if #available(iOS 15.0, *) {
```
simaoseica-dd, line 17: suggestion/ Since this is only for snapshots the availability in the `SwiftUIView` should be enough with `@available(iOS 15.0, *)`

I'm thinking about the tests starting to fail and we can't understand immediately that one block of the view was available in the new test environment.

simaoseica-dd (commented):
Just a few comments/questions

DatadogSessionReplay/SRSnapshotTests/SRFixtures/Sources/SRFixtures/Resources/Assets.xcassets/Flowers_1.imageset/Flowers_1.png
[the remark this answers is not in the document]
mariedm: We only need it to be larger than 100x100 for the test, but its details don't matter. I just grabbed an existing image from the repo to test, and didn't think about it later. It's def a good idea to use a smaller image, so I reduced its size; it's now its less than 50 kb.

DatadogSessionReplay/SRSnapshotTests/SRFixtures/Sources/SRFixtures/ViewControllers/SwiftUIView.swift
mariedm: Totally! I added these UI elements bit by bit to test, and didn't clean up at the end. I removed all these checks to only keep a single one for the whole class.

simaoseica-dd (dismissed):

maxep (approved):
🎉

Issue #2134, opened 2024-12-05
SQ - MessageBus Crash L:97


KVTaniguchi:
### Stack trace

```
EXC_BREAKPOINT: BUG IN CLIENT OF LIBMALLOC: memory corruption of free block

0  libsystem_malloc.dylib +0x4c28  __xzm_xzone_malloc_tiny_outlined
1  libswiftCore.dylib +0x3d3cb4    __swift_allocObject_
2  SPOS-App +0xa645640             MessageBus.send(message:else:) (MessageBus.swift:97:9)
3  SPOS-App +0xa63a128             protocol witness for MessageSending.send(message:else:) in conformance DatadogCore (DatadogCore.swift:307:13)
4  SPOS-App +0xa637c64             closure #1 in DatadogCore.init(directory:dateProvider:initialConsent:performance:httpClient:encryption:contextProvider:applicationVersion:maxBatchesPerUpload:backgroundTasksEnabled:isRunFromExtension:) (DatadogCore.swift:122:19)
5  SPOS-App +0xa6349ac             thunk for @escaping @callee_guaranteed (@in_guaranteed DatadogContext) -> () (<compiler-generated>)
6  SPOS-App +0xa633c00             closure #1 in DatadogContextProvider.context.didset (DatadogContextProvider.swift￼38)
7  SPOS-App +0xa634868             closure #1 in DatadogContextProvider.write(block:) (DatadogContextProvider.swift￼44)
8  SPOS-App +0x43cd80              thunk for @escaping @callee_guaranteed () -> ()
9  libdispatch.dylib +0x236c       __dispatch_call_block_and_release
10 libdispatch.dylib +0x40cc       __dispatch_client_callout
11 libdispatch.dylib +0xb6d4       __dispatch_lane_serial_drain
12 libdispatch.dylib +0xc1dc       __dispatch_lane_invoke
13 libdispatch.dylib +0x17254      __dispatch_root_queue_drain_deferred_wlh
14 libdispatch.dylib +0x16aa0      __dispatch_workloop_worker_thread
15 libsystem_pthread.dylib +0x4c78 __pthread_wqthread
```

### Volume

0.4%

### Affected SDK versions

2.13.0

### Latest working SDK version

2.13.0

### Does the crash manifest in the latest SDK version?

No

### Device Information

> 10 different devices from iPad 8 to iPhone 16 Pro.

PR #2135, opened 2024-12-05
ISSUE-2431-MessageBusCrash

KVTaniguchi:
### What and why?

The purpose of this PR is to try to fix a crash we are seeing.

https://github.com/DataDog/dd-sdk-ios/issues/2134

I suspect that MessageBus is getting deallocated and calling Self inside an asynchronous operation can cause memory issues.  The change I am making is a standard tactic for guarding against this.

### How?

capture self and check that it exists 

### Review checklist
- [x] Feature or bugfix MUST have appropriate tests (unit, integration) - should be covered under existing tests

### DIFF
--- a/DatadogCore/Sources/Core/MessageBus.swift
+++ b/DatadogCore/Sources/Core/MessageBus.swift
@@ -94,7 +94,8 @@ internal final class MessageBus {
             return save(configuration: configuration)
         }
 
-        queue.async {
+        queue.async { [weak self] in
+            guard let self else { return }
             guard let core = self.core else {
                 return
             }


KVTaniguchi:
maxep
  Today at 10:16
Hello :wave:
Thank you for the GH issue, it will be easier for you to track down the progress :+1:
Tho, I'm really not sure the PR will fix it: self is retained by the closure. And the only scenario that would make the reference count reach 0 is when the SDK is deinitialised (when calling Datadog.stopInstance).
The impact table that you have shared suggest that this relates to the release of iOS 18, is that a fair assumption based on your data?
Does the crash happened with different stacktraces? Or is it always with the same [0-3] frames (crashing on __xzm_xzone_malloc_tiny_outlined?
We continue the investigation on our side, it would be super helpful if you could share steps to reproduce if you find any. Thank you for your help!

PR #2136, opened 2024-12-06
RUM-6584: Add dependency between "Publish Cocoapods podspecs" GitLab jobs

simaoseica-dd:
### What and why?

This PR adds relationships between the jobs that publish Cocoapods `podspecs` into the public trunk. 
We want to prevent jobs from running if their dependencies failed to run.

### How?

By using the [needs](https://docs.gitlab.com/ee/ci/yaml/#needs) job keyword we can create relationships between the CI jobs:
- `Publish CP podspecs (dependent)` depends on `Publish CP podspecs (internal)`
- `Publish CP podspecs (legacy)` depends on `Publish CP podspecs (dependent)`

### Review checklist
- [ ] ~~Feature or bugfix MUST have appropriate tests (unit, integration)~~
- [ ] ~~Add Objective-C interface for public APIs (see our [guidelines](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3157787243/RFC+-+Modular+Objective-C+Interface#Recommended-solution) [internal]) and run `make api-surface`)~~

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -359,6 +359,7 @@ Publish CP podspecs (dependent):
     - !reference [.release-pipeline-20m-delayed-job, rules]
   before_script:
     - *export_MAKE_release_params
+  needs: ["Publish CP podspecs (internal)"]
   script:
     - ./tools/runner-setup.sh --xcode "$DEFAULT_XCODE"
     - make env-check
@@ -371,6 +372,7 @@ Publish CP podspecs (legacy):
     - !reference [.release-pipeline-40m-delayed-job, rules]
   before_script:
     - *export_MAKE_release_params
+  needs: ["Publish CP podspecs (dependent)"]
   script:
     - ./tools/runner-setup.sh --xcode "$DEFAULT_XCODE"
     - make env-check


.gitlab-ci.yml
```diff
@@ -359,6 +359,7 @@ Publish CP podspecs (dependent):
     - !reference [.release-pipeline-20m-delayed-job, rules]
```
mariedm, line 359: Do we still need the 20 min delay? This seems redundant and could potentially delay the pipeline unnecessarily.

.gitlab-ci.yml
```diff
@@ -371,6 +372,7 @@ Publish CP podspecs (legacy):
     - !reference [.release-pipeline-40m-delayed-job, rules]
```
mariedm, line 372: Same question here with the 40 min delay

mariedm (approved):
This is a nice addition!

.gitlab-ci.yml
simaoseica-dd: I had this same question!
It seems that after publishing to the CDN with `pod trunk push --synchronous *`, the new version of the pod is not immediately available. Some users have [mentioned](https://stackoverflow.com/questions/41702187/how-long-does-it-take-for-a-new-custom-cocoapod-to-get-uploaded-to-the-pod-trunk) that it may take a while before it becomes accessible via `pod install --repo-update` or `pod repo update`.

.gitlab-ci.yml
simaoseica-dd: Same as above

.gitlab-ci.yml
mariedm: Yes, it takes a while... The best source of truth are the [CP specs](https://github.com/CocoaPods/Specs/tree/master/Specs/d/2/d/DatadogCore) so far. As you can see, even the [website](https://cocoapods.org/pods/DatadogSDK) is not up to date; we released v2.20 3 weeks ago, and yet it shows v2.16.

mariedm: I remember @ncreated added these delays, so he's probably the one with more context on this.

ncreated: @simaoseica-dd is correct — after pod `trunk push --synchronous *` completes, it typically takes around 20 minutes for the pod to become available for use by any pod command, even with `--repo-update`. If we attempt to push dependent pods before this, pod trunk push will fail.

Currently, our release jobs are [designed to be **failure-resistant**](https://github.com/DataDog/dd-sdk-ios/blob/a4c6a1cb49a22eec9f2d5dada8be160082533d35/tools/release/publish-podspec.sh#L213-L218). If a pod isn't yet available and pod trunk push fails, the job won’t fail outright; instead, it will wait for a minute and then retry the push.

This means that while the 20-minute delay is not strictly required in our automation, it's a safeguard. Our CI jobs are constrained by a 1-hour timeout. If we don't introduce the delay, we risk running out of time, leaving only 40 minutes to push all podspecs — which may not be sufficient to complete the entire process.

PR #2143, opened 2024-12-12
RUM-3029: Stopped view shouldn't send values added to the global attributes after it was stopped

simaoseica-dd:
### What and why?

This PR ensures that a `View` collects attributes only during its active state. 
Once the View is stopped, resources, actions, and global attributes are no longer associated with it, preventing potential leaks.

### How?

The `View` attributes' updates are validated through:
- unit tests on `RUMViewScopeTests`
- integration tests on `RUMViewScopeIntegrationTests`

Additionally, it fixes tests that are failing on Xcode 16 due to the use of not declared `ViewControllers` that didn't have an explicit initializer. 
The error that was being thrown is: 
> UIViewController is missing its initial trait collection populated during initialization. This is a serious bug, likely caused by accessing properties or methods on the view controller before calling a UIViewController initializer

### Review checklist


Datadog/IntegrationUnitTests/RUM/RUMViewScopeIntegrationTests.swift
```diff
@@ -0,0 +1,193 @@
...
+
+        let customView = try XCTUnwrap(session.views.first(where: { $0.name == viewName }))
+
+        customView.viewEvents.forEach { viewEvent in
+
+            XCTAssertEqual(viewEvent.context?.contextInfo as? [AttributeKey: AnyCodable], initialAttributes)
+        }
+    }
+
+    func testWhenViewReceivesActions_theViewAttributesAreNotUpdated_withActionAttributes() throws {
```
ncreated, line 120: change-request/ This test seems unusual as it doesn't validate existing behavior but instead asserts that a non-existent behavior doesn't occur. Event attributes are not meant to be inherited by other events, which is intentional. This test essentially verifies that this behavior doesn't happen.

A more valuable approach would be to add coverage for global attributes, which should be included in all events. Specifically, it would be useful to test that:

- Global attributes are added to action events.
- Global attributes are added to resource events.
- Global attributes are added to error events.

Datadog/IntegrationUnitTests/RUM/RUMViewScopeIntegrationTests.swift
```diff
@@ -0,0 +1,193 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import XCTest
+@testable import DatadogRUM
+@testable import DatadogInternal
+
+final class RUMViewScopeIntegrationTests: XCTestCase {
```
ncreated, line 11: suggestion/ Since we're adding an integration test, referring to RUMViewScope—which is an implementation detail—feels out of place. What we're truly testing here is the flow of setting RUM attributes and how they are inherited by different events.

A more descriptive and appropriately scoped name could be:
- `RUMAttributesIntegrationTests`
- `RUMAttributeInheritanceTests`
- `RUMAttributesFlowTests`

DatadogRUM/Tests/Instrumentation/Views/RUMViewsHandlerTests.swift
```diff
@@ -28,8 +28,7 @@ class RUMViewsHandlerTests: XCTestCase {
 
     func testGivenAcceptingPredicate_whenViewDidAppear_itStartsRUMView() throws {
         let viewName: String = .mockRandom()
-        let viewControllerClassName: String = .mockRandom()
-        let view = createMockView(viewControllerClassName: viewControllerClassName)
+        let view = ViewControllerMock()
```
ncreated, line 31: question/ I’m not entirely clear on this change or if it’s missing the removal of the `createMockView()` implementation. From the PR description, I understand that [`createMockView()`](https://github.com/DataDog/dd-sdk-ios/blob/832f47dd8e9df17880750aa50b49a5f04e277d48/DatadogRUM/Tests/Mocks/RUMFeatureMocks.swift#L913-L928) is causing issues in Xcode 16. Ideally, I would expect to see that method either fixed or completely replaced with an alternative approach. While we’ve replaced it in a few places, the original problematic method still appears to be present.

ncreated (changes requested):
Great effort in working on the integration testing 👍👌. It's clear you've put thought into it 🏅.

However, this PR doesn’t address the underlying issue outlined in `RUM-3029`, which still exists. I’ve highlighted the details of the problem in a few comments throughout the PR.

Datadog/IntegrationUnitTests/RUM/RUMViewScopeIntegrationTests.swift
simaoseica-dd: Yes, it is not a behaviour in the product but the idea of the test is to give us confidence we don't have regressions on this regard. 
The attributes are shared through commands for all types (actions, resources, views,..) and so it is very easy to have wrong assumptions by not knowing the playbook.

simaoseica-dd: As suggested I added tests to validate the addition of global attributes to actions, resources and errors

Datadog/IntegrationUnitTests/RUM/RUMAttributesIntegrationTests.swift
```diff
@@ -0,0 +1,412 @@
...
+        let initialAttributes: [AttributeKey: AttributeValue] = ["sameKey": "value1", "key2": "value2"]
+        let viewName = "MyView"
+        RUM.enable(with: rumConfig, in: core)
+
+        let monitor = RUMMonitor.shared(in: core)
+
+        // When
+        monitor.addAttribute(forKey: "sameKey", value: "globalValue")
+        monitor.startView(key: "key", name: viewName, attributes: initialAttributes)
+        //monitor.stopView(key: "key")
```
simaoseica-dd, line 151: The logic behind this test is being discussed

Datadog/IntegrationUnitTests/RUM/RUMAttributesIntegrationTests.swift
```diff
@@ -0,0 +1,450 @@
...
+    var numberOfAttributes: Int { context?.contextInfo.count ?? 0 }
+
+    func attribute<T: Equatable>(forKey key: String) -> T? { (context?.contextInfo[key] as? AnyCodable)?.value as? T }
+}
+
+private extension RUMActionEvent {
+    var numberOfAttributes: Int { context?.contextInfo.count ?? 0 }
+
+    func attribute<T: Equatable>(forKey key: String) -> T? { (context?.contextInfo[key] as? AnyCodable)?.value as? T }
+}
```
ncreated, line 438: 💯 nice

ncreated (dismissed):
Well done 🚀 - it looks good and the coverage is really nice.

DatadogRUM/Sources/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -467,7 +465,15 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
...
-            attributes.merge(rumCommandAttributes: command.attributes)
+            attributes.merge(rumCommandAttributes: command.globalAttributes)
+
+            switch command {
+            case let command as RUMStartViewCommand where identity == command.identity:
+                attributes.merge(rumCommandAttributes: command.attributes)
+            case let command as RUMStopViewCommand where identity == command.identity:
+                attributes.merge(rumCommandAttributes: command.attributes)
+            default: break
+            }
```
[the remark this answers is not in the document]
simaoseica-dd: This logic safeguards local attributes from being updated by commands originating from other Views, external to the current `RUMViewScope` (`identity == command.identity`). 
These 2 commands may be executed in other `RUMViewScope`s due to automated processes that start and stop views.

DatadogRUM/Sources/RUMMonitor/Scopes/RUMResourceScope.swift
```diff
@@ -89,6 +87,10 @@ internal class RUMResourceScope: RUMScope {
     // MARK: - RUMScope
 
     func process(command: RUMCommand, context: DatadogContext, writer: Writer) -> Bool {
+        self.attributes = self.attributes
+            .merging(command.globalAttributes, uniquingKeysWith: { $1 })
+            .merging(command.attributes, uniquingKeysWith: { $1 })
+
```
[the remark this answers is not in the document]
simaoseica-dd: It is redundant. Good catch

Datadog/IntegrationUnitTests/RUM/RUMAttributesIntegrationTests.swift
```diff
@@ -0,0 +1,450 @@
...
+
+        // Then
+        let session = try RUMSessionMatcher
+            .groupMatchersBySessions(try core.waitAndReturnRUMEventMatchers())
+            .takeSingle()
+
+        // It should have the `ApplicationLaunchView` and `MyView` views
+        XCTAssertEqual(session.views.count, 2)
+
+        let applicationView = try XCTUnwrap(session.views.first(where: { $0.name == RUMOffViewEventsHandlingRule.Constants.applicationLaunchViewName }))
```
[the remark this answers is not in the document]
simaoseica-dd: 👌

DatadogRUM/Sources/RUMMonitor/Scopes/RUMViewScope.swift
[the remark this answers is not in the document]
ncreated: I see 👍 . Indeed `identity == command.identity` is not possible with `RUMViewScopePropagatableAttributes`.

ncreated (approved):

PR #2147, opened 2024-12-13
fix: Don't send log spam for PerformanceMetric updates

fuzzybinary:
### What and why?

PerformanceMetric events are triggered automatically by some of the cross platform frameworks, and can result in log spam as views transition or if no view is active. Fix it so that the performance metric event doesn't send log spam.

refs: RUM-6138

### Review checklist

### DIFF
--- a/DatadogRUM/Tests/RUMMonitor/Scopes/RUMSessionScopeTests.swift
+++ b/DatadogRUM/Tests/RUMMonitor/Scopes/RUMSessionScopeTests.swift
@@ -598,7 +598,7 @@ class RUMSessionScopeTests: XCTestCase {
 
     // MARK: - Usage
 
-    func testGivenSessionWithNoActiveScope_whenReceivingRUMCommandOtherThanKeepSessionAliveCommand_itLogsWarning() throws {
+    func testGivenSessionWithNoActiveScope_whenReceivingRUMCommand_itLogsWarning() throws {
         func recordWarningOnReceiving(command: RUMCommand) -> String? {
             // Given
             let scope: RUMSessionScope = .mockWith(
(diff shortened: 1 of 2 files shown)


DatadogRUM/Tests/RUMMonitor/Scopes/RUMSessionScopeTests.swift
```diff
@@ -628,9 +628,55 @@ class RUMSessionScopeTests: XCTestCase {
...
+            let dd = DD.mockWith(logger: CoreLoggerMock())
+            defer { dd.reset() }
+
+            // When
+            _ = scope.process(command: command, context: context, writer: writer)
+
+            // Then
+            XCTAssertEqual(scope.viewScopes.count, 0)
+            return dd.logger.warnLog?.message
+        }
```
ncreated, line 676: nit/ The `recordWarningOnReceiving()` seems to be copy & pasted from previous test. We could reuse it this way:
```swift
func testGivenSessionWithNoActiveScope_whenReceivingCommands_itDoesNotLogWarning() throws {
    func recordWarningOnReceiving(command: RUMCommand) -> String? {
       // ...
    }

    // When
    let keepAliveCommand = RUMUpdatePerformanceMetric(...)
    let keepAliveLog = ...

    // Then
    XCTAssertNil(keepAliveLog, ...)

    // When
    let perfMetricCommand = RUMKeepSessionAliveCommand(...)
    let perfMetricLog = ...

    // Then
    XCTAssertNil(keepAliveLog, ...)
}
```

ncreated (approved):
Looks good 👍 👌

DatadogRUM/Tests/RUMMonitor/Scopes/RUMSessionScopeTests.swift
fuzzybinary: It is copied yes, and I guess it's a style thing. I prefer a smaller number of assertions per test, but I'm good with combining them.  Maybe to make it cleaner or more "single responsibility" we could do:
```swift
let quietCommands = [ RumUpdatePerformanceMetric(...), RUMKeepSessionAliveCommand(...) ]

for command in quietCommands {
  let log = ...
  XCTAssertNil(log, ...)
}
```

What do you think?

ncreated: Sounds good 👌👍. In my opinion, the key is maintaining test clarity and keeping the number of tests manageable to avoid bloating this file. The loop should handle it 👍. `RUM*ScopeTests` often grow large since `RUM*Scopes` mix dozens of responsibilities 😞.

ncreated (approved):

PR #2154, opened 2024-12-18
RUM-7205 Double full snapshot bug

maciejburda:
### What and why?

This PR is essentially reverting https://github.com/DataDog/dd-sdk-ios/pull/1567. Assumption made was wrong. We can expect view that from the wireframe perspective have a different root view (it can be done by fully covering the root with nontransparent view, which will remove the root during flattening process).

The reverted code was fixing some minor edge cases in React Native, although it has a serious penalty of creating double fullsnapshots in others.

### Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - [IMPROVEMENT] Add Datadog Configuration `backgroundTasksEnabled` ObjC API. See [#2148][]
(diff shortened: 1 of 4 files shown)


maxep (approved):

mariedm (approved):

ncreated (approved):
Good finding, good explanation, good trade-off 💪

PR #2160, opened 2024-12-20
RUM-7103 feat: Add API for classifying resources in TTNS metric

ncreated:
### What and why?

📦 ⏱️ This PR introduces an API to customize the classification of resources for the **Time-to-Network-Settled (TTNS)** metric using a predicate-based strategy. The predicate allows users to define which resources should be considered for the TTNS calculation based on their own logic. 

This provides more flexibility in how the metric is computed, enabling users to customize resource classification for different use cases.

**Previously**, resources were included in TTNS calculation solely based on a fixed `100ms` threshold after the view starts:

prev

**Now**, users can customize resource classification by implementing the `NetworkSettledResourcePredicate` protocol, allowing for flexible inclusion or exclusion of resources. For example:

Screenshot 2024-12-20 at 12 09 00

Users can also use the default `TimeBasedTTNSResourcePredicate` with adjusting its `threshold`.

### How?

This PR introduces the `NetworkSettledResourcePredicate` protocol, which allows users to implement their own classification logic for TTNS resources:

```swift
public protocol NetworkSettledResourcePredicate {
    func isInitialResource(resource: TTNSResourceParams) -> Bool
}
```

The default implementation, `TimeBasedTTNSResourcePredicate`, classifies resources based on the time elapsed since the view started (default threshold of `100ms`).

The `TTNSMetric` class is updated with the predicate, making the logic for calculating TTNS more flexible and customizable.

Integration and unit tests have been updated to validate the correct behavior of the default predicate-based approach. An integration test was added to test integration of custom predicate.

### Review checklist


simaoseica-dd (approved):
Great effort on the documentation and illustration 👌

PR #2162, opened 2024-12-30
[APMTI-2602] Move the agent_psr from the root of the span to the metrics map

Reallnas:
### What and why?

This PR moves the `_dd.agent_psr` field from the span's root to the metrics map. 
Also removes the logic related to _dd at the root of the span as it was only used for this field and also because this field shouldn't be set at the root.

When looking to use a stricter json schema (see ticket) for the json payload parser, we noticed that some spans would be rejected due to invalid fields that didn't comply with the established schema. This is one of those spans.

This field was added in PR #1259 but I don't know how this was tested.
I am not familiar with this repository and swift. Feedback is welcome.

### How?

### Review checklist

### DIFF
--- a/DatadogCore/Tests/Datadog/TracerTests.swift
+++ b/DatadogCore/Tests/Datadog/TracerTests.swift
@@ -65,7 +65,6 @@ class TracerTests: XCTestCase {
         {
           "spans": [
             {
-              "_dd.agent_psr": 1,
               "trace_id": "64",
               "span_id": "64",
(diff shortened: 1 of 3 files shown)


ncreated (approved):
👍 Looks good and it is very complete contribution, thanks @Reallnas 👌 

I will first merge this to internal branch so we can run CI and do minor adjustments if necessary. Then, I will open another PR to merge it to `develop`. Will keep you posted 👍.

PR #2163, opened 2024-12-31
RUM-768: Improve telemetry over dropped events due to size limits

simaoseica-dd:
### What and why?

This PR provides detailed information about dropped events when data blocks exceed the limit.
By adding these details, we can gain better insights into the causes of event drops, enabling more effective troubleshooting and optimization.

### How?

The cause of dropped events during writes has been identified and addressed in the [PR](https://github.com/DataDog/dd-sdk-ios/pull/1922/), available since version `2.17.0`.
The primary issue was insufficient storage space on the user’s device, leading to errors such as: 
> The file couldn’t be saved because there isn’t enough space. ([telemetry logs](https://dd-rum-telemetry.datadoghq.com/logs?query=status%3Aerror%20source%3Aios%20%22Failed%20to%20write%22%20%22bytes%20to%20file%22&agg_m=count&agg_m_source=base&agg_t=count&cols=%40org_id%2Cversion&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1734261238455&to_ts=1735557238455&live=true
))

For reading data blocks, this PR enhances logging to provide deeper insights into dropped events. The logs now capture:
- Track name
- Length of the dropped block

### Review checklist


DatadogCore/Sources/Core/Storage/Storage+TLV.swift
```diff
@@ -7,7 +7,7 @@
 import Foundation
 
 /// Default max data length in TLV block for batch file (safety check) - 10 MB
-internal let MAX_DATA_LENGTH: UInt64 = 10.MB.asUInt64()
+internal let MAX_DATA_LENGTH: TLVBlockSize = 10.MB.asUInt32()
```
ncreated, line 10: nit/ 🏕️ Since we are touching this constant, could we take the opportunity to rename it to follow Swift's naming conventions 🙏? The all-uppercase, snake_case format feels out of place, and the current name is a bit ambiguous. Something like `maxTLVDataLength` might be more descriptive and idiomatic.

DatadogCore/Sources/PerformancePreset.swift
```diff
@@ -11,11 +11,11 @@ internal protocol StoragePerformancePreset {
...
-    var maxFileSize: UInt64 { get }
+    var maxFileSize: TLVBlockSize { get }
     /// Maximum size of data directory (in bytes).
     /// Each feature uses separate directory.
     /// If this size is exceeded, the oldest files are deleted until this limit is met again.
-    var maxDirectorySize: UInt64 { get }
+    var maxDirectorySize: TLVBlockSize { get }
     /// Maximum age qualifying given file for reuse (in seconds).
     /// If recently used file is younger than this, it is reused - otherwise: new file is created.
     var maxFileAgeForWrite: TimeInterval { get }
```
ncreated, line 21: blocking/ It seems incorrect to represent these values using `TLVBlockSize`, since they refer to various file and directory size limits rather than strictly to TLV blocks. For instance, a directory size has nothing to do with TLV. If we want to use `UInt32` here, it should be explicit rather than referencing `TLVBlockSize`.

ncreated (changes requested):
Great work on refining the codebase and spotting all these typos 👏 ! I left one blocking comment on using `TLVBlockSize` as a unit - let me know WDYT.

DatadogCore/Sources/Core/Storage/Storage+TLV.swift
simaoseica-dd: Good point! Renamed

DatadogCore/Sources/PerformancePreset.swift
simaoseica-dd: I misjudged this change, and you’re absolutely right. Thank you for pointing this out.

PR #2165, opened 2024-12-31
RUM-7823 feat: Add API for classifying "last interaction" in ITNV metric

ncreated:
### What and why?

📦 ⏱️ This PR introduces an API to customize the classification of interactions for the **Interaction-To-Next-View (ITNV)** metric using a predicate-based strategy. The predicate enables users to define which RUM actions in the previous view should be considered as the "last interaction" for calculating ITNV value (how long it takes to reach the next view).

This approach is similar to the TTNS metric API introduced in #2160.

**Previously**, ITNV was computed by fixed rules, always selecting the last "tap," "click," or "swipe" occurring within 3 seconds before the next view started. For example:

Screenshot 2024-12-31 at 14 06 52

**Now**, with the `NextViewActionPredicate` protocol, developers can implement custom logic to classify the "last interaction." For example, they might choose actions based on a certain name:

Screenshot 2024-12-31 at 14 06 57

Users can also use the default `TimeBasedITNVActionPredicate` and adjust the `maxTimeToNextView` threshold.

### How?

This PR introduces the `NextViewActionPredicate` protocol, enabling custom classification logic:
```swift
public protocol NextViewActionPredicate {
    func isLastAction(from actionParams: ITNVActionParams) -> Bool
}
```
The default implementation, `TimeBasedITNVActionPredicate`, classifies an action as the "last interaction" if:
- It occurs within a configurable time threshold (default: 3s) before the next view starts.
- Its action type is one of: "tap", "click", or "swipe".

The `ITNVMetric` class now consumes this predicate, making ITNV calculations flexible and customizable. If no custom predicate is provided, the default time-based logic applies.

Lastly, integration and unit tests confirm the correct behavior of both the default and custom predicates.

### Review checklist

### DIFF
--- a/DatadogRUM/Sources/RUMMetrics/ITNVMetric.swift
+++ b/DatadogRUM/Sources/RUMMetrics/ITNVMetric.swift
@@ -8,104 +8,165 @@ import Foundation
 
 internal protocol ITNVMetricTracking {
     /// Tracks an action in a given view.
-    /// Actions can be tracked even if the view is no longer active. For example, a "tap" can start in one view, but be tracked
-    /// through this method even after the next view is started.
+    ///
+    /// An action can start in one view but still be tracked through this method even after the next view begins.
+    /// For instance, a tap might begin in one view but end after the next view is already displayed.
+    ///
     /// - Parameters:
-    ///   - startTime: The start time of the action.
-    ///   - endTime: The end time of the action.
-    ///   - actionType: The type of the user action.
-    ///   - viewID: The ID of the view where the action occurred.
-    func trackAction(startTime: Date, endTime: Date, actionType: RUMActionType, in viewID: RUMUUID)
+    ///   - startTime: The time when the action started (device time, no NTP offset).
+    ///   - endTime: The time when the action ended (device time, no NTP offset).
+    ///   - name: A string identifying the action (e.g., button label).
+    ///   - type: The specific type of the action (e.g., tap, swipe).
+    ///   - viewID: The unique identifier of the view where the action was initiated.
+    func trackAction(startTime: Date, endTime: Date, name: String, type: RUMActionType, in viewID: RUMUUID)
 
     /// Tracks the start of a view.
-    /// From this moment, calls to `value(for:)` for this `viewID` will return the value of ITNV.
+    ///
+    /// From this point forward, calling `value(for:)` with this `viewID` can return an ITNV value, until the view is completed.
+    ///
     /// - Parameters:
-    ///   - startTime: The timestamp when the view starts.
-    ///   - viewID: The ID of the view that has just started.
-    func trackViewStart(at startTime: Date, viewID: RUMUUID)
+    ///   - startTime: The time when the view becomes active (device time, no NTP offset).
+    ///   - name: A human-readable name of the view.
+    ///   - viewID: The unique identifier of the newly started view.
+    func trackViewStart(at startTime: Date, name: String, viewID: RUMUUID)
 
     /// Marks the completion of a view.
-    /// Indicates that the view event will no longer be updated and no more calls to `value(for:)` for this `viewID` will be made.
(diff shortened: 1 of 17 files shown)


DatadogRUM/Sources/RUMMetrics/ITNVMetric.swift
```diff
@@ -8,104 +8,165 @@ import Foundation
...
-            return nil // No action recorded in the previous view
+        guard var previousView = viewsByID[previousViewID] else {
+            return nil // The previous view has been removed.
+        }
+        defer { viewsByID[previousViewID] = previousView } // Update the stored view after modifications.
+
+        // Find the most recent action in the previous view that the predicate accepts as "last interaction."
+        let lastAction = previousView.actions
+            .reversed()
+            .first { action in
```
simaoseica-dd, line 135: question/ Even though the `.reversed` is constant and optimized, why `first(where:)` instead of `.last(where:)`? 
In terms of readability it depicts better that we are trying to get the `last` element based on the condition.

DatadogRUM/Sources/RUMMetrics/NextViewActionPredicate.swift
```diff
@@ -0,0 +1,86 @@
...
+/// - The method should always return the same result for identical input parameters to ensure consistent ITNV calculation.
+public protocol NextViewActionPredicate {
+    /// Determines whether the provided action should be classified as the "last interaction" in the previous view for ITNV calculation.
+    ///
+    /// This method is invoked in reverse chronological order for all actions in the previous view. Once `true` is returned,
+    /// the iteration stops, and the accepted action defines the starting point for the ITNV metric.
+    ///
+    /// - Parameter action: The parameters of the action (type, name, time to next view, and next view name).
+    /// - Returns: `true` if this action is the "last interaction" for ITNV, `false` otherwise.
+    func isLastAction(action: ITNVActionParams) -> Bool
```
simaoseica-dd, line 44: nitpicking/ Can the naming be improved? `func isLastAction(from actionParams: ITNVActionParams) -> Bool`?

DatadogRUM/Sources/RUMMetrics/NextViewActionPredicate.swift
```diff
@@ -0,0 +1,86 @@
...
+    public init(maxTimeToNextView: TimeInterval = TimeBasedITNVActionPredicate.defaultMaxTimeToNextView) {
+        self.maxTimeToNextView = maxTimeToNextView
+    }
+
+    /// Determines if the provided action should be considered the "last interaction" for ITNV, based on its action type and timing.
+    ///
+    /// - Parameter action: The parameters of the action (type, name, time to next view, and next view name).
+    /// - Returns: `true` if the action’s `timeToNextView` is within `maxTimeToNextView` and its type is `tap`, `click`, or `swipe`;
+    ///            otherwise, `false`.
+    public func isLastAction(action: ITNVActionParams) -> Bool {
```
simaoseica-dd, line 72: nitpicking/ same naming question here

simaoseica-dd (approved):
Well done! 🚀

DatadogRUM/Sources/RUMMetrics/NextViewActionPredicate.swift
ncreated: 👌  I can see how the `from` label might read more naturally in this context. However, we’ve established a similar naming pattern for the TTNS API ([`isInitialResource(resource:)`](https://github.com/DataDog/dd-sdk-ios/blob/2f13c4ced52bd9c528b4aee26b0938f369e1f470/DatadogRUM/Sources/RUMMetrics/NetworkSettledResourcePredicate.swift#L33)), so renaming just the ITNV method would create an inconsistency between these two predicate interfaces.

Because I like the proposed change, I will rename both in this PR.

PR #2167, opened 2025-01-02
RUM-3029: Fix flaky test while handling a stopped session

simaoseica-dd:
### What and why?

This PR resolves a flaky test issue that occurred when new commands were executed after stopping a session.

### How?

- The test validates that a View scope from a stopped session is immutable from other commands (e.g. adding new global attributes). 
- The test starts two sessions and asserts the attributes added to the view events.  

### Review checklist


ncreated (approved):
🏅💪

PR #2169, opened 2025-01-08
RUM-7937 [SR] Use Deterministic Random Wireframe ID

maxep:
### What and why?

Session Replay of SwiftUI is generating staling views in replays, this is due to duplicate wireframe IDs between UIKit recorder and SwiftUI. UIKit recorders use incremental wireframe ids while SwiftUI is based on a `seed` value of the content, these value can overlap and create duplicates.

### How?

To avoid overlaps, we now use a seeded random generator to generate deterministic IDs base of the content's `seed` value.
The deterministic generator implements the [xoshiro256**](https://prng.di.unimi.it/xoshiro256starstar.c) generator for its speed and low memory footprint.

### Review checklist


maciejburda (approved):
Nice find and witty solution! 🙇

PR #2170, opened 2025-01-09
RUM-7936 [SR] Fix SwiftUI Placeholder with Feature Flag

mariedm:
### What and why?

In pure SwiftUI apps, the feature flag for enabling SwiftUI support in Session Replay was not being respected. The issue stemmed from the recorder only checking for `UIHostingController` to determine if the view hierarchy belonged to SwiftUI, which doesn’t cover all cases.

### How?

- Reused the `Bundle` extension used in RUM view predicates and moved it to the internal module.
- Introduced a `FrameworkUtils` enum to determine if a class belongs to UIKit or SwiftUI. This utility now combines class type checks (this also covers the use case of custom View Controllers) and Bundle detection.

### Review checklist


DatadogInternal/Sources/Utils/SwiftExtensions.swift
```diff
@@ -115,3 +115,17 @@ extension Array {
         0 <= index && index < count ? self[index] : nil
     }
 }
+
+// MARK: - Bundle
+
+extension Bundle {
```
maxep, line 121: /nit I think we should us the `DatadogExtended` pattern when defining `public` extension.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/ViewTreeRecordingContext.swift
```diff
@@ -47,16 +49,23 @@ internal extension ViewTreeRecordingContext {
...
                     self = .safari
-                case is AnyUIHostingViewController:
-                    self = .swiftUI
                 default:
+                    if FrameworkUtils.isSwiftUI(class: type(of: viewController)) {
+                        self = .swiftUI
+                        return
+                    }
                     self = .other
                 }
```
maxep, line 70: /suggestion Lets replace the `switch` statement with a `if else` for readability 👍

maxep (dismissed):
Nice finding and great coverage 👍  I have left minor comments!

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/ViewTreeRecordingContext.swift
```diff
@@ -9,6 +9,8 @@ import Foundation
 import SafariServices
 import SwiftUI
 import WebKit
+@_spi(Internal)
+import DatadogInternal
```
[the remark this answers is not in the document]
mariedm: It's a leftover, good point.

DatadogSessionReplay/Sources/Recorder/Utilities/FrameworkUtils.swift
```diff
@@ -0,0 +1,25 @@
...
+            return `class`.isSubclass(of: UIHostingController<AnyView>.self) || Bundle(for: `class`).isSwiftUI
+        }
+        return Bundle(for: `class`).isSwiftUI
+    }
+
+    /// Checks if the given class belongs to the UIKit framework.
+    static func isUIKit(`class`: AnyClass) -> Bool {
+        /// Supports both native and custom subclasses
+        return `class` is UIViewController.Type || Bundle(for: `class`).isUIKit
+    }
```
maxep, line 24: /question Why not using the Bundle directly?

mariedm: Because it's not working for custom view controllers

PR #2172, opened 2025-01-10
RUM-7795 Anonymous RUM Identifier

maciejburda:
### What and why?

Adds capability of tracking anonymous id (enabled by default). This data allows linking sessions coming from the same device.

### How?

Following proposal from this [RFC (internal)](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/4447830354/RFC+-+Session+Continuity+in+SDKs) it reuses data store mechanism, and based on configuration it generates and reuses stored identifier.

This identifier is attached to `UserInfo` object (schema already updated), which is enriching other SDK events.

> NOTE: The identifier is read asynchronously from a file, which means RUM events created immediately after calling `RUM.enable()` will not include it. However, this is not an issue since the `anonymous_id` belongs to the Session entity. As long as the `anonymous_id` is present in at least in one event, it is sufficient for linking.

Here is corresponding PR for Android:
https://github.com/DataDog/dd-sdk-android/pull/2487

### Review checklist


DatadogRUM/Sources/Integrations/AnonymousIdManager.swift
```diff
@@ -0,0 +1,42 @@
...
+        uuidGenerator: RUMUUIDGenerator
+    ) {
+        self.featureScope = featureScope
+        self.uuidGenerator = uuidGenerator
+    }
+
+    func manageAnonymousId(shouldTrack: Bool) {
+        if shouldTrack {
+            featureScope.rumDataStore.value(forKey: .anonymousId) { [featureScope, uuidGenerator] (anonymousId: String?) in
+                if let anonymousId = anonymousId {
```
mariedm, line 29: nit
```suggestion
                if let anonymousId {
```

DatadogRUM/Sources/RUMConfiguration.swift
```diff
@@ -372,6 +380,7 @@ extension RUM.Configuration {
         longTaskEventMapper: RUM.LongTaskEventMapper? = nil,
         onSessionStart: RUM.SessionListener? = nil,
         customEndpoint: URL? = nil,
+        trackAnonymousUser: Bool = true,
```
mariedm, line 383: nit: You might want to run `make api-surface` since it updates a public API.

DatadogRUM/Tests/Integrations/AnonymousIdentifierManagerTests.swift
```diff
@@ -0,0 +1,59 @@
...
+        XCTAssertNotNil(featureScopeMock.anonymousId)
+    }
+
+    func testWhenShouldNotTrack_itClearsAnonymousID() {
+        anonymousIdentifierManager.manageAnonymousId(shouldTrack: false)
+        XCTAssertNil(featureScopeMock.anonymousId)
+
+        featureScopeMock.set(anonymousId: "test")
+
+        anonymousIdentifierManager.manageAnonymousId(shouldTrack: false)
```
mariedm, line 44: Should we test when the `anonymousId` is set but `shouldTrack` is `false`? In other words, should we delete this line in order to make sure that the `anonymousId` is `nil` even without calling again `anonymousIdentifierManager.manageAnonymousId(shouldTrack: false)`.

mariedm (approved):
LGMT! Only a few minor comments.

simaoseica-dd (dismissed):

DatadogRUM/Sources/RUMConfiguration.swift
```diff
@@ -210,6 +210,14 @@ extension RUM {
         /// Default: `nil`.
         public var customEndpoint: URL?
 
+        /// Determines whether automatic tracking of anonymous user id is enabled.
+        ///
+        /// If enabled, the SDK will automatically track the anonymous user id for each RUM Session, which
+        /// is used to link RUM Sessions belonging to the same anonyomus user.
+        ///
+        /// Default: `true`.
+        public var trackAnonymousUser: Bool
```
ncreated, line 219: suggestion/ I'd consider renaming `trackAnonymousUser` to `trackAnonymousUserID` to emphasize that we’re persisting and using only ID, **rather than tracking any personal user data**. This explicit naming could also helps clarify that the data is purely non-personal, which might avoid confusion and potential privacy concerns. WDYT?

DatadogRUM/Sources/RUMConfiguration.swift
```diff
@@ -210,6 +210,14 @@ extension RUM {
         /// Default: `nil`.
         public var customEndpoint: URL?
 
+        /// Determines whether automatic tracking of anonymous user id is enabled.
+        ///
+        /// If enabled, the SDK will automatically track the anonymous user id for each RUM Session, which
+        /// is used to link RUM Sessions belonging to the same anonyomus user.
```
ncreated, line 216: suggestion/ Let's be more specific about how it works, so we avoid any confusion and we raise no privacy concerns:

```suggestion
        /// Determines whether automatic tracking of an anonymous user ID is enabled.
        ///
        /// When enabled, the SDK generates a unique, non-personal anonymous user ID that is persisted across
        /// app launches. This ID will be attached to each RUM Session, allowing you to link sessions
        /// originating from the same user/device without collecting personal data.
```

ncreated (changes requested):
It looks great overall 👌 - I only left suggestions on naming and documenting the public API.

Requesting a change tho, to add [integration unit tests](https://github.com/DataDog/dd-sdk-ios/tree/develop/Datadog/IntegrationUnitTests/RUM) for this feature. Using real instance of the SDK + proxy and asserting that anonymous ID is generated and inserted into events of two RUM sessions started in two instances of the SDK. I can help and pair on this 🙌

DatadogRUM/Sources/RUMConfiguration.swift
maciejburda: Valid point, but AFAIK that's the API that was used in the browser.

Let me check if we can make an alignment.

If not - I'll make the docs more specific.

maciejburda: It's already out on browser under the same name:
https://docs.datadoghq.com/real_user_monitoring/browser/setup/client/?tab=rum#configuration

DatadogRUM/Tests/Integrations/AnonymousIdentifierManagerTests.swift
maciejburda: Added integration test covering this scenario.

Datadog/IntegrationUnitTests/RUM/AnonymousIdentifierTests.swift
```diff
@@ -0,0 +1,111 @@
...
+    func test_itReusesAnonymousIdentifierOnSubsequentSessions() throws {
+        enableRUM(trackAnonymousUser: true)
+
+        let session1 = try RUMSessionMatcher
+            .groupMatchersBySessions(try core.waitAndReturnRUMEventMatchers())
+            .takeSingle()
+
+        assertAnonymousIdentifier(isSet: true, in: session1)
+        let anonymousIdFromSession1 = session1.views.last?.viewEvents.first?.usr?.anonymousId
+        XCTAssertNotNil(anonymousIdFromSession1)
```
simaoseica-dd, line 55: This assertion is repeated in the above `assertAnonymousIdentifier()`

Datadog/IntegrationUnitTests/RUM/AnonymousIdentifierTests.swift
```diff
@@ -0,0 +1,111 @@
...
+        core = DatadogCoreProxy()
+    }
+
+    private func enableRUM(trackAnonymousUser: Bool) {
+        let rumConfig = RUM.Configuration(applicationID: .mockAny(), trackAnonymousUser: trackAnonymousUser)
+        RUM.enable(with: rumConfig, in: core)
+        // Needs to flush datastore on the caller thread to ensure anonymousId was read from the file.
+        core.scope(for: RUMFeature.self).dataStore.flush()
+        // Create new view that should consist of anonymousId is present.
+        RUMMonitor.shared(in: core).startView(key: .mockRandom(), name: .mockRandom())
```
simaoseica-dd, line 100: question: The initial `ApplicationLaunchView` would be enough right? 
Thinking if you are adding more views, the assert below should assert all of the Views:
`session.views.forEach { XCTAssertNotNil($0.viewEvents.first?.usr?.anonymousId) }`

DatadogInternal/Sources/DatadogCoreProtocol.swift
```diff
@@ -221,8 +221,15 @@ extension BaggageSharing {
     }
 }
 
+/// Provides ability to set or clear the anonymous identifier needed for session linking.
+public protocol AnonymousIdentifiable {
```
simaoseica-dd, line 225: Can this naming be improved? It seems that it is applied to an entity that is `Identifiable` which might not be true. 
`AnonymousIDManager`?

DatadogInternal/Sources/DatadogCoreProtocol.swift
```diff
@@ -221,8 +221,15 @@ extension BaggageSharing {
...
+    /// Sets the anonymous identifier.
+    /// - Parameter anonymousId: The anonymous id to be set. When `nil` it will clear the current anonymous id.
+    func set(anonymousId: String?)
```
simaoseica-dd, line 228: question: Would it be too much to have an explicit `reset` function?

simaoseica-dd (commented):
Can the the anonymous id naming be consistent?
There is `AnonymousIdentifier`, `AnonymousId` and `AnonymousID` in the code.

Datadog/IntegrationUnitTests/RUM/AnonymousIdentifierTests.swift
maciejburda: Yeah - this is important detail of this change.

I added a note in the PR description:
```NOTE: The identifier is read asynchronously from a file, which means RUM events created immediately after calling RUM.enable() will not include it. However, this is not an issue since the anonymous_id belongs to the Session entity. As long as the anonymous_id is present in at least in one event, it is sufficient for linking.```

In this case we're fine with a data store race condition, because synchronisation would impact SDK initialisation time. In the end we need anonymousID to be present in at least one of the event to get propagated to the session.

Hence in the assertion we flush datastore and start mock view to make sure that anonymousId will be attached to it.

Probably it could be fixed on the system level by introducing `SetUser` event, but for now it's out of scope.

File: DatadogInternal/Sources/DatadogCoreProtocol.swift (hunk 8 above)
Comment (maciejburda) [reply DatadogInternal/Sources/DatadogCoreProtocol.swift:225]: Good callout. I changed it to `AnonymousIDManaging` to keep convention with other protocol in the area.

File: DatadogInternal/Sources/DatadogCoreProtocol.swift (hunk 9 above)
Comment (maciejburda) [reply DatadogInternal/Sources/DatadogCoreProtocol.swift:228]: Reset is not really common behaviour. Right now it essentially happens only if from one version to the other customer decides to disable `trackAnonymousUser` API.

The idea was to provide public setter / private getter.

File: Datadog/IntegrationUnitTests/RUM/AnonymousIdentifierTests.swift (hunk 7 above)
Comment (simaoseica-dd) [reply Datadog/IntegrationUnitTests/RUM/AnonymousIdentifierTests.swift:96]: Thank you for the detailed explanation!

**Review by simaoseica-dd — APPROVED** [review]
LGTM! 👌

File: DatadogRUM/Sources/Integrations/AnonymousIdentifierManager.swift [hunk 10]
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    func manageAnonymousIdentifier(shouldTrack: Bool) {
+        if shouldTrack {
+            featureScope.rumDataStore.value(forKey: .anonymousId) { [weak self] (anonymousId: String?) in
+                if let anonymousId {
+                    self?.featureScope.set(anonymousId: anonymousId)
+                } else {
+                    let anonymousId = self?.uuidGenerator.generateUnique().toRUMDataFormat
+                    self?.featureScope.rumDataStore.setValue(anonymousId, forKey: .anonymousId)
```
mariedm, line 33: We don't need to update the `rumDataStore` as well if the `anonymousId` already exists?

mariedm (approved):

Datadog/IntegrationUnitTests/RUM/AnonymousIdentifierTests.swift
```diff
@@ -0,0 +1,110 @@
...
+
+        enableRUM(trackAnonymousUser: true)
+
+        let session2 = try RUMSessionMatcher
+            .groupMatchersBySessions(try core.waitAndReturnRUMEventMatchers())
+            .takeSingle()
+
+        assertAnonymousIdentifier(isSet: true, in: session2)
+        let anonymousId2 = session2.views.last?.viewEvents.first?.usr?.anonymousId
+        XCTAssertEqual(anonymousId1, anonymousId1)
```
simaoseica-dd, line 66: typo here. Always asserts.

DatadogCore/Tests/Datadog/DatadogCore/DatadogCoreTests.swift
```diff
@@ -330,4 +330,87 @@ class DatadogCoreTests: XCTestCase {
         core.flush()
         XCTAssertEqual(requestBuilderSpy.requestParameters.count, 0, "It should not send any request")
     }
+
+    func testItAppendsUserDataIfAnonymousIdentifierExists() {
+        let core = DatadogCore(
```
simaoseica-dd, line 335: nit/ `DatadogCoreProxy()`? and in the other tests too?

DatadogRUM/Sources/Integrations/AnonymousIdentifierManager.swift
simaoseica-dd: `let anonymousId = anonymousId ?? self?.uuidGenerator.generateUnique().toRUMDataFormat`
And then the `rumDataStore.setValue` and `featureScope.set` ?

simaoseica-dd (approved):
Some small nits 
👍

DatadogRUM/Sources/Integrations/AnonymousIdentifierManager.swift
maciejburda: We don't - as long as we can read it's fine.

Generally dataStore generates uuid once and then only reads it
featureScope notifies the SDK that we have anonymousID (regardless the path)

DatadogCore/Tests/Datadog/DatadogCore/DatadogCoreTests.swift
maciejburda: It's Core tests. I believe we should use core directly here. I was following existing pattern.

DatadogRUM/Sources/Integrations/AnonymousIdentifierManager.swift
mariedm: Got it!

mariedm (approved):

simaoseica-dd (approved):

ncreated (approved):

PR #2177, opened 2025-01-22
RUM-7691: Add `addAttributes` and `removeAttributes` APIs

simaoseica-dd:
### What and why?

This PR introduces a new interface to add and remove multiple attributes simultaneously. 
This enhancement reduces threading pressure on the Message Bus when handling multiple attributes at once.

### How?

The new interfaces to handle multiple attributes are:

```swift
func addAttributes(_ attributes: [AttributeKey: AttributeValue])

func removeAttributes(forKeys keys: [AttributeKey])
```

### Review checklist

### DIFF
--- a/DatadogObjc/Sources/RUM/RUM+objc.swift
+++ b/DatadogObjc/Sources/RUM/RUM+objc.swift
@@ -656,8 +656,18 @@ public class DDRUMMonitor: NSObject {
     }
 
     @objc
+    public func addAttributes(_ attributes: [String: Any]) {
+        swiftRUMMonitor.addAttributes(attributes.dd.swiftAttributes)
+    }
+
+    @objc
     public func removeAttribute(forKey key: String) {
         swiftRUMMonitor.removeAttribute(forKey: key)
+    }
+
+    @objc
+    public func removeAttributes(forKeys keys: [String]) {
+        swiftRUMMonitor.removeAttributes(forKeys: keys)
     }
 
     @objc
--- a/DatadogRUM/Sources/RUMMonitor/Monitor.swift
+++ b/DatadogRUM/Sources/RUMMonitor/Monitor.swift
@@ -105,11 +105,7 @@ internal class Monitor: RUMCommandSubscriber {
     private(set) var debugging: RUMDebugging? = nil
 
     @ReadWriteLock
-    private var attributes: [AttributeKey: AttributeValue] = [:] {
-        didSet {
-            fatalErrorContext.globalAttributes = attributes
-        }
-    }
+    private var attributes: [AttributeKey: AttributeValue] = [:]
 
     private let fatalErrorContext: FatalErrorContextNotifying
 
@@ -186,6 +182,10 @@ internal class Monitor: RUMCommandSubscriber {
 
         return mutableCommand
     }
+
+    private func didUpdateAttributes() {
+        fatalErrorContext.globalAttributes = attributes
+    }
 }
 
 /// Declares `Monitor` conformance to public `RUMMonitorProtocol`.
@@ -194,10 +194,24 @@ extension Monitor: RUMMonitorProtocol {
 
     func addAttribute(forKey key: AttributeKey, value: AttributeValue) {
         attributes[key] = value
+        self.didUpdateAttributes()
+    }
+
+    func addAttributes(_ attributes: [AttributeKey: AttributeValue]) {
+        self.attributes.merge(attributes) { $1 }
+        self.didUpdateAttributes()
     }
 
     func removeAttribute(forKey key: AttributeKey) {
         attributes[key] = nil
+        self.didUpdateAttributes()
+    }
+
+    func removeAttributes(forKeys keys: [AttributeKey]) {
+        _attributes.mutate { attributes in
+            keys.forEach { key in attributes.removeValue(forKey: key) }
+        }
+        self.didUpdateAttributes()
     }
 
     // MARK: - session
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - [FEATURE] Add Time To Network Setled metric in RUM. See [#2125][]
 - [FEATURE] Add Interaction To Next View metric in RUM. See [#2153][]
 - [FIX] Fix SwiftUI placeholder in Session Replay when Feature Flag is disabled. See [#2170][]
+- [IMPROVEMENT] Add `addAttributes` and `removeAttributes` APIs. See [#2177][]
 
 # 2.22.0 / 02-01-2025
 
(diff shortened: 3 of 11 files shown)


DatadogRUM/Sources/RUMMonitor/Monitor.swift
```diff
@@ -196,10 +196,18 @@ extension Monitor: RUMMonitorProtocol {
...
+        self.attributes.merge(attributes) { $1 }
+    }
+
     func removeAttribute(forKey key: AttributeKey) {
         attributes[key] = nil
     }
 
+    func removeAttributes(forKeys keys: [AttributeKey]) {
+        keys.forEach { self.attributes.removeValue(forKey: $0) }
+    }
```
ncreated, line 209: blocking/ There are two performance-related issues that need attention:

**Excessive lock acquisitions**: The `self.attributes` property uses the [`@ReadWriteLock`](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogInternal/Sources/Concurrency/ReadWriteLock.swift) property wrapper. As a result, each call to `attributes.removeValue(forKey:)` creates a separate critical section. For `N` attributes, this leads to `N` critical sections. This can be optimized by using [`.mutate {}`](https://github.com/DataDog/dd-sdk-ios/blob/c90c2e4b89ada210628102a6161b0ede6b08983d/DatadogInternal/Sources/Concurrency/ReadWriteLock.swift#L49) to remove multiple keys in a single transaction, reducing the overhead to just one critical section for the removal of all `N` attributes.

**Overloading `messageBus`**: The current implementation does not deliver the intended performance improvement on `messageBus`. Each update to `self.attributes` modifies the `fatalErrorContext`:
https://github.com/DataDog/dd-sdk-ios/blob/c90c2e4b89ada210628102a6161b0ede6b08983d/DatadogRUM/Sources/RUMMonitor/Monitor.swift#L108-L112

This triggers a dispatch of a message to the `messageBus` queue:
https://github.com/DataDog/dd-sdk-ios/blob/c90c2e4b89ada210628102a6161b0ede6b08983d/DatadogRUM/Sources/RUMMonitor/Scopes/FatalErrorContextNotifier.swift#L58-L67

Consequently, calling `removeAttributes(forKeys:)` for `N` keys results in `N` messages being dispatched, offering no improvement over calling `removeAttribute(forKey:)` separately for each key.

ncreated (changes requested):
It's good direction and nice work on integration tests 👌🚀. I left a blocking feedback on the performance of attributes removal.

DatadogRUM/Sources/RUMMonitor/Monitor.swift
simaoseica-dd: Thank you for the detailed review @ncreated. 

As suggested, the removal of attributes is now taking advantage of the [.mutate {}](https://github.com/DataDog/dd-sdk-ios/blob/c90c2e4b89ada210628102a6161b0ede6b08983d/DatadogInternal/Sources/Concurrency/ReadWriteLock.swift#L49), making the whole operation atomically protected by a pthread lock. 

Since the operation is performed internally in the [@ReadWriteLock](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogInternal/Sources/Concurrency/ReadWriteLock.swift), the `didSet` property observer is not triggered. To maintain the same logic, after the removal, the `fatalErrorContext` is explicitly called to update the   global attributes.

ncreated: That makes sense 👍 👍 👍. Let's look at adding coverage, as pointed out in https://github.com/DataDog/dd-sdk-ios/pull/2177#discussion_r1928910418 🙂

DatadogRUM/Sources/RUMMonitor/Monitor.swift
```diff
@@ -196,10 +198,24 @@ extension Monitor: RUMMonitorProtocol {
...
         attributes[key] = nil
     }
 
+    func removeAttributes(forKeys keys: [AttributeKey]) {
+        _attributes.mutate {
+            var mutableAttributes = $0
+            keys.forEach { key in mutableAttributes.removeValue(forKey: key) }
+
+            $0 = mutableAttributes
+        }
```
maxep, line 215: /nit
```suggestion
        _attributes.mutate { attributes in
            keys.forEach { key in attributes.removeValue(forKey: key) }
        }
```

simaoseica-dd: 👌

DatadogRUM/Sources/RUMMonitor/Monitor.swift
```diff
@@ -106,9 +106,7 @@ internal class Monitor: RUMCommandSubscriber {
 
     @ReadWriteLock
     private var attributes: [AttributeKey: AttributeValue] = [:] {
-        didSet {
-            fatalErrorContext.globalAttributes = attributes
```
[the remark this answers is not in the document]
simaoseica-dd: Indeed. Just added 2 new tests to validate the `fatalErrorContext.globalAttributes` update

DatadogRUM/Sources/RUMMonitor/Monitor.swift
```diff
@@ -196,10 +198,21 @@ extension Monitor: RUMMonitorProtocol {
...
+
     func removeAttribute(forKey key: AttributeKey) {
         attributes[key] = nil
     }
 
+    func removeAttributes(forKeys keys: [AttributeKey]) {
+        _attributes.mutate { attributes in
+            keys.forEach { key in attributes.removeValue(forKey: key) }
+        }
+        self.didUpdateAttributes()
```
ncreated, line 213: nit/ I wouldn't mind removing the `didSet { self.didUpdateAttributes() }` and calling 
```swift
self.didUpdateAttributes()
```
explicitly from each flow that makes the update. I believe it can be cleaner control flow than the existing one which assumes that the reader of this code knows when `didSet {}` is called vs when it is not if using `@ReadWriteLock`.

ncreated (dismissed):
Well done! I appreciate the effort on adding coverage in `Monitor+GlobalAttributes` 🙂🏅.

DatadogObjc/Sources/RUM/RUM+objc.swift
```diff
@@ -649,17 +649,30 @@ public class DDRUMMonitor: NSObject {
...
         swiftRUMMonitor.addAttribute(forKey: key, value: AnyEncodable(value))
     }
 
     @objc
-    public func removeAttribute(forKey key: String) {
+    public func addAttributes(_ attributes: [AttributeKey: Any]) {
+        let attributes = attributes.reduce(into: [AttributeKey: AnyEncodable]()) {
+            $0[$1.key] = AnyEncodable($1.value)
+        }
+        swiftRUMMonitor.addAttributes(attributes)
```
maxep, line 663: /nit You have an extension for this:
```suggestion
        swiftRUMMonitor.addAttributes(attributes.dd.swiftAttributes)
```

DatadogObjc/Sources/RUM/RUM+objc.swift
```diff
@@ -649,17 +649,30 @@ public class DDRUMMonitor: NSObject {
 
     @objc
     public func addAttribute(
-        forKey key: String,
+        forKey key: AttributeKey,
```
maxep, line 652: /suggestion Lets keep it as `String`, I'm not 100% this will be properly exposed to objc. Plus, we use `String` for all `attributes`.

maxep (approved):
It looks good, I've left a couple of minor suggestions 👍

DatadogObjc/Sources/RUM/RUM+objc.swift
simaoseica-dd: 👌

DatadogRUM/Sources/RUMMonitor/Monitor.swift
simaoseica-dd: Agree. It was a side effect not clear for the caller of the `@ReadWriteLock`

ncreated (approved):

PR #2180, opened 2025-01-24
RUM-5176 [CP] Allow reporting RUM error in sync

maxep:
### What and why?

Cross-platform SDKs can report fatal errors before the native process crashes. However, RUM and the Core execute processing and write asynchronously, and the crash can suspend these executions, leading to the loss of the cross-platform error information.

We should let cross-platform SDKs to report error in sync, blocking the caller-thread.

### How?

The change involve changes at different layers:

#### Core module

Add `completion` argument to `Writer` interface. The completion callback is called when the data is written to disk.

#### RUM Module

Add `completionHandler` property to `RUMErrorCommand`. The `RUMViewScope` calls the completion handler after writing the error event and the view update.

The RUM Monitor now expose a method to report an error with a completion handler, The method is exposed using the `@_spi(Internal)` attribute.

```swift
/// Adds RUM error to current RUM view.
/// 
/// - Parameters:
///   - error: the `Error` object. It will be used to infer error details.
///   - attributes: custom attributes to attach to this error.
///   - completionHandler: A completion closure called when reporting the error is completed.
@_spi(Internal)
func addError(
    error: Error,
    source: RUMErrorSource,
    attributes: [AttributeKey: AttributeValue],
    completionHandler: @escaping CompletionHandler
)
```

#### Objc Module

The `DDRUMMonitor` now expose a `_internal_sync_addError` to report an error in sync. This method will wait for the underlying completion handler to be called before returning.

### Review checklist

### DIFF
--- a/DatadogInternal/Sources/Concurrency/CompletionHandler.swift
+++ b/DatadogInternal/Sources/Concurrency/CompletionHandler.swift
@@ -1,0 +1,15 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+/// Alias from completion closure with no parameter.
+public typealias CompletionHandler = () -> Void
+
+/// No-op completion function.
+///
+/// Using a function prevent allocating a closure when applying a placeholder.
+public func NOPCompletionHandler() {}
--- a/DatadogInternal/Sources/Storage/Writer.swift
+++ b/DatadogInternal/Sources/Storage/Writer.swift
@@ -11,7 +11,7 @@ public protocol Writer {
     /// Encodes given encodable value and metadata, and writes to the destination.
     /// - Parameters:
     ///   - value: Encodable value to write.
-    ///   -  metadata: Encodable metadata to write.
+    ///   - metadata: Encodable metadata to write.
     ///   - completion: The block to execute after the write task is completed.
     func write<T: Encodable, M: Encodable>(value: T, metadata: M?, completion: @escaping CompletionHandler)
 }
@@ -39,7 +39,7 @@ extension Writer {
     /// Encodes given encodable value and metadata, and writes to the destination.
     /// - Parameters:
     ///   - value: Encodable value to write.
-    ///   -  metadata: Encodable metadata to write.
+    ///   - metadata: Encodable metadata to write.
     public func write<T: Encodable, M: Encodable>(value: T, metadata: M?) {
         write(value: value, metadata: metadata, completion: {})
     }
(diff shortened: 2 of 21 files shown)


simaoseica-dd (dismissed):
Great job. Just some questions

mariedm (approved):
Well done! 🙌 And nice PR description.

mariedm (approved):

DatadogInternal/Sources/Storage/Writer.swift
```diff
@@ -9,17 +9,38 @@ import Foundation
 /// A type, writing data.
 public protocol Writer {
     /// Encodes given encodable value and metadata, and writes to the destination.
-    /// - Parameter value: Encodable value to write.
-    /// - Parameter metadata: Encodable metadata to write.
-    func write<T: Encodable, M: Encodable>(value: T, metadata: M?)
+    /// - Parameters:
+    ///   - value: Encodable value to write.
+    ///   -  metadata: Encodable metadata to write.
```
barboraplasovska: /nit Two spaces.
```suggestion
    ///   - metadata: Encodable metadata to write.
```

DatadogInternal/Sources/Storage/Writer.swift
```diff
@@ -9,17 +9,38 @@ import Foundation
...
+    public func write<T: Encodable>(value: T, completion: @escaping CompletionHandler) {
         let metadata: Data? = nil
-        write(value: value, metadata: metadata)
+        write(value: value, metadata: metadata, completion: completion)
+    }
+
+    /// Encodes given encodable value and metadata, and writes to the destination.
+    /// - Parameters:
+    ///   - value: Encodable value to write.
+    ///   -  metadata: Encodable metadata to write.
```
barboraplasovska: /nit Same as above.

```suggestion
    ///   - metadata: Encodable metadata to write.
```

DatadogRUM/Sources/RUMMonitorProtocol.swift
```diff
@@ -370,6 +387,16 @@ extension RUMMonitorViewProtocol {
...
+        error: Error,
+        source: RUMErrorSource,
+        attributes: [AttributeKey: AttributeValue],
+        completionHandler: @escaping CompletionHandler
+    ) {
+        completionHandler()
+    }
 }
 
 // MARK: - NOP moniotor
```
barboraplasovska: /nit

```suggestion
// MARK: - NOP monitor
```

mariedm (dismissed):

DatadogInternal/Sources/Concurrency/CompletionHandler.swift
```diff
@@ -0,0 +1,15 @@
...
+
+import Foundation
+
+/// Alias from completion closure with no parameter.
+public typealias CompletionHandler = () -> Void
+
+/// No-op completion function.
+///
+/// Using a function prevent allocating a closure when applying a placeholder.
+public func NOPCompletionHandler() {}
```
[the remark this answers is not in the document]
maxep: I find optional closure not convenient to propagate, and I prefer making @escaping explicit (optional closure are implicitly escaping). Also, using a static function by default prevent from initialisation a `Optional<()-> Void> for each function parameter.

mariedm (approved):

PR #2182, opened 2025-01-29
hotfix/2.22.1

mariedm:
- **RUM-8219 [SR] Fix overrides memory leak**
- **Bumped version to 2.22.1**

### What and why?

### How?

### Review checklist


maciejburda (approved):

maxep (commented):
Thanks for the fix 🙏  

I think we can still go further and avoid allocating instances of `SessionReplayPrivacyOverrides` when traversing the tree. We should also avoid retaining the `view` instance and prevent future issues.

Here a proposal:

#### 1. Remove the view instance from overrides:
```swift
/// `UIView` extension  to manage the Session Replay privacy override settings.
public final class SessionReplayPrivacyOverrides {
    /// Text and input privacy override (e.g., mask or unmask specific text fields, labels, etc.).
    public var textAndInputPrivacy: TextAndInputPrivacyLevel?
    /// Image privacy override (e.g., mask or unmask specific images).
    public var imagePrivacy: ImagePrivacyLevel?
    /// Touch privacy override (e.g., hide or show touch interactions on specific views).
    public var touchPrivacy: TouchPrivacyLevel?
    /// Hidden privacy override (e.g., mark a view as hidden, rendering it as an opaque wireframe in replays).
    public var hide: Bool?
}
```

#### 2. Instantiate overrides only when accessed publicly:
```swift
/// Extension to provide access to `SessionReplayPrivacyOverrides` for any `UIView`.
extension DatadogExtension where ExtendedType: UIView {
    /// Provides access to Session Replay override settings for the view.
    /// Usage: `myView.dd.sessionReplayPrivacyOverrides.textAndInputPrivacy = .maskNone`.
    public var sessionReplayPrivacyOverrides: SessionReplayPrivacyOverrides {
        if let overrides = _privacyOverrides {
            return overrides
        }

        let overrides = SessionReplayPrivacyOverrides()
        objc_setAssociatedObject(type, &associatedPrivacyOverridesKey, overrides, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
        return overrides
    }

    /// Internal accessor
    internal var _privacyOverrides: SessionReplayPrivacyOverrides? {
        objc_getAssociatedObject(type, &associatedPrivacyOverridesKey) as? SessionReplayPrivacyOverrides
    }
}
```

The `view.dd._privacyOverrides` can be accessed while traversing the tree. This way, no instance will be allocated if the user didn't set any overrides.

Would that make sense?

maxep (approved):
Nice, thank you for the fix 🙏

mariedm:
@maxep thanks for the review! I updated the PR description to reflect the changes of the 2nd iteration.

PR #2183, opened 2025-01-29
Merge hotfix/2.22.1 to develop

mariedm:
# Problem Description

We identified two memory management issues in Session Replay:

**1. `UIView` Memory Leaks**: The `SessionReplayPrivacyOverrides` class maintains strong references to `UIView`s through the chain `ViewTreeSnapshot → nodes[] → viewAttributes → overrides`. When these are passed to background threads for processing in `SnapshotProcessor`, `UIView` lifecycles are unintentionally extended, leading to memory leaks.

**2. Excessive Memory Allocation**: During view hierarchy traversal (every 100ms), new `SessionReplayPrivacyOverrides` instances are created for every view, even when privacy settings haven't changed. This results in unnecessary allocations.

# Technical Solution

The solution involves two improvements:

**1. Separating Mutable Interface from Immutable Data:**
- Introduced a new immutable `PrivacyOverrideValues` `struct` that captures just the override values
- Modified `ViewAttributes` to store these immutable values instead of the mutable overrides object

**2. Optimizing Override Merging:**

Added logic to avoid creating new instances when no actual merging is needed. If neither parent nor child has overrides, or if only one has overrides, we reuse the appropriate instance instead of creating a new one. A new instance is only created when actual value merging is needed.

# Performance Impact

Benchmarked using Instruments Allocations with a test UI containing 3 buttons over a 30-seconds recording period:

Metric | Original | After Values Struct | After Merge Optimization | Total Improvement
-- | -- | -- | -- | --
Privacy Override Allocations (transient) | 6,595 | 3,200 | 2,600 | -61%
Privacy Override Memory | 206KB | 102KB | 81KB | -61%
Peak Memory Usage | 24-39MB | ~8MB | ~8MB | -67% to -79%
UIView Leaks | Yes | No | No | Fixed

### Review checklist


maciejburda (approved):

maxep (approved):
🚀

PR #2188, opened 2025-02-07
Fix OpenTracing doc

maxep:
### What and why?

The following statement in OpenTracing doc is not true: If span has `nil` parent, it becomes the child of the active span.

```swift
// As `span3` has the explicit parent (nil) it won't become the child of the active span:
let span3 = tracer.startSpan(operationName: "another root", childOf: nil)
```

### How?

Update the doc and mention `startRootSpan` instead.

### Review checklist

### DIFF
--- a/DatadogTrace/Sources/OpenTracing/OTSpan.swift
+++ b/DatadogTrace/Sources/OpenTracing/OTSpan.swift
@@ -52,10 +52,10 @@ public protocol OTSpan {
     ///     let span1 = tracer.startSpan(operationName: "root").setActive()
     ///
     ///     // As `span2` has no explicit parent, it becomes the child of the active `span1`:
-    ///     let span2 = tracer.startSpan(operationName: "child of `span1`")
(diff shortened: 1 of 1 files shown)


DatadogTrace/Sources/OpenTracing/OTSpan.swift
```diff
@@ -52,10 +52,10 @@ public protocol OTSpan {
     ///     let span1 = tracer.startSpan(operationName: "root").setActive()
     ///
     ///     // As `span2` has no explicit parent, it becomes the child of the active `span1`:
```
simaoseica-dd, line 54: The `span2` is now the active span right? 
`As span2 has no explicit parent, it becomes the active span`

ncreated (approved):
Good catch 👍

DatadogTrace/Sources/OpenTracing/OTSpan.swift
maxep: Yes, if you call `.setActive()` on `span2`.

PR #2191, opened 2025-02-12
RUM-8042 Add pending batches to batch deleted metric

maxep:
### What and why?

Add pending batches attribute to Batch Deleted Metric.
This attribute will help drawing distribution of pending batches throughout sessions.

### How?

`FilesOrchestrator` keeps track of pending batches and send it as part of `BatchDeletedMetric` telemetry.

### Review checklist

### DIFF
--- a/DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
+++ b/DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
@@ -187,8 +187,7 @@ internal class FilesOrchestrator: FilesOrchestratorType {
             pendingBatches = files.count
 
             let filesFromOldest = try files
-                .map { (file: $0, fileCreationDate: fileCreationDateFrom(fileName: $0.name)) }
-                .compactMap { try deleteFileIfItsObsolete(file: $0.file, fileCreationDate: $0.fileCreationDate) }
+                .compactMap { try deleteFileIfItsObsolete(file: $0, fileCreationDate: fileCreationDateFrom(fileName: $0.name)) }
                 .sorted(by: { $0.fileCreationDate < $1.fileCreationDate })
 
             if ignoreFilesAgeWhenReading {
--- a/DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
+++ b/DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
@@ -58,11 +58,7 @@ internal class FilesOrchestrator: FilesOrchestratorType {
     }
 
     /// An extra information to include in metrics or `nil` if metrics should not be reported for this orchestrator.
-    var metricsData: MetricsData?
-
-    /// Tracks number of pending batches in the track's directory
-    @ReadWriteLock
-    private var pendingBatches: Int = 0
+    let metricsData: MetricsData?
 
     var trackName: String {
         metricsData?.trackName ?? "Unknown"
@@ -125,8 +121,6 @@ internal class FilesOrchestrator: FilesOrchestratorType {
         lastWritableFileApproximatedSize = writeSize
         lastWritableFileLastWriteDate = dateProvider.now
 
-        // Increment pending batches for telemetry
-        pendingBatches += 1
         return newFile
     }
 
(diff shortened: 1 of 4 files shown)


DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
```diff
@@ -185,12 +197,11 @@ internal class FilesOrchestrator: FilesOrchestratorType {
                     .map { $0.file }
```
simaoseica-dd, line 197: question: Doesn't the `ignoreFilesAgeWhenReading` option need to filter the `excludedFileNames`?

simaoseica-dd (approved):
Looks good. Just some questions

DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
maxep: Good question, it wasn't the case and I'm not changing this behaviour, I have just cleaned up the filter :)

DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
```diff
@@ -58,7 +58,11 @@ internal class FilesOrchestrator: FilesOrchestratorType {
     }
 
     /// An extra information to include in metrics or `nil` if metrics should not be reported for this orchestrator.
-    let metricsData: MetricsData?
+    var metricsData: MetricsData?
+
+    /// Tracks number of pending batches in the track's directory
+    @ReadWriteLock
+    private var pendingBatches: Int = 0
```
maxep, line 65: I'm using a lock here because getting and deleting a file can happen on different threads, but this can add penalty to the call thread.

The other option will be to read the directory each time we send the telemetry:
```swift
BatchDeletedMetric.pendingBatches: try? directory.files()
```
But this also has the drawback of reading the disk each time.

DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
```diff
@@ -258,6 +268,9 @@ internal class FilesOrchestrator: FilesOrchestratorType {
     ///
     /// Note: The `batchFile` doesn't exist at this point.
     private func sendBatchDeletedMetric(batchFile: ReadableFile, deletionReason: BatchDeletedMetric.RemovalReason) {
+        // Decrement pending batches at each batch deletion
+        pendingBatches -= 1
```
mariedm, line 272: Shouldn't the decrementation happen where the batch is deleted rather than when sending the telemetry?

DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
```diff
@@ -174,9 +181,13 @@ internal class FilesOrchestrator: FilesOrchestratorType {
 
     func getReadableFiles(excludingFilesNamed excludedFileNames: Set<String> = [], limit: Int = .max) -> [ReadableFile] {
         do {
-            let filesFromOldest = try directory.files()
-                .map { (file: $0, fileCreationDate: fileCreationDateFrom(fileName: $0.name)) }
-                .compactMap { try deleteFileIfItsObsolete(file: $0.file, fileCreationDate: $0.fileCreationDate) }
+            let files = try directory.files()
+
+            // Reset pending batches for telemetry
+            pendingBatches = files.count
```
mariedm, line 187: Could this lead to discrepancies? Right now this variable can be updated from different places.

mariedm (commented):
Looks good! I left a couple of questions.

DatadogCore/Sources/Core/Storage/FilesOrchestrator.swift
maxep: That's a fair point! Makes me leaning toward the solution proposed [here](https://github.com/DataDog/dd-sdk-ios/pull/2191#discussion_r1954071070). This was, we would be sure that data is accurate.

mariedm: Or could we update `pendingBatches` either only here, or when a file get created & removed? This way we keep the lock.

maxep: Getting the `files.count` while reading is to know the pending batches when the SDK starts. At start, no file has been created or removed but there could be files on disk. So, we need to get the count at read/create/delete operations. I will push the other solution, LMKWYT!

mariedm (dismissed):
🚀

mariedm (approved):

PR #2193, opened 2025-02-14
RUM-8042 Collect Upload Quality Metric

maxep:
### What and why?

Collect, aggregate, and report the Upload Quality Metric as specified in [RFC (internal)](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/4609049834/RFC+-+Additional+telemetry+for+uploader+improvements).

### How?

- The `DataUploadWorker` will report each upload cycle through 'Upload Quality' metric telemetry. This metric will include a failure description if applicable.
- The `TelemetryInterceptor` will intercept and forward the 'Upload Quality' metric's attributes to the `SessionEndedMetricController`.
- The `SessionEndedMetricController` will count upload cycle and failures to report the final `upload_quality` attribute of the 'RUM Session Ended' metric telemetry.

### Review checklist

### DIFF
--- a/DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
+++ b/DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
@@ -232,6 +232,10 @@ internal class DataUploadWorker: DataUploadWorkerType {
     }
 
     private func sendUploadQualityMetric(blockers: [DataUploadConditions.Blocker]) {
+        guard !blockers.isEmpty else {
+            return sendUploadQualityMetric()
+        }
+
         sendUploadQualityMetric(
             failure: "blocker",
             blockers: blockers.map {
@@ -245,17 +249,30 @@ internal class DataUploadWorker: DataUploadWorkerType {
     }
 
     private func sendUploadQualityMetric(status: DataUploadStatus) {
+        guard let error = status.error else {
+            return sendUploadQualityMetric()
+        }
+
         sendUploadQualityMetric(
-            failure: status.error.map {
-                switch $0 {
+            failure: {
+                switch error {
                 case let .httpError(code): return "\(code)"
                 case let .networkError(error): return "\(error.code)"
                 }
-            }
+            }()
         )
     }
 
-    private func sendUploadQualityMetric(failure: String?, blockers: [String] = []) {
+    private func sendUploadQualityMetric() {
+        telemetry.metric(
+            name: UploadQualityMetric.name,
+            attributes: [
+                UploadQualityMetric.track: featureName
+            ]
+        )
+    }
+
+    private func sendUploadQualityMetric(failure: String, blockers: [String] = []) {
         telemetry.metric(
             name: UploadQualityMetric.name,
             attributes: [
--- a/DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
+++ b/DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
@@ -233,7 +233,8 @@ internal class DataUploadWorker: DataUploadWorkerType {
 
     private func sendUploadQualityMetric(blockers: [DataUploadConditions.Blocker]) {
         sendUploadQualityMetric(
-            failure: blockers.first.map {
+            failure: "blocker",
+            blockers: blockers.map {
                 switch $0 {
                 case .battery: return "low_battery"
                 case .lowPowerModeOn: return "lpm"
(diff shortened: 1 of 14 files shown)


DatadogRUM/Sources/Integrations/TelemetryInterceptor.swift
```diff
@@ -20,6 +20,10 @@ internal struct TelemetryInterceptor: FeatureMessageReceiver {
         switch telemetry {
         case .error(let id, let message, let kind, let stack):
             interceptError(id: id, message: message, kind: kind, stack: stack)
+        case .metric(let metric) where metric.name == SDKMetricFields.UploadQuality.name:
+            interceptUploadQualityMetric(attributes: metric.attributes)
+            return true
+
```
ncreated, line 26: question/ The return true statement consumes the metric, meaning it will never be sent as a standalone metric event. As a result, the only effective operation here is intercepting metric.attributes.

Is this an intentional design choice? If so, could we document it more clearly? 🙏

For consistency, we might want to follow the convention of defining a namespace for each metric. Something like this could help clarify its purpose:
```swift
/// ...
/// This metric is not uploaded by the SDK. Instead, it is consumed internally  
/// within the message bus.
/// ...
enum UploadQualityMetric {
   /// Metric name
   public static let name = "upload_quality"

   // ...
}
```

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
```diff
@@ -217,6 +230,39 @@ internal class DataUploadWorker: DataUploadWorkerType {
...
+
+    private func sendUploadQualityMetric(blockers: [DataUploadConditions.Blocker]) {
+        sendUploadQualityMetric(
+            failure: blockers.first.map {
+                switch $0 {
+                case .battery: return "low_battery"
+                case .lowPowerModeOn: return "lpm"
+                case .networkReachability: return "offline"
+                }
+            }
```
ncreated, line 242: concern/ We currently capture only one blocker, even when multiple could apply. This limits our ability to analyze certain cases—for instance, we won’t be able to determine the percentage of uploads blocked solely by `"lpm"`, since some will be reported under `"low_battery"` or `"offline"` instead.

Possible mitigations:
- Concatenate all blockers into a single string (e.g., `"offline+lpm"`).
- Establish a priority order for blockers (e.g., treat `"lpm"` as more important than `"offline"`).

IMO, what we should do depends on the hypothesis we want to test and how we plan to measure it. Concatenated strings allow easy filtering in queries but make it harder to analyze distributions. If doing priorities, we must define most important blockers.

ncreated (commented):
Looks good 🔭👍! I left few minor feedbacks and raised one concern on reporting upload blockers - happy to discuss.

DatadogRUM/Sources/Integrations/TelemetryInterceptor.swift
maxep: Yes, it is by design and lack documentation. It makes sense to have the `UploadQualityMetric` definition, I will add that 👍

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
maxep: That's a fair point! I think applying priority is better for analysis, distribution of failures is important. But the question is what would be the priority 🤔 I will discuss with Android to align!

Valpertui: Or maybe we should report it as 2 fields in case of failure due to a blocker ?  
```
"failure": "blocker",
"blockers": {
  "lpm": true/false,
  "low_power": true/false,
  "offline": true/false
} (or an array if it's easier to use on dashboard)
```

maxep: Good suggestion @Valpertui 👍

maxep: After discussion with @jonathanmos, we are going with @Valpertui proposal to split the failures and blockers and aggregate them separately in the RSE metric.

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
```diff
@@ -86,6 +84,7 @@ internal class DataUploadWorker: DataUploadWorkerType {
                 self.delay.increase()
                 self.backgroundTaskCoordinator?.endBackgroundTask()
                 self.scheduleNextCycle()
+                sendUploadQualityMetric(blockers: blockersForUpload)
```
ncreated, line 87: question/ Shouldn't we differentiate between "no batch" and "blockers"? Here we seem to merge both into one and blockers will be reported even if there were no files to upload, no?

ncreated (approved):
👌 I left one last question - otherwise it looks good 💪.

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
maxep: oh damn, yes! We shouldn't report any failure if there no batch, we just need to increment the cycle count. I will fix that!

maxep: Fixed ✅  It won't report failure when no batch is present.

ncreated (approved):

PR #2195, opened 2025-02-20
RUM-8672 Update UserInfo API to make id mandatory.

maciejburda:
### What and why?

Makes `id` parameter of `setUserInfo()` mandatory. Old signature is kept with deprecation warning.

### Review checklist


DatadogCore/Sources/Datadog.swift
```diff
@@ -269,10 +269,26 @@ public enum Datadog {
     /// Those will be added to logs, traces and RUM events automatically.
     ///
     /// - Parameters:
-    ///   - id: User ID, if any
+    ///   - id: Mandatory User ID
     ///   - name: Name representing the user, if any
     ///   - email: User's email, if any
     ///   - extraInfo: User's custom attributes, if any
+    public static func setUserInfo(
+        id: String,
```
simaoseica-dd, line 277: question: Should it validate empty string ids?

barboraplasovska: I would say no, because an empty string is not a valid ID ?

barboraplasovska (approved):
LGTM👌

DatadogCore/Sources/Datadog.swift
maciejburda: That's a fair point - using empty string is "easy" way around the requirement.

I think it's more about making it clear that we require this value rather than preventing edge cases.

simaoseica-dd (approved):

shagedorn:
What's the motivation behind this change?

We currently set `id` to `nil` when a user logs out of the app, in which case we don't have a useful id anymore. We would now have to come up with an arbitrary string value to signal this, and it will look different than the initial (pre-login) state where no user id was set at all. It complicates something that was never a problem for us.

maciejburda:
Thanks for tuning in, @shagedorn!

The motivation behind this change is purely internal, aimed at aligning our platforms with the introduction of our new product, Product Analytics.

Your logout scenario makes perfect sense, and it raises a valid point—perhaps we need an additional API to clear the entire user object. For now, you can continue using the deprecated function.

We’d really appreciate it if you could reach out through official support and share your exact use cases. This will help us explore whether there’s a better way to address this change while minimizing any disruption on your end.

shagedorn:
Thanks for your response! I described our use case in ticket nr. 2081021.

maciejburda:
Thanks! I'll look into it.

Also, correction - stopping the session won't clear the user automatically. We will likely add an API to clear the user.

siddarthkalra:
@maciejburda we're facing a very similar issue after updating to the latest SDK release. What's the current guidance with handling the logged out state for an app when no user ID is available?

shagedorn:
@maciejburda thanks for following up!

I believe this would help us indeed. We have some properties we set via the `extraInfo:` parameter, which we'd like to retain post logout, but I think we could use `addUserExtraInfo(…)` to re-apply them after calling `clearUserInfo()`.

PR #2200, opened 2025-02-24
RUM-8774 SessionReplay Image recording uses correct frame calculation for scaleAspectFill

Valpertui:
### What and why?

What?
In Session Replay when capturing images from `UIImageView` with `contentMode` `scaleAspectFill`, in some condition the computed `CGRect` which represent what the customer will see on Session Replay Player, did not match what was displayed on device.

Why?
The condition used to find which dimension "scaling factor" to apply (height or width), was comparing the distance (in points) difference between the container and the content, instead of comparing the scaling factors required so that the image fill properly the container for both dimensions.  

### How?

The fix change this condition to now compute both the `x` and `y` scaling factor needed so the content image fills the container, and select the biggest of the two (using `max`), ensuring the image always fill the container in both dimensions.

### Review checklist


maxep (approved):
Good catch, Thanks!!

PR #2201, opened 2025-02-24
RUM-8409 Propagate RUM session ID in request headers

xgouchet:
### What and why?

Propagate the RUM Session ID in traced requests headeres (DD & W3C)

### How?

We use the `FeatureMessage` to track the current RUM Session ID and use it in the tags in DD and W3C headers.

### DIFF
--- a/DatadogInternal/Sources/NetworkInstrumentation/Datadog/HTTPHeadersWriter.swift
+++ b/DatadogInternal/Sources/NetworkInstrumentation/Datadog/HTTPHeadersWriter.swift
@@ -82,6 +82,9 @@ public class HTTPHeadersWriter: TracePropagationHeadersWriter {
             traceHeaderFields[TracingHTTPHeaders.traceIDField] = String(traceContext.traceID.idLo)
             traceHeaderFields[TracingHTTPHeaders.parentSpanIDField] = String(traceContext.spanID, representation: .decimal)
             traceHeaderFields[TracingHTTPHeaders.tagsField] = "_dd.p.tid=\(traceContext.traceID.idHiHex)"
+            if let sessionId = traceContext.rumSessionId {
+                traceHeaderFields[W3CHTTPHeaders.baggage] = "\(W3CHTTPHeaders.Constants.rumSessionBaggageKey)=\(sessionId)"
+            }
         case (.sampled, false):
             break
         }
--- a/DatadogInternal/Sources/NetworkInstrumentation/NetworkContextProvider.swift
+++ b/DatadogInternal/Sources/NetworkInstrumentation/NetworkContextProvider.swift
@@ -1,0 +1,65 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+/// An interface for writing and reading  the `NetworkContext`
+internal protocol NetworkContextProvider: AnyObject {
+    /// Returns current `NetworkContext` value.
+    var currentNetworkContext: NetworkContext? { get }
+}
+
+/// Manages the `NetworkContext` reads and writes in a thread-safe manner.
+internal class NetworkContextCoreProvider: NetworkContextProvider {
+    // MARK: - NetworkContextProviderType
+    @ReadWriteLock
+    var currentNetworkContext: NetworkContext?
+}
+
+extension NetworkContextCoreProvider: FeatureMessageReceiver {
+    /// Defines keys referencing RUM baggage in `DatadogContext.featuresAttributes`.
+    internal enum RUMBaggageKeys {
+        /// The key references RUM view event.
+        /// The view event associated with the key conforms to `Codable`.
+        static let viewEvent = "rum-view-event"
+
+        /// The key references a `true` value if the RUM view is reset.
+        static let viewReset = "rum-view-reset"
+
+        /// The key references RUM session state.
+        /// The state associated with the key conforms to `Codable`.
+        static let sessionState = "rum-session-state"
+
+        /// This key references the global log attributes
+        static let logAttributes = "global-log-attributes"
+
+        /// The key referencing ``DatadogInternal.GlobalRUMAttributes`` value holding RUM global attributes.
+        static let rumAttributes = "global-rum-attributes"
+    }
+
+    func receive(message: FeatureMessage, from core: DatadogCoreProtocol) -> Bool {
+        switch message {
+        case .context(let context):
+            update(context: context, to: core)
+        default:
+            return false
+        }
+
+        return true
+    }
+
+    /// Updates crash context.
+    ///
+    /// - Parameter context: The updated core context.
+    private func update(context: DatadogContext, to core: DatadogCoreProtocol) {
+        do {
+            self.currentNetworkContext = NetworkContext(rumContext: try context.baggages["rum"]?.decode())
+        } catch {
+            core.telemetry
+                .error("Fails to decode RUM Context from Trace", error: error)
+        }
+    }
+}
--- a/DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
+++ b/DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
@@ -29,7 +29,9 @@ internal final class NetworkInstrumentationFeature: DatadogFeature {
     )
 
     /// A no-op message bus receiver.
-    internal let messageReceiver: FeatureMessageReceiver = NOPFeatureMessageReceiver()
+    let messageReceiver: FeatureMessageReceiver
+
+    let networkContextProvider: NetworkContextProvider
 
     /// The list of registered handlers.
     ///
@@ -45,6 +47,14 @@ internal final class NetworkInstrumentationFeature: DatadogFeature {
     ///
     /// The interceptions **must** be accessed using the `queue`.
     private var interceptions: [URLSessionTask: URLSessionTaskInterception] = [:]
+
+    init(
+        networkContextProvider: NetworkContextProvider,
+        messageReceiver: FeatureMessageReceiver
+    ) {
+        self.networkContextProvider = networkContextProvider
+        self.messageReceiver = messageReceiver
+    }
 
     /// Swizzles `URLSessionTaskDelegate`, `URLSessionDataDelegate`, and `URLSessionTask` methods
     /// to intercept `URLSessionTask` lifecycles.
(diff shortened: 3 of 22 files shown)


DatadogInternal/Sources/NetworkInstrumentation/Datadog/HTTPHeadersWriter.swift
```diff
@@ -81,7 +81,11 @@ public class HTTPHeadersWriter: TracePropagationHeadersWriter {
             ]
             traceHeaderFields[TracingHTTPHeaders.traceIDField] = String(traceContext.traceID.idLo)
             traceHeaderFields[TracingHTTPHeaders.parentSpanIDField] = String(traceContext.spanID, representation: .decimal)
-            traceHeaderFields[TracingHTTPHeaders.tagsField] = "_dd.p.tid=\(traceContext.traceID.idHiHex)"
+            if traceContext.rumSessionId == nil {
+                traceHeaderFields[TracingHTTPHeaders.tagsField] = "_dd.p.tid=\(traceContext.traceID.idHiHex)"
+            } else {
+                traceHeaderFields[TracingHTTPHeaders.tagsField] = "_dd.p.tid=\(traceContext.traceID.idHiHex),_dd.p.rsid=\(traceContext.rumSessionId)"
+            }
```
maxep, line 88: /suggestion

```suggestion
            if let sessionId = traceContext.rumSessionId {
                traceHeaderFields[TracingHTTPHeaders.tagsField] = "_dd.p.tid=\(traceContext.traceID.idHiHex),_dd.p.rsid=\(sessionId)"
            } else {
                traceHeaderFields[TracingHTTPHeaders.tagsField] = "_dd.p.tid=\(traceContext.traceID.idHiHex)"
            }
```

maxep (commented):
I have made some more idiomatic suggestions 👍

DatadogInternal/Sources/NetworkInstrumentation/Datadog/HTTPHeadersWriter.swift
```diff
@@ -81,7 +81,11 @@ public class HTTPHeadersWriter: TracePropagationHeadersWriter {
             ]
             traceHeaderFields[TracingHTTPHeaders.traceIDField] = String(traceContext.traceID.idLo)
             traceHeaderFields[TracingHTTPHeaders.parentSpanIDField] = String(traceContext.spanID, representation: .decimal)
-            traceHeaderFields[TracingHTTPHeaders.tagsField] = "_dd.p.tid=\(traceContext.traceID.idHiHex)"
+            if let sessionId = traceContext.rumSessionId {
+                traceHeaderFields[TracingHTTPHeaders.tagsField] = "_dd.p.tid=\(traceContext.traceID.idHiHex),_dd.p.rsid=\(String(describing: sessionId))"
```
simaoseica-dd, line 85: No need for `String(describing:`. It was already unwrapped. 
`_dd.p.rsid=\(sessionId)` is enough

DatadogInternal/Sources/NetworkInstrumentation/W3C/W3CHTTPHeadersWriter.swift
```diff
@@ -97,13 +97,17 @@ public class W3CHTTPHeadersWriter: TracePropagationHeadersWriter {
...
-            let tracestate: [String: String] = [
+            var tracestate: [String: String] = [
                 Constants.sampling: "\(sampled ? 1 : 0)",
                 Constants.parentId: String(traceContext.spanID, representation: .hexadecimal16Chars)
             ].merging(tracestate) { old, new in
                 return new
             }
 
+            if let sessionId = traceContext.rumSessionId {
+                tracestate[Constants.rumSessionId] = String(describing: sessionId)
```
simaoseica-dd, line 108: No need of `String(describing:`

DatadogTrace/Sources/Feature/MessageReceivers.swift
```diff
@@ -28,6 +33,16 @@ internal final class ContextMessageReceiver: FeatureMessageReceiver {
...
+        case .telemetry(let telemetry):
+            return false
+        case .baggage(let label, let baggage) where label == "rum-session-state":
+            do {
+                try _context.mutate { $0.rumSession = try baggage.decode() }
+            } catch {
+                core.telemetry
+                    .error("Fails to decode RUM Session State from Trace", error: error)
+            }
+            return false
```
maxep, line 45: /blocker returning `false` is right so the message can be propagated to other receivers. But it raises the issue of the message being intercepted before `DatadogTrace` by `DatadogCrashReporting`, [here](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogCrashReporting/Sources/CrashContext/CrashContextProvider.swift#L91). We should return `false` there as well to make sure the message is always propagated to both modules. It is a design flaw of the message-bus that we should refactor at some point.

maxep (changes requested):
It looks great 👍  but I noticed an issue with the message-bus propagation that should be addressed

DatadogTrace/Sources/Feature/MessageReceivers.swift
ncreated: @maxep do we need to listen to the `"rum-session-state"` baggage here? Isn't RUM session info already available in `context` message?

maxep: > Please ignore my previous comment, this behaviour has been changed since.

@ncreated Oh that's true, we are already listening to the `context` message on the bus, so we can have the session ID from there. Good catch 👍 

Find below the baggage decoding based on the `context` message:

```swift 
internal struct CoreContext {
    struct RUMContext: Decodable {
            case sessionID = "session.id"
        }
        let sessionID: String
    }

    /// Provides the history of app foreground / background states.
    var applicationStateHistory: AppStateHistory?

    /// Provides the current active RUM Session information, if any
    var rumContext: RUMContext?
}

internal final class ContextMessageReceiver: FeatureMessageReceiver {
    /// The up-to-date core context.
    ///
    /// The context is synchronized using a read-write lock.
    @ReadWriteLock
    var context: CoreContext = .init()

    /// Process messages receives from the bus.
    ///
    /// - Parameters:
    ///   - message: The Feature message
    ///   - core: The core from which the message is transmitted.
        switch message {
        case .context(let context):
            return update(context: context, from: core)
        default:
            return false
        }
    }

    /// Updates context of the `DatadogTracer` if available.
    ///
    /// - Parameter context: The updated core context.
    private func update(context: DatadogContext, from core: DatadogCoreProtocol) -> Bool {
        _context.mutate {
            $0.applicationStateHistory = context.applicationStateHistory
            do {
                $0.rumContext = try context.baggages["rum"]?.decode()
            } catch {
                core.telemetry
                    .error("Fails to decode RUM Context from Trace", error: error)
            }
        }

        return true
    }
}
```

ncreated (changes requested):
Question on definition of done - should this only work for network auto instrumentation in `DatadogTrace` or also in `DatadogRUM`? I suppose it should work in both, but this is not the observation I see when testing this feature manually.

When using `DatadogRUM` + `DatadogTrace` network instrumentation, **it works**:
```swift
RUM.enable(with: .init(applicationID: "app-id"))

Trace.enable(
   with: Trace.Configuration(
      urlSessionTracking: init(
         firstPartyHostsTracing: .trace(hosts: ["my.host"])
      )
   )
)

// X-Datadog-Tags: _dd.p.tid=67dbce2d00000000,_dd.p.rsid=3c53a993-4967-4e86-9b7b-26e02df8d91f
```

When using only `DatadogRUM` with network instrumentation, **it doesn't set session ID in header**.
```swift
RUM.enable(
   with: .init(
      applicationID: "app-id",
      urlSessionTracking: .init(
         firstPartyHostsTracing: .trace(hosts: ["my.host"])
      )
   )
)

// X-Datadog-Tags: _dd.p.tid=67dbd03800000000
```

This parity seems to be important, hence I believe there is some missing implementation.

DatadogInternal/Sources/NetworkInstrumentation/NetworkContextProvider.swift
```diff
@@ -0,0 +1,102 @@
...
+    // MARK: - NetworkContextProviderType
+
+    var currentNetworkContext: NetworkContext? {
+        queue.sync { _context }
+    }
+
+    var onNetworkContextChange: (NetworkContext) -> Void {
+        get { queue.sync { self._callback } }
+        set { queue.async { self._callback = newValue } }
+    }
```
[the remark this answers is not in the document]
xgouchet: Indeed we don't need it anymore, it's a remnant of my explorations

DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
```diff
@@ -148,15 +148,17 @@ internal final class URLSessionRUMResourcesHandler: DatadogURLSessionHandler, RU
 }
 
 extension DistributedTracing {
-    func modify(request: URLRequest, headerTypes: Set<DatadogInternal.TracingHeaderType>) -> (URLRequest, TraceContext?) {
+    func modify(request: URLRequest, headerTypes: Set<DatadogInternal.TracingHeaderType>, rumSessionId: String?) -> (URLRequest, TraceContext?) {
         let traceID = traceIDGenerator.generate()
         let spanID = spanIDGenerator.generate()
+        // TODO RUM-1234
```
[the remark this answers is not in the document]
xgouchet: Same it was just a remnant of my explorations

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
```diff
@@ -148,10 +158,11 @@ extension NetworkInstrumentationFeature {
             return (request, [])
         }
 
+        let networkContext = self.networkContextProvider.currentNetworkContext
```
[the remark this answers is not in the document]
xgouchet: Thanks for the tip, I'll try and make the change :)

ncreated (dismissed):

DatadogInternal/Sources/NetworkInstrumentation/NetworkContextProvider.swift
```diff
@@ -0,0 +1,85 @@
...
+internal class NetworkContextCoreProvider: NetworkContextProvider {
+    /// Queue for synchronizing `unsafeNetworkContext` updates.
+    private let queue = DispatchQueue(
+        label: "com.datadoghq.network-context",
+        target: .global(qos: .utility)
+    )
+
+    // MARK: - NetworkContextProviderType
+    @ReadWriteLock
+    var currentNetworkContext: NetworkContext?
```
maxep, line 25: suggestion/ No need for a queue since the context is protected by the lock 👍

maxep (dismissed):
Looks great! I left a small suggestion

maxep (approved):
Thanks 🙏

PR #2202, opened 2025-02-24
chore: Update RUMSchema to latest version

fuzzybinary:
### What and why?

Adds telemetry for tns / inv thresholds as well as performance members for Flutter.

### Review checklist


DatadogObjc/Sources/RUM/RUMDataModels+objc.swift
```diff
@@ -6756,8 +7118,8 @@ public class DDRUMVitalEventVital: NSObject {
         root.swiftModel.vital.custom as [String: NSNumber]?
     }
 
-    @objc public var details: String? {
-        root.swiftModel.vital.details
+    @objc public var vitalDescription: String? {
+        root.swiftModel.vital.description
```
fuzzybinary, line 7122: This was specifically changed manually after the fact as using `description` creates a conflict with `NSObject`'s description property.

The generator will have to be changed to fix this somehow, but I considered it outside the scope of this change.

maciejburda (dismissed):

DatadogObjc/Sources/RUM/RUMDataModels+objc.swift
ncreated: blocking/ We can't patch generated models manually, otherwise linter phase will not pass (as it already happened to this PR).

We can try adjusting it in [RUMCodeDecorator.swift#L45](https://github.com/DataDog/dd-sdk-ios/blob/41715ccbabefad98633746358286079c21a919e9/tools/rum-models-generator/Sources/CodeDecoration/RUMCodeDecorator.swift#L45) to avoid conflict on Swift reserved property name (`description`). We can rename it to `vitalDescription` with preserving the `"description"` coding key:
```diff
+ /// Description of the vital. It can be used as a secondary identifier (URL, React component name...)
+ public let vitalDescription: String?

// ...

+ case vitalDescription = "description"
```

fuzzybinary:
Fixed. Let me know if that's a workable solution

ncreated (approved):
🎯🏅 Thank you!

PR #2203, opened 2025-02-25
feat: Support custom internal attributes for custom INV and Flutter FBC

fuzzybinary:
### What and why?

This adds an internal method - `setInternalViewAttribute` - which modifies a specific attribute collection for views. This attribute collection is checked for custom INV Values (if INV is disabled) and Flutter First Build Complete metrics.

This also adds the ability to disable INV by setting its last action predicate to `false`

### Review checklist


DatadogRUM/Sources/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -36,6 +36,8 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
     let identity: ViewIdentifier
     /// View attributes.
     private(set) var attributes: [AttributeKey: AttributeValue] = [:]
+    /// Internal view attributes - used by cross platform frameworks and should not be propoagated to events
```
barboraplasovska, line 39:
```suggestion
    /// Internal view attributes - used by cross platform frameworks and should not be propagated to events
```
Just a little typo 😄

barboraplasovska (approved):
Good job👌 Found a small typo, otherwise LGTM

DatadogRUM/Sources/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -523,7 +531,28 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
         let refreshRateInfo = vitalInfoSampler?.refreshRate
         let isSlowRendered = refreshRateInfo?.meanValue.map { $0 < Constants.slowRenderingThresholdFPS }
         let networkSettledTime = networkSettledMetric.value(with: context.applicationStateHistory)
-        let interactionToNextViewTime = interactionToNextViewMetric.value(for: viewUUID)
+        var interactionToNextViewTime = interactionToNextViewMetric.value(for: viewUUID)
+        // Only look for an internal override attribute if native INV calculations are disabled
+        if interactionToNextViewTime == .failure(.disabled),
+           let customInvValue = internalAttributes[RUMInternalAttributes.customINVValue] as? (any BinaryInteger) {
+            let customInvValue = Int64(customInvValue)
```
ncreated, line 538: blocking/ We must check for overflow
```suggestion
            let customInvValue = Int64(withNoOverflow: customInvValue)
```

DatadogRUM/Sources/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -523,7 +531,28 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
...
+            interactionToNextViewTime = .success(TimeInterval(fromNanoseconds: customInvValue))
+        }
+
+        // Only add the performance member if we have a value for it
+        let performance: RUMViewEvent.View.Performance?
+        if let fbcMetric = internalAttributes[RUMInternalAttributes.flutterFirstBuildComplete] as? (any BinaryInteger) {
+            let fbcMetric = Int64(fbcMetric)
```
ncreated, line 545: blocking/ Needs overflow check:
```suggestion
            let fbcMetric = Int64(withNoOverflow: fbcMetric)
```

DatadogRUM/Sources/RUMMonitorProtocol+Internal.swift
```diff
@@ -7,6 +7,18 @@
...
+public struct RUMInternalAttributes {
+    /// Custom Flutter vital - First Build Complete. The amount of time between a route change (the start of a view) and when the first
+    /// `build` method is complete. In nanoseconds since view start
+    public static let flutterFirstBuildComplete: String = "_dd.performance.first_build_complete"
+
+    /// Custom value for Interaction To Next view.
+    /// For Flutter this is the amount of time between an action occurring and the First Build Complete ocurring on the next view.
+    public static let customINVValue: String = "_dd.view.custom_inv_value"
+}
+
```
ncreated, line 21: convention/ Can we move it to [`CrossPlatformAttributes`](https://github.com/DataDog/dd-sdk-ios/blob/363861da11ca1fe0861010bdcbc1402898f03af5/DatadogInternal/Sources/Attributes/Attributes.swift#L67-L69)?

ncreated (changes requested):
👌 Well done, thanks 🚀! I only found two minor issues and left one suggestion on convention.

DatadogRUM/Sources/RUMConfiguration.swift
```diff
@@ -166,9 +166,11 @@ extension RUM {
         /// The `NextViewActionPredicate` determines which action in the previous view should be considered the "last interaction" for INV,
         /// based on properties such as action type, name, or timing relative to the next view’s start.
         ///
+        /// Setting this property to `nil` will disable measuring Interaction to Next View.
```
simaoseica-dd, line 169: What about a `NOPNextViewActionPredicate`? This way you would avoid the dance of the optionals.

DatadogRUM/Sources/RUMMonitor/RUMCommand.swift
```diff
@@ -541,3 +541,15 @@ internal struct RUMUpdatePerformanceMetric: RUMCommand {
...
     let missedEventType: SessionEndedMetric.MissedEventType? = nil
 }
+
+internal struct RUMSetInternalViewAttributeCommand: RUMCommand {
+    let canStartBackgroundView = false
+    let isUserInteraction = false
+    var time: Date
+    var globalAttributes: [AttributeKey: AttributeValue] = [:]
+    var attributes: [AttributeKey: AttributeValue] = [:]
+    let missedEventType: SessionEndedMetric.MissedEventType? = nil
```
simaoseica-dd, line 551: Seems like we are leaking command parameters not related to them. We can improve this

DatadogRUM/Sources/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -523,7 +531,28 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
         let refreshRateInfo = vitalInfoSampler?.refreshRate
         let isSlowRendered = refreshRateInfo?.meanValue.map { $0 < Constants.slowRenderingThresholdFPS }
         let networkSettledTime = networkSettledMetric.value(with: context.applicationStateHistory)
-        let interactionToNextViewTime = interactionToNextViewMetric.value(for: viewUUID)
+        var interactionToNextViewTime = interactionToNextViewMetric.value(for: viewUUID)
+        // Only look for an internal override attribute if native INV calculations are disabled
+        if interactionToNextViewTime == .failure(.disabled),
```
simaoseica-dd, line 536: This is weird from the readability point of view. It failed because it was disabled, but then it is enabled with success. 

Also, is it possible to have value for `interactionToNextViewTime`  and value for the `RUMInternalAttributes.customINVValue` attribute?

simaoseica-dd (commented):
Mostly questions to help improve logic and readability

DatadogRUM/Sources/RUMMonitorProtocol+Internal.swift
ncreated: An argument towards continuing existing convention ([CrossPlatformAttributes](https://github.com/DataDog/dd-sdk-ios/blob/363861da11ca1fe0861010bdcbc1402898f03af5/DatadogInternal/Sources/Attributes/Attributes.swift#L67-L69)) is that this will keep these attributes internal by convention (clients do not depend on `DatadogInternal`). Otherwise, if storing it in `DatadogRUM`, they became a public API that needs to be maintained.

With lack of better alternative at the moment (maybe `@_spi(Internal)`?), I'd recommend following the existing way. Even if not perfect, it's maintainable.

DatadogRUM/Sources/RUMMonitor/RUMCommand.swift
```diff
@@ -541,3 +541,15 @@ internal struct RUMUpdatePerformanceMetric: RUMCommand {
...
+
+    let key: AttributeKey
```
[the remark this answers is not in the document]
fuzzybinary: They could, but I'm trying to keep a separation of concerns. The "attributes" here have a very different meaning than what I'm sending in the event, so a similar logic as to why its separated out in the class applies as well.

DatadogRUM/Sources/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -36,6 +36,8 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
     let identity: ViewIdentifier
     /// View attributes.
     private(set) var attributes: [AttributeKey: AttributeValue] = [:]
+    /// Internal view attributes - used by cross platform frameworks and should not be propagated to events
+    private(set) var internalAttributes: [AttributeKey: AttributeValue] = [:]
```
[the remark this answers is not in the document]
fuzzybinary: This was the original idea when I started on this but two things made me decide to separate them out:
* The weird edge cases I found around view attributes, resulting in the RFC we currently have
* The fact that, in the future, view attributes should be populated down to events

The difference in behavior and meaning let me to keep these attributes outside of the normal `attributes`. This way I easily ensure none of them are ever sent to the backed either accidentally as attributes on a View or as attributes propagated to other Events.

DatadogRUM/Sources/RUMConfiguration.swift
fuzzybinary: That was actually the original plan when the ticket was created. I seem to remember @ncreated wanting to use optionals instead?

DatadogRUM/Sources/RUMMonitor/Scopes/RUMViewScope.swift
fuzzybinary: Yeah, hence the comment. We're essentially preventing using custom INV unless you purposefully disable automatic INV. Maybe I can make that comment clearer?

One reason is because it could be possible to have both a custom INV value and a calculated, non-custom one, especially in Flutter "add-to-app" scenarios. And the way INV is calculated for those two scenarios (Native vs. Flutter) are very different. I'd rather make sure one method or the other is being used instead of attempting to try to figure out which one is "best"

fuzzybinary:
Closing as we're going to take a different approach

PR #2205, opened 2025-02-26
RUM-1871 Reduced view update may miss `hasReplay` flag set before

barboraplasovska:
### What and why?

RUM backend has a special handling for `hasReplay` flag - once it is sent to true in any view update it cannot go back to false.

On the client side, however, we reduce the views, keeping only the last view update for a given `viewId` in the batch. If this view had originally `hasReplay: true` and then last view update has `hasReplay:false` (because by the time view update is sent session replay is disabled) we will miss session replay info for this view.

### How?

Introduction of a new private attribute `hasReplay` in the `RUMViewScope` class that is set each time a new View event is built. In the case that it has once been set to true, each following `RUMViewEvent` that is created will keep `hasReplay` set to true.

### Review checklist

### DIFF
--- a/DatadogRUM/Tests/RUMMonitor/Scopes/RUMViewScopeTests.swift
+++ b/DatadogRUM/Tests/RUMMonitor/Scopes/RUMViewScopeTests.swift
@@ -2833,7 +2833,7 @@ class RUMViewScopeTests: XCTestCase {
 
     // MARK: - Has replay
 
-    func testGivenActiveViewWithoutReplay_viewUpdateWithReplay_itModifiedHasReplayFlag() throws {
+    func testViewUpdate_changesHasReplayFlag_onlyWhenReplayIsPresent() throws {
         // Given
         context.baggages = try .mockSessionReplayAttributes(hasReplay: false)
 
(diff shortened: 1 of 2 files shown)


maxep (dismissed):
To the point, very nice work 🎖️ 👍

DatadogRUM/Tests/RUMMonitor/Scopes/RUMViewScopeTests.swift
```diff
@@ -2830,4 +2830,63 @@ class RUMViewScopeTests: XCTestCase {
         XCTAssertEqual(metric.trackedViewStarts.count, 1)
         XCTAssertEqual(metric.trackedViewCompletes.count, 1)
     }
+
+    // MARK: - Has replay
+
+    func testViewUpdate_changesHasReplayFlag_onlyWhenReplayIsPresent() throws {
```
mariedm, line 2836: nit, just a suggestion to make it clearer, but feel free to disregard

```suggestion
    func testViewUpdate_onceHasReplayIsTrueItRemainsTrue() throws {
```

mariedm (approved):
Looks great, nice job! 👏

maxep (approved):

mariedm (approved):

PR #2208, opened 2025-02-27
RUM-8371 chore: Synchronize app state observers across the SDK

ncreated:
### What and why?

📦 This PR refines how the SDK observes application state changes across Core and all Features. It streamlines this logic to ensure consistent state change notifications across different products

#### 1️⃣ Synchronized App State Updates

The main improvement ensures that all subsystems receive app state updates in a controlled order — **first Core, then Features**, respecting their `enable()` order. This is achieved by synchronizing initialization across all subsystems on the main thread, using a new helper in `DatadogInternal`:
```swift
/// Ensures the provided block is executed on the main thread, synchronously, rethrowing any error.
/// ...
public func runOnMainThreadSync<T>(_ block: () throws -> T) rethrows -> T {
    if Thread.isMainThread {
        return try block()
    } else {
        return try DispatchQueue.main.sync(execute: block)
    }
}
```
⏱️  For clients who already initialize the SDK on the main thread, this introduces no performance impact.

#### 2️⃣ Refactored `ApplicationStatePublisher`

- Removed the `queue`, ensuring state changes are synchronized via the main thread.
- Guarantees that Core processes state changes before Features.

#### 3️⃣ Streamlined `AppStateHistory` Interface

The `AppStateHistory` type has been simplified to better align with the [Interface Segregation Principle (ISP)](https://en.wikipedia.org/wiki/Interface_segregation_principle).

- **Before**, the interface exposed numerous public methods, making it harder to use correctly:
```swift
public struct AppStateHistory {    
    public struct Snapshot {
	public let state: AppState
	public let date: Date
        public init(state:date:)
    }

    public var initialSnapshot: Snapshot { get }
    public var snapshots: [Snapshot] { get }
    public var recentDate: Date { get }
    public var currentSnapshot: Snapshot { get }

    public init(initialSnapshot:recentDate:snapshots:) {}
    public init(initialState:date:snapshots:) {}
    public static func active(since:) -> AppStateHistory {}

    public func take(between:) -> AppStateHistory {}
    public mutating func append(snapshot:) {}
    public var foregroundDuration: TimeInterval { get }
}
```
- **Now**, the API is more concise, with only necessary public methods:
```swift
public struct AppStateHistory {
    public let initialState: AppState
    public var currentState: AppState { get }

    public init(initialState:date:) {}

    public mutating func append(state:at:) {}

    public func state(at:) -> AppState? {}
    public func containsState(during:where:) -> Bool {}
    public func foregroundDuration(during:) -> TimeInterval {}
}
```

#### 🎁 New `DatadogInternal` Assertion Helper
This PR also introduces a new Datadog assertion helper to validate threading assumptions during initialization. It is designed to be generic, allowing other products to use it for validating their own assumptions as well:
```swift
/// Datadog assertion helper.
///
/// - Ensures assumptions that should **never** be violated.
/// - **Fails only in development** to catch mistakes early.
/// - **Never** crashes customer apps in production, even if they enable assertions in Release builds.
public func dd_assert(_:_:) { /**/ }
```

### Review checklist

### DIFF
--- a/Datadog/Datadog.xcodeproj/project.pbxproj
+++ b/Datadog/Datadog.xcodeproj/project.pbxproj
@@ -512,6 +512,8 @@
 		6179FFDE254ADBEF00556A0B /* ObjcAppLaunchHandler.h in Headers */ = {isa = PBXBuildFile; fileRef = 6179FFD1254ADB1100556A0B /* ObjcAppLaunchHandler.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		617B953D24BF4D8F00E6F443 /* RUMMonitorTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 617B953C24BF4D8F00E6F443 /* RUMMonitorTests.swift */; };
 		617B954224BF4E7600E6F443 /* RUMMonitorConfigurationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 617B954124BF4E7600E6F443 /* RUMMonitorConfigurationTests.swift */; };
+		618031F82D6DC430007027E3 /* Threading.swift in Sources */ = {isa = PBXBuildFile; fileRef = 618031F72D6DC430007027E3 /* Threading.swift */; };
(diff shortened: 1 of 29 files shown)


maxep (approved):
Looks great, nothing to add 👍

PR #2211, opened 2025-02-28
[RUM-8413] Move Reflection code to internal module

mariedm:
### What and why?

This PR moves the reflection logic in the `SessionReplay` module to the `Internal` module. The goal is to make this code available to both `SessionReplay` and `RUM` modules, in anticipation of the Auto-instrumentation work.

### How?

- Moving `Reflector` and `ReflectionMirror` to the internal module
- Making core reflection types and methods `public` where necessary

### Review checklist

### DIFF
--- a/DatadogInternal/Sources/Utils/ReflectionMirror.swift
+++ b/DatadogInternal/Sources/Utils/ReflectionMirror.swift
@@ -9,7 +9,7 @@ import Foundation
 /// A representation of the substructure and display style of an instance of
 /// any type.
 ///
-/// The `ReflectionMirror` defers from the standard ``Mirror`` in some key
+/// The `ReflectionMirror` differs from the standard ``Mirror`` in some key
 /// aspect of its implementation:
 ///
 /// ## Display style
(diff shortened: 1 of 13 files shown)


DatadogInternal/Sources/Utils/ReflectionMirror.swift
```diff
@@ -70,25 +70,25 @@ internal struct ReflectionMirror {
...
     /// reflected subject.
-    let children: Children
+    internal let children: Children
 
     /// A mirror of the subject's superclass, if one exists.
-    var superclassMirror: ReflectionMirror? { _superclassMirror.lazy }
+    internal var superclassMirror: ReflectionMirror? { _superclassMirror.lazy }
 
-    var keyPaths: [String: Int]? { _keyPaths.lazy }
+    internal var keyPaths: [String: Int]? { _keyPaths.lazy }
```
maxep, line 91: These can be `public` as well 👍

maxep (commented):
Thanks for fixing my typos 😄

DatadogInternal/Sources/Utils/ReflectionMirror.swift
maxep: They are similar to the standard `Mirror` interface and can be useful during reflection.

maxep (approved):

PR #2212, opened 2025-02-28
feat: Support disabling INV, custom INV values, and FBC for Flutter

fuzzybinary:
### What and why?

This adds an internal method - `setInternalViewAttribute` - which modifies a specific attribute collection for views. This attribute collection is checked for custom INV Values and Flutter First Build Complete metrics.

This also adds the ability to disable INV by setting its `lastActionPredicate` to nil, which has the factory return a nil INVMetricTracking object.

refs: RUM-8548

### How?

### Review checklist


DatadogRUM/Sources/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -531,7 +551,22 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
         let refreshRateInfo = vitalInfoSampler?.refreshRate
         let isSlowRendered = refreshRateInfo?.meanValue.map { $0 < Constants.slowRenderingThresholdFPS }
         let networkSettledTime = networkSettledMetric.value(with: context.applicationStateHistory)
-        let interactionToNextViewTime = interactionToNextViewMetric.value(for: viewUUID)
+        var interactionToNextViewTime = interactionToNextViewMetric?.value(for: viewUUID) ?? .failure(.disabled)
```
ncreated, line 554: question/ The way we extract `fbc` below makes me wonder on the purpose of `CustomValueINVMetric` - do we really need it? Can't we extract custom `inv` in similar fashion as `fbc`, avoiding the `CustomValueINVMetric` implementation?
```swift
let interactionToNextViewTime = customValue(from: internalAttributes) ??
                                interactionToNextViewMetric?.value(for: viewUUID) ??
                                .failure(.disabled)
```

Also, doesn't this raise a compiler warning (_`var interactionToNextViewTime` is never mutated?_ → `let`)?

ncreated (approved):
👍 💪 Looks great, only one question / suggestion left.

DatadogRUM/Sources/RUMMonitor/Scopes/RUMViewScope.swift
fuzzybinary: It's not needed no, we can absolutely extract it the way were were in the previous PR. Using the CustomValueINVMetric simplifies this extraction a little bit at the expense of having the extra class.  Up to you which you prefer.

fuzzybinary: I'm with you, I've switched it back.

ncreated (approved):
Looks good for me 👍 🏅

PR #2214, opened 2025-03-03
RUM-8627: Collect App Hitches in the SDK

simaoseica-dd:
### What and why?

This PR creates the engine to collect View Hitches (Slow frames) that occur during the lifecycle of a View.
The proposed solution is defined in the [internal RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/4722163962/RFC+-+UI+Slowness).

### How?

To scale the entities that observe the render loop, it was created the `DisplayLinker`. 
Any reader (`RenderLoopReader`) that wants to be notified about frame rendering can register to the `DisplayLinker`:
`displayLinker.register(vitalRefreshRateReader)`

This PR lists two `RenderLoopReader`:
- `VitalRefreshRateReader` - A class that reads the refresh rate (frames per second)
- `ViewHitchesReader` - A class that reads View Hitches or Slow frames.

For a better understanding of the relationships, the figure below depicts the idea:
RenderLoopObserver

**Note:**
This is the first part of three (1/3) to integrate the new feature of Slow frames. 
Next tasks will tackle the addition of this information to View events and process ratios of slowness during the view's lifecycle.

### Review checklist


Datadog/Datadog.xcodeproj/project.pbxproj
```diff
@@ -2851,7 +2869,7 @@
 		D2216EC22A96632F00ADAEC8 /* FeatureBaggageTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = FeatureBaggageTests.swift; sourceTree = "<group>"; };
 		D22442C42CA301DA002E71E4 /* UIColor+SessionReplay.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "UIColor+SessionReplay.swift"; sourceTree = "<group>"; };
 		D224430C29E95D6600274EC7 /* CrashReportReceiverTests.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = CrashReportReceiverTests.swift; sourceTree = "<group>"; };
-		D22743E829DEC9A9001A7EF9 /* RUMDataModelMocks.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = RUMDataModelMocks.swift; sourceTree = "<group>"; };
+		D22743E829DEC9A9001A7EF9 /* RUMDataModelMocks.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; name = RUMDataModelMocks.swift; path = ../RUMDataModelMocks.swift; sourceTree = "<group>"; };
```
ncreated, line 2872: question/ Is this intentional change? The `../` path looks odd, what's the rationale?

DatadogRUM/Sources/Feature/RUMFeature.swift
```diff
@@ -97,6 +97,8 @@ internal final class RUMFeature: DatadogRemoteFeature {
                     return nil
                 }
             }(),
+            renderLoopObserver: DisplayLinker(notificationCenter: .default),
```
ncreated, line 100: blocking/ We must use [`configuration.notificationCenter`](https://github.com/DataDog/dd-sdk-ios/blob/5cc5fcd31c1271d3aef92cd8395152ac87dcc836/DatadogRUM/Sources/RUMConfiguration.swift#L316-L317) for dependency injection in integration tests:
```suggestion
            renderLoopObserver: DisplayLinker(notificationCenter: configuration.notificationCenter),
```

This way, when adding integration coverage we can control this `notificationCenter`:
```swift
var config: RUM.Configuration = .mockAny()
config.notificationCenter = custom

RUM.enable(with: config, in: core)
```

DatadogRUM/Sources/Feature/RUMFeature.swift
```diff
@@ -97,6 +97,8 @@ internal final class RUMFeature: DatadogRemoteFeature {
...
+            viewHitchesMetricFactory: { ViewHitchesReader(hangThreshold: configuration.appHangThreshold) },
```
ncreated, line 101: question/ Why using [`appHangTreshold`](https://github.com/DataDog/dd-sdk-ios/blob/5cc5fcd31c1271d3aef92cd8395152ac87dcc836/DatadogRUM/Sources/RUMConfiguration.swift#L129-L141)? It has the default value of `nil` (disabled) and minimum allowed value of `0.1` which might be too high for app hitches, no?

DatadogRUM/Sources/RUMVitals/RenderLoop/RenderLoopObserver.swift
```diff
@@ -0,0 +1,102 @@
...
+    private var displayLink: FrameInfoProvider?
+    private let notificationCenter: NotificationCenter
+
+    init(notificationCenter: NotificationCenter) {
+        self.notificationCenter = notificationCenter
+
+        notificationCenter.addObserver(
+            self,
+            selector: #selector(appWillResignActive),
+            name: UIApplication.willResignActiveNotification,
```
ncreated, line 39: convention/ We should use [`ApplicationNotifications`](https://github.com/DataDog/dd-sdk-ios/blob/5cc5fcd31c1271d3aef92cd8395152ac87dcc836/DatadogInternal/Sources/Context/ApplicationNotifications.swift#L16):
```suggestion
            name: ApplicationNotifications.willResignActive,
```

DatadogRUM/Sources/RUMVitals/RenderLoop/RenderLoopObserver.swift
```diff
@@ -0,0 +1,102 @@
...
+            object: nil
+        )
+        notificationCenter.addObserver(
+            self,
+            selector: #selector(appDidBecomeActive),
+            name: UIApplication.didBecomeActiveNotification,
```
ncreated, line 45: convention/ We should use [`ApplicationNotifications`](https://github.com/DataDog/dd-sdk-ios/blob/5cc5fcd31c1271d3aef92cd8395152ac87dcc836/DatadogInternal/Sources/Context/ApplicationNotifications.swift#L16):
```suggestion
            name: ApplicationNotifications.didBecomeActive,
```

DatadogRUM/Sources/RUMVitals/RenderLoop/ViewHitchesReader.swift
```diff
@@ -0,0 +1,94 @@
...
+ */
+
+import UIKit
+
+public typealias Hitch = (start: Int64, duration: Int64)
+
+public protocol ViewHitchesMetric {
+    var hitchesDuration: Double { get }
+    var hitches: [Hitch] { get }
+}
```
ncreated, line 14: blocking/ It doesn't seem to be needed in RUM's public API, no?
```suggestion
internal typealias Hitch = (start: Int64, duration: Int64)

internal protocol ViewHitchesMetric {
    var hitchesDuration: Double { get }
    var hitches: [Hitch] { get }
}
```

DatadogRUM/Sources/RUMVitals/RenderLoop/ViewHitchesReader.swift
```diff
@@ -0,0 +1,94 @@
...
+    private var nextFrameTimestamp: Double = 0
+
+    private var hangThreshold: TimeInterval
+    private var acceptableLatency: TimeInterval
+
+    /// Amount of time when the frames are rendered too late.
+    private var _hitchesDuration: Double = 0.0
+    private var _hitches: [Hitch] = []
+    var hitchesDuration: Double { queue.sync { self._hitchesDuration } }
+    var hitches: [Hitch] { queue.sync { self._hitches } }
```
ncreated, line 42: question/ Is this ever read? I assume this isn't yet completed feature because we're missing exporting this values to RUM View event, right?

Depending on what's the future plan, I think it would make sense to read both in single `queue.sync {}` rather than blocking the queue twice for information that will always be read in pair. WDYT?

DatadogCore/Tests/Datadog/RUM/RUMVitals/DisplayLinkerTests.swift
```diff
@@ -0,0 +1,103 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import XCTest
+@testable import DatadogRUM
+
+final class DisplayLinkerTests: XCTestCase {
```
ncreated, line 10: sanity-check/ These tests bring symptoms of flakiness - let's make sure it passes 1K reps in local 🙏 so we don't introduce issues in CI.

ncreated (changes requested):
🏅 Really well done ! I like the proposed `DisplayLinker` model 👍 and the reuse we did for FPS reader.

I left few feedbacks, let me know what do you think.

DatadogRUM/Sources/Feature/RUMFeature.swift
simaoseica-dd: The idea is to not record hitches when we are collecting that info via App hangs (so, the defined threshold). If disabled, it will cap hitches using the Frozen frames threshold defined in the long tasks.
`hangThreshold` - max duration a hitch can have 
`acceptableLatency` - min duration a hitch can have

Datadog/Datadog.xcodeproj/project.pbxproj
simaoseica-dd: While organizing the files, it created shortcuts. fixed.

DatadogRUM/Sources/Feature/RUMFeature.swift
simaoseica-dd: Perfect. ty

DatadogRUM/Sources/RUMVitals/RenderLoop/ViewHitchesReader.swift
simaoseica-dd: good catch. ty

DatadogRUM/Sources/RUMVitals/RenderLoop/ViewHitchesReader.swift
simaoseica-dd: Yes, this is the first part of 3 tasks for Slow frames. Added this note to the PR description

DatadogRUM/Sources/Feature/RUMFeature.swift
ncreated: > `hangThreshold` - **max** duration a hitch can have

Ah, so it's **max** threshold, got it 👍. Looks good 👌.

DatadogRUM/Sources/RUMVitals/RenderLoop/ViewHitchesReader.swift
ncreated: > Added this note to the PR description

💯 🏅

ncreated (approved):
🎯 💯

PR #2216, opened 2025-03-04
RUM-8042 Split Upload Quality Metric per Track

maxep:
### What and why?

Aggregated upload quality of all tracks will be difficult to analyse, especially in term of blockers: upload blocker will bock all tracks simultaneously.

We should split de metric per track

### How?

Each upload worker already provide the track name while reporting their upload metric. The aggregation in the `SessionEndedMetric` takes it into account to split the `upload_quality` attribute.

### Review checklist

### DIFF
--- a/DatadogRUM/Sources/SDKMetrics/SessionEndedMetric.swift
+++ b/DatadogRUM/Sources/SDKMetrics/SessionEndedMetric.swift
@@ -103,7 +103,7 @@ internal class SessionEndedMetric {
     private var wasStopped = false
 
     /// Information about the upload quality during the session.
-    private var uploadQuality: Attributes.UploadQuality
+    private var uploadQuality: [String: Attributes.UploadQuality] = [:]
 
     /// If `RUM.Configuration.trackBackgroundEvents` was enabled for this session.
     private let tracksBackgroundEvents: Bool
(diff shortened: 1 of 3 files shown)


DatadogRUM/Sources/SDKMetrics/SessionEndedMetric.swift
```diff
@@ -212,6 +207,16 @@ internal class SessionEndedMetric {
     /// - Parameters:
     ///   - attributes: The upload quality attributes
     func track(uploadQuality attributes: [String: Encodable]) {
+        guard let track = attributes[UploadQualityMetric.track] as? String else {
+            return
+        }
```
mariedm, line 212: I imagine that in theory, the track should always be present, right?

DatadogRUM/Sources/SDKMetrics/SessionEndedMetric.swift
```diff
@@ -344,7 +349,8 @@ internal class SessionEndedMetric {
         }
 
         /// Information about the upload quality during the session.
-        let uploadQuality: UploadQuality
+        /// The upload quality is splitting between upload track name.
+        let uploadQuality: [String: UploadQuality]
```
mariedm, line 353: nit suggestion:

```suggestion
/// Tracks upload quality during the session, aggregating them by track name.
/// Each track reports its own upload quality metrics.
        let uploadQuality: [String: UploadQuality]
```

mariedm (approved):
LGMT! 🙌

Just one question and a nit suggestion, nothing blocking.

DatadogRUM/Sources/SDKMetrics/SessionEndedMetric.swift
maxep: Yes, it should always succeed but it's loosely coupled.

mariedm (approved):

Issue #2190, opened 2025-02-11
Fork of opentelemetry-swift is out of date


jbluntz:
Are there any plans to update http://github.com/datadog/opentelemetry-swift-packages to the latest in upstream, and update the dependency here?  There are a few notable additions that we'd really like to take advantage of, including swift concurrency support (with TaskLocalContextManager) and closure-based APIs for starting and activating spans.

calebhearth:
@ncreated It also appears to no longer compile under Xcode 16.3.0-beta due to the [Span extension for `==` not including `Self` as one of the types](https://github.com/DataDog/opentelemetry-swift-packages/blob/1.6.0/Sources/OpenTelemetryApi/Trace/Span.swift#L77)

> Member operator '==' of protocol 'Span' must have at least one argument of type 'Self'

That's [resolved](https://github.com/DataDog/opentelemetry-swift-packages/blob/main/Sources/OpenTelemetryApi/Trace/Span.swift#L116) on the main branch as well as upstream.

maxep:
Hey @calebhearth 👋 

Thanks for the report, we will update the otel dependency soon. We will keep you posted.

cameroncooke:
My project now won't compile due to a compiler issue in the older version of `opentelemetry-swift` that Datadog utilises. Do we have a timescale for a fix?

maxep:
Hey @cameroncooke,

Can you confirm that it fails to compile on Xcode 16.3 Beta only ?
It should build properly on Xcode 16.2

We are actively working on this and we should be able to release a fix in the coming weeks.
Sorry for the inconvenience.

cameroncooke:
Yes it’s just a beta issue. The issue for me is Xcode 16.2 is so broken I’m
desperate to use any new version including beta versions that Apple
release.

Cameron.

jbluntz:
@maxep can I ask if it's possible to include a podspec for the Concurrency target in the update?  I noticed yesterday that it is accessible via SPM but not cocoapods...

maxep:
Hey @jbluntz 👋 

What Concurrency target? I don't see any in the [Package.swift](https://github.com/DataDog/opentelemetry-swift-packages/blob/main/Package.swift) 🤔

jbluntz:
Looks like it's swift 5.9+: https://github.com/open-telemetry/opentelemetry-swift/blob/main/Package@swift-5.9.swift

ncreated:
Hey everyone! 👋

[...]

PR #1881, opened 2024-06-04
chore: Disable flaky `OTelSpan.testSetActive_givenParentSpan()` test

ncreated:
### What and why?

📦 ❄️ This PR disables flaky OTel test ([`testSetActive_givenParentSpan()`](https://github.com/DataDog/dd-sdk-ios/blob/b142cf9bbd48c97ffbfa9fda3092e52af50b22a4/DatadogTrace/Tests/OpenTelemetry/OTelSpanTests.swift#L252)). This is to address recent CI failures.

### How?

Disabled failed test and created `RUM-4795` to re-enable it when possible.

I also revamped the whole `OTelSpanTests` suite to take dependency on `FeatureScope` rather than core mock. This is to address mid-term goals from https://github.com/DataDog/dd-sdk-ios/pull/1744 - having less and simpler dependency in tests so they run faster, more stable and are easier to debug.

### Review checklist
- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [ ] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes

### Custom CI job configuration (optional)
- [ ] Run unit tests for Core, RUM, Trace, Logs, CR and WVT
- [ ] Run unit tests for Session Replay
- [ ] Run integration tests
- [ ] Run smoke tests
- [ ] Run tests for `tools/`

PR #2217, opened 2025-03-04
RUM-8558 Update DatadogTrace to OTel API `1.13.0`

ncreated:
### What and why?

📦 This PR updates `DatadogTrace` to use `OpenTelemetryApi` [**1.13.0**](https://github.com/DataDog/opentelemetry-swift-packages/releases/tag/1.13.0).

This addresses [#2190](https://github.com/DataDog/dd-sdk-ios/issues/2190).

### How?

The new `OpenTelemetryApi` version was published in [opentelemetry-swift-packages#27](https://github.com/DataDog/opentelemetry-swift-packages/pull/27), and this PR integrates it into `dd-sdk-ios`.

#### Platform Versions Alignment

Since `1.6.0` (our previous `OpenTelemetryApi` dependency version), `open-telemetry-swift` introduced a breaking change in supported platforms, raising the minimum deployment targets from `.iOS(11)` and `.tvOS(11)` to `.iOS(13)` and `.tvOS(13)`. Because our SDK supports `.iOS(12)` and `.tvOS(12)`, this required adding a patch in `Package.swift` to conditionally adjust `platforms` when the `OTEL_SWIFT` environment variable is set. I used this as an opportunity to add comments and better document this flag. In our repository, this is leveraged in `BenchmarkTests`.

#### New `OTelSpanBuilder` APIs

This update introduces the following `OTelSpanBuilder` APIs. Their implementation in this PR follows [otel-swift](https://github.com/open-telemetry/opentelemetry-swift/blob/af8a035d3c462fcbcdb0b96b862966bb7b68eca7/Sources/OpenTelemetrySdk/Trace/SpanBuilderSdk.swift#L164-L187), delegating active span management to `OpenTelemetry.instance.contextProvider`:

```swift
func withActiveSpan<T>(_ operation: (any OpenTelemetryApi.SpanBase) throws -> T) rethrows -> T

func withActiveSpan<T>(_ operation: (any OpenTelemetryApi.SpanBase) async throws -> T) async rethrows -> T
```

This PR adds test coverage for these new APIs and also re-enables `OTelSpanTests/testSetActive_givenParentSpan()`, which was disabled in [#1881](https://github.com/DataDog/dd-sdk-ios/pull/1881) due to flaky performance. The updated `OpenTelemetryApi` refines active span management, so performance should now be stable.

#### New `OTelSpan` APIs

The update also introduces new `OTelSpan` APIs:
```swift
func recordException(_ exception: any OpenTelemetryApi.SpanException, attributes: [String: OpenTelemetryApi.AttributeValue], timestamp: Date)

func recordException(_ exception: any OpenTelemetryApi.SpanException, attributes: [String: OpenTelemetryApi.AttributeValue])

func recordException(_ exception: any OpenTelemetryApi.SpanException, timestamp: Date)

func recordException(_ exception: any OpenTelemetryApi.SpanException)
```

These APIs are **not implemented** in this PR because their proper implementation [depends on](https://github.com/open-telemetry/opentelemetry-swift/blob/1.13.0/Sources/OpenTelemetrySdk/Trace/RecordEventsReadableSpan.swift#L356) the `addEvent()` API, which is not yet supported in our OTel integration.

#### Deprecated OTel APIs

The latest `OpenTelemetryApi` deprecates the following attribute array creation methods:
```swift
.stringArray(_:)
.boolArray(_:)
.intArray(_:)
.doubleArray(_:)
```
These have been replaced with a more uniform `.array(.init(values: []))` API. This PR updates our test suite to use the new `.array()` format and removes dependencies on deprecated APIs.

### Review checklist

### DIFF
--- a/DatadogTrace/Sources/OpenTelemetry/NOPOTelSpanBuilder.swift
+++ b/DatadogTrace/Sources/OpenTelemetry/NOPOTelSpanBuilder.swift
@@ -55,4 +55,20 @@ internal class NOPOTelSpanBuilder: SpanBuilder {
     func setActive(_ active: Bool) -> Self {
         return self
     }
+
+    func withActiveSpan<T>(_ operation: (any OpenTelemetryApi.SpanBase) throws -> T) rethrows -> T {
+        let span = startSpan()
+        defer { span.end() }
+        return try operation(span)
+    }
+
+#if canImport(_Concurrency)
+    /// Ref.: https://github.com/open-telemetry/opentelemetry-swift/issues/578
+    @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+    func withActiveSpan<T>(_ operation: (any OpenTelemetryApi.SpanBase) async throws -> T) async rethrows -> T {
+        let span = startSpan()
+        defer { span.end() }
+        return try await operation(span)
+    }
+#endif
 }
--- a/Package.swift
+++ b/Package.swift
@@ -3,21 +3,34 @@
 import PackageDescription
 import Foundation
 
-let opentelemetry = ProcessInfo.processInfo.environment["OTEL_SWIFT"] != nil ? 
+// If the `OTEL_SWIFT` environment variable is set, `dd-sdk-ios` will be compiled against `OpenTelemetryApi` 
+// from https://github.com/open-telemetry/opentelemetry-swift, which includes the full OpenTelemetry SDK.
+// Otherwise, it will use our lightweight mirror from https://github.com/DataDog/opentelemetry-swift-packages.
+//
+// This split is driven by feedback from https://github.com/DataDog/dd-sdk-ios/issues/1877, where 
+// users reported that fetching the full OpenTelemetry SDK significantly increased dependency size. 
+//
+// By using this environment variable, `dd-sdk-ios` consumers can choose whether to depend on the entire 
+// OpenTelemetry SDK or just the API. This remains necessary until OpenTelemetry officially separates 
+// the API and SDK packages (see https://github.com/open-telemetry/opentelemetry-swift/issues/486).
+let useOTelSwiftPackage = ProcessInfo.processInfo.environment["OTEL_SWIFT"] != nil
+
+let opentelemetry = useOTelSwiftPackage ? 
     (name: "opentelemetry-swift", url: "https://github.com/open-telemetry/opentelemetry-swift.git") :
     (name: "opentelemetry-swift-packages", url: "https://github.com/DataDog/opentelemetry-swift-packages.git")
+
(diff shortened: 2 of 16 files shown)


DatadogTrace/Sources/OpenTelemetry/NOPOTelSpanBuilder.swift
```diff
@@ -55,4 +55,20 @@ internal class NOPOTelSpanBuilder: SpanBuilder {
     func setActive(_ active: Bool) -> Self {
         return self
     }
+
+    func withActiveSpan<T>(_ operation: (any OpenTelemetryApi.SpanBase) throws -> T) rethrows -> T {
+        let span = startSpan()
```
simaoseica-dd, line 60: question: Shouldn't it call `.setActive()` taking into account the caller is expecting it?
(`caller.withActiveSpan()`)

simaoseica-dd (approved):
Great work overall in this update to OTel `1.13.0` 🎖️

DatadogTrace/Sources/OpenTelemetry/NOPOTelSpanBuilder.swift
ncreated: This is no-op implementation, so it doesn't really matter. The only thing that matters is that we invoke the `operation`, which we do 👍 
```swift
try operation(span)
```

Package.swift
```diff
@@ -3,21 +3,34 @@
...
+// If the `OTEL_SWIFT` environment variable is set, `dd-sdk-ios` will be compiled against `OpenTelemetryApi` 
+// from https://github.com/open-telemetry/opentelemetry-swift, which includes the full OpenTelemetry SDK.
+// Otherwise, it will use our lightweight mirror from https://github.com/DataDog/opentelemetry-swift-packages.
+//
+// This split is driven by feedback from https://github.com/DataDog/dd-sdk-ios/issues/1877, where 
+// users reported that fetching the full OpenTelemetry SDK significantly increased dependency size. 
+//
+// By using this environment variable, `dd-sdk-ios` consumers can choose whether to depend on the entire 
+// OpenTelemetry SDK or just the API. This remains necessary until OpenTelemetry officially separates 
+// the API and SDK packages (see https://github.com/open-telemetry/opentelemetry-swift/issues/486).
```
maxep, line 15: Nice, thanks!

maxep (approved):
👌

markst:
Can you explain how we pass the `OTEL_SWIFT` env var when compiling during Xcode? I seems to only be possible to pass when precompiling the package using `xcodebuild`.

markst:
or whether we could use a tool such as:
brew install segment-integrations/formulae/swift-create-xcframework
in order to build Package.swift to XCFramework

PR #2223, opened 2025-03-07
RUM-8931 Simplify host sanitizer logic

maciejburda:
### What and why?

Simplifes host sanitizing logic to include edge cases like `customprotocol://name`, which is valid swift `URL`.

### How?

Simplified regex and updated tests.

### Review checklist


simaoseica-dd (approved):
👌

PR #2229, opened 2025-03-12
RUM-8442 Fix Benchmark Metrics Exporter

maxep:
### What and why?

Fix metrics tags issue introduced in https://github.com/DataDog/dd-sdk-ios/pull/2218

### How?

- Merge the Otel Metric Provider's resources with tags.
- Register the `MeterProvider` and `TracerProvider` 

### Review checklist

### DIFF
--- a/BenchmarkTests/Runner/BenchmarkMeter.swift
+++ b/BenchmarkTests/Runner/BenchmarkMeter.swift
@@ -1,0 +1,83 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+import DatadogInternal
+import DatadogBenchmarks
+import OpenTelemetryApi
+import OpenTelemetrySdk
+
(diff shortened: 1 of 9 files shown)


simaoseica-dd (approved):

BenchmarkTests/Runner/BenchmarkMeter.swift
```diff
@@ -0,0 +1,83 @@
...
+    @discardableResult
+    func observeMemory() -> OpenTelemetryApi.DoubleObserverMetric {
+        let memory = Memory(queue: queue)
+        return meter.createDoubleObservableGauge(name: "ios.benchmark.memory") { metric in
+            // report the maximum memory footprint that was recorded during push interval
+            if let value = memory.aggregation?.max {
+                metric.observe(value: value, labelset: .empty)
+            }
+
+            memory.reset()
```
simaoseica-dd, line 34: There is no need to `reset()` if the value is nil, right?

simaoseica-dd (approved):
Great work! 🎖️

mariedm (approved):
LGMT! 👍

BenchmarkTests/Runner/BenchmarkMeter.swift
maxep: No indeed, in practice it won't have an effect.

PR #2230, opened 2025-03-12
RUM-8042 Replace Upload Quality with Batch Blocked telemetry

maxep:
### What and why?

Specifications of `Upload Quality` attributes in `Session Ended` metric telemetry are not suitable for graphing.
Instead, I propose adding a new Batch metric event to report batches blocked during upload. Currently, we have the `Batch Deleted` telemetry metric, from which we can compute the count of batches deleted due to `intake-code-202` (successful upload). We can add a `Batch Blocked` metric to report when a batch is blocked due to local blockers (low power, offline, etc.) or network failures. This will provide visibility into why batches are not uploaded and allow us to compute the ratio of successful uploads versus failures. This new telemetry event should be sampled at the same rate as `Batch Deleted`: 1.5%.

The complex `rse.upload_quality` attribute can then be replaced by a simple `rse.upload_cycle.<logs|rum|*>` that track upload cycle per track. In this new definition, an Upload Cycle can include multiple batch upload.

### How?

Similar to `BatchDeletedMetric`, the `BatchBlockedMetric` will report reasons for blocking batch upload, or failures to upload (network or http)

### Review checklist


mariedm (approved):
LGMT!

simaoseica-dd (dismissed):
LGTM 👌

barboraplasovska (dismissed):
🏅

DatadogCore/Sources/SDKMetrics/BatchMetrics.swift
```diff
@@ -130,3 +130,23 @@ internal enum BatchClosedMetric {
...
+    /// Note: the "[Mobile Metric]" prefix is added when sending this telemetry in RUM.
+    static let name = "Batch Blocked"
+    /// Metric type value.
+    static let typeValue = "batch blocked"
+    /// The sample rate for this metric.
+    /// It is applied in addition to the telemetry sample rate (20% by default).
+    static let sampleRate: Float = 1.5 // 1.5%
+    /// The key for uploader's current delay.
+    static let uploaderDelayKey = "uploader_delay.current"
+    /// The key for count of bacthes being blocked.
```
jonathanmos, line 146: nit: The key for the number of batches being blocked.

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
mariedm, line 1: If I read this correctly, we use to send `sendUploadQualityMetric` both in cases of success and failure. Now we send `sendUploadCycleMetric` before every upload, and track failures with `sendBatchBlockedMetric`. So we don't explicitly send a telemetry in case of success. I imagine we can still graph the success rate, but is this intentional?

mariedm (commented):
Just a couple of small questions, otherwise LGMT!

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
```diff
@@ -73,20 +73,23 @@ internal class DataUploadWorker: DataUploadWorkerType {
             let context = contextProvider.read()
             let blockersForUpload = uploadConditions.blockersForUpload(with: context)
             let isSystemReady = blockersForUpload.isEmpty
-            let files = isSystemReady ? fileReader.readFiles(limit: maxBatchesPerUpload) : nil
-            if let files = files, !files.isEmpty {
+            let files = fileReader.readFiles(limit: maxBatchesPerUpload)
+
+            if !files.isEmpty && isSystemReady {
```
Valpertui, line 78: Can we try to asses if this change might have a negative performance impact ?  
I don't think it would because `contentsOfDirectory` (urderlying Filesystem method used) only get URLs, and not the content of each file, but maybe it's worth checking using `maxBatchesPerUpload` files everytime ?
Since we only use `files` to get their count for the telemetry, it would be bad if this also had a performance impact.

Valpertui (approved):

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
```diff
@@ -73,21 +73,23 @@ internal class DataUploadWorker: DataUploadWorkerType {
...
                 DD.logger.debug("⏳ (\(self.featureName)) Uploading batches...")
                 self.backgroundTaskCoordinator?.beginBackgroundTask()
                 self.uploadFile(from: files.reversed(), context: context)
                 sendUploadCycleMetric()
             } else {
-                let batchLabel = files?.isEmpty == false ? "YES" : (isSystemReady ? "NO" : "NOT CHECKED")
+                let batchLabel = files.isEmpty ? "NO" : "YES"
```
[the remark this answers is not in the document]
maxep: No it doesn't, it was just for console printing. The drawback of this change is that we will check for batches even if the upload is blocked, but the impact is minimal as we only list the files.

DatadogCore/Sources/Core/Upload/DataUploadWorker.swift
simaoseica-dd: The `files` with their `InputStream` will be needed for the `self.uploadFile(from: files.reversed(), context: context)`

PR #2231, opened 2025-03-13
RUM-8454 Trace Benchmark Scenario

barboraplasovska:
### What and why?

To ensure our SDKs maintain optimal performance and do not introduce regressions, we will establish a robust benchmarking and testing framework. This will include performance benchmarks and end-to-end tests that run for every new release. This PR introduces new scenarios and UI to test the Trace feature.

### How?

Introduction of a configurable scenario that allows us to simulate tracing in a real application.

The UI lets you configure parameters to generate simple or complex traces. You can set the root span name, specify how many child spans are generated at each level, adjust the trace depth, and control the delay between each child to simulate processing time. In the screenshot below, we set children count to 2, depth to 3, and child delay to 2, which produces a trace where each level has 2 child spans.

Trace Scenario UI

Resulting traces

### Review checklist


BenchmarkTests/Runner/Scenarios/Trace/UI/TraceContentView.swift
```diff
@@ -0,0 +1,173 @@
...
+                        OTLogFields.stack: "Foo.swift:42",
+                    ]
+                )
+            }
+
+            wait(seconds: 0.5)
+
+            queue2.sync {
+                sendSpanTree(parent: rootSpan, currentLevel: 0, maxDepth: depth)
+            }
```
maxep, line 135: /suggestion No need to jump on a secondary queue, you can create the children on the same thread.

BenchmarkTests/Runner/Scenarios/Trace/UI/TraceContentView.swift
```diff
@@ -0,0 +1,173 @@
...
+                    fields: [
+                        OTLogFields.event: "error",
+                        OTLogFields.errorKind: "Simulated error",
+                        OTLogFields.message: "Describe what happened",
+                        OTLogFields.stack: "Foo.swift:42",
+                    ]
+                )
+            }
+
+            wait(seconds: 0.5)
```
maxep, line 131: /suggestion Lets apply the `childDelay` as well. Also, there is no need for an instead method here, you can directly call `Thread.sleep`.

BenchmarkTests/Runner/Scenarios/Trace/UI/TraceContentView.swift
```diff
@@ -0,0 +1,173 @@
...
+    private let queue1 = DispatchQueue(label: "com.datadoghq.benchmark-tracing1")
+    private let queue2 = DispatchQueue(label: "com.datadoghq.benchmark-tracing2")
+
+    init() {
+        operationName = "iOS Benchmark span operation"
+        resourceName = "iOS Benchmark span resource"
+        isError = false
+        depth = 1
+        childrenCount = 0
+        childDelay = 1
```
maxep, line 31: /suggestion We should use milliseconds from more realistic traces, we could use 100ms by default with 50ms steps.

BenchmarkTests/Runner/Scenarios/Trace/UI/TraceContentView.swift
```diff
@@ -0,0 +1,173 @@
...
+                        TextField("Depth:", value: $depth, formatter: NumberFormatter())
+                            .textFieldStyle(RoundedBorderTextFieldStyle())
+                            .frame(width: 80)
+                            .keyboardType(.numberPad)
+                        Stepper("", value: $depth, in: 1 ... 100, step: 1)
+                            .frame(width: 80)
+                    }
+
+                    HStack {
+                        Text("Child delay:")
```
maxep, line 70: /suggestion Lets specify the unit here.

maxep (commented):
Amazing, it looks great! I have left few suggestions.

mariedm (approved):
Great work!! 🙌

barboraplasovska:
Since @maxep's PR regarding metrics was merged, I had to solve merge conflicts in the file `BenchmarkTests/BenchmarkTests.xcodeproj/project.pbxproj`. #

mariedm (approved):

Issue #2209, opened 2025-02-27
Wrapping up a SwiftUI view that utilize `interactiveDismissDisabled` with `UIHostingController` causing event drops


danibachar:
### Describe the bug

I have a simple UIKit based app (see attached project), where I wrap my main SwiftUI view in a `UIHostingController`.

When I'm using the `interactiveDismissDisabled` view modifier on the view, events are start dropping - see attached image for the log break point (I also added the warning DadaDog produce).

![Image](https://github.com/user-attachments/assets/e8e76701-6cd8-4e90-b708-7810ef40254e)

It seems like the combination of the `UIHostingController` with the `interactiveDismissDisabled` view modifier causes DataDog to loose tracking of the view hierarchy.

Notes:
- It does not matter if we push or present the wrapper UIHostingController, events are dropped in both cases.
- When using the `interactiveDismissDisabled` in a pure SwiftUI environment, events are not dropped.
- I have added `trackRUMView` on the swiftUI view
- **Super important when I replaced the `trackRUMView` with `onAppear`/`onDisappear` and `RUMMonitor.shared().startView`/`RUMMonitor.shared().stopView`. Events start sending again.**Which is kind of weird? the `trackRUMView` action seems to do something similar
- **For the above workaround to work, I need to call the `RUMMonitor.shared().startView`/`RUMMonitor.shared().stopView` on the exact view that utilize the `interactiveDismissDisabled` view modifier, using it on "upper" view in the heirarchy would not work**

Please see steps to reproduce for more information and example project

### Reproduction steps

1. Download the attached zipped project - [data-dog-rum-bug-uikit.zip](https://github.com/user-attachments/files/19013488/data-dog-rum-bug-uikit.zip)
2. Open the project
3. Go to DatadogManager and assign your rum-id and client-token (see attached image) ![Image](https://github.com/user-attachments/assets/894d3131-785f-4b93-b264-be75d2ae3757)
4. Run the application, try clicking `Push` and `Present`
5. In the popped up screen click the `Click Me` button to trigger event sending.
6. Note the event dropped

### SDK logs

[...]

PR #2236, opened 2025-03-20
RUM-8847 Remove automatic `RUMUntrackedModal`

maxep:
### What and why?

The logic to ignore modal presentation automatically seems wrong and prone to issue such as #2209.
Ignoring a view by returning `isUntrackedModal: true` in the predicate is correct but ignoring a view-controller automatically when it has `isModalInPresentation: true` is not.

From Apple documentation:
> The default value of this property is [false](https://developer.apple.com/documentation/swift/false). When you set it to [true](https://developer.apple.com/documentation/swift/true), UIKit ignores events outside the view controller's bounds and prevents the interactive dismissal of the view controller while it is onscreen.

The `isModalInPresentation` is about dismissal interaction, but the view-controller could be in any navigation stack. 

### How?

Remove ignoring view-controllers with `isModalInPresentation: true` in the view auto-instrumentation.

### Review checklist

### DIFF
--- a/DatadogRUM/Sources/Instrumentation/Views/RUMViewsHandler.swift
+++ b/DatadogRUM/Sources/Instrumentation/Views/RUMViewsHandler.swift
@@ -242,17 +242,6 @@ extension RUMViewsHandler: UIViewControllerHandler {
                     instrumentationType: .swiftui
                 )
             )
-        } else if #available(iOS 13, tvOS 13, *), viewController.isModalInPresentation {
-            add(
-                view: .init(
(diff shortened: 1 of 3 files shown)


fuzzybinary (dismissed):
This will be a breaking change for Flutter hybrid apps for sure, but technically might be breaking for the SDK in general, as this would be changing default behavior.

simaoseica-dd (dismissed):

simaoseica-dd (approved):
LGTM 👍

Issue #2239, opened 2025-03-21
clearAllData emits errors


martin-muller:
### Describe the bug

Calling `clearAllData` on initialised Datadog SDK emits errors to console, all created files seem to be deleted.

### Reproduction steps

Sample project: https://github.com/martin-muller/dd-test

1. Initialise DD SDK
 
```swift    Datadog.verbosityLevel = .debug

    Datadog.initialize(
        with: Datadog.Configuration(
            clientToken: "{redacted}",
            env: "dev",
            site: .us1
        ),
        trackingConsent: .pending
    )

    Logs.enable()
    Trace.enable()
    RUM.enable(with: RUM.Configuration(applicationID: "{redacted}"))

    SessionReplay.enable(
        with: SessionReplay.Configuration(
            textAndInputPrivacyLevel: .maskAllInputs,
            imagePrivacyLevel: .maskNone,
            touchPrivacyLevel: .show,
            featureFlags: [.swiftui: true]
        )
    )
```

2. Call `clearAllData`

```swift
Datadog.clearAllData()
```

### SDK logs

```
[DATADOG SDK] 🐶 → 12:03:07.602 🔥 [Data Store] Error on clearing all data for `logging`

Error details:
→ type: InternalError
→ message: Path doesn't exist or is not a directory: file:///Users/martinmuller/Library/Developer/CoreSimulator/Devices/87DE0516-5283-4F4E-AEAC-2A8EF8B23678/data/Containers/Data/Application/4ECA2FDD-9826-4E65-A8BC-B26D8E826F44/Library/Caches/com.datadoghq/v2/f5da9c34155155fa15554148a8a2d145170e0a082ddf9cba39ea3a198b2bb947/1/logging/
→ stack: Path doesn't exist or is not a directory: file:///Users/martinmuller/Library/Developer/CoreSimulator/Devices/87DE0516-5283-4F4E-AEAC-2A8EF8B23678/data/Containers/Data/Application/4ECA2FDD-9826-4E65-A8BC-B26D8E826F44/Library/Caches/com.datadoghq/v2/f5da9c34155155fa15554148a8a2d145170e0a082ddf9cba39ea3a198b2bb947/1/logging/
[DATADOG SDK] 🐶 → 12:03:07.605 🔥 [Data Store] Error on clearing all data for `session-replay`

Error details:
→ type: InternalError
[...]

PR #2240, opened 2025-03-24
GH-2239 chore: Refine errors printed from `clearAllData()`

ncreated:
### What and why?

📦 This PR limits errors printed from `clearAllData()` API when SDK `.debug` verbosity is enabled.

Fixes https://github.com/DataDog/dd-sdk-ios/issues/2239

Before:
```swift
// Given
Datadog.initialize(with: coreConfig, trackingConsent: .pending)
Logs.enable()
Trace.enable()
RUM.enable(with: rumConfig)

// When
Datadog.clearAllData()

// Then, if run in freshly installed app, it was printing errors like:
// """
/// Path doesn't exist or is not a directory: file:///Users/.../Caches/com.datadoghq/v2/.../1/logging
/// """
```

Now, no errors are printed.

### How?

The issue was caused by an incorrect assumption in `FeatureDataStore.clearAllData()`: it presumed the directory would always exist, which isn't the case when the SDK launches for the first time.

The fix involves adding an existence check (`ifExist`) to verify the directory's presence before attempting to clear its contents.

### Review checklist

### DIFF
--- a/DatadogCore/Sources/Core/DataStore/FeatureDataStore.swift
+++ b/DatadogCore/Sources/Core/DataStore/FeatureDataStore.swift
@@ -90,8 +90,8 @@ internal final class FeatureDataStore: DataStore {
     func clearAllData() {
         queue.async {
             do {
-                let directory = try self.coreDirectory.coreDirectory.subdirectory(path: self.directoryPath)
-                try directory.deleteAllFiles()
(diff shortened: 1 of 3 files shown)


maxep (approved):
That was quick!

simaoseica-dd (approved):

PR #2257, opened 2025-04-04
RUM-8629: Add Telemetry to Slow Frames

simaoseica-dd:
### What and why?

This PR adds telemetry for Slow Frames, in line with the [internal spec](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/4534173727/RUM+View+Ended+Telemetry+-+Spec). 
The main goal is to validate the distribution of slow frames across views and monitor how frequently special cases occur (such as ProMotion usage, slow frames with minimal delays, buffers reaching maximum capacity).

### How?

**1. Telemetry**

The addition of a new Metric provided an opportunity to scale the telemetry Controller at the end of the View's lifecycle. 
The image below illustrates the updated structure:

controller

- `ViewHitchesMetric` - Captures the "RUM UI Slowness" telemetry, which is sent when the view ends. This metric conforms to the `MetricAttributesConvertible` protocol, enabling the Controller to handle it.
```swift
internal protocol MetricAttributesConvertible {
    var metricName: String { get }
    func asMetricAttributes() -> [String: Encodable]?
}
```

- `ViewEndedController` - Coordinates telemetry tracking within the View scope and sends metrics at the end of the view’s lifecycle. The metrics can be added to its stack and conform to the `MetricAttributesConvertible` protocol.
```swift
class ViewEndedController {
  func add(metric: MetricAttributesConvertible)
  func send()
}
```

**2. Slow Frames improvements:**
Additionally, there were some improvements on the `ViewHitchesReader`, after reviewing internal dogfooding data:
- The state of View Hitches is no longer reset by `UIApplication.willResignActiveNotification` if the ViewScope remains active (e.g. when the control center overlays the app).
- The Slow frame rate is no longer affected by tracked App Hangs.

### Review checklist


DatadogRUM/Sources/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -157,9 +157,19 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
...
+            self.viewEndedMetric.add(
+                metric: ViewHitchesMetric(
+                    maxCount: viewHitchesReader.config.maxCollectedHitches,
+                    slowFrameThreshold: viewHitchesReader.config.acceptableLatency.toInt64Nanoseconds,
+                    maxDuration: viewHitchesReader.config.hangThreshold.toInt64Nanoseconds,
+                    viewMinDuration: Constants.minimumTimeSpentForRates.toInt64Nanoseconds
+                )
+            )
+            dependencies.renderLoopObserver?.register(viewHitchesReader)
+        }
```
ncreated, line 172: suggestion/ This setup doesn't utilize any internal state from `RUMViewScope`, which suggests it may not need to reside there. To streamline the architecture, consider moving the registration of the hitches metric (`.add(...)`) to `RUMFeature`, similar to how the "view ended" metric is handled:

```swift
// in RUMFeature.swift
controller.add(metric: ViewEndedMetric(...))
controller.add(metric: ViewHitchesMetric(...))
```

DatadogRUM/Sources/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -169,6 +179,12 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
         }
     }
 
+    deinit {
+        if let viewHitchesReader {
+            dependencies.renderLoopObserver?.unregister(viewHitchesReader)
```
ncreated, line 184: question/ It looks like the reader might be unregistered twice—once in `process(command:)` when `shouldComplete` is true, and again here. Is there a specific reason for this redundancy?

DatadogRUM/Sources/SDKMetrics/ViewEndedController.swift
```diff
@@ -0,0 +1,116 @@
...
+        sampleRate: SampleRate = ViewEndedController.defaultSampleRate
+    ) {
+        self.telemetry = telemetry
+        self.sampleRate = sampleRate
+    }
+
+    /// Adds a metric to be tracked by telemetry.
+    func add(metric: MetricAttributesConvertible) {
+        metrics[metric.metricName] = metric
+    }
```
ncreated, line 40: blocking/ This raises potential data race concerns in the current implementation. One `.add()` call occurs in `RUMFeature.init()`, which runs on the user thread, while the other originates from `RUMViewScope`, running on the SDK context queue. Given that metrics lacks thread synchronization, this could lead to unsafe concurrent access.

We could address this by either introducing proper synchronization (e.g. locks), or—preferably—by refining the controller to be immutable by design. The latter could be achieved by moving the hitches metric initialization to `RUMFeature`, as suggested in the earlier comment.

DatadogRUM/Sources/SDKMetrics/ViewHitchesMetric.swift
```diff
@@ -0,0 +1,135 @@
...
+        ]
+    }
+}
+
+// MARK: - Exporting Attributes
+
+extension ViewHitchesMetric {
+    /// Container to encode "SlowFrames" data according to the spec.
+    internal struct Attributes: Encodable {
+        /// Duration of the view in nanoseconds (equal to `@view.time_spent`.
```
ncreated, line 82:
```suggestion
        /// Duration of the view in nanoseconds (equal to `@view.time_spent`).
```

ncreated (changes requested):
Very neat 👌. I left few feedbacks and questions with one potential blocker on threading concerns.

DatadogRUM/Sources/RUMVitals/RenderLoop/ViewHitchesReader.swift
```diff
@@ -7,30 +7,52 @@
...
-internal protocol ViewHitchesMetric {
-    var hitchesDataModel: HitchesDataModel { get }
+/**
+ - Parameters:
+   - hitchesCount: Total amount of slow frames collected.
+   - ignoredHitchesCount: Total slow frames that were outside of the min and max boundaries.
+   - didApplyDynamicFraming: Indicates that ProMotion was applied.
+   - ignoredDurationNs: Value indicating the extra duration used to calculate the slow frame rate.
+ */
+internal typealias HitchesTelemetryModel = (hitchesCount: Int, ignoredHitchesCount: Int, didApplyDynamicFraming: Bool, ignoredDurationNs: Int64)
```
maxep, line 38: suggestion/ Could be a struct :)

DatadogRUM/Sources/RUMVitals/RenderLoop/ViewHitchesReader.swift
```diff
@@ -42,61 +64,84 @@ internal final class ViewHitchesReader: ViewHitchesMetric {
...
-    var hitchesDataModel: HitchesDataModel { queue.sync { (hitches: self._hitches, hitchesDuration: self._hitchesDuration) } }
+    private var hitchesDuration: Double = 0.0
+    var dataModel: HitchesDataModel { queue.sync { (hitches: self.hitches, hitchesDuration: self.hitchesDuration) } }
+
+    private var removedHitchesCount = 0
+    private var ignoredHitchesCount = 0
+    private var startFrameRate: CFTimeInterval?
+    private var didApplyDynamicFraming = false
+    var telemetryModel: HitchesTelemetryModel {
+        queue.sync {
```
maxep, line 81: question/ I'm not sure about the threading model here, do you know what call thread will be waiting for the `sync`?

maxep (commented):
Looks good 👍  I agree with Maciek's point.
I have left additional minor question/suggestion.

maxep (approved):

PR #2260, opened 2025-04-09
RUM-9395 fix: Fatal App Hang Duplicates

### LINKED ISSUE #2158: After updating to DataDog 2.20.0, seeing lot many App Hang crashes
### Question

Hi Team, 

We recently updated to DataDog 2.20.0 and seeing lot many errors coming our way. One of them are App hang - with crash. First we thought this is real issue and even we believe now, but when we looked to the attached report, there were no crashes seen to the developer and also all of them were reported exactly at same time (check the time stamp).

This makes us think to learn more about this error report. So here are our questions. 

1. Are all these reports are user facing or sometime can be internal sdk related and can be caught. 
2. If they are real crash, how can they be reported at same time. Since once app crash it takes time to launch and crash again. So technically they can not have same crash. 
3. Would love to more about the crash reporting in details, specially on app hang and crash.

maxep:
### What and why?

After reporting the crash-error at launch, the AppHang report is never removed from the persistent DataStore. Meaning that the same error will be reported at each application launch, until the event expires after 18hrs, or until another report overrides it.

Fix #2158

### How?

Remove the stored App Hang after report at launch.

### Review checklist

### DIFF
--- a/DatadogRUM/Sources/Instrumentation/AppHangs/FatalAppHangsHandler.swift
+++ b/DatadogRUM/Sources/Instrumentation/AppHangs/FatalAppHangsHandler.swift
@@ -61,6 +61,7 @@ internal final class FatalAppHangsHandler {
     }
 
     func reportFatalAppHangIfFound() {
+        // Report pending app hang
         featureScope.rumDataStore.value(forKey: .fatalAppHangKey) { [weak self] (fatalHang: FatalAppHang?) in
             guard let fatalHang = fatalHang else {
                 DD.logger.debug("No pending App Hang found")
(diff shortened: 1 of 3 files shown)


mariedm (dismissed):
Good catch! Thanks for the quick fix 🛠️

maciejburda (approved):

DatadogRUM/Sources/Instrumentation/AppHangs/FatalAppHangsHandler.swift
```diff
@@ -71,6 +72,9 @@ internal final class FatalAppHangsHandler {
             }
             self?.send(fatalHang: fatalHang)
         }
+
+        // Remove pending app hang
+        featureScope.rumDataStore.removeValue(forKey: .fatalAppHangKey)
```
simaoseica-dd, line 77: question: What if the `send(fatalHang:)` inside of an escaping closure fails? Is it removed nevertheless?

Issue #2267, opened 2025-04-15
"Missing required module 'DatadogPrivate'" when using SDK in shared / common module


dibeakshay:
### Describe the bug

Hi Datadog team 👋

I’m integrating Datadog into an iOS project, and I’ve run into a build issue when the Datadog SDK is added to a shared module (let’s call it Core) that’s used by both the main app and an app extension.

When compiling the extension, the build fails with:

Missing required module 'DatadogPrivate'

This happens even though I’m guarding the import with Swift compiler flags:

#if !EXTENSION
import Datadog
#endif

I’ve confirmed that the EXTENSION flag is properly set in Other Swift Flags for the extension target, but the linker still seems to expect the DatadogPrivate module during compilation.

Questions:

Is there a recommended way to structure an app that uses Datadog in shared modules but excludes it from app extensions?

Could this be caused by DatadogPrivate being a dependency that leaks beyond the public Datadog import when linked, even if the code path avoids it?

Thanks for your help! Let me know if I can provide a sample project to reproduce.

### Reproduction steps

Create an iOS project with:

- One shared Swift module (Core) that imports Datadog.
- One main app target and one app extension target (e.g., IntentExtension).
- Add Core as a dependency to both the app and the extension.

In Core, guard the Datadog import using:

#if !EXTENSION
import Datadog
#endif

In the extension target’s Build Settings → Other Swift Flags, define EXTENSION:

-DEXTENSION
Build the extension target.

Expected Result:
The extension compiles successfully, since the Datadog code should be excluded via the #if !EXTENSION preprocessor check.

Actual Result:
Build fails with the following error:

Missing required module 'DatadogPrivate'

### Affected SDK versions

2.25.0

### Latest working SDK version

NA

### Did you confirm if the latest SDK version fixes the bug?

Yes

### Integration Methods

SPM

### Xcode Version

Xcode 16.3

ncreated:
Hey @dibeakshay 👋. 

You might want to take a look at [this comment](https://github.com/DataDog/dd-sdk-ios/issues/1402#issuecomment-1680370942), as it addresses a similar use case — using the SDK in a separate target.

If you're integrating the SDK within a separate module, it's important to use the appropriate access-level imports (I assume you want to hide the implementation details of your common module from its consumers). Depending on the Swift version you're using, the syntax will vary, particularly in light of [SE-0409](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0409-access-level-on-imports.md):

[...]

PR #2268, opened 2025-04-16
GH-2277 Fix access level for `DatadogPrivate` imports

ncreated:
### What and why?

📦 This PR updates the access-level used when importing the `DatadogPrivate` module in SPM builds.

It addresses [#2267](https://github.com/DataDog/dd-sdk-ios/issues/2267), where using a direct import DatadogPrivate in DatadogCore could lead to the error:

> "Missing required module 'DatadogPrivate'"

This may occur if `DatadogCore` is imported into a custom module within a client application (rather than the app target).

### How?

The fix introduces Swift version–aware access-level imports, following [SE-0409](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0409-access-level-on-imports.md). Specifically:

```swift
#if swift(>=6.0)
internal import DatadogPrivate
#else
@_implementationOnly import DatadogPrivate
#endif
```

This ensures that `DatadogPrivate` remains hidden from SDK consumers, preventing accidental exposure of internal APIs and resolving module visibility issues across various SPM use cases.

### Review checklist

### DIFF
--- a/DatadogCore/Sources/Core/Context/LaunchTimePublisher.swift
+++ b/DatadogCore/Sources/Core/Context/LaunchTimePublisher.swift
@@ -7,9 +7,15 @@
 import Foundation
 import DatadogInternal
 
+// swiftlint:disable duplicate_imports
 #if SPM_BUILD
-import DatadogPrivate
+    #if swift(>=6.0)
+    internal import DatadogPrivate
+    #else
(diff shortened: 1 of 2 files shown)


simaoseica-dd (approved):
🎖️

maxep (approved):
Thanks!

PR #2265, opened 2025-04-11
Simaoseica/rum 7880/update test utilities and integrations target

simaoseica-dd:
### What and why?

This PR refactors the testing utilities all across the SDK to scale Testing for future developments.
The main advantages about this foundational restructuring and cleanup are:
- Centralize Mocks, Proxies and whatnot in a common place. 
- Refactor Mocks to be reused in different testing frameworks (`XCTest`, [`SwiftTesting`](https://developer.apple.com/xcode/swift-testing/))

### How?

**1. Create new Integration Tests targets**
There are two new targets created to focus on integration tests: `DatadogIntegrationTests iOS`, `DatadogIntegrationTests tvOS`

**2. Centralize testing utilities in the package `TestUtilities`**
The `TestUtilities` is the source of truth for all components related to testing.
TestUtilities

**3. Mocks are clean of `import XCTest`**
To have dynamism selecting the testing framework, the `Mocks` should be clean of associations with testing frameworks. This way, it will be possible to use modern testing techniques like the ones used on [`SwiftTesting`](https://developer.apple.com/xcode/swift-testing/).

**4. Mocks are contextualized by feature**
Given the large number of mocks, they are organized by feature to improve clarity and maintainability.
Mocks

 

### Review checklist
- [ ] Feature or bugfix MUST have appropriate tests (unit, integration)
- [ ] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes
- [ ] Add Objective-C interface for public APIs (see our [guidelines](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3157787243/RFC+-+Modular+Objective-C+Interface#Recommended-solution) (internal) and run `make api-surface`)

PR #2272, opened 2025-04-24
RUM-9521 Move RUM models to DatadogInternal

maxep:
### What and why?

Models that are shared between modules must be shared in `DatadogInternal`. That is the case with RUM generated models.
This PR changes the generator destination for RUM schema.

### How?
- Change generator destination to `DatadogInternal` when generating RUM models.
- Update imports accordingly
- Re-export events used in mappers from RUM module.
- Move/Update mocks accordingly

### Review checklist

### DIFF
--- a/TestUtilities/Sources/Mocks/DatadogInternal/RUMDataModelMocks.swift
+++ b/TestUtilities/Sources/Mocks/DatadogInternal/RUMDataModelMocks.swift
@@ -6,7 +6,20 @@
 
 import Foundation
 
-@testable import DatadogRUM
+import DatadogInternal
+
+/// Creates random RUM event.
+public func randomRUMEvent() -> RUMDataModel {
+    // swiftlint:disable opening_brace
+    return oneOf([
+        { RUMViewEvent.mockRandom() },
+        { RUMActionEvent.mockRandom() },
+        { RUMResourceEvent.mockRandom() },
+        { RUMErrorEvent.mockRandom() },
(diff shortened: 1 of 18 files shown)


simaoseica-dd (dismissed):

TestUtilities/Sources/Mocks/DatadogInternal/RUMDataModelMocks.swift
```diff
@@ -162,7 +175,7 @@ extension RUMViewEvent: RandomMockable {
             service: .mockRandom(),
             session: .init(
                 hasReplay: hasReplay,
-                id: sessionID.toRUMDataFormat,
+                id: sessionID.uuidString.lowercased(),
```
mariedm, line 178: This might be due to resolving conflicts, but I believe @simaoseica-dd made these changes in #2265, so we should use `RUMUUID` and `toRUMDataFormat` now?

mariedm (commented):
I just left a small question, otherwise LGMT!

TestUtilities/Sources/Mocks/DatadogInternal/RUMDataModelMocks.swift
maxep: Yes indeed, but that involves a `@testable import DatadogRUM` which I'm not willing to do in `TestUtilities` module. We should avoid such import in a non-test target.

mariedm (approved):

PR #2274, opened 2025-04-28
RUM-9022 Session consistent trace sampling

xgouchet:
### What and why?

Uses the session id when available to compute the trace sampling on network traces.

### DIFF
--- a/DatadogInternal/Sources/NetworkInstrumentation/TracePropagationHeadersWriter.swift
+++ b/DatadogInternal/Sources/NetworkInstrumentation/TracePropagationHeadersWriter.swift
@@ -25,7 +25,14 @@ public enum TraceSamplingStrategy {
         case .headBased:
             return DeterministicSampler(shouldSample: traceContext.isKept, samplingRate: traceContext.sampleRate)
         case .custom(let sampleRate):
-            return DeterministicSampler(baseId: traceContext.traceID.idLo, samplingRate: sampleRate)
+            // for a UUID with value aaaaaaaa-bbbb-Mccc-Nddd-1234567890ab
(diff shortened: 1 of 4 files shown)


DatadogInternal/Sources/NetworkInstrumentation/TracePropagationHeadersWriter.swift
```diff
@@ -25,7 +25,18 @@ public enum TraceSamplingStrategy {
...
+                // we use as the base id the last part : 0x1234567890ab
+                if let lastSegment = sessionId.split(separator: "-").last {
+                    UInt64(lastSegment, radix: 16)
+                } else {
+                    Optional<UInt64>.none
+                }
+            } else {
+                Optional<UInt64>.none
+            }
+            return DeterministicSampler(baseId: sessionBasedId ?? traceContext.traceID.idLo, samplingRate: sampleRate)
```
maxep, line 39: nit/ Here a more idiomatic way:
```suggestion
            // for a UUID with value aaaaaaaa-bbbb-Mccc-Nddd-1234567890ab
            // we use as the base id the last part : 0x1234567890ab
            let baseId = traceContext.rumSessionId
                .flatMap { $0.split(separator: "-").last }
                .flatMap { UInt64($0, radix: 16) }
            ?? traceContext.traceID.idLo

            return DeterministicSampler(
                baseId: baseId,
                samplingRate: sampleRate
            )
```

xgouchet: Thanks, done :)

maxep (approved):
Thanks for the test coverage 👌

PR #2276, opened 2025-04-29
RUM-9522 Migrate RUM messages to payload format

maxep:
### What and why?

Send RUM shared type as `.payload` message on the bus.

### How?

- Make `.baggage` message deprecated
- Add `.payload` message type
- Replace `.baggage` with `.payload` message from RUM

### Review checklist
- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes
- [ ] Add Objective-C interface for public APIs (see our [guidelines](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3157787243/RFC+-+Modular+Objective-C+Interface#Recommended-solution) (internal) and run `make api-surface`)

PR #2278, opened 2025-04-30
RUM-9522 Migrate Trace messages to payload format

maxep:
### What and why?

Following #2276 

Migrate Trace messages on the message-bus to the new strongly-typed `.payload`

### How?

- Create a shared `LogMessage` type.
- Create a shared `SpanContext` type.

### Review checklist


DatadogInternal/Sources/Models/Trace/SpanContext.swift
```diff
@@ -0,0 +1,23 @@
...
+
+/// The Span context received from `DatadogCore`.
+public struct SpanContext: Codable {
+    public enum CodingKeys: String, CodingKey {
+        case traceID = "dd.trace_id"
+        case spanID = "dd.span_id"
+    }
+
+    public let traceID: String?
+    public let spanID: String?
```
simaoseica-dd, line 17: question: It is weird that this model has all parameters as optional (`spanContext` should be optional when all are). Since we are in the `SpanContext`, can the `spanID` be mandatory at least?

DatadogInternal/Sources/Models/Logs/LogMessage.swift
```diff
@@ -0,0 +1,78 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+public struct LogMessage {
+    /// Log levels ordered by their severity, with `.debug` being the least severe and
+    /// `.critical` being the most severe.
+    public enum Level: Int, Codable {
```
simaoseica-dd, line 12: It should not need to be Codable, right?

simaoseica-dd (dismissed):
Looks good. Just left some questions

DatadogInternal/Sources/Models/Logs/LogMessage.swift
maxep: Should not be indeed.

DatadogInternal/Sources/Models/Trace/SpanContext.swift
maxep: I'm only moving the definition here, I'm not changing the logic. That being said, we actually don't need this anymore, I'm removing it :+1

simaoseica-dd (approved):

PR #2286, opened 2025-05-06
RUM-9496: Migrate RUM, Logs and Core Obj-c interfaces to corresponding modules

simaoseica-dd:
### What and why?

This PR moves the `RUM`, `Logs` and `Core` Objective-C interfaces to their corresponding feature modules.
The main goal is to remove the `DatadogObjc` module, which breaks modularity by importing all features into a single module just to provide an Objective-C interface.

**Note:** This is the first part of 2 (1/2). The second part migrates `Traces` and removes the `DatadogObjc` module.

### How?

Every feature module is responsible by its own API, regardless of the language consuming it. 

- To achieve this, the Objective-C interfaces and models were moved to their correct context. Swift can consume these APIs that have an `objc_` prefix. Objective-C code still consume them in the same way.
- All objc files are next to their Swift counterparts.
- The `api-surface`s are updated to reflect this migration.

All these changes follow @maxep's [RFC internal](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3157787243/RFC+-+Modular+Objective-C+Interface) recommendations.

### Review checklist


DatadogInternal/Sources/NetworkInstrumentation/URLSession/DDURLSessionDelegate+objc.swift
```diff
@@ -5,8 +5,6 @@
  */
 
 import Foundation
-import DatadogCore
-import DatadogInternal
 
 @objc
 @available(*, deprecated, message: "Use `URLSessionInstrumentation.enable(with:)` instead.")
```
maxep, line 11: /question Are we planning to remove this deprecated delegate as part of v3?

DatadogInternal/Sources/NetworkInstrumentation/URLSession/DDURLSessionInstrumentation+objc.swift
```diff
@@ -5,13 +5,11 @@
...
 
 import Foundation
-import DatadogCore
-import DatadogInternal
 
 /// Configuration of URLSession instrumentation.
 @objc
 public class DDURLSessionInstrumentationConfiguration: NSObject {
-    internal var swiftConfig: URLSessionInstrumentation.Configuration
+    public var swiftConfig: URLSessionInstrumentation.Configuration
```
maxep: It should stay `internal` IMO.

DatadogInternal/Sources/Telemetry/CoreLogger.swift
```diff
@@ -111,3 +111,12 @@ extension CoreLogger {
...
 }
+
+@objc
+public enum DDSDKVerbosityLevel: Int {
+    case none
+    case debug
+    case warn
+    case error
+    case critical
+}
```
maxep: This should be part of `DatadogCore` IMO, and follow the new pattern with `objc_` and `@_spi`.

DatadogObjc/Sources/RUM/RUM+objc.swift
```diff
@@ -8,6 +8,7 @@ import Foundation
 import UIKit
 import DatadogInternal
 import DatadogRUM
+import DatadogTrace
```
maxep: Is this intended?

DatadogInternal/Sources/Telemetry/DDInternalLogger+objc.swift
```diff
@@ -5,8 +5,6 @@
  */
 
 import Foundation
-import DatadogInternal
-import DatadogCore
 
 @objc
 public class DDInternalLogger: NSObject {
```
maxep: Same here about `objc_` and `@_spi` convention.

DatadogCore/Sources/DatadogConfiguration.swift
```diff
@@ -0,0 +1,215 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+import DatadogInternal
+
+extension Datadog {
+    /// Configuration of Datadog SDK.
+    public struct Configuration {
```
maxep, line 12: /question Is this change necessary for objc interfaces? 🤔

maxep (changes requested):
Looks great 👍  The main blocker to me is the objc interfaces moved to `DatadogInternal`: we shouldn't expect user to `@import DatadogInternal;`. They should be exposed by Core IMO.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/DDURLSessionInstrumentation+objc.swift
simaoseica-dd: It is `public internal(set)`. The mutations will still be internal

DatadogInternal/Sources/NetworkInstrumentation/URLSession/DDURLSessionDelegate+objc.swift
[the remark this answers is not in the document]
simaoseica-dd: It is a good point. Moved it to `DatadogCore` while we don't have a package focused on `NetworkInstrumentation`

tools/rum-models-generator/run.py
```diff
@@ -23,7 +23,7 @@
 
 # Generated file paths (relative to repository root)
 RUM_SWIFT_GENERATED_FILE_PATH = '/DatadogInternal/Sources/Models/RUM/RUMDataModels.swift'
-RUM_OBJC_GENERATED_FILE_PATH = '/DatadogObjc/Sources/RUM/RUMDataModels+objc.swift'
+RUM_OBJC_GENERATED_FILE_PATH = '/DatadogInternal/Sources/Models/RUM/RUMDataModels+objc.swift'
```
mariedm, line 26: 👌

tools/rum-models-generator/Tests/CodeGenerationTests/Print/ObjcInteropPrinterTests.swift
mariedm, line 1: Once the migration is complete, since both Swift and ObjC interfaces now exist in the same modules, the `make api-surface` command will have to be updated accordingly to scan all modules at once and then separate the output based on language (api-surface-swift or api-surface-objc).

mariedm (approved):
Great work! 🚀

tools/rum-models-generator/Tests/CodeGenerationTests/Print/ObjcInteropPrinterTests.swift
simaoseica-dd: That's a good point. I'll take care of it in the second part of this task

mariedm (approved):

DatadogInternal/Sources/Models/RUM/RUMDataModels+objc.swift
maxep, line 1: /blocker These should be exposed from RUM, not Internal module. The Swift models are part of Internal because they are transmitted over the message-bus and RUM re-export them. The objc definitions do not need to be shared in the Internal module.

DatadogInternal/Sources/NetworkInstrumentation/TracingHeaderType+objc.swift
maxep, line 1: /question Can this one be exposed by Core module instead?

DatadogCore/Sources/NetworkInstrumentation/URLSessionInstrumentation+objc.swift
```diff
@@ -5,39 +5,40 @@
...
 
 /// Configuration of URLSession instrumentation.
-@objc
-public class DDURLSessionInstrumentationConfiguration: NSObject {
-    internal var swiftConfig: URLSessionInstrumentation.Configuration
+@objc(DDURLSessionInstrumentationConfiguration)
+@objcMembers
+@_spi(objc)
+public final class objc_URLSessionInstrumentationConfiguration: NSObject {
+    public internal(set) var swiftConfig: URLSessionInstrumentation.Configuration
```
maxep, line 16: /question Is there a reason for this to be `public`?

maxep (changes requested):
Thanks for the changes 👍 There are still objc interfaces in Internal module that could be moved IMO, especially the RUM models.

DatadogInternal/Sources/NetworkInstrumentation/TracingHeaderType+objc.swift
simaoseica-dd: Apparently not. It is used in [RUM](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogObjc/Sources/RUM/RUM%2Bobjc.swift), [Core](https://github.com/DataDog/dd-sdk-ios/pull/2286/files#diff-d7006addf428aacb11cc338d47edfba77e3f1c25d873b854f6d036a2e1b6db7bL40) and [Trace](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogObjc/Sources/Tracing/Trace%2Bobjc.swift) in the next PR.

maxep (approved):
🚀

PR #2289, opened 2025-05-07
RUM-9747 Add strongly-typed additional context to core

maxep:
### What and why?

`FeatureBaggage` comes with perf implications, we are replacing loosely-typed baggages in core context by strongly-typed additional context.

### How?

Similar to `BaggageSharing` (now deprecated), add `AdditionalContextSharing` for sharing additional context from core and feature-scope.

### Review checklist
- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes
- [ ] Add Objective-C interface for public APIs (see our [guidelines](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3157787243/RFC+-+Modular+Objective-C+Interface#Recommended-solution) (internal) and run `make api-surface`)

PR #2290, opened 2025-05-07
RUM-9747 Strongly-typed RUM Additional Context

maxep:
### What and why?

Following: #2289

`FeatureBaggage` comes with perf implications, we are replacing loosely-typed baggages in core context by strongly-typed additional context.

### How?

Replace RUM baggage by `RUMCoreContext` as additional context.

### Review checklist


DatadogInternal/Sources/Context/DatadogContext.swift
```diff
@@ -186,9 +186,15 @@ public protocol AdditionalContext {
...
     /// - Returns: The `Context` if found
     public func additionalContext<Context>(ofType type: Context.Type = Context.self) -> Context? where Context: AdditionalContext {
         additionalContext[type.key] as? Context
     }
+
+    /// Gets an additional context value.
+    /// - Parameter context: The additional context.
+    public mutating func set<Context>(additionalContext context: Context?) where Context: AdditionalContext {
+        additionalContext[Context.key] = context
+    }
```
ncreated, line 199: api/ The current API comments are too vague and don't provide any meaningful information beyond what the method names already imply. In particular, the comment for `set(additionalContext:)` appears to be incorrect—likely copy-pasted from elsewhere.

Since API comments are currently our primary form of architectural documentation (unfortunately), they warrant careful attention. As a reader, I’d expect the comments to explain key concepts—like what "additional context" actually represents, how to add / update or remove one. Without that, these APIs come across as opaque and disconnected.

To avoid redundancy, this explanation can be centralized in the documentation for the AdditionalContext type itself, and [symbol links](https://www.swift.org/documentation/docc/linking-to-symbols-and-other-content) can be used to reference it from related APIs.

DatadogInternal/Sources/Context/DatadogContext.swift
```diff
@@ -186,9 +186,15 @@ public protocol AdditionalContext {
 extension DatadogContext {
     /// Gets an additional context value.
     ///
-    /// - Parameter type: The value type.
+    /// - Parameter type: The additional context type.
     /// - Returns: The `Context` if found
     public func additionalContext<Context>(ofType type: Context.Type = Context.self) -> Context? where Context: AdditionalContext {
```
ncreated, line 191: nit/ I believe we could win if removing the default `= Context.self`. With the default, callers can write:
```swift
let rum: RUMCoreContext = context.additionalContext()
```
which raises questions like _"Is additionalContext() already returning RUMCoreContext?"_ and _"Why repeat the type annotation?"_.

Making the type explicit keeps intent clear and self-documenting:
```swift
let rum = context.additionalContext(ofType: RUMCoreContext.self)
```

DatadogInternal/Sources/Context/DatadogContext.swift
```diff
@@ -186,9 +186,15 @@ public protocol AdditionalContext {
...
         additionalContext[type.key] as? Context
     }
+
+    /// Gets an additional context value.
+    /// - Parameter context: The additional context.
+    public mutating func set<Context>(additionalContext context: Context?) where Context: AdditionalContext {
```
ncreated, line 197: question/ How can `additionalContext` be removed after it was previously set? 🤔 💭

ncreated (dismissed):
Overall, it looks good 👍 I left a few questions about the internal API and challenged a couple of design choices 🙂.

DatadogInternal/Sources/Context/DatadogContext.swift
maxep: Yes indeed, I will add more details there 👍

DatadogInternal/Sources/Context/DatadogContext.swift
maxep: By using this [method](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogInternal/Sources/DatadogCoreProtocol.swift#L198), I can also add a method to the context itself 👍

DatadogInternal/Sources/Context/DatadogContext.swift
maxep: Done 🙂

DatadogInternal/Sources/Context/DatadogContext.swift
maxep: I have added more doc with link to core's method to mutate the context, LMKWYT!

ncreated (approved):
🎯

PR #2291, opened 2025-05-08
RUM-9066: Fix global attribute propagation in views

simaoseica-dd:
### What and why?

This PR fixes the propagation of global attributes across dependent scopes.
It is part of a broader effort to ensure consistent attribute management across all SDKs (iOS, Android, Browser, etc.).

### How?
This is a task to handle the management of `View` attributes discussed in the  [RFC - internal](https://datadoghq.atlassian.net/wiki/x/QoFRGwE).

To ensure correct propagation of global attributes, they are stored in a single location within the environment (e.g. `Monitor`). Each time an event occurs, the global attributes are propagated via `RUMCommands` to create the event. This approach guarantees that global attributes remain separate from local ones, and that the environment is accurately captured at the moment the event occurs.
For example, this solution ensures that each `View` update has the correct global attributes, even if some attributes were removed earlier in the `View` lifecycle.

**Note:**
This is the first part of 2 (1/2) focused on global attributes. The second part will focus on the View attributes management and will have a proper test coverage to cover all use cases.
 
### Review checklist

### DIFF
--- a/Datadog/IntegrationUnitTests/RUM/RUMAttributesIntegrationTests.swift
+++ b/Datadog/IntegrationUnitTests/RUM/RUMAttributesIntegrationTests.swift
@@ -46,9 +46,8 @@ final class RUMAttributesIntegrationTests: XCTestCase {
             .takeSingle()
 
         let firstView = try XCTUnwrap(session.views.first(where: { $0.name == firstViewName }))
-        firstView.viewEvents.forEach { viewEvent in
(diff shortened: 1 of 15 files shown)


maxep (approved):
Looks great!

fuzzybinary:
This looks good. I didn't see any tests that check for Global -> View -> Event overwrite precedence. Did I just miss them?

fuzzybinary (approved):
Nevermind, the new view attribute behavior looks like it's part of another ticket. Approved.

PR #2301, opened 2025-05-19
RUM-7543 Allocate rwlock on the heap

maxep:
### What and why?

Given this exemple:
```swift
var rwlock = pthread_rwlock_t()
pthread_rwlock_rdlock(&rwlock)
```

`rwlock` is not a real pointer that is long-lasting valid. Instead the lock pointer should be allocated on the heap.

### Review checklist

### DIFF
--- a/DatadogInternal/Sources/Concurrency/ReadWriteLock.swift
+++ b/DatadogInternal/Sources/Concurrency/ReadWriteLock.swift
@@ -18,15 +18,20 @@ public final class ReadWriteLock<Value>: @unchecked Sendable {
     private var value: Value
 
     /// The lock object.
-    private var rwlock = pthread_rwlock_t()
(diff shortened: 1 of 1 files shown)


mariedm (approved):
Nice find! Thanks for this fix. 🛠️ I wonder how often this lead to unexpected crashes given that we use this lock in many parts of the SDK.

maxep:
@mariedm I don't expect it to be much of a problem in its current state since the lock is wrapped within a class, but still it was at risk.

simaoseica-dd (commented):
Can you add more tests to the `ReadWriteLockTests`? For instance, tests that trigger the ReadWriteLock `deinit` concurrently?

maxep:
@simaoseica-dd, the `ReadWriteLock` is now 100% covered 👍

simaoseica-dd (approved):
LGTM

PR #2302, opened 2025-05-19
RUM-4236 Increase RUM batch ttl

maxep:
### What and why?

To reduce data loss, we want to extend the “max age” of uploaded RUM batches to 24h.

### How?

Provide Performance Override in RUM to increase the max file age for reading.

### Review checklist


simaoseica-dd (approved):
cool!

PR #2308, opened 2025-05-22
Collect accessibility attributes and sent as custom attributes

barboraplasovska:
### What and why?

This PR is part of the RUM Data Enrichment project, specifically focused on collecting accessibility attributes on iOS.
Since the RUM event format cannot be modified at this stage, these attributes are temporarily sent as *custom attributes*.
The goal is to start **collecting real data** (via the Datadog mobile app) to inform and refine the development of an accessibility dashboard.
This change is gated behind a **feature flag**, so it *does not affect* customer-facing behavior.

### How?

I created an `AccessibilityReader` class that is initialized in `RUMFeature` when the `collectAccessibilitySettings` feature flag is enabled. This reader is then injected into `RUMScopeDependencies` so that it is accessible across all RUM components. Most of the accessibility attributes have listeners attached, so the internal state of the reader is kept up to date whenever one of these settings changes. The attributes are then added as custom attributes right before a `RUMViewEvent` is sent.

### Review checklist


DatadogRUM/Sources/RUMContext/AccessibilityReader.swift
```diff
@@ -0,0 +1,273 @@
...
+        startObserving()
+        updateState()
+    }
+
+    deinit {
+        stopObserving()
+    }
+
+    func updateState() {
+        DispatchQueue.main.async { [weak self] in
```
ncreated, line 32: suggestion/ To better encapsulate this component, we can make this method `private`. Other components shouldn't need to manually trigger a state update because in real usage this is managed by `AccessibilityReader` internally.

To what I see, the reason for this being `internal` might be related to testing. I've left a separate comment with a suggestion on how we can improve this while keeping the method private.

DatadogRUM/Tests/RUMContext/AccessibilityReaderTests.swift
```diff
@@ -0,0 +1,192 @@
...
+        mock.isGrayscaleEnabled = true
+        mock.isGuidedAccessEnabled = true
+        mock.isOnOffSwitchLabelsEnabled = true
+        mock.isSpeakScreenEnabled = true
+        mock.isSpeakSelectionEnabled = true
+        mock.rtlEnabled = true
+
+        // When
+        let reader = AccessibilityReader(notificationCenter: .init(), accessibilityValues: mock)
+        reader.updateState()
```
ncreated, line 160: question/ Do we need to call it? The `updateState()` is already invoked from `AccessibilityReader.init()`.

DatadogRUM/Tests/RUMMonitor/Scopes/RUMViewScopeTests.swift
```diff
@@ -3390,4 +3390,111 @@ class RUMViewScopeTests: XCTestCase {
...
+        )
+
+        let readerInitExpectation = XCTestExpectation(description: "AccessibilityReader initialized")
+        DispatchQueue.main.async {
+            accessibilityReader?.updateState()
+            DispatchQueue.main.async {
+                readerInitExpectation.fulfill()
+            }
+        }
+        wait(for: [readerInitExpectation], timeout: 2.0)
```
ncreated, line 3435: suggestion/ This is a bit hard to follow without knowing the internals of `AccessibilityReader`. We can likely simplify the model and improve clarity by making queue a dependency of `AccessibilityReader`:
```swift
// In AccessibilityReader.swift

init(
   notificationCenter: NotificationCenter = .default, 
   updateQueue: DispatchQueue = .main,
   accessibilityValues: AccessibilityValues = LiveAccessibilityValues()
)
```
Then in tests:
```swift
let updateQueue = DispatchQueue()
let reader = AccessibilityReader(updateQueue: updateQueue, accessibilityValues: initialMockValues)
```

This allows us to replace the verbose `XCTestExpectation` workaround with a simple:
```swift
updateQueue.sync {}
```
which blocks the caller thread until all pending updates are processed.

DatadogRUM/Sources/RUMContext/AccessibilityReader.swift
```diff
@@ -0,0 +1,273 @@
...
+
+internal final class AccessibilityReader {
+    @ReadWriteLock
+    var state: Accessibility
+
+    private let notificationCenter: NotificationCenter
+    private let accessibilityValues: AccessibilityValues
+    private var observers: [NSObjectProtocol] = []
+
+    init(notificationCenter: NotificationCenter = .default, accessibilityValues: AccessibilityValues = LiveAccessibilityValues()) {
```
ncreated, line 19: convention/ We should avoid providing a default value for `NotificationCenter`. Instead, we should explicitly pass [`RUM.Configuration.notificationCenter`](https://github.com/DataDog/dd-sdk-ios/blob/6763ae261f8f756162dab484ec81d8a144ad9bde/DatadogRUM/Sources/RUMConfiguration.swift#L339-L340) from the caller. Since `RUM.Configuration` is already available in `RUMFeature`, this should be straightforward.

The reason is to ensure consistent dependency injection, especially for integration testing. Reliable integration tests require all internal components to use **the same** instance of `NotificationCenter`. That instance cannot be `.default` if we want full control over simulated notifications.

We recently did a large refactor to enforce this across the SDK in https://github.com/DataDog/dd-sdk-ios/pull/2094 🙂.

ncreated (changes requested):
Really well done 🏅💪, the code reads cleanly, is well-organized, and has a solid structure overall 🚀.

I’ve left a few pieces of feedback, along with one **blocking** question regarding the potential over-allocation of the `AccessibilityReader` instance.

simaoseica-dd (commented):
I left some comments. 
Well done overall 👍

DatadogRUM/Sources/Feature/RUMFeature.swift
```diff
@@ -115,6 +115,11 @@ internal final class RUMFeature: DatadogRemoteFeature {
                     telemetry: core.telemetry
                 )
             },
+            accessibilityReaderFactory: {
+                configuration.featureFlags[.collectAccessibilitySettings]
+                ? AccessibilityReader()
+                : nil
+            },
```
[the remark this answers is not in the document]
barboraplasovska: I completely agree, it is indeed unnecessary to create it for each view. I made the change !

DatadogRUM/Tests/RUMContext/AccessibilityReaderTests.swift
barboraplasovska: Nope, my mistake

DatadogRUM/Tests/RUMMonitor/Scopes/RUMViewScopeTests.swift
barboraplasovska: I've made this change! However since I am using the `updateQueue` inside my `updateState` function:

```Swift
 private func updateState() {
        updateQueue.async { [weak self] in
            guard let self else {
                return
            }
            state = currentState
        }
    }
    
```

I am getting this warning:

```Main actor-isolated property 'currentState' can not be referenced from a Sendable closure; this is an error in the Swift 6 language mode```

Is this something I should be concerned about?

Comment (barboraplasovska) [reply DatadogRUM/Tests/RUMMonitor/Scopes/RUMViewScopeTests.swift]: Following a discussion with @simaoseica-dd  (see his comment [here](https://github.com/DataDog/dd-sdk-ios/pull/2308#discussion_r2111421372)), I’ve made `AccessibilityReader` available only from iOS 13+. As a result, I’ve removed the `updateQueue` and now use only `@MainActor` and `Task`, without mixing in GCD.

File: DatadogRUM/Sources/RUMContext/AccessibilityReader.swift [hunk 6]
```diff
@@ -0,0 +1,273 @@
...
+        }
+        observers.append(speakSelectionObserver)
+    }
+
+    private func stopObserving() {
+        observers.forEach { notificationCenter.removeObserver($0) }
+        observers.removeAll()
+    }
+
+    @MainActor private var currentState: Accessibility {
```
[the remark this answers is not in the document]
barboraplasovska: You're right! Following our discussion, I restricted `AccessibilityReader` to iOS 13+

DatadogRUM/Tests/RUMMonitor/Scopes/RUMViewScopeTests.swift
```diff
@@ -3390,4 +3390,126 @@ class RUMViewScopeTests: XCTestCase {
...
+            interactionToNextViewMetric: INVMetricMock()
+        )
+
+        let readerInitExpectation = XCTestExpectation(description: "AccessibilityReader initialized")
+        DispatchQueue.main.async {
+            DispatchQueue.main.async {
+                readerInitExpectation.fulfill()
+            }
+        }
+        wait(for: [readerInitExpectation], timeout: 2.0)
```
ncreated: question/ Do we actually need this 🤔? Since we're using a mock `AccessibilityReader` that returns state synchronously, do we still need this extra asynchronous waiting? All we want to test is whether `RUMViewScope` reads the accessibility attributes correctly from the reader. While the default reader evaluates the state asynchronously, our mock returns it immediately.

ncreated (approved):

DatadogRUM/Tests/RUMMonitor/Scopes/RUMViewScopeTests.swift
barboraplasovska: You're right! I'll remove it!

ncreated (approved):

DatadogRUM/Sources/RUMContext/AccessibilityReader.swift
```diff
@@ -0,0 +1,265 @@
...
+        var state = Accessibility()
+
+        if let contentSize = UIApplication.dd.managedShared?.preferredContentSizeCategory.rawValue as String? {
+            state.textSize = contentSize
+        } else {
+            state.textSize = UIContentSizeCategory.unspecified.rawValue
+        }
+
+        state.videoAutoplayEnabled = !UIAccessibility.isVideoAutoplayEnabled
```
simaoseica-dd:
```suggestion
        state.videoAutoplayEnabled = UIAccessibility.isVideoAutoplayEnabled
```

DatadogRUM/Sources/RUMContext/AccessibilityReader.swift
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    private func startObserving() {
+        if #available(iOS 14.0, tvOS 14.0, *) {
+            let buttonShapesObserver = notificationCenter.addObserver(
+                forName: UIAccessibility.buttonShapesEnabledStatusDidChangeNotification,
+                object: nil,
+                queue: .main
+            ) { [weak self] _ in
+                self?.updateState()
```
simaoseica-dd, line 47: nit not blocker: It could update only the related value instead of the whole state (with only one set to the Reader `state`).

PR #2292, opened 2025-05-08
RUM-8417 [SwiftUI Auto view tracking] Add UI tests + Improve name extraction logic

mariedm:
### What and why?

This PR improves our SwiftUI view auto-instrumentation logic and adds UI tests.

### How?

**1. Improved SwiftUI View Detection Logic**

- Expanded and renamed reflection paths for better clarity and more reliable view name extraction
- Skipped TabBar controllers from auto-tracking as they interfere with the view stack in `RUMViewsHandler` (customers can still manually track tabs via view modifiers)
- Added filtering logic to ignore UIKit-bundled components
- Implemented a fallback naming strategy when a meaningful name cannot be extracted from reflection

**2. UI Tests**

- Added two test scenarios:
  - Single view at root
  - Tabbar at root with different navigation scenarios in each tab
- Renamed `RUMSwiftUIScenarioTests` to `RUMSwiftUIManualScenarioTests` for better clarity next to the new `RUMSwiftUIAutoScenarioTests`
- Added README documentation to facilitate future test updates

### Review checklist
- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes
- [ ] Add Objective-C interface for public APIs (see our [guidelines](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3157787243/RFC+-+Modular+Objective-C+Interface#Recommended-solution) (internal) and run `make api-surface`)

PR #2275, opened 2025-04-28
RUM-9035 Introduce SwiftUI Automatic `InstrumentationType`

mariedm:
### What and why?

We have three values in the `InstrumentationType` enum:
• `manual`: manual tracking via RUMMonitor
• `uikit`: UIKit predicate tracking (UIKitRUMViewsPredicate)
• `swiftui`: manual SwiftUI tracking (trackRUMView()/trackRUMTapAction() with ViewModifiers)

We recently added automatic SwiftUI tracking via `SwiftUIRUMViewsPredicate`. It was incorrectly labeled as `swiftui`, while it behaves like an automatic tracking based on a predicate, just as our existing UIKit predicate. We need to introduce a new instrumentation type when using the `SwiftUIRUMViewsPredicate`.

### How?

- Set instrumentation type to `swiftUIAutomatic` for views tracked with `SwiftUIRUMViewsPredicate`

### Review checklist
- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes
- [ ] Add Objective-C interface for public APIs (see our [guidelines](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3157787243/RFC+-+Modular+Objective-C+Interface#Recommended-solution) (internal) and run `make api-surface`)

PR #2237, opened 2025-03-20
mariedm/rum 8416 add auto view tracking option in rum config

mariedm:
- **RUM-8416 - Add view tracking option for SwiftUI**
- **RUM-8416 Add auto-tracking option for SwiftUI in RUM Configuration**

### What and why?

A short description of what changes this PR introduces and why.

### How?

A brief description of implementation details of this PR.

### Review checklist
- [ ] Feature or bugfix MUST have appropriate tests (unit, integration)
- [ ] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes
- [ ] Add Objective-C interface for public APIs (see our [guidelines](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3157787243/RFC+-+Modular+Objective-C+Interface#Recommended-solution) [internal]) and run `make api-surface`)

PR #2224, opened 2025-03-10
RUM-8415 Add SwiftUI view auto-instrumentation

mariedm:
### What and why?

A short description of what changes this PR introduces and why.

### How?

A brief description of implementation details of this PR.

### Review checklist
- [ ] Feature or bugfix MUST have appropriate tests (unit, integration)
- [ ] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes
- [ ] Add Objective-C interface for public APIs (see our [guidelines](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3157787243/RFC+-+Modular+Objective-C+Interface#Recommended-solution) [internal]) and run `make api-surface`)

PR #2315, opened 2025-05-27
RUM-8419 Add Automatic Action Tracking for SwiftUI

mariedm:
### What and why?

This PR enhances RUM's SwiftUI support by implementing action tracking, completing our SwiftUI auto-instrumentation initiative (following view tracking in #2224, #2237, #2275, #2292). This brings feature parity with our UIKit auto-instrumentation capabilities.

#### iOS 18 Support

- Leverages gesture information available in iOS 18 to identify SwiftUI `Button` and `NavigationLink` interactions thanks to the presence of `_ButtonGesture` in the touch description, which is only present during the `.began` phase of the `UITouch`
- Uses a two-phase detection approach:
  - Begin phase: captures button gesture information
  - End phase: creates RUM action events using stored gesture data
- Can differentiate between `Button` and `NavigationLink` components based on gestures' names

#### iOS 17 Support

On iOS 17, SwiftUI's internal implementation is different and does not expose the same button gesture information. We use a fallback strategy where precise component identification isn't possible. How it works:
- Filters out container views (`HostingView`, `HostingScrollView`, `PlatformGroupContainer`) since these are not interactive components 
- Reports other interactions as generic SwiftUI elements
- Can be disabled via `isiOS17DetectionEnabled` flag in `DefaultSwiftUIRUMActionsPredicate`

#### Additional support 

- Added support for additional SwiftUI-specific components, regardless of iOS version:
  - Toggle: Detected through `UISwitch` in view hierarchy
  - Menu: Identified via `ViewBasedUIButton` view type, mapped to `SwiftUI_Menu` in the default UIKit predicate
- tvOS support: continues using existing UIKit press event tracking as the interaction model is independent of UI framework

#### SwiftUI Component Detection Behavior

The following table summarizes the detection behavior for different SwiftUI components across iOS versions:

| SwiftUI Component | iOS 18+ Detection | iOS 17 and below Detection | Detection Strategy |
| :---- | :---- | :---- | :---- |
| **Button** | `SwiftUI_Button` | `SwiftUI_Unidentified_Element` | iOS 18+: Identified through `ButtonGesture` signature<br>iOS 17-: Falls back to generic detection |
| **NavigationLink** | `SwiftUI_NavigationLink` | `SwiftUI_Unidentified_Element` | iOS 18+: Identified through gesture pattern<br>iOS 17-: Falls back to generic detection |
| **Toggle** | `SwiftUI_Toggle` | `SwiftUI_Toggle` | Consistent detection via `UISwitch` in view hierarchy |
| **Label** | No action | `SwiftUI_Unidentified_Element` | iOS 17-: Cannot be distinguished from Button |
| **Menu** | `SwiftUI_Menu` | `SwiftUI_Menu` | Consistent detection via `ViewBasedUIButton` type |
| **HostingView** / Background View | No action | No action | iOS 17-: Explicitly filtered out as container view |

### Review checklist


DatadogRUM/Sources/Instrumentation/Actions/RUMActionsHandler.swift
```diff
@@ -8,53 +8,91 @@ import UIKit
...
-        guard let command = uiKitCommandsFactory?.command(from: event) else {
+        guard let command = eventCommandsFactory?.command(from: event) else {
             return // Not a "tap" event or doesn't have the view.
         }
 
         guard let subscriber = subscriber else {
             DD.logger.warn(
                 """
-                A RUM action was detected in UIKit, but RUM tracking appears to be disabled.
+                A RUM action was detected in UIKit or SwiftUI, but RUM tracking appears to be disabled.
```
maciejburda, line 95:
```suggestion
                A RUM action was detected, but RUM tracking appears to be disabled.
```

maciejburda:
```suggestion
                A RUM action was detected, but RUM tracking appears to be disabled.
```

DatadogRUM/Sources/Instrumentation/Actions/SwiftUI/SwiftUIComponentDetector.swift
```diff
@@ -0,0 +1,318 @@
...
+ *    - Creates RUM action if the touch completes successfully
+ *
+ * iOS 17 and below (Legacy):
+ * - Uses view hierarchy analysis for basic detection
+ * - Limited component identification:
+ * reports generic interactions for most components
+ * - May have some false positives: 
+ * a `Button` can't be differentiated from a `Label`, therefore we report both interactions
+ */
+
```
maciejburda:
```suggestion
```

DatadogRUM/Sources/Instrumentation/Actions/SwiftUI/SwiftUIComponentDetector.swift
maciejburda, line 1: Becomes a little big. Could be split to multiple files

DatadogRUM/Sources/Instrumentation/Actions/SwiftUI/SwiftUIRUMActionsPredicate.swift
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// MARK: SwiftUITouchRUMActionsPredicate
+extension DefaultSwiftUIRUMActionsPredicate: SwiftUIRUMActionsPredicate {
+    public func rumAction(with componentName: String) -> RUMAction? {
+        if #available(iOS 18.0, *) {
+            return RUMAction(name: componentName, attributes: [:])
+        } else if isiOS17DetectionEnabled {
+            return RUMAction(name: componentName, attributes: [:])
+        }
```
maciejburda:
```suggestion
        if #available(iOS 18.0, *) || isiOS17DetectionEnabled {
              return RUMAction(name: componentName, attributes: [:])
        }
```

Not 100% sure if it works though.

DatadogRUM/Sources/Instrumentation/Actions/UIKit/UIEventCommandFactory.swift
```diff
@@ -50,16 +44,40 @@ internal struct UITouchCommandFactory: UIEventCommandFactory {
         guard allTouches.count == 1, let tap = allTouches.first else {
             return nil // not a single touch event
         }
+
+        // Detect UIKit interactions first,
+        // as they are more likely to happen.
```
maciejburda, line 49: How does it behave when we nest SwiftUI view in the UIKit context?

Is UIKit predicate always taking a precedence over the SwiftUI one?

maciejburda (approved):
LGTM

Stating the obvious, but we'll need to soon check how it behaves on iOS 19 (or iOS 26 if rumors are true).

Left some minor comments.

DatadogRUM/Sources/Instrumentation/Actions/SwiftUI/SwiftUIRUMActionsPredicate.swift
```diff
@@ -0,0 +1,45 @@
...
+    /// - Returns: RUM Action if it should be recorded, `nil` otherwise.
+    func rumAction(with componentName: String) -> RUMAction?
+}
+
+/// Default implementations of `SwiftUIRUMActionsPredicate`
+public struct DefaultSwiftUIRUMActionsPredicate {
+    /// Whether to enable SwiftUI action detection on iOS 17 and below.
+    /// When set to `false`, actions will only be detected on iOS 18+ where the detection is more reliable.
+    /// Defaults to `true` for backward compatibility.
+    private let isiOS17DetectionEnabled: Bool
```
gonzalezreal: 🤏🏽 I know it's unlikely, but if we ever support other SwiftUI platforms (macOS, watchOS, etc.), the name of this property (and the initializer parameter name) does not hold up. My two-cents `isLegacyDetectionEnabled`.

gonzalezreal (approved):
LGTM

DatadogRUM/Sources/Instrumentation/Actions/SwiftUI/SwiftUIRUMActionsPredicate.swift
```diff
@@ -0,0 +1,45 @@
...
+    /// Creates a default SwiftUI RUM actions predicate.
+    /// - Parameter isiOS17DetectionEnabled: Whether to enable SwiftUI action detection on iOS 17 and below.
+    ///                                     Set to `false` to only use the more reliable iOS 18+ detection.
+    ///                                     Defaults to `true` for backward compatibility.
+    public init(isiOS17DetectionEnabled: Bool = true) {
+        self.isiOS17DetectionEnabled = isiOS17DetectionEnabled
+    }
+}
+
+// MARK: SwiftUITouchRUMActionsPredicate
```
gonzalezreal:
```suggestion
// MARK: SwiftUIRUMActionsPredicate
```

DatadogRUM/Sources/Instrumentation/Actions/UIKit/UIEventCommandFactory.swift
mariedm: I wouldn't say that it takes precedence, it's rather that we first check if the interaction happened on a UIKit element before checking if it happened on a SwiftUI element. And I chose this order as UIKit happens way more often, even in full SwiftUI apps, as many UI components rely on UIKit under the hood. Does it answer your question?

DatadogRUM/Sources/Instrumentation/Actions/SwiftUI/SwiftUIRUMActionsPredicate.swift
mariedm: Unfortunately it's not possible to combine `#available` with another condition :(

DatadogRUM/Sources/Instrumentation/Actions/SwiftUI/SwiftUIRUMActionsPredicate.swift
mariedm: Agreed, that makes sense. I'll rename it.

DatadogRUM/Sources/Instrumentation/Actions/RUMActionsHandler.swift
```diff
@@ -8,53 +8,91 @@ import UIKit
...
+    /// Note: On tvOS, user interactions come through the remote's physical buttons
+    /// as press events. These press events are processed at the system level
+    /// and delivered identically regardless of whether the UI is built with UIKit or SwiftUI.
+    /// Therefore, only one predicate is needed to handle actions from both frameworks.
+    convenience init(
+        dateProvider: DateProvider,
+        uiKitPredicate: UIPressRUMActionsPredicate?,
+        swiftUIPredicate: SwiftUIRUMActionsPredicate?,
+        swiftUIDetector: SwiftUIComponentDetector?
+    ) {
```
maxep: We can remove the `swiftUIPredicate` and `swiftUIDetector` as stated in the comment, right?

DatadogRUM/Sources/Instrumentation/Actions/SwiftUI/LegacySwiftUIComponentDetector.swift
```diff
@@ -0,0 +1,68 @@
...
+    func createActionCommand(
+        from touch: UITouch,
+        predicate: SwiftUIRUMActionsPredicate?,
+        dateProvider: DateProvider
+    ) -> RUMAddUserActionCommand? {
+        guard let predicate = predicate else {
+            return nil
+        }
+
+        if touch.phase == .ended {
```
maxep: nit/ we can combine these 2 in a single guard, for readability ;)

DatadogRUM/Sources/Instrumentation/Actions/SwiftUI/LegacySwiftUIComponentDetector.swift
```diff
@@ -0,0 +1,68 @@
...
+            if let view = touch.view,
+               view.isSwiftUIView,
+               view.isSafeForPrivacy,
+               // For iOS 17 and below, we can't reliably distinguish SwiftUI component types (e.g., Button vs Label).
+               // We exclude hosting views and track other SwiftUI elements with a generic name.
```
maxep: question/ Are we sure we want to do this? Sessions could be flooded with actions. 
Also, we can't infer `isSafeForPrivacy` from the container view which will be mostlikely of type `SwiftUI._GraphicView`. It will always be `true` I guess.

DatadogRUM/Sources/Instrumentation/Actions/SwiftUI/ModernSwiftUIComponentDetector.swift
```diff
@@ -0,0 +1,125 @@
...
+
+import Foundation
+import UIKit
+import DatadogInternal
+
+@available(iOS 18.0, tvOS 18.0, visionOS 18.0, *)
+internal final class ModernSwiftUIComponentDetector: SwiftUIComponentDetector {
+    /// Storage for pending touches that began but haven't ended yet
+    @ReadWriteLock
+    private var pendingSwiftUIActions = [ObjectIdentifier: PendingAction]()
```
maxep: question/ Are we accessing this property from multiple threads? I guess this logic will only be executed on the main thread, so we might not need a lock.

DatadogRUM/Sources/Instrumentation/Actions/SwiftUI/ModernSwiftUIComponentDetector.swift
```diff
@@ -0,0 +1,125 @@
...
+
+        return nil
+    }
+
+    /// Processes a touch in the `.began` phase,
+    /// which is when we can detect the Button gesture.
+    private func handleTouchBegan(_ touch: UITouch, dateProvider: DateProvider) -> Bool {
+        guard let view = touch.view,
+              view.isSwiftUIView,
+              view.isSafeForPrivacy else {
```
maxep: question/ same question here, is it necessary to check for privacy of an opaque SwiftUI view?

DatadogRUM/Sources/Instrumentation/Actions/SwiftUI/SwiftUIComponentDetector.swift
```diff
@@ -0,0 +1,140 @@
...
+    /// as we can collect different information.
+    static func extractComponentName(touch: UITouch, defaultName: String) -> String {
+        // Check gesture recognizers' names for more specific info
+        if let gestures = touch.gestureRecognizers {
+            for gesture in gestures {
+                if let gestureName = gesture.name {
+                    // E.g., Button<ResolvedButtonStyleBody<BorderlessButtonStyleBase>>
+                    if gestureName.hasPrefix("Button<") {
+                        return SwiftUIComponentNames.button
+                    }
```
maxep, line 86: question/ Why do we pick `Button` over `BorderlessButtonStyleBase`? the second seems more specific. Not sure it adds value, but it could help better understand which button was tapped.

DatadogRUM/Sources/Instrumentation/Actions/SwiftUI/SwiftUIComponentDetector.swift
```diff
@@ -0,0 +1,140 @@
...
+    static func extractSwiftUIToggleAction(
+        from touch: UITouch,
+        predicate: SwiftUIRUMActionsPredicate?,
+        dateProvider: DateProvider
+    ) -> RUMAddUserActionCommand? {
+        guard let predicate = predicate else {
+            return nil
+        }
+
+        if let grandSuperview = touch.view?.superview?.superview {
```
maxep: It seems very hacky to access the superview like this and I wonder if we need this 🤔  Isn't the `UISwitch` already covered by the UIKit action instrumentation?

DatadogRUM/Sources/Instrumentation/Actions/SwiftUI/SwiftUIComponentDetector.swift
mariedm: I agree that `BorderlessButtonStyleBase` would be more specific, but I chose to report the generic `SwiftUI_Button` instead of the specific style for several reasons:
 - UIKit action tracking reports generic class names like `UIButton` and I wanted to maintain some consistency across our different instrumentations
 - `BorderlessButtonStyleBase` is a styling detail, it doesn't provide a meaningful name about the component  that was interacted with
 - Button styles can be changed frequently during development, which could "break" the data reported 

Let me know what you think.

DatadogRUM/Sources/Instrumentation/Actions/SwiftUI/ModernSwiftUIComponentDetector.swift
mariedm: You're right, it should only be accessed from the main thread. It was more of a safety measure, but it's probably unnecessary. I'll remove it.

DatadogRUM/Sources/Instrumentation/Actions/RUMActionsHandler.swift
mariedm: We can't remove the SwiftUI parameters from the tvOS initializer because `RUMInstrumentation` calls the same 4-parameter initializer signature regardless of platform. 

I had kept the previous design pattern of having separate convenience initializers for each platform. This made sense when there were only UIKit predicates (which have different types on iOS vs tvOS), but now with SwiftUI parameters added, it's a bit awkward since tvOS ignores the SwiftUI parameters.

I could clean it up by having a single initializer, something like this:

```
convenience init(
    dateProvider: DateProvider,
    uiKitPredicate: UIKitRUMActionsPredicate?,
    swiftUIPredicate: SwiftUIRUMActionsPredicate?,
    swiftUIDetector: SwiftUIComponentDetector?
) {
    #if os(tvOS)
    // SwiftUI parameters ignored on tvOS
    guard let uiKitPredicate else {
        self.init(dateProvider: dateProvider, eventCommandsFactory: nil)
        return
    }
    
    self.init(
        dateProvider: dateProvider,
        eventCommandsFactory: UIPressCommandFactory(
            dateProvider: dateProvider,
            uiKitPredicate: uiKitPredicate
        )
    )
    #else
    // iOS logic...
    #endif
}
```

Let me know what you think.

DatadogRUM/Sources/Instrumentation/Actions/SwiftUI/ModernSwiftUIComponentDetector.swift
mariedm: Indeed, it's likely unnecessary, I added it more as a safety check, just in case. I'll remove it.

DatadogRUM/Sources/Instrumentation/Actions/SwiftUI/LegacySwiftUIComponentDetector.swift
mariedm: Yes, like I said in my reply to your other rcomment, the `isSafeForPrivacy` check is probably useless for SwiftUI views since it only looks for "Keyboard" in the class name, so it (always?) returns `true`. I can remove it here too.

About session flooding, your concern is legit. As said in the code comments, the legacy detection is much less precise since we can't distinguish interactive from non-interactive elements on iOS 17 and below. But that's exactly why I added the `isLegacyDetectionEnabled` parameter in `DefaultSwiftUIRUMActionsPredicate` to let customers decide whether they want to enable this detection or not. That being said, I'll remove the default value so that customers have to make an explicit decision.

DatadogRUM/Sources/Instrumentation/Actions/SwiftUI/SwiftUIComponentDetector.swift
mariedm: I agree it's a bit hacky. The reason why it's not detected by the current UIKit tracking approach is that the actual touch interaction happens on a simple `UIView`, not directly on the `UISwitch` itself. I see two options:
1. Keep the current approach and report the interaction as a `SwiftUI Toggle`
2. Update [bestActionTarget](https://github.com/DataDog/dd-sdk-ios/blob/053b4d6b978ae9dba658b8e44a9212d36d1f5a7b/DatadogRUM/Sources/Instrumentation/Actions/UIKit/UIEventCommandFactory.swift#L90) to also check `UISwitch` types in the view hierarchy and report it as a UIKit component 

Option 2 is cleaner as it deduplicates the similar logic in both UIKit and SwiftUI tracking strategies. But from a customer point of view, option 1 might make more sense.

DatadogRUM/Sources/Instrumentation/Actions/RUMActionsHandler.swift
maxep: I think we should be explicit with what's required for instrumenting each platform and avoid any confusion. If we have a different requirement for tvOS, then we can use `#if os(tvOS)` as you propose, but in the `RUMInstrumentation` to create the action handler. The convenient init for tvOS should also be exposed only when compiled for tvOS.

DatadogRUM/Sources/Instrumentation/Actions/SwiftUI/LegacySwiftUIComponentDetector.swift
maxep: Sounds good 👍

DatadogRUM/Sources/Instrumentation/Actions/SwiftUI/SwiftUIComponentDetector.swift
maxep: I see, It would make sense to add `UIControl` [here](https://github.com/DataDog/dd-sdk-ios/blob/053b4d6b978ae9dba658b8e44a9212d36d1f5a7b/DatadogRUM/Sources/Instrumentation/Actions/UIKit/UIEventCommandFactory.swift#L90) as well. But indeed the name is not representative of a `SwiftUI.Toggle`. I don't have a strong opinion, so I let you decide :) But maybe we could re-use the `findInParentHierarchy` method instead of doing `superview?.superview` which might break in the future.

DatadogRUM/Sources/Instrumentation/Actions/RUMActionsHandler.swift
mariedm: Agreed, I'll go with that!

DatadogRUM/Sources/Instrumentation/Actions/SwiftUI/LegacySwiftUIComponentDetector.swift
```diff
@@ -0,0 +1,68 @@
...
+        from touch: UITouch,
+        predicate: SwiftUIRUMActionsPredicate?,
+        dateProvider: DateProvider
+    ) -> RUMAddUserActionCommand? {
+        guard let predicate,
+              touch.phase == .ended else {
+            return nil
+        }
+
+        if touch.phase == .ended {
```
maxep: This if statement is now redundant ;)

DatadogRUM/Sources/Instrumentation/Actions/SwiftUI/SwiftUIComponentDetector.swift
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        #if !os(tvOS)
+        let uiSwitch = view.findInParentHierarchy { parent in
+            // Note: we might want to extend to `UIControl` in the future.
+            return parent is UISwitch
+        }
+
+        if uiSwitch != nil,
+           touch.phase == .ended,
```
maxep: This `touch.phase == .ended` should be the first statement of the guard above, to not loop through parent if it's ignored anyway.

maxep: That being said, I wonder why we check for the `touch.phase` in here. This method is only about extracting the name, the phase is already checked in the caller.

DatadogRUM/Sources/Instrumentation/Actions/SwiftUI/ModernSwiftUIComponentDetector.swift
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        cleanupStalePendingActions()
+
+        /// Handle `.began` phase:
+        /// store information for later use
+        if touch.phase == .began,
+           handleTouchBegan(touch, dateProvider: dateProvider) {
+            return nil
+        }
```
maxep:
```suggestion
        if touch.phase == .began,
            handleTouchBegan(touch, dateProvider: dateProvider) {
            return nil
        }
```

maxep (commented):
Thanks! I still have a couple of nits

PR #2322, opened 2025-06-05
RUM-9494: Align `os` and `device` attributes across all product features

simaoseica-dd:
### What and why?

This PR aligns the `os` and `device` details sent to the Datadog intake across `Logs`, `RUM` and `Trace` features.
It aims to have consistent models across features to improve discoverability and future additions of more information.

### How?

The main point to address consistency is to have central models that can be reused across the product features. 
This means that we have just one model definition of `Device` and `OperatingSystem`. They are defined on `DatadogInternal` and all redundant models are removed.
The models respect the [schema](https://github.com/DataDog/rum-events-format) defined for RUM models including being set in the root of the event as common parameters across events.
For more details on the properties and types, you can refer to the [Device schema](https://github.com/DataDog/rum-events-format/blob/master/schemas/rum/_common-schema.json#L286) and [OS schema](https://github.com/DataDog/rum-events-format/blob/master/schemas/rum/_common-schema.json#L259).

| RUM Event  | Log Event | Span Event |
| ----------- | ---------- | ----------- |
| rum | log | span |

**Note 1**: The Log events have `_dd` and `http` parameters with extra `device` and `os` information. The `_dd` has the device architecture to help symbolication and deobfuscation and the `http` device info is added by the backend after the intake.
**Note 2**: There is a second `DeviceInfo` model (not `Codable`) but it is used only for internal logic defined in the `DatadogContext`.

### Review checklist


mariedm (approved):

DatadogInternal/Sources/Context/DeviceInfo.swift
```diff
@@ -73,17 +48,49 @@ public struct DeviceInfo: Codable, Equatable {
...
-        self.osBuildNumber = osBuildNumber
+        self.type = .init(modelName: model, osName: osName)
         self.architecture = architecture
         self.isSimulator = isSimulator
         self.vendorId = vendorId
         self.isDebugging = isDebugging
         self.systemBootTime = systemBootTime
     }
+
+    public var rumDevice: Device {
```
maxep: suggestion/ Does the `rum*` properties needs to be defined in the Internal module? Can it be extensions on the RUM module instead?

DatadogInternal/Sources/Context/DeviceInfo.swift
```diff
@@ -73,17 +48,49 @@ public struct DeviceInfo: Codable, Equatable {
...
+    public enum DeviceType: Codable, Equatable {
+        case iPhone
+        case iPod
+        case iPad
+        case appleTV
+        case appleVision
+        case appleWatch
+        case other(modelName: String, osName: String)
+
+        public var rumDeviceType: Device.DeviceType {
```
maxep: same comment about `rum` property here.

DatadogInternal/Sources/Context/OperatingSystem+Utils.swift
```diff
@@ -0,0 +1,65 @@
...
+        version: String,
+        build: String?
+    ) {
+        self.name = name
+        self.version = version
+        self.versionMajor = version.split(separator: ".").first.map { String($0) } ?? version
+        self.build = build
+    }
+
+#if canImport(UIKit)
```
maxep: suggestion/ There is no UIKit import here, and the `_UIDevice` is already a typealias for platforms. I think this should be:
```swift
#if !os(macOS)

#else

#endif
```

DatadogInternal/Sources/Context/DeviceInfo.swift
```diff
@@ -73,17 +48,49 @@ public struct DeviceInfo: Codable, Equatable {
...
+extension DeviceInfo {
+    /// Represents the type of device.
+    public enum DeviceType: Codable, Equatable {
+        case iPhone
+        case iPod
+        case iPad
+        case appleTV
+        case appleVision
+        case appleWatch
+        case other(modelName: String, osName: String)
```
maxep: nit/
```suggestion
        case other(model: String, os: String)
```

DatadogLogs/Sources/Log/LogEventEncoder.swift
```diff
@@ -97,35 +97,15 @@ public struct LogEvent: Encodable {
...
-    }
-
     /// Datadog specific attributes.
     public struct Dd: Codable {
         /// Device information
-        public let device: DeviceInfo
+        public struct Device: Codable {
+            /// The CPU architecture of the device. Used to symbolication and deobfuscation.
+            public let architecture: String
+        }
```
maxep, line 106: So we are duplicating the architecture attribute here for symbolication. Have we consider placing all device info in `_dd.device`, and `_dd.os`?

maxep (commented):
Looks good! I left some minor comments.

DatadogLogs/Sources/Log/LogEventEncoder.swift
simaoseica-dd: Yes, it was considered but was not followed because:
- After discussion with @0xnm we can improve this logic ourselves in the backend and get rid of the `_dd` namespace without much trouble.
- The current organization, `os` and `device` being in the root of the event, is consistent across all feature products. It follows this premise by respecting the [schema](https://github.com/DataDog/rum-events-format/blob/master/schemas/rum/_common-schema.json#L259-L318).

DatadogInternal/Sources/Context/DeviceInfo.swift
simaoseica-dd: I think it is a question of naming since the `Device` and `OS` are now common models for all the feature products. 
I renamed it to `normalizedDevice` to better state this.

DatadogLogs/Sources/Log/LogEventEncoder.swift
maxep: OK 👍  All good if we plan to remove this duplicate.

DatadogInternal/Sources/Context/DeviceInfo.swift
maxep: I see the motivation, the thing is that we  are crossing domains here and we are hiding that fact: this model (same as `OperatingSystem`) is a RUM generated definition but this is no longer explicit. I think we should keep the `RUM*` prefix in the generator and create an extension for casting, such as:
```swift
extension RUMDevice: {
    init(_ deviceInfo: DeviceInfo)
}
```

This way, it will be explicit that we are applying the RUM format across domains. And it will be clear that this definition is coming from the generated models. Would that make sense?
It is a nit pick but I feel it will make thing easier to navigate.

simaoseica-dd: We want to have the feature products isolated from each other. In this case, we want to make sure that the RUM concepts are not leaked to the features using device and operating system models, like Tracing, Logs and Crash Reporting. 
And with this approach we can avoid redundant models at the same time we follow a schema that can be updated in the future.

PR #2324, opened 2025-06-05
RUM-9335 Fix missing app ID in Session Ended metric

mariedm:
### What and why?

This PR fixes the missing application ID in RUM Session Ended telemetry. This can happen when stopping a session manually with `RUMMonitor`.

### How?

For consistency, application ID follows the same override pattern as session ID, using `application_id_override` which gets transformed to `application.id` by the `TelemetryReceiver`.

### Review checklist

### DIFF
--- a/DatadogRUM/Sources/Integrations/TelemetryReceiver.swift
+++ b/DatadogRUM/Sources/Integrations/TelemetryReceiver.swift
@@ -259,6 +259,10 @@ internal final class TelemetryReceiver: FeatureMessageReceiver {
             let sessionIDOverride: String? = attributes.removeValue(forKey: SDKMetricFields.sessionIDOverrideKey)?.dd.decode()
             let sessionID = sessionIDOverride ?? rum?.sessionID
 
+            // Override applicationID using standard `SDKMetricFields`, otherwise use current RUM application ID:
+            let applicationIDOverride: String? = attributes.removeValue(forKey: SDKMetricFields.applicationIDOverrideKey)?.dd.decode()
+            let applicationID = applicationIDOverride ?? rum?.applicationID
+
             // Calculates the composition of sample rates. The metric can have up to 3 layers of sampling.
             var effectiveSampleRate = metric.sampleRate.composed(with: self.sampler.samplingRate)
             if let headSampleRate = attributes.removeValue(forKey: SDKMetricFields.headSampleRate) as? SampleRate {
@@ -268,7 +272,7 @@ internal final class TelemetryReceiver: FeatureMessageReceiver {
             let event = TelemetryDebugEvent(
                 dd: .init(),
                 action: rum?.userActionID.map { .init(id: $0) },
-                application: rum.map { .init(id: $0.applicationID) },
+                application: applicationID.map { .init(id: $0) },
                 date: date.addingTimeInterval(context.serverTimeOffset).timeIntervalSince1970.toInt64Milliseconds,
                 effectiveSampleRate: Double(effectiveSampleRate),
                 experimentalFeatures: nil,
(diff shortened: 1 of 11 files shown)


DatadogRUM/Sources/Integrations/TelemetryReceiver.swift
```diff
@@ -259,6 +259,10 @@ internal final class TelemetryReceiver: FeatureMessageReceiver {
             let sessionIDOverride: String? = attributes.removeValue(forKey: SDKMetricFields.sessionIDOverrideKey)?.dd.decode()
             let sessionID = sessionIDOverride ?? rum?.sessionID
 
+            // Override applicationID using standard `SDKMetricFields`, otherwise use current RUM application ID:
+            let applicationIDOverride: String? = attributes.removeValue(forKey: SDKMetricFields.applicationIDOverrideKey)?.dd.decode()
+            let applicationID = applicationIDOverride ?? rum?.applicationID
+
```
ncreated, line 265: food for thought/ While this approach works and reuses the override model originally introduced for `sessionIDOverrideKey`, I believe the issue with `application.id` is fundamentally different and could be addressed more cleanly - in a way that benefits the broader codebase.

The original purpose of `sessionIDOverrideKey` was to reflect `session.id` in "session ended" telemetry, even after a new session had started. In that case, the RUM context accessed like this (in line 255):
```swift
let rum = context.additionalContext(ofType: RUMCoreContext.self)
```
already reflects the new session’s ID, so we needed a mechanism to override it when reporting data for the previous session.

However, `application.id` doesn’t suffer from the same timing or state issue. It is static across the lifetime of the SDK. The reason it's missing in this context (line 255) isn’t because it's unavailable - it’s because the call to `RUMCoreContext.self` is returning `nil`. That origins [here in `Monitor.swift`](https://github.com/DataDog/dd-sdk-ios/blob/c6c152886994d9b80cd908b0c4cdfb0d3623e2b7/DatadogRUM/Sources/RUMMonitor/Monitor.swift#L152-L153), where we explicitly reset the context when no session is active.

So, using `applicationIDOverrideKey` masks the actual issue and applies a workaround rather than solving the root cause. A more scalable and clean solution might be to make `sessionID` optional in `RUMCoreContext`, allowing us to still propagate `applicationID` even when no session exists. This would likely benefit other scenarios where only partial context is available.

--

I'm not sure how much effort it would take to address this in a more robust way, but it might be worth exploring. That said, if it's non-trivial or time-sensitive, the proposed patch is a reasonable workaround and does solve the immediate problem 👍.

Also cc-ing @maxep and @simaoseica-dd for their opinions.

simaoseica-dd: > already reflects the new session’s ID, so we needed a mechanism to override it when reporting data for the previous session

Isn't it wrong to have a fallback then? `?? rum?.sessionID`

My question here is, what is a `TelemetryDebugEvent` without an `applicationID`? It should exist at this point, and incrementing the channels (`context` + `attributes`) to add that info will make it redundant without the need for it.

maxep: > Isn't it wrong to have a fallback then? ?? rum?.sessionID
@simaoseica-dd No because `SDKMetricFields.applicationIDOverrideKey` means no override, so we should use the current context. I don't understand your second point tho 🤔 

I agreed with @ncreated 's point on not resetting the entire RUM context when there is no active session, the application ID should still be propagated. Now, I'm not sure of about the effort either as it can have many side effects, but I think it would worth it.

mariedm: I have the same question as @simaoseica-dd, and I don't understand your answer @maxep. Why would we use `rum?.sessionID` as a fallback since we know it's the ID of the new session? (and since `session` can be `nil` in `TelemetryDebugEvent`).

mariedm: As for having the app ID propagated correctly, I agree it would be best. I'll try to have a look but this won't be my top priority for now.

mariedm: Following our discussion, I'm closing this PR. To sum up, this is not a priority and we prefer a real fix than a workaround.

PR #2321, opened 2025-06-03
RUM-9997 Adjust "ApplicationLaunch" view for iOS scene-based apps

ncreated:
### What and Why?

📦 This PR adjusts the timing of the "ApplicationLaunch" view for iOS apps that use `UISceneDelegate`.  It ensures the SDK correctly detects user-initiated launches and aligns the view start with the actual process start time.

### How?

The implementation follows the [RFC (internal)](https://datadoghq.atlassian.net/wiki/x/eQHZMAE).

It uses the [`task_role`](https://developer.apple.com/documentation/kernel/task_role_t) kernel API, alongside existing prewarming detection, to determine the **launch reason** at SDK initialization.

**Notable changes:**

- 🔄 `LaunchTime` was renamed to `LaunchInfo` to reflect its broader role in bundling all launch-related attributes.
- ✳️ Introduced a new `LaunchReason` enum to model the reason for process startup:
```swift
public enum LaunchReason {
    case userLaunch         // launched by user
    case backgroundLaunch   // background fetch, silent push, etc.
    case prewarming         // OS prewarming (iOS only)
    case uncertain          // fallback (e.g., tvOS, watchOS, or resolution failure)
}
```
- 🔍 `launchReason` is resolved using:
  - If `processInfo` contains `"ActivePrewarm"` flag → `.prewarming`
  - Else if `task_role == TASK_FOREGROUND_APPLICATION` → `.userLaunch`
  - Else → `.backgroundLaunch`

- 🧩 `task_role` is resolved in `__dd_private_AppLaunchHandler` (Objective-C) and passed to Swift as an `Int` to support iOS, tvOS, and watchOS consistently.

### Tests

🧪 The test suite introduced in [#2299](https://github.com/DataDog/dd-sdk-ios/pull/2299) was expanded with new `given()` branches to simulate `UISceneDelegate`-based launches. Existing `UIApplicationDelegate`-only scenarios are preserved to ensure full backward compatibility.

### Review checklist
- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes
- [ ] Add Objective-C interface for public APIs (see our [guidelines](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3157787243/RFC+-+Modular+Objective-C+Interface#Recommended-solution) (internal) and run `make api-surface`)

PR #2325, opened 2025-06-06
RUM-9997 Adjust "ApplicationLaunch" view for tvOS scene-based apps

ncreated:
### What and why?

📦 This PR adjusts the timing of the "ApplicationLaunch" view for tvOS apps using `UISceneDelegate`. It ensures that user-initiated launches are correctly detected and that the view start aligns with the actual process start time.

This is a counterpart to https://github.com/DataDog/dd-sdk-ios/pull/2321. Unlike iOS, tvOS does not support the `task_role` API, so we cannot determine the launch reason at SDK init. Instead, we apply a **launch window heuristic** that defers `LaunchReason` resolution until enough information can be inferred from app state and lifecycle events.

### How?

The implementation follows the [RFC (internal)](https://datadoghq.atlassian.net/wiki/x/eQHZMAE). It introduces two main components that work together to resolve the launch reason and defer processing of early RUM commands: `AppLaunchWindow` and `LaunchReasonResolver`.

#### 🆕 `AppLaunchWindow`

Buffers RUM commands along with their associated `DatadogContext` and `Writer` while the `LaunchReason` is `.uncertain`.

It resolves to:
- `.userLaunch` if the app starts in `INACTIVE` or `ACTIVE`, or
- `.userLaunch` if a `willEnterForeground` event is received within a 10s threshold, or
- `.backgroundLaunch` if the threshold is exceeded without a foreground transition.

Once resolved, all buffered items are returned via `flushBuffer()` in FIFO order.

#### 🆕 `LaunchReasonResolver`

Uses `AppLaunchWindow` internally to buffer and defer forwarding of commands. Once the `launchReason` is resolved, it:
- Updates the `launchReason` inside each buffered command’s context,
- Forwards all commands to RUM in original order.

Screenshot 2025-06-06 at 08 27 34

`LaunchReasonResolver` is integrated in `RUMApplicationScope`. The `process(command:context:writer:)` method now:
- Always routes through the resolver on **tvOS** and **watchOS**.
- Falls back to it on **iOS** only if `launchReason == .uncertain`, with a one-time debug telemetry log.

#### 🧪 Tests

Both `AppLaunchWindow` and `LaunchReasonResolver` are fully covered with unit tests.

A new integration suite, `RUMSessionTrackingTests`, was added to cover end-to-end session lifecycle scenarios, including:
- User launch + multiple views tracking in foreground + short background interruption.
- Background launch + events tracking remains in background.
- Background launch + app suspended + then moves to foreground where views tracking continues.

This test suite runs on both iOS and tvOS to ensure consistent session tracking behavior for both immediate (iOS) and deferred (tvOS) launch resolution.

#### 🧪 Tests – Alternative Considered

I initially considered reusing the test coverage from https://github.com/DataDog/dd-sdk-ios/pull/2299 to validate deferred `launchReason` resolution on tvOS. While the behaviors should be the same, iOS resolves immediately, whereas tvOS relies on timing - making the existing suite not directly reusable due to short timings used there (under tvOS launch window threshold).

Some scenarios (e.g., short background launches or prewarming) don’t apply to tvOS or fall outside the launch window, and adapting them would require event timing tweaks, more events to be simulated and doing extensive `#if os(tvOS)` branching.

Instead, I chose to:
- Keep the existing test suite focused on iOS,
- Add a new, platform-agnostic `RUMSessionTrackingTests` suite that validates both iOS and tvOS behavior cleanly.

### Review checklist

### DIFF
--- a/Datadog/Datadog.xcodeproj/project.pbxproj
+++ b/Datadog/Datadog.xcodeproj/project.pbxproj
@@ -555,11 +555,6 @@
 		6119DDCC2DD24A9600DA80F9 /* RUMSessionStopTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6119DDC42DD24A9600DA80F9 /* RUMSessionStopTests.swift */; };
 		6119DDCD2DD24A9600DA80F9 /* RUMSessionStartInForegroundTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6119DDC32DD24A9600DA80F9 /* RUMSessionStartInForegroundTests.swift */; };
 		6119DDCE2DD24A9600DA80F9 /* RUMSessionTestsBase.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6119DDC52DD24A9600DA80F9 /* RUMSessionTestsBase.swift */; };
-		6119DDCF2DD24A9600DA80F9 /* RUMSessionWithNoViewTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6119DDC72DD24A9600DA80F9 /* RUMSessionWithNoViewTests.swift */; };
-		6119DDD02DD24A9600DA80F9 /* RUMSessionTimeOutTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6119DDC62DD24A9600DA80F9 /* RUMSessionTimeOutTests.swift */; };
-		6119DDD12DD24A9600DA80F9 /* RUMSessionStartInBackgroundTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6119DDC22DD24A9600DA80F9 /* RUMSessionStartInBackgroundTests.swift */; };
-		6119DDD22DD24A9600DA80F9 /* RUMSessionStopTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6119DDC42DD24A9600DA80F9 /* RUMSessionStopTests.swift */; };
-		6119DDD32DD24A9600DA80F9 /* RUMSessionStartInForegroundTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6119DDC32DD24A9600DA80F9 /* RUMSessionStartInForegroundTests.swift */; };
 		6119DDDA2DD24AB700DA80F9 /* AppRun.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6119DDD42DD24AB700DA80F9 /* AppRun.swift */; };
 		6119DDDB2DD24AB700DA80F9 /* AppRunStep+Fixtures.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6119DDD62DD24AB700DA80F9 /* AppRunStep+Fixtures.swift */; };
 		6119DDDC2DD24AB700DA80F9 /* AppRunner.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6119DDD52DD24AB700DA80F9 /* AppRunner.swift */; };
@@ -638,6 +633,9 @@
 		615192CE2BD6948B0005A782 /* HTTPHeadersWriterTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 615192CC2BD6948B0005A782 /* HTTPHeadersWriterTests.swift */; };
 		615192D02BD6B7C90005A782 /* DatadogTracer+InjectAndExtract.swift in Sources */ = {isa = PBXBuildFile; fileRef = 615192CF2BD6B7C90005A782 /* DatadogTracer+InjectAndExtract.swift */; };
 		615192D12BD6B7C90005A782 /* DatadogTracer+InjectAndExtract.swift in Sources */ = {isa = PBXBuildFile; fileRef = 615192CF2BD6B7C90005A782 /* DatadogTracer+InjectAndExtract.swift */; };
+		61536ABE2DF0A0D100B06D7D /* RUMSessionTrackingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61536ABD2DF0A0D100B06D7D /* RUMSessionTrackingTests.swift */; };
+		61536ABF2DF0A0D100B06D7D /* RUMSessionTrackingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61536ABD2DF0A0D100B06D7D /* RUMSessionTrackingTests.swift */; };
+		61536AC02DF21A0500B06D7D /* LaunchReasonResolverTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 619F1A262DEF0B3F003954BD /* LaunchReasonResolverTests.swift */; };
 		6156A9072BF75A7C00DF66C3 /* ImmutableRequestTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6156A9062BF75A7C00DF66C3 /* ImmutableRequestTests.swift */; };
 		6156A9082BF75A7C00DF66C3 /* ImmutableRequestTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6156A9062BF75A7C00DF66C3 /* ImmutableRequestTests.swift */; };
 		61570005246AADFA00E96950 /* DatadogObjc.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 61133BF0242397DA00786299 /* DatadogObjc.framework */; };
@@ -749,6 +747,9 @@
 		619CE75F2A458CE1005588CB /* TraceConfigurationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 619CE75D2A458CE1005588CB /* TraceConfigurationTests.swift */; };
 		619CE7612A458D66005588CB /* TraceTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 619CE7602A458D66005588CB /* TraceTests.swift */; };
 		619CE7622A458D66005588CB /* TraceTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 619CE7602A458D66005588CB /* TraceTests.swift */; };
+		619F1A242DEE493F003954BD /* LaunchReasonResolver.swift in Sources */ = {isa = PBXBuildFile; fileRef = 619F1A232DEE491D003954BD /* LaunchReasonResolver.swift */; };
+		619F1A252DEE493F003954BD /* LaunchReasonResolver.swift in Sources */ = {isa = PBXBuildFile; fileRef = 619F1A232DEE491D003954BD /* LaunchReasonResolver.swift */; };
+		619F1A282DEF0B3F003954BD /* LaunchReasonResolverTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 619F1A262DEF0B3F003954BD /* LaunchReasonResolverTests.swift */; };
 		61A2CC212A443D330000FF25 /* DDRUMConfigurationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61A2CC202A443D330000FF25 /* DDRUMConfigurationTests.swift */; };
 		61A2CC222A443D330000FF25 /* DDRUMConfigurationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61A2CC202A443D330000FF25 /* DDRUMConfigurationTests.swift */; };
 		61A2CC242A44454D0000FF25 /* DDRUMTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61A2CC232A44454D0000FF25 /* DDRUMTests.swift */; };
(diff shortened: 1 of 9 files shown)


Datadog/Datadog.xcodeproj/project.pbxproj
```diff
@@ -555,11 +555,6 @@
...
 		6119DDCE2DD24A9600DA80F9 /* RUMSessionTestsBase.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6119DDC52DD24A9600DA80F9 /* RUMSessionTestsBase.swift */; };
-		6119DDCF2DD24A9600DA80F9 /* RUMSessionWithNoViewTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6119DDC72DD24A9600DA80F9 /* RUMSessionWithNoViewTests.swift */; };
-		6119DDD02DD24A9600DA80F9 /* RUMSessionTimeOutTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6119DDC62DD24A9600DA80F9 /* RUMSessionTimeOutTests.swift */; };
-		6119DDD12DD24A9600DA80F9 /* RUMSessionStartInBackgroundTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6119DDC22DD24A9600DA80F9 /* RUMSessionStartInBackgroundTests.swift */; };
-		6119DDD22DD24A9600DA80F9 /* RUMSessionStopTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6119DDC42DD24A9600DA80F9 /* RUMSessionStopTests.swift */; };
-		6119DDD32DD24A9600DA80F9 /* RUMSessionStartInForegroundTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6119DDC32DD24A9600DA80F9 /* RUMSessionStartInForegroundTests.swift */; };
```
ncreated, line 562: 💡 Removing this coverage from tvOS, while keeping it for iOS. See PR description rationale.

DatadogRUM/Sources/RUMMonitor/Scopes/RUMApplicationScope.swift
```diff
@@ -56,7 +62,47 @@ internal class RUMApplicationScope: RUMScope, RUMContextProvider {
...
+    /// On tvOS and watchOS, we always use the `LaunchReasonResolver`, since those platforms do not support
+    /// `task_role` or prewarming signals.
+    ///
+    /// - Returns: `true` to indicate that the Application Scope should remain active.
     func process(command: RUMCommand, context: DatadogContext, writer: Writer) -> Bool {
+        #if !os(tvOS) && !os(watchOS)
+        guard context.launchInfo.launchReason != .uncertain else {
+            if !didLogFallbackToResolver {
+                dependencies.telemetry.debug("Falling back unexpectedly to 'launchReasonResolver' due to 'uncertain' launch reason")
+                DD.logger.critical("Falling back unexpectedly to 'launchReasonResolver' due to 'uncertain' launch reason")
```
maxep: question/ This will be seen in user's console, how will this be actionable for them? I wonder if we need something more descriptive, or if we need a console print at all? 🤔

DatadogRUM/Sources/RUMMonitor/Scopes/Utils/LaunchReasonResolver.swift
```diff
@@ -0,0 +1,120 @@
...
+
+    /// Buffers the given command and forwards it once the launch reason is resolved.
+    /// If resolution occurs as a result of this command, all previously buffered commands are forwarded in order,
+    /// with their `DatadogContext.launchInfo.launchReason` updated to match the resolved reason.
+    /// If the context already contains a known `launchReason`, the command is forwarded immediately without modification.
+    func forwardWithLaunchReason(
+        command: RUMCommand,
+        context: DatadogContext,
+        writer: Writer,
+        forwardingBlock: ([(command: RUMCommand, context: DatadogContext, writer: Writer)]) -> Void
```
maxep: suggestion/ It would be more straightforward if the `forwardingBlock` forward a single cmd. A for-each is only required when the buffer is flushed.

DatadogRUM/Sources/RUMMonitor/Scopes/Utils/LaunchReasonResolver.swift
```diff
@@ -0,0 +1,120 @@
...
+    ) {
+        guard context.launchInfo.launchReason == .uncertain else {
+            // Unexpected pre-set reason, forward as-is.
```
maxep: question/ That's not the unexpected state, is it?

DatadogRUM/Sources/RUMMonitor/Scopes/Utils/LaunchReasonResolver.swift
```diff
@@ -0,0 +1,120 @@
...
+            // Buffer command and check if this one resolved the launch reason.
+            window.buffer(command: command, context: context, writer: writer)
+
+            if window.resolvedReason != .uncertain {
+                let resolvedCommands = window
+                    .flushBuffer()
+                    .map { oldCommand, oldContext, oldWriter in (oldCommand, oldContext.replacing(launchReason: window.resolvedReason), oldWriter) }
+                forwardingBlock(resolvedCommands)
+            }
+        }
```
maxep: suggestion/ These statements are difficult to follow,  I think it's because the responsibilities between the resolver and the window are unbalanced. I would suggest to merge the 2 definitions as they are strongly coupled, or resolve the launch reason in the resolver, and keep the window for buffering and track its ttl. Would that make sense?

maxep (commented):
Looks good, it reads well and is very well covered in test 👍  I left a coupld of question/suggestion.

DatadogRUM/Sources/RUMMonitor/Scopes/Utils/LaunchReasonResolver.swift
```diff
@@ -0,0 +1,120 @@
...
+        writer: Writer,
+        forwardingBlock: ([(command: RUMCommand, context: DatadogContext, writer: Writer)]) -> Void
+    ) {
+        guard context.launchInfo.launchReason == .uncertain else {
+            // Unexpected pre-set reason, forward as-is.
+            forwardingBlock([(command, context, writer)])
+            return
+        }
+
+        if window.resolvedReason != .uncertain {
```
mariedm: nit: a `guard let` statement could make it a bit easier to read the rest of the code

DatadogRUM/Sources/RUMMonitor/Scopes/Utils/LaunchReasonResolver.swift
```diff
@@ -0,0 +1,120 @@
...
+            // Launch reason already resolved, inject and forward immediately.
+            let resolvedCommand = (command, context.replacing(launchReason: window.resolvedReason), writer)
+            forwardingBlock([resolvedCommand])
+        } else {
+            // Buffer command and check if this one resolved the launch reason.
+            window.buffer(command: command, context: context, writer: writer)
+
+            if window.resolvedReason != .uncertain {
```
mariedm: Since we already check this at line 41, I’d expect this if statement to never be hit. Am I reading this wrong?

DatadogRUM/Sources/RUMMonitor/Scopes/Utils/LaunchReasonResolver.swift
mariedm: This maybe? 
```suggestion
            // Launch reason resolved, forward as-is.
```

mariedm (commented):
Looks good! 🙌 Only left a few nits.

DatadogRUM/Sources/RUMMonitor/Scopes/RUMApplicationScope.swift
ncreated: Fair point 👍. The only actionable element in this could be "report issue to dd-sdk-ios", but we can already track it automatically with telemetry. Removed 🧹.

DatadogRUM/Sources/RUMMonitor/Scopes/Utils/LaunchReasonResolver.swift
ncreated: ✅ I updated the method signature to:
```swift
func deferUntilLaunchReasonResolved(
    command: RUMCommand,
    context: DatadogContext,
    writer: Writer,
    onReady: (RUMCommand, DatadogContext, Writer) -> Void
)
```
The onReady callback is now only invoked **after the launch reason is resolved**. This means:
- It **won’t be called immediately** for some early commands (they are buffered).
- Once the launch reason is resolved, onReady will be **called multiple times** to flush all buffered commands in FIFO order.
- After that, it will behave **one-to-one**: called exactly once per incoming command.

DatadogRUM/Sources/RUMMonitor/Scopes/Utils/LaunchReasonResolver.swift
ncreated: On tvOS, the `DatadogContext.launchInfo.launchReason` is always `.uncertain` when it reaches `RUMApplicationScope`. That’s why we rely on `LaunchReasonResolver` to infer it using the launch window heuristic 🙂.

That said, this check is **purely defensive** — if the `context` already contains a resolved `launchReason`, we simply forward the command without modification. I’d prefer to keep it, since `LaunchReasonResolver` is also used on iOS as a fallback, in case `launchReason` wasn’t resolved (e.g., if the kernel’s `task_policy_get()` call failed).

I updated the comment for more clarity:
```swift
// If the launch reason was already resolved externally (e.g., via iOS `task_role` or prewarm flag),
// forward the command immediately. This is a defensive check — callers should avoid deferring resolution
// when the reason is already known.
guard context.launchInfo.launchReason == .uncertain else {
    onReady(command, context, writer)
    return
}
```

DatadogRUM/Sources/RUMMonitor/Scopes/Utils/LaunchReasonResolver.swift
ncreated: ✅ Fair point, merged both responsibilities into single `LaunchReasonResolver` - looks simpler 👍.

DatadogRUM/Sources/RUMMonitor/Scopes/Utils/LaunchReasonResolver.swift
ncreated: ✅ True, done

DatadogRUM/Sources/RUMMonitor/Scopes/Utils/LaunchReasonResolver.swift
ncreated: No, the `buffer(command:context:writer:)` makes a side-effect of resolving the _launch reason_. This is now outdated however, I refactored this following other suggestions - should be much simpler and cleaner now 🙂.

DatadogRUM/Sources/RUMMonitor/Scopes/RUMApplicationScope.swift
```diff
@@ -56,7 +62,37 @@ internal class RUMApplicationScope: RUMScope, RUMContextProvider {
...
+    /// On iOS, we expect the `launchReason` to be resolved at SDK initialization using `task_role` or the prewarm flag.
+    /// As a safeguard, we fall back to the `LaunchReasonResolver` if the `launchReason` is still `.uncertain`—
+    /// in case the iOS-side assumptions fail (e.g., unhandled platform edge case or kernel API failure).
+    ///
+    /// On tvOS and watchOS, we always use the `LaunchReasonResolver`, since those platforms do not support
+    /// `task_role` or prewarming signals.
+    ///
+    /// - Returns: `true` to indicate that the Application Scope should remain active.
     func process(command: RUMCommand, context: DatadogContext, writer: Writer) -> Bool {
+        #if !os(tvOS) && !os(watchOS)
```
mariedm, line 76: /question: I know we only officially support iOS and tvOS, but do you know what's the behavior of this logic on macOS, visionOS, etc.? Can we use the `task_role` API? I'm wondering why using `#if !os(tvOS) && !os(watchOS)` and not `#if os(iOS)` here.

DatadogRUM/Sources/RUMMonitor/Scopes/Utils/LaunchReasonResolver.swift
```diff
@@ -0,0 +1,122 @@
...
+/// Once the reason is resolved (based on app state, lifecycle events, or time), all buffered
+/// commands are forwarded in FIFO order with the resolved `launchReason` injected into their contexts.
+internal final class LaunchReasonResolver {
+    /// Launch window configuration constants.
+    enum Constants {
+        /// Time to wait before assuming a background launch.
+        static let launchWindowThreshold: TimeInterval = 10.0
+    }
+
+    /// Launch window duration before resolving `.backgroundLaunch`.
```
mariedm, line 26: nit 
```suggestion
    /// Launch window duration before resolving to `.backgroundLaunch`.
```

DatadogRUM/Sources/RUMMonitor/Scopes/Utils/LaunchReasonResolver.swift
```diff
@@ -0,0 +1,122 @@
...
+    private let threshold: TimeInterval
+    /// Buffer of commands received while the launch reason is unresolved.
+    private var buffer: [(command: RUMCommand, context: DatadogContext, writer: Writer)] = []
+    /// Resolved launch reason, updated once a conclusive condition is met.
+    private var resolvedReason: LaunchReason?
+
+    /// Initializes the resolver with a custom or default threshold.
+    ///
+    /// - Parameter launchWindowThreshold: Time window in seconds before resolving as background launch.
```
mariedm, line 35: nit
```suggestion
    /// - Parameter launchWindowThreshold: Time window in seconds before resolving to a background launch.
```

mariedm (approved):
Thanks for the update, it's much clearer! 👌 LGMT! 🚀

maxep (approved):
Looks great!!

DatadogRUM/Sources/RUMMonitor/Scopes/RUMApplicationScope.swift
ncreated: This code fallbacks to the tvOS `LaunchReasonResolver` if anything goes wrong with the `task_role` API and it sends telemetry error 👍. That means we're covered for other platforms 👌.

For the same reason I decided to use broad check - to use most accurate solution on all OS-es. Alternatively, it could be `os(iOS) || os(visionOS) || ...`, but that's just preference.

DatadogRUM/Sources/RUMMonitor/Scopes/Utils/LaunchReasonResolver.swift
ncreated: Actually, both are correct 🙂

DatadogRUM/Sources/RUMMonitor/Scopes/Utils/LaunchReasonResolver.swift
ncreated: Both are correct 👍

PR #2327, opened 2025-06-06
RUM-10115 Collect locale attributes

barboraplasovska:
### What and why?

This PR adds support for collecting locale-related attributes in the RUM event model:

- `locales`: Ordered list of the user’s preferred system languages as IETF language tags.
- `current_locale`: The user's current locale as a language tag (language + region), computed from their preferences and the app's supported languages, e.g. 'es-FR'.
- `time_zone`: The device’s current time zone identifier, e.g. 'Europe/Berlin'.

These attributes provide important context for reproducing locale-specific bugs (e.g., formatting issues), analyzing user language preferences, and understanding time-related behavior across regions.

### Review checklist

### DIFF
--- a/DatadogCore/Tests/Datadog/DatadogCore/Context/LocaleInfoPublisherTests.swift
+++ b/DatadogCore/Tests/Datadog/DatadogCore/Context/LocaleInfoPublisherTests.swift
@@ -12,35 +12,22 @@ import TestUtilities
 class LocaleInfoPublisherTests: XCTestCase {
     private let notificationCenter = NotificationCenter()
 
-    func testPublishLocaleInfo() throws {
-        let expectation = self.expectation(description: "Locale info published")
-
+    func testInitialValueFormatting() throws {
         // Given
         let initialLocale = LocaleInfo(
             locales: ["en-US"],
-            currentLocale: Locale(identifier: "en-US"),
-            timeZone: TimeZone(identifier: "UTC")!
+            currentLocale: Locale(identifier: "en_US"),
+            timeZone: TimeZone(identifier: "GMT")!
         )
-        let publisher = LocaleInfoPublisher(initialLocale: initialLocale, notificationCenter: notificationCenter)
 
-        // When
-        publisher.publish { locale in
-            // Then
-            XCTAssertEqual(locale.locales, Locale.preferredLanguages)
-            XCTAssertEqual(locale.currentLocale, Locale.current.identifier.replacingOccurrences(of: "_", with: "-"))
-            XCTAssertEqual(locale.timeZoneIdentifier, TimeZone.current.identifier)
-
-            expectation.fulfill()
-        }
-
-        notificationCenter.post(name: NSLocale.currentLocaleDidChangeNotification, object: nil)
-
-        waitForExpectations(timeout: 1.0)
-        publisher.cancel()
+        // Then
+        XCTAssertEqual(initialLocale.locales, ["en-US"])
+        XCTAssertEqual(initialLocale.currentLocale, "en-US")
+        XCTAssertEqual(initialLocale.timeZoneIdentifier, "GMT")
     }
 
-    func testUpdatesLocaleInfoWhenLocaleChanges() throws {
-        let expectation = self.expectation(description: "Locale info updated on locale change")
+    func testPublishLocaleInfoOnNotification() throws {
+        let expectation = self.expectation(description: "Locale info published")
         expectation.expectedFulfillmentCount = 2
 
         // Given


DatadogCore/Sources/Core/DatadogCore.swift
```diff
@@ -529,6 +531,8 @@ extension DatadogContextProvider {
         subscribe(\.brightnessLevel, to: BrightnessLevelPublisher(notificationCenter: notificationCenter))
         #endif
 
+        subscribe(\.localeInfo, to: LocaleInfoPublisher(initialLocale: LocaleInfo(), notificationCenter: notificationCenter))
```
ncreated: We rather want to keep it in-sync with what was passed to `DatadogContext`, right?
```suggestion
        subscribe(\.localeInfo, to: LocaleInfoPublisher(initialLocale: locale, notificationCenter: notificationCenter))
```

DatadogInternal/Sources/Context/LocaleInfo.swift
```diff
@@ -0,0 +1,40 @@
...
+    /// Current Locale (language + region) - active locale that influences system formatting (e.g. "en-US")
+    public let currentLocale: String
+
+    /// Time zone identifier (e.g. "Europe/Berlin")
+    public let timeZoneIdentifier: String
+
+    public init(
+        locales: [String],
+        currentLocale: String,
+        timeZoneIdentifier: String
```
ncreated: nit/ We can improve type safety and apply the [ISP](https://en.wikipedia.org/wiki/Interface_segregation_principle) more effectively here.

#### Use Stronger Types Instead of String
The current API accepts plain `String` values for `currentLocale` and `timeZoneIdentifier`, which makes it easy to misuse and harder to validate during code review (I initially assumed that we pass arbitrary values to here somewhere in `DatadogCore`). By requiring `Locale` and `TimeZone` types instead, we gain stronger type safety and push formatting responsibilities into the initializer itself:
```swift
init(locales: [String], currentLocale: Locale, timeZone: TimeZone)
```
Then within the initializer, we can handle formatting (e.g., converting to "en-US" or "Europe/Berlin") in a single, controlled place. This change makes the interface harder to misuse and easier to maintain.

#### Avoid Making the Custom Initializer Public
If the custom initializer is only used in tests or internal logic, it should not be marked public. Prefer using `@testable import DatadogInternal` in test targets. This keeps the public interface focused and clear—right now, having a public initializer that allows arbitrary values could falsely suggest that this is how `LocaleInfo` is constructed in production paths.

DatadogInternal/Sources/Models/CrashReporting/CrashContext.swift
```diff
@@ -35,6 +35,9 @@ public struct CrashContext: Codable, Equatable {
     /// Current device information.
     public let device: DeviceInfo
 
+    /// Current locale information.
+    public let localeInfo: LocaleInfo
```
ncreated: suggestion/ Consider making this property optional for forward compatibility across SDK versions.

As noted in the `CrashContext` documentation:
```
/// Describes current Datadog SDK context, so the app state information can be attached to
/// the crash report and retrieved back when the application is started again.
```
This type is serialized on crash and deserialized on the next app launch. If `LocaleInfo` is non-optional, then decoding will fail when reading a crash context generated by an older SDK version (one that didn't include this field). This failure will prevent the crash report from being sent after an upgrade.

By declaring `LocaleInfo?`, we allow newer versions of the SDK to gracefully handle crash contexts from older versions, ensuring backwards compatibility and improving crash reporting reliability across version boundaries.

ncreated (dismissed):
Looks neat, clean and well-structured 🎯, great job 🏅 ! I left a few suggestions around type safety, access control, and forward compatibility.

DatadogInternal/Sources/Context/LocaleInfo.swift
maxep: I second @ncreated comment, now that you are using specified types you can expose the initialiser more safely. I think it would make sense to expose a public init with defaults parameters, so the Core module can init easily and tests can provide mocks.

```swift
public init(
    locales: [String] = Locale.preferredLanguages,
    currentLocale: Locale = Locale.current,
    timeZone: TimeZone = TimeZone.current
) {
    self.locales = locales
    self.currentLocale = currentLocale.identifier.replacingOccurrences(of: "_", with: "-")
    self.timeZoneIdentifier = timeZone.identifier
}
```
That's a suggestion, not a change request :) current definition is also fine 👍

DatadogCore/Tests/Datadog/DatadogCore/Context/LocaleInfoPublisherTests.swift
```diff
@@ -0,0 +1,83 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import XCTest
+import TestUtilities
+@testable import DatadogInternal
+@testable import DatadogCore
+
+class LocaleInfoPublisherTests: XCTestCase {
```
maxep, line 12: I don't see the difference in coverage between these 2 tests 🤔  Also, we should test the `initialValue` formatting:
```swift
// Given
let initialLocale = LocaleInfo(
    locales: ["en-US"],
    currentLocale: Locale(identifier: "en-US"),
    timeZone: TimeZone(identifier: "UTC")!
)

XCTAssertEqual(initialLocale.locales, ["en-US"])
XCTAssertEqual(initialLocale.currentLocale, "en-US")
XCTAssertEqual(initialLocale.timeZoneIdentifier, "UTC")
```

PR #2332, opened 2025-06-12
Fix: code generation printer

barboraplasovska:
### What and why?

This PR fixes a bug in the generator that wrongly prints comments when they contained `\n`.

### Review checklist


mariedm (dismissed):
LGMT! Thanks for fixing it.

tools/rum-models-generator/Sources/CodeGeneration/Print/ObjcInteropPrinter.swift
```diff
@@ -528,4 +528,16 @@ public class ObjcInteropPrinter: BasePrinter, CodePrinter {
             throw Exception.unimplemented("Cannot print `objcToSwiftCast()` for \(type(of: swiftType)).")
         }
     }
+
+    // MARK: - Helpers
+
+    private func printComment(_ comment: String?) {
```
ncreated: 🤔 💭 Is this used anywhere?

ncreated (changes requested):
🎯 Looks good, but we need tests 🧪 🙂
- [SwiftPrinterTests.swift](https://github.com/DataDog/dd-sdk-ios/blob/develop/tools/rum-models-generator/Tests/CodeGenerationTests/Print/SwiftPrinterTests.swift)
- [ObjcInteropPrinterTests.swift](https://github.com/DataDog/dd-sdk-ios/blob/develop/tools/rum-models-generator/Tests/CodeGenerationTests/Print/ObjcInteropPrinterTests.swift)

ncreated (dismissed):
🎯 🏅

simaoseica-dd (dismissed):
👌

mariedm:
It seems that the ObjC counterpart is now missing?

barboraplasovska:
> It seems that the ObjC counterpart is now missing?

Well yes, because we don't do any comments in the objc-interop.

PR #2334, opened 2025-06-12
RUM-10234 Add internal telemetry for app launch logic

ncreated:
### What and why?

📦 This PR adds new attributes to the "Session Ended" telemetry for improved observability of app launch tracking. It also introduces a generic process_uptime attribute to all telemetry events to pinpoint their timing relative to process start.

#### ⏱️ `process_uptime` attribute

- **`process_uptime`**: Number of milliseconds from process start to when the telemetry event was recorded. Present on all ERROR, DEBUG, and METRIC telemetry events.

This attribute helps correlate events to process age and reason about the process lifetime.

#### 🚀 `launch_info.*` attributes

- **`launch_reason`**: The reason the app process was launched ("user launch", "background launch", or "prewarming").
- **`task_role`**: The process’s task policy role (mapped from `task_role_t`).
- **`prewarmed`**: Whether the app was prewarmed by the system.
- **`tt_sdk_init`**: Time (ms) from process start to SDK initialization.
- **`tt_become_active`**: Time (ms) from process start to first `applicationDidBecomeActive`, or `null`.
- **`has_scenes_lifecycle`**: Whether the app uses the UIScene lifecycle.
- **`app_state_at_sdk_init`**: App state at the time of SDK init ("active", "background", or "inactive").

These attributes provide insight into how and why the app was launched, key lifecycle timings, and the app state at SDK initialization - crucial for diagnosing start scenarios and classifying launch types.

#### ♻️ `lifecycle_info.*` attributes

- **`tt_session_start`**: Time (ms) from process start to session start.
- **`session_count`**: Number of RUM sessions before this one in the same process.
- **`state_at_start`**: App state when session started.
- **`state_at_end`**: App state when session ended.
- **`fg_coverage`**: Percent of session duration in foreground (`0.0` to `1.0`, optional).

These attributes describe when the session occurred within the process, how many sessions preceded it, and how the app state evolved — useful for understanding and correlating session context and foreground activity to other session timings.

### Review checklist


DatadogRUM/Sources/SDKMetrics/SessionEndedMetric.swift
```diff
@@ -340,10 +351,66 @@ internal class SessionEndedMetric {
...
+        }
+
+        /// Information about the app process launch. Shared between all sessions within the same app process.
+        let launchInfo: LaunchInfo
+
+        struct LifecycleInfo: Encodable {
+            /// Time in milliseconds from process start to the beginning of this session.
+            let timeToSessionStart: Int64
+            /// The number of RUM sessions created within this app process prior to this one.
+            let sessionsCount: Int
```
ncreated, line 422: ❓ Perhaps it could be named better, `previous_sessions`? WDYT?

maciejburda: Is this growing to infinity?

I'm fine with current name. Doc explains this well enough.

DatadogRUM/Sources/SDKMetrics/SessionEndedMetricController.swift
```diff
@@ -18,6 +18,10 @@ internal final class SessionEndedMetricController {
     /// Array to keep track of pending session IDs in their start order.
     private var pendingSessionIDs: [RUMUUID] = []
 
+    /// The number of sessions tracked in this SDK instance.
+    /// Only includes sessions that tracked at least one view.
+    private var validSessionCount = 0
```
maciejburda, line 23: This one is interesting. I like the idea to validate this.

maciejburda (approved):
LGTM

Datadog/IntegrationUnitTests/Public/CoreTelemetryIntegrationTests.swift
```diff
@@ -48,7 +48,7 @@ class CoreTelemetryIntegrationTests: XCTestCase {
 
         let debug = debugEvents[0]
         XCTAssertEqual(debug.telemetry.message, "Debug Telemetry")
-        DDAssertReflectionEqual(debug.telemetry.telemetryInfo, ["debug.attribute": 42])
+        DDAssertReflectionEqual(debug.telemetry.telemetryInfo["debug.attribute"] as? Int, 42)
```
maxep: nit/
```suggestion
        XCTAssertEqual(debug.telemetry.telemetryInfo["debug.attribute"] as? Int, 42)
```

DatadogInternal/Sources/Context/DatadogContext.swift
```diff
@@ -83,6 +83,11 @@ public struct DatadogContext {
     /// Provides the history of app foreground / background states.
     public var applicationStateHistory: AppStateHistory
 
+    /// If the app supports scenes and doesn’t use an app delegate object to manage transitions to and from the foreground or background.
+    /// It doesn't imply whether the app has an actual `UISceneDelegate` implementation or behavior - it only means that app declares `UIApplicationSceneManifest`
+    /// in its `Info.plist`. The presence of that key changes the initiali lifecycle of thea app (in case of user launch, apps with scene manifest start shortly in BACKGROUND before moving to INACTIVE).
+    public let isUsingSceneLifecycle: Bool
+
```
maxep: suggestion/ I'm a bit worried about the core context's increasing number of properties :/ I wonder if we can get this from RUM module only, by adding the `bundle` to the RUM's config and use this for Telemetry, would that work?

DatadogRUM/Sources/SDKMetrics/SessionEndedMetric.swift
ncreated: Yes, it's only incremented - within the current instance of the process ofc, not persisted.

DatadogInternal/Sources/Context/DatadogContext.swift
ncreated: That makes sense 👍, especially that we only need it for telemetry. Moved to `RUM.Configuration.bundle` ✅.

maxep (approved):
👍

PR #2341, opened 2025-06-17
RUM-9540: Introduce new category for network errors

simaoseica-dd:
### What and why?

This PR adds a new category, `Network`, to the [@error.category](https://github.com/DataDog/rum-events-format/blob/master/schemas/rum/error-schema.json#L104).
This new error category aims to tag network errors triggered by no connections, connections cancelled, request timeouts and whatnot.

### How?

This new category was primarily [defined](https://github.com/DataDog/rum-events-format/pull/281) in the error schema. 
The iOS SDK applies it under the `RUMResource` scope by verifying the information stored by the `RUMStopResourceWithError` command.

### Review checklist

### DIFF
--- a/DatadogRUM/Sources/RUMMonitor/RUMCommand.swift
+++ b/DatadogRUM/Sources/RUMMonitor/RUMCommand.swift
@@ -412,6 +412,8 @@ internal struct RUMStopResourceWithErrorCommand: RUMResourceCommand {
     let errorSource: RUMInternalErrorSource
     /// The platform type of the error (iOS, React Native, ...)
     let errorSourceType: RUMErrorEvent.Error.SourceType
+    /// If error is related with the network connection.
+    let isNetworkError: Bool
     /// Error stacktrace.
     let stack: String?
     /// HTTP status code of the Ressource error.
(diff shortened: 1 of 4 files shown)


DatadogRUM/Sources/RUMMonitor/RUMCommand.swift
```diff
@@ -461,6 +464,8 @@ internal struct RUMStopResourceWithErrorCommand: RUMResourceCommand {
         let dderror = DDError(error: error)
         self.errorMessage = dderror.message
         self.errorType = dderror.type
+        let nsError = error as NSError
+        self.isNetworkError = nsError.domain == NSURLErrorDomain
```
maxep, line 468: blocker/ Is this the right condition to categories an error as `network`? I think all resource errors should be categorised as network error, no?

maxep (changes requested):

DatadogRUM/Sources/RUMMonitor/RUMCommand.swift
simaoseica-dd: We already categorize all the resource errors as a `network` source. 
In regards to the category, this new `Network` will subgroup the errors that fail to reach the server from `http errors` that succeed at the transport level.

maxep: Following discussion with the team, we are good to keep `exception` for other error domains.

PR #2345, opened 2025-06-17
Added ability to block uploads on networks with "Low Data Mode" enabled

kmusco-foreflight:
### What and why?

This PR allows us to configure the SDK to block uploads when connected to a network with "Low Data Mode" enabled. 

Why?
A large number of our users rely on satellite internet with plans that charge based on data consumption. There are also scenarios where the connection has limited bandwidth and we only want to allow critical services to make network requests. 

### How?

This PR adds a `constrainedNetworkAccessEnabled` parameter that can be passed to `Datadog.init()`. This name was chosen to based on URLSession's [allowsConstrainedNetworkAccess](https://developer.apple.com/documentation/foundation/urlsessionconfiguration/allowsconstrainednetworkaccess) property which specifies if "Low Data Mode" is enabled. The property defaults to `true` to maintain the current behavior and will disable uploads when set to `false`.

The SDK is already tracking whether or not the network is constrained in [NWPathMonitorPublisher](https://github.com/DataDog/dd-sdk-ios/blob/968bf9586048d2d81aae3fd096bd2fcc0cb9ce8b/DatadogCore/Sources/Core/Context/NetworkConnectionInfoPublisher.swift#L20) with the `isConstrained` property, and has a mechanism for blocking uploads with [DataUploadConditions.blockersForUpload()](https://github.com/DataDog/dd-sdk-ios/blob/968bf9586048d2d81aae3fd096bd2fcc0cb9ce8b/DatadogCore/Sources/Core/Upload/DataUploadConditions.swift#L30). This PR updates `blockersForUpload()` to block uploads if `isConstrained` is `true` and the `allowsConstrainedNetworkAccess` parameter is `false`.

### Review checklist

### DIFF
--- a/DatadogCore/Sources/Core/Upload/DataUploadConditions.swift
+++ b/DatadogCore/Sources/Core/Upload/DataUploadConditions.swift
@@ -10,6 +10,7 @@ import DatadogInternal
 /// Tells if data upload can be performed based on given system conditions.
 internal struct DataUploadConditions {
     enum Blocker {
+        case allowsConstrainedNetworkAccessFalse
         case battery(level: Int, state: BatteryStatus.State)
         case lowPowerModeOn
         case networkReachability(description: String)
@@ -22,9 +23,13 @@ internal struct DataUploadConditions {
 
     /// Battery level above which data upload can be performed.
     let minBatteryLevel: Float
+    
+    /// Blocks uploads on networks with "Low Data Mode" enabled when set to `false`. Defaults to `true`.
+    let allowsConstrainedNetworkAccess: Bool
 
-    init(minBatteryLevel: Float = Constants.minBatteryLevel) {
+    init(minBatteryLevel: Float = Constants.minBatteryLevel, allowsConstrainedNetworkAccess: Bool = true) {
         self.minBatteryLevel = minBatteryLevel
+        self.allowsConstrainedNetworkAccess = allowsConstrainedNetworkAccess
     }
 
     func blockersForUpload(with context: DatadogContext) -> [Blocker] {
@@ -37,6 +42,9 @@ internal struct DataUploadConditions {
         let networkIsReachable = reachability == .yes || reachability == .maybe
         if !networkIsReachable {
             blockers = [.networkReachability(description: reachability.rawValue)]
+        }
+        if let isConstrained = context.networkConnectionInfo?.isConstrained, isConstrained, !allowsConstrainedNetworkAccess {
+            return [.allowsConstrainedNetworkAccessFalse]
         }
         #endif
 
--- a/DatadogCore/Sources/Datadog.swift
+++ b/DatadogCore/Sources/Datadog.swift
@@ -135,49 +135,57 @@ public enum Datadog {
         ///
         /// `false` by default.
         public var backgroundTasksEnabled: Bool
+        
+        /// Allows uploads on networks with "Low Data Mode" enabled.
+        /// When set to `false`, uploads are deferred until the device is connected to an unconstrained network.
+        ///
+        /// Defaults to `true`.
+        public var constrainedNetworkAccessEnabled: Bool
 
         /// Creates a Datadog SDK Configuration object.
         ///
         /// - Parameters:
-        ///   - clientToken:                Either the RUM client token (which supports RUM, Logging and APM) or regular client token,
-        ///                                 only for Logging and APM.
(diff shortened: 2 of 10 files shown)


DatadogCore/Sources/Core/Upload/DataUploadConditions.swift
```diff
@@ -23,8 +24,12 @@ internal struct DataUploadConditions {
     /// Battery level above which data upload can be performed.
     let minBatteryLevel: Float
 
-    init(minBatteryLevel: Float = Constants.minBatteryLevel) {
+    /// Blocks uploads on networks with "Low Data Mode" enabled when set to `false`. Defaults to `true`.
+    let allowsConstrainedNetworkAccess: Bool
```
simaoseica-dd: suggestion: This naming references the [urlsessionconfiguration API](https://developer.apple.com/documentation/foundation/urlsessionconfiguration/allowsconstrainednetworkaccess) and it seems not used in the logic below which can lead to confusion. 
Can you rename the parameters around the API used ([NWPath.isConstrained](https://developer.apple.com/documentation/networkextension/nwpath/isconstrained))?
Here it can be `isNetworkAccessConstrained` with the opposite value of it.

simaoseica-dd: The same for `case allowsConstrainedNetworkAccessFalse`. It could be `case constrainedNetworkAccess`

DatadogCore/Sources/Core/Upload/DataUploadConditions.swift
```diff
@@ -38,6 +43,9 @@ internal struct DataUploadConditions {
         if !networkIsReachable {
             blockers = [.networkReachability(description: reachability.rawValue)]
         }
+        if let isConstrained = context.networkConnectionInfo?.isConstrained, isConstrained, !allowsConstrainedNetworkAccess {
```
simaoseica-dd:
```suggestion
        if context.networkConnectionInfo?.isConstrained == true,
        isNetworkAccessConstrained {
```

DatadogCore/Sources/Datadog.swift
```diff
@@ -136,48 +136,56 @@ public enum Datadog {
         /// `false` by default.
         public var backgroundTasksEnabled: Bool
 
+        /// Allows uploads on networks with "Low Data Mode" enabled.
+        /// When set to `false`, uploads are deferred until the device is connected to an unconstrained network.
+        ///
```
simaoseica-dd, line 142:
```suggestion
        /// This can lead to data loss if "Low Data Mode" remains enabled and the batches become obsolete.
        ///
        /// Defaults to `false`.
```

DatadogCore/Sources/Datadog.swift
```diff
@@ -136,48 +136,56 @@ public enum Datadog {
...
+        ///                                             This can lead to significant time shift in RUM sessions or distributed traces.
+        ///   - backgroundTasksEnabled:                 A flag that determines if `UIApplication` methods
+        ///                                             `beginBackgroundTask(expirationHandler:)` and `endBackgroundTask:`
+        ///                                             are used to perform background uploads.
+        ///                                             It may extend the amount of time the app is operating in background by 30 seconds.
+        ///                                             Tasks are normally stopped when there's nothing to upload or when encountering
+        ///                                             any upload blocker such us no internet connection or low battery.
+        ///                                             By default it's set to `false`.
+        ///   - constrainedNetworkAccessEnabled:        A flag that determines if uploads should be enabled on networks with "Low Data Mode" enabled.
```
simaoseica-dd:
```suggestion
        ///   - isNetworkAccessConstrained: A flag that determines if uploads should be constrained on networks with "Low Data Mode" enabled.
       /// By default it's set to `false`.
```

simaoseica-dd (commented):
Thank you for the contribution @kmusco-foreflight 🙏 
And for providing the use case. It makes sense to have this constraint on data uploads.

I’ve left a few comments regarding naming to help improve readability.

Additionally, I’ll change the base branch to include telemetry for this change so we can monitor it going forward.

DatadogCore/Sources/Core/Upload/DataUploadConditions.swift
kmusco-foreflight: I applied your naming suggestions 👍

PR #2350, opened 2025-06-20
Add SwiftUI privacy overrides to benchmark tests

gonzalezreal:
### What and why?

Sprinkle SwiftUI Session Replay privacy overrides in benchmark tests to be able to detect any performance degradation introduced by this new API.

### How?

- Add privacy override functionality to the internal `DatadogMonitor`
- Use the SwiftUI environment to inject a `DatadogMonitor` instance, instead of using initializer injection
- Add convenience methods that use the `DatadogMonitor` to track or override privacy settings.
- Add privacy overrides in the SwiftUI catalog

### DIFF
--- a/BenchmarkTests/CatalogSwiftUI/ContentView.swift
+++ b/BenchmarkTests/CatalogSwiftUI/ContentView.swift
@@ -40,11 +40,7 @@ public struct ContentView: View {
                                               startRadiusFraction: 0.3,
                                               endRadiusFraction: 5)
 
-    private let monitor: DatadogMonitor
-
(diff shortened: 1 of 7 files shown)


maxep (approved):
Very nice, it is much cleaner using env values!

gonzalezreal:
> Very nice, it is much cleaner using env values!

Thanks @maxep! Added a small new change: Wrap `SessionReplayPrivacyView` in `AnyView` instead of `AnyViewModifier`, to make it closer to the real API.

maxep (approved):

PR #2351, opened 2025-06-20
RUM-10378 Collect battery attributes

barboraplasovska:
### What and why?

This PR adds support for collecting battery-related attributes in the RUM event model:
- `battery_level`: The current battery charge level, expressed as a float from 0.0 (empty) to 1.0 (full).
- `low_power_mode`: A boolean indicating whether the device is currently in Low Power Mode.
- `brightness_level`: The current screen brightness, normalized as a float between 0.0 (darkest) and 1.0 (brightest).

These attributes provide crucial context for analyzing app performance and behavior under different power conditions. They help teams understand usage patterns on low-battery devices, identify potential regressions caused by power-saving configurations, and optimize experiences for energy-constrained environments

### Review checklist


DatadogInternal/Sources/Context/BrightnessStatus.swift
```diff
@@ -0,0 +1,17 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+/// Describe the battery state for mobile devices.
+public struct BrightnessStatus: Codable, Equatable {
+    /// The brightness level
+    public let level: Float
```
maciejburda: If we don't expect more properties - could it be flattened to use float directly?

We could `typealias BrightnessLevel = Float` for readability.

DatadogCore/Sources/Core/DatadogCore.swift
```diff
@@ -525,6 +525,10 @@ extension DatadogContextProvider {
         subscribe(\.isLowPowerModeEnabled, to: LowPowerModePublisher(notificationCenter: notificationCenter, processInfo: processInfo))
         #endif
 
+        #if os(iOS)
```
maciejburda, line 528: Is there anything for tvOS on that front?

maciejburda (dismissed):
LGTM - left some minor questions/suggestions

DatadogCore/Sources/Core/DatadogCore.swift
maxep: Yes, this should be available on tvOS as well!

maxep (dismissed):
Disregard my previous comment, `brightness` is not available on tvOS

DatadogCore/Sources/Core/DatadogCore.swift
barboraplasovska: Hey! Actually I checked on the [documentation](https://developer.apple.com/documentation/uikit/uiscreen/brightness) and it's not available for tvOS.

DatadogInternal/Sources/Context/BrightnessStatus.swift
barboraplasovska: You're right ! I've made the changes.

PR #2359, opened 2025-06-24
RUM-10188: Remove fatal errors from Logs

simaoseica-dd:
### What and why?

This PR stops reporting fatal errors into the `Logs` product feature. 

This change aims to reduce the redundancy of fatal errors in `Error Tracking` and `Logs`. 
From now on all the fatal errors will be centralized on the [Error Tracking for RUM](https://docs.datadoghq.com/real_user_monitoring/error_tracking/mobile/ios/?tab=swiftpackagemanagerspm).  

### How?

It removes the `CrashLogReceiver` from the `Logs` product feature. 

### Review checklist

### DIFF
--- a/DatadogLogs/Sources/RemoteLogger.swift
+++ b/DatadogLogs/Sources/RemoteLogger.swift
@@ -101,9 +101,6 @@ internal final class RemoteLogger: LoggerProtocol, Sendable {
     }
 
     func internalLog(level: LogLevel, message: String, error: DDError?, attributes: [String: Encodable]?) {
-        guard (attributes?[CrossPlatformAttributes.errorLogIsCrash]?.dd.decode() ?? false) == false else {
-            return
-        }
         guard configuration.sampler.sample() else {
             return
         }
@@ -121,7 +118,6 @@ internal final class RemoteLogger: LoggerProtocol, Sendable {
         let loggerAttributes = loggerAttributes.getAttributes()
         var logAttributes = attributes
 
-        let isCrash = logAttributes?.removeValue(forKey: CrossPlatformAttributes.errorLogIsCrash)?.dd.decode() ?? false
         let errorFingerprint: String? = logAttributes?.removeValue(forKey: Logs.Attributes.errorFingerprint)?.dd.decode()
         let addBinaryImages = logAttributes?.removeValue(forKey: CrossPlatformAttributes.includeBinaryImages)?.dd.decode() ?? false
         let userAttributes = loggerAttributes
@@ -184,7 +180,7 @@ internal final class RemoteLogger: LoggerProtocol, Sendable {
             ) { log in
                 writer.write(value: log)
 
-                guard (log.status == .error || log.status == .critical) && !isCrash else {
+                guard log.status == .error || log.status == .critical else {
                     return
                 }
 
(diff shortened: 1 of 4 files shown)


DatadogLogs/Sources/Feature/MessageReceivers.swift
```diff
@@ -61,101 +61,6 @@ internal struct LogMessageReceiver: FeatureMessageReceiver {
...
-        let date = (report.date ?? dateProvider.now)
-            .addingTimeInterval(crashContext.serverTimeOffset)
-
-        var errorAttributes: [AttributeKey: AttributeValue] = [:]
-
-        // Set crash attributes for the error
-        errorAttributes[DDError.threads] = report.threads
-        errorAttributes[DDError.binaryImages] = report.binaryImages
-        errorAttributes[DDError.meta] = report.meta
-        errorAttributes[DDError.wasTruncated] = report.wasTruncated
```
ncreated, line 96: suggestion/ We can remove these `DDError` constants. They were initially used when RUM and Logs relied on error attributes for crash-specific fields. Since we already moved to the type-safe `DDCrashReport` in RUM and now we remove this concept from Logs, they're no longer needed.

DatadogLogs/Sources/RemoteLogger.swift
```diff
@@ -101,6 +101,9 @@ internal final class RemoteLogger: LoggerProtocol, Sendable {
     }
 
     func internalLog(level: LogLevel, message: String, error: DDError?, attributes: [String: Encodable]?) {
+        guard (attributes?[CrossPlatformAttributes.errorLogIsCrash]?.dd.decode() ?? false) == false else {
+            return
+        }
```
ncreated: question/ What’s the motivation for this check? It seems redundant with line 124:
```swift
let isCrash = logAttributes?.removeValue(forKey: CrossPlatformAttributes.errorLogIsCrash)?.dd.decode() ?? false
```

If the intent is to stop CP SDKs from sending crashes as Logs, a better approach might be to remove `CrossPlatformAttributes.errorLogIsCrash` from the native SDK and align with CP SDKs to stop using it in V3. This way, crash logs won’t be sent at all.

Otherwise, this check feels overly defensive—guarding against a case that shouldn't occur and adding unnecessary complexity.

ncreated (approved):
Looks great 👍. I left minor suggestions.

DatadogLogs/Sources/Feature/MessageReceivers.swift
simaoseica-dd: Good catch. ty

DatadogLogs/Sources/RemoteLogger.swift
simaoseica-dd: This is a good point. We want to make sure the CP SDKs are all aligned with this change. 
I also agree that removing the `CrossPlatformAttributes.errorLogIsCrash` attribute brings some benefits:
- It’s a breaking change that makes CP SDKs explicitly aware of the update
- It simplifies and cleans up the CP logic

I’ve followed up with the CP owners and applied the suggested change.

ncreated (approved):
🎯

PR #2362, opened 2025-06-25
RUM-10492: Refactor `DisplayList.ViewInfo` and `DisplayList.Content.Value` for iOS 26 support

gonzalezreal:
### What and why?

iOS 26 introduces some structural changes in the `DisplayList` type we rely on for SwiftUI Session Replay:

- `ViewInfo.view` and `ViewInfo.container` can either be a `UIView` or a `CALayer` (the same as the `ViewInfo.layer` in this case)
- The payload for the `DisplayList.Content.Value.color` case is a `ColorView` instead of a `SwiftUI.Color._Resolved`.

We need to address these differences to improve SwiftUI Session Replay on iOS 26.

### How?

- Make the `ViewInfo.view` property reflection optional, falling back to a `CALayer` for `ViewInfo.container`
- Add an internal `ColorView` type with a `color` property, matching the structure for the payload on iOS 26.

### Review checklist


maxep (approved):
Nice findings!

PR #2364, opened 2025-06-26
RUM-10190: Sample traces according to `session.id`

simaoseica-dd:
### What and why?

This PR ensures that trace sampling is deterministic based on the `session.id` when the RUM feature is enabled.
This guarantees a consistent sampling decision for each session, which is important for features like RUM that rely on [head-based instrumentation](https://github.com/DataDog/dd-sdk-ios/blob/053b4d6b978ae9dba658b8e44a9212d36d1f5a7b/DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift#L168-L193) strategies.

### How?

A `DeterministicSampler` is introduced in the `TracingURLSessionHandler` to make trace sampling consistent per `session.id`. The `session.id` is extracted from the `RUMContext`, which implies that the RUM feature is enabled.
If RUM is not active, the sampling falls back to using the `traceID`.

### Review checklist

### DIFF
--- a/DatadogInternal/Sources/Utils/DeterministicSampler.swift
+++ b/DatadogInternal/Sources/Utils/DeterministicSampler.swift
@@ -1,0 +1,43 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+/// Deterministic sampler that makes consistent sampling decisions for a given `seed`.
+///
+/// This sampler uses Knuth hashing to compute a uniform hash of the `seed`, allowing
+/// deterministic sampling based on a sampling rate `samplingRate`.
+///
+/// Conforms to the `Sampling` protocol.
+public struct DeterministicSampler: Sampling {
+    enum Constants {
+        /// Good number for Knuth hashing (large, prime, fit in 64 bit long)
+        internal static let samplerHasher: UInt64 = 1_111_111_111_111_111_111
+        internal static let maxID: UInt64 = 0xFFFFFFFFFFFFFFFF
+    }
+
+    /// Value between `0.0` and `100.0`, where `0.0` means NO event will be sent and `100.0` means ALL events will be sent.
+    public let samplingRate: SampleRate
+    /// Persisted sampling decision.
+    private let shouldSample: Bool
+
+    /// Initializes a new instance of `DeterministicSampler`.
+    ///
+    /// - Parameters:
+    ///   - seed: A 64-bit unsigned integer used as the base input for Knuth hashing.
+    ///   - samplingRate: A percentage value between `0.0` and `100.0`.
+    ///     - `0.0` disables sampling entirely (no data is sampled).
+    ///     - `100.0` enables full sampling (all data is sampled).
+    public init(seed: UInt64, samplingRate: SampleRate) {
+        // We use overflow multiplication to create a "randomized" hash based on the `seed`
+        let hash = seed &* Constants.samplerHasher
+        let threshold = Float(Constants.maxID) * samplingRate.percentageProportion
+        self.samplingRate = samplingRate
+        self.shouldSample = Float(hash) < threshold
+    }
+
+    /// Based on the `seed` and sampling rate, it returns consistent value decisions.
+    /// - Returns: `true` if data should be "sampled".
+    public func sample() -> Bool { shouldSample }
+}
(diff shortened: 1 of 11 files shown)


DatadogInternal/Sources/Utils/DeterministicSampler.swift
```diff
@@ -0,0 +1,28 @@
...
+        internal static let samplerHasher: UInt64 = 1_111_111_111_111_111_111
+        internal static let maxID: UInt64 = 0xFFFFFFFFFFFFFFFF
+    }
+
+    /// Value between `0.0` and `100.0`, where `0.0` means NO event will be sent and `100.0` means ALL events will be sent.
+    public let samplingRate: SampleRate
+    /// Persisted sampling decision.
+    private let shouldSample: Bool
+
+    public init(baseId: UInt64, samplingRate: SampleRate) {
```
maxep: suggestion/ I would suggest to use `seed` instead of `baseId`, which is generally used for pseudo randomisation. Also, as this can be used by more features, could you add reference documentation to the struct and init that describes what it does?

DatadogTrace/Sources/DatadogTracer.swift
```diff
@@ -77,19 +77,36 @@ internal final class DatadogTracer: OTTracer, OpenTelemetryApi.Tracer {
...
         let parentSpanContext = references?.compactMap { $0.context.dd }.last ?? activeSpan?.context as? DDSpanContext
+        let traceID = parentSpanContext?.traceID ?? defaultTraceID
         return startSpan(
-            spanContext: createSpanContext(parentSpanContext: parentSpanContext, using: localTraceSampler),
+            spanContext: createSpanContext(parentSpanContext: parentSpanContext, traceID: traceID, using: localTraceSampler),
             operationName: operationName,
             tags: tags,
             startTime: startTime
         )
     }
```
maxep: What is the rational here?
- We are in the section of OpenTracing compliance, but this method is not part of the interface.
- It seems to not change the behaviour, `defaultTraceID` will always be `traceIDGenerator.generate()`. Or am I missing something?

DatadogTrace/Sources/Integrations/TracingURLSessionHandler.swift
```diff
@@ -124,15 +132,23 @@ internal struct TracingURLSessionHandler: DatadogURLSessionHandler {
...
+                // Span context may not be injected on iOS13+ if `URLSession.dataTask(...)` for `URL`
+                // was used to create the session task.
+                span = tracer.startSpan(
+                    operationName: "urlsession.request",
+                    defaultTraceID: defaultTraceID,
+                    startTime: resourceMetrics.fetch.start
+                )
+            } else {
+                return
+            }
```
maxep: comment/ I'm not sure this is necessary since we don't complete the interception in `Trace` when `RUM` is enabled. This is part of the `guard` above:
```swift
interception.origin != "rum", // if that request was tracked as RUM resource, the RUM backend will create the span on our behalf
```

maxep: Reading more closely, it seems that we use a random seed for a deterministic sampler, making it non deterministic. I think we can just use a `Sampler` type instead.

My suggestion would be to replace the previous `distributedTraceSampler` with a `defaultSampler` that is used when no RUM or Trace context is available.

DatadogTrace/Sources/Integrations/TracingURLSessionHandler.swift
```diff
@@ -7,43 +7,51 @@
...
+        let traceID = parentSpanContext?.traceID ?? tracer.traceIDGenerator.generate()
+
+        let sampler = samplerBySessionID(defaultID: traceID.idLo)
+        distributedTraceSampler = sampler
+
         let spanContext = tracer.createSpanContext(
             parentSpanContext: parentSpanContext,
-            using: distributedTraceSampler
+            traceID: traceID,
+            using: sampler
```
maxep: suggestion/ I think we can simplify this by not storing the `distributedTraceSampler` property and by not calling `tracer.createSpanContext` method.

We should build the Span context here by using `parentSpanContext`, `tracer.spanIDGenerator`, and a local sampler seeded by either `contextReceiver.context.rumContext?.sessionID` or `parentSpanContext.traceID` OR use a default sampler if non is available.

DatadogTrace/Sources/DatadogTracer.swift
maxep: You can ignore this comment, this change is related to the `TracingURLSessionHandler`. I guess we can avoid this by simplifying the handler 👍

DatadogTrace/Sources/Integrations/TracingURLSessionHandler.swift
simaoseica-dd: As discussed, this step should not be triggered. But if so, we just need a simple `Sampler`.

DatadogTrace/Sources/Integrations/TracingURLSessionHandler.swift
simaoseica-dd: Agree. The suggestion simplifies the logic and readability.

PR #2377, opened 2025-07-01
RUM-9068: Fix propagation of global attributes on expired user actions

simaoseica-dd:
### What and why?

This PR fixes an issue where global attributes were not properly propagated for expired user actions.

### How?

Global attributes are propagated via `RUMCommand` to capture a snapshot of the global environment at the exact moment an event occurs.
However, user actions that time out should also reflect the global environment at their expiration point. To address this, a `RUMStopUserActionCommand` is now propagated to properly close the expired user action with the appropriate snapshot of the global attributes.

### Review checklist

### DIFF
--- a/DatadogRUM/Sources/RUMMonitor/Scopes/RUMUserActionScope.swift
+++ b/DatadogRUM/Sources/RUMMonitor/Scopes/RUMUserActionScope.swift
@@ -106,7 +106,8 @@ internal class RUMUserActionScope: RUMScope, RUMContextProvider {
 
     func process(command: RUMCommand, context: DatadogContext, writer: Writer) -> Bool {
         if let expirationTime = possibleExpirationTime(currentTime: command.time), allResourcesCompletedLoading() {
-            sendActionEvent(completionTime: expirationTime, on: nil, context: context, writer: writer)
+            // Stop user action due to timeout
+            sendActionEvent(completionTime: expirationTime, on: command, context: context, writer: writer)
             return false
         }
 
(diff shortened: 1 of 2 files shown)


DatadogRUM/Sources/RUMMonitor/Scopes/RUMUserActionScope.swift
```diff
@@ -106,7 +106,19 @@ internal class RUMUserActionScope: RUMScope, RUMContextProvider {
...
-            sendActionEvent(completionTime: expirationTime, on: nil, context: context, writer: writer)
+            // Stop user action due to timeout
+            sendActionEvent(
+                completionTime: expirationTime,
+                on: RUMStopUserActionCommand(
+                    time: command.time,
+                    globalAttributes: command.globalAttributes,
+                    attributes: [:],
+                    actionType: .custom, // ignored
+                    name: nil
```
ncreated: suggestion/ It feels a bit hacky, especially given the arbitrary handling - mocking actionType, ignoring current attributes, and forwarding global unconditionally. Ideally, we shouldn't be fabricating command types within the RUM implementation - it's harmful to debugging and can obscure the real source of issues.

A cleaner alternative might be to forward the command directly:
```diff
- on: nil
+ on: command
```
Then, only consider `command.attributes` when the command type isn't `RUMStopUserActionCommand`.

WDYT?

Valpertui (approved):

DatadogRUM/Sources/RUMMonitor/Scopes/RUMUserActionScope.swift
simaoseica-dd: Agree, we should propagate the current command 👍

PR #2386, opened 2025-07-04
Fix support for Toolbar items in Session Replay on iOS 26

barboraplasovska:
### What and why?

In iOS 26, Toolbar items are now rendered as drawings and since in session replay we do not support drawings, this caused them to give `unsupported swiftui component` error in the replay.

### How?

As a temporary solution, we extract the text from these toolbar item drawings and display them as text wireframes to ensure users can see the button labels in session replay instead of having no content.

### Review checklist

### DIFF
--- a/DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/DisplayList+Reflection.swift
+++ b/DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/DisplayList+Reflection.swift
@@ -152,8 +152,23 @@ extension DisplayList.Content.Value: Reflection {
         case let (.enum("image"), image):
             self = try .image(reflector.reflect(image))
 
-        case (.enum("drawing"), _):
-            self = .unknown
+        case let (.enum("drawing"), drawing):
+            // In iOS 26, toolbar items are now rendered as drawings which are not currently supported.
+            // As a temporary fix, we extract the text from the drawing and display it as text content.
+            if #available(iOS 26, tvOS 26, *) {
+                do {
+                    guard let drawingContent = drawing else {
+                        self = .unknown
+                        return
+                    }
+                    let extractedText = try Self.extractTextFromGlyphs(drawingContent)
+                    self = .toolbarItem(extractedText)
+                } catch {
+                    self = .unknown
+                }
+            } else {
+                self = .unknown
+            }
 
         case let (.enum("color"), color):
             if #available(iOS 26, tvOS 26, *) {
@@ -165,6 +180,42 @@ extension DisplayList.Content.Value: Reflection {
         default:
             self = .unknown
         }
+    }
+
+    private static func extractTextFromGlyphs(_ displayListContents: Any) throws -> String {
+        let contentString = String(reflecting: displayListContents)
+
+        // 1. Extract glyphs content using regex
+        guard let regex = try? NSRegularExpression(pattern: #"\(glyphs \"([^\"]+)\"\)"#),
+              let match = regex.firstMatch(in: contentString, range: NSRange(contentString.startIndex..., in: contentString)),
+              let glyphsRange = Range(match.range(at: 1), in: contentString)
+        else {
+            throw InternalError(description: "Failed to extract glyphs from display list contents")
+        }
+
+        let glyphs = String(contentString[glyphsRange])
+
+        // 2. Convert escape sequences like \C into actual characters
+        // Split by backslash and space, then take the first character of each part
+        let components = glyphs.components(separatedBy: "\\").filter { !$0.isEmpty }
+        let reconstructed = components.compactMap { part -> String? in
+            let trimmed = part.trimmingCharacters(in: .whitespaces)
+            guard !trimmed.isEmpty else {
+                return nil
+            }
+
+            if trimmed == "space" {
+                return " "
+            }
+
+            guard let first = trimmed.first else {
+                return nil
+            }
+            return String(first)
+        }
+        .joined()
+
+        return reconstructed
     }
 }
 
@@ -195,4 +246,5 @@ extension DisplayList.Item.Value: Reflection {
         }
     }
 }
+
 #endif
--- a/DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/DisplayListReflectionTests.swift
+++ b/DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/DisplayListReflectionTests.swift
@@ -58,7 +58,7 @@ class DisplayListReflectionTests: XCTestCase {
     func testDisplayList_withShapeContent() throws {
         let color = Color._Resolved.mockRandom()
         let path: SwiftUI.Path = [
-            SwiftUI.Path { $0.move(to: .zero); $0.addLine(to: CGPoint(x: 10, y: 10)) },
+            SwiftUI.Path { $0.move(to: CGPoint.zero); $0.addLine(to: CGPoint(x: 10, y: 10)) },
             SwiftUI.Path(),
             SwiftUI.Path { $0.addRect(CGRect(x: 0, y: 0, width: 50, height: 50)) }
         ].randomElement()!
--- a/DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/DisplayListReflectionTests.swift
+++ b/DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/DisplayListReflectionTests.swift
@@ -164,5 +164,38 @@ class DisplayListReflectionTests: XCTestCase {
             XCTFail("DisplayList should contain a content value.")
         }
     }
+
+    // MARK: Drawing Content (iOS 26+ Toolbar Items)
+    func testDisplayList_withDrawingContent_itHandlesToolbarItemsOniOS26() throws {
+        guard #available(iOS 26, tvOS 26, *) else {
+            return
+        }
+
+        // Given
+        let toolbarContent = DisplayList.Content(
+            seed: .init(value: .mockRandom()),
+            value: .toolbarItem("Save & Continue")
+        )
+        let toolbarDisplayList = DisplayList(items: [
+            DisplayList.Item(
+                identity: .init(value: .mockRandom()),
+                frame: .mockRandom(),
+                value: .content(toolbarContent)
+            )
+        ])
+
+        // When
+        let toolbarReflector = Reflector(subject: toolbarDisplayList, telemetry: NOPTelemetry())
+        let toolbarReflectedList = try DisplayList(from: toolbarReflector)
+
+        // Then
+        XCTAssertEqual(toolbarReflectedList.items.count, 1)
+        if case let .content(content) = toolbarReflectedList.items[0].value,
+           case let .toolbarItem(extractedText) = content.value {
+            XCTAssertEqual(extractedText, "Save & Continue 📱")
+        } else {
+            XCTFail("Toolbar items should extract text correctly")
+        }
+    }
 }
 #endif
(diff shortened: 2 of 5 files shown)


DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/DisplayList+Reflection.swift
```diff
@@ -152,8 +152,23 @@ extension DisplayList.Content.Value: Reflection {
...
+            // In iOS 26, toolbar items are now rendered as drawings which are not currently supported.
+            // As a temporary fix, we extract the text from the drawing and display it as text content.
+            if #available(iOS 26, tvOS 26, *) {
+                do {
+                    guard let drawingContent = drawing else {
+                        self = .unknown
+                        return
+                    }
+                    let extractedText = try Self.extractTextFromGlyphs(drawingContent)
+                    self = .toolbarItem(extractedText)
```
mariedm: I imagine this could work not only for toolbars but any drawing. What about renaming to something more generic, like `glyph` or `drawingWithText`?

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/DisplayList+Reflection.swift
```diff
@@ -152,8 +152,23 @@ extension DisplayList.Content.Value: Reflection {
         case let (.enum("image"), image):
             self = try .image(reflector.reflect(image))
 
-        case (.enum("drawing"), _):
-            self = .unknown
+        case let (.enum("drawing"), drawing):
+            // In iOS 26, toolbar items are now rendered as drawings which are not currently supported.
+            // As a temporary fix, we extract the text from the drawing and display it as text content.
+            if #available(iOS 26, tvOS 26, *) {
```
mariedm, line 158: Do we really need this check? I'm wondering if we could support drawings for all SR, not just iOS 26+. Have you tried this logic on earlier versions?

DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/SwiftUIWireframesBuilderTests.swift
```diff
@@ -52,5 +52,45 @@ class SwiftUIWireframesBuilderTests: XCTestCase {
...
+        let renderer = DisplayList.ViewUpdater(
+            viewCache: DisplayList.ViewUpdater.ViewCache(map: [:]),
+            lastList: DisplayList.Lazy(
+                DisplayList(items: [
+                    DisplayList.Item(
+                        identity: DisplayList.Identity(value: .mockRandom()),
+                        frame: CGRect(x: 0, y: 0, width: 100, height: 44),
+                        value: .content(DisplayList.Content(
+                            seed: DisplayList.Seed(value: .mockRandom()),
+                            value: .toolbarItem("Cancel")
```
mariedm: nit: you could create a variable for the text, which you can use here and later in your assertion. Same for `testDisplayList_withDrawingContent_itHandlesToolbarItemsOniOS26`.

DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/SwiftUIWireframesBuilderTests.swift
```diff
@@ -52,5 +52,45 @@ class SwiftUIWireframesBuilderTests: XCTestCase {
         let wireframe = try XCTUnwrap(wireframes.last?.placeholderWireframe)
         XCTAssertEqual(wireframe.label, "Unsupported SwiftUI component")
     }
+
+    // MARK: Drawing Content (iOS 26+ Toolbar Items)
+    func testDisplayListWithToolbarItem_itCreatesTextWireframeOniOS26() throws {
```
mariedm, line 57: nit: we could test more use cases, e.g., empty text, special characters, failing glyph extraction.

DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/DisplayListReflectionTests.swift
```diff
@@ -164,5 +164,38 @@ class DisplayListReflectionTests: XCTestCase {
...
+
+        // When
+        let toolbarReflector = Reflector(subject: toolbarDisplayList, telemetry: NOPTelemetry())
+        let toolbarReflectedList = try DisplayList(from: toolbarReflector)
+
+        // Then
+        XCTAssertEqual(toolbarReflectedList.items.count, 1)
+        if case let .content(content) = toolbarReflectedList.items[0].value,
+           case let .toolbarItem(extractedText) = content.value {
+            XCTAssertEqual(extractedText, "Save & Continue 📱")
```
mariedm: This is not the same text as the input, we have an additional emoji here 🤔

DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/DisplayListReflectionTests.swift
```diff
@@ -164,5 +164,38 @@ class DisplayListReflectionTests: XCTestCase {
             XCTFail("DisplayList should contain a content value.")
         }
     }
+
+    // MARK: Drawing Content (iOS 26+ Toolbar Items)
+    func testDisplayList_withDrawingContent_itHandlesToolbarItemsOniOS26() throws {
+        guard #available(iOS 26, tvOS 26, *) else {
```
mariedm: For all tests, it would be best to remove this check and test the output in both cases. Something like:
```
if #available(iOS 26, tvOS 26, *) {
    XCTAssertEqual(toolbarReflectedList.items.count, 1)
    // other assertions
} else {
    XCTAssertEqual(toolbarReflectedList.items.count, 0)
    // other assertions
}
```

mariedm (commented):
I left a few comments and questions, but overall it looks good! 🙌

DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/DisplayListReflectionTests.swift
mariedm: I don't think this test is useful. It creates a `toolbarItem` directly and verifies that it reflects back as `toolbarItem`, so it can be removed.

The tests in `SwiftUIWireframesBuilderTests` should be enough with different text inputs & edge cases.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/DisplayList+Reflection.swift
```diff
@@ -152,8 +152,23 @@ extension DisplayList.Content.Value: Reflection {
...
+                do {
+                    guard let drawingContent = drawing else {
+                        self = .unknown
+                        return
+                    }
+                    let extractedText = try Self.extractTextFromGlyphs(drawingContent)
```
maxep, line 164: change-request/ What's problematic is that we are performing the heavy `extractTextFromGlyphs` operation on every single drawing and at each snapshot on the main thread, this will have a negative impact on perfs.

As @mariedm commented, this is assuming that all `drawing` are `toolbarItem` which is not the case. IMO, we shouldn't try to fix Toolbar specifically but instead add general support to `drawing` since it seems to be more widely used on iOS 26. Which mean either adding support for vectorial images, or transform the `drawing` to image resource.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/DisplayList+Reflection.swift
maxep: Yes, the `DisplayList` is a mirror representation and we should keep aligning naming with what's being reflected:
```swift
case let (.enum("drawing"), drawing):
    self = .drawing(...)
```

maxep (changes requested):

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/DisplayList+Reflection.swift
Valpertui: >Which mean either adding support for vectorial images, or transform the drawing to image resource.  
  
Until vectorial images support is built fully (which the team is currently discussing), we do want to give customers an SDK version that is doing its best to solve the recording issues on iOS 26.  
Without this right now we do not capture Navigation bar buttons text anymore, which leads to very degraded replays.  
  
> or transform the drawing to image resource.   

Wouldn't this be even more performance intensive?

DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/DisplayListReflectionTests.swift
barboraplasovska: Deleted it !

DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/SwiftUIWireframesBuilderTests.swift
barboraplasovska: Done

DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/SwiftUIWireframesBuilderTests.swift
barboraplasovska: Done

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/DisplayList+Reflection.swift
maxep: I understand, but `drawing` is a generic content already present in <iOS26 and doing such string processing is to heavy for the main thread. If we don't think we could get vectorial images support before iOS 26 release, we could:
- Offload this string processing to the processing thread
- Transform the drawing to Image resource on the processing thread

>> or transform the drawing to image resource.
>
> Wouldn't this be even more performance intensive?

We do this on a background thread.

gonzalezreal (changes requested):
Hey @barboraplasovska,

Sorry for the late feedback. I had to spend some time reverse-engineering the drawing payload.

The payload is a tuple containing the following:

- An `RBMovedDisplayListContents` instance (`NSObject` subclass) containing the drawing commands. This type is either bridged or wraps a C++ object, so mirroring is not an option. It is amazing that you managed to extract the text glyphs by parsing its debug description, but I don't think that is the long-term solution for performance reasons (as suggested by @mariedm)
- A `CGPoint` value, which is most probably an offset.
- A `SwiftUI.RasterizationOptions`  value.

I did some Objective-C runtime spelunking on `RBMovedDisplayListContents` and found some interesting properties and methods:

- A `boundingRect` property.
- A `renderInContext:options:` method.

The Objective-C runtime lists other properties, like the drawing contents, that is, the commands you see in the `debugDescription`. But accessing that results in an exception.

I think we should abandon trying to extract the drawing commands from this type, and instead render the drawing into an image, like @maxep suggests. It should be feasible to create a bitmap `CGContext`, using the size in `boundingRect`, and render the drawing with the given rasterization options.

The `drawing` case is used for SF Symbols and, in some contexts, for text. So I don't expect any performance issues with this approach.

@Valpertui Taking this route does not invalidate the vector graphics track. I believe we still have other use cases for that. We just need to investigate further.

If you folks agree, I will create a JIRA ticket to implement the suggested approach. I'll be happy to assist @barboraplasovska.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/DisplayList+Reflection.swift
maxep: Sorry @barboraplasovska, I was wrong about this being performed on the main thread. We are actually reflecting lazily on the processing thread. That being said, @gonzalezreal 's proposal sounds more promising to improve the player fidelity overall, I think we should continue exploring that!

PR #2389, opened 2025-07-07
RUM-10723 Add 1ns view filtering to RUMViewEventsFilter

ncreated:
### What and why?

📦 This PR enhances session integrity by filtering out view events with minimal (placeholder) duration. This situation occurs when views receive no follow-up events after being started with a 1ns placeholder duration, representing a failed attempt to reach eventual consistency (where `view.time_spent` should be updated by subsequent events).

### How?

This work leverages the "event metadata" available in the Core interface, which gets written to batches alongside each event. We already use this metadata in RUM to filter redundant view updates from payloads.

The metadata for views is now extended with a `duration` field (optional for backwards compatibility):
```diff
struct Metadata: Codable {
    let id: String
    let documentVersion: Int64
+    /// Duration of the view in nanoseconds.
+    let duration: Int64?
}
```
Based on the `duration` check, views with 1ns `time_spent` are filtered out from the payload in `RUMViewEventsFilter` (used in `RUM.RequestBuilder`).

#### 💭 Alternative considered

I considered change in `RUMViewScope` to not write view updates if their duration is unsatisfactory.

The rationale for rejecting this approach:
- **Scope and complexity:** Implementing this would require significant changes across the codebase and test suite. Many tests assume that `startView()` immediately writes an initial event (with a 1ns duration), which aligns with our eventual consistency model—view durations are later updated as follow-up events arrive. This solution would force updates to dozens of tests that don’t simulate those subsequent events.
- **Design philosophy:** Since 1ns durations are not expected under normal circumstances (due to eventual consistency), filtering them using generic and centralized logic is a cleaner and more maintainable solution.

### Review checklist

### DIFF
--- a/DatadogRUM/Sources/Feature/RUMViewEventsFilter.swift
+++ b/DatadogRUM/Sources/Feature/RUMViewEventsFilter.swift
@@ -8,45 +8,80 @@ import Foundation
 import DatadogInternal
 
 internal struct RUMViewEventsFilter {
-    let decoder: JSONDecoder
+    /// Duration threshold for filtering out 1ns views that represent failed eventual consistency scenarios
+    private static let oneNanosecondDuration: Int64 = 1
 
-    init(decoder: JSONDecoder = JSONDecoder()) {
+    let decoder: JSONDecoder
+    let telemetry: Telemetry
+
+    init(telemetry: Telemetry, decoder: JSONDecoder = JSONDecoder()) {
+        self.telemetry = telemetry
         self.decoder = decoder
     }
 
+    private enum FilterResult {
+        case keep
+        case skipRedundant
+        case skipOneNs
+    }
+
+    /// Filters RUM view events to improve data quality and optimize payload size by removing:
+    /// - Redundant view events (duplicate view IDs)
+    /// - 1ns view events that represent failed eventual consistency scenarios
     func filter(events: [Event]) -> [Event] {
         var seen = Set<String>()
-        var skipped: [String: [Int64]] = [:]
+        var redundantEventsCount = 0
+        var oneNsEventsCount = 0
 
         // reversed is O(1) and no copy because it is view on the original array
         let filtered: [Event] = events.reversed().compactMap { event in
-            guard let metadata = event.metadata else {
-                // If there is no metadata, we can't filter it.
-                return event
-            }
-
-            guard let viewMetadata = try? decoder.decode(RUMViewEvent.Metadata.self, from: metadata) else {
-                // If we can't decode the metadata, we can't filter it.
-                return event
-            }
-
-            guard seen.contains(viewMetadata.id) == false else {
-                // If we've already seen this view, we can skip this
-                if skipped[viewMetadata.id] == nil {
-                    skipped[viewMetadata.id] = []
+            do {
+                let result = try filterEvent(event, seen: &seen)
+                switch result {
+                case .keep:
+                    return event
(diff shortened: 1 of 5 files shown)


DatadogRUM/Sources/Feature/RUMViewEventsFilter.swift
```diff
@@ -8,45 +8,80 @@ import Foundation
...
+        guard let metadata = event.metadata else {
+            // If there is no metadata, we can't filter it.
+            return .keep
         }
 
-        for (id, versions) in skipped {
-            DD.logger.debug("Skipping RUMViewEvent with id: \(id) and versions: \(versions.reversed().map(String.init).joined(separator: ", "))")
+        let viewMetadata = try decoder.decode(RUMViewEvent.Metadata.self, from: metadata)
+
+        if let duration = viewMetadata.duration, duration == RUMViewEventsFilter.oneNanosecondDuration {
```
mariedm: nit: should we reference where this 1ns placeholder is set? Right now we do a simple = operation with a local static value, but we might change the logic later.

DatadogRUM/Sources/Feature/RUMViewEventsFilter.swift
```diff
@@ -8,45 +8,80 @@ import Foundation
...
-            return event
+    private func filterEvent(_ event: Event, seen: inout Set<String>) throws -> FilterResult {
+        guard let metadata = event.metadata else {
+            // If there is no metadata, we can't filter it.
+            return .keep
         }
 
-        for (id, versions) in skipped {
-            DD.logger.debug("Skipping RUMViewEvent with id: \(id) and versions: \(versions.reversed().map(String.init).joined(separator: ", "))")
+        let viewMetadata = try decoder.decode(RUMViewEvent.Metadata.self, from: metadata)
```
mariedm, line 77: Why not throwing anymore in case of failed decoding?

ncreated: It's the opposite - now we throw 🙂 but previously we were not:
```diff
- guard let viewMetadata = try? decoder.decode(RUMViewEvent.Metadata.self, from: metadata) else {
-   // If we can't decode the metadata, we can't filter it.
-   return event
- }
```

With this PR, any failure in metadata decoding is also tracked as telemetry error:
```swift
telemetry.error("Failed to decode RUM view event metadata", error: error)
```

DatadogRUM/Sources/Feature/RUMViewEventsFilter.swift
ncreated: Fair point 👍. I’ve added clarity by introducing a constant in `RUMViewScope` where the `1ns` value is defined. The filter now references this constant directly. I also added a few more code comments to improve readability.

DatadogRUM/Sources/Feature/RUMViewEventsFilter.swift
mariedm: My bad, read too fast 🙃

mariedm (dismissed):

maxep (dismissed):
Looks good!

Does this mean that the last view of a user session won't be tracked if there is no subsequent events?

ncreated:
> Does this mean that the last view of a user session won't be tracked if there is no subsequent events?

Fair point - I updated the implementation to only skip 1ns views if they are **initial** event in the session. Last views in user sessions will be tracked as usual.

PR #2391, opened 2025-07-11
RUM-10768: Render SwiftUI drawings into images

gonzalezreal:
### What and why?

This PR implements rasterization of SwiftUI drawings (primarily SF symbols) into images for Session Replay capture, replacing previously unsupported drawing content with actual rendered images.

### How?

1. Detects SwiftUI drawings in the display list during view tree traversal
2. Rasterizes drawings into bitmap images using Core Graphics when processing wireframes
3. Caches generated images for performance optimization

### Review checklist


DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/DisplayList+Reflection.swift
```diff
@@ -152,8 +152,12 @@ extension DisplayList.Content.Value: Reflection {
         case let (.enum("image"), image):
             self = try .image(reflector.reflect(image))
 
-        case (.enum("drawing"), _):
-            self = .unknown
+        case let (.enum("drawing"), (contents, origin, _) as (NSObject, CGPoint, Any)):
```
gonzalezreal, line 158: See [this comment](https://github.com/DataDog/dd-sdk-ios/pull/2386#pullrequestreview-3000710074) on #2386

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/Image+Drawing.swift
```diff
@@ -0,0 +1,128 @@
...
+            data: nil,
+            width: width,
+            height: height,
+            bitsPerComponent: 8,
+            bytesPerRow: width,
+            space: CGColorSpaceCreateDeviceGray(),
+            bitmapInfo: CGImageAlphaInfo.alphaOnly.rawValue
+        ) else {
+            return nil
+        }
```
gonzalezreal: We could also rasterize drawings into full color bitmaps, allowing [multi-color or palette symbols](https://developer.apple.com/documentation/swiftui/symbolrenderingmode). The problem is that, in some instances, `RBMovedDisplayListContents` data uses `white` color. For instance, when tinting symbols with an accent color, or just arbitrarily as I experienced during my tests.

I couldn't find a way to determine which bitmap format to choose, as I don't have any information about how the symbol is rendered on the screen.

Any ideas are welcome!

mariedm: If I understand correctly, we don't have access to the tint color from the SwiftUI drawing data itself?

If that's the case, a workaround could be to use a static, pre-determined color for all drawings. While it's not ideal (from a visual accuracy point of view), it would ensure _most_ drawings are visible (depending on the background color for example).

gonzalezreal: Exactly. This is what we are doing now by using an alpha-only bitmap. I wish we could find a better solution.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/Image+Drawing.swift
```diff
@@ -0,0 +1,128 @@
...
+
+        // Compute image size
+
+        let width = Int((bounds.width + 1.5) * scale)
+        let height = Int((bounds.height + 1.5) * scale)
+
+        // Create a grayscale, alpha-only bitmap context
+
+        guard let bitmapContext = CGContext(
```
mariedm: I think alpha-only grayscale is a good choice, at least for now. We can always iterate later if needed. From a performance and memory usage point of view, it's actually better!

Some suggestions to improve performance even more, following what we have for UIKit:
- [Add a size limit to images](https://github.com/DataDog/dd-sdk-ios/blob/f3136934e60324c355e5125079766ce6a7fc1647/DatadogSessionReplay/Sources/Recorder/Utilities/UIImage%2BSessionReplay.swift#L41) (max 100px here)
- [Lower the interpolation quality setting](https://github.com/DataDog/dd-sdk-ios/blob/f3136934e60324c355e5125079766ce6a7fc1647/DatadogSessionReplay/Sources/Recorder/Utilities/UIImage%2BSessionReplay.swift#L68)

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/Image+Drawing.swift
mariedm: Note: in UIKit, we [apply the tint color afterward](https://github.com/DataDog/dd-sdk-ios/blob/f3136934e60324c355e5125079766ce6a7fc1647/DatadogSessionReplay/Sources/Recorder/Utilities/UIImage%2BSessionReplay.swift#L90). So we could also have a two-step process for SwiftUI: 1) rasterize, 2) apply tint color.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/Image+Drawing.swift
```diff
@@ -0,0 +1,128 @@
...
+
+        self.init(contents: .cgImage(cgImage), scale: rasterizationScale, orientation: .up)
+    }
+}
+
+@available(iOS 13.0, tvOS 13.0, *)
+extension GraphicsImage {
+    fileprivate class DrawingRasterizer: @unchecked Sendable {
+        private let cache = NSCache<NSNumber, CGImage>()
+        private let lock = NSRecursiveLock()
```
maxep: question/ Do we need a lock? I guess this will always execute on the processing queue.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/DisplayList+Reflection.swift
```diff
@@ -152,8 +152,12 @@ extension DisplayList.Content.Value: Reflection {
...
+            if let image = GraphicsImage(rasterizing: contents, origin: origin) {
+                self = .image(image)
```
maxep: suggestion/ The rendering will occur on the processing thread as we are applying a lazy reflection. But this operation is implicit and the fact that we map `.enum("drawing")` to a `.image` content kind of break the reflection pattern.

I would suggest to map this to a new `self = .drawing(...)` case and do the rendering in the `SwiftUIWireframesBuilder` when mapping the `DisplayList.Content` to a `SRWireframe.image`. This way we are explicitly rendering the `.drawing` content to an `.image` wireframe on the processing thread. WDYT?

maxep (dismissed):
Amazing work! This is a great improvement, well done!

I left one suggestion that I think would clarify the process, LMKWYT!

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/Image+Drawing.swift
gonzalezreal: Good point. I will re-consider it.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/DisplayList+Reflection.swift
gonzalezreal: Yes, that makes sense. I will move the rasterization where you suggest. Thanks!

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/DisplayList+Reflection.swift
```diff
@@ -56,6 +56,9 @@ extension DisplayList.Effect: Reflection {
         case let (.enum("clip"), tuple as (SwiftUI.Path, SwiftUI.FillStyle, Any)):
             self = .clip(tuple.0, tuple.1)
 
+        case let(.enum("filter"), filter):
+            self = try .filter(reflector.reflect(filter))
+
```
gonzalezreal: For tinting SwiftUI applies a `colorMultiply` effect to the drawings.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/ImageRenderer.swift
```diff
@@ -0,0 +1,104 @@
...
+ */
+
+#if os(iOS)
+
+import CoreGraphics
+import Foundation
+
+@available(iOS 13.0, tvOS 13.0, *)
+internal final class ImageRenderer {
+    private class Key: NSObject {
```
gonzalezreal, line 14: `NSCache` requires `NSObject` subclass as the key type.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/SwiftUIWireframesBuilder.swift
```diff
@@ -171,7 +179,36 @@ internal struct SwiftUIWireframesBuilder: NodeWireframesBuilder {
                     label: "Unsupported image type"
                 )
             }
-
+        case let .drawing(contents, origin):
+            if case .maskAll = self.imagePrivacyLevel {
```
gonzalezreal, line 183: We treat drawings as bundled images.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/Image+Drawing.swift
gonzalezreal: Right! This is how I solved it once I could extract the drawing tint color from the parent effect.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/SwiftUIWireframesBuilder.swift
```diff
@@ -70,6 +74,10 @@ internal struct SwiftUIWireframesBuilder: NodeWireframesBuilder {
         case let .clip(path, _):
             let clip = context.convert(frame: path.boundingRect)
             context.clip = context.clip.intersection(clip)
+        case let .filter(filter):
+            if case .colorMultiply(let color) = filter {
+                context.tintColor = color
+            }
```
maxep: nit/ Would that work?
```suggestion
        case let .filter(.colorMultiply(color)):
```

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/ImageRenderer.swift
```diff
@@ -0,0 +1,104 @@
...
+
+    private func makeImage(with contents: DisplayListContents, origin: CGPoint, scale: CGFloat) -> CGImage? {
+        guard let bounds = contents.bounds else {
+            return nil
+        }
+
+        // Compute image size
+
+        let width = Int((bounds.width + 1.5) * scale)
+        let height = Int((bounds.height + 1.5) * scale)
```
maxep: question/ Why `* 1.5`? Maybe we can add an explanation as comment 👍

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/SwiftUIWireframesBuilder.swift
maxep, line 1: suggestion/ The `func contentWireframe(item:,content:,context:) -> SRWireframe?` is growing in complexity, maybe we can start splitting it, WDYT?

maxep (approved):
Amazing! Well done with the `filter` effect and the renderer cache, it reads super well!

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/SwiftUIWireframesBuilder.swift
mariedm: It makes sense, but it might be worth adding this as a comment in the code.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/ImageRenderer.swift
mariedm: Curious as well. It might also make sense to make a constant out of it?

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/ImageRenderer.swift
```diff
@@ -0,0 +1,104 @@
...
+
+        // Create a bitmap context
```
mariedm: Do we want to limit the size here?

DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/SwiftUIWireframesBuilderTests.swift
mariedm, line 1: We could add a test to `UIHostingViewRecorderTests` for this new drawing use case, wdyt?

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/ImageRenderer.swift
mariedm: (previous [comment](https://github.com/DataDog/dd-sdk-ios/pull/2391/files/8e9809e697fa0973fd1c5eb5ff85cc293226759f#r2200626299) suggesting having a max size + lower quality)

gonzalezreal: IMO it is not necessary, even in those cases where users render a big system image. These are monochrome vector graphics with transparency that are going to be very effectively compressed to PNG. What limit do you have in mind?

mariedm: I was thinking of complex drawings with many paths for example, but it's unlikely. As for a size limit, I would have aligned with what we have on UIKit (100px). But probably not necessary.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/ImageRenderer.swift
gonzalezreal: This is adding 0.5 pixel padding (0.5 + 0.5) plus another 0.5 for round-half-up when casting to `Int`. I will add a comment with the explanation.

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/SwiftUIWireframesBuilder.swift
gonzalezreal: I agree! I would prefer to do this in a separate PR if that is OK for you. LMYT

DatadogSessionReplay/Tests/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/SwiftUI/SwiftUIWireframesBuilderTests.swift
gonzalezreal: I added a snapshot test that covers everything

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UIHostingViewRecorder.swift
```diff
@@ -21,6 +21,8 @@ internal class UIHostingViewRecorder: NodeRecorder {
     var semanticsOverride: (UIView, ViewAttributes) -> NodeSemantics?
     var textObfuscator: (ViewTreeRecordingContext, ViewAttributes) -> TextObfuscating
 
+    private let imageRenderer = ImageRenderer()
```
mariedm, line 24: Do we need to store this var or can we just init it when creating `SwiftUIWireframesBuilder`?

mariedm (approved):
Great work! 🙌 The `ImageRepresentable` abstraction will help with future support for other image types. Just a minor nit, otherwise, looks good to go! 🚀

DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeRecorders/UIHostingViewRecorder.swift
gonzalezreal: We need to store it to leverage the cache between view tree captures.

PR #2395, opened 2025-07-15
RUM-10438 Improve backtrace collection and error messages

maxep:
### What and why?

We are seeing telemetry errors implying that `PLCrashReporter` [fails to decode the protobuf encoded stacktrace](https://github.com/microsoft/plcrashreporter/blob/master/Source/PLCrashReport.m#L333) while generating App Hang backtraces.

App Hang is based on `PLCrashReporter`'s 'Live Report' to collect backtraces, Live Report feature will write a report on-disk like during a crash using protobuf serialization and then read the file to decode it. We are reaching the decoding point of failure, meaning that data was corrupted.

While we are not able to reproduce this issue, we have identified a thread-safety issue: Our Crash Context is [injected in the Live Report](https://github.com/microsoft/plcrashreporter/blob/master/Source/PLCrashReporter.m#L751-L754) for encoding but the `customData` property is [nonatomic](https://github.com/microsoft/plcrashreporter/blob/master/Source/PLCrashReporter.h#L127). Context propagation and App Hang occurs on different threads leading to a race condition on `customData`.

### How?

Introduces a secondary `PLCrashReporter` instance for backtrace collection. This secondary instance will only be used for Live Report and will not include the crash-context.

### Review checklist

### DIFF
--- a/DatadogCrashReporting/Sources/PLCrashReporterIntegration/PLCrashReporterIntegration.swift
+++ b/DatadogCrashReporting/Sources/PLCrashReporterIntegration/PLCrashReporterIntegration.swift
@@ -44,11 +44,21 @@ internal extension PLCrashReporterConfig {
 
 internal final class PLCrashReporterIntegration: ThirdPartyCrashReporter {
     private let crashReporter: PLCrashReporter
+    private let backtraceReporter: PLCrashReporter
     private let builder = DDCrashReportBuilder()
 
     init() throws {
-        self.crashReporter = try PLCrashReporter(configuration: .ddConfiguration())
+        let configuration: PLCrashReporterConfig = try .ddConfiguration()
+        self.crashReporter = PLCrashReporter(configuration: configuration)
         try crashReporter.enableAndReturnError()
+
+        // Secondary instance for collecting Live Report for backtraces to prevent
+        // race condition while accessing customData: PLCrashReporter's customData
+        // is not thread-safe and is actually not needed for backtraces.
+        //
+        // This secondary instance doesn't need to and should not be enabled as it
+        // will conflict with the primary one.
+        self.backtraceReporter = PLCrashReporter(configuration: configuration)
     }
 
     func hasPendingCrashReport() -> Bool {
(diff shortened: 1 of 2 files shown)


DatadogCrashReporting/Sources/PLCrashReporterIntegration/PLCrashReporterIntegration.swift
```diff
@@ -44,11 +44,21 @@ internal extension PLCrashReporterConfig {
 
 internal final class PLCrashReporterIntegration: ThirdPartyCrashReporter {
     private let crashReporter: PLCrashReporter
+    private let backtraceReporter: PLCrashReporter
     private let builder = DDCrashReportBuilder()
 
     init() throws {
-        self.crashReporter = try PLCrashReporter(configuration: .ddConfiguration())
+        let configuration: PLCrashReporterConfig = try .ddConfiguration()
```
ncreated: Suggestion/ I think there are parts of the configuration that we may want to differentiate for backtrace reporter, for example the path to write file (to not conflict with regular crashes) or disabling the uncaught exception handler in this instance (that if enabled in second instance, may cause troubles I believe).

ncreated (commented):
I like this reasoning and it makes total sense 👍. I left a strong suggestion on the config.

DatadogCrashReporting/Sources/PLCrashReporterIntegration/PLCrashReporterIntegration.swift
maxep: I think we are good in that regard:
- PLCR will write the Live Report to a dedicated [`live_crash_report.XXXXXX`](https://github.com/microsoft/plcrashreporter/blob/master/Source/PLCrashReporter.m#L736C87-L736C111) file in the `tmp` directory. The base path is not used for Live Report.
- The `signalHandlerType` and `shouldRegisterUncaughtExceptionHandler` are applied when calling `enableCrashReporter` only, which [cannot be called twice](https://github.com/microsoft/plcrashreporter/blob/master/Source/PLCrashReporter.m#L571-L586).

That being said, I think it is ambiguous and I should be more explicit with a dedicated config 👍

ncreated: Ah indeed, both have no practical effect 👌. I agree, the ambiguity is there but it's not blocking definitely.

ncreated (dismissed):

ncreated (approved):
🎯

PR #2396, opened 2025-07-15
RUM-10572: Fix timestamp discrepancies in web-view session replay recordings

gonzalezreal:
### What and why?

This PR fixes a timing discrepancy that occurs between web-view RUM events and Session Replay records when NTP synchronization updates the server time offset after a web-view RUM view is created.

#### The Problem
  - `WebViewEventReceiver` caches `context.serverTimeOffset` when the first web-view event arrives
  - `WebViewRecordReceiver` uses `rumContext.viewServerTimeOffset` which is frozen at native RUM view creation time
  - If an NTP sync occurs between these two timing points, the receivers would apply different offsets to events from the same web-view, creating timing gaps between RUM events and Session Replay records

### How?

Both receivers now use the same cached offset per web-view ID, ensuring consistent timestamps across RUM events and Session Replay records for the same web-view.

- `WebViewEventReceiver` caches offsets in `RUMWebViewContext`
- `WebViewRecordReceiver` uses cached web-view server time offsets.

### Review checklist

### DIFF
--- a/DatadogInternal/Sources/Models/RUM/RUMCoreContext.swift
+++ b/DatadogInternal/Sources/Models/RUM/RUMCoreContext.swift
@@ -20,8 +20,6 @@ public struct RUMCoreContext: AdditionalContext, Equatable {
     public let userActionID: String?
     /// Current view related server time offset
     public let viewServerTimeOffset: TimeInterval?
-    /// Latest web view server time offsets
-    private var webViewServerTimeOffsets: [String: TimeInterval]
 
     /// Creates a RUM context.
     ///
@@ -36,22 +34,12 @@ public struct RUMCoreContext: AdditionalContext, Equatable {
         sessionID: String,
         viewID: String? = nil,
         userActionID: String? = nil,
-        viewServerTimeOffset: TimeInterval? = nil,
-        webViewServerTimeOffsets: [String: TimeInterval] = [:]
+        viewServerTimeOffset: TimeInterval? = nil
     ) {
         self.applicationID = applicationID
         self.sessionID = sessionID
         self.viewID = viewID
         self.userActionID = userActionID
         self.viewServerTimeOffset = viewServerTimeOffset
-        self.webViewServerTimeOffsets = webViewServerTimeOffsets
-    }
-
-    public func serverTimeOffset(forWebView id: String) -> TimeInterval? {
-        webViewServerTimeOffsets[id]
-    }
-
-    public mutating func setServerTimeOffset(_ offset: TimeInterval, forWebView id: String) {
-        webViewServerTimeOffsets[id] = offset
     }
 }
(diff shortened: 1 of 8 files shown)


DatadogRUM/Sources/Integrations/WebViewEventReceiver.swift
```diff
@@ -77,7 +77,19 @@ internal final class WebViewEventReceiver: FeatureMessageReceiver {
...
                let view = event["view"] as? JSON,
                let id = view["id"] as? String {
-                let correctedDate = Int64(date) + self.offset(forView: id, context: context)
+                let offsetMilliseconds: Int64
+
+                if let offset = rum.serverTimeOffset(forWebView: id) {
+                    offsetMilliseconds = offset.toInt64Milliseconds
+                } else {
+                    let offset = context.serverTimeOffset
+                    rum.setServerTimeOffset(offset, forWebView: id)
```
maxep: change-request/ `RUMCoreContext` being a struct, this will set the offset locally but won't be applied to the core-context. I think we should create a `class ViewOffsetCache` and use it as a property of `RUMCoreContext`.

DatadogInternal/Sources/Models/RUM/RUMCoreContext.swift
```diff
@@ -34,12 +36,22 @@ public struct RUMCoreContext: AdditionalContext, Equatable {
...
     ) {
         self.applicationID = applicationID
         self.sessionID = sessionID
         self.viewID = viewID
         self.userActionID = userActionID
         self.viewServerTimeOffset = viewServerTimeOffset
+        self.webViewServerTimeOffsets = webViewServerTimeOffsets
+    }
+
+    public func serverTimeOffset(forWebView id: String) -> TimeInterval? {
```
maxep: suggestion/ The offset is associated to a RUM view, not to the webview :) I would rename this to 
```suggestion
    public func serverTimeOffset(forView id: String) -> TimeInterval? {
```

maxep (changes requested):
Looks great, nice resolution 👍 

Tho there is the problem of `RUMCoreContext` being a struct, we need a way to update the cache so both RUM and SR can access to same values

DatadogRUM/Sources/Integrations/WebViewEventReceiver.swift
gonzalezreal: But the core context is propagated in the next line:

```swift
self.featureScope.set(context: rum)
```

So the cached values are propagated. Is there an issue with that?

gonzalezreal: After looking into this, I guess your concern is about the RUM `Monitor` creating new values of the `RUMCoreContext` without preserving the cached time offsets, right?

gonzalezreal: What do you think about getting the current `RUMCoreContext` value before the `Monitor` creates a new value, and pass the cached time offsets?

maxep: I probably read too quickly and missed that next line 🤦  But indeed, the RUM monitor will override the cache and that's blocking. Maybe the `ViewCacheOffest` could be extracted from the `RUMCoreContext` and be set to the context as an independent additional value.

maxep (approved):
Well done 🎉

PR #2405, opened 2025-07-24
RUM-10501: Improve TNS calculation

simaoseica-dd:
### What and why?

The TNS (Time to Network Settled) measurement can be inaccurate when the instrumentation also captures `URLSessionTaskMetrics`. This is particularly common in hybrid mobile applications.

### How?

- TNS is primarily computed using the network call metrics wrapped in the `ResourceMetrics`, especially the `fetch` phase’s start and end timestamps.
- Additionally, TNS is still computed even if the view experienced an inactive app state during loading.
This state, where the app is in the foreground but not receiving user input (e.g., during an incoming phone call, Notification Center pulled down, etc.), should not delay or invalidate the completion of the view’s initial network calls.

### Review checklist


maxep (approved):
LGTM 👍

PR #2408, opened 2025-07-28
RUM-5993 Refactor dynamic property handling for additionalProperties

maxep:
> [!IMPORTANT]
> This PR changes generated RUM model which is `public` and exposed via the event mappers.
> This breaking change should be considered for v3.

### What and why?

There is a discrepancy in the model generation between iOS and Android. If we have the following [definition](https://github.com/DataDog/rum-events-format/blob/master/schemas/rum/view-schema.json#L177-L186):

```json
"custom_timings": {
  "type": "object",
  "description": "User custom timings of the view. As timing name is used as facet path, it must contain only letters, digits, or the characters - _ . @ $",
  "additionalProperties": {
    "type": "integer",
    "minimum": 0,
    "readOnly": true
  },
  "readOnly": true
}
```

In that case iOS generator prefers to omit dedicated type creation and simply does public let [`customTimings: [String: Int64]?`](https://github.com/DataDog/dd-sdk-ios/blob/d840edcd4ba3ef3493c0b04241e01e2f52547f53/DatadogRUM/Sources/DataModels/RUMDataModels.swift#L2345C9-L2345C51)). On Android, they create a dedicated class:

```kotlin
public data class CustomTimings(
    public val additionalProperties: MutableMap<String, Long> = mutableMapOf(),
)
```

The downside of such inlining is that if a property is added to the object definition, in case of iOS model approach it won’t be backward compatible change (because now we need a dedicated model).

**We need to align and allow background compatible schema evolution.**

### How?

Dynamic properties with `additionalProperties` are now always represented as nested structs with internal dictionaries, regardless of the value type. This change updates the code generation logic, Swift and ObjC interop, and related tests to ensure consistent handling and encoding/decoding of dynamic properties, including for non-'any' types. Also updates generated models to match the latest RUM events schema.

### Review checklist

### DIFF
--- a/tools/rum-models-generator/Sources/CodeGeneration/Generate/Transformers/Swift/JSONToSwiftTypeTransformer.swift
+++ b/tools/rum-models-generator/Sources/CodeGeneration/Generate/Transformers/Swift/JSONToSwiftTypeTransformer.swift
@@ -117,44 +117,36 @@ internal class JSONToSwiftTypeTransformer {
     }
 
     private func transformJSONObject(_ jsonObject: JSONObject) throws -> SwiftType {
-        if let additionalProperties = jsonObject.additionalProperties {
-            if additionalProperties.type == .any {
-                // RUMM-1401: if schema declares `additionalProperties: true` or `additionalProperties: {type: object, ...}`
-                // we model it as a `struct` with nested `<public|public internal(set)><var> <structName>Info: [String: Codable]`
-                // dictionary. In generated encoding code, this dictionary is erased but its keys and values are used as dynamic
-                // properties encoded in JSON.
-                let additionalPropertyName = jsonObject.name + "Info"
-                // RUMM-1420: we noticed that `additionalProperties` is used for custom user attributes which need to be
-                // sanitized by the SDK, hence it's very practical for us to generate `.mutableInternally` modifier for those.
+        var `struct` = try transformJSONObjectToStruct(jsonObject)
 
-                // RUM-5992: we noticed that `additionalProperties` is used for custom user attributes which needs to be
-                // sanitized by the SDK or by user-defined mappers, hence we apply `.mutable` modifier for those.
-                let mutability: SwiftStruct.Property.Mutability = .mutable
-                var `struct` = try transformJSONObjectToStruct(jsonObject)
-                `struct`.properties.append(
-                    SwiftStruct.Property(
-                        name: additionalPropertyName,
-                        comment: additionalProperties.comment,
-                        type: SwiftDictionary(
-                            value: SwiftEncodable()
-                        ),
-                        isOptional: false,
-                        mutability: mutability,
-                        defaultValue: nil,
-                        codingKey: .dynamic
-                    )
+        if let additionalProperties = jsonObject.additionalProperties {
+            // RUMM-1401: if schema declares `additionalProperties: true` or `additionalProperties: {type: object, ...}`
+            // we model it as a `struct` with nested `<public|public internal(set)><var> <structName>Info: [String: Codable]`
+            // dictionary. In generated encoding code, this dictionary is erased but its keys and values are used as dynamic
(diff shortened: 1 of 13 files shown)


tools/rum-models-generator/Sources/CodeGeneration/Generate/Transformers/Swift/JSONToSwiftTypeTransformer.swift
```diff
@@ -117,44 +117,39 @@ internal class JSONToSwiftTypeTransformer {
...
-                // RUMM-1401: if schema declares `additionalProperties: {type: string | bool | integer | double, ...}}`
-                // we model it as dictionary property.
-                return SwiftDictionary(
-                    value: try transformJSONtoPrimitive(additionalProperties.type)
+            // RUMM-1401: if schema declares `additionalProperties: true` or `additionalProperties: {type: object, ...}`
+            // we model it as a `struct` with nested `<public|public internal(set)><var> <structName>Info: [String: Codable]`
+            // dictionary. In generated encoding code, this dictionary is erased but its keys and values are used as dynamic
+            // properties encoded in JSON.
+            let additionalPropertyName = jsonObject.name + "Info"
+            // RUMM-1420: we noticed that `additionalProperties` is used for custom user attributes which need to be
```
simaoseica-dd: Is this comment still relevant? Seems `.mutableInternally` is not used

tools/rum-models-generator/Sources/CodeGeneration/Print/SwiftPrinter.swift
```diff
@@ -237,14 +237,21 @@ public class SwiftPrinter: BasePrinter, CodePrinter {
...
+                    value = "AnyEncodable($1)"
+                default:
+                    value = "$1"
+                }
+
                 writeLine("try \(dynamicProperty.backtickName).forEach {") // dynamic properties are dictionaries
                 indentRight()
-                    writeLine("let key = DynamicCodingKey($0)") // dictionary key is used as coding key
-                    writeLine("try dynamicContainer.encode(AnyEncodable($1), forKey: key)") // encode value
+                    writeLine("try dynamicContainer.encode(\(value), forKey: DynamicCodingKey($0))") // encode value, dictionary key is used as coding ke
```
simaoseica-dd:
```suggestion
                    writeLine("try dynamicContainer.encode(\(value), forKey: DynamicCodingKey($0))") // encode `value` (dictionary `key` is used as coding key)
```

tools/rum-models-generator/Sources/CodeGeneration/Print/SwiftPrinter.swift
```diff
@@ -277,33 +284,41 @@ public class SwiftPrinter: BasePrinter, CodePrinter {
...
             if let dynamicProperty = dynamicallyCodedProperty {
-                writeLine("// Decode other properties into [String: Codable] dictionary:")
+                guard let dict = dynamicProperty.type as? SwiftDictionary else {
+                    throw Exception.unimplemented("Printing dynamic property of type `\(dynamicProperty.type)` is not supported ")
+                }
+
+                let valueType: String
+                switch dict.value {
+                case is SwiftEncodable, is SwiftCodable:
+                    valueType = "AnyCodable"
```
simaoseica-dd, line 294: can it be `AnyEncodable ` only? If not, it is better to have 2 cases.

simaoseica-dd (approved):
Well done 🎖️

tools/rum-models-generator/Sources/CodeGeneration/Print/SwiftPrinter.swift
maxep: No, because this is to generate the `init(from decoder: Decoder)`, so both cases needs to be decoded as `AnyCodable`.

simaoseica-dd (approved):
🎖️

PR #2410, opened 2025-07-29
RUM-10627 Accessibility in View Updates

barboraplasovska:
### What and why?

This PR changes how accessibility attributes are sent in RUM view events. Previously, accessibility data was sent as custom attributes. Now we send all accessibility attributes in the first view update, and subsequent view updates only include the attributes that have changed.

### How?

The `AccessibilityReader` now tracks which accessibility attributes have changed and only includes those in subsequent view updates. The `RUMViewScope` calls the reader to get accessibility data and clears the changed attributes after each view update. When a new view starts, the tracking is reset so the new view sends all its accessibility attributes on the first update.

To enable this functionality, a new `collectAccessibility` flag was added to the `RUMConfiguration`. It is `false` by default and must be explicitly enabled by the user to start collecting accessibility data.

The `ViewEventsFilter` was also updated to retain all view updates that include accessibility changes. Previously, only the last view update in a batch was kept, which could lead to loss of accessibility data as they were not included in all the updates. This change ensures no updates with accessibility attributes are dropped while we wait for a dedicated performance event type to handle them more efficiently.

### Review checklist


DatadogRUM/Sources/DataModels/RUMDataModelsMapping.swift
```diff
@@ -67,21 +67,64 @@ internal extension RUMViewEvent {
     struct Metadata: Codable {
         let id: String
         let documentVersion: Int64
+        let hasAccessibility: Bool
```
maxep: request/ This should be optional to not break backward compatibility. Otherwise, batches written with a previous version of the SDK will fail to decode.

DatadogRUM/Sources/RUMContext/AccessibilityReader.swift
```diff
@@ -9,19 +9,46 @@ import Foundation
...
 internal final class AccessibilityReader: AccessibilityReading {
     @ReadWriteLock
-    private(set) var state: Accessibility
+    private(set) var state: AccessibilityInfo
 
     private let notificationCenter: NotificationCenter
     private var observers: [NSObjectProtocol] = []
 
+    /// Track the initial accessibility state for comparison
+    private var initialState: AccessibilityInfo?
```
maxep: question/ I'm not sure to understand why you need an initial state, because you already have the `state` which is set with an initial value at `init`.

DatadogRUM/Sources/RUMContext/AccessibilityReader.swift
```diff
@@ -33,7 +60,17 @@ internal final class AccessibilityReader: AccessibilityReading {
...
-            self.state = self.currentState
+            let newState = self.currentState
+
+            // If this is the first update, store the initial state
+            if self.initialState == nil {
+                self.initialState = newState
+            } else {
+                // Track changes by comparing with the previous state
+                self.trackChanges(from: self.state, to: newState)
+            }
```
maxep: suggestion/ If we remove the `initialState`, this can be:
```suggestion
            self.trackChanges(from: self.state, to: newState)
```

DatadogRUM/Sources/RUMContext/AccessibilityReader.swift
```diff
@@ -9,19 +9,46 @@ import Foundation
...
 
     private let notificationCenter: NotificationCenter
     private var observers: [NSObjectProtocol] = []
 
+    /// Track the initial accessibility state for comparison
+    private var initialState: AccessibilityInfo?
+    /// Track if this is the first view update
+    private var isFirstViewUpdate = true
+    /// Track changed accessibility attributes
+    private var changedAttributes: [String: Any] = [:]
```
maxep: suggestion\ I think we can use the `RUMViewEvent.View.Accessibility` type here, instead of a dict.

DatadogRUM/Sources/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -736,6 +734,12 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
             // if a watchdog termination occurs in this session, in the next session
             // a watchdog termination event will be sent using saved view event.
             dependencies.watchdogTermination?.update(viewEvent: event)
+
+            // Clear changed attributes after sending the view update
+            if let accessibilityReader = accessibilityReader,
+               accessibilityReader.hasValidAccessibilityData {
```
maxep: question/ why do we check for `hasValidAccessibilityData` here? It could always clear the reader, no?

DatadogRUM/Tests/RUMViewEventsFilterTests.swift
```diff
@@ -105,6 +105,53 @@ final class RUMViewEventsFilterTests: XCTestCase {
...
+        let actual = sut.filter(events: events)
+
+        // Verify that accessibility events are kept
+        let accessibilityEvents = actual.filter { event in
+            guard let metadata = event.metadata,
+                  let viewMetadata = try? JSONDecoder().decode(RUMViewEvent.Metadata.self, from: metadata) else {
+                return false
+            }
+            return viewMetadata.hasAccessibility == true
+        }
```
maxep: suggestion/
```suggestion
        let decoder = JSONDecoder()

        let accessibilityEvents = try actual.filter { event in
            let metadata = try event.metadata.map { try decoder.decode(RUMViewEvent.Metadata.self, from: $0) }
            return metadata?.hasAccessibility == true
        }
```

maxep (commented):
It looks great, well done with the filtering 👍  I have left few questions/suggestions

DatadogRUM/Sources/RUMMonitor/Scopes/RUMViewScope.swift
barboraplasovska: Yes you're right, it should always be cleared.

DatadogRUM/Sources/DataModels/RUMDataModelsMapping.swift
```diff
@@ -67,21 +67,64 @@ internal extension RUMViewEvent {
...
+               accessibility.reducedAnimationsEnabled != nil ||
+               accessibility.rtlEnabled != nil ||
+               accessibility.screenReaderEnabled != nil ||
+               accessibility.shakeToUndoEnabled != nil ||
+               accessibility.shouldDifferentiateWithoutColor != nil ||
+               accessibility.singleAppModeEnabled != nil ||
+               accessibility.speakScreenEnabled != nil ||
+               accessibility.speakSelectionEnabled != nil ||
+               accessibility.textSize != nil ||
+               accessibility.videoAutoplayEnabled != nil
```
Valpertui: Can lead to very slow compilation of the expression (while being instaneous check at runtime), can we look at the compilation time of this?   
If the expression is slow (>1 second), maybe split it into sub expression or maybe consider using an `Any? array` of properties + `.first(where({ $0 != nil}) != nil` solution.

joepeeples (approved):

DatadogRUM/Sources/Feature/RUMFeature.swift
```diff
@@ -73,7 +73,7 @@ internal final class RUMFeature: DatadogRemoteFeature {
         }
 
         var accessibilityReader: AccessibilityReading? = nil
-        if #available(iOS 13.0, tvOS 13.0, *), configuration.featureFlags[.collectAccessibilitySettings] {
+        if configuration.collectAccessibility, #available(iOS 13.0, tvOS 13.0, *) {
```
maciejburda: nit/preference: I think it's better to have #available first. Reads slightly more clearly imo

DatadogRUM/Sources/RUMContext/AccessibilityReader.swift
```diff
@@ -262,4 +292,149 @@ internal final class AccessibilityReader: AccessibilityReading {
...
+               state.speakScreenEnabled != nil ||
+               state.speakSelectionEnabled != nil ||
+               state.rtlEnabled != nil
+    }
+
+    /// Track changes between two accessibility states
+    private func trackChanges(from oldState: AccessibilityInfo, to newState: AccessibilityInfo) {
+        // Only track changes if there are actual differences
+        if hasChanges(between: oldState, and: newState) {
+            self.changedAttributes = RUMViewEvent.View.Accessibility(
```
maciejburda: not sure how often it's called, but maybe we could make the object conform to Equatable protocol and compare if the two are different before merging?

maciejburda (approved):
LGTM - some minor comments / suggestions

joepeeples (approved):

DatadogRUM/Sources/RUMContext/AccessibilityReader.swift
```diff
@@ -9,19 +9,44 @@ import Foundation
...
 internal final class AccessibilityReader: AccessibilityReading {
     @ReadWriteLock
-    private(set) var state: Accessibility
+    private(set) var state: AccessibilityInfo
 
     private let notificationCenter: NotificationCenter
     private var observers: [NSObjectProtocol] = []
 
+    /// Track if this is the first view update
+    private var isFirstViewUpdate = true
```
simaoseica-dd: We are leaking `View` logic to a Reader that should be only focus on `Accessibility`. 
This `Reader` should focus only in maintaining the current accessibility state. Any other logic, like diffing, should be a concern of the entities that read it. Can be a View or not.

I see this a problem with the multi-view concept.

DatadogRUM/Sources/DataModels/RUMDataModelsMapping.swift
simaoseica-dd: I see the the `view` to proactively store the current value of `ValidAccessibility`. This way you can also avoid this noise on the `RUMDataModelsMapping`

DatadogRUM/Sources/RUMContext/AccessibilityReader.swift
barboraplasovska: As you suggested, I moved the logic out of the `AccessibilityReader` and track it directly within the `RUMViewScope`.

DatadogRUM/Sources/DataModels/RUMDataModelsMapping.swift
Valpertui: After verification, this expression isn't slow to compile, resolving

DatadogRUM/Sources/DataModels/Accessibility.swift
```diff
@@ -54,4 +54,123 @@ internal struct Accessibility: Codable {
         case speakSelectionEnabled = "speak_selection_enabled"
         case rtlEnabled = "rtl_enabled"
     }
+
+    static let empty = AccessibilityInfo()
```
simaoseica-dd: If the constructor changes in the future with default values, it will not be empty and we can expect unpredictable behaviour. Also, empty might be misunderstood since here it refers to a nil state.

A state can help here:
`private(set) var hasAccessibilityData = false`

DatadogRUM/Sources/DataModels/Accessibility.swift
```diff
@@ -6,7 +6,7 @@
 
 import Foundation
 
-internal struct Accessibility: Codable {
+internal struct AccessibilityInfo: Codable, Equatable {
```
simaoseica-dd: Can you update the filename as well?

DatadogRUM/Sources/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -643,12 +650,22 @@ extension RUMViewScope {
...
 
-        if let accessibility = self.accessibilityReader?.state {
-            localAttributes["accessibility"] = accessibility
+        var accessibility: RUMViewEvent.View.Accessibility? = nil
+        if isFirstViewUpdate {
+            // For first view update, send the entire state
+            accessibility = currentAccessibilityState?.toRUMViewAccessibility()
+        } else {
+            // For subsequent updates, send only the differences
+            let accessibilityDifferences = currentAccessibilityState?.differences(from: accessibilityState) ?? .empty
```
simaoseica-dd: You can remove the fallback `?? .empty`, since it is expected nil `accessibility`

DatadogRUM/Sources/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -643,12 +650,22 @@ extension RUMViewScope {
...
+        // Take one snapshot of the current accessibility state
+        let currentAccessibilityState = accessibilityReader?.state
 
-        if let accessibility = self.accessibilityReader?.state {
-            localAttributes["accessibility"] = accessibility
+        var accessibility: RUMViewEvent.View.Accessibility? = nil
+        if isFirstViewUpdate {
+            // For first view update, send the entire state
+            accessibility = currentAccessibilityState?.toRUMViewAccessibility()
+        } else {
```
simaoseica-dd: The diffing below can be ignored if the states are equal
`else if currentAccessibilityState != accessibilityState`

DatadogRUM/Sources/Feature/RUMViewEventsFilter.swift
```diff
@@ -76,6 +76,11 @@ internal struct RUMViewEventsFilter {
 
         let viewMetadata = try decoder.decode(RUMViewEvent.Metadata.self, from: metadata)
 
+        if viewMetadata.hasAccessibility == true {
+                // If this event has accessibility information, always keep it
```
simaoseica-dd:
```suggestion
            // If this event has accessibility information, always keep it
```

DatadogRUM/Sources/DataModels/RUMDataModelsMapping.swift
```diff
@@ -75,21 +76,62 @@ internal extension RUMViewEvent {
...
+            self.indexInSession = indexInSession
+        }
     }
 
     enum EventType: String, Codable {
         case view
     }
 
+    /// Checks if the accessibility object has any meaningful data
+    private func hasValidAccessibilityData() -> Bool {
```
simaoseica-dd: Can you remove this function? If the values are all nil, the view can have a `view.accessibility = nil`

`hasAccessibility: view.accessibility != nil`

DatadogRUM/Sources/DataModels/Accessibility.swift
simaoseica-dd: It does not need to be `Codable`, and the `CodingKeys` can be removed

DatadogRUM/Sources/DataModels/Accessibility.swift
```diff
@@ -54,4 +54,123 @@ internal struct Accessibility: Codable {
...
         case rtlEnabled = "rtl_enabled"
     }
+
+    static let empty = AccessibilityInfo()
+
+    var isEmpty: Bool { self == .empty }
+
+    /// Maps `AccessibilityInfo` to the generated RUM model.
+    /// Returns `nil` when the snapshot carries no data (keeps payload lean).
+    func toRUMViewAccessibility() -> RUMViewEvent.View.Accessibility? {
```
simaoseica-dd: Conversion can be a computed var

DatadogRUM/Sources/RUMMonitor/Scopes/RUMViewScope.swift
```diff
@@ -89,6 +89,10 @@ internal class RUMViewScope: RUMScope, RUMContextProvider {
     /// Tells if this scope has received the "start" command.
     /// If `didReceiveStartCommand == true` and another "start" command is received for this View this scope is marked as inactive.
     private var didReceiveStartCommand = false
+    /// Track if this is the first view update for the current view
+    private var isFirstViewUpdate = true
```
simaoseica-dd: You might not need this parameter if you check for `accessibilityState == nil`

barboraplasovska: You're right! I removed it

simaoseica-dd (approved):
Well done! 👏

joepeeples (approved):

PR #2414, opened 2025-07-30
RUM-5176 [CP] Allow reporting critical log in sync

maxep:
### What and why?

Logs counter-part of #2180

Cross-platform SDKs can report fatal errors before the native process crashes. However, Logs execute processing and write asynchronously, and the crash can suspend these executions, leading to the loss of the cross-platform error information.

We should let cross-platform SDKs to report error in sync, blocking the caller-thread.

### How?

The change involve changes at different layers:

#### Logs Module

Add method to `InternalLoggerProtocol` for logging a critical entry with a `completionHandler`. The `RemoteLogger` and `ConsoleLogger` call the completion handler after writing the log.

```swift
/// Internal Logger for Cross-Platform access.
public protocol InternalLoggerProtocol {
    /// Logs a critical entry then call completion.
    ///
    /// This method is meant for non-native or cross platform frameworks (such as KMP) to send error information
    /// synchronously.
    ///
    /// - Parameters:
    ///   - error: the `Error` object. It will be used to infer error details.
    ///   - attributes: a dictionary of attributes (optional) to add for this message. If an attribute with
    ///                 the same key already exist in this logger, it will be overridden (only for this message).
    ///   - completionHandler: A completion closure called when reporting the log is completed.
    func critical(
        message: String,
        error: Error?,
        attributes: [String: Encodable]?,
        completionHandler: @escaping CompletionHandler
    )
}
```

#### Objc Module

The `DDLogger` now expose a `_internal_sync_critical` to report an error in sync. This method will wait for the underlying completion handler to be called before returning.

### Review checklist

### DIFF
--- a/DatadogLogs/Sources/LoggerProtocol.swift
+++ b/DatadogLogs/Sources/LoggerProtocol.swift
@@ -227,7 +227,8 @@ extension CombinedLogger: InternalLoggerProtocol {
         errorKind: String?,
         errorMessage: String?,
         stackTrace: String?,
-        attributes: [String: Encodable]?) {
+        attributes: [String: Encodable]?
+    ) {
         combinedLoggers.forEach {
             $0._internal.log(
                 level: level,
@@ -237,6 +238,34 @@ extension CombinedLogger: InternalLoggerProtocol {
                 stackTrace: stackTrace,
                 attributes: attributes
             )
+        }
+    }
+
+    func critical(
+        message: String,
+        error: Error?,
+        attributes: [String: Encodable]?,
+        completionHandler: @escaping CompletionHandler
+    ) {
+        let group = DispatchGroup()
+
+        combinedLoggers.forEach {
+            group.enter()
+
+            $0._internal.critical(
+                message: message,
+                error: error,
+                attributes: attributes,
+                completionHandler: { group.leave() }
+            )
+        }
+
+        // Sync on a global queue for simplicity.
+        // In practice, users of the internal logger are cross-platform
+        // SDKs which don't use console logger and therefor don't use
+        // the combined-logger.
+        group.notify(queue: .global()) {
+            completionHandler()
         }
     }
 }
--- a/DatadogLogs/Sources/Logs+objc.swift
+++ b/DatadogLogs/Sources/Logs+objc.swift
@@ -323,3 +323,40 @@ public final class objc_Logger: NSObject {
         return objc_Logger(sdkLogger: Logger.create(with: configuration.configuration))
     }
 }
+
+extension objc_Logger {
+    /// **For Datadog internal use only. Subject to changes.**
+    ///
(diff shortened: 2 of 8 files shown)


DatadogLogs/Sources/Logs+objc.swift
```diff
@@ -323,3 +323,40 @@ public final class objc_Logger: NSObject {
...
+        logger.critical(
+            message: message,
+            error: error,
+            attributes: attributes.dd.swiftAttributes,
+            completionHandler: {
+                semaphore.signal()
+            }
+        )
+
+        _ = semaphore.wait(timeout: .now() + .seconds(2))
```
gonzalezreal, line 360: What happens if the wait times out?

DatadogLogs/Sources/LoggerProtocol.swift
```diff
@@ -239,4 +240,32 @@ extension CombinedLogger: InternalLoggerProtocol {
...
+            )
+        }
+
+        // Sync on a global queue for simplicity.
+        // In practice, users of the internal logger are cross-platform
+        // SDKs which don't use console logger and therefor don't use
+        // the combined-logger.
+        group.notify(queue: .global()) {
+            completionHandler()
+        }
```
gonzalezreal, line 269: Should we use a dedicated serial queue instead? Since the global queue is concurrent, many simultaneous critical logs could exhaust the queue's thread pool.

DatadogLogs/Sources/Logs+objc.swift
maxep: If it times out, we **might** loose the error log. This method will be consumed by cross-platform to report an error just before a crash, 2s is our best effort to report the error but we shouldn't block the caller thread longer.

DatadogLogs/Sources/LoggerProtocol.swift
maxep: That's a good point, I will change it. I'm really not a fan of this logic using a `DispatchGroup`, but I also don't want to add much complexity for such a niche feature..

maxep: Actually, I think it is best to use a global queue instead of creating one each time a critical log is reported. The global queue will be invoked only once at completion and we don't expect intensive use of this internal API. This won't be exposed to users, only use by cross-platform for reporting error before a crash.

gonzalezreal: Great. Thanks for clarifying!

DatadogLogs/Sources/Logs+objc.swift
gonzalezreal: Maybe we could add a comment clarifying the 2 second timeout.

gonzalezreal (approved):

PR #2418, opened 2025-08-01
RUM-11166 Mach Sampling Profiler

maxep:
### What and why?

Implement Mach sampling profiler strategy to collect stack traces and binary info of threads. Serialization of samples is not covered in this PR.

### How?

- **Mach Thread Sampling**: Leverages Mach kernel APIs (`task_threads`, `thread_suspend/resume`) to safely capture thread states and walk stack frames using frame pointers
- **Stack Walking**: Performs frame pointer traversal with bounds checking to capture instruction pointers and validate user-space addresses
- **Binary Image Resolution**: Uses `dladdr()` and Mach-O parsing to resolve instruction pointers to binary images, extracting UUIDs and load addresses for symbolication
- **Thread Safety**: Minimizes critical sections during thread suspension to avoid deadlocks, performing stack capture while target threads are suspended

```cpp
// Main sampling loop
void mach_sampling_profiler::main() {
    while (running) {
        uint64_t interval_nanos = config.sampling_interval_nanos;

        // Get all threads in current task
        thread_act_array_t threads;
        mach_msg_type_number_t count;
        task_threads(mach_task_self(), &threads, &count);

        for (uint32_t i = 0; i < count; i++) {
            sample_thread(threads[i], interval_nanos);
        }

        std::this_thread::sleep_for(std::chrono::nanoseconds(interval_nanos));
    }
}

// Critical section: suspend thread, walk stack, resume immediately
void mach_sampling_profiler::sample_thread(thread_t thread, uint64_t interval_nanos) {
    stack_trace_t trace;
    stack_trace_init(&trace, config.max_stack_depth, interval_nanos);

    if (thread_suspend(thread) == KERN_SUCCESS) {
        // CRITICAL: Thread suspended - avoid allocations/locks
        stack_trace_sample_thread(&trace, thread, config.max_stack_depth);
        thread_resume(thread); // Resume ASAP
    }

    // Post-processing: resolve binary images for symbolication
    for (uint32_t i = 0; i < trace.frame_count; i++) {
        binary_image_lookup_pc(&trace.frames[i].image,
                              (void*)trace.frames[i].instruction_ptr);
    }
}
```

### Review checklist


DatadogProfiler/Mach/include/mach_profiler.h
```diff
@@ -7,4 +7,151 @@
...
+    class mach_sampling_profiler;
+}
+
+extern "C" {
+#endif
+
+/**
+ * Structure representing a binary image loaded in memory.
+ */
+typedef struct {
```
awforsythe: nit: this type definition uses an anonymous struct, while other C-style `typedef struct` declarations in this file bind the type name (`binary_image_t`) to a named struct type (`struct binary_image`).

(It's of no real significance in practice, but using `typedef struct binary_image { ... } binary_image_t` would be more consistent and would make the type forward-declarable)

DatadogProfiler/Mach/include/mach_profiler.h
```diff
@@ -7,4 +7,151 @@
...
+ * Default base configuration values
+ */
+#define SAMPLING_CONFIG_DEFAULT { \
+    .sampling_interval_nanos = 1000000, \
+    .profile_current_thread_only = 0, \
+    .max_buffer_size = 1000, \
+    .max_stack_depth = 128, \
+    .max_thread_count = 100, \
+    .qos_class = QOS_CLASS_USER_INTERACTIVE \
+}
```
awforsythe: Designated struct initializers (e.g. `.foo = 1`) have been valid C since C99, but they weren't actually supported in C++ until C++20 🙃 - if you need to target C++17, you may want to consider another means of initializing these default values.

DatadogProfiler/Mach/mach_sampling_profiler.cpp
```diff
@@ -0,0 +1,446 @@
...
+//   - On x86_64: similar layout with kernel space starting around 0x8000000000000000
+//   - This leaves the upper address space for kernel/system use
+//   - Reference: ARM64 memory layout documentation, x86_64 canonical addressing
+//
+// FRAME_POINTER_ALIGNMENT (8 bytes):
+//   - 64-bit systems require 8-byte alignment for pointers
+//   - Stack frame pointers must be properly aligned to avoid bus errors
+//   - Reference: ARM64/x86_64 ABI specifications
+
+#define MIN_USERSPACE_ADDR    0x1000ULL              // 4KB - avoid null deref region
```
awforsythe: Nit: `#defines` in this file could be replaced with `static const` declarations, for stronger typing, easier refactoring, and to avoid potential name collisions (in the event that some other header included in this TU used the same names). This C-style usage is still pretty idiomatic for this kind of low-level code, though.

DatadogProfiler/Mach/mach_sampling_profiler.cpp
```diff
@@ -0,0 +1,446 @@
...
+
+#define MAX_LOAD_COMMANDS     1000                   // Generous upper bound for ncmds
+#define MAX_LOAD_COMMAND_SIZE 0x10000                // 64KB max per load command
+
+/**
+ * Validates if an address is within reasonable user-space bounds.
+ * Rejects null pointers, kernel addresses, and other invalid ranges.
+ */
+#define IS_VALID_USERSPACE_ADDR(addr) \
+    (((uintptr_t)(addr)) >= MIN_USERSPACE_ADDR && ((uintptr_t)(addr)) <= MAX_USERSPACE_ADDR)
```
awforsythe: Nit: Same as above; these macros could be functions for the sake of readability and debuggability... but again, this is a pretty conservative preprocessor usage that's not out of place in this sort of code, so I wouldn't split hairs.

(e.g.)

```cpp
static constexpr bool is_valid_userspace_addr(uintptr_t addr) {
    return addr >= MIN_USERSPACE_ADDR && addr <= MAX_USERSPACE_ADDR;
}
```

DatadogProfiler/Mach/mach_sampling_profiler.cpp
```diff
@@ -0,0 +1,446 @@
...
+bool binary_image_init(binary_image_t* info) {
+    if (!info) return false;
+    memset(info->uuid, 0, sizeof(uuid_t));
+    info->load_address = 0;
+    info->filename = nullptr;
+    return true;
+}
+
+/**
+ * Destroys a binary image structure, freeing any allocated memory.
```
awforsythe: Extreme nit: _"freeing any memory allocated by that struct"_ (or _"within that struct"_) might make it clearer that the `info` struct itself is not being freed (as you'd expect, given that it's passed by the caller on init, not allocated by the profiler implementation)

DatadogProfiler/Mach/mach_sampling_profiler.cpp
```diff
@@ -0,0 +1,446 @@
...
+bool thread_get_frame_pointers(thread_t thread, void** fp, void** pc) {
+#if defined (__i386__) || defined(__x86_64__)
+    x86_thread_state64_t state;
+    mach_msg_type_number_t count = x86_THREAD_STATE64_COUNT;
+    if (thread_get_state(thread, x86_THREAD_STATE64, (thread_state_t)&state, &count) == KERN_SUCCESS) {
+        *fp = (void*)state.__rbp;
+        *pc = (void*)state.__rip;
+        return true;
+    }
+#elif defined (__arm__) || defined (__arm64__)
```
awforsythe: For my own curiosity: do we build and test on both x64 and arm64 architectures in `dd-sdk-ios` CI?

DatadogProfiler/Mach/mach_sampling_profiler.cpp
```diff
@@ -0,0 +1,446 @@
...
+ * @param config The sampling configuration to use
+ * @param callback Function to call with collected stack traces
+ * @param ctx Context to pass to the callback
+ */
+mach_sampling_profiler::mach_sampling_profiler(
+    const sampling_config_t* config,
+    stack_trace_callback_t callback,
+    void* ctx)
+    : running(false)
+    , config(SAMPLING_CONFIG_DEFAULT)
```
awforsythe: As mentioned, this macro expands to a named initializer list, which wouldn't be conformant C++17.

Off the top of my head, you might consider:

1. Defining each default value in the header, so that they're clearly advertised, e.g.
    - `#define SAMPLING_CONFIG_DEFAULT_MAX_STACK_DEPTH 128`, or
    - `static const uint32_t SAMPLING_CONFIG_DEFAULT_MAX_STACK_DEPTH = 128;`, which I believe is still valid C99
2. Declaring a `static sampling_config_t DEFAULT_SAMPLING_CONFIG` in your .cpp file, with its member values initialized from each of those constants
3. Initializing the address held in `this->config` to either the user-supplied `config` pointer or `&DEFAULT_SAMPLING_CONFIG` _(edit: I misread this; the member is a value and the assignment is a dereference-and-copy)_

...but whatever works is fine. C++17 may not even be a serious constraint if the full range of supported iOS build environments supports this feature anyway.

awforsythe (dismissed):
This looks like very competent low-level C/C++ code to me!

DatadogProfiler/Mach/include/mach_profiler.h
maxep: Thanks, that was a miss!

DatadogProfiler/Mach/mach_sampling_profiler.cpp
maxep: No we don't, our CI runs on Apple M1, so arm64 only. That's the only arch we care about in practice.
Now, I realise that the conditions are wrong, 32 and 64 bit should use different state type.

DatadogProfiler/Mach/mach_sampling_profiler.cpp
maxep: Good point, I went for the `static sampling_config_t DEFAULT_SAMPLING_CONFIG` 👍

DatadogProfiler/Mach/include/mach_profiler.h
maxep: Ah, good to know! I have migrated to a static const var 👍

DatadogProfiling/Mach/mach_sampling_profiler.h
```diff
@@ -0,0 +1,125 @@
...
+ * Can be extended for statistical variations.
+ */
+class mach_sampling_profiler {
+public:
+    /**
+     * @brief Constructs a new profiler instance
+     * 
+     * @param config Configuration for the profiler
+     * @param callback Function to call with collected stack traces
+     * @param ctx Contextto pass to the callback
```
gonzalezreal:
```suggestion
     * @param ctx Context to pass to the callback
```

DatadogProfiling/Tests/MachSamplingProfilerTests.swift
```diff
@@ -0,0 +1,499 @@
...
+//            expectation.fulfill()
+//        }
+//
+//        // Then
+//        wait(for: [expectation], timeout: 5.0)
+//
+//        // Cleanup
+//        profiler_stop(profiler) // Ensure stopped
+//        profiler_destroy(profiler)
+//    }
```
gonzalezreal, line 201: Should this test be un-commented or removed?

gonzalezreal (approved):
Excellent work!

Only two minor issues: a typo and a commented test.

DatadogProfiling/Tests/MachSamplingProfilerTests.swift
maxep: I will keep it as commented for now as I need to fix the thread safety, I will tackle that later 👍

PR #2422, opened 2025-08-06
Fix visionOS availability checks

gonzalezreal:
### What and why?

We were mistakenly using `visionOS 18.0` instead of `visionOS 2.0` in availability checks.

### How?

Replace `visionOS 18.0` with `visionOS 2.0`


simaoseica-dd (dismissed):
There is also this mistake in the integration tests
`RUMSwiftUIAutoInstrumentationTests`

gonzalezreal:
> There is also this mistake in the integration tests `RUMSwiftUIAutoInstrumentationTests`

Good catch! Fixed that one as well. Thanks :)

simaoseica-dd (approved):

PR #2427, opened 2025-08-07
RUM-11032 Add Feature Operations steps commands + public APIs

mariedm:
### What and why?

### How?

### Review checklist


DatadogRUM/Sources/RUMMonitor/RUMCommand.swift
```diff
@@ -580,3 +580,60 @@ internal struct RUMUpdatePerformanceMetric: RUMCommand {
...
+    var name: String { get }
+    /// The key of the operation for this operation step (when running several instances of the same operation)
+    var operationKey: String? { get }
+    /// The step type (start, end, retry, etc.)
+    var stepType: RUMVitalEvent.Vital.StepType { get }
+    /// The reason for failure, if applicable
+    var failureReason: RUMVitalEvent.Vital.FailureReason? { get }
+}
+
+internal struct RUMStartStepVitalCommand: RUMFeatureOperationStepVitalCommand {
```
simaoseica-dd: question: Why not only one command detached from the feature (`FeatureOperation`)? 
All of them have the same parameters, just the value changes. I'm thinking on the reuse of the Vital command for other features.

TestUtilities/Sources/Mocks/DatadogRUM/VitalMocks.swift
```diff
@@ -58,3 +58,16 @@ extension ViewHitchesMock: AnyMockable, RandomMockable {
         ViewHitchesMock(hitchesDataModel: (Array(repeating: (.mockAny(), .mockAny()), count: .mockAny()), .mockAny())) as! Self
     }
 }
+
+extension RUMFeatureOperationFailureReason: AnyMockable, RandomMockable {
```
simaoseica-dd: can the `RUMFeatureOperationFailureReason` be `CaseIterable`?

DatadogRUM/Sources/RUMMonitorProtocol.swift
```diff
@@ -405,6 +445,9 @@ internal class NOPMonitor: RUMMonitorProtocol {
     func startAction(type: RUMActionType, name: String, attributes: [AttributeKey: AttributeValue]) { warn() }
     func stopAction(type: RUMActionType, name: String?, attributes: [AttributeKey: AttributeValue]) { warn() }
     func addFeatureFlagEvaluation(name: String, value: Encodable) { warn() }
+    func startFeatureOperation(name: String, operationKey: String?, attributes: [AttributeKey: AttributeValue]) { warn() }
+    func succeedFeatureOperation(name: String, operationKey: String?, attributes: [AttributeKey: AttributeValue]) { warn() }
+    func failFeatureOperation(name: String, operationKey: String?, reason: RUMFeatureOperationFailureReason, attributes: [AttributeKey: AttributeValue]) { warn() }
```
simaoseica-dd, line 450: I'm thinking we can simplify the public interface for the consumers of this API by having only one `start` and `stop`. The `stopFeatureOperation` could have an optional `failureReason` to know if it failed or not.

simaoseica-dd (commented):
Left 2 questions to simplify the public interface and the RUM commands.

DatadogRUM/Sources/RUMMonitor/RUMCommand.swift
mariedm: This could also work, but I think this approach is safer and more explicit. It also seems to follow the existing pattern where start and stop commands are separated (view, resource, etc.). At the same time, we are already planning to introduce new steps in the future (update, retry), so having a single command would make things easier.

DatadogRUM/Sources/RUMMonitorProtocol.swift
mariedm: We need to enforce the `failureReason` when an operation failed, so we need two separate APIs.

TestUtilities/Sources/Mocks/DatadogRUM/VitalMocks.swift
mariedm: I can add it here, but I can't add the conformance directly on the original model since it's automatically generated.

DatadogRUM/Sources/RUMMonitor/RUMCommand.swift
mariedm: Refactored ✅

DatadogRUM/Sources/RUMMonitor/Monitor.swift
```diff
@@ -464,6 +466,63 @@ extension Monitor: RUMMonitorProtocol {
         )
     }
 
+    // MARK: - Feature Operations
+
+    // swiftlint:disable function_default_parameter_at_end
+    func startFeatureOperation(name: String, operationKey: String?, attributes: [AttributeKey: AttributeValue]) {
```
gonzalezreal: Should we have default values for `operationKey` (`nil`) and `attributes` (`[:]`)? This could also allow removing the linter exception.

DatadogRUM/Sources/RUMMonitorProtocol.swift
```diff
@@ -278,6 +281,43 @@ public protocol RUMMonitorProtocol: RUMMonitorViewProtocol, AnyObject {
         value: Encodable
     )
 
+    // MARK: - features
+
+    /// Starts a Feature Operation
+    /// - Parameters:
+    ///   - name: the name of the operation (e.g., `login_flow`)
+    ///   - operationKey: the key of the operation for this step (when running several instances of the same operation)
+    ///   - attributes: custom attributes to attach to this Feature
```
gonzalezreal:
```suggestion
    ///   - attributes: custom attributes to attach to this operation
```

gonzalezreal (commented):
Great work! 🚀
I left a couple of suggestions. Let me know your thoughts.

DatadogRUM/Sources/RUMMonitor/Monitor.swift
mariedm: Actually, it was a mistake that some parameters had default values in `Monitor.swift`. I'm following the existing pattern: no default values in the protocol/implementation, only in the convenience methods. I've corrected the method signatures and removed the swiftlint exceptions.

gonzalezreal: Ah, that makes sense. Thanks for clarifying!

PR #2433, opened 2025-08-11
RUM-11166 Create Profiling Feature

maxep:
### What and why?

This PR introduces the initial implementation of the Datadog Profiling feature. The profiling feature will capture CPU profiles in pprof format.

  > [!IMPORTANT]
  > The public APIs declared in this implementation are not finalized and are subject to changes.

### How?

The implementation includes:

- **Public API**: `Profiling` interface with static methods for configuration, enabling , starting, and stopping profiling sessions
- **Core Components**:
  - `MachProfiler`: Implementation of the sampling profiler using Mach APIs
  - `ProfilerFeature`: Integration with Datadog's core feature system
  - `RequestBuilder`: Handles formatting and uploading profile data
  - `ProfileEvent`: Data structure for profile events with RUM context integration
- **Infrastructure**: Moved `MultipartFormData` utilities to `DatadogInternal` for shared usage across features

The profiling data is automatically enriched with RUM context (application ID, session ID, view ID) when available and tagged with relevant metadata (service, version, environment).

### Review checklist

### DIFF
--- a/DatadogProfiling/Sources/RequestBuilder.swift
+++ b/DatadogProfiling/Sources/RequestBuilder.swift
@@ -39,16 +39,18 @@ internal struct RequestBuilder: FeatureRequestBuilder {
         var multipart = multipartBuilder
 
         multipart.addFormData(
-            name: ProfileEvent.Constants.eventFilename,
+            name: "event",
             filename: ProfileEvent.Constants.eventFilename,
             data: event,
             mimeType: "application/json"
         )
 
-        multipart.addFormData(
+        let decoder = JSONDecoder()
+
+        try multipart.addFormData(
             name: ProfileEvent.Constants.wallFilename,
             filename: ProfileEvent.Constants.wallFilename,
-            data: prof.data,
+            data: decoder.decode(Data.self, from: prof.data),
             mimeType: "application/octet-stream"
         )
 
--- a/DatadogProfiling/Tests/RequestBuilderTests.swift
+++ b/DatadogProfiling/Tests/RequestBuilderTests.swift
@@ -12,17 +12,33 @@ import TestUtilities
 
 class RequestBuilderTests: XCTestCase {
     private let rumContext: RUMCoreContext = .mockRandom()
-    private var event: Event = .mockWith(
-        data: .mockRandom(),
-        metadata: .mockRandom()
+
+    let profileEvent = ProfileEvent(
+        family: .mockRandom(),
+        runtime: .mockRandom(),
+        version: .mockRandom(),
+        start: .mockRandomInThePast(),
+        end: Date(),
+        attachments: [],
+        tags: .mockAny()
     )
+
+    let profileData: Data = .mockRandom()
+
+    private func mockEvent() throws -> Event {
+        let encoder = JSONEncoder.dd.default()
+        return try Event(
+            data: encoder.encode(profileData),
+            metadata: encoder.encode(profileEvent)
+        )
+    }
 
     func testItCreatesPOSTRequest() throws {
         // Given
         let builder = RequestBuilder(customUploadURL: nil, telemetry: TelemetryMock())
 
         // When
-        let request = try builder.request(for: [event], with: .mockAny(), execution: .mockAny())
+        let request = try builder.request(for: [mockEvent()], with: .mockAny(), execution: .mockAny())
 
         // Then
         XCTAssertEqual(request.httpMethod, "POST")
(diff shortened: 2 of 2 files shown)


maciejburda (approved):

gonzalezreal (approved):
Looks good! Clean foundation for the profiling feature with solid tests

DatadogProfiling/Sources/DatadogProfiling.swift
```diff
@@ -0,0 +1,109 @@
...
+    /// - Parameter core: The Datadog core instance to use. Defaults to the default core.
+    public static func stop(in core: DatadogCoreProtocol = CoreRegistry.default) {
+        guard
+            let feature = core.get(feature: ProfilerFeature.self),
+            let profile = try? feature.profiler.stop()
+        else {
+            return
+        }
+
+        core.scope(for: ProfilerFeature.self).eventWriteContext { context, writer in
```
simaoseica-dd, line 79: This logic will be scoped in another entity, right? 
Asking because this file can be clean of events and persistence and focus solely in the public APIs and configurations available to the client apps.

DatadogProfiling/Sources/ProfileEvent.swift
```diff
@@ -0,0 +1,63 @@
...
+    /// The RUM Session
+    internal struct Session: Encodable {
+        /// The RUM Session ID
+        let id: String
+    }
+
+    /// The RUM Views.
+    internal struct Views: Encodable {
+        /// List of RUM View IDs
+        let id: [String]
```
simaoseica-dd, line 30:
```suggestion
        let ids: [String]
```

DatadogProfiling/Sources/ProfileEvent.swift
```diff
@@ -0,0 +1,63 @@
...
+    /// The profiling version.
+    let version: String
+    /// Profile start date
+    let start: Date
+    /// Profile end date
+    let end: Date
+    /// The list of profiles attached to the event.
+    let attachments: [String]
+    /// The profile tags
+    let tags: String
```
simaoseica-dd, line 56: Why doesn't the naming follow the other events? I'm referring to the `ddtags` in the other events

DatadogProfiling/Sources/ProfileEvent.swift
```diff
@@ -0,0 +1,63 @@
...
+    /// The RUM Application
+    var application: Application? = nil
+    /// The RUM Session
+    var session: Session? = nil
+    /// The RUM Views.
+    var view: Views? = nil
```
simaoseica-dd, line 62: Can't it be an array of `View`s? `var views: [Views]? = nil`

DatadogProfiling/Tests/MockProfiler.swift
```diff
@@ -0,0 +1,45 @@
...
+
+import Foundation
+
+@testable import DatadogProfiling
+
+/// A mock implementation of the `Profiler` protocol for testing purposes.
+///
+/// This mock allows tests to simulate different profiler behaviors by controlling
+/// the returned profile data and error conditions without performing actual profiling.
+internal struct MockProfiler: Profiler {
```
simaoseica-dd, line 15: Can you add this mock to the `TestUtilities`?

simaoseica-dd (approved):
Very well done! 🎖️ 
Aligned with the other product features and with the entities with clear responsibilities.

DatadogProfiling/Sources/ProfileEvent.swift
maxep: It's actually called `tags_profiler` in the [Profiling Intake specs (internal)](https://datadoghq.atlassian.net/wiki/spaces/PROF/pages/3237806339/Profiling+Intake+API), I have shortened it to `ProfileEvent\.tags`. So no use of `ddtags` name there.

DatadogProfiling/Tests/MockProfiler.swift
maxep: This is scoped and use by the profiling unit-test only for now. We should move it to`TestUtilities` if we need sharing with other test suites IMHO. I think we will need this sharing for Integration Test with RUM but not sure yet.

DatadogProfiling/Sources/DatadogProfiling.swift
maxep: Yes probably 👍  I'm introducing the components but it subject to implementation change, same for the RUM context in the `ProfileEvent`. It will be revisited on following PRs so I will merge it as is.

PR #2436, opened 2025-08-14
RUM-11136: Move `ddtags` query parameters to RUM events

simaoseica-dd:
### What and why?

This PR moves the `service`, `sdk_version`, `version`, `env`, and `variant` query parameters to RUM events.
This change will ensure that RUM events remain consistently linked to their session values, making persisted batches resilient even if those values change between app launches.

### How?

The `ddtags` from the query parameters are now injected in all RUM events following the format `parameter:value` separated by commas `,`.

### Review checklist


maxep (approved):
LGTM!

PR #2439, opened 2025-08-19
RUM-11048 Profile aggregation and serialization

maxep:
### What and why?

This PR implements pprof data aggregation and serialization for Profiling. It adds the engine to collect, deduplicate, and serialize profiling samples into the standard pprof protobuf format, enabling ingestion from Profiling intake.

#### Key Components:
- Adds `profile` C++ class for sample aggregation with deduplication of strings, mappings, and locations
- Implements `profile_pprof_packer` to convert internal data structures to protobuf format
- Creates C API wrapper (`dd_pprof`) for Swift integration
- Updates `MachProfiler` to use the new aggregation system and generate pprof data

#### Testing:

Uses Cxx interoperability feature introduced in Swift 5.9 for testing aggregation and serialization.

### Review checklist

### DIFF
--- a/DatadogProfiling/Mach/include/profile.h
+++ b/DatadogProfiling/Mach/include/profile.h
@@ -191,6 +191,12 @@ public:
     /** @brief Get cached string ID for "tid" */
     uint32_t tid_str_id() const { return _tid_str_id; }
 
+    /** @brief Get profile start timestamp (mach_absolute_time) */
+    int64_t start_timestamp() const { return mach_time_to_epoch_ns(_start_timestamp); };
+
+    /** @brief Get profile end timestamp (mach_absolute_time) */
+    int64_t end_timestamp() const { return mach_time_to_epoch_ns(_end_timestamp); };
+
 private:
     /** @brief Deduplicated string table (index 0 is always empty string) */
     std::vector<std::string> _strings;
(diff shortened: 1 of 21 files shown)


DatadogProfiling/Mach/include/profile.h
```diff
@@ -0,0 +1,255 @@
...
+};
+
+/**
+ * @brief Individual profiling sample with stack trace and metadata
+ * 
+ * Represents a single profiling sample containing:
+ * - Stack trace as a sequence of location IDs (leaf-to-root order)
+ * - Associated labels (e.g., timestamps, thread info)
+ * - Sample values (e.g., CPU time, wall time, memory allocation)
+ * - Timestamp when the sample was collected
```
awforsythe, line 89: nit: The struct layout doesn't seem to store a "timestamp when the sample was collected," unless that's meant to be encoded in `values` somehow

DatadogProfiling/Mach/profile.cpp
```diff
@@ -0,0 +1,279 @@
...
+ * 
+ * @param timeref Time reference structure to initialize
+ */
+void mach_timeref_init(mach_timeref_t* timeref) {
+    // Get timebase info for mach time conversion
+    mach_timebase_info(&timeref->timebase_info);
+    
+    // Calculate current times
+    uint64_t mach_time = mach_absolute_time();
+    uint64_t mach_time_ns = mach_time * timeref->timebase_info.numer / timeref->timebase_info.denom;
```
awforsythe, line 45: My defensive-programming eyes spot a potential divide-by-zero here, but it seems like `mach_timebase_info` will only ever return `KERN_SUCCESS` and is guaranteed to set `denom` to a nonzero value, so carry on 🙂

maxep: That's a good point! I need to revisit this because of the comment [here](https://developer.apple.com/documentation/kernel/1462446-mach_absolute_time). I will merge it as is and open another PR, to either use `clock_gettime_nsec_np` or add a privacy manifest.

maxep: Hey @awforsythe 👋  I have address this in https://github.com/DataDog/dd-sdk-ios/pull/2488

PR #2440, opened 2025-08-21
RUM-11033 Update Feature Operation logic and add debug logs

mariedm:
### What and why?

This PR introduces a dedicated `RUMFeatureOperationManager`. The `RUMSessionScope` now forwards **Feature Operation** command processing to this manager. Additionally, active operations are tracked to improve developer experience by providing helpful warnings when APIs are misused.

This PR also fixes an issue where `operationStep` vital events were missing the `viewID`. This prevented proper correlation between views and vitals in the RUM UI.

### How?

- Introduced `RUMFeatureOperationManager`: a dedicated class responsible for processing operation commands and creating vital events, replacing the previous inline logic in `RUMSessionScope`
- Updated `RUMSessionScope`: now delegates FO command processing to the manager
- Session scope now extracts the active view from `viewScopes` and passes it explicitly to the FO manager, so that operation step vital events now include the correct `view.id` and `view.url` properties; eliminates the previous issue where FO events inherited stale context from the application scope
- Active, local operation tracking: maintains a set of active operations for troubleshooting purposes only
- Warning system: logs helpful warnings for common API misuse patterns:
  - Starting the same operation multiple times
  - Calling end/update/retry without a corresponding start

### Review checklist

### DIFF
--- a/DatadogRUM/Sources/RUMMonitor/RUMFeatureOperationManager.swift
+++ b/DatadogRUM/Sources/RUMMonitor/RUMFeatureOperationManager.swift
@@ -29,6 +29,7 @@ internal class RUMFeatureOperationManager {
     /// Set of active operation lookup keys for tracking and warning purposes
     /// Key format: "operationName" + "operationKey" (if any)
     private var activeOperations: Set<String> = []
+    private let maxActiveOperations = 500
 
     // MARK: - Dependencies
 
@@ -44,11 +45,25 @@ internal class RUMFeatureOperationManager {
 
     // MARK: - Public Interface
 
-    func process(_ command: RUMOperationStepVitalCommand, context: DatadogContext, writer: Writer, activeViewID: String, activeViewPath: String) {
-        let lookupKey = "\(command.name)\(command.operationKey ?? "")"
+    func process(_ command: RUMOperationStepVitalCommand, context: DatadogContext, writer: Writer, activeView: RUMViewScope?) {
+        // Validate command parameters
+        guard validateCommand(command) else {
+            return
+        }
+
+        if activeView == nil {
+            DD.logger.warn("RUM operation step command received without an active view. This may result in incomplete context information.")
+        }
 
         // Always create and send the vital event - this is the SDK's core responsibility
-        writeVitalEvent(from: command, context: context, writer: writer, activeViewID: activeViewID, activeViewPath: activeViewPath)
+        writeVitalEvent(
+            from: command,
+            context: context,
+            writer: writer,
+            activeView: activeView
+        )
+
+        let lookupKey = "\(command.name)\(command.operationKey ?? "")"
 
         // Track operation state for troubleshooting warnings
         switch command.stepType {
@@ -62,12 +77,7 @@ internal class RUMFeatureOperationManager {
 
     // MARK: - Private Methods
 
-    private func writeVitalEvent(from command: RUMOperationStepVitalCommand, context: DatadogContext, writer: Writer, activeViewID: String, activeViewPath: String) {
-        guard !command.name.isEmpty else {
-            DD.logger.error("Operations must have a non-empty name. \(command) call will be ignored.")
-            return
-        }
-
+    private func writeVitalEvent(from command: RUMOperationStepVitalCommand, context: DatadogContext, writer: Writer, activeView: RUMViewScope?) {
         let vital = RUMVitalEvent.Vital(
             vitalDescription: nil,
             duration: nil,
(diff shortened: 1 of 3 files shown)


ncreated (approved):
Nice 🎯 good cleanup in `SessionScope`, much better SRP 👌

DatadogRUM/Sources/RUMMonitor/RUMFeatureOperationManager.swift
```diff
@@ -0,0 +1,136 @@
...
+
+        case .end, .update, .retry:
+            trackOperationUpdate(name: command.name, operationKey: command.operationKey, lookupKey: lookupKey, stepType: command.stepType)
+        }
+    }
+
+    // MARK: - Private Methods
+
+    private func writeVitalEvent(from command: RUMOperationStepVitalCommand, context: DatadogContext, writer: Writer, activeViewID: String, activeViewPath: String) {
+        guard !command.name.isEmpty else {
```
simaoseica-dd: Can this be validated earlier in the process (before creating the lookupKey)? If it is empty nothing else is important to run.

DatadogRUM/Sources/RUMMonitor/RUMFeatureOperationManager.swift
```diff
@@ -0,0 +1,136 @@
...
+            name: command.name,
+            operationKey: command.operationKey,
+            stepType: command.stepType,
+            type: .operationStep
+        )
+
+        let vitalEvent = RUMVitalEvent(
+            dd: .init(),
+            application: .init(id: parent.context.rumApplicationID),
+            context: .init(contextInfo: command.globalAttributes.merging(command.attributes) { $1 }),
```
simaoseica-dd: All events associated with a view have now View attributes. Shouldn't the vital has them as well?

DatadogRUM/Sources/RUMMonitor/RUMFeatureOperationManager.swift
```diff
@@ -0,0 +1,136 @@
...
+            date: command.time.timeIntervalSince1970.toInt64Milliseconds,
```
simaoseica-dd: Shouldn't it take into account the `serverTimeOffset` like the other events?

DatadogRUM/Sources/RUMMonitor/Scopes/RUMSessionScope.swift
```diff
@@ -216,7 +221,15 @@ internal class RUMSessionScope: RUMScope, RUMContextProvider {
                 // Start view scope explicitly on receiving "start view" command
                 startView(on: startViewCommand, context: context)
             } else if let operationStepVitalCommand = command as? RUMOperationStepVitalCommand {
-                sendFeatureOperationStepVitalEvent(on: operationStepVitalCommand, context: context, writer: writer)
+                let activeView = viewScopes.first { $0.isActiveView }
+                // Forward command to the feature operation manager
+                featureOperationManager.process(
```
simaoseica-dd, line 226: Is it expected to be processed if there is no active view? If so, it would help to have a local warning about it.

simaoseica-dd (approved):
Well done scoping the logic in only one entity. 👌 
Left some validation comments.

DatadogRUM/Sources/RUMMonitor/RUMFeatureOperationManager.swift
mariedm: Yes, we can do that earlier, I'm moving it to the `process` method above.

DatadogRUM/Sources/RUMMonitor/RUMFeatureOperationManager.swift
mariedm: Yes, good idea, I'll add them too

DatadogRUM/Sources/RUMMonitor/RUMFeatureOperationManager.swift
mariedm: Thanks for pointing it out, that's a miss!

DatadogRUM/Sources/RUMMonitor/Scopes/RUMSessionScope.swift
mariedm: Yes, we want to send the event anyway. But that's a good point, I'll add a warning when that happens.

DatadogRUM/Sources/RUMMonitor/RUMFeatureOperationManager.swift
```diff
@@ -0,0 +1,190 @@
...
+ * Local operation tracking is purely for developer experience and debugging.
+ */
+
+internal class RUMFeatureOperationManager {
+    // MARK: - Properties
+
+    /// Set of active operation lookup keys for tracking and warning purposes
+    /// Key format: "operationName" + "operationKey" (if any)
+    private var activeOperations: Set<String> = []
+    private let maxActiveOperations = 500
```
simaoseica-dd, line 32: Since `RUMFeatureOperationManager` allocates strings for the operations, I would say that this limit could be increased.

simaoseica-dd (approved):

DatadogRUM/Sources/RUMMonitor/RUMFeatureOperationManager.swift
simaoseica-dd: You can ignore this comment since the update of the vital is always reported even if it is removed due to this limit

mariedm: > Since RUMFeatureOperationManager allocates strings for the operations, I would say that this limit could be increased.

Yes, we could! cc @satween

mariedm: > You can ignore this comment since the update of the vital is always reported even if it is removed due to this limit

Yes indeed, the vital will always be sent to BE, in any case; this dictionary tracks operations for debugging purposes only

PR #2449, opened 2025-08-26
RUM-9494: Discard locales from span events

simaoseica-dd:
### What and why?

This PR removes `device.locales` from Span events.
This array of strings is preventing custom span events from being ingested.

### How?

We have a centralized function to create `Device` models to be injected on events (`func normalizedDevice()`). 
This development adds an extra configuration to discard `device.locales`, if necessary.

### Review checklist


DatadogInternal/Sources/Context/DeviceInfo.swift
```diff
@@ -227,24 +227,32 @@ extension DatadogExtension where ExtendedType == _UIDevice {
 
 extension DatadogContext {
     /// Current device information to send in the events.
-    public var normalizedDevice: Device {
+    ///
+    /// - Parameter addLocales: Temporary boolean to remove locales from events that don't support array parameters.
```
mariedm, line 231: Should we link to a Jira or some doc for reference, so we know where to look at next time?

DatadogInternal/Sources/Context/DeviceInfo.swift
```diff
@@ -227,24 +227,32 @@ extension DatadogExtension where ExtendedType == _UIDevice {
...
+    /// - Returns: The device model for the events.
+    public func normalizedDevice(addLocales: Bool = true) -> Device {
         var battery: Double?
         if let batteryLevel = batteryStatus?.level {
             battery = Double(batteryLevel)
         }
 
-        var brightness: Double? = nil
+        var brightness: Double?
```
ncreated, line 241: question/ Some properties of `DeviceInfo` are not strings vs there is [this remark](https://github.com/DataDog/dd-sdk-ios/blob/792a6cbb2a076f0c939b7af8618a46c091d05858/DatadogTrace/Sources/Span/SpanEventEncoder.swift#L216) on span meta encoding:
```swift
    /// Encodes default `meta.*` attributes
    private func encodeDefaultMeta(_ span: SpanEvent, to container: inout KeyedEncodingContainer<StaticCodingKeys>) throws {
        // NOTE: RUMM-299 only string values are supported for `meta.*` attributes

        // ...
        try container.encode(span.device, forKey: .device)
```

This `NOTE` makes it clear why the array was breaking parsing. What about `Double` and `Bool`? If this remark is wrong, let's update it too.

simaoseica-dd: The span schema accounts for string, bool and number (double). This info is attached to the span events.
Updated the NOTE to reflect the contract on the span schema.

PR #2462, opened 2025-09-04
chore: Remove unused `objc_SessionReplayConfigurationPrivacyLevel`

simaoseica-dd:
### What and why?

This PR removes the unused `objc_SessionReplayConfigurationPrivacyLevel`. 
It was the Objective-C counterpart of `SessionReplayPrivacyLevel`, which is used for fine-grained masking conversions between the Browser SDK and the `DatadogWebViewTracking`. Since only one structure is required for this purpose, the Objective-C version is no longer needed. 

### Review checklist

### DIFF
--- a/DatadogSessionReplay/Sources/SessionReplay+objc.swift
+++ b/DatadogSessionReplay/Sources/SessionReplay+objc.swift
@@ -161,36 +161,6 @@ public final class objc_SessionReplayConfiguration: NSObject {
     }
 }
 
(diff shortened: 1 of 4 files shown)


gonzalezreal (commented):
LGTM!

My only concern is whether or not this could be a breaking change for some customers and we should deprecate this public API before removing it.

mariedm (approved):

PR #2464, opened 2025-09-05
RUM-11465: Register `FirstFrameReader` entity to the DisplayLinker

simaoseica-dd:
### What and why?

This PR introduces a new `RenderLoopReader` entity to capture information about the first frame rendered in the app.
Tracking this moment allows us to better measure and report app launch metrics.

### How?

The `DisplayLinker` within `RUMScopeDependencies` observes the `CADisplayLink` run loop and notifies any registered readers.
The new `FirstFrameReader` subscribes to these callbacks to detect the first rendered frame. Once this information is collected, it immediately propagates a new command through the chain to the `RUMScopes` responsible for processing commands.

The diagram below illustrates the current hierarchy of readers:

readers

### Review checklist


DatadogInternal/Sources/Context/MediaTimeProvider.swift
```diff
@@ -0,0 +1,29 @@
...
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+#if !os(watchOS)
+import QuartzCore
+#endif
+
+/// Provides current media uptime.
+public protocol MediaTimeProvider: Sendable {
```
maxep: request/ I would suggest to move this to the RUM module, alongside `FirstFrameReader`. There is no use of it outside of RUM.

DatadogCore/Tests/Datadog/RUM/RUMVitals/FirstFrameReaderTests.swift
```diff
@@ -0,0 +1,36 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import XCTest
+import TestUtilities
+
+@testable import DatadogRUM
+
+final class FirstFrameReaderTests: XCTestCase {
```
maxep, line 12: request/ This should be tested in RUM I believe.

DatadogInternal/Sources/Context/MediaTimeProvider.swift
maxep: suggestion/ I think we can add clarity by calling it 
```swift
protocol CACurrentMediaTimeProvider {
    var current: CFTimeInterval { get }
}
```
I was a bit confused with the term 'Media'

DatadogRUM/Sources/RUMVitals/RenderLoop/FirstFrameReader.swift
```diff
@@ -0,0 +1,51 @@
...
+    func publish(to subscriber: RUMCommandSubscriber) {
+        self.subscriber = subscriber
+    }
+}
+
+extension FirstFrameReader: RenderLoopReader {
+    func stop() { self.isActive = false }
+
+    func didUpdateFrame(link: FrameInfoProvider) {
+        if firstFrameTimestamp == nil && isActive {
```
maxep, line 41: question/ Is there a way to unregister the DisplayLink after the first trigger?

iadjivon (dismissed):
All set from docs.

DatadogRUM/Sources/RUMVitals/RenderLoop/FirstFrameReader.swift
simaoseica-dd: Yes, I will do it in the next task after writing the event.

DatadogInternal/Sources/Context/MediaTimeProvider.swift
simaoseica-dd: The initial idea was to have it next to the `DateProvider`. Nevertheless, I moved it to RUM 👍

TestUtilities/Sources/Mocks/DatadogInternal/MediaTimeProviderMock.swift
maxep, line 1: I think this should be moved to the right folder as well :)

maxep (approved):
Looks great! I left a minor comment

PR #2473, opened 2025-09-16
Unify trace context injection behaviour

maciejburda:
### What and why?

Unifies the behaviour of HTTP Header Writers to always inject headers with sampling decision information if `TracingContextInjection` is configured to `all`.

### Review checklist


DatadogInternal/Sources/NetworkInstrumentation/B3/B3HTTPHeadersReader.swift
```diff
@@ -45,7 +45,7 @@ public class B3HTTPHeadersReader: TracePropagationHeadersReader {
 
     public var sampled: Bool? {
         if let single = httpHeaderFields[B3HTTPHeaders.Single.b3Field] {
-            return single != B3HTTPHeaders.Constants.unsampledValue
+            return single.components(separatedBy: B3HTTPHeaders.Constants.b3Separator)[safe: 2] != B3HTTPHeaders.Constants.unsampledValue
```
maciejburda, line 48: I wonder if we should read when B3 when it is just `"0"` as well.

I think that was the main source of confusion, because in public documentation:
https://github.com/openzipkin/b3-propagation
Both `b3=0` and `b3={TraceId}-{SpanId}-{SamplingState}-{ParentSpanId}` are valid.

I think that in our case we always want to use the latter, or not inject header at all.
But maybe if `sampled = false` and `TraceContextInjection = sampled` we should inject `b3=0`?

DatadogInternal/Sources/NetworkInstrumentation/B3/B3HTTPHeadersWriter.swift
```diff
@@ -78,39 +78,38 @@ public class B3HTTPHeadersWriter: TracePropagationHeadersWriter {
...
-                        String(traceContext.traceID, representation: .hexadecimal32Chars),
-                        String(traceContext.spanID, representation: .hexadecimal16Chars),
-                        sampled ? Constants.sampledValue : Constants.unsampledValue,
-                        traceContext.parentSpanID.map { String($0, representation: .hexadecimal16Chars) }
-                    ]
-                    .compactMap { $0 }
-                    .joined(separator: Constants.b3Separator)
-                } else {
-                    traceHeaderFields[B3HTTPHeaders.Single.b3Field] = Constants.unsampledValue
-                }
```
maciejburda, line 110: Here's the issue. Take a look at the dead code in line `103` - dead code because of if statement before.

DatadogInternal/Sources/NetworkInstrumentation/Datadog/HTTPHeadersWriter.swift
```diff
@@ -49,29 +49,37 @@ public class HTTPHeadersWriter: TracePropagationHeadersWriter {
     /// - Parameter spanID: The span ID.
     /// - Parameter parentSpanID: The parent span ID, if applicable.
     public func write(traceContext: TraceContext) {
-        switch (traceContextInjection, traceContext.isKept) {
-        case (.all, _), (.sampled, true):
```
maciejburda, line 53: This was actually correct - but decided to rework the code to match b3 implementation.

I think it's slightly better to read now.

DatadogInternal/Sources/NetworkInstrumentation/W3C/W3CHTTPHeadersWriter.swift
```diff
@@ -61,48 +61,52 @@ public class W3CHTTPHeadersWriter: TracePropagationHeadersWriter {
...
+                case .all:      return true
+                case .sampled:  return sampled
+            }
+        }()
+        guard shouldInject else {
+            return
+        }
 
-        switch (traceContextInjection, sampled) {
-        case (.all, _), (.sampled, true):
```
maciejburda, line 66: Same here. Behaviour was correct according to tests.

I hope it's a bit less cryptic now.

DatadogInternal/Tests/NetworkInstrumentation/B3HTTPHeadersWriterTests.swift
```diff
@@ -44,13 +44,13 @@ class B3HTTPHeadersWriterTests: XCTestCase {
         )
 
         let headers = writer.traceHeaderFields
-        XCTAssertEqual(headers[B3HTTPHeaders.Single.b3Field], "0")
+        XCTAssertEqual(headers[B3HTTPHeaders.Single.b3Field], "00000000000004d200000000000004d2-0000000000000929-0-000000000000162e")
```
maciejburda, line 47: That's the key difference in the behaviour in this PR.

DatadogInternal/Sources/NetworkInstrumentation/B3/B3HTTPHeadersReader.swift
zacharycmontoya: Backend SDK engineer here 👋🏼 

I think we should be flexible on what we extract and accept both `b3=0` and `b3={TraceId}-{SpanId}-{SamplingState}-{ParentSpanId}` styles, but only inject in the `b3={TraceId}-{SpanId}-{SamplingState}-{ParentSpanId}` style

zacharycmontoya: Actually, if the input is `b3=0`, then we don't have a trace-id or span-id at all. For reference in our C# library, we don't accept a value of `b3=0` so if a new span is created then we get a brand new tracecontext. I wonder what the other libraries do...let me get back to you, but I think the injection behavior you have encoded is accurate so let's not block on this

zacharycmontoya (approved):
LGTM. I can follow up on the case where `b3=0` is extracted

simaoseica-dd (dismissed):

urseberry (dismissed):

PR #2477, opened 2025-09-18
feat: Add initial implementation of FlagsEvaluationContext and precomputed assignments support

sameerank:
### What and why?

This PR adds from initial parts to the DatadogFlags module to support feature flagging. The implementation provides:

- `FlagsClient` - Main client for flag evaluation with methods like `getBooleanValue`, `getStringValue`, `getIntegerValue`, `getDoubleValue`, and `getObjectValue`
- `FlagsEvaluationContext` - Context object containing targeting key and attributes for flag evaluation
- Context-driven flag fetching - `setEvaluationContext` method that makes POST requests to /precompute-assignments endpoint to fetch flags based on user context
- Thread-safe flag storage - Concurrent queue-based storage with disk persistence for offline access
- Multi-site support - Endpoint building for all Datadog regions (US, EU, AP1, AP2, US3, US5)
- Enhanced configuration - Support for custom headers, application IDs, flagging proxies, and base URLs

Supported usage
```
Datadog.initialize(with: config)
Flags.enable(with: flagsConfig) // TODO
let client = FlagsClient.create(with: flagsClientConfig)
let context = FlagsEvaluationContext(
    targetingKey: "user-123",
    attributes: [
        "email": "user@example.com",
        "product": "premium",
        "region": "us-east-1"
    ]
)
client.setEvaluationContext(context) { result in
    switch result {
    case .success:
        // Context updated, flags refreshed
    case .failure(let error):
        // Handle error
    }
}
let result = client.getBooleanValue(key: "feature-key", defaultValue: false)
```

### How?

The implementation aims to follow Datadog SDK patterns and mirror the OpenFeature API.

- iOS 12.0+ Compatibility: Uses completion handlers instead of async/await to support the minimum deployment target
- OpenFeature API Alignment: Method names match OpenFeature Swift SDK conventions (`getIntegerValue`, `getDoubleValue`, `getObjectValue`)
- Network Layer: Built on `URLSession` with error handling and response parsing
- Context Serialization: Matches JavaScript client behavior by stringifying non-string attribute values
- Disk Persistence: JSON-based caching with metadata tracking (`fetchedAt` timestamps, evaluation context)
- Testing: With mock implementations and verifying context attribute serialization

### Review checklist


DatadogFlags/Tests/Resources/precomputed-v1.json
sameerank, line 1: Same as https://github.com/DataDog/openfeature-js-client/blob/d1f5c087fc627eef2934a27059b6098a51350492/packages/browser/test/data/precomputed-v1.json

Eventually, we should move test data out of the SDK code and reference a centralized resource

DatadogFlags/Sources/FlagsEndpointBuilder.swift
sameerank, line 1: Borrowed from https://github.com/DataDog/openfeature-js-client/blob/15e44841ecbc3e666e32c1f12c226f4a5aeb3b10/packages/browser/src/transport/endpoint.ts

DatadogFlags/Sources/FlagsHttpClient.swift
sameerank, line 1: Took a bit from https://github.com/DataDog/openfeature-js-client/blob/15e44841ecbc3e666e32c1f12c226f4a5aeb3b10/packages/browser/src/transport/fetchConfiguration.ts

DatadogFlags/Sources/FlagsStore.swift
sameerank, line 1: The current implementation of this store does NOT support multi-client instantiation because they would share the same store in the way that this is set up

https://datadoghq.atlassian.net/browse/FFL-1016

DatadogFlags/Sources/FlagsClient.swift
```diff
@@ -0,0 +1,109 @@
...
+    public func getDoubleValue(key: String, defaultValue: Double) -> Double {
+        let flags = store.getFlags()
+        if let flagData = flags[key] as? [String: Any],
+           let value = flagData["variationValue"] as? NSNumber {
+            return value.doubleValue
+        }
+        return defaultValue
+    }
+
+    public func getObjectValue(key: String, defaultValue: [String: Any]) -> [String: Any] {
```
ncreated, line 128: non-blocking, comment/ `[String: Any]` is a pragmatic first choice, so let’s stick with it for now. That said, we should keep [`OpenFeature.Value`](https://github.com/open-feature/swift-sdk/blob/main/Sources/OpenFeature/Value.swift) in mind - eventually, we’ll want to mirror this container in our SDK to ease migration from Datadog `FlagsClient` to `OpenFeatureClient`. Let’s mark this as a "to do" (maybe add a JIRA?) 👍 and move on, WDYT?

DatadogFlags/Sources/FlagsEndpointBuilder.swift
```diff
@@ -0,0 +1,67 @@
...
+import Foundation
+
+internal struct FlagsEndpointBuilder {
+    /// Builds the complete endpoint URL for precompute-assignments API
+    /// - Parameters:
+    ///   - site: The Datadog site (e.g., "datadoghq.com", "datadoghq.eu")
+    ///   - customerDomain: Optional customer-specific domain prefix
+    /// - Returns: Complete URL string for the flags endpoint
+    /// - Throws: FlagsError.invalidConfiguration if site is unsupported
+    static func buildEndpointURL(site: String, customerDomain: String? = nil) throws -> String {
```
ncreated: change-request/ For `site`, we should rather use the canonical [`DatadogInternal.DatadogSite`](https://github.com/DataDog/dd-sdk-ios/blob/83ec4d43e48086d1843e8929a33f267570ab3633/DatadogInternal/Sources/Context/DatadogSite.swift#L9) enum instead of a plain `String`. This ensures new DCs are compiler-enforced (_`switch` must be exhaustive_).

- Requires `import DatadogInternal`.  
- Don’t rely on [`site.endpoint`](https://github.com/DataDog/dd-sdk-ios/blob/83ec4d43e48086d1843e8929a33f267570ab3633/DatadogInternal/Sources/Context/DatadogSite.swift#L34), since flags endpoints differ we need a dedicated `switch {}` mapping in `DatadogFlags` module.

DatadogFlags/Sources/FlagsEndpointBuilder.swift
```diff
@@ -0,0 +1,67 @@
...
+            return "\(customerDomain).\(baseHost)"
+        }
+
+        return baseHost
+    }
+
+    /// Extracts customer domain from a client token
+    /// - Parameter clientToken: The Datadog client token
+    /// - Returns: Customer domain if extractable, nil otherwise
+    static func extractCustomerDomain(from clientToken: String) -> String? {
```
ncreated, line 60: question/ Interesting 💡! Let’s cover this in the sync meeting 🙏.

DatadogFlags/Sources/FlagsConfiguration.swift
```diff
@@ -0,0 +1,34 @@
...
+import Foundation
+
+public struct FlagsClientConfiguration {
+    public let clientToken: String
+    public let environment: String
+    public let baseURL: String?
+    public let site: String
+    public let applicationId: String?
+    public let customHeaders: [String: String]
+    public let flaggingProxy: String?
```
ncreated: non-blocking, comment/ Nice job here 👍. These parameters are fine for now. Most of this will eventually come automatically from Core - I’ll refine it in my work.

DatadogFlags/Sources/FlagsHttpClient.swift
```diff
@@ -0,0 +1,132 @@
...
+            } else {
+                // JSON serialize non-string values
+                if let data = try? JSONSerialization.data(withJSONObject: value, options: []),
+                   let jsonString = String(data: data, encoding: .utf8) {
+                    return jsonString
+                } else {
+                    return String(describing: value)
+                }
+            }
+        }
```
ncreated: change-request/ Love the instinct to "match the JS client", but Swift gives us better guardrails than `Any` + ad-hoc JSON. Using `context.attributes: [String: Any]` makes serialization brittle and error-prone. It forces manual conversions, can silently produce invalid JSON, and doesn’t reflect the actual API contract. **We can make the attribute type explicit instead of relying on `Any`.**

Few problems with current design:
1. **Silent data loss / invalid JSON**. Non-String values are coerced into strings. If they can’t be JSON-encoded, we fall back to `String(describing:)`, which can emit invalid JSON (e.g., unescaped quotes):
```swift
struct Foo { let bar: String; let buzz: Int }
String(describing: Foo(bar: "abc", buzz: 42))
// → Foo(bar: "abc", buzz: 42)  // not valid JSON
```
2. **Lack of type-safety**. With `[String: Any]`, callers can pass values the API cannot represent (functions, objects, etc.). These problems only surface at runtime.
3. **Contract mismatch with the server**. If the API expects `targeting_attributes` as a `Map<String, String>`, our public API should reflect that. Allowing `Any` suggests "anything goes", which is misleading.

### Recommendation
**Simplest & safest (preferred):** If the API contract is Map<String, String>, we should declare it that way in the public API. This makes the intended usage obvious, prevents callers from accidentally passing unsupported types, and keeps their expectations aligned with what the server actually accepts:
```swift
public class FlagsEvaluationContext {
    public let targetingKey: String
    public let attributes: [String: String]
}
```

**If we need flexibility (later):** we could switch to `[String: Encodable]` and implement a controlled JSON-stringification step. This can be added later in a non-breaking way since `[String: String]` is compatible with `[String: Encodable]`.

DatadogFlags/Sources/FlagsHttpClient.swift
```diff
@@ -0,0 +1,132 @@
...
+                        "name": configuration.environment,
+                        "dd_env": configuration.environment
+                    ],
+                    "subject": [
+                        "targeting_key": context.targetingKey,
+                        "targeting_attributes": stringifiedAttributes
+                    ]
+                ]
+            ]
+        ]
```
ncreated: suggestion/ Building `requestBody` as a dynamic dictionary forfeits type Swift's safety: key typos, wrong shapes, accidental `nil`, or numbers encoded as strings won’t be caught until runtime.

To improve that, we could model the payload with `Encodable` types and use `JSONEncoder.dd.default()` (from `DatadogInternal`) so dates and other nuances match the intake format. This keeps the API contract self-documenting, testable, and refactor-friendly:
```swift
/// Encodable model for the "precompute-assignments-request" payload
struct PrecomputeAssignmentsRequest: Encodable {
    let data: DataContainer

    struct DataContainer: Encodable {
        let type: String
        let attributes: Attributes
    }

    struct Attributes: Encodable {
        let environment: Environment
        let subject: Subject

        struct Environment: Encodable {
            let name: String
            let ddEnv: String

                case name
                case ddEnv = "dd_env"
            }
        }

        struct Subject: Encodable {
            let targetingKey: String
            let targetingAttributes: [String: String]

                case targetingKey = "targeting_key"
                case targetingAttributes = "targeting_attributes"
            }
        }
    }
}

// Build and encode
let body = PrecomputeAssignmentsRequest(...)
let encoder = JSONEncoder.dd.default()
let requestBody = try encoder.encode(body)
request.httpBody = requestBody
```

DatadogFlags/Sources/FlagsHttpClient.swift
```diff
@@ -0,0 +1,132 @@
...
+
+        do {
+            request.httpBody = try JSONSerialization.data(withJSONObject: requestBody)
+        } catch {
+            completion(.failure(error))
+            return
+        }
+
+        URLSession.shared
```
ncreated: blocking/ We should avoid using `URLSession.shared` - it’s not optimized for our needs and should remain dedicated to application code. The SDK should create its own session instance. We can borrow the [Core session setup](https://github.com/DataDog/dd-sdk-ios/blob/c9b7665278b938517089ce5e43c2ff92ab877fc2/DatadogCore/Sources/Core/Upload/URLSessionClient.swift#L14-L31):
```swift
let configuration: URLSessionConfiguration = .ephemeral
configuration.urlCache = nil
let session = URLSession(configuration: configuration)
```

**Note:** Let's skip `configuration.connectionProxyDictionary` for now, but let’s track this in a JIRA task 🙏 (some customers rely on proxies).

DatadogFlags/Sources/FlagsStore.swift
```diff
@@ -0,0 +1,139 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+internal class FlagsStore {
```
ncreated, line 10: non-blocking, comment/ This looks borrowed from [eppo/ConfigurationStore.swift](https://github.com/Eppo-exp/eppo-ios-sdk/blob/main/Sources/eppo/ConfigurationStore.swift), right? Fine for now, but we’ll need to integrate with the Core SDK’s [`DataStore`](https://github.com/DataDog/dd-sdk-ios/blob/c9b7665278b938517089ce5e43c2ff92ab877fc2/DatadogInternal/Sources/DataStore/DataStore.swift#L40-L41). Currently `FlagsStore` uses a single `datadog-flags` directory, but with multiple SDK instances it should use a folder per core instance. Maybe let’s track this in JIRA 🙏 (?).

DatadogFlags/Sources/FlagsStore.swift
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        Self.persistenceQueue.async { [weak self] in
+            guard let self = self else {
+                return
+            }
+            do {
+                var cacheData: [String: Any] = [
+                    "flags": self.cachedFlags
+                ]
```
ncreated, line 80: question/ If multiple `FlagsClient` instances exist, how does `FlagsStore` namespace or partition persisted `"flags"` per client?

DatadogFlags/Sources/FlagsClient.swift
sameerank: Agreed and TODO comment has been added 👍 https://datadoghq.atlassian.net/browse/FFL-1047

DatadogFlags/Sources/FlagsHttpClient.swift
sameerank: Sounds good! Changed `FlagsEvaluationContext.attributes` from `[String: Any]` to `[String: String]` to match the server's `Map<String, String>` contract

DatadogFlags/Sources/FlagsStore.swift
sameerank: Done https://datadoghq.atlassian.net/browse/FFL-1048

DatadogFlags/Sources/FlagsHttpClient.swift
sameerank: 👍 https://datadoghq.atlassian.net/browse/FFL-1049

DatadogFlags/Sources/FlagsHttpClient.swift
sameerank: Appreciate the suggestion! I've refactored `requestBody` to use Swift's types:

https://github.com/DataDog/dd-sdk-ios/blob/707e6cdc87ec617ea79ce39d6cbe045e4d0e41b6/DatadogFlags/Sources/PrecomputeAssignmentsRequest.swift#L12

- Added: `PrecomputeAssignmentsRequest` with nested `Encodable` structs
- Removed: Dynamic `[String: Any]` dictionary building
- Updated: Uses `JSONEncoder.dd.default()` for consistent formatting

DatadogFlags/Sources/FlagsEndpointBuilder.swift
sameerank: https://github.com/DataDog/dd-sdk-ios/pull/2477/commits/228299f9ad93b056d14f4c7faa605aedf485dbd9

DatadogFlags/Sources/FlagsStore.swift
sameerank: I think this is something to solve in follow up work https://datadoghq.atlassian.net/browse/FFL-1016. The idea [in this doc](https://docs.google.com/document/d/1-3F6jAG-sJwuUaB5MeN1PUiaQ2r-JBwy3LHxzMXq-j4/edit?tab=t.0#heading=h.576gaeqdo9z8) to use `clientKey` sounds like a promising strategy to namespace a store to a client.

ncreated (approved):
Great work! 🏅

DatadogFlags/Sources/FlagsConfiguration.swift
```diff
@@ -0,0 +1,35 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+import DatadogInternal
+
+public struct FlagsClientConfiguration {
```
gonzalezreal: ~💅 This could be a nested type (ie. `FlagsClient.Configuration`)~
Oh, I see this is fixed in #2484, nevermind.

DatadogFlags/Sources/FlagsHttpClient.swift
gonzalezreal: 💡 We have the `HTTPClient` protocol and `URLSessionClient` implementation in `DatadogCore`. Can we leverage them to implement the `/precompute-assignments` endpoint?

DatadogFlags/Sources/FlagsHttpClient.swift
```diff
@@ -0,0 +1,129 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+import DatadogInternal
+
+internal protocol FlagsHttpClient {
```
gonzalezreal: 💅🏽 Acronyms that commonly appear as all upper case in American English should be uniformly up- or down-cased according to case conventions (https://www.swift.org/documentation/api-design-guidelines/).

Screenshot 2025-09-19 at 16 07 11

gonzalezreal (commented):
Great foundation work on the `Flags` module! 🎉

Love the clean architecture with proper separation between client, storage, and networking layers. The evaluation logic and caching strategy look solid. 💯

However, I think we could leverage the existing `HTTPClient`/`URLSessionClient` infrastructure for the `/precompute-assignments` endpoint. Let me know your thoughts.

DatadogFlags/Sources/FlagsHttpClient.swift
sameerank: https://github.com/DataDog/dd-sdk-ios/pull/2477/commits/f5ce5d0dce920ae1fae1b1ffd13471c1793dc4a5

DatadogFlags/Sources/FlagsHttpClient.swift
sameerank: https://datadoghq.atlassian.net/browse/FFL-1051

gonzalezreal (approved):

PR #2486, opened 2025-09-19
RUM-11050 Application Launch Profile

maxep:
### What and why?

### How?

### Review checklist


DatadogInternal/Sources/Models/Profiling/ProfilingContext.swift
```diff
@@ -0,0 +1,73 @@
...
+/// RUM might include profiling status in error reports or performance metrics.
+public struct ProfilingContext: AdditionalContext {
+    /// The context key used to identify profiling context in the core context.
+    public static let key = "profiling"
+
+    /// Represents the current status of profiling operations.
+    ///
+    /// This enum maps directly to the underlying C profiler status codes
+    /// from the constructor profiler implementation.
+    public enum Status {
```
simaoseica-dd: Can this status be aligned with the [schema status](https://github.com/DataDog/rum-events-format/blob/master/schemas/rum/_profiling-internal-context-schema.json#L11)?

At least removing the `timedOut`, `prewarmed` and `sampledOut` to a different parameter since they are not status but end reasons.

DatadogInternal/Sources/Models/Profiling/ProfilingMessages.swift
```diff
@@ -0,0 +1,54 @@
...
+/// 3. Serialize it to pprof format
+/// 4. Submit it through the profiling pipeline with the provided context
+///
+/// ## Context Data
+///
+/// The context dictionary can contain any additional metadata about the launch.
+/// The context will be injected into the profile data for query.
+///
+/// This context is included as additional attributes in the final profile event.
+public struct AppLaunchProfileStop {
```
simaoseica-dd: Can this `ProfileStop` be generic enough to not be attached with `AppLaunch`? IIUC, the collection of data cannot be overlapped. Meaning, before a new profiling start, there is a need for a stop call.

DatadogProfiling/Mach/include/ctor_profiler.h
```diff
@@ -0,0 +1,257 @@
...
+ * ```
+ * 
+ * - `AppLaunchProfileSampleRate`: Sample rate percentage (0.0-100.0)
+ *   - 0.0 = profiling disabled
+ *   - 100.0 = always profile
+ *   - Values between 0-100 can be used for probabilistic sampling
+ * 
+ * # Profiling Characteristics
+ * 
+ * - **Timing**: Starts via __attribute__((constructor(101))) - very early in process lifecycle
```
simaoseica-dd: It changed to "65535". It is better to create a constant parameter and reference it in the documentation as a way for it to not be outdated.

DatadogProfiling/Mach/include/ctor_profiler.h
```diff
@@ -0,0 +1,257 @@
...
+ *     // Stop constructor profiling when app launch is complete
+ *     if ctor_profiler_is_active() != 0 {
+ *         ctor_profiler_stop()
+ *     }
+ * }
+ * ```
+ * 
+ * # Performance Considerations
+ * 
+ * - Profiling starts before main() and most static initializers
```
simaoseica-dd:
```suggestion
 * - Profiling starts right before main()
```

DatadogProfiling/Mach/include/ctor_profiler.h
```diff
@@ -0,0 +1,257 @@
...
+    CTOR_PROFILER_STATUS_NOT_STARTED = 0,       ///< Profiler was never started
+    CTOR_PROFILER_STATUS_RUNNING = 1,           ///< Profiler is currently running
+    CTOR_PROFILER_STATUS_STOPPED = 2,           ///< Profiler was stopped manually
+    CTOR_PROFILER_STATUS_TIMEOUT = 3,           ///< Profiler was stopped due to timeout
+    CTOR_PROFILER_STATUS_ERROR = 4,             ///< Profiler encountered an error
+    CTOR_PROFILER_STATUS_PREWARMED = 5,         ///< Profiler was not started due to prewarming
+    CTOR_PROFILER_STATUS_SAMPLED_OUT = 6,       ///< Profiler was not started due to sample rate
+    CTOR_PROFILER_STATUS_ALLOCATION_FAILED = 7, ///< Memory allocation failed
+    CTOR_PROFILER_STATUS_START_FAILED = 8,      ///< Failed to start sampling
+} ctor_profiler_status_t;
```
simaoseica-dd, line 123: Same here. Can we have a new parameter for the end reason?

DatadogProfiling/Sources/AppLaunchProfiler.swift
```diff
@@ -0,0 +1,103 @@
...
+            let event = ProfileEvent(
+                family: "ios",
+                runtime: "ios",
+                version: "4",
+                start: Date(timeIntervalSince1970: start),
+                end: Date(timeIntervalSince1970: end),
+                attachments: [ProfileEvent.Constants.wallFilename],
+                tags: [
+                    "service:\(context.service)",
+                    "version:\(context.version)",
```
simaoseica-dd, line 58: Can it have the `sdk_version` as well?

DatadogProfiling/Sources/AppLaunchProfiler.swift
```diff
@@ -0,0 +1,103 @@
...
+        let pprof = Data(bytes: data, count: size)
+        dd_pprof_free_serialized_data(data)
+
+        core.scope(for: ProfilerFeature.self).eventWriteContext { context, writer in
+            let event = ProfileEvent(
+                family: "ios",
+                runtime: "ios",
+                version: "4",
+                start: Date(timeIntervalSince1970: start),
+                end: Date(timeIntervalSince1970: end),
```
simaoseica-dd, line 54: isn't it better to have the duration to avoid date processing issues on the intake/frontend?

DatadogProfiling/Sources/AppLaunchProfiler.swift
```diff
@@ -0,0 +1,103 @@
...
+
+        let pprof = Data(bytes: data, count: size)
+        dd_pprof_free_serialized_data(data)
+
+        core.scope(for: ProfilerFeature.self).eventWriteContext { context, writer in
+            let event = ProfileEvent(
+                family: "ios",
+                runtime: "ios",
+                version: "4",
+                start: Date(timeIntervalSince1970: start),
```
simaoseica-dd, line 53: This `startDate` should be updated with the NTP time

DatadogProfiling/Mach/ctor_profiler.cpp
```diff
@@ -0,0 +1,244 @@
...
+#ifdef __APPLE__
+
+#include "profile.h"
+#include "mach_sampling_profiler.h"
+
+#include <CoreFoundation/CoreFoundation.h>
+#include <cstdlib>
+#include <cstring>
+#include <random>
+#include <mutex>
```
simaoseica-dd: Do you need to import `mutex`?

DatadogProfiling/Tests/CTorProfilerTests.swift
```diff
@@ -0,0 +1,183 @@
...
+ */
+
+import XCTest
+import DatadogInternal
+import DatadogMachProfiler
+
+final class CTorProfilerTests: XCTestCase {
+    override func setUp() {
+        super.setUp()
+        ctor_profiler_stop()
```
simaoseica-dd: Should not be needed since the `tearDown` takes care of the cleanup.

DatadogProfiling/Mach/ctor_profiler.cpp
maxep: Good catch!

DatadogProfiling/Sources/AppLaunchProfiler.swift
maxep: We can't at this stage, for 2 reasons:
1. The profile samples also include timestamps which are not synced to NTP, the start/end fields need to be synced to samples' timestamp.
2. While it would be possible to sync all samples to NTP, the App Launch profiler starts too early. The NTP offset is simply not available yet.

DatadogProfiling/Sources/AppLaunchProfiler.swift
maxep: Could be, but this is the current expectation of the profiling intake. This event format is applied to all profile families.

DatadogProfiling/Sources/AppLaunchProfiler.swift
maxep: Good idea!

DatadogProfiling/Mach/include/ctor_profiler.h
maxep: From the C layer perspective, it is easier to keep a single status. I will do the change at the Swift layer only 👍

DatadogInternal/Sources/Models/Profiling/ProfilingContext.swift
maxep: I have modified the status to include stop and error reasons. It's not fully aligned with the `_profiling-internal-context-schema` due to our profiler specs. We should update the schema IMO. LMKWYT!

DatadogProfiling/Sources/AppLaunchProfiler.swift
simaoseica-dd: Ok. Let's make it not synced with NTP in this first iteration. TTID will follow the same approach then.

PR #2489, opened 2025-09-23
feat: Name arg for multi-FlagsClient support with data store isolation

sameerank:
### What and why?

This PR implements named instance management for FlagsClient, with a registry pattern inspired by CoreRegistry. The change enables SDK users to create and manage multiple FlagsClient instances with unique names to enable data isolation. 

### How?

- Created `FlagsClientRegistry`: Centralized registry for managing named `FlagsClient` instances with thread-safe operations using `@ReadWriteLock`
- Added `FlagsClient` API:
  - `create()` methods with required naming for each client
  - `instance()` methods to access a named client
- Implemented `NOPFlagsClient`: Safe fallback for error scenarios following NOP pattern
- Updated storage architecture: `FlagsDataStore` with `instanceName` for proper data isolation
- Test coverage: tests covering registry functionality, thread safety, and API behavior

### Review checklist


DatadogFlags/Sources/FlagsClient.swift
```diff
@@ -20,20 +20,126 @@ public class FlagsClient {
...
+    /// - Returns: Default FlagsClient instance if it exists, NOPFlagsClient otherwise.
+    public static func instance() -> FlagsClient {
+        return FlagsClientRegistry.default
+    }
+
+    /// Returns an existing FlagsClient instance by name.
+    ///
+    /// - Parameter name: The name of the instance to retrieve.
+    /// - Returns: FlagsClient instance if it exists, NOPFlagsClient otherwise.
+    public static func instance(named name: String) -> FlagsClient {
```
sameerank: There are a couple things that I would like to do differently from `CoreRegistry`. This is one of them. Core's registry has the following

- `CoreRegistry.default` (property)
- `CoreRegistry.instance(named: String)` (method with required name)
- NO `CoreRegistry.instance()` without parameters

We already have a `FlagsClient.default` property for accessing the main instance, which is similar to `CoreRegistry.default`, but if `name` is an optional argument, IMO I would expect `FlagsClient.instance()` to also return the default. At which point `FlagsClient.default` would be a redundant attribute.

Should we just stick to the patterns that are already in `CoreRegistry` or if it ok for `FlagsClient` to diverge a little, in which case I would keep this method  and remove the `FlagsClient.default` property

DatadogFlags/Sources/FlagsClientRegistry.swift
```diff
@@ -0,0 +1,92 @@
...
+public final class FlagsClientRegistry {
+    /// Returns the default FlagsClient instance if registered, `NOPFlagsClient` instance otherwise.
+    public static var `default`: FlagsClient {
+        instances[defaultInstanceName] ?? NOPFlagsClient()
+    }
+
+    /// The name for the default FlagsClient instance.
+    ///
+    /// Features should use this name as default parameter.
+    public static let defaultInstanceName = "main"
```
sameerank: There are a couple things that I would like to do differently from `CoreRegistry`. This is one of them. And this one is a bit of a nit.

The way this is coded, `defaultInstanceName`'s "main" behaves like a reserved/special name.

If we implement a base name instead of a default name, then the default would just be anonymous, e.g.
- `client`
- `client-firstname`
- `client-secondname`

And so "main" is free to use as a designation for `client-main`. And there's nothing special about naming your instance "main".

On the other hand with this pattern using `defaultInstanceName`, "main" is not free for the user to use as different from the SDK's default flags client. But this conforms more closely with `CoreRegistry` and there's certainly value in consistency, so the right path forward isn't obvious to me: a namespace reserved name (identical to the Core registry) vs letting it be anything (inconsistent with the Core registry) but more free for the customer (though it's just one name 🤷).

DatadogFlags/Sources/FlagsClient.swift
```diff
@@ -20,20 +20,126 @@ public class FlagsClient {
...
+    }
+
+    /// Creates a named FlagsClient instance with custom configuration, or returns existing if one with the same name exists.
+    ///
+    /// - Parameters:
+    ///   - configuration: Custom configuration for the client.
+    ///   - name: The unique name for this instance.
+    ///   - core: The DatadogCore instance to use. Defaults to CoreRegistry.default.
+    /// - Returns: FlagsClient instance (new or existing).
+    public static func createOrGet(with configuration: FlagsClient.Configuration, name: String, in core: DatadogCoreProtocol = CoreRegistry.default) -> FlagsClient {
```
sameerank: We decided on `createOrGet` in our meeting earlier today, but I wonder do we need `.create` and `.instance` and `createOrGet` to all be `public`?

sameerank: I will take this out. This is a confusing API

DatadogFlags/Sources/NOPFlagsClient.swift
```diff
@@ -0,0 +1,74 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+import DatadogInternal
+
+/// A non-functional implementation of FlagsClient that does nothing.
+/// Used as a safe fallback when FlagsClient creation fails.
+internal class NOPFlagsClient: FlagsClient {
```
sameerank: Use a protocol instead. We need a `FlagsClient` protocol first. The `FlagsClient` class should conform to it, and so should `NOPFlagsClient`

DatadogFlags/Sources/FlagsClientProtocol.swift
sameerank, line 1: Following these file patterns

- `LoggerProtocol` + `NOPLogger` → `LoggerProtocol.swift`
- `RUMMonitorProtocol` + `NOPMonitor` → `RUMMonitorProtocol.swift`

DatadogFlags/Sources/FlagsClientProtocol.swift
```diff
@@ -0,0 +1,79 @@
...
+    func getBooleanValue(key: String, defaultValue: Bool) -> Bool
+    func getStringValue(key: String, defaultValue: String) -> String
+    func getIntegerValue(key: String, defaultValue: Int64) -> Int64
+    func getDoubleValue(key: String, defaultValue: Double) -> Double
+    func getObjectValue(key: String, defaultValue: [String: Any]) -> [String: Any]
+}
+
+/// A no-operation implementation of FlagsClient that does nothing.
+/// Used as a safe fallback when FlagsClient creation fails.
+public final class NOPFlagsClient: FlagsClientProtocol {
```
sameerank, line 27: I had to think a bit about whether this should be a class or a struct. Both examples are prevalent in the SDK code

Classes:
1. NOPMonitor - implements RUMMonitorProtocol
2. NOPDatadogCore - implements DatadogCoreProtocol
3. NOPOTelSpan - implements OpenTelemetry Span protocol                                                                            
 
Structs:
1. NOPDataStore - implements DataStore protocol
3. NOPFeatureScope - implements FeatureScope protocol
4. NOPLogger - implements LoggerProtocol protocol
5. NOPTelemetry - implements Telemetry protocol

Reasons for going with a class: to conform to `FlagsClientProtocol: AnyObject`, preserve `.create` and `.instance` return types to be `FlagsClient` (not the protocol) so we can do reference equality for testing (`===` doesn't work for structs), other NOPs that are classes also appear to be client-like. Anything else I should be aware of?

DatadogFlags/Sources/FlagsClient.swift
```diff
@@ -7,33 +7,66 @@
...
-public class FlagsClient {
+public class FlagsClient: FlagsClientProtocol {
     private let configuration: FlagsClient.Configuration
     private let httpClient: FlagsHTTPClient
     private let store: FlagsStore
     private let featureScope: FeatureScope
 
-    internal init(configuration: FlagsClient.Configuration, httpClient: FlagsHTTPClient, store: FlagsStore, featureScope: FeatureScope) {
+    /// The name of this FlagsClient instance
+    public let name: String
```
sameerank, line 17: Saving the name to an instance variable isn't required for multi-client functionality, but it seems like it would be helpful to be able to do `client.name` to identify the client

DatadogFlags/Tests/FlagsClientNamedInstanceTests.swift
```diff
@@ -0,0 +1,156 @@
...
+        let client3 = FlagsClient.create(name: "client3", in: core)
+
+        // Verify each can be retrieved correctly
+        XCTAssertTrue(FlagsClient.instance(named: "client1") === client1)
+        XCTAssertTrue(FlagsClient.instance(named: "client2") === client2)
+        XCTAssertTrue(FlagsClient.instance(named: "client3") === client3)
+
+        // Non-existent instance should return NOP
+        let nonExistent = FlagsClient.instance(named: "nonexistent")
+        XCTAssertTrue(nonExistent is NOPFlagsClient)
```
sameerank, line 77: `nonExistent` is eligible for `===` comparisons, but we don't have a way to check equality with another `NOPFlagsClient` instance because if we generate one, it'll be a different instance

One option I thought about was to do something like `XCTAssertTrue(nonExistent === NOPFlagsClient.shared)` where `NOPFlagsClient.shared` would reference a singleton NOP instance. But that breaks the expectation that `FlagsClient.create` and `FlagsClient.instance` don't return singletons, although one could argue that named clients are singletons within the scope of their name

DatadogFlags/Sources/FlagsClientRegistry.swift
sameerank: By naming `name` required, we can just use the client name as-is and there is no longer any reserved name for the default.

DatadogFlags/Sources/FlagsClient.swift
sameerank: By making `name` required, there's no longer a need to have a "default"

DatadogFlags/Sources/NOPFlagsClient.swift
sameerank: Implemented

DatadogFlags/Sources/FlagsClientRegistry.swift
sameerank, line 1: Based on the implementation of [CoreRegistry](https://github.com/DataDog/dd-sdk-ios/blob/922bba32b92430e04d650f493c6ab58e4c1e6f72/DatadogInternal/Sources/CoreRegistry.swift)

The main difference is that there's no default name

https://github.com/DataDog/dd-sdk-ios/blob/922bba32b92430e04d650f493c6ab58e4c1e6f72/DatadogInternal/Sources/CoreRegistry.swift#L20

This is because we are requiring a name for each client so there's no concept of a default. This also means we don't have any reserved client names, whereas the core registry has the "main" name reserved as the default

gonzalezreal:
Replaced by #2497

PR #2491, opened 2025-09-25
FFL-1022: Send feature flag exposure events to intake

gonzalezreal:
### What and why?

This PR implements the infrastructure for sending exposure events when feature flags are accessed. When users evaluate feature flags through the `FlagsClient`, exposure events are now automatically logged and sent to the corresponding intake endpoint for analytics and monitoring purposes.

This PR implements exposure event logging for the `DatadogFlags` module. When users evaluate feature flags through the `FlagsClient`, exposure events are now automatically logged and sent to the corresponding intake endpoint for analytics and monitoring.

The PR also includes several foundational improvements to support type-safe flag handling and performance optimizations.

#### Exposure Event Infrastructure (FFL-1022)

- `ExposureEvent` model: Complete event structure with allocation, flag, variant, and subject information following the schema used in the [Javascript client](https://github.com/DataDog/openfeature-js-client/blob/ec575fbcb21c1578df4c7c303c06b5f5974c5047/packages/core/src/configuration/exposureEvent.types.ts).
- `ExposureLogger`: Automatic exposure logging when flags are evaluated, integrating with the SDK's event persistence system.
- `ExposureRequestBuilder`: Enhanced to handle batched exposure events using newline-delimited JSON format, with proper headers and endpoint.
- Updated `FlagsFeature` to support batching (up to 50 events) instead of single-event sends

#### Type-Safe Flag System (FFL-1070, FFL-1047)
- `AnyValue` implementation: Comprehensive type system for handling arbitrary flag values with proper JSON encoding/decoding.
- `FlagValue`, `FlagAssignment` and `FlagAssignmentsResponse`: Type-safe models for flag evaluation results and precomputed assignments
- Enhanced `FlagsClient`: Added generic `getValue` method with type safety and improved flag evaluation logic
- Cleaned up and simplified type definitions

#### Performance & Thread Safety (FFL-1052)
- `ReadWriteLock` optimization: Replaced synchronous queue usage with `ReadWriteLock` in `FlagsStore` for better performance

### Review checklist

### DIFF
--- a/DatadogFlags/Sources/Feature/ExposureLogger.swift
+++ b/DatadogFlags/Sources/Feature/ExposureLogger.swift
@@ -8,7 +8,12 @@ import Foundation
 import DatadogInternal
 
 internal protocol ExposureLogging {
-    func logExposure(at date: Date, for flagKey: String, flagAssignment: FlagAssignment)
+    func logExposure(
+        at date: Date,
+        for flagKey: String,
+        assignment: FlagAssignment,
+        context: FlagsEvaluationContext
+    )
 }
 
 internal final class ExposureLogger: ExposureLogging {
@@ -19,21 +24,20 @@ internal final class ExposureLogger: ExposureLogging {
         let variationKey: String
     }
 
-    private let flagsEvaluationContext: FlagsEvaluationContext
-    private let featureScope: FeatureScope
-
+    private let featureScope: any FeatureScope
     private var loggedExposures: Set<Exposure> = []
 
-    init(
-        flagsEvaluationContext: FlagsEvaluationContext,
-        featureScope: FeatureScope
-    ) {
+    init(featureScope: any FeatureScope) {
         self.featureScope = featureScope
-        self.flagsEvaluationContext = flagsEvaluationContext
     }
 
-    func logExposure(at date: Date, for flagKey: String, flagAssignment: FlagAssignment) {
-        guard flagAssignment.doLog else {
+    func logExposure(
+        at date: Date,
+        for flagKey: String,
+        assignment: FlagAssignment,
+        context: FlagsEvaluationContext
+    ) {
+        guard assignment.doLog else {
             return
         }
 
@@ -43,10 +47,10 @@ internal final class ExposureLogger: ExposureLogging {
             }
 
             let exposure = Exposure(
-                targetingKey: flagsEvaluationContext.targetingKey,
+                targetingKey: context.targetingKey,
                 flagKey: flagKey,
-                allocationKey: flagAssignment.allocationKey,
-                variationKey: flagAssignment.variationKey
+                allocationKey: assignment.allocationKey,
+                variationKey: assignment.variationKey
             )
 
             guard !loggedExposures.contains(exposure) else {
@@ -56,12 +60,12 @@ internal final class ExposureLogger: ExposureLogging {
 
             let exposureEvent = ExposureEvent(
                 timestamp: date.timeIntervalSince1970.toInt64Milliseconds,
-                allocation: .init(key: flagAssignment.allocationKey),
+                allocation: .init(key: assignment.allocationKey),
                 flag: .init(key: flagKey),
-                variant: .init(key: flagAssignment.variationKey),
+                variant: .init(key: assignment.variationKey),
                 subject: .init(
-                    id: flagsEvaluationContext.targetingKey,
-                    attributes: flagsEvaluationContext.attributes
+                    id: context.targetingKey,
+                    attributes: context.attributes
                 )
             )
 
--- a/DatadogFlags/Sources/Feature/ExposureLogger.swift
+++ b/DatadogFlags/Sources/Feature/ExposureLogger.swift
@@ -7,11 +7,65 @@
 import Foundation
 import DatadogInternal
 
-internal final class ExposureLogger {
+internal protocol ExposureLogging {
+    func logExposure(at date: Date, for flagKey: String, flagAssignment: FlagAssignment)
+}
+
+internal final class ExposureLogger: ExposureLogging {
     private struct Exposure: Hashable {
         let targetingKey: String
         let flagKey: String
         let allocationKey: String
         let variationKey: String
+    }
+
+    private let flagsEvaluationContext: FlagsEvaluationContext
+    private let featureScope: FeatureScope
+
+    private var loggedExposures: Set<Exposure> = []
+
+    init(
+        flagsEvaluationContext: FlagsEvaluationContext,
+        featureScope: FeatureScope
+    ) {
+        self.featureScope = featureScope
+        self.flagsEvaluationContext = flagsEvaluationContext
+    }
+
+    func logExposure(at date: Date, for flagKey: String, flagAssignment: FlagAssignment) {
+        guard flagAssignment.doLog else {
+            return
+        }
+
+        featureScope.eventWriteContext { [weak self] _, writer in
+            guard let self else {
+                return
+            }
+
+            let exposure = Exposure(
+                targetingKey: flagsEvaluationContext.targetingKey,
+                flagKey: flagKey,
+                allocationKey: flagAssignment.allocationKey,
+                variationKey: flagAssignment.variationKey
+            )
+
+            guard !loggedExposures.contains(exposure) else {
+                return
+            }
+            loggedExposures.insert(exposure)
+
+            let exposureEvent = ExposureEvent(
+                timestamp: date.timeIntervalSince1970.toInt64Milliseconds,
+                allocation: .init(key: flagAssignment.allocationKey),
+                flag: .init(key: flagKey),
+                variant: .init(key: flagAssignment.variationKey),
+                subject: .init(
+                    id: flagsEvaluationContext.targetingKey,
+                    attributes: flagsEvaluationContext.attributes
+                )
+            )
+
+            writer.write(value: exposureEvent)
+        }
     }
 }
--- a/DatadogFlags/Sources/Feature/ExposureLogger.swift
+++ b/DatadogFlags/Sources/Feature/ExposureLogger.swift
@@ -41,7 +41,7 @@ internal final class ExposureLogger: ExposureLogging {
             return
         }
 
-        featureScope.eventWriteContext { [weak self] _, writer in
+        featureScope.eventWriteContext { [weak self] ddContext, writer in
             guard let self else {
                 return
             }
@@ -58,8 +58,9 @@ internal final class ExposureLogger: ExposureLogging {
             }
             loggedExposures.insert(exposure)
 
+            let adjustedDate = date.addingTimeInterval(ddContext.serverTimeOffset)
             let exposureEvent = ExposureEvent(
-                timestamp: date.timeIntervalSince1970.toInt64Milliseconds,
+                timestamp: adjustedDate.timeIntervalSince1970.toInt64Milliseconds,
                 allocation: .init(key: assignment.allocationKey),
                 flag: .init(key: flagKey),
                 variant: .init(key: assignment.variationKey),
--- a/DatadogFlags/Sources/Feature/ExposureLogger.swift
+++ b/DatadogFlags/Sources/Feature/ExposureLogger.swift
@@ -5,11 +5,13 @@
  */
 
 import Foundation
+import DatadogInternal
 
-public protocol FlagValue {}
-
-extension Bool: FlagValue {}
-extension String: FlagValue {}
-extension Int: FlagValue {}
-extension Double: FlagValue {}
-extension AnyValue: FlagValue {}
+internal final class ExposureLogger {
+    private struct Exposure: Hashable {
+        let targetingKey: String
+        let flagKey: String
+        let allocationKey: String
+        let variationKey: String
+    }
+}
--- a/DatadogFlags/Sources/FlagsClient.swift
+++ b/DatadogFlags/Sources/FlagsClient.swift
@@ -11,16 +11,30 @@ public class FlagsClient {
     private let configuration: FlagsClient.Configuration
     private let httpClient: FlagsHTTPClient
     private let store: FlagsStore
-    private let featureScope: FeatureScope
+    private let exposureLogger: any ExposureLogging
+    private let dateProvider: any DateProvider
+    private let featureScope: any FeatureScope
 
-    internal init(configuration: FlagsClient.Configuration, httpClient: FlagsHTTPClient, store: FlagsStore, featureScope: FeatureScope) {
+    internal init(
+        configuration: FlagsClient.Configuration,
+        httpClient: FlagsHTTPClient,
+        store: FlagsStore,
+        exposureLogger: any ExposureLogging,
+        dateProvider: any DateProvider,
+        featureScope: any FeatureScope
+    ) {
         self.configuration = configuration
         self.httpClient = httpClient
         self.store = store
+        self.exposureLogger = exposureLogger
+        self.dateProvider = dateProvider
         self.featureScope = featureScope
     }
 
-    public static func create(with configuration: FlagsClient.Configuration, in core: DatadogCoreProtocol = CoreRegistry.default) -> FlagsClient {
+    public static func create(
+        with configuration: FlagsClient.Configuration,
+        in core: DatadogCoreProtocol = CoreRegistry.default
+    ) -> FlagsClient {
         do {
             // To ensure the correct registration order between Core and Features,
             // the entire initialization flow is synchronized on the main thread.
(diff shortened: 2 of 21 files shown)


DatadogFlags/Sources/Models/FlagAssignment.swift
```diff
@@ -0,0 +1,105 @@
...
+        case "STRING":
+            self.variation = try .string(container.decode(String.self, forKey: .variationValue))
+        case "NUMBER":
+            // Check integer first, then double
+            if let number = try? container.decode(Int.self, forKey: .variationValue) {
+                self.variation = .integer(number)
+            } else {
+                self.variation = try .double(container.decode(Double.self, forKey: .variationValue))
+            }
+        case "OBJECT", "JSON":
```
gonzalezreal: We need to confirm if it is `OBJECT` or `JSON`. The OpenAPI specs says `JSON` but the test endpoint responses are using `OBJECT`.

DatadogFlags/Sources/Models/FlagAssignmentsResponse.swift
```diff
@@ -0,0 +1,27 @@
...
+        case flags
+    }
+
+    init(from decoder: any Decoder) throws {
+        let rootContainer = try decoder.container(keyedBy: CodingKeys.self)
+        let dataContainer = try rootContainer.nestedContainer(keyedBy: CodingKeys.self, forKey: .data)
+        let attributesContainer = try dataContainer.nestedContainer(keyedBy: CodingKeys.self, forKey: .attributes)
+
+        self.flags = try attributesContainer.decode([String: FlagAssignment].self, forKey: .flags)
+    }
```
gonzalezreal, line 26: We do have a [Swift Package](https://github.com/DataDog/swift-jsonapi) for decoding [`{json:api}`](https://jsonapi.org/) responses, but it seems overkill for this use case.

DatadogFlags/Tests/Feature/ExposureLoggerTests.swift
```diff
@@ -0,0 +1,187 @@
...
+            at: .mockAny(),
+            for: .mockAny(),
+            assignment: .mockAnyBoolean(doLog: false),
+            context: .mockAny()
+        )
+
+        // Then
+        XCTAssertTrue(featureScope.eventsWritten.isEmpty)
+    }
+
```
gonzalezreal, line 64: Deduplication behavior tests are the interesting ones 👇🏽

DatadogFlags/Tests/FlagsClientTests.swift
```diff
@@ -38,29 +47,41 @@ final class FlagsClientTests: XCTestCase {
...
         }
 
         waitForExpectations(timeout: 1.0)
+
+        let boolValue = client.getBooleanValue(key: "boolean-flag", defaultValue: false)
+        let stringValue = client.getStringValue(key: "string-flag", defaultValue: "default")
+        let integerValue = client.getIntegerValue(key: "integer-flag", defaultValue: 0)
+        let doubleValue = client.getDoubleValue(key: "numeric-flag", defaultValue: 0.0)
+        let objectValue = client.getObjectValue(key: "json-flag", defaultValue: .null)
+        let typedObjectValue = client.getObjectValue(key: "json-flag", defaultValue: Model(key: "null", prop: 0))
```
gonzalezreal: `AnyValue` is also giving us support for arbitrary `Decodable` types as feature flags values.

DatadogFlags/Tests/FlagsClientTests.swift
```diff
@@ -129,22 +152,28 @@ private class MockFlagsHTTPClient: FlagsHTTPClient {
 }
 
 private class MockFlagsStore: FlagsStore {
-    private var flags: [String: Any] = [:]
+    private struct MockState {
+        var flags: [String: FlagAssignment]
+        var context: FlagsEvaluationContext
+        var date: Date
+    }
```
gonzalezreal, line 182: I will refactor this to use protocol based injection instead of subclassing, in a future PR.

DatadogFlags/Tests/Resources/precomputed-v1.json
```diff
@@ -55,7 +55,7 @@
         "json-flag": {
           "allocationKey": "allocation-127",
           "variationKey": "variation-127",
-          "variationType": "OBJECT",
+          "variationType": "JSON",
```
gonzalezreal, line 58: See discussion [here](https://datadoghq.atlassian.net/wiki/spaces/PANA/pages/5498404973/Flag+Evaluation+API+Precomputed+Assignments?focusedCommentId=5535793777) (internal link).

DatadogFlags/Sources/Models/AnyValue.swift
```diff
@@ -0,0 +1,73 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+public enum AnyValue: Equatable {
```
gonzalezreal, line 9: I just realized we could use `AnyCodable` instead. Is it OK to use `DatadogInternal` types in public APIs? /cc @ncreated

sameerank: Or would it be possible/helpful to make a public `AnyValue` typealias/wrapper that internally uses `AnyCodable` so we are in control of what we expose?

DatadogFlags/Sources/Models/FlagAssignment.swift
sameerank: Just confirmed that the endpoint returns "JSON". Let's remove "OBJECT"

Screenshot 2025-09-25 at 6 44 52 PM

DatadogFlags/Tests/Feature/ExposureLoggerTests.swift
sameerank: I'd consider comparing with [tests in the Eppo SDK](https://github.com/Eppo-exp/eppo-ios-sdk/blob/efdd1b19f66878d82a9558bc19883924800503ef/Tests/eppo/AssignmentCacheTests.swift).

Some ideas
- `testLogExposureOscillatingAssignments`: switch back and forth between two different assignments for same flag or mixed batching, e.g. A, A, B, A, C, B → should result in 3 unique exposures. Might be a little redundant though with `testLogExposureDifferentVariation` and `testLogExposureDifferentAllocation` so it's up to you
- `testLogExposureConcurrentDeduplication`: can verify only one exposure is logged despite concurrent access
- `testLogExposureContextAttributesIndependence`: the inverse of `testLogExposureDifferentTargeting`, should be deduplicated (only targeting key matters in Exposure hash)

DatadogFlags/Sources/FlagsStore.swift
```diff
@@ -5,12 +5,23 @@
...
 
 import Foundation
+import DatadogInternal
 
 // TODO: FFL-1048 Use the Core SDK’s DataStore instead
 internal class FlagsStore {
-    private var cachedFlags: [String: Any] = [:]
-    private var flagsMetadata: FlagsMetadata?
-    private let syncQueue = DispatchQueue(label: "com.datadoghq.flags.store", attributes: .concurrent)
+    private struct State: Codable {
```
sameerank, line 12: Ah, so we would store everything in the same `State`. That's neater than storing flag assignments and metadata separately 👍

DatadogFlags/Tests/Resources/precomputed-v1.json
sameerank: Yup, let's make this change

DatadogFlags/Tests/Feature/FlagsClientMocks.swift
```diff
@@ -0,0 +1,121 @@
...
+    static func mockAnyBoolean(doLog: Bool = true) -> FlagAssignment {
+        .init(
+            allocationKey: .mockAny(),
+            variationKey: .mockAny(),
+            variation: .mockAnyBoolean(),
+            doLog: doLog
+        )
+    }
+
+    static func mockAnyString(doLog: Bool = true) -> FlagAssignment {
```
sameerank, line 48: Some of these (`mockAnyString`, `mockAnyInteger`, `mockAnyDouble`, `mockAnyObject`) appear to be unused, but I guess they're here for completeness

DatadogFlags/Sources/Feature/ExposureLogger.swift
```diff
@@ -0,0 +1,75 @@
...
+internal final class ExposureLogger: ExposureLogging {
+    private struct Exposure: Hashable {
+        let targetingKey: String
+        let flagKey: String
+        let allocationKey: String
+        let variationKey: String
+    }
+
+    private let featureScope: any FeatureScope
+    private var loggedExposures: Set<Exposure> = []
```
sameerank, line 28: Should we be concerned that the cache can grow without bound?

For comparison, `AssignmentCache` in [Eppo's iOS SDK](https://github.com/Eppo-exp/eppo-ios-sdk/blob/efdd1b19f66878d82a9558bc19883924800503ef/Sources/eppo/AssignmentCache.swift) didn't have any mechanisms for controlling the cache size either and it hasn't caused any issues.

sameerank (approved):
Mostly suggestions or questions, but overall looks good to me!

DatadogFlags/Sources/Feature/ExposureLogger.swift
```diff
@@ -0,0 +1,75 @@
...
+                variationKey: assignment.variationKey
+            )
+
+            guard !loggedExposures.contains(exposure) else {
+                return
+            }
+            loggedExposures.insert(exposure)
+
+            let exposureEvent = ExposureEvent(
+                timestamp: date.timeIntervalSince1970.toInt64Milliseconds,
```
ncreated: suggestion/ The `date` is in device time, which doesn't include NTP offset. To ensure this date is real we need: `date.addingTimeInterval(context.serverTimeOffset)`

Reference: [creating `Log` event](https://github.com/DataDog/dd-sdk-ios/blob/5aaa700365ff6a6e0dbd5e2ab273ac0173c36960/DatadogLogs/Sources/Log/LogEventBuilder.swift#L59C19-L59C68).

DatadogFlags/Sources/FlagsClient.swift
```diff
@@ -53,84 +78,82 @@ public class FlagsClient {
...
-        if let flagData = flags[key] as? [String: Any],
-           let value = flagData["variationValue"] as? [String: Any] {
-            return value
+    @inlinable
+    public func getObjectValue(key: String, defaultValue: AnyValue) -> AnyValue {
+        getValue(key: key, defaultValue: defaultValue)
+    }
+
+    @inlinable
+    public func getObjectValue<T: Decodable>(key: String, defaultValue: T, using decoder: JSONDecoder = .init()) -> T {
```
ncreated: question/ Isn't `decoder: JSONDecoder = .init()` allocating new JSONDecoder each time? That might be expensive, no?

DatadogFlags/Sources/FlagsStore.swift
```diff
@@ -40,61 +51,26 @@ internal class FlagsStore {
...
-                            "targetingKey": context.targetingKey,
-                            "attributes": context.attributes
-                        ]
-                    }
-
-                    cacheData["metadata"] = metadataDict
-                }
-
-                let data = try JSONSerialization.data(withJSONObject: cacheData, options: [])
+                let data = try JSONEncoder().encode(state)
```
ncreated, line 73: suggestion/ The `JSONEncoder` / `Decoder` allocation is unnecessary. Better to store it on a property, [like we do in RUM](https://github.com/DataDog/dd-sdk-ios/blob/5aaa700365ff6a6e0dbd5e2ab273ac0173c36960/DatadogRUM/Sources/Feature/RUMDataStore.swift#L38-L41), WDYT?

DatadogFlags/Tests/Feature/ExposureLoggerTests.swift
gonzalezreal: I don't think we need to test concurrent access, since this is guaranteed by the feature scope, which has its own unit tests.

I think the tests in place already cover all the cases you mentioned, as `testLogExposureDeduplication` covers the deduplication, while the rest cover what happens when you change any of the deduplication keys.

DatadogFlags/Sources/Models/AnyValue.swift
gonzalezreal: I discussed this offline with @ncreated and we decided to leave `AnyValue` for our use case. We think that the type specification it provides is better suited for public APIs.

DatadogFlags/Sources/FlagsStore.swift
gonzalezreal: Exactly. On top of that, it avoids impossible states, like non-empty flag assignments with a `nil` evaluation context.

DatadogFlags/Sources/Feature/ExposureLogger.swift
gonzalezreal: That is a fair point. I think we are fine for this first iteration, but we should definitely consider this in the future. Happy to hear any suggestions.

DatadogFlags/Sources/FlagsClient.swift
gonzalezreal: 🤔 I don't think we can use a static `JSONDecoder` instance as a default parameter. As far as I know, `JSONDecoder` is not thread-safe. So, unless we create a static wrapper with a lock, the safest approach is to create a new instance. The time it takes to create a `JSONDecoder` instance is negligible, compared to the time it takes decoding a JSON.

We still have the option to remove this API completely. In the end it is only a nice convenience and not required by our product.

DatadogFlags/Sources/FlagsStore.swift
gonzalezreal: See https://github.com/DataDog/dd-sdk-ios/pull/2491#discussion_r2382139159

DatadogFlags/Sources/Feature/ExposureLogger.swift
gonzalezreal: Done!

DatadogFlags/Sources/FlagsClient.swift
```diff
@@ -53,84 +78,82 @@ public class FlagsClient {
...
+        let anyValue = getValue(key: key, defaultValue: AnyValue.null)
+        let value = try? anyValue.as(T.self, using: decoder)
```
ncreated: question/ The plan is to handle these errors in later `getDetails()` implementation, right?

DatadogFlags/Sources/FlagsClient.swift
ncreated: > We still have the option to remove this API completely. In the end it is only a nice convenience and not required by our product.

Good point. Do we actually have a use case for passing a custom [`JSONDecoder`](https://developer.apple.com/documentation/foundation/jsondecoder)?  

In general-purpose scenarios, it’s useful (e.g. enforcing a custom `dataDecodingStrategy` or handling `Date` formats). But in our case, since all data comes pre-encoded from the Datadog backend and we only decode it, I don’t see a practical reason for customization. Do we?

> As far as I know, `JSONDecoder` is not thread-safe (...), the safest approach is to create a new instance.

Makes sense 👍. I did notice some internal locking/synchronization in [its implementation](https://github.com/swiftlang/swift-foundation/blob/57b6c0c3e56e4d2627ebe02a3d311c2569ded61a/Sources/FoundationEssentials/JSON/JSONDecoder.swift), but without clear guidance from the Swift docs, creating a fresh instance seems like the safest and most predictable strategy.

DatadogFlags/Sources/FlagsStore.swift
```diff
@@ -109,34 +85,7 @@ internal class FlagsStore {
...
-                }
-            } else {
-                // Handle legacy cache format (just flags without metadata)
-                if let flags = try JSONSerialization.jsonObject(with: data, options: []) as? [String: Any] {
-                    self.cachedFlags = flags
-                }
-            }
+            state = try JSONDecoder().decode(State.self, from: data)
         } catch {
             print("No flags found on disk or error decoding: \(error)")
```
ncreated, line 90: improvement/ This should be logged to `featureScope.telemetry`. We need observability for flag encoding/decoding, including both `saveToDisk()` and `loadFromDisk()`. It’s fine to track this as a separate improvement (JIRA) 👌.

DatadogFlags/Sources/Feature/ExposureLogger.swift
ncreated: Wouldn’t tracking an exposure event for a given `flagKey` imply that any previous exposures for that `flagKey` can be dropped from the cache? That way, only the latest exposure per `flagKey` would be stored.

If the logic is more complex, I’d suggest moving this to a separate JIRA task and handling it as an independent improvement to not block this PR. This should also be aligned with the Android team.

ncreated (approved):
Great work 💪 🎯. I left few comments and suggestions, but nothing blocking.

DatadogFlags/Sources/FlagsClient.swift
gonzalezreal: Yes. There are two possible error states: `flagNotFound` and `typeMismatch`.

DatadogFlags/Sources/FlagsStore.swift
gonzalezreal: Yes. Proper handling for this case should be part of using the `DataStore` to store the flag assignments.

DatadogFlags/Sources/FlagsClient.swift
gonzalezreal: I will remove this API as part of FFL-1065. We don't really need it, so why bother with these details.

PR #2497, opened 2025-09-30
FFL-1016: Add support for multiple named clients

gonzalezreal:
### What and why?

This PR implements support for named `FlagsClient` instances, enabling SDK users to create and manage multiple independent feature flag clients with isolated data storage. This work continues the implementation started in #2489.

### How?

Core Implementation:
- Introduced `FlagsClientProtocol` to define the public interface for flag evaluation and define a consistent contract between `FlagsClient` and `NOPFlagsClient`
- Added `FlagsClient.create(name:with:in:)` to create new named client instances
- Added `FlagsClient.instance(named:in:)` to retrieve existing named clients from the registry
- Implemented `FlagsClientRegistry` to manage multiple client instances with thread-safe registration and retrieval

API Design:
- Default client name is `"default"` when no name is specified
- Creating a duplicate client returns `NOPFlagsClient` and logs an error
- Retrieving a non-existent client returns `NOPFlagsClient` (graceful degradation)
- Both methods default to `CoreRegistry.default` for the SDK instance parameter

Testing:
- Added unit tests for `FlagsClient.create(name:with:in:)` covering happy paths, duplicate detection, and error cases
- Added unit tests for `FlagsClient.instance(named:in:)` verifying client retrieval, identity, and error handling

### Review checklist

### DIFF
--- a/DatadogFlags/Sources/Client/FlagsClient.swift
+++ b/DatadogFlags/Sources/Client/FlagsClient.swift
@@ -61,7 +61,12 @@ public class FlagsClient {
                     description: "Flags feature must be enabled before calling `FlagsClient.instance(named:in:)`."
                 )
             }
-            return clientRegistry.client(named: name)
+            guard let client = clientRegistry.client(named: name) else {
+                throw ProgrammerError(
+                    description: "Flags client '\(name)' not found. Make sure that you call `FlagsClient.create(name:with:in:)` first."
+                )
+            }
+            return client
         } catch let error {
             consolePrint("\(error)", .error)
             return NOPFlagsClient()
--- a/DatadogFlags/Sources/Client/FlagsClient.swift
+++ b/DatadogFlags/Sources/Client/FlagsClient.swift
@@ -8,6 +8,8 @@ import Foundation
 import DatadogInternal
 
 public class FlagsClient {
+    public static let defaultName = "default"
+
     private let configuration: FlagsClient.Configuration
     private let httpClient: FlagsHTTPClient
     private let repository: any FlagsRepositoryProtocol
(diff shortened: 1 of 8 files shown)


DatadogFlags/Sources/Client/FlagsClient.swift
```diff
@@ -134,64 +164,3 @@ public class FlagsClient {
...
-    @inlinable
-    public func getDoubleDetails(key: String, defaultValue: Double) -> FlagDetails<Double> {
-        getDetails(key: key, defaultValue: defaultValue)
-    }
-
-    @inlinable
-    public func getObjectDetails(key: String, defaultValue: AnyValue) -> FlagDetails<AnyValue> {
-        getDetails(key: key, defaultValue: defaultValue)
-    }
-}
```
gonzalezreal, line 197: Moved to `FlagsClientProtocol` as extension methods

DatadogFlags/Sources/Client/FlagsClientRegistry.swift
```diff
@@ -0,0 +1,34 @@
...
+        clients[clientName] != nil
+    }
+
+    @discardableResult
+    func unregisterClient(named name: String) -> FlagsClientProtocol? {
+        clients.removeValue(forKey: name)
+    }
+
+    func client(named name: String) -> FlagsClientProtocol {
+        clients[name] ?? NOPFlagsClient()
```
ncreated: suggestion, non-blocking/ It might be cleaner for this API to return `nil` rather than NOP implementation. The defaulting pattern only makes sense if the registry is consumed in a context where a NOP implementation is not available, but that isn’t the case here (nor for [`CoreRegistry`](https://github.com/DataDog/dd-sdk-ios/blob/bc2e7718e5d0c862829c992c750359eb3591f9c5/DatadogInternal/Sources/CoreRegistry.swift#L77), where a similar suggestion could apply).

DatadogFlags/Sources/Client/FlagsClientProtocol.swift
```diff
@@ -0,0 +1,106 @@
...
+        _: FlagsEvaluationContext,
+        completion: @escaping (Result<Void, FlagsError>) -> Void
+    ) {
+        warn()
+        completion(.failure(.clientNotInitialized))
+    }
+
+    func getDetails<T>(key: String, defaultValue: T) -> FlagDetails<T> where T: Equatable, T: FlagValue {
+        warn()
+        return FlagDetails(key: key, value: defaultValue)
```
ncreated: spec/ According to the specification, we should mark this case with an "invalid client" error, allowing upstream implementations (e.g., OpenFeature) to handle it gracefully.

> **Flag detail APIs** → return the default flag value with an "invalid client" error.

ncreated (approved):
Well done and super clean 🎯 👍

I left one feedback on spec alignment, otherwise LGTM 👍.

DatadogFlags/Sources/Client/FlagsClientProtocol.swift
```diff
@@ -0,0 +1,106 @@
...
+import Foundation
+import DatadogInternal
+
+public protocol FlagsClientProtocol: AnyObject {
+    func setEvaluationContext(
+        _ context: FlagsEvaluationContext,
+        completion: @escaping (Result<Void, FlagsError>) -> Void
+    )
+
+    func getDetails<T>(key: String, defaultValue: T) -> FlagDetails<T> where T: Equatable, T: FlagValue
```
sameerank, line 16: Just an FYI, we may add an argument for evaluation hooks later

```
    func getDetails<T: AllowedFlagValueType>(
        key: String, defaultValue: T, options: FlagEvaluationOptions
    ) -> FlagDetails<T>
```
https://github.com/open-feature/swift-sdk/blob/main/Sources/OpenFeature/FlagEvaluationOptions.swift

sameerank (approved):
Nice work!

DatadogFlags/Sources/Client/FlagsClientProtocol.swift
gonzalezreal: Good catch! I updated the implementation and added proper unit tests for `NOPFlagsClient`

ncreated (approved):
👌

PR #2506, opened 2025-10-03
[RUM-11710] Extend resource handling to support multiple MIME types

cdn34dd:
### What and why?

Extend resource handling to support multiple MIME types beyond PNGs to allow for SVG upload in React Native.

### Review checklist


gonzalezreal (approved):

TestUtilities/Sources/Mocks/DatadogSessionReplay/ResourceMocks.swift
```diff
@@ -36,7 +36,7 @@ extension EnrichedResource: RandomMockable, AnyMockable {
         return .init(
             identifier: .mockRandom(),
             data: .mockRandom(),
-            mimeType: .mockRandom(),
+            mimeType: "image/png",
```
mariedm: This seems wrong. Isn't it the point that `EnrichedResource` can take other types than png now?

cdn34dd: Tests were failing if I recall correctly and the reason was that some of them relied on having 'image/png' as th emimeType

PR #2502, opened 2025-10-03
FFL-1017: `DatadogFlags` RUM integration

gonzalezreal:
### What and why?

This PR implements RUM integration for the `DatadogFlags` module, enabling flag evaluations and exposures to be logged to Datadog RUM. This matches the functionality of the JavaScript flags client.

Goals:
- Add flag evaluation tracking - Log every flag evaluation to RUM using `addFeatureFlagEvaluation()`, allowing flag values to be visible in RUM sessions
- Add exposure action logging - Log detailed exposure metadata to RUM as custom `__dd_exposure` actions, including allocation keys, variant keys, subject attributes, and timestamps
- No deduplication - Unlike the existing `ExposureLogger`, RUM integration logs every evaluation to match JavaScript client behavior

### How?

The implementation follows the existing message bus pattern.

```
FlagsClient.getDetails()
    ↓
ExposureLogger (existing) → Exposure intake (deduplicated)
    ↓
RUMExposureLogger (new) → Message Bus → RUM (no deduplication)
```

Key components:

- Message Types
   - `RUMFlagEvaluationMessage` - Carries flag key and evaluated value
   - `RUMFlagExposureMessage` - Carries detailed exposure metadata (timestamp, allocation key, variant key, subject attributes, etc.)
- `RUMExposureLogger`:
   - Sends both flag evaluation and exposure messages via message bus
   - Applies server time offset to timestamps
   - No deduplication (fires on every call)
- Message Receivers:
   - `FlagEvaluationReceiver` - Receives flag evaluation messages and calls `monitor.addFeatureFlagEvaluation()`
   - `FlagExposureReceiver` - Receives exposure messages and calls `monitor.addAction()` with `__dd_exposure` name
   - Registered in `RUMFeature.messageReceivers`
- `FlagsClient` Integration:
   - `RUMExposureLogger` injected alongside existing `ExposureLogger`
   - Both loggers called in `getDetails()` after flag evaluation

### Review checklist
- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes
- [ ] Add Objective-C interface for public APIs (see our [guidelines](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3157787243/RFC+-+Modular+Objective-C+Interface#Recommended-solution) (internal) and run `make api-surface`)

PR #2508, opened 2025-10-06
FFL-1162: Flags / RUM integration tests

gonzalezreal:
### What and why?

This PR adds integration tests that verify the end-to-end communication between the `DatadogFlags` and `DatadogRUM` modules. These tests ensure that flag evaluations correctly trigger RUM events and that all metadata flows through the integration layer properly.

### How?

Created `Datadog/IntegrationUnitTests/Flags/FlagsRUMIntegrationTests.swift` with two test cases:

1. `testWhenFlagIsEvaluated_itAddsFeatureFlagToRUMView()`
   - Verifies flag evaluations appear in `RUMViewEvent.featureFlags`
   - Validates `RUMFlagEvaluationMessage` → `FlagEvaluationReceiver` integration

2. `testWhenFlagIsEvaluated_itCreatesRUMActionWithExposureDetails()`
   - Verifies custom `__dd_exposure` RUM actions are created
   - Validates `RUMFlagExposureMessage` → `FlagExposureReceiver` integration

### Review checklist

### DIFF
--- a/Datadog/IntegrationUnitTests/Flags/FlagsRUMIntegrationTests.swift
+++ b/Datadog/IntegrationUnitTests/Flags/FlagsRUMIntegrationTests.swift
@@ -14,7 +14,7 @@ import DatadogInternal
 /// Covers integration scenarios between Flags and RUM features.
 final class FlagsRUMIntegrationTests: XCTestCase {
     private enum Fixtures {
-        static let flagsRepositoryState = FlagsData(
+        static let flagsData = FlagsData(
             flags: [
                 "string-flag": .init(
                     allocationKey: "allocation-123",
@@ -43,48 +43,41 @@ final class FlagsRUMIntegrationTests: XCTestCase {
 
     override func setUp() {
         super.setUp()
+
         core = DatadogCoreProxy(context: .mockWith(trackingConsent: .granted))
+
+        RUM.enable(with: .init(applicationID: "test-app-id"), in: core)
+        Flags.enable(in: core)
+
+        let featureScope = core.scope(for: FlagsFeature.self)
+        featureScope.flagsDataStore.setFlagsData(Fixtures.flagsData, forClientNamed: FlagsClient.defaultName)
+        featureScope.dataStore.flush()
     }
 
     override func tearDownWithError() throws {
+        let featureScope = core.scope(for: FlagsFeature.self)
+        featureScope.dataStore.clearAllData()
+
         try core.flushAndTearDown()
         core = nil
+
         super.tearDown()
-    }
-
-    private func createFlagsClient(state: FlagsData? = Fixtures.flagsRepositoryState) -> FlagsClientProtocol {
-        let featureScope = core.scope(for: FlagsFeature.self)
-        let dateProvider = SystemDateProvider()
-
-        return FlagsClient(
-            repository: FlagsRepositoryMock(
-                clientName: "default",
-                state: state
-            ),
-            exposureLogger: ExposureLogger(
-                dateProvider: dateProvider,
-                featureScope: featureScope
-            ),
-            rumExposureLogger: RUMExposureLogger(
-                dateProvider: dateProvider,
-                featureScope: featureScope
-            )
-        )
     }
 
     func testWhenFlagIsEvaluated_itAddsFeatureFlagToRUMView() throws {
         // Given
-        RUM.enable(with: .init(applicationID: "test-app-id"), in: core)
-        Flags.enable(in: core)
-
         let monitor = RUMMonitor.shared(in: core)
-        let client = createFlagsClient()
+        let client = FlagsClient.create(in: core)
 
         // When
         monitor.startView(key: "test-view", name: "Test View")
 
+        let featureScope = core.scope(for: FlagsFeature.self)
+        featureScope.dataStore.flush()
+
         let boolValue = client.getBooleanValue(key: "boolean-flag", defaultValue: false)
         let stringValue = client.getStringValue(key: "string-flag", defaultValue: "blue")
+
         core.flush()
 
         monitor.stopView(key: "test-view")
@@ -103,20 +96,22 @@ final class FlagsRUMIntegrationTests: XCTestCase {
             "View should have feature flags"
         )
 
+        XCTAssertEqual(featureFlags.count, 2)
         XCTAssertEqual(featureFlags["boolean-flag"] as? Bool, boolValue)
         XCTAssertEqual(featureFlags["string-flag"] as? String, stringValue)
     }
 
     func testWhenFlagIsEvaluated_itCreatesRUMActionWithExposureDetails() throws {
         // Given
-        RUM.enable(with: .init(applicationID: "test-app-id"), in: core)
-        Flags.enable(in: core)
-
         let monitor = RUMMonitor.shared(in: core)
-        let client = createFlagsClient()
+        let client = FlagsClient.create(in: core)
 
         // When
         monitor.startView(key: "test-view", name: "Test View")
+
+        let featureScope = core.scope(for: FlagsFeature.self)
+        featureScope.dataStore.flush()
+
         _ = client.getBooleanValue(key: "boolean-flag", defaultValue: false)
 
         // Then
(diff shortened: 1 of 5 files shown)


DatadogInternal/Sources/Models/RUM/RUMPayloadMessages.swift
```diff
@@ -109,7 +109,7 @@ public struct RUMFlagEvaluationMessage {
 
 /// Flag exposure message consumed by RUM on the message-bus.
 public struct RUMFlagExposureMessage {
-    /// The timestamp of the exposure
+    /// The timestamp of the exposure (server time with NTP correction)
```
gonzalezreal, line 112: Addressing feedback #2502

DatadogRUM/Sources/Integrations/FlagExposureReceiver.swift
```diff
@@ -20,7 +23,7 @@ internal struct FlagExposureReceiver: FeatureMessageReceiver {
 
         monitor.addAction(
             type: .custom,
-            name: "__dd_exposure",
+            name: Constants.exposureActionName,
```
gonzalezreal, line 26: Addressing feedback #2502

Datadog/IntegrationUnitTests/Flags/FlagsRUMIntegrationTests.swift
```diff
@@ -0,0 +1,181 @@
...
+            rumEvents.last,
+            "Should have at least one view event"
+        )
+        let featureFlags = try XCTUnwrap(
+            viewEvent.featureFlags?.featureFlagsInfo,
+            "View should have feature flags"
+        )
+
+        XCTAssertEqual(featureFlags["boolean-flag"] as? Bool, boolValue)
+        XCTAssertEqual(featureFlags["string-flag"] as? String, stringValue)
```
barboraplasovska, line 101: Suggestion: Check that only 2 feature flags are present?

Datadog/IntegrationUnitTests/Flags/FlagsRUMIntegrationTests.swift
```diff
@@ -0,0 +1,181 @@
...
+        // Given
+        RUM.enable(with: .init(applicationID: "test-app-id"), in: core)
+        Flags.enable(in: core)
+
+        let monitor = RUMMonitor.shared(in: core)
+        let client = createFlagsClient()
+
+        // When
+        monitor.startView(key: "test-view", name: "Test View")
+        _ = client.getBooleanValue(key: "boolean-flag", defaultValue: false)
```
barboraplasovska, line 115: Question: You're not doing core.flush() here, but you have in the first test. Is there a reason why not?

Datadog/IntegrationUnitTests/Flags/FlagsRUMIntegrationTests.swift
```diff
@@ -0,0 +1,181 @@
...
+        XCTAssertEqual(contextInfo["exposure_key"] as? String, "boolean-flag-allocation-124")
+        XCTAssertEqual(contextInfo["subject_key"] as? String, "user-123")
+        XCTAssertEqual(contextInfo["variant_key"] as? String, "variation-124")
+        XCTAssertNotNil(contextInfo["subject_attributes"])
+    }
+}
+
+// MARK: - FlagsRepositoryMock
+
+final class FlagsRepositoryMock: FlagsRepositoryProtocol {
```
barboraplasovska: Suggestion: Move `FlagsRepositoryMock` into a separate file for clarity.

barboraplasovska (dismissed):
Good job!
I've left some questions and suggestions, nothing blocking.

Datadog/IntegrationUnitTests/Flags/FlagsRUMIntegrationTests.swift
```diff
@@ -0,0 +1,181 @@
...
+        XCTAssertEqual(exposureAction.action.target?.name, "__dd_exposure")
+
+        let contextInfo = try XCTUnwrap(exposureAction.context?.contextInfo)
+        XCTAssertNotNil(contextInfo["timestamp"])
+        XCTAssertEqual(contextInfo["flag_key"] as? String, "boolean-flag")
+        XCTAssertEqual(contextInfo["allocation_key"] as? String, "allocation-124")
+        XCTAssertEqual(contextInfo["exposure_key"] as? String, "boolean-flag-allocation-124")
+        XCTAssertEqual(contextInfo["subject_key"] as? String, "user-123")
+        XCTAssertEqual(contextInfo["variant_key"] as? String, "variation-124")
+        XCTAssertNotNil(contextInfo["subject_attributes"])
```
sameerank: Since we're checking all the other fields, why not also verify that this is `["foo": .string("bar")]`?

sameerank (approved):
Nice work!

Datadog/IntegrationUnitTests/Flags/FlagsRUMIntegrationTests.swift
gonzalezreal: The `core.flush()` is needed in the first test because `monitor.stopView()` runs immediately after flag evaluations, and we need to ensure the `RUMFlagEvaluationMessage` is processed before the view event is written.

In the second test, we go straight to `waitAndReturnEvents()` which internally flushes, so no explicit flush is needed since there's no subsequent RUM operation that depends on the messages being processed first.

Datadog/IntegrationUnitTests/Flags/FlagsRUMIntegrationTests.swift
```diff
@@ -0,0 +1,181 @@
...
+            exposureLogger: ExposureLogger(
+                dateProvider: dateProvider,
+                featureScope: featureScope
+            ),
+            rumExposureLogger: RUMExposureLogger(
+                dateProvider: dateProvider,
+                featureScope: featureScope
+            )
+        )
+    }
```
ncreated: suggestion/ It might be worth investigating why we can’t use the real client (`FlagsClient.create(in:)`) here. Re-implementing client creation could indicate that the overall design might benefit from stronger dependency injection. My concern is that as more functionality gets added to `FlagsClient`, this test setup will need to expose more and more of its internals. Over time, that could erode its intended “integration” semantics, turning it into a pseudo–unit test with mixed concerns.

In my view, the minimal dependencies for `FlagsClient` are `DateProvider` and `FlagAssignmentsFetching`. Both could be injected through **internal** variables in `Flags.Configuration`, providing a simple and lightweight form of DI. This is similar to [how RUM handles several dependencies](https://github.com/DataDog/dd-sdk-ios/blob/89490e8c3c6aea06334725712d86178373999b32/DatadogRUM/Sources/RUMConfiguration.swift#L349-L379) for `RUMMonitor`. It’s not a full DI setup, but it’s efficient and keeps dependencies well-contained.

Datadog/IntegrationUnitTests/Flags/FlagsRUMIntegrationTests.swift
gonzalezreal: ~~Other integration tests have their mocks on the same file. I think we are good as long as the mock is local to the test. If we add other `XCTestCase`s that depend on it, then we can move it.~~

I removed the mock :)

Datadog/IntegrationUnitTests/Flags/FlagsRUMIntegrationTests.swift
ncreated: Also, this test currently uses `FlagsRepositoryMock`, but I don’t think there’s anything preventing us from testing against the real repository 🏋️‍♂️ (integrated through the core’s `DataStore`). `DatadogCoreProxy` already provides everything needed for that setup, including proper cleanup routines after the test.

WDYT?

ncreated (dismissed):
Overall, this looks great 💪 - the tested behaviors are exactly what we need 🎯 .

My only concern is that the current setup exposes too many internal implementation details to the test configuration, making it feel less like an integration test and more like a unit test. This boundary could erode further over time. I’ve left an inline comment with a suggestion 💡.

Datadog/IntegrationUnitTests/Flags/FlagsRUMIntegrationTests.swift
gonzalezreal: 🤔 Good point. The goal is to avoid calling `setEvaluationContext`, which triggers a network request. If we can setup a data store for the test client with proper data, we don't need to mock the repository 👍🏽.

gonzalezreal: Solved in [634d9ba](https://github.com/DataDog/dd-sdk-ios/commit/634d9bac3243764bb522cea3239f9f88b25af877)

Datadog/IntegrationUnitTests/Flags/FlagsRUMIntegrationTests.swift
gonzalezreal: Solved in [634d9ba](https://github.com/DataDog/dd-sdk-ios/commit/634d9bac3243764bb522cea3239f9f88b25af877)

Datadog/IntegrationUnitTests/Flags/FlagsRUMIntegrationTests.swift
gonzalezreal: Solved in [634d9ba](https://github.com/DataDog/dd-sdk-ios/commit/634d9bac3243764bb522cea3239f9f88b25af877)

DatadogCore/Sources/SDKMetrics/BatchMetrics.swift
```diff
@@ -19,6 +19,7 @@ internal enum BatchMetric {
         case "tracing":         return "trace"
         case "session-replay":  return "sr"
         case "session-replay-resources":  return "sr-resources"
+        case "flags":           return "flags"
```
ncreated, line 22: 💪

barboraplasovska (approved):

PR #2510, opened 2025-10-06
[FFL-1161] enable exposure logging and rum integration with configuration

sameerank:
### What and why?

- Aligns with cross-platform patterns
  - [OpenFeature JS already has `enableExposureLogging`](https://github.com/DataDog/openfeature-js-client/blob/23f74a9a84718ae471e86255ecabfbc624843929/packages/browser/src/domain/configuration.ts#L47)
  - [Android SDK already has `enableExposureLogging`](https://github.com/DataDog/dd-sdk-android/blob/d8c2c356c3b7cca3941da43dfbf61614cc966038/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsConfiguration.kt#L34)
- Gives users flexibility to independently control exposure logging vs legacy RUM integration
- Allows migration from RUM-based exposure logging to dedicated exposures endpoint or the reverse

### How?

Configuration:
- Added `enableExposureLogging: Bool = true` and `enableRUMIntegration: Bool = true` to `Flags.Configuration`
- Both default to `true` for backward compatibility

Architecture:
- Stored both flags in `FlagsFeature` for access by `FlagsClient`
- Implemented conditional logging in `FlagsClient.swift` around existing `exposureLogger.logExposure()` and `rumExposureLogger.logExposure()` calls
- No changes to existing batching infrastructure or request building

Testing:
- Tests verify independent control of each logging mechanism
- Updated existing test mocks to include new parameters

### Review checklist


DatadogFlags/Sources/Flags.swift
```diff
@@ -28,14 +28,32 @@ public enum Flags {
...
+        ///
+        /// Default: `true`.
+        public var enableExposureLogging: Bool
+
+        /// Enables exposure logging via RUM integration.
+        ///
+        /// When enabled, flag evaluation events are logged as RUM actions for correlation with user sessions.
+        ///
+        /// Default: `true`.
+        public var enableRUMIntegration: Bool
```
gonzalezreal: Suggestion: Consider using `trackExposures` and `rumIntegrationEnabled` instead of `enableExposureLogging` and `enableRUMIntegration` to align with the SDK's configuration naming conventions. RUM uses `track*` prefix for collecting events (`trackFrustrations`, `trackBackgroundEvents`, etc.) and `*Enabled` suffix for integrations (`bundleWithRumEnabled`, `networkInfoEnabled`). This would make the naming more consistent across the codebase.

gonzalezreal (commented):
👌🏽 This is a valid implementation that adds the configuration options we need for cross-platform alignment.

However, I think there's an opportunity to simplify it. Rather than adding the boolean properties to `FlagsClient` and checking them on every flag evaluation, we could use the `Null Object` pattern that's already established in the codebase.

The idea would be to create `NOPExposureLogger` and `NOPRUMExposureLogger`, then inject those in `createOrThrow()` when the corresponding configuration flags are disabled. This keeps `FlagsClient` focused on its core responsibility without needing to track or check these flags. The testing also becomes simpler - you'd just verify that `enableExposureLogging: false` results in a `NOPExposureLogger` being injected, rather than testing all the call count combinations.

DatadogFlags/Sources/Client/FlagsClient.swift
```diff
@@ -24,6 +24,16 @@ public class FlagsClient {
...
 
+    // MARK: - Internal Testing Helpers
+
+    internal var isUsingNOPExposureLogger: Bool {
+        return exposureLogger is NOPExposureLogger
+    }
+
+    internal var isUsingNOPRUMLogger: Bool {
+        return rumExposureLogger is NOPRUMExposureLogger
+    }
```
sameerank: https://github.com/DataDog/dd-sdk-ios/pull/2510/commits/0da26d7506afec8af3ac3fd533fb5a5b936fa600 Regarding

> verify that `enableExposureLogging: false` results in a `NOPExposureLogger` being injected, rather than testing all the call count combinations

I added these internal methods to help us test which logger type was injected, but I'm open to suggestions if there are better ways to verify this. This at least feels better than checking `mockLogger.logExposureCalls.count`, since we are preserving the NOP injection

DatadogFlags/Sources/Flags.swift
sameerank: Thanks! I hadn't noticed this pattern in naming conventions, but I agree that we should aim to be consistent

https://github.com/DataDog/dd-sdk-ios/pull/2510/commits/30342b0606c6db2247f4f5e0f6844629d357c9f1

DatadogFlags/Sources/Client/FlagsClient.swift
gonzalezreal: I realized my initial advice was flawed - checking for NOP types still tests implementation details rather than actual behavior.

I've implemented a proper solution in #2516 that:
- Removes the internal test helper methods
- Uses a factory pattern in `FlagsFeature` to encapsulate the NOP vs real logger decision
- Adds behavioral tests using `SingleFeatureCoreMock` with `DataStoreMock` that verify the actual outcome (whether `ExposureEvent`s are written and RUM messages are sent)

Let me know your thoughts.

PR #2511, opened 2025-10-07
chore: Switch to Dependabot multi-ecosystem group

ncreated:
### What and why?

📦 This PR enables the `multi-ecosystem-groups` setting in Dependabot to simplify dependency management and consolidate updates into a single monthly PR.

### How?

Replaced separate update configurations with one [`multi-ecosystem-group`](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#multi-ecosystem-group).

### Review checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,22 +1,20 @@
 # Dependabot configuration for GitHub Actions and Python dependencies
-# This will automatically create PRs to update GitHub Actions and Python packages to their latest versions
+# This will automatically create PR to update GitHub Actions and Python packages to their latest versions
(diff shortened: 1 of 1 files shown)


mariedm (approved):

mariedm:
Thanks for taking care of it! That will greatly reduce noise. Should we also set it up on other repos (Fastlane, Github action)?

ncreated:
> Thanks for taking care of it! That will greatly reduce noise. Should we also set it up on other repos (Fastlane, Github action)?

On it! 🙂

Issue #2507, opened 2025-10-04
`EXC_BAD_ACCESS` Crash in `CoreLogger.debug()`


msrutek-paylocity:
### Stack trace

```
Thread 1 Queue : com.apple.main-thread (serial)
#0	0x0000000000000000 in 0x00000000 ()
#1	0x000000010cf02210 in InternalLogger.log(_:message:error:) at dd-sdk-ios/DatadogInternal/Sources/Telemetry/InternalLogger.swift:38
#2	0x000000010cf02ee8 in protocol witness for CoreLogger.log(_:message:error:) in conformance InternalLogger ()
#3	0x000000010cf014dc in CoreLogger.debug(_:error:) at dd-sdk-ios/DatadogInternal/Sources/Telemetry/CoreLogger.swift:75
#4	0x000000010cc01b34 in closure #2 in DatadogNTPDateProvider.synchronize(update:) at dd-sdk-ios/DatadogCore/Sources/Core/Context/ServerOffsetPublisher.swift:49
#5	0x000000010cc58f2c in closure #1 in KronosClock.sync(from:samples:first:completion:) at dd-sdk-ios/DatadogCore/Sources/Kronos/KronosClock.swift:142
#6	0x000000010cc59078 in partial apply for closure #1 in KronosClock.sync(from:samples:first:completion:) ()
#7	0x000000010cc61e30 in $defer #1 () in closure #1 in closure #1 in KronosNTPClient.query(pool:version:port:numberOfSamples:maximumServers:timeout:progress:) at dd-sdk-ios/DatadogCore/Sources/Kronos/KronosNTPClient.swift:53
#8	0x000000010cc61930 in closure #1 in closure #1 in KronosNTPClient.query(pool:version:port:numberOfSamples:maximumServers:timeout:progress:) ()
#9	0x000000010cc650b4 in partial apply for closure #1 in closure #1 in KronosNTPClient.query(pool:version:port:numberOfSamples:maximumServers:timeout:progress:) ()
#10	0x000000010cc62b8c in closure #1 in KronosNTPClient.query(ip:port:version:timeout:numberOfSamples:completion:) at dd-sdk-ios/DatadogCore/Sources/Kronos/KronosNTPClient.swift:111
#11	0x000000010cc62d50 in partial apply for closure #1 in KronosNTPClient.query(ip:port:version:timeout:numberOfSamples:completion:) ()
[...]

PR #2514, opened 2025-10-07
[RUM-11957] Fix Logger race condition

mariedm:
### What and why?

Fixes a race condition causing `EXC_BAD_ACCESS` crashes when `DD.logger` is accessed during SDK initialization. This resolves [#2507](https://github.com/DataDog/dd-sdk-ios/issues/2507) where a customer reported crashes in `InternalLogger.log()` when NTP synchronization callbacks attempted to log messages while `Datadog.initialize()` was replacing the logger instance.

The crash occurs when calling `DD.logger` while it's [being replaced during initialization](https://github.com/DataDog/dd-sdk-ios/blob/ff9f01dda6dcccf25a20905f9366e663e733a34e/DatadogCore/Sources/Datadog.swift#L306).

### How?

Added `@ReadWriteLock` property wrapper to the logger to ensure thread-safe access.

### Review checklist

### DIFF
--- a/DatadogCore/Tests/Datadog/Core/DD/InternalLoggerTests.swift
+++ b/DatadogCore/Tests/Datadog/Core/DD/InternalLoggerTests.swift
@@ -189,6 +189,6 @@ class InternalLoggerTests: XCTestCase {
             }
         }
 
-        waitForExpectations(timeout: 60)
+        waitForExpectations(timeout: 2)
     }
 }


DatadogCore/Tests/Datadog/Core/DD/InternalLoggerTests.swift
```diff
@@ -154,4 +154,41 @@ class InternalLoggerTests: XCTestCase {
...
+                        printFunction: { _, _ in /* no-op */ },
+                        verbosityLevel: { .debug }
+                    )
+                    Thread.sleep(forTimeInterval: 0.00001)
+                }
+                expectation.fulfill()
+            }
+        }
+
+        waitForExpectations(timeout: 60)
```
simaoseica-dd: nit: it shouldn't need a 60 second timeout. 2 seconds?

simaoseica-dd (approved):
Well done! 🏅

DatadogCore/Tests/Datadog/Core/DD/InternalLoggerTests.swift
mariedm: Yes! It's a leftover from my local testing :) Thanks for pointing it out.

git-thuerk-done (approved):

PR #2517, opened 2025-10-08
RUM-11469: Report Time to initial display

simaoseica-dd:
### What and why?

This PR introduces a new app launch metric: Time to Initial Display (TTID). 
It follows the proposal outlined in the [internal RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/5433951286/RFC+-+App+Launch+Observability+in+iOS). 

### How?

It adds a new `RUMAppLaunchManager` that centralizes all the logic related to app launch metrics:
- Reports the TTID through an "app_launch" `RUMVitalEvent`.
- Determines the startup type (cold or warm) using `StartupTypeHandler`, which relies on states managed by `AppStateManager`.
- Handles prewarmed launches, adjusting and categorizing TTID as `isPrewarmed`.
- Ensures the metric is reported only once and only for user-triggered launches.
- Detaches the metric from views, as it measures the app launch itself rather than a specific view.

### Review checklist


DatadogRUM/Sources/Instrumentation/AppState/AppStateManager.swift
```diff
@@ -113,4 +134,20 @@ internal final class WatchdogTerminationAppStateManager {
...
+    func storeCurrentAppState() throws {
+        try currentAppStateInfo { [self] appState in
+            featureScope.rumDataStore.setValue(appState, forKey: .appStateKey)
+        }
+    }
+
+    /// Reads the app state from the data store asynchronously.
+    /// - Parameter completion: The completion block called with the app state.
+    private func readAppState(completion: @escaping (AppStateInfo?) -> Void) {
+        featureScope.rumDataStore.value(forKey: .appStateKey) { (state: AppStateInfo?) in
```
simaoseica-dd, line 148: Using the new `.appStateKey` may result in a watchdog termination not being reported. This is a trade-off to avoid introducing complex migration logic here.

DatadogRUM/Sources/RUMMonitor/Dependencies/RUMAppLaunchManager.swift
```diff
@@ -0,0 +1,195 @@
...
+ */
+
+import DatadogInternal
+import Foundation
+
+internal class RUMAppLaunchManager {
+    internal enum Constants {
+        // Maximum time for a long interval between app launches
+        static let maxInactivityDuration: TimeInterval = 604_800 // 1 week
+        // Maximum time for an erroneous ttid
```
mariedm: nit: making the reference explicit for other maintainers
```suggestion
        // Maximum time for an erroneous TTID (Time to Initial Display)
```

DatadogRUM/Sources/RUMMonitor/Scopes/RUMAppLaunchManager.swift
```diff
@@ -0,0 +1,195 @@
...
+                return nil
+            }
+            return command.time.timeIntervalSince(runtimeLoadDate)
+        default:
+            return nil
+        }
+    }
+}
+
+private extension RUMVitalEvent.Vital.AppLaunchProperties.AppLaunchMetric {
```
mariedm, line 143: nit: what about moving the rest of the file to a separate `StartupTypeHandler` file? (I know it means making them `internal` but it would be for more clarity)

DatadogRUM/Sources/RUMMonitor/Scopes/RUMAppLaunchManager.swift
```diff
@@ -0,0 +1,195 @@
...
+        return true
+    }
+
+    private func timeToInitialDisplay(from command: RUMTimeToInitialDisplayCommand, context: DatadogContext) -> TimeInterval? {
+        switch context.launchInfo.launchReason {
+        case .userLaunch:
+            return command.time.timeIntervalSince(context.launchInfo.processLaunchDate)
+        case .prewarming:
+            guard let runtimeLoadDate = context.launchInfo.launchPhaseDates[.runtimeLoad] else {
+                return nil
```
mariedm, line 134: Do we want telemetry for these edge cases? As well as for the different types of cold starts further down?

DatadogRUM/Sources/Instrumentation/AppState/AppStateManager.swift
```diff
@@ -69,39 +105,24 @@ internal final class WatchdogTerminationAppStateManager {
...
-            DD.logger.debug("Reading app state from data store.")
-            completion(state)
+            self.featureScope.rumDataStore.setValue(appState, forKey: .appStateKey)
         }
     }
 
     /// Builds the current app state asynchronously.
     /// - Parameter completion: The completion block called with the app state.
-    func currentAppState(completion: @escaping (WatchdogTerminationAppState) -> Void) throws {
+    func currentAppStateInfo(completion: @escaping (AppStateInfo) -> Void) throws {
```
mariedm: I don't see where this method is throwing; am I missing anything?

DatadogRUM/Sources/RUMMonitor/Scopes/RUMAppLaunchManager.swift
```diff
@@ -0,0 +1,195 @@
...
+    init(parent: RUMContextProvider, dependencies: RUMScopeDependencies) {
+        self.parent = parent
+        self.dependencies = dependencies
+    }
+
+    // MARK: - Internal Interface
+
+    func process(_ command: RUMCommand, context: DatadogContext, writer: Writer, activeView: RUMViewScope?) {
+        do {
+            switch command {
```
mariedm, line 35: There is only one type of command processed here, but I imagine TTFD we be handled here as well?

DatadogCore/Private/ObjcAppLaunchHandler.m
```diff
@@ -57,35 +70,55 @@ - (instancetype)init {
...
+#if TARGET_OS_IOS || TARGET_OS_TV || TARGET_OS_MACCATALYST || TARGET_OS_VISION
+    __block id __unused didFinishLaunchingNotification = [notificationCenter addObserverForName:UIApplicationDidFinishLaunchingNotification
+                                                                object:nil
+                                                                 queue:NSOperationQueue.mainQueue
+                                                            usingBlock:^(NSNotification *_){
+
+        @synchronized(self) {
+            self->_didFinishLaunchingDate = CFAbsoluteTimeGetCurrent();
+            for (UIApplicationNotificationCallback callback in self->_applicationNotificationCallbacks) {
+                callback(self.didFinishLaunchingDate, self.didBecomeActiveDate);
```
maxep: question/ Do we need to expose the `didFinishLaunchingDate` and `didBecomeActiveDate` as public properties? 
This will `@synchronized(self)` 3 times while one is actually needed. By removing public exposure, we can  remove the extra sync.

DatadogRUM/Sources/RUMMonitor/Scopes/RUMAppLaunchManager.swift
```diff
@@ -0,0 +1,195 @@
...
+
+import DatadogInternal
+import Foundation
+
+internal class RUMAppLaunchManager {
+    internal enum Constants {
+        // Maximum time for a long interval between app launches
+        static let maxInactivityDuration: TimeInterval = 604_800 // 1 week
+        // Maximum time for an erroneous ttid
+        static let maxTTIDDuration: TimeInterval = 60 // 1 minute
```
maxep, line 13: Seems high, isn't it? 10s is already erroneous IMO.

DatadogRUM/Sources/RUMMonitor/Dependencies/RUMAppLaunchManager.swift
```diff
@@ -0,0 +1,195 @@
...
+        } catch {
+            dependencies.telemetry.error("RUMAppLaunchManager failed to write the ttid vital event.", error: error)
+        }
+    }
+
+    // MARK: - Private Methods
+
+    private func writeTTIDVitalEvent(from command: RUMTimeToInitialDisplayCommand, context: DatadogContext, writer: Writer, activeView: RUMViewScope?) throws {
+        guard shouldProcess(command: command, context: context),
+              let timeToInitialDisplay = timeToInitialDisplay(from: command, context: context) else { return }
```
maxep: nit/


```suggestion
        guard 
            shouldProcess(command: command, context: context),
            let timeToInitialDisplay = timeToInitialDisplay(from: command, context: context) 
        else {
            return
        }
```

DatadogRUM/Sources/RUMMonitor/Dependencies/RUMAppLaunchManager.swift
```diff
@@ -0,0 +1,195 @@
...
+            default: break
+            }
+        } catch {
+            dependencies.telemetry.error("RUMAppLaunchManager failed to write the ttid vital event.", error: error)
+        }
+    }
+
+    // MARK: - Private Methods
+
+    private func writeTTIDVitalEvent(from command: RUMTimeToInitialDisplayCommand, context: DatadogContext, writer: Writer, activeView: RUMViewScope?) throws {
```
maxep: cr/ I don't see the `activeView` being used, I guess it can be removed.

DatadogRUM/Sources/RUMMonitor/Dependencies/RUMAppLaunchManager.swift
```diff
@@ -0,0 +1,195 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import DatadogInternal
+import Foundation
+
+internal class RUMAppLaunchManager {
```
maxep: suggestion/ This looks like a scope. Do you think it can be turned into a `AppLaunchVitalScope`, or is there specificities blocking that?

DatadogRUM/Sources/RUMMonitor/Dependencies/RUMAppLaunchManager.swift
```diff
@@ -0,0 +1,195 @@
...
+                if let previousAppStateInfo = appStateManager.previousAppStateInfo,
+                   previousAppStateInfo.appVersion != currentAppState.appVersion {
+                    return .coldStart
+                }
+            case .systemRestart:
+                if let previousAppStateInfo = appStateManager.previousAppStateInfo,
+                   previousAppStateInfo.systemBootTime < currentAppState.systemBootTime {
+                    return .coldStart
+                }
+            case .longInactivity:
```
mariedm: While the 3 other cases (System Reboot, Initial App Install, App Upgrade) are mentioned under the cold start scenarios in the RFC, `longInactivity` is not. It's stated that the 4th case is "Memory Pressure Eviction" and that no API or heuristics exist to identify this scenario. So I wonder if it's supposed to map to an "inactivity" scenario and how you came up with 1 week as the threshold for `maxInactivityDuration`?

DatadogRUM/Sources/RUMMonitor/Dependencies/RUMAppLaunchManager.swift
```diff
@@ -0,0 +1,195 @@
...
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import DatadogInternal
+import Foundation
+
+internal class RUMAppLaunchManager {
+    internal enum Constants {
+        // Maximum time for a long interval between app launches
+        static let maxInactivityDuration: TimeInterval = 604_800 // 1 week
```
mariedm: Was it the rationale behind 1 week?

mariedm (commented):
Great work! 🙌  Some questions here and there, but nothing major.

DatadogRUM/Sources/RUMMonitor/Dependencies/RUMAppLaunchManager.swift
mariedm: (Just to add to the above, imo it's better to categorize a scenario as unknown if we are unsure)

DatadogRUM/Sources/RUMMonitor/Scopes/RUMAppLaunchManager.swift
simaoseica-dd: Yes, 10s is a long app launch, but let's dogfood this value

DatadogRUM/Sources/RUMMonitor/Scopes/RUMAppLaunchManager.swift
simaoseica-dd: I added telemetry here but bear in mind that there will be a telemetry PR to address edge cases.

DatadogRUM/Sources/Instrumentation/AppState/AppStateManager.swift
simaoseica-dd: I didn't change the underlying function signature, but your point is correct 👍

DatadogRUM/Sources/RUMMonitor/Dependencies/RUMAppLaunchManager.swift
simaoseica-dd: This is a scenario of big intervals between app launches. The research on this topic captured some pre warmings after a week without app launches, meaning the system needed to allocate resources of the app to memory that might have been released in the meantime. 
This is subject of dogfooding as well, and it is a new use case in the RFC that I will add.

DatadogRUM/Sources/RUMMonitor/Dependencies/RUMAppLaunchManager.swift
simaoseica-dd: Changed to `RUMAppLaunchVitalScope`.  The `RUMFeatureOperationManager` is also under the same perspective. cc @mariedm

DatadogCore/Private/ObjcAppLaunchHandler.m
simaoseica-dd: Good point to have only one channel to share it (and to save syncs)

DatadogRUM/Sources/Instrumentation/WatchdogTerminations/WatchdogTerminationChecker.swift
```diff
@@ -31,18 +31,12 @@ internal final class WatchdogTerminationChecker {
...
     ///   - completion: The completion block called with the result.
     func isWatchdogTermination(launch: LaunchReport, completion: @escaping (Bool, AppStateInfo?) -> Void) throws {
-        do {
-            try appStateManager.currentAppStateInfo { [weak self] current in
-                self?.featureScope.context { [weak self] context in
-                    let previous = self?.appStateManager.previousAppStateInfo
-                    let isWatchdogTermination = self?.isWatchdogTermination(launch: launch, deviceInfo: context.device, from: previous, to: current)
-                    completion(isWatchdogTermination ?? false, previous)
-                }
+        appStateManager.currentAppStateInfo { [weak self] current in
```
mariedm, line 34: I guess this method is not throwing anymore either, same for `storeCurrentAppState()` and `writeTTIDVitalEvent()` I believe.

PR #2518, opened 2025-10-08
FFL-1174: Add bounded cache for exposure event de-duplication

gonzalezreal:
### What and why?

Implements a bounded cache for exposure event de-duplication to prevent unbounded memory growth. Previously, the `ExposureLogger` used an unbounded `Set` to track logged exposures, which could grow indefinitely over the lifetime of the application.

### How?

- Introduces `ExposureTracker` class using `NSCache` with a default limit of 100 entries
- `NSCache` automatically evicts least recently used entries when the limit is reached or there is memory pressure
- Refactors `ExposureLogger` to use `ExposureTracker` instead of unbounded `Set`
- Maintains all existing deduplication behavior and test coverage

### Review checklist

### DIFF
--- a/DatadogFlags/Sources/Client/ExposureTracker.swift
+++ b/DatadogFlags/Sources/Client/ExposureTracker.swift
@@ -1,0 +1,50 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+internal final class ExposureTracker {
+    struct Exposure: Hashable {
+        let targetingKey: String
+        let flagKey: String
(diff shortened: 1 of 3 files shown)


DatadogFlags/Sources/Client/ExposureTracker.swift
```diff
@@ -0,0 +1,50 @@
...
+                return false
+            }
+            return value == other.value
+        }
+    }
+
+    private let cache = NSCache<ExposureBox, NSNumber>()
+    private let sentinel = NSNumber(value: true)
+
+    init(countLimit: Int = 1_000) {
```
sameerank: Given this is a heuristic, I would consider defining this as a constant somewhere, e.g. `private static let defaultCountLimit = 1_000` to make the intent more explicit and maybe even with a comment e.g. so future developers know that this is safe to edit

In the future, this could potentially be made configurable to give customers more control over the memory/functionality tradeoff for their app.

PR #2522, opened 2025-10-09
RUM-11470: Report Time to full display

simaoseica-dd:
### What and why?

This PR introduces a new app launch metric: Time to Full Display (TTFD). 
It follows the proposal outlined in the [internal RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/5433951286/RFC+-+App+Launch+Observability+in+iOS). 

### How?

It introduces a new public interface to report that the app is fully interactive:
```swift
public protocol RUMMonitorProtocol: AnyObject {
    /// Records the time to full display (TTFD) of the current app launch.
    func reportAppFullyDisplayed()
}
```

The TTFD logic is centralized in the `RUMAppLaunchManager` that:
- Reports the TTFD through an "app_launch" `RUMVitalEvent`.
- Decorates the TTFD with the startup type (cold or warm) and `isPrewarmed` value.
- Ensures the metric is reported only once and only for user-triggered launches.

### Review checklist

### DIFF
--- a/DatadogRUM/Sources/Instrumentation/AppState/AppStateManager.swift
+++ b/DatadogRUM/Sources/Instrumentation/AppState/AppStateManager.swift
@@ -22,10 +22,6 @@ internal protocol AppStateManaging {
 
 /// Manages the app state changes observed during application lifecycle events such as application start, resume and termination.
 internal final class AppStateManager: AppStateManaging {
-    enum ErrorMessage: String {
-        case failedToStoreAppState = "Failed to store App State information"
-    }
-
     let featureScope: FeatureScope
 
     /// The last app state observed during application lifecycle events.
--- a/DatadogRUM/Sources/RUMMonitor/Scopes/RUMAppLaunchManager.swift
+++ b/DatadogRUM/Sources/RUMMonitor/Scopes/RUMAppLaunchManager.swift
@@ -21,7 +21,6 @@ internal class RUMAppLaunchManager {
 
     private var timeToInitialDisplay: Double?
     private var timeToFullDisplay: Double?
-    private var viewLoadingTime: Double?
     private var startupType: RUMVitalEvent.Vital.AppLaunchProperties.StartupType?
 
     private lazy var startupTypeHandler = StartupTypeHandler(appStateManager: dependencies.appStateManager)
@@ -41,13 +40,6 @@ internal class RUMAppLaunchManager {
             writeTTIDVitalEvent(from: command, context: context, writer: writer)
         case let command as RUMTimeToFullDisplayCommand:
             writeTTFDVitalEvent(from: command, context: context, writer: writer)
-        case let command as RUMAddViewLoadingTime:
-            registerViewLoadingTime(from: command, context: context)
-        case let command as RUMStopViewCommand:
-            if viewLoadingTime != nil,
-               timeToFullDisplay == nil {
-               writeTTFDVitalEvent(from: command, context: context, writer: writer)
-            }
         default: break
         }
     }
@@ -56,7 +48,7 @@ internal class RUMAppLaunchManager {
 // MARK: - TTID
 
 private extension RUMAppLaunchManager {
-    private func writeTTIDVitalEvent(from command: RUMTimeToInitialDisplayCommand, context: DatadogContext, writer: Writer) {
+    func writeTTIDVitalEvent(from command: RUMTimeToInitialDisplayCommand, context: DatadogContext, writer: Writer) {
         guard shouldProcess(command: command, context: context),
               let ttid = time(from: command, context: context)
         else {
@@ -212,49 +204,7 @@ private extension RUMAppLaunchManager {
     func shouldProcess(command: RUMTimeToFullDisplayCommand, context: DatadogContext) -> Bool {
         // Ignore command if the time to full display was already written
         guard self.timeToFullDisplay == nil else {
-            return false
-        }
-
-        // Ignore command if the time since the SDK load is too big
-        guard let runtimeLoadDate = context.launchInfo.launchPhaseDates[.runtimeLoad],
-              command.time.timeIntervalSince(runtimeLoadDate) < Constants.maxTTFDDuration else {
-            return false
-        }
-
-        return true
-    }
-}
-
-// MARK: - View loading time
-
-private extension RUMAppLaunchManager {
-    func registerViewLoadingTime(from command: RUMAddViewLoadingTime, context: DatadogContext) {
-        guard shouldProcess(command: command, context: context),
-              let viewLoadingTime = time(from: command, context: context) else { return }
-
-        self.viewLoadingTime = viewLoadingTime
-    }
-
-    func writeTTFDVitalEvent(from command: RUMStopViewCommand, context: DatadogContext, writer: Writer) {
-        if let viewLoadingTime, let timeToInitialDisplay, let startupType {
-            let attributes = command.globalAttributes
-                .merging(command.attributes) { $1 }
-            let ttfd = max(timeToInitialDisplay, viewLoadingTime)
-
-            self.writeVitalEvent(
-                duration: Double(ttfd.toInt64Nanoseconds),
-                appLaunchMetric: .ttfd,
-                startupType: startupType,
-                attributes: attributes,
-                context: context,
-                writer: writer
-            )
-        }
-    }
-
-    func shouldProcess(command: RUMAddViewLoadingTime, context: DatadogContext) -> Bool {
-        // Ignore command if the time to full display was already written
-        guard self.timeToFullDisplay == nil else {
+            DD.logger.warn("Time to Full Display was already processed. Make sure the `reportAppFullyDisplayed()` API is only called once.")
             return false
         }
 
--- a/DatadogRUM/Tests/RUMMonitor/Scopes/RUMAppLaunchManagerTests.swift
+++ b/DatadogRUM/Tests/RUMMonitor/Scopes/RUMAppLaunchManagerTests.swift
@@ -9,7 +9,7 @@ import DatadogInternal
 @testable import DatadogRUM
 @testable import TestUtilities
 
-final class RUMAppLaunchVitalScopeTests: XCTestCase {
+final class RUMAppLaunchManagerTests: XCTestCase {
     private var manager: RUMAppLaunchManager! // swiftlint:disable:this implicitly_unwrapped_optional
     private var appStateManager: AppStateManaging! // swiftlint:disable:this implicitly_unwrapped_optional
     private var mockParent: RUMContextProviderMock! // swiftlint:disable:this implicitly_unwrapped_optional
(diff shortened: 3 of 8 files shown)


DatadogRUM/Sources/RUMMonitor/Scopes/RUMAppLaunchManager.swift
```diff
@@ -32,85 +37,80 @@ internal class RUMAppLaunchManager {
...
+    func process(_ command: RUMCommand, context: DatadogContext, writer: Writer) {
         do {
             switch command {
             case let command as RUMTimeToInitialDisplayCommand:
-                try writeTTIDVitalEvent(from: command, context: context, writer: writer, activeView: activeView)
+                try writeTTIDVitalEvent(from: command, context: context, writer: writer)
+            case let command as RUMTimeToFullDisplayCommand:
+                try writeTTFDVitalEvent(from: command, context: context, writer: writer)
+            case let command as RUMAddViewLoadingTime:
+                try registerViewLoadingTime(from: command, context: context)
```
aleksandr-gringauz: I still have doubts about reporting "viewLoadingTime" as TTFD.

IIUC the logic in this PR is the following. We track `viewLoadingTime` for the first 2 views (including or excluding ApplicationLaunch view by the way?). If the `viewLoadingTime` was reported and if at the moment of `RUMStopViewCommand` TTFD hasn't been reported, we report `ViewLoadingTime` as TTFD.

Yes, this way we collect TTFD automatically for those customers who already use `addViewLoadingTime`. But this IMHO isn't a very big help for the customer. Adding 1 line of code is easy, customers don't need help with that.

I have the following concerns here:

1. Suppose the customer wants for some reason to report TTFD later than the first 2 views and at least one of them has `viewLoadingTime`. Maybe the first views are some kind of spinners. Or the customer just wants to report TTFD on the 3rd view for some reason. IIUC with the current implementation this becomes impossible for the customer.
2. We sometimes report TTFD at the stop of the view. If there is a crash before the view ends, TTFD will not be reported but the corresponding `viewLoadingTime` will be. Customers looking at sessions with crash might wonder why. Also on Android we call `stopView` [here](https://github.com/DataDog/dd-sdk-android/blob/4a2b3e4e60c5aa716968f36413ffa31f31e2385f/features/dd-sdk-android-rum/src/main/kotlin/com/datadog/android/rum/tracking/ActivityViewTrackingStrategy.kt#L79). This moment corresponds to "swipe from recents" when the users swipes the app to kill it. There is no guarantee that TTFD event will be written to disk.

simaoseica-dd: Indeed we can have a lot of use cases with `viewLoadingTime` but we can conclude that, if a `addViewLoadingTime`  is called, it is expected that the loading for that view is finished and the app is fully interactive (TTFD definition).
In a multi-view app, like an iPad app, I agree with you that we can improve this and other events better.

DatadogRUM/Sources/Instrumentation/AppState/AppStateManager.swift
```diff
@@ -57,13 +57,7 @@ internal final class AppStateManager: AppStateManaging {
     private func start() {
         self.readAppState { [weak self] in
             self?.previousAppStateInfo = $0
-
-            do {
-                try self?.storeCurrentAppState()
-            } catch {
-                DD.logger.error(ErrorMessage.failedToStoreAppState.rawValue, error: error)
```
mariedm, line 64: Is `ErrorMessage.failedToStoreAppState` still used anywhere or can it be deleted?

DatadogRUM/Tests/RUMMonitor/Scopes/RUMAppLaunchManagerTests.swift
```diff
@@ -9,7 +9,7 @@ import DatadogInternal
 @testable import DatadogRUM
 @testable import TestUtilities
 
-final class RUMAppLaunchManagerTests: XCTestCase {
+final class RUMAppLaunchVitalScopeTests: XCTestCase {
```
mariedm:
```suggestion
final class RUMAppLaunchManagerTests: XCTestCase {
```

DatadogRUM/Tests/RUMMonitor/Monitor+GlobalAttributesTests.swift
```diff
@@ -560,80 +560,6 @@ class Monitor_GlobalAttributesTests: XCTestCase {
         XCTAssertEqual(viewAfterSecondTiming.attribute(forKey: "attribute2"), "value2")
     }
 
-    // MARK: - View Loading Time
```
mariedm, line 563: Why removing these tests?

DatadogRUM/Sources/RUMMonitor/Scopes/RUMAppLaunchManager.swift
```diff
@@ -138,6 +134,138 @@ internal class RUMAppLaunchManager {
...
+                attributes: attributes,
+                context: context,
+                writer: writer
+            )
+        }
+    }
+
+    func shouldProcess(command: RUMTimeToFullDisplayCommand, context: DatadogContext) -> Bool {
+        // Ignore command if the time to full display was already written
+        guard self.timeToFullDisplay == nil else {
```
mariedm, line 206: Should we log a warning when this happens? This would be helpful:
- in case customers call the API multiple times
- because of the fallback to `viewLoadingTime`

DatadogRUM/Sources/RUMMonitor/Scopes/RUMAppLaunchManager.swift
mariedm: I share some of Sasha's concerns. In any case, we should make very clear to customers how this feature works in our public doc and link to it in our release notes.

mariedm (dismissed):
Some minor comments, otherwise looks great!

DatadogRUM/Sources/RUMMonitor/Scopes/RUMAppLaunchManager.swift
simaoseica-dd: I’ll align this with Android to keep it consistent.

DatadogRUM/Sources/RUMMonitor/Scopes/RUMAppLaunchManager.swift
simaoseica-dd: Indeed. Added the console warning.

DatadogRUM/Tests/RUMMonitor/Monitor+GlobalAttributesTests.swift
simaoseica-dd: They were moved to the right context

DatadogRUM/Sources/RUMMonitor/Scopes/RUMAppLaunchManager.swift
```diff
@@ -31,88 +35,72 @@ internal class RUMAppLaunchManager {
...
 
-    // MARK: - Private Methods
+// MARK: - TTID
 
-    private func writeTTIDVitalEvent(from command: RUMTimeToInitialDisplayCommand, context: DatadogContext, writer: Writer) throws {
-        guard
-            shouldProcess(command: command, context: context),
-            let timeToInitialDisplay = timeToInitialDisplay(from: command, context: context)
+private extension RUMAppLaunchManager {
+    private func writeTTIDVitalEvent(from command: RUMTimeToInitialDisplayCommand, context: DatadogContext, writer: Writer) {
```
mariedm: nit: it already inherits the private accessor from the extension
```suggestion
   func writeTTIDVitalEvent(from command: RUMTimeToInitialDisplayCommand, context: DatadogContext, writer: Writer) {
```

DatadogRUM/Sources/RUMMonitor/Scopes/RUMAppLaunchManager.swift
```diff
@@ -138,6 +126,96 @@ internal class RUMAppLaunchManager {
...
+                context: context,
+                writer: writer
+            )
+        }
+    }
+
+    func shouldProcess(command: RUMTimeToFullDisplayCommand, context: DatadogContext) -> Bool {
+        // Ignore command if the time to full display was already written
+        guard self.timeToFullDisplay == nil else {
+            DD.logger.warn("Time to Full Display was already processed. Make sure the `Monitor.reportAppFullyDisplayed()` api is only called once.")
```
mariedm: API should be uppercased. And we should either reference the method name alone or the public accessor `RUMMonitor.shared().reportAppFullyDisplayed()`.

```suggestion
            DD.logger.warn("Time to Full Display was already processed. Make sure the `reportAppFullyDisplayed()` API is only called once.")
```

DatadogRUM/Sources/RUMMonitor/Scopes/RUMAppLaunchManager.swift
```diff
@@ -31,88 +35,72 @@ internal class RUMAppLaunchManager {
...
-                    hasReplay: context.hasReplay,
-                    id: parent.context.sessionID.toRUMDataFormat,
-                    type: dependencies.sessionType
-                ),
-                source: .init(rawValue: context.source) ?? .ios,
-                synthetics: dependencies.syntheticsTest,
-                usr: .init(context: context),
-                version: context.version,
-                vital: vital
+            let startupType = self.startupTypeHandler.startupType(currentAppState: currentAppStateInfo)
```
mariedm, line 68: I know the `StartupTypeHandler` has been introduced in the previous PR, but I think we should add unit tests for all the different use cases it covers.

DatadogRUM/Sources/RUMMonitor/Scopes/RUMAppLaunchManager.swift
```diff
@@ -31,88 +35,72 @@ internal class RUMAppLaunchManager {
...
+                let ttfd = max(ttid, timeToFullDisplay)
+                self.writeVitalEvent(
+                    duration: Double(ttfd.toInt64Nanoseconds),
+                    appLaunchMetric: .ttfd,
+                    startupType: startupType,
+                    attributes: attributes,
+                    context: context,
+                    writer: writer
+                )
+            }
```
maxep, line 91: question/ Is this a valid use case? would mean that the TTFD api was wrongly used, right? Maybe we should just drop a wrong TTFD..

simaoseica-dd: No. It has to do to how we define the view scope. 
The TTFD can be reported even before we start a RUM view. For instance, TTFD ready after `viewDidload` and the SDK only starts the View on `viewDidAppear` which is a similar point as the first frame.

simaoseica-dd:
@maxep 
You can find the integration tests for the different scenarios [here](https://github.com/DataDog/dd-sdk-ios/pull/2533).
The UI Tests also have a task to be handled separately.

DatadogRUM/Sources/RUMMonitor/Scopes/RUMAppLaunchManager.swift
maxep: Yes, but I mean if TTFD is reported by the user before TTID, it is most-likely is misusage of the API, right?

simaoseica-dd: No because one is automatic and the other manual. For the user it might make sense to report in `viewDidload` because the app is ready to be interactable. No need for them to report it much later just for the SDK convenience.

DatadogRUM/Tests/RUMMonitor/Scopes/Utils/StartupTypeHandlerTests.swift
mariedm, line 1: Thank you for adding those 👍

mariedm (approved):

DatadogRUM/Sources/RUMMonitor/Scopes/RUMAppLaunchManager.swift
maxep: OK I see, the `viewDidLoad` can be called before the first render 👍

PR #2525, opened 2025-10-09
RUM-12227: Notify all SDK instances of the `UIApplication.didBecomeActive` notification

simaoseica-dd:
### What and why?

This PR improves how the `UIApplicationDidBecomeActiveNotification` is broadcast to all SDK instances initialized in the app.

### How?

The `AppLaunchHandler` now retains an array of callbacks and notifies all registered instances when the notification is received.

### Review checklist

### DIFF
--- a/DatadogCore/Private/ObjcAppLaunchHandler.m
+++ b/DatadogCore/Private/ObjcAppLaunchHandler.m
@@ -34,7 +34,7 @@ int processStartTimeIntervalSinceReferenceDate(NSTimeInterval *timeInterval);
 @implementation __dd_private_AppLaunchHandler {
     NSTimeInterval _processLaunchDate;
     NSTimeInterval _timeToDidBecomeActive;
-    UIApplicationDidBecomeActiveCallback _applicationDidBecomeActiveCallback;
+    NSMutableArray<UIApplicationDidBecomeActiveCallback> *_applicationDidBecomeActiveCallbacks;
 }
 
 static __dd_private_AppLaunchHandler *_shared;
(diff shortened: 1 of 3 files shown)


DatadogCore/Private/ObjcAppLaunchHandler.m
```diff
@@ -34,7 +34,7 @@
 @implementation __dd_private_AppLaunchHandler {
     NSTimeInterval _processLaunchDate;
     NSTimeInterval _timeToDidBecomeActive;
-    UIApplicationDidBecomeActiveCallback _applicationDidBecomeActiveCallback;
+    NSMutableArray<UIApplicationDidBecomeActiveCallback> *_applicationDidBecomeActiveCallbacks;
```
mariedm, line 37: nit: do we need to update/clean-up the array when an SDK instance is deallocated?

DatadogCore/Private/ObjcAppLaunchHandler.m
```diff
@@ -126,8 +128,11 @@ - (NSNumber *)timeToDidBecomeActive {
 }
 
 - (void)setApplicationDidBecomeActiveCallback:(UIApplicationDidBecomeActiveCallback)callback {
+    if (!callback) {
```
mariedm: nit: we could add a test for this use case

DatadogCore/Private/ObjcAppLaunchHandler.m
simaoseica-dd: Good point on the clean up. I'm removing the callbacks now after the first notification since we don't need them anymore for app launch.

DatadogCore/Private/ObjcAppLaunchHandler.m
simaoseica-dd: Added `nonnull callback` to the interface.

PR #2528, opened 2025-10-13
FFL-1217: Add timeout mechanism to wait for initial flags data read on first access

gonzalezreal:
### What and why?

 This PR adds a timeout mechanism to `FlagsRepository` that waits for the initial flags data read to complete before returning flag assignments. Previously, accessing flags immediately after creating a `FlagsClient` could return `nil` even when data existed in storage, because the async read operation hadn't completed yet.

### How?

- Added a `DispatchSemaphore` to `FlagsRepository` that blocks `flagAssignment(for:)` and `context` accessors until the initial read completes or a 100ms timeout expires
- The semaphore is signaled on a `.userInitiated` QoS queue to avoid priority inversion warnings
- Added a `hasReadFlagsData` flag for fast-path checking to avoid blocking on subsequent accesses

### Review checklist

### DIFF
--- a/DatadogFlags/Sources/Client/FlagsRepository.swift
+++ b/DatadogFlags/Sources/Client/FlagsRepository.swift
@@ -23,6 +23,10 @@ internal protocol FlagsRepositoryProtocol {
 }
 
 internal final class FlagsRepository {
+    private enum Constants {
+        static let readTimeout: TimeInterval = 0.1
+    }
+
     let clientName: String
 
     private let flagAssignmentsFetcher: any FlagAssignmentsFetching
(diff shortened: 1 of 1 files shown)


DatadogFlags/Sources/Client/FlagsRepository.swift
```diff
@@ -23,6 +23,10 @@ internal protocol FlagsRepositoryProtocol {
 }
 
 internal final class FlagsRepository {
+    private enum Constants {
+        static let readTimeout: TimeInterval = 0.1
```
ncreated, line 27: nit/ This could be `initialReadTimeout` or `firstReadTimeout` to better highlight the use case. For adjusting the actual value it might be worth looking at [this benchmark (internal)](https://datadoghq.atlassian.net/wiki/x/dAAj8w), especially results for the Context Queue.

PR #2529, opened 2025-10-13
FFL-1157: Add comprehensive DocC documentation to `DatadogFlags` public API

gonzalezreal:
### What and why?

This PR adds comprehensive DocC documentation to all public types, methods, and properties in the `DatadogFlags` module. The documentation enables developers to understand and use the Flags SDK effectively through inline documentation and code examples.

### How?

Added DocC-compliant documentation throughout the module.

### Review checklist

### DIFF
--- a/DatadogFlags/Sources/Flags.swift
+++ b/DatadogFlags/Sources/Flags.swift
@@ -77,6 +77,15 @@ public enum Flags {
         /// Default: `true`.
         public var rumIntegrationEnabled: Bool
 
+        /// Creates a configuration for the Datadog Flags feature.
+        ///
+        /// - Parameters:
+        ///   - gracefulModeEnabled: Controls error handling behavior for API misuse. Default: `true`.
+        ///   - customFlagsEndpoint: Custom server URL for retrieving flag assignments. Default: `nil`.
+        ///   - customFlagsHeaders: Additional HTTP headers for requests to `customFlagsEndpoint`. Default: `nil`.
+        ///   - customExposureEndpoint: Custom server URL for sending exposure data. Default: `nil`.
+        ///   - trackExposures: Enables exposure logging to the exposures intake endpoint. Default: `true`.
+        ///   - rumIntegrationEnabled: Enables the RUM integration for flag evaluations. Default: `true`.
         public init(
             gracefulModeEnabled: Bool = true,
             customFlagsEndpoint: URL? = nil,


DatadogFlags/Sources/Models/AnyValue.swift
```diff
@@ -6,13 +6,62 @@
 
 import Foundation
 
+/// A type-safe representation of JSON values for attributes and object flags.
+///
+/// `AnyValue` is used to represent arbitrary JSON data in feature flag contexts, particularly for:
```
sameerank, line 11: I think you can use double backticks for `AnyValue` in this file too? Looks like there's a pattern of using double backticks for all DocC references, including self-references within the same file. Some examples:

https://github.com/DataDog/dd-sdk-ios/blob/d1b304e3f3ae4f1330600888658b8a73591d6282/DatadogRUM/Sources/RUMMonitor/Scopes/Utils/ViewCache.swift#L10

https://github.com/DataDog/dd-sdk-ios/blob/57ecc918fa4586f7cbf8bd0c817c4b497330f620/DatadogCore/Sources/Core/Context/BatteryStatusPublisher.swift#L13

https://github.com/DataDog/dd-sdk-ios/blob/40c4736d8e7fd8cfa038a10537400cc825d1aa72/DatadogCore/Sources/Core/Context/ContextValuePublisher.swift#L132

DatadogFlags/Sources/Models/FlagDetails.swift
```diff
@@ -6,19 +6,79 @@
...
+    /// The requested feature flag was not found.
     case flagNotFound
+
+    /// The flag value type doesn't match the requested type.
     case typeMismatch
 }
 
+/// Detailed information about a feature flag evaluation.
+///
+/// `FlagDetails` contains both the evaluated flag value and metadata about the evaluation,
```
sameerank, line 25: Similar comment about using double backticks here

DatadogFlags/Sources/Models/AnyValue.swift
gonzalezreal: DocC uses double backticks to create symbol links in the generated documentation: https://www.swift.org/documentation/docc/formatting-your-documentation-content

IMO I wouldn't add self reference symbol links, and the examples you pointed out should be corrected to use single backticks.

DatadogFlags/Sources/Flags.swift
ncreated, line 1: suggestion/ Should we also add a docstring for the `public init()` method? We do this in other modules, and since our public documentation uses inline configuration in various snippets (`.init(...)`), having inline docstrings would be helpful. No need to repeat the full argument definitions - just a concise summary should suffice.

gonzalezreal: Good catch! Done.

PR #2536, opened 2025-10-16
[FFL-1241] support updated set of variation types

sameerank:
### What and why?

The `variationType` options have changed from uppercase `("STRING", "BOOLEAN", "NUMBER", "JSON")` to lowercase `("string", "boolean", "number", "integer", "float", "object")`. This change adds support for the new specific "integer" and "float" types while maintaining backwards compatibility with the legacy "number" type that older [browser SDK](https://github.com/DataDog/openfeature-js-client) releases still expect.

### How?

- Backwards compatibility: 
  - The `variationType` adjustment is not backwards compatible, but the change is safe because we have not released the product yet.
  - Maintains support for ambiguous "number" type that checks for `Int` first, then `Double`.
- Tests:
  - Updated test cases to cover the new variationType options and added test coverage for the legacy `"number"` type.
  - Verified locally that this branch works with the Shopist demo.

### Review checklist

### DIFF
--- a/DatadogFlags/Sources/Models/FlagAssignment.swift
+++ b/DatadogFlags/Sources/Models/FlagAssignment.swift
@@ -63,7 +63,8 @@ extension FlagAssignment: Codable {
         case "string":
             self.variation = try .string(container.decode(String.self, forKey: .variationValue))
         case "number":
-            // Check integer first, then double for backwards compatibility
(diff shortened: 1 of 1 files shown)


gonzalezreal (approved):
🚀 LGTM!

PR #2539, opened 2025-10-21
Fix OTel parent spans with multiple sequential children

### LINKED ISSUE #2530: Spans created with OpenTelemetry API don't properly connect with more than one child span
### Describe the bug

When using OpenTelemetry APIs to start active spans (via `SpanBuilder.setActive(true)`), if any child span is also activated in this manner, it will cause all subsequent child spans to fail to be part of the same trace.

It appears that this is the case because both the DataDog SDK and the OpenTelemetry API library rely on os_activity tracing to manage context, but activity scopes started by the SDK aren't ended until the underlying ActivityReference is deinitialized, which doesn't happen until the Span itself is dereferenced.

I have [a potential fix](https://github.com/DataDog/dd-sdk-ios/compare/develop...jbluntz:dd-sdk-ios:jbluntz/fix-multiple-child-spans) that I have validated against my test below as well as a running app; I can submit a PR for the same if you're open to it.

### Reproduction steps

```swift
    func testSetActive_givenParentWithMultipleChildren() throws {
        // Given
        let tracer: DatadogTracer = .mockWith(featureScope: featureScope)
        let parentSpan = tracer.spanBuilder(spanName: "Parent").setActive(true).startSpan()
        let childSpan = tracer.spanBuilder(spanName: "Child1").setActive(true).startSpan()
        childSpan.end()

        // When
        let secondChild = tracer.spanBuilder(spanName: "Child2").startSpan()
        secondChild.end()
        parentSpan.end()

        // Then
        let recordedSpans = try featureScope.spanEventsWritten()
        XCTAssertEqual(recordedSpans.count, 3)
        let recordedParent = recordedSpans.first { $0.resource == "Parent" }!
        let recordedChild1 = recordedSpans.first { $0.resource == "Child1" }!
        let recordedChild2 = recordedSpans.first { $0.resource == "Child2" }!
        
        XCTAssertEqual(recordedChild1.traceID, recordedParent.traceID)
        XCTAssertEqual(recordedChild1.parentID, recordedParent.spanID)
        
        //at time of writing, the following two asserts fail
        XCTAssertEqual(recordedChild2.traceID, recordedParent.traceID)
        XCTAssertEqual(recordedChild2.parentID, recordedParent.spanID)

    }

```

### SDK logs

_No response_

### Expected behavior

_No response_

### Affected SDK versions

3.0.0

none

### Did you confirm if the latest SDK version fixes the bug?

Yes

### Integration Methods

SPM

### Xcode Version

_No response_

### Swift Version

_No response_

### MacOS Version

_No response_

### Deployment Target

_No response_

### Device Information

_No response_

_No response_

jbluntz:
### What and why?

This PR addresses two issues with managing Active DDSpans:
1. When using OpenTelemetry APIs to start active spans (via SpanBuilder.setActive(true)), if any child span is also activated in this manner, it will cause all subsequent child spans to fail to be part of the same trace.  
2. When a span is activated multiple times in succession with other spans, old ActivityReferences can linger in the ActiveSpansPool, resulting in a span being returned as the active span even after being finished.

### How?

This PR addresses the issue by immediately leaving the span's ActivityReference scope and removing all references to ActivityReferences when finishing a DDSpan.

Fixes #2530 

### Review checklist

### DIFF
--- a/DatadogTrace/Sources/Utils/ActiveSpansPool.swift
+++ b/DatadogTrace/Sources/Utils/ActiveSpansPool.swift
@@ -14,8 +14,7 @@ internal let OS_ACTIVITY_CURRENT = unsafeBitCast(dlsym(UnsafeMutableRawPointer(b
 /// Used to reference the active span in the current execution context.
 internal class ActivityReference {
     let activityId: os_activity_id_t
-    fileprivate var activityState = os_activity_scope_state_s()
-    
+    private var activityState = os_activity_scope_state_s()
     private var isActive = true
 
     init() {
@@ -25,14 +24,16 @@ internal class ActivityReference {
         os_activity_scope_enter(activity, &activityState)
     }
 
-    deinit {
-        leave()
-    }
-    
     func leave() {
-        guard isActive else { return }
+        guard isActive else {
+            return
+        }
         isActive = false
         os_activity_scope_leave(&activityState)
+    }
+
+    deinit {
+        leave()
     }
 }
 
@@ -40,6 +41,12 @@ internal class ActivityReference {
 internal class ActiveSpansPool {
     private var contextMap = [os_activity_id_t: DDSpan]()
     private let rlock = NSRecursiveLock()
+
+    var isEmpty: Bool {
+        rlock.lock()
+        defer { rlock.unlock() }
+        return contextMap.isEmpty
+    }
 
     /// Returns the Span from the current context
     func getActiveSpan() -> DDSpan? {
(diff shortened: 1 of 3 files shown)


DatadogTrace/Sources/Utils/ActiveSpansPool.swift
```diff
@@ -51,9 +66,11 @@ internal class ActiveSpansPool {
         rlock.unlock()
     }
 
-    func removeSpan(activityReference: ActivityReference) {
+    func removeSpan(span: DDSpan) {
         rlock.lock()
-        contextMap[activityReference.activityId] = nil
+        contextMap = contextMap.filter { key, value in
+            value.ddContext.spanID != span.ddContext.spanID
+        }
```
gonzalezreal, line 73: minor (non-blocking): `filter` is O(n) and creates a new dictionary vs the previous O(1) direct key removal. AFAIK this is totally fine as `contextMap` should be a small dictionary. cc/ @simaoseica-dd

simaoseica-dd: The current design does not help to optimize for O(1) if we want to clean up all invalid spans.

PR #2540, opened 2025-10-22
fix: omit failed flags, don't fail the whole config

typotter:
### What and why?

This PR implements graceful handling of invalid flag assignments during JSON decoding. When the flag assignments response contains flags with unknown/unsupported `variationType` values (e.g., future variation types not yet supported by the SDK), the SDK now:
- Successfully decodes all valid flags
- Tracks failed flags separately
- Logs failures to telemetry for observability
- Continues functioning without disruption

**Why this matters**: As the feature flagging backend evolves and adds new variation types, older SDK versions should gracefully handle flags they don't understand rather than failing entirely. This maintains the SDK's stability principle and prevents complete flag assignment fetch failures due to a single incompatible flag.

### How?

**1. Partial Decoding with `DynamicCodingKey`** (`FlagAssignmentsResponse.swift`)
- Replaced bulk dictionary decoding with individual flag iteration
- Used `DynamicCodingKey` to iterate through dynamic flag names in JSON
- Wrapped each flag decode in try-catch to isolate failures
- Stored successful flags in `flags` dictionary and failures in `failedFlags` dictionary with error descriptions

**2. Telemetry Logging** (`FlagAssignmentsFetcher.swift`)
- Added logging loop after successful response decode
- Reports each failed flag to both `DD.logger` (warn level) and `featureScope.telemetry` (error level)
- Provides visibility into which flags failed and why

**3. Comprehensive Test Coverage** (`FlagAssignmentsResponseTests.swift`)
- Added `testDecodingWithBrokenFlags_healsFromInvalidVariationType()`
- Tests JSON with mix of valid flags (string, boolean, integer) and invalid flags (unknown variation types)
- Verifies valid flags decode correctly
- Verifies invalid flags are captured in `failedFlags` with descriptive error messages

### Review checklist

### DIFF
--- a/DatadogFlags/Sources/Client/FlagAssignmentsFetcher.swift
+++ b/DatadogFlags/Sources/Client/FlagAssignmentsFetcher.swift
@@ -85,7 +85,10 @@ internal final class FlagAssignmentsFetcher: FlagAssignmentsFetching {
                                     )
                                     featureScope.telemetry.debug(
                                         "Failed to decode flag '\(flagKey)' from flag assignments response",
-                                        error: error
+                                        attributes: [
+                                            "flagKey": flagKey,
+                                            "errorDescription": errorDescription
+                                        ]
                                     )
                                 }
                             }


DatadogFlags/Sources/Models/FlagAssignment.swift
```diff
@@ -109,6 +108,9 @@ extension FlagAssignment: Codable {
         case .object(let value):
             try container.encode("object", forKey: .variationType)
             try container.encode(value, forKey: .variationValue)
+        case .unknown(let typeName):
+            try container.encode(typeName, forKey: .variationType)
+            try container.encodeNil(forKey: .variationValue)
```
gonzalezreal, line 114:
```suggestion
        case .unknown(let typeName):
            // Should never happen as unknown variations are filtered out
            try container.encode(typeName, forKey: .variationType)
            try container.encodeNil(forKey: .variationValue)
```

DatadogFlags/Sources/Client/FlagAssignmentsFetcher.swift
```diff
@@ -74,6 +74,22 @@ internal final class FlagAssignmentsFetcher: FlagAssignmentsFetching {
...
+                                for (flagKey, errorDescription) in response.failedFlags {
+                                    let error = InternalError(description: errorDescription)
+                                    DD.logger.warn(
+                                        "Failed to decode flag '\(flagKey)' from flag assignments response. Flag will be dropped from configuration.",
+                                        error: error
+                                    )
+                                    featureScope.telemetry.error(
+                                        "Failed to decode flag '\(flagKey)' from flag assignments response",
+                                        error: error
+                                    )
```
gonzalezreal: minor: I would use `debug` instead of error for the telemetry, since this is a situation we expect in the future.
```suggestion
                                    featureScope.telemetry.debug(
                                        "Failed to decode flag '\(flagKey)' from flag assignments response",
                                        error: error
                                    )
```

DatadogFlags/Tests/Models/FlagAssignmentsResponseTests.swift
```diff
@@ -153,4 +153,148 @@ final class FlagAssignmentsResponseTests: XCTestCase {
             )
         )
     }
+
+    func testDecodingWithBrokenFlags_healsFromInvalidVariationType() throws {
```
gonzalezreal: suggestion: Match existing naming patterns

```suggestion
    func testDecodingFlagAssignmentsResponseWithUnknownVariationTypes() throws {
```

DatadogFlags/Sources/Client/FlagAssignmentsFetcher.swift
typotter: good point!

typotter:
Hey @gonzalezreal 
I made a quick change to the `debug` log call. PTAL

PR #2549, opened 2025-10-28
RUM-7999: Prevent sampled out spans from sending

maciejburda:
### What and why?

This code aligns the implementation on iOS with expectations and lays down foundation for SDK level traces metrics in the future.

Additionally it expands root span creation API with customisable sampling rate that will override global configuration in the context of the single Trace.

### How?

Key change is on Span `finish()` function that now doesn't forward the span for writing when `isKept == false`.

### Review checklist


DatadogTrace/Sources/OpenTracing/OTTracer.swift
```diff
@@ -34,7 +35,8 @@ public protocol OTTracer {
     func startRootSpan(
         operationName: String,
         tags: [String: Encodable]?,
-        startTime: Date?
+        startTime: Date?,
+        customSamplingRate: SampleRate?
```
ncreated: As per our [conventions](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3027638506/Public+API+Convention+in+iOS+SDK+2.x#Rule-7%3A-Use-of-standard-Datadog-and-RUM-terminology), `customSamplingRate` is not coherent with other APIs. It should be `customSampleRate`

Also, this parameter must be explained in inline documentation - what it does and what sampling does it override.

PR #2552, opened 2025-10-29
RUM-12441 Graphql errors

barboraplasovska:
### What and why?

GraphQL APIs follow a different error model than REST APIs, they return HTTP `200 OK` status codes even when operations fail with errors included in the response body under an `errors` key. Without parsing the response body, these errors would go unnoticed in RUM monitoring.
This PR adds support for automatically detecting, parsing, and reporting GraphQL errors in RUM resource events. When a GraphQL request completes with errors in the response, the SDK now captures and includes them in the resource event.

### How?

When a request completes with a JSON response, we check for the presence of an `errors` key. If found, the response body is passed along to the resource scope where it's decoded according to the GraphQL spec and populated into the resource event's `graphql.errors` array and `graphql.errorCount` fields. (cc [PR](https://github.com/DataDog/rum-events-format/pull/308) on `rum_events_format`).

### Review checklist

### DIFF
--- a/DatadogRUM/Tests/Instrumentation/Resources/URLSessionRUMResourcesHandlerTests.swift
+++ b/DatadogRUM/Tests/Instrumentation/Resources/URLSessionRUMResourcesHandlerTests.swift
@@ -1183,7 +1183,6 @@ class URLSessionRUMResourcesHandlerTests: XCTestCase {
 
         XCTAssertNil(stopResourceCommand?.attributes[CrossPlatformAttributes.graphqlErrors])
     }
-    }
 
     // MARK: - Helper Methods
 


DatadogRUM/Tests/RUMMonitor/Scopes/RUMResourceScopeTests.swift
```diff
@@ -9,6 +9,20 @@ import DatadogInternal
...
+        case (.string(let lhsValue), .string(let rhsValue)):
+            return lhsValue == rhsValue
+        case (.integer(let lhsValue), .integer(let rhsValue)):
+            return lhsValue == rhsValue
+        default:
+            return false
+        }
+    }
+}
+
```
barboraplasovska, line 25: I added this extension to be able to do `XCTAssertEqual` on a `Path`. Not sure if there is a better way to make this testable.

git-thuerk-done (dismissed):

DatadogRUM/Sources/DataModels/GraphQLResponse.swift
```diff
@@ -0,0 +1,86 @@
...
+
+/// Lightweight decoder to check if a GraphQL response contains errors.
+/// Only checks for the presence of the "errors" key without decoding the entire array.
+internal struct GraphQLResponseHasErrors: Decodable {
+    let hasErrors: Bool
+    private enum CodingKeys: String, CodingKey { case errors }
+
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        hasErrors = container.contains(.errors)
```
maxep, line 19: smart!

DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
```diff
@@ -169,6 +174,33 @@ internal final class URLSessionRUMResourcesHandler: DatadogURLSessionHandler, RU
             )
         }
     }
+
+    /// Extracts GraphQL errors from JSON response if present.
+    /// Only the errors array is extracted to avoid storing potentially large response data fields.
+    private func extractGraphQLErrorsIfPresent(from interception: URLSessionTaskInterception) -> String? {
```
maxep: change-request/ It is suboptiomal to `JSONDecoder` + `JSONSerialization` + `String`. Lets just check for `GraphQLResponseHasErrors` and pass the `data` directly to `combinedAttributes` and do the decoding in `RUMResourceScope`. `Data` is `Codable` so you can set it directly as attribute.

maxep (changes requested):

DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
barboraplasovska: Done

maxep: You no longer need any of this:
```swift
// Extract only the errors array from the response
guard let jsonObject = try? JSONSerialization.jsonObject(with: data) as? [String: Any],
      let errors = jsonObject["errors"],
      let errorsData = try? JSONSerialization.data(withJSONObject: ["errors": errors]) else {
    return nil
}

return errorsData
```
You can simply do `return data` ;)

maxep (dismissed):
Looks great, well done!

mariedm (dismissed):

barboraplasovska:
Forced pushed to resolve merge conflicts with the develop branch.

rtrieu (approved):

rtrieu:
Hi @barboraplasovska is this ready to merge?

PR #2559, opened 2025-10-31
RUM-12420 Add convenient methods to navigate KSCrash reports

maxep:
### What and why?

`KSCrash` reports are untyped `[String: Any]` dictionaries, making it difficult and error-prone to navigate.
This PR introduce a type-safe dictionary wrapper that uses `KSCrash`'s `CrashField` enum for keys and provides validated accessors for nested values.

### How?

- `CrashFieldDictionary`:
A type alias for `[CrashField: Any]` that provides convenient methods to safely access nested crash report values. It converts untyped string-keyed dictionaries into strongly-typed dictionaries using `KSCrash`'s `CrashField` enum, supporting both optional (`valueIfPresent`) and required (`value`) access patterns with variadic key paths like `dict.value(forKey: .crash, .error, .type)`.

- `DatadogTypeSafeFilter`:
A `KSCrash` filter that performs the conversion from raw crash reports to `CrashFieldDictionary` format. It's placed first in the filter chain to ensure all subsequent filters benefit from type-safe access.

- `AnyCrashReport`:
A generic wrapper that allows arbitrary values to conform to `KSCrash`'s `CrashReport` protocol, enabling the typed dictionaries to flow through the filter pipeline.

### Review checklist

### DIFF
--- a/DatadogCrashReporting/Sources/KSCrashIntegration/DatadogTypeSafeFilter.swift
+++ b/DatadogCrashReporting/Sources/KSCrashIntegration/DatadogTypeSafeFilter.swift
@@ -47,14 +47,14 @@ internal final class DatadogTypeSafeFilter: NSObject, CrashReportFilter {
     ///                   fails for any report (e.g., invalid structure), `nil` is returned
     ///                   along with the error.
     func filterReports(
-        _ reports: [KSCrashRecording.CrashReport],
-        onCompletion: (([KSCrashRecording.CrashReport]?, (Error)?) -> Void)?
+        _ reports: [CrashReport],
+        onCompletion: (([CrashReport]?, (Error)?) -> Void)?
     ) {
         do {
             let reports = try reports.map { report in
                 // Validate and extract the raw report dictionary
                 // KSCrash reports come as untyped dictionaries, we need to ensure it's valid
-                guard let report = report as? KSCrashRecording.CrashReportDictionary else {
+                guard let report = report as? CrashReportDictionary else {
                     throw CrashReportException(description: "KSCrash report untypedValue is not a dictionary")
                 }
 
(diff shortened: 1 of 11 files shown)


DatadogCrashReporting/Sources/PLCrashReporterIntegration/CrashReportInfo.swift
```diff
@@ -11,7 +11,7 @@ import CrashReporter
 ///
 /// It implements preliminary consistency check for Objective-C `PLCrashReport` and provides additional
 /// type-safety for accessing its implicitly unwrapped optional values.
-internal struct CrashReport {
+internal struct CrashReportInfo {
```
maxep, line 14: Avoid conflict with `KSCrash`'s `CrashReport`. We cannot use `KSCrashRecording.` prefix due to CocoaPods specific import :(

DatadogCrashReporting/Sources/KSCrashIntegration/DatadogTypeSafeFilter.swift
```diff
@@ -0,0 +1,71 @@
...
+    /// the same type safety guarantees.
+    ///
+    /// - Parameters:
+    ///   - reports: The raw KSCrash reports to convert
+    ///   - onCompletion: Completion handler called with type-safe reports. If conversion
+    ///                   fails for any report (e.g., invalid structure), `nil` is returned
+    ///                   along with the error.
+    func filterReports(
+        _ reports: [CrashReport],
+        onCompletion: (([CrashReport]?, (Error)?) -> Void)?
```
simaoseica-dd, line 51: question/ why do you need a completion instead of a normal return and throw?

maxep: This is part of `KSCrash` interface: `CrashReportFilter`.
I think the reasoning is that this can be async/escaping.

PR #2561, opened 2025-10-31
RUM-12420 Add KSCrash report minifier

maxep:
### What and why?

This PR adds `DatadogMinifyFilter` for the `KSCrash` integration, which reduces crash report size by removing unnecessary data. This is equivalent to the existing `CrashReportInfoMinifier` used in the `PLCrashReporter` integration. The filter removes unreferenced binary images and limits stack traces to 200 frames to optimize payload size for transmission.

### How?

The filter operates on type-safe `CrashFieldDictionary` crash reports and performs two optimizations:
1.  First, it collects all `object_addr` values from thread backtraces (including main crash and recrash reports) and filters the `binary_images` array to keep only those whose `image_addr` appears in the collected set.
2. Second, when a thread's backtrace exceeds 200 frames, it removes frames from the middle of the stack trace to preserve the most important top and bottom frames, and marks the backtrace as truncated.

### Review checklist

### DIFF
--- a/DatadogCrashReporting/Sources/KSCrashIntegration/DatadogMinifyFilter.swift
+++ b/DatadogCrashReporting/Sources/KSCrashIntegration/DatadogMinifyFilter.swift
@@ -111,7 +111,8 @@ internal final class DatadogMinifyFilter: NSObject, CrashReportFilter {
     }
 
     /// Removes less important stack frames to ensure that their count is equal or below `stackFramesLimit`.
-    /// Frames are removed at the middle of stack trace, which preserves the most important upper and bottom frames.
+    /// Frames are removed at the middle of stack trace, which preserves the most important upper and bottom frames:
+    /// Top frames show the crash site and bottom frames the entry point.
     private func limit(backtrace frames: inout [CrashFieldDictionary]) -> Bool {
         guard frames.count > stackFramesLimit else {
             return false


DatadogCrashReporting/Sources/KSCrashIntegration/DatadogMinifyFilter.swift
```diff
@@ -0,0 +1,128 @@
...
+    func minify(report: CrashFieldDictionary) throws -> CrashFieldDictionary {
+        var minifiedReport = report
+        var objectAddresses: Set<Int64> = []
+
+        if var threads: [CrashFieldDictionary] = try report.valueIfPresent(forKey: .crash, .threads) {
+            threads = try threads.map { try self.minify(thread: $0, objectAddresses: &objectAddresses) }
+            minifiedReport.setValue(forKey: .crash, .threads, value: threads)
+        }
+
+        if var threads: [CrashFieldDictionary] = try report.valueIfPresent(forKey: .recrashReport, .crash, .threads) {
```
maxep, line 81: `KSCrash` also report 're-crash' which are crashes that occurs in KSCrash itself or during the handling. We don't have the pipeline for these yet but they could be reported as telemetry error.

DatadogCrashReporting/Sources/KSCrashIntegration/DatadogMinifyFilter.swift
```diff
@@ -0,0 +1,128 @@
...
+                guard let dict = report.untypedValue as? CrashFieldDictionary else {
+                    throw CrashReportException(description: "KSCrash report untypedValue is not a CrashDictionary")
+                }
+
+                return try AnyCrashReport(minify(report: dict))
+            }
+
+            onCompletion?(reports, nil)
+        } catch {
+            onCompletion?(reports, error)
```
mariedm, line 54: Should we return `reports` or `nil` in case of failure?

DatadogCrashReporting/Sources/KSCrashIntegration/DatadogMinifyFilter.swift
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        var thread = thread
+        thread.setValue(forKey: .backtrace, .contents, value: backtrace)
+        thread.setValue(forKey: .backtrace, .truncated, value: true)
+        return thread
+    }
+
+    /// Removes less important stack frames to ensure that their count is equal or below `stackFramesLimit`.
+    /// Frames are removed at the middle of stack trace, which preserves the most important upper and bottom frames.
```
mariedm: nit: we could explain why -> top frames show the crash site and bottom frames the entry ploint

DatadogCrashReporting/Sources/KSCrashIntegration/DatadogMinifyFilter.swift
mariedm: Nice!

mariedm (dismissed):
LGMT!

DatadogCrashReporting/Sources/KSCrashIntegration/DatadogMinifyFilter.swift
maxep: Good question! I think we can passthrough the reports to give context to the error. Worst case: this will be dropped downstream

PR #2562, opened 2025-10-31
RUM-12420 Add KSCrash report diagnostic

maxep:
### What and why?

This PR adds `DatadogDiagnosticFilter` for the `KSCrash` integration, which generates human-readable diagnostic messages for crash reports. The filter analyzes crash data to determine the crash type (uncaught exceptions or signal-based crashes) and produces descriptive messages that explain the cause in plain language. This mirrors the diagnostic generation already present in `DDCrashReportExporter` for the `PLCrashReporter` integration, and is equivalent to `KSCrash`'s `KSCrashReportFilterDoctor`, ensuring consistent crash descriptions across both crash reporting backends.

### How?

The filter processes crash reports by analyzing the error data within each report:
- For uncaught `NSException` crashes, it extracts the exception name and reason to generate messages like `Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'unrecognized selector'.`
- For signal-based crashes, it maps Unix signal names (`SIGSEGV`, `SIGABRT`, etc.) to human-readable descriptions using a predefined dictionary, generating messages like `Application crash: `SIGSEGV` (Segmentation fault)`. 
 
The filter injects these diagnostic messages into both the main crash report and any recrash reports present, with each report receiving a diagnosis based on its own error data.

### Review checklist

### DIFF
--- a/DatadogCrashReporting/Sources/KSCrashIntegration/DatadogDiagnosticFilter.swift
+++ b/DatadogCrashReporting/Sources/KSCrashIntegration/DatadogDiagnosticFilter.swift
@@ -50,9 +50,6 @@ internal final class DatadogDiagnosticFilter: NSObject, CrashReportFilter {
     /// Placeholder text used when crash information is unavailable.
     private let unknown = "<unknown>"
 
-    /// Placeholder text used for optional fields that are not present.
-    private let unavailable = "???"
-
     /// Mapping of Unix signal names to human-readable descriptions.
     ///
     /// This dictionary provides user-friendly descriptions for all standard POSIX signals,
@@ -114,11 +111,13 @@ internal final class DatadogDiagnosticFilter: NSObject, CrashReportFilter {
                     throw CrashReportException(description: "KSCrash report untypedValue is not a CrashDictionary")
                 }
 
-                if let crash: CrashFieldDictionary = try dict.valueIfPresent(forKey: .crash), let diagnosis = try diagnose(crash: crash) {
+                if let crash: CrashFieldDictionary = try dict.valueIfPresent(forKey: .crash) {
+                    let diagnosis = try diagnose(crash: crash)
                     dict.setValue(forKey: .crash, .diagnosis, value: diagnosis)
                 }
 
-                if let crash: CrashFieldDictionary = try dict.valueIfPresent(forKey: .recrashReport, .crash), let diagnosis = try diagnose(crash: crash) {
+                if let crash: CrashFieldDictionary = try dict.valueIfPresent(forKey: .recrashReport, .crash) {
+                    let diagnosis = try diagnose(crash: crash)
                     dict.setValue(forKey: .recrashReport, .crash, .diagnosis, value: diagnosis)
                 }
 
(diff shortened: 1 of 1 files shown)


DatadogCrashReporting/Sources/KSCrashIntegration/DatadogDiagnosticFilter.swift
```diff
@@ -0,0 +1,170 @@
...
+    /// - Parameter dict: The crash report dictionary containing crash information
+    /// - Returns: A diagnostic message describing the crash, or `nil` if no crash data is found
+    /// - Throws: `CrashReportException` if the report structure is invalid
+    ///
+    /// ## Example Messages
+    ///
+    /// - Exception: `"Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'unrecognized selector'."`
+    /// - Signal: `"Application crash: SIGSEGV (Segmentation fault)"`
+    /// - Unknown: `"Application crash: <unknown>"`
+    func diagnose(crash dict: CrashFieldDictionary) throws -> String? {
```
mariedm: nit: it seems that this method is always returning a string 
```suggestion
    func diagnose(crash dict: CrashFieldDictionary) throws -> String {
```

mariedm: It seems that this method is always returning a `String`.

```suggestion
    func diagnose(crash dict: CrashFieldDictionary) throws -> String {
```

DatadogCrashReporting/Sources/KSCrashIntegration/DatadogDiagnosticFilter.swift
```diff
@@ -0,0 +1,170 @@
...
+        // Check for uncaught exception
+        if let exception: CrashFieldDictionary = try dict.valueIfPresent(forKey: .error, .nsException) {
+            let name: String = try exception.valueIfPresent(forKey: .name) ?? unknown // e.g. `NSInvalidArgumentException`
+            let reason = try dict.valueIfPresent(forKey: .error, .reason) ?? unknown // e.g. `-[NSObject objectForKey:]: unrecognized selector sent to instance 0x...`
+            return "Terminating app due to uncaught exception '\(name)', reason: '\(reason)'."
+        }
+
+        // Check for signal-based crash with known description
+        if
```
mariedm, line 160: Could we have a name without a description? In which case the description could fallback to unknown so we always have a name. Or both name and description have fallback, like you did above for exceptions.

DatadogCrashReporting/Sources/KSCrashIntegration/DatadogDiagnosticFilter.swift
```diff
@@ -0,0 +1,170 @@
...
+/// This filter should be placed early in the KSCrash filter chain, before the
+/// `DatadogCrashReportFilter`, to ensure diagnostic information is available for
+/// subsequent processing and reporting.
+internal final class DatadogDiagnosticFilter: NSObject, CrashReportFilter {
+    /// Placeholder text used when crash information is unavailable.
+    private let unknown = "<unknown>"
+
+    /// Placeholder text used for optional fields that are not present.
+    private let unavailable = "???"
+
```
mariedm: This variable doesn't seem to be used.

```suggestion
```

mariedm (commented):
Thanks for the detailed doc! Only a few nits & questions, otherwise LGTM.

DatadogCrashReporting/Sources/KSCrashIntegration/DatadogDiagnosticFilter.swift
maxep: I was wondering as well :D I'm replicating the exact same logic as [PLCrashReporter integration](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogCrashReporting/Sources/PLCrashReporterIntegration/DDCrashReportExporter.swift#L93-L102), to not break Error Tracking fingerprinting for grouping.

PR #2565, opened 2025-11-04
RUM-12420 Export KSCrash report to Datadog format

maxep:
### What and why?

This PR introduces the `DatadogCrashReportFilter` that converts `KSCrash` crash reports into Datadog's internal `DDCrashReport` format. This filter acts as a transformation layer in the crash reporting pipeline.

### How?

Implements `KSCrash`'s `CrashReportFilter` protocol to convert crash reports from dictionary-based format to Datadog's type-safe` DDCrashReport` structure. The filter extracts crash metadata (timestamp, signal info), stack traces from all threads, binary images with load addresses, system context (OS, device, architecture), and user attributes. 

### Review checklist

### DIFF
--- a/DatadogCrashReporting/Sources/KSCrashIntegration/DatadogCrashReportFilter.swift
+++ b/DatadogCrashReporting/Sources/KSCrashIntegration/DatadogCrashReportFilter.swift
@@ -37,6 +37,14 @@ internal import KSCrashRecording
 /// If a report cannot be converted (invalid format, missing required fields),
 /// the original report is passed through and an error is provided to the completion handler.
 internal final class DatadogCrashReportFilter: NSObject, CrashReportFilter {
+    // Parse timestamp with fractional seconds support
+    // KSCrash timestamps use ISO8601 format with microsecond precision (e.g., "2025-10-22T14:14:12.007336Z")
+    let dateFormatter = {
+        let formatter = ISO8601DateFormatter()
+        formatter.formatOptions = [.withInternetDateTime, .withFractionalSeconds]
+        return formatter
+    }()
+
     /// Filters and converts crash reports to Datadog's format.
     ///
     /// This method processes each report by converting it from KSCrash's format
@@ -74,11 +82,7 @@ internal final class DatadogCrashReportFilter: NSObject, CrashReportFilter {
             throw CrashReportException(description: "KSCrash report untypedValue is not a CrashDictionary")
         }
 
-        // Parse timestamp with fractional seconds support
-        // KSCrash timestamps use ISO8601 format with microsecond precision (e.g., "2025-10-22T14:14:12.007336Z")
-        let formatter = ISO8601DateFormatter()
-        formatter.formatOptions = [.withInternetDateTime, .withFractionalSeconds]
-        let date = try formatter.date(from: dict.value(forKey: .report, .timestamp))
+        let date = try dateFormatter.date(from: dict.value(forKey: .report, .timestamp))
 
         // Extract required crash and system information
         let system: CrashFieldDictionary = try dict.value(forKey: .system)
@@ -118,14 +122,11 @@ internal final class DatadogCrashReportFilter: NSObject, CrashReportFilter {
             }
 
             // Detect system libraries based on path patterns
-            // In device: user images are in /var/containers/Bundle/Application/
-            // System images are in /System/, /usr/lib/, or framework paths
             #if targetEnvironment(simulator)
-            // In simulator: system images are in Xcode.app/Contents/Developer/Platforms/
-            // In Xcode 26: /**/*/iOS *.simruntime
+            // Simulator: system images are in Xcode.app/Contents/Developer/Platforms/ or .simruntime bundles (Xcode 16+)
             let isSystemImage = path.contains("/Contents/Developer/Platforms/") || path.contains("simruntime")
             #else
-            // In device: user images contain /Bundle/Application/
+            // Device: user images are in /var/containers/Bundle/Application/, everything else is system
             let isSystemImage = !path.contains("/Bundle/Application/")
             #endif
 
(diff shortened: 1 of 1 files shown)


DatadogCrashReporting/Sources/KSCrashIntegration/DatadogCrashReportFilter.swift
```diff
@@ -0,0 +1,202 @@
...
+    ///                   receives converted reports array (may include original reports
+    ///                   for items that failed conversion).
+    func filterReports(
+        _ reports: [CrashReport],
+        onCompletion: (([CrashReport]?, (Error)?) -> Void)?
+    ) {
+        do {
+            try onCompletion?(reports.map(parse(report:)), nil)
+        } catch {
+            onCompletion?(reports, error)
```
mariedm, line 67: So if one report fails to convert, all reports are returned unconverted; is this what we want?

DatadogCrashReporting/Sources/KSCrashIntegration/DatadogCrashReportFilter.swift
```diff
@@ -0,0 +1,202 @@
...
+
+        // Extract required crash and system information
+        let system: CrashFieldDictionary = try dict.value(forKey: .system)
+        let crash: CrashFieldDictionary = try dict.value(forKey: .crash)
+        let arch: String = try system.value(forKey: .cpuArch)
+
+        // Build crash type string from signal information
+        // Format: "SIGNAL_NAME (SIGNAL_CODE)" e.g., "SIGSEGV (SEGV_MAPERR)"
+        let signalName: String = try crash.valueIfPresent(forKey: .error, .signal, .name) ?? "<unknown>"
+        let signalCodeName: String = try crash.valueIfPresent(forKey: .error, .signal, .codeName) ?? "#0" // PLCrashReporter defaults to #0
```
mariedm, line 95: > PLCrashReporter defaults to #0

Is it the same for KSCrash?

DatadogCrashReporting/Sources/KSCrashIntegration/DatadogCrashReportFilter.swift
```diff
@@ -0,0 +1,202 @@
...
+        // Validate and extract the type-safe report dictionary
+        guard let dict = report.untypedValue as? CrashFieldDictionary else {
+            throw CrashReportException(description: "KSCrash report untypedValue is not a CrashDictionary")
+        }
+
+        // Parse timestamp with fractional seconds support
+        // KSCrash timestamps use ISO8601 format with microsecond precision (e.g., "2025-10-22T14:14:12.007336Z")
+        let formatter = ISO8601DateFormatter()
+        formatter.formatOptions = [.withInternetDateTime, .withFractionalSeconds]
+        let date = try formatter.date(from: dict.value(forKey: .report, .timestamp))
```
mariedm: `DateFormatter` doesn't throw, does it?

DatadogCrashReporting/Sources/KSCrashIntegration/DatadogCrashReportFilter.swift
```diff
@@ -0,0 +1,202 @@
...
+        // Transform thread information with stack traces
+        // Each thread contains a backtrace showing the call stack at crash time
+        let threads: [DDThread] = try crash.value([CrashFieldDictionary].self, forKey: .threads).map { thread in
+            // Format each stack frame: "index objectName instructionAddr symbolAddr + offset"
+            let backtrace: [String] = try thread.value([CrashFieldDictionary].self, forKey: .backtrace, .contents).enumerated().compactMap { index, frame in
+                guard
+                    let symbolAddr: Int64 = try frame.valueIfPresent(forKey: .symbolAddr),
+                    let objectName: NSString = try frame.valueIfPresent(forKey: .objectName)
+                else {
+                    return nil
```
mariedm: I wonder if we should fallback to placeholders when values are missing instead of dropping the frame entirely?

DatadogCrashReporting/Sources/KSCrashIntegration/DatadogCrashReportFilter.swift
```diff
@@ -0,0 +1,202 @@
...
+    /// - Throws: `CrashReportException` if the report format is invalid or required fields are missing
+    func parse(report: CrashReport) throws -> CrashReport {
+        // Validate and extract the type-safe report dictionary
+        guard let dict = report.untypedValue as? CrashFieldDictionary else {
+            throw CrashReportException(description: "KSCrash report untypedValue is not a CrashDictionary")
+        }
+
+        // Parse timestamp with fractional seconds support
+        // KSCrash timestamps use ISO8601 format with microsecond precision (e.g., "2025-10-22T14:14:12.007336Z")
+        let formatter = ISO8601DateFormatter()
```
mariedm: We could make this a static property instead of recreating on every parse

DatadogCrashReporting/Sources/KSCrashIntegration/DatadogCrashReportFilter.swift
```diff
@@ -0,0 +1,202 @@
...
+            // In device: user images are in /var/containers/Bundle/Application/
+            // System images are in /System/, /usr/lib/, or framework paths
+            #if targetEnvironment(simulator)
+            // In simulator: system images are in Xcode.app/Contents/Developer/Platforms/
+            // In Xcode 26: /**/*/iOS *.simruntime
+            let isSystemImage = path.contains("/Contents/Developer/Platforms/") || path.contains("simruntime")
+            #else
+            // In device: user images contain /Bundle/Application/
+            let isSystemImage = !path.contains("/Bundle/Application/")
+            #endif
```
mariedm, line 131: Comments seem a bit misplaced here, making it a bit confusing.

Also, for the device case, it says that "System images are in /System/, /usr/lib/, or framework paths", so shouldn't we check these paths instead?

Lastly, given that this seems a bit fragile (it changed with Xcode 26), do we want to add logs or telemetry in case of failure?

DatadogCrashReporting/Sources/KSCrashIntegration/DatadogCrashReportFilter.swift
maxep: No, in `KSCrash` it defaults to `nil`. But since we want the exact same reports as we have with `PLCrashReporter`, I'm defaulting to ` ?? "#0"` here.

DatadogCrashReporting/Sources/KSCrashIntegration/DatadogCrashReportFilter.swift
maxep: Indeed, but `dict.value(forKey: .report, .timestamp)` does ;)

DatadogCrashReporting/Sources/KSCrashIntegration/DatadogCrashReportFilter.swift
maxep: Good catch, this is not aligned with PLCrashReport exporter!

DatadogCrashReporting/Sources/KSCrashIntegration/DatadogCrashReportFilter.swift
maxep: Yes, even if it's not used as of now it's best to escalate the reports still. Either so they can be picked up by a following filter if applicable, or to add context to the error

DatadogCrashReporting/Sources/KSCrashIntegration/DatadogCrashReportFilter.swift
maxep: Comments updated 👍 They were actually wrong

Indeed, it's a bit fragile but I'm migrating the existing logic from PLCrashReporter :/ And there is no failure to log, it's either true or false...

mariedm: Thanks, much clearer!

PR #2570, opened 2025-11-06
RUM-12323: Update early launch profiling strategy

simaoseica-dd:
### What and why?

This PR updates the strategy for early app-launch profiling.
The strategy now relies on a fixed 10% sampling rate (to be dogfooded) and the Profiling only starts if it was enabled by the app.

### How?

This PR introduces a new lightweight `DataStore` called `UserDefaultsDataStore` that handles persistence through the `UserDefaults`. Profiling starts very early, before a `FeatureDataStore` exists, so this store provides the required persistence at launch.

This data store will help check if Profiling was enabled by the customers in the app. Before starting an early profile, it will read the persisted flag indicating whether the customers have enabled Profiling via `Profiling.enable()` in the previous session. The previous user facing plist is no longer needed for these decisions.
Additionally, the profile is only uploaded if the stored flag is present and true (`is_profiling_enabled:true`). 

### Review checklist

### DIFF
--- a/DatadogProfiling/Mach/ctor_profiler.cpp
+++ b/DatadogProfiling/Mach/ctor_profiler.cpp
@@ -28,10 +28,6 @@ namespace dd::profiler { class ctor_profiler; }
 
 static dd::profiler::ctor_profiler* g_ctor_profiler = nullptr;
 
-#ifdef __cplusplus
-extern "C" {
-#endif
-
 /**
  * Checks if the current process was launched via pre-warming by examining
  * the ActivePrewarm environment variable.
@@ -62,6 +58,10 @@ bool sample(double sample_rate) {
     double random_value = dis(gen);
     return random_value < sample_rate;
 }
+
+#ifdef __cplusplus
+extern "C" {
+#endif
 
 /**
  * Reads the DatadogProfiling info from the `UserDefaults`
@@ -137,9 +137,6 @@ public:
      * Start the profiler using constructor configuration
      */
     void start() {
-        // Reset profiling defaults to be re-evaluated again
-        delete_profiling_defaults();
-
         if (is_prewarming) {
             status = CTOR_PROFILER_STATUS_PREWARMED;
             return;
@@ -246,6 +243,9 @@ static void ctor_profiler_auto_start() {
         return;
     }
 
+    // Reset profiling defaults to be re-evaluated again
+    delete_profiling_defaults();
+
     // Create profiler and start with sample rate
     g_ctor_profiler = new dd::profiler::ctor_profiler(SAMPLING_RATE, is_active_prewarm());
     g_ctor_profiler->start();
@@ -270,10 +270,18 @@ void ctor_profiler_destroy(void) {
     g_ctor_profiler = nullptr;
 }
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 void ctor_profiler_start_testing(double sample_rate, bool is_prewarming, int64_t timeout_ns) {
     delete g_ctor_profiler;
     g_ctor_profiler = new dd::profiler::ctor_profiler(sample_rate, is_prewarming, timeout_ns);
     g_ctor_profiler->start();
 }
+
+#ifdef __cplusplus
+}
+#endif
 
 #endif // __APPLE__
--- a/DatadogProfiling/Mach/include/ctor_profiler_testing.h
+++ b/DatadogProfiling/Mach/include/ctor_profiler_testing.h
@@ -11,9 +11,56 @@
 extern "C" {
 #endif
 
-// Internal functions exposed only for testing
+/**
+ * @brief Checks if profiling is enabled in UserDefaults
+ *
+ * Reads the profiling enabled state from UserDefaults suite to determine
+ * if the profiling feature was previously enabled via Profiling.enable().
+ *
+ * @return true if profiling is enabled, false otherwise
+ *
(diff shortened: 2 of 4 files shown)


DatadogProfiling/Mach/ctor_profiler.cpp
```diff
@@ -16,12 +16,22 @@
...
+
 static constexpr int64_t CTOR_PROFILER_TIMEOUT_NS = 5000000000ULL; // 5 seconds
 
+static constexpr int64_t SAMPLING_RATE = 10; // 10%
+
 namespace dd::profiler { class ctor_profiler; }
 
 static dd::profiler::ctor_profiler* g_ctor_profiler = nullptr;
 
+#ifdef __cplusplus
```
maxep: You don't need `#ifdef __cplusplus` as you are in a cpp file

DatadogProfiling/Mach/include/ctor_profiler_testing.h
```diff
@@ -0,0 +1,22 @@
...
+#ifndef DD_PROFILER_CTOR_PROFILER_TESTING_H_
+#define DD_PROFILER_CTOR_PROFILER_TESTING_H_
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Internal functions exposed only for testing
+bool is_profiling_enabled();
+void delete_profiling_defaults();
```
maxep: We should move `ctor_profiler_start_testing` here as well

DatadogInternal/Sources/DataStore/UserDefaultsDataStore.swift
```diff
@@ -0,0 +1,67 @@
...
+    }
+    //swiftlint:enable required_reason_api_name
+
+    public func setValue(_ value: Data, forKey key: String, version: DataStoreKeyVersion) {
+        // Encode version (2 bytes) + data
+        var encoded = Data()
+        encoded.append(withUnsafeBytes(of: version) { Data($0) })
+        encoded.append(value)
+
+        userDefaults.set(encoded, forKey: key)
```
maxep, line 31: suggestion/ I think we should set the key and ""\(key.version)"" instead of data concatenation.

DatadogProfiling/Mach/ctor_profiler.cpp
```diff
@@ -54,35 +64,54 @@ bool sample(double sample_rate) {
...
+    if (value) {
+         if (CFGetTypeID(value) == CFDataGetTypeID()) {
+             CFDataRef data = (CFDataRef)value;
+             CFIndex length = CFDataGetLength(data);
+             const CFIndex versionLength = 2;
+
+             // Check we have bytes for version (2 bytes) and for the data (at least 1 byte for bool)
+             if (length >= versionLength + 1) {
+                 const UInt8* bytes = CFDataGetBytePtr(data);
+                 const UInt8* storedData = bytes + versionLength;
```
maxep, line 88: I think we should validate the version as well, no?

DatadogProfiling/Mach/ctor_profiler.cpp
```diff
@@ -108,6 +137,9 @@ class ctor_profiler {
      * Start the profiler using constructor configuration
      */
     void start() {
+        // Reset profiling defaults to be re-evaluated again
+        delete_profiling_defaults();
```
maxep: why? I don't think we need to 🤔  and it's not `ctor_profiler`'s role to do so as the defaults are read outside of its scope

DatadogProfiling/Sources/ProfilerFeature.swift
```diff
@@ -10,19 +10,34 @@ import DatadogInternal
...
+        /// The key to check if Profiling is enabled .
+        static let isProfilingEnabledKey = "is_profiling_enabled"
+    }
+
     let requestBuilder: FeatureRequestBuilder
 
     let messageReceiver: FeatureMessageReceiver
 
+    let dataStore: DataStore
+
```
maxep: Do we need to keep it as property? it is only used during init.

maxep (commented):
This plan makes sense, but have we explored other options than `UserDefaults`? I am wondering if it worth to bring its privacy constraints just to store one boolean. We could simply look for the presence of an empty file on disk, such as `.ctor_enable`. That could be way simpler to manage in both Swift and cpp code. Would that make sense?

DatadogProfiling/Mach/ctor_profiler.cpp
simaoseica-dd: Not important for this feature. The flag is always being added and removed in every session.

DatadogInternal/Sources/DataStore/UserDefaultsDataStore.swift
simaoseica-dd: This is the same logic we have in the other datastore `FeatureDataStore`. The `version` is appended to the data.

DatadogProfiling/Mach/ctor_profiler.cpp
simaoseica-dd: The `Profiling.enable()` can be encapsulated in a feature toggle that changes the profiling dynamically. 

Moved to the static constructor called by the system.

DatadogProfiling/Mach/ctor_profiler.cpp
simaoseica-dd: It is needed because the header "ctor_profiler_testing.h" might be included by both C and C++ code, and C compilers don't understand the "extern "C"".

PR #2564, opened 2025-11-04
RUM-12092: Establish communication between RUM and Profiling

simaoseica-dd:
### What and why?

This PR establishes communication between RUM and Profiling.
In this first iteration, Profiling will only capture the app launch. This PR ensures that profiling stops once the TTID (Time to Initial Display) is recorded.

Additionally, the TTID metric is enriched with the profiling status (whether profiling was running at that time).

### How?

**RUM → Profiling communication**
1. The `RUMAppLaunchManager` sends a `ProfilerStop` message to the message bus whenever the TTID is captured.
2. The `AppLaunchProfiler` listens for `ProfilerStop` messages and stops profiling when one is received.

**Profiling → RUM observation**
1. The `AppLaunchProfiler` and Profiling feature update the `ProfilingContext` in the core context with the current profiling status.
2. The `RUMAppLaunchManager` reads this profiling status to enrich the TTID vital event. At the moment TTID is recorded, the status can be either `running`, `stopped` or `error`.

### Review checklist
- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes
- [ ] Add Objective-C interface for public APIs (see our [guidelines](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3157787243/RFC+-+Modular+Objective-C+Interface#Recommended-solution) (internal) and run `make api-surface`)

PR #2571, opened 2025-11-07
RUM-12323: Skip profiling when TSan is enabled

simaoseica-dd:
### What and why?

This PR follows up on [this comment](https://github.com/DataDog/dd-sdk-ios/pull/2564#discussion_r2499712915).
Profiling is now skipped when the Thread Sanitizer is enabled.
This prevents Xcode from halting or reporting false data races during profiling sessions.

### How?

Before starting, the Profiling feature checks whether the Thread Sanitizer is enabled and skips profiling if it is.

### Review checklist

### DIFF
--- a/DatadogProfiling/Mach/ctor_profiler.cpp
+++ b/DatadogProfiling/Mach/ctor_profiler.cpp
@@ -29,6 +29,24 @@ namespace dd::profiler { class ctor_profiler; }
 static dd::profiler::ctor_profiler* g_ctor_profiler = nullptr;
 
 /**
+ * Checks if ThreadSanitizer is enabled
+ * and without options to avoid halts.
+ *
+ * @return true if ThreadSanitizer is enabled, false otherwise
+ */
+bool is_thread_sanitizer_enabled() {
+#if __has_feature(thread_sanitizer)
+    const char* tsanOptions = getenv("TSAN_OPTIONS");
+    if (tsanOptions != nullptr) {
+        return (strstr(tsanOptions, "halt_on_error=0") == nullptr)
+        || (strstr(tsanOptions, "report_bugs=0") == nullptr);
+    }
+    return true;
+#endif
+    return false;
+}
+
+/**
  * Checks if the current process was launched via pre-warming by examining
  * the ActivePrewarm environment variable.
  *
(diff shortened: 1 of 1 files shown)


DatadogProfiling/Mach/ctor_profiler.cpp
```diff
@@ -206,6 +212,24 @@ class ctor_profiler {
     bool is_prewarming = false;
     int64_t timeout_ns = CTOR_PROFILER_TIMEOUT_NS; // 5 seconds default
 
+    /**
+     * Checks if ThreadSanitizer is enabled
+     * and without options to avoid halts.
+     *
+     * @return true if ThreadSanitizer is enabled, false otherwise
+     */
+    bool is_thread_sanitizer_enabled() {
```
maxep: suggestion/ I think this should be a function instead of a class method, same as `is_active_prewarm`. This will allow you to override the `ctor_profiler_start_testing` and assert the right status depending on the `is_thread_sanitizer_enabled` value.

simaoseica-dd: Moved to function instead 👍 
The status is updated in the class method `void start()`since it is the common place to start Profiling for apps/integration tests and unit tests.

DatadogProfiling/Mach/ctor_profiler.cpp
```diff
@@ -15,14 +15,36 @@
...
 static dd::profiler::ctor_profiler* g_ctor_profiler = nullptr;
-static std::mutex g_ctor_profiler_mutex;
+
+/**
+ * Checks if ThreadSanitizer is enabled
+ * and without options to avoid halts.
+ *
+ * @return true if ThreadSanitizer is enabled, false otherwise
+ */
+bool is_thread_sanitizer_enabled() {
```
mariedm, line 37: Suggestin: the name is slightly misleading as we check if ThreadSanitizer is enabled but with certain options; about something like `should_skip_profiling_for_tsan` or `is_tsan_in_halt_mode`?

simaoseica-dd: Those options are related to it when working together. If they are both set (`"halt_on_error=0" != null && "report_bugs=0" != null`), the thread sanitizer is disabled, otherwise it is enabled.

PR #2574, opened 2025-11-11
RUM-11578 Add Datadog context sharing in obj-c

maciejburda:
### What and why?

Adds spi Internal `DDSharedContextPublisher` that allows subscribing to particular context changes from obj-c. For now only requested userId and accountId are added, but this is easy to extend.


maciejburda:
Closing - we will explore different solution

PR #2587, opened 2025-11-21
RUM-12904 Gate common type extensions behind DatadogExtension

### LINKED ISSUE #2583: Change access control for DatadogInternal API
### Feature description

I'm getting a symbol name collision with a function that's defined in DatadogInternal https://github.com/DataDog/dd-sdk-ios/blob/ca533286e991d4c05096ccbfb6a5309b254036ee/DatadogInternal/Sources/Utils/SwiftExtensions.swift#L118

Since a lot of the API in that module is `public`, it's being imported through my client code as well. Swift 5.9  added a new access control level called `package` [SE-386](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0386-package-access-modifier.md), so the other SPM libraries can use that code without it leaking into the client. The only way I'm able to solve my issue is renaming my subscript extension, which is a bit unfair.

### Proposed solution

_No response_

_No response_

mariedm:
### What and why?

Refactors public extensions on common types in `DatadogInternal` to use the `.dd` namespace pattern, preventing namespace collisions with customer code. This fixes #2583.

### Review checklist

### DIFF
--- a/DatadogInternal/Sources/Models/RUM/RUMDataModels.swift
+++ b/DatadogInternal/Sources/Models/RUM/RUMDataModels.swift
@@ -550,7 +550,7 @@ public struct RUMActionEvent: RUMDataModel {
 
         /// Subset of the SDK configuration options in use during its execution
         public struct Configuration: Codable {
-            /// The percentage of views profiled
+            /// The percentage of sessions profiled
             public let profilingSampleRate: Double?
 
             /// The percentage of sessions with RUM & Session Replay pricing tracked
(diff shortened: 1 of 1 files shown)


Datadog/IntegrationUnitTests/RUM/SDKMetrics/RUMViewEndedMetricIntegrationTests.swift
```diff
@@ -187,5 +187,5 @@ private extension TelemetryDebugEvent {
 }
 
 private extension Int64 {
-    var nanosecondsToSeconds: TimeInterval { TimeInterval(fromNanoseconds: self) }
+    var nanosecondsToSeconds: TimeInterval { TimeInterval.ddFromNanoseconds( self) }
```
gonzalezreal: 👮🏽‍♂️ Extra space
```suggestion
    var nanosecondsToSeconds: TimeInterval { TimeInterval.ddFromNanoseconds(self) }
```

DatadogInternal/Sources/Models/RUM/RUMDataModels.swift
gonzalezreal, line 1: Are these changes related with the PR scope? Shouldn't we address it in a different PR?

DatadogInternal/Sources/Utils/SwiftExtensions.swift
```diff
@@ -7,76 +7,91 @@
...
 // MARK: - Optional
+public struct DDOptionalExtension<Wrapped> {
+    private let optional: Wrapped?
+
+    init(_ optional: Wrapped?) {
+        self.optional = optional
+    }
 
-extension Optional {
     public func ifNotNil(_ closure: (Wrapped) throws -> Void) rethrows {
```
gonzalezreal, line 17: I don't understand why we can't use `DatadogExtension` / `DatadogExtended` in this case. Can you ellaborate?

mariedm: This won't compile: 
```
// Reference to generic type 'Optional' requires arguments in <...>
extension DatadogExtension where ExtendedType == Optional {}
// Cannot find type 'Wrapped' in scope
extension DatadogExtension where ExtendedType == Optional<Wrapped> {}
```

`Optional<Wrapped>` is a generic enum, and we can't reference the generic type parameter `Wrapped` in the extension constraint, hence the custom wrapper. But I'm open to suggestions if you have better idea to work around this as I couldn't think of something else.

DatadogInternal/Sources/Models/RUM/RUMDataModels.swift
mariedm: Not sure why, but CI is failing because of it. I made a separate PR for it: https://github.com/DataDog/dd-sdk-ios/pull/2589

gonzalezreal (dismissed):

PR #2595, opened 2025-11-26
RUM-11578 objc user and account id exposure

maciejburda:
### What and why?

Introduces "new" concept of Internal extension through `CrossPlatformExtension`.

This object is `Objective-C` compatible, single core only, `@_spi(Internal)` and right now allows subscribing to obj-c compatible `SharedContext` which currently exposes `user.id` and `account.id`. This can be easily extended in the future if needed.

Worth noting that internally extension is registering special `ContextSharingFeature` that has single purpose - observing core context on the message bus, but we can add additional `_extension` features if needed.

This change is required by KMP.

### DIFF
--- a/DatadogCore/Sources/Extensions/CrossPlatformExtension.swift
+++ b/DatadogCore/Sources/Extensions/CrossPlatformExtension.swift
@@ -7,10 +7,8 @@
 import Foundation
 import DatadogInternal
 
-///  for additional core capabilties only exposed Internally.
-///
-/// Extension provides cross-platform libraries with additional Internal-only capabilities.
-/// It's also expoed in Objective-C, mainly for KMP support.
+/// Extension that provides cross-platform libraries with additional Internal-only capabilities.
+/// It's Objective-C compatible, mainly for KMP support.
 ///
 /// Note: It only works for single core setup, relying on `CoreRegistry.default` existence.
 @objc(DDCrossPlatformExtension)
--- a/DatadogCore/Tests/Objc/ObjcAPITests/DDCrossPlatformExtension+apiTests.m
+++ b/DatadogCore/Tests/Objc/ObjcAPITests/DDCrossPlatformExtension+apiTests.m
@@ -23,6 +23,8 @@
     [DDCrossPlatformExtension subscribeToSharedContext:^(DDSharedContext * _Nullable context) {
         // Just check API availability in Objective-C
     }];
+
+    [DDCrossPlatformExtension unsubscribeFromSharedContext];
 }
 
 #pragma clang diagnostic pop


DatadogCore/Sources/Datadog+objc.swift
```diff
@@ -103,6 +103,52 @@ public final class objc_Datadog: NSObject {
         Datadog.clearAllData()
     }
 
+    // MARK: - User and Account Info Access
```
maxep: suggestion/ If this is intended for KMP only, maybe we can expose this from a dedicated KMP object only in obj-c. Something like:
```swift
@objc(KMPDatadogExtension)
@objcMembers
@_spi(objc)
public final class kmp_DatadogExtension: NSObject {
    public static func subscribeToUserId(_ receiver: @escaping (String?) -> Void) {
        let core = CoreRegistry.default as? DatadogCore
        return core?.currentUserId()
    }
}
```

DatadogCore/Sources/Datadog.swift
```diff
@@ -341,6 +341,66 @@ public enum Datadog {
...
+    // MARK: - SPI: User and Account Info Access
+
+    /// Returns the current user ID, if any.
+    ///
+    /// This is an SPI method intended for internal use.
+    ///
+    /// - Parameter core: The core instance to query. Defaults to the default core instance.
+    /// - Returns: The current user ID, or `nil` if not set.
+    @_spi(Internal)
+    public static func currentUserId(in core: DatadogCoreProtocol = CoreRegistry.default) -> String? {
```
maxep: question/ Is there a need to define these methods in `Datadog.swift`? I think it could be solely exposed by `Datadog+objc.swift`

maxep (commented):
Looks great. I left a couple of comment on how to expose these. LMKWYT!

DatadogCore/Sources/Extensions/CrossPlatformExtension.swift
```diff
@@ -0,0 +1,48 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+import DatadogInternal
+
+///  for additional core capabilties only exposed Internally.
```
0xnm: it seems something is missing here at the beginning of the sentence

DatadogCore/Tests/Objc/ObjcAPITests/DDCrossPlatformExtension+apiTests.m
```diff
@@ -0,0 +1,30 @@
...
+@implementation DDCrossPlatformExtension_apiTests
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunused-value"
+
+- (void)testDDContextSharingExtensionAPI {
+    [DDCrossPlatformExtension subscribeToSharedContext:^(DDSharedContext * _Nullable context) {
+        // Just check API availability in Objective-C
+    }];
+}
```
ncreated, line 28: We're missing `[DDCrossPlatformExtension unsubscribeFromSharedContext]` here.

DatadogCore/Sources/Extensions/ContextSharing/ContextSharingFeature.swift
```diff
@@ -0,0 +1,17 @@
...
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import DatadogInternal
+
+internal final class ContextSharingFeature: DatadogFeature {
+    static var name: String = "_extension_context_sharing"
+
+    var messageReceiver: FeatureMessageReceiver
```
simaoseica-dd, line 12:
```suggestion
    let messageReceiver: FeatureMessageReceiver
```

DatadogCore/Sources/Extensions/CrossPlatformExtension.swift
```diff
@@ -0,0 +1,46 @@
...
+import DatadogInternal
+
+/// Extension that provides cross-platform libraries with additional Internal-only capabilities.
+/// It's Objective-C compatible, mainly for KMP support.
+///
+/// Note: It only works for single core setup, relying on `CoreRegistry.default` existence.
+@objc(DDCrossPlatformExtension)
+@objcMembers
+@_spi(Internal)
+public final class CrossPlatformExtension: NSObject {
```
simaoseica-dd, line 17: nit: The naming guidelines prefix the public entities with `objc_`
```suggestion
public final class objc_CrossPlatformExtension: NSObject {
```

DatadogCore/Sources/Extensions/CrossPlatformExtension.swift
```diff
@@ -0,0 +1,46 @@
...
+    private static var contextSharingTransformer: ContextSharingTransformer?
+
+    /// Subscribes to shared context updates.
+    ///
+    /// The provided closure will be called immediately with the current context (or `nil` if no context yet),
+    /// and subsequently whenever the context changes. It lazy loads static instance of `ContextSharingTransformer`
+    /// and registers `ContextSharingFeature`.
+    ///
+    /// - Parameter toSharedContext: A closure that receives `SharedContext` updates. Called on context changes.
+    @objc
```
simaoseica-dd, line 27: With the `@objcMembers` above this annotation should not be needed

PR #2598, opened 2025-11-27
RUM-12068 Copy `service` to `ddtags` in Logs

arroz:
### What and why?

Log requests should include the `service` tag for Cloud Cost Management segmentation.

### How?

`LogEventEncoder` now adopts the common `ddTags` getter in `DatadogContext` as the common implementation to obtain the tag string for all Datadog tags. This includes `service` and `sdk_version` which were missing from the Log requests.

### Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - [FEATURE] Add support for configuring a custom version parameter in DatadogConfiguration. See [#2585][] (Thanks [@blimmer][])
 - [IMPROVEMENT] Provide XCFramework without arm64e slice for Xcode 26+ compatibility. See [#2576][]
+- [FIX] Add service and sdk_version tags to log requests. See [#2598][]
 
 # 3.3.0 / 17-11-2025
 


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 - [FEATURE] Add support for configuring a custom version parameter in DatadogConfiguration. See [#2585][] (Thanks [@blimmer][])
 - [IMPROVEMENT] Provide XCFramework without arm64e slice for Xcode 26+ compatibility. See [#2576][]
+- [FIX] Add service and sdk_version tags to log requests. See [#2598][]
```
maciejburda, line 5: We used `CHANGELOG` only for user facing changes.

DatadogLogs/Sources/Log/LogEventEncoder.swift
```diff
@@ -339,11 +341,11 @@ internal struct LogEventEncoder {
...
-        tags.append("env:\(log.environment)") // include default env tag
-        tags.append("version:\(log.applicationVersion)") // include default version tag
-        if let variant = log.variant {
-            tags.append("variant:\(variant)")
-        }
+        // Include dd tags
+        // log.ddTags is already a string with multiple tags
+        // joined by ",". That is OK, as it gets joined with
+        // the specific ones for this log.
+        tags.append(log.ddTags)
```
maciejburda, line 348: So we basically append dd_tags after log's tags?

Did you managed to get confirmation from the backend if they will look at the second occurence if for example `service` is set by the customer in tags?

CHANGELOG.md
arroz: This will add two new tags to the log view, so it was added as user facing.

DatadogLogs/Sources/Log/LogEventEncoder.swift
arroz: If the user tries to add a tag with a reserved name (like service or any of the other dd tags), the sanitization method removes it. So there's no possibility of having two conflicting service names.

PR #2604, opened 2025-12-03
FFL-1460 FlagsClient changes to accommodate Flags RN SDK sync flag evals 

yugisu-flux:
### What and why?

The motivation for this work is to allow the upcoming Flags RN SDK to be able to retrieve complete feature flags state to the JS side at init time ([FFL-1460](https://datadoghq.atlassian.net/browse/FFL-1460)).

This has been developed in parallel with changes in: 
- https://github.com/DataDog/dd-sdk-reactnative/pull/1064
- https://github.com/DataDog/dd-sdk-android/pull/3049

### How?

Changes:
- FlagsClient method for retrieving all feature flags from a FlagsClient.
- FlagsClient method for exposure tracking and RUM logging that will be exposed to the RN SDK.

Please note that the above changes are for internal Datadog usage and are not meant to be used by the wider public.

### Review checklist
- [ ] ~~Add Objective-C interface for public APIs (see our [guidelines](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3157787243/RFC+-+Modular+Objective-C+Interface#Recommended-solution) (internal) and run `make api-surface`)~~
- The changes in this PR are not a part of "public" interface.

[FFL-1460]: https://datadoghq.atlassian.net/browse/FFL-1460?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


DatadogFlags/Sources/Client/FlagsClient.swift
```diff
@@ -211,18 +211,71 @@ extension FlagsClient: FlagsClientProtocol {
...
+                reason: assignment.reason
             )
         }
 
-        return details
+        return result
+    }
+
+    @_spi(Internal)
+    public func trackEvaluation(key: String) {
```
typotter: Tracking needs to be done with information from the initial evaluation of the flag assignment. Changing the context and, in the future, polling for new config may change the assigned value and/or the context associated with it.
If the context does not change, then this will work correctly, however, it is susceptible to incorrectly pairing context with an assignment.

sameerank: Hmm in this case, I think `repository.context` gets the latest context and the context and assignments get updated at the same time when `setEvaluationContext()` is called

DatadogFlags/Tests/Client/FlagsClientTests.swift
```diff
@@ -251,6 +252,184 @@ final class FlagsClientTests: XCTestCase {
         XCTAssertEqual(rumFlagEvaluationReporter.sendFlagEvaluationCalls.count, 6)
     }
 
+    func testAllFlagsEvaluation() {
```
sameerank: Is this test name correct? `testGetFlagsDetails` would indicate that this is testing `client.getFlagsDetails`, and I see that the other test name (`testTrackEvaluation`) matches between `FlagsClientTests.swift` and `FallbackFlagsClientTests.swift`

DatadogFlags/Tests/Client/FlagsClientTests.swift
```diff
@@ -272,10 +451,9 @@ final class FlagsClientTests: XCTestCase {
         // When
         Flags.enable(with: .init(trackExposures: false), in: core)
         let client = FlagsClient.create(in: core)
-        let value = client.getStringValue(key: "test", defaultValue: "")
```
sameerank, line 275: I feel like this is a different kind of test, i.e. that evaluation is automatically tracked when you invoke `client.getStringValue` vs evaluation is explicitly tracked if you invoke `client.trackEvaluation`

Or is the former redundant because other tests also check this? Basically I'm wondering if we should do separate tests or if it's okay to replace the old test like this

DatadogFlags/Sources/Client/FlagsClientProtocol.swift
```diff
@@ -69,6 +69,14 @@ public protocol FlagsClientProtocol: AnyObject {
     ///
     /// - Returns: A ``FlagDetails`` struct containing the evaluated value and metadata.
     func getDetails<T>(key: String, defaultValue: T) -> FlagDetails<T> where T: Equatable, T: FlagValue
+
+    /// Note: This is an internal method. Expect breaking changes in the future.
+    @_spi(Internal)
```
sameerank: And you might need to move these to the protocol extension below to make the smoke tests happy

DatadogFlags/Sources/Client/FlagsClient.swift
```diff
@@ -211,18 +211,71 @@ extension FlagsClient: FlagsClientProtocol {
...
+            case .object(let v):    value = v
+            case .unknown:          continue
+            }
+
+            result[key] = FlagDetails(
+                key: key,
+                value: value,
+                variant: assignment.variationKey,
+                reason: assignment.reason
             )
```
gonzalezreal: 💡 You could encapsulate flag assignment to flag details mapping in a convenience extension that you can re-use in both `getFlagsDetails` and `trackEvaluation` methods:

```swift
extension FlagDetails where T == AnyValue {
    fileprivate init?(key: String, flagAssignment: FlagAssignment) {
        switch flagAssignment.variation {
        case .boolean(let value):
            self.init(key: key, value: .bool(value), variant: flagAssignment.variationKey, reason: flagAssignment.reason)
        case .string(let value):
            self.init(key: key, value: .string(value), variant: flagAssignment.variationKey, reason: flagAssignment.reason)
        case .integer(let value):
            self.init(key: key, value: .int(value), variant: flagAssignment.variationKey, reason: flagAssignment.reason)
        case .double(let value):
            self.init(key: key, value: .double(value), variant: flagAssignment.variationKey, reason: flagAssignment.reason)
        case .object(let value):
            self.init(key: key, value: value, variant: flagAssignment.variationKey, reason: flagAssignment.reason)
        case .unknown:
            return nil
        }
    }
}
```

With this in place, you could construct the flag details dictionary in a leaner way (although the for-loop is totally fine as well):

```swift
let flagDetails: [String: FlagDetails<AnyValue>] = Dictionary(
    uniqueKeysWithValues: flagAssignments.compactMap { key, value in
        guard let details = FlagDetails(key: key, flagAssignment: value) else {
            return nil
        }
        return (key, details)
    }
)
```

DatadogFlags/Sources/Client/FlagsClient.swift
gonzalezreal: Yes, @sameerank is right. The flag assignments are tied to the evaluation context in the repository. The ReactNative wrapper needs to call the new `getDetails` with each new evaluation context.

DatadogFlags/Sources/Client/FlagsClientProtocol.swift
gonzalezreal: 💡 Have you considered creating a dedicated protocol for these public-internal methods? ie. `FlagsClientInternal` or similar.

Then you can make `FlagsClient` conform to both `FlagsClientProtocol` and `FlagsClientInternal`. My guess is that you can query for both protocols from the RN side, right? I think that would be cleaner.

gonzalezreal (commented):
Nice work!

I left a couple of suggestions. Let me know your thoughts.

DatadogFlags/Tests/Client/FlagsClientTests.swift
yugisu-flux: I added a test that explicitly tests this, thank you

DatadogFlags/Sources/Client/FlagsClient.swift
yugisu-flux: That's essentially correct - RN SDK updates its cache (calls `getAllFlagsDetails`) right after `setEvaluationContext` call. So that 

evaluation context update refreshes the state inside iOS repository -> RN DdSdkImplementation awaits on the resolution of the update and calls `getFlagsDetails` right after that to retrieve the updated state -> passes the flags details to JS

Given this flow, JS state should be always up-to-date with current iOS state. If we ever to implement flags polling, then we'll need to pass the updates to JS through events. Are there plans to implement other flags state updates besides "upon evaluation context updated"? @typotter @gonzalezreal

DatadogFlags/Sources/Client/FlagsClient.swift
yugisu-flux: Wow these are some Swift tips and hacks I expected to see in these PRs! Thank you!!

DatadogFlags/Sources/Client/FlagsClientProtocol.swift
yugisu-flux: I have carved out these internal methods to a separate protocol, but I ran into an issue where I need to write `FlagsClientProtocol & FlagsClientInternalProtocol` in the places I don't believe this belongs.

Example of such a place would be the `FlagsClient@create` and `FlagsClient@shared` methods, which are used in the RN SDK along with those "internal" APIs.

For instance, this feels natural in RN SDK code:
Image

And doesn't feel natural here at all:
Image

Please let me know if you know a way to do this 🙏

yugisu-flux: I have added a default no-op implementation for the "internal" methods, hopefully this resolves the smoke test errors

DatadogFlags/Sources/Client/FlagsClient.swift
```diff
@@ -211,18 +211,69 @@ extension FlagsClient: FlagsClientProtocol {
...
+            assignment: flagAssignment,
+            evaluationContext: context
+        )
+
+        let details = FlagDetails(key: key, flagAssignment: flagAssignment)
+
+        rumFlagEvaluationReporter.sendFlagEvaluation(
+            flagKey: key,
+            value: details?.value ?? AnyValue.null
+        )
```
sameerank: Seeing the following [in the CI logs](https://gitlab.ddbuild.io/DataDog/dd-sdk-ios/-/jobs/1276608646)

```
2025-12-08T17:44:39.648510Z 01O     [31m✖[0m testTrackEvaluation, XCTAssertEqual failed: ("nil") is not equal to ("Optional(true)")
2025-12-08T17:44:39.648510Z 01O     [31m✖[0m testTrackEvaluation, XCTAssertEqual failed: ("nil") is not equal to ("Optional("red")")
2025-12-08T17:44:39.648520Z 01O     [31m✖[0m testTrackEvaluation, XCTAssertEqual failed: ("nil") is not equal to ("Optional(42)")
2025-12-08T17:44:39.648520Z 01O     [31m✖[0m testTrackEvaluation, XCTAssertEqual failed: ("nil") is not equal to ("Optional(3.14)")
```

`FlagDetails` sets `details.value` to the `AnyValue` enum. So this casting `rumFlagEvaluationReporter.sendFlagEvaluationCalls[0].1 as? Bool` is not working because `AnyValue.bool(true)` is an enum and so `as? Bool` fails and returns a nil.

Based on what I see in this test

https://github.com/DataDog/dd-sdk-ios/blob/83298e9a177b365376853e5353912bc3c94e0deb/DatadogFlags/Tests/Client/RUMFlagEvaluationReporterTests.swift#L21-L24

It looks like `value` is expected by `sendFlagEvaluation` to be an actual type and not an `AnyValue` enum case, so you might consider doing something like the following

```suggestion
        func send<T: FlagValue>(_ value: T) {
            rumFlagEvaluationReporter.sendFlagEvaluation(flagKey: key, value: value)
        }
        
        switch flagAssignment.variation {
        case .boolean(let value): send(value)
        case .string(let value): send(value)
        case .integer(let value): send(value)
        case .double(let value): send(value)
        case .object(let value): send(value)
        case .unknown: send(AnyValue.null)
        }
```

The switch-case feels a bit messy so feel free to go with a different approach if you think of one. I like how `getDetails` did this

https://github.com/DataDog/dd-sdk-ios/blob/62f86ef10a26673def087d1e8c5436833b15ec3a/DatadogFlags/Sources/Client/FlagsClient.swift#L203-L205

Is `trackEvaluation` invoked in places where we don't know the raw value type `T`, i.e. we can't easily add a `value` or `defaultValue` as an argument?

DatadogFlags/Sources/Client/FlagsClientProtocol.swift
```diff
@@ -292,3 +300,18 @@ extension FlagsClientProtocol {
         getDetails(key: key, defaultValue: defaultValue)
     }
 }
+
+// MARK: - Internal methods consumed by the React Native SDK
```
sameerank: I am okay with this approach since it sounds like implementing a `FlagsClientInternalProtocol` was looking messy. Ideally you'd only reference an internal protocol when it is relevant for internal use

DatadogFlags/Sources/Client/FlagsClientProtocol.swift
gonzalezreal: I would not pollute any public API returning `FlagsClientProtocol` with the internal protocol. Instead I would cast it on demand:

```swift
let client = FlagsClient.shared()
guard let internalClient = client as? FlagsClientInternal else {
  // throw an error
}
// do something with internalClient
let flagDetails = internalClient.getAllFlagsDetails()
```

FYI, the `Protocol` suffix in `FlagsClientProtocol` is used to disambiguate it from `FlagsClient`. You don't need it for the internal protocol.

DatadogFlags/Sources/Client/FlagsClient.swift
typotter: yes. we will want to be able to poll, or update configuration on demand in the future.

Agreed that this will work as is and without the underlying refresh mechanism, this might be overprotective. 
If the RN sdk is (roughly):

```js
getStringAssignment(flagKey, default) {
  _cantAwaitThisPromise = nativeClient.getStringAssignment(flagKey) // during this async call, the effective assignments and what is logged could change after the next line returns.
  return _flagAssignmentSnapshot[flagKey] ?? default
```

If the underlying flags config results in a different assignment, or if the evaluation context has been updated but the RN flag assignment map has not yet been updated, there could be mismatch in either the assignment or the evaluation context.

DatadogFlags/Sources/Client/FlagsClient.swift
```diff
@@ -211,18 +211,81 @@ extension FlagsClient: FlagsClientProtocol {
...
+
+        var value: FlagValue
+        switch flagAssignment.variation {
+        case .boolean(let v): value = v
+        case .string(let v): value = v
+        case .integer(let v): value = v
+        case .double(let v): value = v
+        case .object(let v): value = v
+        case .unknown: value = AnyValue.null
+        }
```
gonzalezreal, line 253: You can reuse the private FlagDetails extension instead of the manual mapping:

```swift
public func trackEvaluation(key: String) {
    guard
        let flagAssignment = repository.flagAssignment(for: key),
        let context = repository.context,
        let details = FlagDetails<AnyValue>(key: key, flagAssignment: flagAssignment)
    else {
        return
    }

    exposureLogger.logExposure(
        for: key,
        assignment: flagAssignment,
        evaluationContext: context
    )
    
    rumFlagEvaluationReporter.sendFlagEvaluation(
        flagKey: key,
        value: details.value
    )
}
```

DatadogFlags/Sources/Client/FlagsClient.swift
```diff
@@ -211,18 +211,81 @@ extension FlagsClient: FlagsClientProtocol {
...
+        trackEvaluation(key: key)
 
         return details
     }
 }
+
+// MARK: - Internal methods consumed by the React Native SDK
+
+extension FlagsClient: FlagsClientInternal {
+    @_spi(Internal)
```
gonzalezreal, line 238: AFAIK you don't need `@_spi(Internal)` on the implementations if it's already on the protocol requirement

DatadogFlags/Sources/Client/FlagsClient.swift
```diff
@@ -211,18 +211,81 @@ extension FlagsClient: FlagsClientProtocol {
...
+                    return nil
+                }
+                return (key, details)
+            }
+        )
+
+        return result
+    }
+
+    @_spi(Internal)
```
gonzalezreal, line 243: Same here. It is not needed.

DatadogFlags/Sources/Client/FlagsClientInternal.swift
```diff
@@ -0,0 +1,31 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+import DatadogInternal
+
+public protocol FlagsClientInternal: AnyObject {
+    /// Note: This is an internal method. Expect breaking changes in the future.
```
gonzalezreal: IMO this is not needed, as it is implied that `@_spi(Internal)` methods can break in the future. If you still want to leave a DocC comment, you can use a more appropriate formatting:

```swift
/// > Warning: This is an internal method and can break in the future.
```

See https://www.swift.org/documentation/docc/other-formatting-options

DatadogFlags/Sources/Client/FlagsClientProtocol.swift
```diff
@@ -292,3 +292,18 @@ extension FlagsClientProtocol {
...
+    public func getAllFlagsDetails() -> [String: FlagDetails<AnyValue>]? {
+        // no-op
+        return nil
+    }
+
+    @_spi(Internal)
+    public func trackEvaluation(key: String) {
+        // no-op
+    }
+}
```
gonzalezreal: This is redundant with `FlagsClientInternal` protocol extension and can be removed.

DatadogFlags/Sources/Client/FlagsClient.swift
yugisu-flux: `FlagDetails`'s value is `AnyValue` here. As Sameeran pointed out in another comment, it feels like `rumFlagEvaluationReporter.sendFlagEvaluation` is expecting "plain" Bool, String, Double, Int values to be passed to it.

I'm just not confident about passing AnyValue enum value to this method, thus the duplication.

gonzalezreal: You're right. Ignore my suggestion.

DatadogFlags/Sources/Models/FlagAssignment.swift
```diff
@@ -6,21 +6,21 @@
 
 import Foundation
 
-internal struct FlagAssignment: Equatable {
-    enum Variation: Equatable {
+@_spi(Internal)
```
yugisu-flux, line 9: This `@_spi` annotation made me to add the SPI annotation in all internal test packages using FlagAssignment struct. Is there a better way to resolve this?

DatadogFlags/Sources/Client/FlagsClientInternal.swift
```diff
@@ -0,0 +1,31 @@
...
+
+public protocol FlagsClientInternal: AnyObject {
+    /// > Warning: This is an internal method and can break in the future.
+    @_spi(Internal)
+    func getFlagAssignmentsSnapshot() -> [String: FlagAssignment]?
+
+    /// > Warning: This is an internal method and can break in the future.
+    @_spi(Internal)
+    func trackFlagSnapshotEvaluation(key: String, assignment: FlagAssignment, context: FlagsEvaluationContext)
+}
```
gonzalezreal, line 18: 🤏🏽 I'm wondering about the choice to use `snapshot` in the method names. It can suggest there are other types of flag retrievals or evaluations that aren't snapshots, but that distinction does not exist.

I suggest using names that are consistent with existing methods and avoid introducing new concepts like `snapshot`:

```swift
func getFlagAssignments() -> [String: FlagAssignment]?
func trackFlagEvaluation(key: String, assignment: FlagAssignment, context: FlagsEvaluationContext)
```

What do you think?

> [!NOTE]
> If the reason is to disambiguate from the internal `trackEvaluation`, Swift handles method overloading (same name, different signature). Or, you can rename the internal method to `sendFlagEvaluation`.

DatadogFlags/Sources/Client/FlagsRepository.swift
```diff
@@ -96,6 +98,11 @@ extension FlagsRepository: FlagsRepositoryProtocol {
         return state?.flags[key]
     }
 
+    func flagAssignmentsSnapshot() -> [String: FlagAssignment]? {
```
gonzalezreal: 🤏🏽 Same feedback here. Consider removing the `Snapshot` suffix.

DatadogFlags/Sources/Models/FlagAssignment.swift
gonzalezreal: I am afraid there is no other way. When you mark a type with `@_spi(Internal)`, any module that uses it must import it with `@_spi(Internal)`

DatadogFlags/Sources/Client/FlagsClientInternal.swift
yugisu-flux: The `Snapshot` naming came from [this comment](https://github.com/DataDog/dd-sdk-ios/pull/2604#discussion_r2604197614). Basically, a snapshot here = current state of the flag assignments

DatadogFlags/Tests/Client/FallbackFlagsClientTests.swift
```diff
@@ -61,4 +61,49 @@ final class FallbackFlagsClientTests: XCTestCase {
             """
         )
     }
+
+    func testGetFlagAssignmentsSnapshot() {
```
gonzalezreal: nit: You can rename this to `testGetFlagAssignments`

DatadogFlags/Tests/Client/FallbackFlagsClientTests.swift
```diff
@@ -61,4 +61,49 @@ final class FallbackFlagsClientTests: XCTestCase {
...
+        XCTAssertEqual(
+            dd.logger.errorMessages.first,
+            """
+            Using fallback client to get all flag values. \
+            Ensure that a client named 'default' is created before using it.
+            """
+        )
+    }
+
+    func testTrackEvaluation() {
```
gonzalezreal: nit: This could be named `testSendFlagEvaluation()`

DatadogFlags/Tests/Client/FlagsClientTests.swift
```diff
@@ -301,12 +336,225 @@ final class FlagsClientTests: XCTestCase {
...
         // Then
-        XCTAssertEqual(value, .mockAny())
-        XCTAssertEqual(messageReceiver.messages.filter(\.isRUMMessage).count, 0, "No RUM messages should be sent")
         XCTAssertEqual(core.events(ofType: ExposureEvent.self).count, 1, "Exposure should still be logged")
+        XCTAssertEqual(messageReceiver.messages.filter(\.isRUMMessage).count, 0, "No RUM messages should be sent")
+    }
+
+    // MARK: - Internal methods consumed by the React Native SDK
+
+    func testGetFlagAssignmentsSnapshot() {
```
gonzalezreal: nit: This should be renamed to `testGetFlagAssignments()`

DatadogFlags/Tests/Client/FlagsClientTests.swift
```diff
@@ -301,12 +336,225 @@ final class FlagsClientTests: XCTestCase {
...
+                variationKey: "variation-124",
+                variation: .boolean(true),
+                reason: "TARGETING_MATCH",
+                doLog: true
+            )
+        )
+        XCTAssertNil(flagAssignments["missing-flag"])
+    }
+
+    func testTrackFlagSnapshotEvaluation() {
```
gonzalezreal: nit: This should be renamed to `testSendFlagEvaluation()`

PR #2605, opened 2025-12-04
RUM-12980: Improve Profiling to handle multiple SDK instances

simaoseica-dd:
### What and why?

This PR enables the `DatadogProfiling` feature to be used across multiple SDK instances concurrently.
The `DatadogProfiling` can now be enabled from different cores, and all concurrent calls to the `ctor_profiler` (get profile, get status, stop, destroy) are handled safely.

**Note**: The current Profiling is focused on App Launch profiling only.
 
 ### How?
 
- Added mutex protection to `ctor_profiler` and `mach_sampling_profiler` exposed functions to handle concurrent access.
- Added `pendingInstances` reference counting in `AppLaunchProfiler` to ensure `ctor_profiler_destroy()` is only called after all profilers have processed the profile.
- Replaced `UserDefaultsDataStore` with a simple boolean in `UserDefaults` to track if profiling was enabled.
- Centralized `UserDefaults` constants in `ctor_profiler.h` for use in both C++ and Swift.
- Main thread stack traces without `thread_name` are now labeled with "com.apple.main-thread" for better visibility in the UI.

### Review checklist


DatadogProfiling/Mach/mach_sampling_profiler.cpp
```diff
@@ -284,19 +296,41 @@ bool stack_trace_get_thread_info(stack_trace_t* trace, thread_t thread) {
...
     // Allocate buffer and get thread name
     trace->thread_name = (char*)malloc(PTHREAD_THREAD_NAME_MAX);
     if (!trace->thread_name) return false;
     
     int result = pthread_getname_np(pthread, (char*)trace->thread_name, PTHREAD_THREAD_NAME_MAX);
+
+    // If it's the main thread and has no name
+    if (pthread == g_main_pthread && trace->thread_name[0] == '\0') {
+        strcpy((char*)trace->thread_name, "com.apple.main-thread");
+    }
```
maxep, line 373: suggestion/ we would need consistencies across devices and always set the same name for the main thread. WDYT?

question/ We discussed adding a `is_main_thread` label to the sample, was this discarded?

idea/ We should evaluate the cost of getting the queue-name when available, there is a solution in [kscrash](https://github.com/kstenerud/KSCrash/blob/master/Sources/KSCrashRecordingCore/KSThread.c#L95) but I'm not sure this is reliable nor performant. Not to do in this PR but to keep in mind.

DatadogProfiling/Mach/mach_sampling_profiler.cpp
```diff
@@ -284,19 +296,41 @@ bool stack_trace_get_thread_info(stack_trace_t* trace, thread_t thread) {
...
     return false;
 }
 
+/**
+ * Safely reads memory from a potentially invalid address.
+ * Uses vm_read_overwrite to avoid crashes on bad memory access.
+ */
+bool safe_read_memory(void* addr, void* buffer, size_t size) {
+    vm_size_t read_size = size;
+    kern_return_t kr = vm_read_overwrite(
```
maxep: cr/  Using `vm_read_overwrite` for every frame during stack unwinding adds significant syscall overhead. As noted in [this post](https://gms.tf/on-the-costs-of-syscalls.html) :

> "So looking at the real syscalls, the user-kernel mode switches cost in the order of a few hundred nanoseconds"

For a 20-frame stack at 100Hz sampling, this means **~2,000 syscalls/sec** = 0.2-0.5% overhead minimum (~500-1000ns per call).

**Suggestion:** (from claude) Use signal-based fault protection for direct memory access:

```cpp
// Signal handler approach - used by production profilers
static sigjmp_buf fault_handler;
static volatile bool in_safe_read = false;

void segv_handler(int sig) {
    if (in_safe_read) siglongjmp(fault_handler, 1);
    // Otherwise, crash normally
}

bool safe_read_memory(void* addr, void* buffer, size_t size) {
    in_safe_read = true;
    if (sigsetjmp(fault_handler, 1) == 0) {
        memcpy(buffer, addr, size);  // Direct read: ~5ns vs syscall: ~500ns
        in_safe_read = false;
        return true;
    }
    in_safe_read = false;
    return false;  // Fault occurred
}
```

This avoids syscall overhead entirely - the signal handler only fires on actual faults, not on every frame. This is the standard approach in profilers like perf and async-profiler.

DatadogProfiling/Mach/mach_sampling_profiler.cpp
```diff
@@ -370,16 +406,22 @@ void* mach_sampling_profiler::sampling_thread_entry(void* arg) {
...
+    std::lock_guard<std::mutex> lock(state_mutex);
+    
     if (running) return false;
 
     if (config.profile_current_thread_only) {
         target_thread = pthread_self();
     }
 
+    // Clear any leftover data from previous runs
+    sample_buffer.clear();
```
maxep, line 492: Good catch!

DatadogProfiling/Sources/AppLaunchProfiler.swift
```diff
@@ -16,21 +16,30 @@ internal import DatadogMachProfiler
 // swiftlint:enable duplicate_imports
 
 internal final class AppLaunchProfiler: FeatureMessageReceiver {
+    /// Shared counter to track pending `AppLaunchProfiler`s from handling the `ProfilerStop` message
+    private static var pendingInstances: Int = 0
+    private static let lock = NSLock()
```
maxep, line 21: suggestion/ Why not using a `@ReadWriteLock` ?

DatadogProfiling/Sources/AppLaunchProfiler.swift
```diff
@@ -16,21 +16,30 @@ internal import DatadogMachProfiler
...
+                || ctor_profiler_get_status() == CTOR_PROFILER_STATUS_STOPPED else {
             return false
         }
 
         ctor_profiler_stop()
         core.set(context: ProfilingContext(status: .current))
-        defer { ctor_profiler_destroy() }
 
         guard let profile = ctor_profiler_get_profile() else {
+            Self.unregisterInstance()
```
maxep: suggestion/ It would be safer to do it in a `defer` block.

DatadogProfiling/Sources/AppLaunchProfiler.swift
```diff
@@ -71,10 +81,54 @@ internal final class AppLaunchProfiler: FeatureMessageReceiver {
...
+    }
+
+    /// Resets the pending instances counter.
+    static func resetPendingInstances() {
+        lock.lock()
+        defer { lock.unlock() }
+        pendingInstances = 0
+    }
+}
+#endif
```
maxep: cr/ The `DD_SDK_COMPILED_FOR_TESTING` is used for integration tests. But this extension is accessed from unit tests only and it's not necessary in unit-tests since you can do `@testable import`.

DatadogProfiling/Sources/AppLaunchProfiler.swift
simaoseica-dd: It needs a static lock shared across all `AppLaunchProfiler` instances. In a multi sdk environment we might have multiple Profiling features, each one with its own `AppLaunchProfiler`.

DatadogProfiling/Mach/mach_sampling_profiler.cpp
simaoseica-dd: It was discussed, but because it required additional backend and frontend work, we decided to proceed with a hardcoded thread name for now.

DatadogProfiling/Mach/mach_sampling_profiler.cpp
simaoseica-dd: Great point here! Thank you.  
Agree that a high-frequency sampling profiler needs to re-consider extra overheads. 
Applied and validated the suggestion having in mind that it will be benchmarked later on.

DatadogProfiling/Sources/AppLaunchProfiler.swift
simaoseica-dd: Yes, it is better.

DatadogProfiling/Sources/AppLaunchProfiler.swift
maxep: But you can apply the `@ReadWriteLock` to a static property as well 👍

DatadogProfiling/Mach/mach_sampling_profiler.cpp
```diff
@@ -66,13 +68,59 @@ static constexpr size_t PTHREAD_THREAD_NAME_MAX = 64;
...
+}
+
+/**
+ * Initialize the main thread pthread identifier
+ * and install signal handlers for safe memory reading.
+ *
  * This constructor runs early to capture the main thread's pthread_t.
  */
 __attribute__((constructor))
 static void init_main_thread_id() {
```
maxep: suggestion/ Maybe we should rename this now.

DatadogProfiling/Mach/mach_sampling_profiler.cpp
```diff
@@ -66,13 +68,59 @@ static constexpr size_t PTHREAD_THREAD_NAME_MAX = 64;
...
+        if (prev->sa_sigaction) {
+            prev->sa_sigaction(sig, info, context);
+        }
+    } else if (prev->sa_handler == SIG_DFL) {
+        // Re-raise with default handler
+        signal(sig, SIG_DFL);
+        raise(sig);
+    } else if (prev->sa_handler != SIG_IGN) {
+        prev->sa_handler(sig);
+    }
```
maxep, line 108: question/ I'm not sure to understand this 🤔  Plus, the `signal` function is not [async-thread-safe](https://man7.org/linux/man-pages/man7/signal-safety.7.html) (it might be on iOS tho), so it could be considered as not safe to be used in a signal handler. Anyway, we can let the OS handle the signal by just returning if it not `SIG_DFL` or `SIG_IGN`:
```c
  static void safe_read_signal_handler(int sig, siginfo_t* info, void* context) {
      // If we're in a safe_read, recover via longjmp
      if (g_is_safe_read) {
          siglongjmp(g_safe_read_handler, 1);
      }

      // Not in safe_read - chain to previous handler
      struct sigaction* prev = (sig == SIGBUS) ? &g_prev_sigbus_handler : &g_prev_sigsegv_handler;

      // Chain to previous SA_SIGINFO handler
      if (prev->sa_flags & SA_SIGINFO && prev->sa_sigaction) {
          prev->sa_sigaction(sig, info, context);
          return;
      }

      // Chain to previous simple handler
      if (prev->sa_handler && prev->sa_handler != SIG_DFL && prev->sa_handler != SIG_IGN) {
          prev->sa_handler(sig);
          return;
      }

      // Previous handler was SIG_DFL or SIG_IGN - just return
  }
```

maxep (dismissed):
LGTM 👍 Great work on the UserDefault and the safe memory access! I have left minor comments

DatadogProfiling/Mach/mach_sampling_profiler.cpp
simaoseica-dd: This is an handler to handle the crash in the `stack_trace_sample_thread` func. If it is in the Profiling `memcpy` of the `fp` (validated by `g_is_safe_read `), the handler will recover it. If not, the context will be restored to the crash be handled as usual. 
I removed the previous `signal` function and added tests for this recover logic in the `SafeReadTests`.

DatadogProfiling/Sources/AppLaunchProfiler.swift
simaoseica-dd: Yes, you right. But the simultaneous read and write operation `+=` is not properly isolated by the `ReadWriteLock`.

maxep (commented):
Looks great, thanks for the optimisation 👍

PR #2615, opened 2025-12-12
RUM-13323: Add dynamic sample rate on the Profiling configuration

simaoseica-dd:
### What and why?

This PR exposes a user-facing API that allows configuring the Profiling sample rate, which was previously fixed at 10%.

### How?

It introduces a new `sampleRate` parameter in `Profiling.Configuration`, which can be set when enabling the `Profiling` feature.

### Review checklist

### DIFF
--- a/DatadogProfiling/Tests/ProfilerFeatureTests.swift
+++ b/DatadogProfiling/Tests/ProfilerFeatureTests.swift
@@ -101,29 +101,4 @@ final class ProfilerFeatureTests: XCTestCase {
         // Then
         XCTAssertEqual(userDefaults.value(forKey: DD_PROFILING_SAMPLE_RATE_KEY) as? SampleRate, previousSampleRate)
     }
-
-    func testInit_usesProvidedUserDefaultsInstance() {
-        // Given
-        let otherSuiteName = "ProfilerFeatureTests-other-\(UUID().uuidString)"
-        let otherUserDefaults = UserDefaults(suiteName: otherSuiteName)!
-        otherUserDefaults.removePersistentDomain(forName: otherSuiteName)
-
-        otherUserDefaults.setValue(false, forKey: DD_PROFILING_IS_ENABLED_KEY)
-        otherUserDefaults.setValue(SampleRate.maxSampleRate, forKey: DD_PROFILING_SAMPLE_RATE_KEY)
-
-        // When
-        _ = ProfilerFeature(
-            requestBuilder: requestBuilder,
-            messageReceiver: messageReceiver,
-            sampleRate: 10,
-            userDefaults: userDefaults
-        )
-
-        XCTAssertEqual(userDefaults.value(forKey: DD_PROFILING_IS_ENABLED_KEY) as? Bool, true)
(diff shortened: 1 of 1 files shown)


DatadogProfiling/Tests/ProfilingConfigurationTests.swift
```diff
@@ -0,0 +1,20 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import XCTest
+import DatadogProfiling
+
+final class ProfilingConfigurationTests: XCTestCase {
+    func testDefaultConfiguration() {
```
maxep, line 11: suggestion/ This test has not much value, we can add it when we have business logic top test.

DatadogProfiling/Tests/ProfilerFeatureTests.swift
```diff
@@ -0,0 +1,129 @@
...
+            messageReceiver: messageReceiver,
+            sampleRate: higherSampleRate,
+            userDefaults: userDefaults
+        )
+
+        // Then
+        XCTAssertEqual(userDefaults.value(forKey: DD_PROFILING_SAMPLE_RATE_KEY) as? SampleRate, previousSampleRate)
+    }
+
+    func testInit_usesProvidedUserDefaultsInstance() {
```
maxep: This test has not much value

DatadogProfiling/Sources/ProfilerFeature.swift
```diff
@@ -29,16 +29,28 @@ internal final class ProfilerFeature: DatadogRemoteFeature {
...
+
+    private func set(sampleRate: SampleRate, in userDefaults: UserDefaults) { //swiftlint:disable:this required_reason_api_name
+        let previousSampleRate = userDefaults.value(forKey: DD_PROFILING_SAMPLE_RATE_KEY) as? SampleRate
+
+        // Profiling will use the lowest sample rate
+        // if there is more than one SDK instance initialized.
+        if previousSampleRate == nil || previousSampleRate ?? .maxSampleRate > sampleRate {
+            userDefaults.setValue(sampleRate, forKey: DD_PROFILING_SAMPLE_RATE_KEY)
+        }
+    }
```
maxep, line 55: question/ Does this means user won't be able to increase the sample rate ? 🤔

maxep (commented):
Looks good! I left a question and a couple of nits

DatadogProfiling/Sources/ProfilerFeature.swift
simaoseica-dd: This logic is a protection for multiple `Profiling` features running at the same time in multiple SDK instances. It takes into account only the current session settings since the `DD_PROFILING_SAMPLE_RATE_KEY` is reset every time the profiler runs.

maxep: Ah ok, I missed the part where `delete_profiling_defaults` is called

maxep (approved):

PR #2616, opened 2025-12-15
RUM-13256 Fix KSCrash context sync

maxep:
### What and why?

Fixes user-info synchronization in KSCrash integration by switching from base64-encoded context to direct JSON injection. The previous approach required triple encoding (JSON → Base64 → JSON), which was too slow to reliably inject context before a crash occurs. Additionally adds telemetry for better observability of KSCrash operations.

### How?

- Changed context injection from base64 string in `userInfo` dictionary to direct JSON via `kscrash_setUserInfoJSON()`. This eliminates the encoding overhead and ensures context is injected quickly enough before crashes.
- Added telemetry events for installation failures and crash report loading errors.
- Configured `KSCrash` to use only signal-based crash monitoring, excluding Mach exceptions and align with `PLCrashReporter` config.

### Review checklist

### DIFF
--- a/DatadogCrashReporting/Sources/KSCrashIntegration/KSCrashPlugin.swift
+++ b/DatadogCrashReporting/Sources/KSCrashIntegration/KSCrashPlugin.swift
@@ -23,16 +23,12 @@ internal import KSCrashFilters
 /// Pass its instance as the crash reporting plugin for Datadog SDK to enable crash reporting feature.
 @objc
 internal class KSCrashPlugin: NSObject, CrashReportingPlugin {
-    private let kscrash: KSCrash
+    private let store: CrashReportStore
     private let telemetry: Telemetry
 
     init(_ kscrash: KSCrash = .shared, telemetry: Telemetry = NOPTelemetry()) throws {
-        self.kscrash = kscrash
-        self.telemetry = telemetry
-
         do {
             try kscrash.install(with: .datadog())
-
             kscrash.reportStore?.sink = CrashReportFilterPipeline(
                 filters: [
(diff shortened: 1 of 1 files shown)


DatadogCrashReporting/Sources/KSCrashIntegration/KSCrashPlugin.swift
```diff
@@ -72,20 +73,19 @@ internal class KSCrashPlugin: NSObject, CrashReportingPlugin {
...
             }
         }
     }
 
     func inject(context: Data) {
-        // Convert Data to base64 string for JSON serialization compatibility
-        // NSJSONSerialization doesn't support NSData directly
-        let contextBase64 = context.base64EncodedString()
-        kscrash.userInfo = [CrashField.dd.rawValue: contextBase64]
+        kscrash_setUserInfoJSON(String(data: context, encoding: .utf8))
```
gonzalezreal: What happens if `context` contains invalid UTF-8? Should we handle the `nil` case from `String(data:encoding:)`?

maxep: Good point! I have an even straighter solution to avoid this:
```swift
context.withUnsafeBytes {
    let c_char = $0.bindMemory(to: CChar.self).baseAddress
    kscrash_setUserInfoJSON(c_char)
}
```

gonzalezreal: Neat! 👌🏽

PR #2618, opened 2025-12-16
RUM-11471: Add app launch telemetry

simaoseica-dd:
### What and why?

This PR adds telemetry for App launch vitals, in line with the [internal spec](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/5903974405/RUM+App+Launch+Telemetry+-+Spec). 
The goal is to understand the timing characteristics of each app launch type, with a particular focus on the variations across cold starts.

### How?

It introduces a new telemetry controller, `AppLaunchMetricController`, responsible for collecting information during app launch through the `AppLaunchMetric` metric.

### Review checklist

### DIFF
--- a/DatadogRUM/Sources/RUMMonitor/Scopes/Utils/StartupTypeHandler.swift
+++ b/DatadogRUM/Sources/RUMMonitor/Scopes/Utils/StartupTypeHandler.swift
@@ -7,7 +7,7 @@
 import DatadogInternal
 import Foundation
 
-internal enum ColdStartRule: CaseIterable {
+internal enum ColdStartRule: String, CaseIterable {
     case freshInstall
     case appUpdate
     case systemRestart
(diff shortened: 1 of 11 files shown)


DatadogRUM/Sources/RUMMonitor/Scopes/Utils/StartupTypeHandler.swift
```diff
@@ -35,21 +38,25 @@ internal final class StartupTypeHandler {
             switch rule {
             case .freshInstall:
                 if appStateManager.previousAppStateInfo == nil {
+                    telemetryController.track(coldStartRule: rule)
```
mariedm, line 41: Why not calling `telemetryController.track(coldStartRule: rule)` once before the switch instead of repeating it in each case statement?

simaoseica-dd: Before the switch it can be a `.warmStart`. The switch returns `.coldStart` only if the conditions are satisfied.

maxep (approved):

PR #2620, opened 2025-12-18
RUM-12429 Add Dual-Mode to Automatic Network Tracking

mariedm:
## What and why?

This PR improves **Automatic Network Instrumentation**, enabling network request tracking **without requiring delegate registration**. This significantly simplifies SDK integration while maintaining backward compatibility with the existing metrics-based approach.

## Key Changes

- **Automatic Mode**: Tracks all `URLSession` tasks without delegate registration
- **Metrics Mode**: Original delegate-based approach that captures detailed timings
- **Dual-Mode Architecture**: Both modes can coexist
  - Automatic mode provides baseline tracking
  - Metrics mode builds on automatic mode by additionally capturing detailed timings (opt-in)

See [RFC](https://datadoghq.atlassian.net/wiki/x/u4DDVQE) (internal) for more details on the decided approach.

## Core Implementation

**1. Tracking Modes** (`TrackingMode` enum in `URLSessionTaskInterception`):
- `automatic`: Tracks all tasks via `setState:` and completion handler swizzling
- `metrics`: Tracks tasks with registered delegates and captures `URLSessionTaskMetrics`

**2. Swizzling Strategy**:
- **setState:**: Detects task completion for delegate-less and async/await tasks
  - New `URLSessionTaskStateSwizzler`
  - Follows the existing swizzling pattern used by existing swizzlers
- **Completion handlers**: Captures data and errors for tasks with completion handlers
- **Delegate methods**: Captures `URLSessionTaskMetrics` and data chunks (metrics mode only)

**3. Updated Completion Logic** (`isDone` in `URLSessionTaskInterception`):
- **Automatic mode**: Complete when `completion != nil` OR `state >= 2` (no metrics required)
- **Metrics mode**: Complete when `completion != nil` AND `metrics != nil` (wait for both)

**4. SDK Request Filtering**:
- Domain-based: Filters Datadog intake domains (`*-intake-datadoghq.com`, `instrumentation.datadoghq.com`, etc.)
- Header-based: Filters requests with `DD-REQUEST-ID` header (for custom endpoints in tests)
- Prevents SDK internal requests from being tracked as RUM resources

## API Changes

**`URLSessionInstrumentation.enableOrThrow(with:in:)` Signature Change:**

```swift
// Before: configuration was required
internal static func enableOrThrow(with configuration: Configuration, in core: DatadogCoreProtocol)

// After: configuration is optional and method is @_spi(Internal) public
@_spi(Internal)
public static func enableOrThrow(with configuration: Configuration?, in core: DatadogCoreProtocol)
```

Optional configuration parameter: `configuration == nil` enables automatic mode, `configuration != nil` enables metrics mode.

1. This design allows:
- RUM and Trace to enable automatic mode by calling `enableOrThrow(with: nil, in: core)`
- Customers to optionally enable metrics mode by calling `enable(with: .init(delegateClass: ...))`

2. Visibility: Made `@_spi(Internal) public` so it can be called from RUM and Trace modules.

**Automatic Mode Enablement** (no code changes required):
- Enabled automatically when `RUM.enable(with:)` or `Trace.enable(with:)` is called with `urlSessionTracking` configuration
- Tracks all `URLSession` tasks globally without any delegate registration

**Metrics Mode Enablement** (optional):
```swift
// Enable metrics mode for detailed timing breakdown
URLSessionInstrumentation.enable(with: .init(delegateClass: MyDelegate.self))
```
Note: this API will be deprecated in a follow-up PR in favor of a new, clearer API. 

## Data Captured

| Feature               | Automatic Mode                    | Metrics Mode                                |
|-----------------------|-----------------------------------|---------------------------------------------|
| Duration              | ✅ Yes                            | ✅ Yes                                      |
| Status code           | ✅ Yes                            | ✅ Yes                                      |
| Response size         | ✅ Yes (via `countOfBytesReceived`) | ✅ Yes (via `URLSessionTaskMetrics`)          |
| Errors                | ✅ Yes                            | ✅ Yes                                      |
| Response data         | Completion handlers only | ✅ Yes |
| Detailed timing       | ❌ No                             | ✅ Yes (DNS, SSL, TTFB, connect, download)  |

Note on data capture: Response data is available in `ResourceAttributesProvider` callback for:
- Automatic mode: Tasks with completion handlers
- Metrics mode: All tasks (via delegate didReceive or completion handler)

Async/await and delegate-only tasks in automatic mode will receive nil for the data parameter, but all other parameters (request, response, error) remain available.

## Tests

**Unit tests**:
- ✅ `setState:` swizzling in `URLSessionTaskStateSwizzlerTests`
- In `NetworkInstrumentationFeatureTests`: 
  - ✅ All mode combinations: automatic only, metrics only, dual-mode
  - ✅ Multiple `URLSession` request types covered:
    - dataTask (with URL, URLRequest, completion handler)
    - uploadTask (with/without completion handler)
    - downloadTask
    - async/await tasks
    - Combine tasks
  - ✅ Intake filtering (domain-based + header-based)

**Integration tests**:
- ❌ No new integration tests specifically for dual-mode, will come-up in a follow-up PR
- ✅ Fixed and validated existing integration tests that now exercise automatic mode (`NetworkInstrumentationIntegrationTests`, `RUMResourcesScenarioTests`, `TracingURLSessionScenarioTests`, etc.) 

## Notes

- **Backward Compatible**: No breaking changes - existing metrics mode behavior unchanged
  - As stated in the [RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/5733843131/RFC+-+iOS+Automatic+Network+Instrumentation) (internal), more requests will be captured by default; customers can filter them out by using the `resourceEventMapper` in [RUMConfiguration](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogRUM/Sources/RUMConfiguration.swift#L28).

- **Distributed Tracing**: when customers call `Trace.enable(with:)`, automatic mode is enabled, but they still need to call `URLSessionInstrumentation.enable(with:)` with their delegate to enable metrics mode if they want distributed tracing to work. Spans need accurate timing from `URLSessionTaskMetrics`. [RUM-13455](https://datadoghq.atlassian.net/browse/RUM-13455) is tracking if we want to add automatic mode support for distributed tracing as well.

- **Delegate Swizzling:** An earlier iteration attempted to clean-up `NetworkInstrumentationFeature`'s implementation and minimize delegate swizzling in metrics mode (only swizzling `didFinishCollecting` for metrics capture). However, the full delegate swizzling was kept for two reasons:
  1. **Data capture for `ResourceAttributesProvider`**: Swizzling `didReceive` is necessary to provide response data to customers via the `resourceAttributesProvider` callback.
  2. **iOS 15 compatibility**: Full delegate swizzling ensures backward compatibility with iOS 15, which couldn't be fully tested due to Xcode version constraints; keeping comprehensive swizzling minimizes risk for older OS versions.

### Review checklist

[RUM-13455]: https://datadoghq.atlassian.net/browse/RUM-13455?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> Introduces dual-mode network tracking with automatic `URLSession` instrumentation and preserves metrics mode for detailed timing.
> 
> - Adds automatic mode: tracks tasks via `setState:` and completion-handler swizzling; no delegate required; filtered out SDK intake requests (via `DD-REQUEST-ID`)
> - Keeps metrics mode: delegate-based; now coexists with automatic mode and captures `URLSessionTaskMetrics` and response data
> - New `URLSessionTaskStateSwizzler` and updates to `NetworkInstrumentationFeature` to manage modes, delegate registration, and interception flow
> - Updates `URLSessionInstrumentation.enableOrThrow` to accept optional config and be `@_spi(Internal) public`; RUM now enables automatic mode by default
> - Enhances `URLSessionTaskInterception`: adds `TrackingMode`, response-size fallback, and mode-specific `isDone` logic; RUM resource size now falls back to `responseSize`
> - Extensive new/updated tests covering automatic, metrics, and dual-mode paths, async/await, Combine, uploads/downloads, and filtering; project files updated accordingly
> - CHANGELOG entry added
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 51df984450a8a9b29488b7d5201063a29819f2fc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
+++ b/DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
@@ -502,7 +502,7 @@ extension NetworkInstrumentationFeature {
     ///
     /// - Parameters:
     ///   - task: The task whose state has changed.
-    ///   - state: The new state of the task (maps to URLSessionTask.State).
+    ///   - state: The new state of the task (maps to URLSessionTask.State: 0=running, 1=suspended, 2=canceling, 3=completed).
     internal func task(_ task: URLSessionTask, didChangeToState state: Int) {
         queue.async { [weak self] in
             guard let self = self, let interception = self.interceptions[task] else {
@@ -512,9 +512,11 @@ extension NetworkInstrumentationFeature {
             // Register the state change
             interception.register(state: state)
 
-            // When task completes (state == 3), also register the response/error if not already done
+            // When task completes, also register the response/error if not already done
             // This ensures completion is recorded even for async/await or delegate-less tasks without completion handlers
-            if state == 3 && interception.completion == nil {
+            // Note: We wait for .completed rather than .canceling because task.error is only populated
+            // after the task fully transitions to completed state
+            if state == URLSessionTask.State.completed.rawValue && interception.completion == nil {
                 interception.register(
                     response: task.response,
                     error: task.error
--- a/DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
+++ b/DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
@@ -313,8 +313,7 @@ internal final class NetworkInstrumentationFeature: DatadogFeature {
 }
 
 extension NetworkInstrumentationFeature {
-
-    /// Determines whether this swizzler should intercept a given task based on the configuration mode.
+    /// Determines whether a task should be intercepted based on the tracking mode.
     ///
     /// - Metrics mode (delegate class configured): Only intercepts tasks with the registered delegate
     /// - Automatic mode (no delegate class): Intercepts all tasks except those with delegates registered in metrics mode
--- a/DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
+++ b/DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
@@ -96,34 +96,23 @@ internal final class NetworkInstrumentationFeature: DatadogFeature {
                     return
                 }
 
-                // Determine if this swizzler should intercept this task
-                let shouldIntercept: Bool
-                if let delegateClass = configuration?.delegateClass {
-                    // Metrics mode: only intercept tasks with the registered delegate
-                    shouldIntercept = task.dd.delegate?.isKind(of: delegateClass) == true
-                } else {
-                    // Automatic mode: skip if task has a delegate registered in metrics mode
-                    if let delegate = task.dd.delegate, self.isRegisteredDelegate(delegate) {
-                        shouldIntercept = false
-                    } else {
-                        shouldIntercept = true
-                    }
+                // Determine if this swizzler should intercept this task based on the configuration
+                guard shouldInterceptTask(task, for: configuration) else {
+                    return
                 }
 
                 // Only perform interception if this swizzler should handle this task
                 // This allows the swizzler chain to continue for tasks we don't handle
-                if shouldIntercept {
-                    var injectedTraceContexts: [TraceContext]?
+                var injectedTraceContexts: [TraceContext]?
 
-                    let configuredFirstPartyHosts = FirstPartyHosts(firstPartyHosts: configuration?.firstPartyHostsTracing) ?? .init()
-                    if let currentRequest = task.currentRequest {
-                        let (request, traceContexts) = self.intercept(request: currentRequest, additionalFirstPartyHosts: configuredFirstPartyHosts)
-                        task.dd.override(currentRequest: request)
-                        injectedTraceContexts = traceContexts
-                    }
-
-                    self.intercept(task: task, with: injectedTraceContexts ?? [], additionalFirstPartyHosts: configuredFirstPartyHosts, trackingMode: trackingMode)
+                let configuredFirstPartyHosts = FirstPartyHosts(firstPartyHosts: configuration?.firstPartyHostsTracing) ?? .init()
+                if let currentRequest = task.currentRequest {
+                    let (request, traceContexts) = self.intercept(request: currentRequest, additionalFirstPartyHosts: configuredFirstPartyHosts)
+                    task.dd.override(currentRequest: request)
+                    injectedTraceContexts = traceContexts
                 }
+
+                self.intercept(task: task, with: injectedTraceContexts ?? [], additionalFirstPartyHosts: configuredFirstPartyHosts, trackingMode: trackingMode)
             }
         )
 
@@ -324,6 +313,31 @@ internal final class NetworkInstrumentationFeature: DatadogFeature {
 }
 
 extension NetworkInstrumentationFeature {
+
+    /// Determines whether this swizzler should intercept a given task based on the configuration mode.
+    ///
+    /// - Metrics mode (delegate class configured): Only intercepts tasks with the registered delegate
+    /// - Automatic mode (no delegate class): Intercepts all tasks except those with delegates registered in metrics mode
+    ///
+    /// This coordination prevents double-processing when both automatic and metrics modes are enabled.
+    ///
+    /// - Parameters:
+    ///   - task: The URLSessionTask to check
+    ///   - configuration: The instrumentation configuration
+    /// - Returns: `true` if this swizzler should intercept the task, `false` otherwise
+    private func shouldInterceptTask( _ task: URLSessionTask, for configuration: URLSessionInstrumentation.Configuration?) -> Bool {
+        if let delegateClass = configuration?.delegateClass {
+            // Metrics mode: only intercept tasks with our registered delegate
+            return task.dd.delegate?.isKind(of: delegateClass) == true
+        } else {
+            // Automatic mode: skip tasks with delegates registered in metrics mode
+            if let delegate = task.dd.delegate, isRegisteredDelegate(delegate) {
+                return false
+            }
+            return true
+        }
+    }
+
     /// Checks if a URLRequest is an SDK internal request that should not be tracked
     ///
     /// - Parameter request: The URLRequest to check.
--- a/DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
+++ b/DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
@@ -324,54 +324,13 @@ internal final class NetworkInstrumentationFeature: DatadogFeature {
 }
 
 extension NetworkInstrumentationFeature {
-    /// Checks if a URLRequest is an SDK internal request that should not be tracked.
-    ///
-    /// Uses two filtering mechanisms:
-    /// 1. **Domain-based filtering**: Checks for production Datadog intake domains (datadoghq.com, datad0g.com)
-    /// 2. **Header-based filtering**: Checks for DD-REQUEST-ID header present in all SDK upload requests
-    ///
-    /// The header-based check is necessary for custom endpoints used in tests (e.g., in `RUMResourcesScenarioTests` in our UI Integration tests),
-    /// where SDK upload requests would otherwise be tracked as RUM resources.
+    /// Checks if a URLRequest is an SDK internal request that should not be tracked
     ///
     /// - Parameter request: The URLRequest to check.
     /// - Returns: `true` if the request is an SDK internal request, `false` otherwise.
-    ///
-    ///   Will be filtered (SDK domains):
-    ///     - browser-intake-datadoghq.com
-    ///     - instrumentation.datadoghq.com
-    ///     - logs.browser-intake-datadoghq.com
-    ///
-    ///   Will NOT be filtered (public domains):
-    ///     - imgix.datadoghq.com (CDN - no "intake" or "instrumentation")
-    ///     - www.datadoghq.com (website)
     private func isDatadogIntakeRequest(_ request: URLRequest?) -> Bool {
-        // Check domain
-        if isDatadogIntakeRequest(request?.url) {
-            return true
-        }
-
         // Check for DD-REQUEST-ID header (present in all SDK upload requests, including custom endpoints)
         return request?.value(forHTTPHeaderField: URLRequestBuilder.HTTPHeader.ddRequestIDHeaderField) != nil
-    }
-
-    /// Checks if the given URL is a Datadog intake request to prevent infinite recursion.
-    ///
-    /// - Parameter url: The URL to check.
-    /// - Returns: `true` if the URL is a Datadog intake request, `false` otherwise.
-    private func isDatadogIntakeRequest(_ url: URL?) -> Bool {
-        guard let host = url?.host else {
-            return false
-        }
-
-        // Filter Datadog intake domains
-        // Must contain "intake" or "instrumentation" to avoid false positives (e.g., imgix.datadoghq.com)
-        let isDatadogDomain = host.hasSuffix("datadoghq.com")
-                           || host.hasSuffix("datadoghq.eu")
-                           || host.hasSuffix("ddog-gov.com")
-                           || host.hasSuffix("datad0g.com")
-        let isIntakeOrInstrumentation = host.contains("intake") || host.contains("instrumentation")
-
-        return isDatadogDomain && isIntakeOrInstrumentation
     }
 
     /// Intercepts the provided request by injecting trace headers based on first-party hosts configuration.
@@ -543,7 +502,7 @@ extension NetworkInstrumentationFeature {
     ///
     /// - Parameters:
     ///   - task: The task whose state has changed.
-    ///   - state: The new state of the task (0=Suspended, 1=Running, 2=Canceling, 3=Completed).
+    ///   - state: The new state of the task (maps to URLSessionTask.State).
     internal func task(_ task: URLSessionTask, didChangeToState state: Int) {
         queue.async { [weak self] in
             guard let self = self, let interception = self.interceptions[task] else {
--- a/DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
+++ b/DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
@@ -14,11 +14,11 @@ import Foundation
 ///     let core: DatadogCoreProtocol
 ///
 ///     let handler: DatadogURLSessionHandler = CustomURLSessionHandler()
-///     core.register(urlSessionInterceptor: handler)
+///     core.register(urlSessionHandler: handler)
 ///
-/// Registering multiple interceptor will aggregate instrumentation.
+/// Registering multiple handlers will aggregate instrumentation.
 internal final class NetworkInstrumentationFeature: DatadogFeature {
-    /// The Feature name: "trace-propagation".
+    /// The Feature name: "network-instrumentation".
     static let name = "network-instrumentation"
 
     /// Network Instrumentation serial queue for safe and serialized access to the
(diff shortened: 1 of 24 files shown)


DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskInterception.swift
```diff
@@ -14,10 +25,31 @@ public class URLSessionTaskInterception {
...
+    ///
+    /// Data is collected in:
+    /// - Metrics mode: via `URLSessionDataDelegate.urlSession(_:dataTask:didReceive:)` swizzling
+    /// - Automatic mode: via completion handler swizzling (only for tasks with completion handlers)
+    ///
+    /// Can be `nil` if:
+    /// - Task completed with error
+    /// - Task has no completion handler in automatic mode (e.g., async/await, tasks without handlers)
+    /// - Task is a download task (data is saved to file instead of captured in memory)
     public private(set) var data: Data?
```
maxep, line 42: question/ Is the `data` used for something else than getting the response size?

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskStateSwizzler.swift
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /// Swizzles `URLSessionTask.setState:` method.
+    class TaskSetState: MethodSwizzler<@convention(c) (URLSessionTask, Selector, Int) -> Void, @convention(block) (URLSessionTask, Int) -> Void> {
+        private static let selector = NSSelectorFromString("setState:")
+
+        private let method: Method
+
+        static func build() throws -> TaskSetState {
+            let className = "__NSCFLocalSessionTask"
```
maxep, line 59: suggestion/ Can we add a comment why we swizzle `__NSCFLocalSessionTask` and not `URLSessionTask`?

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
```diff
@@ -43,6 +43,12 @@ internal final class NetworkInstrumentationFeature: DatadogFeature {
     @ReadWriteLock
     private var swizzlers: [ObjectIdentifier: NetworkInstrumentationSwizzler] = [:]
 
+    /// Tracks delegate classes registered in metrics mode.
+    /// Used to prevent automatic mode from processing tasks that are handled by metrics mode.
+    /// Maps ObjectIdentifier to the actual class type for isKind(of:) checks.
+    @ReadWriteLock
+    private var registeredDelegateClasses: [ObjectIdentifier: AnyClass] = [:]
```
maxep, line 50: question/ From the PR description, this is necessary for backward compatibility with iOS 15, right? Could you elaborate the issue with iOS 15? 
Because, to my understanding, the `metrics` mode could be built on top of the `automatic` mode, where the delegate swizzling only collects the additional metrics. But it seems that they are 2 separated data flows, or am I missing something?

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
```diff
@@ -56,87 +62,174 @@ internal final class NetworkInstrumentationFeature: DatadogFeature {
...
-                    self?.task(task, didCompleteWithError: task.error)
+                    let configuredFirstPartyHosts = FirstPartyHosts(firstPartyHosts: configuration?.firstPartyHostsTracing) ?? .init()
+                    if let currentRequest = task.currentRequest {
+                        let (request, traceContexts) = self.intercept(request: currentRequest, additionalFirstPartyHosts: configuredFirstPartyHosts)
+                        task.dd.override(currentRequest: request)
+                        injectedTraceContexts = traceContexts
+                    }
+
+                    self.intercept(task: task, with: injectedTraceContexts ?? [], additionalFirstPartyHosts: configuredFirstPartyHosts, trackingMode: trackingMode)
                 }
```
maxep: nit/ I think this would read better, feel free to discard:
```swift
// Determine if this swizzler should intercept this task
if let delegateClass = configuration?.delegateClass, task.dd.delegate?.isKind(of: delegateClass) == true {
    return // Metrics mode: only intercept tasks with the registered delegate
} else if let delegate = task.dd.delegate, self.isRegisteredDelegate(delegate) {
    return // Automatic mode: skip if task has a delegate registered in metrics mode
}

// Only perform interception if this swizzler should handle this task
// This allows the swizzler chain to continue for tasks we don't handle
var injectedTraceContexts: [TraceContext]?

let configuredFirstPartyHosts = FirstPartyHosts(firstPartyHosts: configuration?.firstPartyHostsTracing) ?? .init()
if let currentRequest = task.currentRequest {
    let (request, traceContexts) = self.intercept(request: currentRequest, additionalFirstPartyHosts: configuredFirstPartyHosts)
    task.dd.override(currentRequest: request)
    injectedTraceContexts = traceContexts
}

self.intercept(task: task, with: injectedTraceContexts ?? [], additionalFirstPartyHosts: configuredFirstPartyHosts, trackingMode: trackingMode)
```

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
```diff
@@ -150,9 +243,137 @@ internal final class NetworkInstrumentationFeature: DatadogFeature {
...
+    ///   Will NOT be filtered (public domains):
+    ///     - imgix.datadoghq.com (CDN - no "intake" or "instrumentation")
+    ///     - www.datadoghq.com (website)
+    private func isDatadogIntakeRequest(_ request: URLRequest?) -> Bool {
+        // Check domain
+        if isDatadogIntakeRequest(request?.url) {
+            return true
+        }
+
+        // Check for DD-REQUEST-ID header (present in all SDK upload requests, including custom endpoints)
```
maxep, line 345: question/ Why using domains then? we could only rely on `DD-REQUEST-ID`?

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
```diff
@@ -56,87 +62,174 @@ internal final class NetworkInstrumentationFeature: DatadogFeature {
...
+
+        // Swizzle completion handlers for `URLSession.dataTask(with:completionHandler:)` methods
+        //
+        // In dual-mode (automatic + metrics), both modes swizzle completion handlers.
+        // To prevent double-counting, automatic mode skips tasks with registered delegates.
         try swizzler.swizzle(
-            delegateClass: configuration.delegateClass,
-            interceptDidReceive: { [weak self] session, task, data in
-                self?.task(task, didReceive: data)
+            interceptCompletionHandler: { [weak self] task, _, error in
```
maxep, line 164: question/ I don't get why we still need to swizzle `URLSession.dataTask(with:completionHandler:)` 🤔  Can't we solely rely on the `setState:` and wait for the `.completed` state?

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
```diff
@@ -294,6 +538,72 @@ extension NetworkInstrumentationFeature {
         handlers.forEach { $0.interceptionDidComplete(interception: interception) }
         interceptions[task] = nil
     }
+
+    /// Tells the interceptors that the task's state has changed.
+    ///
+    /// - Parameters:
+    ///   - task: The task whose state has changed.
+    ///   - state: The new state of the task (0=Suspended, 1=Running, 2=Canceling, 3=Completed).
```
maxep: Are these different than the `URLSessionTask.State` enum?

maxep (commented):
Great work! The level of comments really helps understanding the logic. I left several questions but I guess the main one being: Can we build the `.metrics` instrumentation on top of the `.automatic` one? So we avoid 2 distinct data flows and minimize the complexity. Is it possible or is there limitations that I am missing?

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskInterception.swift
mariedm: Yes, the data is passed to the customer though `ResourceAttributesProvider` (configured in [RUM Configuration](https://github.com/DataDog/dd-sdk-ios/blob/31d330b91901c0b768ecc768561cd1cebbd6b12f/DatadogRUM/Sources/RUMConfiguration.swift#L341)), allowing them to enrich resource attributes.

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
mariedm: No, they are the same.

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
mariedm: Good point! I initially only had domain filtering and later realised this was not enough.

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
maxep: Then the comment is wrong on the values:
```swift
public enum State : Int, @unchecked Sendable {
    case running = 0
    case suspended = 1
    case canceling = 2
    case completed = 3
}
```

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
mariedm: Not exactly. 

Metrics mode _is_ built on top of the automatic mode where `setState` swizzling is used as the foundation, and metrics mode adds delegate swizzling to collect metrics.

The complexity exists because:
1. Metrics mode also needs to collect response data, which customers rely on in `ResourceAttributesProvider`, and for this we need `didReceive` swizzling
2. API calls are sequential:
- Customers enable automatic mode first (when enabling RUM)
- Later, they _optionally_ enable metrics mode for specific delegate(s) 

Because of this, automatic mode must be able to work alone since we don't know if metrics mode will be added. When both modes are enabled, we have two callback sources (`setState` + delegate) for the same task. `registeredDelegateClasses` prevents double-processing.

Like I tried to explain in the PR description, I tried ti simplify the delegate swizzling logic but couldn't because: 
- `didReceive` swizzling is needed for data collection
- the iOS 15 check prevents removing the rest of delegate swizzling

To eliminate this coordination/double-callback, we'd need to change the API to prevent sequential enabling or simplify delegate swizzling despite the iOS 15 constraints.

mariedm: As for iOS 15, these comments are previous to my work, and I wasn't able to test (can only use XCode 16+ and iOS 16+ on DD laptop).

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
mariedm: Because `setState` can detect completion but cannot capture response body data; and to complete my [previous answer](https://github.com/DataDog/dd-sdk-ios/pull/2620#discussion_r2676339315), data is not only necessary for customers using `ResourceAttributesProvider`, but also to extract GraphQL errors

DatadogTrace/Tests/TracingURLSessionHandlerTests.swift
```diff
@@ -427,7 +434,8 @@ class TracingURLSessionHandlerTests: XCTestCase {
 
         core.context.applicationStateHistory = .mockAppInForeground()
 
-        let interception = URLSessionTaskInterception(request: .mockAny(), isFirstParty: true)
+        // TODO: RUM-13455 - Trace requires metrics mode for now
```
simaoseica-dd: why?

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskInterception.swift
```diff
@@ -27,11 +59,15 @@ public class URLSessionTaskInterception {
     ///
     /// Setting the value to 'rum' will indicate that the span is reported as a RUM Resource.
     public private(set) var origin: String?
+    /// Task state tracked via `setState:` swizzling.
+    /// State values: 0=Suspended, 1=Running, 2=Canceling, 3=Completed
+    private var taskState: Int?
```
ncreated: suggestion/ Using `Int` relies on a Foundation implementation detail and obscures the intent, e.g.:
```swift
let isStateComplete = (taskState ?? 0) == 3`)
```
or
```swift
if state == 3 {
   completionExpectation.fulfill()
}
```
These magic `Int` values in code and tests are hard to understand without additional comments. We could significantly improve readability and maintainability by using [`URLSessionTask.State`](https://developer.apple.com/documentation/foundation/urlsessiontask/state-swift.enum) instead of raw ints.

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
ncreated: > 2. API calls are sequential:
> - Customers enable automatic mode first (when enabling RUM)
> - Later, they optionally enable metrics mode for specific delegate(s)

Since `URLSessionInstrumentation` requires delegate **class** and not an _instance_, I think this sequentiality could be removed in the future. I can imagine future version of RUM requiring list of delegate classes in RUM config.

ncreated (commented):
The implementation looks really good 🎯. I left one piece of feedback regarding using an enum instead of raw integers.

I agree with @maxep that, architecture-wide, the added complexity is concerning, but acceptable as long as it doesn’t leak into public APIs. It would be ideal to track this as technical debt and have "simplification" plan for the future once we drop support for older iOS versions or have an opportunity for an API change in V4.

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
mariedm: That sounds like a great idea! 

To clarify: the sequential constraint is _not_ just about API design; I added it on purpose to simplify implementation, in particular tracking task state. By enforcing automatic --> metrics order, this ensures delegate swizzling is always added on top of automatic tracking. This way, we don't need to handle swizzling order dependencies. But if we require delegate classes upfront, this would allow us to enable both modes at RUM init and potentially simplify greatly implementation.

To go back to your general comment:

> It would be ideal to track this as technical debt and have "simplification" plan for the future once we drop support for older iOS versions or have an opportunity for an API change in V4.

I'll write a short doc about this and link it to our v4 doc.

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
mariedm: I created a helper function to make it easier to read

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
mariedm: Here's [the doc](https://datadoghq.atlassian.net/wiki/x/iQETaQE) (internal)

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
mariedm: Fixed

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionTaskInterception.swift
mariedm: Done 👍

maxep (approved):
LGTM! As we discussed, the complexity of this feature has reached its maximum IMO. we should plan a follow based on your debt doc 👍 
Thanks for navigating in of this :)

ncreated (approved):
🎯

PR #2628, opened 2025-12-24
RUM-6263: Screen change based recording

gonzalezreal:
### What and why?

Session Replay currently captures the screen every 100ms using a fixed timer, regardless of whether anything is
changing. This wastes CPU cycles when the screen is idle. This PR replaces the timer with an event-driven system that
only captures when the screen actually changes, reducing overhead during idle periods while maintaining the same
10fps capture rate during animations and interactions.

This work builds on the ideas from #2074, which demonstrated the performance benefits of event-driven capture.

The change is behind a feature flag (`.screenChangeScheduling`, off by default) so we can dog-food before making it
generally available.

### How?

Rather than removing the existing `Scheduler` abstraction, this implementation preserves it by adding a new
`ScreenChangeScheduler` that fits cleanly into the existing architecture. The scheduler wraps a `ScreenChangeMonitor`
that swizzles three `CALayer` methods (`display()`, `draw(in:)`, and `layoutSublayers()`) to detect when visual
changes occur. This covers every change on the screen, including animations (both content and layout-based) and scrolling.

When a layer change is detected, the monitor aggregates it with other pending changes and delivers them after a 100ms
minimum interval. This rate limiting ensures we never exceed 10fps, even during rapid animations.

> [!NOTE]  
> The layer change snapshot system may seem over-engineered for simple scheduling, but it's designed with the future
> in mind. I'm prototyping a `CALayer`-based capture system that could eventually replace our current UIKit/SwiftUI
> semantic capture, and it will leverage these detailed layer change snapshots to know exactly which layers changed
> and how.

The integration is minimal. `SessionReplayFeature` uses a simple ternary to choose between `ScreenChangeScheduler` and the existing `MainThreadScheduler` based on the feature flag. `RecordingCoordinator` doesn't change at all, which means zero risk to the core recording logic.

Testability was a priority. Everything uses a generic `TimeProvider` protocol, so tests can manually advance time instead of waiting for real timers.

The performance win comes from eliminating those 10 wasted captures per second when nothing is happening on
screen, while still capturing smoothly when things are animating or the user is interacting.

| Timer scheduling  | Screen change scheduling |
| ------------- | ------------- |
|   |   |

### Review checklist


DatadogSessionReplay/Sources/Utilities/Schedulers/ScreenChangeScheduler.swift
```diff
@@ -0,0 +1,85 @@
...
+                return
+            }
+            monitor.stop()
+            self.monitor = nil
+        }
+    }
+
+    private func screenDidChange(_ snapshot: CALayerChangeSnapshot) {
+        // ScreenChangeMonitor notifies on the main thread
+        DD.logger.debug("Screen changed: \(snapshot)")
```
ncreated, line 75: question/ The `snapshot` is currently only used for debug printing. Do we expect it to be used beyond that? If not, I’m wondering whether the complexity of `CALayerChange`, `CALayerChangeAggregator`, and `CALayerChangeSnapshot` is worth it just for conditional debug output.

DatadogSessionReplay/Sources/Recorder/ScreenChangeMonitor/TimeProvider.swift
```diff
@@ -0,0 +1,71 @@
...
+
+    private let queue: DispatchQueue
+
+    init(queue: DispatchQueue = .main) {
+        self.queue = queue
+    }
+
+    func schedule(after interval: TimeInterval, _ action: @escaping () -> Void) -> Task {
+        let timer = DispatchSource.makeTimerSource(queue: queue)
+        timer.schedule(deadline: .now() + interval)
```
ncreated: optimization/ We may want to leverage the [`leeway` parameter](https://developer.apple.com/documentation/dispatch/dispatchsourcetimer/schedule(deadline:repeating:leeway:)-hvhp), which introduces scheduling tolerance and helps reduce battery drain. This is the GCD counterpart of the `tolerance` parameter that [we already adopt in the `MainThreadScheduler`](https://github.com/DataDog/dd-sdk-ios/blob/c08030668e8c7e706309fc655cff6b76356f14e5/DatadogSessionReplay/Sources/Utilities/Schedulers/MainThreadScheduler.swift#L16-L21).

DatadogSessionReplay/Sources/Recorder/ScreenChangeMonitor/TimeProvider.swift
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    private let queue: DispatchQueue
+
+    init(queue: DispatchQueue = .main) {
+        self.queue = queue
+    }
+
+    func schedule(after interval: TimeInterval, _ action: @escaping () -> Void) -> Task {
+        let timer = DispatchSource.makeTimerSource(queue: queue)
```
ncreated, line 62: sanity check/ We’re creating, resuming, handling, and canceling a `DispatchSourceTimer` every time something changes on screen (debounced at ~100ms). Do we know if this introduces any noticeable overhead?

ncreated (approved):
Great idea overall - I’m really happy to see something smarter than the previous timer approach. The code looks neat and well-structured, and the tests are solid 👍  

My only concern is whether `CALayerChangeSnapshot` adds justified complexity (as mentioned in another comment), since the overall concept seems like it could work the same without it.  

With that in mind, I’m good to approve.

DatadogSessionReplay/Sources/Utilities/Schedulers/ScreenChangeScheduler.swift
gonzalezreal: Fair concern. As I mentioned in the description:
> The layer change snapshot system may seem over-engineered for simple scheduling, but it's designed with the future in mind. I'm prototyping a `CALayer`-based capture system that could eventually replace our current UIKit/SwiftUI semantic capture, and it will leverage these detailed layer change snapshots to know exactly which layers changed and how.

DatadogSessionReplay/Sources/Recorder/ScreenChangeMonitor/TimeProvider.swift
gonzalezreal: I took some measurements:
- Timer creation takes on average 0.0148 milliseconds (14.8 microseconds)

So I think we're good with this approach

DatadogSessionReplay/Sources/Recorder/ScreenChangeMonitor/TimeProvider.swift
gonzalezreal: Good catch! Fixed in [f291d4e](https://github.com/DataDog/dd-sdk-ios/pull/2628/commits/f291d4e3be613b7ca4cf29aa48549defc90b277a)

DatadogSessionReplay/Sources/Utilities/Schedulers/ScreenChangeScheduler.swift
ncreated: Ah, got it 👍 - that can justify keeping the snapshot machinery even if it’s a bit more than we need today.

DatadogSessionReplay/Sources/Recorder/ScreenChangeMonitor/CALayerChange.swift
```diff
@@ -0,0 +1,40 @@
...
+
+            init(rawValue: Int8) {
+                self.rawValue = rawValue
+            }
+
+            static let display = Self(rawValue: 1 << Aspect.display.rawValue)
+            static let draw = Self(rawValue: 1 << Aspect.draw.rawValue)
+            static let layout = Self(rawValue: 1 << Aspect.layout.rawValue)
+        }
+    }
```
maxep, line 35: nit/ Seems like the definition is redundant and the `enum Aspect` itself is not used, I think this could be:
```swift
internal struct CALayerChange: Equatable {
    struct AspectSet: OptionSet {
        let rawValue: Int8

        init(rawValue: Int8) {
            self.rawValue = rawValue
        }

        static let display = Self(rawValue: 1 << 0)
        static let draw = Self(rawValue: 1 << 1)
        static let layout = Self(rawValue: 1 << 2)
    }

    weak var layer: CALayer?
    var aspects: AspectSet
}
```

DatadogSessionReplay/Sources/Recorder/ScreenChangeMonitor/CALayerChangeAggregator.swift
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        isRunning = false
+        pendingChanges.removeAll()
+        scheduledDelivery?.cancel()
+        scheduledDelivery = nil
+    }
+
+    private func record(_ layer: CALayer, aspect: CALayerChange.Aspect.Set) {
+        // Only record on the main thread
```
maxep, line 61: question/ Is this a possibility? the swizzled methods should be always called from the main thread.
Not a change-request, I'm just wondering.

maxep (approved):
Looks great, well done! I left a minor nit and a question, but it is good to go IMO 👍

DatadogSessionReplay/Sources/Recorder/ScreenChangeMonitor/CALayerChange.swift
gonzalezreal: You're right, it may seem redundant. I wanted to have a type-safe way of referring to an individual aspect (`.display`) besides the zero or more aspects that the `OptionSet` provides (`[.display, .draw]`, `[]`). I will consider removing the enum if I don't use individual aspect values.

DatadogSessionReplay/Sources/Recorder/ScreenChangeMonitor/CALayerChangeAggregator.swift
gonzalezreal: While UIKit can only be used from the main thread, CoreAnimation can be used from background threads: layer preparation, offscreen rendering, async drawing in `CATiledLayer`. For our use case, we are only interested in `CALayer` changes on the main thread.

PR #2634, opened 2026-01-09
RUM-13456 Introduce new trackMetrics API

mariedm:
### What and why?

This PR is a follow-up from #2620. It introduces a new API: `trackMetrics(with:in:)`. See [RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/5733843131/RFC+-+iOS+Automatic+Network+Instrumentation) (internal) for reference.

### How?

- Added `URLSessionInstrumentation.trackMetrics(with:in:)` API
- Deprecated `URLSessionInstrumentation.enable(with:in:)`API 
- Sends debug telemetry for each API

### Review checklist

### DIFF
--- a/DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionInstrumentation.swift
+++ b/DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionInstrumentation.swift
@@ -24,7 +24,7 @@ public enum URLSessionInstrumentation {
 
             core.telemetry.debug(
                 id: "URLSessionInstrumentation:trackMetrics",
-                message: "URLSession metrics mode enabled"
+                message: "URLSession metrics mode enabled " + isFirstPartyHostsTracing(for: configuration)
             )
         } catch let error {
             consolePrint("\(error)", .error)
@@ -47,7 +47,7 @@ public enum URLSessionInstrumentation {
 
             core.telemetry.debug(
                 id: "URLSessionInstrumentation:enable",
-                message: "URLSession metrics mode enabled (deprecated API)"
+                message: "URLSession metrics mode enabled " + isFirstPartyHostsTracing(for: configuration) + " (deprecated API)"
             )
         } catch let error {
             consolePrint("\(error)", .error)
@@ -89,6 +89,11 @@ public enum URLSessionInstrumentation {
         }
 
         feature.unbind(delegateClass: delegateClass)
+    }
+
+    /* Helper */
+    private static func isFirstPartyHostsTracing(for configuration: URLSessionInstrumentation.Configuration) -> String {
+        return configuration.firstPartyHostsTracing != nil ? "(with FPH Tracing configured)" : "(without FPH Tracing configured)"
     }
 }
 


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - [IMPROVEMENT] Remove `application_start` action from `ApplicationLaunch`. See [#2533][]
 - [FEATURE] Track Slow Frames (view hitches) by default. See [#2631][]
 - [FEATURE] Automatic network instrumentation now tracks `URLSession` requests without requiring delegate registration. See [#2620][]
+- [FEATURE] Deprecate `URLSessionInstrumentation.enable(with:in:)` API in favor of `URLSessionInstrumentation.trackMetrics(with:in:)`. See [#2634][]
```
barboraplasovska: Was it intentional to add it under the release 3.5.0 ?

DatadogCore/Sources/NetworkInstrumentation/URLSessionInstrumentation+objc.swift
```diff
@@ -56,10 +56,24 @@ public final class objc_URLSessionInstrumentationFirstPartyHostsTracing: NSObjec
...
+    /// Enables metrics mode to capture detailed timing breakdowns for URLSession tasks.
+    ///
+    /// This method is optional. Automatic network tracking is already enabled by default when RUM or Trace is initialized with URL session tracking configuration.
+    /// Metrics mode provides additional detailed timing information captured from `URLSessionTaskMetrics` (including DNS, Connection, SSL, First Byte, Download).
+    ///
+    /// Note: Metrics mode involves swizzling `URLSessionDataDelegate` methods to capture `URLSessionTaskMetrics`.
+    ///
+    /// - Parameters:
+    ///   - configuration: Configuration of the feature.
+    public static func trackMetrics(configuration: objc_URLSessionInstrumentationConfiguration) {
```
barboraplasovska: /suggestion To keep the same naming conventions as in the Swift API ?

```suggestion
    public static func trackMetrics(with configuration: objc_URLSessionInstrumentationConfiguration) {
```

barboraplasovska (commented):
Great job!!

CHANGELOG.md
mariedm: No, good catch! I didn't update the changelog after rebasing.

DatadogCore/Sources/NetworkInstrumentation/URLSessionInstrumentation+objc.swift
mariedm: Good point! Looking at other Objc APIs, we actually follow this pattern. The deprecated API was actually inconsistent.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionInstrumentation.swift
```diff
@@ -8,14 +8,47 @@ import Foundation
...
+    ///
+    /// This method is optional. Automatic network tracking is already enabled by default when RUM or Trace is initialized with `urlSessionTracking` configuration.
+    /// Metrics mode provides additional detailed timing information captured from `URLSessionTaskMetrics` (including DNS, Connection, SSL, First Byte, Download).
+    ///
+    /// Note: Metrics mode involves swizzling `URLSessionDataDelegate` methods to capture `URLSessionTaskMetrics`.
+    ///
+    /// - Parameters:
+    ///   - configuration: Configuration of the feature.
+    ///   - core: The instance of Datadog SDK to enable URLSession instrumentation in (global instance by default).
+    public static func trackMetrics(with configuration: URLSessionInstrumentation.Configuration, in core: DatadogCoreProtocol = CoreRegistry.default) {
```
ncreated, line 21: question/ The [`URLSessionInstrumentation.Configuration`](https://github.com/DataDog/dd-sdk-ios/blob/19db3afca5bafd34971e11eb8701f18bd0f699a2/DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionInstrumentation.swift#L64-L68) defines two configuration options:

- `delegateClass`
- `firstPartyHostsTracing`

While `delegateClass` determines which `URLSession` instances will have metrics enabled, it’s unclear to me how `firstPartyHostsTracing` is intended to be used in the `trackMetrics(with:in:)` call. What's the role and expected behavior of `firstPartyHostsTracing` in this context?

mariedm: Following our discussion, we decided to keep the behavior unchanged, meaning the new `trackMetrics` API will still require a `URLSessionInstrumentation.Configuration` parameter. I'll add telemetry to measure how often is `firstPartyHostsTracing` configured per delegate. This will allow us to determine if we can drop this parameter in a future version of the API.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionInstrumentation.swift
```diff
@@ -8,14 +8,47 @@ import Foundation
...
+        do {
+            try enableOrThrow(with: configuration, in: core)
+
+            core.telemetry.debug(
+                id: "URLSessionInstrumentation:trackMetrics",
+                message: "URLSession metrics mode enabled " + isFirstPartyHostsTracing(for: configuration)
+            )
```
ncreated, line 28: It would be more idiomatic to define and send [usage telemetry](https://github.com/DataDog/rum-events-format/blob/master/schemas/telemetry/usage/mobile-features-schema.json), but I get why `DEBUG` is easier here. Given how infrequently this API is called, this seems fine 👍

PR #2638, opened 2026-01-13
RUM-13572: Improve buffer flushing in Profiling

simaoseica-dd:
### What and why?

This PR improves the Profiling feature to handle correctly the flushing if the mach profiler was deleted before.
Additionally it improves the reliability of main thread identification to avoid duplicated `Thread Ids` in the profile.  

### How?
- Removes a deadlock that might happen during timeouts. Also, it only deletes the mach profiler when the ctor profiler is deallocated to avoid crashes when there is no profiler while buffer flushing .
- Moves the initialization of `g_main_pthread` and safe read handlers to the better points in the profiling timeline. This guarantees these are initialized only after the profiler is actually about to run, preventing duplicated `Thread Ids` in the profile.
- Renames the `sampleRate` config parameter to `applicationLaunchSampleRate` to clarify its scope.

### Review checklist

### DIFF
--- a/DatadogProfiling/Mach/ctor_profiler.cpp
+++ b/DatadogProfiling/Mach/ctor_profiler.cpp
@@ -227,8 +227,6 @@ public:
         if (!profiler) return;
         status = CTOR_PROFILER_STATUS_STOPPED;
         profiler->stop_sampling();
-        delete profiler;
-        profiler = nullptr;
     }
 
     profile* get_profile() const { return profile; }


DatadogProfiling/Mach/mach_sampling_profiler.cpp
```diff
@@ -488,6 +507,9 @@ bool mach_sampling_profiler::start_sampling() {
         target_thread = pthread_self();
     }
 
+    // Initialize main thread ID and safe read handlers for sampling
+    init_main_thread_id_and_safe_read_handlers();
```
maxep: There is no guarantee that a start profiling will be called from the main thread. The `mach_sampling_profiler` is not dedicated to App Launch profiling, we should keep the ability to start and  stop from any thread.

DatadogProfiling/Mach/mach_sampling_profiler.cpp
```diff
@@ -230,26 +237,38 @@ bool binary_image_lookup_pc(binary_image_t* info, void* pc) {
...
-    // Get UUID from the image
-    const struct mach_header_64* header = (const struct mach_header_64*)dl_info.dli_fbase;
-    if (!header || header->magic != MH_MAGIC_64) return false;
+    // Get UUID from the image.
+    // The image might be unloaded from memory so a safe_read_memory is called
+    struct mach_header_64 header_buf;
+    if (!safe_read_memory(dl_info.dli_fbase, &header_buf, sizeof(header_buf))) return false;
+    
+    const struct mach_header_64* header = &header_buf;
+    if (header->magic != MH_MAGIC_64) return false;
```
maxep: What does this check do?
If it's platform specific, we would need platform variants like in the `thread_get_frame_pointers`.

DatadogProfiling/Mach/mach_sampling_profiler.cpp
simaoseica-dd: Good point for custom profiling 👍

DatadogProfiling/Mach/mach_sampling_profiler.cpp
maxep: It is actually [this](https://developer.apple.com/documentation/kernel/mach_header_64/1525718-magic). Code was just moved, please ignore..

DatadogProfiling/Mach/mach_sampling_profiler.cpp
```diff
@@ -127,14 +133,10 @@ static void init_safe_read_handlers() {
...
- * This constructor runs early to capture and setup the parameters.
+ * This function should be called from the main thread early in the process lifecycle.
  */
-__attribute__((constructor))
-static void init_main_thread_id_and_safe_read_handlers() {
-    if ((is_profiling_enabled() ||
-        ctor_profiler_get_status() != CTOR_PROFILER_STATUS_NOT_CREATED) == false) {
-        return;
-    }
+void init_main_thread_id_and_safe_read_handlers() {
```
maxep: cr/ Now the problem is that `init_main_thread_id_and_safe_read_handlers` is never called from the `mach_sampling_profiler`, but it needs the `g_is_safe_read` and `g_main_pthread` for walking the stack and thread info.

I actually don't really get why this change is necessary and how it relates to the flushing buffer issue described here, or to duplicate ids. To me, it makes sense to init the `g_main_pthread` and `g_safe_read_handler` in a _constructor_ and this will be executed way before a buffer in flushed.

DatadogProfiling/Mach/mach_sampling_profiler.cpp
```diff
@@ -230,26 +237,38 @@ bool binary_image_lookup_pc(binary_image_t* info, void* pc) {
         }
     }
 
-    // Get UUID from the image
-    const struct mach_header_64* header = (const struct mach_header_64*)dl_info.dli_fbase;
-    if (!header || header->magic != MH_MAGIC_64) return false;
+    // Get UUID from the image.
+    // The image might be unloaded from memory so a safe_read_memory is called
```
maxep: In which case a library can be unloaded? because system libraries cannot be unloaded. In other cases a user will have to call `dlclose`, which sounds unlikely in a mobile app 🤔

DatadogProfiling/Mach/mach_sampling_profiler.cpp
simaoseica-dd: That was the previous solution (2 constructors) where the Profiler initializes the sampling without the `init_main_thread_id_and_safe_read_handlers`. Only after some sampling this init constructor was invoked, creating duplicated ids and no safe reads in this interval. Bear in mind that this second constructor depends on the first one.
This is the best solution taking into account that sampling is managed by a single Profiler.

DatadogProfiling/Mach/mach_sampling_profiler.cpp
simaoseica-dd: Yes, `dlclose` and library unloading are rare but can happen. This saves us from invalid pointers, whether caused by `dlclose` or by random heap corruption.

DatadogProfiling/Mach/mach_sampling_profiler.cpp
maxep: >Only after some sampling this init constructor was invoked, creating duplicated ids and no safe reads in this interval. Bear in mind that this second constructor depends on the first one.

Yes, I understand that. And that's why the `ctor_profiler_auto_start` has the attribute _constructor(65535)_ and the `init_main_thread_id_and_safe_read_handlers` has no priority, so the init is run **before**  start. The order of execution is guaranteed by the priority, or should be.

The issue now is that the `mach_sampling_profiler` depends on the `ctor_profiler.start` to have `g_main_pthread` and `g_safe_read_handler` initialized. This is problematic for custom profiles, if the `ctoc_profiler` was disabled or sampled out, the custom profiles won't have valid  `g_main_pthread` nor `g_safe_read_handler`.

simaoseica-dd: The order of execution is always to run the `ctor_profiler_auto_start` first (independently of the priority). It might has to do to how the profiling relationship entities are defined. 
I split the responsibilities and entry points as discussed.

maxep (approved):
Looks great! Thanks for extra effort here 👍

PR #2639, opened 2026-01-14
RUM-12403 Add `manual.drop` feature to Spans

arroz:
### What and why?

Add support for manually keeping and dropping traces.

### How?

A new entity, `SamplingDecision`, was added to `DDSpanContext` that explicitly models each decision mechanism. This includes the individual decision of each mechanism, mechanism precedence, and the possibility of restoring a decision of a mechanism of lower precedence is a higher precedence override is removed. Currently, 3 mechanisms are supported: fallback (implicitly), agent rate and manual override. The code was updated in multiple places to handle the more expressive representation.

`TracingURLSessionHandler` was improved to handle the creation of a span correctly in every situation, without missing any information regardless of which headers are injected, and is the most heavily modified entity (outside of `DDSpan` and related ones).

Although we support the standard `setTag(key: SpanTags.manualKeep, value: true)` and `setTag(key: SpanTags.manualDrop, value: true)` calls to do this, two convenience methods were added: `keepTrace()` and `dropTrace()`. These avoid the ambiguity in the previous methods (what value to use, etc) and their name implies the entire trace is affected by the call, not just the span.

One note: the way mobile SDKs work (specifically being agent-less), the default sampling configuration is actually closer to "local global sampling rate" than agent rate. However, Datadog expects local global sampling rate to report sampling priorities of -1/2, which are override values, not 0/1. Since the SDKs ends up acting as tracer and agent, we consider our default method to be agent rate, even if it works substantially differently than an actual agent.

### Review checklist


DatadogCore/Tests/Datadog/TracerTests.swift
```diff
@@ -883,7 +887,112 @@ class TracerTests: XCTestCase {
...
+
+    func testMarkingManuallyDroppedChildSpanWithChildren() throws {
+        Trace.enable(with: config, in: core)
+        let tracer = Tracer.shared(in: core)
+
+        let span = tracer.startRootSpan(operationName: .mockAny(), customSampleRate: 100)
+        let child = tracer.startSpan(operationName: .mockAny(), childOf: span.context)
+        let grandchild = tracer.startSpan(operationName: .mockAny(), childOf: child.context)
+
+        span.setTag(key: SpanTags.manualDrop, value: true)
```
simaoseica-dd: Shouldn't the drop be in the `child` or `grandchild`?

DatadogInternal/Sources/Models/Trace/SamplingMechanismType.swift
```diff
@@ -0,0 +1,38 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+/// The mechanism used to make a span sampling decision.
+public enum SamplingMechanismType: Equatable {
```
simaoseica-dd: By extending `String` you don't need to handle the dance of strings in the constructor and the computed var.

```suggestion
public enum SamplingMechanismType: String, Equatable {
```

DatadogInternal/Sources/NetworkInstrumentation/Datadog/HTTPHeadersReader.swift
```diff
@@ -44,10 +44,36 @@ public class HTTPHeadersReader: TracePropagationHeadersReader {
         )
     }
 
-    public var sampled: Bool? {
-        if let sampling = httpHeaderFields[TracingHTTPHeaders.samplingPriorityField] {
-            return sampling == "1"
+    public var samplingPriority: SamplingPriority? {
```
simaoseica-dd, line 47: If `SamplingPriority` extends String it can be simplified
```suggestion
    public var samplingPriority: SamplingPriority? {
        SamplingPriority(rawValue: httpHeaderFields[TracingHTTPHeaders.samplingPriorityField] ?? "")
    }
```

DatadogInternal/Sources/NetworkInstrumentation/Datadog/HTTPHeadersReader.swift
```diff
@@ -44,10 +44,36 @@ public class HTTPHeadersReader: TracePropagationHeadersReader {
...
+        if let sampling = httpHeaderFields[TracingHTTPHeaders.samplingPriorityField],
+           let samplingInteger = Int(sampling),
+           let samplingPriority = SamplingPriority(rawValue: samplingInteger) {
+            return samplingPriority
         }
         return nil
     }
+
+    public var samplingDecisionMaker: SamplingMechanismType? {
```
simaoseica-dd, line 55: q: Is this logic only needed for `DatadogTrace`? Can it be moved there?
The `var samplingDecisionMaker: SamplingMechanismType?` as nil in the other readers seems they should not be forced to depend on interfaces that they do not implement.

DatadogInternal/Sources/NetworkInstrumentation/Datadog/HTTPHeadersReader.swift
```diff
@@ -44,10 +44,36 @@ public class HTTPHeadersReader: TracePropagationHeadersReader {
...
     }
 
-    public var sampled: Bool? {
-        if let sampling = httpHeaderFields[TracingHTTPHeaders.samplingPriorityField] {
-            return sampling == "1"
+    public var samplingPriority: SamplingPriority? {
+        if let sampling = httpHeaderFields[TracingHTTPHeaders.samplingPriorityField],
+           let samplingInteger = Int(sampling),
+           let samplingPriority = SamplingPriority(rawValue: samplingInteger) {
+            return samplingPriority
```
maxep: Can we keep the `sampled` property alongside the `samplingPriority`? The `TracePropagationHeadersReader` and `Writer` are exposed by public modules so removing a property would be a breaking change.

DatadogTrace/Sources/SamplingMechanisms.swift
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+import Foundation
+import DatadogInternal
+
+/// Mechanism used for sampling in ``SamplingDecision``.
+internal protocol SamplingMechanism: Equatable {
+    /// The sampling priority provided by this mechanism.
+    var samplingPriority: SamplingPriority { get }
+}
```
maxep: Are we preparing follow up iteration with this abstraction? Because currently, we could use `SamplingPriority` directly instead of `FixedValueMechanism(samplingPriority: samplingPriority)`.

DatadogInternal/Sources/NetworkInstrumentation/Datadog/HTTPHeadersReader.swift
arroz: Yep, I restored the `sampled` priority.

DatadogTrace/Sources/SamplingMechanisms.swift
arroz: You're right. The initial idea was accounting for the possibility of having mechanisms that are more dynamic than just a value set at creation time. However, that is very likely a bad idea, and I don't see any real use case. Even more dynamic future sampling sources, like remote config or user defaults, should calculate their sampling priority at the time of creation of a sampling decision, or as a result of setting an override explicitly through the API.

Therefore, `SamplingMechanism` was removed, and `SamplingDecision` dictionary is now `var mechanisms: [SamplingMechanismType: SamplingPriority]`.

DatadogTrace/Sources/DDSpanContext.swift
```diff
@@ -20,8 +20,30 @@ internal struct DDSpanContext: OTSpanContext {
...
+    ///    - value: The tag value.
+    /// - returns: `true` if the tag should be kept by the span as it would normally be, `false` if the tag should be ignored
+    /// and not kept by the span.
+    func span(_ span: DDSpan, willSetTagWithKey key: String, value: Encodable) -> Bool {
+        if key == SpanTags.manualDrop && value as? Bool == true {
+            samplingDecision.addManualDropOverride()
+            return false
+        } else if key == SpanTags.manualKeep && value as? Bool == true {
+            samplingDecision.addManualKeepOverride()
+            return false
```
simaoseica-dd, line 43: Shouldn't it return `true` since the tag is kept? 
Also, the func signature has span that is not used and at the call site is not clear it will return a bool.

DatadogInternal/Sources/Models/Trace/SamplingMechanismType.swift
```diff
@@ -0,0 +1,38 @@
...
+    /// Fallback mechanism. This mechanism samples all spans. It should never be used, but it's included for completion.
+    case fallback
+    /// The main decision mechanism. Although the SDK runs in an agent-less scenario, we consider the SDK sampling
+    /// decisions act as the agent rate in scenarios with independent tracers and agent.
+    case agentRate
```
simaoseica-dd: q/ why the introduction of a new term instead of the `headBasedSampling` term used across the SDK?

DatadogTrace/Sources/SamplingMechanismPrecedence.swift
```diff
@@ -0,0 +1,34 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import DatadogInternal
+
+/// Precedence used to decide between multiple mechanisms in a ``SamplingDecision``.
+/// Higher values mean higher precedence.
+internal struct SamplingMechanismPrecedence: Comparable {
```
simaoseica-dd: suggestion: You can simplify the precedence by extending only the `SamplingMechanismType`:
`extension SamplingMechanismType: Comparable { }`

DatadogTrace/Sources/SamplingDecision.swift
```diff
@@ -0,0 +1,112 @@
...
+    /// instances are created with an `.agentRate` mechanism, that should remain there for the lifetime
+    /// of a decision. Conversely, the `.manual` mechanism should only be present when there is a manual
+    /// override. In that situation, both mechanisms are kept, so the decision can revert to the `.agentRate`
+    /// mechanism if the override is removed.
+    ///
+    /// - Remark: A `.fallback` mechanism always exists implicitly. However, as an optimization, it's not
+    /// actually instantiated nor included in this dictionary. Getters like ``samplingPriority`` and
+    /// ``decisionMaker`` should simulate its existence when necessary.
+    @ReadWriteLock
+    private var mechanisms: [SamplingMechanismType: SamplingPriority]
```
simaoseica-dd, line 31: This dictionary is open to every combination, but it seems it shouldn't be. For instance a `[.manual: .autoKeep]`. 
Why not the merge of the `SamplingPriority` with `SamplingMechanismType` in a `SamplingMechanismType` with associated values?

DatadogInternal/Sources/NetworkInstrumentation/TracePropagationHeadersReader.swift
```diff
@@ -6,14 +6,26 @@
...
+
+    /// The sampling decision maker mechanism based on the provided headers.
+    var samplingDecisionMaker: SamplingMechanismType? { get }
+
     /// Indicates whether the trace was sampled based on the provided headers.
     var sampled: Bool? { get }
 }
+
+public extension TracePropagationHeadersReader {
+    var sampled: Bool? {
```
simaoseica-dd: it is losing the public access

DatadogCore/Tests/Datadog/TracerTests.swift
arroz: Yes. 🤦‍♂️ Fixing!

DatadogInternal/Sources/Models/Trace/SamplingMechanismType.swift
arroz: I tend to hesitate when the raw value is not something inherently part of the entity, but in this case I think it's acceptable. Fixing as suggested.

DatadogInternal/Sources/NetworkInstrumentation/Datadog/HTTPHeadersReader.swift
arroz: In this case I prefer to leave it as an Int, since they are conceptually defined as integers that have an intrinsic order with a meaning.

DatadogTrace/Sources/DDSpanContext.swift
arroz: Based on a conversation with the backend ingestion folks, I believe we are not supposed to add the actual tag. This aligns with what other SDKs do, for example: https://github.com/DataDog/dd-trace-py/blob/6796453b62bd10100e0907b21fb4776e70d05fb9/ddtrace/_trace/span.py#L366 and https://datadoghq.atlassian.net/browse/AIT-6623 (internal).

My understanding from that conversations is the tag is just a "fallback API" for situations where there is not a more specific one (like the one I'm adding on this PR) but it should only change the sampling priority and not adding an actual tag to the meta keys.

This is also consistent with the span model, since it would be awkward for a root span (or whatever span the user interacts with) to have such tag, but none of the other spans in the same trace.

DatadogInternal/Sources/Models/Trace/SamplingMechanismType.swift
arroz: Head based sampling is a strategy that defines a sampling decision is done at the root span, and propagated to all the child spans, local or distributed. The sampling mechanisms are methods to make sampling decisions that may, or not, belong to that category. For example, we define agent rate and custom sampling rules as being head based samplers, and manual keep/dropping as nearly head based sampler (it's still head based if defined in the root span or in a situation where no span belonging to the trace has been ingested yet, or any request propagation happened before any span that is part of the trace is marked manually kept or dropped).

As I stated in the comment, we are slightly stretching the definition of agent rate (as we have no agent to automatically raise and lower the sampling rate dynamically based on the number of requests per minute) but it's the most appropriate method for our situation. Since it uses the same identification tag ("1"), it would be confusing to call it something else when all the other SDKs use the agent rate terminology.

DatadogTrace/Sources/SamplingMechanismPrecedence.swift
arroz: Good point, there is no need for the separate priority entity. I moved everything to SamplingMechanismType.

DatadogTrace/Sources/SamplingDecision.swift
arroz: We have intentionally public classes (the header readers like `HTTPHeadersReader`) that may be used by the SDK users to parse the headers of any request. This means our representation needs to be flexible enough to support any combination of mechanisms and priorities, even if they are out of spec from our point of view.

DatadogInternal/Sources/NetworkInstrumentation/Datadog/HTTPHeadersReader.swift
arroz: The `W3CHTTPHeadersReader` was improved with a better implementation of `samplingPriority` and a working implementation of `samplingDecisionMaker`, based on the vendor-specific code. That makes two of the three mechanisms supporting both APIs. As discussed offline, we'll leave the implementation where it is for now.

PR #2644, opened 2026-01-21
RUM-13455 Support Automatic Network Tracking Mode in Tracing

mariedm:
### What and why?

Enables the Trace feature to create spans in automatic network tracking mode by using approximate timing when `URLSessionTaskMetrics` are unavailable. 

Previously, `TracingURLSessionHandler` required metrics mode (delegate-based tracking) because it relied on `URLSessionTaskMetrics` for accurate span timing. This meant Trace instrumentation didn't work for tasks tracked in automatic mode.

### How?

- Added timing capture in `URLSessionTaskInterception`: new `startDate` and `endDate` properties capture approximate timestamps
- Updated `TracingURLSessionHandler` to use metrics timing when available (most accurate) and fallback to approximate timing (`startDate`/`endDate`) for automatic mode

### Review checklist

### DIFF
--- a/DatadogTrace/Sources/Integrations/TracingURLSessionHandler.swift
+++ b/DatadogTrace/Sources/Integrations/TracingURLSessionHandler.swift
@@ -97,16 +97,12 @@ internal struct TracingURLSessionHandler: DatadogURLSessionHandler {
     }
 
     func interceptionDidComplete(interception: DatadogInternal.URLSessionTaskInterception) {
-        // Priority: metrics timing (most accurate) > approximate timing (fallback for automatic mode)
-        let fetchStart = interception.metrics?.fetch.start ?? interception.startDate
-        let fetchEnd = interception.metrics?.fetch.end ?? interception.endDate
-
         guard
             interception.isFirstPartyRequest, // `Span` should be only send for 1st party requests
             interception.origin != "rum", // if that request was tracked as RUM resource, the RUM backend will create the span on our behalf
             let tracer = tracer,
-            let startTime = fetchStart,
-            let endTime = fetchEnd,
+            let startTime = interception.fetchStartDate,
+            let endTime = interception.fetchEndDate,
             let resourceCompletion = interception.completion
         else {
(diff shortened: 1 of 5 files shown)


ncreated (dismissed):

DatadogTrace/Sources/Integrations/TracingURLSessionHandler.swift
```diff
@@ -97,15 +97,16 @@ internal struct TracingURLSessionHandler: DatadogURLSessionHandler {
...
 
     func interceptionDidComplete(interception: DatadogInternal.URLSessionTaskInterception) {
-        // TODO: RUM-13455 - Add support for automatic mode in tracing
-        // Currently, tracing requires metrics mode because spans need accurate timing (fetch start/end).
-        // Automatic mode doesn't capture timing without URLSessionTaskMetrics.
-        // Investigate if we want to support automatic mode for Tracing.
+        // Priority: metrics timing (most accurate) > approximate timing (fallback for automatic mode)
+        let fetchStart = interception.metrics?.fetch.start ?? interception.startDate
+        let fetchEnd = interception.metrics?.fetch.end ?? interception.endDate
+
```
arroz: Question: should the interception have getters that provide the most accurate information available? Meaning moving this logic to the interception, either to a method like `mostAccurateStartDateAvailable` (but _definitely_ with a better name!) or making `startDate` actually be that method, and keeping the lowest precision dates in more explicitly named variables, like `fallbackStartDate`, possibly private. In other words, do we ever want to obtain both dates from outside of the interception, or do we alway want the best available?

arroz (commented):
Looks great overall! Left a question/suggestion regarding the API.

DatadogTrace/Sources/Integrations/TracingURLSessionHandler.swift
mariedm: Yes, that's a good point! I'll move this logic to the interception.

PR #2646, opened 2026-01-22
FFL-1721 Emit `flagevaluation` EVP Track

sameerank:
### What and why?

Introduces evaluation logging for feature flags.

#### Evaluation logging vs Exposure logging:                                                                                                                                                                                                                        
- **Exposure** logging captures when a flag evaluation actually impacts the user experience (`doLog=true`). This is used for experimentation metrics and analysis where you need to know which users were exposed to which variants.
- **Evaluation** logging captures all flag evaluations regardless of `doLog`. This is for operational visibility: for monitoring and debugging.

Evaluation data is aggregated (count, first/last timestamps) rather than logged individually, to minimize overhead.

#### Useful references:                                                                                                                                                                                                                                                      
- [EVALLOG specs](https://datadoghq.atlassian.net/wiki/spaces/PANA/pages/5906009399/FFE+SDK+requirements#Evaluation-logging)
- [EVP schema](https://github.com/DataDog/dd-source/blob/main/domains/evp-workers/apps/flageval-worker/schemas/batchedflagevaluations.json)
- Android implementation:
  - https://github.com/DataDog/dd-sdk-android/pull/3147
  - https://github.com/DataDog/dd-sdk-android/pull/3146
  - https://github.com/DataDog/dd-sdk-android/pull/3145
  - https://github.com/DataDog/dd-sdk-android/pull/3144

### How?

- **`EvaluationAggregator`**: Aggregates flag evaluations by a composite key (flag key, variant key, allocation key, targeting key, error message, and context hash). Tracks `evaluationCount`, `firstEvaluation`, and `lastEvaluation` timestamps per unique combination. 
- **Flush behavior**: Aggregated evaluations are flushed to the backend either:                                                                                                                                                                                    
  - On a configurable timer interval (`evaluationFlushInterval`)                                                                                                                                                                                                 
  - When the max aggregations threshold is reached                                                                                                                                                                                          
  - On aggregator deallocation                                                                                                                                                                                                                                 
- **`FlagEvaluationEvent`**: Event model matching the EVP schema. Fields like `variant`, `allocation`, and `runtimeDefaultUsed` are conditionally included based on EVALLOG spec requirements (e.g., `variant`/`allocation` omitted for runtime defaults per EVALLOG.8, `runtimeDefaultUsed` omitted when false per EVALLOG.13).                
- **Integration**: `FlagsClient` records evaluations to the aggregator for all flag resolutions when evaluation logging is enabled, regardless of the `doLog` flag (unlike exposure logging).  

### Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,10 @@
 # Unreleased
 
 - [FEATURE] Add evaluation logging to `DatadogFlags` module. See [#2646][]
+- [FEATURE] Automatic network instrumentation now tracks `URLSession` requests without requiring delegate registration. See [#2620][]
+- [FEATURE] Deprecate `URLSessionInstrumentation.enable(with:in:)` API in favor of `URLSessionInstrumentation.enableDurationBreakdown(with:in:)`. See [#2634][]
 - [IMPROVEMENT] Skip malformed Logs attributes individually instead of dropping the entire event, and log clear error messages. See [#2665][]
+- [IMPROVEMENT] Expose public entities from `DatadogInternal` to prevent `DatadogInternal` imports in customer code. See [#2666][]
 
 # 3.6.1 / 02-02-2026
 
@@ -1034,15 +1037,18 @@ Release `2.0` introduces breaking changes. Follow the [Migration Guide](MIGRATIO
 [#2599]: https://github.com/DataDog/dd-sdk-ios/pull/2599
 [#2612]: https://github.com/DataDog/dd-sdk-ios/pull/2612
 [#2614]: https://github.com/DataDog/dd-sdk-ios/pull/2614
+[#2620]: https://github.com/DataDog/dd-sdk-ios/pull/2620
 [#2631]: https://github.com/DataDog/dd-sdk-ios/pull/2631
 [#2633]: https://github.com/DataDog/dd-sdk-ios/pull/2633
-[#2647]: https://github.com/DataDog/dd-sdk-ios/pull/2647
+[#2634]: https://github.com/DataDog/dd-sdk-ios/pull/2634
+[#2639]: https://github.com/DataDog/dd-sdk-ios/pull/2639
 [#2640]: https://github.com/DataDog/dd-sdk-ios/pull/2640
 [#2646]: https://github.com/DataDog/dd-sdk-ios/pull/2646
-[#2639]: https://github.com/DataDog/dd-sdk-ios/pull/2639
+[#2647]: https://github.com/DataDog/dd-sdk-ios/pull/2647
 [#2654]: https://github.com/DataDog/dd-sdk-ios/pull/2654
-[#2665]: https://github.com/DataDog/dd-sdk-ios/pull/2665
 [#2661]: https://github.com/DataDog/dd-sdk-ios/pull/2661
+[#2665]: https://github.com/DataDog/dd-sdk-ios/pull/2665
+[#2665]: https://github.com/DataDog/dd-sdk-ios/pull/2666
 
 [@00fa9a]: https://github.com/00FA9A
 [@britton-earnin]: https://github.com/Britton-Earnin
--- a/Datadog/Datadog.xcodeproj/project.pbxproj
+++ b/Datadog/Datadog.xcodeproj/project.pbxproj
@@ -1126,6 +1126,8 @@
 		96155A3E2D7F01250029034E /* CustomDump.swift in Sources */ = {isa = PBXBuildFile; fileRef = 960A0D382D6E2490004BB999 /* CustomDump.swift */; };
 		96155A3F2D7F013A0029034E /* ReflectionMirror.swift in Sources */ = {isa = PBXBuildFile; fileRef = 960A0D392D6E2490004BB999 /* ReflectionMirror.swift */; };
 		96155A402D7F014D0029034E /* Reflector.swift in Sources */ = {isa = PBXBuildFile; fileRef = 960A0D3A2D6E2490004BB999 /* Reflector.swift */; };
+		961583BC2EEAD89C006899B0 /* URLSessionTaskStateSwizzlerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 961583BB2EEAD89C006899B0 /* URLSessionTaskStateSwizzlerTests.swift */; };
+		961583BD2EEAD89C006899B0 /* URLSessionTaskStateSwizzlerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 961583BB2EEAD89C006899B0 /* URLSessionTaskStateSwizzlerTests.swift */; };
 		962223592DFC233000D58EEC /* DDSwiftUIRUMActionsPredicateTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 962223572DFC233000D58EEC /* DDSwiftUIRUMActionsPredicateTests.swift */; };
 		9622235A2DFC233000D58EEC /* DDSwiftUIRUMViewsPredicateTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 962223582DFC233000D58EEC /* DDSwiftUIRUMViewsPredicateTests.swift */; };
 		9622235B2DFC233000D58EEC /* DDSwiftUIRUMActionsPredicateTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 962223572DFC233000D58EEC /* DDSwiftUIRUMActionsPredicateTests.swift */; };
@@ -1161,6 +1163,8 @@
 		96867B9B2D0883DD004AE0BC /* ColorReflectionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 96867B9A2D0883DD004AE0BC /* ColorReflectionTests.swift */; };
 		969B3B212C33F80500D62400 /* UIActivityIndicatorRecorder.swift in Sources */ = {isa = PBXBuildFile; fileRef = 969B3B202C33F80500D62400 /* UIActivityIndicatorRecorder.swift */; };
 		969B3B232C33F81E00D62400 /* UIActivityIndicatorRecorderTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 969B3B222C33F81E00D62400 /* UIActivityIndicatorRecorderTests.swift */; };
+		96CD20642ED4808900B4686E /* URLSessionTaskStateSwizzler.swift in Sources */ = {isa = PBXBuildFile; fileRef = 96CD20632ED4808900B4686E /* URLSessionTaskStateSwizzler.swift */; };
+		96CD20652ED4808900B4686E /* URLSessionTaskStateSwizzler.swift in Sources */ = {isa = PBXBuildFile; fileRef = 96CD20632ED4808900B4686E /* URLSessionTaskStateSwizzler.swift */; };
 		96D331ED2CFF740700649EE8 /* GraphicImagePrivacyTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 96D331EC2CFF740700649EE8 /* GraphicImagePrivacyTests.swift */; };
 		96E414142C2AF56F005A6119 /* UIProgressViewRecorder.swift in Sources */ = {isa = PBXBuildFile; fileRef = 96E414132C2AF56F005A6119 /* UIProgressViewRecorder.swift */; };
 		96E414162C2AF5C1005A6119 /* UIProgressViewRecorderTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 96E414152C2AF5C1005A6119 /* UIProgressViewRecorderTests.swift */; };
@@ -3514,6 +3518,7 @@
 		960B26BF2D0360EE00D7196F /* Assets.xcassets */ = {isa = PBXFileReference; lastKnownFileType = folder.assetcatalog; path = Assets.xcassets; sourceTree = "<group>"; };
 		960B26C22D075BD200D7196F /* DisplayListReflectionTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = DisplayListReflectionTests.swift; sourceTree = "<group>"; };
 		96155A282D79A4DF0029034E /* SwiftUIViewNameExtractorIntegrationTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SwiftUIViewNameExtractorIntegrationTests.swift; sourceTree = "<group>"; };
+		961583BB2EEAD89C006899B0 /* URLSessionTaskStateSwizzlerTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = URLSessionTaskStateSwizzlerTests.swift; sourceTree = "<group>"; };
 		962223572DFC233000D58EEC /* DDSwiftUIRUMActionsPredicateTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = DDSwiftUIRUMActionsPredicateTests.swift; sourceTree = "<group>"; };
 		962223582DFC233000D58EEC /* DDSwiftUIRUMViewsPredicateTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = DDSwiftUIRUMViewsPredicateTests.swift; sourceTree = "<group>"; };
 		962900232D8351A7008DFE39 /* TopLevelReflector.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = TopLevelReflector.swift; sourceTree = "<group>"; };
@@ -3535,6 +3540,7 @@
 		96867B9A2D0883DD004AE0BC /* ColorReflectionTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ColorReflectionTests.swift; sourceTree = "<group>"; };
 		969B3B202C33F80500D62400 /* UIActivityIndicatorRecorder.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = UIActivityIndicatorRecorder.swift; sourceTree = "<group>"; };
 		969B3B222C33F81E00D62400 /* UIActivityIndicatorRecorderTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = UIActivityIndicatorRecorderTests.swift; sourceTree = "<group>"; };
+		96CD20632ED4808900B4686E /* URLSessionTaskStateSwizzler.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = URLSessionTaskStateSwizzler.swift; sourceTree = "<group>"; };
 		96D331EC2CFF740700649EE8 /* GraphicImagePrivacyTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = GraphicImagePrivacyTests.swift; sourceTree = "<group>"; };
 		96E414132C2AF56F005A6119 /* UIProgressViewRecorder.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = UIProgressViewRecorder.swift; sourceTree = "<group>"; };
 		96E414152C2AF5C1005A6119 /* UIProgressViewRecorderTests.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = UIProgressViewRecorderTests.swift; sourceTree = "<group>"; };
@@ -6974,6 +6980,7 @@
 		D2160CC029C0DED100FAA9A5 /* URLSession */ = {
 			isa = PBXGroup;
 			children = (
+				96CD20632ED4808900B4686E /* URLSessionTaskStateSwizzler.swift */,
 				614A708D2BF754D700D9AF42 /* ImmutableRequest.swift */,
 				D2181A8A2B0500BB00A518C0 /* NetworkInstrumentationSwizzler.swift */,
 				D270CDDC2B46E3DB0002EACD /* URLSessionDataDelegateSwizzler.swift */,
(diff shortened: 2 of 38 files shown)


DatadogFlags/Sources/Models/FlagEvaluationEvent.swift
```diff
@@ -0,0 +1,133 @@
...
+}
+
+internal struct EvaluationContext: Codable {
+    let geo: GeoInfo?
+    let device: DeviceInfo
+    let os: OSInfo
+    let service: String
+    let version: String
+    let env: String
+    let rum: RUMInfo?
```
leoromanovsky, line 27: Is the contextual informal that's embedded in `rum` relevant (or possible) to annotate from swift? The desired field is `url` which feels very web centric. It's optional so we don't need to add anything in this PR but wondering if you have insight to take back to augment the schema.

DatadogFlags/Sources/Flags.swift
```diff
@@ -70,6 +70,25 @@ public enum Flags {
...
+        ///
+        /// Default: `true`.
+        public var trackEvaluations: Bool
+
+        /// The interval at which aggregated evaluation data is flushed to the server.
+        ///
+        /// Values are clamped to a minimum of 1 second and maximum of 60 seconds.
+        ///
+        /// Default: `10.0` seconds.
+        public var evaluationFlushInterval: TimeInterval
```
leoromanovsky, line 90: reasonable defaults; thanks

DatadogFlags/Tests/Client/EvaluationLoggingTests.swift
```diff
@@ -0,0 +1,759 @@
...
+        let config = Flags.Configuration(evaluationFlushInterval: 120.0)
+
+        // When
+        Flags.enable(with: config, in: core)
+
+        // Then
+        XCTAssertEqual(
+            dd.logger.warnLog?.message,
+            "`Flags.Configuration.evaluationFlushInterval` cannot exceed 60.0s. A value of 60.0s will be used."
+        )
```
leoromanovsky, line 346: 👍

DatadogFlags/Tests/Client/EvaluationLoggingTests.swift
```diff
@@ -0,0 +1,759 @@
...
+
+        // When - Record one more to reach the limit
+        aggregator.recordEvaluation(
+            for: "flag-3", assignment: .mockAnyBoolean(), evaluationContext: .mockAny(), flagError: nil
+        )
+
+        // Then
+        Thread.sleep(forTimeInterval: 0.1)
+
+        XCTAssertEqual(featureScope.eventsWritten.count, 3, "Should auto-flush when maxAggregations reached")
```
leoromanovsky, line 377: nice!

DatadogFlags/Sources/Feature/EvaluationRequestBuilder.swift
```diff
@@ -0,0 +1,92 @@
...
+        case .iPhone, .iPod:
+            return "mobile"
+        case .iPad:
+            return "tablet"
+        case .appleTV:
+            return "tv"
+        case .appleVision, .appleWatch, .other:
+            return "other"
+        }
+    }
```
leoromanovsky: does this mapping exist somewhere in the sdk already (that's standardized by datadog)?

sameerank: Good catch! Yes looks like `normalizedDeviceType` does this for us already https://github.com/DataDog/dd-sdk-ios/pull/2646/commits/691e1f5a3d8653aa27f9a9e6945cdab614cb9b15

DatadogFlags/Sources/Models/FlagEvaluationEvent.swift
sameerank: Yes, `url` is not required in [the schema](https://github.com/DataDog/dd-source/blob/main/domains/evp-workers/apps/flageval-worker/schemas/batchedflagevaluations.json#L32) and actually neither is the entire `rum` object, which is why I had previously set it to `nil`. I don't see anything comparable to `url` among those available in `RUMCoreContext` for `url`:

https://github.com/DataDog/dd-sdk-ios/blob/54674154872a53bf5fdd84511fb807e771435662/DatadogInternal/Sources/Models/RUM/RUMCoreContext.swift#L10-L22

What I can do with what's currently available is to partially populate `RUMInfo` with the `applicationID`, which would still be compliant with [the schema](https://github.com/DataDog/dd-source/blob/main/domains/evp-workers/apps/flageval-worker/schemas/batchedflagevaluations.json): https://github.com/DataDog/dd-sdk-ios/pull/2646/commits/acba889211ecd0fcee62cc4627b2605f2f2b8dec. 

This deviates slightly from [the Android SDK which is populating both](https://github.com/DataDog/dd-sdk-android/blob/7c8d6b776e2c54467b2556142ed5264bc1f5d033/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/net/EvaluationsRequestFactory.kt#L166). 

My hunch is that a field like `path` would be a bit more appropriate for mobile, though it's worth noting that the field is referred to as "url" in the Android SDK. This `activeViewPath` field looks potentially useable to this end, but it's currently internal to `DatadogRUM`:

https://github.com/DataDog/dd-sdk-ios/blob/792e7134c2815dd870a9ecdd0bafdc00c9b3e713/DatadogRUM/Sources/RUMContext/RUMContext.swift#L23

`RUMCoreContext` in `DatadogInternal` is the bridge that would make `activeViewPath` accessible to `DatadogFlags` but I would need a second opinion about making `activeViewPath` available publicly before I can add it there.

DatadogFlags/Sources/Client/FlagsClient.swift
```diff
@@ -196,11 +200,41 @@ extension FlagsClient: FlagsClientProtocol {
...
+            reason: "DEFAULT",
+            doLog: false
+        )
+
+        guard let context = repository.context else {
+            evaluationLogger.logEvaluation(
+                for: key,
+                assignment: defaultAssignment,
+                evaluationContext: .empty,
+                flagError: "PROVIDER_NOT_READY"
```
sameerank: Previously, with just exposure tracking, we didn't log anything if context was missing, but for evaluation tracking, we need to handle error cases. The Android SDK does something similar for `PROVIDER_NOT_READY`: https://github.com/DataDog/dd-sdk-android/blob/9aac19f5083fd137d6a052e6d4fc23a12d6f998f/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/DatadogFlagsClient.kt#L258

DatadogFlags/Sources/Feature/EvaluationRequestBuilder.swift
```diff
@@ -0,0 +1,89 @@
...
+        )
+    }
+
+    private func buildRumContext(from context: DatadogContext) -> EvaluationContext.RUMInfo? {
+        guard let rum = context.additionalContext(ofType: RUMCoreContext.self) else {
+            return nil
+        }
+        return EvaluationContext.RUMInfo(
+            application: EvaluationContext.RUMInfo.ApplicationInfo(id: rum.applicationID),
+            view: nil // viewURL not available in RUMCoreContext
```
sameerank, line 86: Should this be a TODO to use `activeViewPath`? Or should we settle for partially populating the RUM info?

DatadogFlags/Sources/Feature/FlagsEvaluationFeature.swift
```diff
@@ -0,0 +1,28 @@
...
+    init(
+        customIntakeURL: URL?,
+        telemetry: Telemetry
+    ) {
+        requestBuilder = EvaluationRequestBuilder(
+            customIntakeURL: customIntakeURL,
+            telemetry: telemetry
+        )
+        messageReceiver = NOPFeatureMessageReceiver()
+        performanceOverride = PerformancePresetOverride(maxObjectsInFile: 50)
```
sameerank, line 26: This was set to match

https://github.com/DataDog/dd-sdk-ios/blob/9cd729a41082dfc5f88bd02a2a41c5d6b770389c/DatadogFlags/Sources/Feature/FlagsFeature.swift#L85

The default is 500, and so my understanding is that 50 means that evaluations are batched into smaller files = more frequent uploads. I don't recall what the exact reason was for overriding the default

DatadogFlags/Sources/Client/EvaluationAggregator.swift
```diff
@@ -0,0 +1,205 @@
...
+
+import Foundation
+import DatadogInternal
+
+internal final class EvaluationAggregator {
+    private let flushInterval: TimeInterval
+    private let maxAggregations: Int
+    private let dateProvider: any DateProvider
+    private let featureScope: any FeatureScope
+    private let queue = DispatchQueue(label: "com.datadoghq.flags.evaluation-aggregator")
```
gonzalezreal: This serial dispatch queue is being used purely for synchronization, which is overkill for protecting a dictionary. The codebase already provides [`ReadWriteLock`](https://github.com/DataDog/dd-sdk-ios/blob/c430fca7bd294944d07492fbad14f1aaacfc94e4/DatadogInternal/Sources/Concurrency/ReadWriteLock.swift), a `pthread_rwlock_t`-based property wrapper optimized for this exact use case.

Current approach has unnecessary overhead:

- `DispatchWorkItem` allocation on every recordEvaluation call
- GCD scheduler overhead for trivial dictionary operations
- `.barrier` flag in deinit is meaningless on a serial queue
- `queue.sync` in flush() introduces deadlock risk

Suggested change:

```swift
@ReadWriteLock private var aggregations: [AggregationKey: AggregatedEvaluation] = [:]
```

This is a recurring pattern. If this code was AI-generated, please update the agent's instructions to emphasize following existing codebase conventions for concurrency (see `CLAUDE.md` which mentions `ReadWriteLock`). The agent should search for existing patterns before introducing new synchronization mechanisms.

DatadogFlags/Sources/Client/EvaluationAggregator.swift
```diff
@@ -0,0 +1,205 @@
...
+        self.maxAggregations = maxAggregations
+
+        startFlushTimer()
+    }
+
+    deinit {
+        // Timer uses [weak self], so it becomes a no-op after deallocation.
+        // Copy values before async dispatch - self will be deallocated before the closure runs.
+        let scope = featureScope
+        let snapshot = aggregations
```
gonzalezreal: ⚠️ Race condition: `aggregations` is read here without synchronization, before the async dispatch. If there's a pending `recordEvaluation` block in the queue, this is a data race. Using `ReadWriteLock` would fix this.

DatadogFlags/Sources/Client/EvaluationAggregator.swift
```diff
@@ -0,0 +1,205 @@
...
+            targetingRule: targetingRuleKey.map { .init(key: $0) },
+            targetingKey: targetingKey,
+            runtimeDefaultUsed: runtimeDefaultUsed,
+            error: errorMessage.map { .init(message: $0) },
+            context: eventContext
+        )
+    }
+}
+
+extension EvaluationAggregator: Flushable {
```
gonzalezreal: The `Flushable` protocol is documented as `"Awaits completion of all asynchronous operations"`, other implementations use `queue.sync {}` with an empty block just to drain pending work.

This implementation uses it to send buffered data, which is a different concern. With `ReadWriteLock`, there are no async operations to await, making `Flushable` conformance misleading. Consider removing the conformance and using a dedicated method like `sendAggregatedEvaluations()` for the timer.

DatadogFlags/Tests/Client/EvaluationLoggingTests.swift
```diff
@@ -0,0 +1,827 @@
...
+        // Should not have flushed yet
+        XCTAssertEqual(featureScope.eventsWritten.count, 0)
+
+        // When - Record one more to reach the limit
+        aggregator.recordEvaluation(
+            for: "flag-3", assignment: .mockAnyBoolean(), evaluationContext: .mockAny(), flagError: nil
+        )
+
+        // Then
+        Thread.sleep(forTimeInterval: 0.1)
```
gonzalezreal: ⚠️ Using `Thread.sleep` for async synchronization in tests is fragile and can lead to flaky tests, especially under CI load. Prefer using `XCTestExpectation` with `wait(for:timeout:)` to properly synchronize on async completion.

DatadogFlags/Tests/Client/EvaluationLoggingTests.swift
gonzalezreal, line 1: The test coverage is extensive but shows signs of over-testing. Consider consolidating:

- Tests for trivial default values (`testDefaultEvaluationFlushInterval`, `testEvaluationLoggingEnabledByDefaultAndCanBeDisabled`) provide minimal value
- The three `testRuntimeDefaultUsed*` tests could be one parameterized test
- `testTimestampEqualsFirstEvaluation` is a single assertion that could be part of another test

Focus on testing behavior, not spec checkboxes. A smaller set of well-designed tests is easier to maintain and provides better signal when they fail.

DatadogFlags/Sources/Client/EvaluationAggregator.swift
sameerank: Thanks for the guidance! I saw both patterns (`DispatchQueue` and `@ReadWriteLock`) and wasn't sure which one to use. The new `CLAUDE.md` is helpful as a best practices guide for humans too 😆. Updated: https://github.com/DataDog/dd-sdk-ios/pull/2646/commits/de532eb6df6e1c293414e8f75e990dbe561fe201

DatadogFlags/Sources/Client/EvaluationAggregator.swift
sameerank: Removed in https://github.com/DataDog/dd-sdk-ios/pull/2646/commits/de532eb6df6e1c293414e8f75e990dbe561fe201

DatadogFlags/Sources/Client/EvaluationAggregator.swift
sameerank: Removed the `Flushable` conformance, and the timer and tests now call `sendEvaluations()` directly! Since all the method calls look like `aggregator.sendEvaluations()` it seemed self-evident that the evaluations would be aggregated, so I chose to not rename the method: https://github.com/DataDog/dd-sdk-ios/pull/2646/commits/b0e3095e4165d1ee79c29a6f9d34512cdd761898

DatadogFlags/Tests/Client/EvaluationLoggingTests.swift
sameerank: Actually, since the `ReadWriteLock` operations are synchronous, I can remove these `Thread.sleep` calls! The previous `DispatchQueue` approach for managing aggregations was more complicated to test. https://github.com/DataDog/dd-sdk-ios/pull/2646/commits/832e1716aa3b3e8cfa35e887fca1b8aefa68e4c9

DatadogFlags/Tests/Client/EvaluationLoggingTests.swift
sameerank: In this case, I went with a spec checkbox-style of testing to help ensure we stay aligned in behaviors across SDKs. I saw a similar pattern in the OpenFeature SDKs and found it useful to understand how each SDK's approach to aligning with the specs, e.g. [Ruby does `context "Requirement 1.1.2.1" do`](https://github.com/open-feature/ruby-sdk/blob/8cd7b5ef66cccbcae59dd6ae477ffc301f934968/spec/specification/flag_evaluation_api_spec.rb#L13).

That said, I haven't seen this among the Datadog SDKs, so I see how this is an unusual pattern. I applied your recommended consolidation but haven't removed the spec-to-test mapping (but am open to removing and consolidating further if this still feels unwieldy to maintain): https://github.com/DataDog/dd-sdk-ios/pull/2646/commits/e55c9efca082db981d63ecc97d0756c12d572c77

DatadogFlags/Sources/Feature/FlagsFeature.swift
```diff
@@ -61,10 +61,10 @@ internal struct FlagsFeature: DatadogRemoteFeature {
                 flushInterval = Constants.maxEvaluationFlushInterval
             }
 
-            let evaluationFeatureScope = core.scope(for: FlagsEvaluationFeature.self)
+            // Use closure to defer scope access until first use, avoiding Swift exclusivity conflict.
```
sameerank: This is the issue I'm referring to

Image

DatadogFlags/Sources/Client/EvaluationAggregator.swift
```diff
@@ -0,0 +1,186 @@
...
+        self.featureScope = featureScope
+        self.flushInterval = flushInterval
+        self.maxAggregations = maxAggregations
+
+        startFlushTimer()
+    }
+
+    deinit {
+        flushTimer?.invalidate()
+        // Note: We don't flush here due to exclusivity conflict with DatadogCore.stop()
```
sameerank, line 35: This is the conflict I'm referring to

Screenshot 2026-02-02 at 4 26 42 PM

Also saw it on CI, e.g. [here](https://gitlab.ddbuild.io/DataDog/dd-sdk-ios/-/jobs/1391784948) where `FlagsRUMIntegrationTests.testWhenFlagIsEvaluated_itAddsFeatureFlagToRUMView()` failed due to the test crashing

```
Test Suite 'FlagsRUMIntegrationTests' started at 2026-01-31 08:28:26.829.
note: [DatadogSDKTesting] Crash detected! Exiting...
```

During SDK teardown if we call `sendEvaluations()` here, an exclusivity conflict occurs when `EvaluationAggregator.deinit` tries to flush evaluations while `DatadogCore.stop()` is modifying the features dictionary. The chain is:
1. `deinit` → `sendEvaluations()` → `eventWriteContext` → `feature(named:type:)` → reads `features[name]`
2. Meanwhile, `stop()` is writing `features = [:]`

So I ended up needing to bring back `Flushable` conformance, so we have access to the `Flushable.flush()` pre-shutdown hook.

https://github.com/DataDog/dd-sdk-ios/blob/f9119f539964cd6f1a4527dbb59aaa10921d4877/DatadogCore/Sources/Core/DatadogCore.swift#L597-L598

which is [called before `.stop`](https://github.com/DataDog/dd-sdk-ios/blob/f9119f539964cd6f1a4527dbb59aaa10921d4877/DatadogCore/Sources/Core/DatadogCore.swift#L293-L304).

```swift
    func flushAndTearDown() {
        flush()
        ...
        stop()
    }
```

DatadogFlags/Sources/Feature/FlagsFeature.swift
```diff
@@ -54,3 +87,9 @@ internal struct FlagsFeature: DatadogRemoteFeature {
         issueReporter = IssueReporter.default(isGracefulModeEnabled: configuration.gracefulModeEnabled)
     }
 }
+
+extension FlagsFeature: Flushable {
+    func flush() {
+        evaluationAggregator?.sendEvaluations()
+    }
```
sameerank, line 94: Most `Flushable` implementations use this pattern to drain a `DispatchQueue`:

```swift
func flush() {
    queue.sync {}
}
```

The implementation here differs because `EvaluationAggregator` uses `@ReadWriteLock`, whose operations are all synchronous. Since `sendEvaluations()` is synchronous, calling it directly in `flush()` achieves the same goal.

DatadogFlags/Sources/Client/EvaluationAggregator.swift
sameerank: After removing `Flushable` from `EvaluationAggregator`, I discovered that it is useful for flushing evaluation data without crashing on shutdown: calling `sendEvaluations()` in `deinit` caused a "exclusivity conflict" because `deinit` runs inside `stop()` while it's modifying the `features` dictionary. I described the conflict a little more in https://github.com/DataDog/dd-sdk-ios/pull/2646#discussion_r2759937576

Whereas `Flushable.flush()` runs prior to `.stop()` so it acts as a pre-shutdown hook that we can use to clear the `ReadWriteLock` dictionary, so I've added it here https://github.com/DataDog/dd-sdk-ios/pull/2646/commits/74c7257aea2aee2f503efd7fea9b69277ad60d85

It's possible that I'm still misunderstanding `Flushable`'s purpose so if there are other shutdown-related hooks that I haven't yet considered, I am happy to try those instead.

DatadogFlags/Sources/Client/FlagsClient.swift
gonzalezreal: 🤏🏽 minor: Consider extracting these to constants.

```swift
private enum EvaluationError {
   static let providerNotReady = "PROVIDER_NOT_READY"
   static let flagNotFound = "FLAG_NOT_FOUND"
   static let typeMismatch = "TYPE_MISMATCH"
}
```

Notice that the enum above is just used as a namespace.

gonzalezreal (dismissed):
Thanks for addressing the previous feedback!

The move to `@ReadWriteLock` is exactly what I was hoping for, and I appreciate the detailed explanation of the exclusivity conflict that led you back to `Flushable`, good catch.

A few minor things, none of which are blockers:

1. Test naming use a different style than the SDK's typical `testGivenWhenThen` pattern. I am ok with it if you prefer the spec-aligned approach you mentioned. Just flagging for consistency.
2. Small nit, the hardcoded `"PROVIDER_NOT_READY"`, `"FLAG_NOT_FOUND"`, `"TYPE_MISMATCH"` strings could be constants to avoid typos, but not a big deal.
3. `AnyValue: Hashable`. This is now part of the public API. Totally fine, just making sure it's intentional.

Otherwise, this looks good to me!

sameerank:
All good points!

1. The given-when-then convention is nice and so I’ve renamed the tests: https://github.com/DataDog/dd-sdk-ios/pull/2646/commits/34b4672940d95e25c97b502720fe97e821ad03ec
2. Enum added: https://github.com/DataDog/dd-sdk-ios/pull/2646/commits/eb68983511a97669dd7fe40fcee515a732128e2b
3. `AnyValue`’s being `Hashable` and `Equatable` are both used internally but exposed publicly. The alternative would be to convert a non-hashable `public enum AnyValue: Equatable` to an internal wrapper `internal struct HashableAnyValue: Hashable` before using `evaluationContext` as an aggregation key, but I don’t think the extra code is worth it. We could have also made `Equatable` internal-only, but similarly didn’t. I realized though that I needed to run `make api-surface`: https://github.com/DataDog/dd-sdk-ios/pull/2646/commits/52181ac5c22ae4883bd1a39b35a74b1251b254b2
    * I am surprised that CI didn't notify me about this. [The check only happens during release and hotfix](https://github.com/DataDog/dd-sdk-ios/blob/f2c10e6f7ad72571ee822860f82954ba6cfb14e3/.gitlab-ci.yml#L128-L134)

The changes dismissed your approval, so please take a look again 🙏 @gonzalezreal

Issue #2645, opened 2026-01-21
KSCrash crashing with dynamic frameworks


albsala-backmarket:
### Stack trace

0   DatadogCrashReporting         	       0x117527cb0 Swift runtime failure: arithmetic overflow + 0 [inlined]
1   DatadogCrashReporting         	       0x117527cb0 closure #1 in KSCrashBacktrace.generateBacktrace(threadID:) + 2332 (KSCrashBacktrace.swift:87)
2   DatadogCrashReporting         	       0x117527cd4 partial apply for closure #1 in KSCrashBacktrace.generateBacktrace(threadID:) + 28
3   libswiftCore.dylib            	       0x197861dac Sequence._compactMap<A>(_:) + 784
4   libswiftCore.dylib            	       0x197861a94 Sequence.compactMap<A>(_:) + 12
5   DatadogCrashReporting         	       0x117526cb8 KSCrashBacktrace.generateBacktrace(threadID:) + 1440 (KSCrashBacktrace.swift:53)
6   DatadogCrashReporting         	       0x117528dc4 protocol witness for BacktraceReporting.generateBacktrace(threadID:) in conformance KSCrashBacktrace + 24
7   DatadogInternal               	       0x119bdd5b8 CoreBacktraceReporter.generateBacktrace(threadID:) + 312 (BacktraceReporter.swift:71)
8   DatadogInternal               	       0x119bdd850 protocol witness for BacktraceReporting.generateBacktrace(threadID:) in conformance CoreBacktraceReporter + 24
9   DatadogRUM                    	       0x1192cdf10 AppHangsWatchdogThread.main() + 1548 (AppHangsWatchdogThread.swift:168)
10  DatadogRUM                    	       0x1192ced6c @objc AppHangsWatchdogThread.main() + 44
11  Foundation                    	       0x181132d14 __NSThread__start__ + 716
12  libsystem_pthread.dylib       	       0x10a9965ac _pthread_start + 104
13  libsystem_pthread.dylib       	       0x10a991998 thread_start + 8

### Reproduction steps

- Use dynamic frameworks
- Enable RUM with appHangThreshold to 0.25 (in our case) to force the DatadogCrashReporting on hangs
- iOS 26.2 but I believe it doesn't matter
- Run the app on simulator

In this case, when there's a hang, and KSCrash tries to build the stacktrace, it crashes in some scenarios.

### Volume

In our case only in dev because we use static frameworks in production

### Affected SDK versions

3.5.0

[...]

PR #2647, opened 2026-01-22
RUM-13878 Solve underflow crash when computing the offset of a backtrace frame

maxep:
### What and why?

Solve underflow crash when computing the offset of a backtrace frame.
Solve: https://github.com/DataDog/dd-sdk-ios/issues/2645

### How?

Use standard `dladdr()` instead of `ksdl_dladdr()` to work around a bug where `ksdl_dladdr()` can return the wrong image when looking up addresses in the presence of dynamic frameworks. The bug causes unsigned arithmetic underflow when calculating the instruction offset.

### Review checklist

### DIFF
--- a/DatadogCrashReporting/Sources/KSCrashIntegration/DatadogCrashReportFilter.swift
+++ b/DatadogCrashReporting/Sources/KSCrashIntegration/DatadogCrashReportFilter.swift
@@ -37,13 +37,19 @@ internal import KSCrashRecording
 /// If a report cannot be converted (invalid format, missing required fields),
 /// the original report is passed through and an error is provided to the completion handler.
 internal final class DatadogCrashReportFilter: NSObject, CrashReportFilter {
-    // Parse timestamp with fractional seconds support
-    // KSCrash timestamps use ISO8601 format with microsecond precision (e.g., "2025-10-22T14:14:12.007336Z")
+    /// Parse timestamp with fractional seconds support
+    /// KSCrash timestamps use ISO8601 format with microsecond precision (e.g., "2025-10-22T14:14:12.007336Z")
     let dateFormatter = {
         let formatter = ISO8601DateFormatter()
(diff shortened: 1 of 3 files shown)


gonzalezreal (approved):
🚀

DatadogCrashReporting/Sources/KSCrashIntegration/DatadogCrashReportFilter.swift
```diff
@@ -152,8 +152,9 @@ internal final class DatadogCrashReportFilter: NSObject, CrashReportFilter {
                     return String(format: "%-4ld ??? 0x%016llx 0x0 + 0", index, instructionAddr)
                 }
 
+                let offset = instructionAddr >= objectAddr ? instructionAddr - objectAddr : 0
```
gonzalezreal: Should we add telemetry if the address underflows?

DatadogCrashReporting/Sources/KSCrashIntegration/KSCrashBacktrace.swift
```diff
@@ -65,6 +65,7 @@ internal struct KSCrashBacktrace: BacktraceReporting {
             let libraryName = path.lastPathComponent
             let loadAddress = binaryImage.address
             let uuid = UUID(uuid: imageUUID.withMemoryRebound(to: uuid_t.self, capacity: 1) { $0.pointee })
+            let offset = address >= loadAddress ? UInt64(address) - loadAddress : 0
```
gonzalezreal: Same here.

DatadogCrashReporting/Sources/KSCrashIntegration/DatadogCrashReportFilter.swift
maxep: Yes, I have added error logs 👍

PR #2652, opened 2026-01-23
RUM-13855: Prepare the Profiling feature for release

simaoseica-dd:
### What and why?

This PR prepares the new Datadog Profiling feature for release in the next SDK version.
It includes improvements based on telemetry data and marks the Profiling API as experimental.

### How?
- Marks the public Profiling API as experimental.
- Removes telemetry reporting when the profiler status is sampled out or the app is prewarmed.
- Ignores app launch vitals that occur when the interval between process launch and SDK initialization is too large in non-prewarmed sessions.

### Review checklist

### DIFF
--- a/DatadogRUM/Sources/RUMMonitor/Scopes/RUMAppLaunchManager.swift
+++ b/DatadogRUM/Sources/RUMMonitor/Scopes/RUMAppLaunchManager.swift
@@ -284,6 +284,17 @@ private extension RUMVitalAppLaunchEvent.Vital.AppLaunchMetric {
 }
 
 private extension ProfilingContext {
+    /**
+     * Returns the profiling status reported for app launch.
+     *
+     * Returns:
+     *  - `.running` when the profiler is actively running, or when it was manually stopped or timed out.
+     *  - `.stopped` when the profiler was not started, was sampled out, or the app launch was prewarmed.
+     *  - `.error` when the profiler encountered an error while starting or it is in an unknown status.
+     *
+     * Note: the implementation currently maps `.stopped(.manual)` and `.stopped(.timeout)` to `.running`
+     * because those cases indicate profiling collected enough samples to be associated with the launch.
+     */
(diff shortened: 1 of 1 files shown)


DatadogRUM/Sources/RUMMonitor/Scopes/RUMAppLaunchManager.swift
```diff
@@ -282,10 +286,14 @@ private extension RUMVitalAppLaunchEvent.Vital.AppLaunchMetric {
...
-        case .running: .running
-        case .stopped: .stopped
-        case .error: .error
-        case .unknown: .error
+        case .running: return .running
+        case let .stopped(reason):
+            if reason == .manual || reason == .timeout {
+                return .running
+            }
+            return .stopped
```
maxep, line 305: If profiling is stopped we could still report it as running? that doesn't sounds correct, what is the rational?

simaoseica-dd: If the stopped reason is `manual` or `timeout` we might have the profile. This has to do with use cases such as multiple SDKs configured and waiting for a app launch profile, or too long TTIDs.
And the current FE implementation is to surface profiles only if the profiling status is `running`.

maxep: I think we should expand the `RUMVitalAppLaunchEvent.DD.Profiling.Status` to be more representative of the actual states on Mobile. In the meantime, I would suggest a comment explaining this mapping and potentially a ticket number to address it.

simaoseica-dd: I will propose a `.partial` status to the `DD.Profiling.Status`, since it is used by several SDKs. It will also be revisited since the profiler can be continuous in future iterations. 
As suggested, I added documentation with the mapping.

PR #2653, opened 2026-01-26
RUM-12403 Fixed how the active span is obtained in the TracingURLSessionHandler

arroz:
### What and why?

In situations where there is an active span, that span is not sampled, and `TraceContextInjection` is `.sampled` (the default), a `URLRequest` is not instrumented. However, we need to know the active span in the `interceptionDidComplete` method, otherwise we may create a span to trace the request without the proper parent, and without assigning the same priority and mechanism (in this context, dropping it).

The active span was being obtained on the wrong thread, inside the `interceptionDidStart` method that runs on the trace queue, and not in the thread where the request is created and instrumented. This could be too late (even after the request finished) and we would miss the active span. This was detected when investigating a test failure (`testSendingDroppedDistributedTraceWithParent_throughURLSessionInstrumentationAPI_noInjection`).

### How?

This patch moves the code that obtains the active span to the `TracingURLSessionHandler.modify` method and passes it down to `interceptionDidStart`. To do that, it was necessary to add a bit of infra-structure to allow `DatadogURLSessionHandler` instances to optionally return additional state from the `modify` method and receive it on the `interceptionDidStart`.

For now, we have to handle arrays of these things, but when RUM-13769 (internal) is done, this can be simplified.

### Review checklist

### DIFF
--- a/DatadogProfiling.podspec
+++ b/DatadogProfiling.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = "DatadogProfiling"
-  s.version      = "3.5.0"
+  s.version      = "3.5.1"
   s.summary      = "Official Datadog Profiling module of the Swift SDK."
   
   s.homepage     = "https://www.datadoghq.com"


DatadogTrace/Sources/Integrations/TracingURLSessionHandler.swift
```diff
@@ -101,29 +113,56 @@ internal struct TracingURLSessionHandler: DatadogURLSessionHandler {
...
+         method.
+
+         Note this method (interceptionDidStart) runs on a background thread, so obtaining the active
+         span here would be inaccurate, as this method can run much later, even after the request
+         is finished.
+
+         The comment inside the interceptionDidComplete(…) method explains how the registered active
+         span is used.
+         */
+        capturedStates.compactMap({ ($0 as? TracingURLSessionHandlerCapturedState)?.activeSpan }).first.map {
```
simaoseica-dd, line 166: double checking: Is this correct, where the first `capturedState` is always linked with this `interception`? Is it always aligned?

arroz: In practice, yes. Since we can only have one handler, there's always either zero or one element in that array. Even if we had two handlers of different types, only the `TracingURLSessionHandler` generates captured states of type `TracingURLSessionHandlerCapturedState`.

When we simplify this by removing support for multiple handlers, it's not going to be an array any more.

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
```diff
@@ -153,6 +153,13 @@ internal final class NetworkInstrumentationFeature: DatadogFeature {
 }
 
 extension NetworkInstrumentationFeature {
+    /// Helper structure that optionally contains a trace context and captured state, used to pass this
```
mariedm, line 156: nit: It would be helpful to add a TODO here and/or in `NetworkInstrumentationFeature` referring to the RUM-13769 ticket you mentioned in your PR description

arroz: Done. Thank you!

PR #2655, opened 2026-01-27
RUM-14055: Encoded and decoded response body sizes

barboraplasovska:
### What and why?

We want to support reporting `encodedBodySize` and `decodedBodySize` for network resources in RUM events. 

### How?

The `responseSize` parameter has been removed from `RUMMonitorProtocol._internal.addResourceMetrics()`. This parameter was redundant (duplicate of `responseBodySize.decoded`), never actually used by any SDK (React Native explicitly passes `nil`).

On React Native SDK, we need to remove the `responseSize: nil` line from calls to `addResourceMetrics`. No action needed for Flutter and Unity SDKs.

### Review checklist

### DIFF
--- a/DatadogRUM/Sources/RUMMonitor/Scopes/RUMResourceScope.swift
+++ b/DatadogRUM/Sources/RUMMonitor/Scopes/RUMResourceScope.swift
@@ -160,12 +160,32 @@ internal class RUMResourceScope: RUMScope {
         if let metrics = resourceMetrics {
             resourceStartTime = metrics.fetch.start
             resourceDuration = metrics.fetch.end.timeIntervalSince(metrics.fetch.start)
-            size = metrics.responseSize ?? command.size
+            size = metrics.responseBodySize?.decoded ?? command.size
         } else {
             resourceStartTime = resourceLoadingStartTime
             resourceDuration = command.time.timeIntervalSince(resourceLoadingStartTime)
             size = command.size
         }
+
+        let encodedBodySize: Int64? = {
+            guard let metrics = resourceMetrics else {
+                return nil
+            }
+            let requestEncoded = metrics.requestBodySize?.encoded ?? 0
+            let responseEncoded = metrics.responseBodySize?.encoded ?? 0
+            let total = requestEncoded + responseEncoded
+            return total > 0 ? total : nil
+        }()
+
+        let decodedBodySize: Int64? = {
+            guard let metrics = resourceMetrics else {
+                return nil
+            }
+            let requestDecoded = metrics.requestBodySize?.decoded ?? 0
+            let responseDecoded = metrics.responseBodySize?.decoded ?? 0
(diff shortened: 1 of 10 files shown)


DatadogRUM/Sources/RUMMonitor/Scopes/RUMResourceScope.swift
```diff
@@ -160,13 +160,33 @@ internal class RUMResourceScope: RUMScope {
         if let metrics = resourceMetrics {
             resourceStartTime = metrics.fetch.start
             resourceDuration = metrics.fetch.end.timeIntervalSince(metrics.fetch.start)
-            size = metrics.responseSize ?? command.size
+            size = metrics.responseBodySize?.decoded ?? command.size
```
mariedm: While it's not directly related to your PR, I recently made a change in `URLSessionRUMResourcesHandler` which I believe we should also apply here. If `metrics.responseBodySize?.decoded` equals `0`, it will not fallback to `command.size`. But we can fix it like this:

```suggestion
            let metricsSize = metrics.responseBodySize?.decoded ?? 0
            size = metricsSize > 0 ? metricsSize : command.size
```

Do you mind including it in your PR so we don't forget about it? If it's breaking tests, I'll create a Jira.

DatadogRUM/Sources/RUMMonitor/Scopes/RUMResourceScope.swift
```diff
@@ -160,13 +160,33 @@ internal class RUMResourceScope: RUMScope {
...
         } else {
             resourceStartTime = resourceLoadingStartTime
             resourceDuration = command.time.timeIntervalSince(resourceLoadingStartTime)
             size = command.size
         }
 
+        let encodedBodySize: Int64? = {
```
mariedm: Are you sure that `encodedBodySize = requestEncoded + responseEncoded` and `decodedBodySize = requestDecoded + responseDecoded`? 

Our [shared schema](https://github.com/DataDog/rum-events-format/blob/e39c1de766bf28ead5ceec2bc3da07c531488a6e/lib/cjs/generated/rum.d.ts#L616-L623) says the following, which is a bit unclear to me:
```
/**
         * Size in octet of the resource before removing any applied content encodings
         */
        readonly encoded_body_size?: number;
        /**
         * Size in octet of the resource after removing any applied encoding
         */
        readonly decoded_body_size?: number;
```

So I've looked it up. Multiple sources (W3C, MDN, etc.) say that the `encodedBodySize` corresponds to the **response** payload body before content-coding removal and `decodedBodySize` to **response** message body after content-coding removal. 

In which case I believe we would have the following: 
```
let encodedBodySize = metrics?.responseBodySize?.encoded
let decodedBodySize = metrics?.responseBodySize?.decoded
```

mariedm (approved):
LGMT! And thanks for the size fix fallback in `RUMResourceScope`. 👍

PR #2665, opened 2026-02-02
RUM-13731 Improve Logs attribute encoding error handling

mariedm:
## What and why?

Improves attribute encoding error handling to prevent entire events from being dropped when a single attribute fails to encode.

**Problem**: Currently, when any attribute fails to encode, the entire event is dropped with a generic error message. This makes it difficult for customers to debug which attribute is causing the failure.

**Root causes** (from production telemetry analysis):
1. **Objective-C bridge** - Customers use ObjC APIs like `addAttribute(forKey:value:)` which accepts `[String: Any]`. Our SDK automatically wraps values in `AnyEncodable`, which only fails at encoding time.
2. **Custom non-Encodable types** - Swift API requires `Encodable`, but customers can explicitly wrap non-encodable types using `AnyEncodable(value)` to bypass compile-time checks, e.g., passing closures/blocks, `NSObject`, custom classes, etc.

**Solution**: Implement graceful degradation:
- Skip malformed attributes individually instead of dropping the entire event
- Log clear SDK errors identifying which specific attribute failed and why
- Continue encoding remaining attributes
- Send events successfully with all valid attributes

## How?

1. **Created shared helper in DatadogInternal** (`AttributeEncoding.swift`):
   - `AttributeEncodingContext` enum: defines attribute context (`custom`, `userInfo`, `accountInfo`, `internal`) for clear error messages
   - `KeyedEncodingContainer.encodeAttribute()`: extension method that wraps individual attribute encoding in try-catch, logs errors with customer-facing attribute names, and continues execution

2. **Updated LogEventEncoder** to use `encodeAttribute()` for all attribute sources

**Note**: the same strategy will be later applied to the RUM and Trace features.

### Review checklist


DatadogInternal/Sources/Attributes/AttributeEncoding.swift
maxep, line 1: suggestion/ We could be move this to the `Attributes` folder instead

mariedm: Done 👍

PR #2666, opened 2026-02-02
RUM-14002: Expose public entities from `DatadogInternal`

simaoseica-dd:
### What and why?

Some public symbols defined in `DatadogInternal` are required by SDK features, but importing `DatadogInternal` directly is intended only for internal SDK usage.

This PR makes those symbols available to feature modules without requiring direct `DatadogInternal` imports.

### How?

It re-exports the required dependencies using `@_exported import`, allowing feature modules to access the needed types while keeping `DatadogInternal` as an internal implementation detail.

### Review checklist

### DIFF
--- a/DatadogCore/Sources/DatadogConfiguration.swift
+++ b/DatadogCore/Sources/DatadogConfiguration.swift
@@ -8,7 +8,6 @@ import Foundation
 import DatadogInternal
 
 @_exported import class DatadogInternal.CoreRegistry
-
 @_exported import class DatadogInternal.HTTPHeadersWriter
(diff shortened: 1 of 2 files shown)


DatadogCore/Sources/DatadogConfiguration.swift
```diff
@@ -7,6 +7,12 @@
 import Foundation
 import DatadogInternal
 
+@_exported import class DatadogInternal.CoreRegistry
+
```
mariedm: nit
```suggestion
```

mariedm (commented):
LGMT! 👍 But should we add a changelog entry?

PR #2668, opened 2026-02-03
Fix: Use String-based JSON array for GraphQL errors to support cross-platform SDKs

barboraplasovska:
### What and why?

This PR changes GraphQL error handling to use a String-based JSON array format instead of iOS-specific `Data` type, enabling cross-platform SDKs to easily pass GraphQL errors without platform-specific conversions.

### How?

The iOS SDK now extracts only the errors array from GraphQL responses and converts it to a JSON string, which is decoded directly to `[GraphQLResponseError]` without wrapper structs.

### Review checklist

### DIFF
--- a/DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
+++ b/DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
@@ -157,8 +157,8 @@ internal final class URLSessionRUMResourcesHandler: DatadogURLSessionHandler, RU
         }
 
         // Extract GraphQL errors from response if present
-        if let errorsData = extractGraphQLErrorsIfPresent(from: interception) {
-            combinedAttributes[CrossPlatformAttributes.graphqlErrors] = errorsData
+        if let errorsString = extractGraphQLErrorsIfPresent(from: interception) {
+            combinedAttributes[CrossPlatformAttributes.graphqlErrors] = errorsString
         }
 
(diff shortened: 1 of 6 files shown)


fuzzybinary (dismissed):
This LGTM and should work for Flutter.

cdn34dd (dismissed):

DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
```diff
@@ -191,7 +191,20 @@ internal final class URLSessionRUMResourcesHandler: DatadogURLSessionHandler, RU
             return nil
         }
 
-        return data
+        // Extract just the errors array from the response
+        do {
+            let json = try JSONSerialization.jsonObject(with: data) as? [String: Any]
```
simaoseica-dd: suggestion: Can you use `try JSONDecoder().decode(GraphQLResponse.self, from: data)` instead? 
It would be type safe and you would save an extra serialization. Meaning, you could remove the above ` try? JSONDecoder().decode(GraphQLResponseHasErrors.self, from: data)`

barboraplasovska: Done!

simaoseica-dd (approved):
👌

fuzzybinary (approved):

cdn34dd (approved):

PR #2671, opened 2026-02-04
Add GraphQL to telemetry usage

barboraplasovska:
### What and why?

This PR updates the `RUMDataModels` and then adds telemetry usage for GraphQL applied with the default sample rate of 15%.

### Review checklist

### DIFF
--- a/DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
+++ b/DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
@@ -77,14 +77,6 @@ internal final class URLSessionRUMResourcesHandler: DatadogURLSessionHandler, RU
     // MARK: - DatadogURLSessionHandler
 
     func modify(request: URLRequest, headerTypes: Set<DatadogInternal.TracingHeaderType>, networkContext: NetworkContext?) -> (URLRequest, TraceContext?, URLSessionHandlerCapturedState?) {
-        // Detect GraphQL requests by checking for GraphQL headers
-        let hasGraphQLHeaders = request.value(forHTTPHeaderField: GraphQLHeaders.operationName) != nil ||
-                                request.value(forHTTPHeaderField: GraphQLHeaders.operationType) != nil ||
-                                request.value(forHTTPHeaderField: GraphQLHeaders.variables) != nil ||
-                                request.value(forHTTPHeaderField: GraphQLHeaders.payload) != nil
-
-        let capturedState: URLSessionHandlerCapturedState? = hasGraphQLHeaders ? GraphQLRequestDetectedState() : nil
-
         let (modifiedRequest, traceContext, _) = distributedTracing?.modify(
             request: request,
             headerTypes: headerTypes,
@@ -92,6 +84,10 @@ internal final class URLSessionRUMResourcesHandler: DatadogURLSessionHandler, RU
             userId: networkContext?.userConfigurationContext?.id,
             accountId: networkContext?.accountConfigurationContext?.id
         ) ?? (request, nil, nil)
+
+        // Note: DistributedTracing.modify() currently returns nil for captured state.
+        // If this changes, we'll need to merge captured states instead of replacing.
+        let capturedState: URLSessionHandlerCapturedState? = modifiedRequest.hasGraphQLHeaders ? GraphQLRequestDetectedState() : nil
 
         return (modifiedRequest, traceContext, capturedState)
     }
@@ -102,7 +98,6 @@ internal final class URLSessionRUMResourcesHandler: DatadogURLSessionHandler, RU
 
         // Check if GraphQL was detected in the captured states
         if capturedStates.contains(where: { $0 is GraphQLRequestDetectedState }) {
-            // Track GraphQL request usage (sampled at 15% by default)
             telemetry.send(telemetry: .usage(.init(
                 event: .addGraphQLRequest,
                 sampleRate: UsageTelemetry.defaultSampleRate
@@ -349,5 +344,14 @@ private extension HTTPURLResponse {
         }
         let message = "\(statusCode) " + HTTPURLResponse.localizedString(forStatusCode: statusCode)
         return NSError(domain: "HTTPURLResponse", code: statusCode, userInfo: [NSLocalizedDescriptionKey: message])
+    }
+}
+
+private extension URLRequest {
+    var hasGraphQLHeaders: Bool {
+        value(forHTTPHeaderField: GraphQLHeaders.operationName) != nil ||
+        value(forHTTPHeaderField: GraphQLHeaders.operationType) != nil ||
+        value(forHTTPHeaderField: GraphQLHeaders.variables) != nil ||
+        value(forHTTPHeaderField: GraphQLHeaders.payload) != nil
     }
 }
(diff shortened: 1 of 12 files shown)


DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
```diff
@@ -70,19 +77,38 @@ internal final class URLSessionRUMResourcesHandler: DatadogURLSessionHandler, RU
...
+        return (modifiedRequest, traceContext, capturedState)
     }
 
     func interceptionDidStart(interception: DatadogInternal.URLSessionTaskInterception, capturedStates: [any URLSessionHandlerCapturedState]) {
         let url = interception.request.url?.absoluteString ?? "unknown_url"
         interception.register(origin: "rum")
 
+        // Check if GraphQL was detected in the captured states
+        if capturedStates.contains(where: { $0 is GraphQLRequestDetectedState }) {
+            // Track GraphQL request usage (sampled at 15% by default)
```
mariedm: The sample rate is mentioned in `UsageTelemetry.defaultSampleRate` and this comment could become stale, so I would remove it (at least the 15% bit).
```suggestion
```

DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
mariedm, line 1: Do we want to update `TracingURLSessionHandler` as well to support this when network tracking is enabled with Trace?

DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
```diff
@@ -70,19 +77,38 @@ internal final class URLSessionRUMResourcesHandler: DatadogURLSessionHandler, RU
     // MARK: - DatadogURLSessionHandler
 
     func modify(request: URLRequest, headerTypes: Set<DatadogInternal.TracingHeaderType>, networkContext: NetworkContext?) -> (URLRequest, TraceContext?, URLSessionHandlerCapturedState?) {
-        distributedTracing?.modify(
+        // Detect GraphQL requests by checking for GraphQL headers
+        let hasGraphQLHeaders = request.value(forHTTPHeaderField: GraphQLHeaders.operationName) != nil ||
```
mariedm: I believe this should be done after, using `modifiedRequest` instead of `request`. For now it doesn't change anything since `distributedTracing.modify()` doesn't do anything with GraphQL headers, but in case this changes in the future, it will be taken into account.

DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
```diff
@@ -70,19 +77,38 @@ internal final class URLSessionRUMResourcesHandler: DatadogURLSessionHandler, RU
...
+                                request.value(forHTTPHeaderField: GraphQLHeaders.operationType) != nil ||
+                                request.value(forHTTPHeaderField: GraphQLHeaders.variables) != nil ||
+                                request.value(forHTTPHeaderField: GraphQLHeaders.payload) != nil
+
+        let capturedState: URLSessionHandlerCapturedState? = hasGraphQLHeaders ? GraphQLRequestDetectedState() : nil
+
+        let (modifiedRequest, traceContext, _) = distributedTracing?.modify(
```
mariedm, line 83: It might be worth adding a small note/comment: `DistributedTracing.modify()` currently always returns `nil` for captured state. If this ever changes, we might need to merge captured states instead of replacing/overriding it.

DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
```diff
@@ -70,19 +77,38 @@ internal final class URLSessionRUMResourcesHandler: DatadogURLSessionHandler, RU
     // MARK: - DatadogURLSessionHandler
 
     func modify(request: URLRequest, headerTypes: Set<DatadogInternal.TracingHeaderType>, networkContext: NetworkContext?) -> (URLRequest, TraceContext?, URLSessionHandlerCapturedState?) {
-        distributedTracing?.modify(
+        // Detect GraphQL requests by checking for GraphQL headers
```
mariedm: nit: this logic could be extracted in an extension/helper function; e.g.:
``` 
 extension URLRequest {
      var hasGraphQLHeaders: Bool {
          value(forHTTPHeaderField: GraphQLHeaders.operationName) != nil ||
          value(forHTTPHeaderField: GraphQLHeaders.operationType) != nil ||
          value(forHTTPHeaderField: GraphQLHeaders.variables) != nil ||
          value(forHTTPHeaderField: GraphQLHeaders.payload) != nil
      }
  }
```

mariedm (commented):
Looks good overall! 👍  My question is about supporting this telemetry in Trace as well.

DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
barboraplasovska: yes! good idea, I added it!

DatadogTrace/Sources/Integrations/TracingURLSessionHandler.swift
```diff
@@ -342,3 +366,12 @@ private extension HTTPURLResponse {
         return NSError(domain: "HTTPURLResponse", code: statusCode, userInfo: [NSLocalizedDescriptionKey: message])
     }
 }
+
+private extension URLRequest {
```
mariedm: This extension can be moved to the internal module to avoid duplication since it's used in both Trace and RUM.

DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
```diff
@@ -34,13 +34,18 @@ internal struct DistributedTracing {
 }
 
 internal final class URLSessionRUMResourcesHandler: DatadogURLSessionHandler, RUMCommandPublisher {
+    /// Captured state indicating that a GraphQL request was detected
+    struct GraphQLRequestDetectedState: URLSessionHandlerCapturedState { }
```
mariedm: Maybe it makes sense to rename this structure to `RUMURLSessionHandlerCapturedState` to match the one in Tracing; what do you think? And we could even match the structure:
```
struct RUMURLSessionHandlerCapturedState: URLSessionHandlerCapturedState {
    let hasGraphQLHeaders: Bool
}
```

DatadogTrace/Sources/Integrations/TracingURLSessionHandler.swift
```diff
@@ -163,9 +177,19 @@ internal struct TracingURLSessionHandler: DatadogURLSessionHandler {
...
         // TODO: RUM-13769 This code can be simplified since we never use more than one handler simultaneously.
-        capturedStates.compactMap({ ($0 as? TracingURLSessionHandlerCapturedState)?.activeSpan }).first.map {
+        let capturedState = capturedStates.compactMap({ $0 as? TracingURLSessionHandlerCapturedState }).first
+
+        capturedState?.activeSpan.map {
             interception.register(activeSpanContext: $0.context)
         }
+
+        // Check if GraphQL was detected in the captured state
+        if capturedState?.hasGraphQLHeaders == true {
```
mariedm: You also need to check if the request is not already tracked by RUM (when both RUM and Trace are enabled) with `interception.origin != "rum"` to avoid sending duplicate telemetry. Might be worth adding a couple of unit tests for this.

mariedm (approved):
LGMT! 🚀

PR #2679, opened 2026-02-06
RUM-14354 Add watchOS support to TestUtilities

maxep:
### What and why?

This PR adds watchOS support to the TestUtilities module, which is a prerequisite for enabling watchOS support across other SDK modules.

### How?

- Added watchOS platform support to Package.swift
- Updated mocks to support watchOS platform with appropriate conditional compilation
- Extended UIKit helpers and mocks with watchOS compatibility

### Review checklist

### DIFF
--- a/TestUtilities/Sources/Mocks/DatadogRUM/RUMFeatureMocks.swift
+++ b/TestUtilities/Sources/Mocks/DatadogRUM/RUMFeatureMocks.swift
@@ -5,7 +5,9 @@
  */
 
 import Foundation
+#if !os(watchOS)
 import UIKit
+#endif
 import DatadogInternal
 
 @testable import DatadogRUM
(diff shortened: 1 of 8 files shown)


TestUtilities/Sources/Mocks/DatadogRUM/RUMFeatureMocks.swift
```diff
@@ -1310,7 +1342,7 @@ extension RUMUserActionScope {
         parent: RUMContextProvider,
         dependencies: RUMScopeDependencies = .mockAny(),
         name: String = .mockAny(),
-        actionType: RUMActionType = [.tap, .scroll, .swipe, .custom].randomElement()!,
+        actionType: RUMActionType = .custom,
```
barboraplasovska, line 1345: /question on watchOS we only support `.custom` action type ?

maxep: Yes, we won't support automatic tracking at this stage. So actions will be custom.

PR #2681, opened 2026-02-06
RUM-14354 Enable RUM on watchOS

maxep:
### What and why?

This PR enables RUM support on watchOS, bringing core RUM functionality to Apple Watch applications.

### How?

- Extended `RUMConfiguration` with watchOS-specific options (disabling UIKit/SwiftUI instrumentation, memory warnings, and vitals monitoring)
- Added conditional compilation for UIKit-specific features (UIViewController/UIApplication tracking, accessibility reader)
- Disabled automatic instrumentation of Views and Actions on watchOS
- Added watchOS compatibility for frame rate monitoring APIs

### Review checklist

### DIFF
--- a/DatadogRUM/Sources/Feature/RUMFeature.swift
+++ b/DatadogRUM/Sources/Feature/RUMFeature.swift
@@ -81,12 +81,21 @@ internal final class RUMFeature: DatadogRemoteFeature {
             watchdogTermination = monitor
         }
 
+        var renderLoopObserver: RenderLoopObserver? = nil
         var accessibilityReader: AccessibilityReading? = nil
-        if  #available(iOS 13.0, tvOS 13.0, *), configuration.collectAccessibility {
+
+        let firstFrameReader = FirstFrameReader(dateProvider: configuration.dateProvider, mediaTimeProvider: configuration.mediaTimeProvider)
+
+        #if !os(watchOS)
+        if #available(iOS 13.0, tvOS 13.0, *), configuration.collectAccessibility {
              accessibilityReader = AccessibilityReader(notificationCenter: configuration.notificationCenter)
         }
 
-        let firstFrameReader = FirstFrameReader(dateProvider: configuration.dateProvider, mediaTimeProvider: configuration.mediaTimeProvider)
+        renderLoopObserver = DisplayLinker(
+            notificationCenter: configuration.notificationCenter,
+            frameInfoProviderFactory: configuration.frameInfoProviderFactory
+        )
+        #endif
 
         let dependencies = RUMScopeDependencies(
             featureScope: featureScope,
@@ -119,10 +128,7 @@ internal final class RUMFeature: DatadogRemoteFeature {
                     return nil
                 }
             }(),
-            renderLoopObserver: DisplayLinker(
-                notificationCenter: configuration.notificationCenter,
-                frameInfoProviderFactory: configuration.frameInfoProviderFactory
-            ),
+            renderLoopObserver: renderLoopObserver,
             firstFrameReader: firstFrameReader,
             viewHitchesReaderFactory: {
                 configuration.trackSlowFrames
@@ -192,6 +198,7 @@ internal final class RUMFeature: DatadogRemoteFeature {
         firstFrameReader.publish(to: monitor)
         dependencies.renderLoopObserver?.register(firstFrameReader)
 
+        #if !os(watchOS)
         var memoryWarningMonitor: MemoryWarningMonitor?
         if configuration.trackMemoryWarnings {
             let memoryWarningReporter = MemoryWarningReporter()
@@ -201,7 +208,6 @@ internal final class RUMFeature: DatadogRemoteFeature {
             )
         }
 
-        #if !os(watchOS)
         self.instrumentation = RUMInstrumentation(
             featureScope: featureScope,
             uiKitRUMViewsPredicate: configuration.uiKitViewsPredicate,
--- a/DatadogRUM/Sources/RUMConfiguration.swift
+++ b/DatadogRUM/Sources/RUMConfiguration.swift
@@ -295,12 +295,14 @@ extension RUM {
         /// Default: `true`.
         public var trackAnonymousUser: Bool
 
+        #if !os(watchOS)
         /// Enables the collection of memory warnings.
         ///
         /// When enabled, all the memory warnings are reported as RUM Errors.
         ///
         /// Default: `true`.
         public var trackMemoryWarnings: Bool
+        #endif
 
         /// Enables the collection of slow frames (view hitches).
         ///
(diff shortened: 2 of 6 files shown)


DatadogRUM/Sources/Debugging/RUMDebugging.swift
```diff
@@ -28,12 +28,14 @@ private struct RUMDebugInfo {
 }
 
 internal class RUMDebugging {
+    #if !os(watchOS)
```
simaoseica-dd, line 33: nit: it would be simpler to wrap the whole class with `#if !os(watchOS)`. It is unusable in watchos.

DatadogRUM/Sources/RUMContext/AccessibilityReader.swift
```diff
@@ -264,3 +265,15 @@ internal final class AccessibilityReader: AccessibilityReading {
         return state
     }
 }
+#else
+// No-op implementation for watchOS where UIAccessibility APIs are unavailable
+internal final class AccessibilityReader: AccessibilityReading {
```
simaoseica-dd: suggestion: I would rather have a `NOPAccessibilityReading` to handle these cases. 
The parent entity would decide what to initialize based on the environment and this empty class would not go unnoticed.

DatadogInternal/Tests/NetworkInstrumentation/NetworkInstrumentationFeatureTests.swift
```diff
@@ -1706,6 +1707,7 @@ class NetworkInstrumentationFeatureTests: XCTestCase {
     }
 
     @available(iOS 13.0, tvOS 13.0, *)
```
mariedm:
```suggestion
```

DatadogInternal/Tests/NetworkInstrumentation/NetworkInstrumentationFeatureTests.swift
```diff
@@ -1706,6 +1707,7 @@ class NetworkInstrumentationFeatureTests: XCTestCase {
...
+    @available(iOS 13.0, tvOS 13.0, watchOS 8.0, *)
```
gonzalezreal: ⚠️ You should remove the duplicated availability attribute.

DatadogRUM/Sources/RUMVitals/RenderLoop/RenderLoopObserver.swift
```diff
@@ -112,3 +113,25 @@ extension DisplayLinker: RenderLoopObserver {
         renderLoopReaders.removeAll { $0 === renderLoopReader }
     }
 }
+#else
+
+/// No-op implementation for watchOS where CADisplayLink is unavailable
+internal class DisplayLinker {
```
simaoseica-dd: same here, a `NOPRenderLoopObserver` would be clear to us in the initialization point and this empty class would not go unnoticed.

DatadogRUM/Sources/RUMConfiguration.swift
```diff
@@ -525,6 +555,59 @@ extension RUM.Configuration {
         self.collectAccessibility = collectAccessibility
         self.featureFlags = featureFlags
     }
+    #else
+    public init(
```
mariedm, line 556: nit: you might have to run `make api-surface`

DatadogRUM/Sources/Debugging/RUMDebugging.swift
```diff
@@ -65,26 +67,26 @@ internal class RUMDebugging {
     }
 
     deinit {
-        DispatchQueue.main.async { [weak canvas] in
-            canvas?.removeFromSuperview()
-        }
```
gonzalezreal, line 70: As far as I understand, this is also removing any deinit logic for the canvas view for tvOS and visionOS as as well. Is that intentional?

DatadogInternal/Tests/NetworkInstrumentation/NetworkInstrumentationFeatureTests.swift
```diff
@@ -996,7 +996,7 @@ class NetworkInstrumentationFeatureTests: XCTestCase {
 
     func testGivenBothModesEnabled_whenPerTaskDelegate_itUsesCorrectTrackingMode() throws {
         // pre iOS 15 cannot set delegate per task
-        guard #available(iOS 15, tvOS 15, *) else {
+        guard #available(iOS 15, tvOS 15, watchOS 8, *) else {
```
mariedm, line 972: nit: sometimes `watchOS 8.0` is used, sometimes `watchOS 8`

DatadogRUM/Sources/Instrumentation/MemoryWarnings/MemoryWarningMonitor.swift
```diff
@@ -23,7 +23,11 @@ internal final class MemoryWarningMonitor {
 
     /// Starts monitoring memory warnings by subscribing to `UIApplication.didReceiveMemoryWarningNotification`.
     func start() {
+        #if os(watchOS)
+        consolePrint("Memory warnings instrumentation is not available on watchOS.", .warn)
```
mariedm, line 27: Should we set `trackMemoryWarnings` to `false` by default for the watchOS RUM config?

DatadogRUM/Sources/Feature/RUMFeature.swift
```diff
@@ -277,6 +283,17 @@ internal final class RUMFeature: DatadogRemoteFeature {
         )
 
         // Send configuration telemetry:
+        #if !os(watchOS)
```
simaoseica-dd, line 292: it should not be needed since the configuration already handles it, right?

DatadogRUM/Sources/Debugging/RUMDebugging.swift
maxep: Good catch, not intented!

DatadogRUM/Sources/Debugging/RUMDebugging.swift
maxep: I prefer to keep it so we don't change the calling sites and it keeps the endpoint to add a debug view support on other platforms.

DatadogRUM/Sources/Feature/RUMFeature.swift
maxep: It is because none of the predicate types exist on watchOS, so no `*Predicate` properties are visible in the configuration:
```swift
#if !os(watchOS)
@available(*, message: "This API is experimental and may change in future releases")
public var swiftUIViewsPredicate: SwiftUIRUMViewsPredicate?
#endif
```

DatadogRUM/Sources/RUMVitals/RenderLoop/RenderLoopObserver.swift
maxep: Good catch!

DatadogRUM/Sources/Instrumentation/MemoryWarnings/MemoryWarningMonitor.swift
maxep: Indeed, it could even be completely hidden.

DatadogRUM/Sources/RUMContext/AccessibilityReader.swift
maxep: Indeed, we can even remove the no-op class completely and only rely on the config 👍

DatadogRUM/Sources/RUMConfiguration.swift
maxep: `api-surface` builds for iOS, so there are no changes exposed :/

PR #2682, opened 2026-02-06
RUM-14354 Disable Profiling on watchOS

maxep:
### What and why?

This PR disables the Profiling module on watchOS as the underlying Mach APIs and profiling infrastructure are not supported on watchOS.

### How?

- Added conditional compilation guards throughout C++ profiling code to exclude watchOS
- Updated Swift profiling APIs to be unavailable on watchOS
- Added platform availability annotations to all profiling tests

### Review checklist

### DIFF
--- a/DatadogProfiling/Mach/ctor_profiler.cpp
+++ b/DatadogProfiling/Mach/ctor_profiler.cpp
@@ -6,8 +6,7 @@
 
 #include "ctor_profiler.h"
 
-#ifdef __APPLE__
-#if !TARGET_OS_WATCH
+#if defined(__APPLE__) && !TARGET_OS_WATCH
 
 #include "profile.h"
 #include "mach_sampling_profiler.h"
@@ -327,5 +326,4 @@ void ctor_profiler_start_testing(double sample_rate, bool is_prewarming, int64_t
 }
 #endif
 
-#endif // !TARGET_OS_WATCH
-#endif // __APPLE__
+#endif // __APPLE__ && !TARGET_OS_WATCH
--- a/DatadogProfiling/Sources/Profiling.swift
+++ b/DatadogProfiling/Sources/Profiling.swift
@@ -7,9 +7,8 @@
 import Foundation
 import DatadogInternal
 
-#if os(watchOS)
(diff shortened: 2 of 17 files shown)


DatadogProfiling/Sources/SDKMetrics/ConfigurationMetric.swift
```diff
@@ -5,19 +5,34 @@
...
+
+#if !os(watchOS)
 // swiftlint:disable duplicate_imports
 #if swift(>=6.0)
 internal import DatadogMachProfiler
 #else
 @_implementationOnly import DatadogMachProfiler
 #endif
 // swiftlint:enable duplicate_imports
+#endif
```
simaoseica-dd: Shouldn't the whole file be guarded? It is part of profiling and the duplication of constants below does not seem correct if not necessary

DatadogProfiling/Sources/Profiling.swift
```diff
@@ -7,6 +7,9 @@
 import Foundation
 import DatadogInternal
 
+#if os(watchOS)
+#warning("Datadog Profiling is not supported on watchOS.")
```
simaoseica-dd: question: Is this warning important? `SessionReplay` doesn't seem to need it

DatadogProfiling/Mach/ctor_profiler.cpp
```diff
@@ -7,6 +7,7 @@
 #include "ctor_profiler.h"
 
 #ifdef __APPLE__
```
simaoseica-dd: nit: All files could have a one line guard `#if defined(__APPLE__) && !TARGET_OS_WATCH`

DatadogProfiling/Sources/SDKMetrics/ConfigurationMetric.swift
maxep: Good catch! We don't actually need to wrap this as it is internal.

DatadogProfiling/Sources/Profiling.swift
maxep: Yes, indeed. This warning comes from WebView tracking, I can remove them.

PR #2690, opened 2026-02-12
[PANA-5912] Prepare screen change monitoring for Core Animation recording pipeline integration

gonzalezreal:
### What and why?

This PR is the first stacked PR in the integration of a new Core Animation based Session Replay recording pipeline.

It introduces foundational changes in screen-change monitoring that are required before wiring the new recording strategy:
- make `CALayerChangeset` `Sendable` to support safe cross-concurrency boundaries,
- expose a configurable `handler` on `ScreenChangeMonitor` to simplify integration with different coordinator/recorder implementations.

These changes are intended as scaffolding and do not yet switch recording behavior by themselves.

### How?

- Refactored `CALayerChangeset` to be `Sendable`.
- Added `handler` property on `ScreenChangeMonitor` and delegated delivery through it.
- Kept current monitoring semantics and delivery interval behavior intact.

### Review checklist

### DIFF
--- a/DatadogInternal/Sources/Attributes/Attributes.swift
+++ b/DatadogInternal/Sources/Attributes/Attributes.swift
@@ -145,8 +145,8 @@ public struct CrossPlatformAttributes {
     public static let graphqlVariables = "_dd.graphql.variables"
 
     /// Custom attribute passed when completing GraphQL RUM resources that contain errors in the response.
-    /// It sets the GraphQL errors from the response body as JSON data.
-    /// Expects `Data` value.
+    /// It sets the GraphQL errors array as a JSON string.
+    /// Expects `String` value containing a JSON array of errors.
     public static let graphqlErrors = "_dd.graphql.errors"
 
     /// Override the `source_type` of errors reported by the native crash handler. This is used on
--- a/DatadogProfiling/Mach/ctor_profiler.cpp
+++ b/DatadogProfiling/Mach/ctor_profiler.cpp
@@ -6,7 +6,7 @@
 
 #include "ctor_profiler.h"
 
-#ifdef __APPLE__
+#if defined(__APPLE__) && !TARGET_OS_WATCH
 
 #include "profile.h"
 #include "mach_sampling_profiler.h"
@@ -326,4 +326,4 @@ void ctor_profiler_start_testing(double sample_rate, bool is_prewarming, int64_t
 }
 #endif
 
-#endif // __APPLE__
+#endif // __APPLE__ && !TARGET_OS_WATCH
--- a/DatadogProfiling/Mach/dd_pprof.cpp
+++ b/DatadogProfiling/Mach/dd_pprof.cpp
@@ -5,6 +5,9 @@
  */
 
 #include "dd_pprof.h"
+
+#if defined(__APPLE__) && !TARGET_OS_WATCH
+
 #include "profile.h"
 #include "profile_pprof_packer.h"
 
@@ -61,3 +64,5 @@ double dd_pprof_get_end_timestamp_s(dd_pprof_t* profile) {
 }
 
 } // extern "C"
+
+#endif // __APPLE__ && !TARGET_OS_WATCH
--- a/DatadogProfiling/Mach/include/ctor_profiler.h
+++ b/DatadogProfiling/Mach/include/ctor_profiler.h
@@ -7,7 +7,10 @@
 #ifndef DD_PROFILER_CTOR_PROFILER_H_
 #define DD_PROFILER_CTOR_PROFILER_H_
 
-#include <stdint.h>
+#ifdef __APPLE__
+#include <TargetConditionals.h>
+#if !TARGET_OS_WATCH
+
 #include <stdbool.h>
 
 /**
@@ -78,8 +81,6 @@
  * 
  * All functions are thread-safe and can be called from any thread.
  */
-
-#ifdef __APPLE__
 
 // UserDefaults constants centralized for Profiling
 #define DD_PROFILING_USER_DEFAULTS_SUITE_NAME "com.datadoghq.ios-sdk.profiling"
@@ -230,5 +231,7 @@ void ctor_profiler_destroy(void);
 }
 #endif
 
+#endif // !TARGET_OS_WATCH
 #endif // __APPLE__
+
 #endif // DD_PROFILER_CTOR_PROFILER_H_
--- a/DatadogProfiling/Mach/include/ctor_profiler_testing.h
+++ b/DatadogProfiling/Mach/include/ctor_profiler_testing.h
@@ -7,7 +7,11 @@
 #ifndef DD_PROFILER_CTOR_PROFILER_TESTING_H_
 #define DD_PROFILER_CTOR_PROFILER_TESTING_H_
 
-#include "ctor_profiler.h"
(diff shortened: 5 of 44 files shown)


DatadogSessionReplay/Sources/Recorder/ScreenChangeMonitor/CALayerChangeSnapshot.swift
gonzalezreal, line 1: Refactored into `CALayerChangeset`

DatadogSessionReplay/Sources/Recorder/ScreenChangeMonitor/CALayerChangeAggregator.swift
```diff
@@ -109,14 +105,13 @@ internal final class CALayerChangeAggregator {
     }
 
     private func deliverPendingChanges(_ now: TimeInterval) {
-        let snapshot = CALayerChangeSnapshot(pendingChanges)
-            .removingDeallocatedLayers()
+        let changes = CALayerChangeset(pendingChanges)
```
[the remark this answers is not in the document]
gonzalezreal: We still need to trigger a capture with dead layer references. I removed this because it was premature optimization.

DatadogSessionReplay/Sources/Recorder/ScreenChangeMonitor/CALayerReference.swift
```diff
@@ -0,0 +1,61 @@
...
+import QuartzCore
+
+internal struct CALayerReference: @unchecked Sendable {
+    var identifier: ObjectIdentifier? {
+        layer.map(ObjectIdentifier.init)
+    }
+
+    var isAlive: Bool {
+        layer != nil
+    }
```
sethfowler-datadog, line 25: I think `isAlive` is probably not the best name for this method; even if `isAlive` returns true, the layer may in fact have already been freed, since no strong reference to `layer` remains once the method returns.

I'd suggest renaming this to something like `isKnownDead` or `isKnownDeallocated`, and inverting the sense of the check:

```suggestion
    var isKnownDead: Bool {
        layer == nil
    }
```

That makes it a bit clearer that you can trust the result if it says the layer is dead, but you _can't_ trust the result if it says the layer is alive, which matches the actual behavior a bit better.

DatadogSessionReplay/Sources/Recorder/ScreenChangeMonitor/CALayerChangeset.swift
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    mutating func merge(_ other: CALayerChangeset) {
+        changes.merge(other.changes) {
+            // ObjectIdentifier is only valid during the lifetime of an instance
+            guard $0.layer == $1.layer else {
+                return $1
+            }
+            return CALayerChange(layer: $1.layer, aspects: $0.aspects.union($1.aspects))
+        }
```
sethfowler-datadog, line 45: Two tricky points here:
1. If _both_ `$0` and `$1` are dead, `$0.layer == $1.layer`, right? (Because both will get `nil` when they try to convert their weak references into a strong reference.)
2. If one of `$0` or `$1` is dead, we return `$1`, but it could be `$1` that's dead and `$0` that's alive, right?

If you want to ensure that the key is removed from the dictionary if both `$0` and `$1` are dead, I suppose you may need to use a `for` loop, since it seems like the `Dictionary#merge()` API won't let you do that.

DatadogSessionReplay/Sources/Recorder/ScreenChangeMonitor/CALayerChangeset.swift
```diff
@@ -0,0 +1,74 @@
...
+
+    init(_ changes: [ObjectIdentifier: CALayerChange] = [:]) {
+        self.changes = changes
+    }
+
+    func aspects(for layer: CALayerReference) -> CALayerChange.Aspect.Set? {
+        guard
+            let identifier = layer.identifier,
+            let change = changes[identifier],
+            change.layer == layer // ObjectIdentifier is only valid during the lifetime of an instance
```
sethfowler-datadog, line 31: The layers on both sides of this expression may be `nil`. (After all, although we called `layer.identifier` above, which internally obtained a strong reference to the layer, that strong reference has been released at this point, and another thread could have freed the layer by the time this expression is evaluated.) That'd result in the guard succeeding, even though the underlying `CALayer` has been deinit'd, which might not be what was intended.

DatadogSessionReplay/Sources/Recorder/ScreenChangeMonitor/CALayerReference.swift
```diff
@@ -0,0 +1,61 @@
...
+// A weak, `Sendable` identity wrapper for `CALayer`.
+//
+// Use this for liveness checks and identity comparison without retaining the layer.
+// No direct access is exposed; use `resolve()` on the main actor to work with the
+// layer if it's still alive.
+
+#if os(iOS)
+import QuartzCore
+
+internal struct CALayerReference: @unchecked Sendable {
```
sethfowler-datadog, line 18: Is this `@unchecked Sendable` purely because of the weak var? If so, I guess [once the next version of Swift is released](https://forums.swift.org/t/accepted-se-0481-weak-let/79895), you'll be able to write `weak let layer: CALayer?` and get rid of the `@unchecked`.

sethfowler-datadog (commented):
LGTM! There may be some opportunities to deal with weak vars a bit more safely in this code; some suggestions below.

DatadogSessionReplay/Sources/Recorder/ScreenChangeMonitor/CALayerReference.swift
gonzalezreal: Neat. I will address this in a future PR. Thanks!

DatadogSessionReplay/Sources/Recorder/ScreenChangeMonitor/CALayerReference.swift
gonzalezreal: That's right. With that, maybe we could get rid of the whole `CALayerReference` type, which basically encapsulates a read-only weak reference to a Core Animation layer.

DatadogSessionReplay/Sources/Recorder/ScreenChangeMonitor/CALayerChangeset.swift
gonzalezreal: Totally fair. The good news is that we don't need this method. At some point I used it while prototyping but discarded it and forgot to remove it.

DatadogSessionReplay/Sources/Recorder/ScreenChangeMonitor/CALayerChangeset.swift
gonzalezreal: Good catch. I'll tighten this so that dead references don't pass this guard.

DatadogSessionReplay/Sources/Recorder/ScreenChangeMonitor/CALayerReference.swift
gonzalezreal: This property is not needed and I will remove it. I added the `resolved()` method for that very reason.

PR #2698, opened 2026-02-16
[PANA-5986] Add layer snapshot capture and optimization pipeline for Session Replay

gonzalezreal:
### What and why?

This PR adds the next layer-tree recording building blocks for Session Replay, as part of the integration of the new Core Animation based recording pipeline.

It introduces a full snapshot optimization path for captured layers:

- capture immutable layer snapshots
- remove invisible branches
- flatten the tree into render order
- cull fully obscured snapshots

It also adds stable replay ID generation for `CALayer`.

> [!NOTE]  
> `WKWebView` semantics capture will be addressed in a future PR.

### How?

- Added `CALayer` replay ID support with wraparound and test generator support, similar to existing node ids.
- Added `LayerSnapshot` model and capture logic.
- Added optimization passes:
  - `LayerSnapshot+Visibility.swift`
  - `LayerSnapshot+Flattening.swift`
  - `LayerSnapshot+Culling.swift` (with `OpaqueFrameIndex` spatial index).
- Wired the optimization pipeline into `LayerRecorder.record(...)`:
  - `removingInvisible() -> flattened() -> removingObscured(in:)`.
- Simplified and hardened screen-change layer reference handling:
  - `CALayerChangeset.aspects(for:)` now validates by identifier.
- Removed unused methods from `CALayerChangeset` and `CALayerReference`.
- Added internal overview docs across `LayerTreeRecorder` source files.
- Added dedicated unit tests for:
  - replay ID generation
  - snapshot capture
  - visibility pruning
  - flattening
  - culling

### Review checklist

### DIFF
--- a/DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Culling.swift
+++ b/DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Culling.swift
@@ -57,7 +57,6 @@ extension Array where Element == LayerSnapshot {
             let visibleFrame = snapshot.frame.intersection(snapshot.clipRect)
 
             guard !visibleFrame.isNull, !visibleFrame.isEmpty else {
-                result.append(snapshot)
                 continue
             }
 
@@ -104,7 +103,7 @@ private struct OpaqueFrameIndex {
         frames.append(visibleFrame)
 
         let bandRange = bandRange(for: visibleFrame)
-        let coveredBands = bandRange.upperBound - bandRange.lowerBound + 1
+        let coveredBands = bandRange.count
 
         if coveredBands > maxBandsPerFrame {
             globalFrameIndices.append(frameIndex)
--- a/DatadogSessionReplay/Tests/LayerTreeRecorder/LayerSnapshotCullingTests.swift
+++ b/DatadogSessionReplay/Tests/LayerTreeRecorder/LayerSnapshotCullingTests.swift
@@ -152,6 +152,26 @@ extension LayerSnapshotTests {
 
     @available(iOS 13.0, tvOS 13.0, *)
     @Test
+    func removingObscuredDropsSnapshotsWithoutVisibleFrame() {
+        // given
+        let snapshot = Fixtures.snapshot(
+            replayID: 1,
+            frame: CGRect(x: 200, y: 0, width: 100, height: 100),
+            clipRect: Fixtures.viewportRect,
+            opacity: 1.0,
+            backgroundColor: CGColor(red: 1, green: 0, blue: 0, alpha: 1.0),
+            hasContents: true
+        )
+
+        // when
+        let result = [snapshot].removingObscured(in: Fixtures.viewportRect)
+
+        // then
+        #expect(result.isEmpty)
+    }
+
+    @available(iOS 13.0, tvOS 13.0, *)
+    @Test
     func removingObscuredUsesGlobalIndexBucketForLargeOpaqueFrames() {
         // given
         let viewportRect = CGRect(x: 0, y: 0, width: 100, height: 1_000)


DatadogSessionReplay/Sources/LayerTreeRecorder/LayerRecorder.swift
```diff
@@ -31,10 +40,34 @@ internal actor LayerRecorder: LayerRecording {
...
+                .removingObscured(in: snapshot.clipRect),
+            !targetSnapshots.isEmpty
+        else {
+            // There is nothing visible yet
+            return
+        }
+
+        // Pending stages:
+        // - Render layer bitmaps
+        // - Process layer tree snapshots
```
gonzalezreal, line 59: This will be addressed in future PRs

DatadogSessionReplay/Sources/Recorder/ScreenChangeMonitor/CALayerChangeset.swift
```diff
@@ -28,30 +28,12 @@ internal struct CALayerChangeset: Sendable, Equatable {
...
-        }
-    }
-
-    mutating func removeDeallocatedLayers() {
-        changes = changes.filter(\.value.layer.isAlive)
-    }
-
-    mutating func removeAll() {
-        changes.removeAll()
-    }
```
gonzalezreal, line 54: Removing these are they were unused, addressing this feedback:
- https://github.com/DataDog/dd-sdk-ios/pull/2690#discussion_r2811939510
- https://github.com/DataDog/dd-sdk-ios/pull/2690#discussion_r2812640391

DatadogSessionReplay/Sources/Recorder/ScreenChangeMonitor/CALayerChangeset.swift
```diff
@@ -28,30 +28,12 @@ internal struct CALayerChangeset: Sendable, Equatable {
         guard
             let identifier = layer.identifier,
             let change = changes[identifier],
-            change.layer == layer // ObjectIdentifier is only valid during the lifetime of an instance
+            change.layer.identifier == identifier // ObjectIdentifier is only valid during the lifetime of an instance
```
gonzalezreal, line 31: Addresses this feedback:
- https://github.com/DataDog/dd-sdk-ios/pull/2690#discussion_r2812668108

DatadogSessionReplay/Sources/Recorder/ScreenChangeMonitor/CALayerReference.swift
```diff
@@ -20,10 +20,6 @@ internal struct CALayerReference: @unchecked Sendable {
         layer.map(ObjectIdentifier.init)
     }
 
-    var isAlive: Bool {
-        layer != nil
-    }
-
```
gonzalezreal, line 26: Unused. Prefer the `resolved()` method. Addresses this feedback: https://github.com/DataDog/dd-sdk-ios/pull/2690#discussion_r2811939510

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Culling.swift
```diff
@@ -0,0 +1,156 @@
...
+
+#if os(iOS)
+import Foundation
+import CoreGraphics
+
+@available(iOS 13.0, tvOS 13.0, *)
+extension LayerSnapshot {
+    var isOpaque: Bool {
+        backgroundColor?.alpha == 1.0 && opacity == 1.0 &&
+        !hasMask && cornerRadius == 0 && isAxisAligned
```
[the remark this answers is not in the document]
gonzalezreal: Addressed in [05feda7](https://github.com/DataDog/dd-sdk-ios/pull/2698/commits/05feda72cfcf1f5865cc69517bf047f63a5e081b)

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot.swift
```diff
@@ -0,0 +1,166 @@
...
+    private init(
+        from layer: CALayer,
+        in rootLayer: CALayer,
+        pathComponents: [String],
+        clipRect: CGRect
+    ) {
+        let frame = layer.convert(layer.bounds, to: rootLayer)
+
+        let nextClipRect: CGRect
+        if layer.masksToBounds {
```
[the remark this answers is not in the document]
gonzalezreal: Good catch, and I agree `layer.mask` is not reflected in `clipRect` propagation today.

For this PR, I addressed the culling correctness risk via a simpler path: `hasMask` is now resolved through ancestry during snapshot capture, so any masked branch is treated as masked for `isOpaque` and can’t occlude background layers incorrectly.

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Culling.swift
```diff
@@ -0,0 +1,156 @@
...
+    ) -> [LayerSnapshot] {
+        var result: [LayerSnapshot] = []
+        result.reserveCapacity(count)
+
+        // Process front-to-back
+        for snapshot in reversed() {
+            let visibleFrame = snapshot.frame.intersection(snapshot.clipRect)
+
+            guard !visibleFrame.isNull, !visibleFrame.isEmpty else {
+                result.append(snapshot)
```
sethfowler-datadog: Would it make sense to treat these cases as "obscured"? In other words, should we _not_ append here?

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Culling.swift
```diff
@@ -0,0 +1,156 @@
...
+    mutating func insert(_ frame: CGRect) {
+        guard let visibleFrame = visibleInViewport(frame) else {
+            return
+        }
+
+        let frameIndex = frames.count
+        frames.append(visibleFrame)
+
+        let bandRange = bandRange(for: visibleFrame)
+        let coveredBands = bandRange.upperBound - bandRange.lowerBound + 1
```
sethfowler-datadog: Nit: I _think_ this works:

```suggestion
        let coveredBands = bandRange.count
```

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Culling.swift
```diff
@@ -0,0 +1,156 @@
...
+        // Check large frames first since they can occlude any viewport region.
+        for frameIndex in globalFrameIndices where frames[frameIndex].contains(visibleFrame) {
+            return true
+        }
+
+        for band in bandRange(for: visibleFrame) {
+            for frameIndex in bandFrameIndices[band] where frames[frameIndex].contains(visibleFrame) {
+                return true
+            }
+        }
```
sethfowler-datadog, line 132: There is probably an opportunity to avoid comparing repeatedly against frames that appear in multiple bands, and to generally reduce the number of comparisons, by switching to a BSP-style data structure (i.e. a quadtree), or by at least applying the general principles they use (i.e. comparing against bounding hyperplanes). However, depending on how many frames you are dealing with here in practice, the complexity may not be warranted.

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Culling.swift
```diff
@@ -0,0 +1,156 @@
...
+            globalFrameIndices.append(frameIndex)
+            return
+        }
+
+        for band in bandRange {
+            bandFrameIndices[band].append(frameIndex)
+        }
+    }
+
+    func contains(_ frame: CGRect) -> Bool {
```
sethfowler-datadog, line 118: This method actually checks if any _single_ frame in the `OpaqueFrameIndex` occludes `frame`, right? But if half of `frame` is occluded by OtherFrameA, and the other half is occluded by OtherFrameB, `contains()` will return `false`.

It's considerably more expensive to perform a more accurate check, so I'm not sure if improving on this is worth it, at least without more data about usage. Offhand, the simplest approach I can think of would be a stencil-buffer-like method where you "draw" the AABBs of the frames into a buffer (possibly at a reduced resolution, accepting an accuracy tradeoff) and check for occlusion by seeing if the buffer changed.

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot.swift
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    @MainActor
+    private init(
+        from layer: CALayer,
+        in rootLayer: CALayer,
+        pathComponents: [String],
+        clipRect: CGRect,
+        parentOpacity: Float,
+        parentHasMask: Bool
```
sethfowler-datadog, line 77: Nit: perhaps this would be more accurately named `ancestorHasMask`, since it's propagated transitively.

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot.swift
```diff
@@ -0,0 +1,178 @@
...
+            NSStringFromClass(type(of: self)),
+            delegate.map {
+                NSStringFromClass(type(of: $0))
+            }
+        ].compactMap(\.self).joined(separator: "-")
+    }
+
+    fileprivate func pathComponent(_ index: Int) -> String {
+        "\(componentType)#\(index)"
+    }
```
sethfowler-datadog, line 155: At some point it might be worth reworking things to avoid serializing the same (layer type, delegate type, index) tuples to a string over and over again, but it seems like an optimization that could happen in a future PR.

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Visibility.swift
```diff
@@ -0,0 +1,59 @@
...
+// preserving nodes that can still contribute visible content through themselves or
+// descendants.
+
+#if os(iOS)
+import Foundation
+import CoreGraphics
+
+@available(iOS 13.0, tvOS 13.0, *)
+extension LayerSnapshot {
+    var isVisible: Bool {
```
sethfowler-datadog, line 21: It seems like in `removingInvisible()`, we interpret "invisible" to mean "!isVisible && (rendersContent || hasVisibleDescendants)". The fact that an invisible layer is not simply "!isVisible" suggests to me that we might be able to improve on this naming, since naively, I'd expect that "invisible" means "not visible". Maybe `isVisible()` should be something more like `isPotentiallyVisible()`, since the conditions it checks are necessary but not sufficient?

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Culling.swift
gonzalezreal: Agreed there’s room for a more advanced spatial structure. For now I’m intentionally keeping this pass simple and cheap; I’d prefer to revisit a quadtree/BSP approach only if profiling shows this path is hot enough to justify the added complexity.

For reference, this is what we are doing in the current recording pipeline:
https://github.com/DataDog/dd-sdk-ios/blob/61c3068fcd360a5fa434959ec60287ea47ca8a0f/DatadogSessionReplay/Sources/Processor/Flattening/NodesFlattener.swift

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Culling.swift
gonzalezreal: Good call. These cases are effectively non-visible in the viewport, so keeping them here is not useful for culling and may add noise downstream. I’m aligned with treating them as non-contributing and not appending

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Culling.swift
gonzalezreal: 🤦🏽 Agreed, this is cleaner.

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Culling.swift
gonzalezreal: Correct. Current `contains()` only detects full occlusion by a single opaque frame. That is an intentional conservative tradeoff: it may miss some culling opportunities (false negatives), but it avoids incorrect removals (false positives).

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot.swift
gonzalezreal: Good suggestion. The value is propagated transitively, so ancestorHasMask would be more explicit. I can rename in a follow-up to keep this PR focused

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot.swift
gonzalezreal: Agreed. This looks like a reasonable future optimization. We could leverage the Objective-C runtime to cache the `componentType` value as an associated value of the `CALayer` instance.

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Visibility.swift
gonzalezreal: I agree the naming can be clearer. The check is “potentially visible” (necessary, not sufficient). I’d keep behavior as-is for now and do a focused naming pass in follow-up to avoid broad churn in this PR

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Culling.swift
sethfowler-datadog: Nice; seems like the approach in this PR will be a clear win over the existing algorithm!

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Culling.swift
gonzalezreal: Done in [e31e980](https://github.com/DataDog/dd-sdk-ios/pull/2698/commits/e31e9801d471ff9361a3af5390bdb33530b94ab9)

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Culling.swift
gonzalezreal: Done in [e31e980](https://github.com/DataDog/dd-sdk-ios/pull/2698/commits/e31e9801d471ff9361a3af5390bdb33530b94ab9)

PR #2699, opened 2026-02-16
RUM-14563 Add header capture configuration API and HeaderProcessor

barboraplasovska:
### What and why?

This is the first PR for Network Header Capture. It adds the public configuration API (`TrackResourceHeaders`, `HeaderCaptureRule`) on `URLSessionTracking` to let users opt into capturing HTTP request and response headers on RUM resources. It also introduces `HeaderProcessor`, which resolves the capture list from configuration rules, applies security filtering (regex-based sensitive header detection, reserved iOS headers exclusion), and enforces size limits. RUM models are regenerated to include `request.headers` and `response.headers` on `RUMResourceEvent`.

Follow-up PR will wire the processor into the resource interception pipeline.

### Review checklist
- [ ] Feature or bugfix MUST have appropriate tests (unit, integration)
- [ ] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes
- [ ] Add Objective-C interface for public APIs - see our [guidelines](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3157787243/RFC+-+Modular+Objective-C+Interface#Recommended-solution) (internal)
- [ ] Run `make api-surface` when adding new APIs

PR #2700, opened 2026-02-18
RUM-14563 Wire header capture into resource interception pipeline

barboraplasovska:
### What and why?

This is the second PR for Network Header Capture (following #2699). It wires the `HeaderProcessor` into the resource interception pipeline so that captured headers flow through handler → command → scope → RUM event, and appear as `resource.request.headers` and `resource.response.headers` in the Datadog dashboard.

### Review checklist

### DIFF
--- a/DatadogInternal/Sources/Attributes/Attributes.swift
+++ b/DatadogInternal/Sources/Attributes/Attributes.swift
@@ -163,6 +163,16 @@ public struct CrossPlatformAttributes {
     /// Custom value for Interaction To Next view.
     /// For Flutter this is the amount of time between an action occurring and the First Build Complete occurring on the next view.
     public static let customINVValue: String = "_dd.view.custom_inv_value"
+
+    /// HTTP headers captured from the resource request, serialized as a JSON string.
+    /// Used to transport request headers through the RUM command pipeline.
+    /// Expects `String` value containing a JSON dictionary of `[String: String]`.
+    public static let resourceRequestHeaders = "_dd.resource.request_headers"
+
+    /// HTTP headers captured from the resource response, serialized as a JSON string.
+    /// Used to transport response headers through the RUM command pipeline.
+    /// Expects `String` value containing a JSON dictionary of `[String: String]`.
+    public static let resourceResponseHeaders = "_dd.resource.response_headers"
 }
 
 /// HTTP header names used to pass GraphQL metadata from the application to the SDK.
--- a/DatadogInternal/Sources/Attributes/Attributes.swift
+++ b/DatadogInternal/Sources/Attributes/Attributes.swift
@@ -164,12 +164,12 @@ public struct CrossPlatformAttributes {
     /// For Flutter this is the amount of time between an action occurring and the First Build Complete occurring on the next view.
     public static let customINVValue: String = "_dd.view.custom_inv_value"
 
-    /// Request headers passed from CP SDK or captured from native URLSession interception.
+    /// Request headers passed from Cross-Platform SDK or captured from native URLSession interception.
     /// Used in RUM resources to transport request headers through the RUM command pipeline.
     /// Expects `[String: String]` value containing header keys and values.
     public static let requestHeaders = "_dd.request_headers"
 
-    /// Response headers passed from CP SDK or captured from native URLSession interception.
+    /// Response headers passed from Cross-Platform SDK or captured from native URLSession interception.
     /// Used in RUM resources to transport response headers through the RUM command pipeline.
     /// Expects `[String: String]` value containing header keys and values.
     public static let responseHeaders = "_dd.response_headers"
--- a/DatadogInternal/Sources/Attributes/Attributes.swift
+++ b/DatadogInternal/Sources/Attributes/Attributes.swift
@@ -164,15 +164,15 @@ public struct CrossPlatformAttributes {
     /// For Flutter this is the amount of time between an action occurring and the First Build Complete occurring on the next view.
     public static let customINVValue: String = "_dd.view.custom_inv_value"
 
-    /// HTTP headers captured from the resource request, serialized as a JSON string.
-    /// Used to transport request headers through the RUM command pipeline.
-    /// Expects `String` value containing a JSON dictionary of `[String: String]`.
-    public static let resourceRequestHeaders = "_dd.resource.request_headers"
+    /// Request headers passed from CP SDK or captured from native URLSession interception.
+    /// Used in RUM resources to transport request headers through the RUM command pipeline.
+    /// Expects `[String: String]` value containing header keys and values.
+    public static let requestHeaders = "_dd.request_headers"
 
-    /// HTTP headers captured from the resource response, serialized as a JSON string.
-    /// Used to transport response headers through the RUM command pipeline.
-    /// Expects `String` value containing a JSON dictionary of `[String: String]`.
-    public static let resourceResponseHeaders = "_dd.resource.response_headers"
+    /// Response headers passed from CP SDK or captured from native URLSession interception.
+    /// Used in RUM resources to transport response headers through the RUM command pipeline.
+    /// Expects `[String: String]` value containing header keys and values.
+    public static let responseHeaders = "_dd.response_headers"
 }
 
 /// HTTP header names used to pass GraphQL metadata from the application to the SDK.
--- a/DatadogRUM/Sources/Instrumentation/Resources/HeaderCapture/HeaderProcessor.swift
+++ b/DatadogRUM/Sources/Instrumentation/Resources/HeaderCapture/HeaderProcessor.swift
@@ -85,11 +85,9 @@ internal struct HeaderProcessor {
 
     // MARK: - Private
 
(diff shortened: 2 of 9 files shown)


DatadogInternal/Sources/Attributes/Attributes.swift
```diff
@@ -163,6 +163,16 @@ public struct CrossPlatformAttributes {
     /// Custom value for Interaction To Next view.
     /// For Flutter this is the amount of time between an action occurring and the First Build Complete occurring on the next view.
     public static let customINVValue: String = "_dd.view.custom_inv_value"
+
+    /// Request headers passed from CP SDK or captured from native URLSession interception.
```
mariedm:
```suggestion
    /// Request headers passed from Cross-Platform SDK or captured from native URLSession interception.
```

DatadogInternal/Sources/Attributes/Attributes.swift
```diff
@@ -163,6 +163,16 @@ public struct CrossPlatformAttributes {
...
+    /// Used in RUM resources to transport request headers through the RUM command pipeline.
+    /// Expects `[String: String]` value containing header keys and values.
+    public static let requestHeaders = "_dd.request_headers"
+
+    /// Response headers passed from CP SDK or captured from native URLSession interception.
```
mariedm:
```suggestion
    /// Response headers passed from Cross-Platform SDK or captured from native URLSession interception.
```

DatadogRUM/Sources/RUMMonitor/Scopes/RUMResourceScope.swift
```diff
@@ -171,12 +175,21 @@ internal class RUMResourceScope: RUMScope {
...
             .init(
                 decodedBodySize: size.decoded,
-                encodedBodySize: size.encoded
+                encodedBodySize: size.encoded,
+                headers: requestHeadersObj
             )
         }
+        if request == nil, let requestHeadersObj {
+            request = .init(headers: requestHeadersObj)
+        }
```
mariedm: suggestion: for easier reading, and this way we don't need 2 init paths

```suggestion
    let request: RUMResourceEvent.Resource.Request? = {
        let hasBodySize = resourceMetrics?.requestBodySize != nil
        let hasHeaders = requestHeadersObj != nil
    
        guard hasBodySize || hasHeaders else { return nil }
    
        return .init(
            decodedBodySize: resourceMetrics?.requestBodySize?.decoded,
            encodedBodySize: resourceMetrics?.requestBodySize?.encoded,
            headers: requestHeadersObj
        )
}()
```

DatadogRUM/Tests/RUMMonitor/Scopes/RUMResourceScopeTests.swift
```diff
@@ -1792,4 +1792,113 @@ class RUMResourceScopeTests: XCTestCase {
...
+                writer: writer
+            )
+        )
+
+        // Then
+        let event = try XCTUnwrap(writer.events(ofType: RUMResourceEvent.self).first)
+        let request = try XCTUnwrap(event.resource.request)
+        let headers = try XCTUnwrap(request.headers)
+        XCTAssertEqual(headers.headersInfo["content-type"], "application/json")
+        XCTAssertEqual(headers.headersInfo["cache-control"], "no-cache")
```
mariedm, line 1832: very nit: you could also extend this test (or create a separate one) that also asserts on `encodedBodySize` and `decodedBodySize` since requests can have both the sizes and the headers, or only one or the other.

mariedm (dismissed):
LGMT! 👍 Only nitpicking comments.

DatadogRUM/Sources/Instrumentation/Resources/HeaderCapture/HeaderProcessor.swift
```diff
@@ -101,71 +99,91 @@ internal struct HeaderProcessor {
             rules = customRules
         }
 
+        var requestHeaders: [String] = []
+        var responseHeaders: [String] = []
+        var seenRequest = Set<String>()
+        var seenResponse = Set<String>()
```
mariedm:
```suggestion
        var seenRequests = Set<String>()
        var seenResponses = Set<String>()
```

mariedm (dismissed):
Thanks! And good idea to prioritize default headers 👌

DatadogRUM/Sources/Instrumentation/Resources/HeaderCapture/HeaderProcessor.swift
```diff
@@ -101,71 +99,91 @@ internal struct HeaderProcessor {
...
-                requestHeaders.formUnion(lowercased)
-                responseHeaders.formUnion(lowercased)
+                for name in names {
+                    let lowercased = name.lowercased()
+                    if seenRequest.insert(lowercased).inserted {
+                        requestHeaders.append(lowercased)
+                    }
+                    if seenResponse.insert(lowercased).inserted {
+                        responseHeaders.append(lowercased)
+                    }
```
simaoseica-dd: Is this correct having the same header for requests and headers?

DatadogRUM/Sources/Instrumentation/Resources/HeaderCapture/HeaderProcessor.swift
```diff
@@ -101,71 +99,91 @@ internal struct HeaderProcessor {
...
+                }
             }
         }
 
         return (request: requestHeaders, response: responseHeaders)
     }
 
     /// Filters request headers.
-    private func filterRequestHeaders(_ headers: [String: String]?, capturing: Set<String>) -> [String: String] {
+    private func filterRequestHeaders(_ headers: [String: String]?, capturing: [String]) -> [String: String] {
```
simaoseica-dd: What is `capturing`?
`private func filterRequest(headers: [String: String]?, capturedHeaders: [String]) -> [String: String]`

DatadogRUM/Sources/Instrumentation/Resources/HeaderCapture/HeaderProcessor.swift
```diff
@@ -101,71 +99,91 @@ internal struct HeaderProcessor {
...
         guard let headers else {
             return [:]
         }
-        return filterHeaders(headers.map { ($0.key, $0.value) }, capturing: capturing, excluded: Self.reservedRequestHeaders)
+        // Build a case-insensitive lookup from the raw headers
+        let normalized = Dictionary(headers.map { ($0.key.lowercased(), ($0.key, $0.value)) }, uniquingKeysWith: { _, last in last })
```
simaoseica-dd, line 138: Is it ok to replace the previous key if the keys are colliding because of the cases?

simaoseica-dd (dismissed):
Left some minor questions

DatadogRUM/Sources/Instrumentation/Resources/HeaderCapture/HeaderProcessor.swift
barboraplasovska: Yes this is by design. For the first version, `.matchHeaders` doesn't distinguish between request and response, the header name is matched for both.

DatadogRUM/Sources/Instrumentation/Resources/HeaderCapture/HeaderProcessor.swift
barboraplasovska: Renamed to `capturedHeaders` for clarity!

DatadogRUM/Sources/Instrumentation/Resources/HeaderCapture/HeaderProcessor.swift
barboraplasovska: Yes. Per RFC 7230, HTTP header names are case-insensitive so when there's a se collision we keep the last value which matches the standard HTTP behaviour. In practice, `URLRequest.allHTTPHeaderFields` should already normalize header names so collisions are rare.

mariedm (approved):

simaoseica-dd (approved):

DatadogRUM/Sources/Instrumentation/Resources/HeaderCapture/HeaderProcessor.swift
```diff
@@ -101,71 +99,91 @@ internal struct HeaderProcessor {
...
 
             let truncatedValue = truncateValue(value)
-            let entrySize = key.utf8.count + truncatedValue.utf8.count
+            let entrySize = originalKey.utf8.count + truncatedValue.utf8.count
 
             // Enforce total size budget
             guard totalSize + entrySize <= Self.maxTotalSize else { break }
 
-            result[key] = truncatedValue
+            result[originalKey] = truncatedValue
```
kikoveiga, line 186: Here we use `originalKey` which preserves original casing, should we use `name` to make sure it's lowercased and searching on Datadog is easier and more consistent?

barboraplasovska: Good question! Actually HTTP headers are traditionally displayed in their canonical form.

PR #2701, opened 2026-02-18
RUM-14313: Fix `Failed to decode AppStateInfo` error during app login

simaoseica-dd:
### What and why?

This PR improves determinism around `AppStateManager` startup/update behavior.

It addresses a timing race between `WatchdogTerminationChecker` and `AppStateManager` where the `previousAppStateInfo` might not be available in time to report a watchdog termination.

### How?
- Updates `AppStateManager` to gate dependent work with a `DispatchGroup` so callbacks run only after initial state loading completes.
- Updates `AppStateManaging` to expose `previousAppStateInfo` through a `completion` callback instead of direct property access.

### Review checklist

### DIFF
--- a/DatadogRUM/Sources/Instrumentation/AppState/AppStateManager.swift
+++ b/DatadogRUM/Sources/Instrumentation/AppState/AppStateManager.swift
@@ -8,21 +8,24 @@ import Foundation
 import DatadogInternal
 
 internal protocol AppStateManaging {
-    /// The app state information of the last application run.
-    var previousAppStateInfo: AppStateInfo? { get }
-    /// Deletes the app state from the data store.
-    func deleteAppState()
     /// Updates the app state based on the given application state.
     func updateAppState(state: AppState)
+    /// Returns the previous app state via `completion` callback.
+    func previousAppStateInfo(completion: @escaping (AppStateInfo?) -> Void)
     /// Builds the current app state.
     func currentAppStateInfo(completion: @escaping (AppStateInfo) -> Void)
-    /// Builds the current app state and stores it in the data store.
-    func storeCurrentAppState()
 }
 
 /// Manages the app state changes observed during application lifecycle events such as application start, resume and termination.
 internal final class AppStateManager: AppStateManaging {
-    let featureScope: FeatureScope
+    private static let defaultQueue = DispatchQueue(
+        label: "com.datadoghq.app-state-manager",
+        qos: .utility
+    )
+
+    private let featureScope: FeatureScope
+    private let initialStateGroup = DispatchGroup()
+    private let initialStateQueue: DispatchQueue
 
     /// The last app state observed during application lifecycle events.
     @ReadWriteLock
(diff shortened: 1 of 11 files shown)


DatadogRUM/Sources/RUMMonitor/Scopes/Utils/StartupTypeHandler.swift
```diff
@@ -33,29 +33,29 @@ internal final class StartupTypeHandler {
         self.coldStartRules = coldStartRules
     }
 
-    func startupType(currentAppState: AppStateInfo) -> RUMVitalAppLaunchEvent.Vital.StartupType {
+    func startupType(previousAppState: AppStateInfo?, currentAppState: AppStateInfo) -> RUMVitalAppLaunchEvent.Vital.StartupType {
         for rule in coldStartRules {
```
barboraplasovska, line 34: `appStateManager` is now unused in `StartupType`. We can remove the property and its `init` parameter.

DatadogRUM/Tests/RUMMonitor/Scopes/Utils/StartupTypeHandlerTests.swift
```diff
@@ -29,63 +29,63 @@ final class StartupTypeHandlerTests: XCTestCase {
...
 
     func testSystemRestart_returnsColdStart() {
         // Given
-        (appStateManager as? AppStateManagerMock)?.previousAppStateInfo = .mockWith(systemBootTime: 1_000_000)
+        let previousAppState: AppStateInfo = .mockWith(systemBootTime: 1_000_000)
         let currentAppState: AppStateInfo = .mockWith(systemBootTime: 1_100_000)
 
         // When
-        let result = handler.startupType(currentAppState: currentAppState)
+        let result = handler.startupType(previousAppState: previousAppState,currentAppState: currentAppState)
```
barboraplasovska: nit: missing space after comma

```suggestion
        let result = handler.startupType(previousAppState: previousAppState, currentAppState: currentAppState)
```

DatadogRUM/Sources/Instrumentation/AppState/AppStateManager.swift
barboraplasovska: suggestion: should we wrap `deleteAppState` with `onInitialStateLoaded` too? If it's called before the initial state finishes loading, `start()` could complete afterward and `storeCurrentAppState()` would write a new state back, silently undoing the delete.

simaoseica-dd: The `onInitialStateLoaded` is only guarding the access/mutation of the `AppStateManager` state. 
The `storeCurrentAppState` func always stores the most recent state, that is created on the fly, so the `deleteAppState` func only makes sense for testing since there is always a app state. 
To avoid confusion, I removed `deleteAppState` from the protocol and added documentation.

DatadogRUM/Sources/RUMMonitor/Scopes/Utils/StartupTypeHandler.swift
simaoseica-dd: good call

PR #2704, opened 2026-02-18
[PANA-6007] Integrate layer image rendering stage for the layer-tree recording pipeline

gonzalezreal:
### What and why?

This PR adds the layer image-rendering stage to the new Core Animation layer-tree recording pipeline.

It introduces a dedicated `LayerImageRenderer` and integrates it into `LayerRecorder` with a per-frame rendering budget, so image candidates can be rendered (or reused) within the frame time window while preserving frame cadence constraints.

It also updates wireframe diffing so image wireframe mutations include `resourceId` updates, which keeps image/resource synchronization correct when content changes without requiring full wireframe replacements.

Depends on:
- https://github.com/DataDog/dd-sdk-ios/pull/2698

### How?

- Add `LayerImageRenderer` to:
  - classify/render/reuse image candidates,
  - track partial captures for oversized layers,
  - map timeout and rendering failures into explicit outcomes used by recorder/wireframe generation.
- Add `LayerSnapshot+Rendering` helpers to centralize image-candidate and render-decision logic.
- Integrate renderer into `LayerRecorder` with remaining-time computation from the frame budget.
- Configure the layer-tree recorder budget in RecordingComponents to stay below the monitor delivery interval
- Extend wireframe diffing to mutate `resourceId` for image updates.
- Add focused Swift Testing coverage for rendering decision logic and renderer behavior

### Review checklist

### DIFF
--- a/DatadogSessionReplay/Sources/LayerTreeRecorder/LayerImageRenderer.swift
+++ b/DatadogSessionReplay/Sources/LayerTreeRecorder/LayerImageRenderer.swift
@@ -197,9 +197,17 @@ internal final class LayerImageRenderer: LayerImageRendering {
             )
 
             // If a new render is not needed and we still have the previous image,
-            // return the cached image directly
+            // reuse its resource and update frame only when needed
             if !layerImageChange.needsRender, let layerImage = images.object(forKey: snapshot.replayID as NSNumber) {
-                return layerImage
+                let frame = layerImageChange.rect(in: rootLayer)
+
+                if layerImage.frame.equalTo(frame) {
+                    return layerImage
+                }
+
+                let newLayerImage = LayerImage(resource: layerImage.resource, frame: frame)
+                images.setObject(newLayerImage, forKey: snapshot.replayID as NSNumber)
+                return newLayerImage
             }
 
             let image = renderImage(
--- a/DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Rendering.swift
+++ b/DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Rendering.swift
@@ -57,12 +57,14 @@ extension LayerSnapshot {
         let lastImageRect = imageRects[replayID]
         let isNew = lastImageRect == nil
 
-        // A capture is partial when the previous image did not cover full bounds,
-        // or when the layer is larger than the root viewport on first capture
-        let isPartialImage = lastImageRect.map { rect in
+        let lastRenderWasPartial = lastImageRect.map { rect in
             !rect.equalTo(layer.bounds)
-        } ?? layer.bounds.exceeds(rootLayer.bounds)
-        let imageRectDidChange = isPartialImage && !(lastImageRect?.contains(visibleRect) ?? false)
+        } ?? false
+        let renderWillBePartial = layer.bounds.sizeExceeds(rootLayer.bounds)
+        let isPartialImage = lastRenderWasPartial || (isNew && renderWillBePartial)
+        // Partial captures are content-addressed by local visible rect
+        // Any rect change requires a re-render to keep the rendered pixels aligned
+        let imageRectDidChange = isPartialImage && !(lastImageRect?.equalTo(visibleRect) ?? false)
 
         // Determine when pixel content may have changed
         let needsRender = if type(of: layer) == CALayer.self {
@@ -78,7 +80,7 @@ extension LayerSnapshot {
         return LayerImageChange(
             layer: layer,
             // For oversized layers, render only the visible local rect
-            rect: layer.bounds.exceeds(rootLayer.bounds) ? visibleRect : layer.bounds,
+            rect: renderWillBePartial ? visibleRect : layer.bounds,
             needsRender: needsRender
         )
     }
@@ -110,7 +112,7 @@ extension CALayer {
 }
 
 extension CGRect {
-    fileprivate func exceeds(_ other: CGRect) -> Bool {
+    fileprivate func sizeExceeds(_ other: CGRect) -> Bool {
         width > other.width || height > other.height
     }
 }
--- a/DatadogSessionReplay/Tests/LayerTreeRecorder/LayerImageRendererTests.swift
+++ b/DatadogSessionReplay/Tests/LayerTreeRecorder/LayerImageRendererTests.swift
@@ -126,6 +126,103 @@ struct LayerImageRendererTests {
 
     @available(iOS 13.0, tvOS 13.0, *)
     @Test
+    func rerendersPartialImageWhenVisibleRectChangesWithinCachedRect() async throws {
+        // given
+        let rootLayer = Fixtures.rootLayer
+        let layer = CALayer()
+        layer.bounds = CGRect(x: 0, y: 0, width: 400, height: 120)
+        layer.position = CGPoint(x: 200, y: 60)
+        rootLayer.addSublayer(layer)
+
+        let firstSnapshot = Fixtures.layerSnapshot(
+            for: layer,
+            replayID: 8,
+            in: rootLayer,
+            hasContents: true,
+            clipRect: CGRect(x: 0, y: 0, width: 200, height: 120)
+        )
+        let secondSnapshot = Fixtures.layerSnapshot(
+            for: layer,
+            replayID: 8,
+            in: rootLayer,
+            hasContents: false,
+            clipRect: CGRect(x: 60, y: 0, width: 120, height: 120)
+        )
+
+        let renderer = LayerImageRenderer(scale: 1, timeSource: .constant(0))
+
+        let firstResults = await renderer.renderImages(
+            for: [firstSnapshot],
+            changes: .init(),
+            rootLayer: .init(rootLayer),
+            timeoutInterval: 1
+        )
+
+        // when
+        let secondResults = await renderer.renderImages(
+            for: [secondSnapshot],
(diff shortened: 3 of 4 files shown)


DatadogSessionReplay/Sources/LayerTreeRecorder/LayerImageRenderer.swift
```diff
@@ -0,0 +1,209 @@
...
+
+    private func candidates(
+        for snapshots: [LayerSnapshot],
+        with changes: CALayerChangeset
+    ) -> [LayerSnapshot] {
+        snapshots.filter { snapshot in
+            guard let layerClass = snapshot.layer.class else {
+                imageRects.removeValue(forKey: snapshot.replayID)
+                images.removeObject(forKey: snapshot.replayID as NSNumber)
+                return false
```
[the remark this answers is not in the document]
gonzalezreal: Fixed in https://github.com/DataDog/dd-sdk-ios/pull/2704/commits/6f7d998561878b917d06ddd048259689b8454970

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Rendering.swift
```diff
@@ -0,0 +1,117 @@
...
+        let isPartialImage = lastImageRect.map { rect in
+            !rect.equalTo(layer.bounds)
+        } ?? layer.bounds.exceeds(rootLayer.bounds)
+        let imageRectDidChange = isPartialImage && !(lastImageRect?.contains(visibleRect) ?? false)
+
+        // Determine when pixel content may have changed
+        let needsRender = if type(of: layer) == CALayer.self {
+            // For plain CALayer this means first appearance with layer contents, explicit
+            // content changes, or partial coverage changes
+            (hasContents && isNew) || hasContentChanges || imageRectDidChange
```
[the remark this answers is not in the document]
gonzalezreal: Good point. In this pipeline we enforce a per-frame performance budget and target completion within the monitor cadence, so the “capture in flight while new changes arrive” window should be rare in practice.

Because of that, we’re intentionally keeping the current render-decision logic simple for now and avoiding extra fallback heuristics that would add churn/complexity.

If telemetry later shows budget overruns or dropped schedules, we can revisit this with pending-change merging.

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Rendering.swift
```diff
@@ -0,0 +1,117 @@
...
+
+        // Ignore pure geometry animations as they can be represented by frame updates
+        return animationKeys.contains {
+            !Constants.geometryKeys.contains($0)
+        }
+    }
+}
+
+extension CGRect {
+    fileprivate func exceeds(_ other: CGRect) -> Bool {
```
sethfowler-datadog: Nit: perhaps `sizeExceeds()` would be a little more clear? `exceeds()` sounds like it could also mean "extends outside the bounds of", which could apply to a smaller layer depending on how it was positioned, but this is really just about size.

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerImageRenderer.swift
```diff
@@ -0,0 +1,266 @@
...
+    private let cachePolicy: CachePolicy
+
+    // Stores the last captured image rect in the layer local coordinate space.
+    // This lets us detect when a previously partial capture no longer covers
+    // the currently visible region
+    private var imageRects: [Int64: CGRect] = [:]
+
+    // Tracks when a replay ID was last observed in a captured snapshot, even if
+    // it did not need image rendering in that frame
+    private var frames: [Int64: UInt64] = [:]
```
sethfowler-datadog, line 80: Nit: it might be nice to have a `typealias` or wrapper `struct` for replay IDs to make these types a bit more self-documenting.

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Rendering.swift
```diff
@@ -0,0 +1,117 @@
...
+        let hasContentChanges = changes.hasContentChanges(for: self.layer)
+        let lastImageRect = imageRects[replayID]
+        let isNew = lastImageRect == nil
+
+        // A capture is partial when the previous image did not cover full bounds,
+        // or when the layer is larger than the root viewport on first capture
+        let isPartialImage = lastImageRect.map { rect in
+            !rect.equalTo(layer.bounds)
+        } ?? layer.bounds.exceeds(rootLayer.bounds)
+        let imageRectDidChange = isPartialImage && !(lastImageRect?.contains(visibleRect) ?? false)
```
sethfowler-datadog: So if `lastImageRect?.contains(visibleRect)`, then `imageRectDidChange` is false, and that might mean that we report `needsRender: false`, which will cause us to reuse the cached `LayerImage` if one exists. But doesn't that `LayerImage` have `visibleRect` baked into it for oversized layers? If `visibleRect` has changed, but is still contained in `lastImageRect`, then couldn't we render the wrong thing at replay time?

Think e.g. of a scenario where the left edge of the layer's `clipRect` has moved to the right, but the layer's position has not changed. In that scenario, I think reusing the old `LayerImage` with its old rect would show more of the image than a new rendering would.

You could fix this in different ways, but the most obvious one is just to create a new `LayerImage` with the same `UIImageResource`, but a different rect, when necessary.

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Rendering.swift
```diff
@@ -0,0 +1,117 @@
...
+
+        let hasContentChanges = changes.hasContentChanges(for: self.layer)
+        let lastImageRect = imageRects[replayID]
+        let isNew = lastImageRect == nil
+
+        // A capture is partial when the previous image did not cover full bounds,
+        // or when the layer is larger than the root viewport on first capture
+        let isPartialImage = lastImageRect.map { rect in
+            !rect.equalTo(layer.bounds)
+        } ?? layer.bounds.exceeds(rootLayer.bounds)
```
sethfowler-datadog: Nit: It might be a bit easier to follow the logic in this code if you separate these two concepts; maybe:
```swift
let lastRenderWasPartial = ...
let renderWillBePartial = ...
```

Even with the comment, I found myself devoting a lot of brain power to understanding the implications of this definition. I think that was partially because information about the past and the present is mixed here, and partially because the fact that the information about the present is only checked for new images complicates reasoning about the behavior. That's why I suggest tweaking the definition, but it might just be my issue. 🙂

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerImageRenderer.swift
```diff
@@ -0,0 +1,266 @@
...
+        results.reserveCapacity(candidates.count)
+
+        var firstUnprocessedIndex = candidates.endIndex
+
+        for index in candidates.indices {
+            let now = timeSource.now
+
+            // Stop rendering when the time budget is exhausted and mark the
+            // remaining candidates as timed out
+            if Task.isCancelled || (now - startTime) >= timeoutInterval {
```
sethfowler-datadog, line 136: Does it make sense to "charge" this code for the time when it's yielded? I'm not sure, honestly; I suppose it depends on whether your goal is to avoid occupying the main thread for too long, or to prevent a round of rendering from lasting past the beginning of the next snapshot, or a mixture of both...

gonzalezreal: Yes, it is intentional. `timeoutInterval` is a wall-clock frame budget for the whole rendering pass, not just active CPU time. We yield to avoid staying too long in the main thread (see the 8ms yield threshold).

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerImageRenderer.swift
gonzalezreal: Makes sense. I will address this in a future PR by using a wrapper type (tagged Int64).

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Rendering.swift
gonzalezreal: Agreed. I changed partial-image rect comparison from `contains` to `equalTo`, so any visible-rect change triggers a re-render. I also updated cache reuse, when bitmap re-render is not needed but geometry changed, we now reuse the same `UIImageResource` and create/cache a new `LayerImage` with the updated frame.

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Rendering.swift
gonzalezreal: Nice! The logic is now split into `lastRenderWasPartial` and `renderWillBePartial` before deriving `isPartialImage`, which makes the past vs. current reasoning explicit.

DatadogSessionReplay/Sources/LayerTreeRecorder/LayerSnapshot+Rendering.swift
gonzalezreal: Renamed. Thanks!

PR #2705, opened 2026-02-19
Add support of renaming Objective-C runtime classes for generated RUM models

0xnm:
### What and why?

Inspired by https://github.com/DataDog/dd-sdk-kotlin-multiplatform/issues/233.

There is `DDRUMErrorEventErrorMeta` class in the native iOS SDK which has to be renamed to the `DDRUMErrorEventErrorMetaInfo` in case of usage from KMP SDK, because otherwise there is clash with a class generated by the KMP (it generates `*Meta` classes for every ObjC class).

This worked till Kotlin 2.3.20, where build tooling seems to be updated and now one more workaround has to be added on the client side.

It is simpler just to give this class a different name in iOS SDK to avoid having such issue.

### Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- [IMPROVEMENT] Rename `DDRUMErrorEventErrorMeta` to `DDRUMErrorEventErrorMetaInfo`, add support of custom Objective-C runtime names for generated RUM models. See [#2705][]
+
 # 3.7.0 / 18-02-2026
 
 - [FEATURE] Add evaluation logging to `DatadogFlags` module. See [#2646][]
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - [IMPROVEMENT] Improve span attribute encoding error messages to include attribute name and context. See [#2676][]
 - [IMPROVEMENT] Expose public entities from `DatadogInternal` to prevent `DatadogInternal` imports in customer code. See [#2666][]
 - [FIX] Propagate feature flags to RUM error and view events for crashes, fatal app hangs, and watchdog terminations.
+- [IMPROVEMENT] Rename `DDRUMErrorEventErrorMeta` to `DDRUMErrorEventErrorMetaInfo`, add support of custom Objective-C runtime names for generated RUM models. See [#2705][]
```
simaoseica-dd: Can you rebase with `develop`?  This point should be added to the next release

0xnm: done 👍

PR #2706, opened 2026-02-19
[RUM-13420] Cache DatadogContext.ddTags

tienquocbui:
### What and why?

Cache `DatadogContext.ddTags` as a stored property to avoid rebuilding the same string on every event. The tags are deterministic (derived from `service`, `version`, `sdkVersion`, `env`, `variant`) and only `version` can change after init, so repeated computation is wasteful. Fix copy-paste bug in `CrashContext.init` where `sdkVersion` was incorrectly assigned the `service` value instead of `sdkVersion`.

### How?

* Replace the computed `ddTags` extension property on `DatadogContext` with a `public private(set) var ddTags` stored property, computed once in `init()` via a static `buildDDTags(...)` helper.
* Add a `didSet` observer on `version` to re-sanitize and recompute `ddTags` when the app version changes at runtime.
* Simplify `CrashContext.ddTags` from dictionary-based construction to direct string interpolation for deterministic ordering (all fields are `let`, so no caching needed).
* Add unit tests for `ddTags` correctness, variant handling, sanitization, and version update recomputation.

### Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,6 @@
 - [IMPROVEMENT] Improve span attribute encoding error messages to include attribute name and context. See [#2676][]
 - [IMPROVEMENT] Expose public entities from `DatadogInternal` to prevent `DatadogInternal` imports in customer code. See [#2666][]
 - [FIX] Propagate feature flags to RUM error and view events for crashes, fatal app hangs, and watchdog terminations. See [#2688][]
-- [FIX] Propagate feature flags to RUM error and view events for crashes, fatal app hangs, and watchdog terminations.
 
 # 3.6.1 / 02-02-2026
 
@@ -1061,7 +1060,6 @@ Release `2.0` introduces breaking changes. Follow the [Migration Guide](MIGRATIO
 [#2676]: https://github.com/DataDog/dd-sdk-ios/pull/2676
 [#2688]: https://github.com/DataDog/dd-sdk-ios/pull/2688
 [#2705]: https://github.com/DataDog/dd-sdk-ios/pull/2705
-[#2706]: https://github.com/DataDog/dd-sdk-ios/pull/2706
 
 [@00fa9a]: https://github.com/00FA9A
 [@britton-earnin]: https://github.com/Britton-Earnin


DatadogInternal/Sources/Context/DatadogContext.swift
```diff
@@ -24,7 +24,12 @@ public struct DatadogContext {
     public let env: String
 
     /// The version of the application that data is generated from. Used for [Unified Service Tagging](https://docs.datadoghq.com/getting_started/tagging/unified_service_tagging).
-    public var version: String
+    public var version: String {
+        didSet {
+            version = version.sanitizedToDDTags()
```
simaoseica-dd, line 32: This sanitization and tag rebuilding can be ignored if the `version == oldValue`

DatadogInternal/Sources/Context/DatadogContext.swift
```diff
@@ -187,8 +195,17 @@ public struct DatadogContext {
         self.brightnessLevel = brightnessLevel
         self.isLowPowerModeEnabled = isLowPowerModeEnabled
         self.additionalContext = additionalContext
+        self.ddTags = Self.buildDDTags(service: self.service, version: self.version, sdkVersion: self.sdkVersion, env: self.env, variant: self.variant)
     }
     // swiftlint:enable function_default_parameter_at_end
+
+    private static func buildDDTags(service: String, version: String, sdkVersion: String, env: String, variant: String?) -> String {
```
simaoseica-dd: suggestion: This function is more aligned with the instance instead of the class `DatadogContext`. It is private and the return value changes according to the change in the state of the datadog context instance.

CHANGELOG.md
```diff
@@ -1,15 +1,14 @@
...
 - [FEATURE] Deprecate `URLSessionInstrumentation.enable(with:in:)` API in favor of `URLSessionInstrumentation.enableDurationBreakdown(with:in:)`. See [#2634][]
-- [FEATURE] Session Replay now captures only when the screen changes, reducing work while the app is idle. See [#2674][]
+- [FEATURE] Session Replay now captures only when the screen changes, reducing work while the app is idle.
 - [IMPROVEMENT] Skip malformed Logs attributes individually instead of dropping the entire event, and log clear error messages. See [#2665][]
 - [IMPROVEMENT] Improve span attribute encoding error messages to include attribute name and context. See [#2676][]
 - [IMPROVEMENT] Expose public entities from `DatadogInternal` to prevent `DatadogInternal` imports in customer code. See [#2666][]
-- [FIX] Propagate feature flags to RUM error and view events for crashes, fatal app hangs, and watchdog terminations. See [#2688][]
+- [FIX] Propagate feature flags to RUM error and view events for crashes, fatal app hangs, and watchdog terminations.
+- [IMPROVEMENT] Cache `DatadogContext.ddTags` to avoid repeated string computation on every event. See [#2706][]
```
simaoseica-dd: suggestion: Usually the CHANGELOG entry is only added for user facing changes, to make it cleaner and readable. This task is an improvement on the implementation.

PR #2707, opened 2026-02-19
RUM-13728: Improve Profiling `flush_buffer` with new Binary Image resolver

simaoseica-dd:
### What and why?

This PR moves binary image resolution out of the sampling loop and introduces a cached resolver to handle frame enrichment.

Previously, binary image resolution ran during buffer flush on the sampling thread which increased flush cost and could interfere with the regular sampling.
With this change, sampling remains focused on stack capture and image resolution is handled by the callback consumer.

### How?
- Adds a new `binary_image_resolver` component responsible for binary image lookup and cache management.
- Initializes a `binary_image_cache` in `ctor_profiler` before sampling starts.
- Simplifies the sampling engine to focus on capture and deliver raw frames only (better separation of concerns).
- Resolves frame binary images using the cache before adding samples to the profile.

### Review checklist

### DIFF
--- a/DatadogProfiling/Mach/binary_image_resolver.cpp
+++ b/DatadogProfiling/Mach/binary_image_resolver.cpp
@@ -268,44 +268,29 @@ bool binary_image_cache::lookup(uint64_t instruction_ptr, binary_image_t* out_im
             const cached_image_t& cached = it->second;
             out_image->load_address = cached.load_address;
             memcpy(out_image->uuid, cached.uuid, sizeof(uuid_t));
-
-            if (cached.filename) {
-                size_t len = strlen(cached.filename) + 1;
-                char* fname = (char*)malloc(len);
-                if (fname) {
-                    memcpy(fname, cached.filename, len);
-                    out_image->filename = fname;
-                }
-            }
+            out_image->filename = cached.filename;
             return true;
         }
     }
 
     // Cache miss — fall back to full Mach-O header parsing.
-    // This handles images loaded between start() and now, or images
+    // This handles images loaded between load() and now, or images
     // that the dyld callback missed.
     if (binary_image_lookup_pc(out_image, (void*)instruction_ptr)) {
         // Cache the result for future lookups
         cached_image_t cached;
         cached.load_address = out_image->load_address;
         memcpy(cached.uuid, out_image->uuid, sizeof(uuid_t));
-        cached.filename = nullptr;
-
-        if (out_image->filename) {
-            size_t len = strlen(out_image->filename) + 1;
-            cached.filename = (char*)malloc(len);
-            if (cached.filename) {
-                memcpy(cached.filename, out_image->filename, len);
-            }
-        }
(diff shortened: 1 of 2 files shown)


DatadogProfiling/Mach/binary_image_resolver.cpp
```diff
@@ -0,0 +1,343 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+#include "binary_image_resolver.h"
+
+#if defined(__APPLE__) && !TARGET_OS_WATCH
```
maxep, line 9: Thanks :)

DatadogProfiling/Mach/include/binary_image_resolver.h
```diff
@@ -0,0 +1,138 @@
...
+    binary_image_cache(const binary_image_cache&) = delete;
+    binary_image_cache& operator=(const binary_image_cache&) = delete;
+
+    /**
+     * Populate cache with all currently loaded images and start watching
+     * for new image loads via dyld notifications.
+     *
+     * @return true if dyld register was started successfully, false otherwise
+     */
+    bool start();
```
maxep: nit/ I don't think start reflect what it does, I would suggest `load` or `prepare`

simaoseica-dd: Changed to `load`, since it firstly populates the cache with the current loaded images.

DatadogProfiling/Mach/binary_image_resolver.cpp
```diff
@@ -0,0 +1,346 @@
...
+            memcpy(out_image->uuid, cached.uuid, sizeof(uuid_t));
+
+            if (cached.filename) {
+                size_t len = strlen(cached.filename) + 1;
+                char* fname = (char*)malloc(len);
+                if (fname) {
+                    memcpy(fname, cached.filename, len);
+                    out_image->filename = fname;
+                }
+            }
```
maxep: Do we need to copy, can we just pass the pointer? we should then skip deallocating the image when freeing the trace.

DatadogProfiling/Mach/binary_image_resolver.cpp
```diff
@@ -0,0 +1,346 @@
...
+            return true;
+        }
+    }
+
+    // Cache miss — fall back to full Mach-O header parsing.
+    // This handles images loaded between start() and now, or images
```
maxep:
```suggestion
    // This handles images loaded between load() and now, or images
```

DatadogProfiling/Mach/binary_image_resolver.cpp
simaoseica: Simplified it. thx

PR #2708, opened 2026-02-19
RUM-13728: Rename `ctor_profiler` to `dd_profiler`

simaoseica-dd:
### What and why?

This PR renames `ctor_profiler` to `dd_profiler` as a preparatory step toward making it a continuous profiler. The constructor-based profiler was previously scoped to app launch profiling only, but will evolve into a single long running mach sampling profiler.

### How?

This renaming consolidates the two separate components, the `ctor_profiler` and the thin `mach_profiler` wrapper, into one unified `dd_profiler` component.
All symbols were renamed  to `dd_profiler` and calls in the Swift part were updated as well. It doesn't have any logic changes.

### Review checklist

### DIFF
--- a/DatadogProfiling/Mach/include/dd_profiler.h
+++ b/DatadogProfiling/Mach/include/dd_profiler.h
@@ -143,15 +143,6 @@ typedef struct profiler profiler_t;
 #endif
 
 /**
- * Sets the main thread pthread identifier.
- *
- * This function should be called from the main thread early in the process lifecycle.
- *
- * @param thread The pthread identifier for the main thread
- */
-void set_main_thread(pthread_t thread);
-
-/**
  * Creates a profiler instance.
  *
  * Uses fixed intervals for consistent sampling behavior.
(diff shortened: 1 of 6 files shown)


DatadogProfiling/Mach/include/dd_profiler.h
```diff
@@ -116,11 +117,17 @@ static const sampling_config_t SAMPLING_CONFIG_DEFAULT = {
  */
 typedef void (*stack_trace_callback_t)(stack_trace_t* traces, size_t count, void* ctx);
 
+// UserDefaults constants centralized for Profiling
+#define DD_PROFILING_USER_DEFAULTS_SUITE_NAME "com.datadoghq.ios-sdk.profiling"
+#define DD_PROFILING_IS_ENABLED_KEY "is_profiling_enabled"
+#define DD_PROFILING_SAMPLE_RATE_KEY "profiling_sample_rate"
+
```
fuzzybinary: Where are these used? Depending on how they're used it might be better to have them static `const char*` over `#defines` (which I generally like to avoid when possible).

DatadogProfiling/Mach/include/dd_profiler.h
```diff
@@ -189,6 +196,118 @@ void profiler_stop(profiler_t* profiler);
...
+ *
+ * Reads the profiling enabled state from UserDefaults suite to determine
+ * if the profiling feature was previously enabled via Profiling.enable().
+ *
+ * @return true if profiling is enabled, false otherwise
+ *
+ * @note Reads from suite "com.datadoghq.ios-sdk" with key "is_profiling_enabled"
+ * @note Returns false if the key doesn't exist or on read errors
+ */
+bool is_profiling_enabled(void);
```
fuzzybinary: This might be a bit late for this, but these should probably be prefixed in some way - `dd_is_profiling_enabled` for example, as right now we might conflict with another profiler if they use similar function names.

fuzzybinary: Actually these appear to be internal functions, that likely shouldn't be exported to the `dd_profiler.h`interface. Double check that everything that's in this header and labeled `extern "C"` is part of the needed public interface. Otherwise leaving it out and using default linking / name mangling is preferred.

DatadogProfiling/Mach/include/dd_profiler.h
simaoseica-dd: These are `UserDefaults` keys used in cpp and Swift. The `#define` should be the simplest and cleanest option for this use case.

DatadogProfiling/Mach/include/dd_profiler.h
simaoseica-dd: It is not too late. I moved them to the `dd_profiler_testing.h` since they are internal to dd_profiler and needed for the Swift tests.

maxep (dismissed):
LGTM 👍

fuzzybinary (approved):

PR #2714, opened 2026-02-24
[RUM-14721] Knuth deterministic sampling for RUM sessions, traces, and Session Replay

qsellem:
### What and why?

Extends `DeterministicSampler` to support session-seeded sampling across RUM, traces, and Session Replay. When a RUM session is active, all child features (distributed tracing, replay) use the same Knuth hash seed derived from the session UUID, maximising the probability that a sampled session retains its traces and replay for full observability correlation — consistent with the browser SDK v7 algorithm.

### How?

`DeterministicSampler` gains a public `isSampled` property, and a `combined(with:)` method that composes a parent session rate with a child feature rate while preserving the seed.

`RUMCoreContext` now carries a `sessionSampler: DeterministicSampler`. Consumers call `sessionSampler.combined(with: childRate).sample()` instead of reconstructing a sampler from the session ID string:

- `RecordingCoordinator` uses `sessionSampler.combined(with: replaySampleRate)`
- `URLSessionRUMResourcesHandler` uses `sessionSampler.combined(with: samplingRate)`
- `TracingURLSessionHandler` uses `sessionSampler.combined(with: traceSamplingRate)`

Sessions are now always assigned a real UUID regardless of sampling outcome. `RUMSessionState` carries an explicit `isSampled: Bool` instead of using `.nullUUID` as a sentinel. `SpanEventBuilder`, `RemoteLogger`, and `WebViewLogReceiver` skip RUM context attachment when the session is not sampled.

When no RUM session context is available, consumers fall back to random sampling — intentional, as there is no session to correlate to.

### Review checklist

---
<details><summary>First Iteration</summary>
<p>
## What

  Implements deterministic session sampling for the Datadog iOS SDK using Knuth multiplicative hashing. When a session is sampled, the same deterministic algorithm — applied consistently
  across iOS and server — maximises the probability that the same session retains traces and a replay, enabling full observability correlation.

  ## Implementation

  ### `DeterministicSampler`

  A new `DeterministicSampler` struct in `DatadogInternal` replaces the random `Sampler` for all session-level sampling decisions:

  - `init(seed: UInt64, samplingRate:)` — Knuth multiplicative hash: `(seed * 1_111_111_111_111_111_111) % 2^64 / 2^64 <= samplingRate / 100`, using `Double` precision and `<=` to align
  with the browser SDK v7 reference implementation
  - `init(sessionID: String, samplingRate:)` — convenience init that extracts the last UUID segment (12-hex node component) as the seed. Falls back to `seed=0` for malformed inputs
  (fail-open: a broken UUID is always sampled rather than silently dropped)
  - `seed` property — stored to allow sampler composition
  - `combined(with childRate:)` — returns a new `DeterministicSampler` with the same seed and a composed rate (`sessionRate * childRate / 100`), so child features (traces, replay) remain
  correlated to the session decision

  ### Sampling architecture

  `RUMCoreContext` now carries a `sessionSampler: DeterministicSampler` instance (replacing the raw `sessionSampleRate: SampleRate`). Consumers no longer reconstruct the sampler or parse
  UUIDs:

  - `RecordingCoordinator` → `sessionSampler.combined(with: replaySampleRate).sample()`
  - `TracingURLSessionHandler` → `sessionSampler.combined(with: samplingRate).sample()`
  - `CrashReportReceiver` → `sessionSampler.combined(with: sessionSampler.samplingRate).sample()`

  When no RUM session context is available (e.g. replay before any RUM event), these consumers fall back to random sampling — intentional, as there is no session to correlate to.

  ## Behavioral change

  `DeterministicSampler.init(seed:samplingRate:)` uses `Double`/`<=` for the threshold comparison, replacing the previous `Float`/`<`. This aligns with the browser SDK v7 reference
  implementation.

  **Impact:** For sessions where the Knuth hash of the seed falls exactly on the threshold boundary (`hash == threshold`), the new implementation samples the session whereas the previous
  one would not. In practice this affects an extremely small fraction of sessions (probability ≈ 1 / 2^64) and has no observable impact at normal sampling rates.

  ## Testing

  - Cross-SDK test vectors: 10 known UUID/seed pairs verified against the Knuth formula
  - Determinism test: same session UUID always produces the same sampling decision across multiple calls
  - Child-rate composition: `TracingURLSessionHandlerTests` and `RecordingCoordinatorTests` verify that `combined(with:)` correctly applies both session and child rates
</p>
</details>

### DIFF
--- a/DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
+++ b/DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
@@ -254,13 +254,15 @@ extension DistributedTracing {
         // with the tracing `samplingRate` while preserving the seed, so every resource in
         // the same session receives a consistent sampling decision.
         // When no RUM context exists, fall back to a random `Sampler`.
-        let sampler: Sampling = networkContext?.rumContext?.sessionSampler.combined(with: samplingRate)
+        let sampler: () -> Sampling = {
+            networkContext?.rumContext?.sessionSampler.combined(with: samplingRate)
             ?? Sampler(samplingRate: samplingRate)
+        }
         // In case there is, we use the same traceID so the backend can link the span generated from the RUM resource
         // with the trace.
         let traceID = activeSpanContext?.traceID ?? traceIDGenerator.generate()
         let spanID = spanIDGenerator.generate()
-        let samplingPriority = activeSpanContext?.samplingPriority ?? (sampler.sample() ? .autoKeep : .autoDrop)
+        let samplingPriority = activeSpanContext?.samplingPriority ?? (sampler().sample() ? .autoKeep : .autoDrop)
         let samplingDecisionMaker = activeSpanContext?.samplingMechanismType ?? .agentRate
 
         // Extract GraphQL attributes from request before they are removed


DatadogInternal/Sources/Utils/DeterministicSampler.swift
```diff
@@ -41,3 +41,25 @@ public struct DeterministicSampler: Sampling {
...
+///
+/// This produces consistent results across Datadog SDKs (iOS, browser, server).
+/// The formula is: `(identifier * knuthFactor) % 2^64 / 2^64 <= sampleRate / 100`
+///
+/// - Parameters:
+///   - identifier: A 64-bit integer derived from the session UUID (last 48 bits of the UUID
+///     string's 5th segment, parsed as hexadecimal).
+///   - sampleRate: Percentage between 0.0 and 100.0.
+/// - Returns: `true` if the identifier falls within the sampling threshold.
+public func sampleUsingKnuthFactor(identifier: UInt64, sampleRate: SampleRate) -> Bool {
```
maxep: This is a duplication of the `DeterministicSampler` above. We should either use `DeterministicSampler` as is or updated it to apply the `Double` precision and the `<=` comparison.

DatadogRUM/Sources/Integrations/CrashReportReceiver.swift
```diff
@@ -84,6 +84,17 @@ internal struct CrashReportReceiver: FeatureMessageReceiver {
...
+        let shouldSampleNewSession: Bool = {
+            let newSessionUUID = uuidGenerator.generateUnique()
+            let newSessionIdentifier = newSessionUUID.rawValue.uuidString
+                .split(separator: "-")
+                .last
+                .flatMap { UInt64($0, radix: 16) }
+            return newSessionIdentifier.map {
+                sampleUsingKnuthFactor(identifier: $0, sampleRate: sessionSampler.samplingRate)
+            } ?? sessionSampler.sample()
+        }()
```
maxep: This logic is repeated in multiple places, it should be shared in the `DatadogInternal` module IMO. I could be a convenient initializer of the `DeterministicSampler`

DatadogSessionReplay/Tests/Recorder/RecordingCoordinatorTests.swift
```diff
@@ -307,14 +307,154 @@ class RecordingCoordinatorTests: XCTestCase {
...
             methodCallTelemetrySamplingRate: methodCallTelemetrySamplingRate
         )
     }
 }
 
+// MARK: - Knuth Replay Sampling Tests
+
+/// Tests that `RecordingCoordinator` uses `sampleUsingKnuthFactor` with
+/// the composed (child-corrected) effective rate.
+final class KnuthReplaySamplingTests: XCTestCase {
```
maxep: We should have a single test-suite per file. This one could be merged with `RecordingCoordinatorTests`

DatadogTrace/Tests/TracingURLSessionHandlerTests.swift
```diff
@@ -863,3 +863,201 @@ class TracingURLSessionHandlerTests: XCTestCase {
...
+// MARK: - Knuth trace sampling with child rate correction
+
+/// Tests for `TracingURLSessionHandler` Knuth-based deterministic sampling and child rate correction (CHILD-01).
+///
+/// These tests verify that after replacing `DeterministicSampler` with `sampleUsingKnuthFactor`:
+///  1. Sampling decisions are deterministic for the same sessionID / traceID.
+///  2. The effective rate is `sessionSampleRate.composed(with: traceRate)` — not `traceRate` alone.
+///  3. When no RUM context is present, the handler falls back gracefully (no crash).
+///  4. Boundary values match cross-SDK Knuth vectors (Double / `<=` semantics).
+class KnuthTraceSamplingTests: XCTestCase {
```
maxep: Same here, lets keep a single test-suite per test file

DatadogRUM/Tests/RUMMonitor/Scopes/RUMSessionScopeTests.swift
```diff
@@ -701,3 +701,99 @@ class RUMSessionScopeTests: XCTestCase {
         XCTAssertTrue(result)
     }
 }
+
+class KnuthSessionSamplingTests: XCTestCase {
```
maxep: This suite is not testing sampling using the `Knuth` algorithm. It's not testing much that is not already tested, it can be removed IMO.

DatadogRUM/Tests/RUMMonitor/Scopes/RUMSessionScopeTests.swift
```diff
@@ -701,3 +701,99 @@ class RUMSessionScopeTests: XCTestCase {
...
+            scope1.isSampled,
+            scope2.isSampled,
+            "Same UUID must produce the same keep/drop decision across re-initialisations"
+        )
+    }
+}
+
+// MARK: - RUMCoreContext sessionSampleRate tests
+
+class RUMCoreContextSessionSampleRateTests: XCTestCase {
```
maxep: This suite has not much value either.

DatadogSessionReplay/Tests/Recorder/RecordingCoordinatorTests.swift
maxep: Also, I think only the following tests are valuable:
- test_givenSameSessionID_samplingDecisionIsIdentical_acrossMultipleCalls
- test_knownSessionUUID_matchesPrecomputedKnuthResult

The rest, not so much.

DatadogRUM/Sources/Integrations/CrashReportReceiver.swift
maxep: Note: This might not be needed if the `Sampler` instance is shared in the `RUMCoreContext`.

DatadogInternal/Sources/Models/RUM/RUMCoreContext.swift
```diff
@@ -20,6 +20,10 @@ public struct RUMCoreContext: AdditionalContext, Equatable {
     public let userActionID: String?
     /// Current view related server time offset
     public let viewServerTimeOffset: TimeInterval?
+    /// The sampling rate applied to the current RUM session (0.0–100.0).
+    /// Consumers such as Trace and Session Replay use this value to compute
+    /// the effective child sample rate via `sessionRate.composed(with: childRate)`.
+    public let sessionSampleRate: SampleRate
```
maxep: I think we should share an actual `Sampler` here, instead of a sampler rate and expect consumer to re-apply the same deterministic algorithm. Meaning that we should keep the responsibility of the sampler in RUM only.
That imply having a `combined` method at the `Sampler` level (as opposed to the sample-rate), that would simplify usage in other modules

DatadogInternal/Sources/Utils/DeterministicSampler.swift
maxep: Yes, using an object aligns with the current design using protocol compliance and function programming is less idiomatic in Swift. I wouldn't mind renaming the `DeterministicSampler ` to `KnuthFactorSampler`, but we should stick to the design.

DatadogTrace/Tests/TracingURLSessionHandlerTests.swift
qsellem: Done

DatadogInternal/Sources/Models/RUM/RUMCoreContext.swift
qsellem: Done — `RUMCoreContext.sessionSampleRate: SampleRate` has been replaced by `sessionSampler: DeterministicSampler`. `DeterministicSampler` now stores its `seed` and exposes a
  `combined(with childRate:)` method that returns a new sampler with the same seed and a composed rate.

  Consumers (`RecordingCoordinator`, `TracingURLSessionHandler`, `CrashReportReceiver`) now call `sessionSampler.combined(with: childRate).sample()` — no UUID parsing or sampler
  reconstruction on their side. Sampling responsibility stays in RUM.

DatadogRUM/Sources/Integrations/CrashReportReceiver.swift
qsellem: Done — `DeterministicSampler` now has a `init(sessionID: String, samplingRate:)` convenience initializer in `DatadogInternal` that encapsulates the UUID last-segment extraction. 

The sampler instance is also shared via `RUMCoreContext.sessionSampler`, so `CrashReportReceiver` receives it directly and calls `sessionSampler.combined(with: childRate).sample()` — no UUID parsing on its side.

DatadogInternal/Sources/Utils/DeterministicSampler.swift
qsellem: Done — removed `sampleUsingKnuthFactor` entirely and updated `DeterministicSampler` to use `Double` precision and `<=` comparison. This aligns with the existing protocol-based design.
  The cross-SDK naming concern is set aside for now and can be revisited separately if needed.

DatadogRUM/Sources/Integrations/CrashReportReceiver.swift
qsellem: Addressed on both fronts. `DeterministicSampler` in `DatadogInternal` now has a convenience `init(sessionID:samplingRate:)` that handles the UUID extraction internally. More
  importantly, following your follow-up note, `RUMCoreContext` now carries a `sessionSampler: DeterministicSampler` instance — so consumers no longer need to re-parse the UUID at all.
  `CrashReportReceiver` uses `DeterministicSampler(seed:samplingRate:)` directly, and other consumers (Trace, Session Replay) call `sessionSampler.combined(with: childRate).sample()`.

DatadogSessionReplay/Tests/Recorder/RecordingCoordinatorTests.swift
qsellem: Done — `KnuthReplaySamplingTests` has been merged into `RecordingCoordinatorTests`, keeping only the two tests you flagged as valuable:
  `test_givenSameSessionID_samplingDecisionIsIdentical_acrossMultipleCalls` and `test_knownSessionUUID_matchesPrecomputedKnuthResult`. The standalone file has been removed.

DatadogTrace/Tests/TracingURLSessionHandlerTests.swift
qsellem: Done — `KnuthTraceSamplingTests` has been merged into `TracingURLSessionHandlerTests`. The standalone file has been removed.

DatadogRUM/Tests/RUMMonitor/Scopes/RUMSessionScopeTests.swift
qsellem: Done — `KnuthSessionSamplingTests` has been removed.

DatadogRUM/Tests/RUMMonitor/Scopes/RUMSessionScopeTests.swift
qsellem: Done — `RUMCoreContextSessionSampleRateTests` has been removed.

DatadogInternal/Sources/Models/RUM/RUMCoreContext.swift
qsellem: Done — `sessionSampleRate: SampleRate` has been replaced by `sessionSampler: DeterministicSampler` in `RUMCoreContext`. The sampler is constructed in RUM from the session UUID, keeping
  the sampling responsibility there. `DeterministicSampler` now exposes a `combined(with:)` method that reuses the same seed and composes the rates, so Trace and Session Replay simply
  call `sessionSampler.combined(with: childRate).sample()` — no UUID parsing, no algorithm duplication.

DatadogInternal/Sources/Utils/DeterministicSampler.swift
```diff
@@ -30,14 +32,62 @@ public struct DeterministicSampler: Sampling {
...
+    /// UUIDs are always sampled (fail-open). This is intentional — a broken UUID should
+    /// not silently drop data.
+    ///
+    /// - Parameters:
+    ///   - sessionID: A UUID string in the standard `xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx` format.
+    ///   - samplingRate: A percentage value between `0.0` and `100.0`.
+    public init(sessionID: String, samplingRate: SampleRate) {
+        let seed = sessionID.split(separator: "-").last.flatMap { UInt64($0, radix: 16) } ?? 0
+        self.init(seed: seed, samplingRate: samplingRate)
+    }
```
arroz: 1. ~~Is this algorithm common across Datadog products, or specific for iOS SDK? The reason I'm asking is, 12 hex chars are 48 bits, meaning 16 bits of the UInt64 will always be zero. This leads to a lower quality seed than using the full 64 bits. If we *have* to use this seed because all the other Datadog products use it, it is what it is. But if the seed is limited to iOS, there is no reason to not just use the last 64 bits of the UUID.~~ I found this is common to all platforms, [documented here](https://docs.google.com/document/d/13OtlbdUIZ7xc4IkY6uyk3TmLMtaMu_oWL1wADsABLB4/edit?tab=t.0), so ignore this item.
2. The way this is calculated in this particular implementation (generating a string, parsing the string and converting back to an integer) is extremely expensive, and it's not needed since the internal representation of a UUID is already an array of bytes. It's much faster to create the UInt64 directly from those bytes. The bytes can be obtained either by bridging to NSUUID and calling [getBytes()](https://developer.apple.com/documentation/foundation/nsuuid/getbytes(_:)) or a technique similar to what's [described here](https://gist.github.com/dagronf/bb1d42c5d28a25499c2e1aab9f60f6c6) in pure Swift. I tend to prefer the first implementation since it uses official APIs instead of the second which relies on an assumption of the internal representation, but I don't believe this representation will ever change since too many things depend on it.

DatadogInternal/Sources/Utils/DeterministicSampler.swift
```diff
@@ -10,17 +10,19 @@
 /// deterministic sampling based on a sampling rate `samplingRate`.
 ///
 /// Conforms to the `Sampling` protocol.
-public struct DeterministicSampler: Sampling {
+public struct DeterministicSampler: Sampling, Equatable {
```
arroz: Not a showstopper in any way, but let's start marking entities as `Sendable` as much as possible. The sampler is a good candidate for this.

DatadogLogs/Sources/Feature/MessageReceivers.swift
```diff
@@ -89,7 +89,7 @@ internal struct WebViewLogReceiver: FeatureMessageReceiver {
                 event[dateKey] = correctedTimestamp
             }
 
-            if let rum = context.additionalContext(ofType: RUMCoreContext.self) {
+            if let rum = context.additionalContext(ofType: RUMCoreContext.self), rum.sessionSampler.sample() {
                 event[LogEvent.Attributes.RUM.applicationID] = rum.applicationID
```
arroz, line 93: Do we want to exclude the appID from the attributes if the session isn't sampled? We can still enrich the event with it even we're not sampling.

DatadogLogs/Sources/RemoteLogger.swift
```diff
@@ -136,7 +136,7 @@ internal final class RemoteLogger: LoggerProtocol, Sendable {
             var internalAttributes: [String: Encodable] = [:]
 
             // When bundle with RUM is enabled, link RUM context (if available):
-            if self.rumContextIntegration, let rum = context.additionalContext(ofType: RUMCoreContext.self) {
+            if self.rumContextIntegration, let rum = context.additionalContext(ofType: RUMCoreContext.self), rum.sessionSampler.sample() {
                 internalAttributes[LogEvent.Attributes.RUM.applicationID] = rum.applicationID
```
arroz, line 140: Same as https://github.com/DataDog/dd-sdk-ios/pull/2714/changes#r2965084079

DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
```diff
@@ -249,7 +225,7 @@ internal final class URLSessionRUMResourcesHandler: DatadogURLSessionHandler, RU
 }
 
 extension DistributedTracing {
-    func modify(request: URLRequest, headerTypes: Set<DatadogInternal.TracingHeaderType>, rumSessionId: String?, userId: String?, accountId: String?) -> (URLRequest, TraceContext?, URLSessionHandlerCapturedState?) {
+    func modify(request: URLRequest, headerTypes: Set<DatadogInternal.TracingHeaderType>, networkContext: NetworkContext?) -> (URLRequest, TraceContext?, URLSessionHandlerCapturedState?) {
         let traceID = traceIDGenerator.generate()
```
arroz: This code changed with [my PR](https://github.com/DataDog/dd-sdk-ios/pull/2726) so a bit of merging and rethinking needs to be done here, based on our conversations about this.

DatadogInternal/Sources/Utils/DeterministicSampler.swift
```diff
@@ -30,14 +32,56 @@ public struct DeterministicSampler: Sampling {
...
+    /// **seed=0 fallback:** When UUID parsing fails, seed defaults to 0. The Knuth hash
+    /// of 0 is 0, which is always `<= threshold` for any `samplingRate > 0`, so malformed
+    /// UUIDs are always sampled (fail-open). This is intentional — a broken UUID should
+    /// not silently drop data.
+    ///
+    /// - Parameters:
+    ///   - uuid: A UUID string in the standard `xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx` format.
+    ///   - samplingRate: A percentage value between `0.0` and `100.0`.
+    public init(uuid: String, samplingRate: SampleRate) {
+        let seed = uuid.split(separator: "-").last.flatMap { UInt64($0, radix: 16) } ?? 0
```
arroz: We can have a much faster algorithm here:

```
let seed = withUnsafePointer(to: uuid) { uuidPointer in
    return uuidPointer.withMemoryRebound(to: UInt64.self, capacity: 2) { pointer in
        return UInt64(bigEndian: pointer.successor().pointee) & 0x0000FFFFFFFFFFFF
    }
}
```

All that mess translates to 3 assembly instructions:

```
    0x100000970 <+120>: ldr    x8, [x19, #0x8]
    0x100000974 <+124>: and    x8, x8, #0xffffffffffff0000
    0x100000978 <+128>: rev    x8, x8
```

Comparing both implementations in Release mode, parsing the string and manipulating is 6-7 orders of magnitude slower. I also tested on an Intel Mac just to be sure byte order is not an issue.

I don't think UUIDs will ever change size, but an assert can be added just in case: `assert(MemoryLayout<UUID>.size == 16)`.

DatadogInternal/Sources/Utils/DeterministicSampler.swift
arroz: Ignore this, updated comment about it below.

DatadogInternal/Sources/Utils/DeterministicSampler.swift
arroz: Forgot to say, but this avoids the UUID conversion to string since it seems all the places calling this init method have the UUID in binary form and have to call `uuidString` just to pass it down to this method.

DatadogLogs/Sources/Feature/MessageReceivers.swift
maxep: Could be yes, but I don't know the consequences on the UI. It's best to not change this behaviour for now.

DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
maxep: I rebased and my approach is that the parent trace sampling takes precedence on the session's one. But I am still not sure how to proceed in this case. The other option is to drop the entire trace based on the RUM decision, but I don't see how to do it..

DatadogInternal/Sources/Utils/DeterministicSampler.swift
maxep: You killed me with the asm instructions 😆  That's much better, thanks for the suggestion!

arroz: All the time and money I invest messing with 40-year old computers ought to be useful occasionally. 😅

DatadogInternal/Sources/Models/RUM/RUMPayloadMessages.swift
```diff
@@ -26,21 +28,36 @@ public struct RUMSessionState: Codable, Equatable {
...
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        sessionUUID = try container.decode(UUID.self, forKey: .sessionUUID)
+        isInitialSession = try container.decode(Bool.self, forKey: .isInitialSession)
+        hasTrackedAnyView = try container.decode(Bool.self, forKey: .hasTrackedAnyView)
+        didStartWithReplay = try container.decodeIfPresent(Bool.self, forKey: .didStartWithReplay)
+        // `isSampled` was introduced in a later SDK version. For crash reports persisted by older
+        // SDK versions, fall back to deriving the sampling state from `sessionUUID`: a null UUID
+        // indicated a rejected (non-sampled) session in the previous encoding.
+        isSampled = try container.decodeIfPresent(Bool.self, forKey: .isSampled) ?? (sessionUUID != UUID.dd.nullUUID)
```
arroz, line 59: If I understand correctly, we're moving away from the `nullUUID` magic value? If so (hope so!), can we mark `DatadogExtension.nullUUID` as `@available(*, deprecated, message: "…")`?

DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
```diff
@@ -250,12 +247,20 @@ internal final class URLSessionRUMResourcesHandler: DatadogURLSessionHandler, RU
...
 extension DistributedTracing {
-    func modify(request: URLRequest, headerTypes: Set<DatadogInternal.TracingHeaderType>, rumSessionId: String?, userId: String?, accountId: String?, activeSpanContext: ActiveSpanContext?) -> (URLRequest, TraceContext?, URLSessionHandlerCapturedState?) {
+    func modify(request: URLRequest, headerTypes: Set<DatadogInternal.TracingHeaderType>, networkContext: NetworkContext?) -> (URLRequest, TraceContext?, URLSessionHandlerCapturedState?) {
+        let activeSpanContext = networkContext?.activeSpanProvider?.activeSpanContext()
+        // When a RUM context is available, its `sessionSampler` is a `DeterministicSampler`
+        // seeded from the session ID. Calling `combined(with:)` composes the session rate
+        // with the tracing `samplingRate` while preserving the seed, so every resource in
+        // the same session receives a consistent sampling decision.
+        // When no RUM context exists, fall back to a random `Sampler`.
+        let sampler: Sampling = networkContext?.rumContext?.sessionSampler.combined(with: samplingRate)
```
arroz: Not a show stopper, and the performance gain is mostly irrelevant, but this could be either a `lazy var` or a closure than only gets executed if needed (if `activeSpanContext?.samplingPriority` is nil).

DatadogInternal/Sources/Models/RUM/RUMPayloadMessages.swift
maxep: Yes, we are moving away from this :) but the `nullUUID` is still used as a placeholder for `RUMApplicationScope` when no session is started.

PR #2716, opened 2026-02-25
RUM-14622 Multi-platform integration test target

maxep:
### What and why?

Add multi-platform support to the integration tests target. The integration tests previously had separate iOS and tvOS schemes/targets; this consolidates them into unified multi-platform schemes with iOS and tvOS platform support for now. visionOS and watchOS will be added in further iterations. 

### How?

- **Multi-platform integration tests target**: Consolidated iOS and tvOS `DatadogIntegrationTests` schemes into a single multi-platform scheme.
- **Skip tests on tvOS**: Added `#if !os(tvOS)` guards to skip tests currently failing on tvOS. These were not run on tvOS previously:
  - `RUMSessionStartInBackgroundTests`
  - `RUMSessionStartInForegroundTests`
  - `RUMSessionStopTests`
  - `RUMSessionTimeOutTests`
  - `RUMSessionTrackingTests`
  - `RUMSessionWithNoViewTests`

> [!NOTE]
> These tests were not actually associated with the `DatadogIntegrationTests tvOS` target and are actually failing on tvOS. They relate to the Application states, added to better cover the app state transitions. They are mostly failing due to missing pre-warming state on tvOS, or for different timing expectations.

### Review checklist


Datadog/Example/Debugging/DebugRUMSessionViewController.swift
```diff
@@ -309,17 +309,15 @@ private struct FormItemView: View {
             TextField(placeholder, text: $value)
                 .font(.system(size: 12))
                 .padding(4)
-                .background(Color(UIColor.secondarySystemFill))
```
barboraplasovska, line 312: Was this removal intentional ?

Datadog/IntegrationUnitTests/RUM/RUMSessionStartInBackgroundTests.swift
```diff
@@ -20,9 +20,13 @@ class RUMSessionStartInBackgroundTests: RUMSessionTestsBase {
...
     // MARK: - OS Prewarm Launch
 
+    @available(tvOS, unavailable)
     private var osPrewarmLaunch: AppRunner.ProcessLaunchType { .osPrewarm(processLaunchDate: processLaunchDate, runtimeLoadDate: runtimeLoadDate) }
 
     func testGivenOSPrewarmLaunch_whenNoEventIsTracked() throws {
+        #if os(tvOS)
+        throw XCTSkip("This test is not available on tvOS")
+        #else
         // Given
```
barboraplasovska, line 30: nit/ Could we maybe make this into a small helper function ? It is repeated a lot.

maxep: Not really as it uses macros, and I would like to keep it explicit

Datadog/Example/Debugging/DebugRUMSessionViewController.swift
maxep: Yes, the `secondarySystemFill` is not available on all platforms. We can remove it and rely on the default background

PR #2717, opened 2026-02-26
RUM-14763 Add OOB scroll and swipe action tracking

barboraplasovska:
### What and why?

Adds automatic scroll and swipe action tracking on UIScrollView-based components to match Android SDK behavior. Swizzles `UIScrollView.delegate` setter to wrap delegates with a proxy that intercepts scroll lifecycle callbacks, classifies gestures as scroll or swipe based on velocity at finger lift, and emits RUM start/stop action commands with direction attributes. Follows the existing `MethodSwizzler` pattern and activates when `uiKitActionsPredicate` is configured.

### Review checklist
- [ ] Add Objective-C interface for public APIs

### DIFF
--- a/DatadogRUM/Sources/Instrumentation/Actions/RUMScrollHandler.swift
+++ b/DatadogRUM/Sources/Instrumentation/Actions/RUMScrollHandler.swift
@@ -12,7 +12,7 @@ import DatadogInternal
 /// Handles scroll and swipe gesture detection on UIScrollView-based components.
 /// Tracks scroll lifecycle events, classifies gestures as scroll or swipe based on velocity,
 /// and generates RUM commands for the action pipeline.
-internal final class UIScrollViewScrollHandler: RUMCommandPublisher {
+internal final class RUMScrollHandler: UIScrollViewHandler {
     /// Velocity threshold (in points/second) to classify a gesture as a swipe vs. scroll.
     /// Gestures with velocity magnitude >= this value are classified as swipe.
     static let velocityThreshold: CGFloat = 500
@@ -55,10 +55,10 @@ internal final class UIScrollViewScrollHandler: RUMCommandPublisher {
         self.subscriber = subscriber
     }
 
-    // MARK: - Scroll Lifecycle
+    // MARK: - UIScrollViewHandler
 
     /// Called when the user begins dragging.
-    func scrollViewWillBeginDragging(_ scrollView: UIScrollView) {
+    func notify_scrollViewWillBeginDragging(_ scrollView: UIScrollView) {
         // Only track user-initiated scrolls
         guard scrollView.panGestureRecognizer.state == .began || scrollView.panGestureRecognizer.state == .changed else {
             return
@@ -96,7 +96,7 @@ internal final class UIScrollViewScrollHandler: RUMCommandPublisher {
 
     /// Called when dragging ends. Captures lift velocity while the gesture recognizer still has it.
     /// If no deceleration follows, finalizes the scroll immediately.
-    func scrollViewDidEndDragging(_ scrollView: UIScrollView, willDecelerate decelerate: Bool) {
+    func notify_scrollViewDidEndDragging(_ scrollView: UIScrollView, willDecelerate decelerate: Bool) {
         let id = ObjectIdentifier(scrollView)
 
         // Capture velocity now — after deceleration it will be zero
@@ -111,7 +111,7 @@ internal final class UIScrollViewScrollHandler: RUMCommandPublisher {
     }
 
     /// Called when deceleration completes after a fling. Finalize the gesture.
-    func scrollViewDidEndDecelerating(_ scrollView: UIScrollView) {
+    func notify_scrollViewDidEndDecelerating(_ scrollView: UIScrollView) {
         finalizeScroll(scrollView)
     }
 
--- a/DatadogRUM/Sources/Instrumentation/Actions/RUMScrollHandler.swift
+++ b/DatadogRUM/Sources/Instrumentation/Actions/RUMScrollHandler.swift
@@ -38,9 +38,6 @@ internal final class RUMScrollHandler: UIScrollViewHandler {
     /// Active scrolls keyed by scroll view identity.
     private var activeScrolls: [ObjectIdentifier: ScrollState] = [:]
 
-    /// Lock protecting `activeScrolls`.
-    private let lock = NSLock()
-
     init(
         dateProvider: DateProvider,
         predicate: UITouchRUMActionsPredicate
@@ -78,9 +75,7 @@ internal final class RUMScrollHandler: UIScrollViewHandler {
             actionName: action.name
         )
 
-        lock.lock()
         activeScrolls[ObjectIdentifier(scrollView)] = state
-        lock.unlock()
 
         let command = RUMStartUserActionCommand(
             time: state.startTime,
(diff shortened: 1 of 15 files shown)


DatadogRUM/Sources/Instrumentation/Actions/UIKit/UIScrollViewSwizzler.swift
```diff
@@ -0,0 +1,81 @@
...
+    class SetDelegate: MethodSwizzler<
+        @convention(c) (UIScrollView, Selector, UIScrollViewDelegate?) -> Void,
+        @convention(block) (UIScrollView, UIScrollViewDelegate?) -> Void
+    > {
+        private static let selector = #selector(setter: UIScrollView.delegate)
+        private let method: Method
+        private let handler: UIScrollViewHandler
+
+        /// Proxies keyed by scroll view identity to avoid double-wrapping.
+        private var proxies: [ObjectIdentifier: UIScrollViewDelegateProxy] = [:]
```
gonzalezreal: ⚠️ The proxies dictionary holds strong references to `UIScrollViewDelegateProxy` instances keyed by `ObjectIdentifier(scrollView)`, but entries are never removed. As scroll views are created and destroyed, this dictionary grows unboundedly. The `ObjectIdentifier` keys become stale (or reused) but the proxy values remain alive.

I understand strong retention is necessary, but the current approach ties the proxy's lifetime to the swizzler, not the scroll view.

One potential solution is using `objc_setAssociatedObject` to attach the proxy directly to the scroll view. When the scroll view deallocates, the proxy goes with it automatically. This eliminates the dictionary, the lock, and the leak.

DatadogRUM/Sources/Instrumentation/Actions/RUMScrollHandler.swift
```diff
@@ -0,0 +1,186 @@
...
+
+    /// Called when deceleration completes after a fling. Finalize the gesture.
+    func notify_scrollViewDidEndDecelerating(_ scrollView: UIScrollView) {
+        finalizeScroll(scrollView)
+    }
+
+    /// Cleans up all active scroll states (e.g. on app backgrounding).
+    func cancelAll() {
+        lock.lock()
+        activeScrolls.removeAll()
```
gonzalezreal: ❓This removes all active scroll states without emitting `RUMStopUserActionCommand` for any of them. This means the RUM pipeline will see start actions without matching stops. Is this intentional?

DatadogRUM/Sources/Instrumentation/Actions/RUMScrollHandler.swift
```diff
@@ -0,0 +1,186 @@
...
+        lock.lock()
+        activeScrolls[ObjectIdentifier(scrollView)] = state
+        lock.unlock()
+
+        let command = RUMStartUserActionCommand(
+            time: state.startTime,
+            globalAttributes: [:],
+            attributes: action.attributes,
+            instrumentation: .uikit,
+            actionType: .scroll,
```
gonzalezreal, line 85: ❓The start command always uses `.scroll`, but the stop command may use `.swipe` (determined by velocity at drag end). This means the action type can change mid-lifecycle. Do we support that?

DatadogRUM/Tests/Instrumentation/Actions/RUMScrollHandlerTests.swift
```diff
@@ -0,0 +1,222 @@
...
+ */
+
+#if !os(tvOS)
+
+import XCTest
+import TestUtilities
+import DatadogInternal
+@testable import DatadogRUM
+
+class RUMScrollHandlerTests: XCTestCase {
```
gonzalezreal, line 14: Missing test coverage for a few areas:

- `UIScrollViewDelegateProxy` forwarding logic (`responds(to:)`, `forwardingTarget(for:)`). These are subtle to get right.
- `UIScrollViewSwizzler` (double-wrap prevention, proxy creation)
- Predicate returning `nil` (verifying no commands are emitted)
- `cancelAll()` behavior

DatadogRUM/Sources/Instrumentation/Actions/RUMScrollHandler.swift
```diff
@@ -0,0 +1,186 @@
...
+    private let dateProvider: DateProvider
+    private let predicate: UITouchRUMActionsPredicate
+
+    weak var subscriber: RUMCommandSubscriber?
+
+    /// Active scrolls keyed by scroll view identity.
+    private var activeScrolls: [ObjectIdentifier: ScrollState] = [:]
+
+    /// Lock protecting `activeScrolls`.
+    private let lock = NSLock()
```
gonzalezreal: AFAIK, all `UIScrollViewDelegate` callbacks are dispatched on the main thread. If `cancelAll()` is also main-thread-only, this lock is unnecessary.

If locking is needed, consider using `@ReadWriteLock` from DatadogInternal, it's the established pattern in this codebase and avoids the error-prone manual `lock()`/`unlock()` pairs.

gonzalezreal (changes requested):
Nice work! The architecture is clean and follows existing SDK patterns well.

I left a few comments and questions. The major concern is a memory leak in `UIScrollViewSwizzler.proxies`. I suggest attaching the proxy to the scroll view using `objc_setAssociatedObject` instead of maintaining a `proxies` dictionary.

Let me know your thoughts.

DatadogRUM/Sources/Instrumentation/Actions/UIKit/UIScrollViewSwizzler.swift
```diff
@@ -0,0 +1,81 @@
...
+
+                    // Don't re-wrap if the delegate is already our proxy
+                    if delegate is UIScrollViewDelegateProxy {
+                        previousImplementation(scrollView, Self.selector, delegate)
+                        return
+                    }
+
+                    let proxy = UIScrollViewDelegateProxy(
+                        originalDelegate: delegate,
+                        handler: handler
```
gonzalezreal: A new proxy is allocated every time the delegate setter fires, which can happen frequently for the same scroll view. Consider reusing the existing proxy by updating its `originalDelegate` instead of reallocating.

Also, using `objc_setAssociatedObject` to store the proxy on the scroll view would tie its lifetime to the scroll view automatically, eliminating the proxies dictionary, its lock, and the memory leak from entries never being removed.

DatadogRUM/Sources/Instrumentation/Actions/RUMScrollHandler.swift
barboraplasovska: I removed it, it wasn't used anyway!

DatadogRUM/Sources/Instrumentation/Actions/RUMScrollHandler.swift
barboraplasovska: Since I removed `cancelAll()` this lock was also removed.

DatadogRUM/Sources/Instrumentation/Actions/RUMScrollHandler.swift
barboraplasovska: The action is only sent on the `stop` command so we could theoretically change the action type, we just have to make it mutable.

gonzalezreal (dismissed):
Looks great! All feedback addressed cleanly.

barboraplasovska:
Had to update with `develop` to fix the Snapshots CI.

gonzalezreal (approved):

cdn34dd (approved):

PR #2718, opened 2026-02-26
RUM-14622 Enable visionOS testing

maxep:
### What and why?

Enable running unit and integration tests on visionOS simulator and officially support visionOS by removing the "not officially supported" warning.

### How?

- Added `test-visionos` and `test-visionos-all` make targets (mirroring the existing iOS/tvOS targets).
- Removed the visionOS "not officially supported" warning from `Datadog.swift`.
- Extended `BatteryStatusPublisher` and `ApplicationStatePublisher` compilation to visionOS (previously iOS only).
- Fixed a `DatadogCore` condition to correctly include visionOS for battery status and low-power mode subscriptions.
- Updated `isUIAlertTextField` in `UIKitExtensions.swift` to match the correct private class on visionOS (`_UIAlertControllerTextFieldView` instead of `_UIAlertControllerTextField`), since visionOS wraps text fields inside a `UICollectionView`.
- Updated `UIKitExtensionsTests` to account for visionOS rendering differences: `.actionSheet` style is excluded, text field count assertions are skipped where unreliable, and `buttonCountOffset` accounts for visionOS 26 behavior.

### Review checklist

### DIFF
--- a/DatadogCore/Sources/Core/DatadogCore.swift
+++ b/DatadogCore/Sources/Core/DatadogCore.swift
@@ -548,7 +548,7 @@ extension DatadogContextProvider {
         subscribe(\.carrierInfo, to: CarrierInfoPublisher())
         #endif
 
-        #if (os(iOS) || os(watchOS)) && !targetEnvironment(simulator)
+        #if (os(iOS) || os(visionOS)) && !targetEnvironment(simulator)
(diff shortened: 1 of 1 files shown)


simaoseica-dd (approved):
👌

PR #2719, opened 2026-02-27
FFL-1879 Add FlagsClientState and state management to DatadogFlags

leoromanovsky:
#### Motivation

When a network call to fetch flag assignments fails and the SDK falls back to serving cached flags from disk, there is currently no way for consumers to know they are receiving stale data. [The OpenFeature spec defines a STALE provider status](https://openfeature.dev/specification/types/#provider-status) for exactly this case, and [the Android SDK (`dd-sdk-android`) already implements it](https://github.com/DataDog/dd-sdk-android/blob/develop/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/model/FlagsClientState.kt). This PR brings parity to iOS.

#### Changes

- Added a `FlagsClientState` enum with five states: `notReady`, `ready`, `reconciling`, `stale`, and `error`
- Added a `FlagsStateManager` that tracks the current state and notifies registered listeners on changes (thread-safe, weak references to avoid retain cycles)
- Added `FlagsStateObservable` and `FlagsStateListener` protocols so consumers can observe state
- Modified `FlagsRepository.setEvaluationContext()` to transition through states: it goes to `reconciling` before the network call, then to `ready` on success, `stale` on failure when cached flags exist, or `error` on failure with no cache
- Exposed the `state` property on `FlagsClientProtocol` and `FlagsClient`
- The fallback (no-op) client starts in `error` state
- Added tests for the state manager and all state transition scenarios in the repository

#### Decisions

- Mirrored the Android SDK's `FlagsClientState` enum and `FlagsStateManager` pattern for cross-platform consistency
- Used weak references for listeners to avoid retain cycles, with automatic cleanup of deallocated listeners
- The `stale` vs `error` decision is based on whether `state != nil` (had cached flags) before the network call, matching Android's `hadFlags` check
- `reset()` transitions back to `notReady`
- Renamed `FlagsRepositoryMock.state` to `flagsData` to avoid collision with the new `state` concept on the protocol

#### Merge order

This PR must merge **first**. It adds the `FlagsClientProtocol.state` API that the OpenFeature provider depends on.

**Dependency chain:**
1. **This PR** (dd-sdk-ios) — merge and release
2. **DataDog/dd-openfeature-provider-swift#21** — wires up `observe()` to emit `ProviderEvent.stale` using the new `state` API. Can only merge after this PR is in a released dd-sdk-ios version (its `Package.swift` pins dd-sdk-ios at `3.2.0+`).

#### E2E local testing (`dd-sdk-ios` <-> [`dd-openfeature-provider-swift`](https://github.com/DataDog/dd-openfeature-provider-swift/pull/21) <-> `ToyApp`)

In a toy app, I subscribed to the publisher via `observe`
```
provider.observe() // Returns AnyPublisher<ProviderEvent?, Never>
    .sink { event in // Subscribe and handle each emitted value
        if let event = event {
            // This runs every time an event is published
            print("🚩 [OpenFeature] Provider event: \(event)")
        }
    }
    .store(in: &cancellables) // Keep the subscription alive
```

And then observed the emitted event in the console logs.

Screenshot 2026-03-18 at 11 12 06 AM

### Review checklist
- [ ] ~Add Objective-C interface for public APIs - see our [guidelines](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3157787243/RFC+-+Modular+Objective-C+Interface#Recommended-solution) (internal)~ Did not do because `DatadogFlags` is Swift-only


leoromanovsky (commented):
Nice PR — clean design, good test coverage, and the state machine matches Android well. Two comments:

**1. Consider `NSLock` instead of `NSRecursiveLock` in `FlagsStateManager`**

`updateState` notifies listeners while holding the lock. Since `FlagsStateListener` is a public protocol, an external consumer could (against the documented contract) call `setEvaluationContext` from inside `flagsStateDidChange`, which re-enters `updateState`. With `NSRecursiveLock` this silently delivers out-of-order notifications to subsequent listeners. With a regular `NSLock`, it deadlocks — crash-obvious in development rather than silently wrong in production.

The docs already say listeners should be "fast and non-blocking," so this just enforces that contract.

**2. Document state-before-completion ordering in `FlagsRepository`**

In `setEvaluationContext`, the state manager is updated before `completion` is called:

```swift
self?.stateManager.updateState(.stale)
// ...
completion(.failure(error))
```

The OpenFeature provider in dd-openfeature-provider-swift#21 depends on this — it checks `state.currentState` inside the completion handler to decide whether to throw or swallow the error. Worth a brief comment so a future refactor doesn't accidentally swap the order:

```swift
// State must be updated before calling completion —
// consumers (e.g., OpenFeature provider) check currentState in the callback.
```

leoromanovsky (commented):
Nice PR — clean design, good test coverage, and the state machine matches Android well. Two comments:

**1. Consider `NSLock` instead of `NSRecursiveLock` in `FlagsStateManager`**

`updateState` notifies listeners while holding the lock. Since `FlagsStateListener` is a public protocol, an external consumer could (against the documented contract) call `setEvaluationContext` from inside `flagsStateDidChange`, which re-enters `updateState`. With `NSRecursiveLock` this silently delivers out-of-order notifications to subsequent listeners. With a regular `NSLock`, it deadlocks — crash-obvious in development rather than silently wrong in production.

The docs already say listeners should be "fast and non-blocking," so this just enforces that contract.

**2. Document state-before-completion ordering in `FlagsRepository`**

In `setEvaluationContext`, the state manager is updated before `completion` is called:

```swift
self?.stateManager.updateState(.stale)
// ...
completion(.failure(error))
```

The OpenFeature provider in dd-openfeature-provider-swift#21 depends on this — it checks `state.currentState` inside the completion handler to decide whether to throw or swallow the error. Worth a brief comment so a future refactor doesn't accidentally swap the order:

```swift
// State must be updated before calling completion —
// consumers (e.g., OpenFeature provider) check currentState in the callback.
```

DatadogFlags/Sources/Client/FlagsClientProtocol.swift
```diff
@@ -13,6 +13,12 @@ import DatadogInternal
 /// with detailed evaluation information. The protocol is extended with convenience methods for
 /// common flag types.
 public protocol FlagsClientProtocol: AnyObject {
+    /// An observable for tracking client state changes.
+    ///
+    /// Use this property to observe state transitions such as ``FlagsClientState/ready``,
+    /// ``FlagsClientState/stale``, or ``FlagsClientState/error``.
+    var state: FlagsStateObservable { get }
```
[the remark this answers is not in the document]
sameerank: Given that `val state: StateObservable` was added to the Android SDK in the [3.4.0 release](https://github.com/DataDog/dd-sdk-android/blob/release/3.4.0/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsClient.kt#L169) and was missing in the [3.3.0 release](https://github.com/DataDog/dd-sdk-android/blob/release/3.4.0/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsClient.kt#L169), the Android SDK appears to have been okay with this breaking change to any `FlagsClient` interface conformer

And so this change to any `FlagsClientProtocol` conformer would be following in similar footsteps

DatadogFlags/Sources/Client/FlagsRepository.swift
```diff
@@ -80,55 +85,65 @@ internal final class FlagsRepository {
...
     func flagAssignments() -> [String: FlagAssignment]? {
         waitForFlagsDataRead()
-        return state?.flags
+        return flagsData?.flags
     }
 
     func setEvaluationContext(
         _ context: FlagsEvaluationContext,
         completion: @escaping (Result<Void, FlagsError>) -> Void
     ) {
```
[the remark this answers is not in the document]
sameerank: This suggestion is already implemented

https://github.com/DataDog/dd-sdk-ios/blob/fef4d5fde32f697ec03a3ae391175545a661ab89/DatadogFlags/Sources/Client/FlagsRepository.swift#L114-L115

Plus a race condition test 

https://github.com/DataDog/dd-sdk-ios/blob/fef4d5fde32f697ec03a3ae391175545a661ab89/DatadogFlags/Tests/Client/FlagsRepositoryTests.swift#L257

And a concurrency test

https://github.com/DataDog/dd-sdk-ios/blob/0b873705f847238632ef7c4bcab7564d14da431a/DatadogFlags/Tests/Client/FlagsStateManagerTests.swift#L117

DatadogFlags/Sources/Client/FlagsStateManager.swift
```diff
@@ -0,0 +1,96 @@
...
+            lock.unlock()
+            return
+        }
+        _state = newState
+        let listeners = _listeners
+        lock.unlock()
+
+        for weakListener in listeners {
+            weakListener.value?.flagsStateDidChange(newState)
+        }
```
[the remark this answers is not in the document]
sameerank: This statement "`updateState` releases the lock before notifying" is incorrect. The `defer` statement schedules the unlock to run when the function returns, which is after the notification at

https://github.com/DataDog/dd-sdk-ios/blob/0b873705f847238632ef7c4bcab7564d14da431a/DatadogFlags/Sources/Client/FlagsStateManager.swift#L65

Moreover this test

https://github.com/DataDog/dd-sdk-ios/blob/0b873705f847238632ef7c4bcab7564d14da431a/DatadogFlags/Tests/Client/FlagsStateManagerTests.swift#L117

validates this by firing 1k concurrent updates and verifying no duplicate consecutive states are observed.

Datadog/Datadog.xcodeproj/project.pbxproj
```diff
@@ -1862,7 +1867,7 @@
 		D2C1A50129C4C4CB00946C31 /* DDSpanContext.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61C5A87E24509A0C00DA608C /* DDSpanContext.swift */; };
 		D2C1A50229C4C4CB00946C31 /* Casting.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61C5A87C24509A0C00DA608C /* Casting.swift */; };
 		D2C1A50329C4C4CB00946C31 /* DDFormat.swift in Sources */ = {isa = PBXBuildFile; fileRef = D2EBEDD629B8F08E00B15732 /* DDFormat.swift */; };
-		D2C1A50429C4C4CB00946C31 /* (null) in Sources */ = {isa = PBXBuildFile; };
+		D2C1A50429C4C4CB00946C31 /* BuildFile in Sources */ = {isa = PBXBuildFile; };
 		D2C1A50529C4C4CB00946C31 /* DatadogTracer.swift in Sources */ = {isa = PBXBuildFile; fileRef = D2546BF029AF4F550054E00B /* DatadogTracer.swift */; };
```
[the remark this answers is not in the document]
sameerank: This diff was not relevant, so I reverted it in https://github.com/DataDog/dd-sdk-ios/pull/2719/changes/475c51bc438e64de6301aa685c21139181be72ca

Datadog/Datadog.xcodeproj/project.pbxproj
```diff
@@ -1905,7 +1910,7 @@
 		D2C1A54629C4F2DF00946C31 /* OTConstants.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61E909EB24A24DD3005EA2DE /* OTConstants.swift */; };
 		D2C1A54729C4F2DF00946C31 /* DDSpanContext.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61C5A87E24509A0C00DA608C /* DDSpanContext.swift */; };
 		D2C1A54829C4F2DF00946C31 /* OTTracer.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61E909E924A24DD3005EA2DE /* OTTracer.swift */; };
-		D2C1A54929C4F2DF00946C31 /* (null) in Sources */ = {isa = PBXBuildFile; };
+		D2C1A54929C4F2DF00946C31 /* BuildFile in Sources */ = {isa = PBXBuildFile; };
 		D2C1A54A29C4F2DF00946C31 /* DDSpan.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61C5A87824509A0C00DA608C /* DDSpan.swift */; };
```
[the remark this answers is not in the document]
sameerank: This diff was not relevant, so I reverted it in https://github.com/DataDog/dd-sdk-ios/pull/2719/changes/475c51bc438e64de6301aa685c21139181be72ca

DatadogFlags/Sources/Client/FlagsStateManager.swift
```diff
@@ -0,0 +1,96 @@
...
+/// - Important: Listeners are notified while the lock is held. Calling any
+///   `FlagsStateManager` method from within a listener callback will deadlock.
+///   This enforces the documented contract that listeners should be fast and non-blocking.
+internal final class FlagsStateManager: FlagsStateObservable {
+    /// Guards both `_state` and `_listeners` to ensure atomic state + notify.
+    /// Uses `NSLock` (not `NSRecursiveLock`) to fail fast on re-entrant calls.
+    private let lock = NSLock()
+
+    private var _state: FlagsClientState = .notReady
+    private var _listeners: [WeakListener] = []
```
sameerank: There is a way to get around defining the `WeakListener` wrapper if we do
```
private let _listeners = NSHashTable<AnyObject>.weakObjects()
```
for automatic `nil` cleanup, but casting to `AnyObject` is less type-safe.

I am open to other ideas here

DatadogFlags/Sources/Client/FlagsRepository.swift
```diff
@@ -80,55 +85,71 @@ internal final class FlagsRepository {
...
                 )
                 self.writeState()
+                self.stateManager.updateState(.ready)
                 completion(.success(()))
             case .failure(let error):
+                // State must be updated before calling completion —
+                // dd-openfeature-provider-swift checks currentState in the callback.
+                // Only use cached flags if they match the requested context to avoid
+                // serving flags from a different user/context.
+                let cachedContextMatches = self?.flagsData?.context == context
```
sameerank: I noticed a similar issue in Android, or at least what I considered to be an issue: https://github.com/DataDog/dd-sdk-android/pull/3224

I assume User B should never see the flags meant for User A via cached flags

| Cached Flags | Cached Context | Requested Context | State |
| --- | --- | --- | --- |
| None | - | User B | `Error` |
| Exist | User A | User A | `Stale` |
| Exist | User A | User B | `Error` |

DatadogFlags/Sources/Client/FlagsClientState.swift
```diff
@@ -0,0 +1,36 @@
...
+///
+/// The state transitions follow this lifecycle:
+/// - Initial: ``notReady``
+/// - On `setEvaluationContext()`: ``notReady`` → ``reconciling`` → ``ready`` / ``stale`` / ``error``
+/// - On context change: ``ready`` or ``stale`` → ``reconciling`` → ``ready`` / ``stale`` / ``error``
+/// - On `reset()`: any → ``notReady``
+public enum FlagsClientState: Equatable {
+    /// The provider has not been initialized.
+    ///
+    /// This state occurs before `setEvaluationContext()` is called or after `reset()`.
```
sameerank, line 19: This differs from the Android SDK which doesn't have a `reset()` method. I adjusted the comment in the Android SDK in https://github.com/DataDog/dd-sdk-android/pull/3224 to look similar. The `notReady` behavior between iOS and Android are otherwise identical

| Scenario | iOS | Android |
| --- | --- | --- |
| Fresh install, `NotReady` state | `context` = `nil` → `PROVIDER_NOT_READY` | Same |
| App restart with cached data, `NotReady` state | `context` = cached → flags work | Same |
| After `reset()` | `context` = `nil` → `PROVIDER_NOT_READY` | N/A (no reset method) |

api-surface-swift
```diff
@@ -1300,6 +1302,18 @@ public protocol FlagsClientProtocol: AnyObject
...
+    case ready
+    case reconciling
+    case stale
+    case error
+public protocol FlagsStateListener: AnyObject
+    func flagsStateDidChange(_ newState: FlagsClientState)
+public protocol FlagsStateObservable: AnyObject
+    var currentState: FlagsClientState
+    func addListener(_ listener: FlagsStateListener)
+    func removeListener(_ listener: FlagsStateListener)
```
sameerank, line 1344: I reference all of these in https://github.com/DataDog/dd-openfeature-provider-swift/pull/21

DatadogFlags/Sources/Client/FlagsClientState.swift
```diff
@@ -0,0 +1,36 @@
...
+import Foundation
+
+/// Represents the current state of a ``FlagsClient``.
+///
+/// The state transitions follow this lifecycle:
+/// - Initial: ``notReady``
+/// - On `setEvaluationContext()`: ``notReady`` → ``reconciling`` → ``ready`` / ``stale`` / ``error``
+/// - On context change: ``ready`` or ``stale`` → ``reconciling`` → ``ready`` / ``stale`` / ``error``
+/// - On `reset()`: any → ``notReady``
+public enum FlagsClientState: Equatable {
```
gonzalezreal: Swift infers `Equatable` for enums without associated values, so the explicit conformance is redundant. More importantly, since this value will cross thread boundaries (listeners fire from internal SDK queues), consider adding explicit Sendable conformance. It's trivially satisfied and aligns with the SDK's direction

DatadogFlags/Sources/Client/FlagsClientState.swift
gonzalezreal, line 19: "Provider" is OpenFeature vocabulary. This type describes `FlagsClient` state, not the OpenFeature provider. Leaking the upstream consumer's terminology into the SDK's public API is confusing for consumers who don't use OpenFeature. Suggestion: `The client has not been initialized.`

DatadogFlags/Sources/Client/FlagsClientState.swift
```diff
@@ -0,0 +1,36 @@
...
+    /// The provider has not been initialized.
+    ///
+    /// This state occurs before `setEvaluationContext()` is called or after `reset()`.
+    case notReady
+
+    /// The client has successfully loaded flags and they are available for evaluation.
+    case ready
+
+    /// The client is currently fetching flags for a context change.
+    /// Cached flags may still be available for evaluation during this state.
```
gonzalezreal: This could be misleading on the first `setEvaluationContext` call when there's no cache. Suggestion: `Previously cached flags, if any, may still be available for evaluation during this state.`

DatadogFlags/Sources/Client/FlagsClientProtocol.swift
gonzalezreal, line 20: Adding a requirement to a public protocol is a source-breaking change for any external conformer (custom wrappers, test mocks outside this repo).

Provide a default via protocol extension to maintain backward compatibility:

```swift
extension FlagsClientProtocol {
  var state: FlagsStateObservable { FlagsStateManager() }
}
```

DatadogFlags/Sources/Client/FlagsRepository.swift
```diff
@@ -12,6 +12,8 @@ internal protocol FlagsRepositoryProtocol {
 
     var context: FlagsEvaluationContext? { get }
 
+    var stateManager: FlagsStateManager { get }
```
gonzalezreal: I see a couple of issues here:

- Concrete type in protocol. This exposes `FlagsStateManager` instead of `FlagsStateObservable`. The protocol should use the abstraction.

- Encapsulation violation. The repository should own its state transitions internally. Exposing the manager lets FlagsClient (and anyone else with access to the protocol) read and mutate state, since `FlagsStateManager.updateState` is internal. The client should only need a read-only `FlagsStateObservable`, not the manager itself.

DatadogFlags/Sources/Client/FlagsRepository.swift
```diff
@@ -47,17 +50,19 @@ internal final class FlagsRepository {
         clientName: String,
         flagAssignmentsFetcher: any FlagAssignmentsFetching,
         dateProvider: any DateProvider,
-        featureScope: any FeatureScope
+        featureScope: any FeatureScope,
+        stateManager: FlagsStateManager = FlagsStateManager()
```
gonzalezreal: If the repository encapsulates state management, this parameter shouldn't be needed (the repository should create its own manager). Injecting it couples the caller to the concrete type and lets tests (or production code) share a manager across repositories, which breaks the one-manager-per-client invariant.

DatadogFlags/Sources/Client/FlagsRepository.swift
```diff
@@ -80,55 +85,71 @@ internal final class FlagsRepository {
...
         waitForFlagsDataRead()
-        return state?.flags
+        return flagsData?.flags
     }
 
     func setEvaluationContext(
         _ context: FlagsEvaluationContext,
         completion: @escaping (Result<Void, FlagsError>) -> Void
     ) {
+        waitForFlagsDataRead()
```
gonzalezreal: This blocks the caller's thread with a semaphore wait (up to 0.1s). Since `setEvaluationContext` is called by consumers, potentially from the main thread, this adds another synchronous wait point. The SDK philosophy says "heavy work should be delegated to background threads" and "small footprint — minimize runtime performance impact."

Consider whether `hadFlags` determination can be folded into the existing async readState flow instead of adding a blocking wait.

DatadogFlags/Sources/Client/FlagsRepository.swift
```diff
@@ -80,55 +85,71 @@ internal final class FlagsRepository {
...
+                self.flagsData = .init(
                     flags: flags,
                     context: context,
                     date: self.dateProvider.now
                 )
                 self.writeState()
+                self.stateManager.updateState(.ready)
                 completion(.success(()))
             case .failure(let error):
+                // State must be updated before calling completion —
```
gonzalezreal: `hadFlags` is captured before the fetch, `cachedContextMatches` reads `flagsData` after the fetch. The asymmetry is safe because `flagsData` only mutates on success, but this assumption is implicit and fragile. A future refactor that updates `flagsData` earlier in the success path (or adds a clear-on-failure) would silently break stale-vs-error detection.

I suggest capturing both values at the same point:

```swift
waitForFlagsDataRead()
let hadFlags = flagsData != nil
let cachedContext = flagsData?.context
stateManager.updateState(.reconciling)

// ... in failure:
if hadFlags && cachedContext == context {
```

DatadogFlags/Sources/Client/FlagsRepository.swift
```diff
@@ -80,55 +85,71 @@ internal final class FlagsRepository {
...
                 completion(.failure(error))
             }
         }
     }
 
     func reset() {
-        state = nil
+        flagsData = nil
+        stateManager.updateState(.notReady)
         featureScope.flagsDataStore.removeFlagsData(forClientNamed: clientName)
```
gonzalezreal: `reset()` sets `flagsData = nil`, then updates state to `.notReady`, then removes from disk. If a listener reacts to `.notReady` synchronously and queries the data store, the disk data hasn't been removed yet.

Consider clearing disk first, then nil out memory, then update state. So by the time listeners are notified, all cleanup is complete.

DatadogFlags/Sources/Client/FallbackFlagsClient.swift
```diff
@@ -10,10 +10,14 @@ import DatadogInternal
 internal final class FallbackFlagsClient: FlagsClientProtocol {
     private let name: String
     private weak var core: (any DatadogCoreProtocol)?
+    private let _stateManager = FlagsStateManager()
```
gonzalezreal: FallbackFlagsClient is a no-op client returned when Flags isn't enabled. Allocating a real `FlagsStateManager` for a client that will never transition states is wasteful. The SDK already follows a NOP pattern elsewhere (`NOPExposureLogger`, `NOPEvaluationLogger`, `NOPRUMFlagEvaluationReporter`).

DatadogFlags/Tests/Client/FlagsClientMocks.swift
```diff
@@ -142,20 +142,21 @@ extension AnyValue: AnyMockable, RandomMockable {
 
 final class FlagsRepositoryMock: FlagsRepositoryProtocol {
     var clientName: String
-    var state: FlagsData?
+    var flagsData: FlagsData?
+    let stateManager = FlagsStateManager()
```
gonzalezreal: The mock instantiates a real `FlagsStateManager`, so tests that use `FlagsRepositoryMock` exercise real locking and listener notification code even when testing unrelated concerns. Use a mock or the static NOP observable to keep test concerns isolated.

DatadogFlags/Tests/Client/FlagsRepositoryTests.swift
```diff
@@ -121,4 +121,333 @@ final class FlagsRepositoryTests: XCTestCase {
...
+        XCTAssertEqual(stateInCompletion, .error)
+    }
+}
+
+// MARK: - Helpers
+
+/// A data store mock that dispatches callbacks asynchronously on a background queue,
+/// matching the production `FeatureDataStore` behavior. This enables testing race conditions
+/// where the disk read may not have completed before other operations begin.
+private final class AsyncDataStoreMock: DataStore {
```
gonzalezreal: Check if `TestUtilities` already provides a configurable `DataStore` mock with async dispatch options. If not, consider contributing this there. Other modules may need similar async timing control in tests.

CHANGELOG.md
gonzalezreal, line 1: Commit 11 added the feature entry and duplicated the [IMPROVEMENT] Rename DDRUMErrorEventErrorMeta... line in the 3.8.0 section. Commit 14 removed both when moving the feature entry to Unreleased. Verify the improvement line wasn't unintentionally removed from 3.8.0. It should still appear once.

gonzalezreal (changes requested):
Thanks for this PR. The goal is well-motivated and the context-mismatch safety check is a particularly thoughtful addition that was also pushed upstream to Android.

I have two major concerns and a few smaller ones.

### Using the SDK's existing observation patterns

The SDK already has a mature observer mechanism in `ValuePublisher<Value>` / `ValueObserver` (`DatadogRUM/Sources/Utils/ValuePublisher.swift`). It provides thread-safe observation via a concurrent DispatchQueue, weak references via type-erased AnyObserver, distinct-value filtering for Equatable types, and —critically — notifications outside the lock, which makes deadlocks impossible.

This PR introduces an entirely new pattern (`FlagsStateListener` / `FlagsStateObservable` / `FlagsStateManager`) with `NSLock`-based synchronization and listener notification while holding the lock. This diverges from the SDK's concurrency conventions (which use `@ReadWriteLock` or `DispatchQueue`) and introduces a deadlock hazard on a public API. Since `FlagsStateListener` is public, any consumer whose callback (directly or indirectly) calls back into the state manager will hang. The SDK's philosophy is `Zero Crashes`, a deadlock is effectively a crash for the consumer's app.

I'd recommend one of:

- Extract `ValuePublisher` to `DatadogInternal` and reuse it here. It already solves every problem this PR's `FlagsStateManager` tries to solve, with better safety guarantees.

- Alternatively, if you want to keep a simpler, self-contained solution: use @ReadWriteLock with a grouped struct and notify outside the lock:

```swift
private struct ManagerState {
    var clientState: FlagsClientState = .notReady
    var listeners: [WeakListener] = []
}

@ReadWriteLock
private var managerState = ManagerState()

func updateState(_ newState: FlagsClientState) {
    let listenersToNotify = _managerState.mutate { state -> [WeakListener] in
        guard newState != state.clientState else { return [] }
        state.clientState = newState
        return state.listeners
    }
    for weakListener in listenersToNotify {
        weakListener.value?.flagsStateDidChange(newState)
    }
}
```

This eliminates the deadlock risk, aligns with SDK conventions, and provides the same atomicity guarantees.

### Encapsulation of state management

`FlagsRepositoryProtocol` exposes `var stateManager: FlagsStateManager { get }`,  a concrete internal type. This means `FlagsClient` reaches into the repository to grab the manager and re-expose it as a public property. State management should be an internal concern of the repository:

- The protocol should expose var state: `FlagsStateObservable { get }` (read-only, using the public protocol type)
- The `FlagsStateManager` init parameter on `FlagsRepository` should be removed, the repository should create and own its manager
- `FallbackFlagsClient` should use a lightweight NOP observable

DatadogFlags/Sources/Client/FlagsStateManager.swift
```diff
@@ -0,0 +1,96 @@
...
+
+    func updateState(_ newState: FlagsClientState) {
+        lock.lock()
+        defer { lock.unlock() }
+        guard newState != _state else {
+            return
+        }
+        _state = newState
+        for weakListener in _listeners {
+            weakListener.value?.flagsStateDidChange(newState)
```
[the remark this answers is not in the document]
sameerank: https://github.com/DataDog/dd-sdk-ios/pull/2719/commits/0428cdb2c48a52cc27eb1ed1193118c59630cb0e

Replaced `NSLock`-based synchronization with `@ReadWriteLock` from `DatadogInternal`. The implementation now captures listeners under the lock, then notifies outside the lock to prevent deadlocks when listeners call back into the manager. Added a test (`testListenerCanCallBackIntoManagerWithoutDeadlock`) that verifies listeners can safely call `currentState` from within the callback.

DatadogFlags/Sources/Client/FlagsRepository.swift
sameerank: https://github.com/DataDog/dd-sdk-ios/commit/63d008a063be267697e0d472f56fe4da74995c4a

Changed `FlagsRepositoryProtocol` to expose `var state: FlagsStateObservable` instead of concrete `FlagsStateManager`. Removed `stateManager` init parameter from `FlagsRepository` - it now creates and owns its manager internally. Added `NOPStateObservable` for lightweight no-op clients, used in `FallbackFlagsClient` and `FlagsRepositoryMock`.

DatadogFlags/Sources/Client/FlagsClientState.swift
sameerank: https://github.com/DataDog/dd-sdk-ios/commit/2d5775a3d630443e85690c6dc5b4675252544621

Replaced explicit `Equatable` conformance with `Sendable`

DatadogFlags/Sources/Client/FlagsClientState.swift
sameerank: https://github.com/DataDog/dd-sdk-ios/commit/2d5775a3d630443e85690c6dc5b4675252544621

Changed "The provider has not been initialized" to "The client has not been initialized" in the `notReady` case documentation

DatadogFlags/Sources/Client/FlagsClientState.swift
sameerank: https://github.com/DataDog/dd-sdk-ios/commit/2d5775a3d630443e85690c6dc5b4675252544621

Changed "Cached flags may still be available" to "Previously cached flags, if any, may still be available" to clarify behavior on first call.

DatadogFlags/Sources/Client/FlagsClientProtocol.swift
sameerank: https://github.com/DataDog/dd-sdk-ios/commit/897a7f5128a2bdcba611814aaf171ddd5f222dbe

Added a default implementation of `state` via protocol extension that returns `NOPStateObservable(state: .notReady)`. External conformers of `FlagsClientProtocol` will now get this default behavior without needing to implement the property themselves, preserving source compatibility.

DatadogFlags/Sources/Client/FlagsRepository.swift
sameerank: https://github.com/DataDog/dd-sdk-ios/commit/63d008a063be267697e0d472f56fe4da74995c4a

Addressed with https://github.com/DataDog/dd-sdk-ios/pull/2719#discussion_r3018449580 — removed `stateManager` init parameter. `FlagsRepository` now creates and owns its manager internally.

DatadogFlags/Sources/Client/FlagsRepository.swift
sameerank: https://github.com/DataDog/dd-sdk-ios/commit/8cc5f41085b97d4d19ed156220304e2bc12a327f

I think we initially tried to stick with the pre-existing usage pattern with `waitForFlagsDataRead()`, but it is possible to fold `hadFlags` determination into the async flow if we take a hybrid approach: adding async callback chaining for `setEvaluationContext` and preserving semaphore blocking for sync getters.

**For `setEvaluationContext` (async, no blocking):**
https://github.com/DataDog/dd-sdk-ios/blob/8cc5f41085b97d4d19ed156220304e2bc12a327f/DatadogFlags/Sources/Client/FlagsRepository.swift#L115
- Added `DiskReadState` struct with `isComplete` flag and `pendingCallbacks` array
- Added `whenFlagsDataRead(_:)` method that queues callbacks if disk read is in-flight, executes immediately if complete
- `setEvaluationContext` chains its entire operation via `whenFlagsDataRead`, avoiding blocking per SDK philosophy

**For sync getters (blocking preserved):**
https://github.com/DataDog/dd-sdk-ios/blob/8cc5f41085b97d4d19ed156220304e2bc12a327f/DatadogFlags/Sources/Client/FlagsRepository.swift#L106
- `waitForFlagsDataRead()` with semaphore retained for `context`, `flagAssignment(for:)`, `flagAssignments()`
- These are synchronous APIs where callers explicitly request data and expect cached values if available
- Blocking (up to 0.1s) is acceptable here since the caller chose a synchronous API

DatadogFlags/Sources/Client/FlagsRepository.swift
sameerank: https://github.com/DataDog/dd-sdk-ios/commit/efa661614b9ecc32d96eb030b399af147f43c928

Now capturing both `hadFlags` and `cachedContext` at the same point to ensures both values reflect the same snapshot of state when determining whether to use `.stale` vs `.error` on failure

DatadogFlags/Sources/Client/FlagsRepository.swift
sameerank: https://github.com/DataDog/dd-sdk-ios/commit/efa661614b9ecc32d96eb030b399af147f43c928

Reordered `reset()` to: (1) clear disk first, (2) nil memory, (3) update state. Added a comment to explain the rationale for the ordering.

https://github.com/DataDog/dd-sdk-ios/blob/efa661614b9ecc32d96eb030b399af147f43c928/DatadogFlags/Sources/Client/FlagsRepository.swift#L151-L156

DatadogFlags/Sources/Client/FallbackFlagsClient.swift
sameerank: https://github.com/DataDog/dd-sdk-ios/commit/63d008a063be267697e0d472f56fe4da74995c4a

Addressed with https://github.com/DataDog/dd-sdk-ios/pull/2719#discussion_r3018449580 — `FallbackFlagsClient` now uses `NOPStateObservable(state: .error)` instead of a real `FlagsStateManager`.

DatadogFlags/Tests/Client/FlagsClientMocks.swift
sameerank: Addressed with https://github.com/DataDog/dd-sdk-ios/pull/2719#discussion_r3018449580 — `FlagsRepositoryMock` now uses `NOPStateObservable(state: .notReady)` instead of a real `FlagsStateManager`.

DatadogFlags/Tests/Client/FlagsRepositoryTests.swift
sameerank: https://github.com/DataDog/dd-sdk-ios/commit/62397a251af7a00fe442ce38ae4e511a18c82307

TestUtilities provides `DataStoreAsyncMock` at `TestUtilities/Sources/Mocks/DatadogInternal/DataStoreAsyncMock.swift`. Replaced the local `AsyncDataStoreMock` with the shared one from TestUtilities

CHANGELOG.md
sameerank: Yup, it only shows up once now (see line 24). The duplicate was a mistake during a rebase

https://github.com/DataDog/dd-sdk-ios/blob/4de651126bc6fdc5302580135de41d6c581183c3/CHANGELOG.md#L19-L29

DatadogFlags/Sources/Client/FlagsClientProtocol.swift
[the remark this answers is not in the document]
sameerank: Same issue as https://github.com/DataDog/dd-sdk-ios/pull/2719#discussion_r3018610896 which is addressed by adding a default protocol implementation

https://github.com/DataDog/dd-sdk-ios/commit/897a7f5128a2bdcba611814aaf171ddd5f222dbe

DatadogFlags/Sources/Client/FlagsRepository.swift
```diff
@@ -80,55 +85,71 @@ internal final class FlagsRepository {
...
-        return state?.flags
+        return flagsData?.flags
     }
 
     func setEvaluationContext(
         _ context: FlagsEvaluationContext,
         completion: @escaping (Result<Void, FlagsError>) -> Void
     ) {
+        waitForFlagsDataRead()
+        let hadFlags = flagsData != nil
```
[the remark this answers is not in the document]
sameerank: https://github.com/DataDog/dd-sdk-ios/commit/efa661614b9ecc32d96eb030b399af147f43c928

Addressed in https://github.com/DataDog/dd-sdk-ios/pull/2719#discussion_r3019105233. The `whenFlagsDataRead()` callback ensures disk read has completed before we capture `hadFlags` and `cachedContext`, preventing misclassification

sameerank:
Thank you for all the feedback! And apologies for taking time to get back to addressing everything. I believe it's ready for another look

sameerank:
@codex review

DatadogFlags/Sources/Client/FlagsRepository.swift
```diff
@@ -57,78 +71,140 @@ internal final class FlagsRepository {
...
     ) {
-        flagAssignmentsFetcher.flagAssignments(for: context) { [weak self] result in
-            switch result {
-            case .success(let flags):
-                guard let self else {
-                    completion(.failure(.clientNotInitialized))
-                    return
+        // Chain after disk read completes to ensure correct hadFlags determination
+        whenFlagsDataRead { [weak self] in
+            guard let self else {
```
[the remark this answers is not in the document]
sameerank: https://github.com/DataDog/dd-sdk-ios/commit/3eda988bb7280b4530d70f3d85761ccaaf479e7e

This feedback is valid. However, the issue is not in `FlagsRepository` (which assumes callbacks are always called and this assumption sounds right to me) but in `FlagsDataStore` which is not consistent about always calling the callback, e.g. when decoding fails. Added `callback(nil)` in the catch block to ensure the callback is always invoked.

DatadogFlags/Sources/Client/FlagsRepository.swift
```diff
@@ -57,78 +71,140 @@ internal final class FlagsRepository {
...
+                    completion(.success(()))
+                case .failure(let error):
+                    // State must be updated before calling completion —
+                    // dd-openfeature-provider-swift checks currentState in the callback.
+                    // Only use cached flags if they match the requested context to avoid
+                    // serving flags from a different user/context.
+                    if hadFlags && cachedContext == context {
+                        self?.stateManager.updateState(.stale)
+                    } else {
+                        self?.stateManager.updateState(.error)
```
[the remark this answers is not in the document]
sameerank: https://github.com/DataDog/dd-sdk-ios/commit/677523aeb40fac900741f83fe56e0f843a6b12f7

When we explicitly transition to `.error` because the cached context doesn't match the requested context, we now clear `flagsData` to prevent serving the wrong user's flags. This prevents the scenario where User B could receive User A's flags after a failed context switch.

sameerank:
@codex review

DatadogFlags/Sources/Client/FlagsRepository.swift
```diff
@@ -57,78 +71,144 @@ internal final class FlagsRepository {
...
         completion: @escaping (Result<Void, FlagsError>) -> Void
     ) {
-        flagAssignmentsFetcher.flagAssignments(for: context) { [weak self] result in
-            switch result {
-            case .success(let flags):
-                guard let self else {
-                    completion(.failure(.clientNotInitialized))
-                    return
+        // Chain after disk read completes to ensure correct hadFlags determination
+        whenFlagsDataRead { [weak self] in
```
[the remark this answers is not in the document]
sameerank: https://github.com/DataDog/dd-sdk-ios/commit/2a86ce5efb079c86c6b5886ff11b6f6339cd3d37

This is similar to https://github.com/DataDog/dd-sdk-ios/pull/2719#discussion_r3020413758. The code comment below implies that async callbacks are invoked, but the internal `NOPDataStore` goes against this

https://github.com/DataDog/dd-sdk-ios/blob/ba9ec64d9772f8f2415d3c171fa727d87f51fe1e/DatadogInternal/Sources/DataStore/DataStore.swift#L54

but I chose not to fix it directly in DatadogInternal/Sources/DataStore/DataStore.swift to avoid impacting other modules. Instead, I added a safe wrapper in `FlagsDataStore` with a timeout fallback that guarantees the callback is invoked.

sameerank:
@codex review

DatadogFlags/Sources/Client/FlagsRepository.swift
```diff
@@ -57,78 +71,144 @@ internal final class FlagsRepository {
...
+            guard let self else {
+                completion(.failure(.clientNotInitialized))
+                return
+            }
+
+            let hadFlags = self.flagsData != nil
+            let cachedContext = self.flagsData?.context
```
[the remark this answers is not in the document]
sameerank: https://github.com/DataDog/dd-sdk-ios/commit/109601bcf158dce6d220dd9fa491a0b8c1f40ba0

This is a bit of an edge case. Added a test `testOverlappingContextUpdates_laterSuccessShouldNotBeClearedByEarlierFailure` to reproduce this issue and verify the fix:
1. Request A starts, captures `hadFlags = false`
2. Request B succeeds, writes flags for user-B
3. Request A fails, clears `flagsData` → wipes out user-B's flags
4. Result: `flagsData = nil`, `context = nil` — user-B's successful fetch is lost

And added a version counter to detect when a newer request has succeeded while an older request was in-flight. The failure path now checks if the version changed and skips clearing data if so.

DatadogFlags/Sources/Client/FlagsDataStore.swift
```diff
@@ -36,6 +42,43 @@ internal struct FlagsDataStore {
...
+
+        // Call the underlying data store
+        featureScope.dataStore.value(forKey: key) { result in
+            timeoutWorkItem.cancel()
+            var shouldInvoke = false
+            callbackInvoked.mutate { invoked in
+                if !invoked {
+                    invoked = true
+                    shouldInvoke = true
+                }
```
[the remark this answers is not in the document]
sameerank: This is a deliberate tradeoff from fixing the `NOPDataStore` hang issue: https://github.com/DataDog/dd-sdk-ios/pull/2719#discussion_r3020572332. The timeout prevents indefinite hangs when the SDK isn't properly initialized. I chose to prioritize not blocking over potentially treating slow reads (>100ms) as cache misses. 100ms is generous for disk I/O, and the fallback state (`.error`) is recoverable on the next call.

Also, an indefinite hang could be seen as falling into the same category as a code crash, which goes against the SDK philosophy

https://github.com/DataDog/dd-sdk-ios/blob/20d2733fac489760eb2dfc9f06fe02380322e1ed/ZEN.md#L25

DatadogFlags/Sources/Client/FlagsRepository.swift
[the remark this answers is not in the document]
sameerank: Simplified in https://github.com/DataDog/dd-sdk-ios/pull/2719/commits/755eb236440e667b1d903f4c52667a377c970296 - replaced the timeout workaround with a direct type check for `NOPDataStore`. Same safety guarantee (callbacks always invoked), but eliminates the 100ms delay and complex synchronization logic.

DatadogFlags/Sources/Client/FlagsRepository.swift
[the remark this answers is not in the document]
sameerank: Simplified in https://github.com/DataDog/dd-sdk-ios/pull/2719/commits/755eb236440e667b1d903f4c52667a377c970296 - replaced the timeout workaround with a direct type check for `NOPDataStore`. Same safety guarantee (callbacks always invoked), but eliminates the 100ms delay and complex synchronization logic.

DatadogFlags/Sources/Client/FlagsDataStore.swift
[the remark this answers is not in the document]
sameerank: Resolved in https://github.com/DataDog/dd-sdk-ios/commit/755eb236440e667b1d903f4c52667a377c970296 - replaced the timeout workaround with a direct type check for `NOPDataStore`. The new approach:

- Immediately returns `.noValue` for `NOPDataStore` (no timeout delay)
- Trusts real `DataStore` implementations to always call their callbacks
- Eliminates the risk of dropping valid late-arriving reads since there's no timeout race

This addresses the concern about losing cached flags while still preventing indefinite hangs

DatadogFlags/Sources/Client/FlagsDataStore.swift
```diff
@@ -36,10 +38,24 @@ internal struct FlagsDataStore {
...
+    /// `NOPDataStore` never invokes callbacks, so we short-circuit for that case.
+    private func safeDataStoreValue(forKey key: String, callback: @escaping (DataStoreValueResult) -> Void) {
+        let dataStore = featureScope.dataStore
+        if dataStore is NOPDataStore {
+            callback(.noValue)
+            return
+        }
+        dataStore.value(forKey: key, callback: callback)
+    }
+
```
gonzalezreal, line 58: Why not change `NOPDataStore` to invoke the callback(s) instead with `.noValue`?
Once you do that, this `safeDataStoreValue` wrapper can be removed entirely, and you can go back to calling `featureScope.dataStore.value(forKey:)` directly.

DatadogFlags/Sources/Client/FlagsDataStore.swift
```diff
@@ -36,10 +38,24 @@ internal struct FlagsDataStore {
             } catch let error {
                 DD.logger.error("Failed to decode \(FlagsData.self) from Flags Data Store", error: error)
                 featureScope.telemetry.error("Failed to decode \(FlagsData.self) from Flags Data Store", error: error)
+                callback(nil)
```
gonzalezreal, line 41: Good catch.

gonzalezreal (commented):
The feature itself is sound, the state machine is correct, and the tests are thorough. All previous feedback has been addressed.

One remaining issue: `FlagsDataStore.safeDataStoreValue` works around `NOPDataStore` not calling its callback by doing a type-check. The proper fix is a one-liner in `NOPDataStore` itself (see inline comment). Please fix it at the source and remove the workaround.

The branch is behind develop and has conflicts. Since the last rebase, @maxep merged significant watchOS / visionOS work (#2809, #2817, #2830) that refactored `project.pbxproj` heavily. I see two options to move forward:

1. Merge develop into this branch and resolve the pbxproj conflicts. For the `pbxproj`, take `develop`'s version and re-add your files to the correct build phases.
2. Start fresh from a new branch off `develop`, using this PR's final diff as reference, and incorporate the `NOPDataStore` fix directly. This avoids the conflict resolution and produces a cleaner history.

DatadogFlags/Sources/Client/FlagsDataStore.swift
sameerank: I hesitated for two reasons:
1. Invoking the callback felt like departing from the expected no-op contract.
2. `NOPDataStore` is shared infra and so adjusting it would impact other modules just to make it work for flags.

But I hadn't thought to use the `.noValue` argument, and that sounds like the right approach!

https://github.com/DataDog/dd-sdk-ios/pull/2719/commits/33419b2a501b23faae828fecba154c80e0eb4622

sameerank:
I went with this approach

> 1. Merge develop into this branch and resolve the pbxproj conflicts. For the `pbxproj`, take `develop`'s version and re-add your files to the correct build phases.

To confirm `git diff origin/develop c5cc47a5720c9c0c5aa6c5fcf5894723ebd26ebe -- Datadog/Datadog.xcodeproj/project.pbxproj` returns nothing

I also confirmed that the file hashes match after the merge commit
```
% git show origin/develop:Datadog/Datadog.xcodeproj/project.pbxproj | shasum -a 256
8bcfd75581cda50036c8ba34e0404dbd27fad4e874d799a68e667e91ea44263d  -

% git show c5cc47a57:Datadog/Datadog.xcodeproj/project.pbxproj | shasum -a 256
8bcfd75581cda50036c8ba34e0404dbd27fad4e874d799a68e667e91ea44263d  -
```

And then I re-added the following files in Xcode
1. `DatadogFlags/Sources/Client/FlagsStateManager.swift` to `DatadogFlags` target
2. `DatadogFlags/Sources/Client/FlagsClientState.swift` to `DatadogFlags` target
3. `DatadogFlags/Tests/Client/FlagsStateManagerTests.swift` to `DatadogFlagsTests` target

DatadogFlags/Sources/Client/FlagsClientProtocol.swift
```diff
@@ -71,6 +77,17 @@ public protocol FlagsClientProtocol: AnyObject {
...
 
+extension FlagsClientProtocol {
+    /// Default state observable for backward compatibility with external conformers.
+    ///
+    /// External implementations of ``FlagsClientProtocol`` that don't provide their own
+    /// state management will receive this default implementation, which returns a no-op
+    /// observable in the ``FlagsClientState/notReady`` state.
+    public var state: FlagsStateObservable {
+        NOPStateObservable(state: .notReady)
+    }
```
gonzalezreal, line 88: This default returns a brand-new `NOPStateObservable` on every read. That means `client.state === client.state` is `false`, and a listener registered via `client.state.addListener(...)` is held by an instance that is thrown away on the next line, so a later `client.state.removeListener(...)` runs on a different object.

The simplest fix is to cache the NOP as a `static` singleton on `NOPStateObservable` and return that here (ie. `NOPStateObservable.notReady`). `FallbackFlagsClient.state` can use the same singleton instead of allocating its own.

gonzalezreal (commented):
LGTM! Both earlier review threads are fully addressed.

One small nit left, inline: the default `state` in the protocol extension allocates a fresh `NOPStateObservable` on every read. Happy to approve once that is cached.

DatadogFlags/Sources/Client/FlagsClientProtocol.swift
sameerank: Thanks agreed using singletons makes sense here. Updated in https://github.com/DataDog/dd-sdk-ios/pull/2719/commits/b9addd0293518990eb5abf2f27fd5a495f8760d5

- Added `static let notReady` and `static let error` singletons in `NOPStateObservable` and made `init` private to enforce that external code cannot create new instances
- And replaced `NOPStateObservable` instantiations with singleton references

gonzalezreal (approved):

PR #2723, opened 2026-03-03
Use feature flag variant `key` instead of `value` for RUM Feature Flag Tracking

petzel:
### What does this PR do?
This PR updates how we Datadog Feature Flags are tracked in RUM, using a variant's `key` rather than the variant's `value`.

### Motivation
Previously, when using Datadog Feature Flags, the `value` would be set as the evaluated variant. This is not ideal, particularly when using JSON type flags. JSON flags can be very large, and are not friendly from a UX perspective when typing filters/viewing results. Each variant in a Datadog Feature Flag contains a `key`, which is a string with a maximum length of 64 characters. This `key` is better suited for tracking feature flags in RUM. 

### Review checklist

### DIFF
--- a/DatadogFlags/Sources/Client/FlagsClient.swift
+++ b/DatadogFlags/Sources/Client/FlagsClient.swift
@@ -251,12 +251,12 @@ extension FlagsClient: FlagsClientProtocol {
             reason: flagAssignment.reason
         )
 
-        trackEvaluation(key: key, assignment: flagAssignment, value: value, context: context)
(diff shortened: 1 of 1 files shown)


DatadogFlags/Sources/Client/FlagsClient.swift
simaoseica-dd: private func trackEvaluation(key: String, assignment: FlagAssignment, context: FlagsEvaluationContext)

PR #2725, opened 2026-03-04
RUM-14880: Add clang-tidy linting for C++ sources

simaoseica-dd:
### What and why?

This PR introduces `clang-tidy` linting for the C++ profiler code under `DatadogProfiling/Mach`. 
The goal is to catch potential bugs early and enforce a consistent style across the C++ codebase.

### How?

- Adds a `.clang-tidy` configuration at the repo root, tailored for low-level profiler code with a C ABI boundary.
- Adds a `run-cpp-linter.sh` script and a `make lint-cpp` target in the `Makefile`.
- Adds a build phase to the `DatadogProfiling` Xcode targets.
- Fixes issues reported by the linter.

> [!NOTE]
>  You can automatically fix linting issues by running:
> `./tools/lint/run-cpp-linter.sh --fix`

### Review checklist

### DIFF
--- a/tools/lint/run-cpp-linter.sh
+++ b/tools/lint/run-cpp-linter.sh
@@ -14,9 +14,17 @@ while :; do
     shift
 done
 
-CLANG_TIDY="$(brew --prefix llvm 2>/dev/null)/bin/clang-tidy"
-if [ ! -x "$CLANG_TIDY" ]; then
-	echo "error: clang-tidy not found. Install with: brew install llvm"; exit 1
+# Find clang-tidy in common Homebrew locations (Apple Silicon and Intel)
+CLANG_TIDY=""
+for prefix in /opt/homebrew /usr/local; do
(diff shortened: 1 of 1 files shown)


.clang-tidy
fuzzybinary, line 1: @awforsythe This was copied from the C++ SDK, 

Can take a quick look and point out if any of these should be re-enabled because they were only disabled because some of the specific needs of the C++ SDK?

awforsythe: The C++ code used in `dd-sdk-ios` is similarly low-level in style, so I don't see any issue with the current set of `.clang-tidy` rules. Some of them are just a matter of style preference (e.g. a lot of `readability` rules feel very arbitrary); the rest are mostly just things that you'd inevitably have to suppress when working at the low level of abstraction required in profiling code. LGTM!

awforsythe (approved):

PR #2726, opened 2026-03-04
RUM-14843: Set active span ID as RUM resource parent span ID.

arroz:
### What and why?

Currently, when `URLSession` requests are instrumented by the RUM feature (through the use of the `RUM.Configuration.urlSessionTracking` configuration option) to generate a RUM Resource, that resource is augmented with a trace and span ID, both generated on the fly, and later used by the backend to synthesize an APM trace span from the RUM resource using those IDs. This is not ideal since that span will be detached from any other trace, breaking the continuity of the generated distributed trace.

This PR links both features: if there is a currently active span for the relevant activity, RUM will use it for augmenting the RUM resource. Specifically:

- The active span's trace ID will be used as the RUM resource trace ID;
- The active span's ID will be used as the RUM resource parent span ID;
- The active span sampling information (priority and decision mechanism) will be used for the RUM resource.
- The RUM resource span ID will still be created on the fly and propagated through the request headers, since any child spans created by the customer's backend tracers and agents need to be children of that ID.

The use of the active span's priority and decision mechanism implies the `sampleRate` part of the `FirstPartyHostsTracing` argument of `urlSessionTracking` will be ignored if there is an active span. As an example:

```
RUM.enable(
  with: RUM.Configuration(
    applicationID: appID,
    sessionSampleRate: 20,
    (…)
    urlSessionTracking: .init(
      firstPartyHostsTracing: .trace(
          hosts: ["example.com"],
          sampleRate: 10,  // 👈 This is ignored if there is an active span. It's still used otherwise.
          traceControlInjection: .sampled),
        resourceAttributesProvider: nil)
   )
)
``` 

As a reminder, even if the active span is not sampled, the RUM resource itself will be sampled if the RUM session is sampled. The only thing the active span sampling controls is if Datadog backend generates the APM trace span out of the RUM resource, or not.

### How?

Trace feature initialization adds an instance of `ActiveSpanProvider` as additional context to the core. This entity contains a provider function that is used to dynamically obtain the active span.

The RUM feature creates an `ActiveSpanProviderReceiver` and registers it as a feature message receiver. When the context is updated, this receiver obtains the `ActiveSpanProvider` from the context and stores it. The instance of `ActiveSpanProviderReceiver` is also placed on the `DistributedTracing` instance, allowing `URLSessionRUMResourcesHandler` to finally obtain the active span and properly augment the RUM resource with it.

### Notes

- This is an "always on" feature as long as both RUM and Trace are enabled. A configuration option to turn this off was considered, but there is not an obvious use case where the users would want to disable this feature that justifies the added complexity of yet another configuration option.
- The [W3C trace context headers](https://www.w3.org/TR/trace-context/#traceparent-header) have a `parent-id` key. However, this key expects the RUM resource span ID, not its parent, the active span ID. This is confusing since the other header formats (Datadog and B3) both have explicit fields for the span and parent span IDs, implying the meaning of "parent" ID is different in the W3C case. This, however, makes sense, since the RUM resource span ID will be used as the parent span for any span created in the customers' backends.
- The [function that adds an additional context to the core](https://github.com/DataDog/dd-sdk-ios/pull/2726/changes#diff-540556d7d9c50e6f278da4d5050e571be559749d1c20e2dba47861c5897333e8R84) runs asynchronously, so there's a small chance the context was not yet updated when RUM starts instrumenting requests. The probability of this happening is rather low, but this is something that cannot be avoided with the current architecture. 

### Review checklist

### DIFF
--- a/DatadogRUM/Tests/Instrumentation/Resources/URLSessionRUMResourcesHandlerTests.swift
+++ b/DatadogRUM/Tests/Instrumentation/Resources/URLSessionRUMResourcesHandlerTests.swift
@@ -84,7 +84,7 @@ class URLSessionRUMResourcesHandlerTests: XCTestCase {
                     applicationID: .mockRandom(),
                     sessionID: "abcdef01-2345-6789-abcd-ef0123456789"
                 ),
-                activeSpanProvider: activeSpanProvider,
+                activeSpanProvider: activeSpanProvider
             )
         )
 


gonzalezreal (dismissed):
👌🏽Great work! I think the approach is clean and fits well with the existing architecture.

As a suggestion, consider adding an integration test (in this PR or a follow-up) that enables both RUM and Trace, starts an active span, fires a request within that span's scope, and asserts the RUM resource carries the active span's trace/span IDs and sampling priority. The unit tests cover the logic, but an integration test would validate the full wiring end-to-end, from `Trace.enable()` registering the `ActiveSpanProviderAdditionalContext`, through `ActiveSpanProviderReceiver` picking it up, to `DistributedTracing` using it. The existing infrastructure (`HTTPServerMock`, `RUMSessionMatcher`, `SpanMatcher`) should make this possible.

DatadogRUM/Sources/Feature/RUMFeature.swift
```diff
@@ -23,6 +23,12 @@ internal final class RUMFeature: DatadogRemoteFeature {
 
     let anonymousIdentifierManager: AnonymousIdentifierManaging
 
+    /// If Trace is enabled, this will contain a ``DatadogInternal/ActiveSpanProvider``.
+    ///
+    /// If RUM is running with URLSession tracking, this will be used to obtain the active span and tracer IDs that are
+    /// then used to link RUM Resources to the parent spans.
+    let activeSpanProviderReceiver: ActiveSpanProviderReceiver
```
simaoseica-dd: Can you group the receivers with `CombinedFeatureMessageReceiver` to follow the `DatadogFeature` protocol? This is the pattern when having multiple receivers

DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
```diff
@@ -250,8 +254,15 @@ internal final class URLSessionRUMResourcesHandler: DatadogURLSessionHandler, RU
 
 extension DistributedTracing {
     func modify(request: URLRequest, headerTypes: Set<DatadogInternal.TracingHeaderType>, rumSessionId: String?, userId: String?, accountId: String?) -> (URLRequest, TraceContext?, URLSessionHandlerCapturedState?) {
-        let traceID = traceIDGenerator.generate()
+        // If there is an active trace span, we get the active span and trace ID on traceInfo.
+        let activeSpanContext = activeSpanProviderContainer.activeSpanProvider?.activeSpanContext()
+
+        // In case there is, we use the same traceID so the backend can link the span generated from the RUM resource
+        // with the trace.
+        let traceID = activeSpanContext.map { $0.traceID } ?? traceIDGenerator.generate()
```
simaoseica-dd:
```suggestion
        let traceID = activeSpanContext?.traceID ?? traceIDGenerator.generate()
```

DatadogRUM/Sources/Integrations/ActiveSpanProviderReceiver.swift
```diff
@@ -0,0 +1,34 @@
...
+///
+/// These should be reference types, since we need to refer to them from several places in the code with the
+/// guarantee we're accessing the same instance.
+internal protocol ActiveSpanProviderContainer: Sendable, AnyObject {
+    var activeSpanProvider: ActiveSpanProvider? { get }
+}
+
+/// Handles messages with updated `DatadogCore` instances, and keeps the `ActiveSpanProvider` in a variable.
+/// See ``ActiveSpanProvider`` documentation for details on why this is needed.
+internal final class ActiveSpanProviderReceiver: FeatureMessageReceiver, ActiveSpanProviderContainer, @unchecked Sendable {
```
simaoseica-dd: This logic can be simplified. 
The access to the context should be simple and not repetitive in RUM. For the network calls, there is already the `NetworkContextCoreProvider` that can populate the `NetworkContext` with all the required trace context.

DatadogTrace/Sources/Trace.swift
```diff
@@ -80,5 +80,16 @@ public enum Trace {
             // to capture accurate timing from URLSessionTaskMetrics.
             try URLSessionInstrumentation.enableOrThrow(with: nil, in: core)
         }
+
+        core.set(context: ActiveSpanProviderAdditionalContext { [weak tracer = trace.tracer] in
```
simaoseica-dd: The context is usually grouping the module features context (`RUMCoreContext`,  `SessionReplayCoreContext`). Why not the same here with a unique `TraceContext`?

DatadogTrace/Sources/Trace.swift
```diff
@@ -80,5 +80,16 @@ public enum Trace {
...
+            tracer?.activeSpan?.context.dd.map {
```
simaoseica-dd, line 85: This way of collecting the context is different than the pattern for other contexts. This means that there will be always a need to construct an updated `ActiveSpanContext` instead of having the correct context in the source of truth `datadogcontext.additionalContext`. 
It is not an issue, but introducing new patterns will make the architecture more complex.

DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
```diff
@@ -261,14 +272,15 @@ extension DistributedTracing {
...
 
-        let sampler = sampler(sessionID: rumSessionId)
         let injectedSpanContext = TraceContext(
             traceID: traceID,
             spanID: spanID,
-            parentSpanID: nil,
+            // If there is an active span, use it as parent span for the span
+            // the backend creates out of the RUM resource:
+            parentSpanID: activeSpanContext?.activeSpanID,
             sampleRate: samplingRate,
```
simaoseica-dd: The `samplingRate` might be from Trace or RUM, right?

PR #2727, opened 2026-03-04
RUM-14825: Attach `rum-mobile-events.json` to app-launch profile uploads

simaoseica-dd:
### What and why?

This PR adds a new profiling attachment, `rum-mobile-events.json`, to app-launch profile uploads.
It carries TTID vital metadata so backend processing can correlate profiling samples with RUM vitals.

### How?

- Replaces `ProfilerStop` with `TTIDMessage`, which now includes both correlation context and `Vital` info.
- `RequestBuilder` now builds a `multipart/form-data` request with:
   - `event.json` (ProfileEvent)
   - `wall.pprof` (pprof file)
   - `rum-mobile-events.json` (RUMEvents info when available)
- The current `rum-mobile-events.json` has the format:
```json
{
  "vitals": [
      {
          "id": "abcd-efgh",
          "name": "time_to_initial_display",
          "start_ns": 1500000000,
          "duration_ns": 250000000
      }
  ]
}
```

### Review checklist


DatadogInternal/Sources/Models/RUM/TTIDMessage.swift
maxep, line 1: nit/ Both `TTIDMessage` and `Vital` are RUM concepts, I would move them to `RUM`'s models folder.

DatadogInternal/Sources/Models/RUM/Vital.swift
```diff
@@ -0,0 +1,29 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+public struct Vital: Codable, Equatable {
```
maxep, line 7: suggestion/ This is a very generic name and carries time info that are tied to TTID, I would suggest to define it as an inner type `TTIDMessage.Vital`.

simaoseica-dd: This struct will be generic to all vitals, including RUM operation vitals.

PR #2738, opened 2026-03-10
Fix #2568: Prevent UInt32 overflow crash in KronosNTPPacket.dateToNTPFormat

### LINKED ISSUE #2568: Crash due to Fatal error • KronosNTPPacket.dateToNTPFormat(_:)
### Stack trace

We have started to receive this crash from Datadog code with this message. `Fatal error: Double value cannot be converted to UInt32 because the result would be greater than UInt32.max`

The full crash report is as follows with some information redacted.

```
CrashReporter Key:  6685c0a40ba771116e66b298b9d504e9156eb9cc
Hardware Model:     iPhone14,7
OS Version:         iOS 26.1
Exception Type:     EXC_BREAKPOINT 
Exception Subtype:  KERN_INVALID_ADDRESS

Fatal error: Double value cannot be converted to UInt32 because the result would be greater than UInt32.max

0  libswiftCore.dylib +0x135bf8 _assertionFailure(_:_:file:line:flags:)
1  Glovo +0x1687acc             KronosNTPPacket.dateToNTPFormat(_:)
2  Glovo +0x1686dc0             KronosNTPPacket.prepareToSend(transmitTime:)
3  Glovo +0x1685324             closure #1 in KronosNTPClient.sendAsyncUDPQuery(to:port:timeout:completion:)
4  Glovo +0x16855c8             @objc closure #1 in KronosNTPClient.sendAsyncUDPQuery(to:port:timeout:completion:)
5  CoreFoundation +0xc8084      ___CFSocketPerformV0
6  CoreFoundation +0x68f20      ___CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE0_PERFORM_FUNCTION__
7  CoreFoundation +0x68e94      ___CFRunLoopDoSource0
8  CoreFoundation +0x46b2c      ___CFRunLoopDoSources0
9  CoreFoundation +0x1d6d4      ___CFRunLoopRun
10 CoreFoundation +0x1ca68      __CFRunLoopRunSpecificWithOptions
11 GraphicsServices +0x1494     _GSEventRunModal
12 UIKitCore +0x9dba0           -[UIApplication _run]
13 UIKitCore +0x46a74           _UIApplicationMain
14 Glovo +0x299f0               main (AppDelegate.swift:70:7)
15 dyld +0x4e24                 start
```

The crash seems to be happening in this [function](https://github.com/DataDog/dd-sdk-ios/blob/7677cc0335f0993ffdec911a73c20546a44e014e/DatadogCore/Sources/Kronos/KronosNTPPacket.swift#L160).

### Reproduction steps

I have not tired to reproduce this. But I believe the error message is quite clear on what could be the reproduction steps.

### Volume

For now less. But it is a recurrent crash for users at the start of app.

### Affected SDK versions

2.30.1

2.30.1

### Does the crash manifest in the latest SDK version?

No

### Deployment Target

Minimum Deployment Target is iOS 16

### Device Information

_No response_

_No response_

mariusc83:
## Issue

Fixes #2568: Crash due to Fatal error • KronosNTPPacket.dateToNTPFormat(_:)

**GitHub Issue:** https://github.com/DataDog/dd-sdk-ios/issues/2568

## Root Cause

`dateToNTPFormat()` converts a `TimeInterval` to `UInt32` without bounds checking. When the system clock is set far in the future (or returns a corrupted value), `time + kEpochDelta` exceeds `UInt32.max` (4,294,967,295), causing Swift's runtime to trigger a fatal error: *"Double value cannot be converted to UInt32 because the result would be greater than UInt32.max"*.

## Fix Strategy

Rather than clamping the value (which would silently send corrupted NTP timestamps and produce wrong time offsets), the fix makes `prepareToSend()` return `nil` when any timestamp cannot be represented in NTP format:

1. **`dateToNTPFormat`** now returns `UInt64?` — returns `nil` when `time + kEpochDelta` is outside `[0, UInt32.max]`
2. **`prepareToSend`** now returns `Data?` — validates all 4 timestamps upfront and returns `nil` if any overflow
3. **`KronosNTPClient`** handles `nil` by invalidating the socket, which causes the NTP query to gracefully time out. The app falls back to the system clock — the correct behavior when the clock is unreliable
4. **`intervalToNTPFormat`** uses safe clamping since interval values (rootDelay, rootDispersion) are small by NTP spec

### Files Changed

| File | Change |
|------|--------|
| `DatadogCore/Sources/Kronos/KronosNTPPacket.swift` | `dateToNTPFormat` returns `UInt64?`, `prepareToSend` returns `Data?` with upfront validation |
| `DatadogCore/Sources/Kronos/KronosNTPClient.swift` | Handle `nil` from `prepareToSend` by invalidating socket gracefully |
| `DatadogCore/Tests/Datadog/Kronos/KronosNTPPacketTests.swift` | 6 new tests for overflow, negative, extreme, boundary, and round-trip scenarios |

## Tests

- **Unit tests added:** 5 (far future → nil, negative → nil, extreme → nil, boundary → valid, existing testToData → still passes)
- **Integration tests added:** 1 (round-trip encode → parse with valid timestamp)
- **Module test suite:** PASS (9 tests in KronosNTPPacketTests)
- **TDD iterations:** 1/5
- **Backward compatibility:** Confirmed — existing `testToData` verifies exact byte output is unchanged for normal time values

### DIFF
--- a/DatadogCore/Sources/Kronos/KronosNTPClient.swift
+++ b/DatadogCore/Sources/Kronos/KronosNTPClient.swift
@@ -168,14 +168,14 @@ internal final class KronosNTPClient {
         let callback: CFSocketCallBack = { socket, callbackType, _, data, info in
             if callbackType == .writeCallBack {
                 var packet = KronosNTPPacket()
-                guard let PDU = packet.prepareToSend() as CFData? else {
+                guard let PDU = packet.prepareToSend() else {
                     // System clock is outside representable NTP range (e.g. far future).
                     // Skip sending to avoid corrupted time offsets — the client will
                     // fall back to the system clock.
                     CFSocketInvalidate(socket)
                     return
                 }
-                CFSocketSendData(socket, nil, PDU, kronosDefaultTimeout)
+                CFSocketSendData(socket, nil, PDU as CFData, kronosDefaultTimeout)
                 return
             }
 


DatadogCore/Sources/Kronos/KronosNTPClient.swift
```diff
@@ -168,7 +168,13 @@ internal final class KronosNTPClient {
...
                 var packet = KronosNTPPacket()
-                let PDU = packet.prepareToSend() as CFData
+                guard let PDU = packet.prepareToSend() as CFData? else {
+                    // System clock is outside representable NTP range (e.g. far future).
+                    // Skip sending to avoid corrupted time offsets — the client will
+                    // fall back to the system clock.
+                    CFSocketInvalidate(socket)
+                    return
+                }
                 CFSocketSendData(socket, nil, PDU, kronosDefaultTimeout)
```
maxep: This would be clearer IMO:
```suggestion
                guard let PDU = packet.prepareToSend() else {
                    // System clock is outside representable NTP range (e.g. far future).
                    // Skip sending to avoid corrupted time offsets — the client will
                    // fall back to the system clock.
                    CFSocketInvalidate(socket)
                    return
                }
CFSocketSendData(socket, nil, PDU as CFData, kronosDefaultTimeout)
```

mariusc83: Good call — applied in 7aedaeef4. The guard now binds the `Data?` directly and the `as CFData` cast is at the call site.

PR #2750, opened 2026-03-12
fix: prevent crash when intercepting AVAssetDownloadTask

mariedm:
### What and why?

Fixes a crash when the SDK intercepts an `AVAssetDownloadTask` or `AVAggregateAssetDownloadTask`. These subclasses inherit from `URLSessionTask` but declare most inherited properties (e.g., `currentRequest`, `response`) as unavailable. At runtime, accessing any of these throws an `NSGenericException`, crashing the app.

Our swizzling on `URLSessionTask.resume()` fires for all subclasses, so any app using `AVAssetDownloadTask` with our SDK would crash on `task.resume()`.

### How?

Detect `AVAssetDownloadTask` and `AVAggregateAssetDownloadTask` at `interceptResume` point, before any property is accessed, and return immediately. This is the same approach used by https://github.com/FLEXTool/FLEX/commit/f0b2dc971e84c96808e539cd3c1a6a3bd98ae377. 

### Review checklist

### DIFF
--- a/DatadogInternal/Tests/NetworkInstrumentation/NetworkInstrumentationFeatureTests.swift
+++ b/DatadogInternal/Tests/NetworkInstrumentation/NetworkInstrumentationFeatureTests.swift
@@ -1910,7 +1910,7 @@ class NetworkInstrumentationFeatureTests: XCTestCase {
         XCTAssertTrue(task.isSupportedForInstrumentation)
     }
 
-    @available(iOS 13.0, *)
+    @available(iOS 13.0, tvOS 13.0, *)
     func testIsSupportedForInstrumentation_returnsTrueForWebSocketTask() {
         let session = URLSession(configuration: .ephemeral)
         let task = session.webSocketTask(with: URL(string: "wss://example.com")!)
@@ -1943,7 +1943,7 @@ class NetworkInstrumentationFeatureTests: XCTestCase {
 
     // MARK: - Crash regression: resume() on various task types
 
-    @available(iOS 13.0, *)
+    @available(iOS 13.0, tvOS 13.0, *)
     func testWebSocketTask_resumeDoesNotCrash() throws {
         // Regression: verify that resuming a WebSocketTask with the swizzle installed doesn't crash.
         // The crash in interceptResume is synchronous, so no real connection is needed — we cancel immediately.


DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
```diff
@@ -54,6 +54,20 @@ internal final class NetworkInstrumentationFeature: DatadogFeature {
...
+    static let unsupportedTaskClasses: [AnyClass] = {
+        ["AVAssetDownloadTask", "AVAggregateAssetDownloadTask"]
+            .compactMap { NSClassFromString($0) }
+    }()
+
+    /// Returns `true` if `task` is an instance of a class that doesn't support standard
+    /// `URLSessionTask` properties and must be skipped by the instrumentation.
+    static func isUnsupportedTask(_ task: URLSessionTask) -> Bool {
+        unsupportedTaskClasses.contains { task.isKind(of: $0) }
+    }
```
maxep: cr/ These are public types so we can avoid using String comparison, better to use `(task is AVAssetDownloadTask) || (task is AVAggregateAssetDownloadTask)` I think.

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
```diff
@@ -54,6 +54,20 @@ internal final class NetworkInstrumentationFeature: DatadogFeature {
...
+    /// and throw `NSGenericException` at runtime when those properties are accessed.
+    /// Resolved once at startup using `NSClassFromString` to avoid importing AVFoundation.
+    static let unsupportedTaskClasses: [AnyClass] = {
+        ["AVAssetDownloadTask", "AVAggregateAssetDownloadTask"]
+            .compactMap { NSClassFromString($0) }
+    }()
+
+    /// Returns `true` if `task` is an instance of a class that doesn't support standard
+    /// `URLSessionTask` properties and must be skipped by the instrumentation.
+    static func isUnsupportedTask(_ task: URLSessionTask) -> Bool {
```
simaoseica-dd: nit: These funcs could be contextualized in an `URLSessionTask` extension. And it would improve the readability if it was `supported` instead of `unsupported` :
```swift
extension URLSessionTask {
    private static let unsupportedTaskClasses: [AnyClass] = {
        [
            "AVAssetDownloadTask",
            "AVAggregateAssetDownloadTask"
        ]
        .compactMap(NSClassFromString)
    }()

    var isSupportedForInstrumentation: Bool {
        !Self.unsupportedTaskClasses.contains { self.isKind(of: $0) }
    }
}
```

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
simaoseica-dd: Avoiding the `AVFoundation` import is a good point.  One of the pros can be to avoid `AVFoundation` to initialize static funcs lazily.

PR #2752, opened 2026-03-13
Fix flaky TTID tests by flushing all SDK queues

maxep:
### What and why?

The `testGivenUserSession_whenItIsStopped_andNextEventIsTrackedInBackground` and `testGivenUserSession_whenItTimesOut_andNextEventIsTrackedInBackground` tests were flaky: `XCTAssertNotNil(session.ttidEvent)` would intermittently fail because the TTID vital event write hadn't completed before results were collected.

### How?

`flushDatadogContext()` was only syncing `DatadogContextProvider.defaultQueue`, but the TTID write dispatches through two async hops: context queue → `readWriteQueue`. Under load (CI), the write could still be in-flight when results were read.

Fix: added `flush()` to `AppRunner` (delegates to `core.flush()` which drains all SDK queues in the correct order) and updated `flushDatadogContext()` to use it.

### Review checklist


simaoseica-dd (approved):
Thank you

PR #2794, opened 2026-03-26
RUM-15256 Make trace sampling deterministic based on the RUM session ID

arroz:
### What and why?

As part of improving sampling consistency across all products, Trace will now use a deterministic sampler whenever RUM is active and there is a session ID.

The Trace sampling rate can still be configured independently of RUM. However, we now have several guarantees regarding consistency. For example, for a given session ID, if the RUM session is sampled, and Trace sampling rate is the same or higher than RUM, Trace will be sampled as well. Conversely, if the RUM session is not sampled, and Trace sampling rate is equal or lower than RUM, then Trace wont be sampled as well.

### How?

The Trace context message receiver now keeps the deterministic sampler published by RUM and provides it to tracing as required. It also provides a method to generate a sampler with a custom sampling rate that is still consistent with the RUM sampler seed.

### Review checklist


DatadogTrace/Sources/Utils/SamplerBuilder.swift
```diff
@@ -0,0 +1,30 @@
...
+    ///
+    /// - parameters:
+    ///   - deterministicSampler: If a deterministic sampler provided by RUM exists, it must be
+    ///   passed in this parameter. Otherwise, pass `nil`.
+    ///   - samplingRate: The desired sampling rate. This can either be the value defined in the
+    ///   Trace feature configuration, or a custom value for a situation where a custom sampling rate
+    ///   was requested.
+    static func makeCurrentSamplerFor(deterministicSampler: DeterministicSampler?, using samplingRate: SampleRate) -> Sampling {
+        if let deterministicSampler {
+            return DeterministicSampler(seed: deterministicSampler.seed, samplingRate: samplingRate)
```
maxep: So we are not rebasing the sample rate, is that intended?
Using `deterministicSampler.combined(with: samplingRate)`, so the trace sample rate relates to the session's one.

TestUtilities/Sources/Mocks/DatadogTrace/TracingFeatureMocks.swift
```diff
@@ -390,3 +390,31 @@ public struct MockActiveSpanProvider: TraceActiveSpanProvider {
         storedActiveSpanContext
     }
 }
+
+public struct TestTracerSamplerProvider: TracerSamplerProvider {
```
maxep: To align with our mock convention:
```suggestion
public struct TracerSamplerProviderMock: TracerSamplerProvider {
```

DatadogTrace/Sources/Utils/SamplerBuilder.swift
```diff
@@ -0,0 +1,30 @@
...
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+import DatadogInternal
+
+/// Helper type containing function for creating samplers appropriate for the Trace feature.
+internal enum SamplerBuilder {
```
maxep: Not sure to see the value of this helper, its implementation is only used by `ContextMessageReceiver`. It could be a function there instead.

maxep (commented):
Looks good 👍  I left a question on the sample rate: When the deterministic sampling is applied, the effective sample rate is the composition of the configured sample rate and the RUM session sample rate. I think we should do the same here, right?

DatadogTrace/Sources/Utils/SamplerBuilder.swift
arroz: It's intended, I confirmed that with @qsellem. The logic is, Session Replay requires RUM so it's sort of a "sub-feature" of RUM, therefore the probability should be combined. Trace is independent, so we don't combine (for example, it's possible that Trace has a higher sampling rate than RUM, and that is expected).

DatadogTrace/Sources/Utils/SamplerBuilder.swift
arroz: Good point, I'm moving it there.

DatadogTrace/Sources/Utils/SamplerBuilder.swift
maxep: OK, great! thanks 👍

TestUtilities/Sources/Mocks/DatadogTrace/TracingFeatureMocks.swift
arroz: 👍

arroz:
Ignore previous merge command, clicked the wrong PR

simaoseica-dd (commented):
Great work!
This sampling logic will be used by other SDK features as well, so I'm thinking of extracting it from the context receivers.
WDYT about having a dedicated component responsible only for sampling decisions so the receivers don't have yet another responsibility?

arroz:
> Great work! This sampling logic will be used by other SDK features as well, so I'm thinking of extracting it from the context receivers. WDYT about having a dedicated component responsible only for sampling decisions so the receivers don't have yet another responsibility?

I believe that's the way forward for sure! This is going to be used in many places and it doesn't make to be writing the same code over and over in every feature.

One thing to keep in mind though: depending on the feature, and per business request, we handle the sampling priority differently. For example, despite Trace using the deterministic sampler to be consistent with RUM for a given session ID, the sampling rate is used without being combined with RUM's, since it's not a feature dependent on RUM. On other features, like Session Replay (SR), we combine the sampling rate of both SR and RUM (as in, 50% on Session Replay means 50% of the sampled RUM sessions, so it's SR * RUM), since those features can be looked at "sub-features" or RUM for this purpose.

So a global mechanism must have a flag indicating if the sampling rate is absolute or combined.

maxep (dismissed):

simaoseica-dd (approved):
Looks good! 🚀

PR #2797, opened 2026-03-27
[PANA-6692] Expose internal `addAction` with heatmap data for cross-platform SDKs

gonzalezreal:
### What and why?

Exposes an internal `addAction` method on `DatadogInternalInterface` that accepts `HeatmapAttributes`, enabling cross-platform SDKs (React Native) to pass heatmap data when reporting tap actions.

Also moves `HeatmapAttributes` from `DatadogRUM` to `DatadogInternal` so it can be consumed by cross-platform SDKs without a direct dependency on `DatadogRUM`.

### How?

- Moved `HeatmapAttributes` from `DatadogRUM/Sources/Heatmaps/` to `DatadogInternal/Sources/Heatmaps/` and made it `public`
- Replaced the memberwise initializer with a more ergonomic one accepting `HeatmapIdentifier`, `CGSize`, and `CGPoint`
- Added `addAction(at:type:name:heatmapAttributes:attributes:)` to `DatadogInternalInterface`, which creates a `RUMAddUserActionCommand` with the provided heatmap data
- Updated `UIEventCommandFactory` to use the new `HeatmapAttributes` initializer
- Updated API surface

### Review checklist


DatadogRUM/Sources/RUMMonitorProtocol+Internal.swift
```diff
@@ -163,4 +163,23 @@ public struct DatadogInternalInterface {
             )
         )
     }
+
+    public func addAction(
+        at time: Date,
+        type: RUMActionType,
+        name: String,
+        heatmapAttributes: HeatmapAttributes?,
+        attributes: [AttributeKey: AttributeValue] = [:]
```
simaoseica-dd, line 172: Can the attributes be merged in only one parameter?

DatadogInternal/Sources/Heatmaps/HeatmapAttributes.swift
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+import Foundation
+
+/// Heatmap attributes attached to a user action.
+public struct HeatmapAttributes {
+    /// The permanent identifier of the action target.
+    public let targetPermanentID: String
+    /// The width of the action target, in points.
+    public let targetWidth: Int64
```
simaoseica-dd, line 14: Can the CGSize and CGPoint be used here as well?

simaoseica-dd (approved):
LGTM. Just left 2 questions.

DatadogRUM/Sources/RUMMonitorProtocol+Internal.swift
gonzalezreal: These two parameters serve different purposes and shouldn't be merged. `heatmapAttributes` is typed, structured data that maps directly to the `_dd.action` schema fields in the RUM event. `attributes` is an open-ended dictionary for custom user data that ends up in `context`. Merging them would mean encoding `HeatmapAttributes` fields as untyped key-value pairs in a dictionary, which is exactly what this PR is trying to avoid.

DatadogInternal/Sources/Heatmaps/HeatmapAttributes.swift
gonzalezreal: The initial idea for `HeatmapAttributes` was to hold the values that will map directly to the `_dd.action` schema fields in the RUM event. The initializer is just a convenience to avoid duplicating the conversion logic.

maxep (approved):
LGTM 👍

PR #2799, opened 2026-03-27
[RUM-15318] Revert SR sampler to random sampling pending retention filter readiness

maxep:
### What and why?

Reverts the Session Replay sampler back to random sampling (`Sampler(samplingRate:).sample()`) with a TODO comment for RUM-15318.

The deterministic cross-product sampling (via `sessionSampler.combined(with:)`) was introduced ahead of the Session Replay retention filters being ready. Until those filters are in place, the deterministic approach is premature.

### How?

- Reverts `RecordingCoordinator.onRUMContextChanged` to use `Sampler(samplingRate: replaySampleRate).sample()` with a TODO comment referencing RUM-15318
- Replaces deterministic sampling tests with simpler tests matching the random sampling behaviour: decision is retained for the same session ID, and re-evaluated when the session ID changes

### Review checklist


maxep:
No longer needed

PR #2800, opened 2026-03-27
Fix Carthage rate limiting on raw.githubusercontent.com

maxep:
### What and why?

GitHub introduced rate limits for unauthenticated requests to `raw.githubusercontent.com`
(https://github.blog/changelog/2025-05-08-updated-rate-limits-for-unauthenticated-requests/).
Carthage fetches binary-only framework JSON manifests (e.g. `OpenTelemetryApi.json`) from
that host without authentication by default, causing frequent failures on shared CI runners
that exhaust the unauthenticated quota.

### How?

In `tools/carthage-shim.sh`, when running on CI:
1. Write the existing `GITHUB_ACCESS_TOKEN` to `~/.netrc` for `raw.githubusercontent.com`
2. Pass `--use-netrc` to Carthage so it adds an `Authorization` header when fetching binary JSON manifests
3. Clean up `~/.netrc` via a trap on exit

### Review checklist
- [ ] Add Objective-C interface for public APIs


arroz (approved):
🚢 it!

PR #2802, opened 2026-03-27
RUM-15482: Rename RUM Operations API

simaoseica-dd:
### What and why?

This PR renames the RUM Operations API for consistency and clarity, and introduces a new generic parameter `options: OperationOptions?` to make the API extensible and avoid coupling the public interface to a single feature (profiling).

  **Before:**
  ```swift
func startFeatureOperation(name: String,operationKey: String? = nil,attributes: [AttributeKey: AttributeValue] = [:])
func succeedFeatureOperation(name: String,operationKey: String? = nil,attributes: [AttributeKey: AttributeValue] = [:])
func failFeatureOperation(name: String,operationKey: String? = nil,reason: RUMFeatureOperationFailureReason,attributes: [AttributeKey: AttributeValue] = [:])
```

**After:**
```swift
func startOperation(name: String,operationKey: String?,attributes: [AttributeKey: AttributeValue],options: OperationOptions?)
func succeedOperation(name: String,operationKey: String?,attributes: [AttributeKey: AttributeValue])
func failOperation(name: String,operationKey: String?,reason: RUMFeatureOperationFailureReason,attributes: [AttributeKey: AttributeValue])
```

### How?

- Introduces `OperationOptions` protocol in `DatadogInternal` as the extensible options type for operations.
- Introduces `ProfilingOptions` (conforming to `OperationOptions`) to carry the profiling sample rate.
- Updates `RUMMonitorProtocol`, its default implementations, `Monitor`, and `NOPMonitor` accordingly.
- Updates API surface files to reference the new API naming

### Review checklist

### DIFF
--- a/DatadogRUM/Sources/RUMMonitorProtocol.swift
+++ b/DatadogRUM/Sources/RUMMonitorProtocol.swift
@@ -279,12 +279,37 @@ public protocol RUMMonitorProtocol: RUMMonitorViewProtocol, AnyObject {
 
     // MARK: - features
 
+    /// Starts a RUM Operation
+    /// - Parameters:
+    ///   - name: the name of the operation (e.g., `login_flow`)
+    ///   - operationKey: the key of the operation for this step (when running several instances of the same operation)
+    ///   - attributes: custom attributes to attach to this operation
+    @available(*, message: "This API is in preview and may change in future releases")
+    func startOperation(
+        name: String,
+        operationKey: String?,
+        attributes: [AttributeKey: AttributeValue]
+    )
+
     /// Starts a Feature Operation
     /// - Parameters:
     ///   - name: the name of the operation (e.g., `login_flow`)
     ///   - operationKey: the key of the operation for this step (when running several instances of the same operation)
     ///   - attributes: custom attributes to attach to this operation
+    @available(*, deprecated, renamed: "startOperation(name:operationKey:attributes:)", message: "Use startOperation(name:operationKey:attributes:) instead.")
     func startFeatureOperation(
+        name: String,
+        operationKey: String?,
+        attributes: [AttributeKey: AttributeValue]
+    )
+
+    /// Completes a RUM Operation successfully.
+    /// - Parameters:
+    ///   - name: the name of the operation (e.g., `login_flow`)
+    ///   - operationKey: the key of the operation for this step (when running several instances of the same operation); it should be provided if `operationKey` was provided when invoking `startOperation`
+    ///   - attributes: custom attributes to attach to this operation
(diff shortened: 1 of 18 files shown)


Valpertui:
Can we put deprecate+rename params on the methods to ease the transition for customers?   
We can just make the deprecated method just point to the new method.  

See this blog [post ](https://sarunw.com/posts/deprecate-old-api-in-swift/)
example
```
@available(*, deprecated, renamed: "newMethod")
func oldMethod() {

}

func newMethod() {
    
}
```
It makes Xcode auto provide "rename" callouts

This also works with more complex methods with multiple parameters

DatadogRUM/Sources/RUMMonitorProtocol.swift
```diff
@@ -279,34 +279,77 @@ public protocol RUMMonitorProtocol: RUMMonitorViewProtocol, AnyObject {
...
+    ///   - name: the name of the operation (e.g., `login_flow`)
+    ///   - operationKey: the key of the operation for this step (when running several instances of the same operation)
+    ///   - attributes: custom attributes to attach to this operation
+    ///   - options: options to attach to this operation (e.g. profiling options)
+    @available(*, message: "This API is in preview and may change in future releases")
+    func startOperation(
+        name: String,
+        operationKey: String?,
+        attributes: [AttributeKey: AttributeValue],
+        options: OperationOptions?
```
maxep, line 293: I like this more generatic `options` parameter 👍  But I wonder how we can combine options? Currently, only profiling defines options, but this design is meant to be extensible: How we would combine multiple options? such as `ProfilingOption`+`TraceOptions` for exemple?

maxep (dismissed):
Looks great! I left a question, but it is good to go as is.

DatadogRUM/Sources/RUMMonitorProtocol.swift
simaoseica-dd: We can have an entity that extends `OperationOptions` and stores an array of `OperationOptions`.
We will always need to cast the abstraction but after that, we can have some sugar funcs to help in the implementation.

```swift
struct CombinedOperationOptions: OperationOptions {
    let options: [OperationOptions]

    init(_ options: OperationOptions...) {
        self.options = options
    }

    var profilingOptions: ProfilingOptions? {
        options.first { $0 is ProfilingOptions } as? ProfilingOptions
    }
}

struct TraceOptions: OperationOptions {}

let options = CombinedOperationOptions(ProfilingOptions(sampleRate: 100), TraceOptions())
```

cdn34dd (dismissed):

maxep (dismissed):

maxep (approved):

PR #2803, opened 2026-03-30
Fixes profiler build in Xcode.

arroz:
Xcode 26.4 does not seem to be able to find the `module.modulemap` file recursively, so adjusting the path.


simaoseica-dd (approved):
thank you!

PR #2804, opened 2026-03-30
[RUM-10349]Add telemetry to track WebView instrumentation usage

saraSr5:
### What and why?
Adds trackWebView usage telemetry to WebViewTracking.enable(). We currently have no visibility into how many organizations use WebView tracking, since it is lazily initialized at runtime after the initial Configuration Telemetry is sent. This change fires a UsageTelemetry.Event.trackWebView event every time WebView tracking is successfully enabled, giving us the data needed to measure adoption.

### How?
 - Added case trackWebView to UsageTelemetry.Event in DatadogInternal                                                                                                                                               
 - Added the corresponding mapping in TelemetryReceiver (DatadogRUM) to route the event to .telemetryMobileFeaturesUsage(.trackWebView), matching the mobile usage schema
 - Called core.telemetry.send(telemetry: .usage(.init(event: .trackWebView, sampleRate: UsageTelemetry.defaultSampleRate))) at the end of WebViewTracking.enableOrThrow(), after both the NOPDatadogCore guard and  the duplicate-detection guard, so the event fires only on the first successful enable per controller     

### Review checklist

### DIFF
--- a/DatadogInternal/Sources/Telemetry/Telemetry.swift
+++ b/DatadogInternal/Sources/Telemetry/Telemetry.swift
@@ -120,6 +120,8 @@ public struct UsageTelemetry: SampledTelemetry {
         case addOperationStepVital(TelemetryUsageEvent.Telemetry.Usage.TelemetryCommonFeaturesUsage.AddOperationStepVital)
         /// GraphQL request detected
         case addGraphQLRequest
+        /// trackWebView API
+        case trackWebView
 
         /// Describes the properties of `addViewLoadingTime` usage telemetry.
         public struct ViewLoadingTime {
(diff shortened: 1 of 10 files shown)


DatadogInternal/Sources/Telemetry/Telemetry.swift
```diff
@@ -501,6 +503,11 @@ extension Telemetry {
     public func metric(name: String, attributes: [String: Encodable], sampleRate: SampleRate = MetricTelemetry.defaultSampleRate) {
         send(telemetry: .metric(MetricTelemetry(name: name, attributes: attributes, sampleRate: sampleRate)))
     }
+
+    /// Reports WebView tracking API usage.
+    public func trackWebView() {
+        send(telemetry: .usage(.init(event: .trackWebView, sampleRate: UsageTelemetry.defaultSampleRate)))
+    }
```
maxep, line 520: suggestion/ We can make a more generic method for all usage telemetry:
```swift
public func usage(event: UsageTelemetry.Event, sampleRate: SampleRate = UsageTelemetry.defaultSampleRate)
```

maxep (commented):
Looks great! I left a suggestion

maxep (approved):
LGTM 👍

PR #2807, opened 2026-03-31
RUM-15310: When an active span is not sampled, pass sampling decision to RUM URLSession handler

arroz:
### What and why?

[This PR](https://github.com/DataDog/dd-sdk-ios/pull/2726) changes the behavior of RUM `URLSession` sampling: if there is an active span (APM) when the request happens, the RUM resource span will be attached to it (be a child of it) and uses its sampling decision.

That causes reduction in the number of sampled resource spans compared to before, in the situation where the RUM session is sampled, but the APM span is not.

To fix this, when the span is not sampled, it should not impose its decision, and RUM should still decide based on its own URLSession sampling rate.

### How?

A simple bit of logic was added to `DistributedTracing.modify`: we only consider the active span if it's sampled. Otherwise, we fallback to regular sampling.

### Review checklist

### DIFF
--- a/Datadog/IntegrationUnitTests/RUM/RUMResourceTraceIntegrationTests.swift
+++ b/Datadog/IntegrationUnitTests/RUM/RUMResourceTraceIntegrationTests.swift
@@ -11,8 +11,6 @@ import TestUtilities
 @testable import DatadogRUM
 @testable import DatadogTrace
 
-private class InstrumentedSessionDelegate: NSObject, URLSessionDataDelegate {}
-
 class RUMResourceTraceIntegrationTests: RUMSessionTestsBase {
     private var core: DatadogCoreProxy! // swiftlint:disable:this implicitly_unwrapped_optional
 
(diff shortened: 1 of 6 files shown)


Datadog/IntegrationUnitTests/RUM/RUMResourceTraceIntegrationTests.swift
```diff
@@ -0,0 +1,192 @@
...
+                    sampleRate: urlSessionTrackingSampled ? 70 : 10
+                )
+            )
+        )
+        // This session ID is not sampled at 50%, but it is sampled at 60%.
+        config.uuidGenerator = RUMUUIDGeneratorMock(uuid: RUMUUID(rawValue: UUID(uuidString: "c5b3c4ab-fa4a-4de9-8199-a522131ec48a")!))
+        RUM.enable(with: config, in: core)
+
+        // We need to wait for all the messages between features to stabilize so we have everything setup correctly.
+        core.flush()
```
simaoseica-dd, line 129: Can this flush be removed? 
The stabilization won't happen in a real setup and we might be hiding issues.

arroz: Removing it breaks the tests. This is unavoidable: in a real setup we rely on enough time usually passing between the SDK setup and the first request, so the probability of everything being properly setup by then is high. Here, requests run immediately after the setup, so there's an almost 100% probability of tests breaking from what I saw. We usually don't see it when we test a single feature, but testing feature interaction is showing this problem quite often.

simaoseica-dd (approved):
LGTM!

PR #2809, opened 2026-04-01
[RUM-15258] Enable visionOS and watchOS in SPM and podspecs

maxep:
### What and why?

Enable visionOS and watchOS platform declarations in all podspecs and `Package.swift` files, so that the Datadog iOS SDK can be consumed via SPM or CocoaPods on those platforms.

### How?

- Added `visionos.deployment_target` and `watchos.deployment_target` entries to each podspec that didn't already have them.
- Updated `Package.swift` and `TestUtilities/Package.swift` to include visionOS and watchOS platform targets.

> [!IMPORTANT]
>  Smoke Tests are intentionally left out for the moment, we need to find a way to cover both platform without drastically increasing our CI pipeline duration.

### Review checklist

### DIFF
--- a/TestUtilities/Package.swift
+++ b/TestUtilities/Package.swift
@@ -1,10 +1,16 @@
-// swift-tools-version: 5.7.1
+// swift-tools-version: 5.10
 
 import PackageDescription
 
 let package = Package(
     name: "TestUtilities",
-    platforms: [.iOS(.v12), .tvOS(.v12), .macOS(.v12), .watchOS(.v7)],
+    platforms: [
+        .iOS(.v12),
+        .tvOS(.v12),
+        .macOS(.v12),
+        .watchOS(.v7),
+        .visionOS(.v1)
+    ],
     products: [
         .library(
             name: "TestUtilities",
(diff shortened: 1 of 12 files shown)


TestUtilities/Package.swift
```diff
@@ -1,10 +1,16 @@
-// swift-tools-version: 5.7.1
+// swift-tools-version: 5.10
```
barboraplasovska, line 1: /nit Could we bump to `5.9` instead ? (VisionOS seems to be supported from [there](github.com/swiftlang/swift-package-manager/blob/main/CHANGELOG.md).)

barboraplasovska (commented):
Great job! Just some nits and questions.

`TestUtilities/Package.swift` gets `.visionOS(.v1)` but `TestUtilities.podspec` isn't updated.                                                                                            
                                                                                                                                                                                                                     
Is `DatadogSessionReplay.podspec` intentionally excluded from visionOS and watchOS? If so maybe you could add a note in the PR description so reviewers don't flag it.                                                    
                  
Should we add a **CHANGELOG** entry here or will you do one before merging to develop?

maxep:
Thanks for the review, @barboraplasovska 🙏 
Session Replay is only available on iOS, no need for a note in this PR but it will be specified in the public doc 👍 
The CHANGELOG will be updated before merging the feature branch to `develop`.

barboraplasovska (approved):

PR #2816, opened 2026-04-02
[RUMS-5686] Reproduce: WebView blank screen when enable() called before SR init

mariusc83:
## Reproduction for RUMS-5686

**Jira:** [RUMS-5686](https://datadoghq.atlassian.net/browse/RUMS-5686)

### Issue Summary
When `WebViewTracking.enable()` is called before `SessionReplay.enable()`, the JS bridge capabilities are baked in as `[]` instead of `["records"]`, causing Session Replay to show a blank screen for WebViews.

### Reproduction Tests
- `DatadogWebViewTracking/Tests/WebViewTrackingRUMS5686Tests.swift` — 2 failing tests

### What the Tests Prove
1. `testItAddsUserScriptWithEmptyCapabilitiesWhenSessionReplayIsNotYetInitialized`: When `WebViewTracking.enable()` is called before SR is registered in the core, `getCapabilities()` returns `'[]'` instead of `'["records"]'`. The Browser SDK reads capabilities once at page load; receiving `[]` means it never forwards SR records across the bridge.

2. `testReEnablingWebViewTrackingAfterSRInitDoesNotUpdateCapabilities`: Once a webview is enabled with empty capabilities, there is no recovery path at runtime. Calling `WebViewTracking.enable()` again after SR registration is silently ignored by the `isTracking` guard, permanently leaving `getCapabilities()` returning `'[]'`.

### Root Cause Analysis
Static capabilities bake-in: the capabilities string is computed once at `WebViewTracking.enableOrThrow()` call time via `core.feature(named: SessionReplayFeatureName, type: SessionReplayConfiguration.self)`. If `SessionReplay` is not yet registered at that moment, `capabilities = ""` and the injected `WKUserScript` returns `'[]'` for `getCapabilities()`. The `WKUserScript` is injected at `.atDocumentStart` and cannot be patched after page load. The `isTracking` guard prevents re-invocation without an explicit `disable()` + `enable()` cycle.

### Call Chain
`WebViewTracking.enable()` → `enableOrThrow()` → `core.feature(SessionReplayFeatureName)` returns `nil` → `capabilities = ""` → `getCapabilities()` returns `'[]'` → Browser SDK disables SR record forwarding → native slot exists (blank wireframe) but no browser records → blank screen in SR player

### Failure Output
```
WebViewTrackingRUMS5686Tests.swift:77: error: testItAddsUserScriptWithEmptyCapabilitiesWhenSessionReplayIsNotYetInitialized : XCTAssertTrue failed - RUMS-5686 reproduction: getCapabilities() returned '[]' instead of '["records"]'.

WebViewTrackingRUMS5686Tests.swift:152: error: testReEnablingWebViewTrackingAfterSRInitDoesNotUpdateCapabilities : XCTAssertTrue failed - RUMS-5686 reproduction: After registering SessionReplay, calling WebViewTracking.enable() again was silently ignored by the isTracking guard.
```

---
*Generated by rum:tee-triage-insights*

[RUMS-5686]: https://datadoghq.atlassian.net/browse/RUMS-5686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Datadog/Datadog.xcodeproj/project.pbxproj
+++ b/Datadog/Datadog.xcodeproj/project.pbxproj
@@ -1714,6 +1714,7 @@
 		D297324B2A5C108700827599 /* MessageEmitter.swift in Sources */ = {isa = PBXBuildFile; fileRef = D29732472A5C108700827599 /* MessageEmitter.swift */; };
 		D29732512A5C109A00827599 /* MessageEmitterTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D297324F2A5C109A00827599 /* MessageEmitterTests.swift */; };
 		D29732532A5C109A00827599 /* WebViewTrackingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D29732502A5C109A00827599 /* WebViewTrackingTests.swift */; };
+		129321A0B6144607B30C74AA /* WebViewTrackingRUMS5686Tests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9124FF5108E841CBBC8CD447 /* WebViewTrackingRUMS5686Tests.swift */; };
 		D29A470D2D2ED49F0092BC79 /* Xoshiro.swift in Sources */ = {isa = PBXBuildFile; fileRef = D29A470C2D2ED49F0092BC79 /* Xoshiro.swift */; };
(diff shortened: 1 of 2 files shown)


gonzalezreal:
Closing this as the issue was caused by using a very old version of the Browser SDK that predates Mobile Session Replay

PR #2819, opened 2026-04-03
Client-side stats proof of concept: feature scaffolding + span lifecycle hook

tienquocbui:
### What and why?

Proof-of-concept for client-side stats (CSS) for APM on the iOS SDK. This is the first agent-less CSS implementation across Datadog tracers. Client-side stats compute trace metrics (hits, errors, duration) locally on all eligible spans — including those sampled out — so that RED metrics remain accurate regardless of client-side sampling rates.

This is a throwaway PoC validating the architecture. Remaining work (DDSketch, MessagePack encoding, flush-to-storage pipeline) will follow in subsequent PRs.

### How?

- Introduces `ClientStatsFeature` as a separate `DatadogRemoteFeature` alongside `TraceFeature`, with its own storage/upload pipeline targeting `/api/v0.2/stats`
- Hooks into `DDSpan.finish()` to create lightweight `SpanSnapshot` value types **before** the sampling decision, ensuring all spans (including sampled-out) contribute to stats
- Adds `StatsConcentrator` that aggregates snapshots into time-bucketed stats on a dedicated background queue, with span eligibility filtering per the Client-Side Stats v1.2.0 spec (top-level, measured, or OTel span_kind of server/consumer/client/producer)
- Gated behind `Trace.Configuration.statsComputationEnabled` (default `false`)

### Review checklist
  - `ClientStatsFeatureTests` — registration, config defaults, request builder wiring (7 tests)
  - `SpanSnapshotTests` — snapshot creation, tag extraction, sampled-out spans (9 tests)
  - `StatsConcentratorTests` — eligibility, aggregation, flushing, key differentiation (12 tests)
  - Existing `TraceTests` and `DDSpanTests` pass with zero regressions

### DIFF
--- a/DatadogTrace/Sources/ClientStats/StatsConcentrator.swift
+++ b/DatadogTrace/Sources/ClientStats/StatsConcentrator.swift
@@ -7,26 +7,32 @@
 import Foundation
 import DatadogInternal
 
+/// Nanosecond-precision timestamp or duration.
+internal typealias Nanoseconds = UInt64
+
 /// Aggregates `SpanSnapshot`s into time-bucketed stats (hit counts, error counts,
 /// duration distributions) keyed by an `AggregationKey`.
 ///
 /// Thread-safety: All public methods dispatch onto a dedicated serial queue
 /// so that `DDSpan.finish()` pays only the cost of enqueuing.
-internal final class StatsConcentrator {
+internal final class StatsConcentrator: @unchecked Sendable {
     /// Default bucket width: 10 seconds (in nanoseconds).
-    static let defaultBucketDuration: UInt64 = 10_000_000_000
+    static let defaultBucketDuration: Nanoseconds = 10_000_000_000
 
-    private let bucketDuration: UInt64
+    /// OTel span kinds that qualify a span for stats, regardless of top-level or measured status.
+    static let eligibleSpanKinds: Set<String> = ["server", "consumer", "client", "producer"]
+
+    private let bucketDuration: Nanoseconds
     private let queue: DispatchQueue
 
     @ReadWriteLock
-    private var buckets: [UInt64: StatsBucket] = [:]
+    private var buckets: [Nanoseconds: StatsBucket] = [:]
 
     /// Peer tag keys considered for downstream-service aggregation.
     private let peerTagKeys: [String]
 
     init(
-        bucketDuration: UInt64 = StatsConcentrator.defaultBucketDuration,
+        bucketDuration: Nanoseconds = StatsConcentrator.defaultBucketDuration,
         peerTagKeys: [String] = StatsConcentrator.defaultPeerTagKeys
     ) {
         self.bucketDuration = bucketDuration
@@ -46,7 +52,7 @@ internal final class StatsConcentrator {
 
     /// Returns all buckets whose start time is strictly older than `cutoff`,
     /// removing them from the internal storage.
-    func flush(olderThan cutoff: UInt64) -> [StatsBucket] {
+    func flush(olderThan cutoff: Nanoseconds) -> [StatsBucket] {
         var flushed: [StatsBucket] = []
         _buckets.mutate { buckets in
             let staleKeys = buckets.keys.filter { $0 < cutoff }
@@ -68,9 +74,9 @@ internal final class StatsConcentrator {
 
         let bucketKey = alignedBucketStart(for: snapshot)
         _buckets.mutate { buckets in
-            var bucket = buckets[bucketKey] ?? StatsBucket(start: bucketKey, duration: self.bucketDuration)
             let key = AggregationKey(snapshot: snapshot, peerTagKeys: self.peerTagKeys)
-            var group = bucket.groups[key] ?? ClientGroupedStats()
+            var group = buckets[bucketKey, default: StatsBucket(start: bucketKey, duration: self.bucketDuration)]
+                .groups[key, default: ClientGroupedStats()]
 
             group.hits += 1
             if snapshot.isError {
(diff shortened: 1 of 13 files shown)


DatadogTrace/Sources/ClientStats/StatsConcentrator.swift
```diff
@@ -0,0 +1,188 @@
...
+    // MARK: - Private
+
+    private func aggregate(_ snapshot: SpanSnapshot) {
+        guard Self.isEligible(snapshot) else {
+            return
+        }
+
+        let bucketKey = alignedBucketStart(for: snapshot)
+        _buckets.mutate { buckets in
+            var bucket = buckets[bucketKey] ?? StatsBucket(start: bucketKey, duration: self.bucketDuration)
```
arroz: These can be simplified to `buckets[bucketKey, default: StatsBucket(…)`

DatadogTrace/Sources/ClientStats/StatsConcentrator.swift
```diff
@@ -0,0 +1,188 @@
...
+    /// qualifying `span_kind` (server, consumer, client, producer).
+    static func isEligible(_ snapshot: SpanSnapshot) -> Bool {
+        if snapshot.isTopLevel {
+            return true
+        }
+        if snapshot.isMeasured {
+            return true
+        }
+        if let kind = snapshot.spanKind?.lowercased(),
+           ["server", "consumer", "client", "producer"].contains(kind) {
```
arroz: These can be constants or an enum (although it should support unknown values).

DatadogTrace/Sources/ClientStats/StatsConcentrator.swift
```diff
@@ -0,0 +1,188 @@
...
+/// Aggregates `SpanSnapshot`s into time-bucketed stats (hit counts, error counts,
+/// duration distributions) keyed by an `AggregationKey`.
+///
+/// Thread-safety: All public methods dispatch onto a dedicated serial queue
+/// so that `DDSpan.finish()` pays only the cost of enqueuing.
+internal final class StatsConcentrator {
+    /// Default bucket width: 10 seconds (in nanoseconds).
+    static let defaultBucketDuration: UInt64 = 10_000_000_000
+
+    private let bucketDuration: UInt64
```
arroz: Ideally create a typealias for the time unit (assuming we don't have it already for other payloads).

DatadogTrace/Sources/ClientStats/StatsConcentrator.swift
```diff
@@ -0,0 +1,188 @@
...
+        var hash: UInt64 = 14_695_981_039_346_656_037 // FNV offset basis
+        for (key, value) in relevant {
+            for byte in "\(key)=\(value),".utf8 {
+                hash ^= UInt64(byte)
+                hash &*= 1_099_511_628_211 // FNV prime
+            }
+        }
+        return hash
+    }
+}
```
arroz: This can be moved to an utility class.

DatadogTrace/Sources/ClientStats/StatsConcentrator.swift
```diff
@@ -0,0 +1,188 @@
...
+
+import Foundation
+import DatadogInternal
+
+/// Aggregates `SpanSnapshot`s into time-bucketed stats (hit counts, error counts,
+/// duration distributions) keyed by an `AggregationKey`.
+///
+/// Thread-safety: All public methods dispatch onto a dedicated serial queue
+/// so that `DDSpan.finish()` pays only the cost of enqueuing.
+internal final class StatsConcentrator {
```
arroz: Generic comment, mark things as Sendable as much as possible so we can have as much compile-time checking as possible.

DatadogTrace/Sources/Span/SpanSnapshot.swift
```diff
@@ -0,0 +1,40 @@
...
+    let httpStatusCode: UInt32
+    let isError: Bool
+    /// Span start time in nanoseconds since Unix epoch.
+    let startTime: UInt64
+    /// Span duration in nanoseconds.
+    let duration: UInt64
+    let isTopLevel: Bool
+    let isMeasured: Bool
+    /// Peer tag values for downstream-service aggregation (e.g., `peer.service`, `out.host`).
+    let peerTags: [String: String]
```
arroz, line 36: We have some typealiases defined for this, `[AttributeKey: AttributeValue]`.

DatadogTrace/Sources/DDSpan.swift
```diff
@@ -55,6 +58,7 @@ internal final class DDSpan: OTSpan {
         self.isFinished = false
         self.eventBuilder = eventBuilder
         self.eventWriter = eventWriter
+        self.onSpanFinished = onSpanFinished
```
arroz: Do we need to carry `onSpanFinished` around like this? Seems unnecessary for every single span to have a reference to it.

DatadogTrace/Sources/Span/SpanSnapshot.swift
tienquocbui: I considered this, the existing `AttributeKey/AttributeValue` typealiases map to `String/Encodable`, but `peerTags` is intentionally `[String: String]` since we only extract string values from tags for hashing purposes. I think for using `[AttributeKey: AttributeValue]` would lose that type safety. Happy to switch if you prefer consistency over strictness here though.

DatadogTrace/Sources/DDSpan.swift
tienquocbui: You're right, the callback could live on `DatadogTracer` instead (it already holds it as `onSpanFinished`), and `DDSpan.finish()` can call through `ddTracer` directly. That way the span stays clean and the tracer is the single owner of the callback. I'll refactor so `DDSpan.finish()` calls something like `ddTracer.notifySpanFinished(snapshot)`.

tienquocbui:
Closing PoC PR — the code will continue on the feature branch and merge to develop once the feature is end-to-end functional.

PR #2820, opened 2026-04-06
Parallelizes CocoaPods publishing.

arroz:
Currently, all the pods except `DatadogInternal` are published serially in the same job. CocoaPods upload takes an average of 12 minutes per pod (not counting linting time). Since we have 9 pods, this takes a very long time and jobs tend to fail with timeouts. This commit parallelizes the publishing of those 9 pods, by creating a different CI job for each one. These will run in parallel based on the CI queue. It should decrease the release time and also allow us to retry individual pods instead of the entire batch.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -427,9 +427,9 @@ Publish CP podspecs (internal):
     - make clean
     - make release-publish-internal-podspecs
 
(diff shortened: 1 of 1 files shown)


simaoseica-dd (approved):
Thank you!

maxep (approved):
🫶

PR #2822, opened 2026-04-07
RUM-8113 Align retry_count request tag with other SDKs

saraSr5:
### What and why?                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   
The iOS SDK was incorrectly emitting `retry_count:1` on every request including the first one.                                                                                                                                                                               
Unlike Android and Browser SDKs, which only set `retry_count` after an initial failure, iOS was                                                                                                                                                                            
setting it unconditionally. This PR fixes the behaviour to align with other SDKs.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                               
  ### How?                                                                                                                                                                                                                                                                                                                                                               

Extracted shared retry tag logic into two computed properties on `ExecutionContext` in `DatadogInternal`:                                                                                                                                                                  
  - `retryTags` (private) — returns `nil` on the first request (`attempt == 0`), otherwise builds `["retry_count:<n>"]` plus `"retry_after:<status>"` when an HTTP response code is available
  - `retryQueryItems` (public) — returns `[]` on first request or `[.ddtags(tags: retryTags)]` on retries                                                                                                                                                                                                                                                                                                                                                                                                                                                     
  This correctly handles both retry scenarios:                                                                                                                                                                                                                                 
  - **HTTP error retry** (`attempt > 0`, `previousResponseCode != nil`): emits `retry_count` + `retry_after`                                                                                                                                                           
  - **Network error retry** (`attempt > 0`, `previousResponseCode == nil`): emits `retry_count` only
  
Updated RUM, SessionReplay (Segment + Resource), and Profiling request builders to use `execution.retryQueryItems` instead of their own inline tag construction.

Logs and Trace builders are unaffected as they do not use retry tags.
                                                                                                                                                                                                                                                                                                                                                                                                              
  ### Review checklist                                                                                                                                                                                                                                                         
  - [ ] Add CHANGELOG entry for user facing changes — not needed                                                                                                                                                                                                               
  - [ ] Add Objective-C interface for public APIs — no customer-facing public APIs added
  - [ ] Run `make api-surface` when adding new APIs — verified, no surface changes

### DIFF
--- a/DatadogInternal/Sources/Upload/FeatureRequestBuilder.swift
+++ b/DatadogInternal/Sources/Upload/FeatureRequestBuilder.swift
@@ -40,6 +40,25 @@ public struct ExecutionContext {
     /// The current attempt number.
     public let attempt: UInt
 
+    /// Tags to include in the request when this is a retry attempt.
+    /// Returns `nil` on the first request (`attempt == 0`).
+    private var retryTags: [String]? {
+        guard attempt > 0 else {
+            return nil
+        }
+        var tags = ["retry_count:\(attempt)"]
+        if let code = previousResponseCode {
+            tags.append("retry_after:\(code)")
+        }
+        return tags
+    }
+
+    /// Query items to include in the request when this is a retry attempt.
+    /// Returns an empty array on the first request (`attempt == 0`).
+    public var retryQueryItems: [URLRequestBuilder.QueryItem] {
+        retryTags.map { [.ddtags(tags: $0)] } ?? []
+    }
+
     /// Initializes the execution context.
     /// - Parameters:
     ///   - previousResponseCode: Previous HTTP status code, if available.
--- a/DatadogCore/Tests/Datadog/RUM/RUMFeatureTests.swift
+++ b/DatadogCore/Tests/Datadog/RUM/RUMFeatureTests.swift
@@ -91,7 +91,7 @@ class RUMFeatureTests: XCTestCase {
         XCTAssertEqual(
             requestURL.query,
             """
-            ddsource=\(randomSource)&ddtags=retry_count:1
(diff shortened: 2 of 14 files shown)


DatadogInternal/Sources/Upload/FeatureRequestBuilder.swift
```diff
@@ -40,6 +40,24 @@ public struct ExecutionContext {
...
+
+    /// Query items to include in the request when this is a retry attempt.
+    /// Returns an empty array on the first request (when `previousResponseCode` is `nil`).
+    public var retryQueryItems: [URLRequestBuilder.QueryItem] {
+        guard previousResponseCode != nil else {
+            return []
+        }
+        return [.ddtags(tags: retryTags)]
+    }
+
```
maxep, line 61: nit/

```suggestion
    /// Tags to include in the request when this is a retry attempt.
    /// Returns an `nil` on the first request (when `previousResponseCode` is `nil`).
    private var retryTags: [String]? {
       previousResponseCode.map { ["retry_count:\(attempt)", "last_failure_status:\($0)"] } 
    }

    /// Query items to include in the request when this is a retry attempt.
    /// Returns an empty array on the first request (when `previousResponseCode` is `nil`).
    public var retryQueryItems: [URLRequestBuilder.QueryItem] {
       retryTags.map { [.ddtags(tags: $0)] } ?? []
    }

```

maxep (dismissed):
Looks great! Just a minor suggestion, but it is good to go as is 👍

maxep (dismissed):
Thanks 🙏

maxep (approved):

PR #2830, opened 2026-04-10
[RUM-14354] Enable SwiftUI manual instrumentation on watchOS

maxep:
### What and why?

watchOS doesn't have `UIApplication` but it supports SwiftUI. This PR partially enables RUM instrumentation on watchOS by making the SwiftUI manual modifiers (`.trackRUMTapAction()` and `.trackRUMView()`) available on watchOS 7+.

Automatic instrumentation (UIKit swizzling) remains unavailable on watchOS.

Additionally, this PR fixes a crash when initializing the SDK in a SwiftUI lifecycle watchOS app (using `@main App`). The root cause was `WKExtension.shared()` being called during SDK init — this API only works in the legacy WatchKit Extension lifecycle and aborts when called from a SwiftUI `@main` app. The fix replaces it with `WKApplication.shared()`, which works across both lifecycles.

### How?

The SwiftUI manual view and action modifiers are now compiled for watchOS 7+. UIKit-based automatic instrumentation is unchanged and continues to be excluded on watchOS.

`DefaultAppStateProvider` on watchOS now uses `WKApplication.shared().applicationState` instead of `WKExtension.shared().applicationState` to safely support both the legacy WatchKit Extension and the modern SwiftUI `@main` app lifecycle.

### Review checklist

### DIFF
--- a/DatadogInternal/Sources/Context/AppState.swift
+++ b/DatadogInternal/Sources/Context/AppState.swift
@@ -158,8 +158,7 @@ public struct DefaultAppStateProvider: AppStateProvider {
     ///
     /// **Note**: Must be called on the main thread.
     public var current: AppState {
-        let wkState = WKExtension.dd.shared.applicationState
-        return AppState(wkState)
+        return AppState(WKApplication.shared().applicationState)
     }
 }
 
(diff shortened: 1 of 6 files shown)


gonzalezreal (approved):
🚀

PR #2832, opened 2026-04-10
[RUM-744] Update hardcoded data limits

saraSr5:
### What and why?
Enforce hardcoded backend data limits client-side to give developers explicit warnings instead of silently losing data. Specifically: truncate string attribute values exceeding 25,600 characters (the backend indexing limit) in Logs, update the attribute count limit to match the actual backend limit, update batch size constraints to match the server's 5MB per-batch limit, and extract the TLV batch size as a named constant.

### How?
  - Added maxAttributeValueLength = 25_600 to AttributesSanitizer.Constraints and a new sanitizeValues() method that truncates string values exceeding the limit (handles both String and bridged NSString values)
  - Wired sanitizeValues() into LogEventSanitizer only. RUM and Trace have no documented attribute value length limit in their backend schemas                                                                       
  - Updated maxNumberOfAttributes from 128 → 256 to match the actual backend limit documented in the Platform Limitations page                
  - Updated PerformancePreset defaults to match the server's documented batch capacity: maxFileSize 4MB → 5MB, maxObjectsInFile 500 → 1,000, maxObjectSize 512KB → 1MB. The previous values were conservative        
  estimates below the actual backend limits                                                                                                                                                                  
  - Extracted maxTLVDataLength as a named constant in Storage+TLV.swift, reduced from 10MB → 5MB to align with the server's 5MB per-batch limit, with a separate maxDataStoreTLVDataLength constant (10MB) scoped to 
  data-store files                                                                                                          

### Review checklist
  - [ ] Feature or bugfix MUST have appropriate tests (unit, integration) → new value truncation tests in LogSanitizerTests (including NSString bridging), testMaxTLVDataLengthMatchesBatchSizeLimit added               
  - [ ] Add CHANGELOG entry for user facing changes → not needed                   
  - [ ] Add Objective-C interface for public APIs → not applicable                                                                                                                                                       
  - [ ] Run make api-surface when adding new APIs → ran, no changes to API surface

### DIFF
--- a/DatadogCore/Sources/PerformancePreset.swift
+++ b/DatadogCore/Sources/PerformancePreset.swift
@@ -127,13 +127,13 @@ internal extension PerformancePreset {
         uploadDelayFactors: (initial: Double, min: Double, max: Double, changeRate: Double),
         maxBatchesPerUpload: Int
     ) {
-        self.maxFileSize = 4.MB.asUInt32()
+        self.maxFileSize = 5.MB.asUInt32()
         self.maxDirectorySize = 512.MB.asUInt32()
         self.maxFileAgeForWrite = meanFileAge * 0.95 // 5% below the mean age
         self.minFileAgeForRead = meanFileAge * 1.05 //  5% above the mean age
         self.maxFileAgeForRead = 18.hours
-        self.maxObjectsInFile = 500
-        self.maxObjectSize = 512.KB.asUInt32()
+        self.maxObjectsInFile = 1_000
+        self.maxObjectSize = 1.MB.asUInt32()
         self.initialUploadDelay = minUploadDelay * uploadDelayFactors.initial
         self.minUploadDelay = minUploadDelay * uploadDelayFactors.min
         self.maxUploadDelay = minUploadDelay * uploadDelayFactors.max
--- a/DatadogCore/Sources/PerformancePreset.swift
+++ b/DatadogCore/Sources/PerformancePreset.swift
@@ -132,8 +132,8 @@ internal extension PerformancePreset {
         self.maxFileAgeForWrite = meanFileAge * 0.95 // 5% below the mean age
         self.minFileAgeForRead = meanFileAge * 1.05 //  5% above the mean age
         self.maxFileAgeForRead = 18.hours
-        self.maxObjectsInFile = 1_000 // updated from 500 to match the backend's documented 1,000 events-per-batch limit
-        self.maxObjectSize = 1.MB.asUInt32() // updated from 512 KB to match the backend's documented 1 MB per-event limit
+        self.maxObjectsInFile = 1_000 // Backend limit: max 1,000 events per batch
+        self.maxObjectSize = 1.MB.asUInt32() // Backend limit: max 1 MB per event
         self.initialUploadDelay = minUploadDelay * uploadDelayFactors.initial
         self.minUploadDelay = minUploadDelay * uploadDelayFactors.min
         self.maxUploadDelay = minUploadDelay * uploadDelayFactors.max
--- a/DatadogCore/Sources/PerformancePreset.swift
+++ b/DatadogCore/Sources/PerformancePreset.swift
@@ -127,13 +127,13 @@ internal extension PerformancePreset {
         uploadDelayFactors: (initial: Double, min: Double, max: Double, changeRate: Double),
         maxBatchesPerUpload: Int
     ) {
-        self.maxFileSize = 5.MB.asUInt32()
+        self.maxFileSize = 5.MB.asUInt32() // matches the server's 5 MB per-batch limit
         self.maxDirectorySize = 512.MB.asUInt32()
         self.maxFileAgeForWrite = meanFileAge * 0.95 // 5% below the mean age
         self.minFileAgeForRead = meanFileAge * 1.05 //  5% above the mean age
         self.maxFileAgeForRead = 18.hours
-        self.maxObjectsInFile = 1_000
-        self.maxObjectSize = 1.MB.asUInt32()
+        self.maxObjectsInFile = 1_000 // updated from 500 to match the backend's documented 1,000 events-per-batch limit
+        self.maxObjectSize = 1.MB.asUInt32() // updated from 512 KB to match the backend's documented 1 MB per-event limit
         self.initialUploadDelay = minUploadDelay * uploadDelayFactors.initial
         self.minUploadDelay = minUploadDelay * uploadDelayFactors.min
         self.maxUploadDelay = minUploadDelay * uploadDelayFactors.max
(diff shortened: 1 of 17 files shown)


DatadogLogs/Tests/Log/LogSanitizerTests.swift
```diff
@@ -175,6 +175,42 @@ class LogSanitizerTests: XCTestCase {
         }
     }
 
+    func testWhenAttributeValueExceeds25600Characters_itIsTruncated() {
```
mariedm: nit: I wouldn't hardcode the 25600 limit in the test titles in case it changes
```suggestion
    func testWhenAttributeValueExceedsCharacterLimit_itIsTruncated() {
```

DatadogLogs/Tests/Log/LogSanitizerTests.swift
```diff
@@ -175,6 +175,42 @@ class LogSanitizerTests: XCTestCase {
...
+    func testWhenAttributeValueIsWithinLimit_itIsNotModified() {
+        let value = String(repeating: "a", count: 25_600)
+        let log = LogEvent.mockWith(
+            attributes: .mockWith(userAttributes: ["key": value])
+        )
+
+        let sanitized = LogEventSanitizer().sanitize(log: log)
+
+        let result = sanitized.attributes.userAttributes["key"] as? String
+        XCTAssertEqual(result?.count, 25_600)
```
mariedm, line 227: extra nit: you could additionally assert on the value itself since it's not modified

DatadogInternal/Sources/Attributes/AttributesSanitizer.swift
```diff
@@ -65,6 +68,25 @@ public struct AttributesSanitizer {
         return sanitized
     }
 
+    // MARK: - Attribute values sanitization
+
+    /// Truncates string attribute values exceeding `Constraints.maxAttributeValueLength`.
+    public func sanitizeValues(for attributes: [String: Encodable]) -> [String: Encodable] {
+        return Dictionary(uniqueKeysWithValues: attributes.map { key, value in
+            guard let string = value.dd.decode(String.self),
+                  string.count > Constraints.maxAttributeValueLength else {
```
mariedm: What is the unit of `maxAttributeValueLength`? Is it characters, bytes, or something else? Swift counts characters (and more specifically Unicode grapheme clusters). This means an emoji counts for 1 character with `Swift.count`, but 4 UTF-8 bytes (or more, like 8 bytes with skin tone for example). I doubt customers often use emojis in attributes, but we should align on backend's unit to avoid wrong truncations.

mariedm (changes requested):
Good job overall! 🙌  

Aside from the few comments I left, I have one concern:

> RUM and Trace have no documented attribute value length limit in their backend schemas

By "backend schemas", do you mean the [rum-events-format](https://github.com/DataDog/rum-events-format) repo? This schema mainly serves as a contract between the SDKS but is actually not enforced by the backend (no obvious, I know).

In any case, changing the `maxNumberOfAttributes` in `AttributesSanitizer` **does affect** RUM and Trace as both `RUMEventSanitizer` and `SpanSanitizer` use it. Can we check with backend if the RUM and Trace endpoints have the same requirements as Logs? If yes, we can keep the same constants for all features. If not, we'll need to split them.

saraSr5:
> Good job overall! 🙌
> 
> Aside from the few comments I left, I have one concern:
> 
> > RUM and Trace have no documented attribute value length limit in their backend schemas
> 
> By "backend schemas", do you mean the [rum-events-format](https://github.com/DataDog/rum-events-format) repo? This schema mainly serves as a contract between the SDKS but is actually not enforced by the backend (no obvious, I know).
> 
> In any case, changing the `maxNumberOfAttributes` in `AttributesSanitizer` **does affect** RUM and Trace as both `RUMEventSanitizer` and `SpanSanitizer` use it. Can we check with backend if the RUM and Trace endpoints have the same requirements as Logs? If yes, we can keep the same constants for all features. If not, we'll need to split them.


Hi! I checked with backend on both!           
                                                                                                                                                                                                                                                                               
For RUM, confirmed that 256 is the correct limit. The RUM backend parser enforces 256 attributes by default (same as the EP processing layer), so the shared constant is correct.                                                                                 
                                                                                                                                                                                                                                                                               
For Trace/APM, I checked in #apm-trace-storage. Bumping from 128 → 256 is fine. They did raise a concern about the priority order of which span tags get dropped when the limit is hit. This is pre-existing behaviour not introduced by this PR, but wanted to flag it in case it's worth a follow-up.

DatadogCore/Sources/PerformancePreset.swift
```diff
@@ -127,13 +127,13 @@ internal extension PerformancePreset {
...
     ) {
-        self.maxFileSize = 5.MB.asUInt32()
+        self.maxFileSize = 5.MB.asUInt32() // matches the server's 5 MB per-batch limit
         self.maxDirectorySize = 512.MB.asUInt32()
         self.maxFileAgeForWrite = meanFileAge * 0.95 // 5% below the mean age
         self.minFileAgeForRead = meanFileAge * 1.05 //  5% above the mean age
         self.maxFileAgeForRead = 18.hours
-        self.maxObjectsInFile = 1_000
-        self.maxObjectSize = 1.MB.asUInt32()
+        self.maxObjectsInFile = 1_000 // updated from 500 to match the backend's documented 1,000 events-per-batch limit
```
mariedm: I'd avoid mentioning the previous value since it will become stale.
```suggestion
        self.maxObjectsInFile = 1_000 // Backend limit: max 1,000 events per batch
```

DatadogInternal/Sources/Attributes/AttributesSanitizer.swift
```diff
@@ -15,8 +15,11 @@ public struct AttributesSanitizer {
         /// Maximum number of attributes in log.
         /// If this number is exceeded, extra attributes will be ignored.
         public static let maxNumberOfAttributes: Int = 256
-        /// Maximum length of a string attribute value.
-        /// Values exceeding this will be truncated. This matches the backend hard limit, anything beyond 25,600 chars is discarded server-side regardless of facet status.
+        /// Maximum length of a string attribute value, measured in Unicode grapheme clusters (Swift `String.count`).
+        /// Values exceeding this will be truncated. This matches the backend hard limit: anything beyond 25,600
+        /// characters is discarded server-side regardless of facet status. Note that the backend limit is defined
+        /// in characters, not UTF-8 bytes — for ASCII-only values the two are equivalent, but multi-byte characters
+        /// (e.g. emoji) may differ in byte count while still counting as a single character here.
```
mariedm: If I understand correctly, we still don't know how backend counts characters, just that it's not using UTF-8 bytes, so backend and iOS string counting might differ. Can we get an answer on that so we can align? Depending on what backend says, we could use `string.utf8.count` or `string.unicodeScalars.count`. 

I'd say there is a low probability that customers use emojis in attributes, but they could still use non-ASCII characters, especially as we have many international customers.

mariedm:
> I checked with backend on both!

Thanks for confirming!

DatadogInternal/Sources/Attributes/AttributesSanitizer.swift
saraSr5: Got confirmation from RUM backend: the backend measures length using Java's String.length(), which counts UTF-16 code units, not UTF-8 bytes or Unicode scalars. Updated the truncation to use **string.utf16.count** and truncate at a grapheme cluster boundary within that limit

mariedm (dismissed):
Great! Thanks for confirming with backend and updating the logic accordingly. Not a blocker, but it would have be nice to add a unit test reflecting this as well.

mariedm (approved):
🚀🚀

PR #2834, opened 2026-04-10
RUMS-5762: Fix blank SwiftUI TabView when Session Replay is enabled

simaoseica-dd:
### What and why?

This PR fixes a Session Replay issue introduced in [3.7.0](https://github.com/DataDog/dd-sdk-ios/releases/tag/3.7.0) where `SwiftUI` `TabView` could fail to render if it appeared after Session Replay was enabled. The root cause was reading `intrinsicContentSize` on `SwiftUI` owned views while they still had pending layout, which could trigger an `AttributeGraph` cycle.

### How?

Session Replay now avoids reading `intrinsicContentSize` for `SwiftUI` owned views with pending layout and uses the current bounds size instead during snapshot capture. 
Additionally, the `SwiftUI` view detection helper was centralized in `DatadogInternal`.

### Review checklist

### DIFF
--- a/Datadog/Datadog.xcodeproj/project.pbxproj
+++ b/Datadog/Datadog.xcodeproj/project.pbxproj
@@ -9,6 +9,7 @@
 /* Begin PBXBuildFile section */
 		0904F9F42EE1DA6800ED9A22 /* UIKitExtensionsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0904F9F32EE1DA6800ED9A22 /* UIKitExtensionsTests.swift */; };
 		0904F9F62EE1DA6800ED9A22 /* UIKitExtensionsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0904F9F32EE1DA6800ED9A22 /* UIKitExtensionsTests.swift */; };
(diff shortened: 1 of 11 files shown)


Valpertui (dismissed):
Seems like a good fix

gonzalezreal (commented):
Thanks for the investigation, @simaoseica-dd. I agree with the root cause, reading `intrinsicContentSize` on SwiftUI views with pending layout triggers an AttributeGraph cycle.

However, I think the fix is incomplete. `DisplayList+Reflection.swift` also reads `intrinsicContentSize` and `preferredFrameSize` on SwiftUI containers during the same snapshot capture path, and those aren't protected by the `safeIntrinsicContentSize` guard. We'd need to audit and patch every property read that can trigger layout work, and we might miss some.

I'd prefer to fix this at the architectural level instead. The root problem is that the `CALayer` swizzle can trigger snapshot capture synchronously, on the same call stack as the layout pass. This means we're reading view properties while SwiftUI's layout engine is still active.

If we decouple the swizzle from the capture so that the swizzle only marks the screen as dirty and a separate timer drives the actual capture, snapshot capture will always run on a different run loop turn. By that point layout has settled and all these property reads are safe. This eliminates the entire class of bugs without needing per-property guards.

I can take care of implementing this. The change should be internal to `CALayerChangeAggregator` so the ScreenChangeMonitor interface stays the same.

Let me know your thoughts.

PR #2840, opened 2026-04-14
[PANA-6952] Defer `CALayer` change delivery off the swizzled call stack

gonzalezreal:
### What and why?

The `ScreenChangeMonitor` uses a process-wide `CALayer` swizzle to detect screen changes. When the minimum delivery interval had already elapsed, `CALayerChangeAggregator` delivered changes synchronously on the same call stack as the swizzled `layoutSublayers`. This meant snapshot capture could run inside an active layout pass, causing:

- An infinite loop where snapshot reads triggered new layer changes, re-entering the delivery path (fixed in 3.8.2)
- Blank SwiftUI `TabView` when reading `intrinsicContentSize` re-entered SwiftUI's `AttributeGraph` during an active layout pass (RUMS-5762)

### How?

`CALayerChangeAggregator.scheduleDeliveryIfNeeded()` now always delivers through a scheduled timer, even when the minimum interval has elapsed. When the interval is already past, the timer fires with zero delay, which defers delivery to the next run loop turn. This guarantees snapshot capture never runs on the same call stack as the swizzled layout pass.

### Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- [FIX] Prevent Session Replay capture from running on the swizzled `CALayer` layout call stack, avoiding AttributeGraph cycles and other errors. See [#2840][]
+
 # 3.9.0 / 02-04-2026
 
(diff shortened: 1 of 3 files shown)


simaoseica-dd (approved):
LGTM!

maxep (approved):

PR #2841, opened 2026-04-14
RUM-15513: Add deterministic sampling based on session id to Profiling

simaoseica-dd:
### What and why?

This PR aligns `DatadogProfiling` continuous profiling sampling with the RUM session sampling decision, so profiling follows the same cross-feature deterministic strategy already used in tracing.

The goal is to make profiling behavior more consistent across features where sampling is impacted by RUM configurations.

### How?

It introduces a dedicated `ProfilingSamplerProvider` updated by a `ProfilingContextMessageReceiver`, so profiling sampling state is derived from the RUM session sampler instead of making a one-off local sampling decision when the profiler feature starts.

`ContinuousProfiler` and `AppLaunchProfiler` now consume that shared provider, which centralizes the sampling logic and keeps the profiler components focused on lifecycle and writing behavior. Continuous profiling can start normally, but continuous profile writes are gated by the RUM-linked sampling result.

### Review checklist

### DIFF
--- a/DatadogInternal/Sources/Utils/Sampler.swift
+++ b/DatadogInternal/Sources/Utils/Sampler.swift
@@ -51,7 +51,7 @@ extension SampleRate {
     }
 
     /// Clamps the value between 0 and 100
-    internal var normalized: Self {
+    public var normalized: Self {
         max(0, min(100, self))
     }
 }
--- a/DatadogInternal/Sources/Utils/Sampler.swift
+++ b/DatadogInternal/Sources/Utils/Sampler.swift
@@ -28,7 +28,7 @@ public struct Sampler: Sampling {
     public let samplingRate: SampleRate
 
     public init(samplingRate: SampleRate) {
-        self.samplingRate = samplingRate.normalized
+        self.samplingRate = samplingRate.normalizedSampleRate
     }
 
     /// Based on the sampling rate, it returns random value deciding if an event should be "sampled" or not.
@@ -51,7 +51,7 @@ extension SampleRate {
     }
 
     /// Clamps the value between 0 and 100
-    public var normalized: Self {
+    public var normalizedSampleRate: Self {
         max(0, min(100, self))
     }
 }
--- a/DatadogProfiling/Sources/DatadogProfiler.swift
+++ b/DatadogProfiling/Sources/DatadogProfiler.swift
@@ -327,9 +327,9 @@ private extension DatadogProfiler {
     func shouldKeepProfilerRunning() -> Bool {
         if profilingSamplerProvider.isContinuousProfilingConfigured {
             switch profilingSamplerProvider.continuousProfilingSampled {
-            case true: // It is Continuous Profiling running
+            case .some(true): // It is Continuous Profiling running
                 return hasConditionsToProfile
-            case false: // It is Custom Profiling running
+            case .some(false): // It is Custom Profiling running
                 return hasConditionsToProfile && currentRUMVitals.ongoingOperations().isEmpty == false
             case .none: // Waiting for RUM context info
                 return hasConditionsToProfile && isContinuousProfilingGraceAvailable
--- a/DatadogProfiling/Sources/DatadogProfiler.swift
+++ b/DatadogProfiling/Sources/DatadogProfiler.swift
@@ -38,7 +38,7 @@ internal final class DatadogProfiler: ProfilingHandler {
 
     /// The queue used to synchronize the profiling data and the writes.
     private let queue: DispatchQueue
-    private let isContinuousProfiling: Bool
+    private let profilingSamplerProvider: ProfilingSamplerProvider
     private let profilingConditions: ProfilingConditions
     private let profilingInterval: TimeInterval
     private var timer: DispatchSourceTimer?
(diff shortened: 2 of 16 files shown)


DatadogProfiling/Sources/DatadogProfiler.swift
```diff
@@ -175,22 +184,31 @@ private extension DatadogProfiler {
             guard let self else {
                 return
             }
-            // Check, based on the context, if the profiler has conditions to profile
+
```
simaoseica-dd, line 187: Fetching these parameters to filter out context updates that are not important. 
The ones important are background state and `currentAppState != previousAppState || hasConditionsToProfile != previousConditions`

DatadogProfiling/Sources/DatadogProfiler.swift
```diff
@@ -237,7 +238,7 @@ private extension DatadogProfiler {
                     currentRUMVitals[message.operation.key] = startVital
 
                     // Stop profiler if it is a custom profiler and the operations have completed
-                    if currentRUMVitals.didCompleteOperations() && isContinuousProfiling == false {
+                    if currentRUMVitals.didCompleteOperations() && profilingSamplerProvider.isContinuousProfilingConfigured == false {
```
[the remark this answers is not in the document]
simaoseica-dd: Fixed here https://github.com/DataDog/dd-sdk-ios/pull/2841/changes/ae8ba1de60e34d95a799e318729db05594e0ddb7

DatadogProfiling/Sources/DatadogProfiler.swift
```diff
@@ -305,14 +304,30 @@ private extension DatadogProfiler {
...
+
+    func shouldKeepProfilerRunning() -> Bool {
+        if profilingSamplerProvider.isContinuousProfilingConfigured {
+            switch profilingSamplerProvider.continuousProfilingSampled {
+            case true?: // It is Continuous Profiling running
+                return hasConditionsToProfile
+            case false?: // It is Custom Profiling running
+                return hasConditionsToProfile && currentRUMVitals.ongoingOperations().isEmpty == false
+            case .none: // Waiting for RUM context info
+                return hasConditionsToProfile
```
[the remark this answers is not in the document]
simaoseica-dd: Fixed here https://github.com/DataDog/dd-sdk-ios/pull/2841/changes/ae8ba1de60e34d95a799e318729db05594e0ddb7

DatadogInternal/Sources/Utils/Sampler.swift
```diff
@@ -51,7 +51,7 @@ extension SampleRate {
     }
 
     /// Clamps the value between 0 and 100
-    internal var normalized: Self {
+    public var normalized: Self {
```
arroz, line 54: Making `normalized` public makes me a little nervous because `SampleRate` is just a Float. It's prooooobably OK, but maybe we should change the name to `normalizedSampleRate` to avoid any conflict with customer extensions, or future Apple APIs?

DatadogProfiling/Sources/DatadogProfiler.swift
```diff
@@ -305,14 +317,35 @@ private extension DatadogProfiler {
...
+        // At least Custom Profiling is running
+        self.currentRUMVitals.count > 0
+        // Continuous Profiling is sampled in and there are events of interest
+        || (profilingSamplerProvider.continuousProfilingSampled == true && (hangs.count > 0 || longTasks.count > 0))
+    }
+
+    func shouldKeepProfilerRunning() -> Bool {
+        if profilingSamplerProvider.isContinuousProfilingConfigured {
+            switch profilingSamplerProvider.continuousProfilingSampled {
+            case true?: // It is Continuous Profiling running
```
arroz, line 330: The question marks aren't wrong, but are not needed. Alternatively it can be written as `.some(true)`. No changes needed, just a matter of preference (personally, the question mark there confuses me a little).

DatadogProfiling/Sources/ProfilingSamplerProvider.swift
```diff
@@ -0,0 +1,44 @@
...
+
+    /// Session-linked sampling result for continuous profiling.
+    ///
+    /// The value is `nil` until a RUM context provides a `sessionSampler`. Once a context is received,
+    /// the value becomes the result of composing the RUM session sampler with `continuousSampleRate`.
+    @ReadWriteLock
+    private(set) var continuousProfilingSampled: Bool?
+
+    /// `true` when continuous profiling is configured with a sample rate greater than zero.
+    let isContinuousProfilingConfigured: Bool
```
arroz, line 32: This may have a more explicit name like `isContinuousProfilingDisabled` although that would require inverting all the logic that uses it. I don't have a suggestion for a name keeping the positive logic because `isContinuousProfilingEnabled` wouldn't be accurate. Naming things is hard. :) It's not a show stopper as it is, though.

arroz (approved):
All looks good! Left a few minor points, none is a show stopper.

DatadogProfiling/Sources/ProfilingSamplerProvider.swift
simaoseica-dd: `Enable` and `disable` should also be avoided to prevent confusion with feature enablement.`Profiling.enable()`.

arroz: No worries, the documentation is enough to clarify any possible doubts.

arroz (dismissed):
🚢 it!

arroz (approved):
🚀

PR #2676, opened 2026-02-06
RUM-13731 Improve attribute encoding error message for Trace

mariedm:
### What and why?

Follow-up of #2665. 

Unlike Logs (see #2665), Trace already handles errors gracefully, but error messages lack context and don't include the attribute name in the main message. This PR improves error messaging with names and context (custom, userInfo, accountInfo), consistent with what has been done for Logs. 

### How?

- Enhanced existing error handling in `SpanEventBuilder.castValuesToString()` with context awareness and improved error message format
- Integration tests in `DDSpanTests` simulate ObjC bridge with closures/NSObject

Note: also merged 2 tests in one in `RemoteLoggerTests`

### Review checklist
- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference
- [x] Add CHANGELOG entry for user facing changes
- [ ] Add Objective-C interface for public APIs - see our [guidelines](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3157787243/RFC+-+Modular+Objective-C+Interface#Recommended-solution) (internal) 
- [ ] Run `make api-surface` when adding new APIs

PR #2844, opened 2026-04-15
[RUM-13731] Skip malformed RUM attributes instead of dropping the entire event

mariedm:
### What and why?

When a customer passes a non-encodable value as a RUM attribute — most commonly via the ObjC bridge, where `addAttribute(forKey:value:)` accepts `Any` and the SDK wraps it in `AnyEncodable` — the entire RUM event is dropped instead of just skipping the bad attribute.

This is the same issue fixed for Logs in #2665 and Trace in #2676. This PR closes the gap for RUM.

### How?

RUM event models are auto-generated from the `rum-events-format` schema via `tools/rum-models-generator`. Patching `RUMDataModels.swift` directly would be overwritten on the next schema update, so the fix goes into `SwiftPrinter.swift` (the code generator): the `forEach { try encode() }` pattern is replaced with `forEach { encodeAttribute() }`, using the `KeyedEncodingContainer` extension introduced in #2665. It catches encoding errors internally, logs a clear message naming the specific attribute, and skips only that key — the event is always sent.

Property names are mapped to `AttributeEncodingContext` for precise error messages:
- `usrInfo` → `.userInfo` — "Failed to encode user info attribute 'X'"
- `accountInfo` → `.accountInfo` — "Failed to encode account attribute 'X'"
- `telemetryInfo`, `syntheticsInfo` → `.internal` — "Failed to encode internal attribute 'X'"
- `contextInfo`, `featureFlagsInfo`, others → `.custom` — "Failed to encode attribute 'X'"

The same fix is applied to the hand-written `encode(to:)` in `UserInfo.swift` and `AccountInfo.swift`.

### Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-- [IMPROVEMENT] Skip malformed RUM attributes individually instead of dropping the entire event, and log clear error messages. See [#xxx][]
+- [IMPROVEMENT] Skip malformed RUM attributes individually instead of dropping the entire event, and log clear error messages. See [#2844][]
 - [FEATURE] Add Mobile Heatmaps support. See [#2829][]
 - [FEATURE] Add watchOS and visionOS support. See [#2817][]
 - [FEATURE] Add support for the FedRAMP-compatible `fed2.ddog-gov.com` site. See [#2827][]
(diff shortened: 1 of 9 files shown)


maxep (approved):
Looks great, well done!

PR #2845, opened 2026-04-15
[RUM-8315] Add configurable redactedStatusCodes to URLSessionTracking

saraSr5:
### What and why?

By default, the tracer replaces the resource.name span tag with "404" for 404 responses, with no way to disable it. This PR surfaces a redactedStatusCodes: Set<Int> property on Trace.Configuration.URLSessionTracking so customers can control which HTTP status codes have their resource URL redacted.

### How?

  - Added redactedStatusCodes: Set<Int> (default [404]) to URLSessionTracking with a defaultRedactedStatusCodes static constant as the single source of truth for the default value
  - Threaded the value from config → Trace.swift → TracingURLSessionHandler                                                                                                                                          
  - Decoupled URL redaction from asClientError() in interceptionDidComplete, redaction now runs independently for any status code, not just 4xx
  - Added ObjC bridge via setRedactedStatusCodes(_ codes: [NSNumber])                                                                                                                                                
  - Regenerated API surface (Swift + ObjC)    

### Review checklist
- [x] Feature or bugfix MUST have appropriate tests (unit, integration) -Tests added (default, empty set, custom set, propagation, default value pinned)
- [x] Add Objective-C interface for public APIs - see our [guidelines](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3157787243/RFC+-+Modular+Objective-C+Interface#Recommended-solution) (internal) - Objective-C interface added


maxep (approved):
LGTM 👍

PR #2846, opened 2026-04-15
[PANA-7031] Gate heatmap identifier computation behind a Session Replay feature flag

gonzalezreal:
### What and why?

The heatmaps feature computes a heatmap identifier for every view in the UIKit hierarchy on every Session Replay recording tick. Since heatmaps is shipping as a preview feature, this computation should be opt-in.

### How?

Added a `.heatmaps` case to `SessionReplay.Configuration.FeatureFlag` (defaults to `false`) and used it to gate the heatmap computation:

- `ViewTreeSnapshotBuilder` stores `featureFlags` and only creates a `HeatmapCache` when `.heatmaps` is enabled. When disabled, `nil` is passed to the recording context and the registry is not updated.
- `ViewTreeRecorder` skips the heatmap identifier computation (MD5 hashing, path accumulation, cache writes) when `heatmapCache` is `nil`.
- `HeatmapCache` in `ViewTreeRecordingContext` is now optional, acting as the signal for whether heatmap computation should run.

When the flag is off, `permanentId` is omitted from wireframes and `dd.action.target.permanent_id` is absent from RUM action events. Both fields are already optional, so backends handle this gracefully.

Users opt in with:
```swift
SessionReplay.Configuration(featureFlags: [.heatmaps: true])
```

### Review checklist

### DIFF
--- a/DatadogSessionReplay/Sources/Recorder/Recorder.swift
+++ b/DatadogSessionReplay/Sources/Recorder/Recorder.swift
@@ -91,7 +91,8 @@ public class Recorder: Recording {
                     nodeRecorders: createDefaultNodeRecorders(featureFlags: featureFlags) + additionalNodeRecorders
                 ),
                 idsGenerator: NodeIDGenerator(),
-                core: core
+                core: core,
+                featureFlags: featureFlags
             )
         )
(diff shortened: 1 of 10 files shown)


arroz (approved):
Looks good! 🚀

maxep (approved):

PR #2847, opened 2026-04-15
RUM-15279: Propagate native anonymous_id to webview events

kikoveiga:
### What and why?

- Propagate the native `anonymousId` into webview RUM and Log events (overriding browser one).
- Why: [RUMS-5561](https://datadoghq.atlassian.net/browse/RUMS-5661)
- dd-sdk-android counterpart: [3306](https://github.com/DataDog/dd-sdk-android/pull/3306)

Tested on Shopist, here is an [example session](https://mobile-integration.datadoghq.com/rum/sessions?query=%40type%3Asession%20%40application.id%3A88ac1787-811c-467d-98d5-48247d1d11d7&agg_m=count&agg_m_source=base&agg_t=count&cols=&event=AwAAAZ2WRZGuqYDrOAAAABhBWjJXUlpHdUFBQXp3d3hwS1l4QVVQVGgAAAAkMDE5ZDk2NGYtNzBkYy00YmNmLWJiNDktZDNiZmJjNGEwYzRiAAAAAA&fromUser=false&graphType=flamegraph&shouldShowLegend=true&spanID=6532790809895426692&traceQuery=&from_ts=1775737865662&to_ts=1776342665662&live=true):
image

### Review checklist

[RUMS-5561]: https://datadoghq.atlassian.net/browse/RUMS-5561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/DatadogLogs/Sources/Feature/MessageReceivers.swift
+++ b/DatadogLogs/Sources/Feature/MessageReceivers.swift
@@ -96,6 +96,13 @@ internal struct WebViewLogReceiver: FeatureMessageReceiver {
                 event[LogEvent.Attributes.RUM.actionID] = rum.userActionID
             }
 
+            // Add native anonymous_id to the event's usr object
+            if let anonymousId = context.userInfo?.anonymousId {
+                var usr = event[LogEvent.Attributes.User.key] as? [String: Any] ?? [:]
+                usr[LogEvent.Attributes.User.anonymousId] = anonymousId
+                event[LogEvent.Attributes.User.key] = usr
+            }
+
             writer.write(value: AnyEncodable(event))
         }
 
--- a/DatadogLogs/Tests/WebViewLogReceiverTests.swift
+++ b/DatadogLogs/Tests/WebViewLogReceiverTests.swift
@@ -260,4 +260,102 @@ class WebViewLogReceiverTests: XCTestCase {
         XCTAssertNil(log["user_action.id"])
         XCTAssertTrue(telemetryReceiver.messages.isEmpty)
     }
+
+    // MARK: - Anonymous ID Propagation
+
+    func testWhenAnonymousIdIsAvailable_itAddsAnonymousIdAndPreservesExistingWebUsrFields() throws {
+        // Given
+        let messageReceiver = WebViewLogReceiver()
+        let fakeAnonymousId: String = .mockRandom()
+        let webUsrId: String = .mockRandom()
+        let webUsrName: String = .mockRandom()
+
+        let expectation = expectation(description: "Send log")
+        let core = PassthroughCoreMock(
+            context: .mockWith(
+                userInfo: UserInfo(anonymousId: fakeAnonymousId)
+            )
+        )
+        core.onEventWriteContext = { _ in expectation.fulfill() }
+
+        // When
+        XCTAssert(
+            messageReceiver.receive(
+                message: .webview(.log([
+                    "test": "value",
+                    "usr": ["id": webUsrId, "name": webUsrName]
+                ])),
+                from: core
+            )
+        )
+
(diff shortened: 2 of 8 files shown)


DatadogRUM/Sources/Integrations/WebViewEventReceiver.swift
```diff
@@ -123,6 +123,13 @@ internal final class WebViewEventReceiver: FeatureMessageReceiver {
                 event["_dd"] = dd
             }
 
+            // Add native anonymous_id to the event's usr object
+            if let anonymousId = context.userInfo?.anonymousId {
+                var usr = event["usr"] as? JSON ?? [:]
+                usr["anonymous_id"] = anonymousId
+                event["usr"] = usr
```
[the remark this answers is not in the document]
kikoveiga: Fixed.

DatadogLogs/Sources/Feature/MessageReceivers.swift
```diff
@@ -96,6 +96,15 @@ internal struct WebViewLogReceiver: FeatureMessageReceiver {
                 event[LogEvent.Attributes.RUM.actionID] = rum.userActionID
             }
 
+            // Add native anonymous_id to the event's usr object if not already present
+            if let anonymousId = context.userInfo?.anonymousId {
+                var usr = event[LogEvent.Attributes.User.key] as? [String: Any] ?? [:]
+                if usr[LogEvent.Attributes.User.anonymousId] == nil {
```
mariedm: Here we are first checking if `anonymousId` is present. But currently the PR on Android doesn't do this check, it simply overwrites it. I see there has been some conversations on the behavior to apply. What was decided in the end? We should have the same behavior on both platforms, which doesn't seem to be the case right now.

DatadogLogs/Tests/WebViewLogReceiverTests.swift
mariedm, line 1: For both unit test classes (here and in `WebViewEventReceiverTests`), it would be good to also test:
- when the `anonymousId` was already set by the browser SDK
- when no `usr` was set, to make sure we don't overwrite these values

mariedm (changes requested):
The PR looks good, but the current behavior seems to differ from its Android counterpart. The expected behavior should be clarified so that both platforms can align.

If this a bug fix, we should also add an entry to the changelog.

DatadogLogs/Sources/Feature/MessageReceivers.swift
kikoveiga: Thank you for pointing this out! Yes we should check if it's present, updated the Android pr.

0xnm: I think that **unconditional** overwrite is the valid behavior, because `anonymousId` is a persistent ID (we store it) for a _given_ RUM application ID.

We set mobile RUM application ID, it means we should always set mobile `anonymousId` if present as well.

But I'm open to discuss it.

mariedm: I agree, it makes sense

kikoveiga: Discussed with @cdn34dd, we also agree! Updating the PR, thank you!

DatadogLogs/Tests/WebViewLogReceiverTests.swift
kikoveiga: Added those, thank you! Also fixed the failing integration tests.

mariedm (approved):
Thank you! 🚀

PR #2859, opened 2026-04-21
RUM-15702 Add trace sampling decision to DatadogEventBridge

arroz:
### What and why?

With deterministic sampling, the decision of sampling requests on native clients became consistent, based on the sampling rate and session ID. However, when a native client loads an instrumented web site on a WebView, there is a disconnect between the sampling decision used by the browser SDK of the page being viewed in the WebView, and the native client's decision. This cannot be fixed by post-processing on the native client when the data is funneled through it, since the browser SDK needs to know if headers should be injected on requests, or not. More details [here](https://datadoghq.atlassian.net/browse/RUM-15702) (internal).

This is the iOS counterpart to https://github.com/DataDog/browser-sdk/pull/4516.

### How?

To fix this problem, this patch introduces a way for iOS to send the sampling decision to the Browser SDK running on the WebView through the JS bridge.

On the JS side, depending on the decision being injected, tracing sample rate will be set to 100 (when the decision is `true`) or 0 (when it's `false`). Because of that, the iOS SDK re-writes the `_dd.rule_psr` value with the actual sampling rate configured in the iOS SDK.

### Review checklist


arroz:
@codex review

arroz:
@codex review

arroz:
@codex review

arroz:
@codex review

DatadogInternal/Sources/NetworkInstrumentation/DistributedTracingSampleRateProvider.swift
```diff
@@ -0,0 +1,27 @@
...
+    /// was not configured in RUM.
+    var distributedTracingSampleRate: SampleRate? { get }
+}
+
+extension NetworkInstrumentationFeature: DistributedTracingSampleRateProvider {
+    var distributedTracingSampleRate: SampleRate? {
+        guard let urlSessionHandler = handlers
+            .lazy
+            .compactMap({ $0 as? DatadogURLSessionHandlerSupportingDistributedTracing })
+            .first
```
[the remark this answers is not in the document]
arroz: This cannot happen, `RUM._internal.enableURLSessionTracking` is only called by other SDKs that run on top of iOS, and we either have a configuration or a late start, not both.

DatadogInternal/Sources/Concurrency/Threading.swift
```diff
@@ -10,9 +10,19 @@ import Foundation
...
-public func runOnMainThreadSync<T>(_ block: () throws -> T) rethrows -> T {
+public func runOnMainThreadSync<T>(_ block: @MainActor () throws -> T) rethrows -> T {
     if Thread.isMainThread {
-        return try block()
+        if #available(iOS 13.0, *) {
+            return try MainActor.assumeIsolated {
+                return try block()
+            }
+        } else {
+            // Does the same as MainActor.assumeIsolated on iOS 12. 🙈
```
arroz: Sorry about this… I need to enforce some stuff runs on the main thread so I had to do this. iOS 12 support can't go away soon enough.

DatadogRUM/Sources/RUMMonitor/Monitor.swift
```diff
@@ -152,9 +157,7 @@ internal class Monitor: RUMCommandSubscriber {
                     return nil
                 }
 
-                let context = activeSession.viewScopes.last?.context ??
-                                activeSession.context ??
-                                self.scopes.context
+                let context = activeSession.viewScopes.last?.context ?? activeSession.context
```
arroz, line 155: `activeSession.context` is never nil at this point, so this was causing a compiler warning.

DatadogWebViewTracking/Sources/WebViewTracking.swift
```diff
@@ -60,23 +60,33 @@ public enum WebViewTracking {
...
-    /// - Note: This method **must** be called when the webview can be deinitialized.
-    /// 
-    /// - Parameter webView: The web-view to stop tracking.
-    public static func disable(webView: WKWebView) {
+    /// - Note: This method **must** be called when the WebView can be deinitialized.
+    ///
+    /// - Parameters:
+    ///   - webView: The web-view to stop tracking.
+    ///   - core: Datadog SDK core where the WebView was tracked.
+    public static func disable(webView: WKWebView, in core: DatadogCoreProtocol = CoreRegistry.default) {
```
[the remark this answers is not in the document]
arroz: This is not a problem in our use case, iOS apps are always built against the SDKs.

DatadogWebViewTracking/Sources/WebViewTracking.swift
```diff
@@ -143,26 +180,160 @@ public enum WebViewTracking {
...
+
+        // Run code to update the current page
+        let js =
+        """
+        if (window.\(DDScriptMessageHandler.name)) {
+            window.\(DDScriptMessageHandler.name).getIsTraceSampled = () => '\(isTraceSampled)'
+        }
+        """
+
+        webView.evaluateJavaScript(js)
```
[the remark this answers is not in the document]
arroz: Ok, we're supporting iFrames and *nested* iFrames now. I hope you're happy. 😛

arroz:
@codex review

DatadogInternal/Sources/Models/RUM/RUMSessionSamplerProvider.swift
```diff
@@ -0,0 +1,19 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+public let RUMFeatureName = "rum"
```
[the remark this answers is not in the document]
arroz: That's not accurate, most files in that directory are hand written.

DatadogInternal/Sources/Models/RUM/RUMSessionSamplerProvider.swift
```diff
@@ -0,0 +1,19 @@
...
+
+/// Provides the RUM session deterministic sampler for the active session.
+public protocol RUMSessionSamplerProvider {
+    /// The RUM session deterministic sampler for the active session. `nil` if there is no active session.
+    var rumSessionSampler: DeterministicSampler? { get }
+}
+
+public extension DatadogFeature where Self: RUMSessionSamplerProvider {
+    static var name: String { RUMFeatureName }
```
simaoseica-dd: This definition is repeated in `RUMFeature`. Can `DatadogInternal` avoid this link with RUM and be more agnostic?

DatadogInternal/Sources/NetworkInstrumentation/NetworkInstrumentationFeature.swift
```diff
@@ -6,6 +6,8 @@
 
 import Foundation
 
+public let NetworkInstrumentationFeatureName = "network-instrumentation"
```
simaoseica-dd: nit: These spread constants would benefit from being scoped: `Feature.networkInstrumentation`

DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
```diff
@@ -33,7 +33,7 @@ internal struct DistributedTracing {
     }
 }
 
-internal final class URLSessionRUMResourcesHandler: DatadogURLSessionHandler, RUMCommandPublisher {
+internal final class URLSessionRUMResourcesHandler: DatadogURLSessionHandlerSupportingDistributedTracing, RUMCommandPublisher {
```
simaoseica-dd, line 36: Why isn't `DatadogURLSessionHandlerSupportingDistributedTracing` also extending "TracingURLSessionHandler" since it has also a role on distributed tracing?

DatadogRUM/Sources/Feature/RUMFeature.swift
```diff
@@ -23,6 +23,10 @@ internal final class RUMFeature: DatadogRemoteFeature {
 
     let anonymousIdentifierManager: AnonymousIdentifierManaging
 
+    /// Used by WebViewTracking to obtain the RUM session sampler synchronously.
+    @ReadWriteLock
+    var rumSessionSampler: DeterministicSampler?
```
simaoseica-dd:
```suggestion
    private(set) var rumSessionSampler: DeterministicSampler?
```

DatadogRUM/Sources/Integrations/WebViewEventReceiver.swift
```diff
@@ -117,10 +119,21 @@ internal final class WebViewEventReceiver: FeatureMessageReceiver {
...
+                    dd["replay_stats"] = nil
+                    event["_dd"] = dd
+                }
+                if dd["rule_psr"] != nil,
+                   let networkInstrumentation = core.feature(
+                    named: NetworkInstrumentationFeatureName,
+                    type: DistributedTracingSampleRateProvider.self
+                   ),
+                   let distributedTracingSampleRate = networkInstrumentation.distributedTracingSampleRate {
+                    dd["rule_psr"] = distributedTracingSampleRate.percentageProportion
```
simaoseica-dd, line 134: question: Why is the `rule_psr` overwritten? Doesn't BrowserSDK have everything for this `rule_psr` to be correct?

DatadogRUM/Sources/RUMMonitor/Monitor.swift
```diff
@@ -136,6 +144,8 @@ internal class Monitor: RUMCommandSubscriber {
 
             _ = self.scopes.process(command: transformedCommand, context: context, writer: writer)
 
+            self.onActiveSessionUpdate(self.scopes.activeSession?.sampler)
```
simaoseica-dd: This callback might be called too late and the Monitor should be agnostic of samplers or sessions. 

My suggestion, Monitor stays clean, the sampler update is done on the `RUMApplicationScope`  that has session ownership and detects session lifecycle changes.
The `RUMScopeDependencies` would have an entity to `setActiveSessionSampler` and `RUMFeature` could look into that.

DatadogWebViewTracking/Sources/WebViewSessionRolloverHandler.swift
```diff
@@ -0,0 +1,178 @@
...
+         be unregistered, if any, in a set, and delete them after the fact.
+
+         - NSMapTable.dictionaryRepresentation() cannot be used because it throws an
+         exception, since it requires the map Key type to implement NSCopying and WKWebView
+         does not. This hints at another reason to not use dictionaryRepresentation:
+         it creates an entire copy of the dictionary, which is wasteful.
+
+         Using weak boxes on modern Swift collections was considered, but due to the
+         manual work required to remove the boxes containing deallocated objects from
+         the collection, it would be even grotesquier.
```
simaoseica-dd, line 75: Why grotesquier? The Swift implementation using weak boxes would be more Swifty and readable, without exposing ObjC semantics.
And the deallocated webView entries  could be removed in the loop of this func

TestUtilities/Sources/Mocks/DatadogRUM/RUMFeatureMocks.swift
```diff
@@ -183,7 +183,7 @@ extension WebViewEventReceiver: AnyMockable {
             featureScope: featureScope,
             dateProvider: dateProvider,
             commandSubscriber: commandSubscriber,
-            viewCache: viewCache
+            viewCache: viewCache,
```
simaoseica-dd:
```suggestion
            viewCache: viewCache
```

DatadogWebViewTracking/Sources/WebViewTracking.swift
```diff
@@ -120,6 +128,35 @@ public enum WebViewTracking {
...
+        try WebViewSessionRolloverHandler.register(webView: webView, in: core, using: elements)
+
+        injectUserScript(on: webView, in: core, using: elements, isTraceSampled: isTraceSampled)
+
+        core.telemetry.usage(event: .trackWebView)
+    }
+
+    /// Injects the Javascript bridge code in the WebView user scripts.
+    ///
+    /// - Important: This function does not check if the script is already there and should be called *only* when
```
simaoseica-dd, line 138: The Datadog bridge is installed on `webView.configuration.userContentController`, not really on the WKWebView. If multiple webviews share the same controller, the script might already be there and fail to set. This happened in the past escalations.

DatadogWebViewTracking/Sources/WebViewTracking.swift
```diff
@@ -120,6 +128,35 @@ public enum WebViewTracking {
             .map { return "\"\($0)\"" }
             .joined(separator: ",")
 
+        let elements = WebViewTrackingElements(allowedWebViewHostsString: allowedWebViewHostsString)
+        let isTraceSampled = WebViewTracking.isTraceSampledStringValue(for: core)
+
+        try WebViewSessionRolloverHandler.register(webView: webView, in: core, using: elements)
```
simaoseica-dd, line 129: The register by core, opening the possibility to associate the webView with multiple cores, might lead to misuse by the customer and unexpected behavior. It would be better to register only for one core at a time.

simaoseica-dd (commented):
Left a few comments. My main concern is that `WebViewTracking` has now a webview and core association, which makes it easy to misuse and can lead to inconsistent behavior. 
Also, I think there's opportunity to better scope responsibilities across the RUM entities.

DatadogInternal/Sources/NetworkInstrumentation/DistributedTracingSampleRateProvider.swift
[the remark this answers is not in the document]
maxep: Yes, but the `first` filter is still problematic, because it assumes that only one `DistributedTracingSampleRateProvider` is present. It is the case in practice, only RUM is compliant, but it is still an assumption at this level. We need to make the code path explicit because it is very difficult to navigate: I would suggest to merge `RUMSessionSamplerProvider` and `DistributedTracingSampleRateProvider` into one `RUMSampleRateProvider` which will contain both.

DatadogWebViewTracking/Sources/WebViewTracking.swift
```diff
@@ -143,26 +180,168 @@ public enum WebViewTracking {
...
+                return '[\(elements.allowedWebViewHostsString)]'
             },
             getCapabilities() {
                 return '[\(capabilities)]'
             },
             getPrivacyLevel() {
                 return '\(privacyLevel.rawValue)'
+            },
+            getIsTraceSampled() {
+                return '\(isTraceSampled)'
```
maxep, line 187: How important is it to have this at enablement? could this be set async?
If so, we could solely rely on the core context and message-bus to inject the `isTraceSampled`, without the need to extend the `DatadogFeature`.

DatadogRUM/Sources/RUMMonitor/Monitor.swift
```diff
@@ -111,16 +111,24 @@ internal class Monitor: RUMCommandSubscriber {
     private let rumUUIDGenerator: RUMUUIDGenerator
     private let telemetry: Telemetry
 
+    /// When the active session changes, this function should be called with the session's sampler.
+    ///
+    /// Pass the session's `DeterministicSampler` or `nil` if there is no active session (usually
+    /// as a result of a call to `Monitor.stopSession()`.
+    private let onActiveSessionUpdate: (DeterministicSampler?) -> Void
```
maxep: I think you should re-use the existing `onSessionStart` callback. It is called when the `RUMSessionState` changes which include the sampler.

maxep (commented):
I was not expected such level of complexity to be honest. It highlights missing capabalities in the core that we should address in my opinion, such as:
- message-bus registration management (even subscription)
- sync access to core context

Not to be done in this PR but we should work on new specs for the core.

DatadogWebViewTracking/Sources/WebViewTracking.swift
```diff
@@ -60,23 +60,28 @@ public enum WebViewTracking {
...
+    /// - Parameters:
+    ///   - webView: The web-view to stop tracking.
     public static func disable(webView: WKWebView) {
-        let controller = webView.configuration.userContentController
-        controller.removeScriptMessageHandler(forName: DDScriptMessageHandler.name)
-        let others = controller.userScripts.filter { !$0.source.starts(with: Self.jsCodePrefix) }
-        controller.removeAllUserScripts()
-        others.forEach(controller.addUserScript)
+        runOnMainThreadSync {
+            let controller = webView.configuration.userContentController
```
arroz, line 69: This (lines 69-73) was intentionally moved into the `runOnMainThreadSync` since, by consistency, we should always do these modifications in the same thread.

DatadogInternal/Sources/NetworkInstrumentation/DistributedTracingSampleRateProvider.swift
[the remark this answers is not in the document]
arroz: I'm handwaving this for now since I have a ticket to improve the `DatadogURLSessionHandler` situation (including the problems created by both being on at the same time) and I'll likely follow Android's SDK approach of moving all this to a unique implementation, so those two classes will disappear at least the way they exist today. Since the scope of this PR is already way too large, I prefer to leave it like this for now.

DatadogRUM/Sources/Instrumentation/Resources/URLSessionRUMResourcesHandler.swift
arroz: I'm handwaving this for now since I have a ticket to improve the DatadogURLSessionHandler situation (including the problems created by both being on at the same time) and I'll likely follow Android's SDK approach of moving all this to a unique implementation, so those two classes will disappear at least the way they exist today. Since the scope of this PR is already way too large, I prefer to leave it like this for now.

arroz: (Also, part of the reason is `URLSessionRUMResourcesHandler` contains a `DistributedTracing` that `TracingURLSessionHandler` does not, which is questionable but just another reason it needs to be cleaned up).

DatadogRUM/Sources/Integrations/WebViewEventReceiver.swift
arroz: It doesn't. Per business, the native SDK should override the trace sampling rate configured in the Browser SDK. If the Browser SDK already supported distributed tracing, this would be a matter of just passing the rate. However, it doesn't, and we need this feature ASAP. So the decision made with the Browser SDK team was passing the sampling decision to the Browser SDK as a simple boolean, that will fake the sampling on the Browser side to either 0 or 100. Because of this, the `rule_psr` that comes from the browser is one of those values, and we need to overwrite with the correct one.

DatadogWebViewTracking/Sources/WebViewSessionRolloverHandler.swift
arroz: There are multiple reasons:
- Based on a conversation I had a while ago with a member of the Swift stdlib team, the reason they did not yet implement a weak collection in Swift is it breaks a lot of the language model. By its nature, a weak collection cannot be immutable since it must mutate when a value is deallocated (it may not technically mutate at that exact time, since the language offers no way to trigger code when an object held by a weak ref is deallocated, but it mutates conceptually). However in Swift collections (and generally value types) must support immutability. Although this could be trivially solved by implementing as a class, that would be very problematic when implementing all the Collection protocols.
- I'm using weak _keys_, not (just) values. A dictionary key must maintain its hash and consistency in the result of equality testing per contract, otherwise the dictionary will corrupt. So a weak box for a key would have to keep its own values for hash and use the object pointer for comparison.

The `NSMapTable` solves this by:
1. Not being a real Swift collection, and not using one for its implementation, as we would have to use if we created a Swift weak collection.
2. The weak boxes of `NSMapTable` do exactly what I described already.

Given that an implementation of this in Swift would not bring any real benefit, could not comply with any of the Collection protocols since it cannot implement them properly (the inconsistency between `count` and object iteration that `NSMapTable` presents would also exist in the Swift version, and if it implemented the `Collection` protocols would cause all sorts of troubles), would have to use the exact same tricks `NSMapTable` uses, there is no value in implementing and maintaining that code with `NSMapTable` exists and has been battle-tested for more than a decade (and supports many options like weak keys and/or values at will).

DatadogWebViewTracking/Sources/WebViewTracking.swift
arroz: A `WebView` cannot be registered in multiple cores, both conceptually (cores can have different configurations and the `WebView` cannot work differently depending on the core it's sending events to), and practically (the code now unregisters it from one core before registering it in another). See the implementation of the method called above.

DatadogWebViewTracking/Sources/WebViewTracking.swift
arroz: We cannot support `userContentController` sharing. Right now the code does a best effort to prevent that from happening (by not instrumenting a second `WebView` sharing a controller with one already instrumented). But this cannot happen conceptually, since each `WebView.enable` call will try to configure the controller to instrument based on the parameters passed in to that call. In the current SDK version, this will only work by accident, if the parameters passed to the `enable` call are the same. Otherwise the behaviour is already incorrect.

I don't propose solving this problem now, since this PR is already quite complex, but I propose solving this by triggering a fatal assertion in dev mode if users try to share controllers between WebViews, and a log/telemetry at runtime so we can easily know why tracking is not working as expected in that edge case.

DatadogWebViewTracking/Sources/WebViewTracking.swift
arroz: It must be done synchronously, otherwise the `WebView` will not be configured correctly between the moment it's instrumented and the context propagates. This is annoying but required, since the `WebViewTrackingFeature` is created and registered on the fly, not at app startup time, with ample time for messages to arrive.

Datadog/IntegrationUnitTests/Public/WebLogIntegrationTests.swift
```diff
@@ -10,11 +10,13 @@ import XCTest
...
+import WebKit
 
 @testable import DatadogLogs
 @testable import DatadogRUM
 @testable import DatadogWebViewTracking
 
+@MainActor
 class WebLogIntegrationTests: XCTestCase {
     private var core: DatadogCoreProxy! // swiftlint:disable:this implicitly_unwrapped_optional
     private var controller: WKUserContentControllerMock! // swiftlint:disable:this implicitly_unwrapped_optional
```
maciejburda, line 22: we can probably remove controller, maybe replace with config? (here in the rest of the tests)

DatadogInternal/Sources/Concurrency/Threading.swift
maciejburda: I would put something more descriptive. Maybe something along the lines of:
```swift
// This erases the @MainActor annotation from `block` and calls it synchronously.
// Equivalent in intent to `MainActor.assumeIsolated { try block() }`,
// but unlike `assumeIsolated`, this performs no runtime check that we are
// actually on the MainActor, so it is only safe if the caller guarantees that.
```

DatadogInternal/Sources/Feature.swift
```diff
@@ -0,0 +1,15 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+public enum Feature {
```
maciejburda, line 9: I like that. Actually makes me thing that we should have all of them in `DatadogInternal`

DatadogWebViewTracking/Sources/WebViewTracking.swift
```diff
@@ -88,11 +93,12 @@ public enum WebViewTracking {
             )
         }
 
+        let controller = webView.configuration.userContentController
         let isTracking = controller.userScripts.contains { $0.source.starts(with: Self.jsCodePrefix) }
         guard !isTracking else {
             DD.logger.warn("`startTrackingDatadogEvents(core:hosts:)` was called more than once for the same WebView. Second call will be ignored. Make sure you call it only once.")
             return
-       }
+        }
```
maciejburda, line 101:
```suggestion
       }
```

maciejburda (dismissed):
LGTM!

Maybe feature introduction and the implementation could have been split to two PRs, but not a big deal.

I think the answers to Maxime's and Simão's comments are reasonable.

Left some nits.

DatadogWebViewTracking/Sources/WebViewTracking.swift
arroz: The added space was intentional, since it was missing before.

maciejburda (dismissed):
LGTM

maciejburda (dismissed):

maciejburda (dismissed):

ncreated (dismissed):

ncreated (dismissed):

ncreated (approved):

PR #2864, opened 2026-04-23
Enforce schema character set on vital.name in Feature Operation APIs

Valpertui:
### What and why?

The authoritative `_vital-common-schema.json` restricts `vital.name` to the character set `[\w.@$-]*` (letters, digits, `_`, `.`, `@`, `$`, `-`) and the backend enforces this as a server-side regex on the facet path. Until now the iOS SDK only rejected blank names; non-conforming values (e.g. `"user login"`, `"api/v1"`, `"ログイン"`) were silently accepted and shipped to intake.

This PR validates the character set at the API boundary so developers catch the problem at dev time instead of discovering it only after events are dropped at intake.

### How?

Split `RUMFeatureOperationManager`'s single `validateString` helper into:

- `validateName` — rejects blank names with `DD.logger.error` (drops the event), then checks each unicode scalar against an explicit ASCII `CharacterSet(charactersIn: "A-Za-z0-9_.@$-")`. Using `.alphanumerics` would have been wrong — it is Unicode-aware and would accept Japanese/Cyrillic/etc.
- `validateOperationKey` — blank-check only (schema does not restrict `operation_key`).

When `name` contains characters outside the allowed set, `DD.logger.warn` is emitted quoting the exact backend pattern `[\w.@$-]*`, but the event is still emitted. The rationale: the backend is the source of truth on the policy, so a client-side drop would force customers to bump the SDK if the server rule is ever relaxed.

Updated `RUMFeatureMocks.mockRandom()` to use `.alphanumerics` so existing forgery-based tests stay within the allowed set.

Tests cover:

- blank / whitespace-only name → dropped + error
- invalid ASCII chars (space, slash, colon, comma, plus, tab) → warn + emitted
- non-ASCII name `"ログイン"` and emoji `"login🔐"` → warn + emitted (pins ASCII-only semantics against future `.alphanumerics` regressions)
- every allowed-character class → silent emit
- `operationKey` with space/slash/etc. → silent emit (no char-set rule)
- applies to all three methods: start/succeed/fail

### Review checklist
- [x] Add Objective-C interface for public APIs — not applicable, change is internal
- [x] Run `make api-surface` when adding new APIs — not applicable, no new public API

### DIFF
--- a/DatadogRUM/Sources/RUMMonitor/Scopes/RUMFeatureOperationManager.swift
+++ b/DatadogRUM/Sources/RUMMonitor/Scopes/RUMFeatureOperationManager.swift
@@ -176,8 +176,9 @@ internal class RUMFeatureOperationManager {
     /// (which `CharacterSet.alphanumerics` would pull in) do not mask
     /// non-conforming names — e.g. `ログイン` is all Unicode "Letter, other"
     /// and must still trigger the warning.
-    private static let validOperationNameCharacters: CharacterSet =
-        CharacterSet(charactersIn: "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_.@$-")
+    private static let validOperationNameCharacters = CharacterSet(
+        charactersIn: "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_.@$-"
+    )
 
     /// Validates the operation name.
     ///
--- a/DatadogRUM/Sources/RUMMonitor/Scopes/RUMFeatureOperationManager.swift
+++ b/DatadogRUM/Sources/RUMMonitor/Scopes/RUMFeatureOperationManager.swift
@@ -157,30 +157,61 @@ internal class RUMFeatureOperationManager {
     /// Validates the `name` and `operationKey` of the command
     private func validateCommand(_ command: RUMOperationStepVitalCommand) -> Bool {
         // Validate name (required)
-        guard validateString(command.name, stepType: command.stepType) else {
+        guard validateName(command.name, stepType: command.stepType) else {
             return false
         }
 
         // Validate operationKey if present (optional)
         if let operationKey = command.operationKey,
-           !validateString(operationKey, stepType: command.stepType) {
+           !validateOperationKey(operationKey, stepType: command.stepType) {
             return false
         }
 
         return true
     }
 
-    /// Validates the string is not empty
-    // or contains only whitespace/line breaks
-    private func validateString(_ value: String, stepType: RUMVitalOperationStepEvent.Vital.StepType) -> Bool {
+    /// ASCII character set accepted by the backend for `vital.name`, matching
+    /// the server-side regex `[\w.@$-]` (letters, digits, `_`, `.`, `@`, `$`,
+    /// `-`). Built explicitly from ASCII so that Unicode-aware categories
+    /// (which `CharacterSet.alphanumerics` would pull in) do not mask
+    /// non-conforming names — e.g. `ログイン` is all Unicode "Letter, other"
+    /// and must still trigger the warning.
+    private static let validOperationNameCharacters: CharacterSet =
+        CharacterSet(charactersIn: "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_.@$-")
+
+    /// Validates the operation name.
+    ///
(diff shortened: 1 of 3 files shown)


DatadogRUM/Sources/RUMMonitor/Scopes/RUMFeatureOperationManager.swift
```diff
@@ -157,30 +157,61 @@ internal class RUMFeatureOperationManager {
...
-            DD.logger.error("Operation `name` and `operationKey` cannot be empty or contain only whitespace/line breaks. \(stepType) command will be ignored.")
+            DD.logger.error("Operation `name` cannot be empty or contain only whitespace/line breaks. \(stepType) command will be ignored.")
+            return false
+        }
+
+        if !value.unicodeScalars.allSatisfy(Self.validOperationNameCharacters.contains) {
+            DD.logger.warn("Operation `name` '\(value)' does not match the backend-accepted pattern [\\w.@$-]* (letters, digits, _ . @ $ -). The \(stepType) command will still be emitted and may be rejected by the backend.")
+        }
+
+        return true
```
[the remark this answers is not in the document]
Valpertui: It's on purpose we still send the event, we don't want to block the event. If the backend relaxes the requirement, the event will then be accepted.

DatadogRUM/Sources/RUMMonitor/Scopes/RUMFeatureOperationManager.swift
```diff
@@ -157,30 +157,62 @@ internal class RUMFeatureOperationManager {
...
+    }
+
+    /// Validates the operation key: non-blank. The schema does not restrict
+    /// the character set for `operation_key`.
+    private func validateOperationKey(_ value: String, stepType: RUMVitalOperationStepEvent.Vital.StepType) -> Bool {
+        let trimmed = value.trimmingCharacters(in: .whitespacesAndNewlines)
+
+        guard !trimmed.isEmpty else {
+            DD.logger.error("Operation `operationKey`, when provided, cannot be empty or contain only whitespace/line breaks. \(stepType) command will be ignored.")
             return false
```
simaoseica-dd: Shouldn't the validation of the `operationKey` be relaxed? It is an optional parameter, so it could also be ignored if empty. This way the operation step would not be discarded completely.

DatadogRUM/Tests/RUMMonitor/RUMFeatureOperationManagerTests.swift
```diff
@@ -148,6 +150,81 @@ class RUMFeatureOperationManagerTests: XCTestCase {
...
+                command,
+                context: mockContext,
+                writer: mockWriter,
+                activeView: .mockAny()
+            )
+        }
+
+        // Then — every event is emitted; the name is preserved verbatim.
+        let vitalEvents = mockWriter.events(ofType: RUMVitalOperationStepEvent.self)
+        XCTAssertEqual(vitalEvents.count, invalidCharacterSetNames.count)
```
simaoseica-dd, line 188: Can you validate the warning messages being logged with `DD.mockWith(logger: CoreLoggerMock())`?

DatadogRUM/Tests/RUMMonitor/RUMFeatureOperationManagerTests.swift
```diff
@@ -148,6 +150,81 @@ class RUMFeatureOperationManagerTests: XCTestCase {
...
+                command,
+                context: mockContext,
+                writer: mockWriter,
+                activeView: .mockAny()
+            )
+        }
+
+        // Then
+        let vitalEvents = mockWriter.events(ofType: RUMVitalOperationStepEvent.self)
+        XCTAssertEqual(vitalEvents.count, validNames.count)
```
simaoseica-dd, line 220: And here no warnings in the `dd.logger.warnLog`

Valpertui: Fixed in latest version

DatadogRUM/Tests/RUMMonitor/RUMFeatureOperationManagerTests.swift
Valpertui: Fixed in latest version

DatadogRUM/Sources/RUMMonitor/Scopes/RUMFeatureOperationManager.swift
Valpertui: Yes, good catch

simaoseica-dd (approved):
Looks good!

PR #2868, opened 2026-04-27
RUM-14274: Create continuous Benchmarks for continuous Profiling

simaoseica-dd:
### What and why?

This PR adds an interactive Profiling benchmark scenario to `BenchmarkTests` so we can generate deterministic profiling signals on demand. The scenario lets us explicitly trigger long tasks, app hangs and RUM operations from the UI, which makes continuous profiling behavior easier to benchmark and compare across runs.

### How?

`ProfilingScenario` now enables RUM with explicit long-task and app-hang thresholds, tags data with the `Profiling` scenario attribute, and enables Profiling with max sample rates for both app launch and continuous profiling. `ProfilingContentView` was expanded into a small control panel where the user can select a signal type, configure durations and operation settings, and send one signal per tap.

### Review checklist

### DIFF
--- a/BenchmarkTests/Runner/Scenarios/Profiling/ProfilingScenario.swift
+++ b/BenchmarkTests/Runner/Scenarios/Profiling/ProfilingScenario.swift
@@ -1,0 +1,37 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+import SwiftUI
+
+import DatadogCore
+import DatadogRUM
+import DatadogProfiling
+
+struct ProfilingScenario: Scenario {
+    var initialViewController: UIViewController {
+        UIHostingController(rootView: ProfilingContentView())
+    }
+
+    func instrument(with info: AppInfo) {
+        Datadog.initialize(
+            with: .benchmark(info: info),
+            trackingConsent: .granted
+        )
+
+        RUM.enable(
+            with: RUM.Configuration(
+                applicationID: info.applicationID,
+                longTaskThreshold: 0.1,
+                appHangThreshold: 0.4
+            )
+        )
+
+        RUMMonitor.shared().addAttribute(forKey: "scenario", value: "Profiling")
+
+        Profiling.enable(with: .init(applicationLaunchSampleRate: .maxSampleRate, continuousSampleRate: .maxSampleRate))
+    }
+}
--- a/BenchmarkTests/Runner/Scenarios/Profiling/ProfilingScenario.swift
+++ b/BenchmarkTests/Runner/Scenarios/Profiling/ProfilingScenario.swift
@@ -30,7 +30,7 @@ struct ProfilingScenario: Scenario {
             )
         )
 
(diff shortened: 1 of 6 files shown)


BenchmarkTests/Runner/Scenarios/Profiling/UI/ProfilingContentView.swift
```diff
@@ -0,0 +1,305 @@
...
+            "scenario": "Profiling"
+        ]
+
+        rumMonitor.startOperation(
+            name: name,
+            attributes: attributes,
+            options: ProfilingOptions(sampleRate: .maxSampleRate)
+        )
+
+        runHighProcessingBlock(for: configuration.operationDuration)
```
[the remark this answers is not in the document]
simaoseica-dd: Not needed, but can be another approach to route work

BenchmarkTests/Runner/Scenarios/SyntheticScenario.swift
```diff
@@ -19,6 +19,7 @@ internal struct SyntheticScenario: Scenario {
         case trace
         case rumManual
         case rumAuto
+        case profiling
```
mariedm: nit: given we have an `profiling` case for the `SyntheticRun` enum, we could name it differently to avoid confusion, e.g., `continuousProfiling`

BenchmarkTests/Runner/Scenarios/Profiling/ProfilingScenario.swift
```diff
@@ -0,0 +1,37 @@
...
+
+        RUM.enable(
+            with: RUM.Configuration(
+                applicationID: info.applicationID,
+                longTaskThreshold: 0.1,
+                appHangThreshold: 0.4
+            )
+        )
+
+        RUMMonitor.shared().addAttribute(forKey: "scenario", value: "Profiling")
```
mariedm: nit: this is also set at the operation level when calling `startOperation`

BenchmarkTests/Runner/Scenarios/Profiling/UI/ProfilingContentView.swift
```diff
@@ -0,0 +1,307 @@
...
+        }
+
+        recordSignal(
+            label: configuration.shouldFailOperation ? "Failed operation" : "Operation",
+            duration: configuration.operationDuration
+        )
+    }
+
+    @discardableResult
+    func runHighProcessingBlock(for duration: TimeInterval) -> Double {
```
mariedm, line 248: nit: can be `private` or `static`

mariedm (dismissed):
LGMT! Only nits, nothing blocking.

BenchmarkTests/Runner/Scenarios/SyntheticScenario.swift
simaoseica-dd: Agree. We should avoid confusion 👍

BenchmarkTests/Runner/Scenarios/Profiling/UI/ProfilingContentView.swift
simaoseica-dd: It is already in a private extension

BenchmarkTests/Runner/Scenarios/Profiling/ProfilingScenario.swift
simaoseica-dd: good catch

mariedm (approved):

PR #2869, opened 2026-04-27
RUM-15698: Align profiling timestamps with RUM server time offset

simaoseica-dd:
### What and why?

This PR aligns Datadog Profiling timestamps with the SDK server time offset used by RUM, so profiles and attached RUM events use consistent wall-clock dates when device time drifts from server time.

### How?

Profiling samples now keep their raw uptime timestamp and the C++ profiler stores the latest server time offset. During `pprof` serialization, each sample's `end_timestamp_ns` label is computed from raw uptime plus the epoch offset and server time offset.

Swift propagates the server time offset to the C++ profiler when context updates are received and applies the current offset again before profile export as a final guard. App launch profiling uses the offset carried by the TTID vital so the harvested profile and attached RUM vital are aligned. RUM app launch vital and RUM operation vitals now carry the server offset when encoding their start timestamps.

### Review checklist

### DIFF
--- a/DatadogInternal/Sources/Models/RUM/Vital.swift
+++ b/DatadogInternal/Sources/Models/RUM/Vital.swift
@@ -27,7 +27,7 @@ public struct Vital: Encodable, Equatable {
     /// Date when the vital was created
     public let date: Date
     /// Interval between device and server time.
-    public var serverTimeOffset: TimeInterval
+    public let serverTimeOffset: TimeInterval
     /// Duration of the vital in nanoseconds
     public var duration: Int64?
     /// Key identifier of the vital
--- a/DatadogInternal/Sources/Models/RUM/Vital.swift
+++ b/DatadogInternal/Sources/Models/RUM/Vital.swift
@@ -26,6 +26,8 @@ public struct Vital: Encodable, Equatable {
     public let stepType: RUMVitalOperationStepEvent.Vital.StepType?
     /// Date when the vital was created
     public let date: Date
+    /// Interval between device and server time.
+    public var serverTimeOffset: TimeInterval
     /// Duration of the vital in nanoseconds
     public var duration: Int64?
     /// Key identifier of the vital
@@ -37,6 +39,7 @@ public struct Vital: Encodable, Equatable {
         operationKey: String? = nil,
         stepType: RUMVitalOperationStepEvent.Vital.StepType? = nil,
         date: Date = Date(),
+        serverTimeOffset: TimeInterval = .zero,
         duration: Int64? = nil
     ) {
         self.id = id
@@ -44,6 +47,7 @@ public struct Vital: Encodable, Equatable {
         self.operationKey = operationKey
         self.stepType = stepType
         self.date = date
+        self.serverTimeOffset = serverTimeOffset
         self.duration = duration
         self.key = "\(name)-\(operationKey ?? "")"
     }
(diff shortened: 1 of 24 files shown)


DatadogProfiling/Mach/include/profile.h
```diff
@@ -177,12 +179,24 @@ class profile {
...
     int64_t start_timestamp() const { return uptime_ns_to_epoch_ns(_start_timestamp); };
 
     /** @brief Get profile end timestamp (uptime nanoseconds converted to epoch) */
     int64_t end_timestamp() const { return uptime_ns_to_epoch_ns(_end_timestamp); };
 
+    /**
+     * @brief Set the server time correction used for exported timestamps.
+     * @param offset_ns Server time offset in nanoseconds
+     */
+    void set_server_time_offset_ns(int64_t offset_ns);
```
maxep: nit/ This can be inline like other accessor methods

```cpp
    void profile::set_server_time_offset_ns(int64_t offset_ns) { _server_time_offset_ns = offset_ns; };
```

DatadogProfiling/Mach/profile_pprof_packer.cpp
```diff
@@ -282,22 +283,31 @@ void perftools_profiles_profile_add_samples(const std::vector<sample_t>& samples
...
+
+        auto* timestamp_label = static_cast<Perftools__Profiles__Label*>(
+            pb_alloc(allocator, sizeof(Perftools__Profiles__Label))
+        );
+        perftools__profiles__label__init(timestamp_label);
+        timestamp_label->key = prof.end_timestamp_ns_str_id();
+        timestamp_label->str = 0;
+        timestamp_label->num = prof.epoch_timestamp_ns(src_sample.timestamp_uptime_ns);
+        timestamp_label->num_unit = prof.nanoseconds_str_id();
+        sample->label[0] = timestamp_label;
```
maxep: suggestion/ The packer's responsibility is pprof aggregation only; the actual format is owned by the `profile` class. The packer now reaches into `prof` to call `end_timestamp_ns_str_id()`, `nanoseconds_str_id()`, and `epoch_timestamp_ns()` to build the timestamp label and `profile.h` had to expose those accessors solely to serve it.

Consider exposing `profile::labels_for(const sample_t&)` returning the full label set (including the computed timestamp label). The packer would iterate labels without knowing their semantics, with no extra allocation or pass.

DatadogInternal/Sources/Models/RUM/Vital.swift
```diff
@@ -26,6 +26,8 @@ public struct Vital: Encodable, Equatable {
     public let stepType: RUMVitalOperationStepEvent.Vital.StepType?
     /// Date when the vital was created
     public let date: Date
+    /// Interval between device and server time.
+    public var serverTimeOffset: TimeInterval
```
maxep:
```suggestion
    public let serverTimeOffset: TimeInterval
```
Nothing in the PR mutates it.

DatadogProfiling/Mach/profile_pprof_packer.cpp
simaoseica-dd: Good point on the separation of concerns.  👌 
Moved the timestamp label semantics into profile and made clear that no additional allocations are needed per sample.

maxep (approved):
LGTM 👍

PR #2871, opened 2026-04-28
Add StatsConcentrator for time-bucketed span aggregation

tienquocbui:
### What and why?

Implements the `StatsConcentrator`, the core aggregation engine for client-side stats. This is the component that receives span snapshots from the `onSpanFinished` callback and groups them into time-bucketed stats (hit counts, error counts, duration totals) keyed by aggregation dimensions (service, operation, resource, HTTP status, span kind, etc.).

This PR also wires the concentrator into the existing `ClientStatsFeature` with periodic timer-based flushing and `Flushable` conformance for SDK teardown.

### How?

**New types** (`StatsConcentrator.swift`):
- `StatsConcentrator`: aggregates `SpanSnapshot`s into 10-second time buckets, matching the Go reference implementation. Uses `@ReadWriteLock` for thread safety. Supports eligibility filtering (top-level, measured, or qualifying span kind), `oldestTs` tracking to prevent late-arriving spans from creating already-flushed buckets, and periodic + force flush.
- `AggregationKey`: the set of dimensions by which spans are grouped (service, operation, resource, HTTP status, type, span kind, is_trace_root, peer tags hash, service source).
- `Trilean`: matches the protobuf enum for `is_trace_root` (notSet / true / false).
- `GroupedStats`, `StatsBucket`: internal mutable types for running counters within a bucket.
- `ExportedBucket`, `ExportedGroupedStats`: immutable `Encodable` + `Sendable` structs for serialization.
- `fnv64a()`: FNV-64a hash for peer tag aggregation keys, matching Go's `tagsFnvHash`.
- `stochasticRound()`: converts floating-point counts to UInt64 (effectively truncation since weight is always 1.0 for now).

**Wiring changes**:
- `ClientStatsFeature` now holds a `StatsConcentrator`, runs a 30-second flush timer, conforms to `Flushable`, and writes exported buckets to feature storage via `eventWriteContext`.
- `Trace.enableOrThrow()` wires `trace.tracer.onSpanFinished` to `stats.concentrator.add(snapshot)` and passes `configuration.dateProvider` to ensure the stats pipeline shares the same clock as the tracer.
- `SpanSnapshot` gains a `type` field (populated from `tags["span.type"]` or defaulting to `"custom"` in `DDSpan.createSnapshot()`).
- `SpanSnapshot.mockAny()` / `.mockWith()` helpers added to `TestUtilities`.

**Known limitations** (to be addressed in subsequent PRs):
- `okSummary` / `errorSummary` are `Data()` placeholders; DDSketch implementation is next.
- `StatsRequestBuilder` still concatenates raw event bytes with `application/json` content type. Multi-bucket uploads will produce invalid JSON. The real MessagePack encoder replaces this in a later PR.
- When `Trace.Configuration.service` is nil, `snapshot.service` resolves to `""` instead of falling back to `DatadogContext.service` (which is only available asynchronously via `eventWriteContext`). This means stats may be grouped under an empty service name for default-config users. A proper fix requires resolving the core context service at snapshot creation or aggregation time, tracked for a follow-up PR.

### Review checklist
- [x] Add Objective-C interface for public APIs

### DIFF
--- a/DatadogTrace/Sources/Feature/ClientStatsFeature.swift
+++ b/DatadogTrace/Sources/Feature/ClientStatsFeature.swift
@@ -19,12 +19,68 @@ internal final class ClientStatsFeature: DatadogRemoteFeature {
     let messageReceiver: FeatureMessageReceiver
     let performanceOverride: PerformancePresetOverride?
 
-    init(core: DatadogCoreProtocol, configuration: Trace.Configuration) {
+    let concentrator: StatsConcentrator
+    private let featureScope: FeatureScope
+    private let dateProvider: DateProvider
+    private var flushTimer: Timer?
+
+    /// Interval between periodic flushes (default: 30 seconds).
+    let flushInterval: TimeInterval
+
+    init(
+        core: DatadogCoreProtocol,
+        configuration: Trace.Configuration,
+        dateProvider: DateProvider,
+        flushInterval: TimeInterval = 30
+    ) {
         self.requestBuilder = StatsRequestBuilder(
             customIntakeURL: configuration.customEndpoint,
             telemetry: core.telemetry
         )
         self.messageReceiver = NOPFeatureMessageReceiver()
         self.performanceOverride = nil
+        self.dateProvider = dateProvider
+        self.flushInterval = flushInterval
+        self.featureScope = core.scope(for: ClientStatsFeature.self)
+
+        let now = dateProvider.now.timeIntervalSince1970.dd.toNanoseconds
+        self.concentrator = StatsConcentrator(now: now)
+
+        startFlushTimer()
+    }
+
+    deinit {
+        flushTimer?.invalidate()
+    }
+
+    /// Flushes completed buckets and writes them to the feature storage for upload.
+    func flushStats(force: Bool = false) {
+        let now = dateProvider.now.timeIntervalSince1970.dd.toNanoseconds
+        let exportedBuckets = concentrator.flush(now: now, force: force)
+
+        guard !exportedBuckets.isEmpty else {
+            return
+        }
+
+        featureScope.eventWriteContext { _, writer in
+            for bucket in exportedBuckets {
+                writer.write(value: bucket)
+            }
+        }
+    }
+
+    private func startFlushTimer() {
+        flushTimer = Timer.scheduledTimer(
+            withTimeInterval: flushInterval,
+            repeats: true
+        ) { [weak self] _ in
+            self?.flushStats()
+        }
+    }
+}
+
+extension ClientStatsFeature: Flushable {
+    func flush() {
+        flushStats(force: true)
     }
 }
--- a/DatadogTrace/Sources/Feature/ClientStatsFeature.swift
+++ b/DatadogTrace/Sources/Feature/ClientStatsFeature.swift
@@ -54,6 +54,10 @@ internal final class ClientStatsFeature: DatadogRemoteFeature {
     }
 
     /// Flushes completed buckets and writes them to the feature storage for upload.
+    ///
+    /// - Parameter force: When `true`, flushes all buckets regardless of age (used during
+    ///   SDK teardown via `Flushable`). When `false`, only buckets older than the buffer
+    ///   window are flushed (normal periodic flush).
     func flushStats(force: Bool = false) {
         let now = dateProvider.now.timeIntervalSince1970.dd.toNanoseconds
         let exportedBuckets = concentrator.flush(now: now, force: force)
--- a/DatadogTrace/Sources/Feature/StatsConcentrator.swift
+++ b/DatadogTrace/Sources/Feature/StatsConcentrator.swift
@@ -1,0 +1,349 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+import DatadogInternal
+
+// MARK: - Trilean
+
+/// Matches the protobuf `Trilean` enum used in `ClientGroupedStats.is_trace_root`.
+internal enum Trilean: Int, Encodable, Hashable, Sendable {
+    case notSet = 0
+    case `true` = 1
+    case `false` = 2
+}
+
+// MARK: - Aggregation Key
+
+/// The set of dimensions by which spans are grouped within a time bucket.
+/// Matches the Go reference `BucketsAggregationKey` in `aggregation.go`.
+internal struct AggregationKey: Hashable, Sendable {
+    let service: String
+    let operationName: String
+    let resource: String
+    let httpStatusCode: UInt32
+    let type: String
+    let spanKind: String
+    let isTraceRoot: Trilean
+    let synthetics: Bool
+    let peerTagsHash: UInt64
+    let serviceSource: String
+}
+
+// MARK: - Grouped Stats
+
+/// Running counters for a single aggregation key within a time bucket.
+internal final class GroupedStats {
+    var hits: Double = 0
+    var topLevelHits: Double = 0
+    var errors: Double = 0
+    var duration: Double = 0
+    /// Peer tags stored as `"key:value"` pairs for export, matching Go's `matchingPeerTags`.
+    let peerTags: [String]
+
+    init(peerTags: [String]) {
+        self.peerTags = peerTags
+    }
+}
+
+// MARK: - Stats Bucket
+
+/// A single time bucket holding grouped stats keyed by aggregation dimensions.
+internal final class StatsBucket {
+    let start: UInt64
+    let duration: UInt64
+    var groups: [AggregationKey: GroupedStats] = [:]
+
+    init(start: UInt64, duration: UInt64) {
+        self.start = start
+        self.duration = duration
+    }
+}
+
+// MARK: - Exportable Bucket
+
+/// A flushed bucket ready for serialization and upload.
+internal struct ExportedBucket: Encodable, Sendable {
+    let start: UInt64
+    let duration: UInt64
+    let stats: [ExportedGroupedStats]
+}
+
+/// A single grouped stats entry ready for serialization.
+internal struct ExportedGroupedStats: Encodable, Sendable {
+    let service: String
+    let name: String
+    let resource: String
+    let httpStatusCode: UInt32
+    let type: String
+    let spanKind: String
+    let isTraceRoot: Trilean
+    let synthetics: Bool
+    let hits: UInt64
+    let errors: UInt64
+    let duration: UInt64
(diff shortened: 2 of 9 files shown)


tienquocbui:
@codex review

tienquocbui:
@codex review

DatadogTrace/Sources/Feature/ClientStatsFeature.swift
```diff
@@ -19,12 +19,68 @@ internal final class ClientStatsFeature: DatadogRemoteFeature {
...
+        self.concentrator = StatsConcentrator(now: now)
+
+        startFlushTimer()
+    }
+
+    deinit {
+        flushTimer?.invalidate()
+    }
+
+    /// Flushes completed buckets and writes them to the feature storage for upload.
```
arroz, line 56: The documentation should mention what `force` does and when to use it.

DatadogTrace/Sources/Feature/StatsConcentrator.swift
```diff
@@ -0,0 +1,349 @@
...
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+import DatadogInternal
+
+// MARK: - Trilean
+
+/// Matches the protobuf `Trilean` enum used in `ClientGroupedStats.is_trace_root`.
+internal enum Trilean: Int, Encodable, Hashable, Sendable {
```
arroz, line 13: Do we need this type, for encoding maybe? Or can this be just an Optional Bool?

DatadogTrace/Sources/Feature/StatsConcentrator.swift
```diff
@@ -0,0 +1,349 @@
...
+            hash = hash &* prime
+        }
+    }
+    return hash
+}
+
+// MARK: - Stochastic Rounding
+
+/// Converts a floating-point count to UInt64 using stochastic rounding.
+/// When weight is always 1.0, this is effectively a truncation to integer.
```
arroz: What weight?

DatadogTrace/Sources/Feature/StatsConcentrator.swift
```diff
@@ -0,0 +1,349 @@
...
+/// Span kinds eligible for stats computation per the v1.2.0 spec.
+private let eligibleSpanKinds: Set<String> = ["server", "consumer", "client", "producer"]
+
+/// Span kinds for which peer tags should be included in the aggregation key.
+private let peerTagSpanKinds: Set<String> = ["client", "producer", "consumer"]
+
+// MARK: - FNV-64a Hash
+
+/// Computes an FNV-64a hash of sorted tag strings, matching Go's `tagsFnvHash`.
+internal func fnv64a(_ tags: [String]) -> UInt64 {
```
arroz: This function and the next should probably be in an utility entity.

DatadogTrace/Sources/Feature/StatsConcentrator.swift
```diff
@@ -0,0 +1,349 @@
...
+            )
+            let bucket = buckets[bucketKey, default: StatsBucket(start: bucketKey, duration: self.bucketDuration)]
+
+            let group = bucket.groups[aggregationKey, default: GroupedStats(peerTags: peerTagStrings)]
+
+            group.hits += 1
+            if snapshot.isError {
+                group.errors += 1
+            }
+            group.duration += Double(snapshot.duration)
```
arroz, line 225: Isn't group.duration a UInt64? Why converting to Double?

DatadogTrace/Sources/Feature/StatsConcentrator.swift
```diff
@@ -0,0 +1,349 @@
...
+    ///
+    /// - Parameter force: When `true`, flushes all buckets regardless of age.
+    ///   Used during SDK teardown.
+    /// - Parameter now: Current time in nanoseconds.
+    /// - Returns: Array of exported buckets ready for serialization.
+    func flush(now: Nanoseconds, force: Bool) -> [ExportedBucket] {
+        var flushed: [ExportedBucket] = []
+
+        _buckets.mutate { buckets in
+            let cutoff = Int64(now) - Int64(self.bufferLen) * Int64(self.bucketDuration)
```
arroz: Two comments:

1. Just a nitpick, but you can probably simplify a bit (arguably, I admit) if you do something like `let cutoff = force ? .max : Int64(now) - Int64(self.bufferLen) * Int64(self.bucketDuration)`. That way you're explicitly saying your cutoff date is the distant future and avoiding a check below. But up to you, depends on you think about it.
2. Why does the cutoff date depend on `bufferLen`? Not saying it's wrong, but `bufferLen` should be documented explaining what it is and how it impacts the cutoff.

DatadogTrace/Sources/Feature/StatsConcentrator.swift
```diff
@@ -0,0 +1,349 @@
...
+            }
+
+            for key in keysToRemove {
+                buckets.removeValue(forKey: key)
+            }
+
+            let aligned = Self.alignTimestamp(now, bucketDuration: self.bucketDuration)
+            let offset = UInt64(self.bufferLen - 1) * self.bucketDuration
+            let newOldestTs = aligned >= offset ? aligned - offset : 0
+            self._oldestTs.mutate { oldest in
```
arroz: ⚠️ Be very careful with this. Two locks inside each other is half way into deadlocking. If these two vars (oldestTs and buckets) need to be updated at the same time, they probably should be members of another struct (can be an inner one defined inside this one), and that one should be locked, using a single lock.

DatadogTrace/Sources/Feature/StatsConcentrator.swift
```diff
@@ -0,0 +1,349 @@
...
+        let isTraceRoot: Trilean = snapshot.parentSpanID == nil ? .true : .false
+        let peerTagStrings = peerTags.map { "\($0.key):\($0.value)" }
+
+        return AggregationKey(
+            service: snapshot.service,
+            operationName: snapshot.operationName,
+            resource: snapshot.resource,
+            httpStatusCode: snapshot.httpStatusCode,
+            type: snapshot.type,
+            spanKind: snapshot.spanKind ?? "",
```
arroz, line 350: Is "" valid?

DatadogTrace/Sources/Trace.swift
```diff
@@ -46,11 +46,15 @@ public enum Trace {
 
         // Register Client-Side Stats feature if enabled:
         if configuration.statsComputationEnabled {
-            let stats = ClientStatsFeature(core: core, configuration: configuration)
+            let stats = ClientStatsFeature(core: core, configuration: configuration, dateProvider: configuration.dateProvider)
             try core.register(feature: stats)
 
-            // [CSS] Wire StatsConcentrator here to aggregate snapshots into
-            // time-bucketed stats payloads before writing to the stats feature scope.
+            trace.tracer.onSpanFinished = { [weak core] snapshot in
```
arroz: Why getting a reference to `core` here if it's not used in the function body? OTOH `stats` should probably be a weak ref to avoid holding the feature.

DatadogTrace/Sources/Feature/StatsConcentrator.swift
```diff
@@ -0,0 +1,349 @@
...
+                group.errors += 1
+            }
+            group.duration += Double(snapshot.duration)
+            if snapshot.isTopLevel {
+                group.topLevelHits += 1
+            }
+
+            bucket.groups[aggregationKey] = group
+            buckets[bucketKey] = bucket
+        }
```
ncreated: I have performance concerns about the threading model used here. From what I can see, in the worst case this can block or introduce jitter on the application's main thread because both `add()` and `flush()` acquire an exclusive write lock on `buckets` (they share a critical section)

🐌 Imagine this scenario: `span.finish()` is called on the main thread, which executes `add()` on the main thread too. If `flush()` is already holding the lock on a background thread at that moment, **the main thread will block until `flush()` leaves its critical section**.

🐌🐌 Or another scenario: `span.finish()` is called simultaneously from multiple threads, including main. In the worst case, the main thread will be blocked until the last competing thread leaves the critical section.

To understand the impact, here's O-complexity estimate by Claude (I believe it's correct):

> - `flush()` is **O(B × G)**: it iterates over every active bucket and every aggregation group within each bucket to build `ExportedGroupedStats` values. B is the number of active time buckets; G is the number of distinct `(service, operation, resource, status, spanKind, ...)` combinations seen in that window. G is unbounded and grows with application diversity — an app with many endpoints can produce a large G.
> - `add()` is **O(1)**: a couple of dictionary lookups and counter increments.


Can we do better? Given that `add()` is fire-and-forget (no return value) and `flush()` is driven by a periodic timer rather than a latency-sensitive caller, a serial `DispatchQueue` would IMO be a better fit than a lock. `add()` would dispatch asynchronously (returning immediately to the caller), while `flush()` could use `queue.sync` to drain any in-flight adds before collecting results.

This would eliminate lock contention entirely: the user's thread is never blocked, and `buckets`/`oldestTs` could be plain vars without any `@ReadWriteLock` overhead. It would also simplify the threading model, which — as noted by @arroz — is prone to deadlocks if used carelessly.

DatadogTrace/Sources/Feature/StatsConcentrator.swift
tienquocbui: The protobuf schema defines `is_trace_root` as a `Trilean` enum with explicit wire values: `NOT_SET=0`, `TRUE=1`, `FALSE=2`. We need the raw integer mapping for correct MessagePack encoding in a later PR. An `Optional<Bool>` would work semantically, but we'd need a custom encoder to produce the right values. Keeping the enum makes the wire format explicit.

DatadogTrace/Sources/Feature/StatsConcentrator.swift
tienquocbui: The Go reference accumulates all counters (hits, errors, duration, topLevelHits) as floats because weighted sampling multiplies each span's contribution by a float weight. We always use weight 1.0 today, so UInt64 would work for now. I kept Double to match the Go reference and to avoid a refactor when weighted sampling is added later. Happy to change if you prefer UInt64.

DatadogTrace/Sources/Feature/StatsConcentrator.swift
tienquocbui: Yes, empty string is intentional for unset span kind. The Go reference uses empty string for unset values in aggregation keys rather than optionals. This keeps `AggregationKey` `Hashable` without needing optional handling in the hash.

tienquocbui:
@codex review

DatadogTrace/Sources/Feature/StatsConcentrator.swift
arroz: Nevermind, I thought "group" here was a `ExportedGroupedStats` instead of `GroupedStats`.

arroz (approved):
Looks good! :shipit:

PR #2878, opened 2026-05-05
RUM-15735: Offload Mach profile aggregation

simaoseica-dd:
### What and why?

This PR improves the iOS Mach profiler so aggregation work no longer blocks the sampling loop. It also reduces aggregation cost and adds bounded overload handling so normal profiling can keep sampling while heavy profile processing catches up.

### How?

Mach profiling now separates the sampling hot path from the aggregation cold path. The sampling thread only captures raw stack samples and hands off completed buffers, avoiding expensive profile mutation, binary image resolution and flush/profile-rotation work on the sampling loop.

Aggregation runs on a long-lived serialized `aggregation_worker`. The worker drains sample batches in order, executes flush requests as ordered barriers and performs profile swaps on the same serialized stream so `flush_and_get_profile()` produces a deterministic profile boundary while sampling continues.

Additionally, the `aggregation_worker` also:
- resolves binary image data lazily for first seen instruction pointers instead of resolving every frame in every batch.
- excludes profiler owned threads from collected profiles.
- uses reusable buffers plus temporary overflow buffers so the sampler does not pause when aggregation falls behind.
- applies a hard byte ceiling to queued aggregation work; if the ceiling is reached, new batches are dropped and internal diagnostics counters are recorded.
- exposes dropped batch diagnostics through a C API. This diagnostics are collected in the `Profiling Session` telemetry ([internal spec](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/edit-v2/5892964530)).

### Review checklist

### DIFF
--- a/DatadogProfiling/Mach/aggregation_worker.cpp
+++ b/DatadogProfiling/Mach/aggregation_worker.cpp
@@ -1,0 +1,385 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+#include "aggregation_worker.h"
+
+#if defined(__APPLE__) && !TARGET_OS_WATCH
+
+#include <algorithm>
+#include <cstdio>
+#include <cstdlib>
+#include <new>
+
+namespace dd::profiler {
+
+static void destroy_stack_trace_payload(stack_trace_t* trace) {
+    if (!trace) return;
+
+    if (trace->thread_name) {
+        std::free((void*)trace->thread_name);
+        trace->thread_name = nullptr;
+    }
+
+    if (trace->frames) {
+        std::free(trace->frames);
+        trace->frames = nullptr;
+    }
+}
+
+aggregation_worker::aggregation_worker(
+    size_t buffer_capacity,
+    uint32_t max_stack_depth,
+    stack_trace_callback_t callback,
+    void* ctx,
+    uint64_t hard_limit_bytes,
+    qos_class_t worker_qos)
+    : buffer_capacity(buffer_capacity)
+    , max_stack_depth(max_stack_depth)
+    , worker_qos(worker_qos)
+    , callback(callback)
+    , ctx(ctx)
+    , hard_limit_bytes(hard_limit_bytes) {
+}
+
+aggregation_worker::~aggregation_worker() {
+    stop();
+}
+
+void* aggregation_worker::worker_thread_entry(void* arg) {
+    pthread_setname_np("com.datadoghq.profiler.aggregate");
+    static_cast<aggregation_worker*>(arg)->worker_main();
+    return nullptr;
+}
+
+bool aggregation_worker::start() {
+    std::lock_guard<std::mutex> lock(work_mutex);
+
+    if (worker_thread_started) {
+        return false;
+    }
+
+    clear_pending_work_locked();
+    reusable_buffers.clear();
+    next_flush_id = 0;
+    completed_flush_id = 0;
+    pending_bytes = 0;
+    dropped_batch_count = 0;
+    dropped_sample_count = 0;
+    max_pending_bytes = 0;
+    producer_finished = false;
+    worker_finished = false;
+    worker_mach_thread.store(MACH_PORT_NULL, std::memory_order_relaxed);
+
+    pthread_attr_t worker_attr;
+    pthread_attr_init(&worker_attr);
+    pthread_attr_set_qos_class_np(&worker_attr, worker_qos, 0);
+
+    int result = pthread_create(&worker_thread, &worker_attr, worker_thread_entry, this);
+    pthread_attr_destroy(&worker_attr);
+    if (result != 0) {
+        producer_finished = true;
+        worker_finished = true;
+        flush_cv.notify_all();
+        return false;
+    }
+
+    worker_mach_thread.store(pthread_mach_thread_np(worker_thread), std::memory_order_relaxed);
+    worker_thread_started = true;
+    return true;
+}
+
+void aggregation_worker::stop() {
+    if (is_worker_thread()) {
+        std::lock_guard<std::mutex> lock(work_mutex);
+        producer_finished = true;
+        work_cv.notify_all();
+        flush_cv.notify_all();
+        return;
+    }
+
+    {
+        std::lock_guard<std::mutex> lock(work_mutex);
+        if (!worker_thread_started && worker_finished) {
+            clear_pending_work_locked();
+            reusable_buffers.clear();
+            worker_mach_thread.store(MACH_PORT_NULL, std::memory_order_relaxed);
+            return;
+        }
+
+        producer_finished = true;
+        work_cv.notify_all();
+        flush_cv.notify_all();
+    }
+
+    if (worker_thread_started) {
+        pthread_join(worker_thread, nullptr);
+    }
+
+    {
+        std::lock_guard<std::mutex> lock(work_mutex);
+        worker_thread_started = false;
+        worker_mach_thread.store(MACH_PORT_NULL, std::memory_order_relaxed);
+        clear_pending_work_locked();
+        reusable_buffers.clear();
+        worker_finished = true;
+        producer_finished = true;
+    }
+}
+
+void aggregation_worker::request_flush(flush_action_t action, void* action_ctx) {
+    std::unique_lock<std::mutex> lock(work_mutex);
+    bool execute_inline = false;
+
+    const uint64_t flush_id = ++next_flush_id;
+    if (worker_finished) {
+        completed_flush_id = flush_id;
+        execute_inline = true;
+    } else {
+        work_item barrier{
+            work_item::kind::flush_barrier,
+            {},
+            flush_id,
+            action,
+            action_ctx
+        };
+
+        if (producer_finished) {
+            pending_work.push_back(std::move(barrier));
+            work_cv.notify_one();
+        } else {
+            requested_flushes.push_back(std::move(barrier));
+        }
+
+        flush_cv.wait(lock, [this, flush_id] {
+            return completed_flush_id >= flush_id || worker_finished;
+        });
+
+        if (completed_flush_id < flush_id && worker_finished) {
+            completed_flush_id = flush_id;
+            execute_inline = true;
+        }
+    }
+
+    lock.unlock();
+    if (execute_inline && action) {
+        action(action_ctx);
+    }
+}
+
+void aggregation_worker::enqueue_active_buffer(std::vector<stack_trace_t>& active_buffer) {
+    if (active_buffer.empty()) {
+        return;
+    }
+
+    std::vector<stack_trace_t> dropped_batch;
+    bool did_drop = false;
+
+    {
+        std::lock_guard<std::mutex> lock(work_mutex);
+        std::vector<stack_trace_t> batch;
+        batch.swap(active_buffer);
+
+        if (!reusable_buffers.empty()) {
+            active_buffer = std::move(reusable_buffers.back());
+            reusable_buffers.pop_back();
+        }
+
+        const uint64_t batch_bytes = batch_footprint_bytes(batch);
+        if (pending_bytes + batch_bytes > hard_limit_bytes) {
+            dropped_batch_count += 1;
+            dropped_sample_count += batch.size();
+            max_pending_bytes = std::max(max_pending_bytes, pending_bytes + batch_bytes);
+            dropped_batch = std::move(batch);
+            did_drop = true;
+        } else {
+            pending_bytes += batch_bytes;
+            max_pending_bytes = std::max(max_pending_bytes, pending_bytes);
+            pending_work.push_back({
+                work_item::kind::batch,
+                std::move(batch),
+                0,
+                nullptr,
+                nullptr,
+                batch_bytes
+            });
+        }
+    }
+
+    if (did_drop) {
+        destroy_batch(dropped_batch);
+    }
+
+    if (!did_drop) {
+        work_cv.notify_one();
+    }
+
+    if (active_buffer.capacity() == 0) {
+        try {
+            // Best effort: overflow handoff keeps sampling moving even if we
+            // cannot immediately reserve the next full-capacity buffer.
+            active_buffer.reserve(buffer_capacity);
+        } catch (const std::bad_alloc&) {
+            return;
+        }
+    }
+}
+
+void aggregation_worker::service_pending_flush_request(std::vector<stack_trace_t>& active_buffer) {
+    {
+        std::lock_guard<std::mutex> lock(work_mutex);
+        if (requested_flushes.empty()) {
+            return;
+        }
+    }
+
+    enqueue_active_buffer(active_buffer);
+
+    {
+        std::lock_guard<std::mutex> lock(work_mutex);
+        while (!requested_flushes.empty()) {
+            pending_work.push_back(std::move(requested_flushes.front()));
+            requested_flushes.pop_front();
+        }
+    }
+
+    work_cv.notify_one();
+}
+
+void aggregation_worker::finish_producer(std::vector<stack_trace_t>& active_buffer) {
+    service_pending_flush_request(active_buffer);
+    enqueue_active_buffer(active_buffer);
+
+    {
+        std::lock_guard<std::mutex> lock(work_mutex);
+        while (!requested_flushes.empty()) {
+            pending_work.push_back(std::move(requested_flushes.front()));
+            requested_flushes.pop_front();
+        }
+
+        producer_finished = true;
+    }
+
+    work_cv.notify_all();
+    flush_cv.notify_all();
+}
+
+bool aggregation_worker::is_worker_thread() {
+    std::lock_guard<std::mutex> lock(work_mutex);
+    return worker_thread_started && pthread_equal(pthread_self(), worker_thread);
+}
+
+bool aggregation_worker::is_worker_thread(thread_t thread) {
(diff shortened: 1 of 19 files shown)


DatadogProfiling/Mach/include/mach_sampling_profiler.h
```diff
@@ -60,7 +64,8 @@ class mach_sampling_profiler {
     mach_sampling_profiler(
         const sampling_config_t* config,
         stack_trace_callback_t callback,
-        void* ctx);
+        void* ctx,
+        uint64_t hard_limit_bytes);
```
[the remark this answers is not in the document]
simaoseica-dd: `mach_sampling_profiler` is not part of the SDK's supported public API.

DatadogProfiling/Mach/include/dd_profiler_testing.h
```diff
@@ -49,6 +49,9 @@ void dd_delete_profiling_defaults(void);
  *                   - Default: 5000000000ULL (5 seconds)
  *                   - Timeout checking occurs during sample processing
  *
```
arroz: Nitpick: empty line between these two parameters but not the others.

DatadogProfiling/Mach/dd_profiler.cpp
```diff
@@ -15,14 +15,21 @@
...
 #include <cstring>
-#include <random>
 #include <mutex>
+#include <new>
+#include <random>
 
 // Profiling sampling backstop (see `callback`).
 // Typical profile span is ~1 minute; this cutoff includes additional slack beyond that.
 // The extra time avoids stopping sampling while the profile is still being processed.
 static constexpr int64_t DD_PROFILER_TIMEOUT_NS = 90000000000ULL; // 1:30 minutes
```
arroz: There is a mismatch between the value and the variable type (signed vs unsigned). Although the value fits within the signed bits, both types should match.

DatadogProfiling/Mach/include/aggregation_worker.h
```diff
@@ -0,0 +1,159 @@
...
+    std::atomic<thread_t> worker_mach_thread{MACH_PORT_NULL};
+
+    /// Ordered stream consumed by the worker thread.
+    std::deque<work_item> pending_work;
+    /// Flush barriers requested by external callers and awaiting a producer safe point.
+    std::deque<work_item> requested_flushes;
+    /// Recycled buffers retained for future producer handoff.
+    std::vector<std::vector<stack_trace_t>> reusable_buffers;
+
+    std::mutex work_mutex;
```
arroz, line 157: This mutex could have a bit of documentation regarding which variables it protects.

arroz (commented):
Overall it seems OK! I took some time to validate locking as I'm not a C++ expert, I don't see any obvious deadlocks or anything like that.

Just left some minor comments.

DatadogProfiling/Mach/include/aggregation_worker.h
```diff
@@ -0,0 +1,159 @@
...
+     * @brief Returns and resets diagnostics accumulated since the last consume.
+     */
+    void consume_diagnostics(dd_profiler_diagnostics_t* out);
+
+private:
+    struct work_item {
+        enum class kind {
+            batch,
+            flush_barrier
+        };
```
fuzzybinary: If we're able to use C++17 (which I think we are) this would be better served as an `std::variant` which would also likely save a bunch of memory.

DatadogProfiling/Mach/include/aggregation_worker.h
```diff
@@ -0,0 +1,159 @@
...
+    /// Wakes the aggregation thread when new batches or flush barriers are queued.
+    std::condition_variable work_cv;
+    /// Wakes flush callers when their requested flush barrier has been completed.
+    std::condition_variable flush_cv;
+    uint64_t next_flush_id = 0;
+    uint64_t completed_flush_id = 0;
+    uint64_t pending_bytes = 0;
+    uint64_t dropped_batch_count = 0;
+    uint64_t dropped_sample_count = 0;
+    uint64_t max_pending_bytes = 0;
```
fuzzybinary: We already have a structure that stores these together - maybe just make these a `dd_profiler_diagnositcs` object?

DatadogProfiling/Mach/aggregation_worker.cpp
```diff
@@ -0,0 +1,385 @@
...
+#if defined(__APPLE__) && !TARGET_OS_WATCH
+
+#include <algorithm>
+#include <cstdio>
+#include <cstdlib>
+#include <new>
+
+namespace dd::profiler {
+
+static void destroy_stack_trace_payload(stack_trace_t* trace) {
```
fuzzybinary: nit: This feels like a weird place for this. Shouldn't this live somewhere with the declaration of `stack_trace_t`?

DatadogProfiling/Mach/aggregation_worker.cpp
```diff
@@ -0,0 +1,385 @@
...
+        return false;
+    }
+
+    worker_mach_thread.store(pthread_mach_thread_np(worker_thread), std::memory_order_relaxed);
+    worker_thread_started = true;
+    return true;
+}
+
+void aggregation_worker::stop() {
+    if (is_worker_thread()) {
```
fuzzybinary: Does it really make sense to allow the worker thread to call `stop`?

Even if it does, I think this whole function could use some documentation, explaining what it's doing under what circumstances and why..

DatadogProfiling/Tests/DDProfilerTests.swift
awforsythe: This is a pre-existing test that wasn't modified in these changes, but it seems like it may be relevant, since both `dd_profiler_flush_and_get_profile` and `dd_profiler_stop` now require coordination with the aggregation worker.

The test name ("stop unblocks pending flush request") seems to suggest that when you call `dd_profiler_stop`, you expect any pending call to `dd_profiler_flush_and_get_profile` to be _unblocked_, as in canceled immediately... but the test doesn't seem to validate that behavior. A flush call now waits on `flush_cv` while holding `g_dd_profiler_mutex`, meaning that a subsequent stop call will block (waiting for `g_dd_profiler_mutex`) until the flush completes.

i.e. from what I understand, if Thread A calls `dd_profiler_flush_and_get_profile()` and that call takes 100ms, a call to `dd_profiler_stop()` in Thread B will block until that 100ms flush operation is finished. Is this the expected behavior?

(If so, the name of the test seems a bit misleading, as the call to stop has no effect on whether the pending flush is unblocked: the test is just asserting that when you call both functions concurrently, they both return and you end up in the stopped state.)

DatadogProfiling/Mach/aggregation_worker.cpp
```diff
@@ -0,0 +1,385 @@
...
+        if (completed_flush_id < flush_id && worker_finished) {
+            completed_flush_id = flush_id;
+            execute_inline = true;
+        }
+    }
+
+    lock.unlock();
+    if (execute_inline && action) {
+        action(action_ctx);
+    }
```
fuzzybinary: I'm curious if this `execute_inline` special case can be avoided somehow. It would be nice if whatever is processing the work_items could reliably execute the flush action.

DatadogProfiling/Mach/aggregation_worker.cpp
```diff
@@ -0,0 +1,385 @@
...
+
+    if (!did_drop) {
+        work_cv.notify_one();
+    }
+
+    if (active_buffer.capacity() == 0) {
+        try {
+            // Best effort: overflow handoff keeps sampling moving even if we
+            // cannot immediately reserve the next full-capacity buffer.
+            active_buffer.reserve(buffer_capacity);
```
fuzzybinary: I feel like `buffer_capacity` is misnamed here, as it's more an initial size for a given buffer not a limit.

DatadogProfiling/Mach/aggregation_worker.cpp
```diff
@@ -0,0 +1,385 @@
...
+        }
+    }
+
+    lock.unlock();
+    if (execute_inline && action) {
+        action(action_ctx);
+    }
+}
+
+void aggregation_worker::enqueue_active_buffer(std::vector<stack_trace_t>& active_buffer) {
```
fuzzybinary, line 133: I had to stare at this function for quite a while to figure out what it was doing. Comments here would be greatly appreciated.

awforsythe (approved):
The level of complexity required here for proper synchronization is pretty impressive. I've done my best to understand the mental model for everything that's happening here re: concurrency, but I think this is the sort of code where you can't go wrong with more inline comments 🙂

As far as C++ code quality goes, LGTM.

DatadogProfiling/Mach/aggregation_worker.cpp
```diff
@@ -0,0 +1,385 @@
...
+    batch.clear();
+
+    std::lock_guard<std::mutex> lock(work_mutex);
+    pending_bytes = pending_bytes > footprint_bytes ? pending_bytes - footprint_bytes : 0;
+    if (reusable_buffers.size() < max_reusable_buffers) {
+        reusable_buffers.push_back(std::move(batch));
+    }
+}
+
+void aggregation_worker::destroy_batch(std::vector<stack_trace_t>& batch) {
```
fuzzybinary: nit: naming - this isn't destroying the batch, it's clearing the data in the individual items. It looks really weird to see `destroy_batch` immediately followed by `recycle_batch`.

DatadogProfiling/Mach/aggregation_worker.cpp
```diff
@@ -0,0 +1,385 @@
...
+}
+
+void aggregation_worker::recycle_batch(std::vector<stack_trace_t>&& batch, uint64_t footprint_bytes) {
+    batch.clear();
+
+    std::lock_guard<std::mutex> lock(work_mutex);
+    pending_bytes = pending_bytes > footprint_bytes ? pending_bytes - footprint_bytes : 0;
+    if (reusable_buffers.size() < max_reusable_buffers) {
+        reusable_buffers.push_back(std::move(batch));
+    }
```
fuzzybinary: Did we do any profiling / checks on how much time we save performing recycling on these vectors? It seems we only allocate once during the buffer swap, which, because these are all C structs, should be a single allocation without any constructor calls. I feel like the added complexity of recycling the buffers might not be worth it, especially as the number of buffers available isn't a hard limit. You'll still get a new buffer even if we're at the max number of reusable buffers.

DatadogProfiling/Mach/dd_profiler.cpp
```diff
@@ -247,22 +257,78 @@ class dd_profiler {
...
-        profile->set_server_time_offset_ns(server_time_offset_ns);
-        return harvested;
+        return swap_context.flushed_profile;
+    }
+
+    dd_profiler_diagnostics_t diagnostics() {
+        dd_profiler_diagnostics_t result = empty_diagnostics();
+
+        if (profiler) {
+            profiler->consume_diagnostics(&result);
```
fuzzybinary: Since this is all C++, any reason this couldn't be pass by reference instead of taking the pointer?

DatadogProfiling/Mach/dd_profiler.cpp
```diff
@@ -247,22 +257,78 @@ class dd_profiler {
...
     profile* flush_and_get_profile() {
-        if (profiler) profiler->request_flush();
+        profile_swap_context swap_context{
+            this,
+            new (std::nothrow) dd::profiler::profile(sampling_interval_ns),
+            nullptr
+        };
+
+        if (profiler) {
+            profiler->request_flush(swap_profile_action, &swap_context);
```
fuzzybinary: Does `request_flush` ever take an action that isn't `swap_profile_aciton`? If not, I feel like we might be able to add some type safety here by letting this take an `std::function` over a C-style funciton pointer.

fuzzybinary (commented):
This is some good work! A few comments that I think should be addressed, but I won't hold up merge.

DatadogProfiling/Mach/include/aggregation_worker.h
```diff
@@ -0,0 +1,159 @@
...
+
+/**
+ * @brief Serialized worker that drains sampled stack-trace batches in-order.
+ *
+ * The aggregation worker owns the aggregation thread, flush barriers, and a
+ * small reusable buffer pool used to decouple sample capture from heavy
+ * callback work. Under bursty load it can temporarily accept additional
+ * overflow buffers, then shrink back to the reusable pool as the worker
+ * catches up.
+ */
```
maxep, line 38: suggest/ The two-queue flush design is non-obvious and worth documenting here.

A `request_flush` call enqueues the barrier into `requested_flushes`, not directly into `pending_work`. It only moves into the ordered stream once the sampling thread reaches a safe point via `service_pending_flush_request()` (or `finish_producer()` on shutdown). This ensures the barrier lands *after* the current in-progress batch, so the flush action (e.g. `swap_profile_action`) sees a complete sample boundary.

Suggest adding a `## Flush ordering` paragraph to this class doc explaining this.

DatadogProfiling/Mach/include/dd_profiler.h
```diff
@@ -181,6 +181,18 @@ typedef enum {
     DD_PROFILER_STATUS_ALREADY_STARTED = 7,   ///< Failed to start profiler because it is already started
```
maxep: `DD_PROFILER_STATUS_ALREADY_STARTED` is now dead code. The PR changed the status assigned when `start_sampling()` returns false from `ALREADY_STARTED` to `NOT_STARTED`, so this value can no longer be returned.

DatadogProfiling/Mach/dd_profiler.cpp
```diff
@@ -209,17 +217,19 @@ class dd_profiler {
...
         }
 
+        status = DD_PROFILER_STATUS_RUNNING;
         return 1;
     }
 
     void stop() {
         if (!profiler) return;
         status = DD_PROFILER_STATUS_STOPPED;
+        profiler->request_stop();
```
maxep: cr/ `stop_sampling()` already calls `request_stop()` unconditionally at the top of its lock-guarded path. Either drop this line, or add a comment explaining the intent (e.g. "signal the loop early so it can exit before we block on the join").

maxep (commented):
I would suggest more documentation on this because it is difficult to grasp. 

The PR adds a meaningful layer of concurrency with the`aggregation_worker` but the overall architecture now has three threads, two queues, and a global singleton wrapped by an instance-based Swift feature. None of which is documented in RFC or Architecture doc. A reviewer/contributor has to reconstruct the full picture by reading the code.

A short architecture note at the top of `dd_profiler.cpp` (or in the module `README`) describing the three-thread model, the singleton ownership, and how the Swift feature layer relates to the C++ singleton would go a long way.

DatadogProfiling/Mach/dd_profiler.cpp
simaoseica-dd: It is not needed. ty

DatadogProfiling/Mach/aggregation_worker.cpp
simaoseica-dd: Renamed to `free_batch_payloads`

DatadogProfiling/Mach/aggregation_worker.cpp
simaoseica-dd: Good point. This logic can be simplified. It was part of an initial version where the aggregation would avoid dropping samples and allow the temporary queued memory to grow without limits. But there are no infinite resources.
We can simplify by removing the `reusable_buffers` 👍

DatadogProfiling/Mach/dd_profiler.cpp
simaoseica-dd: No strong reason. Changed

DatadogProfiling/Mach/dd_profiler.cpp
simaoseica-dd: Yes, it is better. ty

DatadogProfiling/Mach/aggregation_worker.cpp
simaoseica-dd: Also, it was duplicated. Removed

DatadogProfiling/Mach/include/aggregation_worker.h
simaoseica-dd: Sounds good. I also grouped the constants that are only config params

DatadogProfiling/Mach/include/aggregation_worker.h
simaoseica-dd: Yes, much better

DatadogProfiling/Mach/aggregation_worker.cpp
simaoseica-dd: You're right. We don't need to be so defensive against self-join deadlock, because it should not happen

DatadogProfiling/Mach/aggregation_worker.cpp
simaoseica-dd: Changed the `request_flush` to return `bool` where `true` means the flush was processed and false means the worker didn't processed it. 
Then `dd_profiler` performs the direct profile swap fallback only if false is returned.

DatadogProfiling/Tests/DDProfilerTests.swift
simaoseica-dd: Yes, that is the expected behavior with the current API. The Thread B will block until Thread A releases the mutex.
Good catch, changed the name and expectation.

simaoseica-dd (commented):
@fuzzybinary great catches and suggestions. The code is in much better shape now, ty.
@maxep thanks for the review. I also created a [ticket](https://datadoghq.atlassian.net/browse/RUM-16423) to focus on updating the documentation.
@arroz @awforsythe thanks for the careful read. 
All addressed.

maxep (approved):

PR #2882, opened 2026-05-05
Generate timeseries memory and CPU RUM event models

barboraplasovska:
### What and why?

Regenerated `RUMDataModels.swift` and `RUMDataModels+objc.swift` from the `bplasovska/timeseries` branch of `rum-events-format`, adding `RUMTimeseriesMemoryEvent` and `RUMTimeseriesCpuEvent` to the SDK.

### Review checklist
- [ ] Feature or bugfix MUST have appropriate tests (unit, integration)
- [ ] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes
- [ ] Add Objective-C interface for public APIs - see our [guidelines](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3157787243/RFC+-+Modular+Objective-C+Interface#Recommended-solution) (internal)
- [ ] Run `make api-surface` when adding new APIs

PR #2881, opened 2026-05-05
Fix ObjC bridge generator for nested DataPoint types

barboraplasovska:
### What and why?

The ObjC interop code generator was walking all the way up to the top-level event class when resolving the root class for nested structs inside array items. This caused `DataPoint` wrapper classes to be typed as `objc_RUMTimeseriesCpuEvent` instead of `objc_RUMTimeseriesCpuEventTimeseriesData`, breaking the generated bridge.

### Review checklist
- [ ] Feature or bugfix MUST have appropriate tests (unit, integration)
- [ ] Make sure each commit and the PR mention the Issue number or JIRA reference
- [ ] Add CHANGELOG entry for user facing changes
- [ ] Add Objective-C interface for public APIs - see our [guidelines](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3157787243/RFC+-+Modular+Objective-C+Interface#Recommended-solution) (internal)
- [ ] Run `make api-surface` when adding new APIs

PR #2883, opened 2026-05-05
Add delta encoder for timeseries compression

barboraplasovska:
### What and why?

Adds `DeltaEncoder`, a pure value-type encoder converting `(timestamp, value)` samples into a columnar delta format: absolute value at index 0, deltas thereafter, scaled to fixed-point integers at `precision = 4`. Gives 2.4–3.3× uncompressed size reduction.

### Review checklist


barboraplasovska:
* **#2991** <a href="https://app.graphite.com/github/pr/DataDog/dd-sdk-ios/2991?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2885** <a href="https://app.graphite.com/github/pr/DataDog/dd-sdk-ios/2885?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2884** <a href="https://app.graphite.com/github/pr/DataDog/dd-sdk-ios/2884?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2883** <a href="https://app.graphite.com/github/pr/DataDog/dd-sdk-ios/2883?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/dd-sdk-ios/2883?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2882** <a href="https://app.graphite.com/github/pr/DataDog/dd-sdk-ios/2882?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2881** <a href="https://app.graphite.com/github/pr/DataDog/dd-sdk-ios/2881?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `feature/timeseries`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

DatadogRUM/Sources/Timeseries/DeltaEncoder.swift
```diff
@@ -0,0 +1,109 @@
...
+
+        var ts: [Int64] = []
+        var memoryMax: [Int64] = []
+        var memoryPercent: [Int64] = []
+
+        for (index, sample) in batch.enumerated() {
+            if index == 0 {
+                ts.append(sample.timestamp)
+                memoryMax.append(Int64(round(sample.dataPoint.memoryMax * Double(scale))))
+                memoryPercent.append(Int64(round(sample.dataPoint.memoryPercent * Double(scale))))
```
maxep: Arithmetics and conversation can overflow, lets use the `multipliedReportingOverflow` or the `ddWithNoOverflow` extension method to make it safer.

Datadog/Datadog.xcodeproj/project.pbxproj
```diff
@@ -8551,6 +8665,229 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
+		D23F8E5129DDCD28001CFAE8 /* Sources */ = {
```
maxep: I don't think this block should be added, looks like a conflict artifact.

DatadogRUM/Tests/Timeseries/DeltaEncoderTests.swift
```diff
@@ -0,0 +1,90 @@
...
+    func testEncodeMemory_correctDeltaEncoding() {
+        // Given
+        let samples: [RUMTimeseriesMemoryEvent.Timeseries.Data] = [
+            .init(dataPoint: .init(memoryMax: 100.0, memoryPercent: 10.0), timestamp: 1_000_000_000),
+            .init(dataPoint: .init(memoryMax: 200.5, memoryPercent: 20.0), timestamp: 2_000_000_000),
+            .init(dataPoint: .init(memoryMax: 200.5, memoryPercent: 20.5), timestamp: 3_000_000_000)
+        ]
+
+        // When
+        let result = try! XCTUnwrap(DeltaEncoder.encodeMemory(samples))
```
maxep: Lets the test fail with `testEncodeMemory_correctDeltaEncoding() throws` instead of force unwrapping, which will crash in case of an exception.

maxep (changes requested):
Looks great, good test coverage 👍  Lets make it safer by managing overflowing and use error escalation in tests :)

Datadog/Datadog.xcodeproj/project.pbxproj
```diff
@@ -923,6 +927,93 @@
 		D233E8002E4B588B00E7CFDE /* dd_pprof.h in Headers */ = {isa = PBXBuildFile; fileRef = D233E7FE2E4B588B00E7CFDE /* dd_pprof.h */; settings = {ATTRIBUTES = (Private, ); }; };
 		D233E8022E4B653F00E7CFDE /* dd_pprof.cpp in Sources */ = {isa = PBXBuildFile; fileRef = D233E8012E4B653F00E7CFDE /* dd_pprof.cpp */; };
 		D234613228B7713000055D4C /* FeatureContextTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D234613028B7712F00055D4C /* FeatureContextTests.swift */; };
+		D23F8E5229DDCD28001CFAE8 /* UIViewControllerHandler.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61F3CDA2251118FB00C816E5 /* UIViewControllerHandler.swift */; };
```
arroz: Are all these new entries expected?

DatadogRUM/Sources/Timeseries/DeltaEncoder.swift
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        var ts: [Int64] = []
+        var memoryMax: [Int64] = []
+        var memoryPercent: [Int64] = []
+
+        for (index, sample) in batch.enumerated() {
+            if index == 0 {
+                ts.append(sample.timestamp)
+                memoryMax.append(Int64.ddWithNoOverflow(sample.dataPoint.memoryMax * Double(scale)))
```
arroz: Since `scale` is a constant, this is likely optimized out by the compiler, but just in case it's not, the conversion to `Double(scale)` is being done repeatedly. Since `scale` is never used in the `Int64` type, changing the `scale` type to `Double` would avoid all these conversions.

DatadogRUM/Sources/Timeseries/DeltaEncoder.swift
```diff
@@ -0,0 +1,109 @@
...
+    /// Output format:
+    /// ```
+    /// {
+    ///   "precision": 4,
+    ///   "ts": [absoluteNs, delta1, delta2, ...],
+    ///   "memory_max": [scaledInt64, delta1, delta2, ...],
+    ///   "memory_percent": [scaledInt64, delta1, ...]
+    /// }
+    /// ```
+    static func encodeMemory(_ batch: [RUMTimeseriesMemoryEvent.Timeseries.Data]) -> [String: Any]? {
```
arroz, line 31: The logic is repeated in both methods. Can it be extracted to a method that takes a collection and returns the final array? Performance wise it's a bit tricky because creating a temporary array (for example `batch.map { $0.dataPoint.memoryMax }` to pass to an utility function is costly, but regardless of how it's done, it seems it can be simplified.

DatadogRUM/Sources/Timeseries/DeltaEncoder.swift
```diff
@@ -0,0 +1,109 @@
...
+        var memoryPercent: [Int64] = []
+
+        for (index, sample) in batch.enumerated() {
+            if index == 0 {
+                ts.append(sample.timestamp)
+                memoryMax.append(Int64.ddWithNoOverflow(sample.dataPoint.memoryMax * Double(scale)))
+                memoryPercent.append(Int64.ddWithNoOverflow(sample.dataPoint.memoryPercent * Double(scale)))
+            } else {
+                let prev = batch[index - 1]
+                ts.append(sample.timestamp - prev.timestamp)
```
gonzalezreal: This still has a crash path. The conversion is clamped, but the subtraction can still overflow and trap in Swift.

Could we use `subtractingReportingOverflow` for the timestamp and value deltas too, and add a small boundary test for `Int64.min` / `Int64.max`?

Datadog/Datadog.xcodeproj/project.pbxproj
```diff
@@ -5828,6 +5921,24 @@
 			path = RUMVitals;
 			sourceTree = "<group>";
 		};
+		BB1A2B3C4D5E6F7800000002 /* Timeseries */ = {
+			isa = PBXGroup;
+			children = (
+				BB1A2B3C4D5E6F7800000003 /* DeltaEncoder.swift */,
+				BB1A2B3C4D5E6F7800000001 /* TimeseriesSessionCollector.swift */,
```
gonzalezreal: Could we remove the leftover `TimeseriesSessionCollector` references from the project file?
The target still builds, but these IDs do not point to real project objects, so keeping them makes the Xcode project harder to trust and can cause trouble when it is edited later.

gonzalezreal (commented):
Thanks for the updates here.

I still see two things worth fixing before merging: the delta subtraction can still overflow and crash, and the project file still has leftover `TimeseriesSessionCollector` references.

DatadogRUM/Sources/Timeseries/DeltaEncoder.swift
barboraplasovska: thanks for the suggestion! I think it's better to keep them separate as the two schemas are structurally very different and as you mentioned passing temporary arrays has a real allocation cost per field per batch.

gonzalezreal (approved):
Thanks for fixing the overflow path. LGTM!

The project file still has leftover `TimeseriesSessionCollectorTests` references and a couple of unused source phases.

I do not think that should block approval, but it would be good to clean it up before merging.

DatadogRUM/Sources/Timeseries/DeltaEncoder.swift
arroz: Yeah… based on a few personal projects I've done before, using lazy collections or closures also has some impact. Your version is likely the fastest one from all the reasonable alternatives.

arroz (approved):
As Guille mentioned, there are still some entires in the Xcode file that are not related to this PR, they should probably be removed unless they were missing (sorry, parsing pbxproj files is a nightmare so it's hard to say if it's one way or the other 😅).

Aside from that looks good. Let me know if you need another approval after cleaning.

PR #2889, opened 2026-05-06
Add result bundles to unit tests when running in CI.

arroz:
### What and why?

Generate test result bundles in CI saved as artifacts.

This allows us to easily download and inspect result bundles, which is specially useful when the test runner crashes. In that case there is no way to obtain the crash log except through the result bundle.

### Review checklist


mariedm (approved):

maxep (approved):
Great addition, thanks!

PR #2897, opened 2026-05-08
[PANA-7300] Anchor heatmap identifiers to views with a Session Replay wireframe

gonzalezreal:
### What and why?

RUM tap actions could report a heatmap ID with no matching Session Replay wireframe. The recorder cached IDs for every view, but only views that rendered attached them to a wireframe. RUM looked up by the action target, often a transparent container, so the lookup returned an ID no wireframe carried.

### How?

The recorder now caches heatmap identifiers only for views that produce a wireframe. RUM now looks up by `tap.view` instead of the action target. If `tap.view` has no entry, the action ships without heatmap data; the action itself still fires.

### Review checklist


mariedm (approved):
LGMT!

PR #2914, opened 2026-05-12
Add DDSketch implementation for latency distributions

tienquocbui:
### What and why?

Adds a self-contained DDSketch implementation in `DatadogTrace/Sources/DDSketch/` for
computing approximate latency percentiles in client-side stats. The `okSummary` and
`errorSummary` fields in the stats payload require DDSketch data serialized as protobuf bytes.

DDSketch lives in `DatadogTrace` (not `DatadogInternal`) because it has exactly one consumer
today. The code is self-contained with no SDK dependencies, so extracting it to its own
package later remains straightforward.

### How?

Four source files, ported from the [Go reference implementation](https://github.com/DataDog/sketches-go):

- **`ProtoEncoder`**: Minimal protobuf encoder supporting only the subset needed by
  `ddsketch.proto`: varint, fixed64, sint32/zigzag, length-delimited, and packed doubles.
- **`LogarithmicMapping`**: Maps positive doubles to integer bin indices. 1% relative
  accuracy, `index(for:)` is a line-for-line match with Go's `LogarithmicMapping.Index()`.
- **`CollapsingLowestDenseStore`**: Contiguous bin array that collapses lowest bins when
  exceeding `maxNumBins` (2048). Trades lowest-quantile accuracy for bounded memory.
  Initializer clamps `maxNumBins` to `max(1, ...)` rather than asserting - an SDK must
  never crash the host app on internal misuse.
- **`DDSketch`**: Internal struct with `makeForStats()` factory (1% accuracy, 2048 bins),
  `add()` for recording values, and `toProtoBytes()` for protobuf serialization. All proto
  field numbers verified against [`ddsketch.proto`](https://github.com/DataDog/sketches-go/blob/master/ddsketch/pb/ddsketch.proto).
  Initializer clamps `relativeAccuracy` to `[0.0001, 0.9999]` - the mathematical `(0, 1)`
  boundary is unsafe at runtime because it leaves `multiplier` non-finite.

72 unit tests: property assertions (relative accuracy guarantees, bin bounds, monotonic
indexing, round-trip), edge cases (NaN, infinity, zero, very small values), collapse
correctness in both directions, protobuf wire-format correctness, and fixture-style
example tests for debuggability.

### Review checklist
- [x] Add CHANGELOG entry for user facing changes - N/A (internal, not user-facing)
- [x] Add Objective-C interface for public APIs - N/A (internal to DatadogTrace)
- [x] Run `make api-surface` when adding new APIs - N/A (internal to DatadogTrace)

### DIFF
--- a/DatadogTrace/Sources/DDSketch/CollapsingLowestDenseStore.swift
+++ b/DatadogTrace/Sources/DDSketch/CollapsingLowestDenseStore.swift
@@ -1,0 +1,150 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+/// A contiguous array of bin counts that collapses the lowest bins when the
+/// number of bins exceeds `maxNumBins`.
+///
+/// Ported from the Go reference:
+/// https://github.com/DataDog/sketches-go/blob/master/ddsketch/store/collapsing_lowest_dense_store.go
+///
+/// Collapsing the lowest bins trades accuracy on the lowest quantiles for bounded
+/// memory usage. This is the correct trade-off for latency distributions where
+/// higher percentiles (p50, p90, p99) matter more than p1.
+internal struct CollapsingLowestDenseStore {
+    private(set) var bins: [Double]
+    private(set) var count: Double = 0
+    private(set) var minIndex: Int = 0
+    private(set) var maxIndex: Int = 0
+    private(set) var offset: Int = 0
+    private(set) var isCollapsed: Bool = false
+    let maxNumBins: Int
+    private var isEmpty: Bool = true
+
+    init(maxNumBins: Int) {
+        precondition(maxNumBins > 0, "maxNumBins must be positive")
+        self.maxNumBins = maxNumBins
+        self.bins = []
+    }
+
+    /// Adds `count` to the bin at the given index, extending or collapsing as needed.
+    mutating func add(index: Int, count: Double) {
+        if count == 0 {
+            return
+        }
+
+        if isEmpty {
+            setupFirstValue(index: index)
+        }
+
+        if index < minIndex {
+            if isCollapsed {
+                bins[minIndex - offset] += count
+                self.count += count
+                return
+            }
+            extendRange(newMin: index, newMax: maxIndex)
+            if isCollapsed {
+                bins[minIndex - offset] += count
+                self.count += count
+                return
+            }
+        } else if index > maxIndex {
+            extendRange(newMin: minIndex, newMax: index)
+        }
+
+        let arrayIndex = index - offset
+        bins[arrayIndex] += count
+        self.count += count
+    }
+
+    /// Returns the contiguous bin data for protobuf serialization.
+    /// The `offset` is the index of the first bin in the contiguous array.
+    func contiguousBins() -> (counts: [Double], indexOffset: Int32) {
+        if isEmpty {
+            return ([], 0)
+        }
+
+        let startArrayIndex = minIndex - offset
+        let endArrayIndex = maxIndex - offset
+        let slice = Array(bins[startArrayIndex...endArrayIndex])
+        return (slice, Int32(minIndex))
+    }
+
+    // MARK: - Private
+
+    private mutating func setupFirstValue(index: Int) {
+        isEmpty = false
+        minIndex = index
+        maxIndex = index
+        offset = index
+        bins = [0]
+    }
+
+    private mutating func extendRange(newMin: Int, newMax: Int) {
+        let requiredBins = newMax - newMin + 1
+
+        if requiredBins > maxNumBins {
+            collapse(newMin: newMin, newMax: newMax)
+            return
+        }
+
+        if newMax > maxIndex {
+            let neededCapacity = newMax - offset + 1
+            if neededCapacity > bins.count {
+                bins.append(contentsOf: [Double](repeating: 0, count: neededCapacity - bins.count))
+            }
+            maxIndex = newMax
+        }
+
+        if newMin < minIndex {
+            let shift = offset - newMin
+            if shift > 0 {
+                let newBins = [Double](repeating: 0, count: shift) + bins
+                bins = newBins
+                offset = newMin
+            }
+            minIndex = newMin
+        }
+    }
+
+    /// Rebuilds the bins array to cover exactly `adjustedMin...newMax` within `maxNumBins`.
+    /// Values below `adjustedMin` are folded into the lowest surviving bin.
+    private mutating func collapse(newMin: Int, newMax: Int) {
+        let adjustedMin = newMax - maxNumBins + 1
+
+        if bins.isEmpty || adjustedMin >= maxIndex {
+            let totalCount = count
+            bins = [Double](repeating: 0, count: maxNumBins)
+            bins[0] = totalCount
+            offset = adjustedMin
+            minIndex = adjustedMin
+            maxIndex = newMax
+            isCollapsed = true
+            return
+        }
+
+        var newBins = [Double](repeating: 0, count: maxNumBins)
+        for oldIdx in minIndex...maxIndex {
+            let srcArrayIdx = oldIdx - offset
+            if srcArrayIdx < 0 || srcArrayIdx >= bins.count {
+                continue
+            }
+            let dstIdx = max(oldIdx, adjustedMin)
+            let dstArrayIdx = dstIdx - adjustedMin
+            if dstArrayIdx >= 0 && dstArrayIdx < newBins.count {
+                newBins[dstArrayIdx] += bins[srcArrayIdx]
+            }
+        }
+
+        bins = newBins
+        offset = adjustedMin
+        minIndex = adjustedMin
+        maxIndex = newMax
+        isCollapsed = true
+    }
+}
--- a/DatadogTrace/Sources/DDSketch/CollapsingLowestDenseStore.swift
+++ b/DatadogTrace/Sources/DDSketch/CollapsingLowestDenseStore.swift
@@ -63,8 +63,12 @@ internal struct CollapsingLowestDenseStore {
     private var isEmpty: Bool = true
 
     init(maxNumBins: Int) {
-        precondition(maxNumBins > 0, "maxNumBins must be positive")
-        self.maxNumBins = maxNumBins
+        // An SDK must never crash the host app on internal misuse. A non-positive
+        // `maxNumBins` has no sensible interpretation, so we clamp to 1 and keep
+        // the store operational. The resulting store will only ever hold one bin,
+        // which makes percentiles degenerate, but no metric outside this sketch
+        // is affected.
+        self.maxNumBins = Swift.max(1, maxNumBins)
         self.bins = []
     }
 
--- a/DatadogTrace/Sources/DDSketch/CollapsingLowestDenseStore.swift
+++ b/DatadogTrace/Sources/DDSketch/CollapsingLowestDenseStore.swift
@@ -15,14 +15,51 @@ import Foundation
 /// Collapsing the lowest bins trades accuracy on the lowest quantiles for bounded
 /// memory usage. This is the correct trade-off for latency distributions where
 /// higher percentiles (p50, p90, p99) matter more than p1.
+///
+/// ### Storage model
+///
+/// Bin indices can be any signed integer (positive or negative), but the backing
+/// array `bins` is a flat `[Double]` indexed from `0`. The mapping between a logical
+/// bin index `i` and its position in `bins` is:
+///
+///     bins[i - offset]
+///
+/// `offset` is the logical index of `bins[0]`. `minIndex` and `maxIndex` track the
+/// range of logical indices that currently hold non-zero counts. The store may
+/// reserve more array slots than this range (e.g. after collapsing).
+///
+/// ### Collapsing
+///
+/// When inserting a new index would require more than `maxNumBins` bins to span
+/// `[minIndex, maxIndex]`, the store collapses the lowest bins: the new logical
+/// range becomes `[newMax - maxNumBins + 1, newMax]`, and all counts below that
+/// new minimum are folded into the lowest surviving bin. Once collapsed,
+/// `isCollapsed` stays `true` for the lifetime of the store.
 internal struct CollapsingLowestDenseStore {
+    /// Flat array of bin counts. `bins[i - offset]` holds the count for logical bin `i`.
+    /// May contain trailing zeros beyond `maxIndex` due to capacity reservation.
     private(set) var bins: [Double]
+
+    /// Sum of all bin counts. Updated incrementally by `add`.
     private(set) var count: Double = 0
+
+    /// Lowest logical bin index currently in use. Undefined when the store is empty.
     private(set) var minIndex: Int = 0
+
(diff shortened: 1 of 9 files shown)


tienquocbui:
@codex review

maxep:
QQ: Why hosting it in `DatadogInternal`? who will be the consumers of the `DDSketch`?

tienquocbui:
@maxep The consumer is `StatsConcentrator` in `DatadogTrace`, which needs `DDSketch` to compute the `okSummary`/`errorSummary` latency distributions in client-side stats payloads. Since feature modules can only import `DatadogInternal`, it lives here. We discussed making it a standalone module but agreed with @arroz that a self-contained subdirectory in `DatadogInternal` avoids the extra build target overhead while keeping the code isolated for potential future extraction.

maxep:
If it's only consumed by `DatadogTrace`, can we define it in `DatadogTrace` instead? `DatadogInternal` is only for shared definition with minimal logic (NetworkInstrumentation being the exception)

tienquocbui:
Good point! We discussed this with Miguel and Maciek earlier and agreed on `DatadogInternal` as a middle ground. The code is fully self-contained (no SDK dependencies, own subdirectory) so it's halfway to being extracted into a standalone Swift package later.

tienquocbui:
Moved `DDSketch` from `DatadogInternal` to `DatadogTrace` per @maxep's feedback, since `DatadogTrace` is the only consumer, it fits better here. `DatadogInternal` should stay focused on shared interfaces, not standalone utilities. The code remains self-contained (no SDK dependencies) so future extraction is still straightforward.

DatadogTrace/Sources/DDSketch/CollapsingLowestDenseStore.swift
```diff
@@ -0,0 +1,150 @@
...
+            if isCollapsed {
+                bins[minIndex - offset] += count
+                self.count += count
+                return
+            }
+        } else if index > maxIndex {
+            extendRange(newMin: minIndex, newMax: index)
+        }
+
+        let arrayIndex = index - offset
```
arroz: Nitpick: just a matter of style, but on the other assignments above, the index calculation was done inline, like `bins[minIndex - offset]`. Why doing it differently here?

DatadogTrace/Sources/DDSketch/CollapsingLowestDenseStore.swift
```diff
@@ -0,0 +1,150 @@
...
+        bins[arrayIndex] += count
+        self.count += count
+    }
+
+    /// Returns the contiguous bin data for protobuf serialization.
+    /// The `offset` is the index of the first bin in the contiguous array.
+    func contiguousBins() -> (counts: [Double], indexOffset: Int32) {
```
arroz: You can save the creation of an array (and copying memory, etc) by returning `Array<Double>.SubSequence` instead of [Double]. Then, in `ProtoEncoder.encodePackedDoubles(…)` you can change `values` type to `Array<Double>.SubSequence` as well.

DatadogTrace/Sources/DDSketch/CollapsingLowestDenseStore.swift
```diff
@@ -0,0 +1,150 @@
...
+        if isEmpty {
+            return ([], 0)
+        }
+
+        let startArrayIndex = minIndex - offset
+        let endArrayIndex = maxIndex - offset
```
arroz: When calculating indexes, it's always good practice to base them on offsets from the collection start or end index. Although skipping that works here, this makes the code safer if, in the future, an optimization is added where the array turns into a subarray, like what I suggested above. Never assume the `startIndex` of a collection is 0 (and in many collections the `Index` type may be something other than an `Int`).

```suggestion
        let startArrayIndex = bins.startIndex.advanced(by: minIndex - offset)
        let endArrayIndex = bins.startIndex.advanced(by: maxIndex - offset)
```

DatadogTrace/Sources/DDSketch/CollapsingLowestDenseStore.swift
```diff
@@ -0,0 +1,150 @@
...
+        }
+    }
+
+    /// Rebuilds the bins array to cover exactly `adjustedMin...newMax` within `maxNumBins`.
+    /// Values below `adjustedMin` are folded into the lowest surviving bin.
+    private mutating func collapse(newMin: Int, newMax: Int) {
+        let adjustedMin = newMax - maxNumBins + 1
+
+        if bins.isEmpty || adjustedMin >= maxIndex {
+            let totalCount = count
```
arroz: Why creating an additional var and not just using `count` directly below?

DatadogTrace/Sources/DDSketch/CollapsingLowestDenseStore.swift
```diff
@@ -0,0 +1,150 @@
...
+        var newBins = [Double](repeating: 0, count: maxNumBins)
+        for oldIdx in minIndex...maxIndex {
+            let srcArrayIdx = oldIdx - offset
+            if srcArrayIdx < 0 || srcArrayIdx >= bins.count {
+                continue
+            }
+            let dstIdx = max(oldIdx, adjustedMin)
+            let dstArrayIdx = dstIdx - adjustedMin
+            if dstArrayIdx >= 0 && dstArrayIdx < newBins.count {
+                newBins[dstArrayIdx] += bins[srcArrayIdx]
```
arroz: Same comment as before about basing indexes on the array's `startIndex`.

DatadogTrace/Sources/DDSketch/CollapsingLowestDenseStore.swift
```diff
@@ -0,0 +1,150 @@
...
+            bins[0] = totalCount
+            offset = adjustedMin
+            minIndex = adjustedMin
+            maxIndex = newMax
+            isCollapsed = true
+            return
+        }
+
+        var newBins = [Double](repeating: 0, count: maxNumBins)
+        for oldIdx in minIndex...maxIndex {
```
arroz: Is this easier or harder than just calculating two offsets and the count of values to copy, and then copying those? Maybe even doing `newBins: = [Double](repeating: 0, count: x) + Array(bins[y..<z] + [Double](repeating: 0, count: w)`. Might be a bit more performant as well, but I'm not 100% sure, profiling would be needed to confirm. But it would at least be easier to read and understand the algorithm (since the construction of the new array is clearly 0 or more of zeros, then part of the older one, followed by 0 or more zeros).

DatadogTrace/Sources/DDSketch/CollapsingLowestDenseStore.swift
```diff
@@ -0,0 +1,150 @@
...
+/// number of bins exceeds `maxNumBins`.
+///
+/// Ported from the Go reference:
+/// https://github.com/DataDog/sketches-go/blob/master/ddsketch/store/collapsing_lowest_dense_store.go
+///
+/// Collapsing the lowest bins trades accuracy on the lowest quantiles for bounded
+/// memory usage. This is the correct trade-off for latency distributions where
+/// higher percentiles (p50, p90, p99) matter more than p1.
+internal struct CollapsingLowestDenseStore {
+    private(set) var bins: [Double]
```
arroz, line 41: This struct could use more documentation, on variables and functions, explaining what they are used for in the algorithm.

DatadogTrace/Sources/DDSketch/DDSketch.swift
```diff
@@ -0,0 +1,134 @@
...
+/// so it can be extracted to a standalone repository.
+///
+/// Usage for client-side stats:
+/// ```
+/// var sketch = DDSketch.makeForStats()
+/// sketch.add(durationInNanoseconds)
+/// let protoBytes = sketch.toProtoBytes()
+/// ```
+internal struct DDSketch {
+    let mapping: LogarithmicMapping
```
arroz, line 39: Same comment as above about documentation.

DatadogTrace/Sources/DDSketch/LogarithmicMapping.swift
```diff
@@ -0,0 +1,73 @@
...
+    let minIndexableValue: Double
+
+    /// Largest positive value that can be mapped to a valid `Int32` index.
+    let maxIndexableValue: Double
+
+    /// The target relative accuracy, in (0, 1). A value of 0.01 means 1% accuracy.
+    let relativeAccuracy: Double
+
+    init(relativeAccuracy: Double) {
+        precondition(relativeAccuracy > 0 && relativeAccuracy < 1, "relativeAccuracy must be in (0, 1)")
```
arroz:
```suggestion
        precondition(relativeAccuracy > 0 && relativeAccuracy < 1, "relativeAccuracy must be in [0, 1]")
```

DatadogTrace/Sources/DDSketch/LogarithmicMapping.swift
```diff
@@ -0,0 +1,73 @@
...
+        self.indexOffset = 0.0
+
+        self.minIndexableValue = max(
+            exp((Double(Int32.min) - indexOffset) / multiplier + 1),
+            Double.leastNormalMagnitude * gamma
+        )
+
+        self.maxIndexableValue = min(
+            exp((Double(Int32.max) - indexOffset) / multiplier - 1),
+            exp(709.0) / (2.0 * gamma / (1.0 + gamma))
```
arroz: What is this 709 number? Can it be moved to a constant with a name that expresses its meaning?

DatadogTrace/Sources/DDSketch/ProtoEncoder.swift
```diff
@@ -0,0 +1,109 @@
...
+
+/// Minimal protobuf encoder for the DDSketch wire format.
+///
+/// Supports only the subset of proto3 needed by `ddsketch.proto`:
+/// varint, fixed64 (double), sint32 (zigzag), length-delimited (nested messages),
+/// and packed repeated doubles.
+///
+/// This encoder is intentionally self-contained with no SDK dependencies
+/// so the DDSketch code can be extracted to a standalone repository.
+internal struct ProtoEncoder {
```
arroz, line 17: Have we considered the pros and cons of using [Swift Protobuf](https://github.com/apple/swift-protobuf) instead? Specifically, how much does the compiled binary grow in `Release` mode, after the usual dead code removing, compared to our version? I'm not against leaving our implementation in, but if the resulting different in the binary size is not very significative, I would be happy with not having to maintain this code.

tienquocbui: We considered it. The main reasons for a custom encoder: 
- (1) adding `apple/swift-protobuf` is a new dependency, which goes against the SDK's small-footprint principle - the library adds ~200KB to binary size even after dead code stripping, 
- (2) we only need ~100 lines of write-only encoding for 5 wire types, no decoding, 
- (3) the maintenance burden is minimal since the `DDSketch` proto schema is stable and unlikely to change

DatadogTrace/Sources/DDSketch/LogarithmicMapping.swift
tienquocbui: `709` is the largest safe exponent for `exp()` before IEEE 754 double overflow (`exp(709.78...) ≈ Double.greatestFiniteMagnitude`). It comes directly from the [Go reference](https://github.com/DataDog/sketches-go/blob/master/ddsketch/mapping/logarithmic_mapping.go#L49). I'll extract it to a named constant.

maxep (approved):
LGTM, well done 🎖️

DatadogTrace/Sources/DDSketch/LogarithmicMapping.swift
tienquocbui: The parentheses are intentional here, `(0, 1)` is open-interval notation meaning both endpoints are excluded, which matches the `> 0 && < 1` check (strict on both sides). `[0, 1]` would imply 0 and 1 are valid, but they aren't: `at relativeAccuracy = 0,` `multiplier = 1/log(1) = 1/0` is undefined, and at 1, `gamma = 2/0` blows up. So the open interval is mathematically required.

DatadogTrace/Sources/DDSketch/CollapsingLowestDenseStore.swift
```diff
@@ -0,0 +1,227 @@
...
+    private(set) var isCollapsed: Bool = false
+
+    /// Maximum number of bins the store will allocate before collapsing.
+    let maxNumBins: Int
+
+    /// `true` until the first non-zero `add(index:count:)` call.
+    private var isEmpty: Bool = true
+
+    init(maxNumBins: Int) {
+        precondition(maxNumBins > 0, "maxNumBins must be positive")
```
ncreated: blocking/ Unless we have extremely good reasons, we should avoid running `precondition()` in customer app code. Otherwise, the app may crash in production if the precondition is not met.

Using `assertion()` here would be a slightly better option, since assertions are disabled in the `Release` configuration by default.

However, IMO neither option is the right choice here, since this failure is purely internal to `DatadogTrace` and the user cannot do anything about it. In this case, the SDK should be able to recover from the failure and not punish the app.

DatadogTrace/Sources/DDSketch/CollapsingLowestDenseStore.swift
```diff
@@ -0,0 +1,227 @@
...
+
+        if bins.isEmpty || adjustedMin >= maxIndex {
+            bins = [Double](repeating: 0, count: maxNumBins)
+            bins[0] = count
+            offset = adjustedMin
+            minIndex = adjustedMin
+            maxIndex = newMax
+            isCollapsed = true
+            return
+        }
```
ncreated, line 188: nit/ Test coverage is a great metric for assessing the completeness of a PR like this (almost pure arithmetic, fully deterministic, manageable problem space). The coverage in this PR is amazing, except for this early return, which isn’t checked 🙂

Image

DatadogTrace/Tests/DDSketch/DDSketchTests.swift
```diff
@@ -0,0 +1,248 @@
...
+        let durations: [Double] = [
+            1_000_000,    // 1ms
+            5_000_000,    // 5ms
+            10_000_000,   // 10ms
+            50_000_000,   // 50ms
+            100_000_000,  // 100ms
+            500_000_000,  // 500ms
+            1_000_000_000, // 1s
+            5_000_000_000  // 5s
+        ]
```
ncreated, line 155: question/ Where do these fixtures come from? How does our testing strategy compare to the one in [ddsketch_test.go](https://github.com/DataDog/sketches-go/blob/master/ddsketch/ddsketch_test.go)?

I’m wondering if we can prove that Swift implementation produces the same results as the Go one.

ncreated (changes requested):
Great work and excellent test coverage 🏅! I left a few comments, including blocking feedback on the use of `precondition()`.

DatadogTrace/Tests/DDSketch/DDSketchTests.swift
tienquocbui: The fixtures are synthetic, I picked values that mirror typical iOS span durations (1ms -> 5s spread) rather than lifting from `ddsketch_test.go`. The other tests are property-style (relative accuracy, bin bounds, deterministic output, round-trip through protobuf), which is the same shape as the Go reference tests, just not the same inputs.

For proving parity with Go, I see three levels:
1. Port one or two property tests from `ddsketch_test.go` directly.
2. Offline byte-equality: run identical sequences through Swift + Go, diff the protobuf bytes. One-off script, not CI.
3. End-to-end through the APM stats intake, the canonical proof, lands with the integration PR.

 I DM’d you to align on which of these we want in this PR vs follow-up.

ncreated (approved):
Great work! 🚀 🏅

PR #2919, opened 2026-05-13
[RUM-16084] Fetch and store remote configuration file at SDK init

saraSr5:
### What and why?

Implements the first building block of the Remote configuration: fetching the remote config document from the Datadog CDN at SDK startup and caching the raw JSON to disk.                                                                                                                                                                                                                                                            
  
When a `remoteConfigurationID` is provided in `Datadog.Configuration`, the SDK asynchronously fetches the corresponding config document and writes it atomically to `remote-config.json` inside the SDK's private core directory. On any failure (network error, non-2xx, empty body, invalid JSON), a telemetry error is reported and any existing cached file is left untouched.                                                                                                                                                                     
Parsing the cached file and applying its values to SDK configuration is **out of scope** for this PR.                                                                                                                                                                        
                                            
  ### How?

  - Added `remoteConfigurationID: String?` to `Datadog.Configuration` (defaults to `nil`, no behaviour change for existing apps)
  - Added `DatadogSite.remoteConfigurationURL(for:)` to build the CDN URL from the ID and resolved site                                                                                                                                                                        
  - Added `RemoteConfigurationCache` — thin wrapper around an atomic file write/read in the core directory
  - Added `RemoteConfigurationFetcher` — fires a one-shot `URLSession` `dataTask`, validates the response (2xx, non-empty, valid JSON), and delegates storage to the cache                                                                                                     
  - Wired everything in `Datadog.initialize()` — fire-and-forget, does not block SDK init
                                                                                                                                                                                                                                                                               
  ### Review checklist                      


saraSr5:
@codex review

saraSr5:
@codex review

DatadogCore/Sources/Core/RemoteConfigurationFetcher.swift
```diff
@@ -0,0 +1,84 @@
...
+
+            // 4. Invalid JSON
+            // Intentional allocate-and-discard: we only need to validate the bytes
+            // are well-formed JSON before caching. The parsed object is thrown away.
+            // This guarantees the cache never contains non-JSON data, so future
+            // parsing layers can trust the cached bytes without re-validating.
+            guard (try? JSONSerialization.jsonObject(with: data)) != nil else {
+                telemetry.error("[RemoteConfig] Response is not valid JSON")
+                return
+            }
```
maxep: Lets remove this check for now, we will validate against the actual schema later. Lets create a follow up Jira ticket and add a TODO comment:
```suggestion
            // TODO RUM-XXXX: Validate the schema before saving
```

DatadogCore/Sources/Core/RemoteConfigurationFetcher.swift
```diff
@@ -0,0 +1,84 @@
...
+                let code = (response as? HTTPURLResponse)?.statusCode ?? -1
+                // Use a fixed message so all HTTP errors bucket together in telemetry;
+                // the status code lives in the error object, not the message string.
+                telemetry.error("[RemoteConfig] Non-2xx response", error: NSError(
+                    domain: "RemoteConfiguration",
+                    code: code,
+                    userInfo: [NSLocalizedDescriptionKey: "HTTP \(code)"]
+                ))
+                return
+            }
```
maxep: I have added these 2 conditions in the [RFC (internal)](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/6587975318/RFC+-+Mobile+Remote+Configuration#Caching-Strategy):

> The following conditions will be applied at each fetch:
> 
> ETag-based conditional requests: Store the RC response alongside the ETag header from CDN. Background fetch sends If-None-Match. CDN returns 304 if unchanged, saving bandwidth.
> 
> TTL-based revalidation: Only revalidate if the cached config is older than a TTL, e.g. 5 mns, to skip fast background/foreground navigations which are frequent on mobile.

Can we address them or add a TODO with a follow up Jira ticket?

DatadogCore/Sources/Core/RemoteConfigurationCache.swift
```diff
@@ -0,0 +1,64 @@
...
+
+    private let fileURL: URL
+
+    /// Raw JSON bytes from the previous CDN fetch, read synchronously at init.
+    /// `nil` when no cache exists yet (first launch, or remoteConfigurationID was never set).
+    /// Consumed by the config-application layer in a follow-on ticket (out of scope for RUM-16084).
+    private(set) var data: Data?
+
+    init(directory: Directory) {
+        self.fileURL = directory.url.appendingPathComponent(Self.fileName)
```
maxep: lets use the config-is as file name, the [RFC (internal)](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/6587975318/RFC+-+Mobile+Remote+Configuration#File-location) was recently updated with that spec

DatadogCore/Sources/Core/RemoteConfigurationCache.swift
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    // MARK: - Private
+
+    private static func readFromDisk(at url: URL) -> Data? {
+        guard FileManager.default.fileExists(atPath: url.path)
+        else {
+            return nil
+        }
+        return try? Data(contentsOf: url)
```
maxep: We can report a telemetry error if the file failed to load

DatadogInternal/Sources/Context/DatadogSite.swift
```diff
@@ -48,4 +48,21 @@ extension DatadogSite {
...
+        // Format: https://sdk-configuration.browser-intake-{site}/v1/{id}.json
+        // `.urlPathAllowed` leaves `/`, `?`, and `#` unencoded (they are legal in a URL).
+        // Subtract them so an ID containing those characters doesn't produce extra path
+        // segments, a query string, or a fragment.
+        let pathSegmentAllowed = CharacterSet.urlPathAllowed.subtracting(CharacterSet(charactersIn: "/?#"))
+        guard let encoded = id.addingPercentEncoding(withAllowedCharacters: pathSegmentAllowed) else {
+            return nil
+        }
+        // swiftlint:disable:next force_unwrapping
+        return URL(string: "https://sdk-configuration.\(endpoint.host!)/v1/\(encoded).json")
```
maxep: suggestion/ Lets move this to the `RemoteConfigurationFetcher` and use `URLComponents` to construct the URL.

saraSr5:
@codex review

saraSr5:
@codex review

DatadogCore/Sources/Core/DatadogCore.swift
```diff
@@ -255,6 +262,19 @@ internal final class DatadogCore {
...
+    /// Called from `Datadog.initialize()` when `remoteConfigurationID` is set.
+    /// The fetch is fire-and-forget — SDK init does not wait for it.
+    internal func fetchRemoteConfiguration(from endpoint: URL, session: URLSession) {
+        let fetcher = RemoteConfigurationFetcher(
+            cache: remoteConfigCache,
+            telemetry: telemetry,
+            session: session
+        )
+        fetcher.fetch(from: endpoint)
+    }
```
maxep: suggestion/ Lets use a convenience `RemoteConfigurationFetcher.init` instead of a `fetchRemoteConfiguration` method in core. Something like:
```swift
init(
  cache: RemoteConfigurationCache,
  connectionProxyDictionary: [:],
  telemetry: Telemetry
)
```

The `URLSession` can be initialize directly in the convenience init.

DatadogCore/Sources/Core/DatadogCore.swift
```diff
@@ -114,7 +121,7 @@ internal final class DatadogCore {
         self.contextProvider.subscribe(\.accountInfo, to: accountInfoPublisher)
         self.contextProvider.subscribe(\.version, to: applicationVersionPublisher)
         self.contextProvider.subscribe(\.trackingConsent, to: consentPublisher)
-
+        self.remoteConfigCache = RemoteConfigurationCache(directory: directory.coreDirectory)
```
maxep: not/ Lets keep it in `directory.coreDirectory` for now, but I think we would need a more persistent folder: `directory.coreDirectory` is in cache which can be cleared by the OS

maxep (commented):
Looks good! I left some suggestions

DatadogCore/Sources/Core/DatadogCore.swift
maxep: Also, I think the RemoteConfigurationCache should be an init parameter instead

DatadogCore/Sources/Datadog.swift
```diff
@@ -415,12 +444,22 @@ extension DatadogCore {
...
 
+        let directory = try CoreDirectory(
+            in: configuration.systemDirectory(),
+            instanceName: instanceName,
+            site: configuration.site
+        )
+
+        let remoteConfigCache: RemoteConfigurationCache?
+        if let rawID = configuration.remoteConfigurationID {
+            let id = rawID.trimmingCharacters(in: .whitespacesAndNewlines)
```
maxep: We don't need to sanitize the ID

DatadogCore/Sources/DatadogConfiguration.swift
```diff
@@ -228,5 +244,10 @@ extension Datadog {
 
         /// The default application state provider for accessing [application state](https://developer.apple.com/documentation/uikit/uiapplication/state).
         internal var appStateProvider: AppStateProvider = DefaultAppStateProvider()
+
+        /// Override the URLSession used for remote configuration fetching. For tests only.
+        /// When `nil` (the default), `Datadog.initialize()` builds a session from
+        /// `proxyConfiguration` — matching the proxy behaviour of the main HTTP client.
+        internal var remoteConfigurationSession: URLSession? = nil
```
maxep: I don't think we need this, we should inject the URLSession into the underlying remote-config object instead of in the core config

DatadogCore/Sources/Datadog.swift
```diff
@@ -297,6 +298,34 @@ public enum Datadog {
...
+                        fetcher = RemoteConfigurationFetcher(cache: cache, telemetry: core.telemetry, session: session)
+                    } else {
+                        fetcher = RemoteConfigurationFetcher(cache: cache, connectionProxyDictionary: configuration.proxyConfiguration, telemetry: core.telemetry)
+                    }
+                    fetcher.fetch(from: endpoint)
+                }
+            } else {
+                core.telemetry.error("[RemoteConfig] Could not build CDN URL for remoteConfigurationID '\(id)'")
+            }
+        }
```
maxep: There is too much logic in this method, we should have very little complexity in `Datadog.initializeOrThrow`, we need to isolate this. I would suggest to create a `RemoteConfiguration` object instead, that is responsible for both Fetch and Cache and that is own by the Core. Similar to how `UserInfoPublisher` or `ApplicationVersionPublisher` works. This will both clarify the initialization flow and ease testability.

maxep (changes requested):

saraSr5:
@codex review

DatadogInternal/Sources/Context/DatadogSite.swift
```diff
@@ -48,4 +48,13 @@ extension DatadogSite {
         // swiftlint:enable force_unwrapping
         }
     }
+
+    /// The CDN hostname used for fetching remote configuration documents.
+    /// Format: `sdk-configuration.<intake-host>`
+    ///
+    /// The full URL (with path and ID) is constructed by `RemoteConfigurationSynchronizer`.
+    @_spi(Internal)
```
maxep: No need for _spi accessor here, we are already in the Internal module
```suggestion
```

DatadogInternal/Tests/Context/DatadogSiteTests.swift
```diff
@@ -0,0 +1,45 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import XCTest
+@_spi(Internal)
```
maxep:
```suggestion
```

DatadogInternal/Sources/Context/DatadogSite.swift
```diff
@@ -48,4 +48,13 @@ extension DatadogSite {
...
+    public var remoteConfigurationHost: String? {
```
maxep: What I meant in my feedback was this:
```swift
extension DatadogSite {
  public var host: String {
    ...
  }
  
  public var endpoint: URL {
    return URL(string: "https://\(host)/")!
  }

  public var remoteConfigurationEndpoint: URL {
    return URL(string: "https://sdk-configuration.\(host)/")!
  }
}
```

DatadogCore/Sources/Core/RemoteConfigurationSynchronizer.swift
```diff
@@ -0,0 +1,212 @@
...
+// MARK: - RemoteConfigurationSynchronizer
+
+/// Owns the remote configuration cache and fetch lifecycle.
+///
+/// Created and started by `DatadogCore` during initialization when a `remoteConfigurationID`
+/// is provided. `start(from:connectionProxyDictionary:telemetry:)` fires the initial CDN fetch.
+///
+/// TODO: Also trigger `start()` on every app foreground transition so remote config
+/// is refreshed while the app is in use, not only at SDK init (RFC §Caching Strategy).
+internal final class RemoteConfigurationSynchronizer {
```
maxep, line 17: Lets not overly complexify this with 3 layers. I would propose the following simple design that uses a single object:
```swift
internal final class RemoteConfigurationSynchronizer {
    let id: String
    let site: DatadogSite
    let directory: Directory
    let httpClient: HTTPClient

    @ReadWriteLock
    private(set) var cache: Result<Data, Error>

    init(
        id: String,
        site: DatadogSite,
        directory: Directory,
        httpClient: HTTPClient
    ) {
        self.id = id
        self.site = site
        self.directory = directory
        self.httpClient = httpClient

        self.cache = // 0: read the cache directory
    }

    func sync(_ completionHandler: (Result<Data, Error>) -> Void) {
        let endpoint = site.remoteConfigurationEnpoint.appendingPathComponent("\(id).json")
        httpClient.send(request: URLRequest(url: endpoint) { result in
            // 1: process the response
            // 2: save the config in cache
            // 3: set the cache property
            // 4: call completionHandler with the new cache
        }
    }
}
```

Not that I am using the `HTTPClient` which will already be configured with the proxy.

DatadogCore/Sources/Core/RemoteConfigurationSynchronizer.swift
```diff
@@ -0,0 +1,212 @@
...
+        // present after a previous successful fetch — absent on first launch.
+        let (data, error) = Self.readFromDisk(at: fileURL)
+        self.data = data
+        self.loadError = error
+    }
+
+    // MARK: - Private
+
+    private static func readFromDisk(at url: URL) -> (Data?, Error?) {
+        guard FileManager.default.fileExists(atPath: url.path) else {
```
maxep: The `Directory` abstraction already provide all interfaces to interact with folders and files, it should be used instead of `FileManager`. Same when saving the file.

maxep (changes requested):
We are getting closer :) I still see opportunities to simplify and reuse existing utilities.

maxep (approved):
Great work 🎖️ Thank you for the follow ups on the feedback 🙏

saraSr5:
> Great work 🎖️ Thank you for the follow ups on the feedback 🙏

@maxep Thanks for your help!! :)

PR #2924, opened 2026-05-18
 RUM-4011: add DDAssertDiff() test helper

saraSr5:
### What and why?

Adds `DDAssertDiff()` to `TestUtilities`. It is a helper for strict structural assertions on the diff between two `Encodable` values and leverages it in the crash and fatal-app-hang recovery tests.
The recovery code that bypasses tracking consent on app restart could silently mutate extra fields, leaking data collected without granted consent. `DDAssertDiff` enables tests to assert exactly which fields are mutated, nothing more.

### How?

 Encodes both values via `JSONEncoder`, deserializes through `JSONSerialization`, flattens to dot-separated key paths, and reports the diff via `Set` operations. Adds a `DiffResult.assertExact(different:added:removed:)` convenience used by both real-test conversions.
Seeds `viewIsActive: true` in both target tests so the recovery code's `isActive → false` flip is deterministically observable in the diff.
Pre-existing field-by-field assertions in the converted tests are kept alongside `DDAssertDiff` for clearer failure messages on individual fields.

### Review checklist


saraSr5:
@codex review

saraSr5:
@codex review

TestUtilities/Sources/Helpers/DDAssert.swift
```diff
@@ -252,3 +252,132 @@ public func DDAssertEqual(_ date1: Date?, _ date2: Date, accuracy: TimeInterval,
...
+    if lhsIsBool != rhsIsBool {
+        return false
+    }
+    return lhs.isEqual(rhs)
+}
+
+private func _DDAssertDiff<T: Encodable>(
+    _ expression1: @autoclosure () throws -> T,
+    _ expression2: @autoclosure () throws -> T
+) throws -> DiffResult {
```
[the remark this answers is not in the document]
saraSr5: Keeping a single T on purpose, if you diff two different types you'd just get everything as added/removed, which isn't useful. We can add a second generic later if there's a real need for it.

TestUtilities/Sources/Helpers/DDAssert.swift
```diff
@@ -252,3 +252,132 @@ public func DDAssertEqual(_ date1: Date?, _ date2: Date, accuracy: TimeInterval,
...
+    public let removedKeyPaths: [String]
+    /// Key paths present in `expression2` but missing in `expression1`.
+    public let addedKeyPaths: [String]
+}
+
+private func _flatten(_ value: Any, prefix: String) -> [String: NSObject] {
+    if let dict = value as? [String: Any] {
+        if dict.isEmpty {
+            return [prefix: NSDictionary()]
+        }
```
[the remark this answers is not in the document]
saraSr5: This is an edge case that doesn't arise with any actual SDK model type, they all encode to non-empty JSON objects. Adding complexity to handle it isn't worth it right now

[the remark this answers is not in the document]
simaoseica-dd: It might not happen, but it is verified below before prefixing: `prefix.isEmpty ?`
The suggestion seems clean

TestUtilities/Sources/Helpers/DDAssert.swift
```diff
@@ -252,3 +252,132 @@ public func DDAssertEqual(_ date1: Date?, _ date2: Date, accuracy: TimeInterval,
         XCTAssertEqual(date1.timeIntervalSince1970, date2.timeIntervalSince1970, accuracy: accuracy, message(), file: file, line: line)
     }
 }
+
+public struct DiffResult {
```
simaoseica-dd: Seems too generic for a public symbol 
```suggestion
public struct JSONDiff {
```

TestUtilities/Sources/Helpers/DDAssert.swift
```diff
@@ -252,3 +252,132 @@ public func DDAssertEqual(_ date1: Date?, _ date2: Date, accuracy: TimeInterval,
...
+    /// Asserts the diff contains exactly these key paths in each category.
+    /// Pass empty arrays (or omit) for categories you expect to be empty.
+    func assertExact(
+        different: [String] = [],
+        added: [String] = [],
+        removed: [String] = [],
+        file: StaticString = #fileID,
+        line: UInt = #line
+    ) {
+        XCTAssertEqual(differentKeyPaths, different.sorted(), file: file, line: line)
```
simaoseica-dd: Can the `XCTest` dependency be removed? I am thinking on decoupling it from `XCTest` to be used in [Swift Testing](https://developer.apple.com/xcode/swift-testing/).

maxep: The entire `DDAsset` suite is targeting `XCTest`. We should use new utilities for Swift Testing.

DatadogCore/Tests/Datadog/TestUtilities/DDAssertDiffTests.swift
```diff
@@ -0,0 +1,286 @@
...
+// Test structs and enums in this file use single-letter property names (a, b, x, etc.) for brevity —
+// these types are local to each test method and never escape, so descriptive names would only add noise.
+
+private struct ThrowingEncodable: Encodable {
+    func encode(to encoder: Encoder) throws {
+        throw NSError(domain: "test", code: 0)
+    }
+}
+
+class DDAssertDiffTests: XCTestCase {
```
maxep: `DDAssert` is part of the `TestUtilities` module, it doesn't need tests. We can considered that it is tested while used.

TestUtilities/Sources/Helpers/DDAssert.swift
maxep: I would also suggest to place all functions (`_flatten`, `leafEqual`) as static methods of that object, to place them under the same domain.

TestUtilities/Sources/Helpers/DDAssert.swift
```diff
@@ -252,3 +252,146 @@ public func DDAssertEqual(_ date1: Date?, _ date2: Date, accuracy: TimeInterval,
...
+    let addedKeys = keys2.subtracting(keys1)
+
+    return JSONDiff(
+        differentKeyPaths: differentKeys.sorted(),
+        removedKeyPaths: removedKeys.sorted(),
+        addedKeyPaths: addedKeys.sorted()
+    )
+}
+
+public func DDAssertDiff<T: Encodable>(
```
maxep: I would suggest to call it `DDAssertJSONDiff` to align with the existing `DDAssertJSONEqual`

simaoseica-dd (approved):

maxep (approved):

PR #2927, opened 2026-05-19
[chain-pr 1/10] Typed MessageBus protocol and shared message types

maxep:
> This PR is part of a chain split from _maxep/message-bus-subscription_ by [chain-pr](https://github.com/maxep/chain-pr).

### What and why?

The existing bus delivers all messages through a `FeatureMessage` enum. Core-level messages have typed cases (`.context`, `.telemetry`, `.webview`), but feature-to-feature messages all go through `.payload(Any)` — the sender wraps an arbitrary value, every `FeatureMessageReceiver` is called for every delivery, and each must attempt a runtime type-cast to decide whether the message is relevant. The new `MessageBus` protocol routes by Swift type: subscribers register for exactly the concrete type they handle and only receive deliveries for that type, eliminating the runtime casts and the fan-out.

This PR is the **foundation layer**: it defines the new protocols, conforms all existing shared message types to `BusMessage`, introduces typed WebView bridge message types, and deprecates the legacy API surface so subsequent PRs can migrate each SDK module without touching this one.

#### New protocol surface

Defined in `DatadogInternal/Sources/MessageBus/MessageBus.swift`:

| Type | Role |
|---|---|
| `MessageBus` | Protocol with `subscribe(receiver:)`, `unsubscribe(receiver:)`, and `send(message:else:)`, all typed on the concrete `BusMessage` conformance |
| `BusMessage` | Marker protocol for any value that can travel on the bus; each conformance declares a stable `key` for diagnostics |
| `BusMessageReceiver` | Class-bound protocol with an associated `Message` type; identity (`===`) is used to track subscriptions |
| `MessageBusSubscription` | Opaque handle returned by the closure-based `subscribe(block:)` convenience; pass to `unsubscribe(_:)` for teardown |
| `NOPMessageBus` | No-op default; always invokes the `fallback` closure on `send` |

#### Core protocol

`DatadogCoreProtocol` gains a `messageBus: MessageBus` property with a default `NOPMessageBus()` — all existing cores, mocks, and test utilities compile without modification. Concrete cores will wire up the real bus in PR #3.

#### Deprecations

`FeatureMessage`, `FeatureMessageReceiver`, `NOPFeatureMessageReceiver`, `CombinedFeatureMessageReceiver`, `MessageSending`, and `DatadogFeature.messageReceiver` are marked `@available(*, deprecated)` with inline guidance pointing to the replacement types.

#### Message type conformances

Existing types conformed to `BusMessage` (no routing changes — purely additive):

`DatadogContext` · `Crash` · `LogMessage` · `LogEventAttributes` · `TelemetryMessage` · `RUMSessionState` · `RUMErrorMessage` · `RUMFlagEvaluationMessage` · `RUMViewEvent` · `RUMEventAttributes`

New types added:
- `RUMViewReset` — signals the active view stopped with no successor
- `WebViewLogMessage`, `WebViewRUMMessage`, `WebViewRecordMessage` — typed replacements for the untyped `.payload` WebView bridge messages

#### Other

- `DatadogCore.MessageBus` gains a `Telemetry` conformance in production code (previously a retroactive conformance in the test target).
- New unit tests in `DatadogInternal/Tests/MessageBus/MessageBusTests.swift` covering closure-based subscription, type-based routing, and unsubscription via handle.

### Review checklist

---

## Chain overview

| # | PR | Title |
|---|---|---|
| **1** | ← _this PR_ | Typed MessageBus protocol and shared message types |
| 2 | — | Test infrastructure for typed message bus |
| 3 | — | DatadogCore: CoreMessageBus implementation and rename |
| 4 | — | DatadogCrashReporting: typed-bus migration |
| 5 | — | DatadogLogs: typed-bus migration |
| 6 | — | DatadogTrace: typed-bus migration |
| 7 | — | DatadogFlags: typed-bus migration |
| 8 | — | DatadogRUM: typed-bus migration and TelemetryInterceptor merge |
| 9 | — | DatadogSessionReplay and DatadogWebViewTracking: typed-bus migration |
| 10 | — | Documentation updates |

```mermaid
graph LR
    G1["1 — Typed MessageBus protocol and shared message types"]
    G2["2 — Test infrastructure for typed message bus"]
    G3["3 — DatadogCore: CoreMessageBus implementation and rename"]
    G4["4 — DatadogCrashReporting: typed-bus migration"]
    G5["5 — DatadogLogs: typed-bus migration"]
    G6["6 — DatadogTrace: typed-bus migration"]
    G7["7 — DatadogFlags: typed-bus migration"]
    G8["8 — DatadogRUM: typed-bus migration and TelemetryInterceptor merge"]
    G9["9 — DatadogSessionReplay and DatadogWebViewTracking: typed-bus migration"]
    G10["10 — Documentation updates"]
    G1 --> G2
    G2 --> G3
    G3 --> G4
    G3 --> G5
    G5 --> G6
    G3 --> G7
    G6 --> G8
    G7 --> G8
    G8 --> G9
    G1 --> G10
```

_Merge in dependency order. PRs with no incoming edges from un-merged PRs can be reviewed in parallel._

### DIFF
--- a/DatadogInternal/Sources/MessageBus/MessageBus.swift
+++ b/DatadogInternal/Sources/MessageBus/MessageBus.swift
@@ -1,0 +1,198 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+/// A typed publish/subscribe channel for messages exchanged between Features
+/// registered to the same core.
+///
+/// Every message is a nominal type conforming to `BusMessage`. Subscribers declare
+/// the message type they handle and the bus uses Swift's type system for routing —
+/// no runtime casts at the call site, no shared enum to extend when a new message
+/// is introduced.
+///
+/// ## Subscription lifetime
+///
+/// The bus retains each subscriber until it is explicitly removed via `unsubscribe`.
+/// Long-lived subscribers should arrange for unsubscription at teardown. The
+/// closure-based convenience (`subscribe(block:)`) returns a `MessageBusSubscription`
+/// that callers pass back to `unsubscribe(_:)` to end the subscription.
+///
+/// ## Threading
+///
+/// Implementations deliver messages on a serial queue. Receivers may treat state
+/// mutations across `receive(message:from:)` calls as serialised, but must not block
+/// the caller — doing so delays delivery to every other subscriber.
+///
+/// ## Choosing receiver- vs closure-based subscription
+///
+/// - `subscribe(receiver:)` — for long-lived objects that already manage their own
+///   lifecycle (Features, instrumentation components).
+/// - `subscribe(block:)` — for ad-hoc subscriptions where no natural owner exists;
+///   the returned `MessageBusSubscription` is the cleanup anchor.
+///
+/// - SeeAlso: `BusMessage`, `BusMessageReceiver`, `MessageBusSubscription`.
+public protocol MessageBus: Sendable {
+    /// Subscribes `receiver` to messages of type `Receiver.Message`.
+    ///
+    /// The bus retains `receiver` until `unsubscribe(receiver:)` is called.
+    /// Receivers are identified by object identity (`===`); subscribing the same
+    /// instance twice is treated as a single subscription.
+    func subscribe<Receiver>(receiver: Receiver) where Receiver: BusMessageReceiver
+
+    /// Removes a previously subscribed receiver.
+    ///
+    /// Identity (`===`) is used to locate the subscription; the call is a no-op if
+    /// `receiver` is not currently subscribed.
+    func unsubscribe<Receiver>(receiver: Receiver) where Receiver: BusMessageReceiver
+
+    /// Publishes `message` to every subscriber registered for `Message`.
+    ///
+    /// If no subscriber is registered for the message kind, `fallback` is invoked
+    /// instead. Do not assume which thread the fallback runs on.
+    ///
+    /// - Parameters:
+    ///   - message: The message to publish.
+    ///   - fallback: Closure invoked when the message had no subscribers.
+    func send<Message>(message: Message, else fallback: @escaping () -> Void) where Message: BusMessage
+}
+
+extension MessageBus {
+    /// Subscribes a closure to messages of type `Message`.
+    ///
+    /// A convenience over `subscribe(receiver:)` for the common case where no
+    /// natural receiver object exists. The closure is wrapped in an internal
+    /// receiver retained by the returned `MessageBusSubscription`; pass that
+    /// handle to `unsubscribe(_:)` when the subscription is no longer needed.
+    ///
+    /// ```swift
+    /// let subscription = bus.subscribe { (message: MyMessage, core) in
+    ///     // handle message
+    /// }
+    /// // ... later
+    /// bus.unsubscribe(subscription)
+    /// ```
+    ///
+    /// - Parameter block: Called on the bus's delivery queue. Must not block.
+    /// - Returns: A handle to pass to `unsubscribe(_:)`.
+    public func subscribe<Message>(block: @escaping (Message, DatadogCoreProtocol) -> Void) -> MessageBusSubscription where Message: BusMessage {
+        let receiver = BusMessageReceiverCallback(block: block)
+        self.subscribe(receiver: receiver)
+        return MessageBusSubscription(receiver)
+    }
+
+    /// Removes the subscription created by `subscribe(block:)`.
+    ///
+    /// Idempotent — additional calls with the same `subscription` are no-ops.
+    public func unsubscribe(_ subscription: MessageBusSubscription) {
+        subscription.unsubscribe(from: self)
+    }
+
+    /// Publishes `message` to every subscriber registered for `Message`.
+    ///
+    /// Convenience over `send(message:else:)` for the common case where the
+    /// caller doesn't care whether the message was consumed.
+    public func send<Message>(message: Message) where Message: BusMessage {
+        send(message: message, else: {})
+    }
+}
+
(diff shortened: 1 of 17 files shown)


DatadogInternal/Sources/Models/WebViewTracking/WebViewBusMessages.swift
```diff
@@ -0,0 +1,55 @@
+/*
+ * Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+ * This product includes software developed at Datadog (https://www.datadoghq.com/).
+ * Copyright 2019-Present Datadog, Inc.
+ */
+
+import Foundation
+
+/// Typed-bus message for a Browser SDK log event forwarded through the WebView bridge.
+public struct WebViewLogMessage: BusMessage {
```
simaoseica-dd, line 10: nit: The naming of the messages is not consistent. All these have `Message` suffix. But there are others that don't have. 
The same for the keys. Some use `.`, others `-`, others have `message` suffix.

DatadogInternal/Sources/MessageBus/MessageBus.swift
```diff
@@ -0,0 +1,198 @@
...
+    /// let subscription = bus.subscribe { (message: MyMessage, core) in
+    ///     // handle message
+    /// }
+    /// // ... later
+    /// bus.unsubscribe(subscription)
+    /// ```
+    ///
+    /// - Parameter block: Called on the bus's delivery queue. Must not block.
+    /// - Returns: A handle to pass to `unsubscribe(_:)`.
+    public func subscribe<Message>(block: @escaping (Message, DatadogCoreProtocol) -> Void) -> MessageBusSubscription where Message: BusMessage {
```
simaoseica-dd, line 81: For the entities that don't need to unsubscribe in the SDK lifecycle
```suggestion
    @discardableResult
    public func subscribe<Message>(block: @escaping (Message, DatadogCoreProtocol) -> Void) -> MessageBusSubscription where Message: BusMessage {
```

DatadogInternal/Sources/MessageBus/MessageBus.swift
```diff
@@ -0,0 +1,198 @@
...
+///
+/// Each conforming type represents a distinct kind of message. The bus routes by
+/// type — conformance is the only registration step required.
+///
+/// ## Conformance guidance
+///
+/// - Prefer immutable value types (`struct`, or `enum` for closed variants). A
+///   `BusMessage` should survive queue hops without aliasing surprises.
+/// - `key` is a stable identifier used for diagnostics and telemetry. It must be
+///   globally unique across the SDK; namespaced identifiers
```
simaoseica-dd: > namespaced identifiers

It is not the strategy in place

DatadogInternal/Sources/MessageBus/MessageBus.swift
```diff
@@ -0,0 +1,198 @@
...
+///   (e.g. `"rum.session.start"`) are recommended.
+public protocol BusMessage {
+    /// A stable, globally unique identifier for this message kind.
+    ///
+    /// Used by logs, telemetry, and any out-of-band consumer that refers to messages
+    /// by name. Treat the string as part of the public contract — changing it is a
+    /// breaking change for downstream tooling.
+    static var key: String { get }
```
simaoseica-dd, line 138: Is it a breaking change because the value is shared to telemetry?

DatadogInternal/Sources/MessageBus/MessageBus.swift
```diff
@@ -0,0 +1,198 @@
...
+/// The protocol is class-bound (`AnyObject`) because the bus tracks subscriptions
+/// by object identity. The concrete `Message` is fixed per receiver — to handle
+/// multiple message kinds, register multiple receivers.
+public protocol BusMessageReceiver: AnyObject {
+    /// The message kind this receiver consumes.
+    associatedtype Message: BusMessage
+
+    /// Handles a delivered message.
+    ///
+    /// Called on the bus's delivery queue. Implementations must not block the
```
simaoseica-dd, line 152: This re-architecture is an excellent opportunity to improve the receivers that spend extra time handling the messages. This point can improve much more the current broadcast of messages.
Some points that could be considered:
- `MessageBus` logs a warning in DEBUG if the receiver took more than a `threshold` to handle the message.
- `MessageBus` redirects the receiver message work for a previous shared queue in `subscribe<Receiver>(receiver: Receiver, on queue: DispatchQueue? = nil)`. This will explicitly open a new  possibility to async work in a new queue.

simaoseica-dd (commented):
This is a great step forward for the Message Bus 👌
The main improvement I see is moving away from broad broadcast toward explicit typed routing but I'd still like us to think around the receiver cost on the shared MessageBus queue.
Can we also have an high-level diagram of the new architecture?

Also, given the amount of deprecations introduced here, I'd prefer merging this branch into a feature branch rather than develop, otherwise the Xcode will be flooded of warnings.

DatadogInternal/Sources/MessageBus/MessageBus.swift
maxep: I would prefer to keep this explicit with `_ = subscribe { }` and avoid any leaks.

DatadogInternal/Sources/MessageBus/MessageBus.swift
maxep: This comment is wrong indeed!

DatadogInternal/Sources/MessageBus/MessageBus.swift
maxep: Those are good points and suggestions 👍 Lets keep them for a follow up iteration, because there 9 other PRs on top of this already 😄

DatadogInternal/Sources/Models/WebViewTracking/WebViewBusMessages.swift
maxep: Good catch!

simaoseica-dd (approved):
Looks good!

Issue #2867, opened 2026-04-27
NSInvalidArgumentException - -[DatadogRUM.UIScrollViewDelegateProxy scrollViewDidScroll:]: unrecognized selector sent to instance


ashman09:
### Stack trace

We are using 3.10.0 SPM.

### Reproduction steps

[com.shaadi.iphone_issue_49a18ab8433052b243527542acbf01dd_crash_session_9328e0a8d016407bad72cf5b9159bc7c_DNE_1_v2_stacktrace.txt](https://github.com/user-attachments/files/27123702/com.shaadi.iphone_issue_49a18ab8433052b243527542acbf01dd_crash_session_9328e0a8d016407bad72cf5b9159bc7c_DNE_1_v2_stacktrace.txt)

### Volume

4 events 3 users

### Affected SDK versions

3.10.0

### Latest working SDK version

3.10.0

### Does the crash manifest in the latest SDK version?

Yes

### Deployment Target

iOS 16, iPhone

gonzalezreal:
Hi @ashman09, thanks for the report and the stack trace.

This looks very similar to a `UIScrollViewDelegateProxy` crash we fixed in #2776, which was released in `3.8.3` and is included in `3.10.0`.

From the stack trace, UIKit appears to call `scrollViewDidScroll:` on our delegate proxy while a `UICollectionView` is being removed during view/controller deallocation. One possible explanation is that UIKit previously cached that the delegate responds to this selector, but during teardown the original delegate is already deallocating. In that state, the proxy may still be alive, while its weak reference to the original delegate already reads as `nil`, so the proxy can no longer forward the call.

Could you please double-check that the app build producing this crash is definitely using `3.10.0`?  We would like to rule out a stale/resolved SDK version before investigating this as a new remaining edge case.

PR #2942, opened 2026-05-21
[RUM-16361] Fix UIScrollViewDelegateProxy crash during VC teardown

mariedm:
### What and why?

Fixes a recurring `NSInvalidArgumentException` ("unrecognized selector") crash in `UIScrollViewDelegateProxy` reported in:
- [RUMS-5962](https://datadoghq.atlassian.net/browse/RUMS-5962)
- [RUMS-5941](https://datadoghq.atlassian.net/browse/RUMS-5941)
- [#2867](https://github.com/DataDog/dd-sdk-ios/issues/2867)

Crash signature: `-[DatadogRUM.UIScrollViewDelegateProxy <selector>]: unrecognized selector`, dispatched from `_notifyDidScroll` during view-controller teardown.

**Mechanism:** when a `UIViewController` owns its `UIScrollView` AND is its delegate, the two tear down together. The VC's `deallocating` flag is set BEFORE its associated objects are released. Inside that window, UIKit can dispatch a cached delegate selector to the still-alive proxy — whose weak `originalDelegate` reads as `nil`. The proxy didn't implement that selector directly, so the ObjC runtime invoked `forwardingTarget(for:)`, which returned `nil`, and raised "unrecognized selector".

**Why prior fixes missed it:**
- #2776 (proxy lifetime tied to delegate) — the proxy is still alive in this race window.
- #2791 (setter re-entrancy guard) — orthogonal code path.

### How?

Introduces a small Objective-C base class `__dd_private_DDForwardingProxyBase` in a new `DatadogRUMPrivate` SPM target (mirrors the `DatadogPrivate` pattern used by `DatadogCore`). It overrides three ObjC runtime hooks:

- `forwardingTargetForSelector:` — fast-forwards to a live target.
- `methodSignatureForSelector:` — returns the selector's real signature from known UIKit delegate protocols (`UIScrollViewDelegate`, `UITableViewDelegate`, `UICollectionViewDelegate`, `UITextViewDelegate`) via `protocol_getMethodDescription`; falls back to `v@:@` for unknown selectors.
- `forwardInvocation:` — invokes on a live target, or silently drops the invocation if the target is gone.

`UIScrollViewDelegateProxy` now inherits from the base class, overrides `forwardingTargetOrNil` to return `originalDelegate`, and drops its own `forwardingTarget(for:)` override (now in the base). `responds(to:)` is unchanged, so the RxSwift re-entrancy guard from #2761 still applies on the alive-delegate path.

This closes the entire forwarding-crash family uniformly — any selector, any UIKit delegate protocol, present or future.

This closes the forwarding-crash family for selectors in the enumerated UIKit delegate protocols (`UIScrollViewDelegate`, `UITableViewDelegate`, `UICollectionViewDelegate`, `UICollectionViewDelegateFlowLayout`, `UITextViewDelegate`) — covering all customer crashes seen so far. Selectors outside those protocols still drop without crashing, but use a generic v@:@ fallback signature; non-void returns with non-matching ABIs may have partially-defined values. Adding a new protocol is a one-line change if future selectors fall outside the current list.

**Build wiring:**
- New `DatadogRUMPrivate` SPM target at `DatadogRUM/Private/`.
- `DatadogRUM.podspec` `source_files` glob extended to include the new ObjC files.
- New umbrella header `Datadog/TargetSupport/DatadogRUM/DatadogRUM.h` to expose the ObjC class to Swift in the Xcode framework build (Carthage / XCFramework).
- `Datadog/Datadog.xcodeproj` updated: new `DatadogRUMPrivate` group, files added to `DatadogRUM` target's Compile Sources + Headers phase (Public). Some unrelated `RecordingController`/`RecordingComponents` line reorderings appear in the diff — Xcode auto-sorts these on save; they're noise, not functional changes.

**Tests** (`UIScrollViewDelegateProxyTests`, 4 new tests covering a 2×2 matrix of {alive, deallocated} × {UIScrollViewDelegate, UITableViewDelegate}):
- `..._dispatchingForwardedSelector_doesNotCrash` — UIScrollViewDelegate, deallocated. Reproduces the production crash; fails on `hotfix/3.11.1` baseline.
- `..._dispatchingUITableViewDelegateSelector_doesNotCrash` — same with a UITableViewDelegate selector (covers the RxCocoa-wrapped path documented in RUMS-5941).
- `..._dispatchingForwardedSelector_reachesTheOriginalDelegate` — alive-delegate happy path, UIScrollViewDelegate.
- `..._dispatchingUITableViewDelegateSelector_reachesTheOriginalDelegate` — alive-delegate happy path, UITableViewDelegate (exercises protocol-lookup branch).

**Testing:**
- All `UIScrollViewDelegateProxyTests` pass on iOS / tvOS / watchOS / visionOS.
- Full `DatadogRUM` module test suite passes on iOS / tvOS / watchOS.
- `UIKitExtensionsTests` shows pre-existing failures on visionOS that also fail on `hotfix/3.11.1` baseline — unrelated to this PR.

### Review checklist
- [ ] Add Objective-C interface for public APIs — N/A (new ObjC is module-internal; \`__dd_private_\` prefix matches existing convention)
- [ ] Run \`make api-surface\` when adding new APIs — N/A (no new public Swift APIs)

[RUMS-5962]: https://datadoghq.atlassian.net/browse/RUMS-5962?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RUMS-5941]: https://datadoghq.atlassian.net/browse/RUMS-5941?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- [FIX] Fix `NSInvalidArgumentException` ("unrecognized selector") crash in automatic scroll action tracking when UIKit dispatches a cached delegate method after the original `UIScrollViewDelegate` has been deallocated. See [#2942][].
+
 # 3.11.0 / 12-05-2026
 
 - [IMPROVEMENT] Add support for `maui` source for cross-platform RUM events from .NET MAUI applications. See [#2891][]
@@ -1144,6 +1146,7 @@ Release `2.0` introduces breaking changes. Follow the [Migration Guide](MIGRATIO
 [#2856]: https://github.com/DataDog/dd-sdk-ios/pull/2856
 [#2866]: https://github.com/DataDog/dd-sdk-ios/pull/2866
 [#2891]: https://github.com/DataDog/dd-sdk-ios/pull/2891
+[#2942]: https://github.com/DataDog/dd-sdk-ios/pull/2942
 
 [@00fa9a]: https://github.com/00FA9A
 [@britton-earnin]: https://github.com/Britton-Earnin
--- a/Datadog/Datadog.xcodeproj/project.pbxproj
+++ b/Datadog/Datadog.xcodeproj/project.pbxproj
@@ -401,8 +401,6 @@
 		61054E8D2A6EE10A00AAA894 /* RUMContextReceiver.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61054E3E2A6EE10A00AAA894 /* RUMContextReceiver.swift */; };
 		61054E8E2A6EE10A00AAA894 /* SRContextPublisher.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61054E3F2A6EE10A00AAA894 /* SRContextPublisher.swift */; };
 		61054E8F2A6EE10A00AAA894 /* SegmentRequestBuilder.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61054E412A6EE10A00AAA894 /* SegmentRequestBuilder.swift */; };
-		6AF71E3149694FD0972B41C5 /* RecordingController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7D0DE1D82F8B4F7FA24BD2C9 /* RecordingController.swift */; };
-		B5B633BD7EC649AC9F29F246 /* RecordingComponents.swift in Sources */ = {isa = PBXBuildFile; fileRef = 333D0D8EAED64F559835F384 /* RecordingComponents.swift */; };
 		61054E942A6EE10A00AAA894 /* TextObfuscator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61054E4A2A6EE10A00AAA894 /* TextObfuscator.swift */; };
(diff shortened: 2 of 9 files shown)


maxep (approved):

barboraplasovska (approved):
Really clean fix, thanks for handling this! 🙌🏻 
Should we update `docs/SWIZZLING.md` in this or a separate PR ? It currently covers setter-time recursion and forwarding-time re-entrancy, but not the dealloc-time forwarding race. Since `AGENTS.md` points there as required reading before any swizzle work, I think it could be worth adding a short section on this pattern.

maxep:
@barboraplasovska , I think we can rely on this current doc statement while we migrate to a new solution:
> Proxy-setter swizzles are a last-resort, legacy pattern. Do not add new ones.

Once migrated, we should update the `SWIZZLING.md` and remove the `Proxy-setter swizzle` entirely, and add a requirement to always call the `previousImplementation` with the exact same set of arguments. So we don't change the behaviour of a swizzled method.

PR #2947, opened 2026-05-26
Wire DDSketch into StatsConcentrator for latency summaries

tienquocbui:
### What and why?

The `okSummary` and `errorSummary` fields in `ExportedGroupedStats` have been empty `Data()` placeholders since [PR #2871 (StatsConcentrator)](https://github.com/DataDog/dd-sdk-ios/pull/2871), which landed the aggregation pipeline before the DDSketch implementation was ready. Now that [PR #2914 (Add DDSketch implementation for latency distributions)](https://github.com/DataDog/dd-sdk-ios/pull/2914) has merged, this PR plugs DDSketch into the aggregation path so the latency distributions are actually populated when buckets are flushed.

### How?

- `GroupedStats` now owns two `DDSketch` instances, `okSummary` and `errorSummary`, created via `DDSketch.makeForStats()` (1% relative accuracy, 2048 max bins).
- A new `update(with: SpanSnapshot)` method on `GroupedStats` absorbs each span in one place: bumps the relevant counters and routes the duration into exactly one sketch based on `isError`. This replaces the inline mutations that previously sat in `StatsConcentrator.add()`, keeping the per-span state transition local to `GroupedStats`.
- `StatsConcentrator.add()` is now a clean "find or create bucket, find or create group, hand the snapshot to the group".
- `StatsConcentrator.flush()` calls `toProtoBytes()` on each sketch when assembling `ExportedGroupedStats`, replacing the placeholders.

### Tests

Six new tests on `StatsConcentratorTests`:

- `testOkSpanPopulatesOkSummaryNotErrorSummary`: a non-error span produces an `okSummary` larger than the empty-sketch baseline and an `errorSummary` equal to it.
- `testErrorSpanPopulatesErrorSummaryNotOkSummary`: mirror case.
- `testMixedOkAndErrorSpans_populateBothSummaries`: a group that sees both span kinds populates both sketches.
- `testIneligibleSpansDoNotContributeToSummaries`: the eligibility filter still drops non-qualifying spans before any sketch update.
- `testOkSummaryBytesMatchIndependentlyBuiltSketch`: fixture-style cross-check. The emitted bytes are byte-equal to a `DDSketch` built independently from the same input, so any future regression in the wiring shows up here immediately.
- `testDistinctAggregationKeysGetIndependentSketches`: two spans with different resources produce two distinct groups, each carrying its own sketch.

The previous `testExportedGroupedStatsHasEmptySketchPlaceholders` is removed; it asserted the placeholder behavior this PR is replacing.

### Review checklist
- [x] Add CHANGELOG entry for user facing changes. N/A (internal, not user-facing)
- [x] Add Objective-C interface for public APIs. N/A (internal to DatadogTrace)
- [x] Run `make api-surface` when adding new APIs. N/A (internal to DatadogTrace)

### DIFF
--- a/DatadogTrace/Sources/Feature/StatsConcentrator.swift
+++ b/DatadogTrace/Sources/Feature/StatsConcentrator.swift
@@ -35,19 +35,48 @@ internal struct AggregationKey: Hashable, Sendable {
 
 // MARK: - Grouped Stats
 
-/// Running counters for a single aggregation key within a time bucket.
+/// Running counters and latency distributions for a single aggregation key within a
+/// time bucket.
+///
 /// Counters use `Double` to support fractional weighted sampling (weight per span).
 /// Currently weight is always 1.0, so these are effectively integers.
+///
+/// `duration` is the total nanoseconds across all spans in the group; `okSummary` and
+/// `errorSummary` are DDSketches tracking the distribution of those same durations
+/// split by `isError`. Both representations are required by the v1.2.0 stats payload.
+///
+/// Each sketch is bounded by `DDSketch.makeForStats()` to 2048 bins (max ~16 KB per
+/// sketch), so a fully saturated group caps at roughly 32 KB of sketch memory.
 internal final class GroupedStats {
     var hits: Double = 0
     var topLevelHits: Double = 0
     var errors: Double = 0
     var duration: Double = 0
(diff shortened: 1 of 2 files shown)


simaoseica-dd (approved):
Looks good!

barboraplasovska (approved):
🚀

PR #2948, opened 2026-05-27
RUM-16112 Stop overwriting fatal error context view from inactive RUM view scopes

barboraplasovska:
### What and why?

When a view became inactive (because another view started) but stayed open for a pending resource, the terminal event emitted on `stopResource()` would overwrite `fatalErrorContext.view`, causing crashes to be misattributed to the deactivated view.

### How?

Gates the `fatalErrorContext.view` assignment in `RUMViewScope` on `isActiveView`. Writes to `sessionEndedMetric`, `viewEndedMetric`, and `watchdogTermination` are outside the gate and still fire for every event. The `nil` write in `RUMSessionScope` (which drives the `rum-view-reset` sentinel) is left untouched.

### Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- [FIX] Prevent crash misattribution when an inactive RUM view emits a terminal event after `stopResource()`. See [#2948][]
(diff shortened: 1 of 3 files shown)


maxep (approved):
👍

tienquocbui (approved):
LGTM!

PR #2949, opened 2026-05-27
[RUM-16386] Add ETag conditional requests to remote config sync

saraSr5:
### What and why?

The remote config fetcher made an unconditional GET on every SDK init. This PR adds ETag-based conditional requests to avoid downloading the full config when it has not changed.

- **ETag conditional requests**: send `If-None-Match` on the next fetch; handle 304 Not Modified by keeping the existing cache untouched
- **Safety guard**: only send `If-None-Match` when the in-memory cache is usable (`.success`) — a 304 with a failed cache would leave the SDK unable to recover

Note: TTL-based revalidation is out of scope and will be addressed in a separate ticket.

### How?

- ETag stored in a separate `{id}.etag` file alongside `{id}.json`; read from disk directly in `sync()` — no in-memory property
- 304 response returns `self.cache` directly
- Config file written via `File(url:).write(data:)` with `.atomic` option (write to temp, then rename) — prevents a data-loss window
- `If-None-Match` only sent when `cache` is `.success` to prevent unrecoverable state on 304

### Review checklist
- [x] Tests added — ETag stored, `If-None-Match` sent, 304 handling, 304 with failed cache, `If-None-Match` not sent when cache is `.failure`
- [x] JIRA reference in commit and PR title
- [x] CHANGELOG — N/A, internal API
- [x] Objective-C interface — N/A, internal
- [x] `make api-surface` — N/A, no public API changes

### DIFF
--- a/DatadogCore/Sources/Core/RemoteConfigurationSynchronizer.swift
+++ b/DatadogCore/Sources/Core/RemoteConfigurationSynchronizer.swift
@@ -103,11 +103,15 @@ internal final class RemoteConfigurationSynchronizer {
                     // Store ETag for conditional requests on the next sync.
                     // If the response has no ETag, delete any stale validator so we
                     // never send If-None-Match for a different representation.
-                    let etagFile = File(url: self.directory.url.appendingPathComponent("\(self.id).etag"))
+                    let etagFileName = "\(self.id).etag"
                     if let etag = http.allHeaderFields.first(where: { ($0.key as? String)?.lowercased() == "etag" })?.value as? String {
-                        try? etagFile.write(data: Data(etag.utf8))
+                        let etagFile = self.directory.hasFile(named: etagFileName)
+                            ? (try? self.directory.file(named: etagFileName))
+                            : (try? self.directory.createFile(named: etagFileName))
+                        try? etagFile?.write(data: Data(etag.utf8))
                     } else {
-                        try? etagFile.delete()
+                        // try? silently handles the case where the file does not exist
+                        try? self.directory.file(named: etagFileName).delete()
                     }
 
                     completionHandler(.success(data))
--- a/DatadogCore/Sources/Core/RemoteConfigurationSynchronizer.swift
+++ b/DatadogCore/Sources/Core/RemoteConfigurationSynchronizer.swift
@@ -26,6 +26,8 @@ internal final class RemoteConfigurationSynchronizer {
     @ReadWriteLock
     private(set) var cache: Result<Data, Error>
 
+    private static let ttl: TimeInterval = 5 * 60
+
     init(id: String, site: DatadogSite, directory: Directory, httpClient: HTTPClient) {
         self.id = id
         self.site = site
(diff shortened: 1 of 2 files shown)


saraSr5:
@codex review

DatadogCore/Sources/Core/RemoteConfigurationSynchronizer.swift
```diff
@@ -49,31 +51,55 @@ internal final class RemoteConfigurationSynchronizer {
...
     /// - Parameter completionHandler: Called with the fetch result when the operation (and any cache write) is done.
     func sync(_ completionHandler: @escaping (Result<Data, Error>) -> Void) {
-        // TODO RUM-16386: Add ETag-based conditional requests (If-None-Match / 304) and
-        // TTL-based revalidation (skip fetch if cached config is < 5 min old).
+        // Skip fetch if cached config is less than 5 minutes old.
+        // The TTL clock resets only when new data is written (2xx response).
+        // A 304 does not update modifiedAt, so the next sync will still hit the network —
+        // this is intentional: the 5-minute window measures time since the last confirmed write.
+        if case .success = cache,
+           let date = try? directory.file(named: "\(id).json").modifiedAt(),
```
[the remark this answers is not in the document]
saraSr5: try? applied to a function returning T? is automatically flattened to T?, not T?? (Swift 5.0+)

saraSr5:
@codex review

saraSr5:
@codex review

DatadogCore/Sources/Core/RemoteConfigurationSynchronizer.swift
```diff
@@ -49,31 +51,58 @@ internal final class RemoteConfigurationSynchronizer {
...
+        httpClient.fetch(request: request) { result in
             switch result {
             case .failure(let error):
                 completionHandler(.failure(error))
 
             case .success(let (http, data)):
-                // 1. Non-2xx HTTP status
+                // 1. Not Modified — existing cache is still valid
+                if http.statusCode == 304 {
+                    completionHandler(self.cache)
```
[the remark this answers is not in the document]
saraSr5: This is related to having foreground refresh case

saraSr5:
@codex review

saraSr5:
@codex review

DatadogCore/Sources/Core/RemoteConfigurationSynchronizer.swift
```diff
@@ -82,11 +94,22 @@ internal final class RemoteConfigurationSynchronizer {
...
-                    let file = try self.directory.createFile(named: "\(self.id).json")
-                    try file.write(data: data)
+                    try File(url: self.directory.url.appendingPathComponent("\(self.id).json")).write(data: data)
                     self.cache = .success(data)
+
+                    // Store ETag for conditional requests on the next sync.
+                    // If the response has no ETag, delete any stale validator so we
+                    // never send If-None-Match for a different representation.
+                    let etagFile = File(url: self.directory.url.appendingPathComponent("\(self.id).etag"))
+                    if let etag = http.allHeaderFields.first(where: { ($0.key as? String)?.lowercased() == "etag" })?.value as? String {
```
barboraplasovska, line 107: You don't need to iterate `allHeaderFields` manually, `HTTPURLResponse.value(forHTTPHeaderField:)` already does a case-insensitive lookup.

DatadogCore/Sources/Core/RemoteConfigurationSynchronizer.swift
```diff
@@ -82,11 +94,22 @@ internal final class RemoteConfigurationSynchronizer {
...
                 do {
-                    let file = try self.directory.createFile(named: "\(self.id).json")
-                    try file.write(data: data)
+                    try File(url: self.directory.url.appendingPathComponent("\(self.id).json")).write(data: data)
                     self.cache = .success(data)
+
+                    // Store ETag for conditional requests on the next sync.
+                    // If the response has no ETag, delete any stale validator so we
+                    // never send If-None-Match for a different representation.
+                    let etagFile = File(url: self.directory.url.appendingPathComponent("\(self.id).etag"))
```
barboraplasovska: You can use `directory.file(named:)` for both the read and write, since that's already used for the read above. ([line](https://github.com/DataDog/dd-sdk-ios/pull/2949/changes#diff-e480ed31ddb56dc78623dbf4b3b88e62f40b874327276142eddda08beaf873b2R65))

DatadogCore/Tests/Datadog/Core/RemoteConfigurationTests.swift
```diff
@@ -270,6 +270,146 @@ class RemoteConfigurationTests: XCTestCase {
...
+        wait(for: [expectation], timeout: 2)
+
+        XCTAssertEqual(try? rc.cache.get(), existing, "Cache must be unchanged after 304")
+    }
+
+    func test304ResponseCallsCompletionEvenWhenCacheIsFailure() throws {
+        // Given — unreadable .json so cache is .failure
+        let fileURL = coreDir.coreDirectory.url.appendingPathComponent("test-id.json")
+        try Data("{\"v\":1}".utf8).write(to: fileURL, options: .atomic)
+        try FileManager.default.setAttributes([.posixPermissions: 0o000], ofItemAtPath: fileURL.path)
```
barboraplasovska: Setting permissions to `0o000` won't make the file unreadable when tests run as root on CI simulators. To get a `.failure` cache, just don't write the `.json` file at all, like `testSyncDoesNotSendIfNoneMatchWhenCacheIsFailure` does.

barboraplasovska (commented):
Great job!

DatadogCore/Sources/Core/RemoteConfigurationSynchronizer.swift
saraSr5: Hii! HTTPURLResponse.value(forHTTPHeaderField:) requires iOS 13.0+, but the SDK declares .iOS(.v12) in Package.swift, so this fails to compile:
image


The manual allHeaderFields iteration with .lowercased() is the iOS 12 compatible alternative :(
If the minimum deployment  target is bumped to iOS 13, this can be simplified :)

DatadogCore/Sources/Core/RemoteConfigurationSynchronizer.swift
saraSr5: Good point! I have just modify it using hasFile + file/createFile because directory.file(named:) throws when the file doesn't exist yet (first sync). Happy to simplify further if you have a better approach in mind!

barboraplasovska (approved):

PR #2954, opened 2026-05-28
[PANA-7550] Add layer tree occlusion culling to Session Replay

gonzalezreal:
### What and why?

This is the next PR in the series integrating the Core Animation recording pipeline.

It adds an occlusion culling step that removes hidden layers from each captured snapshot before it reaches wireframe generation.

### How?

The PR adds `OcclusionMap`, a tile bitmap over the viewport. Insert marks only tiles fully inside a rect. Query requires every overlapping tile to be set. This asymmetry is what makes the map safe: a layer can only be reported as covered when something in front of it completely fills each tile it touches.

It extends `CALayerSnapshot` with `removingOccluded()`, a single recursive pass that sorts siblings by `zPosition`, walks the tree front-to-back, and drops any layer whose visible frame is fully covered. Structural nodes with no content and no surviving children are also dropped. Rounded-corner layers contribute band rects instead of their full frame. `preservesSublayerOcclusion` propagates an ancestry flag down the tree so descendants of semi-transparent, masked, rotated, or filtered ancestors cannot contribute opaque coverage.

It wires `removingOccluded()` into `LayerRecorder.record`, which stops recording early if nothing survives culling.

### Review checklist


DatadogSessionReplay/Sources/LayerTreeRecorder/CALayerSnapshot+Occlusion.swift
```diff
@@ -0,0 +1,193 @@
...
+            }
+        }
+
+        if visibleLayers.isEmpty {
+            if drawsContent {
+                if !hasVisibleShadow && occlusionMap.isCovered(visibleFrame) {
+                    return nil
+                }
+            } else {
+                return nil
```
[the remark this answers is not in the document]
gonzalezreal: I agree we should not rely on contents for semantic leaves. Even if many UIKit-backed layers expose content today, `ignoreSubtree` means the snapshot owns its representation through semantics / privacy rather than descendants. I’ll treat ignored subtree snapshots as content so the structural container cull only removes nodes that are truly structural.

DatadogSessionReplay/Sources/LayerTreeRecorder/CALayerSnapshot+Occlusion.swift
```diff
@@ -0,0 +1,193 @@
...
+    var drawsContent: Bool {
+        observation.ignoreSubtree || contentsClass != nil || hasBackgroundColor || hasBorder || hasVisibleShadow
+    }
+
+    /// A Boolean value indicating whether the layer paints its frame as an opaque rectangle.
+    var isOccluder: Bool {
+        opacity == 1
+            && backgroundColor?.alpha == 1
+            && mask == nil
+            && transform.isAxisAligned
```
sethfowler-datadog, line 23: Technically, this is conservative (the non-AA transform will still enclose an AA region that we could treat as occluded) but in practice I'd guess being more precise here would not have a big impact.

DatadogSessionReplay/Sources/LayerTreeRecorder/CALayerSnapshot+Occlusion.swift
```diff
@@ -0,0 +1,193 @@
...
+    func occlusionRects(in frame: CGRect) -> [CGRect] {
+        if self == .zero {
+            return [frame]
+        }
+
+        var rects: [CGRect] = []
+
+        let topInset = max(topLeft.height, topRight.height)
+        let bottomInset = max(bottomLeft.height, bottomRight.height)
+        let horizontalHeight = frame.height - topInset - bottomInset
```
sethfowler-datadog: This is a case where the CSS box model actually gives us nice terminology -- the distinction between the "content box" and the "border box" is nice. If you wanted to use that terminology by analogy here, you could describe this as `contentHeight`.

DatadogSessionReplay/Sources/LayerTreeRecorder/CALayerSnapshot+Occlusion.swift
```diff
@@ -0,0 +1,193 @@
...
+                    y: frame.minY + topInset,
+                    width: frame.width,
+                    height: horizontalHeight
+                )
+            )
+        }
+
+        let leftInset = max(topLeft.width, bottomLeft.width)
+        let rightInset = max(topRight.width, bottomRight.width)
+        let verticalWidth = frame.width - leftInset - rightInset
```
sethfowler-datadog: Similarly, this could be `contentWidth` if we adopted the CSS box model terminology.

DatadogSessionReplay/Sources/LayerTreeRecorder/OcclusionMap.swift
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        return TileRange(
+            columnRange: max(0, minColumn)...min(columns - 1, maxColumn),
+            rowRange: max(0, minRow)...min(rows - 1, maxRow)
+        )
+    }
+
+    /// Tiles overlapping `rect`, clipped to the canvas.
+    private func queryTileRange(for rect: CGRect) -> TileRange? {
```
sethfowler-datadog, line 144: Is it possible to share more code between `insertTileRange(for:)` and `queryTileRange(for:)`? It seems like they are totally identical except for the rounding direction, so maybe there could be a helper both call which computes an `UnsnappedTileRange` which would use `ClosedRange<Float>` for the ranges, and then you could round to produce a final `TileRange`.

DatadogSessionReplay/Sources/LayerTreeRecorder/OcclusionMap.swift
```diff
@@ -0,0 +1,176 @@
...
+                if words[row * stride + word] & mask != mask {
+                    return false
+                }
+            }
+        }
+        return true
+    }
+
+    /// Yields each `(word, mask)` pair covering the given tile column range.
+    private struct WordSpan: Sequence {
```
sethfowler-datadog, line 68: If `CFMutableBitVector` had a nicer API, I might suggest using it, but I'm not sure it'd be a win overall.

sethfowler-datadog (dismissed):
Nice one! 👍

DatadogSessionReplay/Sources/LayerTreeRecorder/CALayerSnapshot+Occlusion.swift
gonzalezreal: Agreed. This is intentionally conservative for the first pass. Computing the inscribed AA region for transformed layers could improve culling, but I’d rather keep the geometry simple until data shows this matters.

DatadogSessionReplay/Sources/LayerTreeRecorder/CALayerSnapshot+Occlusion.swift
gonzalezreal: Yep, that's much better naming. Done in 4850b2f1ea047d76e3dcbdc7832134c2207d4967

DatadogSessionReplay/Sources/LayerTreeRecorder/OcclusionMap.swift
gonzalezreal: It is possible, but I’d keep these two methods separate for now. The duplicated setup is small, while the different rounding rules are the important part. A shared unsnapped range helper would save a few lines but make that distinction less direct.

DatadogSessionReplay/Sources/LayerTreeRecorder/OcclusionMap.swift
gonzalezreal: I don’t think `CFMutableBitVector` improves this. The local `UInt64` storage keeps the hot path explicit and avoids bridging and API overhead.

sethfowler-datadog: The advantage of using a bit vector implementation, in general, would be that you could directly address ranges of bits and wouldn't need to handle the chunking into `UInt64`-sized ranges yourself, but yeah, `CFMutableBitVector` just doesn't seem like a good choice given its limited API and the requirement to bridge to Objective C.

sethfowler-datadog (approved):

DatadogSessionReplay/Sources/LayerTreeRecorder/OcclusionMap.swift
sethfowler-datadog: I kinda perceive it the other way -- factoring out the common code leaves each of the top-level functions containing only the part of the implementation that's different, so the differences are highlighted a bit more. However, I can totally see the argument in the other direction; keeping them separate is fine by me!

PR #2955, opened 2026-05-29
RUM-4288 Add instanceName overloads to Obj-C API surface

barboraplasovska:
### What and why?

Adds `instanceName:` parameter overloads to the Objective-C API surface so Obj-C and KMP iOS consumers can initialize and operate on a non-default SDK core instance, achieving parity with the Android SDK and the existing Swift API.

Affected classes: `DDDatadog`, `DDLogs`, `DDLogger`, `DDRUM`, `DDRUMMonitor`, `DDTrace`, `DDTracer`, `DDSessionReplay`, `DDWebViewTracking`, `DDURLSessionInstrumentation`.

### How?

Each new overload accepts a trailing `instanceName: String` parameter, resolves the named core via `CoreRegistry.instance(named:)` and forwards to the same Swift API the existing zero-arg method already calls.

`WebViewTracking.disable(webView:)` and `URLSessionInstrumentation.disable(delegateClass:)` were intentionally left without `instanceName` overloads since their Swift counterparts do not accept `in core:`.

Note: `api-surface-swift` also removes the stale `in core:` entry for `WebViewTracking.disable` — this reflects a pre-existing Swift source change from a prior PR and is correctly cleaned up here.

### Review checklist

### DIFF
--- a/DatadogCore/Sources/Datadog+objc.swift
+++ b/DatadogCore/Sources/Datadog+objc.swift
@@ -79,80 +79,80 @@ public final class objc_Datadog: NSObject {
         Datadog.setUserInfo(id: userId, name: name, email: email, extraInfo: extraInfo.dd.swiftAttributes)
     }
 
-    public static func setUserInfo(userId: String, instanceName: String, name: String? = nil, email: String? = nil, extraInfo: [String: Any] = [:]) {
-        Datadog.setUserInfo(id: userId, name: name, email: email, extraInfo: extraInfo.dd.swiftAttributes, in: CoreRegistry.instance(named: instanceName))
+    public static func setUserInfo(userId: String, instanceName: String?, name: String? = nil, email: String? = nil, extraInfo: [String: Any] = [:]) {
+        Datadog.setUserInfo(id: userId, name: name, email: email, extraInfo: extraInfo.dd.swiftAttributes, in: CoreRegistry.instance(named: instanceName ?? CoreRegistry.defaultInstanceName))
     }
 
     public static func clearUserInfo() {
         Datadog.clearUserInfo()
     }
 
-    public static func clearUserInfo(instanceName: String) {
-        Datadog.clearUserInfo(in: CoreRegistry.instance(named: instanceName))
+    public static func clearUserInfo(instanceName: String?) {
+        Datadog.clearUserInfo(in: CoreRegistry.instance(named: instanceName ?? CoreRegistry.defaultInstanceName))
     }
 
     public static func addUserExtraInfo(_ extraInfo: [String: Any]) {
         Datadog.addUserExtraInfo(extraInfo.dd.swiftAttributes)
     }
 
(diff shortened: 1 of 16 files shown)


DatadogCore/Sources/Datadog+objc.swift
```diff
@@ -67,45 +79,89 @@ public final class objc_Datadog: NSObject {
         Datadog.setUserInfo(id: userId, name: name, email: email, extraInfo: extraInfo.dd.swiftAttributes)
     }
 
+    public static func setUserInfo(userId: String, instanceName: String, name: String? = nil, email: String? = nil, extraInfo: [String: Any] = [:]) {
+        Datadog.setUserInfo(id: userId, name: name, email: email, extraInfo: extraInfo.dd.swiftAttributes, in: CoreRegistry.instance(named: instanceName))
+    }
```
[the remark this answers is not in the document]
barboraplasovska: This won't pass the linter checks as parameter with defaults should be located at the end of the list

DatadogCore/Sources/Datadog+objc.swift
maxep, line 1: ~~cr/ All of the public methods here should have an optional `instanceName` instead, otherwise it is a breaking change of the objc api.~~

maxep (changes requested):
This is a great improvement 👍  ~~The only blocker is the non-optional `instanceName`, which is a breaking change we can avoid.~~

DatadogCore/Sources/Datadog+objc.swift
maxep: Nevermind, these are additions, not changes 👍  sorry!

maxep (dismissed):

DatadogWebViewTracking/Sources/ObjC/WebViewTracking+objc.swift
```diff
@@ -59,6 +59,32 @@ public final class objc_WebViewTracking: NSObject {
...
+    /// `hosts`, web events will be correlated with the RUM session from native SDK.
+    ///
+    /// - Parameters:
+    ///   - webView: The web-view to track.
+    ///   - instanceName: The name of the SDK instance to use for tracking.
+    ///   - hosts: A set of hosts instrumented with Browser SDK to capture Datadog events from.
+    ///   - logsSampleRate: The sampling rate for logs coming from the WebView. Must be a value between `0` and `100`,
+    ///   where 0 means no logs will be sent and 100 means all will be uploaded. Default: `100`.
+    @objc
+    public static func enable(
```
gonzalezreal, line 98: Could we also add the `instanceName` overload for the `hostPatterns` variant?
Swift already supports `WebViewTracking.enable(webView:hostPatterns:logsSampleRate:in:)`, but Obj-C/KMP callers using wildcard host patterns still only have the default-core API.
That would keep it in parity with the new hosts overload.

gonzalezreal (approved):
🥇 LGTM!

maxep (approved):

PR #2958, opened 2026-05-29
Apply EventMapper before SpanSnapshot creation

tienquocbui:
### What and why?

The `SpanSnapshot` used for client-side stats was being built **before** the user-configured `SpanEventMapper` ran, while trace uploads carried the **post-mapper** values. This is a correctness gap: a mapper that rewrites resource names (for example, to redact PII, normalize URL paths, or reduce cardinality) would key stats off the pre-mapper values while traces carry the post-mapper values. The result is dashboard inconsistency (stats and traces disagree on `resource.name` for the same span) and potential cardinality explosions in the stats pipeline whenever a mapper merges many request resources into one normalized name.

### How?

Refactored `DDSpan.finish()` to gate on whether the stats hook is registered:

- **Stats enabled** (`ddTracer.onSpanFinished != nil`): build the `SpanEvent` first (which runs the `SpanEventMapper`) and derive the `SpanSnapshot` from it. The same event is reused for the upload when the span is sampled in, so the mapper still runs **at most once per finished span**.
- **Stats disabled**: original flow is preserved. The mapper only runs on the sampled-in upload path, so customers who do not opt into stats see no change in per-span cost.

The snapshot derivation moves into a new `SpanSnapshot.init(from: SpanEvent)` extension that reads every field from the post-mapper event (and its `tags`). This is a single source of truth and replaces the parallel field-extraction logic that previously lived in `DDSpan.createSnapshot(finishTime:)` (now removed). The snapshot type was introduced in [PR #2853 (Add SpanSnapshot type and span finish notification hook)](https://github.com/DataDog/dd-sdk-ios/pull/2853).

### Tests

**`SpanSnapshotTests`** (6 new tests covering mapper consistency):

- `testSnapshotReflectsMappedResourceName`: mapper rewrites `resource`, snapshot carries the new resource.
- `testSnapshotReflectsMappedOperationName`: mapper rewrites `operationName`, snapshot reflects it.
- `testSnapshotReflectsMappedPeerTags`: mapper rewrites `peer.service`, snapshot's `peerTags` reflects the new value.
- `testSnapshotReflectsMapperRemovingPeerTag`: mapper removes a peer tag, snapshot's `peerTags` no longer contains it.
- `testSnapshotReflectsMappedHTTPStatusCode`: mapper rewrites the status tag, snapshot's `httpStatusCode` reflects it.
- `testSnapshotMatchesUploadedSpanEvent_whenMapperRewritesResource`: end-to-end invariant. The snapshot the stats hook receives and the `SpanEvent` written to the upload agree on `resource`, `operationName`, `service`, and `isError`.

**`DDSpanTests`** (4 new tests covering mapper invocation count and sampling interplay):

- `testWhenStatsEnabled_eventMapperRunsExactlyOnce_forSampledInSpan`: with stats on, a sampled-in span invokes the mapper exactly once.
- `testWhenStatsEnabled_eventMapperRunsExactlyOnce_forSampledOutSpan`: with stats on, a sampled-out span still invokes the mapper once (for the snapshot) and produces no upload.
- `testWhenStatsDisabled_eventMapperOnlyRunsOnUploadPath_forSampledInSpan`: with stats off, behavior unchanged.
- `testWhenStatsDisabled_eventMapperDoesNotRun_forSampledOutSpan`: with stats off, sampled-out spans skip the mapper.

### Review checklist
- [x] Add CHANGELOG entry for user facing changes. N/A (internal, not user-facing)
- [x] Add Objective-C interface for public APIs. N/A (internal to DatadogTrace)
- [x] Run `make api-surface` when adding new APIs. N/A (internal to DatadogTrace)

### DIFF
--- a/DatadogTrace/Sources/Span/SpanSnapshot.swift
+++ b/DatadogTrace/Sources/Span/SpanSnapshot.swift
@@ -14,7 +14,25 @@ internal typealias Nanoseconds = UInt64
 ///
 /// Created in `DDSpan.finish()` **before** the sampling decision, so that all spans
 /// (including sampled-out ones) contribute to accurate aggregate metrics.
+///
+/// The snapshot is derived from a post-`SpanEventMapper` `SpanEvent`, which guarantees
+/// that stats and trace uploads agree on resource name, service name, tags, and the
+/// `isError` flag. Without this, a customer-supplied mapper that rewrites resource
+/// names (e.g. to redact PII or normalize cardinality) would key stats off the
+/// pre-mapper values while traces carry post-mapper values, leading to dashboard
+/// inconsistencies and potential cardinality explosions in the stats pipeline.
 internal struct SpanSnapshot: Encodable, Sendable {
+    /// Span tag keys that participate in peer-service aggregation. Mirrors the
+    /// `peer_tags_aggregation` set used by the Datadog Agent.
+    static let peerTagKeys: [String] = [
+        "peer.service",
+        "db.instance",
+        "db.system",
+        "out.host",
+        "net.peer.name",
+        "server.address"
+    ]
+
     let traceID: TraceID
     let spanID: SpanID
     let parentSpanID: SpanID?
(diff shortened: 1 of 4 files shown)


DatadogTrace/Sources/Span/SpanSnapshot.swift
```diff
@@ -37,3 +55,50 @@ internal struct SpanSnapshot: Encodable, Sendable {
...
+        let spanKind = event.tags[SpanTags.kind]
+        let httpStatusCode: UInt32 = {
+            guard let raw = event.tags[OTTags.httpStatusCode],
+                  let intValue = Int(raw),
+                  let uint32 = UInt32(exactly: intValue) else {
+                return 0
+            }
+            return uint32
+        }()
+        let isTopLevel = event.parentID == nil || event.tags["_dd.top_level"] == "1"
```
arroz, line 89: Can it ever be a top level event if `parentID` is not `nil`?

DatadogTrace/Sources/Span/SpanSnapshot.swift
```diff
@@ -59,9 +59,18 @@ internal struct SpanSnapshot: Encodable, Sendable {
...
     ///
-    /// All fields are read from the event (and its `tags`), so any mutation applied by
-    /// the user-configured `SpanEventMapper` is reflected in the stats aggregation.
-    init(from event: SpanEvent) {
+    /// All field values that the `SpanEventMapper` may mutate (resource, service,
+    /// operation name, tags, isError) are read from the event so that any mutation
+    /// applied by the user-configured mapper is reflected in stats aggregation.
+    ///
+    /// - Parameter event: the post-mapper `SpanEvent`.
+    /// - Parameter startTime: the raw device-local span start time. `SpanEvent.startTime`
```
arroz: Very minor suggestion: keep `Parameter` short (just the first sentence) and move the rest to a `Note` or `Remark` or some other appropriate markdown entry.

arroz (approved):
Looks good, just left a very minor thing.

arroz (approved):
🚀

PR #2956, opened 2026-05-29
chore: Align public RUM session IDs with event formatting

simaoseica-dd:
> [!NOTE]
> This PR follows up on the contribution from @noremac in https://github.com/DataDog/dd-sdk-ios/pull/2953.

### What and why?

This PR aligns RUM session IDs exposed by iOS with the canonical RUM data format used in emitted events and core context.

Previously, some public and internal surfaces were building UUID strings directly, while RUM events used the normalized `toRUMDataFormat` representation. That could lead to mismatches when comparing IDs across those paths. This change makes them consistent.

### How?

This PR updates the RUM Monitor and session update flow to use `toRUMDataFormat` when constructing:
- `currentSessionID`
- `onSessionStart`
- `RUMCoreContext`

### Review checklist
- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] Make sure each commit and the PR mention the Issue number or JIRA reference
- [x] Add CHANGELOG entry for user facing changes
- [ ] Add Objective-C interface for public APIs - see our [guidelines](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3157787243/RFC+-+Modular+Objective-C+Interface#Recommended-solution) (internal) 
- [x] Run `make api-surface` when adding new APIs

PR #2963, opened 2026-06-01
[RUM-16633] Add wildcard host pattern matching to WebView tracking

barboraplasovska:
### What and why?

Adds a new `WebViewTracking.enable(webView:hostPatterns:[String])` overload that accepts wildcard pattern strings like `"*.example.com"` or `"preview-*.shopist.io"`. Customers currently cannot configure WebView tracking for dynamic hostnames (ephemeral preview URLs, multi-tenant subdomain fleets) — the existing `hosts:` API only supports exact matches and subdomain-suffix matching.

This is the iOS side of [RUM-15826](https://datadoghq.atlassian.net/browse/RUM-15826). Browser SDK counterpart: https://github.com/DataDog/browser-sdk/pull/4703

### How?

A new optional `getAllowedWebViewHostPatterns?()` method is injected into the bridge alongside the existing `getAllowedWebViewHosts()`, which keeps its current shape forever. Browser SDK detects the new method via `typeof === 'function'` and evaluates patterns using plain string operations — no regex engine involved on any platform.

Patterns with more than one `*` are dropped with a console warning. Patterns are lowercased. The existing `enable(webView:hosts:)` path is byte-for-byte unchanged.

### Review checklist

[RUM-15826]: https://datadoghq.atlassian.net/browse/RUM-15826?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 - [FIX] Expose RUM operation options to Objective-C from `DatadogRUM`. See [#2969][]
 - [FEATURE] Instrumented Web Views now have their tracing decision consistent with the native SDK. See [#2859][]
 - [IMPROVEMENT] Align public RUM session IDs with event formatting. See [#2956][]
-- [FEATURE] Add wildcard host pattern support to WebView tracking via `WebViewTracking.enable(webView:hostPatterns:)`. Patterns like `"*.example.com"` and `"preview-*.shopist.io"` are matched in Browser SDK using plain string operations. See [#2959][]
+- [FEATURE] Add wildcard host pattern support to WebView tracking via `WebViewTracking.enable(webView:hostPatterns:)`. See [#2963][]
 
 # 3.11.1 / 28-05-2026
 
@@ -1173,6 +1173,7 @@ Release `2.0` introduces breaking changes. Follow the [Migration Guide](MIGRATIO
 [#2952]: https://github.com/DataDog/dd-sdk-ios/pull/2952
 [#2968]: https://github.com/DataDog/dd-sdk-ios/pull/2968
 [#2969]: https://github.com/DataDog/dd-sdk-ios/pull/2969
+[#2963]: https://github.com/DataDog/dd-sdk-ios/pull/2963
 
 [@00fa9a]: https://github.com/00FA9A
 [@britton-earnin]: https://github.com/Britton-Earnin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@
 - [FIX] Prevent crash misattribution when an inactive RUM view emits a terminal event after `stopResource()`. See [#2948][]
 - [FIX] Fix wrong types in the `objc_LogEventDevice` properties definition. See [#2966][]
 - [FIX] Expose RUM operation options to Objective-C from `DatadogRUM`. See [#2969][]
+- [FEATURE] Instrumented Web Views now have their tracing decision consistent with the native SDK. See [#2859][]
+- [IMPROVEMENT] Align public RUM session IDs with event formatting. See [#2956][]
+- [FEATURE] Add wildcard host pattern support to WebView tracking via `WebViewTracking.enable(webView:hostPatterns:)`. Patterns like `"*.example.com"` and `"preview-*.shopist.io"` are matched in Browser SDK using plain string operations. See [#2959][]
 
 # 3.11.1 / 28-05-2026
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- [FEATURE] Add wildcard host pattern support to WebView tracking via `WebViewTracking.enable(webView:hostPatterns:)`. See [#2963][]
+
 # 3.12.0 / 04-06-2026
 
 - [FEATURE] Instrumented Web Views now have their tracing decision consistent with the native SDK. See [#2859][]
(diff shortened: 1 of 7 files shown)


DatadogWebViewTracking/Sources/WebViewTracking.swift
```diff
@@ -133,6 +168,61 @@ public enum WebViewTracking {
...
+            emitter: MessageEmitter(
+                logsSampler: Sampler(samplingRate: logsSampleRate),
+                core: core
+            )
+        )
+
+        controller.removeScriptMessageHandler(forName: bridgeName)
+        controller.add(messageHandler, name: bridgeName)
+
+        let validPatterns: [String] = hostPatterns.compactMap { pattern in
```
gonzalezreal, line 160: Could we validate or safely encode these patterns before putting them into the injected JavaScript?
Right now a value with a quote, like `foo'bar.com`, can break the script.

CHANGELOG.md
```diff
@@ -1,13 +1,10 @@
...
 
-- [FIX] Prevent a crash from `VitalCPUReader` when the CPU tick counter rolls over. See [#2968][]
 - [FIX] Prevent crash misattribution when an inactive RUM view emits a terminal event after `stopResource()`. See [#2948][]
 - [FIX] Fix wrong types in the `objc_LogEventDevice` properties definition. See [#2966][]
-- [IMPROVEMENT] Add `logger` case to `RUMErrorSource` (Swift) and `DDRUMErrorSource` (Obj-C) for cross-platform parity. See [#2949][]
-- [FIX] Add `logger` case to `RUMErrorSource` (Swift) and `DDRUMErrorSource` (Obj-C) for cross-platform parity. See [#2952][]
-- [FIX] Expose RUM operation options to Objective-C from `DatadogRUM`. See [#2969][]
 - [FEATURE] Instrumented Web Views now have their tracing decision consistent with the native SDK. See [#2859][]
 - [IMPROVEMENT] Align public RUM session IDs with event formatting. See [#2956][]
+- [FEATURE] Add wildcard host pattern support to WebView tracking via `WebViewTracking.enable(webView:hostPatterns:)`. Patterns like `"*.example.com"` and `"preview-*.shopist.io"` are matched in Browser SDK using plain string operations. See [#2959][]
```
gonzalezreal: 🤏🏽 This points to `#2959`, but this PR is `#2963`. Could you update the link and add the matching reference at the bottom of the file?

gonzalezreal (commented):
Could you also update `api-surface-objc` for the new Obj-C overload?

api-surface-swift has `enable(webView:hostPatterns:logsSampleRate:in:)`, but `api-surface-objc` still only has the old `hosts:` overload.

gonzalezreal (dismissed):

DatadogWebViewTracking/Sources/WebViewTracking.swift
```diff
@@ -133,6 +168,67 @@ public enum WebViewTracking {
...
+        guard !(core is NOPDatadogCore) else {
+            throw ProgrammerError(
+                description: "Datadog SDK must be initialized before calling `WebViewTracking.enable(webView:)`."
+            )
+        }
+
+        let controller = webView.configuration.userContentController
+        let isTracking = controller.userScripts.contains { $0.source.starts(with: Self.jsCodePrefix) }
+        guard !isTracking else {
+            DD.logger.warn("`startTrackingDatadogEvents(core:hosts:)` was called more than once for the same WebView. Second call will be ignored. Make sure you call it only once.")
```
kikoveiga: I think this is using an old API name `startTrackingDatadogEvents`, could you update to the new `WebViewTracking.enable` please? Same in L134.

DatadogWebViewTracking/Sources/WebViewTracking.swift
```diff
@@ -133,6 +168,67 @@ public enum WebViewTracking {
         core.telemetry.usage(event: .trackWebView)
     }
 
+    @MainActor
+    static func enableOrThrow(
```
kikoveiga, line 150: Most of the code in this function is the same as the existing `enableOrThrow`, except for the validation and serialization step. Could we extract and reuse some code?

DatadogWebViewTracking/Sources/WebViewTracking.swift
```diff
@@ -133,6 +168,67 @@ public enum WebViewTracking {
...
+            .joined(separator: ",")
+
+        let elements = WebViewTrackingElements(allowedWebViewHostsString: allowedWebViewHostsString)
+        let isTraceSampled = WebViewTracking.isTraceSampledStringValue(for: core)
+
+        try WebViewSessionRolloverHandler.register(webView: webView, in: core, using: elements)
+
+        injectUserScript(on: webView, in: core, using: elements, isTraceSampled: isTraceSampled)
+
+        core.telemetry.usage(event: .trackWebView)
```
kikoveiga, line 186: We can't distinguish the usage of the new vs old API. Should we think about adding a flag `usesPatterns` to the event?

DatadogWebViewTracking/Tests/WebViewTrackingTests.swift
kikoveiga, line 1: Could you add more tests to make sure serialization handles them well, such as `*`, `"" (empty-string)`, `https://` and mixed inputs (plain + patterns) please?

DatadogWebViewTracking/Sources/WebViewTracking.swift
barboraplasovska: Good idea! We can do this in a follow-up PR since it required a schema change.

gonzalezreal (dismissed):
LGTM!
One small changelog issue remains: this PR is listed under 3.12.0, which is already released, and it also duplicates the #2859 and #2956 entries.

Could you move the #2963 entry to Unreleased and remove the duplicate lines

gonzalezreal (approved):

DatadogWebViewTracking/Sources/WebViewTracking.swift
```diff
@@ -87,6 +122,77 @@ public enum WebViewTracking {
...
+            if wildcardCount > 1 {
+                DD.logger.warn("WebView host pattern \"\(pattern)\" contains more than one wildcard and will be ignored.")
+                return nil
+            }
+            return lowercased
+        }
+
+        let allowedWebViewHostsString = validPatterns
+            .map { "\"\($0)\"" }
+            .joined(separator: ",")
```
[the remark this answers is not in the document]
kikoveiga: This is not correct, there's no `getAllowedWebViewHostPatterns()` in the browser-sdk PR.

kikoveiga (approved):

PR #2964, opened 2026-06-01
fix: Account for cpu_ticks rollover.

fuzzybinary:
### What and why?

VitalCPUReader assumes that the value returned from `cpu_ticks.0` is always increasing, which is not true, as it is a 32-bit integer of the number of CPU Ticks used by User applications since device boot. For a device that has been online for a while, or has seen in heavy use by user applications, this can cause the value to overflow.

In these cases, the VitalCPUReader could crash with an overflow exception when it attempted to subtract out the inactive ticks from the active ones.

### How?

To avoid the overflow and to simplify the logic I've removed tracking "inactive" ticks and changed it to tracking "active" ticks  VitalCPUReader will always return an increasing number of ticks since the first `appDidBecomeActive` notification from NotificationCenter.

### Review checklist

### DIFF
--- a/DatadogRUM/Sources/RUMVitals/VitalCPUReader.swift
+++ b/DatadogRUM/Sources/RUMVitals/VitalCPUReader.swift
@@ -24,6 +24,7 @@ internal class VitalCPUReader: SamplingBasedVitalReader {
         self.telemetry = telemetry
         notificationCenter.addObserver(self, selector: #selector(appWillResignActive), name: ApplicationNotifications.willResignActive, object: nil)
         notificationCenter.addObserver(self, selector: #selector(appDidBecomeActive), name: ApplicationNotifications.didBecomeActive, object: nil)
+        lastReadActiveTicks = readUtilizedTicks()
     }
 
     func readVitalData() -> Double? {
--- a/DatadogRUM/Sources/RUMVitals/VitalCPUReader.swift
+++ b/DatadogRUM/Sources/RUMVitals/VitalCPUReader.swift
@@ -11,8 +11,9 @@ import DatadogInternal
 internal class VitalCPUReader: SamplingBasedVitalReader {
     /// host_cpu_load_info_count is 4 (tested in iOS 14.4)
     private static let host_cpu_load_info_count = MemoryLayout<host_cpu_load_info>.stride / MemoryLayout<integer_t>.stride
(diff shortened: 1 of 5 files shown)


fuzzybinary:
After attempting to deal with having the lastReadActiveTicks only be initialized when the application is Active, I've defaulted to what the previous behavior was, which is to track CPU ticks whether the app is launched in the foreground or in the background.

There is a potential improvement here to not track this vital when the application is launched in the background but I think it is out of the scope of this change.

DatadogRUM/Sources/RUMVitals/VitalCPUReader.swift
```diff
@@ -23,34 +24,56 @@ internal class VitalCPUReader: SamplingBasedVitalReader {
...
-            return Double(ticks - inactiveTicks)
+        addActiveTicks()
+        if let totalActiveTicks {
+            return Double(totalActiveTicks)
         }
         return nil
     }
 
+    private func addActiveTicks() {
+        // If lastReadActiveTicks isn't set, we're currently not the active application.
```
simaoseica-dd, line 39: This comment is incomplete. The nil can also happen if it fails to read ticks if `host_statistics()` fails (`result != KERN_SUCCESS ` below).

DatadogRUM/Sources/RUMVitals/VitalCPUReader.swift
```diff
@@ -23,34 +24,56 @@ internal class VitalCPUReader: SamplingBasedVitalReader {
...
     }
+
     @objc
     private func appDidBecomeActive() {
-        if let previouslyReadTicks = utilizedTicksWhenResigningActive,
-           let currentTicks = readUtilizedTicks() {
-            utilizedTicksWhenResigningActive = nil
-            totalInactiveTicks += currentTicks - previouslyReadTicks
-        }
+        lastReadActiveTicks = readUtilizedTicks()
```
simaoseica-dd, line 73: If the app is already active and another `didBecomeActive` notification arrives, the `lastReadActiveTicks` will be replaced with the new value, losing completely the previous value.

simaoseica-dd (commented):
This PR seems to change more than necessary and introduces gaps in a few edge cases.
My suggestion is to keep the current business logic intact and fix the issue in `readUtilizedTicks()` by normalizing the 32-bit `cpu_ticks.0` value.

PR #2967, opened 2026-06-02
RUM-16442: Fix profiler safe memory reads

simaoseica-dd:
### What and why?

Fixes a crash in the iOS Mach profiler caused by reading stale or invalid frame pointers during stack unwinding.

The previous safe-read path, introduced in [#2605](https://github.com/DataDog/dd-sdk-ios/pull/2605), relied on faulting memory access recovery but invalid stack memory could still terminate the app with `EXC_BAD_ACCESS` / `SIGSEGV`. 
This PR makes stack memory reads non-faulting so the profiler degrades safely instead of crashing.

### How?

Replaces direct memory reads during stack sampling with `vm_read_overwrite`.
The profiler now first tries to read the expected stack region in one batch for performance. If that read fails, it falls back to page-by-page reads from the stack pointer and keeps the readable prefix, allowing partial stack recovery without touching invalid memory.

### Review checklist

### DIFF
--- a/DatadogProfiling/Mach/mach_sampling_profiler.cpp
+++ b/DatadogProfiling/Mach/mach_sampling_profiler.cpp
@@ -67,6 +67,13 @@ static constexpr size_t STACK_REGION_MAX_READ = 65536; // 64 KB safety cap
 // 128 bytes covers the vast majority of real-world frames without over-reading.
 static constexpr size_t STACK_REGION_FRAME_SIZE_ESTIMATE = 128;
 
+struct frame_pointer_pair_t {
+    void* next_frame_pointer;
+    void* return_address;
+};
+
+static_assert(sizeof(frame_pointer_pair_t) == sizeof(void*) * 2, "Frame pointer pair must stay two pointers.");
+
 // Main thread pthread identifier for comparison
 static pthread_t g_main_pthread = NULL;
 
@@ -98,7 +105,8 @@ static size_t read_memory_region(void* addr, uint8_t* buf, size_t size) {
     }
 
     const uintptr_t address = reinterpret_cast<uintptr_t>(addr);
-    if (address > UINTPTR_MAX - size) {
+    const uintptr_t buffer = reinterpret_cast<uintptr_t>(buf);
+    if (address > UINTPTR_MAX - size || buffer > UINTPTR_MAX - size) {
         return 0;
     }
 
@@ -146,7 +154,7 @@ static size_t read_stack_region_by_pages(void* sp, uint8_t* buf, size_t size) {
         }
 
         total_read += bytes_read;
-        if (total_read == size || current_address > UINTPTR_MAX - bytes_read) {
+        if (total_read == size) {
             break;
         }
 
@@ -321,18 +329,18 @@ static bool read_frame_pair_from_snapshot(
     uintptr_t stack_base,
     const uint8_t* stack_buf,
     size_t bytes_read,
-    void* next_frame[2]
+    frame_pointer_pair_t* next_frame
 ) {
-    if (stack_base == 0 || stack_buf == nullptr || bytes_read < sizeof(void*[2]) || fp < stack_base) {
+    if (stack_base == 0 || stack_buf == nullptr || bytes_read < sizeof(frame_pointer_pair_t) || fp < stack_base) {
         return false;
     }
 
     const uintptr_t fp_offset = fp - stack_base;
-    if (fp_offset > bytes_read - sizeof(void*[2])) {
+    if (fp_offset > bytes_read - sizeof(frame_pointer_pair_t)) {
         return false;
     }
 
-    memcpy(static_cast<void*>(next_frame), stack_buf + fp_offset, sizeof(void*[2]));
+    memcpy(static_cast<void*>(next_frame), stack_buf + fp_offset, sizeof(frame_pointer_pair_t));
     return true;
 }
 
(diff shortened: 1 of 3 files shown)


DatadogProfiling/Mach/mach_sampling_profiler.cpp
```diff
@@ -110,24 +82,94 @@ void set_main_thread(pthread_t thread) {
...
-        return;
+static size_t read_memory_region(void* addr, uint8_t* buf, size_t size) {
+    if (addr == nullptr || buf == nullptr || size == 0) {
+        return 0;
+    }
+
+    const uintptr_t address = reinterpret_cast<uintptr_t>(addr);
+    if (address > UINTPTR_MAX - size) {
+        return 0;
+    }
```
fuzzybinary, line 111: Is this supposed to be a check on `addr` or on `buf`? Or should it check both? I would think since `addr` is the read address, if were going to pass `UINTPTR_MAX` than it would just return the amount read when calling `vm_read_overwirte`.

DatadogProfiling/Mach/mach_sampling_profiler.cpp
```diff
@@ -110,24 +82,94 @@ void set_main_thread(pthread_t thread) {
...
+            buf + total_read,
+            chunk_size
+        );
+
+        if (bytes_read != chunk_size) {
+            break;
+        }
+
+        total_read += bytes_read;
+        if (total_read == size || current_address > UINTPTR_MAX - bytes_read) {
```
fuzzybinary: I'm confused about this check:
```
current_address > UINTPTR_MAX - bytes_read
```

If this is true, doesn't that mean `read_memory_region` has already read past UINTPTR_MAX?

DatadogProfiling/Mach/mach_sampling_profiler.cpp
```diff
@@ -266,67 +314,200 @@ bool stack_trace_get_thread_info(stack_trace_t* trace, thread_t thread) {
...
+ * Reads a saved frame pointer / program counter pair from a stack snapshot.
+ */
+static bool read_frame_pair_from_snapshot(
+    uintptr_t fp,
+    uintptr_t stack_base,
+    const uint8_t* stack_buf,
+    size_t bytes_read,
+    void* next_frame[2]
+) {
+    if (stack_base == 0 || stack_buf == nullptr || bytes_read < sizeof(void*[2]) || fp < stack_base) {
```
fuzzybinary: if you can uses `constexpr` it might be worth factoring out the repeated `sizeof` checks:

```c++
constexpr size_t read_size = sizeof(void*[2]);
```

fuzzybinary: Actually, this might be a larger refactor, but I'm seeing a lot of use of void*[2] -- refactoring this into a `struct` would make this easier to read with no impact on overhead:

```c++
struct Frame {
  void* next_frame_pointer;
  void* return _addresss;
}
```

Maybe worth a the change sometime in the future.

fuzzybinary (approved):
This LGTM other than some minor nits about style, but I'm trusting your tests exercise this properly as I'm a little out of my depth on the specifics of stack walking for profiling.

DatadogProfiling/Mach/mach_sampling_profiler.cpp
simaoseica-dd: Good point. It is better to check both to have a strict contract and avoid passing any overflowing range.

DatadogProfiling/Mach/mach_sampling_profiler.cpp
simaoseica-dd: It was intended to guard the next line `current_address += bytes_read`. But you are right, the `read_memory_region()` already validates the ranges.

DatadogProfiling/Mach/mach_sampling_profiler.cpp
simaoseica-dd: Good suggestion. I agree that a struct makes the saved frame layout clearer. I added the `struct frame_pointer_pair_t`

DatadogProfiling/Mach/mach_sampling_profiler.cpp
```diff
@@ -266,67 +314,126 @@ bool stack_trace_get_thread_info(stack_trace_t* trace, thread_t thread) {
...
+        stack_base = reinterpret_cast<uintptr_t>(sp);
+
+        const uintptr_t fp_addr = reinterpret_cast<uintptr_t>(fp);
+        const size_t current_frame_size = (fp_addr > stack_base) ? (fp_addr - stack_base) : 0;
+        const size_t region_size = std::min(
+            current_frame_size + (static_cast<size_t>(max_depth) * STACK_REGION_FRAME_SIZE_ESTIMATE),
+            STACK_REGION_MAX_READ
+        );
+
+        bytes_read = read_stack_region(sp, stack_buf, region_size);
```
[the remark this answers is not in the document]
simaoseica-dd: Moving the VM read after `thread_resume` was considered but that will make the sample inconsistent like codex [commented here](https://github.com/DataDog/dd-sdk-ios/pull/2967/changes#r3350259836).
So the read stays inside the suspended window to keep the stack snapshot coherent. To reduce the risk, the suspended section is kept as minimal as possible.

maxep (approved):
Nice work!

PR #2946, opened 2026-05-26
Wire DDSketch into StatsConcentrator for latency summaries

tienquocbui:
### What and why?

The `okSummary` / `errorSummary` fields in `ExportedGroupedStats` have been empty `Data()` placeholders since PR 3 (the stats concentrator landed before the DDSketch implementation). Now that [PR 4](https://github.com/DataDog/dd-sdk-ios/pull/2914) has merged, this PR plugs DDSketch into the aggregation path so the latency distributions are actually populated when buckets are flushed.

### How?

- `GroupedStats` now owns two `DDSketch` instances — `okSummary` and `errorSummary` — created via `DDSketch.makeForStats()` (1% relative accuracy, 2048 max bins).
- A new `record(duration:isError:)` method on `GroupedStats` routes each span's duration into exactly one of the two sketches based on `isError`, mirroring the [v1.2.0 spec](https://github.com/DataDog/architecture/blob/main/rfcs/apm/integrations/client-side-stats/rfc.md).
- `StatsConcentrator.add()` calls `record(...)` alongside the existing counter bumps. No change to the eligibility filter or the bucket layout.
- `StatsConcentrator.flush()` calls `toProtoBytes()` on each sketch when assembling `ExportedGroupedStats`, replacing the placeholders.

### Tests

Four new tests covering the routing rule and the empty-sketch baseline:

- `testOkSpanPopulatesOkSummaryNotErrorSummary` — a non-error span exceeds the empty-sketch byte length on `okSummary` and equals it on `errorSummary`.
- `testErrorSpanPopulatesErrorSummaryNotOkSummary` — mirror case.
- `testMixedOkAndErrorSpans_populateBothSummaries` — both sketches contain values for a group that sees both span kinds.
- `testIneligibleSpansDoNotContributeToSummaries` — eligibility filter still drops non-qualifying spans before any sketch update.

The previous `testExportedGroupedStatsHasEmptySketchPlaceholders` is removed; it asserted the placeholder behavior this PR is replacing.

### Review checklist
- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [ ] Make sure each commit and the PR mention the Issue number or JIRA reference
- [x] Add CHANGELOG entry for user facing changes — N/A (internal, not user-facing)
- [x] Add Objective-C interface for public APIs — N/A (internal to DatadogTrace)
- [x] Run `make api-surface` when adding new APIs — N/A (internal to DatadogTrace)

PR #2971, opened 2026-06-03
[RUM-16605] Add APM client-side stats models, MessagePack encoder, and customStatsEndpoint

tienquocbui:
### What and why?

This PR adds the data models, MessagePack encoder, and configuration surface that the Client-Side Stats feature needs in order to upload to the `/api/v0.2/stats` intake. It is the iOS counterpart to Android [#3487 (RUM-16535)](https://github.com/DataDog/dd-sdk-android/pull/3487).

It also fixes a latent scaffolding bug: `ClientStatsFeature` was constructing its request builder from `Trace.Configuration.customEndpoint` (the spans endpoint override). Spans and stats go to different intakes (`/api/v2/spans` vs `/api/v0.2/stats`), so a customer enabling `statsComputationEnabled: true` together with `customEndpoint` would have stats POSTed to the trace intake and silently rejected.

### How?

1. **`customStatsEndpoint` configuration.** New `Trace.Configuration.customStatsEndpoint: URL?` (Swift + ObjC bridge). `ClientStatsFeature` switched to read `customStatsEndpoint` instead of `customEndpoint`.
2. **`MsgPackEncoder`.** Hand-rolled minimal MessagePack encoder, internal, no third-party runtime dependency. Implements only the subset of the spec used by the stats payload (null, bool, int32, int64, fixstr/str8/16/32, bin8/16/32, fixmap/map16/32, fixarray/array16/32). Ported from the Android Kotlin implementation with field-level parity.
3. **Stats payload models.** `StatsPayload` (outer envelope), `ClientStatsPayload` (per-tracer payload), `ClientStatsBucket` (time-bucketed group), `ClientGroupedStats` (per `(service, name, resource, ...)` aggregate). Field names and order match `stats_gen.go` in `datadog-agent`, ensuring byte-level cross-platform parity with Android. `agentHostname`, `agentEnv`, and `agentVersion` are hardcoded empty (no agent on mobile). `clientComputed` is always `true`. `lang` is `"ios"`.
4. **Regenerate `api-surface`.** Reflects the new `customStatsEndpoint` plus the previously missing `statsComputationEnabled` entry (housekeeping fix from the earlier scaffolding PR).

### Tests

**`MsgPackEncoderTests`** (60 tests):
- Per-method format boundary coverage: `fixstr` upper, `str8` lower/upper, `str16` lower/upper, `str32` lower; same shape for binary, int, long, array, map.
- Signed integer two's-complement edge cases for `Int32.min`, `Int32.max`, `Int64.min`, `Int64.max`, and the negfixnum and fixnum boundaries.
- UTF-8 multibyte string round-trip.
- Composite assertions for nested map + array + mixed-type entries.
- First-byte marker sanity table to catch accidental format-constant drift.

**`ClientStatsPayloadTests`** (14 tests) and **`StatsPayloadTests`** (6 tests):
- Verify map element counts (9 for `ClientStatsPayload`, 3 for `ClientStatsBucket`, 17 for `ClientGroupedStats`, 6 for `StatsPayload`).
- Verify field order matches the protobuf definition byte-for-byte.
- Verify scalar values, binary payloads (`okSummary`, `errorSummary`), `peerTags` arrays, `Trilean.isTraceRoot` mapping, hardcoded fields (`Synthetics=false`, `GRPCStatusCode=""`, `ClientComputed=true`, `lang="ios"`).
- Verify `StatsPayload.appendRawBytes` splices pre-encoded `ClientStatsPayload` bytes correctly.

**Independent decoder for cross-check.** A test-only `MsgPackTestDecoder` (separate from the production encoder) decodes encoder output to verify the structural and value-level correctness. Same intent as the `msgpack-java` cross-check on Android, without pulling in a third-party Swift dependency.

### Not in scope (separate tickets)

- Request builder content-type and gzip compression for the stats payload.
- Flush hook in `StatsConcentrator` that converts `ExportedBucket` to `StatsPayload`.
- Storage / upload pipeline wiring.

### Review checklist
- [ ] Add CHANGELOG entry for user facing changes. Deferred to the feature branch merge into `develop`, matching the pattern set by #2946 and #2947.
- [x] Add Objective-C interface for public APIs. `DDTraceConfiguration.customStatsEndpoint` added.
- [x] Run `make api-surface` when adding new APIs. Diff included.


DatadogTrace/Sources/Stats/ClientStatsPayload.swift
```diff
@@ -0,0 +1,213 @@
...
+        /// not language.
+        static let langValue = Data("ios".utf8)
+        static let tracerVersion = Data("TracerVersion".utf8)
+        static let runtimeID = Data("RuntimeID".utf8)
+        static let sequence = Data("Sequence".utf8)
+        static let service = Data("Service".utf8)
+    }
+}
+
+/// A time-bucketed group of `ClientGroupedStats` for a fixed `[start, start + duration)` window.
```
arroz: Should this be `[start, start + duration]` or `[start, start + duration[`?

DatadogTrace/Sources/Stats/MsgPackEncoder.swift
```diff
@@ -0,0 +1,341 @@
...
+    func writeString(_ string: String?) {
+        guard let string = string else {
+            writeNull()
+            return
+        }
+        writeRawString(Data(string.utf8))
+    }
+
+    /// Writes the raw bytes of a UTF-8 string. The caller is responsible for ensuring `utf8Bytes` is valid UTF-8.
+    func writeRawString(_ utf8Bytes: Data) {
```
arroz: Since the constants we pass to this function (like `static let resource = Data("Resource".utf8)`) are essentially doing the same thing as `writeString(_ string: String?)`, why having this function at all? Might make more sense to have a function for `String` and another one for `String?` but none for raw, as that should be handled by the encoder.

DatadogTrace/Sources/Stats/MsgPackEncoder.swift
```diff
@@ -0,0 +1,341 @@
...
+import Foundation
+
+/// Subset of the MessagePack specification required to encode `StatsPayload`.
+///
+/// See https://github.com/msgpack/msgpack/blob/master/spec.md
+///
+/// The encoder is intentionally minimal: only the formats actually used by the
+/// client-side stats payload are supported. Cross-SDK parity with the Android
+/// implementation is verified by `MsgPackEncoderTests`.
+internal final class MsgPackEncoder {
```
arroz: Was it considered to build this as a Swift `Encoder`, and integrate with `Encodable` architecture? It would save all the manual encoding code on the structs above, and allow us to encode any type as MsgPack in the future. Unless there's any strong reasons against it, that should likely be the way forward.

arroz (commented):
Biggest thing is why not writing the encoder implementing Swift's `Encoder` protocol. This would be beneficial for current and future use.

DatadogTrace/Sources/Stats/ClientStatsPayload.swift
```diff
@@ -0,0 +1,148 @@
...
+
+    func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encode(hostname, forKey: .hostname)
+        try container.encode(env, forKey: .env)
+        try container.encode(version, forKey: .version)
+        try container.encode(stats, forKey: .stats)
+        // Platform identifier on the wire, matching Android's `"android"`. Distinct from
+        // the spans `lang` tag (`"swift"`); the stats intake keys off platform.
+        try container.encode("ios", forKey: .lang)
```
arroz: I think you can avoid the custom encode method by adding this as a private variable to the object, like `private let lang = "ios"`. Since it's a constant, it should not have any significative impact on the memory usage as the compiler optimizes it out (I think). As long as you keep it in `CodingKeys` it should be encoded, even if private (test it to make sure!).

DatadogTrace/Sources/Stats/ClientStatsPayload.swift
```diff
@@ -0,0 +1,148 @@
...
+    let duration: UInt64
+    let stats: [ClientGroupedStats]
+
+    private enum CodingKeys: String, CodingKey {
+        case start = "Start"
+        case duration = "Duration"
+        case stats = "Stats"
+    }
+
+    func encode(to encoder: Encoder) throws {
```
arroz: You don't need this method, I think. `CodingKeys` here is enough.

DatadogTrace/Sources/Stats/ClientStatsPayload.swift
```diff
@@ -0,0 +1,148 @@
...
+        case topLevelHits = "TopLevelHits"
+        case spanKind = "SpanKind"
+        case peerTags = "PeerTags"
+        case isTraceRoot = "IsTraceRoot"
+        case grpcStatusCode = "GRPCStatusCode"
+        case serviceSource = "srv_src"
+    }
+
+    func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
```
arroz: Same as above, you can encode the constants as private lets.

ncreated (approved):
Nice work 👍. Switching to Swift's `Encodable` was the right call 👌 . Thanks for scoping the PR very well it makes it easy to review 🙂.

arroz (approved):
Looks good! 🎉

Issue #2004, opened 2024-08-20
Logs are never uploaded on real Apple Watch device


jfiser-paylocity:
### Describe the bug

When datadog is used on a real Apple Watch device, no logs are ever uploaded to API. The issue is that `NWPathMonitor` is used to evaluate network reachability. While this works flawlessly on a simulator (as it reads macOS status instead), it always reports as unreachable on a real device. That leads to `DataUploadConditions` blocking every data upload attempt.

### Reproduction steps

Integrate datadog in any watchOS application and run on a real Apple Watch device.

### SDK logs

No visible logs in the watchOS application.

### Expected behavior

Logs are eventually uploaded to API and visible on Datadog dashboard

### Affected SDK versions

2.15.0

### Latest working SDK version

none

### Did you confirm if the latest SDK version fixes the bug?

Yes

### Integration Methods

SPM

### Xcode Version

15.4

### Swift Version

5.10

### MacOS Version

Sonoma 14.6.1

### Deployment Target

watchOS 9.0

maciejburda:
Thank you for reporting the issue @jfiser-paylocity 

Apparently [there're some issues](https://developer.apple.com/forums/thread/729568) with `NWPathMonitor` on watchOS.

I tried looking for some simple mitigation, but unfortunately could make it work in the timely manner. We'll need to discuss this internally with the team and see how we can proceed.

We'll keep the GH Issue posted.

mariedm:
Hello @jfiser-paylocity 👋

We released [version 2.17.0](https://github.com/DataDog/dd-sdk-ios/releases/tag/2.17.0), which includes #2005. Thanks a lot for your contribution! 🙌 🚀

PR #2975, opened 2026-06-04
Fix watchOS uploads blocked by NWPathMonitor always reporting no reachability

### LINKED ISSUE #2974: Logs not uploading on watchOS
### Describe the bug

Logs being created via the SDK (i.e an `DatadogLogs.LoggerProtocol` instance) do not seem to be getting uploaded to Datadog, and are not not findable in the Log Explorer, when the logging occurs in our Apple Watch application (watchOS). We had somewhat recently updated the SDK version from 3.7.0 to 3.10.0, and logs for any users that are still on a version of our app that uses 3.7.0 are still uploading + findable in Log Explorer.

It seems like when I debug by setting `Datadog.verbosityLevel = .debug`, I can see perpetual logs in the Datadog SDK that basically say `[upload was skipped because: 📡 Network reachability is: no]` continuously, even if the watch is both on wifi and/or paired to a phone with a connection.

Am I simply missing a newer config/setup component that would resolve this? Otherwise in the 'Other relevant information' section of this issue, I did put my hypothesis + research surrounding how reachability on watchOS may be causing the issue.

### Reproduction steps

In a watchOS Application:
 
- initialize Datadog SDK on launch
```swift
Datadog.initialize(
	with: Datadog.Configuration(
		clientToken: "insert-our-token-here",
		env: "production",
		site: .us1,
		service: "insert-our-service-name-here"
	),
	trackingConsent: .granted
)

#if DEBUG
Datadog.verbosityLevel = .debug
#endif

Logs.enable(with: Logs.Configuration())

Trace.enable(
	with: Trace.Configuration(
		sampleRate: 20,
		networkInfoEnabled: true
	)
)
```

- later, create a logger and try to send a log message
```swift
public final class DatadogLogger {
	private let sdkLogger: LoggerProtocol

	public init(name: String) {
		let config = DatadogLogs.Logger.Configuration(
			name: name,
			networkInfoEnabled: true
		)
		self.sdkLogger = DatadogLogs.Logger.create(with: config)
	}

	public func logTest() {
		sdkLogger.info("Test log message on watchOS")
	}
}

// usage somewhere in the app
let testLogger = DatadogLogger(name: "test-watchos")
testLogger.logTest()
```

### SDK logs

Throughout app usage + logging attemps, I just perpetually see a bunch of these, regardless of having network connectivity:

```
[DATADOG SDK] 🐶 → 15:10:28.596 💡 (logging) No upload. Batch to upload: NOT CHECKED, System conditions: ❌ [upload was skipped because: 📡 Network reachability is: no]
[DATADOG SDK] 🐶 → 15:10:29.200 💡 (logging) No upload. Batch to upload: NOT CHECKED, System conditions: ❌ [upload was skipped because: 📡 Network reachability is: no]
[DATADOG SDK] 🐶 → 15:10:30.118 💡 (logging) No upload. Batch to upload: NOT CHECKED, System conditions: ❌ [upload was skipped because: 📡 Network reachability is: no]
[DATADOG SDK] 🐶 → 15:10:31.501 💡 (logging) No upload. Batch to upload: NOT CHECKED, System conditions: ❌ [upload was skipped because: 📡 Network reachability is: no]
[DATADOG SDK] 🐶 → 15:10:33.419 💡 (logging) No upload. Batch to upload: NOT CHECKED, System conditions: ❌ [upload was skipped because: 📡 Network reachability is: no]
```

### Expected behavior

Ideally the logs would upload when there's network connectivity

### Affected SDK versions

3.10.0

3.7.0 (but possibly others, this was the one that was working for us before 3.10.0)

### Did you confirm if the latest SDK version fixes the bug?

No

### Integration Methods

SPM

### Xcode Version

Xcode 26.1

### Swift Version

Swift 6.2.1

### MacOS Version

macOS Tahoe 26.5

### Deployment Target

watchOS 8+

### Device Information

e.g. Apple Watch Series 9, watchOS 26.2.1, on wifi + paired to iPhone 15 Pro

This is not necessarily the issue for sure, but I did notice there was a [change to DataUploadConditions.swift](https://github.com/DataDog/dd-sdk-ios/commit/53c7e55313f2b4020fb5132381fec31f00b78942#diff-673d8f45bfb5193d24bf384905c8f5603acbf98af0f2d9bd9b8653ddf2e4725fL32-L41) which turned on network reachability as a blocker for uploads on watchOS, in the SDK version we started seeing the issue. Additionally, if reachability is ultimately being determined/provided by [NWPathMonitor](https://developer.apple.com/documentation/Network/NWPathMonitor), that seems like it may always be 'unreachable'/'unsatisfied' per [TN3135: Low-level networking on watchOS](https://developer.apple.com/documentation/technotes/tn3135-low-level-networking-on-watchos)
> watchOS blocks low-level networking outside of these specific circumstances. For example, if a normal app attempts to start an [NWConnection](https://developer.apple.com/documentation/Network/NWConnection), that connection will stay in the [.waiting(_:)](https://developer.apple.com/documentation/Network/NWConnection/State-swift.enum/waiting(_:)) state with an error of ENETDOWN. **Similarly, an [NWPathMonitor](https://developer.apple.com/documentation/Network/NWPathMonitor) will remain in the [.unsatisfied](https://developer.apple.com/documentation/Network/NWPath/Status-swift.enum/unsatisfied) state.**

It looks like this may have came up in [the past here](https://github.com/DataDog/dd-sdk-ios/issues/2004), which was solved by the change that whose removal I linked above.

barboraplasovska:
### What and why?

Regression introduced in 3.10.0 (PR #2768): `NWPathMonitor` always reports `.unsatisfied` on watchOS for non-audio-streaming apps (Apple TN3135), causing `DataUploadConditions` to permanently block all uploads even when the watch has connectivity. This was the same root cause as issue #2004, fixed in 2.17.0, and accidentally reverted in #2768.

Closes #2974.

### How?

On watchOS, replace `NWPathMonitorPublisher` with a `NOPContextValuePublisher` that statically reports `.maybe` reachability. `NWPathMonitor` is never started, and `DataUploadConditions` stays platform-agnostic.

### Review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1179,6 +1179,7 @@ Release `2.0` introduces breaking changes. Follow the [Migration Guide](MIGRATIO
 [#2963]: https://github.com/DataDog/dd-sdk-ios/pull/2963
 [#2955]: https://github.com/DataDog/dd-sdk-ios/pull/2955
 [#2975]: https://github.com/DataDog/dd-sdk-ios/pull/2975 
+[#2975]: https://github.com/DataDog/dd-sdk-ios/pull/2975
 
 [@00fa9a]: https://github.com/00FA9A
 [@britton-earnin]: https://github.com/Britton-Earnin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@
 - [FIX] Fix wrong types in the `objc_LogEventDevice` properties definition. See [#2966][]
 - [FIX] Expose RUM operation options to Objective-C from `DatadogRUM`. See [#2969][]
 - [IMPROVEMENT] Add Objective-C API support for custom SDK instance names across all modules. See [#2955][]
+- [FEATURE] Instrumented Web Views now have their tracing decision consistent with the native SDK. See [#2859][]
+- [IMPROVEMENT] Align public RUM session IDs with event formatting. See [#2956][]
+- [FIX] Fix watchOS uploads blocked by NWPathMonitor always reporting no reachability. See [#2975][]
 
 # 3.11.1 / 28-05-2026
 
@@ -1175,6 +1178,7 @@ Release `2.0` introduces breaking changes. Follow the [Migration Guide](MIGRATIO
 [#2969]: https://github.com/DataDog/dd-sdk-ios/pull/2969
 [#2963]: https://github.com/DataDog/dd-sdk-ios/pull/2963
 [#2955]: https://github.com/DataDog/dd-sdk-ios/pull/2955
+[#2975]: https://github.com/DataDog/dd-sdk-ios/pull/2975 
 
 [@00fa9a]: https://github.com/00FA9A
 [@britton-earnin]: https://github.com/Britton-Earnin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,6 @@
 - [FIX] Fix wrong types in the `objc_LogEventDevice` properties definition. See [#2966][]
 - [FIX] Expose RUM operation options to Objective-C from `DatadogRUM`. See [#2969][]
 - [IMPROVEMENT] Add Objective-C API support for custom SDK instance names across all modules. See [#2955][]
-- [FEATURE] Instrumented Web Views now have their tracing decision consistent with the native SDK. See [#2859][]
-- [IMPROVEMENT] Align public RUM session IDs with event formatting. See [#2956][]
 
 # 3.11.1 / 28-05-2026
 
@@ -1178,7 +1176,6 @@ Release `2.0` introduces breaking changes. Follow the [Migration Guide](MIGRATIO
 [#2969]: https://github.com/DataDog/dd-sdk-ios/pull/2969
 [#2963]: https://github.com/DataDog/dd-sdk-ios/pull/2963
 [#2955]: https://github.com/DataDog/dd-sdk-ios/pull/2955
-[#2975]: https://github.com/DataDog/dd-sdk-ios/pull/2975 
 [#2975]: https://github.com/DataDog/dd-sdk-ios/pull/2975
 
 [@00fa9a]: https://github.com/00FA9A
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - [FEATURE] Add wildcard host pattern support to WebView tracking via `WebViewTracking.enable(webView:hostPatterns:)`. See [#2963][]
+- [FIX] Fix watchOS uploads blocked by NWPathMonitor always reporting no reachability. See [#2975][]
 
 # 3.12.0 / 04-06-2026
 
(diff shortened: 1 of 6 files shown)


gonzalezreal (commented):
Could you add a changelog entry for this fix?

This changes user-facing watchOS upload behavior, so it should be listed in the release notes.

gonzalezreal (dismissed):

DatadogCore/Sources/Core/DatadogCore.swift
```diff
@@ -553,7 +553,15 @@ extension DatadogContextProvider {
...
 
+        #if os(watchOS)
+        // NWPathMonitor always reports `.unsatisfied` on watchOS for non-audio-streaming apps (Apple TN3135).
+        // URLSession uploads work fine on watchOS (proxied via paired iPhone), so assume reachable.
+        subscribe(\.networkConnectionInfo, to: NOPContextValuePublisher(initialValue: NetworkConnectionInfo?.some(
+            NetworkConnectionInfo(reachability: .maybe, availableInterfaces: nil, supportsIPv4: nil, supportsIPv6: nil, isExpensive: nil, isConstrained: nil)
+        )))
+        #else
         subscribe(\.networkConnectionInfo, to: NWPathMonitorPublisher())
+        #endif
```
maxep: cr/ This will change the connectivity reported in events, which won't reflect the actual [TN3135](https://developer.apple.com/documentation/technotes/tn3135-low-level-networking-on-watchos). Can we revert [this change](https://github.com/DataDog/dd-sdk-ios/pull/2768/changes#diff-673d8f45bfb5193d24bf384905c8f5603acbf98af0f2d9bd9b8653ddf2e4725fL32-L41) instead?

DatadogCore/Sources/Core/Upload/DataUploadConditions.swift
```diff
@@ -29,6 +29,7 @@ internal struct DataUploadConditions {
 
     func blockersForUpload(with context: DatadogContext) -> [Blocker] {
         var blockers: [Blocker] = []
+        #if !os(watchOS)
```
gonzalezreal, line 32: Could we add or adjust a watchOS test for this path?
`DatadogCoreTests` runs on watchOS, and the current upload-condition tests still expect `.no` reachability to block uploads.

CHANGELOG.md
```diff
@@ -1173,6 +1174,7 @@ Release `2.0` introduces breaking changes. Follow the [Migration Guide](MIGRATIO
 [#2968]: https://github.com/DataDog/dd-sdk-ios/pull/2968
 [#2969]: https://github.com/DataDog/dd-sdk-ios/pull/2969
 [#2963]: https://github.com/DataDog/dd-sdk-ios/pull/2963
+[#2975]: https://github.com/DataDog/dd-sdk-ios/pull/2975
```
gonzalezreal: Small nit: this line has trailing whitespace.

gonzalezreal (dismissed):

barboraplasovska:
I had to rebase to fix merge conflicts in CHANGELOG.

gonzalezreal (approved):

maxep (approved):

PR #2977, opened 2026-06-05
[PANA-7550] Preserve layer subclass content during occlusion pruning

gonzalezreal:
### What and why?

This PR fixes the layer tree optimization step so it does not remove visible Core Animation subclass layers.

SwiftUI can use layer subclasses, such as shape layers, that draw content through subclass-specific state instead of `contents`, background, border, or shadow.

### How?

`CALayerSnapshot.drawsContent` now treats non-`CALayer` subclasses as content-bearing.

This keeps pruning conservative for custom or framework-owned layers, while still allowing empty plain `CALayer` containers to be removed.

Added focused occlusion tests for `CAShapeLayer`.

### Review checklist

### DIFF
--- a/DatadogSessionReplay/Sources/LayerTreeRecorder/CALayerSnapshot+Occlusion.swift
+++ b/DatadogSessionReplay/Sources/LayerTreeRecorder/CALayerSnapshot+Occlusion.swift
@@ -5,14 +5,19 @@
  */
 
 #if os(iOS)
-import CoreGraphics
 import Foundation
+import QuartzCore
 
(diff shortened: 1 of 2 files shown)


barboraplasovska (approved):
👍🏻 
Should we add a CHANGELOG entry ?

gonzalezreal:
> 👍🏻 Should we add a CHANGELOG entry ?

@barboraplasovska Thanks for taking a look. No need to add a CHANGELOG entry, as this is still under an internal feature flag. Once we open this for preview, we will add an entry.

PR #2981, opened 2026-06-09
RUM-16667 Add wildcard host pattern matching to first-party hosts tracing

barboraplasovska:
### What and why?

Adds wildcard host pattern support to first-party host tracing. Customers can now use `*` as a wildcard in `firstPartyHostsTracing` (e.g. `*.example.com`, `preview-*.shopist.io`) instead of enumerating every host explicitly.

### How?

- Added `case traceWithPatterns(hostPatterns: [String])` to the three `FirstPartyHostsTracing` enums and their Objective-C wrappers
- New `HostPatternSanitizer` validates patterns at config time (character allowlist, single wildcard, no bare `*`)
- `FirstPartyHosts` extended with a second storage field and a plain-string wildcard matcher (prefix/suffix anchoring, no regex, apex excluded)
- Merge operators (`+=`, `+`) updated to carry patterns alongside plain hosts

### Review checklist


DatadogInternal/Sources/NetworkInstrumentation/HostPatternSanitizer.swift
```diff
@@ -0,0 +1,63 @@
...
+    guard lowercased.unicodeScalars.allSatisfy({ allowedCharacters.contains($0) }) else {
+        warnings.append("'\(pattern)' is not a valid host pattern and will be dropped.")
+        return nil
+    }
+    let wildcardCount = lowercased.filter({ $0 == "*" }).count
+    guard wildcardCount <= 1 else {
+        warnings.append("'\(pattern)' is not a valid host pattern and will be dropped.")
+        return nil
+    }
+    guard wildcardCount == 0 || lowercased.contains(".") else {
```
kikoveiga: For plain (wildcard-free) entries, we should keep parity with the old `hosts:`/`enable` API. Because of this, I think we can route wildcard-free entires back through `HostsSanitizer` and only send `*` entries through this pattern validator. We would be able to drop entries like `com`, which are actually passing now, wdyt?

kikoveiga: Also, here we are now dropping a bare `*`, which could make sense as an allow-all hosts match. Should we add a `lowercased == "*"` statement to the `guard`?

barboraplasovska: I fixed the first comment, on the other hand we never want to support bare `*`.

DatadogWebViewTracking/Sources/WebViewTracking.swift
```diff
@@ -157,21 +157,10 @@ public enum WebViewTracking {
...
-                DD.logger.warn("WebView host pattern \"\(pattern)\" contains more than one wildcard and will be ignored.")
-                return nil
-            }
-            return lowercased
-        }
-
+        let validPatterns = sanitizeHostPatterns(
+            hostPatterns,
+            warningMessage: "The WebView host pattern configured for Datadog SDK is not valid"
+        )
```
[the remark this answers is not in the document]
barboraplasovska: `URL(string: "https://foo.com%27")` returns nil in Foundation since `%27` in a hostname isn't RFC-valid, so this path isn't reachable.

DatadogInternal/Sources/NetworkInstrumentation/URLSession/URLSessionInstrumentation.swift
```diff
@@ -133,5 +133,12 @@ extension URLSessionInstrumentation {
         ///
         /// - `hostsWithHeaders` - Dictionary of hosts and tracing header types to use. Note: Hosts must not include "http(s)://" prefix.
         case traceWithHeaders(hostsWithHeaders: [String: Set<TracingHeaderType>])
+
+        /// Trace hosts matching the given wildcard patterns using Datadog and W3C `tracecontext` tracing headers.
+        ///
+        /// - Parameters:
+        ///   - hostPatterns: Wildcard patterns for hosts to trace. Use `*` as a wildcard (e.g. `"*.example.com"`).
+        ///     Patterns must not include the "http(s)://" prefix.
+        case traceWithPatterns(hostPatterns: [String])
```
maxep: suggestion/ rather than adding a new case, consider letting the existing `.trace(hosts:)` and `.traceWithHeaders(hostsWithHeaders:)` cases accept wildcard patterns alongside plain hosts. `FirstPartyHosts` can detect `*` at init time (the split logic already exists in `sanitizeHostPatterns`)

This removes three new enum cases, three new ObjC wrappers, and three near-identical switch arms across RUM/Trace/URLSession, and lets users naturally mix plain hosts and patterns in a single call

DatadogInternal/Sources/NetworkInstrumentation/FirstPartyHosts.swift
```diff
@@ -61,18 +81,37 @@ public struct FirstPartyHosts: Equatable {
...
+        guard pattern.contains("*") else {
+            return host == pattern || host.hasSuffix(".\(pattern)")
+        }
+        let parts = pattern.split(separator: "*", maxSplits: 1, omittingEmptySubsequences: false).map(String.init)
+        guard parts.count == 2 else {
+            return false
+        }
+        return host.hasPrefix(parts[0]) && host.hasSuffix(parts[1]) && host.count > parts[0].count + parts[1].count
+    }
+
```
maxep: This could use a regex to stay consistent with the plain-host matching on line 85 (which already uses `NSRegularExpression.escapedPattern`). Since the pattern is already sanitized, the conversion is simple:

```swift
private func matchesWildcardPattern(host: String, pattern: String) -> Bool {
    let parts = pattern.split(separator: "*", maxSplits: 1, omittingEmptySubsequences: false).map(String.init)
    let regex = "^\(NSRegularExpression.escapedPattern(for: parts[0])).+\(NSRegularExpression.escapedPattern(for: parts[1]))$"
    return host.range(of: regex, options: .regularExpression) != nil
}
```

`.+` naturally enforces "wildcard must match at least one character", replacing the manual count check. `escapedPattern` also future-proofs the literal prefix/suffix.

DatadogInternal/Sources/NetworkInstrumentation/HostPatternSanitizer.swift
```diff
@@ -0,0 +1,73 @@
...
+        return nil
+    }
+    guard lowercased.filter({ $0 == "*" }).count == 1 else {
+        warnings.append("'\(pattern)' is not a valid host pattern and will be dropped.")
+        return nil
+    }
+    guard lowercased.contains(".") else {
+        warnings.append("'\(pattern)' is not a valid host pattern and will be dropped.")
+        return nil
+    }
```
maxep: Edge case: a pattern like `*.` passes all three guards (valid chars, one `*`, contains `.`) but is not a valid host. Consider checking that neither side of the `*` is empty after the split. e.g. that the suffix after `*` is not just `.` and the pattern doesn't end with `*`.

maxep (commented):
Looks good, great feature! 

Two design suggestions before this lands:

**1. No new config case needed**

The existing `.trace(hosts:)` and `.traceWithHeaders(hostsWithHeaders:)` cases can absorb wildcard patterns transparently. `FirstPartyHosts` already has all the plumbing to split plain hosts from patterns at init time via `sanitizeHostPatterns`, it just needs to be wired up to the existing inits instead of gated behind a new case.

2. Use regex for wildcard matching

The plain-host path already uses `NSRegularExpression.escapedPattern` at match time. The new wildcard path should do the same rather than introducing a split/hasPrefix/hasSuffix approach.

DatadogInternal/Sources/NetworkInstrumentation/FirstPartyHosts.swift
```diff
@@ -89,21 +96,20 @@ public struct FirstPartyHosts: Equatable {
...
-
-    return FirstPartyHosts(
-        left.hostsWithTracingHeaderTypes.merging(right.hostsWithTracingHeaderTypes, uniquingKeysWith: { left, right in
-            left.union(right)
-        })
+    var result = FirstPartyHosts()
+    result.hostsWithTracingHeaderTypes = left.hostsWithTracingHeaderTypes.merging(
+        right.hostsWithTracingHeaderTypes, uniquingKeysWith: { $0.union($1) }
     )
+    return result
```
maxep: I don't see logic change here, and previous implementation was more readable IMHO :)

maxep (dismissed):
Looks great 👍  Reusing pre-existing config and logic makes it clearer!

maxep (approved):

PR #2984, opened 2026-06-10
[RUM-11641] Speed up api-surface generation and run it on PRs

tienquocbui:
### What and why?
`make api-surface` was too slow (~8 min locally, ~3 min in CI) to run on every PR, so it only ran on release branches. As a result the committed API surface references drifted from the actual public API. This makes generation about 5x faster so the check can run on every PR and stay accurate.

### How?
- Build each module once into a shared workspace and derived data path, so shared dependencies (like DatadogInternal) compile only once instead of once per module per language.
- Parse each module once and write both the Swift and ObjC surfaces in a single run, instead of two full passes.
- Parse modules in parallel using one `parse-module` child process per module, each with its own sourcekitd. In-process thread parallelism is unsafe (sourcekitd silently drops results), so it falls back to serial in-process parsing under `swift test`.
- Enable the `API Surface Verify` job on PRs.

Local timing went from 478s to ~95s. Output is byte for byte identical to the previous implementation.

### Review checklist
- [x] Tests updated (api-surface tool integration tests)
- [x] PR mentions the JIRA reference (RUM-11641)
- [ ] CHANGELOG entry: N/A, tooling and CI only
- [ ] Objective-C interface: N/A
- [x] Ran make api-surface-verify (passes)

### DIFF
--- a/tools/api-surface/Sources/APISurfaceCore/APISurface.swift
+++ b/tools/api-surface/Sources/APISurfaceCore/APISurface.swift
@@ -65,7 +65,7 @@ public struct APISurface {
 /// This is created inside a `parse-module` child process (one per module, each with its own sourcekitd), which is
 /// the only safe way to parallelize parsing: the in-process SourceKit daemon silently drops results when hit from
 /// multiple threads in the same process.
-struct ParsedAPISurface {
+internal struct ParsedAPISurface {
     private let docs: [SwiftDocs]
 
     /// Reconstructs the module from its name and compiler arguments and parses it. No build is performed here; it
@@ -85,7 +85,7 @@ struct ParsedAPISurface {
 
 /// Holds a temporary patched package workspace used to run `xcodebuild` and
 /// automatically removes it when no longer referenced.
-final class PatchedPackageWorkspace {
+internal final class PatchedPackageWorkspace {
     let path: String
 
     init(originalPath: String) throws {
--- a/tools/api-surface/Sources/APISurfaceCore/APISurface.swift
+++ b/tools/api-surface/Sources/APISurfaceCore/APISurface.swift
@@ -62,9 +62,10 @@ public struct APISurface {
 
 /// The parsed SourceKit documentation for a single module, ready to print for any language.
 ///
-/// This is created inside a `parse-module` child process (one per module, each with its own sourcekitd), which is
-/// the only safe way to parallelize parsing: the in-process SourceKit daemon silently drops results when hit from
-/// multiple threads in the same process.
+/// Created inside a `parse-module` child process (one per module). Parsing is isolated per process rather than
+/// parallelized across threads in a single process because, in our own testing, parsing modules concurrently on
+/// multiple threads produced incomplete output: declarations were silently dropped (e.g. the ObjC surface shrank
+/// from 3566 to 1204 lines). A fresh SourceKit state per child process avoids that.
 internal struct ParsedAPISurface {
     private let docs: [SwiftDocs]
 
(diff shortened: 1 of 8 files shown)


tools/api-surface/Sources/APISurfaceCore/APISurface.swift
```diff
@@ -11,53 +11,81 @@ public struct APISurfaceError: Error, CustomStringConvertible {
...
+    /// These are captured once from the build above and can be handed to a separate `parse-module` process so the
+    /// expensive SourceKit parsing runs in parallel without rebuilding (each process gets its own sourcekitd).
+    var compilerArguments: [String] { module.compilerArguments }
+}
+
+/// The parsed SourceKit documentation for a single module, ready to print for any language.
+///
+/// This is created inside a `parse-module` child process (one per module, each with its own sourcekitd), which is
+/// the only safe way to parallelize parsing: the in-process SourceKit daemon silently drops results when hit from
+/// multiple threads in the same process.
```
arroz: Minor thing, but if you have a source for that claim (bug tracker issue, web page, etc), add it to the comment.

tools/api-surface/Sources/APISurfaceCore/Commands.swift
```diff
@@ -90,45 +134,243 @@ public struct VerifyCommand: ParsableCommand {
...
+
+/// Runs a single `parse-module` child process for one module and returns its printed body per language.
+private func runParseModuleChild(
+    executablePath: String,
+    surface: APISurface,
+    languages: [Language],
+    tempDir: URL,
+    index: Int
+) throws -> [Language: String] {
+    let argsFile = tempDir.appendingPathComponent("args-\(index).json")
```
arroz, line 329: Have you considered using [swift-subprocess](https://github.com/swiftlang/swift-subprocess)? Might save a bunch of code and given it has an async interface, might make locking simpler on the Commands file as well (potentially eliminating them). It has the disadvantage of adding another dependency.

tools/api-surface/Sources/APISurfaceCore/APISurface.swift
tienquocbui: Good point. I don't have an authoritative source to cite (bug report / docs) for a specific sourcekitd mechanism, so I reworded it to describe what we actually observed rather than assert a cause: parsing modules concurrently on multiple threads in one process produced incomplete output (the ObjC surface shrank from 3566 to 1204 lines), and a fresh SourceKit state per child process avoids it. Updated in 41b1b7c.

tools/api-surface/Sources/APISurfaceCore/Commands.swift
tienquocbui: I did look at it, and I agree the async interface is nicer here. I'm going to keep Foundation `Process` for now though:

- It's a pre-1.0 (0.x) package whose policy is that any minor release may bump the required Swift toolchain, which is more churn than I'd like to take on for a build tool.
- It adds a dependency to the api-surface tool, whereas the current orchestration (`parseModulesInParallel`) is small and self-contained.
- GCD handles the blocking child waits (`waitUntilExit`) well, so the locking we have is minimal.

Compatibility isn't the blocker (CI runs on Xcode 26 / Swift 6.2, which it supports), so if the team would rather standardize on swift-subprocess for child-process handling I'm happy to switch in a follow-up. For this PR I'd prefer to keep the dependency surface unchanged.

arroz (dismissed):

arroz (approved):

PR #2987, opened 2026-06-10
fix(flags): dedupe exposures by latest assignment

leoromanovsky:
## Motivation

DatadogFlags exposure deduplication stored the full exposure tuple as already seen. That suppressed a valid exposure when a subject and flag cycled back to a previous assignment, for example `A -> B -> A`, because the final `A` had been seen before.

The cache also needs to protect mobile devices while supporting the expected client high-water mark of two subjects across 2,500 flags.

Related Android PR: DataDog/dd-sdk-android#3526 applies the same mobile exposure-cache contract.

## Changes

- Add a regression test proving `A -> B -> A` writes three exposure events.
- Add focused exposure tracker tests covering repeated assignments, assignment cycles, and 2 subjects x 2,500 flags.
- Change the exposure tracker cache to use `targetingKey + flagKey` as the key and `allocationKey + variationKey` as the latest cached value.
- Keep duplicate suppression only for repeated evaluations of the current assignment.
- Use platform `NSCache` with a 5,000-entry `countLimit` for the in-memory exposure assignment cache.

## Decisions

- Kept the change internal with no public API changes.
- Used a count limit, not a byte limit, to avoid approximate object-size accounting while preserving a reviewer-friendly bound.
- Sized the count limit around the expected two-subject, 2,500-flag high-water mark.
- Used iOS's platform cache primitive instead of a custom deterministic LRU implementation; eviction order is intentionally not part of the correctness contract.
- Matched the Android PR's count-based mobile policy while using iOS's platform cache primitive.
- Matched `dd-trace-go` and `openfeature-js-client` semantics, where the dedupe key and assignment value are modeled separately so assignment changes re-emit exposures.

Validation:

- `xcodebuild -workspace Datadog.xcworkspace -scheme DatadogFlags -destination "platform=iOS Simulator,id=32AE07AE-F88E-4FF4-A384-F6A347AA3B98" -only-testing:DatadogFlagsTests/ExposureTrackerTests -only-testing:DatadogFlagsTests/ExposureLoggerTests/testLogExposureAssignmentCycle test`
- `SKIP_LINT=1 xcodebuild -workspace Datadog.xcworkspace -scheme DatadogFlags -destination "platform=iOS Simulator,id=32AE07AE-F88E-4FF4-A384-F6A347AA3B98" test`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - [FEATURE] Add wildcard host pattern support to WebView tracking via `WebViewTracking.enable(webView:hostPatterns:)`. See [#2963][]
+- [FIX] Fix `DatadogFlags` exposure deduplication so assignment changes for the same subject and flag emit new exposure events. See [#2987][]
 
 # 3.12.0 / 04-06-2026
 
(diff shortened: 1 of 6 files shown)


barboraplasovska (commented):
Looks great! But you should add a CHANGELOG entry.

barboraplasovska (approved):
LGTM

PR #2989, opened 2026-06-11
FFL-2510: Add allocationKey to FlagDetails

typotter:
### What and why?

FFL-2510 — \`FlagDetails\` lacked a top-level \`allocationKey\` field, making the allocation key from the Precomputed Flags response inaccessible to callers using \`getDetails\`.

### How?

- \`FlagDetails.swift\`: Added \`public var allocationKey: String?\` (default \`nil\`). \`nil\` means no allocation was associated with the evaluation — never empty string.
- \`FlagsClient.getDetails\`: populates \`allocationKey\` from the typed assignment field on success; error paths return \`nil\`
- API surface regenerated

### Review checklist
- [ ] Run \`make api-surface\` when adding new APIs

### DIFF
--- a/DatadogFlags/Sources/Client/FlagsClient.swift
+++ b/DatadogFlags/Sources/Client/FlagsClient.swift
@@ -252,7 +252,8 @@ extension FlagsClient: FlagsClientProtocol {
             key: key,
             value: value,
             variant: flagAssignment.variationKey,
-            reason: flagAssignment.reason
+            reason: flagAssignment.reason,
+            allocationKey: flagAssignment.allocationKey.isEmpty ? nil : flagAssignment.allocationKey
         )
 
         trackEvaluation(key: key, assignment: flagAssignment, context: context)
--- a/DatadogFlags/Sources/Models/FlagDetails.swift
+++ b/DatadogFlags/Sources/Models/FlagDetails.swift
@@ -101,5 +101,4 @@ public struct FlagDetails<T>: Equatable where T: Equatable {
         self.error = error
         self.allocationKey = allocationKey
     }
-
 }
--- a/DatadogFlags/Sources/Models/FlagDetails.swift
+++ b/DatadogFlags/Sources/Models/FlagDetails.swift
@@ -102,23 +102,4 @@ public struct FlagDetails<T>: Equatable where T: Equatable {
         self.allocationKey = allocationKey
     }
 
-    /// Creates detailed flag evaluation information.
-    ///
-    /// This overload preserves binary compatibility with prior SDK versions that did not include `allocationKey`.
-    ///
-    /// - Parameters:
-    ///   - key: The feature flag key.
-    ///   - value: The evaluated or default value.
-    ///   - variant: The variant key served, if any.
-    ///   - reason: The evaluation reason, if available.
-    ///   - error: Any error that occurred during evaluation.
-    public init(
-        key: String,
-        value: T,
-        variant: String? = nil,
-        reason: String? = nil,
-        error: FlagEvaluationError? = nil
-    ) {
-        self.init(key: key, value: value, variant: variant, reason: reason, error: error, allocationKey: nil)
-    }
 }
--- a/DatadogFlags/Sources/Models/FlagDetails.swift
+++ b/DatadogFlags/Sources/Models/FlagDetails.swift
@@ -101,4 +101,24 @@ public struct FlagDetails<T>: Equatable where T: Equatable {
         self.error = error
         self.allocationKey = allocationKey
     }
+
+    /// Creates detailed flag evaluation information.
+    ///
+    /// This overload preserves binary compatibility with prior SDK versions that did not include `allocationKey`.
+    ///
+    /// - Parameters:
+    ///   - key: The feature flag key.
+    ///   - value: The evaluated or default value.
+    ///   - variant: The variant key served, if any.
+    ///   - reason: The evaluation reason, if available.
+    ///   - error: Any error that occurred during evaluation.
+    public init(
+        key: String,
+        value: T,
+        variant: String? = nil,
+        reason: String? = nil,
+        error: FlagEvaluationError? = nil
+    ) {
+        self.init(key: key, value: value, variant: variant, reason: reason, error: error, allocationKey: nil)
+    }
 }
--- a/DatadogFlags/Sources/Models/FlagDetails.swift
+++ b/DatadogFlags/Sources/Models/FlagDetails.swift
@@ -71,6 +71,12 @@ public struct FlagDetails<T>: Equatable where T: Equatable {
     /// value is from a successful evaluation or a fallback to the default value.
     public var error: FlagEvaluationError?
 
+    /// The allocation key associated with this flag evaluation.
+    ///
+    /// Identifies the allocation bucket used when evaluating the flag. `nil` if the flag
+    /// was not found, evaluation failed, or no allocation was associated with this evaluation.
+    public var allocationKey: String?
+
     /// Creates detailed flag evaluation information.
     ///
     /// - Parameters:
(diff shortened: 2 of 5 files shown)


typotter:
**Required by:** DataDog/dd-openfeature-provider-swift#23 — the Swift OpenFeature provider depends on `FlagDetails.metadata` introduced here to populate `ProviderEvaluation.flagMetadata`. That PR currently pins `dd-sdk-ios` to this branch; once this PR merges and a release is cut, the provider's `Package.swift` should be updated to the new version tag.

DatadogFlags/Sources/Models/FlagDetails.swift
```diff
@@ -79,17 +84,29 @@ public struct FlagDetails<T>: Equatable where T: Equatable {
...
     ///   - error: Any error that occurred during evaluation.
+    ///   - metadata: Additional metadata about the evaluation, such as `allocationKey`.
     public init(
         key: String,
         value: T,
         variant: String? = nil,
         reason: String? = nil,
-        error: FlagEvaluationError? = nil
+        error: FlagEvaluationError? = nil,
+        metadata: [String: Any] = [:]
```
sameerank: You'll need to run `make api-surface` to update the surface file with the new public `init` signature

Also I believe we prefer to use `AnyValue` instead of `Any`, if possible. It has a lot of properties that work better in Swift: it's Equatable, Sendable, Codable, Hashable, etc. and distinguishes .int/.bool/.double

https://github.com/DataDog/dd-sdk-ios/blob/f61b3833ab9526bb1c3a5b906feb8fb95978bcaf/DatadogFlags/Sources/Models/AnyValue.swift#L31

DatadogFlags/Sources/Client/FlagsClient.swift
```diff
@@ -248,11 +248,27 @@ extension FlagsClient: FlagsClientProtocol {
...
+        var metadata: [String: Any] = [:]
+        // Write extraLogging primitives first; skip "allocationKey" since it is always
+        // sourced from the typed field below so it cannot clobber the actual allocation key.
+        for (logKey, value) in flagAssignment.extraLogging where logKey != "allocationKey" {
+            switch value {
+            case .string(let s): metadata[logKey] = s
+            case .int(let i): metadata[logKey] = i
+            case .double(let d): metadata[logKey] = d
+            case .bool(let b): metadata[logKey] = b
+            default: break // skip complex types (arrays, dictionaries, null)
```
sameerank: We might be able to avoid dropping these if we use `AnyValue`

DatadogFlags/Sources/Models/FlagAssignment.swift
```diff
@@ -39,12 +40,20 @@ public struct FlagAssignment: Equatable {
...
     }
 
-    public init(allocationKey: String, variationKey: String, variation: Variation, reason: String, doLog: Bool) {
+    public init(
+        allocationKey: String,
+        variationKey: String,
+        variation: Variation,
+        reason: String,
+        doLog: Bool,
+        extraLogging: [String: AnyValue] = [:]
```
sameerank: Might need to run `make api-surface` to update the surface file

typotter:
FFL-2510

DatadogFlags/Sources/Models/FlagAssignment.swift
typotter: done

DatadogFlags/Sources/Client/FlagsClient.swift
typotter: done

DatadogFlags/Sources/Models/FlagDetails.swift
typotter: done

api-surface-swift
```diff
@@ -1530,7 +1530,8 @@ public struct FlagAssignment: Equatable
     public var variation: Variation
     public var reason: String
     public var doLog: Bool
-    public init(allocationKey: String, variationKey: String, variation: Variation, reason: String, doLog: Bool)
+    public var extraLogging: [String: AnyValue]
+    public init(allocationKey: String,variationKey: String,variation: Variation,reason: String,doLog: Bool,extraLogging: [String: AnyValue] = [:])
```
sameerank: There used to be spaces between the function arguments and now there aren't. I wonder if there is some configuration mismatch in the `make api-surface` run that created the pervious version vs yours?

DatadogFlags/Sources/Client/FlagsClient.swift
```diff
@@ -248,11 +248,21 @@ extension FlagsClient: FlagsClientProtocol {
             return FlagDetails(key: key, value: defaultValue, error: .typeMismatch)
         }
 
+        var metadata: [String: AnyValue] = [:]
+        // Write extraLogging values first; skip "allocationKey" since it is always
+        // sourced from the typed field below so it cannot clobber the actual allocation key.
```
sameerank: If `flagAssignment` always has a `.allocationKey` then is there even a concern about clobbering?

You either don't need the `logKey != "allocationKey"` check because it'll get overwritten later or you could remove this comment

>  // allocationKey is always written last so it wins over any extraLogging value

Because you're already explicitly ensured that  "allocationKey" is skipped from `extraLogging`

sameerank (dismissed):
I think you may also need to add an entry to the "Unreleased" section of CHANGELOG.md

api-surface-swift
sameerank: There's still an inconsistency in the formatting here by removing the spaces between the arguments

typotter: From what I can tell, the spacing difference is caused by the extraLogging parameter having a default value (= [:]). The tool omits spaces between parameters when any parameter has a default; FlagDetails.init already uses the same format (variant: String? = nil etc). This is the output of running `make api-surface`

typotter:
Thanks @sameerank . Ready for another look

DatadogFlags/Sources/Client/FlagsClient.swift
```diff
@@ -248,11 +248,15 @@ extension FlagsClient: FlagsClientProtocol {
             return FlagDetails(key: key, value: defaultValue, error: .typeMismatch)
         }
 
+        var metadata: [String: AnyValue] = [:]
+        metadata["allocationKey"] = .string(flagAssignment.allocationKey)
```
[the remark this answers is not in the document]
typotter: the `extraLogging` field is deprecated

DatadogFlags/Sources/Models/FlagDetails.swift
```diff
@@ -79,17 +85,20 @@ public struct FlagDetails<T>: Equatable where T: Equatable {
...
     ///   - error: Any error that occurred during evaluation.
+    ///   - allocationKey: The allocation key for the evaluation, or `nil` if not available.
     public init(
         key: String,
         value: T,
         variant: String? = nil,
         reason: String? = nil,
-        error: FlagEvaluationError? = nil
+        error: FlagEvaluationError? = nil,
+        allocationKey: String? = nil
```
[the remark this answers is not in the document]
typotter: Added the 5-arg overload in 72c158eb that delegates to the 6-arg init with `allocationKey: nil`. Binary symbol preserved. Also added a CHANGELOG entry under Unreleased.

gonzalezreal (dismissed):

DatadogFlags/Sources/Models/FlagDetails.swift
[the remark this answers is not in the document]
simaoseica-dd: We don't preserve ABI compatibility for unrecompiled downstream binaries in this case.

Some v3 examples:
```
3.8.0 RUM.Configuration.URLSessionTracking.init(...)
Added trackResourceHeaders: TrackResourceHeaders = .disabled
3.10.0 (SRShapeWireframe|SRTextWireframe|SRImageWireframe|...).init(...)
Added permanentId: String? = nil
3.11.0 Trace.Configuration.URLSessionTracking.init(...)
Added redactedStatusCodes: Set<Int> = defaultRedactedStatusCodes
```

If we want to keep the old init, please deprecate it.

simaoseica-dd (dismissed):
Left a small comment. Other than that, the PR description does not reflect the logic in review

DatadogFlags/Sources/Models/FlagDetails.swift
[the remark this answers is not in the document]
typotter: Removed the overload in 5133d929d. Only the 6-arg init with `allocationKey: String? = nil` remains, consistent with the SDK's convention.

simaoseica-dd (approved):

PR #2992, opened 2026-06-12
[RUM-16711] Add foreground refresh to RemoteConfigurationProvider

saraSr5:
### What and why?

Adds foreground refresh to RemoteConfigurationProvider so remote configuration stays fresh while the app is in use, not only at SDK initialization.

When the app returns to the foreground, the provider re-syncs with the CDN. If the configuration has changed, the new document is fetched, decoded, persisted to disk, and the in-memory cache is updated. If nothing has changed, the server returns 304 and no work is done thanks to the ETag conditional request mechanism.

### How?

- RemoteConfigurationProvider observes ApplicationNotifications.willEnterForeground via NotificationCenter, registered in init and removed in deinit.
- The observer is only registered on platforms that support UIKit (#if canImport(UIKit)).
- notificationCenter is injected in the RemoteConfigurationProvider init and passed from Datagdog.swift via configuration.notificationCenter.
- sync() is now private — the provider is fully autonomous: it syncs on init and on every foreground transition without any external orchestration.
- Telemetry errors are reported internally via an injected Telemetry instance (NOPTelemetry() by default for tests).

### Review checklist

Note: TTL-based suppression (skip the CDN request if last sync was recent) is intentionally out of scope and will be tracked in a separate ticket.

### DIFF
--- a/DatadogCore/Sources/Core/DatadogCore.swift
+++ b/DatadogCore/Sources/Core/DatadogCore.swift
@@ -56,8 +56,7 @@ internal final class DatadogCore {
     let bus = MessageBus()
 
     /// The remote configuration provider, if configured.
-    @ReadWriteLock
-    var remoteConfigurationProvider: RemoteConfigurationProvider?
+    let remoteConfigurationProvider: RemoteConfigurationProvider?
 
     /// The last successfully fetched remote configuration, if any.
     @ReadWriteLock
--- a/DatadogCore/Sources/Core/DatadogCore.swift
+++ b/DatadogCore/Sources/Core/DatadogCore.swift
@@ -56,7 +56,8 @@ internal final class DatadogCore {
     let bus = MessageBus()
 
     /// The remote configuration provider, if configured.
-    var remoteConfigurationProvider: RemoteConfigurationProvider?
+    @ReadWriteLock
+    var remoteConfigurationProvider: RemoteConfigurationProvider? = nil
 
     /// The last successfully fetched remote configuration, if any.
     var remoteConfiguration: RemoteConfiguration? {
@@ -99,9 +100,9 @@ internal final class DatadogCore {
         directory: CoreDirectory,
         dateProvider: DateProvider,
         initialConsent: TrackingConsent,
-    	performance: PerformancePreset,
-    	httpClient: HTTPClient,
-    	encryption: DataEncryption?,
+        performance: PerformancePreset,
+        httpClient: HTTPClient,
+        encryption: DataEncryption?,
         contextProvider: DatadogContextProvider,
         applicationVersion: String,
         maxBatchesPerUpload: Int,
@@ -327,6 +328,7 @@ internal final class DatadogCore {
     /// Stops all processes for this instance of the Datadog core by
     /// deallocating all Features and their storage & upload units.
     func stop() {
+        remoteConfigurationProvider?.stop()
         stores = [:]
         features = [:]
     }
--- a/DatadogCore/Sources/Core/DatadogCore.swift
+++ b/DatadogCore/Sources/Core/DatadogCore.swift
@@ -57,12 +57,11 @@ internal final class DatadogCore {
 
     /// The remote configuration provider, if configured.
     @ReadWriteLock
-    var remoteConfigurationProvider: RemoteConfigurationProvider? = nil
+    var remoteConfigurationProvider: RemoteConfigurationProvider?
 
     /// The last successfully fetched remote configuration, if any.
-    var remoteConfiguration: RemoteConfiguration? {
-        remoteConfigurationProvider?.remoteConfiguration
-    }
+    @ReadWriteLock
+    var remoteConfiguration: RemoteConfiguration?
 
     /// Registry for Features.
     @ReadWriteLock
@@ -133,6 +132,15 @@ internal final class DatadogCore {
         // forward any context change on the message-bus
         self.contextProvider.publish { [weak self] context in
             self?.send(message: .context(context))
+        }
+
+        self.remoteConfigurationProvider?.start { [weak self] result in
+            switch result {
+            case .success(let remoteConfiguration):
+                self?.remoteConfiguration = remoteConfiguration
+            case .failure(let error):
+                self?.telemetry.error("[RemoteConfig] Sync failed", error: error)
+            }
         }
     }
 
--- a/DatadogCore/Sources/Core/RemoteConfigurationProvider.swift
+++ b/DatadogCore/Sources/Core/RemoteConfigurationProvider.swift
@@ -7,22 +7,22 @@
 import Foundation
 import DatadogInternal
 
-/// Provides the last successfully fetched remote configuration and refreshes its cache.
-///
-/// TODO: Also trigger `sync()` on every app foreground transition so remote config
-/// is refreshed while the app is in use, not only at SDK init (RFC §Caching Strategy).
+/// Provides the last successfully fetched remote configuration.
+/// Refreshes when started and when the app enters foreground.
 internal final class RemoteConfigurationProvider {
     let id: String
     let site: DatadogSite
     let directory: Directory
     let httpClient: HTTPClient
-    private let telemetry: Telemetry
+    private let notificationCenter: NotificationCenter
+    @ReadWriteLock
+    private var foregroundObserver: NSObjectProtocol?
 
     /// The result of the last cache read or CDN fetch.
     /// `.success(remoteConfiguration)` — remote configuration is available.
     /// `.failure` — no cache yet, or a read/write error.
     @ReadWriteLock
-    private(set) var cache: Result<RemoteConfiguration, RemoteConfigurationError>
+    private(set) var cache: Result<RemoteConfiguration, RemoteConfigurationError> = .failure(.diskError)
 
     var remoteConfiguration: RemoteConfiguration? {
         try? cache.get()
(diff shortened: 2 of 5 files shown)


saraSr5:
@codex review

DatadogCore/Sources/Core/RemoteConfigurationProvider.swift
```diff
@@ -67,38 +113,45 @@ internal final class RemoteConfigurationSynchronizer {
...
+                    return
+                }
 
                 // All checks passed — persist to disk and update in-memory cache.
                 // File.write uses .atomic (write to temp, then rename), so the update is
                 // all-or-nothing: the existing file is never left in a truncated state.
                 do {
                     try File(url: self.directory.url.appendingPathComponent("\(self.id).json")).write(data: data)
-                    self.cache = .success(data)
+                    self.cache = .success(remoteConfiguration)
```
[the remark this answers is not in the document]
saraSr5: TTL-based rate-limiting will address this. Tracking it in a separate ticket.

saraSr5:
@codex review

DatadogCore/Sources/Core/RemoteConfigurationProvider.swift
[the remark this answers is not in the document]
saraSr5: TTL-based rate-limiting will address this. Tracking it in a separate ticket.

DatadogCore/Sources/Datadog.swift
```diff
@@ -461,14 +462,18 @@ extension DatadogCore {
...
-            remoteConfiguration: configuration.remoteConfigurationID.map { ($0, configuration.site) }
+            isRunFromExtension: isRunFromExtension
         )
 
-        synchronizer?.sync { [weak self] result in
-            if case .failure(let error) = result {
-                self?.telemetry.error("[RemoteConfig] Sync failed", error: error)
-            }
+        if let remoteConfigurationID = configuration.remoteConfigurationID {
+            self.remoteConfigurationProvider = RemoteConfigurationProvider(
```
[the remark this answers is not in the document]
saraSr5: Good point

saraSr5:
@codex review

DatadogCore/Sources/Core/DatadogCore.swift
```diff
@@ -329,6 +328,7 @@ internal final class DatadogCore {
     /// Stops all processes for this instance of the Datadog core by
     /// deallocating all Features and their storage & upload units.
     func stop() {
+        remoteConfigurationProvider = nil
```
maxep: no need to nillify the provider here.

DatadogCore/Sources/Core/DatadogCore.swift
```diff
@@ -55,9 +55,14 @@ internal final class DatadogCore {
     /// The message-bus instance.
     let bus = MessageBus()
 
-    /// Owns the remote configuration cache and fetch lifecycle.
-    /// `nil` when `remoteConfiguration` was not set at init.
-    internal let synchronizer: RemoteConfigurationSynchronizer?
+    /// The remote configuration provider, if configured.
+    @ReadWriteLock
+    var remoteConfigurationProvider: RemoteConfigurationProvider? = nil
```
maxep: No need for a default here, the default is in the init.
```suggestion
    var remoteConfigurationProvider: RemoteConfigurationProvider?
```

saraSr5:
@codex review

saraSr5:
@codex review

DatadogCore/Sources/Core/RemoteConfigurationProvider.swift
```diff
@@ -67,38 +124,48 @@ internal final class RemoteConfigurationSynchronizer {
             request.setValue(etag, forHTTPHeaderField: "If-None-Match")
         }
 
-        httpClient.fetch(request: request) { result in
+        httpClient.send(request: request, delegate: nil) { [weak self] result in
```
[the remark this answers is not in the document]
saraSr5: TTL-based rate-limiting (RUM-16753) will significantly reduce the window for this race, once a sync completes and sets lastSyncDate, subsequent foreground notifications within 5 minutes are suppressed, so concurrent requests can only occur between the init sync and the very first foreground transition. This edge case is intentionally deferred to RUM-16753.

DatadogCore/Sources/Core/RemoteConfigurationProvider.swift
```diff
@@ -67,38 +124,48 @@ internal final class RemoteConfigurationSynchronizer {
...
-                guard !data.isEmpty else {
-                    completionHandler(.failure(RemoteConfigurationError.emptyBody))
+                guard let data = data, !data.isEmpty else {
+                    self.reportSyncFailure(.emptyBody)
                     return
                 }
 
-                // TODO RUM-16387: Validate the schema before saving
+                let remoteConfiguration: RemoteConfiguration
+                switch Self.decode(data) {
```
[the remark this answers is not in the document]
saraSr5: This is a valid forward-compatibility concern not currently covered. The generated closed enums will fail on unknown values from newer server schemas.

DatadogCore/Sources/Core/DatadogCore.swift
```diff
@@ -327,6 +328,7 @@ internal final class DatadogCore {
     /// Stops all processes for this instance of the Datadog core by
     /// deallocating all Features and their storage & upload units.
     func stop() {
+        remoteConfigurationProvider?.stop()
```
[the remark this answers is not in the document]
saraSr5: This is intentional. stop() removes the foreground observer so no new syncs are triggered after the SDK is stopped. The provider is kept alive so remoteConfiguration remains readable until the core is deallocated. Any in-flight request uses [weak self] so if the provider is released it will be a no-op.

DatadogCore/Sources/Core/RemoteConfigurationProvider.swift
```diff
@@ -33,38 +33,68 @@ internal final class RemoteConfigurationProvider {
...
+        foregroundObserver = notificationCenter.addObserver(
+            forName: ApplicationNotifications.willEnterForeground,
+            object: nil,
+            queue: nil
+        ) { [weak self] _ in
+            self?.sync { _ in }
+        }
+        #endif
+
+        sync(cached == nil ? completionHandler : { _ in })
```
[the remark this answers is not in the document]
saraSr5: If a refresh fails, the cached config remains valid and the SDK continues operating normally. Reporting every failure would generate noise for expected conditions like poor network coverage. If the CDN is consistently unavailable, the initial sync failure (which is reported) would already surface the issue

DatadogCore/Sources/Core/RemoteConfigurationProvider.swift
```diff
@@ -7,22 +7,22 @@
...
+    private let notificationCenter: NotificationCenter
+    @ReadWriteLock
+    private var foregroundObserver: NSObjectProtocol?
 
     /// The result of the last cache read or CDN fetch.
     /// `.success(remoteConfiguration)` — remote configuration is available.
     /// `.failure` — no cache yet, or a read/write error.
     @ReadWriteLock
-    private(set) var cache: Result<RemoteConfiguration, RemoteConfigurationError>
+    private(set) var cache: Result<RemoteConfiguration, RemoteConfigurationError> = .failure(.diskError)
```
maxep: Lets remove the cache here and let the Core store the configuration or report the error. For that, the `start` method should be called from the `Core.init`, like we do [here](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogCore/Sources/Core/DatadogCore.swift#L122-L125)

DatadogCore/Sources/Core/RemoteConfigurationProvider.swift
```diff
@@ -33,38 +33,68 @@ internal final class RemoteConfigurationProvider {
...
         // Acceptable because the file is small (a single JSON document) and only
         // present after a previous successful fetch — absent on first launch.
-        self._cache = ReadWriteLock(wrappedValue: Self.readCache(id: id, from: directory, telemetry: telemetry))
+        let cached = Self.readCache(id: id, from: directory)
+        if let cached {
+            cache = cached
+            completionHandler(cached)
+        }
+
+        #if canImport(UIKit)
```
maxep: All `#if canImport(UIKit)` can be removed, none of this code depends on UIKit

DatadogCore/Sources/Core/RemoteConfigurationProvider.swift
```diff
@@ -33,38 +33,68 @@ internal final class RemoteConfigurationProvider {
...
+        foregroundObserver = notificationCenter.addObserver(
+            forName: ApplicationNotifications.willEnterForeground,
+            object: nil,
+            queue: nil
+        ) { [weak self] _ in
+            self?.sync { _ in }
```
maxep: The `completionHandler` should be called here, to propagate the new configuration

DatadogCore/Sources/Core/RemoteConfigurationProvider.swift
```diff
@@ -33,38 +33,68 @@ internal final class RemoteConfigurationProvider {
...
+        var observer: NSObjectProtocol?
+        _foregroundObserver.mutate {
+            observer = $0
+            $0 = nil
+        }
+
+        if let observer {
+            notificationCenter.removeObserver(observer)
+        }
+        #endif
```
maxep: You can simply do:

```suggestion
        _foregroundObserver.mutate { observer in
            observer.map { notificationCenter.removeObserver($0) }
            observer = nil
        }
```

DatadogCore/Tests/Datadog/DatadogCore/DatadogCoreTests.swift
```diff
@@ -385,6 +412,53 @@ class DatadogCoreTests: XCTestCase {
         XCTAssertEqual(requestBuilderSpy.requestParameters.count, 0, "It should not send any request")
     }
 
+#if canImport(UIKit)
```
maxep:
```suggestion
```

maxep (commented):
Looks good 👍  I have left few feedbacks

DatadogCore/Sources/Core/RemoteConfigurationProvider.swift
```diff
@@ -78,35 +87,37 @@ internal final class RemoteConfigurationProvider {
...
-        // a 304 response would leave us with no data to serve.
         var request = URLRequest(url: site.remoteConfigurationEndpoint
             .appendingPathComponent("v1")
             .appendingPathComponent(id)
             .appendingPathExtension("json"))
 
-        if case .success = cache,
+        if directory.hasFile(named: "\(id).json"),
            let data = try? directory.file(named: "\(id).etag").read(),
            let etag = String(data: data, encoding: .utf8) {
```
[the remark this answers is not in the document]
saraSr5: I intentionally used hasFile instead of a full decode to avoid synchronous JSON decoding on main thread during foreground transitions. The accepted tradeoff is that a corrupted .json with a stale .etag would result in a 304 leaving us without config until the next refresh. In practice this is very unlikely since the SDK uses atomic writes for persistence.

DatadogCore/Sources/Core/DatadogCore.swift
```diff
@@ -56,12 +56,12 @@ internal final class DatadogCore {
     let bus = MessageBus()
 
     /// The remote configuration provider, if configured.
+    @ReadWriteLock
     var remoteConfigurationProvider: RemoteConfigurationProvider?
```
maxep: This value is never mutated, no need for a lock nor a var :) 
```suggestion
    let remoteConfigurationProvider: RemoteConfigurationProvider?
```

DatadogCore/Sources/Core/RemoteConfigurationProvider.swift
```diff
@@ -7,64 +7,77 @@
...
+        self.notificationCenter = notificationCenter
+    }
+
+    func start(_ completionHandler: @escaping (Result<RemoteConfiguration, RemoteConfigurationError>) -> Void) {
         // Synchronous read on the caller's thread (main thread during SDK init).
         // Acceptable because the file is small (a single JSON document) and only
         // present after a previous successful fetch — absent on first launch.
-        self._cache = ReadWriteLock(wrappedValue: Self.readCache(id: id, from: directory, telemetry: telemetry))
+        let cached = Self.readCache(id: id, from: directory)
+        if let cached {
```
maxep: nit
```suggestion
        if let cached = Self.readCache(id: id, from: directory){
```

DatadogCore/Sources/Core/RemoteConfigurationProvider.swift
```diff
@@ -78,35 +91,37 @@ internal final class RemoteConfigurationProvider {
...
             case .success(let (http, data)):
-                // 1. Not Modified — existing cache is still valid
+                // 1. Not Modified — existing persisted configuration is still valid.
                 if http.statusCode == 304 {
-                    completionHandler(self.cache)
+                    if let cached = Self.readCache(id: self.id, from: self.directory), case .success = cached {
+                        completionHandler(cached)
+                    } else {
+                        completionHandler(.failure(.httpError(http.statusCode)))
+                    }
```
maxep: We should simply `return` when status is 304 (not changed). No need to re-read the cache and call `completionHandler`, the value was already shared after launch or after a previous foreground fetch.

maxep (approved):
Looks good 👍  I have few minor feedback

DatadogCore/Sources/Core/RemoteConfigurationProvider.swift
```diff
@@ -78,35 +90,32 @@ internal final class RemoteConfigurationProvider {
...
             switch result {
             case .failure(let error):
                 completionHandler(.failure(.networkError(error)))
 
             case .success(let (http, data)):
-                // 1. Not Modified — existing cache is still valid
+                // 1. Not Modified — existing persisted configuration is still valid.
                 if http.statusCode == 304 {
-                    completionHandler(self.cache)
                     return
```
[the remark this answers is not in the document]
saraSr5: This is intentional on 304, the configuration was already shared with the caller after the initial start() or a previous foreground fetch. There's no new value to deliver

maxep (approved):
👍

Issue #2994, opened 2026-06-12
Crash: "Swift/UnsafeRawPointer.swift:449: Fatal error: load from misaligned raw pointer" on Simulator iOS 27 beta 1


davidpisoni-payjoy:
### Stack trace

```
* thread #14, queue = 'com.datadoghq.ios-sdk-read-write', stop reason = Fatal error: load from misaligned raw pointer
     frame #0: 0x0000000180110f90 libswiftCore.dylib`_swift_runtime_on_report
     frame #1: 0x0000000180182ea0 libswiftCore.dylib`_swift_stdlib_reportFatalErrorInFile + 200
     frame #2: 0x00000001801aa850 libswiftCore.dylib`closure #1 in closure #1 in _assertionFailure(_:_:file:line:flags:) + 348
     frame #3: 0x00000001801aa3cc libswiftCore.dylib`_assertionFailure(_:_:file:line:flags:) + 188
     frame #4: 0x00000001801aa6c8 libswiftCore.dylib`_fatalErrorMessage(_:_:file:line:flags:) + 40
     frame #5: 0x00000001802a2fbc libswiftCore.dylib`UnsafeRawBufferPointer.load(fromByteOffset:as:) + 224
   * frame #6: 0x0000000109123674 PayJoy.debug.dylib`closure #1 in TLVBlockReader.readType($0=2 values (0x16c478963)) at TLVBlockReader.swift:131:42
  ﹉ frame #11: 0x0000000109122d48 PayJoy.debug.dylib`TLVBlockReader.readType() at TLVBlockReader.swift:131:21
     frame #12: 0x0000000109120bd4 PayJoy.debug.dylib`TLVBlockReader.readBlock() at TLVBlockReader.swift:118:24
     frame #13: 0x0000000109120654 PayJoy.debug.dylib`TLVBlockReader.next() at TLVBlockReader.swift:50:28
     frame #14: 0x0000000109121124 PayJoy.debug.dylib`TLVBlockReader.all() at TLVBlockReader.swift:69:31
     frame #15: 0x00000001091156ac PayJoy.debug.dylib`FileReader.decode(stream=0x000000010d022080) at FileReader.swift:67:27
[...]

PR #2995, opened 2026-06-15
RUM-16755: Fix loading of misaligned data.

arroz:
### What and why?

Fixes crash reported in #2994. Data loaded in the corrected functions is not guaranteed to be aligned as the loaded type expects, since they may be local (stack) copies of single bytes (Inline Data representations) the compiler may store on any address. Although this has not been a problem so far, Xcode 27 compiler aligns data differently and causes crashes.

### How?

Since we cannot assume data is aligned, `load` calls were replaced by `loadUnaligned`.

### Review checklist


simaoseica-dd (approved):

maxep (approved):
🙏

PR #2997, opened 2026-06-15
[RUM-16774] Encode and upload client-side stats via MessagePack

tienquocbui:
### What and why?

Wires the client-side stats pipeline end to end. Building on [RUM-16605](https://github.com/DataDog/dd-sdk-ios/pull/2971), this replaces the `StatsRequestBuilder`
stub - which emitted raw JSON - with the real upload path: the `ExportedBucket`s
persisted by `ClientStatsFeature` are decoded, mapped onto the wire-format
`StatsPayload`, MessagePack-encoded, and uploaded deflate-compressed to the
`/api/v0.2/stats` intake.

### How?

- **Content type**: add `URLRequestBuilder.HTTPHeader.ContentType.applicationMsgPack`
  (`application/msgpack`).
- **Storage round-trip**: make `ExportedBucket` / `ExportedGroupedStats` `Codable`
  and `Trilean` `Decodable` so stored events decode back out of feature storage.
- **Mapping**: add `ExportedBucket -> ClientStatsBucket` and
  `ExportedGroupedStats -> ClientGroupedStats` initializers (kept in the trace
  module, no SDK-context coupling).
- **Envelope**: `StatsRequestBuilder` now assembles `ClientStatsPayload` with
  `hostname` empty, `env` / `version` / `service` / `tracerVersion` from
  `DatadogContext`, a per-process `runtimeID` (UUID), and a monotonic
  `sequenceNumber` from a thread-safe `StatsSequenceNumberProvider`. The builder
  throws on an empty batch or undecodable events so the core drops them rather
  than uploading a partial payload.
- **Wiring**: `ClientStatsFeature` owns the `runtimeID` and sequence provider,
  injecting both into the request builder.
- **E2E**: enable `statsComputationEnabled` in the dogfood trace scenario so the
  Synthetics run validates real intake acceptance of the deflate + MessagePack
  envelope.

The `application/msgpack` case is on `URLRequestBuilder` (public but not part of
the tracked api-surface), so the public API surface is unchanged.

### Review checklist
- [ ] Add CHANGELOG entry for user facing changes (n/a - internal feature, not yet public)
- [ ] Add Objective-C interface for public APIs (n/a - no new public Swift API)
- [x] Run \`make api-surface\` when adding new APIs (n/a - api-surface unchanged)

[RUM-16605]: https://datadoghq.atlassian.net/browse/RUM-16605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/DatadogTrace/Sources/Feature/StatsRequestBuilder.swift
+++ b/DatadogTrace/Sources/Feature/StatsRequestBuilder.swift
@@ -7,21 +7,76 @@
 import Foundation
 import DatadogInternal
 
+/// Provides monotonically increasing sequence numbers for uploaded stats payloads.
+///
+/// The sequence number lets the intake order and uniquely identify the payloads sent by a
+/// single tracer (`runtimeID`). It is incremented once per built request; retries of the same
+/// batch therefore get distinct sequence numbers, which is acceptable because the value is used
+/// for identification, not for stats aggregation.
+internal final class StatsSequenceNumberProvider {
+    @ReadWriteLock private var value: UInt64 = 0
+
+    /// Returns the next sequence number, starting at `1` for the first call.
+    func next() -> UInt64 {
+        var next: UInt64 = 0
+        _value.mutate {
+            $0 += 1
+            next = $0
+        }
+        return next
+    }
+}
+
 /// Builds HTTP requests for uploading client-side stats to the `/api/v0.2/stats` intake.
+///
+/// Decodes the `ExportedBucket`s persisted by `ClientStatsFeature`, maps them onto the
+/// wire-format `StatsPayload`, encodes that to MessagePack, and uploads it deflate-compressed.
 internal struct StatsRequestBuilder: FeatureRequestBuilder {
+    /// Custom stats intake URL, overriding the site default when set.
     let customIntakeURL: URL?
+    /// Sends telemetry through the SDK core.
     let telemetry: Telemetry
+    /// Identifies the tracer instance uniquely across the payloads it sends.
+    let runtimeID: String
+    /// Supplies the per-payload sequence number.
+    let sequenceNumberProvider: StatsSequenceNumberProvider
+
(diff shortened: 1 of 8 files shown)


DatadogTrace/Sources/Feature/StatsRequestBuilder.swift
```diff
@@ -7,21 +7,77 @@
...
+        }
+
+        // If a stored bucket cannot be decoded there is no way to recover it, so we throw
+        // and let the core drop the batch rather than uploading a partial payload.
+        let buckets = try events
+            .map { try decoder.decode(ExportedBucket.self, from: $0.data) }
+            .map(ClientStatsBucket.init)
+
+        let clientStats = ClientStatsPayload(
+            hostname: "",
```
arroz, line 68: Should we obtain the real hostname from the device?

DatadogTrace/Tests/ClientStatsFeatureTests.swift
```diff
@@ -138,6 +138,251 @@ class ClientStatsFeatureTests: XCTestCase {
...
+        XCTAssertThrowsError(
+            try builder.request(
+                for: [],
+                with: .mockAny(),
+                execution: .init(previousResponseCode: nil, attempt: 0)
+            )
+        )
+    }
+
+    func testSequenceNumberProviderIncrementsMonotonically() {
```
arroz, line 284: It would be nice to have a test that launches a bunch of threads, runs this 1000 times or so on each thread, and verifies if the next number after all the threads finish is the expected one, to guarantee we are supporting concurrent access properly.

arroz (commented):
Congrats, this feature is almost done! 😃 Left two minor comments.

arroz (approved):
All good!

PR #3001, opened 2026-06-17
[PANA-7720] Fix text detection in transparent images

gonzalezreal:
### What and why?

This PR fixes text redaction for layer snapshot images with transparent backgrounds.

Vision can miss dark text when it is rendered over transparency. In that case, text detection can fail and the text may not be redacted in the layer tree recording pipeline.

### How?

For non-opaque images, the redaction path now composites the image over a solid background before running Vision text detection.

The background color is chosen to contrast with the image dominant color, so dark text is placed over a light background and light text is placed over a dark background.

### Review checklist

### DIFF
--- a/DatadogSessionReplay/Sources/Recorder/Utilities/UIImage+FeatureDetection.swift
+++ b/DatadogSessionReplay/Sources/Recorder/Utilities/UIImage+FeatureDetection.swift
@@ -97,7 +97,7 @@ extension UIColor {
 
         self.getRed(&r, green: &g, blue: &b, alpha: nil)
 
-        // W3C Luminance Formula
+        // Approximate perceived brightness to choose a contrasting background
         return (0.299 * r) + (0.587 * g) + (0.114 * b)
     }
 }
--- a/DatadogSessionReplay/Sources/Recorder/Utilities/UIImage+FeatureDetection.swift
+++ b/DatadogSessionReplay/Sources/Recorder/Utilities/UIImage+FeatureDetection.swift
@@ -10,10 +10,24 @@ import Vision
 
 extension UIImage {
     func textRectangles() throws -> [CGRect] {
+        let targetImage = if imageRendererFormat.opaque {
+            self
+        } else {
+            // Composite transparent images over a contrasting background so Vision can detect text
+            UIGraphicsImageRenderer(size: size, format: imageRendererFormat).image { context in
+                let foregroundColor = dominantColor ?? .black
+                let backgroundColor = foregroundColor.luminance < 0.5 ? UIColor.white : .black
+
+                backgroundColor.setFill()
+                UIRectFill(CGRect(origin: .zero, size: size))
+                draw(at: .zero)
+            }
+        }
+
         let request = VNDetectTextRectanglesRequest()
         request.reportCharacterBoxes = false
 
-        try performFeatureDetection(with: request)
+        try targetImage.performFeatureDetection(with: request)
 
         guard let results = request.results else {
             return []
@@ -72,6 +86,19 @@ extension CGImagePropertyOrientation {
         default:
             self = .up
         }
+    }
+}
+
+extension UIColor {
+    fileprivate var luminance: CGFloat {
+        var r: CGFloat = 0
+        var g: CGFloat = 0
+        var b: CGFloat = 0
+
+        self.getRed(&r, green: &g, blue: &b, alpha: nil)
+
+        // W3C Luminance Formula
+        return (0.299 * r) + (0.587 * g) + (0.114 * b)
     }
 }
 #endif
--- a/DatadogSessionReplay/Tests/Utilities/UIImage+SessionReplayTests.swift
+++ b/DatadogSessionReplay/Tests/Utilities/UIImage+SessionReplayTests.swift
@@ -6,6 +6,7 @@
 
 #if os(iOS)
 import Foundation
+import UIKit
 import XCTest
 
 @testable import DatadogSessionReplay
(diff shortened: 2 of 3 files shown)


DatadogSessionReplay/Sources/Recorder/Utilities/UIImage+FeatureDetection.swift
```diff
@@ -10,10 +10,24 @@ import Vision
...
+            self
+        } else {
+            // Composite transparent images over a contrasting background so Vision can detect text
+            UIGraphicsImageRenderer(size: size, format: imageRendererFormat).image { context in
+                let foregroundColor = dominantColor ?? .black
+                let backgroundColor = foregroundColor.luminance < 0.5 ? UIColor.white : .black
+
+                backgroundColor.setFill()
+                UIRectFill(CGRect(origin: .zero, size: size))
+                draw(at: .zero)
```
maxep, line 23: This is run once per UIImage instance, right?

DatadogSessionReplay/Sources/Recorder/Utilities/UIImage+FeatureDetection.swift
```diff
@@ -74,4 +88,17 @@ extension CGImagePropertyOrientation {
...
+extension UIColor {
+    fileprivate var luminance: CGFloat {
+        var r: CGFloat = 0
+        var g: CGFloat = 0
+        var b: CGFloat = 0
+
+        self.getRed(&r, green: &g, blue: &b, alpha: nil)
+
+        // W3C Luminance Formula
+        return (0.299 * r) + (0.587 * g) + (0.114 * b)
```
sethfowler-datadog, line 101: This is a perceptual value (i.e., these coefficients are derived from the human visual system). I'm honestly not sure whether this is the right choice for the "visual system" of the text detection model. An alternative that may or may not be better would be to call `getHue()` and use the returned brightness value.

sethfowler-datadog (dismissed):

DatadogSessionReplay/Sources/Recorder/Utilities/UIImage+FeatureDetection.swift
```diff
@@ -10,10 +10,24 @@ import Vision
 
 extension UIImage {
     func textRectangles() throws -> [CGRect] {
+        let targetImage = if imageRendererFormat.opaque {
+            self
+        } else {
+            // Composite transparent images over a contrasting background so Vision can detect text
+            UIGraphicsImageRenderer(size: size, format: imageRendererFormat).image { context in
+                let foregroundColor = dominantColor ?? .black
+                let backgroundColor = foregroundColor.luminance < 0.5 ? UIColor.white : .black
```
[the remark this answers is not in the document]
sethfowler-datadog: I was a bit worried about this too, though I'm not sure of the details of how `dominantColor` is computed -- do transparent black and transparent white produce different results?

DatadogSessionReplay/Sources/Recorder/Utilities/UIImage+FeatureDetection.swift
gonzalezreal: Yes. Only once, because redacted images are cached in an associated property. Notice that this won't be run on all image snapshots taken from `CALayer`s, only those that meet the criteria based on the layer class and the layer delegate class. See https://github.com/DataDog/dd-sdk-ios/blob/f5464fe3c6ebec40a80bce262d64a8381c800261/DatadogSessionReplay/Sources/LayerTreeRecorder/ImageSnapshot%2BRedaction.swift

DatadogSessionReplay/Sources/Recorder/Utilities/UIImage+FeatureDetection.swift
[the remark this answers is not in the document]
gonzalezreal: I am double-checking this. It may have a point. I was under the impression that the `CIKMeans` filter didn't take transparent pixels into account, since it always produce opaque colors.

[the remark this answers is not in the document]
gonzalezreal: I checked this against the `CIKMeans` docs and repros.

The palette alpha is the color weight, and the RGB we render is premultiplied by that weight. So a small white foreground over transparency comes back as `[3, 3, 3, 3]`; `dominantColor` reads that as near-black, but after `settingAlphaOne(in:)` it becomes `[255, 255, 255, 255]`.

I’ll update `dominantColor` to set alpha to 1 before reading the palette color. That gives us the actual dominant color, so the luminance check can choose the right contrasting background.

See: https://developer.apple.com/documentation/coreimage/cifilter-swift.class/kmeans()

DatadogSessionReplay/Sources/Recorder/Utilities/UIImage+FeatureDetection.swift
gonzalezreal: I think luminance is still the better fit here. `brightness` treats saturated colors like red and blue as bright, so it can choose black even when white would give better contrast.

I’ll keep luminance, but I can make the comment less tied to W3C and more explicit about the goal: choosing a contrasting background for Vision.

DatadogSessionReplay/Tests/Utilities/UIImage+SessionReplayTests.swift
```diff
@@ -24,6 +25,17 @@ class UIImageSessionReplayTests: XCTestCase {
         let scaledData = try XCTUnwrap(image.dd.pngData(maxSize: CGSize(width: 25, height: 25)))
         XCTAssertLessThan(scaledData.count, imageData.count)
     }
+
+    func testDominantColor_ReturnsOpaqueColorForLowWeightPaletteColor() throws {
```
sethfowler-datadog, line 29: Just as a sanity check, I'd suggest splitting this into two variants:
1. Background is transparent black and the smaller opaque rect is white. (I believe this is what you have now.)
2. Background is transparent white and the smaller opaque rect is black.

Past experiences implementing scaling algorithms have made me wary of the transparent white edge case. 🙂

sethfowler-datadog (approved):

maxep (approved):
👍

DatadogSessionReplay/Tests/Utilities/UIImage+SessionReplayTests.swift
gonzalezreal: Good idea. I will do that in a follow up PR.

